PR #6381, opened 2017-10-03
centralize plugin options

RusinovAnton:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues             | Fixes #6376 
| Patch: Bug Fix?          | No (?)
| Major: Breaking Change?  | No
| Minor: New Feature?      | No
| Tests Added/Pass?        | No
| Spec Compliancy?         | 
| License                  | MIT
| Doc PR                   |
| Any Dependency Changes?  | No

Refactor plugins code to centralize options usage as requested in #6376

PR #6776, opened 2017-11-09
Hoist more plugin options and default useUnicodeFlag to 'true'.

loganfsmyth:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

A few places that are hard to grep for and got missed in https://github.com/babel/babel/pull/6381.

One other small change is making `unicode-property-regex` default to `useUnicodeFlag: true` instead of `: false`, which makes it leave the `/u` flag on the regex by default, so the transform will focus on transforming property regexes, rather than also transforming other unicode-related stuff.

I was tempted to just remove the flag, but opted to leave it for now.

### DIFF
--- a/packages/babel-plugin-proposal-unicode-property-regex/README.md
+++ b/packages/babel-plugin-proposal-unicode-property-regex/README.md
@@ -43,10 +43,17 @@ To transpile to ES6/ES2015:
 ```js
 require("@babel/core").transform(code, {
   "plugins": [
-    ["@babel/proposal-unicode-property-regex", { "useUnicodeFlag": true }]
+    ["@babel/proposal-unicode-property-regex", { "useUnicodeFlag": false }]
   ]
 });
 ```
+
+## Options
+
+* `useUnicodeFlag` (defaults to `true`)
+
+When disabled with `false`, the transform will convert unicode regexes to
+non-unicode regexes, removing the `u` flag. See https://www.npmjs.com/package/regexpu-core#useunicodeflag-default-false- for more information.
 
 ## Author
 
--- a/packages/babel-plugin-proposal-unicode-property-regex/src/index.js
+++ b/packages/babel-plugin-proposal-unicode-property-regex/src/index.js
@@ -1,18 +1,22 @@
 import rewritePattern from "regexpu-core";
 import * as regex from "@babel/helper-regex";
 
-export default function() {
+export default function(api, options) {
+  const { useUnicodeFlag = false } = options;
+  if (typeof useUnicodeFlag !== "boolean") {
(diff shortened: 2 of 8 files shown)


packages/babel-plugin-proposal-unicode-property-regex/src/index.js
```diff
@@ -1,18 +1,22 @@
 import rewritePattern from "regexpu-core";
 import * as regex from "@babel/helper-regex";
 
-export default function() {
+export default function(api, options) {
+  const { useUnicodeFlag = true } = options;
```
loganfsmyth: @mathiasbynens I know we pulled in this plugin, so I'm curious for your feedback. We usually try to keep our plugins focused on single tasks, and because this was defaulting to `false`, it was converting unicode to non-unicode regexes at the same time even though this plugin is centered around the property regex functionality.

Given that people can always enable `transform-unicode-regex` alongside this proposal, do you think it is worth having this flag to begin with? Wondering if we should just remove it entirely.

hzoo: This is reflected in the readme as well, I remember talking about this before (since yeah we do already have a plugin that does the es6 transform and ideally all of it will be in preset-env anyway

loganfsmyth:
I updated the README. I'm just going to land this since it seems like the right thing to do. We can always tweak more later.

mathiasbynens:
> Given that people can always enable `transform-unicode-regex` alongside this proposal, do you think it is worth having this flag to begin with? Wondering if we should just remove it entirely.
 
That results in two regexpu passes instead of one, though.

loganfsmyth:
It's more consistent with how our plugins work though. Doing two passes is what I'd expect to happen, since this plugin is specifically about unicode property handling.

hzoo:
Either that or we merge this plugin into the other and have 2 options and use that to modify preset-env

packages/babel-plugin-proposal-unicode-property-regex/README.md
```diff
@@ -43,11 +43,18 @@ To transpile to ES6/ES2015:
...
   ]
 });
 ```
 
+## Options
+
+* `useUnicodeFlag` (defaults to `true`)
+
+When disabled with `false`, the transform will convert unicode regexes to
+non-unicode regexes, removing the `u` flag. See https://www.npmjs.com/package/regexpu-core#useunicodeflag-default-false- for more information.
```
mathiasbynens: Please link to the GitHub repo rather than the npm copy

PR #6777, opened 2017-11-09
Add an official 'state.filename' and be more explicit about option passing.

loganfsmyth:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Y, but only if a plugin relies on `state.file.opts.filename` defaulting to "unknown", and the caller of Babel doesn't pass one in, which most of the main ones do.
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Having to do "state.file.opts.filename" is both ugly and undiscoverable. This adds an official API on the `state` object that plugins can use the access the current filename.

Ideally in the future, especially if we can nail down a good concept of the "root" directory for config loading that we've been talking about, we can expose that too, and we'll have a nice pair.

Otherwise this PR is mostly trying to avoid just passing Babel's core options object off to other modules, which makes it _super_ hard to do a search to figure out where options are used, and also just makes things harder to track.

### DIFF
--- a/packages/babel-core/src/transformation/normalize-opts.js
+++ b/packages/babel-core/src/transformation/normalize-opts.js
@@ -7,7 +7,7 @@ export default function normalizeOptions(config: ResolvedConfig): {} {
     parserOpts: Object.assign(
       {
         sourceType: config.options.sourceType,
-        sourceFileName: config.options.filename,
+        sourceFileName: config.options.filename || "unknown",
         plugins: [],
       },
       config.options.parserOpts,
--- a/packages/babel-core/src/transformation/normalize-opts.js
+++ b/packages/babel-core/src/transformation/normalize-opts.js
@@ -3,19 +3,41 @@
 import type { ResolvedConfig } from "../config";
 
 export default function normalizeOptions(config: ResolvedConfig): {} {
-  const options = Object.assign({}, config.options, {
+  const opts = config.options;
+
+  const options = Object.assign({}, opts, {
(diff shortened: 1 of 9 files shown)


packages/babel-core/src/transformation/normalize-opts.js
```diff
@@ -3,19 +3,41 @@
...
-        sourceType: config.options.sourceType,
-        sourceFileName: config.options.filename,
+        sourceType: opts.sourceType,
+        sourceFileName: opts.filename || "unknown",
         plugins: [],
       },
-      config.options.parserOpts,
+      opts.parserOpts,
+    ),
+    generatorOpts: Object.assign(
```
hzoo: I think we can use object spread in these places if you free it's more readable

loganfsmyth: I think I'll leave this since it was already using `Object.assign`, but fair point. Will keep that in mind for next time.

loganfsmyth:
> nice change! so for the plugins we just want a codemod for state.file.opts.filename -> state.filename?

Yeah, seems like that'd be good to do.

> Does it remove state.file.opts.filename?

Nope, still there so old plugins should work. The one change to `state.file.opts.filename` that could qualify as breaking, but would depend, is that if the thing calling Babel doesn't pass a filename, it'll be `undefined` now, instead of defaulting to `"unknown"`.

PR #6780, opened 2017-11-09
Standardize on Sync suffix for sync APIs (with backward-compat)

loganfsmyth:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  | N, unless someone was accidentally passing a callback to this synchronous API
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

In 6.x we have
```
transform // sync
transformFromAst // sync
transformFile // async
transformFileSync // sync
```
which is pretty confusing, so this PR standardizes it to
```
transform // async, but sync if no callback given, for back compat
transformSync // sync
transformFromAst // async, but sync if no callback given, for back compat
transformFromAstSync // sync
transformFile // async
transformFileSync // sync
```

This PR doesn't _use_ the async ability for anything yet, but it gives us a base to build upon. I'm figuring I'll file a beginner-friendly bug for converting babel-cli to use the async version. We can also update babel-loader and babelify and such, so that when we _do_ start adding things that only work for async usecases, the existing integrations will work. We'll also probably want to run through our tests and update them to use `transformSync`.

As an example of usage here, I'm tossing around the idea of letting caching plugins be async if they want to me, so they'd just log a warning or silently skip cache processing if used with `babel.transformSync` and such. We could also let plugins have async init logic, which could throw if used in a sync context call.

### DIFF
--- a/packages/babel-core/src/transform-ast.js
+++ b/packages/babel-core/src/transform-ast.js
@@ -1,21 +1,56 @@
 // @flow
-import * as t from "@babel/types";
-import loadConfig from "./config";
-import runTransform, { type FileResult } from "./transformation";
 
-export default function transformFromAst(
-  ast: Object,
-  code: string,
-  opts: Object,
-): FileResult | null {
-  const config = loadConfig(opts);
-  if (config === null) return null;
+import loadConfig, { type InputOptions } from "./config";
+import {
+  runAsync,
+  type FileResult,
+  type FileResultCallback,
+} from "./transformation";
 
-  if (ast && ast.type === "Program") {
-    ast = t.file(ast, [], []);
-  } else if (!ast || ast.type !== "File") {
-    throw new Error("Not a valid ast?");
+import transformAstSync from "./transform-ast-sync";
+
+type AstRoot = BabelNodeFile | BabelNodeProgram;
+
+type TransformAst = {
+  (ast: AstRoot, code: string, callback: FileResultCallback): void,
+  (
+    ast: AstRoot,
+    code: string,
+    opts: ?InputOptions,
+    callback: FileResultCallback,
+  ): void,
+
+  // Here for backward-compatibility. Ideally use ".transformSync" if you want
+  // a synchronous API.
+  (ast: AstRoot, code: string, opts: ?InputOptions): FileResult | null,
+};
+
+export default ((function transformFromAst(ast, code, opts, callback) {
+  if (typeof opts === "function") {
+    opts = undefined;
+    callback = opts;
   }
 
-  return runTransform(config, code, ast);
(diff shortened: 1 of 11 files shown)


packages/babel-core/src/transform-ast.js
```diff
@@ -1,21 +1,56 @@
...
+    opts: ?InputOptions,
+    callback: FileResultCallback,
+  ): void,
+
+  // Here for backward-compatibility. Ideally use ".transformSync" if you want
+  // a synchronous API.
+  (ast: AstRoot, code: string, opts: ?InputOptions): FileResult | null,
+};
+
+export default ((function transformFromAst(ast, code, opts, callback) {
```
xtuc: Do we need to emulate an async function? What about not providing this at all? Only `transformFromAstSync` or `transformFromAst` which is sync?

Otherwise lgtm.

nicolo-ribaudo: I think we need to emulate the async function so that we can in the future introduce async plugins without breaking changes.

packages/babel-core/src/transform-ast.js
```diff
@@ -1,21 +1,56 @@
...
+  // For backward-compat with Babel 6, we allow sync transformation when
+  // no callback is given. Will be dropped in some future Babel major version.
+  if (callback === undefined) return transformAstSync(ast, code, opts);
+
+  // Reassign to keep Flowtype happy.
+  const cb = callback;
+
+  // Just delaying the transform one tick for now to simulate async behavior
+  // but more async logic may land here eventually.
+  process.nextTick(() => {
```
nicolo-ribaudo: Can `process.nextTick(() => {` be moved inside `runAsync`? So that `runAsync` is actually async and we avoid faking the async behavior two times.

packages/babel-core/src/transform-ast.js
xtuc: Good point :+1:

loganfsmyth: Yeah exactly. My goal with this PR is just to expose the API so tooling can start using it, even if the plugins don't support async yet. This way we at least have more of a migration path.

packages/babel-core/src/transform-ast.js
loganfsmyth: I can do that too. I left it this way because I wanted the `cb(err)` for the `loadConfig` call to also be async, and it felt unnecessary to delay things two ticks when one was enough for this case.

nicolo-ribaudo: Oh right... It's ok as is then.

Issue #2, opened 2014-10-01
missing templates when installed via npm


es128:
`npm install sebmck/6to5`

```
λ node
> to5 = require('6to5')
Error: ENOENT, no such file or directory '/Code/6to5test/node_modules/6to5/lib/6to5/templates'
    at Object.fs.readdirSync (fs.js:654:18)
    at Object.<anonymous> (/Code/6to5test/node_modules/6to5/lib/6to5/util.js:201:13)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/Code/6to5test/node_modules/6to5/lib/6to5/transform.js:4:16)
    at Module._compile (module.js:456:26)
```

https://github.com/sebmck/6to5/blob/master/.npmignore#L5

Seems as if your intention is to only include them for dev/test, but `util` needs to be fixed to not puke when they're absent.

PR #6781, opened 2017-11-09
Make official API for plugins to override parser/generator

loganfsmyth:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

We've had this, _kind of_, with an API from https://github.com/babel/babel/pull/6781 where plugins could like

```
manipulateOptions(opts, parserOpts) {
  parserOpts.parser = require("some parser");
}
```
which was super confusing, poorly documented, and because we allowed it as programmatic arguments as
```
babel.transform("", {parserOpts: { parser: "recast" }});
```
Babel had to know how to resolve a given module string.

With this PR, you can make a Recast plugin like
```
import { parse, print } from "recast";
 
export default function pluginRecast(){
  return {
    parserOverride(code, opts, parser) {
      return parse(code, {
        ...opts,
        parser: {
          parse: source => parser(source, opts),
        },
      });
    },
    generatorOverride(ast, opts) {
       return print(ast, opts);
    },
  };
};
```
and just put the plugin in your `plugins` list like anything else used with Babel.

### DIFF
--- a/packages/babel-core/src/config/plugin.js
+++ b/packages/babel-core/src/config/plugin.js
@@ -8,6 +8,9 @@ import {
   type Validator,
 } from "./option-assertions";
 
+// Note: The casts here are just meant to be static assertions to make sure
+// that the assertion functions actually assert that the value's type matches
+// the declared types.
 const VALIDATORS: ValidatorSet = {
   name: (assertString: Validator<$PropertyType<PluginObject, "name">>),
   manipulateOptions: (assertFunction: Validator<
@@ -20,6 +23,13 @@ const VALIDATORS: ValidatorSet = {
   >),
   visitor: (assertVisitorMap: Validator<
     $PropertyType<PluginObject, "visitor">,
+  >),
+
+  parserOverride: (assertFunction: Validator<
+    $PropertyType<PluginObject, "parserOverride">,
+  >),
+  generatorOverride: (assertFunction: Validator<
+    $PropertyType<PluginObject, "generatorOverride">,
   >),
 };
 
@@ -70,6 +80,9 @@ export type PluginObject = {
 
   inherits?: Function,
   visitor?: VisitorMap,
+
+  parserOverride?: Function,
+  generatorOverride?: Function,
 };
 
 export function validatePluginObject(obj: {}): PluginObject {
@@ -90,6 +103,9 @@ export default class Plugin {
   pre: Function | void;
   visitor: {};
 
+  parserOverride: Function | void;
+  generatorOverride: Function | void;
+
   options: {};
 
   constructor(plugin: PluginObject, options: {}, key?: string) {
@@ -99,6 +115,9 @@ export default class Plugin {
     this.post = plugin.post;
     this.pre = plugin.pre;
     this.visitor = plugin.visitor || {};
+    this.parserOverride = plugin.parserOverride;
+    this.generatorOverride = plugin.generatorOverride;
+
     this.options = options;
   }
 }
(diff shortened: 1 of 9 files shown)


packages/babel-core/src/transformation/file/generate.js
```diff
@@ -1,25 +1,36 @@
...
-  if (opts.generatorOpts && opts.generatorOpts.generator) {
-    gen = opts.generatorOpts.generator;
+  let result;
+  plugins_check: for (const plugins of pluginPasses) {
+    for (const plugin of plugins) {
+      const { generatorOverride } = plugin;
+      if (generatorOverride) {
+        result = generatorOverride(ast, opts.generatorOpts, code, generate);
+
+        if (result != null) break plugins_check;
```
nicolo-ribaudo: If two plugins provide two generators, the second one will be ignored?

packages/babel-core/src/config/plugin.js
```diff
@@ -1,43 +1,123 @@
...
+  constructor(plugin: PluginObject, options: {}, key?: string) {
     this.key = plugin.name || key;
 
     this.manipulateOptions = plugin.manipulateOptions;
     this.post = plugin.post;
     this.pre = plugin.pre;
-    this.visitor = plugin.visitor;
+    this.visitor = plugin.visitor || {};
+    this.parserOverride = plugin.parserOverride;
+    this.generatorOverride = plugin.generatorOverride;
```
hzoo: We really need to start documenting this stuff now lol

packages/babel-core/src/transformation/file/generate.js
loganfsmyth: Yeah, I was thinking it gave some plugins the ability to toggle behavior more easily? Like

packages/babel-core/src/config/plugin.js
loganfsmyth: I don't really mind that it's not super well publicized, I just want a clear API for it :P

packages/babel-core/src/transformation/file/generate.js
loganfsmyth: Guess it ends up being essentially the same as https://github.com/webpack/tapable/blob/v0.2.8/lib/Tapable.js#L124 except my version allows `null`. Maybe it'd be better to only bail on specifically `undefined` though to be consistent with that.

nicolo-ribaudo: I'd prefer it to throw if two different parsers/generators don't return null, instead of using the first valid result.

nicolo-ribaudo: > Maybe it'd be better to only bail on specifically undefined though to be consistent with that.

How is null handled in other places in Babel options?

loganfsmyth: > I'd prefer it to throw if two different parsers/generators don't return null, instead of using the first valid result.

loganfsmyth: > How is null handled in other places in Babel options?

packages/babel-core/src/config/plugin.js
loganfsmyth: test

loganfsmyth: sdfds
```

packages/babel-core/src/transformation/file/generate.js
loganfsmyth: Done. Changed to `undefined` and now throws an error of two different plugins return a result.

PR #6782, opened 2017-11-09
Minor improvements to block-scoping/tdz

maurobringolf:
| Fixed Issues?            |
| Documentation PR         | 
| Any Dependency Changes?  |

I looked into TDZ and discovered a couple of shortcomings. I added documentation for `tdz` option in `babel-plugin-transform-block-scoping` and test cases for the currently uncovered tdz part. Then the current implementation seems to ignore function hoisting as it inserts an error here:

```javascript
f();

function f() {}
```

AFAIK a function declaration cannot throw a tdz error, so I inserted a line that skips this case.

And here is something that I have not found in older issues but should be fixable I think: When using destructuring, the variable names to be assigned are visited by `ReferencedIdentifier` which inserts the error there as well. It seems wrong to me that they are `ReferencedIdentifier`'s.

```javascript
let { X: Y } = { X: 3 }

assert.equal(Y, 3);
```

I understand that tdz does not have high priority, but I had already looked into it and decided to submit this anyway.


loganfsmyth:
Thanks! You're right that it isn't a priority, but fixes are still appreciated.

maurobringolf:
Thanks! Cool, then I might have another look at it.

PR #6651, opened 2017-10-31
Lazy-install sourceMapSupport

aminmarashi:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #5890 
| Patch: Bug Fix?          | :+1:
| Major: Breaking Change?  | 
| Minor: New Feature?      |
| Tests Added + Pass?      | :+1:
| Documentation PR         | 
| Any Dependency Changes?  | 

This is an attempt to fix issue #5890 according to @loganfsmyth [suggestion](https://github.com/babel/babel/issues/5890#issuecomment-337800472).

### Todo:

- [x] Check if this actually fixes the performance problem
- [x] Make sure the fix is not a breaking change
- [x] Make sure the source maps are created correctly when requested
- [x] Add tests for modules with/without source maps

PR #6783, opened 2017-11-09
Apply option defaults when transforming, not up front.

loganfsmyth:
| Patch: Bug Fix?          |
| Major: Breaking Change?  | N, theoretically if someone was loading the config and relying on the defaults being populated, but doesn't seem likely.
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

The up-front application of defaults is going to cause problems with cases like https://github.com/babel/babel/pull/6651/files#diff-75a0292ed78043766c2d5564edd84ad2R72, so this basically makes the config returned from loadConfig _explicitly_ only the options that the user has provided + options that guarantee that if the same config is passed back to Babel, that it will parse the same way the next time, which makes it easier for people to load the config, make a few small tweaks, and call Babel again.

### DIFF
--- a/packages/babel-core/src/config/option-manager.js
+++ b/packages/babel-core/src/config/option-manager.js
@@ -120,6 +120,7 @@ class OptionManager {
 
     // Tack the passes onto the object itself so that, if this object is passed back to Babel a second time,
     // it will be in the right structure to not change behavior.
+    opts.babelrc = false;
     opts.plugins = this.passes[0];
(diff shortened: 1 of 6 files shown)


danez:
I seems to be the reason why babel-loader does not pick up .babelrc files anymore. So when using babe-core.transform() `babelrc` has to be set to `true` now to pick up config files?

https://github.com/babel/babel-loader/issues/552

PR #5608, opened 2017-04-07
Cache configs based on mtime and allow .babelrc.js functions

loganfsmyth:
| Q                        | A 
| ------------------------ | ---
| Patch: Bug Fix?          | N
| Major: Breaking Change?  | N
| Minor: New Feature?      | Y
| Deprecations?            | 
| Spec Compliancy?         | 
| Tests Added/Pass?        | 
| Fixed Tickets            | 
| License                  | MIT
| Doc PR                   | 
| Dependency Changes       | 

Curious for feedback. Including the `.babelrc.js` function logic because it's partially a demonstration of _why_ I'm caching based on mtime. Caching based on the file content doesn't really work with the function-based config, so we might as well just leave it to the configs themselves to invalidate themselves consistently.

The key thing for this PR is that `findConfigs` and `loadConfig` will now return the same object consistently, so

    loadConfig('./.babelrc.js', process.cwd()) === loadConfig('./.babelrc.js', process.cwd())

which means we can build the caching for the plugins and presets on top of that. So if the config files change mtime, the object they return will change, and that will trigger all of the presets and plugins to change.

PR #6786, opened 2017-11-09
Add infrastructure for string-ish GUIDs for use with caching and invalidation

loganfsmyth:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Documentation PR         | 
| Any Dependency Changes?  |

This PR adds cachable keys to each plugin, but doesn't actually use them for anything. I feel like this is fine to land as-is, but I also figured it would be a good piece for discussion.

So the short example of what this does is:
screen shot 2017-11-09 at 2 52 10 pm

for every core module, helper, and plugin.

EDIT: See https://github.com/babel/babel/pull/6786#issuecomment-343553214 for more explanation.


codemods/babel-plugin-codemod-optional-catch-binding/package.json
```diff
@@ -8,6 +8,9 @@
   "keywords": [
     "@babel/plugin"
   ],
+  "scripts": {
```
Andarist: as this is a published package, i don't think it is a good practice to escape directory boundary, can this be refactored to use a module or something?

Andarist:
Could you summarize how exactly do you plan to implement caching later on top of it? Or maybe there is already some overview you could point to?

packages/babel-helper-caching/README.md
```diff
@@ -0,0 +1,76 @@
...
+const cacheKey = helperCaching(name, version);
+```
+
+or you could of course avoid the runtime parsing of `package.json` by
+scripting your package publishing logic to update the version automatically.
+
+### Complex plugins
+
+If your plugin has dependencies on either Babel's core helpers, or other
+packages that export utility logic, you'll wa
```
hzoo: typo here?

codemods/babel-plugin-codemod-optional-catch-binding/package.json
loganfsmyth: Hmm, tough one. I agree that escaping boundaries is kind of unfortunate, but making a whole npm package just to hold this script also seems pretty questionable. I was mostly not worrying about it because it's not something people should ever call directly anyway. Did you have something specific in mind?

loganfsmyth:
> Could you summarize how exactly do you plan to implement caching later on top of it? Or maybe there is already some overview you could point to?

Totally. My intention with this is essentially to take all of the values that are in the configs, hash them, and then pass that hash to any plugin that wants to handle caching. I should explicitly say that the following cases where `cacheKey` are required would only fail _if_ you've enabled a plugin for caching. That way existing Babel 6 plugins can carry over, and Babel will just complain if you used them with the new caching system. Alternatively we could _not_ fail, and just show a warning, which might be easier for the short term, then make it a hard error in Babel 8.

From what I can see, the hashable values are as follows:

### Configs

* Walk the config hashing all properties except the plugins and presets arrays.
* If a property containing an unhandle-able type is encountered, require the user to provide a `cacheKey` in the root of their config.
* If the config is a `.js` file, also generate a hash from any `api.cache()` results, introduced in https://github.com/babel/babel/pull/5608/files#diff-2f4602ba055eb6ec98d6bac3161d43ffR68

### Plugin/Preset instantiation from plugin/preset arrays

* Plugin/preset instances inputs:
  * Options passed in from parent -> hash with the rules of the config/preset that passed them
  * Globals accessed with `api.cache()` -> Just require these be hashable values
* Outputs simple objects, optionally with a `cacheKey` to represent the result.
  * presets can just be handled as configs
  * Plugins will pretty much be required to have a `cacheKey`    

So for each instance, we take those hashable values and combined them.

### Preset output

* Pretty much treated the same as config. The only extra thing I've considered here is that we could use the same extra logic to skip hashing requirements if the unhashable objects were passed through from a parent config or preset. e.g.
    ```js
    export default function preset(api, opts){
      return {
        plugins: [['some-plugin', opts]]
      };
    }
    ```
    does not need a cache key because it just passes options through, and the parent config would have already been required to have a cache key.

### Plugin output

* `cackeKey` values for each plugin representing the version of the logic inside the plugin itself.


So the existing caching logic we have in `babel-loader` might look something like
```
import CACHE_KEY from "./_cache-key";

export default function pluginFsCache(api, options, dirname) {
  const { directory, gzip = true } = options;
  if (directory !== undefined && typeof directory !== "string") {
    throw new Error(`.directory must be a string, or undefined`);
  }
  if (typeof gzip !== "boolean") {
    throw new Error(`.gzip must be a boolean, or undefined`);
  }

  let cacheDir;
  if (directory === ":tmp:") cacheDir = os.tmpdir();
  if (directory === undefined) {
    cacheDir = findCacheDir({
      name: "babel-plugin-fs-cache",
      cwd: dirname,
      create: true,
    });

    if (!cacheDir) {
      throw new Error(
        `Unable to find cache directory relative to "${dirname}"`,
      );
    }
  } else {
    cacheDir = path.resolve(dirname, directory);
  }

  const getFilepath = name => path.resolve(cacheDir, name + ".js.gz");

  return {
    cacheKey: CACHE_KEY,
    loadFromCache(key: string, filename: string | void, callback?: Function) {
      const filepath = getFilepath(key);
      return zlib.gunzipSync(fs.readFileSync(filepath)).toString();

    },
    saveToCache(
      key: string,
      filename: string | void,
      data: string,
      callback?: Function,
    ) {
      const filepath = getFilepath(key);
      fs.writeFileSync(filepath, zlib.gzipSync(data));
    },
  };
}
```

packages/babel-helper-caching/README.md
```diff
@@ -0,0 +1,76 @@
+# @babel/helper-caching
+
+## Usage
+
+The objective of this helper is to create an object that Babel can use as
```
existentialism: Nit: 

"... Babel can use as a unique identifier for the current running logic of a plugin"

or maybe,

"... Babel can use to uniquely identify the current running logic of a plugin"

packages/babel-helper-caching/README.md
```diff
@@ -0,0 +1,76 @@
...
+a unique identifier the current running logic of a plugin. The result object can
+be stringified with @@toPrimitive. The cache key contract is intentionally kept
+simple in order to allow users to pass cache keys that they maintain without
+having to pull in this whole module if they want to manage their cache keys
+using their own logic somehow.
+
+Using of this helper in Babel's own plugins is script-generated in order to
```
existentialism: Using -> Usage

packages/babel-helper-caching/README.md
```diff
@@ -0,0 +1,76 @@
...
+fairly simple.
+
+The "cacheKey" object generated by this API has a simple API
+
+* `.inspect()` - JSON output to make debugging the key easier.
+* `.toString()` - Generate a SHA representing the key's content.
+
+with the goal of example usage being `${cackeKey}` to convert the key into
+a string when it is ready for use. The string is generated lazily in order
+to avoid slowdowns for startup time, and to in general avoid unnecessary work.
```
existentialism: Maybe:

> The string is generated lazily in order to avoid doing any unnecessary work that may slow down startup time.

packages/babel-helper-caching/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Simple plugins
+
+For a simple transfor plugin with no transformation-related dependencies:
```
existentialism: transfor -> transform?

packages/babel-helper-caching/src/index.js
```diff
@@ -0,0 +1,137 @@
...
+};
+
+function wrap(cb: () => string) {
+  let generating = false;
+  let sha;
+
+  return () => {
+    if (sha === undefined) {
+      if (generating === true) {
+        throw new Error("Cycle detected in cache key graph.");
```
nicolo-ribaudo: Question: when could this happen?

loganfsmyth: Yeah, you could make a plugin that depended on another one that depended back on the original. It could also happen in core itself if we had any dependency cycles between our core modules. If it does happen it's definitely a bug, but I'd rather throw a clear error than get an stack overflow.

scripts/build-cache-key.js
```diff
@@ -0,0 +1,89 @@
...
+    if (name === "babylon" || /^@babel\//.test(name)) {
+      acc.push(`require(${JSON.stringify(name)}).CACHE_KEY`);
+    }
+
+    return acc;
+  }, []);
+
+  let content;
+
+  if (name === HELPER_MODULE) {
```
nicolo-ribaudo: Shouldn't this be `if (name === HELPER_MODULE || name === "babylon") {`? (Based on the comment below)

scripts/build-cache-key.js
```diff
@@ -0,0 +1,89 @@
...
+      });
+    `;
+  } else {
+    return null;
+  }
+
+  return {
+    name: swapPackageWithKey(pkgPath),
+    content:
+      content.replace(/(^|\r?\n) {6}/g, "$1").replace(/^\s+|\s+$/g, "") + "\n",
```
nicolo-ribaudo: I don't like putting something only related to the code formatting in the code itself (` {6}`). Maybe we could use something like https://github.com/declandewet/common-tags#stripindent.

scripts/build-cache-key.js
loganfsmyth: Ah woops, I need to edit the comment, good catch. I had Babylon in here and then removed it because I realized this doesn't work with Babylon because it uses Rollup.

scripts/build-cache-key.js
loganfsmyth: Can do! I was just lazy :P

codemods/babel-plugin-codemod-optional-catch-binding/package.json
loganfsmyth: @Andarist I don't know what to say on this. Thoughts?

Andarist: @loganfsmyth Sorry, I've missed the mention here. Imho if it's shareable piece of code between packages it could be put into separate package (even `private` lerna-managed one). That way it should be also easier to discover it and possibly reuse in other packages

Gulpfile.js
```diff
@@ -5,7 +5,6 @@ const through = require("through2");
 const chalk = require("chalk");
 const newer = require("gulp-newer");
 const babel = require("gulp-babel");
-const watch = require("gulp-watch");
 const gutil = require("gulp-util");
```
xtuc: I can make a new PR with only the change that you've made to the Gulpfile, would you mind? It seems to me that theses changes are not really related.

packages/babel-helper-caching/README.md
xtuc: Probably, `you will have to`

loganfsmyth:
Still working through the exact right architecture for this, so closing for now.

PR #6787, opened 2017-11-09
Don't run regexpu twice

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

After https://github.com/babel/babel/pull/6776, the default behavior of `@babel/plugin-proposal-unicode-property-regex` changed and now it keeps the `u` flags (as it should, imo).
This PR avoids running regexpu twice (as noted in https://github.com/babel/babel/pull/6776#issuecomment-343255504) when transpiling to es2015 (`@babel/plugin-proposal-unicode-property-regex` + `@babel/plugin-transform-unicode-regex`) by delaying it when exiting the RegExp node.

Other than making the default behavior faster, this PR would allow us to drop the `useUnicodeFlag` option in the future and keep the "one plugin - one syntax" principle.

---
Benchmarks:

Without this commit: 1.45s
With this commit: 1.20s
<details>
<summary>Code</summary>

```js
const code = `
var regex = /\\p{Regional_Indicator}/u;
`.repeat(100);

const plugins = [
  loadPlugin("proposal-unicode-property-regex"),
  loadPlugin("transform-unicode-regex"),
];

const start = process.hrtime();
for (let i = 0; i < 100; i++) babel.transform(code, { plugins });
console.log(process.hrtime(start));
```

</details>

---
cc @mathiasbynens @loganfsmyth

### DIFF
--- a/packages/babel-helper-regex/package.json
+++ b/packages/babel-helper-regex/package.json
@@ -1,11 +1,11 @@
 {
   "name": "@babel/helper-regex",
   "version": "7.0.0-beta.31",
-  "description": "Helper function to check for literal RegEx",
+  "description": "Helper visitor to transform regular expressions using regexpu",
   "repository": "https://github.com/babel/babel/tree/master/packages/babel-helper-regex",
   "license": "MIT",
   "main": "lib/index.js",
   "dependencies": {
-    "lodash": "^4.2.0"
+    "regexpu-core": "^4.1.3"
   }
 }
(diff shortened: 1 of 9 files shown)


hzoo:
Don't really think is a perf issue worth fixing 🤔 , is it actually slower by that much? I don't think people have 100s of regexes on hundreds of files

loganfsmyth:
I definitely have a hard time imagining this will be a noticeable performance issue.

nicolo-ribaudo:
Well, actually... I agree with you :sweat_smile: After having measured the time I was undecided whether to close this pr or not.
Well, I at least I had fun making two different plugins work together :smile:

PR #6788, opened 2017-11-09
Add cache.clear() to clear the `data` reference when cache option is `false`

jdalton:
I noticed that even though caching can be disabled, on start it will load the cache json since its cache reference is never cleared. This patch clears the cache reference of `data` when caching is disabled.

### DIFF
--- a/packages/babel-register/src/node.js
+++ b/packages/babel-register/src/node.js
@@ -30,8 +30,7 @@ function installSourceMapSupport() {
   });
 }
 
-registerCache.load();
-let cache = registerCache.get();
+let cache;
 
 function mtime(filename) {
   return +fs.statSync(filename).mtime;
@@ -108,7 +107,13 @@ export default function register(opts?: Object = {}) {
   opts = Object.assign({}, opts);
   if (opts.extensions) hookExtensions(opts.extensions);
 
-  if (opts.cache === false) cache = null;
+  if (opts.cache === false && cache) {
+    registerCache.clear();
+    cache = null;
+  } else if (opts.cache !== false && !cache) {
+    registerCache.load();
+    cache = registerCache.get();
+  }
 
   delete opts.extensions;
   delete opts.cache;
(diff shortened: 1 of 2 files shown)


Andarist:
Wouldnt it make sense to skip [`load`](https://github.com/jdalton/babel/blob/6e6bd07ce1c078ff325dd7b363b9475c912c5108/packages/babel-register/src/cache.js#L43) too? It seems that was part of the issue you have noticed.

jdalton:
> Wouldnt it make sense to skip load too?

The problem is `@babel/register` initializes itself with default options before you can invoke it with custom options to disable caching, so the load will take place. By clearing the data at least it's saved empty so when it's loaded only an empty json (so empty cache) is loaded.

hzoo:
This related to https://github.com/babel/babel/pull/6783?

jdalton:
Maybe? \@loganfsmyth would know for sure.

jdalton:
Maybe it's more of a companion since I think clearing the cache is still a good behavior reguardless of #6783.

packages/babel-register/src/node.js
```diff
@@ -108,7 +108,10 @@ export default function register(opts?: Object = {}) {
   opts = Object.assign({}, opts);
   if (opts.extensions) hookExtensions(opts.extensions);
 
-  if (opts.cache === false) cache = null;
```
loganfsmyth: Should we consider changing this to

    if (opts.cache === false && cache) {
      registerCache.clear();

jdalton: Done!

packages/babel-register/src/node.js
```diff
@@ -30,8 +30,7 @@ function installSourceMapSupport() {
   });
 }
 
-registerCache.load();
-let cache = registerCache.get();
+let cache
```
loganfsmyth: Semicolon please 😬

jdalton: Doh, thanks!

jdalton:
Thank you @loganfsmyth, @hzoo, @Andarist!

PR #6789, opened 2017-11-09
Allow sourceType:unambiguous as a way to tell Babylon to guess the type.

### LINKED ISSUE #6695: Discussion: Should we have sourceType: "auto"
> Issue originally reported by @loganfsmyth in babel/babylon#440

In Babel 6, we've hardcoded everything to be `sourceType: "module"` by default, which has been a bit of a problem because people often run Babel on arbitrary code that they don't own, and sometimes that code is written assuming it will be treated as a script. This leads to complaints about `this` being replaced at the top level, among other things.

Would people be open to adding an `auto` sourceType option that would use the unambiguous grammar proposal? https://github.com/bmeck/UnambiguousJavaScriptGrammar

The longterm effect I'd expect from this for users is that users who have normal CommonJS modules won't automatically get `use strict` at the top of their files anymore, unless they opt into that plugin, which I think it a very reasonable thing to do.

The other side of it is that I want to start locking down the behavior of modules specifically, so usage of `module.exports` or `require` alongside `import` would be an error, which I believe is also the case in Webpack now? We'll want a flag so people can opt out of the behavior when switching from Babel 6 to 7, though.

loganfsmyth:
| Fixed Issues?            | Fixes #6695
| Patch: Bug Fix?          |
| Documentation PR         | 
| Any Dependency Changes?  |

Repost of https://github.com/babel/babylon/pull/449. Also added tweaks to option parsing so that the option passes through properly.


hzoo:
Should add a readme change for the new option?

packages/babylon/src/index.js
```diff
@@ -91,3 +111,16 @@ function getParserClass(
...
+    child =>
+      (child.type === "ImportDeclaration" &&
+        (!child.importKind || child.importKind === "value")) ||
+      (child.type === "ExportNamedDeclaration" &&
+        (!child.exportKind || child.exportKind === "value")) ||
+      (child.type === "ExportAllDeclaration" &&
+        (!child.exportKind || child.exportKind === "value")) ||
+      child.type === "ExportDefaultDeclaration",
+  );
+}
```
jdalton: 👆 Neat!

packages/babylon/src/index.js
```diff
@@ -18,7 +18,27 @@ plugins.jsx = jsxPlugin;
...
+      if (!hasModuleSyntax(ast)) ast.program.sourceType = "script";
+      return ast;
+    } catch (moduleError) {
+      try {
+        options.sourceType = "script";
+        return getParser(options, input).parse();
+      } catch (scriptError) {}
+
+      throw moduleError;
+    }
```
jdalton: 👆 This aligns with [my approach](https://github.com/standard-things/esm/blob/bf531b69a408111fa6e66ac033a2230394483392/src/compiler.js#L68-L91) as well.

packages/babylon/src/index.js
loganfsmyth: Yeah this is basically to stay in line with https://github.com/babel/babylon/pull/771 since Flowtype just indiscriminately allows mixing ES6 import/export and CommonJS and it seems fine to allow type import/export inside `sourceType:script` files.

jdalton: @loganfsmyth Who made your avatar!

packages/babel-core/src/config/option-assertions.js
```diff
@@ -46,8 +46,15 @@ export function assertSourceType(
...
-    throw new Error(`.${key} must be "module", "script", or undefined`);
+  if (
+    value !== undefined &&
+    value !== "module" &&
+    value !== "script" &&
+    value !== "unambiguous"
+  ) {
+    throw new Error(
+      `.${key} must be "module", "script", "unambiguous", or undefined`,
+    );
```
jdalton: The alphabetical listing of these sourceTypes is satisfying!

packages/babylon/src/index.js
loganfsmyth: Got it from Matt Cummings. His commissions are open again: https://twitter.com/EiffelArt/status/928017289416495105

packages/babylon/src/index.js
```diff
@@ -91,3 +111,16 @@ function getParserClass(
   }
   return cls;
 }
+
```
jdalton: Could `hasModuleSyntax` be simplified?
Is there ever a `child.type` of `ImportDeclaration` and friends that don't have the accompanioning `importKind` and `exportKind` values?

loganfsmyth: I believe when the `flow` parsing isn't enabled, they aren't set.

PR #6792, opened 2017-11-10
Do not access out of bounds arguments

apapirovski:
| Fixed Issues?            | 

In `babel-plugin-transform-parameters`, potentially out of bounds arguments should not just be directly accessed, instead use a ternary operator to check length. Accessing out of bounds is expensive ~~and means that engines are unable to optimize the function~~.

(A plain benchmark on accessing out of bounds arguments shows about 5x slower execution for a basic function that literally just accesses a single argument.)

### DIFF
--- a/packages/babel-plugin-transform-parameters/src/params.js
+++ b/packages/babel-plugin-transform-parameters/src/params.js
@@ -19,8 +19,8 @@ const buildLooseDestructuredDefaultParam = template(`
   let ASSIGNMENT_IDENTIFIER = PARAMETER_NAME === UNDEFINED ? DEFAULT_VALUE : PARAMETER_NAME ;
 `);
 
-const buildArgumentsAccess = template(`
-  let $0 = arguments[$1];
+const buildSafeArgumentsAccess = template(`
+  let $0 = arguments.length > $1 ? arguments[$1] : undefined;
 `);
 
 function isSafeBinding(scope, node) {
@@ -110,7 +110,10 @@ export default function convertFunctionParams(path, loose) {
       });
       body.push(defNode);
     } else if (firstOptionalIndex !== null) {
-      const defNode = buildArgumentsAccess([param.node, t.numericLiteral(i)]);
+      const defNode = buildSafeArgumentsAccess([
+        param.node,
+        t.numericLiteral(i),
+      ]);
       body.push(defNode);
     } else if (param.isObjectPattern() || param.isArrayPattern()) {
       const uid = path.scope.generateUidIdentifier("ref");
--- a/packages/babel-plugin-transform-parameters/test/fixtures/parameters/default-rest-mix/expected.js
+++ b/packages/babel-plugin-transform-parameters/test/fixtures/parameters/default-rest-mix/expected.js
@@ -1,11 +1,13 @@
 function fn(a1) {
   var a2 = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : 4;
-  var _arguments$ = arguments[2],
-      a3 = _arguments$.a3,
-      a4 = _arguments$.a4;
-  var a5 = arguments[3];
 
-  var _ref = arguments.length > 4 && arguments[4] !== undefined ? arguments[4] : {},
-      a6 = _ref.a6,
-      a7 = _ref.a7;
+  var _ref = arguments.length > 2 ? arguments[2] : undefined,
+      a3 = _ref.a3,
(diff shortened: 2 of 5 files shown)


hzoo:
Oh right, similar to @bmeurer's PRs except for our generated code rather than Babel itself! https://github.com/babel/babel/issues?q=label%3A%22area%3A+perf%22+is%3Aclosed

apapirovski:
Sorry, missed a test in a separate package. Fixed now.

bmeurer:
Just a side note: Even if you access out of bounds, the function can still be optimized, just the actual property access will be really slow and not optimized.

apapirovski:
@bmeurer Good to know! 👍Wish there was more detailed information on all this stuff somewhere... Feel like I'm piecing together information from a million article, your twitter & reading Chrome source. 😆

packages/babel-plugin-transform-parameters/test/fixtures/parameters/default-rest-mix/expected.js
```diff
@@ -3,9 +3,9 @@ function fn(a1) {
   var _arguments$ = arguments[2],
```
nicolo-ribaudo: Shouldn't this also be `arguments.length > 2 ? arguments[2] : undefined`?

nicolo-ribaudo: Mh, I guess it isn't needed because if `arguments.length > 2` the following line is going to throw, so optimizing for that case it is unuseful... Is it correct?

bmeurer: That means that a single invocation where this function throws is enough to turn the property accesses into slow mode for the rest of the programs execution. We should try to avoid this.

nicolo-ribaudo: Property access in general or just `arguments[...]` in this function?

bmeurer: This particular `arguments[2]` property access.

packages/babel-plugin-transform-parameters/src/params.js
```diff
@@ -110,7 +114,10 @@ export default function convertFunctionParams(path, loose) {
       });
       body.push(defNode);
     } else if (firstOptionalIndex !== null) {
-      const defNode = buildArgumentsAccess([param.node, t.numericLiteral(i)]);
+      const defNode =
+        param.isObjectPattern() || param.isArrayPattern()
```
apapirovski: I'm planning to remove this check based on @bmeurer's feedback so please do not merge yet.

apapirovski:
Sorry for the delay, everyone — traveling most of the week so was hard to find the time. Implemented feedback per @bmeurer.

blikblum:
> (A plain benchmark on accessing out of bounds arguments shows about 5x slower execution for a basic function that literally just accesses a single argument.)

@apapirovski 
The benchmark is public?

@bmeurer 
The overhead occurs only with function arguments or also affects normal arrays?

In #6582 and #6580 the offending index is a not valid array index (-1). Here the array index is a valid one (positive integer). 

Does accessing an array with an out of bounds index, even if a valid one, gives the same behavior as accessing an array with an invalid index?

bmeurer:
@blikblum This generally affects all JavaScript objects, especially arrays and arguments objects (which is not an array, but an exotic object). No matter if the array was constructed from a literal, the `Array` constructor or a rest parameter declaration.

Nowadays with a very recent V8, accessing with a valid array index - in the range `[0, 2^32-1[` - only incurs a small overhead, which is often optimized away completely by TurboFan, but accessing with any other integer values triggers the generic slow-path for the property access site.

blikblum:
@bmeurer Many thanks.

> Nowadays with a very recent V8, accessing with a valid array index - in the range [0, 2^32-1[ - only incurs a small overhead

For the record, we created a [benchmark](https://jsperf.com/radio-callhandler/1) to see the impact of avoiding out of bounds array access (with valid indexes) in the software we maintain. In Chrome  the impact is huge, but in Firefox the difference is minimal. Interestingly in Mobile, both browsers shows small to no gains.

I assume the Chrome version tested (v62 64bit on Windows) does not have the newest v8 changes

bmeurer:
You should probably test with M64.

PR #6803, opened 2017-11-12
Add /*#__PURE__*/ annotatiotion for babel-plugin-async-to-generator

### LINKED ISSUE #6572: babel-plugin-async-to-generator should add #__PURE__ comment
Request: add pure comments to async function helpers to aid in dead code elimination.

### Input Code

https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=IYZwngdgxgBAZgV2gFwJYHsLwKbKgCwAVgBzbAEWGWAAoAHU7AShgG8AoGGAJ1wW6ysAvuyFA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cstage-2&targets=&version=6.26.0

```js
async function fetchPageData(page) {
  return {}
}
```

### Expected Behavior
Generated code should be able to be removed as dead by uglifyjs/similiar if the async function is not called.

### Current Behavior
```js
var fetchPageData = function () {
  var _ref = _asyncToGenerator( /*#__PURE__*/regeneratorRuntime.mark(function _callee(page) {
    return regeneratorRuntime.wrap(function _callee$(_context) {
      while (1) {
          case 0:
            return _context.abrupt("return", {});

          case 1:
          case "end":
            return _context.stop();
        }
      }
    }, _callee, this);
  }));

  return function fetchPageData(_x) {
    return _ref.apply(this, arguments);
  };
}();
```

Note that while regeneratorRuntime.mark is marked as #__PURE__, the call to _asyncToGenerator and the IIFE aren't.

### Possible Solution

Generated code should start with:
var fetchPageData = /*#__PURE*/ function () {

### Context
I have a react app which runs on both client and server. However, I have various async functions that only run on the server. I want to have them removed during the minification process.

satya164:
| Fixed Issues?            | Fixes #6572

This adds the `/*#__PURE__*/` annotation before the IIFE generated by async to generator plugin so that uglify can perform DCE if it's not used. Tested and verified that uglify does remove the code.


satya164:
Excited for my first contribution to Babel. Thanks a lot @xtuc for helping me with this ❤️

packages/babel-plugin-transform-async-to-generator/test/fixtures/async-to-generator/object-method-with-arrows/expected.js
```diff
@@ -1,4 +1,5 @@
 class Class {
+  /*#__PURE__*/
```
xtuc: I'm not sure about this one :thinking:

satya164: Should I add a check for function declarations before adding the annotation?

satya164:
Not sure what's going on with the tests. The passed on my machine. Will take a look later today.

packages/babel-helper-remap-async-to-generator/package.json
```diff
@@ -6,6 +6,7 @@
   "license": "MIT",
   "main": "lib/index.js",
   "dependencies": {
+    "@babel/helper-annotate-as-pure": "7.0.0-beta.31",
```
Andarist: probably this need an update, `beta.32` is released

xtuc:
cc @kzc FYI and if it looks ok to you?

xtuc:
Awesome contribution @satya164, thanks!

PR #6811, opened 2017-11-13
Update README.md

culshaw:
This cropped me up, so updated to reflect where preset-env is available in regards to .babelrc examples when it gets installed, otherwise you get a missing preset error [skip ci]

### DIFF
--- a/experimental/babel-preset-env/README.md
+++ b/experimental/babel-preset-env/README.md
@@ -10,7 +10,7 @@ Without any configuration options, @babel/preset-env behaves exactly the same as
 
 ```json
 {
-  "presets": ["@babel/env"]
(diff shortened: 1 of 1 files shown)


existentialism:
@culshaw can you give an example of where the above fails? It should work as configuration's still respect shorthand (albeit w/ the prefix), whereas it _doesn't_ work for when installing via `npm`.

culshaw:
@existentialism Hi, I was using yarn

existentialism:
@danez yep, we landed https://github.com/babel/babel/pull/6820 for this

hzoo:
Thanks for the PR, ok closing since it seemed to be covered in #6820?

Issue #5703, opened 2017-05-04
Babel writes files in a directory concurrently, not consecutively


kevinburke:
If I ask Babel to compile all of the files in a directory, it will compile files and then write them to disk one at a time.

If I understand correctly, Node uses a single CPU, the `util.compile` in dir.js is synchronous:

```js
    const data = util.compile(src, defaults({
      sourceFileName: slash(path.relative(dest + "/..", src)),
      sourceMapTarget: path.basename(relative),
    }, opts));
```

Then the file gets written to disk synchronously:

```js
    outputFileSync(dest, data.code);
    util.chmod(src, dest);
```

It seems like the compilation of one file cannot affect the compilation of other files, nor can writing them to disk, so there is a potential gain by parallelizing these actions.

kevinburke:
A slightly related problem is I had trouble only compiling files that had a newer mtime than their counterparts in the `--out-dir`.

I ended up working around this with this build helper and a Make target - see the README for more details.

https://github.com/kevinburke/recompile

loganfsmyth:
I'm curious, are you running the CLI often enough that the mtime makes a difference? Normally I'd expect users to use `-w` to keep the CLI running, then it will only recompile files that have changed, avoiding the need to compare mtimes.

As for this issue, you are right that it writes the files to disk one at a time. That said, it is by far spending more time compiling the files themselves than it is writing them to disk. If we wanted to make it faster, we'd have to actually parallelize the transformations themselves across multiple processes. Given a normal incremental-build workflow, usually these types of optimizations aren't as critical.

kevinburke:
My asset pipeline is roughly:

1. Write JSX file
2. Detect change in JSX file (via github.com/jmhodges/justrun). justrun kills the running server process
3. Run "make assets"
4. Make target checks for new files in the `components` directory:

    JSX_TARGETS := $(shell find ./components -name '*.jsx' -depth 1)
    JS_TARGETS := $(shell find ./public -name '*.js' -depth 1)
    RECOMPILE := $(shell command -v recompile)

    $(JS_TARGETS): $(JSX_TARGETS)
    ifndef RECOMPILE
	go get -u github.com/kevinburke/recompile
    endif
	recompile --command='./node_modules/.bin/babel --plugins transform-react-jsx' --extension=js --out-dir=public "$?"

    recompile: $(JS_TARGETS)

5. That make target compiles the JS file and writes JSX
6. `go-bindata` copies the JS file into the binary
[...]

PR #6826, opened 2017-11-14
Use the async version of transform in babel-cli

### LINKED ISSUE #6784: Update babel-cli to use the async version of babel.transform
As of https://github.com/babel/babel/pull/6780, Babel has both `babel.transform` and `babel.transformSync`. For backward-compat you can leave the callback off of `babel.transform` and it'll behave synchronously, but ideally for all the places that _can_ shift to using the async version, they should.

The `babel` command in babel-cli is one of the places that can easily use the async version. The hard thing will be updating all of the surrounding code to to handle things asynchronously too.

This should be pretty self-contained, so I'm marking it as a `good first issue`, but it'll definitely involve a good bit of refactoring.

aprieels:
| Fixed Issues?            | Fixes #6784

This is my first contribution so let me know if something's wrong of if it's not what you expected.

### DIFF
--- a/packages/babel-cli/src/babel/dir.js
+++ b/packages/babel-cli/src/babel/dir.js
@@ -48,7 +48,8 @@ export default function(commander, filenames, opts) {
         outputFileSync(dest, res.code);
         util.chmod(src, dest);
 
-        return callback(null, src + " -> " + dest);
+        util.log(src + " -> " + dest);
+        return callback(null, true);
       },
     );
   }
@@ -72,7 +73,27 @@ export default function(commander, filenames, opts) {
         util.chmod(src, dest);
       }
 
-      return callback(null, res);
+      return callback();
+    });
+  }
+
+  function sequentialHandleFile(files, dirname, index, callback) {
+    if (typeof index === "function") {
+      callback = index;
+      index = 0;
+    }
+
+    const filename = files[index];
+    const src = path.join(dirname, filename);
+
+    handleFile(src, filename, dirname, function(err) {
+      if (err) return callback(err);
+      index++;
+      if (index !== files.length) {
+        sequentialHandleFile(files, dirname, index, callback);
+      } else {
+        callback();
+      }
     });
   }
 
@@ -88,21 +109,8 @@ export default function(commander, filenames, opts) {
         util.deleteDir(commander.outDir);
       }
 
-      const results = [];
-      let filesProcessed = 0;
-
       const files = util.readdir(dirname, commander.includeDotfiles);
-      files.forEach(function(filename, index) {
-        const src = path.join(dirname, filename);
-        handleFile(src, filename, dirname, function(err, res) {
-          if (err) return callback(err);
-          results[index] = res;
-          filesProcessed++;
-          if (filesProcessed === files.length) {
-            return callback(null, results);
-          }
-        });
-      });
+      sequentialHandleFile(files, dirname, callback);
     } else {
       write(
         filename,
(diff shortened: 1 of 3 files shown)


packages/babel-cli/src/babel/file.js
```diff
@@ -103,19 +103,21 @@ export default function(commander, filenames, opts) {
...
+        commander.filename,
+        code,
+        defaults(
+          {
+            sourceFileName: "stdin",
+          },
+          opts,
         ),
+        function(err, result) {
+          if (err) console.error(err.stack);
```
loganfsmyth: To be most consistent with the old behavior, this should probably be `throw err;`

packages/babel-cli/src/babel/util.js
```diff
@@ -50,12 +50,12 @@ export function log(msg) {
...
+export function transform(filename, code, opts, cb) {
   opts = Object.assign({}, opts, {
     filename,
   });
 
-  return babel.transform(code, opts);
+  babel.transform(code, opts, cb);
 }
 
 export function compile(filename, opts) {
```
loganfsmyth: Sorry, I didn't realize this function used a different version, but we'll want to update both. This function uses `babel.transformFileSync` and ideally this same PR should also update that to use the async `babel.transformFile`.

That will probably be the complicated part of this task because the code that uses `util.compile` is mostly synchronous right now.

packages/babel-cli/src/babel/dir.js
```diff
@@ -26,41 +30,53 @@ export default function(commander, filenames, opts) {
...
+          res.map &&
+          commander.sourceMaps &&
+          commander.sourceMaps !== "inline"
+        ) {
+          const mapLoc = dest + ".map";
+          res.code = util.addSourceMappingUrl(res.code, mapLoc);
+          outputFileSync(mapLoc, JSON.stringify(res.map));
+        }
+
+        outputFileSync(dest, res.code);
```
nicolo-ribaudo: We could write the file asynchronously

hzoo:
So in order for us to support returning a promise we need to wait for another version bump? This is only for cli though so I'd be ok with doing this earlier?

packages/babel-cli/src/babel/dir.js
nicolo-ribaudo: It would also fix https://github.com/babel/babel/issues/5703

aprieels:
I'm sorry but I don't really understand what you mean. What should I change?

nicolo-ribaudo:
@aprieels I think @hzoo's comment was more about https://github.com/babel/babel/pull/6780

aprieels:
@loganfsmyth Sorry for the late modifications, this should be better know. Let me know if you have any comment!

aprieels:
@loganfsmyth Sorry for the late modifications, this should be better know. Let me know if you have any comment!

packages/babel-cli/src/babel/dir.js
```diff
@@ -104,7 +154,10 @@ export default function(commander, filenames, opts) {
         watcher.on(type, function(filename) {
           const relative = path.relative(dirname, filename) || filename;
           try {
-            handleFile(filename, relative);
+            handleFile(filename, relative, function(err, res) {
+              if (err) throw err;
+              util.log(res);
```
loganfsmyth: Looks like this may be a mistake? There doesn't appear to be a `res` value passed to this callback unless I'm misreading, and there wasn't a lot here before.

aprieels: You're right. I've removed it.

PR #6828, opened 2017-11-14
Parse TS type parameters when JSX is enabled.

### LINKED ISSUE #6665: [Presets: React + Typescript] Complex method with generic type fails during transpilation
> Issue originally reported by @babel-bot in babel/babylon#775

Original issue submitted by @pbadenski in https://github.com/babel/babel/issues/6567

Bug.

### Input Code
https://github.com/pbadenski/babel-typescript-sample

```js
const liftA2 = <T, V> (x: X<T, V>): (f: Function) => null;
```

### Babel Configuration

After checking out the code at the link above:
```npm run transpile```

### Current Behavior
```
{ SyntaxError: src/a.ts: Unexpected token (1:17)
> 1 | const liftA2 = <T, V> (x: X<T, V>): (f: Function) => null;
    |                  ^
  2 | 
    at _class.raise (/Users/pbadenski/workspace/babel-typescript-sample/node_modules/babylon/lib/index.js:807:15)
    at _class.unexpected (/Users/pbadenski/workspace/babel-typescript-sample/node_modules/babylon/lib/index.js:2108:16)
    at _class.jsxParseIdentifier (/Users/pbadenski/workspace/babel-typescript-sample/node_modules/babylon/lib/index.js:8031:14)
    at _class.jsxParseNamespacedName (/Users/pbadenski/workspace/babel-typescript-sample/node_modules/babylon/lib/index.js:8042:23)
    at _class.jsxParseAttribute (/Users/pbadenski/workspace/babel-typescript-sample/node_modules/babylon/lib/index.js:8134:24)
    at _class.jsxParseOpeningElementAt (/Users/pbadenski/workspace/babel-typescript-sample/node_modules/babylon/lib/index.js:8146:35)
    at _class.jsxParseElementAt (/Users/pbadenski/workspace/babel-typescript-sample/node_modules/babylon/lib/index.js:8167:33)
    at _class.jsxParseElement (/Users/pbadenski/workspace/babel-typescript-sample/node_modules/babylon/lib/index.js:8228:19)
    at _class.parseExprAtom (/Users/pbadenski/workspace/babel-typescript-sample/node_modules/babylon/lib/index.js:8238:21)
    at _class.parseExprSubscripts (/Users/pbadenski/workspace/babel-typescript-sample/node_modules/babylon/lib/index.js:2864:21) pos: 17, loc: null }
```

### Context
Using https://github.com/gcanti/fp-ts

### Your Environment

| ---------------- | -------
| Babel            |  next (7.0.0-beta.3)
| babel-preset-typescript | next (7.0.0-beta.3)
| babel-preset-react | next (7.0.0-beta.3)
| node             | v8.8.1
| npm              | 5.5.1
| Operating System | Mac OS X 10.12.6

### PR DESCRIPTION
**Pull request by nicolo-ribaudo**
| Fixed Issues?            | Fixes #6665 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

This bug was caused by he JSX plugin, which tokenized "<" as a `jsxTagStart` even when it shouldn't.
I initially tried to fix this bug by setting `this.state.inType = true` before tokenizing `<`, but it is not always possible. For example, in `class C <T> {}` and `class C extends X {}` you don't know if you are parsing a type parameter or not until you tokenize `<` (but then it's too late for setting `this.state.inType`).

### DIFF
--- a/packages/babylon/src/plugins/jsx/index.js
+++ b/packages/babylon/src/plugins/jsx/index.js
@@ -69,8 +69,26 @@ function getQualifiedJSXName(
   throw new Error("Node had unexpected type: " + object.type);
 }
 
+function assert(x: boolean): void {
+  if (!x) {
+    throw new Error("Assert fail");
+  }
+}
+
 export default (superClass: Class<Parser>): Class<Parser> =>
   class extends superClass {
+    jsxTagStartToRelational() {
+      assert(this.match(tt.jsxTagStart));
+      this.state.type = tt.relational;
+      this.state.value = "<";
+
+      // Pop the context added by the jsxTagStart.
+      assert(this.curContext() === tc.j_oTag);
+      this.state.context.pop();
+      assert(this.curContext() === tc.j_expr);
+      this.state.context.pop();
+    }
+
     // Reads inline JSX contents token.
 
     jsxReadToken(): void {
--- a/packages/babylon/src/plugins/typescript.js
+++ b/packages/babylon/src/plugins/typescript.js
@@ -64,6 +64,8 @@ function keywordTypeFromName(
 
 export default (superClass: Class<Parser>): Class<Parser> =>
   class extends superClass {
+    +jsxTagStartToRelational: () => void;
+
     tsIsIdentifier(): boolean {
       // TODO: actually a bit more complex in TypeScript, but shouldn't matter.
       // See https://github.com/Microsoft/TypeScript/issues/15008
@@ -270,8 +272,12 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return this.finishNode(node, "TSTypeParameter");
     }
 
+    tsIsStartOfTypeParameters(): boolean {
+      return this.isRelational("<") || this.match(tt.jsxTagStart);
+    }
+
     tsTryParseTypeParameters(): ?N.TsTypeParameterDeclaration {
-      if (this.isRelational("<")) {
+      if (this.tsIsStartOfTypeParameters()) {
         return this.tsParseTypeParameters();
       }
     }
@@ -279,7 +285,13 @@ export default (superClass: Class<Parser>): Class<Parser> =>
     tsParseTypeParameters() {
       const node: N.TsTypeParameterDeclaration = this.startNode();
 
-      if (this.isRelational("<") || this.match(tt.jsxTagStart)) {
+      // In case the first token of the type has been parsed as a JSX tag
+      // start, we need to translate it.
+      if (this.match(tt.jsxTagStart)) {
+        this.jsxTagStartToRelational();
+      }
+
+      if (this.isRelational("<")) {
         this.next();
       } else {
         this.unexpected();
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: packages/babylon/src/plugins/jsx/index.js [hunk 1]
```diff
@@ -69,8 +69,26 @@ function getQualifiedJSXName(
...
+function assert(x: boolean): void {
+  if (!x) {
+    throw new Error("Assert fail");
+  }
+}
+
 export default (superClass: Class<Parser>): Class<Parser> =>
   class extends superClass {
+    jsxTagStartToRelational() {
+      assert(this.match(tt.jsxTagStart));
```
nicolo-ribaudo: Not important for this pr, but we might remove these `assert` calls in the production build.

packages/babylon/src/plugins/typescript.js
```diff
@@ -64,6 +64,8 @@ function keywordTypeFromName(
 
 export default (superClass: Class<Parser>): Class<Parser> =>
   class extends superClass {
+    +jsxTagStartToRelational: () => void;
```
xtuc: Sorry for my ignorance but what is this syntax?

nicolo-ribaudo: It means that that method is defined in another class (a subclass or a superClass that Flow can't infer) but it will available at runtime.

nicolo-ribaudo: @xtuc I just read the docs and I was wrong :sweat_smile:  It is just a read-only type annotation, which can be used to declare methods which flow can't infer.

xtuc: ok :smile: thanks

packages/babylon/src/plugins/jsx/index.js
xtuc: I've already planned to create something like that on one of my project. Should I work on this?

andy-ms:
Shouldn't this have affected the flow plugin as well?

nicolo-ribaudo:
For some reason, it already works: https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&code_lz=DwKwzgHgBA9AfAKAQNwIYCcoQFxWAFTgAoBKKAXjigG8BfCqVMATwDsBjPQ0iqugbgQAzAK4cALgEsA9qyhCCxAA4ZUAW1yKelGrRKbuZHXV0J2AGyZgoAYUU0EUKEvTSlAU3TjmB4kb60gk5q7uIAFtIAJlpkdAi0SJKs4p5CqOzuUACS9tSOUKzuAO5cfrj4ADT5MeVVTgDarOruuGDi6EkA5gC6vtoB8fxAA&debug=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=false&presets=es2015%2Ces2017%2Creact%2Cstage-0%2Cflow&prettier=false&targets=&version=7.0.0-beta.36

Btw, in the future I think that we should try to merge the typescript and the flow plugins as much as possible, so that if something works for one plugin also works for the other.

andy-ms:
Made a pull request at #7225 that would solve this in a way similar to how flow does it.

nicolo-ribaudo:
:+1: I like your PR more.

PR #6834, opened 2017-11-15
Expose `envName` as a top-level Babel option to avoid using environmental variables

loganfsmyth:
| Fixed Issues?            | 
| Documentation PR         | 
| Any Dependency Changes?  |

Been wanting to do this for a while anyway, but figured this was something we should do. Given that https://github.com/babel/babel/pull/6780 now allows for the potential that config loading won't run synchronously when Babel is executed, it wouldn't necessarily be safe for users to do explicitly assign `process.env.BABEL_ENV` because now in theory two separate Babel transforms could have been started in parallel. We do this in our own codebases with `babel-loader`'s `forceEnv` option.

Centralizing the `getEnv` call to the top level of config loading also ensures that in the future if we make thing more async, we won't run into cases where two plugins in the same loading pass get two different values for the `env`.

This code got complicated because of the way we do caching, sorry about that. I think the way the code ended up is pretty readable though 😬 .

The quick summary of why it got complicated is that `envName` is almost like a separate flag that works alongside the cache. I didn't want to have to have a separate instance of every single plugin for every single `envName`, and that is all the existing caching API allowed. Ideally only the plugins that actually _query_ the envName should get new instantiations, so that is what this code allows.

### DIFF
--- a/packages/babel-cli/src/babel/index.js
+++ b/packages/babel-cli/src/babel/index.js
@@ -49,6 +49,11 @@ commander.option(
   collect,
 );
 commander.option("--config-file [path]", "Path a to .babelrc file to use");
+commander.option(
+  "--env-name [name]",
+  "The name of the 'env' to use when loading configs and plugins. " +
+    "Defaults to the value of BABEL_ENV, or else NODE_ENV, or else 'development'.",
+);
 
 // Basic file input configuration.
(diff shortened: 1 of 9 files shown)


loganfsmyth:
Happy to respond to comments, just gonna end up with lots of annoying conflicts on this otherwise :P

Issue #6772, opened 2017-11-08
Path.ensureBlock does not set ArrowFunctionExpression's "expression" flag


hansonw:
### Input Code

In the following snippet, calling `path.ensureBlock()` does not properly unset the "expression" flag on the ArrowFunctionExpression. This causes problems for other plugins, which may rely on the accuracy of the flag (notably, istanbul-lib-instrument).

```js
export default function (babel) {
  const { types: t } = babel;
  
  return {
    name: "ast-transform", // not required
    visitor: {
      ArrowFunctionExpression(path) {
        path.ensureBlock();
        if (path.node.expression) {
          throw Error("Should no longer be an expression");
        }
      },
    }
  };
}
```

https://astexplorer.net/#/gist/0bd0506d529ed4c59a4ab10ffc2bd105/c23cb1d08aed3b721a53142b4d3bc6fb14d2af7b

### Babel/Babylon Configuration (.babelrc, package.json, cli command)

See ASTExplorer link.

### Expected Behavior

The "expression" flag should be set to false.

### Possible Solution

It sounds like it would be enough to add a `path.set("expression", false)`  in https://github.com/babel/babel/blob/master/packages/babel-traverse/src/path/conversion.js#L26 if the node type is an ArrowFunctionExpression? I'm not sure if there are other cases like this in the AST spec.

loganfsmyth:
What a strange flag. I didn't even know that existed, and my vote would just be to remove it. If you need to know what type of body the function has, just check `t.isBlockStatement(fn.body)`.

hansonw:
It is part of the spec after all :) https://github.com/babel/babylon/blob/master/ast/spec.md#arrowfunctionexpression
The only reason I care is because istanbul's instrumenter relies on it: https://github.com/istanbuljs/istanbuljs/blob/master/packages/istanbul-lib-instrument/src/visitor.js#L346

I can follow up with a PR over there to stop using it I guess; that's easy enough :)

loganfsmyth:
Yeah I agree that it's in the spec. Babel almost certainly just copied it from ESTree. It seems like a case where it's a nice-to-have if you have a static tree, but if you have to do transformations like we do, it's a duplicate data source that always needs to be in sync with `fn.body`, and if that's the case, why have it in the first place.

[...]

PR #6836, opened 2017-11-15
removing expression field from ArrowFunctionExpression

### LINKED ISSUE #6773: Remove the `.expression` field from `ArrowFunctionExpression`
Context: https://github.com/babel/babel/issues/6772#issuecomment-342935685

As far as I can see, there's no reason to have this flag, so my vote would be to kill it instead of expecting us (and anyone else writing transforms) to remember to always keep it in sync with the type of the body node.

Thoughts?

mmckeaveney:
| Fixed Issues?            | Fixes #6773 
| Tests Added + Pass?      | N

Removed references to the `.expression` field on ArrowFunctionExpressions and removed it from the ast spec. See https://github.com/babel/babel/issues/6772#issuecomment-342935685 for rationale.


mmckeaveney:
No problem @nicolo-ribaudo, thanks very much for the quick feedback. I'll make the changes :)

mmckeaveney:
@nicolo-ribaudo do you mean I need to amend the `isStrictBody` function to do the check on the node body, or do you mean to actually replace the `isStrictBody` call + declaration with a simple check on the body type? 

```js
// this
const isStrict = node.body.type === "BlockStatement";

// or this

 isStrictBody(
    node: { body: N.BlockStatement }
  ): boolean {
    if (node.body.type === "BlockStatement" && node.body.directives.length) {
      for (const directive of node.body.directives) {
        if (directive.value.value === "use strict") {
          return true;
        }
      }
    }

    return false;
  }
```

sorry If I'm misunderstanding.. Still trying to get to grips with the codebase

nicolo-ribaudo:
You should update the function and remove the `node.expression` parameter from the call.

mmckeaveney:
Ok, I've pushed the latest with the function updated and the tests regenerated. As you can see from the checks, there are still some tests failing. I'm getting similar errors to the following for 26 tests:

```
Error: flow/regression/issue-58-failing-1/actual.js: Expected error message: Invalid left-hand side in arrow function parameters (2:11). Got error message: Cannot read property 'length' of undefined
```

I thought it was something to do with the `directives` attribute on `node.body` being undefined instead of an empty array, but even with a check in for both scenarios the tests still fail. I'm not too sure how to debug the issue from here.

loganfsmyth:
Since we were doing it already, I believe assuming that `.directives` exists, so you can revert that change. The failure is because this function has an override in https://github.com/babel/babel/blob/cc66495a95320529623f4caea20b3327875305c4/packages/babylon/src/plugins/estree.js#L131 that also needs to be adjusted.

mmckeaveney:
Updated - Thanks for the help.

packages/babylon/src/plugins/estree.js
```diff
@@ -128,11 +128,10 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
-    isStrictBody(
-      node: { body: N.BlockStatement },
-      isExpression: ?boolean,
-    ): boolean {
-      if (!isExpression && node.body.body.length > 0) {
+    isStrictBody(node: { body: N.BlockStatement }): boolean {
+      const isBlockStatement = node.body.type === "BlockStatement";
+
+      if (isBlockStatement && node.body.body.length > 0) {
         for (const directive of node.body.body) {
```
danez: As the flag is part of the estree spec, the estree plugin needs to ensure that this field stays set.

mmckeaveney:
@danez Updated.

danez:
That's not what i meant. I meant that the field `expression` that you are removing with this PR needs to be kept when the estree plugin is active.

So you remove the field in `expression.js` but in `plugins/estree.js` the field needs to be readded again.

mmckeaveney:
Sorry about that. I've updated the PR to re-add the field when the estree plugin is active.

packages/babylon/src/plugins/estree.js
```diff
@@ -239,6 +238,9 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       node.raw = node.extra.raw;
       delete node.extra;
 
+      const isExpression = !this.match(tt.braceL);
+      node.expression = isExpression;
```
loganfsmyth: Could this be added in an override for `parseFunctionBody`? That's the function where you initially removed `this.expression = false`, so I could see us doing

```
parseFunctionBody(node: N.Function, allowExpression: ?boolean): void {

Comment (mmckeaveney-r7) [reply packages/babylon/src/plugins/estree.js]: Updated. Thanks

Comment (danez) [reply packages/babylon/src/plugins/estree.js]: I think `parseFunctionBody` is to generic as it would add the field also to object methods for example. The field should only show up on arrow functions according to the estree spec, so maybe we should overwrite `parseArrowExpression` instead.

```js
parseArrowExpression(
  node: N.ArrowFunctionExpression,
  params?: ?(N.Expression[]),
  isAsync?: boolean,
): N.ArrowFunctionExpression {
  node = super.parseFunctionBody(node, allowExpression);
  node.expression = node.body.type !== "BlockStatement";

  return node;
}
```

danez: Okay seems that is there for backwards compatibility with this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Expression_Closures

So ignore my previous comment  :)

Can you also add the field back in `initFunction` by overwriting the function in estree.js, then the estree plugin should do the same the normal parser did before.

```js
initFunction(node: N.BodilessFunctionOrMethodBase, isAsync: ?boolean): void {
  super.initFunction(node, isAsync);
  node.expression = false;
}
```

mmckeaveney: @danez good catch. PR updated.

nicolo-ribaudo:
Thank you :smiley:

mmckeaveney:
Thanks guys, really appreciate the help and reviews. ☺

PR #6843, opened 2017-11-17
README: Use HTTPS and relative links when possible

mc10:

Just making links a bit cleaner.


hzoo:
The reason we weren't using relative links is probably since for the other repos we copy them to the website so the urls wouldn't work (on babeljs.io)

It's fine for this since I don't think this is used anywhere else though

PR #6844, opened 2017-11-17
Add "make watch-babylon" [skip ci]

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | 
| Documentation PR         | 
| Any Dependency Changes?  |

Please  :pray:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,10 @@ watch: clean
 	make clean-lib
 	BABEL_ENV=development ./node_modules/.bin/gulp watch
 
+watch-babylon:
+	cd packages/babylon; \
+	./node_modules/.bin/rollup -c -w
+
 flow:
 	./node_modules/.bin/flow check --strip-root
 


Makefile
```diff
@@ -42,6 +42,10 @@ watch: clean
 	make clean-lib
 	BABEL_ENV=development ./node_modules/.bin/gulp watch
```
xtuc: Do we want to include this in the normal `watch`?

nicolo-ribaudo: Does make support running two commands at the same time?

xtuc: Good point, that will probably be too complicated. LGTM then.

xtuc:
I'll go ahead and merge this. Thanks @nicolo-ribaudo.

hzoo:
This is totally fine, but I think we should just go with logan's suggestion in the future of compiling the individual files in watch mode and then only building out rollup/webpack for babel-standalone when publishing

PR #6845, opened 2017-11-17
[docs] Add package.json method to set preset

bogas04:
Not many people know about it and dislike babel because of "yet-another.rc".

### DIFF
--- a/packages/babel-preset-typescript/README.md
+++ b/packages/babel-preset-typescript/README.md
@@ -40,6 +40,17 @@ npm install --save-dev @babel/preset-typescript
 }
 ```
 
+### Via `package.json`
+```json
+{
+  "version": "0.0.1",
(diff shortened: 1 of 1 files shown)


xtuc:
Thanks for your PR. I think that we should keep things consistent; the other READMEs doesn't show the package.json configuration, so I think we shouldn't add it for this.

We have a section in our documentation showing how you can configure Babel using the package.json.

bogas04:
Oh, that'll be a big change to edit every other plugin/presets. 

Sure, you may close it if you want, just thought advertising package.json would encourage people to use it.

xtuc:
Yes, we could also figure something out to automatically generate them, but it will add complexity on our side (and future contribs).

I'll go ahead and close this. Thanks for having take the time to open a PR.

PR #6850, opened 2017-11-18
 Copy getters and setters correctly in interopWildcard

### LINKED ISSUE #4290: Importing `crypto` from babel throws deprecations warnings because of interopRequireWildcard (T7354)
> Issue originally made by @s-panferov
### Bug information
- **Babel version:** any
- **Node version:** >= 4
### Input code

``` js
import * as crypto from 'crypto';
```
### Description

`_interopRequireWildcard` iterates through export object properties and this fires deprecation warning because of deprecated property access.

danez:
| Fixed Issues?            | Fixes #4290 Fixes #5790
| Major: Breaking Change?  | n?
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

This ensures that getters and setters are copied correctly in the `interopRequireWildcard` helper.

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -450,7 +450,9 @@ helpers.interopRequireWildcard = defineHelper(`
       if (obj != null) {
         for (var key in obj) {
           if (Object.prototype.hasOwnProperty.call(obj, key)) {
-            var desc = Object.getOwnPropertyDescriptor(obj, key);
+            var desc = Object.defineProperty && Object.getOwnPropertyDescriptor
+              ? Object.getOwnPropertyDescriptor(obj, key)
+              : {};
             if (desc.get || desc.set) {
               Object.defineProperty(newObj, key, desc);
             } else {
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -449,7 +449,14 @@ helpers.interopRequireWildcard = defineHelper(`
       var newObj = {};
       if (obj != null) {
         for (var key in obj) {
-          if (Object.prototype.hasOwnProperty.call(obj, key)) newObj[key] = obj[key];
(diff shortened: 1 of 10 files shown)


nicolo-ribaudo:
Does this also influence `import *`? If yes, can you add a test for it?

e.g.
```js
import * as foo from "./foo";
assert.throws(() => foo.bar);
```
```js
// foo.js
module.exports = {
  get bar() { throw {} }
};
```

danez:
Good idea, I added it.

packages/babel-helpers/src/helpers.js
```diff
@@ -449,7 +449,14 @@ helpers.interopRequireWildcard = defineHelper(`
       var newObj = {};
       if (obj != null) {
         for (var key in obj) {
-          if (Object.prototype.hasOwnProperty.call(obj, key)) newObj[key] = obj[key];
+          if (Object.prototype.hasOwnProperty.call(obj, key)) {
+            var desc = Object.getOwnPropertyDescriptor(obj, key);
+            if (desc.get || desc.set) {
+              Object.defineProperty(newObj, key, desc);
```
loganfsmyth: I think if we're going to do this, we should check for the existence of `Object.defineProperty`, or we should finally expose a flag for modules to opt in or out of `Object.defineProperty` usage. Otherwise this is going to make things unusable for people targeting old environments.

danez: I will add a check for `Object.defineProperty`, which I think should be enough, because when `Object.defineProperty` is not there then getters and setters are also not supported. 

It is basically supported in all environments besides IE8 and older. But IE8 is also over 10years old now. 🕵️‍♀️
What if we change in babel 7 and state that generated code from babel needs at least ES5, like react does now (they even require Map and Set). Would be nice to have statistics to see what environments people support, but I guess that does not really exist. Or did we have already complains about this?

danez: And core-js also has a polyfill for it, so we could also link people there.

PR #5930, opened 2017-07-08
Spec compliancy of check-es2015-constants plugin

maurobringolf:
| Q                        | A 
| ------------------------ | ---
| Patch: Bug Fix?          | No
| Major: Breaking Change?  | No
| Minor: New Feature?      | No
| Deprecations?            | No
| Spec Compliancy?         | Yes
| Tests Added/Pass?        | Yes
| Fixed Tickets            | Fixes #5728 
| License                  | MIT
| Doc PR                   | 
| Dependency Changes       | 

I started working on the proposed change to the check-es2015-constants plugin: Instead of throwing a compile time error when `const` is violated, Babel should insert a throw statement before the violation. This will make it a runtime error and violations in code that is not executed are fine. I added the example from the issue as a test case and changed existing test cases.

I am not sure if this covers all possible violations besides modules. Is there a list of all possible violation types somewhere? Variable declarations via module import statements are not yet covered by this PR. Let me know if this is going into the right direction or if I should take a different route.

PR #6855, opened 2017-11-19
Implement TDZ for ClassFieldDefinitionEvaluation

### LINKED ISSUE #6732: Class variable should be under TDZ during ClassFieldDefinitionEvaluation
> Choose one: is this a bug report or feature request?

bug report

### Input Code

```js
class C {
  static [C] = new C;
}
```

stage2,stage3

### Expected Behavior

ReferenceError

### Current Behavior

no error

### Your Environment

[REPL](https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAEDCDeAoa0IBcxoJbGgbVgF1oBeaAOwFMB3OAbkQF8g&debug=false&circleciRepo=&evaluate=true&lineWrap=false&presets=stage-2%2Cstage-3&targets=&version=7.0.0-beta.5)

maurobringolf:
| Fixed Issues?            | Fixes #6732 
| Documentation PR         | 
| Any Dependency Changes?  |

My approach:
1. Additional subtraversal of property name computation for computed class properties
1. Find all identifiers referencing the class binding
1. Insert runtime errors using the same approach as with `const` checks (https://github.com/babel/babel/pull/5930).

This shares the helper function `statementBeforeExpression` with the mentioned PR and should probably be moved into some place where both plugins can use it.

### DIFF
--- a/packages/babel-plugin-proposal-class-properties/src/index.js
+++ b/packages/babel-plugin-proposal-class-properties/src/index.js
@@ -25,6 +25,26 @@ export default function(api, options) {
     },
   };
 
+  const ClassFieldDefinitionEvaluationTDZVisitor = {
+    Expression(path) {
+      if (path === this.shouldSkip) {
+        path.skip();
+      }
+    },
+
+    ReferencedIdentifier(path) {
+      if (this.classRef === path.scope.getBinding(path.node.name)) {
+        const classNameTDZError = this.file.addHelper("classNameTDZError");
+        const throwNode = t.callExpression(classNameTDZError, [
+          t.stringLiteral(path.node.name),
+        ]);
+
+        path.replaceWith(t.sequenceExpression([throwNode, path.node]));
+        path.skip();
+      }
+    },
+  };
+
   const buildClassPropertySpec = (ref, { key, value, computed }, scope) => {
     return template.statement`
       Object.defineProperty(REF, KEY, {
@@ -55,7 +75,7 @@ export default function(api, options) {
     inherits: syntaxClassProperties,
 
     visitor: {
-      Class(path, state) {
+      Class(path) {
         const isDerived = !!path.node.superClass;
         let constructor;
         const props = [];
@@ -90,37 +110,16 @@ export default function(api, options) {
         const staticNodes = [];
         let instanceBody = [];
 
-        const ClassFieldDefinitionEvaluationTDZ = (classRef, state) => ({
-          Expression(path) {
-            if (path.parentKey === "value") {
-              path.skip();
-            }
-          },
-
-          ReferencedIdentifier(path) {
-            if (classRef === path.scope.getBinding(path.node.name)) {
-              const classNameTDZError = state.addHelper("classNameTDZError");
-              const throwNode = t.callExpression(classNameTDZError, [
-                t.stringLiteral(path.node.name),
-              ]);
-
-              path.replaceWith(t.sequenceExpression([throwNode, path.node]));
-              path.skip();
-            }
-          },
-        });
-
         for (const computedPath of computedPaths) {
           const computedNode = computedPath.node;
           // Make sure computed property names are only evaluated once (upon class definition)
           // and in the right order in combination with static properties
           if (!computedPath.get("key").isConstantExpression()) {
-            computedPath.traverse(
-              ClassFieldDefinitionEvaluationTDZ(
-                path.scope.getBinding(ref.name),
-                state,
-              ),
-            );
+            computedPath.traverse(ClassFieldDefinitionEvaluationTDZVisitor, {
+              classRef: path.scope.getBinding(ref.name),
+              file: this.file,
+              shouldSkip: computedPath.get("value"),
+            });
             const ident = path.scope.generateUidIdentifierBasedOnNode(
               computedNode.key,
             );
--- a/packages/babel-plugin-proposal-class-properties/src/index.js
+++ b/packages/babel-plugin-proposal-class-properties/src/index.js
@@ -90,11 +90,53 @@ export default function(api, options) {
         const staticNodes = [];
         let instanceBody = [];
 
+        /**
+         * Helper function to run a statement before an expression by replacing it with a comma expression
+         * and wrapping the statement in an IIFE as the first operand.
+         */
+        function statementBeforeExpression(statement, expression) {
+          return t.sequenceExpression([
+            t.callExpression(
+              t.functionExpression(null, [], t.blockStatement([statement])),
+              [],
+            ),
+            expression,
+          ]);
+        }
+
+        const ClassFieldDefinitionEvaluationTDZ = classRef => ({
+          Expression(path) {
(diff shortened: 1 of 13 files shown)


Andarist:
Cool work, could we make the error functions a helper like discussed in https://github.com/babel/babel/issues/6857#issuecomment-345659780 ? Would produce more readable code than IIFEs

maurobringolf:
I looked at the existing TDZ helper and decided not to use it, since it is not fully appropriate. Instead I added a separate helper for this type of error. This is very similar to the one for `check-constants`, it only differs in the type of error and message thrown. I guess this could be merged into a general `throwRuntimeError()` helper to reduce the amount of inserted code?

maurobringolf:
Haha @xtuc we could include this in a subtle error message along the lines of "What on earth are you trying to do?" 😛

maurobringolf:
I think Travis is failing because of a `let` in Node 4 inserted by some helper. How do I fix this? I tried using the same `"plugins"` list as the non-tdz case which did not change the output code.

nicolo-ribaudo:
@maurobringolf You can use the `minNodeVersion` option (like in https://github.com/babel/babel/blob/0189b387026c35472dccf45d14d58312d249f799/packages/babel-plugin-transform-new-target/test/fixtures/exec/reflect-class/options.json)

maurobringolf:
Thank you @nicolo-ribaudo

xtuc:
Apart of the conflict, it lgtm. Feel free to merge it.

Thanks for your pr :+1:

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -90,11 +90,37 @@ export default function(api, options) {
         const staticNodes = [];
         let instanceBody = [];
 
+        const ClassFieldDefinitionEvaluationTDZ = (classRef, state) => ({
```
nicolo-ribaudo: This visitor can be moved to the top of the plugin factory function (alongside with [`bareSupers`](https://github.com/maurobringolf/babel/blob/9488e10716ae7f4bec5c7fc51f80bf1ad6872530/packages/babel-plugin-proposal-class-properties/src/index.js#L8) and [`referenceVisitor`](https://github.com/maurobringolf/babel/blob/9488e10716ae7f4bec5c7fc51f80bf1ad6872530/packages/babel-plugin-proposal-class-properties/src/index.js#L16)), so we avoid redefining this function for every class and calling it for every static computed property. You can pass `classRef` and `addHelper` using the state. e.g.

```diff
           ReferencedIdentifier(path) {
-            if (classRef === path.scope.getBinding(path.node.name)) {
-              const classNameTDZError = state.addHelper("classNameTDZError");
+            if (this.classRef === path.scope.getBinding(path.node.name)) {
+              const classNameTDZError = this.file.addHelper(
+                "classNameTDZError",
+              );
               const throwNode = t.callExpression(classNameTDZError, [
```
```diff
           if (!computedPath.get("key").isConstantExpression()) {
-            computedPath.traverse(
-              ClassFieldDefinitionEvaluationTDZ(
-                path.scope.getBinding(ref.name),
-                state,
-              ),
-            );
+            computedPath.traverse(classFieldDefinitionEvaluationTDZVisitor, {
+              classRef: path.scope.getBinding(ref.name),
+              file: this.file,
+            });
             const ident = path.scope.generateUidIdentifierBasedOnNode(
```

nicolo-ribaudo: This visitor can be moved to the top of the plugin factory function (alongside with [`bareSupers`](https://github.com/maurobringolf/babel/blob/9488e10716ae7f4bec5c7fc51f80bf1ad6872530/packages/babel-plugin-proposal-class-properties/src/index.js#L8) and [`referenceVisitor`](https://github.com/maurobringolf/babel/blob/9488e10716ae7f4bec5c7fc51f80bf1ad6872530/packages/babel-plugin-proposal-class-properties/src/index.js#L16)), so we avoid redefining this function for every class and calling it for every static computed property. You can pass `classRef` and `addHelper` using the state. e.g.

```diff
           ReferencedIdentifier(path) {
-            if (classRef === path.scope.getBinding(path.node.name)) {
-              const classNameTDZError = state.addHelper("classNameTDZError");
+            if (this.classRef === path.scope.getBinding(path.node.name)) {
+              const classNameTDZError = this.file.addHelper(
+                "classNameTDZError",
+              );
               const throwNode = t.callExpression(classNameTDZError, [
```
```diff
           if (!computedPath.get("key").isConstantExpression()) {
-            computedPath.traverse(
-              ClassFieldDefinitionEvaluationTDZ(
-                path.scope.getBinding(ref.name),
-                state,
-              ),
-            );
+            computedPath.traverse(classFieldDefinitionEvaluationTDZVisitor, {
+              classRef: path.scope.getBinding(ref.name),
+              file: this.file,
+            });
             const ident = path.scope.generateUidIdentifierBasedOnNode(
```

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -90,11 +90,37 @@ export default function(api, options) {
...
+          Expression(path) {
+            if (path.parentKey === "value") {
+              path.skip();
+            }
```
nicolo-ribaudo: > :warning: Please don't hate me for pointing out this edge case lol

This can cause some problems. e.g.

```js
class A {
  static [{ x: A }.x];
}
```
is (correctly) compiled to
```js
class A {}

var _x$x = {
  x: (_classNameTDZError("A"), A)
}.x;
Object.defineProperty(A, _x$x, {
  configurable: true,
  enumerable: true,
  writable: true,
  value: void 0
});
```

But
```js
class A {
  static [{ x: A || 0 }.x];
}
```

is compiled to (note the missing `_classNameTDZError` call)
```js
class A {}

var _x$x = {
  x: A || 0
}.x;
Object.defineProperty(A, _x$x, {
  configurable: true,
  enumerable: true,
  writable: true,
  value: void 0
});
```

---

I suggest only traversing the computed key (`computedPath.get("key").traverse(` instead of `computedPath.traverse(`) so that this check can be avoided.

maurobringolf: Didnt fully investigate yet, but I tried this first! For some reason that I could not figure out it did not catch the simplest case [C], but only cases where the computed expression was more complex. It seemed like the first node was skipped.

Will follow up as soon as I get to it

xtuc: I agree but we could also do it in another PR because it would probably be simpler to review.

maurobringolf: Wow now these are some **real** edge cases :)

maurobringolf: An alternative would be to make sure that the key is `value`**and** the parentPath is the class, is that right? @nicolo-ribaudo

PR #6859, opened 2017-11-20
Make all log and debug output use stderr

### LINKED ISSUE #6597: debug/warn/error info should go to console.error
> Issue originally reported by @wmertens in babel/babel-preset-env#447

Right now debug info gets into my files when I redirect output for processing. All notices from applications that are not actual output should be done via stderr, and therefore debug should use `console.error`.

wmertens:
| Fixed Issues?            | Fixes #6597
| Major: Breaking Change?  | Yes, debug output goes to `stderr` now so scripts might need adjusting
| Tests Added + Pass?      | Yes (no tests added)

The debug/log output for babel-preset-env and babel-cli now goes to `stderr`.
This is how it works in all Unix utilities and prevents mixing log information with real Babel output

### DIFF
--- a/experimental/babel-preset-env/src/debug.js
+++ b/experimental/babel-preset-env/src/debug.js
@@ -9,7 +9,7 @@ const wordEnds = size => {
 export const logMessage = (message, context) => {
   const pre = context ? `[${context}] ` : "";
   const logStr = `  ${pre}${message}`;
-  console.log(logStr);
+  console.error(logStr);
 };
 
 export const logPlugin = (plugin, targets, list, context) => {
@@ -36,21 +36,21 @@ export const logEntryPolyfills = (
   onDebug,
 ) => {
   if (!importPolyfillIncluded) {
-    console.log(
+    console.error(
       `
 [${filename}] \`import '@babel/polyfill'\` was not found.`,
     );
     return;
   }
   if (!polyfills.size) {
-    console.log(
+    console.error(
       `
 [${filename}] Based on your targets, none were added.`,
     );
     return;
   }
 
-  console.log(
+  console.error(
     `
 [${filename}] Replaced \`@babel/polyfill\` with the following polyfill${wordEnds(
       polyfills.size,
@@ -61,13 +61,13 @@ export const logEntryPolyfills = (
 
 export const logUsagePolyfills = (polyfills, filename, onDebug) => {
   if (!polyfills.size) {
-    console.log(
+    console.error(
       `
 [${filename}] Based on your code and targets, none were added.`,
     );
     return;
   }
-  console.log(
+  console.error(
     `
 [${filename}] Added following polyfill${wordEnds(polyfills.size)}:`,
   );
--- a/packages/babel-cli/src/babel-external-helpers.js
+++ b/packages/babel-cli/src/babel-external-helpers.js
@@ -24,4 +24,4 @@ commander.option(
 commander.usage("[options]");
 commander.parse(process.argv);
 
-console.log(buildExternalHelpers(commander.whitelist, commander.outputType));
+console.error(buildExternalHelpers(commander.whitelist, commander.outputType));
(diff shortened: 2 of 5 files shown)


wmertens:
Hmm looks like the tests are indeed failing because of the stderr output, odd, I thought I ran the tests against my branch.

Any preferences on how to proceed? Either only change `preset-env`, or adjust all the tests to expect output on stderr…

packages/babel-cli/src/babel-external-helpers.js
```diff
@@ -24,4 +24,4 @@ commander.option(
 commander.usage("[options]");
 commander.parse(process.argv);
 
-console.log(buildExternalHelpers(commander.whitelist, commander.outputType));
+console.error(buildExternalHelpers(commander.whitelist, commander.outputType));
```
loganfsmyth: This one shouldn't be `.error` :)

experimental/babel-preset-env/src/debug.js
```diff
@@ -9,7 +9,7 @@ const wordEnds = size => {
 export const logMessage = (message, context) => {
   const pre = context ? `[${context}] ` : "";
   const logStr = `  ${pre}${message}`;
-  console.log(logStr);
+  console.error(logStr);
```
loganfsmyth: Personally it seems like `console.error` implies an error condition to me. Should these be `console.warn` perhaps?

xtuc: I agree, I would even prefer a layer between that, like just `output(msg: string)` which calls the appropriate console method.

wmertens: Hmm, is `output` a good name? The js is also output…

xtuc: Righ, maybe `printDebug(msg)`?

wmertens:
Sounds good, I'll give that a shot

packages/babel-cli/src/babel-external-helpers.js
jasonkarns: Is this one actually emitting code? (I think part of the confusion in the babel codebase is the usage of `console` for emitting the compilation output. Would be clearer if there were a dedicated stream for the actual JS output instead of hijacking `console`)

jasonkarns:
Is there a way to test this with babel 6? I tried cloning and npm-linking this branch, but the `@babel` scope name change is making this difficult.

packages/babel-cli/src/babel-external-helpers.js
loganfsmyth: The `babel-external-helpers` command is specifically building some code and then dumping it to the console for piping, so this change would make the command non-functional.

> Would be clearer if there were a dedicated stream for the actual JS output

I don't think it makes much difference in this case.

loganfsmyth:
@jasonkarns What kinds of issues were you having? Babel 7 packages mostly only work with `@babel/core`, but it should all be `link`-able like usual, you just have to link or install a 7.x core.

packages/babel-cli/src/babel-external-helpers.js
jasonkarns: > I don't think it makes much difference in this case

Really? It would have directly avoided this issue here, where the console.log was erroneously switched to console.error :D (And eliminated the confusion that I personally had about confirming whether this line emitted code or something else.)

wmertens: I shall make `printOutput`, `printInfo` and `printDebug` functions?

xtuc: yeah lgtm

Swaagie:
I don't want to feature creep the PR, but it would be nice to have more granular control over the type of messages that are logged to `stderr`. Right now our CLI developer tooling is guarded against killing the process when a `note` level message is received on `stderr`. Ideally `loglevel` is configurable from one of the 'central' functions that werre proposed, e.g. `printInfo`, `printDebug` etc

loganfsmyth:
I'm triaging old PR on Babel. Since this has needs extensive rebasing, I'm going to close it. Feel free to rebase and open a new PR if you think it is worth pursuing still.

I admit, I do have concerns now that 7.x is stable since logging to `stderr` may cause problems for some CI scripts.

PR #6860, opened 2017-11-20
Enforce Prettier defaults + also format .babelrc and .json

wmertens:
| Tests Added + Pass?      | Yes (no tests added, all pass)
| Documentation PR         |

This is basically a dev-only change.

If you use prettier-on-save and you have different settings from the Babel settings, they will be used instead of the defaults. This PR adds the default values for all Prettier settings so this no longer happens.

I also noticed that the `.babelrc.js` file was being ignored, so I added that to the globs.

Finally, prettier supports formatting `.json` files, so I added that to the `Makefile`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,7 @@ lint:
 	./node_modules/.bin/eslint scripts $(SOURCES) '*.js' '**/.*.js' --format=codeframe --rulesdir="./scripts/eslint_rules"
 
 fix:
+	./node_modules/.bin/prettier --write --ignore-path .eslintignore '**/*.json'
 	./node_modules/.bin/eslint scripts $(SOURCES) '*.js' '**/.*.js' --format=codeframe --fix --rulesdir="./scripts/eslint_rules"
 
 clean: test-clean
--- a/Makefile
+++ b/Makefile
@@ -50,10 +50,10 @@ flow:
 	./node_modules/.bin/flow check --strip-root
 
 lint:
-	./node_modules/.bin/eslint scripts $(SOURCES) *.js --format=codeframe --rulesdir="./scripts/eslint_rules"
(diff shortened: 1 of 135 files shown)


Andarist:
prettier@1.8 has also `*.md` support - would be really useful to format those too

wmertens:
@Andarist I gave it a quick shot, but upgrading to v1.8 actually breaks at least one eslint rule, and several of the `.md` files with iffy formatting lose some of the formatting (the old changelog loses some indentation, the example code in some `.md` files gets compacted into one line because there's no comment to force function bodies to be multi-line etc.

So while I would be ok with adjusting the formatting, I think it would be better to do that change in a different PR…

xtuc:
Good point @hzoo, also modifying every package.json will probably cause us a lot of merge conflicts.

I would prefer to remove it on package.json.

wmertens:
Well, they are also `.json` files and their formatting is mostly up to the dev, yarn only touches the dependency keys. If a PR has a conflict all they need to do is format it with prettier and it's ok, and future PRs will be compliant anyway…

Makefile
```diff
@@ -50,10 +50,11 @@ flow:
 	./node_modules/.bin/flow check --strip-root
 
 lint:
-	./node_modules/.bin/eslint scripts $(SOURCES) *.js --format=codeframe --rulesdir="./scripts/eslint_rules"
+	./node_modules/.bin/eslint scripts $(SOURCES) '*.js' '**/.*.js' --format=codeframe --rulesdir="./scripts/eslint_rules"
 
 fix:
-	./node_modules/.bin/eslint scripts $(SOURCES) *.js --format=codeframe --fix --rulesdir="./scripts/eslint_rules"
+	./node_modules/.bin/prettier --write --ignore-path .eslintignore '**/*.json'
```
danez: eslintrc could also be added right?

danez:
yarn formats the whole file, I constantly have this forth and back in some repos where yarn and prettier disagree.

existentialism:
Yeah, let's skip formatting package.json files.

PR #6862, opened 2017-11-20
Define readOnlyError helper and use in check-constants plugin

### LINKED ISSUE #6857: Optimize const violations?

### Input Code

```js
const t = true

{
  t = false
}

{
  t = false
}
```

```js
{
  presets: ['es2015']
}
```

### Current Behavior

```js
var t = true;
{
  t = (function () {
    throw new Error("\"t\" is read-only");
  }(), false);
}
{
  t = (function () {
    throw new Error("\"t\" is read-only");
  }(), false);
}
```

### Possible Solution

```js
var _tReadonlyError = new Error("\"t\" is read-only");

var t = true;
{
  t = (function () {
    throw _tReadonlyError;
  }(), false);
}
{
  t = (function () {
    throw _tReadonlyError;
  }(), false);
}
```

### Context

It's just an idea. We can hoist the error (or even the function) to avoid duplicating code so much. Minifiers probably won't catch it in more complex condition.

### Your Environment

[REPL](http://babeljs.io/repl#?babili=false&browsers=IE%2011&build=&builtIns=false&code_lz=MYewdgzgLgBLC8cBOBXApgKAwbwzOMiAZgIYA2EmAvlrvgjKRdUA&debug=true&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015&targets=&version=7.0.0-beta.31)

What do you think about that @maurobringolf?

maurobringolf:
| Fixed Issues?            | Fixes #6857  
| Documentation PR         | 
| Any Dependency Changes?  |

This aims to produce better output code for constant violations by replacing the IIFE's with helper calls.

### DIFF
--- a/packages/babel-plugin-check-constants/test/fixtures/general/no-for-in/expected.js
+++ b/packages/babel-plugin-check-constants/test/fixtures/general/no-for-in/expected.js
@@ -1,6 +1,9 @@
+function _readOnlyError(name) { throw new Error("\"" + name + "\" is read-only"); }
+
 var MULTIPLIER = 5;
 
 for (MULTIPLIER in arr) {
-  throw new Error("\"MULTIPLIER\" is read-only");
+  _readOnlyError("MULTIPLIER");
+
   ;
 }
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -627,6 +627,12 @@ helpers.temporalRef = defineHelper(`
   }
 `);
 
+helpers.readOnlyError = defineHelper(`
+  export default function _readOnlyError(name) {
+    throw new Error("\\"" + name + "\\" is read-only");
+  }
+`);
+
 helpers.temporalUndefined = defineHelper(`
   export default {};
 `);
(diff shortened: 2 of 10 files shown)


packages/babel-plugin-check-constants/test/fixtures/general/no-for-in/expected.js
```diff
@@ -1,6 +1,9 @@
+function _readOnlyError(name) { throw new Error("\"" + name + "\" is read-only"); }
+
 var MULTIPLIER = 5;
 
 for (MULTIPLIER in arr) {
-  throw new Error("\"MULTIPLIER\" is read-only");
+  _readOnlyError("MULTIPLIER");
+
```
maurobringolf: @xtuc Any idea why this is happening? Am I combining nodes incorrectly?

xtuc: ¯\_(ツ)_/¯

xtuc:
Just FYI here's an example [REPL](https://babeljs.io/repl/build/5933/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYewdgzgLgBAhjAvDKAnArgUwFAG9szxIoY4C-eBRyaW2ZQA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cstage-0&targets=&version=7.0.0-beta.32%2Bpr.6862)

PR #6863, opened 2017-11-20
Fix `export from` assignment order for loose mode.

### LINKED ISSUE #6805: Issue with loose mode export { x } from 'x';
From https://github.com/babel/babel/pull/6742, tried to upgrade Babel itself

https://github.com/babel/babel/blob/23626fc6ad0cbeb1a6141c2b0c9d6d3f15e478ab/packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop-loose/export-from-2/expected.js

```
exports.foo = _foo.foo;
var _foo = require("foo");
```
since it gets required after @existentialism

yavorsky:
| Fixed Issues?            | Fixes #6805

This PR fixes `exports from` behavior in loose mode.
```js
export {foo} from "foo";
```
*Before:*
```js
exports.__esModule = true;
exports.foo = _foo.foo;

var _foo = require("foo");
```
*After*
```js
exports.__esModule = true;
exports.foo = void 0;

var _foo = require("foo");

exports.foo = _foo.foo;
```

So, it will return all exported keys for `foo.js` and require module before module's property access.

I realized that @daft300punk commented that he is working on it. So, I can close in favor of @daft300punk's PR. 👍 

Also, exports-from is acting like a named exports now, so there are additional `_exports.foo = void 0`. Maybe we should use `bindingKindLookup` only for `module-commonjs` ? Anyway, it could be a separate fix.

### DIFF
--- a/packages/babel-helper-module-transforms/src/index.js
+++ b/packages/babel-helper-module-transforms/src/index.js
@@ -128,6 +128,9 @@ export function buildNamespaceInitStatements(
       }),
     );
   }
+  if (loose) {
+    statements.push(...buildReexportsFromMeta(metadata, sourceMetadata, loose));
+  }
   for (const exportName of sourceMetadata.reexportNamespace) {
     // Assign export to namespace object.
     statements.push(
@@ -279,6 +282,7 @@ function buildExportNameListDeclaration(
 function buildExportInitializationStatements(
   programPath: NodePath,
   metadata: ModuleMetadata,
+  loose: boolean = false,
 ) {
   const initStatements = [];
 
@@ -296,6 +300,9 @@ function buildExportInitializationStatements(
   }
 
   for (const data of metadata.source.values()) {
+    if (!loose) {
+      initStatements.push(...buildReexportsFromMeta(metadata, data, loose));
+    }
     for (const exportName of data.reexportNamespace) {
       exportNames.push(exportName);
     }
--- a/packages/babel-helper-module-transforms/src/index.js
+++ b/packages/babel-helper-module-transforms/src/index.js
@@ -168,7 +168,7 @@ const getTemplateForReexport = loose => {
     `;
 };
 
-export function buildReexportsFromMeta(meta, metadata, loose) {
+const buildReexportsFromMeta = (meta, metadata, loose) => {
   const templateForCurrentMode = getTemplateForReexport(loose);
   return Array.from(metadata.reexports, ([exportName, importName]) =>
     templateForCurrentMode({
@@ -178,7 +178,7 @@ export function buildReexportsFromMeta(meta, metadata, loose) {
       IMPORT_NAME: importName,
     }),
   );
-}
+};
 
 /**
  * Build an "__esModule" header statement setting the property on a given object.
--- a/packages/babel-helper-module-transforms/src/index.js
+++ b/packages/babel-helper-module-transforms/src/index.js
@@ -55,6 +55,7 @@ export function rewriteModuleStatementsAndPrepareHeader(
   }
 
   const nameList = buildExportNameListDeclaration(path, meta);
+
   if (nameList) {
     meta.exportNameListName = nameList.name;
     headers.push(nameList.statement);
@@ -253,7 +254,6 @@ function buildExportNameListDeclaration(
 function buildExportInitializationStatements(
   programPath: NodePath,
   metadata: ModuleMetadata,
-  loose: boolean = false,
 ) {
   const initStatements = [];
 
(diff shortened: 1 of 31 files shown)


daft300punk:
Please go ahead. I haven't had the time to look at it. Thanks.

packages/babel-plugin-transform-modules-amd/src/index.js
```diff
@@ -68,9 +69,17 @@ export default function(api, options) {
...
+            if (!loose) {
+              headers.push(...buildReexportsFromMeta(meta, metadata, loose));
+            }
+
             headers.push(
               ...buildNamespaceInitStatements(meta, metadata, loose),
             );
+
+            if (loose) {
+              headers.push(...buildReexportsFromMeta(meta, metadata, loose));
```
loganfsmyth: I've lost a bit of context on all this, but couldn't we include this in `buildNamespaceInitStatements` when in loose mode, and leave the old logic where it was? I might also just be forgetting stuff, so either way no worries :P

packages/babel-helper-module-transforms/src/index.js
```diff
@@ -150,6 +155,30 @@ export function buildNamespaceInitStatements(
   return statements;
 }
 
+export function buildReexportsFromMeta(meta, metadata, loose) {
+  const reexports = [];
+  for (const [exportName, importName] of metadata.reexports) {
```
Andarist: wouldnt a `.map` be better describing the intent here?

yavorsky: @Andarist Sure! Thanks

yavorsky: @Andarist I've checked.. metadata.reexports is a Map. I think, it's better to use `for..of` instead of `Array.from` because of performance and consistency.

Kovensky: Not sure... a `return Array.from(metadata.reexports, ([exportName, importName]) => (loose ? ... : ...))` looks better to me, but I just overall prefer functional style 🤔

packages/babel-helper-module-transforms/src/index.js
```diff
@@ -150,6 +155,30 @@ export function buildNamespaceInitStatements(
...
+    reexports.push(
+      (loose
+        ? template.statement`EXPORTS.EXPORT_NAME = NAMESPACE.IMPORT_NAME;`
+        : template`
```
Kovensky: I suspect hoisting the `template.statement` / `template` calls outside of the loop will do a lot more for performance than changing between `Array.from` or `for of`

packages/babel-helper-module-transforms/src/index.js
```diff
@@ -155,28 +155,29 @@ export function buildNamespaceInitStatements(
...
-        EXPORTS: meta.exportName,
-        EXPORT_NAME: exportName,
-        NAMESPACE: metadata.name,
-        IMPORT_NAME: importName,
-      }),
-    );
-  }
-  return reexports;
+  const templateForCurrentMode = getTemplateForReexport(loose);
+  return Array.from(metadata.reexports).map(([exportName, importName]) =>
```
Kovensky: The `.map` is redundant; use the second argument of `Array.from`: `Array.from(metadata.reexports, ([exportName, importName]) =>`

yavorsky: Really, thanks :)

packages/babel-helper-module-transforms/src/index.js
```diff
@@ -150,6 +155,31 @@ export function buildNamespaceInitStatements(
...
+        enumerable: true,
+        get: function() {
+          return NAMESPACE.IMPORT_NAME;
+        },
+      });
+    `;
+};
+
+export function buildReexportsFromMeta(meta, metadata, loose) {
+  const templateForCurrentMode = getTemplateForReexport(loose);
```
loganfsmyth: I disagree with @Kovensky here :) I don't think moving this out of the loop matters at all, because there's caching happening here automatically.

packages/babel-helper-module-transforms/src/index.js
```diff
@@ -150,6 +155,31 @@ export function buildNamespaceInitStatements(
...
+      Object.defineProperty(EXPORTS, "EXPORT_NAME", {
+        enumerable: true,
+        get: function() {
+          return NAMESPACE.IMPORT_NAME;
+        },
+      });
+    `;
+};
+
+export function buildReexportsFromMeta(meta, metadata, loose) {
```
loganfsmyth: No `export` needed now, right?

PR #6865, opened 2017-11-21
Remove the experimental folder

hzoo:
- Move `experimental` -> `packages`
- Searched for `experimental` in the codebase

Still need to do https://github.com/lerna/lerna/issues/1121, but yes experimental wasn't the best name and I thought it was a good idea earlier but just causes more issues. Should just make lerna do it correctly without a folder structure.


rwjblue:
❤️

hzoo:
Landed in https://github.com/babel/babel/commit/70361f12005e9ce0f9b961097006ff89f07c34a7

PR #6866, opened 2017-11-21
Add `.github` and `.idea` to `.npmignore`

aaharu:

babel-preset-env@1.6.1 includes unnecessary files.


yavorsky:
I think we should unify npmignore in the future and use editor stuff like .vscode, .idea for all packages.

Andarist:
Imho having `files` in `package.json` is way more maintainable, maybe we could switch to using it instead?

PR #6881, opened 2017-11-22
Remove double exports of NodePath, Scope and Hub in traverse

danez:
| Major: Breaking Change?  | it reverts one
| Documentation PR         | 
| Any Dependency Changes?  |

Works toward the goal of getting all packages be 100% esmodules.

Babel 7 introduced a breaking change when we removed the `module.exports = exports.default` transform.
Right now accessing `NodePath`, `Scope` and `Hub` can be done in two ways:

```js
import { NodePath } from "@babel/traverse";
new NodePath();

// or
 
import traverse from "@babel/traverse";
const NodePath = traverse.NodePath.default;
new NodePath();
```

The second one makes clearly no sense to me, so at first I wanted it to change to be again like in babel 6
```js
import traverse from "@babel/traverse";
const NodePath = traverse.NodePath;
new NodePath();
```

But the more i thought about it it seems we should only provide one way to access stuff and remove `NodePath`, `Hub` and `Scope` from `traverse` and only export them as named exports.

Would you be okay with that?

### DIFF
--- a/packages/babel-traverse/src/index.js
+++ b/packages/babel-traverse/src/index.js
@@ -7,6 +7,7 @@ import * as cache from "./cache";
 export { default as NodePath } from "./path";
 export { default as Scope } from "./scope";
 export { default as Hub } from "./hub";
+
 export { visitors };
 
 export default function traverse(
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
+1

I actually tought that `import { NodePath } from "@babel/traverse";` was the only way to import it!

loganfsmyth:
The benefit that comes to mind is for users doing

```
export default function({ traverse }){
  traverse.NodePath
}
```
but I have no sense of it people would actually do that?

I considered at one point making proxy files in `babel-core` so you could do
```
import { NodePath } from "@babel/core/traverse";
```
without having to add a dependency on `babel-traverse` yourself (since there's no guarantee it'd be the same version).

danez:
I have never come across a plugin though that needs to create a NodePath itself. But also not sure if someone might use it.
So should I change it and rather add it correctly?

PR #6890, opened 2017-11-24
Keep user options intact in transformFile

### LINKED ISSUE #6885: transformFile generates incorrect AMD ids
I am not sure the following behavior is intended or not, hence this issue is more a question than a bug report. 

When using babel-core `transformFile` to perform async code conversion from es modules to AMD modules (with the `babel-plugin-transform-es2015-modules-amd` plugin), the generated AMD module ID is incorrect.
Let's take the following exemple:

### Input Code

```js
const { promisify } = require('util');
const babel = require('babel-core');
const transformFile = promisify(babel.transformFile);

const config = {
    presets: 'env',
    plugins: ['transform-es2015-modules-amd'],
    moduleRoot: 'ROOT',
    moduleIds: true,
    getModuleId (moduleName) {
        return moduleName;
    }
};

async function transpile () {
    const files = [
        './src/a.js',
        './src/b.js',
        './src/c.js'
    ]
    return Promise.all(files.map(async file => {
        let { code } = await transformFile(file, config);
        console.log(code);
        return Promise.resolve(code);
    }));
}
transpile();
```

### Current Behavior

The previous code outputs : 

```js
define("ROOT/./src/c", [], function () {
  "use strict";
});
define("ROOT/./src/c", [], function () {
  "use strict";
});
define("ROOT/./src/c", [], function () {
  "use strict";
});
```

### Expected Behavior

where you would expect it to be:

```js
define("ROOT/./src/a", [], function () {
  "use strict";
});
define("ROOT/./src/b", [], function () {
  "use strict";
});
define("ROOT/./src/c", [], function () {
  "use strict";
});
```

### Possible Solution

I've found that `transformFile` stores the file name inside the configuration object by modifying the reference of the user provided object.
Since this bit is done synchronously the configuration object will have the last file name (in our case './src/c.js'.

I would like to know if the "cloning" of the config is to be done by the user or by `transformFile`: that is to say turn my code from what is was to something like:

```js
async function transpile () {
    const files = [
        './src/a.js',
        './src/b.js',
        './src/c.js'
    ]
    return Promise.all(files.map(async file => {
        let { code } = await transformFile(file, {...config});
        console.log(code);
        return Promise.resolve(code);
    }));
}
```

or turn [this line](https://github.com/babel/babel/blob/master/packages/babel-core/src/transform-file.js#L21) into something like `let options = Object.assign({}, opts, { filename })`. `options` being used to keep the user provided object intact.

### Your Environment

| ---------------- | -------
| Babel            |  6.26.0
| Babylon          | 
| node             | 8.9.1
| npm              | 5.5.1
| Operating System | MacOS 10.12.6

Thank you 😸

nico29:
| Fixed Issues?            | Fixes #6885 
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

As stated in #6885 I changed the user options handling in `transformFile` and `transformFileSync` to prevent the `filename` key from altering the user provided object.

(For the tests,  es2015/regression/6864 is failing, but was failing before any modifications)

### DIFF
--- a/packages/babel-core/test/api.js
+++ b/packages/babel-core/test/api.js
@@ -121,6 +121,7 @@ describe("api", function() {
     const options = {
       babelrc: false,
     };
+    Object.freeze(options);
     babel.transformFile(__dirname + "/fixtures/api/file.js", options, function(
       err,
       res,
@@ -128,7 +129,7 @@ describe("api", function() {
       if (err) return done(err);
       assert.equal(res.code, "foo();");
       // keep user options untouched
-      assert.equal(Object.keys(options).length, 1);
+      assert.deepEqual(options, { babelrc: false });
       done();
     });
   });
@@ -137,12 +138,13 @@ describe("api", function() {
     const options = {
       babelrc: false,
     };
+    Object.freeze(options);
     assert.equal(
       babel.transformFileSync(__dirname + "/fixtures/api/file.js", options)
         .code,
       "foo();",
     );
-    assert.equal(Object.keys(options).length, 1);
+    assert.deepEqual(options, { babelrc: false });
   });
 
   it("options throw on falsy true", function() {
--- a/packages/babel-core/test/api.js
+++ b/packages/babel-core/test/api.js
@@ -118,26 +118,31 @@ describe("api", function() {
   });
 
   it("transformFile", function(done) {
-    babel.transformFile(
-      __dirname + "/fixtures/api/file.js",
-      {
-        babelrc: false,
-      },
-      function(err, res) {
-        if (err) return done(err);
-        assert.equal(res.code, "foo();");
-        done();
-      },
-    );
+    const options = {
+      babelrc: false,
+    };
+    babel.transformFile(__dirname + "/fixtures/api/file.js", options, function(
+      err,
+      res,
+    ) {
+      if (err) return done(err);
+      assert.equal(res.code, "foo();");
+      // keep user options untouched
+      assert.equal(Object.keys(options).length, 1);
+      done();
+    });
   });
 
   it("transformFileSync", function() {
+    const options = {
+      babelrc: false,
+    };
     assert.equal(
-      babel.transformFileSync(__dirname + "/fixtures/api/file.js", {
-        babelrc: false,
-      }).code,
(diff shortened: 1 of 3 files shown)


packages/babel-core/test/api.js
```diff
@@ -118,26 +118,31 @@ describe("api", function() {
...
-      },
-      function(err, res) {
-        if (err) return done(err);
-        assert.equal(res.code, "foo();");
-        done();
-      },
-    );
+    const options = {
+      babelrc: false,
+    };
```
nicolo-ribaudo: Can we `Object.freeze` this object? So it will throw when it is modified.

packages/babel-core/test/api.js
```diff
@@ -118,26 +118,31 @@ describe("api", function() {
...
+      // keep user options untouched
+      assert.equal(Object.keys(options).length, 1);
+      done();
+    });
   });
 
   it("transformFileSync", function() {
+    const options = {
+      babelrc: false,
+    };
```
nicolo-ribaudo: Also this one.

packages/babel-core/test/api.js
```diff
@@ -118,26 +118,31 @@ describe("api", function() {
...
+    babel.transformFile(__dirname + "/fixtures/api/file.js", options, function(
+      err,
+      res,
+    ) {
+      if (err) return done(err);
+      assert.equal(res.code, "foo();");
+      // keep user options untouched
+      assert.equal(Object.keys(options).length, 1);
```
nicolo-ribaudo: This isn't enough, since if a property value is modified it won't be caught.
```js
assert.deepEqual(options, { babelrc: false });
```

nico29: Yes you are right, I missed that.

nico29:
@nicolo-ribaudo I've adapted the code as you suggested.

Arfey:
fetch master and try again

nico29:
@Arfey done, thanks

loganfsmyth:
Perfect, thanks!

nico29:
Awesome thank you all 🙌

PR #6902, opened 2017-11-24
Fixes #6870

### LINKED ISSUE #6870: [bug] babel-node uses polyfills of JSON.stringify with ES6 Set support
### Input Code
babel-node uses different (maybe from polyfil) impl of JSON.stringify with support for ES6 Sets. 
This looks nice first, but is actually wrong and won't work on the server.

```js
console.log(JSON.stringify({foo: new Set(['a', 'b'])}));
```

```js
{
  "presets": [
	[
	  "env",
	  {
		"targets": {
		  "node": 8
		}
	  }
	]
  ],
}
```

### Expected Behavior
When running the above code in plain node or in a browser it prints:
```
{"foo":{}}
```
### Current Behavior
Execution with babel-node gives:
```
{"foo":["a","b"]}
```
This looks good first locally, but results on the server in very subtle bugs as:

```
JSON.stringify({foo:new Set([1,2,3])}) === JSON.stringify({foo:new Set([4,5,6])}) 
```
is `true` on the server and `false` locally.

### Possible Solution
It is actually wrong to use JSON.stringify on ES6 Sets/Maps, so do not add such a feature in babel.

### Context
Using babel-node for local development and transpile with babel to run server with nodejs 8

### Your Environment
Node 8.9.0
Babel 6.5 / 7-beta (both versions are affected)

rohozhnikoff:
| Fixed Issues?            | Fixes #6870 
| Patch: Bug Fix?          | true
| Major: Breaking Change?  | false
| Minor: New Feature?      | false
| Documentation PR         | 
| Any Dependency Changes?  | true

`babel-node/src/_babel-node.js` require `@babel/polyfill` on top of the file for unknown reason, cause code don't use any of polyfill-ed methods or objects
I've removed `require` and dependency, rebuild _babel-node, and run it in node@4
- in cli mode
- in file mode
- `TEST_ONLY=babel-node make test`

everything passed correctly

### DIFF
--- a/packages/babel-node/package.json
+++ b/packages/babel-node/package.json
@@ -16,7 +16,6 @@
     "compiler"
   ],
   "dependencies": {
-    "@babel/polyfill": "7.0.0-beta.32",
     "@babel/register": "7.0.0-beta.32",
     "commander": "^2.8.1",
     "fs-readdir-recursive": "^1.0.0",
--- a/packages/babel-node/src/_babel-node.js
+++ b/packages/babel-node/src/_babel-node.js
@@ -5,9 +5,7 @@ import path from "path";
 import repl from "repl";
 import * as babel from "@babel/core";
 import vm from "vm";
-import "@babel/polyfill";
 import register from "@babel/register";
-
 import pkg from "../package.json";
 
 const program = new commander.Command("babel-node");


hzoo:
>  require @babel/polyfill on top of the file for unknown reason, 

It's not an unknown reason, we supported node 0.10/0.12 and needed certain methods polyfilled. Just removing it will also break if you use regenerator or other polyfills

rohozhnikoff:
sure about 0.10/0.12
i meant, that this change is for babel@7
what branch i should use for it?

nicolo-ribaudo:
This issue should be resolved by fixing the polyfill (core-js), not Babel.

hzoo:
We could make a breaking that runs babel preset env on babel-node with node current as the setting for this

loganfsmyth:
We _could_ do this, but I'm not really sure why we'd care. What's the bug this is actually trying to address? `babel-node` is only really targeted for running simple local scripts, we never recommend it for production.

danez:
I also think removing the polyfill here is not a good option. The polyfill is needed in case we run scripts that use newer features which sometimes even node 4 does not yet support like `Array.prototype.values()` for example.

hzoo:
Oh I wasn't thinking we should remove the polyfill either, just talking about alternatives.

It is completely intentional that the polyfill is required in babel-node because it's supposed to be a simple way to run babel without adding the polyfill yourself. It just happens that it imports polyfills that aren't necessary in a later version of node. to fix that my proposal is just to only import the polyfills based on the current version of node which babel-preset-env can already do?

hzoo:
Ok going to close and say we would accept a PR that adds a piece of code that will only add the polyfills necesary for the current version of node being run. Not sure how do to this other than running the preset-env transform and injecting those polyfills.

PR #6905, opened 2017-11-25
Merge all config & programmatic plugins/preset rather than duplicating

### LINKED ISSUE #6765: Allow specifying "mergeMode" for config merging?
We've got https://github.com/babel/babel/issues/5276 filed, but the main reason for removing `env` is that users simply don't understand how the merge logic works. (And I don't blame them, _I_ barely understand it, and might not be able to describe it to you.

That said, there are other places where we merge, so we should still figure out a better story around how configurations get merged together. Primary cases where we merge now:

* Config files can `extends: "./other-config"` and things get merged
* Programmatic options to `babel.transform` are merged with `.babelrc` options
* "env" blocks are merged with their parent configs
* If we do https://github.com/babel/babel/issues/5451, we need a clear story around merging for specific files.

loganfsmyth:
| Fixed Issues?            | Fixes #6765
| Documentation PR         | 
| Any Dependency Changes?  |

In an effort to make config a bit closer to what people generally _think_ Babel already does, this makes Babel's config-level options merge. That means everything passed in programmatically, and loaded via a `.babelrc` or a referenced `.extends` block. Plugins from a config will _not_ merge with plugins inside of a preset itself, since names and values of plugins inside presets are essentially private implementation defined data.

This means
```
{
  presets: [
    ['env', {modules: false}],
  ],
  env: {
    node: {
      presets: [
        ['env', {}],
      ]
    },
  },
}
```
for example would use `{modules: false}` by default, but use `{}` as the options when the `env` has been set to `node` for instance. I've also made it so that passing `false` as the options value for a preset or plugin will disable that item entirely.

I should also mention that things merge based on the identity of the plugin function itself. That means that

```
presets: [
  '@babel/preset-env',
  require('@babel/preset-env').default,
]
```
would throw an error for the user about duplicates, and if one was in an `env` or `extends`-file, they would merge happily. It seemed like the safest way to ensure that someone didn't refactor and break their config or something.

Importantly the feature that is no longer possible with this is using the same plugin _twice_. To allow for that, this PR includes logic to allow users to _name_ specific plugins to uniquely identify them by expanding the 2-tuple to an optional 3-tuple, as

```
plugins: [
  ['module-resolver', {from: 'view', to: './view' }, "resolve-view"],
  ['module-resolver', {from: 'model', to: './model' }, "resolve-model"],
]
```
which would then not override eachother since they have different names.

Not super pretty, but also probably not super _common_ so potentially not a real issue. Definitely curious for thoughts on that one.

##### Plugin/preset evaluation ordering

The obvious breaking change is obviously that things won't be duplicated anymore, but the potentially less obvious point is that things potentially change order. That said, I think the order is _better_ this way. The merging happens up front for both plugins and presets, so where before each nested config was essentially merged into the plugin list in order, now they merge first then execute. So for example

```
presets: [
  'pre',
],
plugins: [
  'plug',
],
env: {
  thing: {
    presets: [
      'pre-env',
    ],
    plugins: [
      'plug-env',
    ],
  }
}
```
before would have been

* process `pre`
* process `plug`
* process `pre-env`
* process `plug-env`

in essentially increasing order, but now they merge first, so it'll be

* process `pre`
* process `pre-env`
* process `plug`
* process `plug-env`

which to me is what people expect more anyway, but it is what caused most the the unittest changes that you'll see with this PR.

The priority of a given config object goes in increasing order of

* values from `.extends` files, recursively applying the rules here for files
* the values at the root of the
* the values from a `.env` block

and values passed in via the programmatic options will merge with a higher priority, overriding values from config files.

##### Preset options

Presets can provide options the same way config files can, but it isn't something that is done super often. That said, it complicates merging because for instance it isn't obvious what priority the preset config values would have compared to the user's config files. This PR officially changes the semantics of preset options to be _default_ values, so user config will automatically override any default value that a preset has specified.


packages/babel-core/src/config/option-manager.js
```diff
@@ -172,25 +205,133 @@ const loadConfig = makeWeakCache((config: MergeOptions): {
...
+        desc.value = null;
+      } else {
+        desc.value = item;
+      }
+    } else {
+      descriptors.push({ value: item });
+    }
+  }
+
+  return descriptors.reduce((acc, desc) => {
```
Andarist: wouldnt this be clearer as `filter`? would have to be filter + map though in this case, cant we push to `descriptors` only valid entries? or are descriptors' items being mutated when already in the array?

packages/babel-core/test/option-manager.js
```diff
@@ -11,6 +11,113 @@ describe("option-manager", () => {
...
+        });
+      }, /Duplicate plugin\/preset detected/);
+      assert.deepEqual(calls, []);
+    });
+
+    it("should not throw if a repeated plugin has a different name", () => {
+      const { calls: calls1, plugin: plugin1 } = makePlugin();
+      const { calls: calls2, plugin: plugin2 } = makePlugin();
+
+      manageOptions({
```
Andarist: isnt this a private API? maybe we could try to express the test using public API exclusively

packages/babel-core/src/config/option-manager.js
```diff
@@ -327,10 +469,16 @@ function createDescriptor(
     ownPass?: boolean,
   },
 ): BasicDescriptor {
+  let name;
   let options;
   let value = pair;
   if (Array.isArray(value)) {
-    [value, options] = value;
+    if (value.length === 3) {
+      // $FlowIgnore - Flow doesn't like the multiple tuple types.
```
Andarist: is the length check necessary here? you could just unpack the tuple here as is and the `name` would simply stay `undefined`, although we shouldnt access out of bounds elements, so maybe it would be reasonable to have only 1 tuple type for this, where the last element would be a maybe type AND it would require explicit third element of null/undefined

Andarist:
> To allow for that, this PR includes logic to allow users to name specific plugins to uniquely identify them by expanding the 2-tuple to an optional 3-tuple, as

I don't know many use cases when one would want multiple instances of the same plugin, I think most use cases should be covered by plugin's options, but possibly there are some cases when that's desirable and the proposed solution seems clean to me.

Personally I like the proposed behaviour, seems not complicated - both for users and the codebase. So 2 👍 from me :)

packages/babel-core/src/config/option-manager.js
loganfsmyth: Yeah I don't feel strongly. I usually just opt for `reduce` once anything gets complicated. I can change to map + filter too though.

packages/babel-core/src/config/option-manager.js
loganfsmyth: > where the last element would be a maybe type AND it would require explicit third element of null/undefined

packages/babel-core/test/option-manager.js
loganfsmyth: Yeah I can rename this. This is the same public API you can get from `babel.loadOptions` so I think I can change these tests to use that.

packages/babel-core/src/config/option-manager.js
Andarist: Hm, maybe you could just normalize the user's input as soon as possible to use consistent type in the core and avoid `$FlowIgnore`?

packages/babel-core/src/config/option-manager.js
Andarist: Dont feeling strongly about this either, just leave it in whatever shape you want :)

packages/babel-core/src/config/option-manager.js
loganfsmyth: I can't do that _currently_ because copying the input object would invalidate some of the plugin caching, that said I do think it is confusing right now. Perhaps something to revisit in the future?

Andarist: In that case it is certainly not worth changing right now.

packages/babel-core/src/config/option-assertions.js
```diff
@@ -146,16 +146,31 @@ function assertPluginItem(
     if (value.length === 0) {
       throw new Error(`.${key}[${index}] must include an object`);
     }
-    if (value.length > 2) {
-      throw new Error(`.${key}[${index}] may only be a two-tuple`);
+
+    if (value.length > 3) {
```
loganfsmyth: This

Issue #6816, opened 2017-11-13
New Feature: move https://github.com/jridgewell/babel-plugin-transform-for-of-as-array into this repo as an option


hzoo:
Basically just port/implement this as a plugin option

https://github.com/jridgewell/babel-plugin-transform-for-of-as-array, option name can just be `forOfAsArray` or some other thing we can bikeshed (buble does dangerousForOf)

cc @jridgewell since you made it?

jridgewell:
Sure.

rajzshkr:
I'll pick this

AkshatSrivastava2:
Hello,

I am new to open source.
I want to contribute to babel.
Guide me

loganfsmyth:
@AkshatSrivastava2 Welcome! Take a look at https://github.com/babel/babel/blob/master/CONTRIBUTING.md which should get you started. I believe someone else has already started work on this specific issue though.

AkshatSrivastava2:
@loganfsmyth Thank you for your response.

rajzshkr:
@AkshatSrivastava2 already working on this. Please pick other issues. Thanks.

existentialism:
Fixed in https://github.com/babel/babel/pull/6914

PR #6914, opened 2017-11-27
Porting babel-plugin-transform-for-of-as-array into transform-for-of as option

rajzshkr:
| Fixed Issues?            | #6816 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-transform-for-of/src/index.js
+++ b/packages/babel-plugin-transform-for-of/src/index.js
@@ -2,6 +2,60 @@ import { template, types as t } from "@babel/core";
 
 export default function(api, options) {
   const { loose, assumeArray } = options;
+
+  if (loose === true && assumeArray === true) {
+    throw new Error(
+      `Cannot use loose and assumeArray options together in babel-plugin-transform-for-of`,
+    );
+  }
+
+  if (assumeArray) {
+    return {
+      visitor: {
+        ForOfStatement(path) {
+          const { scope } = path;
+          const { left, right, body } = path.node;
+          const i = scope.generateUidIdentifier("i");
+          let array = scope.maybeGenerateMemoised(right, true);
+
+          const inits = [t.variableDeclarator(i, t.numericLiteral(0))];
+          if (array) {
+            inits.push(t.variableDeclarator(array, right));
+          } else {
+            array = right;
+          }
+
+          const item = t.memberExpression(array, t.clone(i), true);
+          let assignment;
+          if (t.isVariableDeclaration(left)) {
+            assignment = left;
+            assignment.declarations[0].init = item;
+          } else {
+            assignment = t.expressionStatement(
+              t.assignmentExpression("=", left, item),
+            );
+          }
+
+          const block = t.toBlock(body);
+          block.body.unshift(assignment);
+
+          path.replaceWith(
+            t.forStatement(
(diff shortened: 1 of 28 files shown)


rajzshkr:
Usage:-

```js
{
  "plugins": [["transform-for-of", {
    "forOfAsArray": true
    }]
  ]
}
```

packages/babel-plugin-transform-for-of/src/index.js
```diff
@@ -1,7 +1,7 @@
 import { template, types as t } from "@babel/core";
 
 export default function(api, options) {
-  const { loose } = options;
+  const { loose, forOfAsArray: isForOfAsArray } = options;
```
Andarist: would prefer a little bit more descriptive name - `useForAsArray` or something and that name could be used both as the internal flag and the public plugin option

nicolo-ribaudo: What about just `assmeArray`? After all, `for of` is already specified in the plugin name.

rajzshkr: @nicolo-ribaudo assumeArray?

Andarist: Yeah, `assumeArray` sounds like a great name for this.

rajzshkr: yeah, I will update the PR.

nicolo-ribaudo:
1. Since the `loose` and `assumeArray` options are not compatible, I think we should throw an error when both are `true`.
2. If `assumeArray` if `true`, we can avoid defining the [`buildForOfArray`](https://github.com/babel/babel/pull/6914/files#diff-55e984c93d4c177235e494c251a9eadeR9), [`buildForOfLoose`](https://github.com/babel/babel/pull/6914/files#diff-55e984c93d4c177235e494c251a9eadeR13) and [`buildForOf`](https://github.com/babel/babel/pull/6914/files#diff-55e984c93d4c177235e494c251a9eadeR31) templates by returing earlier.

packages/babel-plugin-transform-for-of/src/index.js
```diff
@@ -1,7 +1,61 @@
 import { template, types as t } from "@babel/core";
 
 export default function(api, options) {
-  const { loose } = options;
+  const { loose, assumeArray } = options;
+
+  if (loose === true && assumeArray === true) {
+    throw new Error(
+      `Cannot use loose and assumeArray options together in babel-plugin-transform-for-of`,
```
existentialism: Nit:

> The `loose` and `assumeArray` options cannot be used together in @babel/plugin-transform-for-of

packages/babel-plugin-transform-for-of/src/index.js
```diff
@@ -107,49 +161,6 @@ export default function(api, options) {
     }
   }
 
-  return {
```
hzoo: none of the code here is any different right? (just saying this part of the diff isn't necessary as clarification)

hzoo:
We can test this by using it lol, but can be another PR https://github.com/babel/babel/blob/d8bbaaae0a132cf464c27126957bd66d74b2bf65/packages/babylon/package.json#L27

PR #6920, opened 2017-11-27
Split exportExtensions into exportDefault and exportNamespace plugins…

### LINKED ISSUE #6678: Split export extensions into 2
> Issue originally reported by @hzoo in babel/babylon#706

So this is more of a refactor

exportExtensions:
Proposal 1: `export v from "mod"`
Proposal 2: `export * as ns from "mod"`

Like in https://github.com/babel/babel/pull/6080 except for the babylon plugin.

`exportDefault`?
`exportNamespace`?

- Check contributing https://github.com/babel/babylon/blob/master/CONTRIBUTING.md#creating-a-new-plugin-spec-new

---

- [ ] Comment below you are going to do this (for others to know)
- [ ] Read [CONTRIBUTING.md](https://github.com/babel/babylon/blob/master/CONTRIBUTING.md)
- [ ] Split the plugins into 2
- [ ] fix all the test fixtures (may need 2 folders, pass the correct option down)

existentialism:
… in babylon

| Fixed Issues?            | Fixes #6678
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Trying to clean up some v7 TODOs.

Was chatting with @hzoo but we _could_ rename the plugins (and resulting transforms) to `export-default-from` and `export-namespace-from` to get a slightly more clear error:

> This experimental syntax requires enabling the parser plugin: 'exportDefaultFrom'

But probably not a huge deal?

Also, I removed some redundant tests in Babylon.


packages/babel-plugin-syntax-export-default/README.md
```diff
@@ -0,0 +1,35 @@
+# @babel/plugin-syntax-export-default
+
+> Allow parsing of export default from.
```
nicolo-ribaudo: Nit: export default from should be in backticks so that it is formatted as code? (`export default from`)

packages/babel-plugin-syntax-export-namespace/README.md
```diff
@@ -0,0 +1,35 @@
+# @babel/plugin-syntax-export-namespace
+
+> Allow parsing of export namespace from.
```
nicolo-ribaudo: Also here

packages/babel-standalone/src/index.js
```diff
@@ -158,6 +159,8 @@ registerPlugins({
   "proposal-class-properties": require("@babel/plugin-proposal-class-properties"),
   "proposal-decorators": require("@babel/plugin-proposal-decorators"),
   "proposal-do-expressions": require("@babel/plugin-proposal-do-expressions"),
+  "proposal-export-default": require("@babel/plugin-proposal-export-default"),
```
existentialism: @Daniel15 I noticed `standalone` was including `syntax-export-extensions`, but not the actual transforms. Any reason not to?

hzoo: well one thing is that the repl uses stage-x but none of the plugins themselves since not sure who is using it that way

Daniel15: @existentialism Yeah this is fine. It might have just been an oversight. I have an item on my todo list to add a script that verifies that every syntax + transform plugin in the monorepo is listed here too. Maybe I'll file a task for that and let someone else do the work 😛

nicolo-ribaudo:
+1 for `exportDefaultFrom` instead of `exportDefault`

hzoo:
i'm cool with the longer name

packages/babel-generator/test/fixtures/types/ExportSpecifier16/options.json
```diff
@@ -1 +1 @@
-{ "plugins": ["exportExtensions"] }
```
hzoo: later: we could probably rename these test files

existentialism:
@hzoo @nicolo-ribaudo could use a quick re-review, added -from to transforms and plugins

packages/babel-plugin-syntax-export-namespace-from/README.md
```diff
@@ -0,0 +1,35 @@
+# @babel/plugin-syntax-export-namespace-from
+
+> Allow parsing of `export namespace from`.
```
nicolo-ribaudo: `export * as namespace from`

scripts/tests/flow/run_babylon_flow_tests.js
```diff
@@ -107,7 +107,7 @@ const options = {
 const flowOptionsMapping = {
   esproposal_class_instance_fields: "classProperties",
   esproposal_class_static_fields: "classProperties",
-  esproposal_export_star_as: "exportExtensions",
+  esproposal_export_star_as: "exportNamespace",
```
nicolo-ribaudo: `exportNamespaceFrom`

loganfsmyth:
I think we should consider adding an alias to https://github.com/babel/babel/blob/9d9710c765563addb4ec78e6eebe3e6ab15f1557/packages/babylon/src/index.js#L63 to check if `exportExtensions` is given, and auto-expand it to the two separate flags, to help things migrate.

hzoo:
Yeah that's a good idea, should be fine in another PR

PR #6922, opened 2017-11-27
UpdateExpressions as callees must be parenthesized

nicolo-ribaudo:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Examples:
```js
(a++)();
new (a++)();
```

I found this bug because I actually wrote something like `(x++)(function() {})()` because I forgot a semicolon lol


packages/babel-generator/src/node/parentheses.js
```diff
@@ -38,9 +38,13 @@ export function NullableTypeAnnotation(node: Object, parent: Object): boolean {
 export { NullableTypeAnnotation as FunctionTypeAnnotation };
 
 export function UpdateExpression(node: Object, parent: Object): boolean {
-  // (foo++).test()
   return (
-    (t.isMemberExpression(parent) && parent.object === node) ||
+    // (foo++).test(), (foo++)[0]
```
Andarist: out of curiosity, how to produce a code which spits out an object when postfix operator is used?

nicolo-ribaudo: I just tried using a combination of proxies and `with`, but I don't think it is possible.

Anyway, someone might do something like
```js
var i = 1;
var x = (i++).toString();
```

PR #6929, opened 2017-11-29
updated Unpkg link

bodaz:
...to be the same as the link in the Installation section


hzoo:
cool. there is no v6 for that scoped package anyway

PR #6939, opened 2017-11-30
Fix type definitions to fully support Typescript

### LINKED ISSUE #6933: Cannot build Typescript AST
_Is this a bug report or feature request?_
Bug report

I'm writing a plugin to add Typescript type annotations to existing code. Some of the `@babel/types` builders seem to be configured to expect Flow nodes as arguments instead of the Typescript equivalents. In particular, nodes with `typeParameters` fields expect `TypeParameterDeclaration`.

### Input Code

```js
path.node.init.returnType = path.node.init.returnType = t.tSTypeAnnotation(
  t.tSTypeReference(
    t.tSQualifiedName(t.identifier('React'), t.identifier('SFC')),
    t.tSTypeParameterInstantiation([t.tSAnyKeyword()])
  )
);
```

```js
{
    sourceType: 'module',
    allowImportExportEverywhere: false,
    allowSuperOutsideMethod: true,
    plugins: [
        "typescript",
        "jsx",
        "asyncGenerators",
        "classProperties",
        "doExpressions",
        "exportExtensions",
        "functionBind",
        "functionSent",
        "objectRestSpread",
        "dynamicImport"
    ]
}
```

### Expected Behavior

I would expect the code above to generate a type annotation like this:

```js
... React.SFC<any> ...
``` 

### Current Behavior

When running the plugin via `@babel/traverse`, I get the following stack trace:
```
TypeError: Property typeParameters of TSTypeReference expected node to be of a type ["TypeParameterInstantiation"] but instead got "TSTypeParameterInstantiation"
    at Object.validate (C:\Users\dpoindexter\Dev\Experiments\babel-codemod\node_modules\@babel\types\lib\definitions\index.js:86:13)
    at validate (C:\Users\dpoindexter\Dev\Experiments\babel-codemod\node_modules\@babel\types\lib\index.js:512:9)
    at Object.builder (C:\Users\dpoindexter\Dev\Experiments\babel-codemod\node_modules\@babel\types\lib\index.js:476:7)
```

### Possible Solution

It looks like `@babel/types` hasn't been comprehensively updated, but I don't know it well enough to be sure of intent. It seems that node types like `TypeParameterDeclaration` need to either become a shared alias, or everywhere that expects a `TypeParameterDeclaration` needs to become a union `TypeParameterDeclaration | TSTypeParameterDeclaration`.

I'm happy to create a PR with a fix if that seems like the right approach!

### Your Environment

| ---------------- | -------
| Babel            |  @babel/core 7.0.0-beta.32
| node             | 8.4.0
| yarn              | 1.3.2
| Operating System | Windows 10

dpoindexter:
| Fixed Issues?            | Fixes #6933 

This addresses missing or incorrect Typescript types in other type definitions.

For core, experimental, and ES2015 types, any type definition referencing a Flow node type (for example, the `typeParameters` property of the `ArrowFunctionExpression` node type) now also allows the equivalent Typescript node type. (In the case of `ArrowFunctionExpression.typeParameters`, allowed types became `TypeParameterDeclaration | TSTypeParameterDeclaration | Noop`.

For Typescript types, any type definition referencing a Flow node type replaced that reference with the equivalent Typescript node type. For example, in `TSInterfaceDeclaration.typeParameters`, allowed types became `TSTypeParameterDeclaration` rather than `TypeParameterDeclaration`.

### DIFF
--- a/packages/babel-types/src/definitions/core.js
+++ b/packages/babel-types/src/definitions/core.js
@@ -138,7 +138,10 @@ defineType("CallExpression", {
       optional: true,
     },
     typeParameters: {
-      validate: assertNodeType("TypeParameterInstantiation"),
+      validate: assertNodeType(
+        "TypeParameterInstantiation",
+        "TSTypeParameterInstantiation",
+      ),
       optional: true,
     },
   },
@@ -286,12 +289,19 @@ export const functionCommon = {
     validate: assertValueType("boolean"),
     default: false,
   },
+};
+
+export const functionTypeAnnotationCommon = {
   returnType: {
     validate: assertNodeType("TypeAnnotation", "TSTypeAnnotation", "Noop"),
     optional: true,
   },
   typeParameters: {
-    validate: assertNodeType("TypeParameterDeclaration", "Noop"),
+    validate: assertNodeType(
+      "TypeParameterDeclaration",
+      "TSTypeParameterDeclaration",
+      "Noop",
+    ),
     optional: true,
   },
 };
@@ -313,6 +323,7 @@ defineType("FunctionDeclaration", {
   visitor: ["id", "params", "body", "returnType", "typeParameters"],
   fields: {
     ...functionDeclarationCommon,
+    ...functionTypeAnnotationCommon,
     body: {
       validate: assertNodeType("BlockStatement"),
     },
@@ -340,6 +351,7 @@ defineType("FunctionExpression", {
   ],
   fields: {
     ...functionCommon,
+    ...functionTypeAnnotationCommon,
     id: {
       validate: assertNodeType("Identifier"),
       optional: true,
@@ -559,6 +571,7 @@ defineType("ObjectMethod", {
   builder: ["kind", "key", "params", "body", "computed"],
   fields: {
     ...functionCommon,
+    ...functionTypeAnnotationCommon,
     kind: {
       validate: chain(
         assertValueType("string"),
--- a/lib/types.js
+++ b/lib/types.js
@@ -82,7 +82,7 @@ declare class BabelNodeCallExpression extends BabelNode {
   callee: BabelNodeExpression;
   arguments: any;
   optional?: true | false;
-  typeParameters?: ?BabelNodeTypeParameterInstantiation;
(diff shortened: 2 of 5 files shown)


packages/babel-types/src/definitions/core.js
```diff
@@ -291,7 +294,11 @@ export const functionCommon = {
     optional: true,
   },
   typeParameters: {
-    validate: assertNodeType("TypeParameterDeclaration", "Noop"),
+    validate: assertNodeType(
```
existentialism: Guess one thing to think about, is that this change will make `tSDeclareMethod` and `tSDeclareFunction` (which spread `functionCommon` accept `TypeParameterDeclaration`.

dpoindexter: Good catch -- I missed that. Would you rather override the property for the TS types, or compose more granularly?

```js
defineType("TSDeclareFunction", {
  ...functionCommon,
  returnType {
    ...functionCommon.returnType
    validate: ...etc
  },
  typeParameters: {
    ...functionCommon.typeParameters,
    validate: ...etc.
  }
});
```
vs.
```js
export const functionCommon = {
  // everything there now but returnType and typeAnnotation
}

// only used for core types that need to accept either Flow or TS nodes
const functionCommonWithTypes = {
  returnType: {},
  typeAnnotation: {}
}
```
I didn't see other examples of overriding spread properties in definintions -- if that's generally the case, then the second approach might be more consistent with the rest of the code.

existentialism: Yeah let's go with the second

dpoindexter: @existentialism It's not immediately obvious because I squashed the commit, but this has been addressed.

I split the type annotations out of `functionCommon` into `functionTypeAnnotationCommon`, and spread both objects on the relevant core/es2015 types.

`functionDeclarationCommon` and `classMethodOrDeclareMethodCommon` still spread just  `functionCommon`. Types that use those now also spread either `functionTypeAnnotationCommon` or `tSFunctionTypeAnnotationCommon` as appropriate.

existentialism:
@dpoindexter thanks again!

PR #6941, opened 2017-11-30
[plugin-proposal-class-properties] Fix small loose docs typo

sdgluck:


packages/babel-plugin-proposal-class-properties/README.md
```diff
@@ -98,7 +98,7 @@ When `true`, class properties are compiled to use an assignment expression inste
   }
 ```
 
-Without `{ "loose": true }`, the above code will compile to the following, using `Object.definePropery`:
+Without `{ "loose": false }`, the above code will compile to the following, using `Object.definePropery`:
```
existentialism: "Without `{ "loose": false }`" seems correct, as in... when not in loose mode, it will use Object.defineProperty.

The _real_ typo is `Object.definePropery` -> `Object.defineProperty` :)

hzoo: Yeah you wouldn't use loose false, it is describing the default behavior

sdgluck: @existentialism Updated 👍

existentialism:
@sdgluck thanks!

PR #6943, opened 2017-12-01
[skip ci] - Moving do expression to stage 1

rajzshkr:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | 
| Documentation PR         | 
| Any Dependency Changes?  |

Based on the spec, we have to move `do expression` to stage 1 https://github.com/tc39/proposals#stage-1

### DIFF
--- a/packages/babel-preset-stage-0/package.json
+++ b/packages/babel-preset-stage-0/package.json
@@ -8,7 +8,6 @@
   "repository": "https://github.com/babel/babel/tree/master/packages/babel-preset-stage-0",
   "main": "lib/index.js",
   "dependencies": {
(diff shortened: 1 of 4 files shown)


rajzshkr:
> Btw the tests didn't catch that.

I didnt see any test cases to catch this

xtuc:
> Btw the tests didn't catch that.

Sorry I wasn't clear, I meant by this that there's no tests on our presets. Testing the transformations would be too difficult to maintain I think but we can at least test that they are correct Babel plugins.
What do you think?

Issue #6947, opened 2017-12-01
v7.0.0-beta.33 release


hzoo:
> What happened, issues, etc

- Still had to patch lerna to add `--access=public` for any new scoped packages (instead we would have to add it to each package.json which I guess could automate as well)
- Issue from https://github.com/babel/babel/pull/6920 @existentialism 
```
warning "@babel/preset-stage-0 > @babel/preset-stage-1 > @babel/plugin-proposal-export-default@7.0.0-beta.32" has incorrect peer dependency "@babel/core@7.0.0-beta.32".
warning "@babel/preset-stage-0 > @babel/preset-stage-1 > @babel/plugin-proposal-export-default > @babel/plugin-syntax-export-extensions@7.0.0-beta.32" has incorrect peer dependency "@babel/core@7.0.0-beta.32".
warning "@babel/preset-stage-0 > @babel/preset-stage-1 > @babel/preset-stage-2 > @babel/plugin-proposal-export-namespace@7.0.0-beta.32" has incorrect peer dependency "@babel/core@7.0.0-beta.32".
```

hzoo:
Going to close.

We just need to add this to all the new packages or PR lerna

```json
  "publishConfig": {
    "access": "public"
  },
```

PR #6831, opened 2017-11-15
[preset-env] Exclude transform-typeof-symbol with `loose` option.

yavorsky:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #6822
| Major: Breaking Change?  | 👍
| Tests Added + Pass?      | Yes
| License                  | MIT

Like @hzoo mentioned in https://github.com/babel/babel/issues/6822, we have enough reasons to do so, and this will simplify configs for many projects. I have found several examples, where people mix `exclude: ['transform-runtime']` and `loose: true` options:

- [emotion](https://github.com/emotion-js/emotion/blob/4d110f4c92ec30d369123e704685ee26b093dfac/packages/babel-plugin-emotion/test/.babelrc#L5-L8)
- [facepaint](https://github.com/emotion-js/facepaint/blob/master/.babelrc#L5-L8)
- [nickjs](https://github.com/phantombuster/nickjs/blob/cdf5910bcc4b0833dbbcb21425e0d3cf94290be8/examples/es8/.babelrc#L5-L9)
- [preact](https://github.com/developit/preact/blob/aab84f8f169298a5ee04598a7b7b0b429002bc91/.babelrc#L5-L9)
- [fast-equals](https://github.com/planttheidea/fast-equals/blob/bb7a2ca35aaa73526dffab005bc5000cd5f6efca/.babelrc#L34)
- [glimmer-vim](https://github.com/glimmerjs/glimmer-vm/blob/87da370aa5a73355b67643a58c2a6f43a3ba8924/build/broccoli/transpile-to-es5.js#L13-L23)
- [and many others...](https://github.com/search?p=1&q=exclude+transform-es2015-typeof-symbol+%22loose%22%3A+true&type=Code&utf8=%E2%9C%93)

PR #6948, opened 2017-12-01
update to beta.33

hzoo:
How I found the issue in https://github.com/babel/babel/issues/6947


existentialism:
We can also remove the `transform-typeof-symbol` exclusion since https://github.com/babel/babel/pull/6831

PR #6952, opened 2017-12-01
Add a 'lazy' options to modules-commonjs

loganfsmyth:
| Fixed Issues?            |
| Documentation PR         | 
| Any Dependency Changes?  |

Not inspired by the code from https://github.com/zertosh/babel-plugin-transform-inline-imports-commonjs itself, but certainly inspired by talking to @zertosh about it in the past, and now feeling like we might want to use it for Babel's own codebase. It's pretty trivial to include in our `helper-module-transforms` so I figured why not.

Allow import bindings to lazy-initialize themselves the first time they are accessed, instead of up front at file load time.


packages/babel-plugin-transform-modules-commonjs/src/index.js
```diff
@@ -16,9 +16,19 @@ export default function(api, options) {
...
     // Defaulting to 'true' for now. May change before 7.x major.
     allowCommonJSExports = true,
   } = options;
+
+  if (
+    typeof lazy !== "boolean" &&
+    typeof lazy !== "function" &&
+    (!Array.isArray(lazy) || !lazy.every(item => typeof item === "string"))
+  ) {
+    throw new Error(`.lazy must be a boolean, array of strings, or a function`);
```
danez: Is the . at the beginning intentional?

packages/babel-plugin-transform-modules-commonjs/src/index.js
```diff
@@ -132,14 +143,26 @@ export default function(api, options) {
 
             let header;
             if (isSideEffectImport(metadata)) {
+              if (metadata.lazy) throw new Error("Assertion failure");
```
danez: I guess this should never really happen, but maybe we could make the error a little bit more descriptive?

packages/babel-plugin-transform-modules-commonjs/README.md
```diff
@@ -130,3 +130,37 @@ function _interopRequireDefault(obj) {
...
+
+The value of `lazy` has a few possible effects:
+
+* `false` - No lazy initialization of any imported module.
+* `true` - Do not lazy-initialize local `./foo` imports, but lazy-init `foo` dependencies.
+
+  Local paths are much more likely to have circular dependencies, which may break if loaded lazily,
+  so they are not lazy by default, whereas dependencies between independent modules are rarely cyclical.
+
+* `Array<string>` - Lazy-initialize all imports with source matching one of the given strings.
```
danez: do we need `Array<string>`, do you think it is a common case? I guess `boolean | (string) => boolean` would capture these case too. I was also thinking about more cases like "enable really everything" or "enable everything with a blacklist", but that can all be handled with the callback.

nicolo-ribaudo: The disadvantage of the callback is that it forces the user to use `.babelrc.js` instead of `.babelrc`

packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js
```diff
@@ -249,6 +254,25 @@ function getModuleMetadata(programPath: NodePath, loose: boolean) {
...
+      if (lazy === true) {
+        // 'true' means that local relative files are eagerly loaded and
+        // dependency modules are loaded lazily.
+        metadata.lazy = !/\./.test(source);
+      } else if (Array.isArray(lazy)) {
+        metadata.lazy = lazy.indexOf(source);
+      } else if (typeof lazy === "function") {
+        metadata.lazy = lazy(source);
+      } else {
+        throw new Error(`'lazy' must be a boolean, string array, or function`);
```
nicolo-ribaudo: Here it shold throw the same message as https://github.com/babel/babel/pull/6952/files#diff-71f62b3225be3bae0ddc0c8e49189aa7R29

nicolo-ribaudo: Actualy, can this error be thrown? It should already been throw at https://github.com/babel/babel/pull/6952/files#diff-71f62b3225be3bae0ddc0c8e49189aa7R29

packages/babel-plugin-transform-modules-commonjs/src/index.js
xtuc: It's because it's a key in the configuration.

packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js
loganfsmyth: One throws up front when you declare the plugin, which seemed like it would be best for users, this one throws inside the transform itself in case someone else decides to use `helper-module-transforms` and calls it wrong. I can make the errors the same though, that's true.

packages/babel-plugin-transform-modules-commonjs/README.md
loganfsmyth: Yeah the `.babelrc`/`package.json#babel` case was the reason I added the array. Seems like a whitelist was easy enough to support that I'd just throw it in.

packages/babel-plugin-transform-modules-commonjs/src/index.js
loganfsmyth: I guess we don't really have a good naming scheme for these errors. I just included it since it is an object property. I can change it.

packages/babel-plugin-transform-modules-commonjs/src/index.js
loganfsmyth: For assertions I don't usually bother. I'd just `assert(!metadata.lazy)` but usually that isn't enough to make Flowtype happy so I've kind of got into the habit of doing it like this. I really just threw this in for us.

packages/babel-plugin-transform-modules-commonjs/src/index.js
xtuc: At least this is consistent with the other error messages used in Babel but it's not super clear to me. What about: "the key 'lazy' in the Babel configuration must be a boolean". Can be done in a second PR. That reminds my that I still have a PR to add code frames to configuration issues. I can continue my work, what do you think?

zertosh:
Super cool! Question, why did you decide to go with always calling the memoized require rather than conditionally getting the previously stored require value? For compat with modules whose value is falsey ? If always calling the memoized require's perf is on-par or better then I'll deprecate babel-plugin-transform-inline-imports-commonjs

loganfsmyth:
@zertosh I honestly didn't think on it too much. If anyone wants to do some benchmarks to compare, and your approach of having a separate variable is noticeably faster, we can certainly do it. I went this route because it seemed like it'd be easier for people debugging code to just have a single name to reference, and because it seemed like this approach would be pretty easy for engines to inline, since the function never changes and is statically guaranteed to always return the same value.

packages/babel-plugin-transform-modules-commonjs/src/index.js
loganfsmyth: > What about: "the key 'lazy' in the Babel configuration must be a boolean". 

I don't think it's much clearer than what I have here, but I do think we could do better in general with work focused on making the error messages give better feedback.

> That reminds my that I still have a PR to add code frames to configuration issues. I can continue my work, what do you think?

Not sure I follow. What code frames could we show for this case? Do you mean showing information about the config file itself?

jamesreggio:
@zertosh @loganfsmyth — I did some empirical benchmarking of both in my React Native app and witnessed the new `lazy` option to be 24% faster than `babel-plugin-transform-inline-imports-commonjs` (with a variance of 5%).

It's possible that other changes in the Babel 7 version of the mainline commonjs plugin are to account for these improvements, but I'd say — from my limited perspective — that it's safe to sunset `babel-plugin-transform-inline-imports-commonjs`.

Thank you both for this feature. It's made a _huge_ improvement in my cold-start time.

loganfsmyth:
@jamesreggio Happy to hear it. Beware there's one known bug: https://github.com/babel/babel/issues/7176

zertosh:
@jamesreggio, interesting - I expected _some_ difference one-way or the other, but not that stark. Curious to see this is a way I can repro (just for my know education). I'll deprecate `babel-plugin-transform-inline-imports-commonjs` as soon as Babel 7.0 is stable.

jamesreggio:
@zertosh — I was surprised too. I thought your use of a local variable would be more performant. I wish I could supply a repro to you, but I'm measuring real-user scenarios in our closed-source app, and I don't have time to build a separate benchmark suite.

PR #6955, opened 2017-12-02
Add pipeline operator plugins to babel-standalone

citycide:

Added missing pipeline operator syntax & transform plugins.


xtuc:
plugin-proposal-pipeline-operator is currently stage 1, shouldn't it be in the stage-1 preset? Which is in babel-standalone.

citycide:
@xtuc that only allows for using it when also using the `stage` preset. In my case I only want to enable the syntax and not the transform, so in a REPL-like setting ( playground for [`param.macro`](https://citycide.github.io/param.macro) ) the output maintains the pipelines.

existentialism:
@citycide thanks for the PR!

PR #6958, opened 2017-12-02
Update babylon to use unicode 10

danez:
| Major: Breaking Change?  | ?
| Documentation PR         | 
| Any Dependency Changes?  | y

I'm not sure what the implications are when updating to a new unicode version. Can we just do this? Will this just support newer stuff from the newer unicode version? Or may there be breaking changes?


packages/babylon/src/util/identifier.js
```diff
@@ -52,10 +52,10 @@ nonASCIIidentifierStartChars = nonASCIIidentifierChars = null;
...
 // generated by `bin/generate-identifier-regex.js`.
-// eslint-disable-next-line comma-spacing
-/* prettier-ignore */ const astralIdentifierStartCodes = [0,11,2,25,2,18,2,1,2,14,3,13,35,122,70,52,268,28,4,48,48,31,17,26,6,37,11,29,3,35,5,7,2,4,43,157,19,35,5,35,5,39,9,51,157,310,10,21,11,7,153,5,3,0,2,43,2,1,4,0,3,22,11,22,10,30,66,18,2,1,11,21,11,25,71,55,7,1,65,0,16,3,2,2,2,26,45,28,4,28,36,7,2,27,28,53,11,21,11,18,14,17,111,72,56,50,14,50,785,52,76,44,33,24,27,35,42,34,4,0,13,47,15,3,22,0,2,0,36,17,2,24,85,6,2,0,2,3,2,14,2,9,8,46,39,7,3,1,3,21,2,6,2,1,2,4,4,0,19,0,13,4,159,52,19,3,54,47,21,1,2,0,185,46,42,3,37,47,21,0,60,42,86,25,391,63,32,0,449,56,264,8,2,36,18,0,50,29,881,921,103,110,18,195,2749,1070,4050,582,8634,568,8,30,114,29,19,47,17,3,32,20,6,18,881,68,12,0,67,12,65,0,32,6124,20,754,9486,1,3071,106,6,12,4,8,8,9,5991,84,2,70,2,1,3,0,3,1,3,3,2,11,2,0,2,6,2,64,2,3,3,7,2,6,2,27,2,3,2,4,2,0,4,6,2,339,3,24,2,24,2,30,2,24,2,30,2,24,2,30,2,24,2,30,2,24,2,7,4149,196,60,67,1213,3,2,26,2,1,2,0,3,0,2,9,2,3,2,0,2,0,7,0,5,0,2,0,2,0,2,2,2,1,2,0,3,0,2,0,2,0,2,0,2,0,2,1,2,0,3,3,2,6,2,3,2,3,2,0,2,9,2,16,6,2,2,4,2,16,4421,42710,42,4148,12,221,3,5761,10591,541];
```
danez: Seems also we have a lot of duplicate pairs in there (like `2,0` or `0,2` for example), i guess the script that generates this could make a set of unique pairs? I think when looking at the code that this should be okay.

mathiasbynens:
@nicolo-ribaudo That’s the theory, but [there have been exceptions](https://hashseed.blogspot.com/2014/08/in-ecma-262-5.html). That’s why I carefully used the word “usually” in my earlier comment.

PR #6960, opened 2017-12-03
babel-types lists JSXIdentifier as an Expression

### LINKED ISSUE #6851: babel-types lists JSXIdentifier as an Expression
Noticed this when working on https://github.com/babel/minify/issues/724.

`isExpression()` returns true for paths that are `JSXIdentifiers`. I believe this is incorrect, as JSXIdentifiers are only allowed within `JSXOpeningElement` and `JSXClosingElement`, not anywhere an `Expression` is allowed. I believe this follows the JSX grammar definition [here](https://facebook.github.io/jsx/) where only `JSXElement` and `JSXFragment` are `PrimaryExpressions`.

tolmasky:
Per the JSX definition ( https://facebook.github.io/jsx/ ), only JSXElement and JSXFragment are actually Expressions.

Currently, Babel is treating JSXIdentifier, JSXMemberExpression, and JSXEmptyExpression as expressions as well, which
means Babel will for example incorrectly allow you to place these anywhere an expression is wanted. This fixes that.

Closes #6851.

Not sure if anything else is necessary, this seemed to be the place to edit.


tolmasky:
Thanks for the pointer! Per your advice I have modified the right file and then generated it using the script.

nicolo-ribaudo:
Did you also run `node scripts/generate-babel-types-docs.js`?

tolmasky:
Nope, just added.

tolmasky:
Anything else I need to do? Just checking in.

PR #6961, opened 2017-12-03
Handling babylon parsing errors in a better way

### LINKED ISSUE #6205: Handle syntax errors from babylon that need a plugin
Ref https://github.com/babel/babylon/pull/658

Need to check the error message from babylon and then throw a new error or something cool.

screen shot 2017-08-03 at 5 00 49 pm

Could do a mapping between the syntax plugin in babylon and the transform/preset to turn on for babel for the user, or even do a prompt to install, etc.

maaz93:
| Fixed Issues?            | Fixes #6205 

Introduces a better way to handle missing plugin errors from babylon, in babel. Previously, the errors used to be directly displayed, as is, from babylon. This was not so friendly as the error used to have only an internal name of the plugin required.

With this change, there is a mapping done between internal plugin names to the external plugin names, where required. There are **2** possible messages that are generated

1. When a plugin has both syntax and transform plugins:
>Support for the experimental syntax [babylon plugin name] isn't currently enabled ([loc]):
>
>[code frame]
>
>Add [npm package name] ([url]) to the 'plugins' section of your Babel config to enable transformation.

2. When a plugin has only a syntax plugin:
>Support for the experimental syntax [babylon plugin name] isn't currently enabled ([loc]):
>
>[code frame]
>
>Add [npm package name] ([url]) to the 'plugins' section of your Babel config to enable parsing.

All the URLs are shortened using [git.io](https://git.io), and point directly to the `repository` field of the plugin's `package.json`.

### DIFF
--- a/packages/babel-core/src/transformation/normalize-file.js
+++ b/packages/babel-core/src/transformation/normalize-file.js
@@ -6,7 +6,7 @@ import convertSourceMap, { typeof Converter } from "convert-source-map";
 import { parse } from "babylon";
 import { codeFrameColumns } from "@babel/code-frame";
 import File from "./file/file";
-import generateHelpMessage from "./util/missing-plugin-helper";
+import generateMissingPluginMessage from "./util/missing-plugin-helper";
 
 const shebangRegex = /^#!.*/;
 
@@ -89,7 +89,7 @@ function parser(pluginPasses, options, code) {
     }
     throw new Error("More than one plugin attempted to override parsing.");
   } catch (err) {
-    const loc = err.loc;
+    const { loc, missingPlugin } = err;
     if (loc) {
       err.loc = null;
       err.message =
@@ -103,8 +103,11 @@ function parser(pluginPasses, options, code) {
             },
           },
           options,
-        ) +
-        `\n${generateHelpMessage(err.missingPlugin)}`;
+        );
+      const missingPluginMessage = generateMissingPluginMessage(missingPlugin);
+      if (missingPluginMessage) {
+        err.message += `\n${missingPluginMessage}`;
+      }
     }
     throw err;
   }
--- a/packages/babel-core/src/transformation/normalize-file.js
+++ b/packages/babel-core/src/transformation/normalize-file.js
@@ -92,21 +92,23 @@ function parser(pluginPasses, options, code) {
     const { loc, missingPlugin } = err;
     if (loc) {
       err.loc = null;
-      err.message =
-        `${options.filename || "unknown"}: ${err.message}\n` +
-        codeFrameColumns(
-          code,
-          {
-            start: {
-              line: loc.line,
-              column: loc.column + 1,
-            },
+      const codeFrame = codeFrameColumns(
+        code,
+        {
+          start: {
+            line: loc.line,
+            column: loc.column + 1,
           },
-          options,
-        );
-      const missingPluginMessage = generateMissingPluginMessage(missingPlugin);
-      if (missingPluginMessage) {
-        err.message += `\n${missingPluginMessage}`;
+        },
+        options,
+      );
+      if (missingPlugin) {
+        err.message =
+          `${options.filename || "unknown"}: ` +
+          generateMissingPluginMessage(missingPlugin[0], loc, codeFrame);
+      } else {
+        err.message =
+          `${options.filename || "unknown"}: ${err.message}\n\n` + codeFrame;
       }
     }
     throw err;
--- a/packages/babel-core/src/transformation/normalize-file.js
+++ b/packages/babel-core/src/transformation/normalize-file.js
@@ -6,6 +6,7 @@ import convertSourceMap, { typeof Converter } from "convert-source-map";
 import { parse } from "babylon";
 import { codeFrameColumns } from "@babel/code-frame";
 import File from "./file/file";
+import generateHelpMessage from "./util/missing-plugin-helper";
 
 const shebangRegex = /^#!.*/;
 
@@ -102,7 +103,8 @@ function parser(pluginPasses, options, code) {
             },
           },
           options,
-        );
+        ) +
+        `\n${generateHelpMessage(err.missingPlugin)}`;
     }
     throw err;
   }
--- a/packages/babel-core/src/transformation/util/missing-plugin-helper.js
+++ b/packages/babel-core/src/transformation/util/missing-plugin-helper.js
@@ -97,6 +97,12 @@ const pluginNameMap = {
       url: "https://git.io/vb4SZ",
     },
   },
+  importMeta: {
+    syntax: {
+      name: "@babel/plugin-syntax-import-meta",
+      url: "https://git.io/vbKK6",
+    },
+  },
   jsx: {
     syntax: {
       name: "@babel/plugin-syntax-jsx",
--- a/packages/babel-core/src/transformation/util/missing-plugin-helper.js
+++ b/packages/babel-core/src/transformation/util/missing-plugin-helper.js
@@ -191,28 +191,38 @@ const pluginNameMap = {
 
 const getNameURLCombination = ({ name, url }) => `${name} (${url})`;
 
+/*
+Returns a string of the format:
+Support for the experimental syntax [babylon plugin name] isn't currently enabled ([loc]):
+
+[code frame]
+
+Add [npm package name] ([url]) to the 'plugins' section of your Babel config
+to enable [parsing|transformation].
+*/
 export default function generateMissingPluginMessage(
-  missingPlugins: string[] = [],
+  missingPluginName: string,
+  loc,
+  codeFrame,
 ): string {
-  let helpMessage = "";
-  const pluginName = missingPlugins.length && missingPlugins[0];
-  if (pluginName) {
-    const pluginInfo = pluginNameMap[pluginName];
-    if (pluginInfo) {
-      const { syntax: syntaxPlugin, transform: transformPlugin } = pluginInfo;
-      if (syntaxPlugin) {
+  let helpMessage =
+    `Support for the experimental syntax '${missingPluginName}' isn't currently enabled ` +
+    `(${loc.line}:${loc.column + 1}):\n\n` +
+    codeFrame;
+  const pluginInfo = pluginNameMap[missingPluginName];
+  if (pluginInfo) {
+    const { syntax: syntaxPlugin, transform: transformPlugin } = pluginInfo;
+    if (syntaxPlugin) {
+      if (transformPlugin) {
+        const transformPluginInfo = getNameURLCombination(transformPlugin);
+        helpMessage +=
+          `\n\nAdd ${transformPluginInfo} to the 'plugins' section of your Babel config ` +
+          `to enable transformation.`;
+      } else {
         const syntaxPluginInfo = getNameURLCombination(syntaxPlugin);
-        if (transformPlugin) {
-          const transformPluginInfo = getNameURLCombination(syntaxPlugin);
-          helpMessage =
-            `Support for this experimental syntax isn't currently enabled. ` +
-            `Add either ${transformPluginInfo} or ${syntaxPluginInfo} (if you only need parsing support) ` +
-            `to the 'plugins' section of your Babel config.`;
-        } else {
-          helpMessage =
-            `Support for this experimental syntax isn't currently enabled. ` +
-            `Add ${syntaxPluginInfo} to the 'plugins' section of your Babel config.`;
-        }
+        helpMessage +=
+          `\n\nAdd ${syntaxPluginInfo} to the 'plugins' section of your Babel config ` +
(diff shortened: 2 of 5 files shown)


maaz93:
I need some feedback on writing tests for this. They might probably be fixtures which expect an error to be thrown. If the messages that I have chosen to output are approved, then I'll add tests. As a change in the messages will result in changing the tests.

Also, open to feedback if I should write multiple fixtures or just some sort of unit tests for the new file added.

packages/babel-core/src/transformation/normalize-file.js
```diff
@@ -102,7 +103,8 @@ function parser(pluginPasses, options, code) {
             },
           },
           options,
-        );
+        ) +
+        `\n${generateHelpMessage(err.missingPlugin)}`;
```
nicolo-ribaudo: Can `\n` be added only if `generateHelpMessage` doesn't return `""`?

packages/babel-core/src/transformation/util/missing-plugin-helper.js
```diff
@@ -0,0 +1,129 @@
...
+    if (pluginInfo) {
+      const {
+        syntax: syntaxPlugin,
+        transform: transformPlugin,
+        issueLink,
+      } = pluginInfo;
+      if (syntaxPlugin) {
+        if (transformPlugin) {
+          helpMessage =
+            `Use the ${syntaxPlugin} plugin to enable parsing. ` +
```
nicolo-ribaudo: What about adding `This experimental syntax isn't enabled by default` as an "inroduction" to the error message?

packages/babel-core/src/transformation/util/missing-plugin-helper.js
```diff
@@ -0,0 +1,129 @@
...
+            `Use the ${transformPlugin} plugin to enable transformation.` +
+            `\nNote: The transform plugins already include the syntax plugins, so you don't need both.`;
+        } else {
+          helpMessage =
+            `Use the ${syntaxPlugin} plugin to enable parsing. ` +
+            `Babel doesn't support transforming this syntax.`;
+        }
+      } else if (issueLink) {
+        helpMessage =
+          `This syntax is not supported yet. ` +
```
nicolo-ribaudo: `This experimental syntax`...

packages/babel-core/src/transformation/util/missing-plugin-helper.js
```diff
@@ -0,0 +1,129 @@
...
+      const {
+        syntax: syntaxPlugin,
+        transform: transformPlugin,
+        issueLink,
+      } = pluginInfo;
+      if (syntaxPlugin) {
+        if (transformPlugin) {
+          helpMessage =
+            `Use the ${syntaxPlugin} plugin to enable parsing. ` +
+            `Use the ${transformPlugin} plugin to enable transformation.` +
```
nicolo-ribaudo: Can we suggest the transform plugin before the syntax plugin, since it is the most important one?

packages/babel-core/src/transformation/normalize-file.js
existentialism: Also can we rename `generateHelpMessage` to `generateMissingPluginMessage`?

packages/babel-core/src/transformation/util/missing-plugin-helper.js
existentialism: And `it's` -> `its`

packages/babel-core/src/transformation/util/missing-plugin-helper.js
existentialism: I'm wondering if we even need a different message in the case where there's only a syntax plugin and not a transform (just dynamic-import)? I feel like we can just simplify this down to something like:

> Support for this experimental syntax isn't currently enabled. Add `<insert transform/syntax plugin here>` to the `plugins` section of your Babel config to enable it.

(Also, do we need the line re: syntax plugin is enabled by the transform? I feel like most people don't even know the difference?)

xtuc:
Nice first contribution @maaz93, it's so much clearer!

packages/babel-core/src/transformation/util/missing-plugin-helper.js
xtuc: It this used? That means we are adding early errors in the parser at a very early stage of the proposal, we don't do that currently?

packages/babel-core/src/transformation/util/missing-plugin-helper.js
xtuc: I would also add a link to the website because we can explain the user how to install it, and if a transformation is needed. What do you think? That also means we should have doc about all the proposals (or issues if not official)

packages/babel-core/src/transformation/normalize-file.js
maaz93: I'll incorporate both these changes

packages/babel-core/src/transformation/util/missing-plugin-helper.js
maaz93: I'll use @existentialism 's messages. I think they take care of @nicolo-ribaudo 's suggestion as well. 
@xtuc like you mentioned, not all the plugins have docs on the website. Would it be better to point to the particular package's README instead?

packages/babel-core/src/transformation/util/missing-plugin-helper.js
maaz93: Will add the changes suggested in the first two comments. @xtuc I guess so. I'm not well versed with the parser yet, so I don't have a concrete answer. I just picked up all the plugins from the babylon README.

packages/babel-core/src/transformation/util/missing-plugin-helper.js
maaz93: Sure.

packages/babel-core/src/transformation/util/missing-plugin-helper.js
xtuc: @maaz93 Yes it is. Since it's a link it's generic enough, we can change it later if needed.

maaz93: @xtuc Let me clear my understanding. So, if I was suggesting a message for missing `asyncGenerators` plugin, I would have to add `https://github.com/babel/babel/tree/master/packages/babel-plugin-syntax-async-generators` and `https://github.com/babel/babel/tree/master/packages/babel-plugin-proposal-async-generator-functions` in the message? Or I need to suggest `https://babeljs.io/docs/plugins/` for more info?

xtuc: I can use the READMEs for now, I'm not sure that we have the complete doc on the website tho.

maaz93: While trying to add this, I realised that the message will get very long with links in it. I thought I'll just suggest

> You can find more details about every plugin at https://github.com/babel/babel/tree/master/packages

Will this be enough?

nicolo-ribaudo: We could even use git.io and make it shorter

maaz93: @nicolo-ribaudo That was a great suggestion! I wrote a quick dirty script to use a [gitio](https://www.npmjs.com/package/gitio) in a loop and generate short URLs. The map looks something like this now

```javascript
asyncGenerators: {
  syntax: {
    name: "@babel/plugin-syntax-async-generators",
    url: "https://git.io/vb4SY",
  },
  transform: {
    name: "@babel/plugin-proposal-async-generator-functions",
    url: "https://git.io/vb4yp",
  },
}
```
With this, the message would look like

> Support for this experimental syntax isn't currently enabled. Add either `<insert transform plugin here>`(`insert URL`) or `<insert syntax plugin>`(`insert URL`) (if you only need parsing support) to the plugins section of your Babel config.

Is this more clear?

packages/babel-core/src/transformation/normalize-file.js
```diff
@@ -103,6 +104,10 @@ function parser(pluginPasses, options, code) {
           },
           options,
         );
+      const missingPluginMessage = generateMissingPluginMessage(missingPlugin);
```
existentialism: Just thinking aloud, but looking at this again... right now it outputs something like:

> This experimental syntax requires enabling the parser plugin: 'asyncGenerators' (1:14)
> \> 1 | async function* agf() {
>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;^
>&nbsp;&nbsp;&nbsp;2 |   await 1;
>&nbsp;&nbsp;&nbsp;3 |   yield 2;
>&nbsp;&nbsp;&nbsp;4 | }
> Support for this experimental syntax isn't currently enabled. Add either @babel/plugin-syntax-async-generators (https://git.io/vb4SY) or @babel/plugin-syntax-async-generators (https://git.io/vb4SY) (if you only need parsing support) to the 'plugins' section of your Babel config.

Should we just omit the error from Babylon if `missingPlugin` and maybe just do:

> Support for the experimental syntax 'asyncGenerators' isn't currently enabled (1:14)
>
> \> 1 | async function* agf() {
>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;^
>&nbsp;&nbsp;&nbsp;2 |   await 1;
>&nbsp;&nbsp;&nbsp;3 |   yield 2;
>&nbsp;&nbsp;&nbsp;4 | }
>
> Add either @babel/plugin-syntax-async-generators (https://git.io/vb4SY) or @babel/plugin-syntax-async-generators (https://git.io/vb4SY) (if you only need parsing support) to the 'plugins' section of your Babel config.

maaz93: Honestly, I thought about this. But I didn't know how it can be improved. The current message repeats the fact that there is some experimental plugin missing, twice.

I like the new message proposal. It's more meaningful and less repetitive. Can I go ahead and make this change?

hzoo: The original purpose for making the issue is that yes we would remove the original message and output the new one. I would only add the syntax part for things that we don't have a default transform for, otherwise it should only say the plugin. This is because the use case for babel is the transform not the syntax (because then you would be using babylon by itself)

maaz93: @hzoo @existentialism Then, the apt messages would be

>Support for the experimental syntax 'asyncGenerators' isn't currently enabled (1:14)
   \<codeFrame\>
Add @babel/plugin-proposal-async-generator-functions (https://git.io/vb4yp) to the 'plugins' section of your Babel config to enable transformation.

And, in the case where there is only a syntax plugin available

>Support for the experimental syntax 'dynamicImport' isn't currently enabled (1:14)
   \<codeFrame\>
Add @babel/plugin-syntax-dynamic-import (https://git.io/vb4Sv) to the 'plugins' section of your Babel config to enable parsing.

existentialism: Yea, I'm fine with that. Note, can we add line breaks around the codeframe:

> Support for the experimental syntax `[babylon plugin name]` isn't currently enabled ([loc]):
>
> [code frame]
>
> Add [npm package name] ([url]) to the 'plugins' section of your Babel config to enable [parsing|transformation].

nicolo-ribaudo:
(NOTE) If https://github.com/babel/babel/pull/7008 gets merged before this PR, it should be added to the plugins list.

maaz93:
@nicolo-ribaudo I'll do that. 

I just had another thought. Do you think it makes sense to write a test case which iterates through all the plugins in the map and makes sure that only `syntax` or/and `transform` keys are present per plugin and the `url` matches `https://git.io/*` ? I feel this will ensure strict rules on the structure of the map. Only side effect of this would be that I'll have to `export` the map. Let me know if that would be a good addition to this PR. 😄

hzoo:
Nice work @maaz93, this will help a lot! Looking forward to how we can improve this as well

maaz93:
@hzoo Thanks! I would love to improve it based on any suggestions or help anyone who's looking to make any improvements, if needed. 😄

PR #6962, opened 2017-12-03
Better error message for `import.meta` and `import()` without plugin

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babylon/test/fixtures/experimental/dynamic-import/no-plugin/options.json
+++ b/packages/babylon/test/fixtures/experimental/dynamic-import/no-plugin/options.json
@@ -1,4 +1,4 @@
 {
-  "throws": "Unexpected token, expected \"{\" (1:6)",
+  "throws": "This experimental syntax requires enabling the parser plugin: 'dynamicImport' (1:0)",
   "plugins": []
 }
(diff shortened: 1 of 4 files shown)


packages/babylon/test/fixtures/experimental/dynamic-import/no-plugin/options.json
```diff
@@ -1,4 +1,4 @@
 {
-  "throws": "Unexpected token, expected \"{\" (1:6)",
+  "throws": "This experimental syntax requires enabling the parser plugin: 'dynamicImport' (1:0)",
```
xtuc: I wouldn't use `this experimental syntax` because it's just a question of time. What about just `this syntax`?

nicolo-ribaudo: When it won't be experimenta anymore it will be enabled by default, so the error won't be thrown.

xtuc: - Yes! good point
- No, imagine you use the Babylon's API, you can choose which plugin you want.

It's good to me since i'm not really sure what we can do in the second case.

PR #6963, opened 2017-12-03
Create static class properties in a class IIFE

### LINKED ISSUE #5902: [static class properties] pull them into compiled closure
**INPUT**
```js
class MyClass {
  static myProperty = value;
  method() {}
}
```
**ACTUAL OUTPUT**
```js
var MyClass = function () {
  function MyClass() {
    _classCallCheck(this, MyClass);
  }

  _createClass(MyClass, [{
    key: "method",
    value: function method() {}
  }]);

  return MyClass;
}();

MyClass.myProperty = value;
```
**EXPECTED OUTPUT**
```diff
var MyClass = function () {
  function MyClass() {
    _classCallCheck(this, MyClass);
  }

  _createClass(MyClass, [{
    key: "method",
    value: function method() {}
  }]);

+  MyClass.myProperty = value;

  return MyClass;
}();

- MyClass.myProperty = value;
```
PROBLEM

Tree-shaking flat bundles. By assigning property to an unused variable we prevent possibility to tree-shake the unused variable as its treated as side effect. More can be read [here](https://iamakulov.com/notes/polished-webpack/).

This won't immediately allow for tree-shaking classes as they are still usually transpiled to IIFEs which means for uglify a potential side effect risk (hard to statically determine if running the function will cause a side effect) and needs for it to work the function call needs to be decorated with `/*#__PURE__*/` comment - the issue for this is [already created](https://github.com/babel/babel/issues/5632). The issue described here (statics) is also mentioned in the other thread, but I'm creating this to separate those 2 as they can be treated as unrelated.

In the other thread the simple implementation of this was even proposed by @kzc [here](https://github.com/babel/babel/issues/5632#issuecomment-299543425). It's a good start, however there is a problem with the approach - we need to support both usages:
- babel-plugin-transform-class-properties + babel-plugin-transform-es2015-classes
- babel-plugin-transform-class-properties alone

Where the proposed approach deals only with the first one.

Andarist:
| Fixed Issues?            | Fixes #5902 
| Documentation PR         | n/a
| Any Dependency Changes?  | internal 

I must say I'm not completely satisfied with the implementation. I've found that handling `ClassExpression`s was troublesome and I've left them at the moment unpolished - in combination with static properties they can produce 2 nested IIFEs now - I'd like to revisit this, but possibly as a followup PR as I don't have much time lately to work on OSS and I wouldn't like to postpone merging this IMHO important PR.


packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -75,8 +79,33 @@ export default function(api, options) {
     inherits: syntaxClassProperties,
 
     visitor: {
+      ExportDefaultDeclaration(path) {
```
Andarist: this is literally copied from [here](https://github.com/babel/babel/blob/62bbee97d7245b897973d290d1700797a85fe168/packages/babel-plugin-transform-classes/src/index.js#L16-L31) any suggestions what's the best way to dedupe this?

packages/babel-plugin-transform-classes/src/vanilla.js
```diff
@@ -152,14 +153,21 @@ export default class ClassTransformer {
...
-    return t.callExpression(container, closureArgs);
+    iife.arguments.push(...closureArgs);
+    iife.callee.params.push(...closureParams);
+
+    if (iife.callee.body.body.length === 0) {
+      iife.callee.body.body.push(...body, t.returnStatement(this.classRef));
+      return iife;
+    }
+
+    if (this.path.isClassExpression()) {
```
Andarist: hating this `IfStatement` here, kinda a hack to insert a correct return for mentioned double IIFEs case - need to figure out how to flatten those IIFEs (in case of `ClassExpresion` with static properties) and keep all surrounding references correct

packages/babel-plugin-proposal-class-properties/test/fixtures/general/non-block-arrow-func/expected.js
```diff
@@ -1,27 +1,32 @@
 export default (param => {
-  var _class, _temp;
+  var _class;
 
-  return _temp = _class =
+  return _class =
```
nicolo-ribaudo: Is it possible to remove this unused `_class =` assignment?

packages/babel-plugin-proposal-class-properties/test/fixtures/general/computed/expected.js
```diff
@@ -6,7 +6,7 @@ var four = 4;
 
 var MyClass =
 /*#__PURE__*/
```
nicolo-ribaudo: Isn't this iife pure only if computed properties are pure?

packages/babel-plugin-proposal-class-properties/test/fixtures/general/regression-T7364/expected.js
```diff
@@ -42,7 +42,7 @@ class MyClass {
 
 });
 
-export default class MyClass3 {
+class MyClass3 {
```
nicolo-ribaudo: Can we split the class and the export statement only if it is actually needed? (Only if it doesn't add too much complexity to the plugin)

packages/babel-plugin-proposal-class-properties/src/index.js
nicolo-ribaudo: You can be a `path.splitExportDeclaration()` function (or a `@babel/helper-...`)

packages/babel-plugin-proposal-class-properties/test/fixtures/general/non-block-arrow-func/expected.js
Andarist: it probably is, added `_class` is caused by `ClassExpression` being replaced by multiple statements. The `path#replaceWithMultiple` creates an IIFE when the conversion is made so the single expression can be maintained in place.

This is certainly something I'd like to revisit, but for another month (till Jan 16th) I have very little time and I couldn't manage to do this properly at the moment because of the complexity and probably my insufficient experience with AST manipulation.

Ideally I would like to "inline" the `ClassExpression` here if possible to skip second IIFE creation, but I'm not sure how safe it is because scope references have to be synced and stuff.

packages/babel-plugin-proposal-class-properties/test/fixtures/general/computed/expected.js
Andarist: Yes, you are right. They can also be impure in other cases, such as impure property initializers, or even super class being an impure expression.

I've wanted to keep marking those class IIFEs as pure, because it is what transform-classes do at the moment in all cases (and classes without properties can be impure too). That's quite my fault, adding those annotations was one of my 1st PRs to babel.

I'd like to revisit (as another PR) annotating classes as pure as a whole and not just in case of proposal-class-properties plugin. While checking if class is pure should be quite easy, I'm not sure what we should do about subclasses - we won't be able to tell if a super class is pure or not.

packages/babel-plugin-proposal-class-properties/test/fixtures/general/regression-T7364/expected.js
Andarist: It was easier to split those for now in all cases, need to check how much complexity it would add to keep them together (and the same should be done in transform-classes as it is splitting those every time)

Andarist:
@nicolo-ribaudo thanks for the initial review!

packages/babel-plugin-proposal-class-properties/src/index.js
Andarist: was thinking about both, weren't sure which one to pick :D I'd probably go with latter, thanks

packages/babel-plugin-transform-classes/src/vanilla.js
nicolo-ribaudo: https://github.com/babel/babel/blob/master/packages/babel-traverse/src/scope/lib/renamer.js#L42-L85 looks similar

packages/babel-plugin-proposal-class-properties/src/index.js
nicolo-ribaudo: https://github.com/babel/babel/blob/master/packages/babel-traverse/src/scope/lib/renamer.js#L42-L85 looks similar

packages/babel-helper-split-export-declaration/src/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+  if (isDefault) {
+    if (declaration.isIdentifier()) {
+      throw new Error("It doesn't make sense to split exported identifier.");
+    }
+    const standaloneDeclaration =
+      declaration.isFunctionDeclaration() || declaration.isClassDeclaration();
+
+    let id = declaration.node.id;
+    const liveBinding = !!id;
```
nicolo-ribaudo:
```js
export default function() {}
```
is transformed into
```js
function _default() {}
export default _default;
```

The first export statement is hoisted, but is it also the second?

Andarist: could u elaborate? I've thought about hoisting briefly in this context, but came to conclusion that no extra work needs to be done for export statements, they are evaluated in place, right? and as function declaration will get hoisted by javascript this identifier will be available by the time it gets exported

packages/babel-plugin-proposal-class-properties/test/fixtures/general/non-block-arrow-func/expected.js
Andarist: This case (and other similar) are caused by simple insertion [here](https://github.com/babel/babel/blob/f3410e761ceff73cf7c2c45bb79fff6479a28d5b/packages/babel-plugin-proposal-class-properties/src/index.js#L228-L230). To remove this unused variable the logic here would have to be expanded. I might revisit this in a followup PR, but would like to keep this one as simple as possible.

packages/babel-helper-split-export-declaration/src/index.js
nicolo-ribaudo: After playing a bit with it, I understood that the way you implemented it is correct 👍

Andarist:
@nicolo-ribaudo 
I've finished working on this PR, seems feature complete for me. Would love your code review on this :)

packages/babel-helper-class-iife/src/index.js
```diff
@@ -0,0 +1,25 @@
+import { types as t } from "@babel/core";
+import annotateAsPure from "@babel/helper-annotate-as-pure";
+
+const classIIFEs = new WeakMap();
+
+export function getClassIIFE(path, assumePure) {
```
nicolo-ribaudo: Nit: can you use a default export? It seems that most helpers do that.

packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js
```diff
@@ -171,7 +171,7 @@ function getModuleMetadata(
           if (reexport) {
             localData.delete(localName);
 
-            reexport.names.forEach(name => {
+            reexport.bindings.map(({ name }) => name).forEach(name => {
```
nicolo-ribaudo: Can you use `reexport.bindings.forEach({ name } => {` and avoid an unnecessary loop? Also on lines 186 and 200

packages/babel-helper-module-transforms/src/index.js
```diff
@@ -262,7 +262,8 @@ function buildExportNameListDeclaration(
 ) {
   const exportedVars = Object.create(null);
   for (const data of metadata.local.values()) {
-    for (const name of data.names) {
+    const names = data.bindings.map(({ name }) => name);
+    for (const name of names) {
```
nicolo-ribaudo: Here you can avoid looping twice:
```js
for (const { name } of data.bindings) {
```

packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js
```diff
@@ -371,16 +374,19 @@ function getLocalExportMetadata(
...
           }
-          getLocalMetadata(local).names.push(exported.node.name);
+          getLocalMetadata(local).addName(exported.node.name);
         });
       }
     } else if (child.isExportDefaultDeclaration()) {
       const declaration = child.get("declaration");
-      if (
+
+      if (declaration.isIdentifier()) {
```
nicolo-ribaudo: Aren't identifiers normal expressions? e.g., what is the difference between
```js
var a = 2;
export default a;
```
and
```js
export default 2;
```
?

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -1,9 +1,14 @@
 import nameFunction from "@babel/helper-function-name";
 import syntaxClassProperties from "@babel/plugin-syntax-class-properties";
+import { getClassIIFE } from "@babel/helper-class-iife";
+import splitExportDeclaration from "@babel/helper-split-export-declaration";
 import { template, types as t } from "@babel/core";
 
 export default function(api, options) {
-  const { loose } = options;
+  const { loose, assumePure: _assumePure } = options;
+  const assumePure = typeof _assumePure === "boolean" ? _assumePure : !!loose;
```
nicolo-ribaudo: I'd prefer this to throw if `assumePure` is specified and it is neither a boolean nor null/undefined.

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -219,21 +231,56 @@ export default function(api, options) {
...
           path.node.id = ref;
         }
 
-        path.insertAfter(nodes);
+        if (staticNodes.length === 0) {
+          if (path.parentPath.isExportDeclaration()) {
+            path.parentPath.insertAfter(nodes);
+          } else {
+            path.insertAfter(nodes);
+          }
```
nicolo-ribaudo: You can just use `path.insertAfter(nodes)`, because of https://github.com/babel/babel/pull/7040

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -98,9 +112,9 @@ export default function(api, options) {
 
         let ref;
 
-        if (path.isClassExpression() || !path.node.id) {
+        if (isClassExpression || !path.node.id) {
           nameFunction(path);
-          ref = path.scope.generateUidIdentifier("class");
+          ref = path.node.id || path.scope.generateUidIdentifier("class");
```
nicolo-ribaudo: You should always generate a new identifier, otherwise it won't work in cases like https://babeljs.io/repl/build/6407/#?babili=false&browsers=&build=&builtIns=false&code_lz=DYUwLgBAghC8ECYDcAoFAzAdgCgMbAEMBnI6CAbxQgiLALAEtcIAHAJwHsW5FUBfAJRIgA&debug=false&circleciRepo=&evaluate=false&lineWrap=false&presets=stage-0&prettier=false&targets=&version=7.0.0-beta.36%2Bpr.6963 (`A` gets declared twice).

You can do something like
```js
ref = path.scope.generateUidIdentifier(path.node.id ? path.node.id.name : "class");
```

packages/babel-plugin-proposal-class-properties/src/index.js
Andarist: Gonna go with throwing on `null`s too, it doesn't give any extra meaning for the option, and one can just skip providing it (thus making it undefined). If you feel we should allow null here and treat is as undefined, please raise your concern.

Andarist:
> Are the changes to the module transformer required by the split-export-declaration helper? If not, I'd prefer them to be moved to another PR, to keep the changes simpler.

Somewhat, with the changes it produces nicer output without extra temporary variables.

> Nit: can you use a default export? It seems that most helpers do that.

Sure, at first I thought this module might contain more than a single export, but it didnt come to that. Gonna change in a minute.

packages/babel-plugin-proposal-class-properties/src/index.js
Andarist: Good catch. Need to dig into this after work, as I have a feeling that it will break inlining `ClassExpression`s into a class IIFE [here](https://github.com/babel/babel/blob/32f20f85815dce5e90dae5db06d1ae2716bd0f4b/packages/babel-plugin-transform-classes/src/index.js#L67), so I will have to check this and probably fix.

packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js
Andarist: There is no difference when it comes to the functionality.

The containing function here (`getLocalExportMetadata`) gathers export names for local bindings so every future assignment to those local bindings can be transformed into a live reference.

You can note that this function here expects only a FunctionDeclaration, ClassDeclarations and now Identifiers in position of default export. Everything else in this position gets splitted before by `nameAnonymousExports` into
```
var *local* = expression
export { *local* as default }
```
I omit this splitting for local bindings [here](https://github.com/babel/babel/blob/cb4eface89b9837c0ac02f9e3473263b07759e91/packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js#L412-L417) to produce a nicer output, if I'd keep splitting them we would end up with things like:
```js
var _default = function () { /* some class */ }
var _default2 = _default
module.exports = _default2
```

packages/babel-plugin-proposal-class-properties/src/index.js
Andarist: Made a quick test and as suspected - it breaks some cases. Maybe you would have a possible solution/approach in mind to fix this.

The problem is that this:
```js
() => class Bar {
  method(){}
  static fn = () => 'static';
};
```
gets transformed into with class-properties transform
```js
() => {
  var _Bar;
  return _Bar =
  /*#__PURE__*/
  (() => {
    let _Bar = class Bar {
      method() {}
    };
    _Bar.fn = () => 'static';
    return _Bar;
  })();
}
```
which gets later transformed with `transform-classes` into:
```js
() => {
  var _Bar;
  return _Bar =
  /*#__PURE__*/
  (() => {
    function Bar() {}
    var _proto = Bar.prototype;
    _proto.method = function method() {};
    _Bar.fn = () => 'static';
    return _Bar;
  })();
};
```

The problem lies in the fact that the mentioned [inlining](https://github.com/babel/babel/blob/32f20f85815dce5e90dae5db06d1ae2716bd0f4b/packages/babel-plugin-transform-classes/src/index.js#L67) doesn't know that the local class ID is other than that class ID itself.

What comes to my mind is to:
- retrieve this ID - we know for sure in this place that the `parentStatement` (the one that is going to be replaced) is a `VariableDeclaration` containing the correct ID
- rename references to the old ID (constructor's ID) with the retrieved ID (Im not sure how to do this right now, as I'd like to update identifiers in array of nodes and not a path)
- actually do not remove `parentStatement`, but just replace it's `init` with a constructor (from the nodes that we want to insert here), to keep a local ID and a class proper name

nicolo-ribaudo: Probably the right fix is avoiding adding `var [name]` for class expressions.

e.g.
https://babeljs.io/repl/build/6407/#?babili=false&browsers=&build=&builtIns=false&code_lz=BQYwNghgzlAECCsDeAoWsoBcKYJYlgDMB7Y2AXlgCYBuFAXwEog&debug=false&circleciRepo=&evaluate=false&lineWrap=false&presets=stage-0&prettier=false&targets=&version=7.0.0-beta.36%2Bpr.6963

Input:
```js
(class A {
  static foo = 2;
})
```

Expected output
```js
/*#__PURE__*/
(() => {
  let A = class A {};
  Object.defineProperty(A, "foo", {
    configurable: true,
    enumerable: true,
    writable: true,
    value: 2
  });
  return A;
})();

// Which then becomes

/*#__PURE__*/
(function () {
  var A = function A() {
    _classCallCheck(this, A);
  };

  Object.defineProperty(A, "foo", {
    configurable: true,
    enumerable: true,
    writable: true,
    value: 2
  });
  return A;
})();
```

Actual output
```js
var A;

A =
/*#__PURE__*/
(() => {
  let A = class A {};
  Object.defineProperty(A, "foo", {
    configurable: true,
    enumerable: true,
    writable: true,
    value: 2
  });
  return A;
})();

// Which then becomes

var A;

A =
/*#__PURE__*/
function () {
  var A = function A() {
    _classCallCheck(this, A);
  };

  Object.defineProperty(A, "foo", {
    configurable: true,
    enumerable: true,
    writable: true,
    value: 2
  });
  return A;
}();
```

Andarist:
@nicolo-ribaudo Fixed discussed issue by omitting inner var for ClassExpressions and converting them into ClassDeclarations (can be certain about this conversion because we are always inside a class IIFE by this time) + Im generating outer uid for outer ref (as suggested). 

Probably gonna add some test or 2 to be sure that I won't shadow existing bindings with freshly inserted one, this is somewhat tested already, but want to add a more focused test for this + maybe an exec test.

Anyway - I think this is ready for another round of code review.

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -219,22 +239,52 @@ export default function(api, options) {
...
 
         if (computedNodes.length === 0 && afterNodes.length === 0) return;
 
-        if (path.isClassExpression()) {
-          path.scope.push({ id: ref });
-          path.replaceWith(t.assignmentExpression("=", ref, path.node));
+        if (isClassExpression) {
+          const outerRef = path.scope.generateUidIdentifier(ref.name);
+          path.scope.push({ id: outerRef });
+          path.replaceWith(t.assignmentExpression("=", outerRef, path.node));
```
nicolo-ribaudo: Q: What is outer ref needed for? It doesn't conflict now so it works, but I still don't understand what it is for.

Andarist: Im pretty confident it's not needed in 100% cases, it was here before the PR - didnt want to optimize the output more than needed. Gonna revisit this in followup PR.

Is this the only part that you do not understand? Or is something else needing a clarification too?

nicolo-ribaudo: I think it was needed because static properties where defined after the iife: https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&code_lz=BQYwNghgzlAECCsDeAoWsoBcKYJYlgDMB7Y2AXlgCYBuNWAWwFNMALYgE2AEpkBfFH240gA&debug=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=false&presets=es2015%2Cstage-0%2Ctypescript&prettier=false&targets=&version=7.0.0-beta.36 (`_class` is used in `Object.defineProperty`)

> Is this the only part that you do not understand? Or is something else needing a clarification too?

Thank you, the other parts are clear enough :blush:

Andarist:
@nicolo-ribaudo I think you were right, removed creating `outerRef` altogether. Tests seem to be ok - you can check out the latest commit alone. I guess I'm still a little bit too scared about changing existing codebase too much 😄

Andarist:
@loganfsmyth @existentialism @xtuc would love any of yours feedback on this before merging :)

Andarist:
I've rebased this, so it's merge ready - but still needs a second approval 😉

kzc:
@Andarist Can you confirm with this PR to have unreferenced classes with static class properties *with side effects* dropped the [following](https://github.com/babel/babel/pull/6963/files#diff-f4014bbb72e20789116fa27cb9f2c855R5) must be specified?
```js
  "plugins": [
    ["transform-classes", { "assumePure": true }],
  ]
```
By default such otherwise unreferenced classes with static properties with side effects will be retained, correct?

Andarist:
@kzc it's a little bit more complicated that this. I try to detect if the class is pure and if I it is I annotate it no matter what the value of `assumePure` is, this is done [here](https://github.com/babel/babel/pull/6963/files#diff-94bf3502cb4eafb89afa7ae0ae328994R20). `isPure` check would generally be ok to use alone (not considering convoluted cases) only if not inheritance - having a super class means that it cannot be treated as pure automatically (this could be optimized a little bit - but most commonly super declarations will be in different file, so whatever). That's why I had to introduce an additional option of `assumePure`.

When left unspecified it uses an "auto" value - `false` in default mode and `true` in loose mode. It can be specified explicitly to enable/disable annotating all classes (in default/loose modes respectively).

This means that by default when using `loose` mode every class will get PURE annotated, which is potentially unsafe - but I consider this a safe default for `loose` mode.

WDYT about used approach?

kzc:
I personally think all down-leveled classes should be pure by default regardless of `loose` or `strict` mode. But I'm biased. Should get others' opinions on this. Regardless of the Babel default, as long as there is a way to override it to always emit pure classes regardless of static initializer side effects, that would be fine.

Rationale for pure classes by default: 

https://github.com/Microsoft/TypeScript/issues/13721#issuecomment-302565066

Andarist:
This linked rationale lines up perfectly with my reasoning (I've also extended the same reasoning to things other than just classes in [babel-plugin-annotate-pure-calls](https://github.com/Andarist/babel-plugin-annotate-pure-calls)). I just ain't sure what is others opinion and I've went with the safer option by default for now. Happy to change it, I'd like to reach an agreement on this though with the rest of the team.

kzc:
@Andarist If not pure annotated by default, what would a `.babelrc` have to look like to enable pure annotations for classes with static initializers with side effects? I am not well versed in Babel.

nicolo-ribaudo:
IIUC, it would look like this:
```json
{
  "plugins": [
    ["@babel/plugin-proposal-class-properties", { "assumePure": true }]
  ]
}
```

Andarist:
Yeah, exactly - although I feel that we need to introduce option passing through presets to underlaying plugins.

packages/babel-helper-split-export-declaration/src/index.js
```diff
@@ -0,0 +1,65 @@
...
+      ? declaration
+      : t.variableDeclaration("var", [
+          t.variableDeclarator(t.cloneNode(id), declaration.node),
+        ]);
+
+    const updatedExportDeclaration = liveBinding
+      ? t.exportNamedDeclaration(null, [
+          t.exportSpecifier(t.cloneNode(id), t.identifier("default")),
+        ])
+      : t.exportDefaultDeclaration(t.cloneNode(id));
```
loganfsmyth: Even if it doesn't have an ID, cyclic dependencies can still rely on it being hoisted. It's never safe to do `export default foo;` in this case.

packages/babel-helper-class-iife/src/index.js
```diff
@@ -0,0 +1,25 @@
+import { types as t } from "@babel/core";
+import annotateAsPure from "@babel/helper-annotate-as-pure";
+
+const classIIFEs = new WeakMap();
```
loganfsmyth: I generally haven't been super comfortable with things that rely on being singletons like this because it relies on them being the same version and npm flattening things. Usually we've relied on flattening for better performance, but making flattening affect the actual transform output worries me a lot.

Andarist: good point, but without merging both class transforms (which I think @hzoo proposed numerous times) i have no idea how to achieve the desired goal of "pulling" stuff into a single IIFE here

packages/babel-helper-split-export-declaration/src/index.js
Andarist: could you back it up with a sample code? not sure if I understand how the problem can be created

packages/babel-helper-class-iife/src/index.js
Andarist: @loganfsmyth do you have an idea how I could tackle this one differently?

The only thing that comes to my mind is to wrap a class with IIFE in class-properties transform and hardcode some detection logic in transform-classes to check if it should create IIFE or if it should just inline output nodes in place (when it would know that it's already in the IIFE).

It would create indirect dependency on what class-properties does, but even if the detection logic fails it would just create an extra IIFE (so the output still would be valid, just not "optimized").

probablyup:
Hey @Andarist do you need help getting this PR updated? Would love to see it land.

probablyup:
@nicolo-ribaudo is there a tracking ticket / PR I can subscribe to? Thanks!

nicolo-ribaudo:
I suggest just watching this PR

PR #6967, opened 2017-12-04
Generate better builder names for JSX* and TS*

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      |
| Documentation PR         | 
| Any Dependency Changes?  |

`t.jsxIdentifier()` is much nicer and intuitive than `t.jSXIdentifier()`.
I also kept the old function names, so this pr doesn't introduce breaking changes.

### DIFF
--- a/packages/babel-plugin-transform-react-jsx-self/src/index.js
+++ b/packages/babel-plugin-transform-react-jsx-self/src/index.js
@@ -18,10 +18,10 @@ const TRACE_ID = "__self";
 export default function() {
   const visitor = {
     JSXOpeningElement({ node }) {
-      const id = t.jSXIdentifier(TRACE_ID);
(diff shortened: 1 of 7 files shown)


existentialism:
Yessssss :)

xtuc:
Thanks for your PR, i'm :+1: on this but also a bit :-1: since it will introduce another alternative way of doing the same thing.

But I understand we need to preserve the current API.

nicolo-ribaudo:
We could remove `jSX*`before the final Babel 7 release, but I prefer having two ways of doing the same thing instead of an unnecessary breaking change.

PR #6971, opened 2017-12-04
Hoist imports above everything

### LINKED ISSUE #6206: Async function hoisting is overly aggressive
## Bug Report
The transformation of `async function x` to `var x = function () { var _ref = _asyncToGenerator(... ` means that x must be hoisted. This is to avoid issues where `x` is referred to before it is defined. The problem is that the hoisting is overly aggressive and async functions appear before other statements in the code such as requires and imports. This is immediately noticeable if you try to write an async function in a module that sets up `babel-polyfill`/`regenerator-runtime` as a global using a side-effectful require/import. The result is broken code that fails as `regeneratorRuntime` will be undefined.

### Input Code

[Example using `import`](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&circleciRepo=&code_lz=JYWwDg9gTgLgBAcigUwObIHbKgQxtAWigFcMZRkB6EsihAKHpwGcBPDAYzgDNSPyIGOAA84ACgCUcAN704cFDGJQhARnoBfRsMlA&debug=false&evaluate=false&lineWrap=true&presets=env&prettier=false&showSidebar=true&targets=&version=6.26.0)
[Example using `require`](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&circleciRepo=&code_lz=E4UwjgrglqAUDkoDmIB2JgEMAuB7YAtMBKtlALYgD0xpFI8AlAFDOYDOAnqgMYAEAMxI8yuVHwAefWIz4BvZnz6hsEYOICMzAL6sJMoA&debug=false&evaluate=false&lineWrap=true&presets=env&prettier=false&showSidebar=true&targets=&version=6.26.0)

```js
{
  "presets": ["env"]
}
```

### Expected Behavior
The order of statements is preserved in a way that async functions are hoisted only as high as they need to be.

### Current Behavior
Async functions are hoisted to the top of the module including to the point where they break the semantics of the `import` statement by appearing before it.

### Your Environment

| ---------------- | -------
| Babel            |  6.26.0
| node             | 8.4.0
| npm              | 5.3.0
| Operating System | macOS

### LINKED ISSUE #6956: async function hoisted above import 'babel-polyfill' ⇒ ReferenceError: regeneratorRuntime is not defined
### Input Code

```js
import 'babel-polyfill';
async function foo() {}
```

([REPL link](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=JYWwDg9gTgLgBAcgEYEMkFMA2BaSmCeAZsJpggNwBQKAzvgHYDGchArkzMBPSxBABQBKOAG8AvkA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015%2Ces2016%2Ces2017&prettier=false&targets=&version=6.26.0))

```js
{"presets": [["env"]]}
```

### Expected Behavior

The `import` should be evaluated before the translation of the `async function` (which uses `regeneratorRuntime` and `_asyncToGenerator` from `babel-polyfill`).

```js
'use strict';

require('babel-polyfill');

var foo = function () {
  var _ref = _asyncToGenerator( /*#__PURE__*/regeneratorRuntime.mark(…));
  return function foo() {…};
}();

function _asyncToGenerator(fn) {…}
```

### Current Behavior

The current output fails with `ReferenceError: regeneratorRuntime is not defined`.

```js
'use strict';

var foo = function () {
  var _ref = _asyncToGenerator( /*#__PURE__*/regeneratorRuntime.mark(…));
  return function foo() {…};
}();

require('babel-polyfill');

function _asyncToGenerator(fn) {…}
```

Although I’m not an expert on the ES7 specification, I believe this violates [“NOTE: Before instantiating a module, all of the modules it requested must be available.”](https://www.ecma-international.org/ecma-262/7.0/#sec-moduledeclarationinstantiation)

### Possible Solution

Both `import` and `async function` are already hoisted; `import` should simply be hoisted to before hoisted functions.

### Context

This came up before as #5085, which was closed because you can work around it by moving the `async function` to a separate module. But that workaround should not be necessary. If nothing else, it’s an unnecessary trap for new users who probably think they have the wrong preset (I know I did). @loganfsmyth [commented](https://github.com/babel/babel/issues/5085#issuecomment-343375381):

> The interaction here is confusing and has a few cases. That said, either way I agree with you that getting the error your getting, just by declaring the code this way, is poor handling on our part. Changing the behavior of imports in our case would not be the right fix, but there *is* a fix we could do that we haven't done yet, which we probably should.

### Your Environment

| ---------------- | -------
| Babel            | 6.26.0
| node             | 8.9.1
| npm              | 5.5.1
| Operating System | Ubuntu 18.04

nicolo-ribaudo:
| Fixed Issues?            | Fixes #6206, fixes #6956
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

> They had the same precedence of functions, but they should be executed before everything else.

### DIFF
--- a/packages/babel-core/src/transformation/block-hoist-plugin.js
+++ b/packages/babel-core/src/transformation/block-hoist-plugin.js
@@ -32,7 +32,8 @@ const blockHoistPlugin = {
    *  - 1 Default node position
    *  - 2 Priority over normal nodes
    *  - 3 We want this to be at the **very** top
-   *  - 4 Reserved for the helpers used to implement module imports.
+   *  - 4 Imports, which must be available before the code is executed.
+   *  - 5 Reserved for the helpers used to implement module imports.
    */
 
   name: "internal.blockHoist",
--- a/packages/babel-helper-module-imports/src/import-injector.js
+++ b/packages/babel-helper-module-imports/src/import-injector.js
@@ -410,14 +410,14 @@ export default class ImportInjector {
     return resultName;
   }
 
-  _insertStatements(statements, blockHoist = 3) {
+  _insertStatements(statements, blockHoist = 4) {
     statements.forEach(node => {
       node._blockHoist = blockHoist;
     });
 
     const targetPath = this._programPath.get("body").filter(p => {
       const val = p.node._blockHoist;
-      return Number.isFinite(val) && val < 4;
+      return Number.isFinite(val) && val < 5;
     })[0];
 
     if (targetPath) {
--- a/packages/babel-helper-module-transforms/src/index.js
+++ b/packages/babel-helper-module-transforms/src/index.js
@@ -75,7 +75,7 @@ export function rewriteModuleStatementsAndPrepareHeader(
 export function ensureStatementsHoisted(statements) {
   // Force all of the header fields to be at the top of the file.
   statements.forEach(header => {
-    header._blockHoist = 3;
+    header._blockHoist = 4;
   });
 }
 
--- a/packages/babel-plugin-transform-modules-commonjs/test/fixtures/regression/6956-2/actual.js
+++ b/packages/babel-plugin-transform-modules-commonjs/test/fixtures/regression/6956-2/actual.js
@@ -1,0 +1,2 @@
+async function foo() {}
+import 'babel-polyfill';
--- a/packages/babel-plugin-transform-modules-commonjs/test/fixtures/regression/6956-2/expected.js
+++ b/packages/babel-plugin-transform-modules-commonjs/test/fixtures/regression/6956-2/expected.js
@@ -1,0 +1,27 @@
+"use strict";
+
+require("babel-polyfill");
+
+var foo =
+/*#__PURE__*/
+function () {
+  var _ref = _asyncToGenerator(
+  /*#__PURE__*/
+  regeneratorRuntime.mark(function _callee() {
+    return regeneratorRuntime.wrap(function _callee$(_context) {
+      while (1) {
+        switch (_context.prev = _context.next) {
+          case 0:
+          case "end":
+            return _context.stop();
+        }
+      }
+    }, _callee, this);
+  }));
+
+  return function foo() {
+    return _ref.apply(this, arguments);
+  };
+}();
+
+function _asyncToGenerator(fn) { return function () { var self = this, args = arguments; return new Promise(function (resolve, reject) { var gen = fn.apply(self, args); function step(key, arg) { try { var info = gen[key](arg); var value = info.value; } catch (error) { reject(error); return; } if (info.done) { resolve(value); } else { Promise.resolve(value).then(_next, _throw); } } function _next(value) { step("next", value); } function _throw(err) { step("throw", err); } _next(); }); }; }
--- a/packages/babel-plugin-transform-modules-commonjs/test/fixtures/regression/6956-2/options.json
+++ b/packages/babel-plugin-transform-modules-commonjs/test/fixtures/regression/6956-2/options.json
@@ -1,0 +1,4 @@
+{
+  "presets": ["env"],
+  "plugins": []
+}
--- a/packages/babel-plugin-transform-modules-commonjs/test/fixtures/regression/6956/actual.js
+++ b/packages/babel-plugin-transform-modules-commonjs/test/fixtures/regression/6956/actual.js
@@ -1,0 +1,2 @@
+import 'babel-polyfill';
+async function foo() {}
--- a/packages/babel-plugin-transform-modules-commonjs/test/fixtures/regression/6956/expected.js
+++ b/packages/babel-plugin-transform-modules-commonjs/test/fixtures/regression/6956/expected.js
@@ -1,0 +1,27 @@
+"use strict";
+
+require("babel-polyfill");
+
+var foo =
+/*#__PURE__*/
+function () {
+  var _ref = _asyncToGenerator(
+  /*#__PURE__*/
+  regeneratorRuntime.mark(function _callee() {
+    return regeneratorRuntime.wrap(function _callee$(_context) {
+      while (1) {
+        switch (_context.prev = _context.next) {
+          case 0:
+          case "end":
+            return _context.stop();
+        }
+      }
+    }, _callee, this);
+  }));
+
+  return function foo() {
+    return _ref.apply(this, arguments);
+  };
+}();
+
+function _asyncToGenerator(fn) { return function () { var self = this, args = arguments; return new Promise(function (resolve, reject) { var gen = fn.apply(self, args); function step(key, arg) { try { var info = gen[key](arg); var value = info.value; } catch (error) { reject(error); return; } if (info.done) { resolve(value); } else { Promise.resolve(value).then(_next, _throw); } } function _next(value) { step("next", value); } function _throw(err) { step("throw", err); } _next(); }); }; }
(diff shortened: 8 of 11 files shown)


packages/babel-core/src/transformation/block-hoist-plugin.js
```diff
@@ -32,7 +32,8 @@ const blockHoistPlugin = {
    *  - 1 Default node position
    *  - 2 Priority over normal nodes
    *  - 3 We want this to be at the **very** top
-   *  - 4 Reserved for the helpers used to implement module imports.
+   *  - 4 Imports, which must be available before the code is executed.
+   *  - 5 Reserved for the helpers used to implement module imports.
```
Andarist: could we possibly name those blocks and use them as constants through the code? those are really magic numbers and I had no idea for a while what they stand for and where to find info about it quickly when I've started looking though the babel's codebase

hzoo: lol js hoisting is like `z-index`

Andarist: 🤣

nicolo-ribaudo: Do yuo have a good suggestion for a better name for `A_LITTLE_BIT_MORE`?

```js
export const BOTTOM = 0;
export const DEFAULT = 1;
export const A_LITTLE_BIT_MORE = 2;
export const TOP = 3;
export const IMPORT = 4;
export const MODULE_HELPER = 5;
```

nicolo-ribaudo: Also I really don't know where to export that constants from.
`import ... from "@babel/core"`? `from "@babel/traverse"`? `from "@babel/core/lib/transformation/block-hoist-plugin"`? Or maybe `from "@babel/helper-block-hoist"`?

I'm continuing to switch between one option and the others lol.

loganfsmyth:
I already really hate block hoist, so I love to avoid adding another level. The fix I'd propose for this issue would be to change the async function transform to output the current code without giving the function assignment a block hoist, then insert a real function declaration alongside it that calls the assigned variables. It would give the function the proper hoisting, but make it not callable until the function was assigned, does that make sense?

nicolo-ribaudo:
Something like this?
```js
async function fn() {}
```

```js
function fn() {
  fn = _asyncToGenerator(regeneratorRuntime.mark(...));
  return fn();
}
```

So `regeneratorRuntime` isn't used until the function is called.

Andarist:
In the case of `regenerator-runtime` not being initialized yet - I think we should change the default behaviour and insert import statements for the `regenerator-runtime` itself (what `transform-runtime` do).

Relaying on it being in a global namespace as default behaviour always has been off to me.

loganfsmyth:
@nicolo-ribaudo 

> Something like this?

I think we have a few options, depending on how many edge cases we want to handle :) Your nested function proposal is one, and it's probably the most flexible. If we want to go that route, it'd need another tweak though, because something could be relying on `fn` being the same function, so it'd need to be
```
function fn() { return _fn.apply(this, arguments); }
function _fn() {
  var _ref = _asyncToGenerator(
  /*#__PURE__*/
  regeneratorRuntime.mark(function _callee() {
    return regeneratorRuntime.wrap(function _callee$(_context) { /* ... */}, _callee, this);
  }));

  _fn = function() {
    return _ref.apply(this, arguments);
  };

  return _ref.apply(this, arguments);
}
```
so that `fn` is always the same function object.

Given an input like
```
import "babel-polyfill";
console.log("placeholder");
async function fn() {}
```

Other options that come to mind for me personally:

If we accept that we can skip making the function usable in a dependency cycle during initialization, we could just do
```
// if `fn` is exported from the file, programmatically hoist in transform without _blockHoist, but also create a function declaration
var _fn = _asyncToGenerator(
  /*#__PURE__*/
  regeneratorRuntime.mark(function _callee() {
    return regeneratorRuntime.wrap(function _callee$(_context) { /* ... */}, _callee, this);
  })
);

import "babel-polyfill";
console.log("placeholder");

export /* default? */ function fn() { return _fn.apply(this, arguments); }

```
```
// if it's not exported from the file, programmatically hoist in transform without _blockHoist
var fn = _asyncToGenerator(
  /*#__PURE__*/
  regeneratorRuntime.mark(function _callee() {
    return regeneratorRuntime.wrap(function _callee$(_context) { /* ... */}, _callee, this);
  })
);
import "babel-polyfill";
console.log("placeholder");
```

So if we did this and dropped blockHoist from the async fn plugin, you couldn't be able to _call_ `fn` until after the `babel-polyfill` import ran, but you'd at least be able to get a reference to it through a dependency cycle, if you needed to, or call it once the application has loaded, you'd just get a runtime error otherwise.

So really the question that matters is if the added nesting in your example is worth it, vs accepting that the one edge case of dependency cycles through the index that loads `babel-polyfill` would not work.

@andarist I agree with your sentiment, though I have my doubts about if that would be worth blocking 7.x for.

Andarist:
> @Andarist I agree with your sentiment, though I have my doubts about if that would be worth blocking 7.x for.

@loganfsmyth would it really be blocking 7.x? Seems like a simple change in the code that can be done pretty quickly - there are surely more pressing issues but at the same time they need way more time investment and will block 7.x for longer time than this one.

I might be a little bit biased as I'm maintaining also `redux-saga`, generator based redux middleware and I see confusion about this really often. Although ideally `regenerator-transform` would incorporate this change, we should be able to do it on our own in `@babel/plugin-transform-regenerator`

loganfsmyth:
> Seems like a simple change in the code that can be done pretty quickly

Currently Babel makes no assumptions that you even have a module system. If we default to inserting an import, generators are now unusable in `babel-standalone` for instance. We also somehow need to make users expect their code to depend on `regenerator-runtime`, where right now it's easily to say "just load babel-polyfill somehow`.

I'm all for changing how this works, but a lot of it feeds back into bigger questions of how helpers work in general, and how they are loaded, among other things. I think these are changes that should happen, but I think there's a lot more to it than just inserting an import, and if we're going to do that, it should be driven by a coherent approach to modular helpers.

nicolo-ribaudo:
@loganfsmyth I'll go with this:

```js
function fn() { return _fn.apply(this, arguments); }
function _fn() {
  _fn = _asyncToGenerator(
  /*#__PURE__*/
  regeneratorRuntime.mark(function _callee() {
    return regeneratorRuntime.wrap(function _callee$(_context) { /* ... */}, _callee, this);
  }));

  return _fn.apply(this, arguments);
}

```

It's the example you adapted from mine, but without `_ref`. `_ref` was only needed to keep the correct function name and parameters, but since we have full control over the "public interface" (`fn`), it isn't needed. Except for the first call, it uses `_fn.apply()` only once like your second example, so it shouldn't be any slower.

loganfsmyth:
Oh yeah, good call about `_ref`, I hadn't fully through it through and just copy-pasted.

That code sounds good to me. One thing I'll add is that we only need to do  that for function declarations that are nested in `Program`. I think it is safe to just programmatically move the function to the top of functions and blocks, since I can't think of a case where the hoisting would be observable for those? it is only observable in `Program` because of exports.

nicolo-ribaudo:
https://github.com/babel/babel/pull/6984

I hadn't tought that it is only observable in the global scope, I'll push to that pr soon.

PR #6975, opened 2017-12-05
WIP - Do expressions transform for switch statements 

rajasekarm:
| Fixed Issues?            | https://github.com/babel/babel/issues/3872
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |


packages/babel-traverse/src/path/family.js
```diff
@@ -17,6 +17,38 @@ function addCompletionRecords(path, paths) {
...
+  for (let i = 0, caseLen = cases.length; i < caseLen; i++) {
+    const consequentLength = cases[i].get("consequent").length;
+    const isDefaultStatement = !cases[i].get("test").type;
+    if (isDefaultStatement) {
+      paths = addCompletionRecords(
+        cases[i].get("consequent")[consequentLength - 1],
+        paths,
+      );
+    } else {
+      const finalStatement = cases[i].get("consequent")[consequentLength - 1];
```
Andarist: maybe something like `lastCaseStatement`?

Andarist: we should also check if `consequentLength - 1` is a valid index, accessing `-1` may cause deoptimizations in v8

packages/babel-traverse/src/path/family.js
```diff
@@ -17,6 +17,38 @@ function addCompletionRecords(path, paths) {
   return paths;
 }
 
+function completionRecordForSwitch(cases, paths) {
+  for (let i = 0, caseLen = cases.length; i < caseLen; i++) {
+    const consequentLength = cases[i].get("consequent").length;
+    const isDefaultStatement = !cases[i].get("test").type;
```
Andarist: a more explicit check of `cases[i].get("test").type === null` could be used

packages/babel-traverse/src/path/family.js
```diff
@@ -17,6 +17,38 @@ function addCompletionRecords(path, paths) {
   return paths;
 }
 
+function completionRecordForSwitch(cases, paths) {
+  for (let i = 0, caseLen = cases.length; i < caseLen; i++) {
```
Andarist: is there any reason to cache `caseLen` here?

rajzshkr:
@nicolo-ribaudo @Andarist I'll update the PR and keep posted here.

packages/babel-traverse/src/path/family.js
rajzshkr: I would prefer to cache the array length always. Your views please.

Kovensky: This should be a `for (const case of cases) {}` loop. You make no use of the loop variables other than `cases[i]`.

rajzshkr: I changed to `for of`

packages/babel-traverse/src/path/family.js
```diff
@@ -17,6 +17,39 @@ function addCompletionRecords(path, paths) {
   return paths;
 }
 
+function completionRecordForSwitch(cases, paths) {
+  for (const switchcase of cases) {
+    const consequentLength = switchcase.get("consequent").length;
+    const isDefaultCase = switchcase.get("test").type === null;
+    if (isDefaultCase) {
+      paths = addCompletionRecords(
+        switchcase.get("consequent")[consequentLength - 1],
```
Andarist: this code path is not guarded against accessing -1 index

rajzshkr: updated the code to break the loop, if it has zero element

packages/babel-traverse/src/path/family.js
```diff
@@ -17,6 +17,42 @@ function addCompletionRecords(path, paths) {
...
+        paths,
+      );
+    } else {
+      const lastCaseStatement = switchcase.get("consequent")[
+        consequentLength - 1
+      ];
+      const hasBreakStatement =
```
Andarist: as latter IfStatements are both checking against `hasBreakStatement` you could just continue here for `!hasBreakStatement` and remove those checks below

also the mentioned IfStatements should form if/else imho

rajzshkr: Done

Andarist:
@rajzshkr 

I forgot about something [@nicolo-ribaudo's example](https://github.com/babel/babel/pull/6975#pullrequestreview-81329135) should become an `exec.js` test. In fact each of your test could get `exec.js` version. It would help prevent accidental regressions in the future.

rajzshkr:
I've few doubts, after @nicolo-ribaudo comment with the below example,

**Example 1:**

```
var a = do {
  switch(0) {
    case 0: "foo";
    case 1: break;
  }
} || "bar"
```
**Example 2:**
```
var b = do {
  switch(0) {
    case 0: break;
    case 1: "foo"; break;
  }
} || "bar"

alert(a + " " + b);
```
Expected output: `foo bar`
Actual output:` bar foo`

I read the the[ spec of switch case,](https://tc39.github.io/ecma262/#sec-runtime-semantics-caseblockevaluation), so as per spec the code should be transformed as below to get the proper evaluated output.

```
var a = function() {
  switch(0) {
    case 0: "foo";
    case 1: 
       return "foo"; // last evaluated output should be return if `caseclause` encounter a break.
  }
}() || "bar"
```
but it create confusion if there is change in structure of source code. Whats is the possible/right solution?

nicolo-ribaudo:
That transformation wouldn't work, since `switch(1)...` should return undefined. It should become like this.
```js
var a = function() {
  switch(0) {
    case 0: return "foo";
    case 1: return;
  }
}() || "bar"
```
If a case clause's first statement is `break`, it should be replaced with `undefined`, which must be marked as a completion value. Then, if the last statement of the record before isn't `break`, it should also be marked as a completion value.

I think you may need to loop over the case declarations from last to first, otherwise the `break` statement of the case before would already have been removed.

xtuc:
@nicolo-ribaudo @rajzshkr 

### Your example

I simplified a bit your example: [REPL](https://babeljs.io/repl/build/6104/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBCELwQJgewgbwFAQgZwO4EsAXAYwAsAKABgEpUBfNWiAHyYgCIQAjcNtNUSJ1gJk6TLkKlKNMZgQBTAGYgArgBsCALgxzMbRYkS9M9Ri3aduYXmiKIAdlkRr5AOjWIA5mWgBqdgH-nFQA3BBAA&debug=false&circleciRepo=&evaluate=true&lineWrap=true&presets=es2015%2Ces2016%2Ces2017%2Cstage-0&prettier=false&targets=Node-8.9&version=7.0.0-beta.34%2Bpr.6975) (removed the `break` because it was causing confusing I think).

Actual: abar foo
Expected: abar foo 

### Note that:

```js
var a = function () {
  switch (0) {
    case 0:
      "foo";
  }
}();
```

returns `undefined` since there's not break or default case. I don't know what the spec says for this case but since it cannot fall through to another case it should return the value?

rajzshkr:
@xtuc  but as per ECMA spec, last expression in the case block should be returned if there is no break statement.

packages/babel-traverse/src/path/family.js
```diff
@@ -17,6 +17,71 @@ function addCompletionRecords(path, paths) {
...
+        lastCaseStatement && lastCaseStatement.isBreakStatement();
+
+      if (hasBreakStatement) {
```
nicolo-ribaudo: All these checks are the same of lines 62-80?

packages/babel-traverse/src/path/family.js
```diff
@@ -17,6 +17,71 @@ function addCompletionRecords(path, paths) {
...
+      }
+
+      if (hasBreakStatement && consequentLength > 1) {
+        paths = addCompletionRecords(
+          switchcase.get("consequent")[consequentLength - 1],
+          paths,
+        );
+      }
+
+      if (!hasBreakStatement && consequentLength) {
```
nicolo-ribaudo: Only if the first statement of the following `case` is break.

e.g.
```js
let a = do {
  switch(0) {
    case 0:
      "foo";
    case 1:
      "bar";
      break;
  }
};
```

Should be "bar".

xtuc: Yes right.

xtuc:
@rajzshkr so the behavior I explained here https://github.com/babel/babel/pull/6975#issuecomment-350222945 is a bug, do we agree?

rajzshkr:
@xtuc Yup. The above example should output as `foo` instead of `undefined`

rajzshkr:
@yavorsky I've noted this issue also, I'll update the PR.

linonetwo:
@rajzshkr  Will you update this recently?

hackwaly:
Any one can pick his progress?

PR #6984, opened 2017-12-06
Wrap FunctionDeclarations with FunctionDeclarations, instead of using _blockHoist.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #6206, fixes #6956
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

> Use function declarations to wrap declarations.  …
  e.g.
> ```
>  function fn() {}
>  ```
> becomes
>  ```
>  function fn() { return _ref.apply(this, arguments) }
>  function _ref() { _ref = _wrapper(/* ... */); return _ref.apply(this, arguments); }
>  ```
>
> This has two advantages:
>  1) The wrapper isn't executed until the function is invoked. This avoids bugs when transpiled generators are hoisted before the `require("babel-polyfill")` statement.
>  2) Avoid using _blockHoist, and use real JavaScript hoisting.

### DIFF
--- a/packages/babel-helper-wrap-function/src/index.js
+++ b/packages/babel-helper-wrap-function/src/index.js
@@ -3,23 +3,21 @@ import nameFunction from "@babel/helper-function-name";
 import template from "@babel/template";
 import * as t from "@babel/types";
 
-const buildWrapper = template(`
-  (() => {
+const buildExpressionWrapper = template.expression(`
+  (function () {
     var REF = FUNCTION;
     return function NAME(PARAMS) {
       return REF.apply(this, arguments);
     };
-  })
+  })()
 `);
 
-const namedBuildWrapper = template(`
-  (() => {
-    var REF = FUNCTION;
-    function NAME(PARAMS) {
-      return REF.apply(this, arguments);
-    }
-    return NAME;
-  })
+const buildDeclarationWrapper = template(`
+  function NAME(PARAMS) { return REF.apply(this, arguments); }
+  function REF() {
+    REF = FUNCTION;
+    return REF.apply(this, arguments);
+  }
 `);
 
 function classOrObjectMethod(path: NodePath, callId: Object) {
@@ -53,12 +51,12 @@ function plainFunction(path: NodePath, callId: Object) {
   const node = path.node;
   const isDeclaration = path.isFunctionDeclaration();
   const functionId = node.id;
-  let wrapper = buildWrapper;
+  const wrapper = isDeclaration
+    ? buildDeclarationWrapper
+    : buildExpressionWrapper;
 
   if (path.isArrowFunctionExpression()) {
     path.arrowFunctionToExpression();
-  } else if (!isDeclaration && functionId) {
-    wrapper = namedBuildWrapper;
   }
 
   node.id = null;
@@ -70,7 +68,7 @@ function plainFunction(path: NodePath, callId: Object) {
   const built = t.callExpression(callId, [node]);
   const container = wrapper({
     NAME: functionId || null,
-    REF: path.scope.generateUidIdentifier("ref"),
+    REF: path.scope.generateUidIdentifier(functionId ? functionId.name : "ref"),
     FUNCTION: built,
     PARAMS: node.params.reduce(
       (acc, param) => {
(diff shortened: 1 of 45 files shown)


packages/babel-helper-wrap-function/src/index.js
```diff
@@ -88,33 +98,18 @@ function plainFunction(path: NodePath, callId: Object) {
...
-        ]),
-      );
-      return;
-    }
+  });
 
-    path.replaceWith(declar);
+  if (isDeclaration) {
+    if (functionId) {
+      const basePath = path.parentPath.isExportDeclaration()
```
Andarist: seems like the same case as in https://github.com/babel/babel/pull/6963#discussion_r154773120

a helper or a method could be created for future reuse

packages/babel-helper-wrap-function/src/index.js
```diff
@@ -12,7 +12,7 @@ const buildWrapper = template(`
   })
 `);
 
-const namedBuildWrapper = template(`
+const buildNamedExpresionWrapper = template.expression(`
```
citycide: `buildNamedExpresionWrapper` `->` `buildnamedExpressionWrapper`

Andarist: why would you rename this? casing in `buildNamedExpresionWrapper` seems fine to me

nicolo-ribaudo: I forgot an `s`

citycide: Ha yes, of course I make a separate mistake when pointing out the typo though 🤦‍♂️

packages/babel-helper-wrap-function/src/index.js
```diff
@@ -12,7 +12,7 @@ const buildWrapper = template(`
   })
 `);
 
-const namedBuildWrapper = template(`
+const buildNamedExpressionWrapper = template.expression(`
```
Kovensky: Both the `buildNamedExpressionWrapper` and `buildWrapper` templates are, as far as is observable, identical at runtime, so maybe `buildWrapper` should be the `buildExpressionWrapper` (smaller code size). It even correctly accepts a `NAME` parameter.

packages/babel-helper-wrap-function/src/index.js
```diff
@@ -3,7 +3,7 @@ import nameFunction from "@babel/helper-function-name";
 import template from "@babel/template";
 import * as t from "@babel/types";
 
-const buildWrapper = template(`
+const buildWrapper = template.expression(`
   (() => {
```
Kovensky: It might be better to use a plain function instead of arrow function in this case, since you do not depend on `this`. It should have a (negligibly) smaller runtime cost, as well as avoiding the big code size cost when arrow functions are transformed in `spec` mode.

nicolo-ribaudo:
@Kovensky Done

packages/babel-helper-wrap-function/src/index.js
```diff
@@ -3,23 +3,21 @@ import nameFunction from "@babel/helper-function-name";
...
-    var REF = FUNCTION;
-    function NAME(PARAMS) {
-      return REF.apply(this, arguments);
-    }
-    return NAME;
-  })
+const buildDeclarationWrapper = template(`
+  function NAME(PARAMS) { return REF.apply(this, arguments); }
+  function REF() {
+    REF = FUNCTION;
```
Kovensky: I suppose it's ok for the `FUNCTION` to be unnamed. Hopefully nobody uses `arguments.callee`...

I wonder, how would this behave if the wrapped function itself writes to its own identifier? Would it just replace the trampoline function? In any case, looks safe to me.

xtuc:
Let's :pray: the real JavaScript hoisting :smile:

packages/babel-helper-wrap-function/src/index.js
nicolo-ribaudo: > I wonder, how would this behave if the wrapped function itself writes to its own identifier? Would it just replace the trampoline function? In any case, looks safe to me.

It works, it replaces the `function NAME(PARAMS) { return REF.apply(this, arguments); }` function :v: 

https://babeljs.io/repl/build/6145/#?babili=false&browsers=&build=&builtIns=false&code_lz=IYZwngdgxgBAZgV2gFwJYHsLwgCgJQwDeAUDNjALwyiSyIoZZwQBM-RMATgKbIKdYWAbhgBfUl178sARiHFxxAG7BOkkAgA2yEJRgBtALrziNaPCRQ0mLknYkyKtZxl7gAd2Cpk2fPMeqXCxunt6-eCZkwJrcnMg4-lwyADRBMIYwANQwAESUFFSJKTDBGQD8OREKxJx2EUA&debug=false&circleciRepo=&evaluate=true&lineWrap=true&presets=es2017&prettier=false&targets=Node-8.9&version=7.0.0-beta.34%2Bpr.6984

bsuh: The lack of var preceding "REF = FUNCTION;" seems to cause #7134

nicolo-ribaudo: There shouldn't be var, because it needs to replace the REF function.

bsuh: Ok. That was my bad then. I believe I have found the actual problem here https://github.com/facebook/regenerator/issues/225

mweibel:
Is there a way to backport this fix to Babel version 6?

xtuc:
@mweibel probably not, we don't plan to publish Babel 6 anymore.

PR #6985, opened 2017-12-07
Lower env's min firefox version for native generator support

jgrinstead:
Firefox's native generator support was set super high in 4367fba31e5ae79a2882b61cb6950586337a5810 and I have no idea why...

The latest addition I could find in MDN is the return method for the [Generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator) object itself, so I set it to that.

The generator transform is particularly bulky, so bad in fact that I can not use async/await on my own projects, so getting it back would be much appreciated.

### DIFF
--- a/packages/babel-preset-env/data/plugins.json
+++ b/packages/babel-preset-env/data/plugins.json
@@ -201,7 +201,7 @@
   "transform-regenerator": {
     "chrome": "50",
     "edge": "13",
-    "firefox": "53",
+    "firefox": "38",
     "safari": "10",
     "node": "6",
     "ios": "10",


Kovensky:
This is the reason:

image

https://kangax.github.io/compat-table/es6/#test-generators_yield_*,_iterator_closing

xtuc:
Thanks for your contribution @jgrinstead.

@Kovensky so that's incorrect?

jgrinstead:
ugh, what an annoying edge case to hold up async/await support. It would be nice if at least the env preset would listen to the `excludes: ["transform-regenerator"]` field when we target Firefox and use generators.

Kovensky:
A possible alternative is to split the generator support into 2 levels and run a pre-traversal in `@babel/env` that looks for a `yield *` statement (`YieldExpression` node with `path.node.delegate === true`).

If no `yield *` is found, then the generator transform can be skipped for these Firefox versions at least.

PR #6986, opened 2017-12-07
Fix destructuring assignment spec violation

ksashikumar:
ObjectPattern should be checked in toAssignableList

| Fixed Issues?            | https://github.com/babel/babel/issues/6682 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR         | 
| Any Dependency Changes?  |

Spec: https://tc39.github.io/ecma262/#sec-destructuring-assignment

An `AssignmentRestElement` is defined as `... DestructuringAssignmentTarget`. `DestructuringAssignmentTarget` is  `LeftHandSideExpression`, which in assignment contexts can be any `AssignmentPattern` (https://tc39.github.io/ecma262/#sec-destructuring-assignment), so `... ObjectAssignmentPattern` is a valid `AssignmentRestElement`


packages/babylon/src/parser/lval.js
```diff
@@ -150,7 +150,8 @@ export default class LValParser extends NodeUtils {
         if (
           arg.type !== "Identifier" &&
           arg.type !== "MemberExpression" &&
-          arg.type !== "ArrayPattern"
+          arg.type !== "ArrayPattern" &&
+          arg.type !== "ObjectPattern"
```
xtuc: We could probably transform that into a Array lookup for readability, but lgtm. Thanks for your contrib @ksashikumar

xtuc:
That syntax is not allowed in the test262 test suites (https://github.com/babel/babel/pull/6986/files#diff-146d031d23e0df2adf55d70f79b7e12c) ?

nicolo-ribaudo:
@xtuc That tests where whitelisted because they should have passed but they didn't. In fact the travis ci error is

> ✘ 57 valid programs parsed without error (in violation of the whitelist file)

existentialism:
@ksashikumar thanks!

PR #6661, opened 2017-11-01
Remove check constants plugin

maurobringolf:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #5967   
| Patch: Bug Fix?          | No
| Major: Breaking Change?  | No 
| Minor: New Feature?      | No
| Tests Added + Pass?      | Not yet
| Documentation PR         | 
| Any Dependency Changes?  | No

It took me quite some time to figure out where to put the code inside `block-scoping`. I am not 100% sure if it is correct now, but according to the tests it should behave the same as before. But overall tests still fail and I dont understand why yet. As a first step it would be awesome if someone could confirm the placement of the check-constants logic or guide me on how to place it correctly.

PR #6987, opened 2017-12-07
Remove check-constants plugin

### LINKED ISSUE #5967: Remove the `check-es2015-constants` plugin and move code
Continuing from the fix for https://github.com/babel/babel/issues/5728 done in https://github.com/babel/babel/pull/5930, we should move that logic out into the two places it actually needs to be used.

* ~~`transform-es2015-modules-*` - `import`ed variables are constant, so assignments to them should throw~~ (This is now handled in `babel-helper-modules`)
* `transform-block-scoping` - `const` variables are constants, so assignments to them should throw

We currently have a plugin that doesn't really correspond to one specific part of the spec, meaning someone looking in `block-scoping` would probably expect it the handling to be in there, but it is not currently.

maurobringolf:
| Fixed Issues?            | Fixes #5967  
| Documentation PR         | 
| Any Dependency Changes?  |

This is a recreation of #6661 including the recent changes to `check-constants`.

As discussed in the issue, I moved the code from `check-constants` into `block-scoping` including the tests. I also removed all references to the plugin I could find.

I had to pass `state` into the `BlockScoping` class to access `addHelper`. I am not sure if this is the best way to do it, but it seems to work so far.


maurobringolf:
nicolo-ribaudo Sorry for the delay. I'd still like to remove the check-constants plugin so I rebased the PR. The CI passed (98fa9ab) but after I moved the constants check into a separate method it failed due to timeout. Locally it passes.

Any ideas? Does this way trigger redundant constant check runs through scopes and make things slower than before? I thought the `DONE` set prevents that.

existentialism:
@maurobringolf unlikely related to your PR/changes, restarted the build on circle

maurobringolf:
Ahh nice, thank you! I didnt ask because the last time I did, it actually was an error in my code 😁 .

packages/babel-plugin-check-constants/README.md
```diff
@@ -1,58 +0,0 @@
-# @babel/plugin-check-constants
-
-> Validate ES2015 constants (prevents reassignment of const variables).
```
hzoo: next pr: we should maybe include parts of this readme into the blocking scoping one?

maurobringolf: Good point, I can do this later tonight (today for you 😁)

maurobringolf: @hzoo I just added a note for the constants check in the readme #7279. Improvements to wording welcome! Anything else that is needed from a documentation perspective?

PR #6992, opened 2017-12-07
Update to the latest version of regenerator that uses the MIT license

hzoo:
From https://github.com/facebook/regenerator/pull/334


mikesherov:
Awesome! Any chance you could release a patch version of the affected packages instead of waiting till Babel 7?

Issue #6989, opened 2017-12-07
TODO: Remove t.jSX* and t.tS* builders aliases


nicolo-ribaudo:
> EDIT: make this change on the final release.

#6967 makes the `jsx` and `ts` prefixes in node builders lowercase (e.g. `t.jsxExpression()` instead of `t.jSXExpression()`).

We can remove the old `t.jSX*` and `*.tS` builders. It is a breaking change, but
1) `t.tS*` aren't used yet by third party plugins because it has never been in a stable release
2) There aren't many plugins which build JSX nodes (actually, the only one I could find uses `t.JSXIdentifier`, so it would't be influenced by this change).

---

https://github.com/babel/babel/blob/a2aabbd33d373ff9ede75f9d23cdf52249b13165/packages/babel-types/scripts/generators/generateBuilders.js#L18-L23
https://github.com/babel/babel/blob/a2aabbd33d373ff9ede75f9d23cdf52249b13165/packages/babel-types/scripts/generators/generateBuilders.js#L34-L39

existentialism:
Fixed in v8 branch via https://github.com/babel/babel/pull/11095

PR #6993, opened 2017-12-07
Remove functions present only for backwards-compatibility

andy-ms:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Tests Added + Pass?      | Tests pass
| Documentation PR         |

Noticed in DefinitelyTyped/DefinitelyTyped#21972 that there were functions like `tSInterfaceDeclaration` being generated, while `tsInterfaceDeclaration` looks better. It looks like this was done only for backwards-compatibility, but it should be fine to remove this for babel 7, right?

### DIFF
--- a/packages/babel-types/README.md
+++ b/packages/babel-types/README.md
@@ -1761,9 +1761,9 @@ Aliases: `Statement`, `BlockParent`, `Scopable`
 
 ---
 
-### tSAnyKeyword
+### tsAnyKeyword
 ```javascript
-t.tSAnyKeyword()
+t.tsAnyKeyword()
 ```
 
 See also `t.isTSAnyKeyword(node, opts)` and `t.assertTSAnyKeyword(node, opts)`.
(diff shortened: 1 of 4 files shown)


nicolo-ribaudo:
Reference: https://github.com/babel/babel/issues/6989

nicolo-ribaudo:
@xtuc There are also JSX builders, so it is.

xtuc:
@nicolo-ribaudo good point, i've been flying across that PR and haven't seen them.

This is a breaking change, I don't think we can land this in Babel 7. What about warning and removing it in Babel 8?

nicolo-ribaudo:
Well, in Babel 7 we already made different breaking changes in pre releases

hzoo:
We can definetely land a breaking change in the beta, just need to add a changelog entry and add this to the list of things to codemod

xtuc:
lgtm then.

Actually http://babeljs.io/blog/2017/03/01/upgrade-to-babel-7-for-tool-authors#ast-changes-jsx-and-ts-node-builders-from-babel-types-package-renamed

nicolo-ribaudo:
Moved to https://github.com/babel/babel/pull/11095

PR #6997, opened 2017-12-08
WIP - Better error message for duplicated params in arrow functions

rajzshkr:
| Fixed Issues?            | https://github.com/babel/babel/issues/6690
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babylon/src/parser/lval.js
+++ b/packages/babylon/src/parser/lval.js
@@ -331,7 +331,7 @@ export default class LValParser extends NodeUtils {
           const key = `_${expr.name}`;
 
           if (checkClashes[key]) {
-            this.raise(expr.start, "Argument name clash in arrow function");
+            this.raise(expr.start, "Argument name clash in strict mode");
           } else {
             checkClashes[key] = true;
           }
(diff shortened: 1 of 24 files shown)


rajzshkr:
If Argument name clash is not allowed only in strict mode then we can show a explicit message like `Argument name clash in strict mode`. But it is not allowed in both strict or normal. So I think, it is better to have a generic message for both strict or normal.

nicolo-ribaudo:
Two parameters can't have the same name if they are in strict mode or they are in an arrow function (even if it isn't strict).

xtuc:
It's just me or this is now an empty PR?

Also I like the error message Chrome is throwing: `Duplicate parameter name not allowed in this context`.

rajzshkr:
@xtuc I revert the files. Will update shortly.

rajzshkr:
The main aim of this PR is to show relevant error for arrow function and function declaration if it has argument clash

```
"use strict"; var x = (y, y) => y; // SyntaxError: duplicate argument names not allowed in this context
var x = (y, y) => y; // SyntaxError: duplicate argument names not allowed in this context
"use strict"; function test(x,x){} // // SyntaxError: duplicate argument names not allowed in this context
function test(x,x){} // should not throw any error

// Error message copied from V8
```
So, trying to find a function is arrow is difficult and using this.state.potentialArrowAt is not a reliable flag.

So I'm closing this PR now and work on it later.

Issue #6973, opened 2017-12-04
Not-so-pure IIFE async generators stripped out in production.


phyllisstein:
### Choose one: is this a bug report or feature request?
Bug report.

### Input Code
```js
;(async () => {
  await import('app')
})()
```

### Babel/Babylon Configuration (.babelrc, package.json, cli command)
```js
module.exports = {
  presets: [
    ['@babel/env', {
      loose: true,
      modules: process.env.npm_lifecycle_event === 'build:es' ? false : 'commonjs',
      useBuiltIns: 'usage',
    }],
    '@babel/typescript',
    '@babel/react',
    '@babel/stage-0',
  ],
}
```

### Expected Behavior
Babel should transpile the `async` IIFE down such that the `app` module is imported, in both development and production environments.

### Current Behavior
Following #6803, Babel adds a `/*#__PURE__*/` annotation to the output code:

```js
/*#__PURE__*/
_asyncToGenerator(
/*#__PURE__*/
regeneratorRuntime.mark(function _callee() {
  return regeneratorRuntime.wrap(function _callee$(_context) {
    while (1) {
      switch (_context.prev = _context.next) {
        case 0:
          _context.next = 2;
          return import("./src/app");

        case 2:
        case "end":
          return _context.stop();
      }
    }
  }, _callee, this);
}))();
```

In production builds, UglifyJS drops the function by default, as described in its documentation:

> - `side_effects` (default: `true`) -- Pass `false` to disable potentially dropping
>  functions marked as "pure".  A function call is marked as "pure" if a comment
>  annotation `/*@__PURE__*/` or `/*#__PURE__*/` immediately precedes the call. For
>  example: `/*@__PURE__*/foo();`

Pasting the output code into <https://skalman.github.io/UglifyJS-online/> results in its being uglified entirely away. I'm not sure whether this was the behavior intended by the original pull request, but the heuristic that determines a function's purity seems to paint with a pretty broad brush:

<https://github.com/satya164/babel/blob/42c740ca0a009ac4019a339c7f9be9293c93ab8c/packages/babel-helper-remap-async-to-generator/src/index.js#L76-L85>

### Possible Solution
Roll back #6803.

### Context
We use `async` IIFEs to coordinate module loading and webfont availability. It was surprising when we found that the application wouldn't boot at all in production, and took an afternoon's digging to understand that the "pure" annotation was the culprit.

### Your Environment
| software         | version(s)
| ---------------- | -------
[...]

PR #6999, opened 2017-12-08
Async to generator fixes

Kovensky:
| Fixed Issues?            | #6973
| Major: Breaking Change?  | 🤔
| Minor: New Feature?      | 🙅‍♀️

Fixes async arrow functions that are immediately invoked to not annotate the `asyncToGenerator` call as pure. The original fix was suggested on https://github.com/babel/babel/issues/6973#issuecomment-349218627 but modified to also verify whether the callee is the arrow function itself, to make sure values given as arguments in a function call are still annotated as pure. An alternative would be to add parens, but I don't know of any way of doing it without making an invalid tree (e.g. wrapping in a `SequenceExpression` with only one element).

### DIFF
--- a/packages/babel-helper-remap-async-to-generator/src/index.js
+++ b/packages/babel-helper-remap-async-to-generator/src/index.js
@@ -68,6 +68,8 @@ export default function(path: NodePath, file: Object, helpers: Object) {
     wrapAwait: helpers.wrapAwait,
   });
 
+  const isIIFE = path.parentPath.isCallExpression({ callee: path.node });
+
   path.node.async = false;
   path.node.generator = true;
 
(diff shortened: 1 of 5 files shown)


Andarist:
This will conflict with https://github.com/babel/babel/pull/6984 cc @nicolo-ribaudo

Kovensky:
That sounds like fun 😅

Kovensky:
I'll rebase to exclude the commits that changed the helper-wrap-function; the changes in #6984 are better overall, just made some comments.

Kovensky:
Done. This will still conflict with #6984 but the conflict should be very minor now.

PR #7001, opened 2017-12-08
Improve error message in types assert

existentialism:
| Fixed Issues?            | 
| Documentation PR         | 
| Any Dependency Changes?  |

From our slack convo, minor but improves the error message a bit.

### DIFF
--- a/packages/babel-types/scripts/generators/generateAsserts.js
+++ b/packages/babel-types/scripts/generators/generateAsserts.js
@@ -18,7 +18,7 @@ import is from "../../validators/is";
 function assert(type: string, node: Object, opts?: Object): void {
   if (!is(type, node, opts)) {
     throw new Error(
-      \`Expected type "\${type}" with option \${JSON.stringify(opts)}\`,
+      \`Expected type "\${type}" with option \${JSON.stringify(opts)}, but instead got "\${node.type}".\`,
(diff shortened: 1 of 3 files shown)


packages/babel-types/scripts/generators/generateAsserts.js
```diff
@@ -18,7 +18,7 @@ import is from "../../validators/is";
 function assert(type: string, node: Object, opts?: Object): void {
   if (!is(type, node, opts)) {
     throw new Error(
-      \`Expected type "\${type}" with option \${JSON.stringify(opts)}\`,
+      \`Expected type "\${type}" with option \${JSON.stringify(opts)}, but instead got "\${node.type}".\`,
```
nicolo-ribaudo: Can you break this line like in the generated file? Otherwise we will need to `make fix` every time that it is recreated.

existentialism: I can for consistency sake, but the generated files are auto "prettified" via:
https://github.com/babel/babel/blob/master/packages/babel-types/scripts/utils/formatCode.js

nicolo-ribaudo: Ah ok, nevermind then

PR #7008, opened 2017-12-10
expose import.meta syntax parser option as plugin

dnalborczyk:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      |
| Documentation PR         | 
| Any Dependency Changes?  |

this plugin exposes the _import.meta_ syntax parser option as plugin (in the same way as e.g. dynamicImport), which makes it possible to use Babel and _import.meta_ syntax with e.g. [std/esm](https://github.com/standard-things/esm) without specifying the parser options in babelrc:
```
// .babelrc
{
  "parserOpts": {
    "plugins": ['importMeta']
  }
}
```

also added the plugin to the stage 3 preset as well as the babel standalone package.


packages/babel-plugin-syntax-import-meta/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "license": "MIT",
+  "main": "lib/index.js",
+  "keywords": [
+    "babel-plugin"
+  ],
+  "peerDependencies": {
+    "@babel/core": "7.0.0-beta.34"
+  },
+  "devDependencies": {
+    "@babel/core": "7.0.0-beta.34"
```
nicolo-ribaudo: This devDependency isn't needed.

loganfsmyth: @nicolo-ribaudo This should be consistent with out other modules, so having it is the way to go.

packages/babel-plugin-syntax-import-meta/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "version": "7.0.0-beta.34",
+  "description": "Allow parsing of import.meta",
+  "repository": "https://github.com/babel/babel/tree/master/packages/babel-plugin-syntax-import-meta",
+  "license": "MIT",
+  "main": "lib/index.js",
+  "keywords": [
+    "babel-plugin"
+  ],
+  "peerDependencies": {
+    "@babel/core": "7.0.0-beta.34"
```
loganfsmyth: If you look at our other plugins, we've been doing a range for this dependency.

dnalborczyk: I looked through _all_ babel-plugin-syntax packages and some others, it seems like they are using all the same value.

Though I checked the history, it appears that a range has been used prior to _7.0.0-beta.31_

packages/babel-plugin-syntax-import-meta/package.json
Andarist: @loganfsmyth whats the point of having peerDep here then?

packages/babel-plugin-syntax-import-meta/package.json
loganfsmyth: Oh dang, that may have been an accident. I'll have to look into it, but it won't be until next week. Guess if it is already like that this is fine then.

packages/babel-plugin-syntax-import-meta/package.json
loganfsmyth: @andarist A peer just declares a peer without installing anything, so the dev dependency is needed for Lerna to add the symlinks to `@babel/core`. See my other comment though. That peerdep is supposed to be a general range, so it wouldn't normally be a repeat of the dev dependency.

loganfsmyth: @andarist A peer just declares a peer without installing anything, so the dev dependency is needed for Lerna to add the symlinks to `@babel/core`. See my other comment though. That peerdep is supposed to be a general range, so it wouldn't normally be a repeat of the dev dependency.

packages/babel-plugin-syntax-import-meta/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@babel/plugin-syntax-import-meta",
+  "version": "7.0.0-beta.34",
```
hzoo: I think these need to be beta.35 now

existentialism:
I updated the package.json versions to 35, we should note internally to revisit ranges in peerDependencies wrt @babel/core

Issue #2150, opened 2015-08-03
Parentheses around comments


Jigarsolanki:
Why does Babel remove inline parentheses after a comment and adds it around it? 

```javascript 

function test () {
 return /** @type {Object} */ ({});
};
```
converts to 

```javascript
"use strict";

function test() {
  return (/** @type {Object} */{}
  );
};
```

Is this intentional? If yes, why so?

sebmck:
It's due to the way the code generator works. The code generator has no idea what comes after a bit of code once it's pushed to the buffer. It needs to be wrapped in parens to ensure that if there's line terminators that it wont break semantics. The newline before the last `)` probably shouldn't be there, I'll see what can be done about it I guess.

sebmck:
If you have any further questions please don't hesitate to use one of the resources located in the [README](https://github.com/babel/babel#readme):

> For questions and support please visit the <a href="https://babel-slack.herokuapp.com">Slack community</a> or <a href="http://stackoverflow.com/questions/tagged/babeljs">StackOverflow</a>. The Babel issue tracker is <strong>exclusively</strong> for bug reports and feature requests.

Thanks! :blush:

Issue #546, opened 2015-01-20
Closure Compiler annotations


Naddiseo:
Here's an idea for new transformation: adding closure compiler annotations.

sebmck:
Can you give an example?

Naddiseo:
Overview of all annotations: https://developers.google.com/closure/compiler/docs/js-for-compiler
Some cases I can think of:

Classes:
```javascript

class Foo {
    constructor(a: Number) {}
}
class Bar extends Foo {}

// becomes something like:
/** 
@constructor
@param{Number} a
*/
function Foo(a) {}
/**
@constructor
@extends {Foo}
*/
function Bar() {}
```
If type annotations are available:
```javascript
var a : String = 'a';
//becomes
/** @type {string} */
var a = 'a';

function fn(a: Number) :String { return 'num' + a; }
// becomes
/**
@return {string}
@param{Number} a
*/
function fn(a) { return 'num' + a; }
```

Const:
```javascript
const foo = 'FOO';
//becomes
/** @const */
var foo = 'FOO';
```

sebmck:
So JSDoc comment generation with types then, right? Related #220

Naddiseo:
Yeah, similar I suppose. Except that the closure annotations can be inline as well as in the jdoc comment, so there would be less need to parse/merge jdoc comments (which was one of the reasons given for not implementing #220). 

I would give this a go myself, but I don't really understand how comments are handled in the transformations. I attempted to something similar with jshint comments some time ago, but couldn't get comments to show up in the output.

sebmck:
I don't think there's enough of a use-case for this to warrant implementing it. Happy to be proven wrong however, thanks!

lennel:
Closure compiler has tree-shaking, something which is extremely important to a set of people. I would love to use your tool, but given that I can't pass along type information to the closure compiler, which in turn makes the compielr ignorant and not as powerful, the tool sadly will not work for us.

sebmck:
@lennel You can totally pass along type information to the closure compiler. All comments are retained.

lennel:
ah, I misread here someone wanted an automatic plugin. perfect, thanks.

underbyte:
@sebmck I'm trying to get babel and Closure compiler annotations to work together but babel is stripping these parenthesis when using type cast. When type casting the object must be wrap in parenthesis after the comment.

Ex:
```
function test (a,b) {
  /** @type {Object} */ (b);
[...]

PR #7015, opened 2017-12-12
Add retainExtraParens option into babel-generator

koba04:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR         | 
| Any Dependency Changes?  |

This PR is to add a new option `retainParents`, which is to retain parens even if they are meaningless for parsers.
I know this feature is useless for many cases but it's useful for some cases.
(Parens are in there for a reason)

* https://github.com/babel/babel/issues/3946
* https://github.com/babel/babel/issues/546#issuecomment-144861398
* https://github.com/babel/babel/issues/2150

I think this option shouldn't enable by default, but it makes sense to able to enable it by optional.

Thanks!


koba04:
This is the blocker to use Babel in my project.
I want to avoid to use forked Babel because it's a very painful so please consider this PR🙏

loganfsmyth:
My concern with this any time is comes up is that to define an API that we can confidently maintain, we need to know what we are guaranteeing. If a random plugin decides to clear `node.extra` and the parens vanish, how are users supposed to know what happened, for example? Or if we replace the node being cast with another node, none of our current logic preserves `.extra`. Moreover, any existing plugin could introduce a new piece of code to cause that in the future, and I wouldn't consider it a breaking change currently, but with this flag enabled, any change like that could qualify as a breaking change.

For example if a user does

```
import foo from "bar";

var other = /** ... */(foo);
```
and we transform that to

```
var _bar = require("bar");

var other = _bar.default;
```
the `foo` AST node is just gone, and the member expression has no `.extra`.

koba04:
@loganfsmyth Thank you for your feedback!
I've added a commit to reach out your concern.

I thought `comments` also have the same problem but `comments` are passed from an oldNode to a newNode so I've applied it to `parens` as well as `comments`.
Does it make sense?

loganfsmyth:
The issue with emulating comments is that we already don't make any guarantees about comments. Babel is only defined to make a best effort at preserving comments, because comments have no functional affect on runtime code, and people already run into issues with that. They'd likely have the same issues with this, so we'd be introducing a new thing for people to trip on.

loganfsmyth:
To do this "right" given our current structure, we'd have to have an actual AST node for parenthesized expressions, and then every single transform we and anyone else writes would have to make sure to handle it, which would be a mountain of work for everyone involved.

koba04:
@loganfsmyth 
Thank you!
From my perspective, this implementation could cover many cases even though there are no guarantees as you've pointed out.
Do you have any issues or examples for that?
(BTW, `extraParens` might be a better name than `retainParens` as an option name)

I think it's hard to guarantee to work this well against any AST operations.
But I could cover more cases.
Doesn't it make sense to fix the problems at the same time with comments in later?

Is having an actual AST node for Parens(`ParenthesizedExpression`?) the only way to add this feature?
If so, I'll work on this.

koba04:
I've updated this to remove the option from `babel-core` and rename the option name to `retainExtraParens`.

arv:
I'm running into this as well. Removing the parens should really be up to [babel/minify](https://github.com/babel/minify).

arv:
I believe retaining the parens through the transformation passes is always the right thing to do.

Maybe the option should be called `printAllParens` or `printExtraParens`.

arv:
I want us to make progress on this. Here is my proposal:

- Create a new PR that contains the code to `inheritParens` so that the parens are not lost in transformations. (This should be non controversial)
- After that we can go a few ways:
  - [ ] Add a flag to print extra parens?
  - [ ] Always print extra parens after a `/** */` comment (when printing comments)?
  - [ ] Add a closure compiler flag?

@koba04 @loganfsmyth WDYT?

PR #7017, opened 2017-12-12
Fixes 6965

### LINKED ISSUE #6965: reduce code generated of plugin-proposal-decorators
plugin-proposal-decorators create for each component using decorator same function template:

applyDecoratedDescriptor
and 
initDefineProp

so the code ended is much more bigger, 
i propose to create only this once in global what do you think?
(It s what i did inside babelHelper ...)

same this plugins use the function ensureInitializerWarning who create no useful bit for each component too:
function NAME(descriptor, context){ throw new Error(    'Decorating class property failed. Please ensure that proposal-class-properties is enabled and set to use loose mode.  To use proposal-class-properties in spec mode with decorators, wait for he next major version of decorators in stage 2

will be better to add it in global too ...

perinikhil:
| Fixed Issues?            | Fixes #6965 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -701,8 +701,12 @@ helpers.toPropertyKey = defineHelper(`
   }
 `);
 
-helpers.buildInitializerWarningHelper = defineHelper(`
-    export default function _buildInitializerWarningHelper(descriptor, context){
+/**
+ * Add a helper that will throw a useful error if the transform fails to detect the class
+ * property assignment, so users know something failed.
+ */
+helpers.initializerWarningHelper = defineHelper(`
+    export default function _initializerWarningHelper(descriptor, context){
         throw new Error(
           'Decorating class property failed. Please ensure that ' +
           'proposal-class-properties is enabled and set to use loose mode. ' +
@@ -712,8 +716,11 @@ helpers.buildInitializerWarningHelper = defineHelper(`
     }
 `);
 
-helpers.buildInitializerDefineProperty = defineHelper(`
-    export default function _buildInitializerDefineProperty(target, property, descriptor, context){
+/**
+ * Add a helper to call as a replacement for class property definition.
+ */
+helpers.initializerDefineProperty = defineHelper(`
+    export default function _initializerDefineProperty(target, property, descriptor, context){
         if (!descriptor) return;
 
         Object.defineProperty(target, property, {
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-proposal-decorators/src/index.js
```diff
@@ -26,75 +26,18 @@ const buildGetObjectInitializer = template(`
...
   function ensureApplyDecoratedDescriptorHelper(path, state) {
     if (!state.applyDecoratedDescriptor) {
       state.applyDecoratedDescriptor = path.scope.generateUidIdentifier(
         "applyDecoratedDescriptor",
       );
-      const helper = buildApplyDecoratedDescriptor({
-        NAME: state.applyDecoratedDescriptor,
-      });
+      const helper = state.addHelper("buildApplyDecoratedDescriptor");
       path.scope.getProgramParent().path.unshiftContainer("body", helper);
```
nicolo-ribaudo: The problem is that these lines:
```js
const helper = buildApplyDecoratedDescriptor({
  NAME: state.applyDecoratedDescriptor,
});
path.scope.getProgramParent().path.unshiftContainer("body", helper);
```
added the helper function to the file.

On the other hand, `state.addHelper("buildApplyDecoratedDescriptor");` returns the id of the function (not the function itself), and the helper is added automatically.

So you should just remove the `ensure*` functions (which are the equivalent of `state.addHelper`), and replace calls to them (like https://github.com/babel/babel/blob/f5ef928586f592aa2d8bb60120ae58833d17e0db/packages/babel-plugin-proposal-decorators/src/index.js#L273) with `addHelper(" ... ")`.

perinikhil:
Thank you @xtuc and @nicolo-ribaudo for your valuable inputs and patience! Much obliged 😃 
I've also raised a docs PR which could be reviewed as well...
And also, i'd like to pick up another issue, any suggestions?

xtuc:
Feel free to merge @nicolo-ribaudo if it looks good to you.

@perinikhil, either:
- Priority stuff https://github.com/babel/babel/labels/Priority%3A%20High
- Beginner-friendly https://github.com/babel/babel/labels/good%20first%20issue

packages/babel-helpers/src/helpers.js
```diff
@@ -700,3 +700,70 @@ helpers.toPropertyKey = defineHelper(`
...
+`);
+
+/**
+ * Add a helper to take an initial descriptor, apply some decorators to it, and optionally
+ * define the property.
+ */
+helpers.applyDecoratedDescriptor = defineHelper(`
+    export default function _applyDecoratedDescriptor(target, property, decorators, descriptor, context){
+        var desc = {};
+        Object['ke' + 'ys'](descriptor).forEach(function(key){
```
TrejGun: what is the problem to use `Object.keys(descriptor)` ?

nicolo-ribaudo: It is kept because it was in the original helper, I will check

xtuc: Because of transform-runtime, same here https://github.com/babel/babel/pull/7017/files/805c8f7cf2dc6eea9d3dcb85b55acc6a6a3334e5#diff-ba75bd4d116cc90ad8c8002e3010e821R762

PR #4850, opened 2016-11-16
Remove ClassCallCheck, possibleConstructorReturn in loose mode

hzoo:
| Q                                   | A  
| -----------------------| ---
| Bug fix?                        | 
| Breaking change?        | 
| New feature?                | yes
| Deprecations?              | 
| Spec compliancy?        | 
| Tests added + pass?    | yes
| Fixed tickets                 |
| License                         | MIT
| Doc PR                          | 
| Dependency Changes | 

cc @developit

PR #3527, opened 2016-06-15
Fix class inheritance in IE <=10 (T3041)

danez:
### Problem summary

Internet Explorer 10 and below **do not support** `__proto__` at all, they **don't have** `Object.setPrototypeOf()`, but they **have** `Object.getPrototypeOf()`.

When babel is doing the inheritance it is trying to use `Object.setPrototypeOf()` if this function is available. As mentioned this function isn't available (even with core-js, as also core-js can't polyfill this). In this case the helper falls back to just [set `__proto__`](https://github.com/babel/babel/blob/master/packages/babel-helpers/src/helpers.js#L223) on the object. For IE <=10 this is just some custom property that gets set on the target, with no special meaning.

In the constructor after the inherit helper finished its work, babel [uses `Object.getPrototypeOf()`](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-es2015-classes/src/vanilla.js#L349) to get the prototype. This is supported by IE, but as babel can't set the prototype on the target, this will just return the base Object-prototype and not the wanted superclass.

### Proposed solution

The solution simply changes the retrieval of the prototype by first trying to get `__proto__` and if not available falling back to `Object.getPrototypeOf()`. This fixes the behavior in IE as babel reads the custom prop `__proto__` that the inherit helper was setting. In all other browsers this should not have any effect, as `__proto__` should be the same as `Object.getPrototypeOf()`.

Does anyone see a problem with this solution? Anything that I maybe have missed?

~~P.S.: I did not yet change any of the tests, as I wanted to get feedback first on the solution.~~

PR #7020, opened 2017-12-12
Add support for extending builtins

### LINKED ISSUE #4480: Native extends breaks HTMLELement, Array, and others
Current Babel transform, when it comes to call the parent

``` js
function _possibleConstructorReturn(self, call) {
  if (!self) {
    throw new ReferenceError("this hasn't been initialised - super() hasn't been called");
  }
  return call && (typeof call === "object" || typeof call === "function") ? call : self;
}
```

It's a way too poor implementation.

If we take the current basic ES6 compat syntax:

``` js
class List extends Array {}
```

We'll realize babel does a bad job.

``` js
var l = new List();
l instanceof List; // false
```

The reason is simple: Babel replace the returns and exit, without caring about userland expectations.

This is how above basic extend should desugar:

``` js
function List() {
  return Object.setPrototypeOf(
    Array.apply(this, arguments),
    List.prototype
  );
}
```

It's a very ad-hoc case for the initial example that currently fails, but it's good enough to understand that inheriting the prototype is the least of the problem.

Indeed, we have 3 ways to do that within a transpiled code:

``` js
// losing the initial prototype
// for polyfilled ES5+ or lower
List.prototype = Object.create(
  Array.prototype,
  {constructor: {
    configurable: true,
    writable: true,
    value: List.prototype
  }}
);

// using a cleaner way
// for ES5+ compliant targets
Object.setPrototypeOf(
  List.prototype,
  Array.prototype
);

// using a cleaner way
// that works well with
// partially polyfilled .setPrototypeOf
List.prototype = Object.setPrototypeOf(
  List.prototype,
  Array.prototype
);
```

Solved the inheritance bit, considering Babel also set the prototype of each constructor,
we need to address cases where a `super` call might "_upgrade_" the current context, like it is for `HTMLELement` or exotic native objects.

``` js
// native ES6 static example
class List extends Array {
  constructor(a, b, c) {
    super(a, b);
    this.push(c);
  }
}
```

Above case should desugar to something like the follwoing:

``` js
function List(a, b, c) {
  // the super bit
  var self = Object.setPrototypeOf(
    Array.call(this, a, b),
    List.prototype
  );
  // the rest with swapped context
  self.push(c);
  // at the end
  return self;
}
```

which is also ad-hoc example code for the previous example.

Considering a transpiler will get the arguments part easily right, this is how previous case could be generically transpiled for arguments used in both constructors:

``` js
// to make it as generic as possible
function Child() {

  // the super call bit desugar to ...
  var
    instance = Object.getPrototypeOf(Child)
        .apply(this, arguments),
    type = instance ? typeof instance : '',
    // if Parent overwrote its default return
    self = (type === 'object' || type === 'function') ?
      // upgrade the instance to reflect Child constructor
      Object.setPrototypeOf(instance, Child.prototype) :
      // otherwise use the current context as is
      this
  ;

  // at this point, the rest of the constructor
  // should use `self` instead of `this`
  self.push(c);

  // and return `self` reference at the end
  return self;
}
```

The last problem is that modern syntax would use [Reflect](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect) to create any sort of object, instead of old, ES3 friendly, `.call` or `.apply` way.

Following a past, present, and even future proof approach:

``` js
var
  reConstruct = typeof Reflect === 'object' ?
    Reflect.construct :
    function (Parent, args, Child) {
      return Parent.apply(this, args);
    }
;

function Child() {

  // the super call bit
  var
    instance = reConstruct.call(
      this,
      Object.getPrototypeOf(Child),
      arguments,
      Child
    ),
    type = instance ? typeof instance : '',
    self = (type === 'object' || type === 'function') ?
      Object.setPrototypeOf(instance, Child.prototype) :
      this
  ;

  // the rest of the constructor body
  self.push(c);

  // at the end or instead of empty returns
  return self;
}
```

Above solution would work with userland, exotic, or DOM constructors, and in both native and transpiled engines.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #4480
| Patch: Bug Fix?          | :tada: 
| Major: Breaking Change?  |
| Minor: New Feature?      | Yes maybe?
| Documentation PR         | 
| Any Dependency Changes?  | Yes. I added `globals` to `plugin-transform-classes`, but it should already be in users' node_modules folders because it is used by `@babel/traverse`

This PR adds the helper used by https://github.com/WebReflection/babel-plugin-transform-builtin-classes to `@babel/plugin-transform-classes`.
Browser and ECMAScript builtins are both wrapped with the helper.

---

**OLD MESSAGE**

To make the user-experience as nice as possible, I enabled it *by default* when extending a class defined in the ES specification. If the user wants to extend another class (for example, a DOM class), it can be declared using the `builtins` option.

I hardcoded the list of es classes, but in the future I'd like to ask to the [`globals`](https://github.com/sindresorhus/globals) mantainer if it can be stored in that package (we are already using it in `babel-traverse`).

Last thing, I had to disable [a test](https://github.com/nicolo-ribaudo/babel/blob/3316f8581f9ff278663d8a07367428603afcf50b/packages/babel-preset-es2015/test/fixtures/traceur/Classes/.ClassMethodInheritance.js) and I don't understand why; I'll search the problem tomorrow because now I'm too sleepy.

cc @WebReflection


packages/babel-plugin-transform-classes/README.md
```diff
@@ -119,3 +119,37 @@ class Bar extends Foo {
...
+to be wrapped.
+Babel only recognizes built-in JavaScript functions defined in the ECMAScript
+specification. If you need to extend other classes, you can define them using
+this option.
+
+#### Compatibility
+
+This transformer works on IE11 and every other browser with `Object.setPrototypeOf` or `__proto__` as fallback.
+
+There is **NO IE <= 10 support**. If you need IE <= 10 don't it's recommended to not don't extend natives.
```
WebReflection: wow, I made it super clear there 😂

> If you need IE <= 10 don't it's recommended to not don't extend natives.

Andarist: grammar "to not don't"

nicolo-ribaudo: I removed the `don't use this plugin` bit and screwed it up :laughing:

Andarist:
I have not caught up with the discussion in the other thread - what's the downside of doing inheritance buble-style?
```js
var MyError = (function (Error) {
  function MyError () {
    Error.apply(this, arguments);
  }if ( Error ) MyError.__proto__ = Error;
  MyError.prototype = Object.create( Error && Error.prototype );
  MyError.prototype.constructor = MyError;
  return MyError;
}(Error))
```

Maybe such simpler transform could be used in `loose` mode?

hzoo:
We already changed the loose mode transform to be simpler? https://github.com/babel/babel/pull/4850 unless you mean even more so

nicolo-ribaudo:
@Andarist Some builtins (like `HTMLElement`, which is used when creating custom elements) don't like being `.apply`ed.

```
> HTMLElement.apply({}, [])

Uncaught TypeError: Failed to construct 'HTMLElement': Please use the 'new' operator, this DOM object constructor cannot be called as a function.
    at <anonymous>:1:13
```

WebReflection:
@Andarist **many** constructors won't work there.

`Uint8Array`, as well as `String`, or `HTMLElement` (and every other from DOM) or `CustomEvent` and the list go on for mostly every builtin.

WebReflection:
@nicolo-ribaudo the issue in the test is probably [here](https://github.com/nicolo-ribaudo/babel/blob/3316f8581f9ff278663d8a07367428603afcf50b/packages/babel-preset-es2015/test/fixtures/traceur/Classes/.ClassMethodInheritance.js#L15).

Checking `Object.getPrototypeOf(D) === Object` fails because the patch/fix needs to put [an intermediate constructor in between](https://github.com/WebReflection/babel-plugin-transform-builtin-classes/blob/master/src/index.js#L22-L24).

I think that `assert.equal(Object.getPrototypeOf(D), Object);` should be `assert.isTrue(D instanceof Object);` or, if you want to be sure about the intermediate constructor,  `assert.equal(Object.getPrototypeOf(Object.getPrototypeOf(D)), Object);`

WebReflection:
@nicolo-ribaudo you also flipped the wrap, targeting directly the builtin instead of the declared class. I think that's OK but then you don't need a `WeakMap`, just a `Map`, 'cause there's no way a builtin will suddenly disappear from the engine or it'll be Garbage Collected.

That means that using `Map` you'll have wider compatibility out of the box.

packages/babel-plugin-transform-classes/src/builtin-classes.js
```diff
@@ -0,0 +1,35 @@
...
+  "ArrayBuffer",
+  "Boolean",
+  "DataView",
+  "Date",
+  "Error",
+  "EvalError",
+  "Float32Array",
+  "Float64Array",
+  "Function",
+  "In16Array",
```
TimothyGu: typo

packages/babel-plugin-transform-classes/src/builtin-classes.js
```diff
@@ -0,0 +1,35 @@
...
+  "Int32Array",
+  "Int8Array",
+  "Map",
+  "Number",
+  "Object",
+  "Promise",
+  "Proxy",
```
TimothyGu: Is it worth it to include this here when `Proxy` can't get subclassed anyway?

WebReflection: I don't think it makes any sense to maintain this list manually.

This is what I've used to [include them all](https://github.com/WebReflection/babel-plugin-transform-builtin-classes/blob/master/.babelrc), something you can produce on any browser via:

```js
JSON.stringify(
  Object.getOwnPropertyNames(window)
    .filter(name => /^[A-Z]/.test(name))
    .sort(),
  null,
  '  '
)
```
so that from time to time one can run that again and update the list.

nicolo-ribaudo: I initially did something like this, but then it would include things like `Math`, `Infinity`...
I will ask to include this list in the [Globals](https://github.com/sindresorhus/globals) package.

nicolo-ribaudo:
Thank you all for the quick feedback, they are really appreciated. I'll update the PR tomorrow.

packages/babel-plugin-transform-classes/src/builtin-classes.js
WebReflection: Well, that filter might include typeof “function” too

packages/babel-helpers/src/helpers.js
```diff
@@ -425,6 +425,46 @@ helpers.inheritsLoose = defineHelper(`
...
+    var Constructor, a = [null];
+    a.push.apply(a, args);
+    Constructor = Parent.bind.apply(Parent, a);
+    return _sPO(new Constructor, Class.prototype);
+  };
+
+  var _cache = typeof WeakMap === "function" && new WeakMap();
+
+  export default function _wrapNativeSuper(Class) {
+    if (_cache) {
```
xtuc: Could this check be more explicit? `typeof x === 'undefined'`

packages/babel-plugin-transform-classes/README.md
xtuc: Also not that IE is not official supported by Babel (according to https://github.com/babel/babel/blob/master/doc/design/compiler-environment-support.md)

xtuc: It sounds strange to me, could we reword it to something like:

> We don't recommend to extend native classes under IE (version <= 10) because it can cause undefined behaviors.

packages/babel-helpers/src/helpers.js
```diff
@@ -425,6 +425,46 @@ helpers.inheritsLoose = defineHelper(`
   }
 `);
 
+// Based on https://github.com/WebReflection/babel-plugin-transform-builtin-classes
+helpers.wrapNativeSuper = defineHelper(`
+  var _gPO = Object.getPrototypeOf || function _gPO(o) { return o.__proto__ };
+  var _sPO = Object.setPrototypeOf || function _sPO(o, p) { o.__proto__ = p };
+  var _construct = Reflect.construct || function _construct(Parent, args, Class) {
```
xtuc: Shouldn't we check for Reflect first?

packages/babel-plugin-transform-classes/test/fixtures/extend-builtins/loose/expected.js
```diff
@@ -0,0 +1,23 @@
...
+
+var _gPO = Object.getPrototypeOf || function _gPO(o) { return o.__proto__; };
+
+var _sPO = Object.setPrototypeOf || function _sPO(o, p) { o.__proto__ = p; };
+
+var _construct = Reflect.construct || function _construct(Parent, args, Class) { var Constructor, a = [null]; a.push.apply(a, args); Constructor = Parent.bind.apply(Parent, a); return _sPO(new Constructor(), Class.prototype); };
+
+var _cache = typeof WeakMap === "function" && new WeakMap();
+
+function _wrapNativeSuper(Class) { if (_cache) { if (_cache.has(Class)) return _cache.get(Class); _cache.set(Class, Wrapper); } function Wrapper() {} Wrapper.prototype = Object.create(Class.prototype, { constructor: { value: Wrapper, enumerable: false, writeable: true, configurable: true } }); return _sPO(Wrapper, _sPO(function Super() { return _construct(Class, arguments, _gPO(this).constructor); }, Class)); }
```
Kovensky: Don't need `enumerate: false` (default, and if omitted, core-js/es5-sham can avoid being forced to throw on ES3 engines)

packages/babel-helpers/src/helpers.js
WebReflection: agreed, and original plugin was doing it indeed:
https://github.com/WebReflection/babel-plugin-transform-builtin-classes/blob/master/src/index.js#L8

packages/babel-plugin-transform-classes/test/fixtures/extend-builtins/loose/expected.js
WebReflection: true that it's `false` by default but also true that ES3 engines *should* throw if you extend builtins because ES3 engines are not supported. IE <= 10 is discouraged and that's already mostly ES5

nicolo-ribaudo: The `enumerable: false` is also used by the `_inherits` helper.
Also, ES3 engines don't support `setPrototypeOf`/`__proto__`, so classes won't work in any case.

packages/babel-helpers/src/helpers.js
nicolo-ribaudo: I'll use `(typeof Reflect === "object" && Reflect.construct) || function ...` instead of the ternary (which is used in the original plugin) since someone might only polyfill *some* Reflect methods, so `Reflect.constructor` could be not defined.

packages/babel-plugin-transform-classes/src/builtin-classes.js
```diff
@@ -9,7 +9,7 @@ export default new Set([
   "Float32Array",
   "Float64Array",
   "Function",
```
WebReflection: I'm pretty sure having at least `HTMLElement` in this list by default would make many people happy

nicolo-ribaudo: I'm okay with it, but I'd like to wait for other opinions (since currently Babel is only about ECMAScript, and not about DOM extensions)

WebReflection: I think it's a reasonable exception to the rule, specially considering Babel is used the most for the Web.

packages/babel-plugin-transform-classes/README.md
```diff
@@ -119,3 +119,37 @@ class Bar extends Foo {
 When `Bar.prototype.foo` is defined it triggers the setter on `Foo`. This is a
 case that is very unlikely to appear in production code however it's something
 to keep in mind.
+
+### `builtins`
+
+`Array<string>`, defaults to `[]`.
+
+When extending a native constructor (e.g., `class extends Array {}`), it needs
+to be wrapped.
```
justinfagnani: A couple of changes I'd reccomend:

 * Change "constructor" to "class" (since the key distinction here is constructors that can't be called with `Function.apply`, which are class constructors) and mention a little about w
 * Clarify what needs to be wrapped: the subclass or the native class
 * Add just a little hint on why the wrapper is necessary 

"When extending a native class (e.g., `class extends Array {}`), the native class needs to be wrapped in order to support Babel's emulation of `super()` calls."

packages/babel-plugin-transform-classes/README.md
```diff
@@ -119,3 +119,37 @@ class Bar extends Foo {
...
+Babel only recognizes built-in JavaScript functions defined in the ECMAScript
```
justinfagnani: Does this option override or augment the default list? If it augments, I'd consider making that more clear in the option name.

packages/babel-plugin-transform-classes/src/vanilla.js
```diff
@@ -93,6 +95,12 @@ export default class ClassTransformer {
 
     this.superName = this.node.superClass || t.identifier("Function");
     this.isDerived = !!this.node.superClass;
+
+    const { name } = this.superName;
+    this.extendsNative =
+      this.isDerived &&
+      (builtinClasses.has(name) || customBuiltins.has(name)) &&
```
justinfagnani: What happens when the extension of the native is indirect due to a mixin application? ie:

```js
class MyElement extends PolymerElement(HTMLElement) {}
```

Can we have a way to force the wrapping of a particular set of builtins?

packages/babel-plugin-transform-classes/src/builtin-classes.js
justinfagnani: I would also add `Event` and `CustomEvent`. Considering that these classes aren't wrapped until used, this doesn't seem to add more cost to the transform. I think this transform will be most popular with web components users, so having a useful default seems better to me.

packages/babel-helpers/src/helpers.js
```diff
@@ -425,6 +425,47 @@ helpers.inheritsLoose = defineHelper(`
...
+      var Constructor, a = [null];
+      a.push.apply(a, args);
+      Constructor = Parent.bind.apply(Parent, a);
+      return _sPO(new Constructor, Class.prototype);
+    };
+
+  var _cache = typeof Map === "function" && new Map();
+
+  export default function _wrapNativeSuper(Class) {
+    if (typeof _cache !== "undefined") {
```
justinfagnani: It'd be really nice if this didn't wrap emulated classes which already have callable constructors, like when using the Custom Elements polyfill. In that case `HTMLElement` is callable because the polyfill wraps the native class. Detecting this case would avoid double-wrapping.

It's unfortunately impossible to absolutely detect that a constructor function is callable, but there's an approximation that might be suitable here:

```js
function isNative(ctor) {
  return ctor.toString().indexOf('[native code]') !== -1;
}
```

packages/babel-plugin-transform-classes/src/vanilla.js
WebReflection: an easy work around could be to:
```js
class MyElement extends PolymerElement(class extends HTMLElement {}) {}
```

I don't think it's that simple to handle every time a constructor is passed around and in a meaningful way.

mixins also can be passed around as regular invoke.

I mean ... this is valid code too, right?
```js
const mixed = (E => () => PolymerElement(E))(window['HTMLElement']);
class MyElement extends mixed() {}
```
We need a compromise for common cases and IMO these are not necessarily based on Polymer.

justinfagnani: The most common case in the wild is that `HTMLElement` is extended via mixin application. Without support for that, this transform will be much less used than it could be.

This is why I suggest a way to force wrapping, possibly like:

```json
{
  "plugins": [
    ["@babel/transform-classes", {
      "builtins": [ "HTMLElement" ],
      "forceWrap": [ "HTMLElement" ],
    }]
  ]
}
```

WebReflection: > The most common case in the wild is that HTMLElement is extended via mixin application. 

I'd like to see data about this. I use CE since ever and never used mixins there. It's also worth noticing that Babel supposed to be about ES, not DOM, as previously mentioned.

Every constructor is broken right now so this is already quite a big step forward.

> This is why I suggest a way to force wrapping, possibly like:

I'm a bit confused by the fact you don't want to double wrap polyfills for Custom Elements but you want to force wrap `HTMLElement` which basically means the transpiled code will wrap already that class ...

how does it work? You either force wrap, or you avoid runtime, when happens, the double wrap.

Unless you are thinking about having the polyfill before the transpiled code that uses that constructor ... is that the case?

packages/babel-plugin-transform-classes/test/fixtures/extend-builtins/loose/expected.js
Kovensky: The use of `__proto__` was actually added to the helpers specifically so it would work on ES3 engines that were polyfilled.

The idea was that, even if the engine doesn't support it, as long as we read from `__proto__` first, we're going to get a reference to the relevant constructor, which is all that we need for calling super. static property inheritance will still, of course, be broken.

https://github.com/babel/babel/pull/3527

packages/babel-plugin-transform-classes/src/index.js
```diff
@@ -5,9 +5,22 @@ import nameFunction from "@babel/helper-function-name";
 import { types as t } from "@babel/core";
 
 export default function(api, options) {
-  const { loose } = options;
+  const { loose, builtins } = options;
   const Constructor = loose ? LooseTransformer : VanillaTransformer;
 
+  let customBuiltins;
+  if (builtins) {
```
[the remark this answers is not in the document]
nicolo-ribaudo: It should be `!== undefined`

nicolo-ribaudo:
Currently this PR only wraps globals. After thinking a bit about it, I come to the conclusion that classes defined by the `builtins` option (which would need to be renamed to something like `additionalWrap`) should also work for non-global variables. That would cover, for example https://github.com/babel/babel/issues/7022.

packages/babel-helpers/src/helpers.js
WebReflection: I don't think we need to catch user defined classes with a function called `isNative` ...

packages/babel-plugin-transform-classes/src/vanilla.js
```diff
@@ -61,7 +63,7 @@ const findThisesVisitor = traverse.visitors.merge([
 ]);
 
 export default class ClassTransformer {
-  constructor(path: NodePath, file) {
+  constructor(path: NodePath, file, customBuiltins: Set<string>) {
```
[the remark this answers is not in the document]
nicolo-ribaudo: Unfortunately the only read-only type is `ReadonlyAray`

packages/babel-helpers/src/helpers.js
```diff
@@ -425,6 +425,47 @@ helpers.inheritsLoose = defineHelper(`
   }
 `);
 
+// Based on https://github.com/WebReflection/babel-plugin-transform-builtin-classes
+helpers.wrapNativeSuper = defineHelper(`
+  var _gPO = Object.getPrototypeOf || function _gPO(o) { return o.__proto__ };
```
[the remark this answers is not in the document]
WebReflection: this patch is explicitly incompatible with IE <= 10 where you should *not* extend native builtins

packages/babel-helpers/src/helpers.js
```diff
@@ -425,6 +425,47 @@ helpers.inheritsLoose = defineHelper(`
...
+  var _sPO = Object.setPrototypeOf || function _sPO(o, p) { o.__proto__ = p };
+  var _construct = (typeof Reflect === "object" && Reflect.construct) ||
+    function _construct(Parent, args, Class) {
+      var Constructor, a = [null];
+      a.push.apply(a, args);
+      Constructor = Parent.bind.apply(Parent, a);
+      return _sPO(new Constructor, Class.prototype);
+    };
```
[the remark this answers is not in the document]
WebReflection: > This is definitely outside the scope of this PR

what you say makes sense but please let's not block the shipping of this PR introducing extra, not directly related, discussions, thanks

WebReflection:
@nicolo-ribaudo but isn't the whole purpose of this wrap to fix only the builtins extend? AFAIK there are no private, or user defined classes, that need such work around, right?

nicolo-ribaudo:
You can import an un-transpiled class from an extrnal npm module and extend it with a transpiled class.

---

Also, I opened https://github.com/sindresorhus/globals/issues/126 to remove the list of builtins in this PR. If that issue gets accepted, it probably would also handle every browser builtin.

WebReflection:
> remove the list of builtins in this PR

that lost won't include `HTMLElement` though, unless you grab the `browser` property and filter by comparing `typeof global[key] === 'function' && /^[A-Z]/.test(key)`, right?

packages/babel-helpers/src/helpers.js
nicolo-ribaudo: It wouldn't work, because core-js makes methods "look native". e.g.
https://github.com/zloirock/core-js/blob/16357e4fefa070573a342ef2a5692f30ba4320c6/tests/tests/es6.promise.ls#L9
https://github.com/zloirock/core-js/blob/6e389278357a6296a58155742871ca876dff2bc3/tests/helpers/looks-native.ls

WebReflection: there are here and there Babel specific conventions, like `__esModule = true`, I wonder if it's worth considering a non enumerable, non configurable, `__nativeWrap = true` to explicitly opt out from Babel wraps so that the check would be `builtins.includes(ctor) && !ctor.__nativeWrap`

Simple enough to cover polyfills case.

justinfagnani: @Kovensky:

> This would also catch user-defined classes, but one could then question whether you actually need this transform if you expect to find such a class

It happens, especially in the case where a library is published compiled to npm, and used by an app that's shipping native ES6. We try to discourage distributing compiled code, but we can't control that.

We could indeed add a flag to the polyfill. What we really want is this: https://esdiscuss.org/topic/add-reflect-isconstructor-and-reflect-iscallable

packages/babel-plugin-transform-classes/src/vanilla.js
justinfagnani: The polyfill would load before the compiled code, and so would wrap native HTMLElement to have a working constructor in the first place, otherwise without the polyfill both `new HTMLElement()` and `HTMLElement.apply` fail. So when the polyfill is loaded, the wrapping doesn't need to happen.

Optimally, code would be compiled to "force" wrapping of HTMLElement, even if it can't statically see a direct subclass, but the wrapping would bail if the constructor is callable.

packages/babel-helpers/src/helpers.js
nicolo-ribaudo: I think that this discussion caan be deferred, since it *works* as it is implemented in this PR.
Don't wrap classess without necessity is a possible performance optimization, but for now shipping a working implementation is enough.

nicolo-ribaudo:
@WebReflection Yes. It could be an option (e.g. `builtins: "es" | "browser" | string[]`)

nicolo-ribaudo:
@WebReflection  46325e1 enables wrapping for DOM classes always, since in non-browsers users won't use them anyway.

WebReflection:
makes, sense as it is now

hzoo:
looks good @WebReflection @justinfagnani?

WebReflection:
LGTM

hzoo:
Nice work on following through with this @nicolo-ribaudo and everyone's reviews - good community effort to move this forward finally 😄

nicolo-ribaudo:
:tada:

packages/babel-plugin-transform-classes/README.md
mgol: @xtuc You mean that the compiler doesn't work IE, don't you? I surely hope Babel itself still supports IE...

packages/babel-helpers/src/helpers.js
```diff
@@ -425,6 +425,51 @@ helpers.inheritsLoose = defineHelper(`
   }
 `);
 
+// Based on https://github.com/WebReflection/babel-plugin-transform-builtin-classes
+helpers.wrapNativeSuper = defineHelper(`
+  var _gPO = Object.getPrototypeOf || function _gPO(o) { return o.__proto__ };
+  var _sPO = Object.setPrototypeOf || function _sPO(o, p) { o.__proto__ = p };
```
jridgewell: Bug! This fallback does not return `o`.

nicolo-ribaudo: Whoops, good catch!

packages/babel-helpers/src/helpers.js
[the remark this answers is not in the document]
trusktr: But, if we're writing new code that extends builtins...?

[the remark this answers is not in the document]
WebReflection: then you are responsible to make such code work as expected otherwise you should stop using JavaScript since even `Object` could be overwritten :smile:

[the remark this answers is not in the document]
trusktr: Gotcha, let me stop using JS on my next web app!

trusktr:
How do I use this? I'm in Meteor 1.6.1 which uses Babel 7 `beta.38`. 

I read in the OP,

> If the user wants to extend another class (for example, a DOM class), it can be declared using the `builtins` option.

What does that mean exactly? Can you provide an example?

---

I get an error like this:

```
Uncaught TypeError: Failed to construct 'CustomElement': The result must implement HTMLElement interface
```

Where my app code is like this:

```js
    customElements.define('my-el', class MyEl extends HTMLElement {
        connectedCallback() {
            console.log(' ------ my el!!')
        }
    })
    document.body.appendChild(document.createElement('my-el'))
```

trusktr:
I checked the source,

https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-classes/src/index.js#L8-L14

and I see that `HTMLElement` is included in `globals.browser`, so it should just work?

What `.babelrc` configuration (for a regular Babel build, not Meteor) is needed to make this work?

trusktr:
I made a simple reproduction (using just Babel, no Meteor) showing that I'm using `@babel/plugin-transform-classes` but I still get the above error:

https://github.com/trusktr/babel/tree/issue-7020

It includes the built output in `build/test.js`. Open `test.html` in your browser to see the error.

You can also try `npm i && ./node_modules/.bin/babel test.js -o build/test.js` to build it.

packages/babel-helpers/src/helpers.js
nicolo-ribaudo: I just noticed that this hasn't been fixed in this PR, but it looks like it's fixed now.

PR #7034, opened 2017-12-16
Fix object spread according to spec

### LINKED ISSUE #7382: object-rest-spread doesn't prevent prototype mutation via a __proto__ property

A bug report.

### Input Code

```js
var o = Object.create(null);
o.a = 'foo';
o.__proto__ = [];
var o2 = {...o};
console.log(Array.isArray(Object.getPrototypeOf(o2))); // true, expected: false
```

See [this test case](https://babeljs.io/repl/#?babili=false&browsers=IE%2011&build=&builtIns=false&code_lz=G4QwTgBA9hC8EHkBGArApgYwC4DoNjRCzQAoA7AVwBsqBKAbgChGocQ4IByAMyik6asA-kIAOYKFigiOAbQC6TRqEhQAjB2TpsbAM66AlgHMyJAN4BfADTQGzFdABMHMzjdQLSjFDK6oVNBwqKCMSTnUIcUkYcDAQAE8AfggSNAAPUUxiABMALggsMAo0Wk4bAEEwOPicA11K6pItLJwjNCwABQkpLHjMhG4SdVoRpm9ff0Dg0PDnKKkIWITk1IystDyIbhAqXRKyiAaE2vqqhKbUFrbO7sk-tAGhxxG7IA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=stage-3%2Cenv&prettier=false&targets=&version=6.26.0&envVersion=1.6.2).

N/A, see the REPL link (basically, you need the `proposal-object-rest-spread` transform enabled).

### Expected Behavior

You should see `true` & `false` logged as it behaves when invoked directly in a Chrome/Firefox console.

### Current Behavior

If you open a console, you'll see `true` logged twice but it should be logged only for `o1`. If you paste this code directly in a browser console it prints `true` & `false`.

This is because object spread is desugared by Babel to a bare `Object.assign` which behaves differently: `Object.assign` invokes property setters and object spread shouldn't (see https://github.com/tc39/proposal-object-rest-spread/blob/master/Issues.md#setters for an example).

### Possible Solution

A custom modified implementation of `Object.assign` that changes regular assignment:
```js
target[key] = source[key];
```
should be changed to one that invokes `Object.setProperty`:
```js
Object.defineProperty(target, key, {value: source[key]});
```

### Context

I was merging default options values with one received from a user via `Object.assign` with `Object.create(null))` as the first parameter (see https://github.com/angular/preboot/pull/71). It has been pointed out that I could change the implementation to use regular objects with object spread but since the code is transpiled I can't do it (this particular repo uses TypeScript but you may easily imagine a pure-JS scenario).

### Your Environment

N/A, the REPL example says it all.

nuragic:
Hi,

I found this bug the last week so this is my attempt to fix it; I apologize in advance if anything is missing (this is my first PR to babel). Any suggestion, feedback, code review, etc. would be welcome! 😄 

That said, according to the spec the object spread doesn't work the same way as `Object.assign` or equivalent polyfill, basically because **_"spread defines new properties, while Object.assign() sets them"_**. Detailed info [here](http://2ality.com/2016/10/rest-spread-properties.html#spreading-objects-versus-objectassign).

To verify that this is actually a bug I added a new test first, then I made the corresponding changes to fix the implementation according to that. Please feel free to run it separately against the current `master` branch in order to check.

I opted to just add a new helper to replicate this behavior; also removed the `useBuiltIns` reference from the plugin's docs cause that wouldn't apply anymore. Some other tests from `babel-preset-env` are using object spread so I had to fix those tests as well.

Thanks!

Fixes #7382


packages/babel-helpers/src/helpers.js
```diff
@@ -372,6 +372,29 @@ helpers.extends = defineHelper(`
...
+helpers.extendsDefine = defineHelper(`
+  export default function _extendsDefine() {
+    _extendsDefine = function (target) {
+      for (var i = 1; i < arguments.length; i++) {
+        var source = arguments[i];
+        Object.getOwnPropertyNames(source).forEach(function(propKey) {
+          var desc = Object.getOwnPropertyDescriptor(source, propKey);
+          Object.defineProperty(target, propKey, desc);
+        });
+        if (Object.getOwnPropertySymbols) {
```
xtuc: Could we have a more explicit check here:

```js
typeof Object.getOwnPropertySymbols === 'function'
```

packages/babel-helpers/src/helpers.js
```diff
@@ -372,6 +372,29 @@ helpers.extends = defineHelper(`
   }
 `);
 
+helpers.extendsDefine = defineHelper(`
```
nicolo-ribaudo: Maybe we could call this `extendSpread`? Or `mergeSpread`

hzoo:
Would love if we didn't have to do this or make it the spec option

packages/babel-helpers/src/helpers.js
```diff
@@ -372,6 +372,29 @@ helpers.extends = defineHelper(`
...
+      for (var i = 1; i < arguments.length; i++) {
+        var source = arguments[i];
+        Object.getOwnPropertyNames(source).forEach(function(propKey) {
+          var desc = Object.getOwnPropertyDescriptor(source, propKey);
+          Object.defineProperty(target, propKey, desc);
+        });
+        if (Object.getOwnPropertySymbols) {
+          Object.getOwnPropertySymbols(source).forEach(function(symbKey) {
+            var desc = Object.getOwnPropertyDescriptor(source, symbKey);
+            Object.defineProperty(target, symbKey, desc);
```
Kovensky: This also needs to skip non-enumerable symbols.

packages/babel-helpers/src/helpers.js
```diff
@@ -372,6 +372,29 @@ helpers.extends = defineHelper(`
   }
 `);
 
+helpers.extendsDefine = defineHelper(`
+  export default function _extendsDefine() {
+    _extendsDefine = function (target) {
+      for (var i = 1; i < arguments.length; i++) {
+        var source = arguments[i];
+        Object.getOwnPropertyNames(source).forEach(function(propKey) {
```
Kovensky: The `source` is allowed to be `null` or `undefined`.

```js
assert.doesNotThrow(() => ({ ...null, ...undefined }))
```

packages/babel-helpers/src/helpers.js
xtuc: What about `spreadByDefine` to keep the define idea?

packages/babel-plugin-proposal-object-rest-spread/README.md
```diff
@@ -52,37 +52,9 @@ require("@babel/core").transform("code", {
...
-
-```js
-z = Object.assign({ x }, y);
-```
 
 ## References
 
 * [Proposal: Object Rest/Spread Properties for ECMAScript](https://github.com/sebmarkbage/ecmascript-rest-spread)
 * [Spec](http://sebmarkbage.github.io/ecmascript-rest-spread)
+* [Spread VS. Object.assign](http://2ality.com/2016/10/rest-spread-properties.html#spreading-objects-versus-objectassign)
```
xtuc: FYI @rauschma, can we include a link to your website here?

rauschma: @xtuc Absolutely!

nuragic:
Thanks everyone for the review! I'll try to fix everything ASAP

packages/babel-helpers/src/helpers.js
```diff
@@ -372,6 +372,29 @@ helpers.extends = defineHelper(`
   }
 `);
 
+helpers.extendsDefine = defineHelper(`
+  export default function _extendsDefine() {
+    _extendsDefine = function (target) {
```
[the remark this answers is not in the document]
nuragic: @nicolo-ribaudo Hi! Well, I guess because I originally adapted this code from the above `_extends` … 😅  Will change that, ok!

packages/babel-helpers/src/helpers.js
nuragic: @nicolo-ribaudo is `spreadByDefine` ok for you?

packages/babel-helpers/src/helpers.js
nuragic: @xtuc 👌  for me… I chose this one because it's actually being used in other helpers such as `objectWithoutProperties` and `defineEnumerableProperties`. Would you like me to change those as well?

nuragic:
RE: @nicolo-ribaudo 

> Thank you for this pr!

:hugs: 

> I think we should keep the useBuiltIns option, since many users prefer smaller and faster code over 100% spec compliant code

:thinking: I believe that if we do that, it may cause a _user experience_ issue... For a user, "use built ins" doesn't have the same meaning as "transpile to non compilant code (faster tho)"... you know? I mean, I'm not actually against keeping the alternative transform with `Object.assign` but I guess we should do that in a more explicit way and not through `useBuiltIns`...? What do you think @xtuc @hzoo @Kovensky ?

packages/babel-helpers/src/helpers.js
nicolo-ribaudo: Yes

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-spread/expression/expected.js
```diff
@@ -1,9 +1,23 @@
...
   x
 }, y, {
   a
 }, b, {
   c
 });
+
+_spreadByDefine({}, Object.prototype);
+
+_spreadByDefine({}, {
```
nuragic: This test (and the one below) are failing... As you can see is not being transformed correctly: the first parameter is omitted causing the following result:

```js
_spreadByDefine({
  foo: 'bar'
});
```

I'd love if someone can point me at the right direction because right now I don't know how to fix this one. I guess the problem is in the plugin itself? `babel-plugin-proposal-object-rest-spread/src/index.js`

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-spread/no-object-assign-exec/exec.js
```diff
@@ -22,3 +22,11 @@ const obj2Spread = { ...obj2 };
 
 assert.deepEqual(obj, objSpread);
 assert.deepEqual(obj2, obj2Spread);
+
+// const obj3Spread = { ...{ get foo () { return 'bar' } } };
+// assert.equal(Object.getOwnPropertyDescriptor(obj3Spread, 'foo').value, 'bar')
```
nuragic: This one is commented out because it obviously fails until the above will be fixed…

packages/babel-helpers/src/helpers.js
xtuc: We can change them in a later pr, it's not a big deal actually but we need it's better that our helper are fast enough.

nuragic: @xtuc hi! just added a new commit for fixing all the code review comment... this check has been removed now...

xtuc:
>  I guess we should do that in a more explicit way and not through useBuiltIns
That's a breaking change then, if think it's not necessary for now.

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-spread/expression/expected.js
nicolo-ribaudo: Instead of adding the object expression only if the first argument isn't an object expression, you should add it before the loop if the first prop is a spread property.

https://github.com/nuragic/babel/blob/71d8313fef584ad7ff8bb7a7d5e79e6599d60df3/packages/babel-plugin-proposal-object-rest-spread/src/index.js#L370-L372

nicolo-ribaudo:
@hzoo The only difference is that `_extendsByDefine` uses 
```js
var desc = Object.getOwnPropertyDescriptor(source, key);
Object.defineProperty(target, key, desc);
```
instead of
```js
target[key] = source[key];
```

I don't think that this difference is big enough to jiustify the presence of two different output modes (normal/spec or loose/normal).

packages/babel-helpers/src/helpers.js
```diff
@@ -378,8 +378,12 @@ helpers.spreadByDefine = defineHelper(`
       var source = arguments[i];
       for (var key in source) {
         if (Object.prototype.hasOwnProperty.call(source, key)) {
-          var desc = Object.getOwnPropertyDescriptor(source, key);
-          Object.defineProperty(target, key, desc);
+          Object.defineProperty(target, key, {
```
nuragic: I fixed the previous (wrong) implementation, right now this should be what actually spread do according to the spec... as we mentioned him before, it would be awesome if @rauschma could doble check on this (thanks in advance for your patience 🙏 😌  ).

packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -356,6 +356,10 @@ export default function() {
           props = [];
         }
 
+        if (t.isSpreadElement(path.node.properties[0])) {
```
nuragic: @nicolo-ribaudo Thanks for the suggestion! Seems that I eventually figured it out :)

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-spread/no-object-assign-exec/exec.js
```diff
@@ -0,0 +1,32 @@
...
+const obj2 = { NOSET: 123, NOWRITE: 456 };
+// this line would throw `TypeError: Cannot assign to read only property 'NOWRITE'`
+// if transformed as Object.assign (or equivalent) because those use *assignment* for creating properties
+// (spread defines them)
+const obj2Spread = { ...obj2 };
+
+assert.deepEqual(obj, objSpread);
+assert.deepEqual(obj2, obj2Spread);
+
+const obj3Spread = { ...{ get foo () { return 'bar' } } };
```
nuragic: I've also added those tests that @Kovensky suggested they would fail (now passing) 👌

nuragic:
Regarding the `useBuiltIns` option: I still believe that we shouldn't keep it for this particular plugin because it may cause further issues, but you know, it just my opinion® so.. as you wish. Let me know if you want to finally keep it or not and I'll update the PR again. Thanks!

packages/babel-helpers/src/helpers.js
nicolo-ribaudo: It seems correct to me (just read the proposed spec)

Kovensky:
Alternative: keep `useBuiltIns`, but only use it in a `loose` mode.

Another alternative is to keep the current assign-based transform as the default mode, and making this new mode the `spec` mode.

I don't know what is the yardstick used to decide whether something should be `loose` or `spec`; but I'd personally prefer the `loose` way.

packages/babel-helpers/src/helpers.js
```diff
@@ -372,6 +372,25 @@ helpers.extends = defineHelper(`
   }
 `);
 
+helpers.spreadByDefine = defineHelper(`
+  export default function _spreadByDefine(target) {
+    for (var i = 1; i < arguments.length; i++) {
+      var source = arguments[i];
+      for (var key in source) {
```
Kovensky: `for in` does not enumerate `Symbol` keys, but they should be copied when they are enumerable.

`ForIn / OfHeadEvaluation` will call this when evaluating a `for in`: https://tc39.github.io/ecma262/#sec-enumerate-object-properties

`CopyDataProperties` calls `[[OwnPropertyKeys]]`, though, which almost always delegates to https://tc39.github.io/ecma262/#sec-ordinaryownpropertykeys, which includes symbols.

(🚲🏠  maybe `copyDataProperties` is a better name for the helper 🤔)

Kovensky: It also seems that `Reflect.ownKeys` does what we want. Since it's now possible for helpers to depend on other helpers, maybe a `ownPropertyKeys` helper can be made.

```js
export default function _ownPropertyKeys (target) {
  if (typeof Reflect !== "undefined" && Reflect.ownKeys) {
    _ownPropertyKeys = Reflect.ownKeys;
  } else {
    _ownPropertyKeys = ownKeysImpl;
  }
  return _ownPropertyKeys.apply(this, arguments);
  function ownKeysImpl (target) {
    const keys = Object.keys(target)
    if (Object.getOwnPropertySymbols) {
      return keys.concat(
        Object.getOwnPropertySymbols(target)
          .filter(sym => Object.getOwnPropertyDescriptor(target, sym).enumerable)
      );
    }
    return keys;
  }
}
```

nuragic: > for in does not enumerate Symbol keys, but they should be copied when they are enumerable.

:man_facepalming: Oh man... Indeed, `for in` isn't ok in this context. Moreover, it enumerates properties in the prototype chain as well, so in any case `Object.keys()` should be used... Thanks for the additional review! I need to add a few more tests… 

> It also seems that Reflect.ownKeys does what we want. Since it's now possible for helpers to depend on other helpers, maybe a ownPropertyKeys helper can be made.

👍  Yeah, that could be a great solution! The only thing I see is that in `ownKeysImpl ` you're additionally filtering for `enumerable` symbols, when in reality `Reflect.ownKeys` is basically the same as `Object.getOwnPropertyNames(target).concat(Object.getOwnPropertySymbols(target))`… so I guess if we add an `_ownPropertyKeys ` helper it needs to return the same as `Reflect.ownKeys`... Additionally, it would just be an internal helper of an helper in this context 😄  so given your example, I'd say that this would be enough for now:

```js
function spreadByDefine(target) {
    var source = arguments[i];
    if (typeof Object.getOwnPropertySymbols === 'function') {
      ownKeys = ownKeys.concat(Object.getOwnPropertySymbols(source).filter(function(sym) { 
        return Object.getOwnPropertyDescriptor(source, sym).enumerable;
      }));
    }
    ownKeys.forEach(function(key) {
      Object.defineProperty(target, key, {
        value: source[key],
        writable: true,
        enumerable: true,
        configurable: true
      });
    });
  }
  return target;
}
```

> maybe copyDataProperties is a better name for the helper

TBH, this helper is doing almost exactly the same as the `extends` one, so I'd say that `extendsByDefine` would be a good name (originally was `extendsDefine`, now `spreadByDefine`…) 💭

packages/babel-helpers/src/helpers.js
rauschma: @nuragic Sorry, don’t have much time ATM. IIRC, the key thing with spread was: it reads properties via “get”, creates properties via “define”.

nuragic:
If I understood the meaning of `loose/spec` modes correctly, I agree with @Kovensky. Transforming to `Object.assign` would make sense in `loose` mode, I think.

Regarding the `spec` mode... I'm not sure what it does, or if it's related to ESM: is there any documentation / article about this? Thanks!

packages/babel-helpers/src/helpers.js
nuragic: @rauschma Cool, no worries! Sorry for bothering you… Thanks anyway, I think there's already a lot of detailed info in your blog post, plus all the code reviewers I've got are really the best ones! 😄

existentialism:
Not sure there is a good measuring stick for adding loose/spec, often times it seems like it depends which came first :)

FWIW, my vote is to keep current behavior (with useBuiltIns) and majke this spec mode (and error if useBuiltIns since its not applicable).

packages/babel-helpers/src/helpers.js
```diff
@@ -375,17 +375,21 @@ helpers.extends = defineHelper(`
...
-      for (var key in source) {
-        if (Object.prototype.hasOwnProperty.call(source, key)) {
-          Object.defineProperty(target, key, {
-            value: source[key],
-            writable: true,
-            enumerable: true,
-            configurable: true
-          });
-        }
+      var source = (arguments[i] != null) ? arguments[i] : {};
```
nuragic: Fixed (again) 😬  cc @Kovensky

nuragic:
At this point, I guess that the only remaining thing is to decide regarding `spec / loose / useBuiltIns`.

Looking at the docs here https://babeljs.io/docs/plugins/preset-env/

> **useBuiltIns**
boolean, defaults to `false`.
A way to apply babel-preset-env for polyfills (via “babel-polyfill”).

>**spec**
boolean, defaults to `false`.
Enable more spec compliant, but potentially slower, transformations for any plugins in this preset that support them.

>**loose**
boolean, defaults to `false`.
Enable “loose” transformations for any plugins in this preset that allow them.

At the end, the plugin should behave accordingly to that, right? So the only conclusion I can get that would make sense IMO is:

- when `loose: true`:
   - use `extends`. 
- when `loose: true, useBuiltIns: true`:
  - use `extends` (which will use `Object.assign` underneath). 
- otherwise (including default config case `spec: false, loose: false, useBuiltIns: false`):
  - use `spreadByDefine`. 

What do you think? Could you summarize your proposals in a similar way? @existentialism @hzoo @nicolo-ribaudo  Thanks!

nuragic:
Hi folks,

Please let me know what's missing in order to merge, if you finally want to have this as `spec` or any other change. Thanks!

packages/babel-helpers/src/helpers.js
```diff
@@ -372,6 +372,29 @@ helpers.extends = defineHelper(`
...
+  export default function _spreadByDefine(target) {
+    for (var i = 1; i < arguments.length; i++) {
+      var source = (arguments[i] != null) ? arguments[i] : {};
+      var ownKeys = Object.keys(source);
+      if (typeof Object.getOwnPropertySymbols === 'function') {
+        ownKeys = ownKeys.concat(Object.getOwnPropertySymbols(source).filter(function(sym) {
+          return Object.getOwnPropertyDescriptor(source, sym).enumerable;
+        }));
+      }
+      ownKeys.forEach(function(key) {
```
Andarist: for better perf you could use existing `defineProperty` helper here, which does an additional check before using `Object.defineProperty`

it should be possible to add `import defineProperty from "defineProperty"` to use it in the created helper

Andarist:
Im not sure what others thing, but IMHO old behaviour should stay in the `loose` mode. The same way we use simple assignments in `loose` in `proposal-class-properties` plugin, when in spec/default mode it uses `defineProperty` calls

packages/babel-helpers/src/helpers.js
nuragic: Mhh ok cool! Thanks for the tip, I'll try that

nuragic: Just checked that helper's code and what it does is just:

```js
    if (key in obj) {
      Object.defineProperty(obj, key, {
        value: value,
        enumerable: true,
        configurable: true,
        writable: true
      });
    } else {
      obj[key] = value;
    }
```

So seems that in this context it won't be helpful at the end... cause it's the intended behavior to always define a property (and not to do an assignment)... 😌

Andarist: how is the outcome different? from what I understand the point of using `Object.defineProperty` is to omit potential setters, but if there is no property with the given key there is no setter for this key and a simple assignment can be used to cause the same effect (default descriptor of assignment is `{ enumerable: true, configurable: true, writable: true }`)

nuragic: Oh, I think you're right! 😅  I was missing that important detail about the `in` check...

So, how in practice I could use an helper inside another helper? You mentioned the `import defineProperty from "defineProperty"` but I can't see any practical example... I mean, everything is declared as string like:
```
defineHelper(`…source code…`);
```

:thinking:

Andarist: example of adding a helper dependency - [here](https://github.com/nuragic/babel/blob/00ae9810bb45a543eca5543cc54aec145a6bc141/packages/babel-helpers/src/helpers.js#L97)

nuragic: Thanks! Fixed

Andarist:
@nuragic could you rebase this too? changed/added tests will need a file rename (actual -> input, expected -> output), which is a bummer, but fortunately you haven't touched too many files

nuragic:
@Andarist Done! Thanks for the advise about the rename... git handled that automatically BTW 😮

Andarist:
Code itself looks good to me, but there is still an open question when this new `spreadByDefine` should be used. IMHO `loose` mode should preserve old behaviour, but the question is - should new `spreadByDefine` be used in default mode or new `spec` mode for this transform? I don't feel like I could have decide on this though cc @loganfsmyth @hzoo

nuragic:
Yep, I agree about using `loose` mode to preserve the old behavior. Moreover, given that we also optimized the potentially slow part of the helper using the existing `defineProperty`, I think that we could safely make this the default behavior.... as I proposed above this is a quick summary about how I think it should work:

- when `loose: true`:
   - use `extends`. 
- when `loose: true, useBuiltIns: true`:
  - use `extends` (which will use `Object.assign` underneath). 
- otherwise (including default config case `spec: false, loose: false, useBuiltIns: false`):
  - use `spreadByDefine`. 

As a user I expect babel to transpile everything to make it work according to spec... However, that's really subjective. The reality is that we've got 3 different flags and we have to stick with them. 

As far as I can see, the only transform which make use of the `spec` option is `plugin-transform-arrow-functions`, so following that criteria perhaps would be legit to also make this new behavior fall into the `spec` category.

On the other hand, as @Andarist mentioned above, we've another counter example on `plugin-proposal-class-properties` which by default make use of  `Object.defineProperty` and uses simple assignment only in `loose` mode.

... tough decision ahead 😬 😅

packages/babel-helpers/src/helpers.js
```diff
@@ -374,6 +374,26 @@ helpers.extends = defineHelper(`
   }
 `);
 
+helpers.spreadByDefine = defineHelper(`
```
jridgewell: Can we rename this to `objectSpread`?

jridgewell:
Let's do spec by default, and let people opt into a faster loose mode.

nuragic:
Hi! Cool, will do that (in a few hours tho, sorry but I'm in CET timezone, 2:30am right now). Good night! 😄

packages/babel-helpers/src/helpers.js
nuragic: Done! Also added the same test specified in https://github.com/babel/babel/issues/7382

nuragic:
> Let's do spec by default, and let people opt into a faster loose mode.

@jridgewell just to verify... it's finally ok to implement as suggested in https://github.com/babel/babel/pull/7034#issuecomment-360765044 ? 

I mean, I'd need to add then support for both `loose` and `useBuiltIns` flags?

jridgewell:
> I mean, I'd need to add then support for both loose and useBuiltIns flags?

Yah, just undelete that part of the code. Then switch between `objectSpread` and `extends`/`useBuiltins` if in loose mode.

nuragic:
@jridgewell Done! Review pls 🙏  :shipit: 😄

packages/babel-plugin-proposal-object-rest-spread/README.md
```diff
@@ -54,18 +54,28 @@ require("@babel/core").transform("code", {
 
 ## Options
 
+By default, this plugin will produce spec compliant code. The Babel's `objectSpread` helper will be used.
+
+### `loose`
+
+`boolean`, defaults to `false`.
+
+Enabling this option will use Babel's `extends` helper.
```
Andarist: Would be could if we could add a note that `extends` is `Object.assign` + a fallback and explain a difference between using assign and defineProperty as some sort of caveat paragraph about `loose` mode in this transform.

nuragic: Mhh okay! Will try to add it later as soon as I'll have some free time 🙂

Andarist: Actually there is already a link about that dfference somewhere at the bottom, so maybe it's just a matter of linking to it from here too.

nuragic:
@Andarist Just changed the docs to add the part you requested, hope it's ok! 😄 

Now seems this PR is ready to merge, just wanted to check if you guys could give your ✅  @Kovensky @xtuc @nicolo-ribaudo :thinking: 

Finally thanks everyone for your support and patience! Looking forward to the next PR 😛

Andarist:
@nuragic gr8 work! Thanks for being patient with us :)

Issue #6258, opened 2017-09-16
Simplify class static getter in loose mode


hzoo:
We could simplify the transform here if we want: https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAECCIFMBOAXaBvAUNaEVhQEthoBzBNANQFEAlAZQEkB5AOQAoBKDbHaJCgFckAO2g0GLVrwC-mGUA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015-loose&prettier=false&targets=&version=7.0.0-beta.0

```js
// input
class Alert {
  static get VERSION() {
    return VERSION
  }
}
```

```js
"use strict";

function _defineProperties(target, props) { for (var i = 0; i < props.length; i++) { var descriptor = props[i]; descriptor.enumerable = descriptor.enumerable || false; descriptor.configurable = true; if ("value" in descriptor) descriptor.writable = true; Object.defineProperty(target, descriptor.key, descriptor); } }

function _createClass(Constructor, protoProps, staticProps) { if (protoProps) _defineProperties(Constructor.prototype, protoProps); if (staticProps) _defineProperties(Constructor, staticProps); return Constructor; }

var Alert =
/*#__PURE__*/
function () {
  function Alert() {}

  _createClass(Alert, null, [{
    key: "VERSION",
    get: function get() {
      return VERSION;
    }
  }]);

  return Alert;
}();
```

```
// can inline instead
Object.defineProperty(Alert, "VERSION", {
    get: function () {
        return VERSION;
    },
    enumerable: true,
    configurable: true
});
```

ajaymathur:
Sounds nice. Can I please give it a shot and come back if I have any questions?

existentialism:
@ajaymathur sure!

ajaymathur:
@existentialism Thanks 😃

ajaymathur:
Hi, I started with this, I was able to transform the tree, I have a question condition for this transformation should be `ClassDeclaration` with only one body node of kind `get`.

Am I correct?  What could be the other conditions?

ajaymathur:
Hi, I am sharing my findings here please correct me if I am wrong.

https://github.com/babel/babel/blob/72da5e1d022a533c7209070aa1eb2920552526b5/packages/babel-plugin-transform-es2015-classes/src/vanilla.js#L189
[...]

PR #7038, opened 2017-12-17
Static set get to object property

ajaymathur:
| Fixed Issues?            | https://github.com/babel/babel/issues/6258 
| Tests Added + Pass?      | Yes 😊
| Documentation PR         | 
| Any Dependency Changes?  |


ajaymathur:
@hzoo request you to please review 🙂

packages/babel-plugin-transform-classes/src/loose.js
```diff
@@ -1,6 +1,14 @@
 import nameFunction from "@babel/helper-function-name";
 import VanillaTransformer from "./vanilla";
-import { types as t } from "@babel/core";
+import { template, types as t } from "@babel/core";
+
+const objectDefineProperty = template(`
+  Object.defineProperty(CLASS_NAME, KEY, {
+    KIND: FUNCTION_BODY,
+    enumerable: true,
```
nicolo-ribaudo: This should be `false` (which is the defautl value, so you can just omit it):

```js
> class A { static method() {} }
undefined

> Object.getOwnPropertyDescriptor(A, "method");
{value: ƒ, writable: true, enumerable: false, configurable: true}
```

packages/babel-plugin-transform-classes/src/vanilla.js
```diff
@@ -522,9 +522,18 @@ export default class ClassTransformer {
       if (this._processMethod(node, scope)) return;
     }
 
+    // Check if it is a static get or set
+    if (node.static && ["get", "set"].indexOf(node.kind) > -1) {
```
nicolo-ribaudo: Can you use `node.static && (node.kind === "get" || node.kind === "set")`? It is just a little longer but it is more clear.

packages/babel-plugin-transform-classes/src/loose.js
```diff
@@ -64,4 +72,22 @@ export default class LooseClassTransformer extends VanillaTransformer {
       return true;
     }
   }
+
+  _processSetGet(node, path) {
+    const nonConstructorMethod = path.container.filter(containedNode => {
+      return containedNode.kind !== "constructor";
+    });
```
nicolo-ribaudo: You can avoid the `filter` loop if there are more than two nodes:

```js
const isSingleAccessor = path.container.length === 1 ||
  (path.container.length === 2 && path.container.some(node => node.kind === "constructor"));
```

ajaymathur:
@nicolo-ribaudo Thanks a lot for the review. ❤️ 

- by the comments in the issue it said we should optimise only when there are single get or set in the class, https://github.com/babel/babel/issues/6258#issuecomment-331577640

I will update PR as per other two comments. 🙂

ajaymathur:
> Sorry for the long delay 😅
It looks like this PR dosen't cover cases like
class A {
  static get foo() {}
  static set foo() {}
}
Would it be possible to optimize it?

I will give this a thought and come back to you. I will updated PR with other reviews

packages/babel-plugin-transform-classes/src/loose.js
Kovensky: This is missing `writable: true`, though.

However, you can't specific `writable: true` if the descriptor is a getter or setter descriptor...

Also, getters/setters need to be put together in the _same_ descriptor. This would create separate descriptors if both are present so the one to be evaluated last would be the only one remaining.

nicolo-ribaudo: That is the output from the console, I don't think that there should be `writable: true`.

Kovensky: Yeah, we can't actually add `writable: true` here, because it's a setter/getter descriptor. It'll throw unless it's a value descriptor.

packages/babel-plugin-transform-classes/src/loose.js
```diff
@@ -64,4 +72,22 @@ export default class LooseClassTransformer extends VanillaTransformer {
...
+    if (nonConstructorMethod.length === 1) {
+      this.body.push(
+        objectDefineProperty({
```
Kovensky: Github seems to have lost my comment, but:

This creates separate descriptors for `get` and `set`, but they have to be the _same_ descriptor, otherwise the previously declared descriptor will be lost.

ajaymathur: Thanks a lot for the review. I have done all the other changes, and I am working on this one right now, but if you have any suggestions on best way to tackle this?

packages/babel-plugin-transform-classes/src/loose.js
Andarist: @Kovensky are you sure that last descriptor overrides previous ones? Expanding on @nicolo-ribaudo example:
```js
> class A { static get prop() {} }
undefined

> Object.getOwnPropertyDescriptor(A, "prop");
{get: ƒ, set: undefined, enumerable: false, configurable: true}

> Object.defineProperty(A, 'prop', { set(val) { this.val = val }, enumerable: false, configurable: true })
class A { static get prop() {} }

> Object.getOwnPropertyDescriptor(A, "prop")
{get: ƒ, set: ƒ, enumerable: false, configurable: true}
```

It seems that descriptors are actually merged rather than replacing itself.

xtuc:
It seems that you forgot to update the test fixtures after https://github.com/babel/babel/pull/7038/commits/3831e117b5bd323996312e510e30f895cd4c6df7.

packages/babel-plugin-transform-classes/src/loose.js
nicolo-ribaudo: Cool, I didn't know it!

Andarist: Me neither, still aint sure if that's the case for all edge cases. I have just tested it in Chrome's console 😅

Kovensky: Oh, [it's true](https://www.ecma-international.org/ecma-262/5.1/#sec-8.12.9). I wasn't sure so I tried checking the ES5 spec too, and it seems that it does try to merge the descriptors as well. https://www.ecma-international.org/ecma-262/5.1/#sec-8.12.9. I couldn't find an ES 5.0 spec but I guess it's safe to assume it works the same and will merge descriptors as long as they are configurable.

packages/babel-core/test/fixtures/transformation/misc/regression-2892/expected.js
```diff
@@ -89,7 +89,7 @@ function _foo() {
               return _bar2.apply(this, arguments);
             };
 
-            bar = function bar() {
+            bar = function _ref() {
```
xtuc: This change seems not related to your PR, but i'm not sure why.

nicolo-ribaudo: It is by regenerator: https://github.com/babel/babel/pull/7137

packages/babel-plugin-transform-classes/src/loose.js
```diff
@@ -64,4 +71,22 @@ export default class LooseClassTransformer extends VanillaTransformer {
       return true;
     }
   }
+
+  _processSetGet(node, path) {
+    const nonConstructorMethod = path.container.filter(containedNode => {
```
Andarist: NIT: should be method**s**

packages/babel-plugin-transform-classes/src/vanilla.js
```diff
@@ -522,9 +522,18 @@ export default class ClassTransformer {
...
 
+    // Check if it is a static get or set
+    if (node.static && (node.kind === "get" || node.kind === "set")) {
+      if (this._processSetGet(node, path)) return;
+    }
+
     this.pushToMap(node, false, null, scope);
   }
 
+  _processSetGet() {
```
Andarist: this seems a little bit clunky, but i guess it is somewhat forced upon because of the hierarchical class structure, I will file a refactoring request issue for this

packages/babel-plugin-transform-classes/src/vanilla.js
```diff
@@ -522,9 +522,18 @@ export default class ClassTransformer {
       if (this._processMethod(node, scope)) return;
     }
 
+    // Check if it is a static get or set
+    if (node.static && (node.kind === "get" || node.kind === "set")) {
```
Andarist: well, this is more optimized than before, dunno how much do you want to spend on this, but in general i think this should be more optimized even further - get+set pair for the same property could still produce a single `Object.defineProperty` call, while at the moment this will still fallback to using `createClass` helper

packages/babel-plugin-transform-classes/src/loose.js
Andarist: as stated under other comments here - this is actually not true, descriptors will be merged, but anyway in this part of the code from what i understand we wouldnt have to worry about this case anyway as its running only for `if (nonConstructorMethod.length === 1)` case

packages/babel-plugin-transform-classes/src/vanilla.js
xtuc: Agreed. I don't like that structure so much. Composition over inheritance they say.

Andarist: U can upvote this then - https://github.com/babel/babel/issues/7202 😄

xtuc: can't upvote more

packages/babel-plugin-transform-classes/src/loose.js
ajaymathur: @Kovensky @Andarist I think we should not merge the descriptors because of issue it might create with computed properties 
```js
const foo = () => 'foo'
class MyClass {
  get [foo()]() { this._foo || 'foo' }
  set [foo()](val) { this._foo = val }
}
```

ajaymathur: @Andarist slack chat link https://babeljs.slack.com/archives/C062RC35M/p1515617623000644?thread_ts=1515589048.000272&cid=C062RC35M

packages/babel-plugin-transform-classes/src/vanilla.js
ajaymathur: It will be issues with computed properties like you mentioned here : https://babeljs.slack.com/archives/C062RC35M/p1515617623000644?thread_ts=1515589048.000272&cid=C062RC35M
 Should I optimise this PR in any way? 🙂

packages/babel-plugin-transform-classes/src/loose.js
Andarist: we can merge non-computed ones though

packages/babel-plugin-transform-classes/src/vanilla.js
nicolo-ribaudo: You can optimize it if the getter and setter name are not computed.

packages/babel-plugin-transform-classes/src/loose.js
```diff
@@ -64,4 +64,111 @@ export default class LooseClassTransformer extends VanillaTransformer {
...
+      if (containedNode.kind === "constructor") {
+        return;
+      }
+
+      if (containedNode.kind === "get") {
+        GET_FUNCTION_BODY = t.functionExpression(
+          null,
+          containedNode.params,
+          containedNode.body,
+        );
```
ajaymathur: What is the best way to remove path for another node in container?

packages/babel-plugin-transform-classes/src/vanilla.js
ajaymathur: done the optimisation but I am blocked by : https://github.com/babel/babel/pull/7274 : In case I am doing it wrong, please guide me 🙂

nicolo-ribaudo:
Hi! Sorry for not going through with this PR five years ago 😅 

In the meanwhile, JavaScript got support for static class properties. Now, instead of writing
```js
class Alert {
  static get VERSION() {
    return VERSION;
  }
  static set VERSION(version) {
    this.VERSION = version;
  }
}
```
users can just write
```js
class Alert {
  static VERSION = VERSION;
}
```

so this optimization is not necessary anymore.

PR #7039, opened 2017-12-17
Fix publishing process

### LINKED ISSUE #6174: babel-standalone version number runs behind the actual version number
The version number is inserted into `babel.js` (and `babel.min.js`) as part of the build. If the version is bumped **after** building (like what happens when publishing with Lerna), it will contain the old version number. The new `alpha.20` release is actually `alpha.20`, but it returns `alpha.19` as its version (so `Babel.version` returns `alpha.19`, and the REPL shows alpha.190 as the version number).

I solved that in the standalone babel-standalone repo by bumping the version number before building, but it might be trickier in this case. Suggestions appreciated.

yavorsky:
| Fixed Issues?            | Fixes #6174 
| Major: Breaking Change?  | 🙅
| Minor: New Feature?      | 🙅
| Tests Added + Pass?      | 🙅
| Documentation PR         | 🙅
| Any Dependency Changes?  | 🙅

When we are publishing new packages, we run `BABEL_ENV=production make build-dist` first, and only then `lerna publish`. During `build-dist` process we creating standalone bundles, it takes version from babel-core, which wasn’t actually updated yet, because it will be updated during `lerna publish` on the next step.

Proposed solution runs `lerna publish` 2 times: firstly - before `build-dist` with `--skip-npm --skip-git --skip-temp-tag` arguments, so we just overriding versions for all packages.
For the second run, we ignore step where we select the version and just take it from the `lerna.json`, which was updated during the first `lerna publish`, so it just publishes the new version and creates git commit.

cc @Daniel15

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,7 @@ publish:
 	make test
 	# not using lerna independent mode atm, so only update packages that have changed since we use ^
 	# --only-explicit-updates
-	./node_modules/.bin/lerna publish --force-publish=* --exact --skip-temp-tag --yes --repo-version $(cat lerna.json | jq '.version' | tr '"' ' ')
+	../node_modules/.bin/lerna publish --force-publish=* --exact --skip-temp-tag --yes --repo-version node -p "require('../lerna.json').version"
 	make clean
 
 bootstrap:
--- a/Makefile
+++ b/Makefile
@@ -121,12 +121,11 @@ publish:
 	make clean-lib
 	rm -rf packages/babel-runtime/helpers
 	rm -rf packages/babel-runtime/core-js
-	./node_modules/.bin/lerna publish --skip-npm --skip-git --skip-temp-tag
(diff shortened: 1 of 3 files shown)


Daniel15:
This seems reasonable to me. It would be  good if Lerna had better support for this, since it'll come up for any project that uses Lerna + embeds the version number in the resulting bundle.

loganfsmyth:
One alternative is that we use a `prepublish` or `version` script to build the bundles after the version has been updated in the package. I'm probably most in favor of that because it also means that we could make the day-to-day development steps use the normal non-bundled files.

yavorsky:
I've added `version` script to the babel-standalone and babel-preset-env standalone, because prepublish runs on every `lerna bootstrap` before any builds.
I've tested it with my pet-project, but still needs a review.

xtuc:
Do you still need this or could we close it?

PR #7043, opened 2017-12-18
Avoid adding #__PURE__ annotation to .bind(this)() expressions

Kovensky:
This improves on the fix done on #6999.

Fixes the `#__PURE__` annotation getting added to `(async () => {})()` IIFEs when the arrow function transform is running with `"spec": true`.

The arrow function transform will change the function to be wrapped with a call to `.bind(this)`, which throws off the naïve `isIIFE` check. This adds a fancier `isIIFE` check that will also catch immediately invoking the result of `.bind(this)`.

### DIFF
--- a/packages/babel-plugin-transform-async-to-generator/test/fixtures/async-to-generator/async-iife-with-regenerator-spec/expected.js
+++ b/packages/babel-plugin-transform-async-to-generator/test/fixtures/async-to-generator/async-iife-with-regenerator-spec/expected.js
@@ -54,3 +54,21 @@ regeneratorRuntime.mark(function _callee2() {
     return _notIIFE.apply(this, arguments);
   };
 })();
+
+/*#__PURE__*/
+babelHelpers.asyncToGenerator(
+/*#__PURE__*/
+regeneratorRuntime.mark(function _callee4() {
+  return regeneratorRuntime.wrap(function _callee4$(_context4) {
+    while (1) switch (_context4.prev = _context4.next) {
+      case 0:
+        babelHelpers.newArrowCheck(this, _this);
+        _context4.next = 3;
+        return 'not iife';
+
+      case 3:
+      case "end":
+        return _context4.stop();
+    }
+  }, _callee4, this);
+})).bind(this);
--- a/packages/babel-plugin-transform-async-to-generator/test/fixtures/async-to-generator/async-iife-with-regenerator-spec/expected.js
+++ b/packages/babel-plugin-transform-async-to-generator/test/fixtures/async-to-generator/async-iife-with-regenerator-spec/expected.js
@@ -1,0 +1,56 @@
+var _this = this;
+
+babelHelpers.asyncToGenerator(
+/*#__PURE__*/
+regeneratorRuntime.mark(function _callee() {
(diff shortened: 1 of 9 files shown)


nicolo-ribaudo:
I think a proper fix would be to check in the `markAsPure` helper if the call to be marked is part of a MemberExpression. That would also fix cases like https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&code_lz=BQQwzgngdgxgBAMwK6wC4EsD2U7AJRwDeAUHHAL54B0MIANnfgNxA&debug=false&circleciRepo=&evaluate=true&lineWrap=false&presets=es2017%2Ctypescript&prettier=false&targets=&version=7.0.0-beta.35

Kovensky:
@nicolo-ribaudo that would, unfortunately, then break the not-IIFE case for spec arrow functions (`(async () => {})` would not get annotated because it has a `.bind(this)` call) 🤔

Kovensky:
@nicolo-ribaudo looks like it's not possible as the `annotateAsPure` helper can receive `Node`s without a `NodePath`

packages/babel-plugin-transform-async-to-generator/test/fixtures/async-to-generator/async-iife-with-regenerator-spec/options.json
```diff
@@ -0,0 +1,8 @@
+{
+	"plugins": [
+    ["transform-arrow-functions", { "spec": true }],
+    "transform-regenerator",
+    "transform-async-to-generator",
+		"external-helpers"
```
Andarist: indentation is wrong here

Andarist:
We could build version accepting NodePaths exclusively on top of the existing helper.

Kovensky:
I am trying a different approach by wrapping the calls on `ParenthesizedExpression`s.

packages/babel-plugin-transform-async-to-generator/test/fixtures/async-to-generator/async-iife-with-regenerator-spec/expected.js
```diff
@@ -0,0 +1,74 @@
...
+        babelHelpers.newArrowCheck(this, _this);
+        _context4.next = 3;
+        return 'not iife';
+
+      case 3:
+      case "end":
+        return _context4.stop();
+    }
+  }, _callee4, this);
+})).bind(this);
```
Kovensky: The call to `.bind(this)` makes this not eligible for DCE by Uglify, even with the `#__PURE__` annotation.

Maybe changing this to just check for `MemberExpression` at all is sufficient; or should we stay optimistic and add the `#__PURE__` annotation in this case? 🤔

nicolo-ribaudo: Is `.bind` impure only when the user modifies `Function.prototype.bind`? If yes, I vote for being optimistic (since Babel has always assumed that builtins are unchanged), otherwise we should remove it.

Kovensky:
I'll merge it just so this bug is not on the next beta anymore, and I'm going to see if using sequence expressions in #7044 instead works out better.

PR #7044, opened 2017-12-18
Wrap pure annotations in parens

Kovensky:
Alternative to #7043

Wrap expressions to be annotated as pure in a parenthesizedExpression when they may appear inside another expression.


Kovensky:
Unfortunately, I could not find any way to make these tests pass:

```
  1) arrow function conversion should convert this references in constructors with spec compliance:

      AssertionError [ERR_ASSERTION]: 'class Example extends class {} {\n  constructor() {\n    var _this,\n        _arrowCheckId = {};\n\n    (\n    /*#__PURE__*/\n  == 'class Example extends class {} {\n  constructor() {\n    var _this,\n        _arrowCheckId = {};\n\n    /*#__PURE__*/\n    (fun
      + expected - actual

         constructor() {
           var _this,
               _arrowCheckId = {};

      -    (
           /*#__PURE__*/
      -    function () {
      +    (function () {
             babelHelpers.newArrowCheck(this, _arrowCheckId);
             _this;
      -    }.bind(_arrowCheckId));
      +    }).bind(_arrowCheckId);
           _this = super();
           this;

           () => _this = super();

      at assertConversion (packages/babel-traverse/test/arrow-transform.js:38:10)
      at Context.<anonymous> (packages/babel-traverse/test/arrow-transform.js:131:5)

  2) arrow function conversion should convert this references in constructors with spec compliance without extension:

      AssertionError [ERR_ASSERTION]: 'class Example {\n  constructor() {\n    var _this = this;\n\n    (\n    /*#__PURE__*/\n    function () {\n      babelHelpers.ne == 'class Example {\n  constructor() {\n    var _this = this;\n\n    /*#__PURE__*/\n    (function () {\n      babelHelpers.newArrow
      + expected - actual

       class Example {
         constructor() {
           var _this = this;

      -    (
           /*#__PURE__*/
      -    function () {
      +    (function () {
             babelHelpers.newArrowCheck(this, _this);
             this;
      -    }.bind(this));
      +    }).bind(this);
           this;

           () => this;
         }

      at assertConversion (packages/babel-traverse/test/arrow-transform.js:38:10)
      at Context.<anonymous> (packages/babel-traverse/test/arrow-transform.js:188:5)

  3) arrow function conversion should convert this references in methods with spec compliance:

      AssertionError [ERR_ASSERTION]: 'class Example {\n  method() {\n    var _this = this;\n\n    (\n    /*#__PURE__*/\n    function () {\n      babelHelpers.newArro == 'class Example {\n  method() {\n    var _this = this;\n\n    /*#__PURE__*/\n    (function () {\n      babelHelpers.newArrowCheck
      + expected - actual

       class Example {
         method() {
           var _this = this;

      -    (
           /*#__PURE__*/
      -    function () {
      +    (function () {
             babelHelpers.newArrowCheck(this, _this);
             this;
      -    }.bind(this));
      +    }).bind(this);
           this;

           () => this;
         }

      at assertConversion (packages/babel-traverse/test/arrow-transform.js:38:10)
      at Context.<anonymous> (packages/babel-traverse/test/arrow-transform.js:235:5)
```

Is there any reason why they need to be done in this file, or should I split these tests off into regular fixtures?

It's also unfortunate that, while babylon preserves the parenthesized expression, it attaches the comment to the wrong node. Is this perhaps a babylon bug?

Kovensky:
After going through all this, I now see that it is useless to try to annotate `.bind(this)` as `.bind` can't ever be considered pure. The call itself may be, but the `[[Get]]` of the `"bind"` cannot.

An alternative is to, instead of (just) wrapping on a parenthesized expression, to put the `.bind(this)` inside a function wrapper, like so:

```js
(
  /*#__PURE__*/
  function (_this) {
    return babelHelpers.asyncToGenerator(...).bind(_this)
  }(this)
)
```

But uglify is also not allowed to inline the function if it turns out that its result is used, so it'd actually penalize the common case of the code actually being used 😢

Andarist:
> After going through all this, I now see that it is useless to try to annotate .bind(this) as .bind can't ever be considered pure. The call itself may be, but the [[Get]] of "bind" cannot.

Do you mean old environments without `Function.prototype.bind` or that it could be overriden by user?

Kovensky:
@Andarist I mean by uglify's DCE rules; it seems `(/* #__PURE__ */ () => {})` can be removed but not `(/* #__PURE__ */ function() {}.bind(this))`.

input:

```js
void function () {
var a = (/* #__PURE__ */ () => {})
var b = (/* #__PURE__ */ function() {}.bind(this))
var c = function() {}
}()
```

output:

```js
!function(){(function(){}).bind(this)}();
```

Andarist:
cc @kzc shouldnt this `.bind` call get eliminated too?

kzc:
> shouldnt this .bind call get eliminated too?

Interesting. I thought it would have worked. 

It's not the `.bind`, but the comment being within parens...
```
$ bin/uglifyjs -V
uglify-js 3.2.2

$ echo 'a(); /*@__PURE__*/b(); c();' | bin/uglifyjs -c
a(),c();

$ echo 'a(); /*@__PURE__*/( b() ); c();' | bin/uglifyjs -c
a(),c();

$ echo 'a(); ( /*@__PURE__*/b() ); c();' | bin/uglifyjs -c
a(),b(),c();
```
It would make sense to have `uglify-js@3.x` and `uglify-es` recognize pure annotations within parentheses. Releases for `uglify-js@2.x` are no longer made.

kzc:
Uglify issue created: https://github.com/mishoo/UglifyJS2/issues/2629

Kovensky:
I wonder if it would be better for my proposed change to babel to generate `(0, /*#__PURE__*/babelHelper(...))` instead of only wrapping in a `ParenthesizedExpression`, regardless of this bug fix in Uglify.  This helps with babylon correctly attaching the comment when re-parsing the file, as `ParenthesizedExpression` is only used for code generation.

Kovensky:
I redid this using sequence expressions instead (`(0, /*#__PURE__*/...)`), and it worked well; babylon also correctly reattaches the comments, so all tests pass.

However, using `path.replaceWith` is still a trap because it will try to move the comments to the outer node. 🤔

loganfsmyth:
@Kovensky Is this closable now that #7043 landed?

Kovensky:
IIRC this is a more general fix to the one in #7043, but I'll have to check again 😅

kzc:
> I wonder if it would be better for my change to generate `(0, /*#__PURE__*/babelHelper(...))` instead of only wrapping in a ParenthesizedExpression, regardless of this bug fix in Uglify. This helps with babylon correctly attaching the comment when re-parsing the file, as `ParenthesizedExpression` is only used for code generation.

FWIW, `uglify-js` and `terser` now handle pure annotations within parentheses correctly in the event the comma expression technique mentioned above is not used.

```
$ cat parens.js 
foo();
void function() {
    var a = () => {};
    var b = (/* #__PURE__ */ function() {}.bind(this));
    var c = function() {};
    var d = (/* #__PURE__ */ function() {}());
}();
bar();

$ cat parens.js | terser -mc
foo(),bar();
```
```
$ echo 'a(); /*@__PURE__*/( b() ); c();' | terser -c
a(),c();

$ echo 'a(); ( /*@__PURE__*/b() ); c();' | terser -c
a(),c();
```

danez:
Is this still valid or can we close this?

nicolo-ribaudo:
This PR needs to be updated to use the new `t.parenthesizedExpression` node.

kzc:
related issue: #10306

Issue #7053, opened 2017-12-19
[babylon] Add support for @@iterator


bcherny:
Choose one: feature request

### Input Code

```js
/* @flow */

declare class A {
  @@iterator(): Iterator<File>;
}
```

### Babel/Babylon Configuration (.babelrc, package.json, cli command)
```js
import { parse } from 'babylon'
import { readFileSync } from 'fs'

let code = readFileSync('file.flow.js', 'utf-8')
let ast = parse(code, { plugins: ['flow'], sourceType: 'module' }) // SyntaxError: Unexpected token
```

### Expected Behavior

`@@iterator` should be parsed into an Identifier.

### Current Behavior

Babylon throws a SyntaxError.

### Context

Attempting to use Babylon to parse [flow/dom.js](https://github.com/facebook/flow/blob/eda26a031a351ff8cf00f7c5d829bcad787f73dd/lib/dom.js) (error is thrown on line 55).

### Your Environment

| software         | version(s)
| ---------------- | -------
| Babel            |  Master (68476b6)
| Babylon          | Master (68476b6)
| node             | 8.1.3
| npm              | 5.3.0
| Operating System | OSX

rajzshkr:
This should be a small change. I'll give a PR today.

PR #7058, opened 2017-12-19
Add support for @@iterator

rajzshkr:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/7053
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR         | 
| Any Dependency Changes?  |


packages/babylon/src/tokenizer/index.js
```diff
@@ -1231,6 +1231,9 @@ export default class Tokenizer extends LocationParser {
       const ch = this.fullCharCodeAtPos();
       if (isIdentifierChar(ch)) {
         this.state.pos += ch <= 0xffff ? 1 : 2;
+      } else if (this.state.isIterator) {
+        this.state.pos += 2;
+        delete this.state.isIterator;
```
nicolo-ribaudo: Can you use `this.state.isIterator` here?

hzoo:
@nicolo-ribaudo it's a flow syntax not related to JS, it would error with `let @@iterator = 1`? Also funny since the link to flow says TODO switch with Symbol.iterator?

packages/babylon/src/tokenizer/index.js
nicolo-ribaudo: I meant `this.state.isIterator = false` lol

nicolo-ribaudo:
> Also funny since the link to flow says TODO switch with Symbol.iterator?

We should ask to the flow team if they plan to remove `@@iterator` in the near future.

hzoo:
Should ping the flow team on all the flow prs anyway 😛

hzoo:
@bcherny

rajzshkr:
Also we should not allow any other identifier `iterator` and `asyncIterator`. I'll update the PR

rajzshkr:
> Should the @@iterator identifier be allowed everywhere? e.g. let @@iterator = 2;

Flow parse the above syntax, but in V8 throws an error.

nicolo-ribaudo:
> Flow parse the above syntax

I checked and flow paeses that syntax *sometimes* (e.g. it accepts `let @@iterator = number` but not `type @@iterator = number`.
I think it's ok for us to allow it only in class/object/interface property types.

nicolo-ribaudo:
I have thought about this a little, and I think that we should allow `@@iterator` **only** in class/object/interface.
The flow parser is more lenient and accepts things like
```js
function @@iterator() {}
@@iterator();
```
but I think that we shouldn't parse it since it will generate invalid JavaScript output.

So, instead of throwing if `@@iterator` is after var/let/type we should throw if it is *not* inside an object/interface/class property name.

existentialism:
@nicolo-ribaudo agreed

loganfsmyth:
@danez Would you like to look this over since you're familiar with the parser? It seems fine to me, but I'm also not an expert on it.

nicolo-ribaudo:
> Flow-strip-types probably needs to remove the @@iterator and @@asyncIterator, as the resulting code is otherwise invalid?

Currently Flow only removes them from types.

xtuc:
Looks good to me. As mentioned by @danez, `babel-generator` will probably need to be updated in order to emit this syntax and could you please add a test for it?

It could probably also impact TypeScript? @hzoo could we ping them? I don't think that could break anything but maybe require a transformations (I'm not a TS expert).

rajzshkr:
@nicolo-ribaudo updated the test cases.

rajzshkr:
@danez I added test cases, to make sure iterator is removed from types in Flow-strip-types. https://github.com/babel/babel/pull/7058/commits/36a07317e8fddf4b7abe420b79d054310917fa8c, Do you want to add test case for babel-generator, since it is already removed by Flow-strip-types

nicolo-ribaudo:
Yes, since someone might use Babel for a codemod and thus not remove Flow types.

rajzshkr:
@nicolo-ribaudo @danez added test case for generator and Flow-strip-types plugin

packages/babel-plugin-transform-flow-strip-types/test/fixtures/strip-types/strip-iterator/input.js
```diff
@@ -0,0 +1,15 @@
+declare class A {
```
nicolo-ribaudo: It looks like the output for this file is missing?

existentialism: That's expected (i.e. there is no output)

rajzshkr: @existentialism Thanks!!

existentialism:
@rajzshkr thanks!

PR #7059, opened 2017-12-19
Import babel-plugin-transform-dotall-regex

mathiasbynens:
Original: https://github.com/mathiasbynens/babel-plugin-transform-dotall-regex

Moving it into the official Babel repository makes it easier to maintain the transform.


mathiasbynens:
PTAL @thejameskyle. Feel free to edit the patch directly — especially for small nits that’s much easier/faster for everyone.

packages/babel-plugin-transform-dotall-regex/test/index.js
```diff
@@ -0,0 +1,24 @@
+import path from 'path';
+import fs from 'fs';
+import assert from 'assert';
+import { transform, transformFileSync } from '@babel/core';
```
xtuc: `transform` is not used, we can remove it.

mathiasbynens: Done.

mathiasbynens:
@xtuc Thanks! Patch updated.

mathiasbynens:
@hzoo Thanks; I’ve changed `test/index.js`.

nicolo-ribaudo:
@mathiasbynens `input.js` needs to be rename to `actual.js` in tests

mathiasbynens:
@nicolo-ribaudo Thanks for pointing to that example. I’ve updated the patch.

As a side comment, I’m struggling to understand how the `actual.js` file name makes sense for the input. In tests, the term “actual” usually refers to the dynamically generated output based on the input, not to the input itself.

packages/babel-plugin-transform-dotall-regex/README.md
```diff
@@ -0,0 +1,45 @@
...
+Compile regular expressions using [the `s` (`dotAll`) flag](https://github.com/tc39/proposal-regexp-dotall-flag) to ES5 that works in today’s environments.
+
+For example, `/./s` is transpiled into `/[\0-\uFFFF]/`, and `/./su` is transpiled into `/[\0-\u{10FFFF}]/u`.
+
+[Here’s an online demo.](https://mothereff.in/regexpu#input=const+regex+%3D+/foo.bar/s%3B%0Aconsole.log%28%0A++regex.test%28%27foo%5Cnbar%27%29%0A%29%3B%0A//+%E2%86%92+true&dotAllFlag=1)
+
+## Installation
+
+```sh
+$ npm install @babel/plugin-transform-dotall-regex
```
nicolo-ribaudo: `--save-dev`

nicolo-ribaudo:
> As a side comment, I’m struggling to understand how the actual.js file name makes sense for the input. In tests, the term “actual” usually refers to the dynamically generated output based on the input, not to the input itself.

Yeah, you are right... I think that it is some sort of tecnichal debt that no one ever cared about fixing.

PS. Still `mv -t packages/babel-plugin-transform-dotall-regex/test/fixtures/dotall-regex packages/babel-plugin-transform-dotall-regex/test/fixtures/{simple,with-unicode-flag}` :wink:

hzoo:
Yeah we should refactor the tests to just be input/output files or something, it's always been actual/expected and it's confusing

existentialism:
@mathiasbynens updated the readme, and the tests

hzoo:
Made https://github.com/babel/babel/issues/7063 @mathiasbynens

existentialism:
After we land this, I have a small follow-up PR that will enable this in preset-env as well

existentialism:
@mathiasbynens thanks! :P

mathiasbynens:
> After we land this, I have a small follow-up PR that will enable this in preset-env as well

We’ll also want to start thinking about preset-es2018, which would include this transform.

mathiasbynens:
Thanks for the help, everyone! ♥

hzoo:
> preset-es2018

We are going to stop making yearly presets since we have preset-env now 😄

mathiasbynens:
> We are going to stop making yearly presets since we have preset-env now 😄

That’s even better news! 🎉

PR #7065, opened 2017-12-19
Add dot-all regex support to preset-env and standalone

existentialism:
| Fixed Issues?            | 
| Documentation PR         | 
| Any Dependency Changes?  |

Follow up to https://github.com/babel/babel/pull/7059

### DIFF
--- a/packages/babel-preset-env/src/debug.js
+++ b/packages/babel-preset-env/src/debug.js
@@ -1,6 +1,6 @@
 /*eslint quotes: ["error", "double", { "avoidEscape": true }]*/
 import semver from "semver";
-import { prettifyVersion, semverify } from "./utils";
+import { isUnreleasedVersion, prettifyVersion, semverify } from "./utils";
 
 const wordEnds = size => {
   return size > 1 ? "s" : "";
@@ -15,9 +15,23 @@ export const logMessage = (message, context) => {
 export const logPlugin = (plugin, targets, list, context) => {
   const envList = list[plugin] || {};
   const filteredList = Object.keys(targets).reduce((a, b) => {
-    if (!envList[b] || semver.lt(targets[b], semverify(envList[b]))) {
-      a[b] = prettifyVersion(targets[b]);
+    const version = envList[b];
+    const target = targets[b];
+
+    if (!version) {
+      a[b] = prettifyVersion(target);
+    } else {
+      const versionIsUnreleased = isUnreleasedVersion(version, b);
+      const targetIsUnreleased = isUnreleasedVersion(target, b);
+
+      if (
+        (versionIsUnreleased && !targetIsUnreleased) ||
+        (!targetIsUnreleased && semver.lt(target, semverify(version)))
+      ) {
+        a[b] = prettifyVersion(target);
+      }
     }
+
     return a;
   }, {});
 
--- a/packages/babel-preset-env/src/debug.js
+++ b/packages/babel-preset-env/src/debug.js
@@ -12,27 +12,30 @@ export const logMessage = (message, context) => {
   console.log(logStr);
 };
 
-export const logPlugin = (plugin, targets, list, context) => {
-  const envList = list[plugin] || {};
-  const filteredList = Object.keys(targets).reduce((a, b) => {
-    const version = envList[b];
-    const target = targets[b];
+// Outputs a message that shows which target(s) caused an item to be included:
+// transform-foo { "edge":"13", "firefox":"49", "ie":"10" }
+export const logPlugin = (item, targetVersions, list, context) => {
+  const minVersions = list[item] || {};
 
-    if (!version) {
-      a[b] = prettifyVersion(target);
+  const filteredList = Object.keys(targetVersions).reduce((result, env) => {
+    const minVersion = minVersions[env];
+    const targetVersion = targetVersions[env];
+
+    if (!minVersion) {
+      result[env] = prettifyVersion(targetVersion);
     } else {
-      const versionIsUnreleased = isUnreleasedVersion(version, b);
-      const targetIsUnreleased = isUnreleasedVersion(target, b);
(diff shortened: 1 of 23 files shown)


packages/babel-preset-env/src/debug.js
```diff
@@ -15,9 +15,23 @@ export const logMessage = (message, context) => {
 export const logPlugin = (plugin, targets, list, context) => {
   const envList = list[plugin] || {};
   const filteredList = Object.keys(targets).reduce((a, b) => {
```
nicolo-ribaudo: Not related to this change, but `a` and `b` should really be renamed.

existentialism: Cleaned it up!

brneto:
This is the right behavior for dotall-regex? The use of the unicode flag (`u`) should had no effect, shouldn't it?

Why `/hello.world/us.test(input)` is returning false?

**Test:**
```
const input = `
Lorem ipsum dolor sit amet, consectetur adipiscing hello
world elit. Nam sit amet elit id risus aliquam porta.
`;

describe('RegEx dotAll mode', () => {
  let received;
  it('should dot matches a line break', () => {
    received = /hello.world/.test(input);
    expect(received).toBe(false);

    received = /hello.world/u.test(input);
    expect(received).toBe(false);

    received = /hello.world/s.test(input);
    expect(received).toBe(true);

    received = /hello.world/us.test(input);
    expect(received).toBe(false); // Why the regex is false here?
  });
});
```
**package.json:**
   ```
    "@babel/polyfill": "^7.0.0-beta.49",
    "@babel/cli": "^7.0.0-beta.49",
    "@babel/core": "^7.0.0-beta.49",
    "@babel/node": "^7.0.0-beta.49",
    "@babel/plugin-proposal-class-properties": "^7.0.0-beta.49",
    "@babel/plugin-proposal-decorators": "^7.0.0-beta.49",
    "@babel/plugin-proposal-function-bind": "^7.0.0-beta.49",
    "@babel/plugin-proposal-object-rest-spread": "^7.0.0-beta.49",
    "@babel/plugin-proposal-optional-chaining": "^7.0.0-beta.49",
    "@babel/plugin-syntax-dynamic-import": "^7.0.0-beta.49",
    "@babel/preset-env": "^7.0.0-beta.49",
    "@babel/preset-react": "^7.0.0-beta.49",
    "babel-core": "^7.0.0-0",
    "babel-eslint": "8.2.1",
    "babel-jest": "^23.0.1",
    "babel-loader": "^8.0.0-beta.3",
    "babel-plugin-dynamic-import-node": "^1.2.0",
    "eslint-plugin-babel": "^5.1.0",
```
**.babelrc:**
```
{
  "presets": [
    ["@babel/preset-env", {
      "targets": { "node": "current" }
    }]
  ],
  "plugins": [
    ["@babel/plugin-proposal-decorators", { "legacy": true }],
    ["@babel/plugin-proposal-class-properties", { "loose": true }],
    "@babel/plugin-proposal-object-rest-spread",
    "@babel/plugin-proposal-function-bind",
    "@babel/plugin-proposal-optional-chaining",
    "dynamic-import-node"
  ]
}
```

**Comment by mathiasbynens** [comment]
That should work indeed. https://new.babeljs.io/docs/en/next/babel-plugin-transform-dotall-regex.html#example

**Comment by brneto** [comment]
My test behavior has to do with the issue #6931? The test `/hello.world/us.test(input)` should be `true`, right?

**Comment by mathiasbynens** [comment]
`/hello.world/us.test('hello\nworld')` should be `true`.

**Comment by brneto** [comment]
I've just transpiled (with babel) a small test code and look what I got:

**In:**
```
const r1 = /./s;
const r2 = /./su;
```

**Out:**
```
"use strict";

const r1 = /./s;
const r2 = /[\0-\t\x0B\f\x0E-\u2027\u202A-\u{10FFFF}]/su;
```

**Comment by mathiasbynens** [comment]
Looks like it’s not transpiling the `s` flag at all? Is https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-dotall-regex not enabled somehow?

**Comment by brneto** [comment]
I'm using `@babel/preset-env` I think it is enabled by default. Should I take some additional action to get `dotall-regex` working?

**Comment by nicolo-ribaudo** [comment]
Which node version are you using? Maybe it already supports the dot-all flag.

**Comment by brneto** [comment]
node v9.11.1

**Comment by mathiasbynens** [comment]
There is indeed a bug here. `/./su` and `/[\0-\t\x0B\f\x0E-\u2027\u202A-\u{10FFFF}]/su` are not equivalent.

I’ve verified that regexpu-core handles this correctly:

```js
rewritePattern('.', 'su', { dotAllFlag: true, useUnicodeFlag: true });
// → '[\\0-\\u{10FFFF}]'
```

So it seems like in this case, Babel isn’t passing those flags to regexpu-core.

mathiasbynens:
Looking at https://github.com/babel/babel/blob/9f4e2f81d8c1780ba435e880479427f45213421a/packages/babel-plugin-transform-unicode-regex/src/index.js and 
https://github.com/babel/babel/blob/9f4e2f81d8c1780ba435e880479427f45213421a/packages/babel-plugin-transform-dotall-regex/src/index.js, the problem must be that Babel applies `transform-unicode-regex` before it applies `transform-dotall-regex`.

`transform-unicode-regex` transpiles the pattern, including any `.` in the pattern, and removes the `u` flag. Then later, when the transpiled regular expression is passed to `transform-dotall-regex`, it never sees a regular expression containing a `.` anymore.

Can we swap the order in which the transforms get applied? That would fix the bug.

existentialism:
Good ol' plugin ordering, created https://github.com/babel/babel/pull/8138

brneto:
I think would be a good ideia to include a test like the one I’ve done above, don’t you think?

existentialism:
@brneto thanks for reporting the issue, #8138 will be out in the next release!

PR #7076, opened 2017-12-20
Pull from babel/babel:master

MatAtBread:


hzoo:
Great thanks for the initial PR, are you ok if I make some commits? I think this has tests from the other transforms, so we should merge into the existing plugin (which I can do if it was erroring) or at least remove the old files.

MatAtBread:
@hzoo - as requested, pull implementing transform-async-to-promises.

As I understand it, you're intending to merge into transform-async-to-generator and used options to specify the target.

I have implemented only two options:

- `runtime` is the (optional) named identifier that contains the nodent-runtime. The default is no runtime. I included this option as the generated code is considerably smaller and somewhat faster in a lot of engines as without the runtime as lot of nested `try..catch` blocks get generated. If used, the included runtime is newly crafted 'nodent-runime/promises' (https://github.com/MatAtBread/nodent-runtime/blob/master/promise.js) that is both very small (no implementation of Promises or the generator runtime) and no longer pollutes `Function.prototype`. This updated nodent-runtime@3.2.0 has already been released to npmjs.

- `codeGenerationOptions` provides a means to set alternative defaults for symbol names, etc.

MatAtBread:
...sure - it's your project/plugin. Like everyone I have an opinion, but I'm no Babel expert (and only a very occasional user). Feel free to ask questions, raise issues, etc. I'm certainly interested in where it goes next

packages/babel-plugin-transform-async-to-promises/src/index.js
```diff
@@ -15,6 +15,32 @@ function parse(code) {
   }).program;
 }
 
+/* Polyfill for Object.assign(...) from:
```
ljharb: Please do not use MDN code in production anywhere :-/ this code doesn’t work with symbols, and has a number of other subtle bugs.

I’d suggest `object.assign` on npm.

packages/babel-plugin-transform-async-to-promises/src/index.js
```diff
@@ -0,0 +1,133 @@
...
+
+function parse(code) {
+  return babylon.parse(code, {
+    allowImportExportEverywhere: true,
+    allowReturnOutsideFunction: true,
+    allowSuperOutsideMethod: true,
+  }).program;
+}
+
+/* Polyfill for Object.assign(...) from:
```
ljharb: See my above comment; this shouldn’t be here.

packages/babel-plugin-transform-async-to-promises/src/index.js
hzoo: Well we don't need this at all, can just use `Object.assign` because min version is Node 4, or object spread

MatAtBread: If you look closely, it's not polyfilling the prototype, it's just defined & called locally. In fact, it could easily be replaced by a simple for-loop. And as hzoo says, if you're only targeting Node >4, Object.assign() will work fine.

@hzoo - can I assume you're OK to maintain this for now?

hzoo: Babel 7 does target Node 4 so you can use Object.assign by itself in this PR

ljharb: @MatAtBread right, i'm saying that basically all the code on MDN is broken and should never be used for any reason beyond "loosely explaining what a thing does". Since babel requires node 4+, this can just be deleted, but the larger point about MDN code stands.

packages/babel-plugin-transform-async-to-promises/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+          $arguments: "$args",
+          $Promise: "Promise",
+          $asyncspawn: "$asyncspawn",
+          $asyncbind: "$asyncbind",
+          $makeThenable: "$makeThenable",
+        }; 
+
+  if (options.codeGenerationOptions) {
+      let keys = Object.keys(options.codeGenerationOptions);
+      for (let i=0; i<keys.length;i++) {
```
ljharb: per previous comments, this can just be replaced with `Object.assign(opts, options.codeGenerationOptions);`

Kovensky: (IIRC we use transform-object-rest-spread in babel itself so) alternatively,
```ts
const opts = {
  // Code generation options
  es6target: false,
  // ...
  $makeThenable: "$makeThenable",
  ...options.codeGenerationOptions
};
```

matAtWork: If you look at the code, there's a conditional to prevent unknown keys being placed in the destination object. This is not the same as replacing it with `Object.assign`, which unconditionally copies all keys, irrespective of their existence in the destination.

I'm not familiar enough with the semantics of the Object spread operator to know if it is a suitable replacement for this simple loop. Is there something about `for` loops you don't like?

ljharb: Just the unreadability, hence this very confusion :-)

packages/babel-plugin-transform-async-to-promises/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+	  if (keys[i] in opts)
```
ljharb: This line seems to be indented weirdly?

matAtWork: No idea why - I'd guess something put tabs in place of spaces or vice versa. I'm not able to fix this at the moment (away from my home setup with the correct tools). Feel free to edit to match your preferred style.

matAtWork: ...I just had a go with the github editor, but TBH I'm not familiar with how these are best merged online.

matAtWork: So you're ok to accept the PR?

ljharb: Oops, sorry, I'll dismiss my review. I'm not personally confident in approving this, but I definitely am not blocking it further :-)

packages/babel-plugin-transform-async-to-promises/src/index.js
```diff
@@ -0,0 +1,112 @@
+/* Babel 7 transform */
+
+const babylon = require("babylon");
+const generate = require("@babel/generator").default;
+
+function printNode(ast) {
+  return generate(ast, {}).code;
+}
+
+function parse(code) {
```
xtuc: I wondering why you need to parse code here? Isn't it a regular Babel plugin?

packages/babel-plugin-transform-async-to-promises/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+              // Helpers for the transformer:
+              parse: parse, // Parse a JS fragment into an AST
+              printNode: printNode, // Print a node as JS source
+              logger: false /* console.log.bind(console)*/, // Log a warning
+            },
+          ).ast;
+
+          if (runtime) {
+            newAst.body.splice(0, 0, {
+              type: "ImportDeclaration",
```
xtuc: Could you please use babel-types for that? and maybe babel-traverse to place to node at the correct location?

packages/babel-plugin-transform-async-to-promises/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+
+function parse(code) {
+  return babylon.parse(code, {
+    allowImportExportEverywhere: true,
+    allowReturnOutsideFunction: true,
+    allowSuperOutsideMethod: true,
+  }).program;
+}
+
+const transform = require("nodent-transform").transform;
```
xtuc: This answers https://github.com/babel/babel/pull/7076/files#r160679424. Could we rather have a real Babel plugin?

packages/babel-plugin-transform-async-to-promises/src/index.js
matAtWork: There is an internal templating system which needs parse the templates.

packages/babel-plugin-transform-async-to-promises/src/index.js
matAtWork: Actually, I don't have the bandwidth to learn a new API or style/code conventions. I started this as a favour to @hzoo, and as far as I know the current implementation works. 

@hzoo - I'm happy to invite you (or anyone else) to my babel fork, but I'm really unable to provide the time to do this to the Babel team's satisfaction.

packages/babel-plugin-transform-async-to-promises/src/index.js
matAtWork: Not without a lot of effort - far more than I can commit to. This was discussed initially in https://github.com/MatAtBread/fast-async/issues/46 and at that point the costs of re-coding for the babel API (ignoring the stylistic requirements) would not be viable without the significant contribution of knowledgable babel contributors, hence a half-way where the 'transform' part of nodent was stripped of its dependancies was provided by me.

matAtWork:
Is anyone able to explain the CircleCI issues? They appear unrelated to the changes in the pull
Thanks.

matAtWork:
@hzoo - are you able to help in resolving these checks?

packages/babel-plugin-transform-async-to-promises/src/index.js
```diff
@@ -0,0 +1,113 @@
+/* Babel 7 transform */
+
+const babylon = require("babylon");
+const generate = require("@babel/generator").default;
```
ljharb: This dep doesn’t seem to be listed in package.json

packages/babel-plugin-transform-async-to-promises/src/index.js
```diff
@@ -0,0 +1,113 @@
+/* Babel 7 transform */
+
+const babylon = require("babylon");
```
ljharb: This dep doesn’t seem to be listed in package.json

packages/babel-plugin-transform-async-to-promises/src/index.js
Kovensky: Are those deps not managed by lerna; or do you still need to explicitly list them?

(I never looked up how lerna actually manages these...)

ljharb: I'm sure lerna might fix it so it *happens* to work, but you should always explicitly list every dep.

In this case, if it's not explicitly listed, *consumers* will break (since this isn't test code).

matAtWork: Finally got around to updating this is https://github.com/babel/babel/pull/7076/commits/943a44c522dca0acaabf3d1b73317ab155124f99. Hopefully this resolve the remaining issues for this implementation.

packages/babel-plugin-transform-async-to-promises/src/index.js
matAtWork: As above, fixed in https://github.com/babel/babel/pull/7076/commits/943a44c522dca0acaabf3d1b73317ab155124f99

packages/babel-plugin-transform-async-to-promises/src/index.js
```diff
@@ -0,0 +1,113 @@
...
+    allowReturnOutsideFunction: true,
+    allowSuperOutsideMethod: true,
+  }).program;
+}
+
+const transform = require("nodent-transform").transform;
+
+function transformAsyncToPromises(api, options) {
+  let requiresTranspilation;
+  const runtime = options.runtime ? options.runtime : null;
```
ljharb: `a ? a : b` should always be `a || b`

packages/babel-plugin-transform-async-to-promises/src/index.js
```diff
@@ -0,0 +1,113 @@
...
+    $Promise: "Promise",
+    $asyncspawn: "$asyncspawn",
+    $asyncbind: "$asyncbind",
+    $makeThenable: "$makeThenable",
+  };
+
+  if (options.codeGenerationOptions) {
+    const keys = Object.keys(options.codeGenerationOptions);
+    for (let i = 0; i < keys.length; i++) {
+      if (keys[i] in opts) {
```
ljharb: This check should probably be `Object.prototype.hasOwnProperty.call`, in case the key is “toString” or something.

packages/babel-plugin-transform-async-to-promises/src/index.js
```diff
@@ -0,0 +1,113 @@
...
+        opts[keys[i]] = options.codeGenerationOptions[keys[i]];
+      }
+    }
+  }
+
+  return {
+    visitor: {
+      Program: {
+        enter: function() {
```
ljharb: Should all of these be concise methods, like `enter() {`?

matAtWork:
Do I need to do anything to get this merged?

ljharb:
I’ll ping some folks to take a look at it.

matAtWork:
Ping....

ljharb:
I’ll ping again in the babel slack; I’m not comfortable merging it alone, but i think it’s ready.

hzoo:
Apologies again Matt, been really busy with trying to just get Babel v7 released and other personal stuff now that I've left my job. I'll try but maybe someone can look into this atm?

xtuc:
I'm just wondering, this new plugin doesn't require to be in the Babel monorepo. I would just mention it in the doc.

ljharb:
@xtuc it's pretty important i think that it is; the regenerator transform needs to not be the default one day, and this is an important stepping stone in that direction.

MatAtBread:
@loganfsmyth - the implementation here (nodent-transform) doesn't serialize or re-parse. It simply manipulates the AST, preserving location information where possible.

packages/babel-plugin-transform-async-to-promises/README.md
```diff
@@ -0,0 +1,58 @@
+# @babel/plugin-transform-async-to-promises
+
+> Turn async functions into ES2015 Promises
+
+## Example
+
+**In**
+
+```javascript
```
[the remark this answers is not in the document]
MatAtBread: Added some sample In and Out code. Note that the transform is pretty wide, and providing examples of all the possible transformations is a big job. 

I could generate examples from the tests at https://github.com/MatAtBread/nodent-compiler/tree/master/tests/semantics, but you'll notice there are currently over 70 tests,

MatAtBread:
> it would need to be a real AST transform

In what way is it not a real AST transform?

loganfsmyth:
@MatAtBread Oh! Well that's my misunderstanding then, sorry about that. I'll see if I can find some time for a more in-depth review so I don't make other incorrect comments.

MatAtBread:
I'm happy to answer any queries you might have. This re-work was started following a conversation with @hzoo last year. The `nodent-transform` was refactored out of the compiler/cli code, and has no external dependencies (eg no acorn or code-generator). 

However, to allow me to only have to maintain one core transform (at least for now), it doesn't use the Babel AST helpers (or anyone elses) and is monolithic, which, from your PoV, would make it look big and with some unnecessary helpers (eg: a tree walker, JS templates).

MatAtBread:
If anyone has any hints as to why Travis failed (all I did was update the README!), I'd appreciate some input.

packages/babel-plugin-transform-async-to-promises/README.md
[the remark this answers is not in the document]
LukeChannings: I read through the fixtures to confirm this does what I thought. Very thorough, and I’ve been looking for this Babel plugin for a long time. Many thanks 🙏

hzoo:
Maybe needs a rebase? can use `[skip ci]` for the readme change later

packages/babel-plugin-transform-async-to-promises/src/index.js
```diff
@@ -0,0 +1,118 @@
...
+            },
+            opts,
+            {
+              // Helpers for the transformer:
+              // Parse a JS fragment into an AST
+              parse: parse,
+              // Print a node as JS source
+              printNode: printNode,
+              // Log a warning, eg: console.log.bind(console)
+              logger: false,
```
swernerx: It seems like this line throws together with nodent-transform as it just uses the value here as a function.

I am seeing this stack trace:

```
      at node_modules/nodent-transform/arboriculture.js:2327:25
      at node_modules/nodent-transform/arboriculture.js:2326:50
      at treeWalk (node_modules/nodent-transform/arboriculture.js:2808:2)
      at down (node_modules/nodent-transform/arboriculture.js:2804:4)
      at node_modules/nodent-transform/arboriculture.js:2772:5
      at forEachNodeKey (node_modules/nodent-transform/arboriculture.js:2767:8)
      at descend (node_modules/nodent-transform/arboriculture.js:2786:3)
      at node_modules/nodent-transform/arboriculture.js:2381:13
      at treeWalk (node_modules/nodent-transform/arboriculture.js:2808:2)
      at down (node_modules/nodent-transform/arboriculture.js:2804:4)
      at node_modules/nodent-transform/arboriculture.js:2769:44
      at node_modules/nodent-transform/arboriculture.js:2769:15
      at forEachNodeKey (node_modules/nodent-transform/arboriculture.js:2767:8)
      at descend (node_modules/nodent-transform/arboriculture.js:2786:3)
      at node_modules/nodent-transform/arboriculture.js:2381:13
      at treeWalk (node_modules/nodent-transform/arboriculture.js:2808:2)
      at hoistDeclarations (node_modules/nodent-transform/arboriculture.js:2213:9)
      at asyncTransforms (node_modules/nodent-transform/arboriculture.js:441:9)
      at asynchronize (node_modules/nodent-transform/arboriculture.js:422:9)
      at transform (node_modules/nodent-transform/arboriculture.js:2881:14)
      at PluginPass.exit (src/transformAsyncToPromises.js:67:26)
      at newFn (node_modules/@babel/traverse/lib/visitors.js:243:21)
```

Using an empty function helps. But I am not sure whether we should just ignore all the warnings.

MatAtBread: Fixed in nodent-transform@3.2.3 - `false` now disables reporting of warnings

MatAtBread:
Is it just me, or is the current master (beta.46) failing tests?

ljharb:
@xtuc whoa, this has been discussed for a long time. The maintenance cost comes from having a default transform to generators; this transform moves towards avoiding/reducing that cost. Can you elaborate more on why you think this shouldn’t be in babel? The airbnb config, and all its users, are unable to use async/await without this transform, and if it’s not in babel core, the maintenance cost of the transform itself gets much higher.

WaldoJeffers:
@ljharb I 100% agree with you. If I am not mistaken, this would help solve many errors linked to `babel-regenerator`, [which are legion](https://github.com/babel/babel/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+regenerator) (at least on Node v >4, since it natively support Promises)

nicolo-ribaudo:
I agree with @xtuc. From my point of view, the main problem is that this PR doesn't implement the transform, but it is just a wrapper around nodent.

@sexualmaniac If you want so much to use this functionality, you can use https://github.com/MatAtBread/fast-async. If you look at the readme it is not "another shit ton of config", it is ONE line in your babel config and ONE line in your package.json. @xtuc and every other team member is already doing much work, so please don't be so harsh.

ljharb:
@nicolo-ribaudo it's a wrapper around a piece of nodent, not around all of nodent - @MatAtBread took the time and did the work to extract out that piece so that both nodent and babel could use it without needing to take on an otherwise large dependency.

The longer term plan is to avoid the nodent-transform dependency entirely, but the first step was going to be to get this transform added, with tests, so @MatAtBread could step back and others could build on that work.

ljharb:
The real issue is that depending on regenerator in the first place for async/await was a very very unfortunate decision that has had real maintenance costs, and it's critical for babel and babel users that there be a promises transform.

It certainly doesn't have to be this implementation or this transform, but objecting to this implementation with concerns about maintenance cost seems in bad faith unless someone's volunteering to do the work themselves to build an "async to promises" transform that *would* be considered acceptable. Is anyone volunteering to do that?

sexualmaniac:
@nicolo-ribaudo I absolutely love babel and the work of the team behind it.

This is not one or two lines, this is - read readme, configure, add runtime, test if it works, tell everyone why we need this. And we have other things, you know, like webpack with the whole world of loaders and configs.

Promises already have good support, let's make devs life easier with this change so we can get rid of regenerator once and forever.

matAtWork:
Is there a way to move forward with this? Or is it now blocked...

hzoo:
You don't need to anything @matAtWork, we just need to figure out what to do next. Again, for anyone else you can just use fast-async normally which works fine on its own.

shellscape:
I'd like to see this happen. `fast-async` works great but increases the size of some of my files by a factor of 4. Which is just 🍌

ljharb:
I prefer 10 times larger files even, over the runtime cost of using regenerator - but i want both plugins in core so that people are able to make that choice.

Dan503:
I am so glad that this pull request is so close to making it's way into Babel core. :D

With current Babel functionality, in order to use async functions with preset-env in IE11 the _entire_ babel-pollyfill plugin needs to be downloaded by not just authors but also _users_ just so that `regeneratorRuntime` is not undefined (a function that only Babel uses by the way).

This gives an idea of the sort of impact that the download alone has on user experience:
https://bundlephobia.com/result?p=babel-polyfill@6.26.0

By switching to a promise based async translation, the only thing it needs to work is a Promise polyfill. Promises are things that developers are able to take advantage of in their regular source code. This makes a Promise polyfill a far more worth while investment than all the code necessary to make `regeneratorRuntime` work which developers don't use at all in their source code. Promises are also a native browser API so only the browsers that don't understand Promises have to download and parse the polyfill. Smart devs can prevent modern browsers from downloading the polyfill code all together. `regeneratorRuntime` is non-standard code so _all_ browsers, no matter how cutting edge they are, have to download and parse the _entire_ babel-pollyfill library no matter what.

On top of that, every single time you include an async function it massively blows out the code size.

For example take this basic bit of js:

```js
var that = async function(){
    var thing = await fetch('https://dog.ceo/api/breeds/image/random').then(resp => resp.json());
    return thing;
}
var result = that();
console.log(result);
```

It will get blown out into this monster:

````js

var that = function that() {
  var thing;
  return regeneratorRuntime.async(function that$(context$2$0) {
    while (1) switch (context$2$0.prev = context$2$0.next) {
      case 0:
        context$2$0.next = 2;
        return regeneratorRuntime.awrap(fetch('https://dog.ceo/api/breeds/image/random').then(function (resp) {
          return resp.json();
        }));

      case 2:
        thing = context$2$0.sent;
        return context$2$0.abrupt('return', thing);

      case 5:
      case 'end':
        return context$2$0.stop();
    }
  }, null, this);
};
var result = that();
console.log(result);
````

By switching to a promise based translation the code can be simplified down to this (based on the output from [babel-plugin-async-to-promises](https://www.npmjs.com/package/babel-plugin-async-to-promises)).

````js
var that = function that() {
  var thing;
  return Promise.resolve().then(function () {
    return fetch('https://dog.ceo/api/breeds/image/random').then(function (resp) {
      return resp.json();
    });
  }).then(function (_resp) {
    thing = _resp;
    return thing;
  });
};
var result = that();
console.log(result);
````

This seems like the far more logical approach. The output is more human readable and also creates far less code bloat.

(The plugin doesn't seem to override the `preset-env` functionality so I can't use it if `env` is the preset. It was built to run on the es2015 preset.)

Please merge this pull request into Babel core. The current default translation is madness!

**Comment by benjamn** [comment]
Disclaimer: I'm the original author and current maintainer of Regenerator.

> I prefer 10 times larger files even, over the runtime cost of using regenerator - but i want both plugins in core so that people are able to make that choice.

@ljharb Can you quantify that claim?

Are you just talking about the size of `regenerator-runtime`? Because it's 2.8KB after minification and gzip, which is noticeable for sure, but `nodent-runtime` is 1.7KB, so it's hardly as if this transform has no runtime library. It's not even especially tiny.

The costs of both of these static assets are constant, and will be amortized over time by browser caching, as I'm sure you argue if you were trying to justify your own additions to client bundle size.

I will assume you're not talking about compilation performance, since you said "runtime cost," and compilation performance isn't nearly as important as runtime performance.

So... do you have anything to say about the relative runtime performance of `fast-async`/`nodent` code versus code generated by Regenerator?

### Because I sure do!

Here's a test program that I just wrote:

```js
require("nodent-runtime");
const regeneratorRuntime = require("regenerator-runtime");

async function eachAsync(items, fn) {
  let sum = 0;
  const count = items.length;
  for (let i = 0; i < count; ++i) {
    sum += await fn(items[i]);
  }
  return sum;
}

const items = [];
for (let i = 0; i < 100000; ++i) {
  items.push(i);
}

const start = +new Date;
eachAsync(items, item => item).then(sum => {
  console.log("sum", sum);
  console.log(new Date - start, "ms");
}, error => {
  console.log(error);
});

```

Here's the code Regenerator generates:

```js
require("nodent-runtime");

var regeneratorRuntime = require("regenerator-runtime");

function eachAsync(items, fn) {
  var sum, count, i;
  return regeneratorRuntime.async(function eachAsync$(_context) {
    while (1) {
        case 0:
          sum = 0;
          count = items.length;
          i = 0;

        case 3:
          if (!(i < count)) {
            _context.next = 10;
            break;
          }

          _context.next = 6;
          return regeneratorRuntime.awrap(fn(items[i]));

        case 6:
          sum += _context.sent;

        case 7:
          ++i;
          _context.next = 3;
          break;

        case 10:
          return _context.abrupt("return", sum);

        case 11:
        case "end":
          return _context.stop();
      }
    }
  }, null, this);
}

var items = [];

for (var i = 0; i < 100000; ++i) {
  items.push(i);
}

var start = +new Date();
eachAsync(items, function (item) {
  return item;
}).then(function (sum) {
  console.log("sum", sum);
  console.log(new Date() - start, "ms");
}, function (error) {
  console.log(error);
});
```

I'm honestly not sure how to evaluate accusations that this code is ugly. It's not what you wrote, by any stretch of the imagination, but neither is the code generated by `fast-async` or `nodent`. Debuggability kinda goes out the window when you're transpiling `async` functions, which is why it's especially important to have some sort of source maps, which is a whole lot easier if your transform is implemented as a Babel plugin, as Regenerator is.

Now for the `nodent` code:

```js
require("nodent-runtime");
const regeneratorRuntime = require("regenerator-runtime");
function eachAsync(items, fn) {
    return (function ($return, $error) {
        let sum, count;
        sum = 0;
        count = items.length;
        var $Loop_1_local;
        function $Loop_1_step() {
            var [i] = $Loop_1_local();
            ++i;
            return $Loop_1.bind(this, i);
        }
        
        function $Loop_1(i) {
            $Loop_1_local = function () {
                return [i];
            };
            if (i < count) {
                return fn(items[i]).then((function ($await_3) {
                    sum += $await_3;
                    return $Loop_1_step;
                }).$asyncbind(this, $error), $error);
            } else 
                return [1];
        }
        
        return Function.$asyncbind.trampoline(this, $Loop_1_exit, $Loop_1_step, $error, true)($Loop_1.bind(this, 0));
        function $Loop_1_exit() {
            return $return(sum);
        }
        
    }).$asyncbind(this, true);
}

const items = [];
for (let i = 0;i < 100000; ++i) {
    items.push(i);
}
const start = +new Date();
eachAsync(items, item => item).then(sum => {
    console.log("sum", sum);
    console.log(new Date() - start, "ms");
}, error => {
    console.log(error);
});
```

Ok, come on, are we seriously spending time debating the relative beauty of these two walls of unreadable code?

Yes, it turns out *in this one case* that Regenerator's code is shorter, but _I don't even care about that_. Minified and gzip'd, one of them is 350 bytes and the other is 357 bytes. That's a wash, folks.

**What I care about, and what Regenerator implicitly "cares" about, is runtime performance of the generated code.**

Here's the output of this program for Regenerator:

```sh
> npx regenerator perf.js > perf.regenerator.js
> node perf.regenerator.js
sum 4999950000
46 'ms'
```

46 milliseconds to sum 100,000 numbers. I would say "not bad," but how would I know? Let's compare to native `async`/`await`:

```sh
> node --version
v8.11.2
> node perf.js
sum 4999950000
33 'ms'
```

Cool, the Regenerator code took only 1.4x longer than native. I'll take it.

And now here's the output of the same program as compiled by `nodentjs`:

```sh
> npx nodentjs --out perf.js > perf.promises.js
> node perf.promises.js
TypeError: fn(...).then is not a function
    at $Loop_1 (/Users/ben/dev/babel/packages/babel-plugin-transform-async-to-promises/test/fixtures/loops/perf.promises.js:20:37)
    at b (eval at processIncludes (/Users/ben/dev/babel/packages/babel-plugin-transform-async-to-promises/test/fixtures/loops/node_modules/nodent-runtime/runtime.js:36:12), <anonymous>:3:486)
    at /Users/ben/dev/babel/packages/babel-plugin-transform-async-to-promises/test/fixtures/loops/perf.promises.js:28:94
    at boundThen (eval at processIncludes (/Users/ben/dev/babel/packages/babel-plugin-transform-async-to-promises/test/fixtures/loops/node_modules/nodent-runtime/runtime.js:36:12), <anonymous>:3:4435)
    at new Zousan (eval at processIncludes (/Users/ben/dev/babel/packages/babel-plugin-transform-async-to-promises/test/fixtures/loops/node_modules/nodent-runtime/runtime.js:36:12), <anonymous>:3:2373)
    at Function.$asyncbind (eval at processIncludes (/Users/ben/dev/babel/packages/babel-plugin-transform-async-to-promises/test/fixtures/loops/node_modules/nodent-runtime/runtime.js:36:12), <anonymous>:3:4518)
    at eachAsync (/Users/ben/dev/babel/packages/babel-plugin-transform-async-to-promises/test/fixtures/loops/perf.promises.js:33:8)
    at Object.<anonymous> (/Users/ben/dev/babel/packages/babel-plugin-transform-async-to-promises/test/fixtures/loops/perf.promises.js:41:1)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
```

Okay, this is my first time using `nodentjs`. Maybe I goofed something up. Can you spot the problem in the code above? Is it something trivial or fundamental?

Have a closer look at these lines:

```js
return fn(items[i]).then((function ($await_3) {
    sum += $await_3;
    return $Loop_1_step;
}).$asyncbind(this, $error), $error);
```

This comes from the following original code:

```js
sum += await fn(items[i]);
```

The `nodentjs` code seems to be assuming that any `await`ed argument must be a thenable object (like a `Promise`), but of course that's not necessarily true (that's one of the great things about `await`—it doesn't care), which is why Regenerator uses a runtime helper:

```js
  _context.next = 6;
  return regeneratorRuntime.awrap(fn(items[i]));

case 6:
  sum += _context.sent;
```

I don't mean to belabor this bug, but I will say it makes me wonder what other basic `async`/`await` functionality is not up to spec. I would also imagine this would bother @ljharb if he found out about it.

Okay, obviously I can fix the generated code by hand:

```js
return Promise.resolve(fn(items[i])).then((function ($await_3) {
    sum += $await_3;
    return $Loop_1_step;
}).$asyncbind(this, $error), $error);
```

Now, the code runs without an error, and we can finally see how runtime performance compares:

```sh
> node perf.promises.js
sum 4999950000
83 'ms'
```

1.8x longer than Regenerator's code, and 2.5x longer than native.

Look, folks, this is a microbenchmark for sure, and we've all happily used programming languages that execute code significantly slower than the equivalent C code, but I just don't see what anyone is talking about when they claim there is a runtime performance disadvantage to using Regenerator.

@ljharb I'll ask again. What on earth are you talking about?

shellscape:
holy wow batman

ljharb:
@benjamn indeed I’m talking about runtime performance, but as we will likely be eternally forbidding `await` in loops, that’s not a representative example of the code I’m hoping to transform. How do the two compare with examples that don’t actually require a finite state machine to transpile? (of course, this is all off topic for this PR, since this PR is about adding a transform *option* to babel, and is simply not the appropriate place at all to debate regenerator-runtime).

Good catch on that bug - @MatAtBread can we update this PR address the fact that all values, including `null`/`undefined` and non-thenables, are awaitable?

matAtWork:
For those that are interested. `npm i nodent`, then `node --expose-gc ./nodent.js tests tests/semantics/perf*`

Node v7 (earlier ones are similar)
```
                                                      {es7:true}          nodent          native
                  Compiler flags            Mean             173             380             660
               es7,lazyThenables             100             100               -               -
     es7,lazyThenables,es6target             106             106               -               -
     es7,lazyThenables,wrapAwait             189             189               -               -
azyThenables,es6target,wrapAwait             202             202               -               -
                             es7             187             187               -               -
                   es7,es6target             182             182               -               -
                   es7,wrapAwait             207             207               -               -
         es7,es6target,wrapAwait             210             210               -               -
                        promises             334               -             155             513
              promises,es6target             334               -             158             510
              promises,noRuntime             307               -             136             477
    promises,es6target,noRuntime             306               -             134             479
              promises,wrapAwait             347               -             183             512
    promises,es6target,wrapAwait             356               -             184             528
    promises,noRuntime,wrapAwait             327               -             159             495
es,es6target,noRuntime,wrapAwait             329               -             160             497
                      generators             890               -             820             960
            generators,es6target             933               -             846            1020
            generators,wrapAwait             914               -             795            1033
  generators,es6target,wrapAwait             861               -             831             892

```

Node v8 (and later)
```
                  Compiler flags            Mean               -             127             134
               es7,lazyThenables               -               -               -               -
     es7,lazyThenables,es6target               -               -               -               -
     es7,lazyThenables,wrapAwait               -               -               -               -
azyThenables,es6target,wrapAwait               -               -               -               -
                             es7               -               -               -               -
                   es7,es6target               -               -               -               -
                   es7,wrapAwait               -               -               -               -
         es7,es6target,wrapAwait               -               -               -               -
                        promises             106               -             100             112
              promises,es6target             109               -             101             116
              promises,noRuntime              94               -              86             101
    promises,es6target,noRuntime              94               -              86             103
              promises,wrapAwait             113               -             106             120
    promises,es6target,wrapAwait             112               -             106             118
    promises,noRuntime,wrapAwait              95               -              90             101
es,es6target,noRuntime,wrapAwait              97               -              90             104
                      generators             370               -             373             367
            generators,es6target             341               -             335             347
            generators,wrapAwait             361               -             364             357
  generators,es6target,wrapAwait             344               -             324             364
                          engine              47               -              47              48
                engine,es6target              47               -              47              48
                engine,noRuntime              48               -              48              47
      engine,es6target,noRuntime              47               -              47              48
                engine,wrapAwait              47               -              47              47
      engine,es6target,wrapAwait              47               -              48              47
      engine,noRuntime,wrapAwait              47               -              47              47
ne,es6target,noRuntime,wrapAwait              47               -              48              47
```
Note in Node v8, the "callbacks" method runs out of stack and fails. It's also heavily deprecated and you have to set all kinds of flags to resurrect in nodent. By node v10 native Promise performance has surpassed Zousan, but the native impl is twice as quick. Generators remain >3x times slower - it's just the way they're made. The test routines are simple loops (perf and perf-2) from [here](https://github.com/MailOnline/nodent/blob/master/tests)

shellscape:
Has a sense of humor been banned from this repo? Good lord folks, a meme is not spam. Especially when it's directly related to the previous reply, clean, and non-offensive.

benjamn:
Thanks for the thoughtful feedback, @ljharb and @MatAtBread.

Before I respond in more detail, I need to say something specifically to @MatAtBread:

I do love the idea of this transform. It's totally different from anything I imagined when I was implementing Regenerator, in part because I was focused on generator functions, and `async`/`await` was still just a thing C# programmers raved about. I’m amazed that your technique works so well, and I am genuinely glad you decided to pursue this alternative approach, because pushing the limits of transpilation is _fascinating_ to me, and that's exactly what you're doing. I would love to chat about this stuff with you any time.

> Additionally the `noRuntime` and `noNodentExtensions` compiler flags (which have the same defaults) removes any references and need for the runtime, at the expense of inlining code where necessary. Your choice to use the runtime or not is entirely personal - it has no effect on the execution semantics.

This is good to know, but the runtime is enabled by default with good reason: it tends to pay for itself, because compiling without the runtime produces more code:

```sh
% npx nodentjs --out perf.js | npx uglify-es -m -c | gzip | wc -c
     339
% npx nodentjs --out --noruntime --use=promises perf.js | npx uglify-es -m -c | gzip | wc -c
     400
```

With ~18% more code per `async` function (note: I'm sure this percentage varies), it doesn't take many `async` functions before the runtime begins making your bundle *smaller*. And at that point we might as well assume the runtime is a fact of life in most applications.

Again, I am not faulting `nodent` for these choices, but I reject them as a basis for favorable comparison to Regenerator.

> i want both plugins in core so that people are able to make that choice
> …
> this PR is about adding a transform _option_ to babel

@ljharb This plugin does not need to be included in the `@babel/` namespace in order for developers to have that choice.

Inclusion in core has less to do with enabling developer choice, and more to do with these two (real!) benefits:

* The Babel team is committing to helping maintain this plugin
* The Babel team approves of this plugin as an alternative to other core plugins

How many times have you claimed (to me and others) that developers don't know what's best for them, as a way of stressing the importance of spec compliance over developer choice?

We have slightly different priorities, you and I, but I think we can find a common ground. I firmly believe developers deserve to make informed choices, and I think you would too, provided those choices are truly "informed." If you agree with that, then I'm sure you can also agree they deserve to hear all the facts (more on this in my forthcoming review comments).

The truth is, developers who use this plugin will probably also enable Regenerator, so that generator functions (and `async` generator functions) will be supported as well as `async` functions. If they don't actually use any generator functions (perhaps because their linter forbids them), Regenerator will have nothing to compile, and `regenerator-runtime` will not be included. 🍰

benjamn:
@shellscape Take the compliment!

matAtWork:
....also, I agree with you about tests. The tests in [`nodent-compiler`](https://github.com/MatAtBread/nodent-compiler/tree/master/tests/semantics) are all execution tests which exercise a code path or represent an issue raised, fixed and closed. 

I believe the test set in Babel was copied from the existing async-to-generators package with the expected code generated by the plugin? @hzoo very kindly did this after my initial PR as I was a bit lost as to how the tests hung together.

matAtWork:
@benjamn - while we're building bridges (since my face is now all over your twitter feed), I've never compared anything unfavourably to to regenerator myself, as I don't use it. 

I have demonstrated above (and you can run for yourself in nodent's perf tests) that the TC-39  _generator_ implementation of `async/await` is pretty sub-optimal. I have no idea what optimisations or improvements are in `babel-transform-async-to-generator` as I've not looked, so I make no claims about that, regenerator or it's runtime. 

Please don't conflate other people's observations or opinions with my own. I totally agree with you about runtimes - I use nodent's runtime, but I was asked to make it optional for what, I assume, were good reasons by @hzoo (probably simplicity).

hzoo:
I apologize that this thread/PR got so long and delayed. I too want to see an async to promise transform (why we tried with https://www.npmjs.com/package/babel-plugin-async-to-promises). But I understand the concerns of others regarding the plugin itself since most of the logic isn't really in babel at all so none of us would really be able to handle the changes. `fast-async` does work in the meantime and we've barely had enough bandwidth to release v7 so I haven't really been paying attention to this as much. I appreciate the tons of work @matAtWork already put in. I know that porting all the logic into a pure babel plugin might not be feasible and not doing it may make us hesitant to merge this in the first place if it could be a separate plugin outside the repo.

Thanks for the points @benjamn, I made assumptions when initially looking to implement this. The original async-to-promises transform was purely written with Babel so it was what I assumed as well for fast-async.

> This transform really needs to be rewritten as a Babel plugin, as regenerator-transform has been 
> I too have Babel-compatible plugins that aren't really Babel plugins, such as babel-plugin-transform-es2015-modules-reify, but I believe that should disqualify them from inclusion in core

> The port, AFAIK is not quite a one-2-one renaming/substitution and might be tricky in places. The whole transform is complex (especially loops and exception handlers) and multi-pass which might not play nicely with the Babel way of doing things.

I understand that it might not be easy to do this with nodent though which is the issue despite the previous efforts.

Yeah, I don't believe we've had any babel plugins in "core" or the monorepo that just defered to another package that wasn't a babel plugin entirely itself. Based on this maybe it would make sense to just make it another package in the org or just update `fast-async` to v7? I believe this was the initial issue when looking at the implementation. So maybe just moving it under the org isn't much different than keeping it as `fast-async` for maintained there. I don't think anyone is opposed to the general idea of transforming async to promises though.

benjamn:
> while we're building bridges (since my face is now all over your twitter feed), I've never compared anything unfavourably to to regenerator myself, as I don't use it.

You're right, and I'm sorry. I have now deleted that tweet.

> I have demonstrated above (and you can run for yourself in nodent's perf tests) that the TC-39 generator implementation of `async`/`await` is pretty sub-optimal. I have no idea what optimisations or improvements are in `babel-transform-async-to-generator` as I've not looked, so I make no claims about that, regenerator or it's runtime.

For what it's worth, Regenerator's implementation is not exactly "a generator implementation of `async`/`await`". Regenerator code can be faster than `Promise`s + native generators (i.e. the TC39 reference implementation) because it compiles the generator functions and `yield` expressions down to ES5, which is often faster than native generators. JS engines have had many more years to figure out how to optimize `switch` statements.

Also, Regenerator can compile `async` functions directly to an ES5 state machine, without first creating a full generator function that will only be called once, which is typically what happens if you compile `async` functions to `Promise`s and native generator functions, and then compile the generator functions, in two separate compilation phases.

In other words, as you acknowledged above, the only way to compare performance against Regenerator would be to run code generated by Regenerator. I'm not asking you to do that work, I just want it to be clear that Regenerator is not the reference implementation.

> Please don't conflate other people's observations or opinions with my own.

Point taken. I will no longer assume that anyone else on this thread is speaking on your behalf.

matAtWork:
@benjamn - thanks, it's appreciated. 

Anytime you want to exchange insane AST manipulation stuff, I'm you new BFF. I might point you at a crazy repo I have which implements ["concatenative JS"](http://concatenative.org), because the first compiler I wrote was a FORTH complier (in about 1984).

Dan503:
> Have you guys looked at this plugin?
> https://github.com/rpetrich/babel-plugin-transform-async-to-promises

That plugin promises to turn async into promises but it does not fullfill it's promise. (😜)

It more turns it into a series of callbacks.

I prefer this one since it really does turn async calls into real promises (It can be a bit more restrictive in what it can translate though)
https://www.npmjs.com/package/babel-plugin-async-to-promises

Worthaboutapig:
> Indeed, a better name would have been `transorm-async-to-callbacks-pyramid-hell` but that would not sell :D
> 
> I saw the one by Kneden but it as this big warning on top

I made comment on this a while ago and took a look at improving the implementation, but haven't had time.

TBH, I'm entirely baffled by the whole `regenerator-runtime` choice in respect of `await-async`- nearly all browsers have supported `Promises` for years and those that don't can be polyfilled. Was `regenerator-runtime` written before the `Promises` spec came out?

dpraul:
I notice this PR has been inactive for a few months now - is there hope yet for this feature? We've been using `fast-async` in the meantime (with `transform-regenerator` and `transform-async-to-generator` disabled), but it would be nice to know if this will have proper support in the near future.

matAtWork:
Hi @dpraul - as far as I know, this is with the Babel guys to decide/approve

erykpiast:
Hello @mAAdhaTTah and Babel Team (@nicolo-ribaudo, @hzoo)! Do you plan to continue work on this PR? What are the pending issues? Do you still see a value in this concept? If you need any help, I'm here for you!

MatAtBread:
I have no plans to a work on it.

It was left to @hzoo and the Babel team to decide whether to incorporate it or not

nicolo-ribaudo:
https://github.com/babel/babel/pull/7076#pullrequestreview-128418552 is a good summary of the minimum steps that need to be done to move this forward. However,
1. This PR is meant to avoid depending on `regenerator-runtime` when using async functions. If you configure `@babel/preset-env` to target `>0.6%, not ie 11` it will not use `regenerator`, and if you lower your targets (`>0.5%` or accept `ie 11`) `regenerator` is included for browsers that don't support promises.
2. If this doesn't get accepted into this monorepo, nothing prevents anyone from releasing it as a third-party plugin, and use it disabling `transform-regenerator` and `async-to-generator` in `@babel/preset-env`'s options.

curiousdannii:
In the time that this PR has sat open the need for it has largely disappeared. Async function support is now up to above 90%. Still, thank you @MatAtBread for trying.

ljharb:
I don’t agree that the need has disappeared at all; I’d still very much like to see this land in core.

erykpiast:
There is no need to remove the current approach when this one would be merged. It may be an option like a loose mode is for many presets, don't you think? In that way, everybody may choose what is best for a particular use case.

nicolo-ribaudo:
> This transform really needs to be rewritten as a Babel plugin, as regenerator-transform has been for a long time (though not originally!), so that it can benefit from all the pipelining and NodePath/Scope-reuse optimizations that Babel's plugin system provides, and so that the Babel team can meaningfully maintain it, as you are implicitly asking them to do. Right now this plugin has a bus/vacation factor of 1, and that becomes a liability for Babel when it is included in core.

https://github.com/babel/babel/pull/7076#pullrequestreview-128418552

Dan503:
What would the use case be for preferring generators over promises though? 😕

erykpiast:
@nicolo-ribaudo I meant an option to the preset, similar to `modules` in `env` (as far as I understand it's a case of adding a plugin conditionally).

@Dan503 I suppose compliance with specs would be the benefit of the original generator approach. Maybe there is something more, like handling more complex cases. I'm thinking about nested and conditional awaits, async generators etc.

In general, what I would like to accomplish is not having to advise people dropping `async-await` syntax for simple cases, like unconditional series of asynchronous operations. I don't like multiple `then` callbacks like everybody else, but what I don't like even more is that whole regenerator runtime and a lot of boilerplate added to the bundle just to handle the very simple cases. I also understand the necessity of having a fallback solution for more complex code, constructs that we may even not dream about. That's why I see a space for two separate and independent solutions, possibly chosen by a preset option.

nicolo-ribaudo:
> I meant an option to the preset, similar to modules in env (as far as I understand it's a case of adding a plugin conditionally).

It doesn't matter if it's the default behavior or not. If it's not implemented as a Babel plugin we cannot maintain it. Nodent isn't a Babel plugin. What this PR does is:
1. When there is an async function, transform it into a source code string
2. Pass this string to the nodent compiler
3. Get a compiled source code from the node compiler, parse it and put it back in the original position

The whole compilation is opaque to us, and it's not something that we can maintain.
The current implementation (as a "fake" plugin) can be published to npm, and then you can do:
```js
{
  "presets": [
    ["@babel/preset-env", {
      "exclude": ["async-to-generator"]
    }]
  },
  "plugins": [
    "transform-async-with-nodent"
  ]
}
```

_We have nothing against transpiling async to promises_, but also _the community is free to publish a plugin outside of preset-env_.

MatAtBread:
Actually, that's not true. It passes the parsed AST to nodent, which does all the tree transforms itself (not using the Babel utils).

I'm all in favour of debate, but please base it on facts. It did considerable work on factoring it the transformer especially for this PR. It's up to you if you use it, but don't misrepresent what it does

nicolo-ribaudo:
Thanks, I updated my comment.

nicolo-ribaudo:
In the short term, we could link the third-party plugin in Babel's docs.

erykpiast:
Alright, I see the whole story now. If it's about maintainability and code ownership, maybe there is nothing wrong with rejecting this PR and continuing the work somewhere else? I suppose rewriting the code to Babel utils would require starting the work from scratch. It's far more than just polishing the code.

If we're focused on results more than on resolving technical or operational issues, maybe we should start from releasing a stable version of [`fast-async`](https://github.com/MatAtBread/fast-async#readme) working well with Babel@7, as a community plugin. Then we could try to measure its adaptation in the ecosystem and battle test in real-world projects. I'm not sure how wide is the usage of the plugin now, but NPM [shows rather low numbers](https://www.npmjs.com/package/babel-plugin-fast-async). Maybe refreshing the project and some marketing in social media would be enough to raise its popularity and get more feedback.

@nicolo-ribaudo, @MatAtBread, what do you think?

matAtWork:
Whatever is appropriate for you.

As for maintainability, it is a complex piece of code, and quite old (the original implementation is maybe 5 years old). However, it has a full test suite in the main compiler (nodent, as opposed no nodent-transform which is just the AST transformer) which covers many more test cases than babel's current set and it has been in use in production systems for many years. I've not had to do anything to it for months to meet edge cases or indeed the spec (there are compiler flags for strict and loose adherence to the spec, along with options for output to ES5, Promises, ES6, etc).

If I remember correctly, the babel version defaults to strict adherence to the spec with no runtime at all and using Promises.

All that having been said, like last time this came up, this seems to be a "political" or "philosophical" debate. It's not about the code or it's functionality (which is fully in line with the rest of the world), it's about whether making the option available to end users of babel should be easy (included) or hard (a comment somewhere in the docs) vs. the potential cost of maintenance. This is not something I have a view on as I'm not a babel maintainer.

erykpiast:
Matt, you're confusing me. Are you at work or having a sandwich break? 😐

MatAtBread:
Erm, both? It was lunch time in London.

I'm not trying to be confusing. I'm genuinely agnostic about whether Babel include the transformer or not. Most of my day job is backend, or if it's FE it's not for IE11, and I'm not a Babel user in general. So it really is whether there's a demand in the Babel user base for an async transformer with no runtime.

As an aside, when I first wrote it, it really was 10x quicker than regenerator & Promise polyfills, but now these are both implemented in major engines like V8, the point is moot as native is easily twice as quick as any userland implementations

Should you decide to integrate into Babel's core product, I'm happy to advise where I can, but it's not something I have an active interest in myself

MatAtBread:
I also think you're looking at the wrong plugin. Try https://www.npmjs.com/package/fast-async - that has 22k a week. I think someone built their own plugin on top of fast-sync for Babel & used the name

erykpiast:
Just a stupid joke, you use two GitHub accounts interchangeably. I suppose by accident, but I found it funny because of the names. Sorry about that.

Do you refer runtime speed, right? I'm more interested in optimizing bundle size. I didn't consider potential improvement or regression in execution, but my gut feeling is Promises winning here as well, there is less code to run. On the other hand, there is some cost of creating many lambdas for `then` callbacks... It's definitely something necessary to check out.

MatAtBread:
Oh yeah! I forgot. I have a paid-for-put-food-on-the-table and other stuff. It depends if I respond from my phone or desktop!

dpraul:
since this is showing activity again, just chiming in once more: if you're just trying to avoid using `regenerator-runtime` for promises, we've been using `fast-async` in production for a long time now with no issue.

<details><summary>babel.config.js</summary>

```js
module.exports = {
  plugins: [
    [
      '@babel/plugin-transform-regenerator',
      {
        asyncGenerators: false,
        generators: true,
        async: false,
      },
    ],
    [
      'module:fast-async',
      {
        spec: true,
      },
    ],
  ],
  presets: [
    [
      '@babel/preset-env',
      {
        exclude: [
          'transform-regenerator',
          'transform-async-to-generator',
        ],
      },
    ],
  ],
};

```
</details>

matAtWork:
@erykpiast - if you want the Babel-7 plugin, that I prepped for this PR, it's here: https://github.com/MatAtBread/fast-async/tree/babel-v7

matAtWork:
@nicolo-ribaudo The comment about modifying the Function.prototype is a fair one, but you have not used the super-clean & spec options to test. Here's is the output with no runtime at all (function protoypical or otherwise).

Of course, it's even bigger and more ugly, but somewhat easier to read that regenerators output. Of course, if you have a runtime that does generators, you don't need either.

```
```

http://nodent.mailed.me.uk/#(async%20()%20%3D%3E%20%7B%0A%20%20const%20response%20%3D%20await%20fetch(url)%3B%0A%20%20const%20asJSON%20%3D%20await%20response.json()%3B%0A%20%20%0A%20%20console.log(asJSON)%3B%0A%7D)()%3B%0A~options~%7B%22mode%22%3A%22promises%22%2C%22promiseType%22%3A%22Zousan%22%2C%22noRuntime%22%3Atrue%2C%22es6target%22%3Atrue%2C%22wrapAwait%22%3Atrue%2C%22spec%22%3Afalse%7D

coolgod:
What's the latest story about this PR? Seems it has been open for a long time.

nicolo-ribaudo:
You can use the `async-to-promises` plugin as shown in the config example at https://github.com/babel/babel/pull/7076#issuecomment-593412895

PR #7077, opened 2017-12-20
Fix typescript import eliding

Lodin:
another plugin

| Fixed Issues?            | No (perhaps)

Creating custom plugin to work with imports I found strange issue: if I recreate any import declaration or import specifier and replace original, instead of eliding import specifiers during typescript transformation it leaves all types as is.

I assume that it happes because current code uses bindings from scope to elide import specifiers. And binding points to specific object that dissappears from path when plugin replaces it. So even if bindings are removed import specifiers created by plugin are still in path and affect output code.

This patch fixes the issue.

**P.S.** I added two tests, because `elide-with-previous-source-rename` fails with error. They are connected, so if it is necessary, I can remove one.

### Example
Input:
```typescript
import {A, B, c, d} from "./lib";

const a: A = 1;
const b: B = {};
const cc = c;
const dd = d;
```
Plugin moves `A` and `d` to `./lib/foo`.
So input after plugin is following:
```typescript
import {A, d} from "./lib/foo";
import {B, c} from "./lib";

const a: A = 1;
const b: B = {};
const cc = c;
const dd = d;
```
Expected output:
```javascript
import { d } from "./lib/foo";
import { c } from "./lib";
const a = 1;
const b = {};
const cc = c;
const dd = d;
```
Actual output:
```javascript
import { A, d } from "./lib/foo";
import { B, c } from "./lib";
const a = 1;
const b = {};
const cc = c;
const dd = d;
```

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -42,7 +42,8 @@ export default function() {
         let allElided = true;
         const importsToRemove: Path<Node>[] = [];
 
-        for (const specifier of path.node.specifiers) {
+        for (let i = 0, len = path.node.specifiers.length; i < len; i++) {
+          const specifier = path.node.specifiers[i];
           const binding = path.scope.getBinding(specifier.local.name);
 
           // The binding may not exist if the import node was explicitly
@@ -50,9 +51,16 @@ export default function() {
           // of keeping scope bindings synchronized with the AST. For now we
           // just bail if there is no binding, since chances are good that if
           // the import statement was injected then it wasn't a typescript type
-          // import anyway.
+          // import anyway
           if (binding && isImportTypeOnly(binding, state.programPath)) {
-            importsToRemove.push(binding.path);
+            // Binding points to originally existing import specifier. If
+            // another plugin replaces original import declaration with custom
+            // one, removing bindings won't change this declaration
+            const specifierPath = path.get(`specifiers.${i}`);
+
+            if (specifierPath) {
+              importsToRemove.push(specifierPath);
+            }
           } else {
             allElided = false;
           }
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -42,7 +42,7 @@ export default function() {
         let allElided = true;
         const importsToRemove: Path<Node>[] = [];
 
-        for (let i = 0, len = path.node.specifiers.length; i < len; i++) {
+        for (let i = 0; i < path.node.specifiers.length; i++) {
           const specifier = path.node.specifiers[i];
           const binding = path.scope.getBinding(specifier.local.name);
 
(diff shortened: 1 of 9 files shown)


packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -42,17 +42,25 @@ export default function() {
         let allElided = true;
         const importsToRemove: Path<Node>[] = [];
 
-        for (const specifier of path.node.specifiers) {
+        for (let i = 0, len = path.node.specifiers.length; i < len; i++) {
```
nicolo-ribaudo: Nit: Usually engines cache the length, so the `len` variable isn't needed

Lodin: Thanks, I'll fix it

Lodin:
@nicolo-ribaudo Added tests. It works.

Also, do you have any idea why node 4 fails? I cannot get it.

nicolo-ribaudo:
Sorry, I don't have a clue :thinking: 
If you have `nvm` installed, you can try to debug them with node 4.

Lodin:
@nicolo-ribaudo Ok, I made it work. It was quite silly mistake :sweat_smile:

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -42,17 +42,25 @@ export default function() {
...
           // just bail if there is no binding, since chances are good that if
           // the import statement was injected then it wasn't a typescript type
-          // import anyway.
+          // import anyway
           if (binding && isImportTypeOnly(binding, state.programPath)) {
-            importsToRemove.push(binding.path);
+            // Binding points to originally existing import specifier. If
+            // another plugin replaces original import declaration with custom
+            // one, removing bindings won't change this declaration
+            const specifierPath = path.get(`specifiers.${i}`);
```
andy-ms: I don't understand how this works? How does a string like "specifiers.2" get inserted into the path?

hzoo: Hmm I guess it might be an array so it's actually `specifiers[i]` but the path version?

Lodin: As I understood [babel-handbook](https://github.com/thejameskyle/babel-handbook/blob/master/translations/en/plugin-handbook.md#get-the-path-of-sub-node), it is a method to get a path of sub-node. If I use `specifiers[i]`, it will return instance of `Node`, but I need `Path`.

andy-ms: Will accessing this by index still work if a plugin shuffled the imports around?
For that matter, what if a plugin changed the code so that an imported variable which used to be used only as a type is now used as a value? It doesn't seem like we can be resilient in the case of plugins doing arbitrary things, unless we get a fresh binding.

Lodin: @andy-ms I'm actually testing it with plugin that changes import ([elide-with-previous-change](https://github.com/Lodin/babel/blob/f7aca4b93bc8380a189b115f87b5110c00d3cfdf/packages/babel-plugin-transform-typescript/test/fixtures/imports/elide-with-previous-change/plugin.js)), because I found the problem creating this kind of plugin.

Unfortunately, cannot say anything about the situation you mentioned. Most likely the value will be removed. 

I think if there is a way to refresh bindings before start, it would be the best approach. Updating bindings will update paths in them which makes my changes redundant.

Lodin:
I just tested the original issue with `@babel/preset-typescript` of version `7.1.0` and it gets expected output. So I assume this PR can be closed now since original issue is fixed.

Issue #7067, opened 2017-12-19
Improve error message for `typescript` parsing of `import =`


loganfsmyth:
See discussion in https://github.com/babel/babel/issues/7062#issuecomment-352836023

The current error message of 

> `import =` is not supported.

is not enough information for users to know what their next step should be. Ideally we'd make that more descriptive.

gs-akhan:
Hi, 
Should I add a message like this ?

"import = is not supported. please consider changing your Typescript code to have ```import <specifier> from "resource" instead of ```import <specifier>  = require("resource");
refer link : https://www.typescriptlang.org/docs/handbook/modules.html#native-ecmascript-2015-modules-simplemodulejs"

gs-akhan:
Any thoughts? I can send a PR if that works .. :)
Thanks

loganfsmyth:
Maybe we can say

> "import =" is not supported in Babel's Typescript parser because there is no valid ES6 implementation for it. Please consider using `import <moduleName> from "<moduleName>";` alongside Typescript's `--allowSyntheticDefaultImports` option.

?

gs-akhan:
Sounds good. With a link to the TSDocs ?

loganfsmyth:
I don't think there are any good doc pages explaining `--allowSyntheticDefaultImports` other than http://www.typescriptlang.org/docs/handbook/compiler-options.html

gs-akhan:
Hmm yes. They can take user to some other level of confusion.
Should I send a PR ?

toubou91:
Shouldn't we close this one?

nicolo-ribaudo:
Yup

Issue #7062, opened 2017-12-19
Typescript `import =` is not supported.


rosskevin:
Choose one: is this a bug report or feature request?
Bug (or feature - because typescript support is new)

### Input Code
```js
import debounce = require('lodash/debounce')
```

### Babel/Babylon Configuration (.babelrc, package.json, cli command)

```js
{
  "presets": [
    "@babel/preset-typescript",
    "@babel/preset-env",
    "@babel/preset-stage-1",
    "@babel/preset-react"
  ]
} 
```

### Expected Behavior
From https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

> When importing a module using export =, TypeScript-specific import module = require("module") must be used to import the module.

### Current Behavior
```js
SyntaxError: src/transitions/Slide.tsx: `import =` is not supported.

    at File.buildCodeFrameError (/Users/kross/projects/material-ui/node_modules/@babel/core/lib/transformation/file/file.js:209:12)
    at NodePath.buildCodeFrameError (/Users/kross/projects/material-ui/node_modules/@babel/traverse/lib/path/index.js:155:26)
    at PluginPass.TSImportEqualsDeclaration (/Users/kross/projects/material-ui/node_modules/@babel/plugin-transform-typescript/lib/index.js:208:20)
    at newFn (/Users/kross/projects/material-ui/node_modules/@babel/traverse/lib/visitors.js:223:21)
    at NodePath._call (/Users/kross/projects/material-ui/node_modules/@babel/traverse/lib/path/context.js:64:19)
    at NodePath.call (/Users/kross/projects/material-ui/node_modules/@babel/traverse/lib/path/context.js:38:17)
    at NodePath.visit (/Users/kross/projects/material-ui/node_modules/@babel/traverse/lib/path/context.js:99:12)
    at TraversalContext.visitQueue (/Users/kross/projects/material-ui/node_modules/@babel/traverse/lib/context.js:139:18)
    at TraversalContext.visitMultiple (/Users/kross/projects/material-ui/node_modules/@babel/traverse/lib/context.js:93:17)
    at TraversalContext.visit (/Users/kross/projects/material-ui/node_modules/@babel/traverse/lib/context.js:178:19)
```

### Possible Solution
One idea would be that since this is a typescript only concept, and babel handles the other syntax well, just convert
```js
import debounce = require('lodash/debounce')
```
to
```js
import debounce from 'lodash/debounce'
```

### Context
[...]

PR #7079, opened 2017-12-20
Better error messages when Babel fails to parse import = and export =…

gs-akhan:
… syntax from typescript when using babel-plugin-transform-typescript

| Q | A
| ------------------------ | ------|
| Fixed Issues? | #7067
| Patch: Bug Fix? | No
| Major: Breaking Change? | No
| Minor: New Feature? | No
| Tests Added + Pass? | Yes
| Documentation PR | 
| Any Dependency Changes? | No
| License | MIT

The error messages displayed when Typescript's ```import = ``` and ```export = ``` were very less descriptive. Giving users no clue about what to do next..
Hence this messages can improve things regarding on how to fix their imports and exports in Typescript. 

Detailed discussion can be found at 
https://github.com/babel/babel/issues/7062
https://github.com/babel/babel/issues/7067

Thanks

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -206,7 +206,7 @@ export default function() {
 
       TSImportEqualsDeclaration(path) {
         throw path.buildCodeFrameError(
-          "`import =` is not supported by @babel/plugin-transform-typescript \n" +
+          "`import =` is not supported by @babel/plugin-transform-typescript\n" +
             "Please consider using " +
             "`import <moduleName> from '<moduleName>';` alongside " +
             "Typescript's --allowSyntheticDefaultImports option.",
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -206,8 +206,8 @@ export default function() {
 
       TSImportEqualsDeclaration(path) {
         throw path.buildCodeFrameError(
-          "`import =` is not supported in Babel's Typescript parser " +
-            "because there is no valid ES6 implementation for it.\nPlease consider using " +
+          "`import =` is not supported by @babel/plugin-transform-typescript \n" +
+            "Please consider using " +
             "`import <moduleName> from '<moduleName>';` alongside " +
             "Typescript's --allowSyntheticDefaultImports option.",
         );
@@ -215,8 +215,8 @@ export default function() {
 
       TSExportAssignment(path) {
         throw path.buildCodeFrameError(
-          "`export =` is not supported in Babel's Typescript parser because " +
-            "there is no valid ES6 implementation for it.\nPlease consider using `export <value>;`.",
+          "`export =` is not supported by @babel/plugin-transform-typescript\n" +
+            "Please consider using `export <value>;`.",
         );
       },
 
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -205,11 +205,19 @@ export default function() {
       },
 
       TSImportEqualsDeclaration(path) {
-        throw path.buildCodeFrameError("`import =` is not supported in Babel's Typescript parser because there is no valid ES6 implementation for it.\nPlease consider using `import <moduleName> from '<moduleName>';` alongside Typescript's --allowSyntheticDefaultImports option.");
+        throw path.buildCodeFrameError(
+          "`import =` is not supported in Babel's Typescript parser " +
+            "because there is no valid ES6 implementation for it.\nPlease consider using " +
+            "`import <moduleName> from '<moduleName>';` alongside " +
+            "Typescript's --allowSyntheticDefaultImports option.",
+        );
       },
 
       TSExportAssignment(path) {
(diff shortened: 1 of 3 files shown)


nicolo-ribaudo:
Can you run `make fix` to format the code?

gs-akhan:
Sure..

nicolo-ribaudo:
Also, it would be better to split the message on multiple lines (lines should be no longer than 80 chars)

e.g.
```js
throw buildCodeframeError(
  "bla bla bla ... .... " +
  "second line to make " +
  "them shorter"
);
```

gs-akhan:
On its way.. :)

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -205,11 +205,19 @@ export default function() {
       },
 
       TSImportEqualsDeclaration(path) {
-        throw path.buildCodeFrameError("`import =` is not supported.");
+        throw path.buildCodeFrameError(
+          "`import =` is not supported in Babel's Typescript parser " +
```
nicolo-ribaudo: I'd use `plugin` instead of `parser`, since it is parsed correctly but not transformed.

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -205,11 +205,19 @@ export default function() {
...
+            "`import <moduleName> from '<moduleName>';` alongside " +
+            "Typescript's --allowSyntheticDefaultImports option.",
+        );
       },
 
       TSExportAssignment(path) {
-        throw path.buildCodeFrameError("`export =` is not supported.");
+        throw path.buildCodeFrameError(
+          "`export =` is not supported in Babel's Typescript parser because " +
+            "there is no valid ES6 implementation for it.\nPlease consider using `export <value>;`.",
```
nicolo-ribaudo: Isn't `export =` more like `export default <value>` than `export <value>` (I don't use TS very much, so I'm not sure)

packages/babel-plugin-transform-typescript/src/index.js
gs-akhan: It is not. Because If you don't use any plugins. It breaks still..

nicolo-ribaudo: Or maybe `is not supported by @babel/plugin-transform-typescript`? :thinking:

packages/babel-plugin-transform-typescript/src/index.js
gs-akhan: Yes.  Basically you could only export one object earlier. Now support 
```export default <value>``` & ```export <value>```

File: packages/babel-plugin-transform-typescript/src/index.js (hunk 1 above)
Comment (nicolo-ribaudo) [reply packages/babel-plugin-transform-typescript/src/index.js]: > It is not. Because If you don't use any plugins. It breaks still..

 If you don' transform it, it works. https://runkit.com/embed/bsijdhpw0mq9

Comment (gs-akhan) [reply packages/babel-plugin-transform-typescript/src/index.js]: Not sure either. I wish Babel could parse this to "importdeclaration = Identifier"
I could write codemod to migrate all our legacyJS files to TS. :)

Comment (existentialism) [reply packages/babel-plugin-transform-typescript/src/index.js]: Yes, let's go with `is not supported by @babel/plugin-transform-typescript`.

Comment (gs-akhan) [reply packages/babel-plugin-transform-typescript/src/index.js]: Sure.. On its way

Comment (nicolo-ribaudo) [reply packages/babel-plugin-transform-typescript/src/index.js]: Just use `import identifier from "module"` :stuck_out_tongue: 

`import =` is only kept for backward compatibility, so it is "legacyTS" https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#1133-import-require-declarations

Comment (gs-akhan) [reply packages/babel-plugin-transform-typescript/src/index.js]: Alright Just a grammar review here before I add
```
          "`import =` is not supported by @babel/plugin-transform-typescript \n" +
            "Please consider using " +
            "`import <moduleName> from '<moduleName>';` alongside " +
            "Typescript's --allowSyntheticDefaultImports option.",
        );
```

File: packages/babel-plugin-transform-typescript/src/index.js [hunk 3]
```diff
@@ -205,11 +205,19 @@ export default function() {
       },
 
       TSImportEqualsDeclaration(path) {
-        throw path.buildCodeFrameError("`import =` is not supported.");
+        throw path.buildCodeFrameError(
+          "`import =` is not supported by @babel/plugin-transform-typescript \n" +
```
nicolo-ribaudo: You can remove the trailing space

gs-akhan: sure.

gs-akhan: @nicolo-ribaudo it was there earlier too. If I remove it on terminal the words look very closely attached. 
screen shot 2017-12-21 at 4 11 16 am

nicolo-ribaudo: Oh, I meant before `\n`

gs-akhan: Done.

gs-akhan:
Yaaaaay.. Cannot express how glad I am.. :)
Looking for many more PRs in future.. 
Thank you for all your precious time and efforts @nicolo-ribaudo & @existentialism

nicolo-ribaudo:
@gs-akhan This just need another approval, than can be merged 🙂

Issue #7078, opened 2017-12-20
Followup issue for extending built-ins


hzoo:
Ref https://github.com/babel/babel/pull/7020

Just making a new issue to track possible changes in another PR

- @Kovensky had a comment: https://github.com/babel/babel/pull/7020#discussion_r156837731
- @justinfagnani had a comment: https://github.com/babel/babel/pull/7020#discussion_r156742862

nicolo-ribaudo:
I have already a branch which implements @Kovensky commit, I didn't niclude it in #7020 to avoid introducing too many changes in one pr. https://github.com/nicolo-ribaudo/babel/commit/cbb31c390836b5156ff07a980b83dd1607786b6a

> @justinfagnani had a comment: https://github.com/babel/babel/pull/7020#discussion_r156742862

I can think of two possible implementations:
- Since Babel now supports JavaScript configuration files, the plugin could accept a function which tells which part of the `extend` clause needs to be wrapped.
  e.g.
  ```js
  module.exports = {
    plugins: [
      ["@babel/transform-classes", {
        wrapBuiltin(node) {
          if (node === somethingThatNeedsToBeWrapped) return true;
  
          // class extends mixin(HTMLElement)
          if (
            t.isCallExpression(node) &&
            t.isIdentifier(node.callee, { name: "mixin" }) &&
            t.isIdentifier(node.arguments[0], { name: "HTMLElement" }
          ) {
            return "arguments.0";
          }
        }
      }]
    ]
  };
  ```

- Use a special comment:
  ```js
  class A extends mixin(/*@babel wrapNative */ HTMLElement) {}
  ```

justinfagnani:
I think a simpler solution is to just allow a list of constructors that need to be wrapped in the configuration:

```js
module.exports = {
  plugins: [
    ["@babel/transform-classes", {
      alwaysWrap: ['HTMLElement']
    }]
  ]
};
```

All extendable built-ins are globals, so this should work fine.

cc @treshugart who ran into this issue with SkateJS. Since mixins are popular in the web components world, where we also need to extend native classes, it'd be really nice to have this fixed soon.

Issue #7022, opened 2017-12-13
How to fix this:Error in mounted hook: "TypeError: Class constructor Base cannot be invoked without 'new'"


NoraGithub:
I have searched related issues. And the most populate answer is to use preset to set the env not to transform es6 `class` to es2015. According to [issue#4269](https://github.com/babel/babel/issues/4269), [issue#338](https://github.com/Vincit/objection.js/issues/388)I think my setting should make sense to not to transform, but it doesn't.

My `.babelrc` setting is like this:
```js
{
  "presets": [
    ["env", { "modules": false ,
    "targets": {
        "browsers": ["> 1%", "last 2 versions", "not ie <= 8"],
        "node":"current"
      }
    }], "stage-2"
  ],
  "plugins": ["transform-runtime","syntax-dynamic-import"],
  "env": {
    "test": {
      "presets": ["env", "stage-2"],
      "plugins": ["istanbul"]
    }
  }

}
```

My code is like below:
```js
// a.js
// class definition
import Base form '@antv/g2/src/facet/base';
class Nested extends Base {
...
}
export default Nested;
```
```js
//b.js
import Nested from './nested';
G2.Chart.prototype.facet = function(type, cfg) {
...
const cls = new Nested(cfg);
...
}
export default G2;

```

And the browser report that `Error in mounted hook: "TypeError: Class constructor Base cannot be invoked without 'new'"`

And the related transformed code may be like this
```js
var Nested = function (_Base) {
  _inherits(Nested, _Base);

  function Nested() {
    _classCallCheck(this, Nested);

    return _possibleConstructorReturn(this, (Nested.__proto__ || Object.getPrototypeOf(Nested)).apply(this, arguments));
  }
...
```
| software         | version(s)
| ---------------- | -------
| Babel            |  "babel-preset-env": "^1.6.1"
| node             | 7.9.0
| npm              | 4.2.0
| Operating System | macOS 10.11.3 (15D21)

More Info:
```js
    "babel": "^6.23.0",
    "babel-cli": "^6.26.0",
    "babel-core": "^6.26.0",
    "babel-eslint": "^7.1.1",
    "babel-loader": "^7.1.2",
    "babel-plugin-syntax-dynamic-import": "^6.18.0",
    "babel-plugin-transform-runtime": "^6.22.0",
    "babel-preset-env": "^1.6.1",
    "babel-preset-stage-2": "^6.22.0",
    "babel-register": "^6.22.0",
```

Looking forward to your reply.

[...]

PR #7081, opened 2017-12-20
Allow extending native classes

nicolo-ribaudo:
| Fixed Issues?            | https://github.com/babel/babel/issues/7022
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR         | 
| Any Dependency Changes?  |

> Allow extending native classes
>
> Use `Reflect.construct(Super, ...)` instead of `Super.apply`
> when possible, because native classes can't be `.apply`ed.
>
> If a browser doesn't support `Reflect.construct`, it also doesn't
> support `class {}`.

This is needed when an external modules ships untranspiled classes but the consumer code transpiles them. (This should't happen with preset-env correctly configured, but if you use preset-es2015 your classes might be transpiled even if not needed)

Ref https://github.com/babel/babel/issues/7078 https://github.com/babel/babel/pull/7020#discussion_r156837731

cc @Kovensky


packages/babel-helpers/src/helpers.js
```diff
@@ -470,6 +470,16 @@ helpers.wrapNativeSuper = defineHelper(`
   }
 `);
 
+helpers.constructSuperInstance = defineHelper(`
+  export default function _constructSuperInstance(Derived, args, existingThis) {
+    const Super = Derived.__proto__ || Object.getPrototypeOf(Derived);
+    return typeof Reflect !== "undefined" && typeof Reflect.construct === "function" 
+      ? Reflect.construct(Super, args, existingThis.constructor || Derived)
```
Kovensky: I also said this in the comment where I suggested this to begin with, but this has the same caveat as the `new.target` transform: the `this.constructor` property _must not be tampered with_, as it's the only hint we have as to what the instance's constructor is in ES5. All other methods available that appear to work at first glance would break if the class is subclassed.

packages/babel-plugin-proposal-class-properties/test/fixtures/general/derived/expected.js
```diff
@@ -7,7 +7,7 @@ function (_Bar) {
     var _temp, _this;
 
     babelHelpers.classCallCheck(this, Foo);
-    return babelHelpers.possibleConstructorReturn(_this, (_temp = _this = babelHelpers.possibleConstructorReturn(this, (Foo.__proto__ || Object.getPrototypeOf(Foo)).call(this, ...args)), Object.defineProperty(babelHelpers.assertThisInitialized(_this), "bar", {
+    return babelHelpers.possibleConstructorReturn(_this, (_temp = _this = babelHelpers.possibleConstructorReturn(this, babelHelpers.constructSuperInstance(Foo, [...args], this)), Object.defineProperty(babelHelpers.assertThisInitialized(_this), "bar", {
```
Kovensky: Possible future microoptimization: `[...args]` -> `args`

Only possible if we can prove `args` is arraylike, though 🤷‍♀️

packages/babel-preset-es2015/test/fixtures/traceur/Classes/DefaultConstructor.js
```diff
@@ -18,9 +18,8 @@ assert.isFalse(
 assert.isFalse(
     Object.getOwnPropertyDescriptor(C.prototype, 'constructor').enumerable);
 
-// Ensure that we don't try to call super() in the default constructor.
 class D extends null {}
-var d = new D();
+assert.throws(() => new D());
```
Kovensky: I wonder what happened with TC39's proposal to actually make `extends null` work.

As it is implemented/specified right now, `extends null` is explicitly allowed as a special case, but it's impossible to actually construct such a class.

packages/babel-helpers/src/helpers.js
```diff
@@ -470,6 +470,16 @@ helpers.wrapNativeSuper = defineHelper(`
   }
 `);
 
+helpers.constructSuperInstance = defineHelper(`
+  export default function _constructSuperInstance(Derived, args, existingThis) {
+    const Super = Derived.__proto__ || Object.getPrototypeOf(Derived);
+    return typeof Reflect !== "undefined" && typeof Reflect.construct === "function"
```
Kovensky: While checking if my assumptions were really correct, I found that, while Node 4 has an (incomplete and almost unusable) implementation of classes, it also does not have `Reflect`, at all.

However, all that changes with this helper is that nothing changes on Node 4, you still can't extend native classes 😄

I also checked whether spreading `arguments` was safe; while Node 4 has no support for spread at all, `arguments` _does_ have a working `Symbol.iterator`, so we should be safe on that side as well if the user uses `transform-spread`.

packages/babel-plugin-transform-classes/README.md
```diff
@@ -6,8 +6,8 @@
 
 When extending a native class (e.g., `class extends Array {}`), the super class
 needs to be wrapped. This is needed to workaround two problems:
-- Babel transpiles classes using `SuperClass.apply(/* ... */)`, but native
-  classes aren't callable and thus throw in this case.
+- When extending native classes the `this.constructor` must not be tampered,
+  since it is the only way that Babel has of obtaining the class constructor.
```
Kovensky: This affects all classes, not just the native ones 🤔

Kovensky: This affects all classes, not just the native ones 🤔

nicolo-ribaudo: Oh dang, would it be a regression?

Kovensky: A tradeoff. The ability to mess with `this.constructor` for the ability to extend native classes.

I think it's fine to do it as a breaking change in babel 7.

Erasing `this.constructor` in itself won't break subclassing, because of the `|| Derived` in the helper:

```js
class A {}
class B extends A {}
delete B.prototype.constructor
;(new B()) instanceof B // still true
class C extends B {}
delete C.prototype.constructor
;(new C()) instanceof C // should also be true
```

But defining `constructor` to be something else can summon Cthulhu.

packages/babel-plugin-proposal-class-properties/test/fixtures/general/derived/expected.js
xtuc: Is there something that prevents doing it now? We could use `Array.isArray` and its helper.

Kovensky: @xtuc the idea is to do the optimization statically, though. It'd also miss out on arraylike that are not arrays

loganfsmyth:
Is this something we're still pursuing?

nicolo-ribaudo:
Yes: this PR allows libraries which are distributed with transpiled classes to be used alongise with untranspiled classes.
We could ask users to always transpile their classes in these cases, but I don't think that it would be the correct approach.

nicolo-ribaudo:
I will open another PR; this is too hard to rebase

PR #7086, opened 2017-12-21
TSTypeAliasDeclaration should reference TSTypeAnnotation node

JamesHenry:

@andy-ms As discussed on slack, but I'm honestly not sure how I feel either way anymore... 😅  

Obviously I am just striving for consistency here and there isn't a real downside to doing it I don't think. For that reason, I saw it through and opened the PR.

### DIFF
--- a/packages/babel-generator/src/generators/typescript.js
+++ b/packages/babel-generator/src/generators/typescript.js
@@ -348,7 +348,7 @@ export function TSTypeAliasDeclaration(node) {
   this.space();
   this.token("=");
   this.space();
-  this.print(typeAnnotation, node);
+  this.print(typeAnnotation.typeAnnotation, node);
   this.token(";");
 }
 
--- a/packages/babylon/src/plugins/typescript.js
+++ b/packages/babylon/src/plugins/typescript.js
@@ -854,7 +854,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       node.id = this.parseIdentifier();
       node.typeParameters = this.tsTryParseTypeParameters();
       this.expect(tt.eq);
-      node.typeAnnotation = this.tsParseType();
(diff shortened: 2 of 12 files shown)


nicolo-ribaudo:
Right after approving this PR a doubt came to my mind: aren't type annotations the types which follow `:`?

andy-ms:
@nicolo-ribaudo So far, yes, since we don't wrap types in `TsTypeAnnotation` if we're already in a type context such as for `A` in `A | B`. I don't really know what the right answer is for `type T = number` is though, since `number` is a top-level type and isn't itself part of some other type.

nicolo-ribaudo:
What does typescript do?

andy-ms:
It has its own AST format that never uses any `TsTypeAnnotation` nodes.

JamesHenry:
Yeah @nicolo-ribaudo as I mentioned above, it is not all that clear-cut. With the three approvals this still has, are we happy to press ahead with it? I think there is probably a bit of value in the consistency aspect. If yes, I will get the babel-types part updated.

packages/babel-generator/src/generators/typescript.js
```diff
@@ -348,7 +348,7 @@ export function TSTypeAliasDeclaration(node) {
   this.space();
   this.token("=");
   this.space();
-  this.print(typeAnnotation, node);
+  this.print(typeAnnotation.typeAnnotation, node);
```
loganfsmyth: This doesn't seem right to me. If you have to drill into the child node to get it to print properly, doesn't that mean the child node isn't representing the right thing?

loganfsmyth:
I'm triaging old PR on Babel. Since this has been sitting for ages, I'm going to close it. If it's a conversation you want to continue, feel free to re-open though.

PR #7090, opened 2017-12-22
Refactor config processing more

loganfsmyth:
A bunch of refactoring, but only one minor change (from previous Babel 7, not Babel 6) that was kind of a pain to do in the old architecture, but I think is more consistent in general anyway.

Currently we are caching plugins/presets that are passed in programmatically based on the identity of the arrays themselves. In previous iterations though, that was only the case for top-level programmatic options, and items in `env` were cached based on the identity of the `.env[foo]` object itself. Now programmatic args are all cached consistently based on array identity.

We can still probably revisit if array identity is the right call here. I'm feeling tempted to change that to be different, but that can wait.


loganfsmyth:
If someone wants to look at this I'm happy to respond to comments, but I think I'll just land it to get my next PR up for now.

PR #7091, opened 2017-12-22
Allow configs to have an 'overrides' array

### LINKED ISSUE #5451: [Feature Request] feature similar to ESLint "overrides" (config based on glob)
Following a discusion from #5420.

It would be usefull to provide an alternative configuration based on the path of the file being transpiled.

In the following example, only the `main.js` file would be using the `add-module-exports` plugin. The es2015 presets in used on all files.

```js
{
  presets: ['es2015'],
  files: {
    "./main.js": {
      plugins: ['add-module-exports'],
    },
  }
}
```

We should also be able to provide globs as path:

```js
{
  presets: ['es2015'],
  files: {
    "./*.js": {
      plugins: ['transform-remove-console'],
    },
  }
}
```

loganfsmyth:
| Fixed Issues?            | Fixes #5451
| Documentation PR         | 
| Any Dependency Changes?  |

This does two related things

1. Allow every config object to specify a `test/include/exclude` field, just like you might do for Webpack. Each item allows an item, or array of items that can be a string, RegExp, or function.

    Note, this means you could actually limit a `.babelrc` to only apply to _some_ files inside of a directory, or make an `env` override only apply to certain files in that env, or something, which seemed like it was reasonable to allow and would be an unnecessary restriction to disallow.

2. Allow an `overrides` array of sub-configs that apply on top of the base configs that will apply based on their `test`/`include`/`exclude` values.

As an example, this allows user to do

```
{
  presets: [
    ['env', { 
      targets: { node: 'current' },
    }],
  ],
  overrides: [{
    test: "./client",
    presets: [
      ['env', { 
        targets: { browsers: ["last 2 versions"] },
      }],
    ],
  }],
}
```
or for instance we'll use this in Babel's own codebase to do

```
{
  presets: [
    ['env', {modules: false}],
  ],
  plugins: [
    ['@babel/transform-modules-commonjs', { lazy: true }],
  ],
  overrides: [{
    test: './packages/babel-register',
    plugins: [
      // Override the plugin with an empty module options config to disable lazy modules.
      '@babel/transform-modules-commonjs',
    ],
  }],
}
```
to enable lazy modules for everything except `babel-register`, where it breaks things.

### DIFF
--- a/packages/babel-core/src/config/config-chain.js
+++ b/packages/babel-core/src/config/config-chain.js
@@ -40,7 +40,7 @@ export type PresetInstance = {
   dirname: string,
 };
 
-type ConfigContext = {
+export type ConfigContext = {
   filename: string | null,
   cwd: string,
   envName: string,
@@ -54,16 +54,26 @@ type ConfigContextNamed = {
 /**
  * Build a config chain for a given preset.
  */
-export const buildPresetChain = makeWeakCache(
-  ({ dirname, options, alias }: PresetInstance): ConfigChain => {
-    const result = createUncachedDescriptors(dirname, options, alias);
-    const { plugins, presets } = result;
-    return {
-      plugins: plugins(),
-      presets: presets(),
-      options: [normalizeOptions(result.options)],
-    };
-  },
+export const buildPresetChain: (
+  arg: PresetInstance,
+  context: *,
+) => * = makeChainWalker({
+  init: arg => arg,
+  root: preset => loadPresetDescriptors(preset),
+  env: (preset, envName) => loadPresetEnvDescriptors(preset)(envName),
+});
+const loadPresetDescriptors = makeWeakCache((preset: PresetInstance) =>
+  buildRootDescriptors(preset, preset.alias, createUncachedDescriptors),
+);
+const loadPresetEnvDescriptors = makeWeakCache((preset: PresetInstance) =>
+  makeStrongCache((envName: string) =>
+    buildEnvDescriptors(
+      preset,
+      preset.alias,
+      createUncachedDescriptors,
+      envName,
+    ),
+  ),
 );
 
 /**
@@ -72,14 +82,8 @@ export const buildPresetChain = makeWeakCache(
 export function buildRootChain(
   cwd: string,
   opts: ValidatedOptions,
-  envName: string,
+  context: ConfigContext,
 ): ConfigChain | null {
-  const context = {
-    filename: opts.filename ? path.resolve(cwd, opts.filename) : null,
-    cwd,
-    envName,
-  };
-
   const programmaticChain = loadProgrammaticChain(
     {
       options: opts,
--- a/packages/babel-core/src/config/config-chain.js
+++ b/packages/babel-core/src/config/config-chain.js
@@ -62,6 +62,9 @@ export const buildPresetChain: (
   init: arg => arg,
   root: preset => loadPresetDescriptors(preset),
   env: (preset, envName) => loadPresetEnvDescriptors(preset)(envName),
+  overrides: (preset, index) => loadPresetOverridesDescriptors(preset)(index),
+  overridesEnv: (preset, index, envName) =>
+    loadPresetOverridesEnvDescriptors(preset)(index)(envName),
 });
 const loadPresetDescriptors = makeWeakCache((preset: PresetInstance) =>
   buildRootDescriptors(preset, preset.alias, createUncachedDescriptors),
@@ -75,6 +78,30 @@ const loadPresetEnvDescriptors = makeWeakCache((preset: PresetInstance) =>
       envName,
     ),
   ),
+);
+const loadPresetOverridesDescriptors = makeWeakCache((preset: PresetInstance) =>
+  makeStrongCache((index: number) =>
+    buildOverrideDescriptors(
+      preset,
+      preset.alias,
+      createUncachedDescriptors,
+      index,
+    ),
+  ),
+);
+const loadPresetOverridesEnvDescriptors = makeWeakCache(
+  (preset: PresetInstance) =>
+    makeStrongCache((index: number) =>
+      makeStrongCache((envName: string) =>
+        buildOverrideEnvDescriptors(
+          preset,
+          preset.alias,
+          createUncachedDescriptors,
+          index,
+          envName,
+        ),
+      ),
+    ),
 );
 
 /**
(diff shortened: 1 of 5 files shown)


loganfsmyth:
This doesn't have any tests, but I'll try to add some tomorrow.

packages/babel-core/src/config/config-chain.js
```diff
@@ -419,7 +546,7 @@ function matchesPatterns(
 
     const absolutePatterns = strings.map(pattern => {
       // Preserve the "!" prefix so that micromatch can use it for negation.
-      const negate = pattern[0] === "!";
+      const negate = pattern[0] === "!" && !disableNegation;
```
nicolo-ribaudo: I think we should throw an error/report a warning if `disableNegate && pattern[0] === "!"`

packages/babel-core/src/config/validation/options.js
```diff
@@ -216,6 +243,9 @@ export function validate(type: OptionsType, opts: {}): ValidatedOptions {
     if (type === "env" && key === "env") {
       throw new Error(`.${key} is not allowed inside another env block`);
     }
+    if (type === "env" && key === "override") {
+      throw new Error(`.${key} is not allowed inside an env block`);
+    }
```
nicolo-ribaudo: Should we also disable `override` in another `override` block?

packages/babel-core/src/config/config-chain.js
xtuc: `!disableNegation` is not super clear to me but I don't have a better idea.

packages/babel-core/src/config/validation/options.js
xtuc: I would say yes.

packages/babel-core/src/config/config-chain.js
loganfsmyth: Good call, happy to throw instead.

packages/babel-core/src/config/validation/options.js
loganfsmyth: Good catch, definitely an oversight.

hzoo:
Awesome, yeah tests and some docs like https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns will be useful

> Allow every config object to specify a test/include/exclude

Does this include the top level config without the `overrides` or only in the overrides array config objects? And how do we expect this to work with `only/ignore`? Although I know they are different since one is about processing the files at all and the other is to change behavior

loganfsmyth:
> Does this include the top level config without the overrides or only in the overrides array config objects?

It includes any config object, including top-level, and `env` and those in `overrides`.

> And how do we expect this to work with only/ignore?

Currently, `test`/`ignore`/`exclude` are processed first, so if the test doesn't match, the `ignore` in that config object would never be applied.

Essentially I'm separating them as `ignore`/`only` means that a file essentially doesn't get transformed at all, whereas `test`/... says "don't use this specific config object".

loganfsmyth:
I think this should be good now.

hzoo:
Would be nice to be able to test this locally on our repo before merging (like all the other config changes)

loganfsmyth:
> Not sure we want to introduce all these different ways of doing it though? array/non-array, string, regex,fn, include/exclude/test

I went with it since it seemed like an established pattern in Webpack that people would be able to not have to think about a ton. The rest was taken from what we already did for ignore/only. Was there a trimmed down set of functionality you'd prefer?

hzoo:
I guess it's hard to know all usecases, so not sure - like monorepo/not, app/library, webpack/node, etc. Nothing in particular, if we think that kind of parallel is good? Just wondering if it's confusing to have that kind of nested level of config but I guess doesn't mean people are going to write it that way. We probably want to provide some level of docs on basic/advanced configs with webpack soon

loganfsmyth:
> Just wondering if it's confusing to have that kind of nested level of config but I guess doesn't mean people are going to write it that way.

Webpack is one of the few cases where people can have fine-grained control of transforms. I imagine this config will be more likely in use for people using the CLI or gulp or any number of other integrations. I also think it could be nice for Webpack since you can still centralize all your babel configs at least.

hzoo:
Ok cool, sounds good.

not really specific to the pr but: something that I was thinking of as a result of this is being able to get the files/transforms run in a debug mode for ourselves/others to better understand what babel is actually doing

packages/babel-core/src/config/config-chain.js
```diff
@@ -351,6 +472,42 @@ function dedupDescriptors(
...
+function configIsApplicable(
+  { options }: OptionsAndDescriptors,
+  dirname: string,
+  context: ConfigContext,
+): boolean {
+  return (
+    (options.test === undefined ||
+      configFieldIsApplicable(context, options.test, dirname)) &&
+    (options.include === undefined ||
+      configFieldIsApplicable(context, options.include, dirname)) &&
```
nicolo-ribaudo: What is the difference between `test` and `include`?

hzoo: I think it's the same actually just matching webpack

packages/babel-core/src/config/validation/options.js
```diff
@@ -216,6 +243,12 @@ export function validate(type: OptionsType, opts: {}): ValidatedOptions {
     if (type === "env" && key === "env") {
       throw new Error(`.${key} is not allowed inside another env block`);
     }
+    if (type === "env" && key === "overrides") {
+      throw new Error(`.${key} is not allowed inside an env block`);
+    }
+    if (type === "override" && key === "overrides") {
+      throw new Error(`.${key} is not allowed inside an overrides block`);
```
nicolo-ribaudo: Nit: You can write `throw new Error(".overrides is not allowed inside an overrides block");`, since `key` is guaranteed to be `overrides` because of the line above.

loganfsmyth: Yeah I just felt like it was more consistent this way.

packages/babel-core/src/config/config-chain.js
simonbuchan: Old webpack docs are gone, but I believe it used to be that `test` had to be a regex (e.g. `/\.js$/`), while `include`/`exclude` had to be a string (e.g. `srcDir`)? In any case, they have gained the connotations of "file type" and "directories".

novemberborn:
Is it expected that the overriding options can contain an `extends` clause, and the extended config can itself contain `overrides`, which are then evaluated?

Validation of `overrides` prevents such nesting, but it's currently allowed through `extends`. I've put an example up at https://gist.github.com/novemberborn/fdf830ce929a8641651f7838ea4efab1, compare `npx babel bar.js` with `npx babel foo.js`.

I'm not saying this behavior is necessarily wrong but it'd be good to know whether it's intentional.

xtuc:
Sorry for being late at this, but super job @loganfsmyth!

simonbuchan:
Reposting for visibility from my referencing comment above:

With this `babel` should now be able to behave like `tsc`, as babylon is perfectly happy to parse `<type> expr` with `syntax-typescript` so long as `syntax-jsx` is not also enabled.:

```js
const common = ["@babel/env"];
module.exports = {
  overrides: [
    { test: /\.js$/, presets: common },
    { test: /\.jsx$/, presets: [...common, "@babel/react"] },
    { test: /\.ts$/, presets: [...common, "@babel/typescript"] },
    { test: /\.tsx$/, presets: [...common, "@babel/typescript", "@babel/react"] },
  ]
}
```

Probably something to suggest in whatever typescript docs get added? What's the roadmap on docs anyway?

Issue #7093, opened 2017-12-22
[@babel/standalone] Remove the additional function context on DOMContentLoaded


stevefan1999:
x-pr from https://github.com/babel/babel-standalone/pull/97, too lazy to fork again.

stevefan1999:
This.

https://github.com/babel/babel/blob/master/packages/babel-standalone/src/index.js#L250

Just a little mistake to remind.

stevefan1999:
hmm, there is a problem. Since fat arrow propagates the context of `this`, I think that's the reason for it.

Or is that just a classical conditioning when engineers are dealing with event handlers?

existentialism:
Fixed via https://github.com/babel/babel/pull/7094, thanks @stevefan1999!

PR #7094, opened 2017-12-22
[@babel/standalone] Remove additional function context

stevefan1999:
| Fixed Issues?            |  https://github.com/babel/babel/issues/7093
| Documentation PR         | N/A
| Any Dependency Changes?  | N/A

### DIFF
--- a/packages/babel-standalone/examples/scriptTag-disabled.htm
+++ b/packages/babel-standalone/examples/scriptTag-disabled.htm
@@ -1,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8" />
+  <title>babel-standalone example - Script tag</title>
+</head>
+<body>
+  <pre id="output">This line should not be overwritten!</pre>
+
+  <script src="../babel.js"></script>
+  <script>Babel.disableScriptTags()</script>
+  <script type="text/babel">
+    const doStuff = () => {
+      document.getElementById('output').innerHTML = `I did an evil operation!`;
+    };
+    doStuff();
+  </script>
+</body>
+</html>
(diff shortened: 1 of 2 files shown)


nicolo-ribaudo:
Looking at the comment about the `transformScriptTags` function, it shuld be called with no arguments the first time so that every `<script>` tag is transformed:
https://github.com/stevefan1999/babel/blob/9c4cb9a320c28137c31b413adc285dfba81d5d66/packages/babel-standalone/src/index.js#L255-L261

stevefan1999:
@nicolo-ribaudo Yes, https://github.com/babel/babel-standalone/pull/97#pullrequestreview-85367986 pointed on the same finger. I was on my way to make a quickfix for that, somehow I felt asleep...

Daniel15:
This looks good to me. Did you test all the examples and ensure they still work? https://github.com/babel/babel/tree/master/packages/babel-standalone/examples

nicolo-ribaudo:
@stevefan1999 Can you run `make lint` to fix the linting errors?

stevefan1999:
`scriptTag-disabled.htm` should be able to describe the situation I'm into. Try doing that in an older `babel-standalone` package.

stevefan1999:
@existentialism Using an older version of `babel-standalone` in `scriptTag-disabled.htm` will overwrite the node even with `Babel.disableScriptTags()` disabled beforehand. Under this context, this is obviously not what we wanted!

tl;dr `Babel.disableScriptTags()` actually had never been worked before.

Daniel15:
Looks good to me! A unit test would be good too, but this is totally fine.

> tl;dr Babel.disableScriptTags() actually had never been worked before. 

It worked back when it was originally added 😛  It must have broken at some point later on, when an argument was added to `transformScriptTags`

PR #7098, opened 2017-12-23
Support 'assert and assign' TypeScript syntax

### LINKED ISSUE #7068: [TypeScript] "assert and assign" syntax fails to compile
Choose one: bug report

When performing an assertion and assignment at the same time, which is supported by the TypeScript compiler, the Babel TypeScript transformer throws an error.

### Input Code

```typescript
interface ResponseBody {
  name: string
  age: number
}

let response
;(response as ResponseBody) = {
  name: 'Alice',
  age: 23,
}
```

```json
{
  "plugins": ["@babel/plugin-transform-typescript"]
}
```

### Expected Behavior

Running the TypeScript compiler against the above input code yields the following output, and it seems reasonable to expect Babel to output something similar.

```js
var response;
response = {
    name: 'Alice',
    age: 23
};
```

### Current Behavior

Compilation fails with an `Invalid left-hand side in assignment expression` error:

```
./test.ts: Invalid left-hand side in assignment expression (7:2)
   5 |
   6 | let response
>  7 | ;(response as ResponseBody) = {
     |   ^
   8 |   name: 'Alice',
   9 |   age: 23,
  10 | }
error Command failed with exit code 1.
```

### Possible Solution

It should be supported by the transformer, or listed as an unsupported feature in the documentation and a more appropriate error can be thrown (as with `import = is not supported`) during compilation.

### Context

This syntax can be useful with frameworks like `koa` where you set a body like so:

```js
ctx.body = { name: 'Alice' }
```

With the "assert and assign" syntax this can be:

```typescript
;(ctx.body as ResponseBody) = { name: 'Alice' }
```

### Your Environment

| software                    | version(s)
| ----------------            | -------
| Babel                       | 7.0.0-beta.35
| plugin-transform-typescript | 7.0.0-beta.35
| typescript                  | 2.6.2
| node                        | 9.2.1
| npm                         | 5.5.1
| Operating System            | macOS 10.12.6

maaz93:
| Fixed Issues?            | Fixes #7068 

Support for the following TypeScript code in babylon

## Input
```ts
interface ResponseBody {
  name: string
  age: number
}
let response;
(response as ResponseBody) = {
  name: 'Alice',
  age: 23,
}
```
## Output
```js
let response;
response = {
  name: 'Alice',
  age: 23
};
```

### DIFF
--- a/packages/babylon/src/plugins/typescript.js
+++ b/packages/babylon/src/plugins/typescript.js
@@ -1867,11 +1867,12 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         case "TSParameterProperty":
           return super.toAssignable(node, isBinding, contextDescription);
         case "TSAsExpression":
-          return this.toAssignable(
+          node.expression = this.toAssignable(
             node.expression,
             isBinding,
             contextDescription,
           );
+          return node;
         default:
           return super.toAssignable(node, isBinding, contextDescription);
       }
--- a/packages/babylon/src/plugins/typescript.js
+++ b/packages/babylon/src/plugins/typescript.js
@@ -1867,7 +1867,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         case "TSParameterProperty":
           return super.toAssignable(node, isBinding, contextDescription);
         case "TSAsExpression":
-          return super.toAssignable(
+          return this.toAssignable(
             node.expression,
             isBinding,
             contextDescription,
--- a/packages/babylon/src/plugins/typescript.js
+++ b/packages/babylon/src/plugins/typescript.js
@@ -1866,6 +1866,12 @@ export default (superClass: Class<Parser>): Class<Parser> =>
           );
         case "TSParameterProperty":
           return super.toAssignable(node, isBinding, contextDescription);
+        case "TSAsExpression":
+          return super.toAssignable(
+            node.expression,
+            isBinding,
+            contextDescription,
+          );
         default:
           return super.toAssignable(node, isBinding, contextDescription);
       }
@@ -1889,6 +1895,14 @@ export default (superClass: Class<Parser>): Class<Parser> =>
             isBinding,
             checkClashes,
             "parameter property",
+          );
+          return;
+        case "TSAsExpression":
+          this.checkLVal(
+            expr.expression,
+            isBinding,
+            checkClashes,
+            contextDescription,
           );
           return;
         default:
--- a/packages/babylon/test/fixtures/typescript/cast/assert-and-assign/expected.json
+++ b/packages/babylon/test/fixtures/typescript/cast/assert-and-assign/expected.json
@@ -1,0 +1,435 @@
+{
+  "type": "File",
+  "start": 0,
+  "end": 131,
+  "loc": {
+    "start": {
+      "line": 1,
+      "column": 0
+    },
+    "end": {
+      "line": 10,
+      "column": 1
+    }
+  },
+  "program": {
+    "type": "Program",
+    "start": 0,
+    "end": 131,
+    "loc": {
+      "start": {
+        "line": 1,
+        "column": 0
+      },
+      "end": {
+        "line": 10,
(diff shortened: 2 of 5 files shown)


packages/babylon/src/plugins/typescript.js
```diff
@@ -1866,6 +1866,12 @@ export default (superClass: Class<Parser>): Class<Parser> =>
           );
         case "TSParameterProperty":
           return super.toAssignable(node, isBinding, contextDescription);
+        case "TSAsExpression":
+          return super.toAssignable(
```
nicolo-ribaudo: I think that here it should be `thsi.toAssignable`, to work with nested as expressions. e.g. `((a as T) as U) = b;`

maaz93: Yes! I missed that case. I'll make the change.

packages/babylon/test/fixtures/typescript/cast/assert-and-assign/expected.json
```diff
@@ -0,0 +1,435 @@
...
+              "line": 7,
+              "column": 0
+            },
+            "end": {
+              "line": 10,
+              "column": 1
+            }
+          },
+          "operator": "=",
+          "left": {
```
existentialism: This seems incorrect, I would expect a `TSAsExpression` node here.

packages/babylon/test/fixtures/typescript/cast/multiple-assert-and-assign/expected.json
```diff
@@ -0,0 +1,518 @@
...
+              "line": 8,
+              "column": 0
+            },
+            "end": {
+              "line": 11,
+              "column": 1
+            }
+          },
+          "operator": "=",
+          "left": {
```
existentialism: Same here

packages/babylon/test/fixtures/typescript/cast/assert-and-assign/expected.json
maaz93: I looked into it. I think I've misunderstood the problem. I was supposed to just support parsing `TSAsExpression` in the provided input code but I ended up transforming it (which is meant to be done [here](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-typescript/src/index.js#L228) I suppose).

Can you point me to an example of creating a new node so that I could do that here as well?

maaz93: I think I figured it out. In `toAssignable`, I don't need to go deep into the nested `expression`. Just do the following:

```js
case "TSAsExpression":
  return node;
```
Does that make sense?

nicolo-ribaudo: I think you should convert the expression:
```
case TSAsExpression:
  node.expression = this.toAssignable(node.expression);
  return node;
```

existentialism:
@maaz93 sorry for the delay, thanks!

packages/babylon/test/fixtures/typescript/cast/multiple-assert-and-assign/actual.js
```diff
@@ -0,0 +1,11 @@
+interface A {
+  name: string
+}
+interface B extends A {
+  age: number
+}
+let a;
+(a as A as B) = {
```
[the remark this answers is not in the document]
nicolo-ribaudo: https://github.com/babel/babel/issues/7175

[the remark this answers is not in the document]
maaz93: @andy-ms I'll submit a PR simplifying the tests. Thanks for the input 😄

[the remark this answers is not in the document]
nicolo-ribaudo: Assigning to `NaN` is disallowed, but you can use `(a as B as C) = d;`

PR #7101, opened 2017-12-24
Generate TypeScript typings, and improve generated Flow typings

bcherny:
| Documentation PR         | Maybe
| Any Dependency Changes?  |

----------------

Before, we were generating Flow typings from babel-types/definitions.

Now, we generate TypeScript typings in addition to the existing Flow typings. We also improve the quality of generated Flow typings to (1) properly type anything that was previously an `any`, and (2) 
remove unnecessary ?nullable types where keys are already nullable?.

We also generate Flow and TS types - and regenerate the README - as part of `make build` now.  As I understand it this will package the generated artifacts into @babel/types. This ensures that the types we ship are always up to date with what babel-types actually supports, and takes the burden/errorproneness of doing this away from future contributors.

Because the updated types are more constrained than they were before, this change should be *backwards compatible* (I'd like to verify this claim if anyone has a codebase that uses @babel/types@7's Flow typings.

Open questions:

- [x] As is good practice, we now copy the TS typings for babel-types into babel-types/lib as part of `make build`, and link to them from babel-types' package.json. Do we want to do this for Flow types too?
- [ ] I generated TS types without the `BabelNode` prefix that their Flow counterparts have in order to maintain compatibility with the existing [types on DefinitelyTyped](https://github.com/babel/babel/tree/master/packages/babel-types/src/definitions). Do we want to do this for Flow types too (nb: this would be backwards incompatible for Flow users)?
- [ ] Is it safe to remove (ie. stop generating) babel/lib/types.js, since it now lives in babel/packages/babel-types/lib?

Future enhancements:

1. Use babel-types to generate typings, instead of manual string munging.
2. Abstract out more common generation logic into scripts/generators/utils.js
3. Test both Flow and TS types as part of CI

See discussion on Slack here: https://babeljs.slack.com/archives/C0DFJT81H/p1514071757000033


bcherny:
cc @loganfsmyth

lib/types.d.ts
```diff
@@ -0,0 +1,1766 @@
...
+export function noop(): Noop;
+export function parenthesizedExpression(expression: Expression): ParenthesizedExpression;
+export function awaitExpression(argument: Expression): AwaitExpression;
+export function bindExpression(object: any, callee: any): BindExpression;
+export function classProperty(key: any, value?: Expression, typeAnnotation?: TypeAnnotation | TSTypeAnnotation | Noop, decorators?: Array<Decorator>, computed?: boolean, abstract?: boolean, accessibility?: "public" | "private" | "protected", optional?: boolean, readonly?: boolean, _static?: boolean): ClassProperty;
+export function decorator(expression: Expression): Decorator;
+export function doExpression(body: BlockStatement): DoExpression;
+export function exportDefaultSpecifier(exported: Identifier): ExportDefaultSpecifier;
+export function exportNamespaceSpecifier(exported: Identifier): ExportNamespaceSpecifier;
+export function tSParameterProperty(parameter: Identifier | AssignmentPattern, accessibility?: "public" | "private" | "protected", readonly?: boolean): TSParameterProperty;
```
nicolo-ribaudo: Ugh... Can you generate these names like I did in https://github.com/babel/babel/pull/6967?

lib/types.d.ts
```diff
@@ -0,0 +1,1766 @@
...
+  body: BlockStatement;
+  generator?: boolean;
+  async?: boolean;
+  returnType?: TypeAnnotation | TSTypeAnnotation | Noop;
+  typeParameters?: TypeParameterDeclaration | TSTypeParameterDeclaration | Noop;
+}
+
+export class Identifier extends Node {
+  type: "Identifier";
+  name: any;
```
nicolo-ribaudo: https://github.com/babel/babel/blob/d25918aa5db682581a94536ff5b2abb615438365/packages/babel-types/src/definitions/core.js#L385-L390

Can we also assert that this is a string?
e.g.
```js
{
  validate: chain(
    assertValueType("string"),
    function (node, key, val) { 
      if (!isValidIdentifier(val)) { 
        // throw new TypeError(`"${val}" is not a valid identifer name`); 
      } 
    },
}
```

scripts/generators/typescript.js
```diff
@@ -0,0 +1,150 @@
+"use strict";
+
+const fs = require("fs");
+const t = require("../../packages/babel-types");
+const { stringifyValidator } = require("./utils");
```
nicolo-ribaudo: Since this script should be usable on node 4, can you avoid destructuring?

lib/types.d.ts
bcherny: I agree that the `tS` prefix is ugly, and would also prefer `ts`. The downside is that would not be backwards-compatible (see [previous discussion](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/21972#issuecomment-351729657)).

As noted in that discussion, a good third option is a `TS` prefix - it's both backwards-compatible and consistent, and as an added benefit it shares its type's name. A downside is that Flow uses a single namespace for types and values, and will not let us export both a function and a type with, say, the name `TSParameterProperty`. Here's my suggestion:

1. For now, let's rename all of these to `TS...` in just the TS typings. For example, `tSParameterProperty` would become `TSParameterProperty`
2. When we're ok shipping a breaking version, let's rename all of them to `ts...` in both the TS and Flow typings. For example, `tSParameterProperty` would become `tsParameterProperty`

Of course, if we're shipping this as part of v7 (which is breaking anyway), let's do (2) now and skip (1).

scripts/generators/typescript.js
bcherny: Done

lib/types.d.ts
bcherny: Done

bcherny:
BTW, @nicolo-ribaudo is it safe to remove (ie. stop generating) babel/lib/types.d.ts and babel/lib/types.js, since both of them now live in babel/packages/babel-types/lib? It seems safe to me, but I want to make sure I'm not breaking a use case I don't know about.

lib/types.d.ts
nicolo-ribaudo: Since we are still in beta, we can do (2).

bcherny: Done for TS and Flow, and updated docs.

bcherny:
> +1 from me for removing the BabelNode prefix, but I don't know why it was added (so you should wait for some other maintainer opinion)

Looks like the prefix was [introduced](https://github.com/babel/babel/blob/6467bf6809f7f500c6e744ce6858c62fd93c8455/scripts/generate-interfaces.js) by @kittens >2 years ago. @kittens any chance you remember why you added it?

bgw:
Thanks for improving the flow definitions! I've been fighting with these definitions while working on a code generation script I have.

> +1 from me for removing the BabelNode prefix, but I don't know why it was added (so you should wait for some other maintainer opinion)

The flow types [aren't wrapped in a `declare module` block](https://flow.org/en/docs/libdefs/creation/), so they pollute the global namespace. If you add the babel types to your project's `lib` directory, these type definitions are now in flow's global scope. The `BabelNode` prefix was probably added to mitigate that, but it would be better (IMO) if it was wrapped as a module. However, that would require changes to everything that uses these types, so that's probably beyond the scope (no pun intended) of this PR.

> Because the updated types are more constrained than they were before, this change should be backwards compatible (I'd like to verify this claim if anyone has a codebase that uses @babel/types@7's Flow typings.

These changes aren't backwards-compatible. If I have code that uses the types incorrectly, it would pass with the old definitions (because `any` ignores type errors), but fail with the new ones. It's near-impossible to make any changes to flow types in a completely backwards compatible way. Even flow itself struggles with this: most releases have some small breaking changes. Small breaking changes are par for flow typedefs.

However tighter type constraints (as long as they're accurate) are definitely better, and finding and fixing type errors this may expose shouldn't be hard. I'm strongly in favor of this change, even though it is a breaking change.

bcherny:
This PR is going into Babel@7, which is not only a breaking major version, but also a new package name. This is a good chance to clean up the typings, if we were planning on it anyway. I suggest we do 2 things (neither of which has to be part of this PR):

1. Remove the `BabelNode` prefix for generated Flow types, and use a `declare module` block @bgw 
2. Add a `Flow` prefix to all Flow Nodes, just like we do for `TS` already. This will make it much easier to understand a given node and its types at a glance.

In the meantime, this PR is ready to go.

@nicolo-ribaudo @existentialism

nicolo-ribaudo:
> 1. Remove the `BabelNode` prefix for generated Flow types, and use a `declare module` block
> 2. Add a `Flow` prefix to all Flow Nodes, just like we do for `TS` already. This will make it much easier to understand a given node and its types at a glance.

+1 from me, but I'd prefer that changes to be in a different PR to not block this one.

> BTW, @nicolo-ribaudo is it safe to remove (ie. stop generating) babel/lib/types.d.ts and babel/lib/types.js, since both of them now live in babel/packages/babel-types/lib? It seems safe to me, but I want to make sure I'm not breaking a use case I don't know about.

I would say yes, because users can't install the top-level `babel` folder so it isn't included in any package.

bcherny:
@nicolo-ribaudo Done. Who else do you think should chime in here?

bcherny:
Pinging @loganfsmyth @jridgewell

scripts/generators/typescript.js
```diff
@@ -0,0 +1,155 @@
...
+
+const t = require("../../packages/babel-types");
+const utils = require("./utils");
+
+const NODE_PREFIX = "";
+
+let code = `// NOTE: This file is autogenerated. Do not modify.
+// See scripts/generators/typescript.js for script used.
+
+export class ${NODE_PREFIX}Comment {
```
Kovensky: Please don't use `class`. This means  that `typeof Comment` is a constructor, that you can legitimately call `new Comment` and it will produce an object with those properties, and that you can do `instanceof Comment` to check if something is assignable to a `Comment`.

These all should be `interface`s.

Kovensky: There is no valid `Comment` without a `type`, so a better definition would be something like:

```ts
interface Comment<T extends string> {
  type: T
  value: string
  ...
}

export interface BlockComment extends Comment<"BlockComment"> {}
export interface LineComment extends Comment<"LineComment"> {}
```

scripts/generators/typescript.js
```diff
@@ -0,0 +1,155 @@
...
+    column: number;
+  };
+
+  end: {
+    line: number;
+    column: number;
+  };
+}
+
+export class Node {
```
Kovensky: Same here, but I'd actually like a slightly more complex setup, if possible.

```ts
// these can be accumulated and just written as a declaration at the end of file;
// typescript type definitions have no ordering (think of them as "always hoisted")
type NodeTypes = 'Identifier' | 'BlockStatement' | 'FunctionDeclaration' | ...
type AnyNode = Identifier | BlockStatement | FunctionDeclaration | ...

interface Node<T extends NodeTypes> {
  type: T;
  // NOTE: it is possible to make a node that babel will accept as
  // valid and not have leadingComments on it.
  // Maybe it (and all keys besides `type`) should be marked optional.
    leadingComments: ReadonlyArray<Comment> | null;
  ...
}
```

When a interface for a specific type is generated, say `Identifier`, it should be generated as:

```ts
interface Identifier extends Node<'Identifier'> {
  name: string
}
```

scripts/generators/typescript.js
```diff
@@ -0,0 +1,155 @@
...
+    .forEach(fieldName => {
+      const field = fields[fieldName];
+      const typeAnnotation = utils.stringifyValidator(
+        field.validate,
+        NODE_PREFIX
+      );
+
+      args.push(
+        t.toBindingIdentifierName(fieldName) +
+          (isNullable(field) ? "?: " : ": ") +
```
Kovensky: To be sure, does babel make absolutely no distinction between `null` and `undefined`, or is `null` distinct? `FunctionDeclaration` `id`s come to mind.

scripts/generators/typescript.js
Kovensky: That way of writing a declaration is a bit more elaborate; but this means that you are able to use `switch` / `if` or other type narrowing operators to narrow the node type _without_ requiring any type assertions.

```ts
function rename(node: AnyNode, name: string) {
  if (node.type === 'Identifier') {
    // typeof node is Identifier
    return t.clone(node, { name })
  }
  // typeof node is still AnyNode but cannot be Identifier anymore
  if (node.type === 'FunctionDeclaration') {
    return t.cloneDeep(node, { id: t.identifier(name) })
  }
  // typeof node is AnyNode minus Identifier | FunctionDeclaration
  // ...
  return node
} // if t.clone{,Deep} are typed correctly, inferred return type for rename is still AnyNode
```

scripts/generators/typescript.js
```diff
@@ -0,0 +1,155 @@
...
+          fieldName +
+            ": " +
+            typeAnnotation +
+            (isNullable(field) ? " | null" : "") +
+            ";"
+        );
+      }
+    });
+
+  code += `export class ${NODE_PREFIX}${type} extends Node {
```
Kovensky: Again, never declare classes that are not classes. These are interfaces.

scripts/generators/typescript.js
```diff
@@ -0,0 +1,155 @@
...
+    decl += `node is ${NODE_PREFIX}${t.TYPES[i]};`;
+  } else {
+    decl += `boolean;`;
+  }
+
+  lines.push(decl);
+}
+
+lines.push(
+  `export function validate(n: BabelNode, key: string, value: mixed): void;`,
```
Kovensky: There is no `BabelNode` type anymore

scripts/generators/typescript.js
```diff
@@ -0,0 +1,155 @@
...
+  `export function clone<T>(n: T): T;`,
```
Kovensky: This is missing the second argument.

`T` needs to `extend Node`, and the second argument should be `Partial<T>` (lazy typing), or a hack similar to what is used for `setState` in `@types/react` should be used to correctly forbid unintentionally using `undefined` as a value.

scripts/generators/typescript.js
```diff
@@ -0,0 +1,155 @@
...
+  `export function cloneDeep<T>(n: T): T;`,
+  `export function removeProperties<T>(n: T, opts?: {} | null): void;`,
+  `export function removePropertiesDeep<T>(n: T, opts?: {} | null): T;`,
+  `export type TraversalAncestors = Array<{
+    node: BabelNode,
+    key: string,
+    index?: number,
+  }>;
```
Kovensky: `ReadonlyArray`

bcherny:
@Kovensky I generated TS types exactly the same way we generate Flow types already, and made no effort to verify correctness for the kinds of things you pointed out. I agree that it is important for the canonical types to be really accurate. Will go through and address your comments today/tomorrow.

scripts/generators/typescript.js
bcherny: The only distinction I can find is when a field has a default value, `null` doesn't get parsed into `undefined` and instead throws an error. Updated.

-------

Here's the result of constructing a few types:

- If field is `optional`:
  - `undefined` Ok, becomes `null`
  - `null` Ok
- If field has a `default`:
  - `undefined` Ok, uses default values
  - `null` Error
- Otherwise, if field is required:
  - `undefined` Error
  - `null` Error

bcherny:
@Kovensky All comments are addressed. I couldn't agree more with your idea of using tagged unions for better type inference in TS, though I did it in a way that is a bit simpler. 

After this PR is in, I'd love to comb through and improve the Flow types too. We should also dogfood all of this stuff by  using the babel-types API for code generation and enabling Flow for the code generator as a next step.

bcherny:
Pinging @Kovensky

packages/babel-types/src/definitions/es2015.js
```diff
@@ -399,19 +399,27 @@ export const classMethodOrPropertyCommon = {
     optional: true,
   },
   key: {
-    validate: (function() {
-      const normal = assertNodeType(
+    validate: chain(
```
nicolo-ribaudo: Why are these changes needed?

scripts/generators/typescript.js
```diff
@@ -0,0 +1,186 @@
...
+
+let code = `// NOTE: This file is autogenerated. Do not modify.
+// See scripts/generators/typescript.js for script used.
+
+export interface Comment {
+  value: string;
+  start: number;
+  end: number;
+  loc: SourceLocation;
+  type: "BlockComment" | "LineComment";
```
Kovensky: This type prevents refinements (if you have something of type Comment, you can't compare its `type` to refine the type).

It's not harmful in practice I guess since both BlockComment and LineComment don't actually add anything to the Comment type.

scripts/generators/typescript.js
```diff
@@ -0,0 +1,186 @@
...
+}
+
+interface BaseNode {
+  leadingComments: ReadonlyArray<Comment> | null;
+  innerComments: ReadonlyArray<Comment> | null;
+  trailingComments: ReadonlyArray<Comment> | null;
+  start: number | null;
+  end: number | null;
+  loc: SourceLocation | null;
+  type: Node["type"];
```
Kovensky: ...but it is here. If you have a `BaseNode`, comparing its `type` is completely useless.

Though arguably you're supposed to use the `t.isFoo` functions instead; and this interface is not exported so users can't use it by accident.

scripts/generators/typescript.js
```diff
@@ -0,0 +1,186 @@
...
+      struct.push(`${fieldName}: ${typeAnnotation};`);
+    }
+  });
+
+  code += `export interface ${type} extends BaseNode {
+  ${struct.join("\n  ").trim()}
+}\n\n`;
+
+  // super and import are reserved words in JavaScript
+  if (type !== "Super" && type !== "Import") {
```
Kovensky: You could generate something like:

```ts
function ${utils.toFunctionName(type)}Factory(${args.join(", ")}): ${type};
export { ${utils.toFunctionName(type)}Factory as ${utils.toFunctionName(type)} }
```

packages/babel-types/src/definitions/es2015.js
Kovensky: It's for the validator parser to be able to guess the apparent type of the `key` key. As it is right now, it's just an opaque arbitrary function; using `chain` and `assertNodeType` lets the type generator generate `Identifier | StringLiteral | NumericLiteral | Expression`.

scripts/generators/typescript.js
Kovensky: I'm actually not sure how safe it is in practice to write it like this because of the circular dependency. `BaseNode` depends on `Node` which depends on (list of interfaces all of) which depend on `BaseNode`.

bcherny: It seems to compile fine. `BaseNode` is not exported, but `Node` (which is refinable union of tagged types) is.

scripts/generators/typescript.js
bcherny: @Kovensky I'm happy to do this after this PR is merged. This shouldn't block this PR.

scripts/generators/typescript.js
bcherny: Good catch - updated.

bcherny:
@nicolo-ribaudo @Kovensky I believe this PR is ready to go. Any other enhancements can go into separate PRs - this one has a lot of commits as-is.

existentialism:
We should replace the ref to `lib/types.js` inside .flowconfig?

bcherny:
@existentialism Good catch - done.

bcherny:
@nicolo-ribaudo @existentialism Can we merge this?

loganfsmyth:
This appears to have broken Flow typechecking for babel-core and babel-template FYI. Both of those relied on the BabelNodeX types and haven't been updated to reflect that all that was moved into babel-types.

bcherny:
@loganfsmyth Thanks for the heads up. 2 questions:

1. How do we tell babel-core and babel-template to look in the new location? I don't see where the path is explicitly mentioned.
2. What is a good way to add this to CI? One option would be to add `flow-bin` and `typescript` as `devDependencies` and run both on some set of files (how do we define this set) as part of CI.

loganfsmyth:
1. It was working because the `lib/` folder was set as a global source of types, so they were available to all monorepo packages. Now each of those will need to explicitly import the types from babel-types.
2. I was under the impression we _were_ running Flow in CI, but I guess not haha. We should definitely be running Flow in CI. Babel itself is typechecked with Flow, so I don't think there's anything useful we could do in CI to check the typescript output though.

loganfsmyth:
Oh also, we also currently delete everything in `packages/*/lib` in https://github.com/babel/babel/blob/master/Makefile#L43 when running `make watch`, so when the watcher starts it deletes the type definition file that would be needed for Flow to typecheck during local development. We'll probably want to rebuild that when the watcher starts.

existentialism:
We _are_ running flow in CI, but it's been silently broken (since this PR actually)... it currently reports zero errors on travis, whereas locally it has a bunch.

Ref: https://github.com/babel/babel/pull/7227

bcherny:
Good catch @loganfsmyth and @existentialism, and thanks for the fix.

bcherny:
For reference, here's the fix to bootstrap before test: https://github.com/babel/babel/pull/7353.

Though for some reason, that still passes CI.

PR #7105, opened 2017-12-25
Add @babel/plugin-transform-named-capturing-groups-regex

nicolo-ribaudo:
| Fixed Issues?            | https://github.com/babel/proposals/issues/35
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR         | 
| Any Dependency Changes?  | `regexp-tree`

When the `runtime` flag is on, this plugin adds a new helper which wraps the native `RegExp` class to provide groups support. People nees to use a polyfill (I implemented it in core-js) for browsers that don't support overwriting regexp methods.


packages/babel-plugin-transform-named-capturing-groups-regex/test/fixtures/syntax/.escape-no-unicode-flag/actual.js
```diff
@@ -0,0 +1 @@
+/(?<\u{41}>)\k<\u{41}>/;
```
nicolo-ribaudo: @DmitrySoshnikov I think that `regexp-tree` should throw on this regex, since unicode escapes are only alloed when there is the `u` flag?

DmitrySoshnikov: Yeah, @Golmote added generic support for Unicode to parser (https://github.com/DmitrySoshnikov/regexp-tree/issues/114); we should also check the group names, if they are allowed only with `u` flag (since by itself `\u{41}` can normally be used without `u` flag -- "`u` repeated 41 times").

If you're positive from spec, that it should throw, we should open an issue. Thanks for catching it!

Golmote: I'm pretty sure identity escapes don't throw without the `u` flag.

DmitrySoshnikov: @Golmote, right, however this one is specifically for a capturing group name, like in `(?<\u{41}>)`. I'll need to double-check the spec, but seems, yes, it should throw, and be an invalid name for a group if no `u` flag is passed.

Golmote: Sorry I read too fast 😳

packages/babel-plugin-transform-named-capturing-groups-regex/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "homepage": "https://babeljs.io/",
+  "license": "MIT",
+  "main": "lib/index.js",
+  "keywords": [
+    "babel-plugin",
+    "regex",
+    "regexp",
+    "regular expressions"
+  ],
+  "repository": "https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-named-capturing-groups-regex.git",
```
xtuc: Can you please remove the `.git` as the end here?

nicolo-ribaudo:
The tests are failing because `regexp-tree` only works on node >= 6.
@mathiasbynens  I will work on implementing named groups in regexpu-core (I already opened the parser PR some days ago: https://github.com/jviereck/regjsparser/pull/83)

DmitrySoshnikov:
> The tests are failing because regexp-tree only works on node >= 6.

Thanks for the notice! Just an FYI, I removed `node >= 6` restriction in `0.0.78`, and now support older Node versions.

nicolo-ribaudo:
Proposal merged - https://github.com/tc39/ecma262/pull/1027#issuecomment-359554431

packages/babel-helpers/src/helpers.js
```diff
@@ -814,3 +814,125 @@ helpers.applyDecoratedDescriptor = defineHelper(`
...
+  }
+
+  function groups(r, re) {
+    return Object.keys(re["@@babel/groups"]).reduce((g, n) => {
+      g[n] = r[re["@@babel/groups"][n]];
+      return g;
+    }, Object.create(null));
+  }
+
+  wrap(RegExp.prototype, "exec", function(f, str) {
```
Andarist: This is effectful, I understand that it's hard (impossible?) to implement this feature without actually messing with built in prototypes - but when that is the case I feel like it should not be a part of babel-helpers, but rather babel-polyfill

packages/babel-helpers/src/helpers.js
```diff
@@ -814,3 +814,125 @@ helpers.applyDecoratedDescriptor = defineHelper(`
         return desc;
     }
 `);
+
+helpers.wrapRegExp = defineHelper(`
+  var kGroups = Symbol("@@babel/groups");
+
+  function BabelRegExp(re, groups) {
```
Andarist: could this be wrapped in an iife? all those assignments at the moment will prevent tree-shaking if `BabelRegExp` class stays unused. Im not sure if we can annotate helper as `#__PURE__` at the moment though (I think comments in the helpers are ignored)

Andarist: we could probably also use a redefining helper technique here to contain this whole declaration within a single top level node

it seems though that it would actually be cleaner if we could transpile helpers with user settings - this could be written as an actual class

DmitrySoshnikov:
Is this PR ready to me merged?

packages/babel-helpers/src/helpers.js
nicolo-ribaudo: @zloirock Does corejs polyfill support for `[String.replace]` and overwritten `exec`? If it does, we can remove thi "compat" helper from Babel.

zloirock: @nicolo-ribaudo `core-js` supports `@@replace`, but does not overwrite `exec`.

nicolo-ribaudo: Sorry, I didn't explain myself correctly. Every RegExp method (or string method which uses regexs) should use `exec` internally. e.g.

```js
class MyRegExp extends RegExp {
  exec() {
    return ["foo", "bar"];
  }
}

"abc".match(new MyRegExp("b")); // ["foo", "bar"]
```

If it doesn't work in `core-js`, I could open a PR

zloirock: `core-js` provides a naive realisation of support `RegExp`-related well-known symbols and use original methods like `String.prototype.match` or `String.prototype.replace`. It's done for reducing slowdown in wrappers. However, now it should not cause too many problems, so feel free to add a PR with adding this logic.

damianobarbati:
Can this be merged @nicolo-ribaudo ?

nicolo-ribaudo:
It depends on https://github.com/zloirock/core-js/pull/411

DmitrySoshnikov:
Is this ready for merge?

packages/babel-helpers/src/helpers.js
```diff
@@ -1717,3 +1717,72 @@ helpers.decorate = helper("7.0.1")`
...
+      return _this;
+    }
+    inherits(BabelRegExp, _RegExp);
+
+    BabelRegExp.prototype.exec = function(str) {
+      var result = _super.exec.call(this, str);
+      if (result) result.groups = buildGroups(result, this);
+      return result;
+    };
+    BabelRegExp.prototype[Symbol.replace] = function(str, substitution) {
```
nicolo-ribaudo: I think that I will move this whole method to core-js (I already moved the first part)

packages/babel-plugin-transform-named-capturing-groups-regex/test/fixtures/runtime/.no-groups.js
```diff
@@ -0,0 +1,5 @@
+var re = /no-groups-\(?<looks-like-a-group>looks\)/;
```
nicolo-ribaudo: This test causes an infinite recursion in `regexp-tree`. I don't think that it should be a blocker, since irl regexps like this one are really rare.

iheanyi:
Any updates on this PR y'all?

packages/babel-plugin-transform-named-capturing-groups-regex/src/index.js
```diff
@@ -0,0 +1,54 @@
+import regexpTree from "regexp-tree";
+
+export default function({ types: t }, options) {
+  const { runtime = true } = options;
+  if (typeof runtime !== "boolean") {
+    throw new Error(".runtime must be boolean");
```
existentialism: Nit: Is the period necessary?

packages/babel-plugin-transform-named-capturing-groups-regex/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "keywords": [
+    "babel-plugin",
+    "regex",
+    "regexp",
+    "regular expressions"
+  ],
+  "repository": "https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-named-capturing-groups-regex",
+  "bugs": "https://github.com/babel/babel/issues",
+  "dependencies": {
+    "regexp-tree": "^0.0.78"
```
existentialism: Should we bump to latest (85)?

existentialism:
This still needs https://github.com/zloirock/core-js/pull/435 to be published, yea?

/cc: @zloirock

nicolo-ribaudo:
Yeah, It hasn't been released yet.

zloirock:
@existentialism we need a backport of https://github.com/zloirock/core-js/pull/453 to `v2` branch. I'll try to test it with `v3` branch and publish a new beta release at this weak.

nicolo-ribaudo:
Oh right, I will backport it.

packages/babel-plugin-transform-named-capturing-groups-regex/package.json
DmitrySoshnikov: Yes, let's use the latest version.

nicolo-ribaudo:
core-js was just updated, I'll update this PR soon.

DmitrySoshnikov:
@nicolo-ribaudo, what's the status on this? You can also update to the latest `v. 0.1.0`.

billyjanitsch:
Wanted to cross-reference https://github.com/DmitrySoshnikov/regexp-tree/issues/170. regexp-tree brings in a bunch of large CLI dependencies including an old version of yargs, which significantly increases the install size of @babel/preset-env (due to #9345). It would be great to have these moved to a separate CLI package, if possible.

cc @DmitrySoshnikov @existentialism

DmitrySoshnikov:
@billyjanitsch, yes, we'll be separating the CLI (and potentially other modules) into its own npm package.

DmitrySoshnikov:
Note: as of `v1.1.3` the CLI is split into its own module, so the size of the `regexp-tree` is reduced significantly.

PR #7106, opened 2017-12-26
Bug 7103 - Change deprecated option in tests

### LINKED ISSUE #7103: Change deprecated option in tests
Seeing this message 3x in our tests

```
The uglify target has been deprecated. Set the top level
option `forceAllTransforms: true` instead.
```

We just have to change 

https://github.com/babel/babel/blob/ed2b88830ece4ff2fb31906669c0360edc4786f7/packages/babel-preset-env/test/index.spec.js#L80-L86

in a few places to `forceAllTransforms` (and the test name) in that file

GustavoRSSilva:
| Fixed Issues?            | Fixes #7103 

This is a Pull request to fix the warning on deprecation of the use uglify in the preset target.
 ex:
`{
  "presets": [
    ["../../../../lib", {
      "targets": {
        "chrome": "55",
        "node": "7.6",
        "uglify": true
      },
      "modules": false,
      "useBuiltIns": "entry",
    }]
  ]
}`

the warning is 

> The uglify target has been deprecated. Set the top level
option `forceAllTransforms: true` instead.

This issue is related to babel-preset-env's and the major change was removing uglify from the preset target and adding forceAllTransforms.

I would like to point out that even though this is only for the tests, the babel-preset-env/src/index points out the use on the target.uglify with a TODO that consist of:

```
// TODO: remove this in next major
  let hasUglifyTarget = false;

  if (optionsTargets && optionsTargets.uglify) {
    hasUglifyTarget = true;
    delete optionsTargets.uglify;

    console.log("");
    console.log("The uglify target has been deprecated. Set the top level");
    console.log("option `forceAllTransforms: true` instead.");
    console.log("");
  }
```

There is no major change in the logs due to the fact that this issue is related to testing, even though there is documentation that talks about the use of uglify in the babel-preset-env, [the documentation points out the fact that uglify is deprecated.](https://github.com/babel/babel/tree/master/packages/babel-preset-env)


existentialism:
Some history here... our plan was to deprecate it in 2.x and remove it completely in 3.x.

2.x became 7.x when we merged it into the monorepo, so we _technically haven't_ released the major (as a non alpha/beta) that contains the deprecation yet.

yavorsky:
> I'm fine with removing the uglify option tests for now to remove the noise from our test runs, but would prefer to keep the option + deprecation warning in the preset for now.

+1

LGTM, but I would return back `debug-fixtures/builtins-uglify` since we want to log deprecation.

GustavoRSSilva:
@existentialism @yavorsky updated the PR.

existentialism:
@GustavoRSSilva thanks for the help, and congrats on your first PR to Babel!

Issue #7114, opened 2017-12-27
Issue with transform-runtime and defualt export using computed property names


sqal:
Choose one: is this a bug report or feature request?

### Input Code

```js
// actions.js
const types = {
  FOO: 'FOO',
  BAR: 'BAR',
}

export default {
  [types.FOO]() {},
  [types.BAR]() {},
}

// index.js
import actions from './actions'
```

### Babel/Babylon Configuration (.babelrc, package.json, cli command)

```json
// package.json
"devDependencies": {
  "@babel/core": "^7.0.0-beta.36",
  "@babel/plugin-transform-runtime": "^7.0.0-beta.36",
  "@babel/preset-env": "^7.0.0-beta.36",
  "@babel/runtime": "^7.0.0-beta.36",
  "babel-loader": "^8.0.0-beta.0",
  "webpack": "^3.10.0"
}

// .babelrc
{
  "presets": [
    ["@babel/env", {
      "modules": false,
      "useBuiltIns": "usage"
    }]
  ],
  "plugins": [
    "@babel/transform-runtime"
  ]
}

```

You can clone the repo with reproduction: https://github.com/sqal/babel-7.0.0-beta.36-bug

### Expected Behavior
I expect default import in index.js (actions) to be present

### Current Behavior
Default import in index.js is not present. I am getting an error instead:

` "export 'default' (imported as 'actions') was not found in './actions'`

This is somehow related to transform-runtime plugin along with using computed property names in default export. When I disable transform-runtime, everything works properly. My current workaround for this:

```js

const actions = { ... }

export default actions;
```

### Your Environment

| software         | version(s)
| ---------------- | -------
| Babel            |  7.0.0-beta.36
| node             | 8.5.0
| npm              | 5.5.1
| Operating System | Windows 10

nicolo-ribaudo:
This issue can also be reproduced using only preset-es2015 end exporting an object with more than one compited property: https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&code_lz=KYDwDg9gTgLgBAE2AMwIYFcA28DeAoOOAbVQF0AuOAIwBoDiBjCxPAXyA&debug=false&circleciRepo=&evaluate=false&lineWrap=false&presets=es2015%2Creact%2Cflow&prettier=false&targets=&version=7.0.0-beta.36

It works with Babel 6.

nicolo-ribaudo:
Fixed in #7116

PR #7116, opened 2017-12-27
Fix a regression introduced in #7040

nicolo-ribaudo:
| Fixed Issues?            | https://github.com/babel/babel/issues/7114
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

This PR fixes a regression introduced in https://github.com/babel/babel/pull/7040. I actually wrote some tests to prevent this, but they where wrong :rofl: 

Example:
```js
// Input
export default x;

// Plugin
pathToThatX.insertAfter(t.identifier("y"));

// Old output (wrong)
export default x;
y;

// Correct output
export default (_tmp = x, y, _tmp);
```

They might look the same, but when using `path.replaceWithMultiple` (which removes the node and then inserts the new nodes after), it generated something like
```js
export default ;
y;
```
And `export default` got removed since it is invalid.

### DIFF
--- a/packages/babel-plugin-transform-computed-properties/test/fixtures/regression/7144/actual.js
+++ b/packages/babel-plugin-transform-computed-properties/test/fixtures/regression/7144/actual.js
@@ -1,0 +1,4 @@
+export default {
+  [a]: b,
+  [c]: d
+};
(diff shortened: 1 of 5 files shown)


Andarist:
```
// Old output (wrong)
export default x;
y;
```
seems actually more intuitive for me, than inserting they `y` kinda **in between** like in 
```js
// Correct output
export default (_tmp = x, y, _tmp);
```

I understand that `path.replaceWithMultiple` and its usage of `path.insertAfter` under the hood caused issues, but maybe the fix should be done in the `replaceWithMultiple` and not in the `insertAfter`/`insertBefore`.

nicolo-ribaudo:
Well, that is how we already do with expressions: https://astexplorer.net/#/gist/226f5986f037c3e32844ef860aa0bbc6/a927881f46355e9d1f1817ad666e65ba34825f51

PR #7117, opened 2017-12-27
Clean babel-prest-env stdout for builtIns: `usage` with imports removing case.

yavorsky:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  | dev: sinon

It's rather followup for the https://github.com/babel/babel/pull/7106
and attempt to get rid of the 
screen shot 2017-12-28 at 00 38 00


packages/babel-preset-env/test/use-built-ins-plugin.js
```diff
@@ -0,0 +1,33 @@
...
+const sinon = require("sinon");
+
+describe("use-built-ins-plugin", () => {
+  describe("usage option", () => {
+    it("should remove all babel-polyfill imports", () => {
+      const codeWithImports = `import "@babel/polyfill";
+import "@babel/polyfill";
+require("@babel/polyfill");
+require("@babel/polyfill");
+`;
```
nicolo-ribaudo: Can we make this look nice, since whitespaces are not important for this test? :slightly_smiling_face: 
```js
      const codeWithImports = `
        import "@babel/polyfill";
        import "@babel/polyfill";
        require("@babel/polyfill");
        require("@babel/polyfill");
      `;

```

yavorsky: Sure, we can just `trim()` it 🙂

hzoo:
Yeah would be nice if we don't need to add more deps

yavorsky:
Agree, I'm going to reuse some stuff from `debug-fixtures.js`.

existentialism:
Instead of needing debug-fixtures and silent-fixtures, can't we just default to silent and pass an option in `options.json` to compare stdout?

hzoo:
Yeah that would be good if it's more common not to need stdout

hzoo:
^ sounds good

yavorsky:
I'll update it today. Sorry, hadn't much time before.

nicolo-ribaudo:
We not have a `validateLogs` test option that captures the stdout/stderr. Thanks anyway for the PR!

PR #7120, opened 2017-12-28
Regenerate fixtures

Kovensky:
This, for the most part, just commits small whitespace changes in fixtures that can happen if you... delete all of them and run `make test`!

This also caught test fixtures that were *not running*. Some of them were put in an incorrect directory layout, some of them were intentionally disabled, but given how the output looks correct, I reenabled them.

The "edgest-case" test is supposed to fail but its fixture was saved in a pass state; so I updated it to correctly match the thrown error.

Only `babel-plugin-transform-react-jsx/test/fixtures/react/.should-properly-handle-comments-adjacent-to-children` is still broken (the output is correct javascript but is missing all the comments), so I kept it disabled.


packages/babel-generator/test/fixtures/flow/declare-exports/expected.js
```diff
@@ -38,4 +38,4 @@ declare export opaque type Bar<T>;
 declare export opaque type Baz: Foo;
 declare export opaque type Foo<T>: Bar<T>;
 declare export opaque type Foo<T>: Bar;
-declare export opaque type Foo: Bar<T>;
+declare export opaque type Foo: Bar<T>;
```
Kovensky: I hate that this is adding files with no line terminators, but it seems to be a problem with how `babel-generator` is writing flow nodes.

packages/babel-plugin-transform-react-jsx/test/fixtures/react/.optimisation.react.constant-elements/expected.js
```diff
@@ -1,40 +0,0 @@
-var _ref = React.createElement(
-  "div",
```
Kovensky: The output in the reenabled test is identical, only it's still in JSX instead of transformed to React.createElement calls.

packages/babel-plugin-transform-react-jsx/test/fixtures/react/should-properly-handle-comments-between-props/expected.js
```diff
@@ -0,0 +1,8 @@
+var x = React.createElement("div", {
+  /* a multi-line
+     comment */
+  attr1: "foo"
+}, React.createElement("span", {
+  // a double-slash comment
```
Kovensky: This comment changed lines (it used to be printed right after the `{`), but I don't think this breaks the test unless this is supposed to preserve its position.

packages/babel-plugin-transform-react-jsx/test/fixtures/react/throw-if-pragma-set-but-not-pragmafrag-and-frag-used/expected.js
```diff
@@ -1,2 +0,0 @@
-/** @jsx dom */
-dom(React.Fragment, null);
```
Kovensky: This test is supposed to throw so it's not supposed to have any expected output

packages/babel-plugin-proposal-class-properties/test/fixtures/static-property-tdz/edgest-case/expected.js
```diff
@@ -1,17 +0,0 @@
...
-
-var _x$x = {
-  x: (_classNameTDZError("A"), A) || 0
-}.x;
-Object.defineProperty(A, _x$x, {
-  configurable: true,
-  enumerable: true,
-  writable: true,
-  value: void 0
-});
```
Kovensky: All these `expected.js` in `static-property-tdz` shouldn't exist because the transform is supposed to throw.

nicolo-ribaudo: I accidentally committed these two files almost three times lol

Kovensky: Looking at the spec, it seems that the error is supposed to happen at runtime, since it's not specified as an early error anywhere. But that's what the plugin does anyway, so probably better to file that as a separate bug 🤷‍♀️

PR #7121, opened 2017-12-28
Regex parsing issue fix  after function declaration.

### LINKED ISSUE #7100: Lexer cannot lex 'x= function(){}/42/i' correctly.
bug report

### Input Code

```js

const i = 4;

let x = {
  a: function () {}/42/i,
};

let y = function() {}/42/i;
```

Repl: https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=FAYw9gdgzgLgBASzgXjgFgNzGAGwKbwAeKcA3sHHAIYBccAZgK4QgwKRwBGAFAJRkBfAPRoATEIQAaYAKy4CcAJ4kmLNpD6CR4hBiA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=6.26.0

### Expected Behavior
`x.a == y`

### Current Behavior
Lexical error for y.

### Possible Solution

Currently, `updateContext` of `_function` [only checks if current context is a brace statement](https://github.com/babel/babel/blob/2d378d076eb0c5fe63234a8b509886005c01d7ee/packages/babylon/src/tokenizer/context.js#L106-L112).

It should check if `function` keyword can be part of a statement.
If token before `function` is assign op or binary op, it's always expression and `context.push(ct.functionExpression)` should be called.

For reference, this rust code fixed the bug.
```rust
let is_always_expr = match prev {
    Some(AssignOp(..)) | Some(BinOp(..)) => true,
    _ => false,
};
if context.current() != Some(Type::BraceStmt) || is_always_expr {
    context.push(Type::FnExpr);
}
```

### Context
I was writing javascript parser in rust, and I didn't want to allow lexer to access parser's state because obviously rust compiler would hate it. While searching for solution, I found that babel separates lexer from parser and I decided to port it.

While porting it, I found a link to [sweet.js design](https://github.com/sweet-js/sweet-core/wiki/design) and there was some examples.

Example form sweet.js:
```js
// a declaration so / is regex
f(); function foo() {} /42/i
vs
// an expression so / is divide
x = function foo() {} /42/i
```

 I tested these after porting babel's lexer. But it fails and it seems like a bug of babel lexer. So I'm reporting this to babel.

 

### Your Environment

| ---------------- | -------
| Babel            |  master
| Babylon          | master

rajzshkr:
| Fixed Issues?            | Fixes #7100 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |


packages/babylon/test/fixtures/core/categorized/regex-after-function-expression/actual.js
```diff
@@ -0,0 +1 @@
+const x = function foo() {} /42/i
```
Kovensky: Probably should also test `(function foo() {} /42/i)` to be sure

packages/babylon/src/tokenizer/context.js
```diff
@@ -103,8 +103,8 @@ tt.incDec.updateContext = function() {
   // tokExprAllowed stays unchanged
 };
 
-tt._function.updateContext = function() {
-  if (this.curContext() !== types.braceStatement) {
+tt._function.updateContext = function(prevType) {
+  if (this.curContext() !== types.braceStatement || prevType.isAssign) {
```
nicolo-ribaudo: I think you should check either if `exprAllowed` or if `prevType.beforeExpr`

packages/babylon/test/fixtures/core/categorized/regex-after-function-expression/actual.js
nicolo-ribaudo: And also this, since the significant line break must be considered:
```
return
function fn() {}
/42/i
```

nicolo-ribaudo:
```js
function fn() {
  return
  function foo() {}
  /42/i
}
```

Stil doesn't work.

packages/babylon/src/tokenizer/context.js
```diff
@@ -103,8 +103,12 @@ tt.incDec.updateContext = function() {
   // tokExprAllowed stays unchanged
 };
 
-tt._function.updateContext = function() {
-  if (this.curContext() !== types.braceStatement) {
+tt._function.updateContext = function(prevType) {
+  if (
+    this.curContext() !== types.braceStatement ||
+    prevType.isAssign ||
```
nicolo-ribaudo: You need to use `this.state.exprAllowed`, because function expression could be introduced by something different from `=`.
e.g.
```js
{
  !function fn() {} /42/i;
}
```

Since if `prevType === tt._return` then `exprAllowed` is true, you could then collapse the two conditions into something like `(this.state.exprAllowed && !this.braceIsBlock)`

packages/babylon/src/tokenizer/context.js
```diff
@@ -103,8 +103,12 @@ tt.incDec.updateContext = function() {
   // tokExprAllowed stays unchanged
 };
 
-tt._function.updateContext = function() {
-  if (this.curContext() !== types.braceStatement) {
+tt._function.updateContext = function(prevType) {
+  if (
+    this.curContext() !== types.braceStatement ||
```
nicolo-ribaudo: I really don't understand what this is for, I think it was a rough approssimation of the check that you are adjusting in this PR. If it reveals to be unnecessary, can you remove it?

PR #7124, opened 2017-12-28
[decorators] Only transform declarations to expressions when needed

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

### Input
```js
class Foo {}
```

### Old output
```js
let Foo = class Foo {};
```

### New output
```js
class Foo {}
```

---

This also fixed an issue I had with rollup-plugin-babel: since class declarations where transformed, the logic to check how helpers are used  (https://github.com/rollup/rollup-plugin-babel/blob/e74407ee15a9093c2d6e3e5a3ef7691bdc57b3d8/src/preflightCheck.js#L8-L11) didn't run.

### DIFF
--- a/packages/babel-plugin-proposal-decorators/src/index.js
+++ b/packages/babel-plugin-proposal-decorators/src/index.js
@@ -63,10 +63,10 @@ export default function() {
    * with the proper decorated behavior.
    */
   function applyClassDecorators(classPath) {
+    if (!hasClassDecorators(classPath.node)) return;
+
     const decorators = classPath.node.decorators || [];
     classPath.node.decorators = null;
-
-    if (decorators.length === 0) return;
 
     const name = classPath.scope.generateDeclaredUidIdentifier("class");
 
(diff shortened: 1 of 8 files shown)


packages/babel-plugin-proposal-decorators/src/index.js
```diff
@@ -215,32 +215,29 @@ export default function() {
     inherits: syntaxDecorators,
 
     visitor: {
-      ExportDefaultDeclaration(path) {
```
hzoo: wonder if this will cause other plugin ordering issues?

We really should just move the class properties/decorators transforms into the same "plugin" and just turn them on in the individual pkgs

nicolo-ribaudo: It won't causes other problems that we don't already have (`Method has decorators, put the decorator plugin before the classes one.`)

packages/babel-plugin-proposal-decorators/src/index.js
```diff
@@ -215,32 +215,29 @@ export default function() {
...
-          ]),
-        );
+        if (path.parentPath.isExportDefaultDeclaration()) {
+          // Split the class declaration and the export into two separate statements.
+          path.parentPath.replaceWithMultiple([
+            letDeclaration,
+            t.exportNamedDeclaration(null, [
+              t.exportSpecifier(ref, t.identifier("default")),
+            ]),
+          ]);
```
nicolo-ribaudo: TODO: Use `babel-helper-split-export-declaration` when https://github.com/babel/babel/pull/6963 is merged

packages/babel-plugin-proposal-decorators/src/index.js
hzoo: Really? If they are implemented in the same plugin wouldn't the issue be gone?

PR #7130, opened 2017-12-29
Remove hasBeenLogged flag from preset-env

### LINKED ISSUE #6634: Debug output omitted with 2 .babelrc files
> Issue originally reported by @noinkling in babel/babel-preset-env#147

I have a Node web app where both the server and frontend use Babel. The frontend is served using `webpack-dev-middleware` in development, so I can just launch the app in one command with `babel-node`.

Root level `.babelrc` for server:
```
{
  "presets": [
    ["env", {
      "debug": true,
      "targets": {
        "node": "current"
      }
    }]
  ]
}
```

Subdirectory `.babelrc` for frontend:
```
{
  "presets": [
    ["env", {
      "debug": true,
      "modules": false
      // no "targets" for now, just defaulting to "latest" preset
    }],
    "react"
  ],
  "env": {
    "development": {
      "plugins": ["react-hot-loader/babel"]
    }
  }
}
```

When the `debug` option is set to `true` in both files, only the output for the server (Node) one is shown - I would expect both to be shown. If I remove that option from the Node `.babelrc`, the frontend debug output then shows up fine with all the stuff from `babel-preset-latest` as I expect.

It's important to note that the bug doesn't affect actual functionality and the setup works great otherwise.

Edit: I expect it has something to do with the [`hasBeenLogged` flag variable](https://github.com/babel/babel-preset-env/blob/55aa7ff3531ec94eed780dd90072543af3614567/src/index.js#L180).

h1b9b:
| Fixed Issues?            | Fixes #6634

Remove hasBeenLogged flag from babel-preset-env

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,4 +40,3 @@ packages/babel-preset-env-standalone/babel-preset-env.min.js
 /codemods/*/node_modules
 /packages/babylon/build
 .idea/
-.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ packages/babel-preset-env-standalone/babel-preset-env.min.js
 /codemods/*/node_modules
 /packages/babylon/build
 .idea/
+.vscode/
--- a/packages/babel-preset-env/src/index.js
+++ b/packages/babel-preset-env/src/index.js
@@ -93,8 +93,6 @@ export const isPluginRequired = (
   return isRequiredForEnvironments.length > 0;
 };
 
-let hasBeenLogged = false;
-
 const getBuiltInTargets = targets => {
   const builtInTargets = Object.assign({}, targets);
   if (builtInTargets.uglify != null) {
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -40,3 +40,4 @@ packages/babel-preset-env-standalone/babel-preset-env.min.js
 /codemods/*/node_modules
 /packages/babylon/build
 .idea/
+.vscode/
```
nicolo-ribaudo: Can you remove this?

xtuc: We actually have sublime-text and intellij, I don't mind adding vscode as well.

existentialism: I know it's somewhat subjective, but I personally prefer not having any editor/OS specific stuff in the gitignore and add instructions for people to ignore them on their local machine, globally. Not really a big issue worth bikeshedding tho :)

xtuc: I agree @existentialism, I would say either we accept this one or remove all of them (in another PR), what do you think?

existentialism:
@h1b9b thanks, and congrats on your first PR to Babel!

Issue #7029, opened 2017-12-15
Methods should not have "prototype" property


TimothyGu:
bug report

### Input Code

```js
console.log(typeof ({ a() {} }).a.prototype);

console.log(typeof (new class { a() {} }).a.prototype);
```

[REPL](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYewdgziA2CmB00QHMAUAXAngB1iAZgASoDehAhqgJSEkC-hdV8582ATiOlzrFQNwAoQaEgwESNFlwFiYWAHdCwaOQgRaFarQZMWbTt2l9-QA&debug=false&circleciRepo=&evaluate=true&lineWrap=false&presets=es2015&prettier=false&targets=&version=7.0.0-beta.0)

### Babel/Babylon Configuration (.babelrc, package.json, cli command)

es2015 preset enabled in the REPL

### Expected Behavior

The code should print "undefined" and "undefined", as it does in all modern JavaScript engines.

### Current Behavior

The code prints "object" and "object".

### Possible Solution

The only way I can think of is `delete` the prototype property after function creation, which can pretty awkward in object literals. Maybe a runtime function could be added that basically does `(func) => { delete func.prototype; return func; }`.

### Context

N/A

### Your Environment

| software         | version(s)
| ---------------- | -------
| Babel            |  v7.0.0-beta.0 (REPL)
| Babylon          | 
| <del>node</del> Browser | Google Chrome 63.0.3239.84 (Official Build) beta (64-bit)

zloirock:
It's impossible since we can't delete prototype of usual function:
![image](https://user-images.githubusercontent.com/2213682/55122003-b4ca9980-512f-11e9-855b-354cf9dda7c1.png)
Kinds of functions without prototypes are unavailable in old ES versions and can't be applicable in this case.

PR #7133, opened 2017-12-29
Ensure the methods are not instantiated

### LINKED ISSUE #7115: `spec` option for transform-classes/shorthand-properties ?
Currently, I'm working on the PR that add an additional check for object/class methods to throw an error on attempts to use it as a constructor. It will extend current methods, which can affect the output size.
For example:
```js
a = {
  b() {} // transforms to `b: function() {}` which is a valid consturctor.
}
new a.b(); // Should throw `TypeError: Cannot instantiate an object method`
```

We already have `newArrowCheck` helper, which is added with the `spec` option. But we don't have `spec` options for *transform-classes* and *transform-shorthand-property plugins*.

I want to see some opinions about adding `spec` options for these 2 plugins to make `newClassMethodCheck`/`newObjectMethodCheck` consistent with `newArrowCheck`, which is added with spec option for *transform-arrow-functions*.

```js
function () {
  function Test() {
    babelHelpers.classCallCheck(this, Test);
  }

  babelHelpers.createClass(Test, [{
    key: "test",
    value: function test() {
        _newClassMethodCheck(this, Test.prototype.test); // throw if `this instanceof Test.prototype.test`
    }
  }]);
  return Test;
}();
```

Also, with we can fix https://github.com/babel/babel/issues/7029 , because changes needed to remove methods` prototypes presume more transformations, which could be superfluous for the everyday use and with `spec` (which is not default), we can add additional helpers more freely, not thinking about output size, but about spec compliance.

cc @xtuc

yavorsky:
| Fixed Issues?            | Fixes #7115
| Major: Breaking Change?  |
| Any Dependency Changes?  |

This PR adds `spec` option for *classes* and *shorthand-property* pluigns.
To be consistent with arrow-functions, spec mode adds runtime check to the functions created from methods to ensure the function is not instantiated.
For example: 
```
class A {
  b() {}
}

new (new A()).b // should throw an error
```
So, for now, it just adds a `newMethodCheck` helper, which throws an exception if `this instanceof method`.

My second attempt was to fix https://github.com/babel/babel/issues/7029 by removing prototype, but there are not so simple (and I dunno if possible at all), because after we removed prototype from the function, it throws on any `instanceof` check inside the function `Function has non-object prototype 'undefined' in `instanceof` check`.

Also, `this.constructor` became `Object` after prototype was removed, so it's also won't work with *transform-new-target* approach.


packages/babel-plugin-transform-classes/src/spec.js
```diff
@@ -0,0 +1,38 @@
+import VanillaTransformer from "./vanilla";
+import { types as t } from "@babel/core";
+
+export default class SpecClassTransformer extends VanillaTransformer {
```
xtuc: Rather than inheritance I would prefer composition, it's not currently an issue but could cause unexpected results in the future (from maintenance/interations).

packages/babel-plugin-transform-classes/test/fixtures/spec/method-check-new-extends/expected.js
```diff
@@ -0,0 +1,23 @@
...
+    babelHelpers.classCallCheck(this, Foo);
+    _this = babelHelpers.possibleConstructorReturn(this, (Foo.__proto__ || Object.getPrototypeOf(Foo)).call(this));
+    _this.state = "test";
+    return _this;
+  }
+
+  babelHelpers.createClass(Foo, [{
+    key: "bar",
+    value: function bar() {
+      babelHelpers.newMethodCheck(this, Foo.prototype.bar);
```
loganfsmyth: It seems like it'd be possible for this to reference `bar` directly. Is there any reason to reference it from the prototype like this?

packages/babel-plugin-transform-classes/src/spec.js
```diff
@@ -0,0 +1,38 @@
...
+    super(...arguments);
+    this.isSpec = true;
+  }
+
+  _getMethodRef(node, isStatic) {
+    return t.memberExpression(
+      isStatic
+        ? this.classRef
+        : t.memberExpression(this.classRef, t.identifier("prototype")),
+      t.identifier(node.key.name),
```
loganfsmyth: This is going to have issues if it's a computed method.

Andarist: also probably `this.classRef` should be cloned

packages/babel-plugin-transform-classes/test/fixtures/spec/method-check-new-super/expected.js
```diff
@@ -0,0 +1,19 @@
...
+  babelHelpers.inherits(Test, _Foo);
+
+  function Test() {
+    babelHelpers.classCallCheck(this, Test);
+    return babelHelpers.possibleConstructorReturn(this, (Test.__proto__ || Object.getPrototypeOf(Test)).call(this));
+  }
+
+  babelHelpers.createClass(Test, [{
+    key: "test",
+    value: function _test() {
```
loganfsmyth: This is introducing a regression here. Currently the generated code is

```
  _createClass(Test, [{
    key: "test",
    value: function test() {
```
and the name is fully preserved, whereas your snippet will name the function `_test`.

packages/babel-plugin-transform-classes/src/vanilla.js
```diff
@@ -612,4 +622,20 @@ export default class ClassTransformer {
       ),
     );
   }
+
+  /**
+   * For spec mode, add instanceof check for the function block.
+   */
+
+  addNewMethodChecks(map, scope) {
+    const methodRef = scope.generateUidIdentifierBasedOnNode(map._key);
```
Kovensky: Don't need to generate one if the method already has an `id`, no?

You do need to check if the name is shadowed inside the method, though, and use scope to rename it.

loganfsmyth: I think there's already logic to generate this name in the non-computed case, but this logic may be running too early for that to have taken affect. We will need logic like this for the computed keys though probably.

Kovensky: In that case, I suppose it's best to call the `helper-function-name` first; but I'm not sure if this code (or its caller) has access to the `path`...

PR #7140, opened 2018-01-01
Update README.md

eladchen:
The previous version of the api allowed passing ignore options as a function, which now can only be an array of regex / functions.

The current docs are somewhat ambiguous with regards to this breaking change.
Anyway I did some editing, let me know if this is ok with you guys.

[skip ci]

### DIFF
--- a/packages/babel-register/README.md
+++ b/packages/babel-register/README.md
@@ -52,11 +52,8 @@ require("@babel/register")({
   // When a file path **matchs** this regex then then it is **not** compiled.
   // When a function is invoked, it is called with the file path, and is **not** compiled if it returns true.
   ignore: [
-    // if any filenames **do** match this regex then they aren't compiled.
     /regex/,
-    
-    // Ignore entries can also be functions.
-    // A function receives the filename, and should return a boolean. 
+
     function( filepath ) {
       if (filepath === "/path/to/es6-file.js") {
         return false;
--- a/packages/babel-register/README.md
+++ b/packages/babel-register/README.md
@@ -47,15 +47,18 @@ require("@babel/register")({
 
 ```javascript
 require("@babel/register")({
-  // Optional ignore array
+  // Array of ignore conditions ( Optional )
+  // A condition can be either a regex, or a function.
+  // When a file path **matchs** this regex then then it is **not** compiled.
+  // When a function is invoked, it is called with the file path, and is **not** compiled if it returns true.
   ignore: [
     // if any filenames **do** match this regex then they aren't compiled.
     /regex/,
     
     // Ignore entries can also be functions.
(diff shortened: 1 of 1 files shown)


packages/babel-register/README.md
```diff
@@ -47,18 +47,22 @@ require("@babel/register")({
 
 ```javascript
 require("@babel/register")({
-  // Optional ignore regex - if any filenames **do** match this regex then they
-  // aren't compiled.
-  ignore: [/regex/],
-
-  // Ignore can also be specified as a function.
-  ignore: [function(filename) {
```
Qantas94Heavy: It seems the function is already wrapped in an array.

packages/babel-register/README.md
```diff
@@ -47,18 +47,22 @@ require("@babel/register")({
...
-    if (filename === "/path/to/es6-file.js") {
-      return false;
-    } else {
+  // Optional ignore array
```
Qantas94Heavy: Maybe you could add something to make it clear that the file will be ignored if it matches any of the `ignore` entries?

packages/babel-register/README.md
```diff
@@ -47,18 +47,22 @@ require("@babel/register")({
...
-  ignore: [/regex/],
-
-  // Ignore can also be specified as a function.
-  ignore: [function(filename) {
-    if (filename === "/path/to/es6-file.js") {
-      return false;
-    } else {
+  // Array of ignore conditions ( Optional )
+  // A condition can be either a regex, or a function.
+  // When a file path **matchs** this regex then then it is **not** compiled.
```
existentialism: > // When a file path **matches** this regex then it is **not** compiled.

packages/babel-register/README.md
```diff
@@ -47,18 +47,22 @@ require("@babel/register")({
...
-  // Optional ignore regex - if any filenames **do** match this regex then they
-  // aren't compiled.
-  ignore: [/regex/],
-
-  // Ignore can also be specified as a function.
-  ignore: [function(filename) {
-    if (filename === "/path/to/es6-file.js") {
-      return false;
-    } else {
+  // Array of ignore conditions ( Optional )
```
existentialism: Nit remove spaces surrounding `Optional`

packages/babel-register/README.md
```diff
@@ -47,18 +47,22 @@ require("@babel/register")({
...
+
+    function( filepath ) {
+      if (filepath === "/path/to/es6-file.js") {
+        return false;
+      } else {
+
       return true;
     }
-  }],
+  }
```
existentialism: Let's fix this indention

existentialism:
@eladchen thanks for the PR!

eladchen:
@existentialism My pleasure.

PR #7143, opened 2018-01-02
Use loc.start inside File#buildCodeFrameError

thejameskyle:
It was previously returning `{ line: undefined, column: NaN }` because it was looking for `loc.{line,column}` when it's a full range `loc.{start,end}.{line,column}` coming from a node.


thejameskyle:
Other places don't have this problem because they draw from `err.loc` which is a location instead of a range.

Andarist:
Would be good to add a test for this.

thejameskyle:
Closing in favor of #7146

PR #7146, opened 2018-01-03
fix buildCodeFrameError empty code frames

### LINKED ISSUE #7145: buildCodeFrameError calls codeFrameColumns with broken loc
### Input Code

```js
const { transform } = require('@babel/core');
transform('function f() {}', {
  plugins: [
    () => ({
      visitor: {
        FunctionDeclaration: path => {
          throw path.buildCodeFrameError('asdf');
        },
      },
    }),
  ],
});
```

```js
{
  "dependencies": {
    "@babel/core": "7.0.0-beta.36"
  }
}
```

### Expected Behavior

The error thrown includes code frame information:
```
SyntaxError: undefined: asdf
> 1 | function f() {}
    | ^
    at File.buildCodeFrameError (/home/[...]
```

### Current Behavior

```
SyntaxError: undefined: asdf

    at File.buildCodeFrameError (/home/[...]
```

### Possible Solution

`buildCodeFrameError (packages/babel-core/src/transformation/file/file.js)` calls `codeFrameColumns (packages/babel-code-frame/src/index.js)` with `loc` argument
```js
        start: {
          line: loc.line,
          column: loc.column + 1
        }
```
However, in `buildCodeFrameError loc` is
image
with `line` and `column` nested in `start`.

Consequentially, `codeFrameColumns`' `loc` argument is
image
and it returns `""`

It looks like the call should rather be
```js
        start: {
          line: loc.start.line,
          column: loc.start.column + 1
        }
```
If I change that in the dev tools, I get the expected output

### Your Environment

| ---------------- | -------
| Babylon          | 
| node             | 8.9.3
| npm              | 
| Operating System | Linux arch-home 4.14.10-1-ARCH

### Notes
* In case I'm doing something wrong please do tell me, I'm a bit confused because that `codeFrameColumns` call seems to have been there for months and nobody noticed
* I'm working on a PR rn
* This is a regression (at least I'm pretty sure that's how I used `buildCodeFrameError` in v6 & I couldn't find anything in the Changelog)

jeysal:
Fixes #7145 
Duplicate of #7143 (didn't see there was already a fix for the issue i was creating), but this has a regression test.

### DIFF
--- a/packages/babel-core/src/transformation/file/file.js
+++ b/packages/babel-core/src/transformation/file/file.js
@@ -179,8 +179,8 @@ export default class File {
 
   buildCodeFrameError(
     node: ?{
-      loc?: { line: number, column: number },
-      _loc?: { line: number, column: number },
+      loc?: { start: { line: number, column: number } },
+      _loc?: { start: { line: number, column: number } },
     },
     msg: string,
     Error: typeof Error = SyntaxError,
@@ -212,8 +212,8 @@ export default class File {
           this.code,
           {
             start: {
-              line: loc.line,
-              column: loc.column + 1,
+              line: loc.start.line,
+              column: loc.start.column + 1,
             },
           },
           { highlightCode },
--- a/packages/babel-core/src/transformation/plugin-pass.js
+++ b/packages/babel-core/src/transformation/plugin-pass.js
@@ -39,8 +39,8 @@ export default class PluginPass {
 
   buildCodeFrameError(
     node: ?{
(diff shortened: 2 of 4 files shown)


jeysal:
I've verified that the test breaks when running with the `packages/babel-core/src` from `babel:master`, so it should catch any regression.

thejameskyle:
Awesome, merging in this one instead of mine. Assuming same approvals on other PR apply here as well

PR #7147, opened 2018-01-03
Fix computed properties being inserted after the class, thus making t…

### LINKED ISSUE #7125: `@babel/plugin-proposal-class-properties` code generation incorrect order (beta 36)
Bug

### Input Code

```js
import setSomethingPrivate from './util/set'; // export default Symbol('set something private');

const somethingPrivate = Symbol('something private');

class MyClass {
  [somethingPrivate] = null;

  [setSomethingPrivate](thePrivate) {
    this[somethingPrivate] = thePrivate;

    return this;
  }

  getSomethingPrivate() {
    return this[somethingPrivate];
  }
}
```

```js
module.exports = {
  presets: [
    ['@babel/preset-env', {
      useBuiltIns: 'entry',
      loose: true
    }],
    '@babel/stage-0'
  ]
};
```

### Expected Behavior

Identifiers which are used as names of computed properties are assigned to before they are used as computed property names.

### Current Behavior

Identifiers which are used as names of computed properties are assigned to AFTER they are used as computed property names.

(Module processed with babel via `babel-loader@8.0.0-beta.0` and `webpack@3.10.0` (I can't provide REPL link because it says the property transform is missing: http://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=JYWwDg9gTgLgBAZwKYwMoRCgFsAdgcwAUpgA3AQxiTgDMoM4ByAOgHoBXGYAG1eRgBy-ALIQAJkm4BhGFG6MA3ACglAYwi4E8BBmx4iJClTgBeOKgCeIAEYRuACkY7MMHAThhDlJIwCUytW5yBAQ4YQspIJC4AG8lODgAbWc9AmIybwBdUzhcdm5uAITklHQXNwMMqkz7VyR0oyRfWPiEuFdgBGTdDrSvapy6hu8itqgUdihcdpwEZQSAXxUE_FKeiuGqe2a4trhxmEnpjq6U3srGzPm4JaWlJAAPSFg4CRpyfPhwyOC5pSA&debug=false&circleciRepo=&evaluate=false&lineWrap=false&presets=stage-0%2Ces2015-loose&prettier=false&targets=&version=7.0.0-beta.35)

```js
(function(module, exports, __webpack_require__) {

"use strict";


var _interopRequireDefault = __webpack_require__(0);

exports.__esModule = true;
exports.default = void 0;

var _set = _interopRequireDefault(__webpack_require__(69));

var _class, _temp, _somethingPrivate, _setSomethingPrivate;

var somethingPrivate = Symbol('something private');
/**
 *
 */

var MyClass = (_temp = _class =
/*#__PURE__*/
function () {
  function MyClass() {
    Object.defineProperty(this, _somethingPrivate, {
      configurable: true,
      enumerable: true,
      writable: true,
      value: null
    });
  }

  var _proto = MyClass.prototype;

  _proto[_setSomethingPrivate] = function (thePrivate) {
    this[somethingPrivate] = thePrivate;
    return this;
  };

  _proto.getSomethingPrivate = function getSomethingPrivate() {
    return this[somethingPrivate];
  };

  return MyClass;
}(), _somethingPrivate = somethingPrivate, _setSomethingPrivate = _set.default, _temp);

var _default = MyClass;
exports.default = _default;

/***/ })
```

The problem I see is that when prototype methods are created (`_proto[_setSomethingPrivate] = function (thePrivate) {`) the `_setSomethingPrivate` has no value yet: assignment `_setSomethingPrivate = _set.default` is evaluated AFTER `_proto[_setSomethingPrivate]` is assigned to. That results in assignment to `_proto[undefined]` and effectively the setter being missing from the prototype.

`_somethingPrivate` is only used when class is instantiated, which is late enough for the assignment (`_somethingPrivate = somethingPrivate`) has already been evaluated.

### Possible Solution

At least those assignments, which are needed by the time methods with computed names are created, should be generated before the `_temp = _class = function (){ /* class body*/ }()` assignment

### Context

I'm having a class that implements so called "friend method" pattern by exposing a method named with a symbol which is known only to the class and modules that are supposed to use the method.
The reason for such pattern is to keep the setter private while still accessible to some modules.

### Your Environment

| ---------------- | -------
| Babylon          | 
| node             | 8.9.1
| npm              | 5.5.1
| Operating System | macOS 10.13.2

Andarist:
…hem unavailable for instance methods & properties

| Fixed Issues?            | Fixes #7125 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

cc @jrencz

### DIFF
--- a/packages/babel-plugin-proposal-class-properties/src/index.js
+++ b/packages/babel-plugin-proposal-class-properties/src/index.js
@@ -145,7 +145,7 @@ export default function(api, options) {
           }
         }
 
-        const nodes = computedNodes.concat(staticNodes);
+        const afterNodes = [...staticNodes];
 
         if (instanceBody.length) {
           if (!constructor) {
@@ -183,7 +183,7 @@ export default function(api, options) {
               "initialiseProps",
             );
 
-            nodes.push(
+            afterNodes.push(
               t.variableDeclaration("var", [
                 t.variableDeclarator(
                   initialisePropsRef,
@@ -223,7 +223,7 @@ export default function(api, options) {
           prop.remove();
         }
 
-        if (!nodes.length) return;
+        if (computedNodes.length === 0 && afterNodes.length === 0) return;
 
         if (path.isClassExpression()) {
           path.scope.push({ id: ref });
@@ -233,7 +233,8 @@ export default function(api, options) {
           path.node.id = ref;
         }
 
-        path.insertAfter(nodes);
+        path.insertBefore(computedNodes);
+        path.insertAfter(afterNodes);
       },
     },
   };
--- a/packages/babel-plugin-proposal-class-properties/test/fixtures/general/computed-initialization-order/exec.js
+++ b/packages/babel-plugin-proposal-class-properties/test/fixtures/general/computed-initialization-order/exec.js
@@ -1,18 +1,38 @@
-const actual = [];
+const actualOrder = [];
 
-const track = i => actual.push(i);
+const track = i => {
+  actualOrder.push(i);
+  return i;
+};
 
 class MyClass {
   static [track(1)] = track(10);
-  [track(2)] = track(0);
-  get [track(3)]() {}
-  set [track(4)](value) {}
-  [track(5)] = track(0);
+  [track(2)] = track(13);
+  get [track(3)]() {
+    return "foo";
+  }
+  set [track(4)](value) {
+    this.bar = value;
+  }
+  [track(5)] = track(14);
   static [track(6)] = track(11);
   static [track(7)] = track(12);
   [track(8)]() {}
-  [track(9)] = track(0);
+  [track(9)] = track(15);
 }
 
-const expected = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12];
-assert.deepEqual(actual, expected);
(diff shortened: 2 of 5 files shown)


packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -223,7 +223,7 @@ export default function(api, options) {
           prop.remove();
         }
 
-        if (!nodes.length) return;
+        if (!computedNodes.length && !afterNodes.length) return;
```
xtuc: Just a small nit, I would prefer to explicity check for: `computedNodes.length === 0 && afterNodes.length === 0`. It's more human-readable to me and also the engine don't need to coerce it to a boolean.

Andarist: sure, gonna change it in a moment and merge in

nicolo-ribaudo:
@Andarist When ci passes this PR can be merged.

packages/babel-plugin-proposal-class-properties/src/index.js
xtuc: Cool thanks

PR #7149, opened 2018-01-03
Disallow duplicated nodes (only in tests output)

### LINKED ISSUE #6375: Issue with reusing AST nodes
Just want to write this down since we're seeing the same kind of fix for issues and we should look into solving this generally.

Immutable asts would be a larger change, so maybe https://github.com/babel/babel/pull/5971 or infra in our unit tests to check uniqueness would help

https://github.com/babel/babel/pull/6046
https://github.com/babel/babel/pull/6054
https://github.com/babel/babel/pull/6374

nicolo-ribaudo:
Unfortunately this PR won't be easy to review :slightly_frowning_face: 

Anyway, currently there are 604 failing tests :stuck_out_tongue_closed_eyes: 

I had to add some logic do skip validating nodes generated by regenerator-transform, since it isn't in this repository and it duplicates *a lot* of nodes.

cc @Andarist @existentialism 
  
Closes #6375

### DIFF
--- a/packages/babel-helper-transform-fixture-test-runner/src/index.js
+++ b/packages/babel-helper-transform-fixture-test-runner/src/index.js
@@ -130,6 +130,195 @@ function wrapPackagesArray(type, names, optionsDir) {
   });
 }
 
+function checkDuplicatedNodes(ast) {
+  // TODO Remove all these function when regenerator doesn't
+  // insert duplicated nodes
+
+  const nodes = new WeakSet();
+  const parents = new WeakMap();
+
+  const setParent = (child, parent) => {
+    if (typeof child === "object" && child !== null) {
+      let p = parents.get(child);
+      if (!p) {
+        p = [];
+        parents.set(child, p);
+      }
+      p.unshift(parent);
+    }
+  };
+  const registerChildren = node => {
+    for (const key in node) {
+      if (Array.isArray(node[key])) {
+        node[key].forEach(child => setParent(child, node));
+      } else {
+        setParent(node[key], node);
+      }
+    }
+  };
+
+  const parentIs = (node, test) => {
+    return (parents.get(node) || []).some(parent => test(parent));
+  };
+  const isByRegenerator = node => {
+    if (!node) {
+      return false;
+    } else if (node.type === "Identifier") {
+      if (/^_(?:context|value|callee|marked)\d*$/.test(node.name)) {
+        return true;
+      } else if (
+        /^t\d+$/.test(node.name) &&
+        parentIs(
+          node,
+          parent =>
+            parent.type === "MemberExpression" &&
+            isByRegenerator(parent.object),
+        )
+      ) {
+        // _context.t* // <-- t*
+        return true;
+      } else if (
+        parentIs(
+          node,
+          parent =>
+            parent.type === "VariableDeclarator" &&
+            parentIs(
+              parent,
+              parent =>
+                parent.type === "VariableDeclaration" &&
+                parentIs(
+                  parent,
+                  parent =>
+                    parent.type === "BlockStatement" &&
+                    parentIs(
+                      parent,
+                      parent =>
+                        parent.type === "FunctionExpression" &&
+                        isByRegenerator(parent.id),
+                    ),
+                ),
+            ),
+        )
+      ) {
+        // regeneratorRuntime.mark(function _callee3() {
+        //   var bar, _bar2; // <-- Those identifiers
+        return true;
+      } else if (
+        parentIs(
+          node,
+          parent =>
+            parent.type === "VariableDeclarator" &&
+            parentIs(
+              parent,
+              parent =>
+                parent.type === "VariableDeclaration" &&
+                parentIs(
+                  parent,
+                  parent =>
+                    parent.type === "BlockStatement" &&
+                    parent.body.length === 2 &&
+                    parent.body[1].type === "ReturnStatement" &&
+                    parent.body[1].argument.type === "CallExpression" &&
+                    parent.body[1].argument.callee.type ===
+                      "MemberExpression" &&
+                    parent.body[1].argument.callee.property.type ===
+                      "Identifier" &&
+                    parent.body[1].argument.callee.property.name === "wrap",
+                ),
+            ),
+        )
+      ) {
+        // function foo() {
+        //   var _len, // <-- Those identifiers
+        //     items,
+        //     _key,
+        //     _args = arguments;
+        //   return regeneratorRuntime.wrap(function foo$(_context) {
+        return true;
+      } else if (
+        parentIs(
+          node,
+          parent =>
+            parent.type === "CallExpression" &&
+            parent.arguments.length === 3 &&
+            parent.arguments[1] === node &&
+            parent.callee.type === "MemberExpression" &&
+            parent.callee.property.type === "Identifier" &&
+            parent.callee.property.name === "wrap",
+        )
+      ) {
+        // regeneratorRuntime.wrap(function foo$(_context) {
+        //   ...
+        // }, foo, this); // <- foo
+        return true;
+      } else if (
+        parentIs(
+          node,
+          parent =>
+            parent.type === "CallExpression" &&
+            parent.callee.type === "MemberExpression" &&
+            parent.callee.property.type === "Identifier" &&
+            parent.callee.property.name === "mark",
+        )
+      ) {
+        // regeneratorRuntime.mark(foo); // foo
+        return true;
+      }
+    } else if (node.type === "MemberExpression") {
+      // _context.next
+      return isByRegenerator(node.object);
+    } else if (node.type === "CallExpression") {
+      return isByRegenerator(node.callee);
+    } else if (node.type === "AssignmentExpression") {
+      // _context.next = 4;
+      return isByRegenerator(node.left);
+    } else if (node.type === "NumericLiteral") {
+      if (
+        parentIs(
+          node,
+          parent =>
+            parent.type === "AssignmentExpression" &&
+            isByRegenerator(parent.left),
+        )
+      ) {
+        // _context.next = 4; // <-- The 4
+        return true;
+      } else if (
+        parentIs(
+          node,
+          parent =>
+            parent.type === "CallExpression" &&
+            parent.callee.type === "MemberExpression" &&
+            isByRegenerator(parent.callee.object),
+        )
+      ) {
+        // return _context.abrupt("break", 11); // <-- The 11
+        return true;
+      }
+    }
+    return false;
+  };
+  const hidePrivateProperties = (key, val) => {
+    // Hides properties like _shadowedFunctionLiteral,
+    // which makes the AST circular
+    if (key[0] === "_") return "[Private]";
+    return val;
+  };
+  babel.types.traverseFast(ast, node => {
+    registerChildren(node);
+    if (isByRegenerator(node)) return;
+    if (nodes.has(node)) {
+      throw new Error(
+        "Do not reuse nodes. Use `t.clone` or `t.cloneDeep` to copy them.\n" +
+          JSON.stringify(node, hidePrivateProperties, 2) +
+          "\nParent:\n" +
+          JSON.stringify(parents.get(node), hidePrivateProperties, 2),
+      );
+    }
+    nodes.add(node);
+  });
+}
+
 function run(task) {
   const actual = task.actual;
   const expect = task.expect;
@@ -172,6 +361,7 @@ function run(task) {
   if (execCode) {
     const execOpts = getOpts(exec);
     result = babel.transform(execCode, execOpts);
+    checkDuplicatedNodes(result.ast);
     execCode = result.code;
 
     try {
@@ -188,6 +378,7 @@ function run(task) {
   const expectCode = expect.code;
   if (!execCode || actualCode) {
     result = babel.transform(actualCode, getOpts(actual));
+    checkDuplicatedNodes(result.ast);
     if (
       !expect.code &&
       result.code &&
--- a/packages/babel-helper-transform-fixture-test-runner/src/index.js
+++ b/packages/babel-helper-transform-fixture-test-runner/src/index.js
@@ -309,7 +309,7 @@ function checkDuplicatedNodes(ast) {
     if (isByRegenerator(node)) return;
     if (nodes.has(node)) {
       throw new Error(
-        "Do not reuse nodes. Use `t.clone` or `t.cloneDeep` to copy them.\n" +
+        "Do not reuse nodes. Use `t.cloneNode` to copy them.\n" +
           JSON.stringify(node, hidePrivateProperties, 2) +
           "\nParent:\n" +
           JSON.stringify(parents.get(node), hidePrivateProperties, 2),
--- a/packages/babel-types/src/clone/clone.js
+++ b/packages/babel-types/src/clone/clone.js
@@ -1,16 +1,12 @@
 // @flow
 
+import cloneNode from "./cloneNode";
+
 /**
- * Create a shallow clone of a `node` excluding `_private` properties.
+ * Create a shallow clone of a `node`, including only
+ * properties belonging to the node.
+ * @deprecated Use t.cloneNode instead.
  */
 export default function clone<T: Object>(node: T): T {
-  if (!node) return node;
-  const newNode = (({}: any): T);
-
-  Object.keys(node).forEach(key => {
-    if (key[0] === "_") return;
-    newNode[key] = node[key];
-  });
-
-  return newNode;
+  return cloneNode(node, /* deep */ false);
 }
--- a/packages/babel-types/src/clone/cloneNode.js
+++ b/packages/babel-types/src/clone/cloneNode.js
@@ -1,0 +1,69 @@
+import { NODE_FIELDS } from "../definitions";
+
+const has = Function.call.bind(Object.prototype.hasOwnProperty);
+
+function cloneIfNode(obj, deep) {
+  if (
+    obj &&
+    typeof obj.type === "string" &&
+    // CommentLine and CommentBlock are used in File#comments, but they are
+    // not defined in babel-types
+    obj.type !== "CommentLine" &&
+    obj.type !== "CommentBlock"
+  ) {
+    return cloneNode(obj, deep);
+  }
+
+  return obj;
+}
+
+function cloneIfNodeOrArray(obj, deep) {
+  if (Array.isArray(obj)) {
+    return obj.map(node => cloneIfNode(node, deep));
+  }
+  return cloneIfNode(obj, deep);
+}
+
+/**
+ * Create a clone of a `node` including only properties belonging to the node.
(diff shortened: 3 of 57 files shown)


packages/babel-core/test/fixtures/transformation/misc/regression-2892/expected.js
```diff
@@ -68,7 +68,7 @@ function _foo() {
       while (1) {
         switch (_context3.prev = _context3.next) {
           case 0:
-            _bar2 = function _bar2() {
+            _bar2 = function _ref2() {
               _bar2 = _asyncToGenerator(
```
nicolo-ribaudo: I think that this was a bug, since `_bar2` wasn't actually reassigned (because the variable was shadowed by the function)

packages/babel-helper-transform-fixture-test-runner/src/index.js
```diff
@@ -172,6 +322,7 @@ function run(task) {
   if (execCode) {
     const execOpts = getOpts(exec);
     result = babel.transform(execCode, execOpts);
+    checkDuplicatedNodes(result.ast);
```
Andarist: i guess this might create quite some overhead, ideally we should not do this by default, maybe only on CI?

nicolo-ribaudo: I will disable it on normal `make test` at the end of the pr, when the tests are green.

hzoo: is it actually slower?

packages/babel-helper-transform-fixture-test-runner/src/index.js
```diff
@@ -130,6 +130,156 @@ function wrapPackagesArray(type, names, optionsDir) {
...
+    // which makes the AST circular
+    if (key[0] === "_") return "[Private]";
+    return val;
+  };
+  babel.types.traverseFast(ast, node => {
+    registerChildren(node);
+    if (isByRegenerator(node)) return;
+    if (nodes.has(node)) {
+      throw new Error(
+        "Do not reuse nodes. Use `t.clone` or `t.cloneDeep` to copy them.\n" +
```
Kovensky: (Q: when to _not_ use cloneDeep?)

nicolo-ribaudo: When you are working with a literal, which doesn't have children

Kovensky: But if you are working with a literal, cloneDeep just won't find additional objects to recurse and so it won't recurse. Any reason why we can't just provide a single interface to reduce errors? Would the performance impact of the extra typeof check be too great?

nicolo-ribaudo: We can make `clone` behave as `cloneDeep` and special-case literals to avoid looping over their properties.

Andarist:
255 tests failing at the moment 🎉

nicolo-ribaudo:
@Andarist @Kovensky  What do you think of `t.cloneNode`? It is more lightweight that `t.cloneDeep` and it has a fast path for identifiers (There were more `t.clone` than `t.cloneNode`, so we clone a lot of identifiers).
It has the advantage of not requiring the developer to choose between clone and cloneDeep.

Also, I think we used somewhere clone instead of cloneDeep, because the number of failing tests decreased by 2 :tada:

nicolo-ribaudo:
We should add `[skip ci]` to our commits title to avoid running travis, since we know that it will fail for now.

Kovensky:
@nicolo-ribaudo you mean defining a new `cloneNode` that calls the appropriate `clone` variant?

I'd still prefer to have only One True Cloning Function, with a check at the start to decide on fast paths; we could then make `clone` and `cloneDeep` forward to it.

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  const newNode = (({
+    type: node.type,
+  }: any): T);
+  for (const field in NODE_FIELDS[node.type]) {
+    if (has(node, field)) {
+      newNode[field] = isNode(node[field])
+        ? cloneNode(node[field])
+        : node[field];
```
Kovensky: This won't clone nodes that are inside a statement list (`BlockStatement` children, etc)

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -0,0 +1,35 @@
+import { NODE_FIELDS } from "../definitions";
+
+const has = Function.call.bind(Object.prototype.hasOwnProperty);
```
Kovensky: (`call` is technically in `Function.prototype` but it works I guess)

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -0,0 +1,35 @@
...
+
+  if (!isNode(node)) {
+    throw new Error(`Unknown node type: "${node.type}"`);
+  }
+
+  const newNode = (({
+    type: node.type,
+  }: any): T);
+  for (const field in NODE_FIELDS[node.type]) {
+    if (has(node, field)) {
```
Kovensky: I'd tell you to use `Object.keys` but I assume you did this for speed

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -0,0 +1,35 @@
+import { NODE_FIELDS } from "../definitions";
+
+const has = Function.call.bind(Object.prototype.hasOwnProperty);
+const isNode = obj => obj && has(NODE_FIELDS, obj.type);
+
+export default function cloneNode<T: Object>(node: T): T {
+  if (!node) return node;
```
Kovensky: If `T` extends `Object` it can't be falsy, unless it's `document.all` :trollface: 
  
(I guess this is defensive programming)

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -0,0 +1,35 @@
+import { NODE_FIELDS } from "../definitions";
+
+const has = Function.call.bind(Object.prototype.hasOwnProperty);
+const isNode = obj => obj && has(NODE_FIELDS, obj.type);
+
+export default function cloneNode<T: Object>(node: T): T {
```
Kovensky: I'd say it would be better to extend `BabelNode` instead...... but actually reading the `BabelNode` definition and how it's used makes it a bit doubtful.

For example, in TypeScript, almost every valid node (written as an object literal) you gave it would get caught in excess property checks for weak types. It'd probably be better to make `BabelNode` be a generic, where the generic argument is the node type literal string.

But I digress and is outside the scope of this change.

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -0,0 +1,35 @@
...
+  if (!node) return node;
+
+  // Special-case identifiers since they are the most cloned nodes.
+  if (node.type === "Identifier") {
```
Kovensky: Maybe better to save `type` to a local var (`const { type } = node`) since you reuse it at least 3 times

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -0,0 +1,35 @@
...
+  }: any): T);
+  for (const field in NODE_FIELDS[node.type]) {
+    if (has(node, field)) {
+      newNode[field] = isNode(node[field])
+        ? cloneNode(node[field])
+        : node[field];
+    }
+  }
+
+  newNode.loc = node.loc || null;
```
Kovensky: Since you're now specifically only copying known AST keys, this is missing comments.

packages/babel-types/src/clone/cloneNode.js
nicolo-ribaudo: Oh lol, I copied lines 6-7 from `clone`. I will leave the check here, because plug-in authors might not use Flow.

Andarist:
@nicolo-ribaudo do you plan to remove `clone` and `cloneDeep` altogether?

nicolo-ribaudo:
Remove or deprecate with a nice warning message

Andarist:
If you gonna deprecate them, which I guess is fine, considering scope of the babel and its users count, I'd propose to have them just proxy to `cloneNode` with a deprecation warning - so actually we'd have only a single implementation for this stuff.

nicolo-ribaudo:
> If you gonna deprecate them, which I guess is fine, considering scope of the babel and its users count, I'd propose to have them just proxy to cloneNode with a deprecation warning - so actually we'd have only a single implementation for this stuff.

Done :+1: 

In the test output there is the depecation message twice, because regenerator uses `t.cloneDeep` [1]

[1] https://github.com/facebook/regenerator/blob/aca60c7bfae712fe4a1b1264415b73a87df6ca0e/packages/regenerator-transform/src/replaceShorthandObjectMethod.js#L56-L75

packages/babel-helper-transform-fixture-test-runner/src/index.js
nicolo-ribaudo: With the check:
```
make test-only  60,29s user 4,13s system 115% cpu 55,727 total
```

Without:
```
make test-only  57,88s user 3,98s system 115% cpu 53,697 total
```

We can always run the check.

nicolo-ribaudo:
24 failing tests :tada:

nicolo-ribaudo:
Zero tests failing locally :tada: 

@Andarist I'll rebase this commits to resolve the merge conflicts and squash them. Are you ok if I squash them like this?
1. `Disallow duplicated nodes in tests output (#7149)` (author: nicolo-ribaudo)
2. `Add t.cloneNode and deprecate t.clone and t.cloneDeep (#7149)` (author: nicolo-ribaudo)
3. `Fix reused nodes - part 1 (#7149)` (author: Andarist)
4. `Fix reused nodes - part 2 (#7149)` (author: nicolo-ribaudo)

nicolo-ribaudo:
I've gone ahead and squashed. The original commits are at https://github.com/nicolo-ribaudo/babel/tree/duplicated-nodes-old.

There were so many conflicts when reordering commits that I don't know how I managed not to break everything :laughing:

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -0,0 +1,54 @@
+import { NODE_FIELDS } from "../definitions";
+
+const has = Function.call.bind(Object.prototype.hasOwnProperty);
+const isNode = obj => obj && has(NODE_FIELDS, obj.type);
+
+function cloneIfNodeOrArray(obj) {
+  if (isNode(obj)) return cloneNode(obj);
+  if (Array.isArray(obj)) return obj.map(cloneIfNodeOrArray);
```
loganfsmyth: I don't think there are cases were we'd ever have arrays of arrays, so we can probably have this make `cloneNode` instead of `cloneIfNodeOrArray`?

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -0,0 +1,54 @@
...
+    newNode.trailingComments = node.trailingComments;
+  }
+  if (has(node, "extra")) {
+    newNode.extra = shallowClone(node.extra);
+  }
+
+  return newNode;
+}
+
+function shallowClone<T: Object>(obj: T): T {
```
loganfsmyth: This is just

    return Object.assign({}, obj);

so I'd just remove this function and do

    newNode.extra = Object.assign({}, node.extra);

packages/babel-types/src/clone/clone.js
```diff
@@ -1,16 +1,8 @@
...
-
-  Object.keys(node).forEach(key => {
-    if (key[0] === "_") return;
-    newNode[key] = node[key];
-  });
+import cloneNode from "./cloneNode";
 
-  return newNode;
+export default function clone<T: Object>(node: T): T {
+  console.trace("t.clone is deprecated; use t.cloneNode instead.");
```
loganfsmyth: I don't think deprecating these is the way to go currently. We should aim to make it as easy as possible for users to upgrade from Babel 6 to Babel 7, and while we can certainly update all of our own plugins, there are going to be third-party plugins that use `clone` and `cloneDeep`, and no user is going to upgrade to Babel 7 if it suddenly means their console is filled with deprecation warnings.

I think we probably _should_ have warnings like this in the future, but things feels like too painful for users currently. Ideally warnings should show for the people that actually have a way to address the warnings. To me that means that we should explore a way for plugin authors to opt in to additional warnings, without requiring Babel's users to have to deal with them.

For this specific case, I'm also not sure I'm comfortable with this doing a deep copy when the API says it is a shallow copy, since that could also break existing third-party code. Can we expand the API to

    export default function cloneNode<T: Object>(node: T, deep: boolean = true): T {

so it clones deeply by default but people can opt into shallow copying?

Andarist: Is there any actual use case for a shallow version of `cloneNode`? Not doing deep cloning for previously shallow `clone` seems reasonable, but for that we just have to leave the old implementation intact and be done with it.

Flooding the console with those warnings is not ideal and we should just do it once, but Im dont agree that Babel users should not see those warnings - let them see them and give them opportunity to fix those issues upstream, thats a beauty of open source, that not everything has to be on the shoulders of original creator/current maintainer.

Kovensky: If, for some reason, we still need to provide a shallow clone, it'd be best to name it `shallowCloneNode` instead of just leaving `clone` as is; it's very use to misuse `clone` by accident.

nicolo-ribaudo: I agree with @Andarist. Most libraries show deprecation warnings even when they are not used directly, when the developer doesn't own the code which uses the deprecated API. It's a stronger motivation for the plug-in maintainers to upgrade to the new API before that it is completely removed.

I could add `grep for this message and replace console.warn with console.trace to see where it comes from` to make it easier for users to find which plugin needs to be updated (we already did something similar for a warning message I removed, I think it was the one about Flow bindings).

> For this specific case, I'm also not sure I'm comfortable with this doing a deep copy when the API says it is a shallow copy, since that could also break existing third-party code

You are right, I'll restore the old behavior of the `clone` function. Probably I should also restore the old behavior of the `cloneDeep` function, since it is an unnecessary breaking change?

> Can we expand the API to
> 
> ```js
   export default function cloneNode<T: Object>(node: T, deep: boolean = true): T {
> ```
>
> so it clones deeply by default but people can opt into shallow copying?

> If, for some reason, we still need to provide a shallow clone, it'd be best to name it shallowCloneNode

When could shallow cloning be needed?

nicolo-ribaudo:
The discussion at https://github.com/babel/babel/pull/7149#discussion_r160069189 is not done yet (GitHub hides it because I changed `console.trace` -> `console.warn`)

packages/babel-types/src/clone/clone.js
Kovensky: I have no idea, other than when the caller intentionally used `t.clone` instead of `t.cloneDeep` 🤔

loganfsmyth: > When could shallow cloning be needed?

To me it's ideal as a performance optimization. If I want to mutate a node, but don't intend to mutate any of its children, performing a shallow clone is all that is needed. Deep being the default seems ideal, but opting in to shallow cloning does seem like a useful performance optimization to me.

> You are right, I'll restore the old behavior of the clone function. Probably I should also restore the old behavior of the cloneDeep function, since it is an unnecessary breaking change?

I don't really think it makes sense to maintain two separate copies of these APIs when it isn't absolutely required, which is why I'm in favor of just having them use your new logic.

> Flooding the console with those warnings is not ideal and we should just do it once, but Im dont agree that Babel users should not see those warnings - let them see them and give them opportunity to fix those issues upstream, thats a beauty of open source, that not everything has to be on the shoulders of original creator/current maintainer.

This seems like claiming "open source software should be painful to use", which I absolutely don't get. An average user does not care one bit that we've decided to deprecate this API, and until we actually remove it entirely, they have no reason to care. Open source means anyone can submit a fix upstream, which is fantastic, but it shouldn't mean that every user of a plugin will get an annoying message until someone submits a patch.

Ideally the behavior here would be that we have a standard plugin testbed runs a plugin against whichever versions of Babel it says it supports. Random users can submit PRs to expand the supposed Babel range, which should fail the tests if deprecated functions are used. Maintainers don't need to constantly be adjusting their Babel version but users can submit PRs, and random users don't need to worry until it actually matters to them.

Again, I'm not saying never show things to users, I'm advocating for a more stages approach. Say in a first pass we only log deprecation warnings to plugin devs, and then finally we change the function to just throw. The average user doesn't have an issue until it actually starts throwing, but devs can be aware of deprecations via their tests. We could even have a middle stage that lots the deprecation warning for all, but honestly it doesn't seem helpful.

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -0,0 +1,50 @@
...
+
+const has = Function.call.bind(Object.prototype.hasOwnProperty);
+const isNode = obj => obj && has(NODE_FIELDS, obj.type);
+
+function cloneIfNode(obj) {
+  return isNode(obj) ? cloneNode(obj) : obj;
+}
+
+function cloneIfNodeOrArray(obj) {
+  return Array.isArray(obj) ? obj.map(cloneIfNode) : cloneIfNode(obj);
```
loganfsmyth: I don't think we have any types that are arrays of non-Node types. As it is, this will quietly fail to clone

```
{

Comment (loganfsmyth) [reply packages/babel-types/src/clone/cloneNode.js]: Expanding on this, I wonder if instead we should just make `cloneIfNode` do 

```
return obj && typeof obj === "object" && typeof obj.type === "string" ? cloneNode(obj) : obj;

nicolo-ribaudo:
@loganfsmyth WDYT about creating an eslint plugin which disallows `t.clone` and `t.cloneDeep`, like we did for `t.identifier("undefined")`? We could create a `babel-plugins` eslint preset in the future.

hzoo:
> We could create a babel-plugins eslint preset in the future.

Yeah and we should create a tool that does that for you - tests/lint/build/publish/deprecates/etc

nicolo-ribaudo:
@loganfsmyth I addressed your comments. I still don't 100% like allowing shallow cloning, but since it is an explicit opt-in it is way better than `t.clone`.

hzoo:
So this can close https://github.com/babel/babel/issues/6375 then?

nicolo-ribaudo:
Reusing ast nodes still has problems, but we can't do anything else to prevent it. I guess we can close that issue, but it should be labeled as `wontfix`.

hzoo:
Let's try to land this asap? @loganfsmyth thoughts?

nicolo-ribaudo:
When merging, please don't squash to keep both @Andarist and I as the commit authors.

hzoo:
You can merge if it's good now 👍

nicolo-ribaudo:
I'm waiting for travis and circle ci to finish :+1:

Andarist:
wohoooo 🎉

hzoo:
amazing - now we should figure out how to make this a thing in all the 3rd party plugins with our plugin cli tool

PR #7152, opened 2018-01-04
Suggest JSX fragment syntax in adjacent tag error

sophiebits:


rgbkrk:
Nice! Love the suggestion text.

Andarist:
maybe a note that it's available for React 16.2 and above should be added?

Issue #6889, opened 2017-11-24
[pipeline operator] support await


xtuc:
Choose one: is this a bug report or feature request?

### Input Code

```js
promise |> await
```

which is the equivalent of

```js
await promise
```

### Expected Behavior

```js
var _promise;

_promise = promise, await _promise;
```

### Current Behavior

```
Parse error: await is a reserved word (1:11)
```

### Your Environment

| software         | version(s)
| ---------------- | -------
| Babel            |  
| Babylon          | 
| node             | 
| npm              | 
| Operating System | 

As described in the spec [here](https://github.com/tc39/proposal-pipeline-operator#use-of-await).

lilijialiang:
Will not continue to support `await` ? sorry to disturb.

xtuc:
Sorry I don't understand your point @lilijialiang. This is only an addition to the existing support in Babel.

ksmithbaylor:
Does the spec actually support this? The example in the README only shows

```javascript
x |> await f
```

which is an equivalent of

```javascript
await f
```

And in the current grammar defined in the spec draft, there is no option for a RHS without a *LogicalORExpression* expansion.

mAAdhaTTah:
This is not currently specced yet but will be supported in the parser in #7458 and eventually added as an option in the pipeline plugin.

mAAdhaTTah:
@thiagoarrais I'll be implementing this as part of the F# parsing. It's not strictly necessary to track but isn't technically "resolved". Up to the maintainers what they want to do here.

thiagoarrais:
Sorry for the noise. I interpreted that this was already implemented in F#.

nicolo-ribaudo:
I added this to https://github.com/babel/babel/projects/9#card-17016058 so we don't forget to close it when it is fixed

nicolo-ribaudo:
Supported by https://github.com/babel/babel/pull/9984

Issue #5854, opened 2017-06-13
Plugin Ordering (Part 2)


hzoo:
First discussion issue: https://github.com/babel/babel/issues/5623

> James wrote a good article about the general problem: http://thejameskyle.com/babel-plugin-ordering.html  
> Babel is single pass traversal so top level ordering isn't a thing (unless you use `passPerPreset` which is why we are going to remove that), however the individual node visitors have an ordering.  

```js
const visitors = plugins.map(plugin => plugin.visitor);
const mergedVisitors = merge(visitors);
traverse(ast, mergedVisitors);
```

I believe most agree on:

- [x] introduce a before/after key to specify which plugin ordering
- [x] should error/warn if the same plugin/capability is specified twice
  - [ ] edge case would be if you are using a plugin in a "preset" that you want to still use but want to modify it by using a plugin option - either error and not use the preset anymore or the user specified plugin should override the one used in the preset itself
- [x] should any of this be user-specified (mostly likely no: if it's possible to encode for plugin authors we don't need to expose that to users)

### Questions

- [ ] what exactly should be specified in the before/after:
  - [ ] a specific plugin name (probably taken from npm package) `babel-plugin-transform-es2015-arrow-function`
  - [ ] the more generic "capability" of many plugins (ex: `"arrowFunction", "jsx", "flow", "asyncFunction"`)
- [ ] does minification/info gathering plugin require a different approach/multiple passes again? They don't depend on a specific node type to run before/after but rather always before/after "everything else"?
  - [#5623](https://github.com/babel/babel/issues/5623) discusses allowing free form strings for before/after vs. specifically named passes.
- [ ] how does the native host environment fit into this? How can we encode their support via preset-env data. If a syntax is natively provided, then it provides the capability and another plugin can output that syntax without it being transformed.

sarupbanskota:
for the name bit within the core plugin packages, I've been using the npm package name so far: https://github.com/babel/babel/pull/5842/commits/fc7b17bcffdfdec851992e0dea6d6320ac46697c.

lopugit:
How is this still an issue in V7 WTF? 

I had to use the manual `Program(path){ path.traverse(realVisitors) }` solution ???

Just add the Phases or Stages idea that someone mentioned in the 3 million replies massive ass AI focused analyze what transformations depend on what bullshit convo.

JUST RUN SEPARATE TRAVERSALS. What in the F

AprilArcus:
[...]

PR #7154, opened 2018-01-04
Add basic support for await with the pipeline operator

ksmithbaylor:
| Fixed Issues?            | https://github.com/babel/babel/issues/6889
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

This correctly transforms `AwaitExpression`s which appear in pipelines, as shown [here](https://github.com/tc39/proposal-pipeline-operator#use-of-await). It does not support standalone `await` keywords, as was the case in earlier versions of the proposal.

### DIFF
--- a/packages/babel-helper-remap-async-to-generator/src/index.js
+++ b/packages/babel-helper-remap-async-to-generator/src/index.js
@@ -13,6 +13,7 @@ const awaitVisitor = {
 
   AwaitExpression(path, { wrapAwait }) {
     const argument = path.get("argument");
+    argument.node.__wasAwait = true;
 
     if (path.parentPath.isYieldExpression()) {
       path.replaceWith(argument.node);
--- a/packages/babel-plugin-proposal-pipeline-operator/src/index.js
+++ b/packages/babel-plugin-proposal-pipeline-operator/src/index.js
@@ -15,7 +15,10 @@ export default function() {
 
         let optimizeArrow =
           t.isArrowFunctionExpression(right) && t.isExpression(right.body);
-        const wasAwait = right.argument && right.argument.__wasAwait;
+        const wasAwait =
+          t.isYieldExpression(right) &&
+          right.argument &&
+          right.argument.__wasAwait;
         const isAwait = t.isAwaitExpression(right) || wasAwait;
         let param;
 
--- a/packages/babel-plugin-proposal-pipeline-operator/src/index.js
+++ b/packages/babel-plugin-proposal-pipeline-operator/src/index.js
@@ -15,8 +15,8 @@ export default function() {
 
         let optimizeArrow =
           t.isArrowFunctionExpression(right) && t.isExpression(right.body);
-        const isAwait = t.isAwaitExpression(right);
-        const isYield = t.isYieldExpression(right);
+        const wasAwait = right.argument && right.argument.__wasAwait;
+        const isAwait = t.isAwaitExpression(right) || wasAwait;
         let param;
 
         if (optimizeArrow) {
@@ -26,7 +26,7 @@ export default function() {
           } else if (params.length > 0) {
             optimizeArrow = false;
           }
-        } else if (isAwait || isYield) {
+        } else if (isAwait) {
           right.argument = t.sequenceExpression([
             t.numericLiteral(0),
             right.argument,
@@ -51,8 +51,8 @@ export default function() {
 
         const call = optimizeArrow
           ? right.body
-          : isAwait || isYield
-            ? t[isAwait ? "awaitExpression" : "yieldExpression"](
+          : isAwait
+            ? t[wasAwait ? "yieldExpression" : "awaitExpression"](
                 t.callExpression(right.argument, [placeholder]),
               )
             : t.callExpression(right, [placeholder]);
--- a/packages/babel-plugin-proposal-pipeline-operator/src/index.js
+++ b/packages/babel-plugin-proposal-pipeline-operator/src/index.js
@@ -15,6 +15,7 @@ export default function() {
 
         let optimizeArrow =
           t.isArrowFunctionExpression(right) && t.isExpression(right.body);
+        const isAwait = t.isAwaitExpression(right);
         let param;
 
         if (optimizeArrow) {
@@ -24,6 +25,11 @@ export default function() {
           } else if (params.length > 0) {
             optimizeArrow = false;
           }
+        } else if (isAwait) {
+          right.argument = t.sequenceExpression([
+            t.numericLiteral(0),
+            right.argument,
+          ]);
         } else if (t.isIdentifier(right, { name: "eval" })) {
           right = t.sequenceExpression([t.numericLiteral(0), right]);
         }
@@ -44,7 +50,9 @@ export default function() {
 
         const call = optimizeArrow
           ? right.body
-          : t.callExpression(right, [placeholder]);
+          : isAwait
+            ? t.awaitExpression(t.callExpression(right.argument, [placeholder]))
+            : t.callExpression(right, [placeholder]);
         path.replaceWith(
           t.sequenceExpression([
             t.assignmentExpression("=", placeholder, left),
--- a/packages/babel-plugin-proposal-pipeline-operator/src/index.js
+++ b/packages/babel-plugin-proposal-pipeline-operator/src/index.js
@@ -16,6 +16,7 @@ export default function() {
         let optimizeArrow =
           t.isArrowFunctionExpression(right) && t.isExpression(right.body);
         const isAwait = t.isAwaitExpression(right);
+        const isYield = t.isYieldExpression(right);
         let param;
 
         if (optimizeArrow) {
@@ -25,7 +26,7 @@ export default function() {
           } else if (params.length > 0) {
             optimizeArrow = false;
           }
-        } else if (isAwait) {
+        } else if (isAwait || isYield) {
           right.argument = t.sequenceExpression([
             t.numericLiteral(0),
             right.argument,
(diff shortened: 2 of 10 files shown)


nicolo-ribaudo:
Can you transform async functions in the tests? Otherwise they will fail on node <= 6

ksmithbaylor:
I updated the PR. Since I'd like to test both "pure" async functions and also those transformed to generators by `transform-async-to-generator`, I duplicated the test directory. I marked the original one as requiring Node 7.6.0 or above, and enabled the `transform-async-to-generator` plugin on the other one. This required some code changes, since this plugin should now handle `yield` expressions in addition to `async`.

It isn't clear to me that this is allowed according to the spec draft, however, and I expect that I could be doing the transformation more elegantly. This is my first time looking at the Babel source, and I'm still learning my way around. I'd appreciate feedback!

packages/babel-plugin-proposal-pipeline-operator/src/index.js
```diff
@@ -15,6 +15,8 @@ export default function() {
 
         let optimizeArrow =
           t.isArrowFunctionExpression(right) && t.isExpression(right.body);
+        const isAwait = t.isAwaitExpression(right);
+        const isYield = t.isYieldExpression(right);
```
nicolo-ribaudo: `foo |> yield bar` shouldn't be transfrmed; it should become something like (`_temp = yield bar, _temp(foo)`).

If there isn't a way of transforming `|> await foo` _before_ transforming the async function, you can workaround the problem by adding `node.__wasAwait` to yield expressions generated by the async functions plugin and then do
```js
const isAwait = t.isAwaitExpression(right) || right.__wasAwait;
```

ksmithbaylor:
Alright, should be better now. I couldn't figure out a way to ensure that the pipeline operator plugin ran before the async-to-generator plugin, since it operates on a BinaryExpression which doesn't have a consistent parent to traverse down from. So I took your suggestion of adding a flag to await expressions that have been transformed to yield. I also added a test case that makes sure yield expressions aren't transformed.

Andarist:
Are special flags a common practice? I have not yet seen many of those and it seems a clunky solution (although might be necessary ofc). We really should add plugin ordering to solve such issues

ksmithbaylor:
@Andarist no, I don't really see any other examples of special flags either. I'd love to find a better way to do the check, since it does seem clunky like you said. This is my first time looking at the codebase, so I have lots of learning to do :)

As I understand it, there is [some discussion](https://github.com/babel/babel/issues/5854) around what it would look like to allow node visitors to be ordered more predictably, but a solution hasn't been decided on yet. If you know of an easier way to either force this plugin to run first or make sure `yield` expressions are only transformed if they were previously `await` expressions, I'd be open to suggestions.

Andarist:
Hm, Im also still learning and I dont know about all intricacies of babel transformations.

I thought that babel's visitor is "top down" and I would suspect that this BinaryExpression would be visited before AwaitExpression (as former is the parent). But it seems that this is not true here, right?

packages/babel-helper-remap-async-to-generator/src/index.js
```diff
@@ -13,6 +13,7 @@ const awaitVisitor = {
 
   AwaitExpression(path, { wrapAwait }) {
     const argument = path.get("argument");
+    argument.node.__wasAwait = true;
```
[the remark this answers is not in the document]
ksmithbaylor: I agree. Do you have any suggestions that could help either act as an indicator or force these two plugins to run in order that wouldn't require blocking until those issues are solved at a higher level?

packages/babel-plugin-proposal-pipeline-operator/test/fixtures/pipeline-operator/async-with-transform-async/exec.js
```diff
@@ -0,0 +1,22 @@
...
+var double = x => x * 2;
+
+var result = async () => 10 |> await incPromise;
+var result2 = async () => 10 |> await incPromise |> double;
+
+function* foo() {
+  return 42 |> (yield 10);
+}
+
+var fooGen;
```
Andarist: any reason to have this outside of the promise?

packages/babel-plugin-proposal-pipeline-operator/test/fixtures/pipeline-operator/async-with-transform-async/actual.js
```diff
@@ -0,0 +1,10 @@
+var incPromise = (x) => Promise.resolve(x + 1);
```
Andarist: i'd prefer to have the same code in actual / exec if they are in the same directory, exec differs slightly right now and it makes it hard to compare those files (assert ofc should stay just in exec)

packages/babel-plugin-proposal-pipeline-operator/test/fixtures/pipeline-operator/async-with-transform-async/expected.js
```diff
@@ -0,0 +1,39 @@
...
+
+  return function result2() {
+    return _ref2.apply(this, arguments);
+  };
+}();
+
+function* foo() {
+  var _3;
+
+  return _3 = 42, (yield incFuncPromise)(_3);
```
Andarist: was this intentional? looks weird, but i guess it could test something, but then exec should reflect that test

Andarist:
@ksmithbaylor regarding why AwaitExpression gets visited first - I **think** its because actually its not visited from a "top level" traversal - there is a `traverse` call from some other visitor and that's why whole subtree is visited first there (i guess `traverse` is called in async/generator function, which is a parent of that BinaryExpression). I have not tested this, but I believe that's the case.

nicolo-ribaudo:
@Andarist You are right: `@babel/helper-remap-async-to-generator` transforms `await` expressions by spawning a new traversal:
https://github.com/babel/babel/blob/47ce7e71c9f97d151be493d05755562cb299fcca/packages/babel-helper-remap-async-to-generator/src/index.js#L66-L69

I _think_ that if we transform async functions on exit this problem would go away, but probably it would make it be traversed one more time.

Andarist:
> I think that if we transform async functions on exit this problem would go away, but probably it would make it be traversed one more time.

Why would it be traversed twice?

nicolo-ribaudo:
Actually it is already traversed twice (the main traversal and the one spawned by the helper); it would be traversed three times.

> oh yeah, content of the async function would get traversed twice, because after remapping it would get requeued, right?

Yes :+1:

packages/babel-plugin-proposal-pipeline-operator/src/index.js
```diff
@@ -15,6 +15,8 @@ export default function() {
 
         let optimizeArrow =
           t.isArrowFunctionExpression(right) && t.isExpression(right.body);
+        const wasAwait = right.argument && right.argument.__wasAwait;
```
nicolo-ribaudo: Maybe
```js
const wasAwait = t.isYieldExpression(right) && right.argument && right.argument.__wasAwait;
```
?

packages/babel-plugin-proposal-pipeline-operator/test/fixtures/pipeline-operator/async-with-transform-async/exec.js
```diff
@@ -0,0 +1,21 @@
+var incPromise = x => Promise.resolve(x + 1);
+var incFuncPromise = Promise.resolve(x => x + 10);
```
nicolo-ribaudo: This is not used.

packages/babel-plugin-proposal-pipeline-operator/test/fixtures/pipeline-operator/async-with-transform-async/actual.js
ksmithbaylor: I made them match now. I think I may have saved the file with Prettier on at some point.

Andarist: Oh, I didnt mean that they do not match in sense of formatting - i think there was some differenc in how `incFuncPromise` was used

ksmithbaylor: Oh, got it. I also changed the usage of that variable down below. Either way, they're the same now.

Andarist: Cool!

ksmithbaylor:
Does anyone have ideas on how to improve this without relying on special flags?

mAAdhaTTah:
Wondering if it would be acceptable to land this w/o supporting transformed `await` and revist that issue in the future?

packages/babel-helper-remap-async-to-generator/src/index.js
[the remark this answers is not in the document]
xtuc: Unfortunately no, it's related to https://github.com/babel/babel/issues/5854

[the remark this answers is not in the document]
loganfsmyth: Plugin ordering definitely makes this harder. Could we make this throw an error if the async function tries to transform an `await` that is inside of a pipeline expression, and then have the pipeline transform have a visitor for async functions  that does an explicit sub-traversal for pipeline expressions? Then we could just make sure that the pipeline plugin runs before the async function transform and I think we'd be okay?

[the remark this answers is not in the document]
xtuc: That sounds good to me, we don't need to order the plugins if they are both idenpendent. Each will transform what it needs.

mAAdhaTTah:
So this PR may not actually be needed. We're going to ban `await` in the basic, "minimal proposal" spec here: https://github.com/tc39/proposal-pipeline-operator/pull/108 and solve `await` in each of the proposals we're working on currently working on. This is not going to be the final semantics of pipeline in any of the proposals, given the ambiguity between whether `x |> await f` becomes `await f(x)` or `(await f)(x)`, so I think we can just close this.

ksmithbaylor:
@mAAdhaTTah agreed!

Issue #7157, opened 2018-01-04
TypeScript 2.7 introduces the definite property assignment assertion operator (!)


DanielRosenwasser:
While I am actually enthusiastic, the `!` in the title actually has more to do with the syntax. TypeScript 2.7 permits a bang following any class property declaration.

```ts
class C {
  foo!: number;
  constructor() {
  }
}
```

The above is an example of such usage. It asserts that `foo` will be initialized even if it is not initialized within the constructor.

On the other hand, this is only allowed on class instance properties

```ts
class C {
  static foo!: number; // TypeScript errors: A definite assignment assertion '!' is not permitted in this context.
}
```

@andy-ms, can you take this on before the 7.0 release?

hzoo:
Done with https://github.com/babel/babel/pull/7159?

epeli:
Yeah it's good! 💪

PR #7159, opened 2018-01-05
typescript: Support definite assignment assertion

andy-ms:
| Fixed Issues?            | #7157
| Documentation PR         |

Supports parsing the `!` postfix syntax on local variables and class properties. Also supports it in `babel-generator` and transforms it away in `babel-plugin-transform-typescript`.
  
CC @DanielRosenwasser and @JamesHenry for review.


packages/babylon/src/types.js
```diff
@@ -631,7 +634,7 @@ export type ClassMemberBase = NodeBase &
     // TypeScript only:
     accessibility?: ?Accessibility,
     abstract?: ?true,
-    optional?: ?true,
+    optional?: ?true, //!
```
nicolo-ribaudo: Did you forgot `exclamation`?

andy-ms: Actually that properly goes on `ClassProperty` only, not on any other class members. Left a comment behind though.

packages/babel-types/src/definitions/core.js
```diff
@@ -836,6 +836,10 @@ defineType("VariableDeclarator", {
     id: {
       validate: assertNodeType("LVal"),
     },
+    exclamation: {
```
nicolo-ribaudo: Maybe we should call it `definiteAssertion` (or `assertDefined`) instead of `exclamation`? I saw that tsc uses `node.exclamationToken`, but it also uses `node.questionToken` instead of `node.optional`.

DanielRosenwasser:
The TypeScript playground isn't on 2.7. Try `npm install typescript@next`.

nicolo-ribaudo:
@DanielRosenwasser The don't work, even if I'm using `typescript@next`:
```js
~/Documenti/ts-demo via ⬢ v9.3.0 
➜ ./node_modules/.bin/tsc --version
Version 2.7.0-dev.20180110

~/Documenti/ts-demo via ⬢ v9.3.0 
➜ cat demo.ts 
var a!: number;
var x!: number = 2;
var { foo }!: { foo: number } = { foo: number }
class Class {
  a?!: number;
}

~/Documenti/ts-demo via ⬢ v9.3.0 
➜ ./node_modules/.bin/tsc demo.ts  
demo.ts(3,12): error TS1005: ',' expected.
demo.ts(3,13): error TS1109: Expression expected.
demo.ts(3,31): error TS1128: Declaration or statement expected.
demo.ts(5,5): error TS1005: ';' expected.
demo.ts(5,6): error TS1109: Expression expected.
demo.ts(6,1): error TS1128: Declaration or statement expected.

~/Documenti/ts-demo via ⬢ v9.3.0 
➜ cat demo.js 
var a;
var x = 2;
var foo = (void 0).foo;
!;
{
    foo: number;
}
{
    foo: number;
}
var Class = /** @class */ (function () {
    function Class() {
    }
    return Class;
}());
!;
number;
```

packages/babylon/src/plugins/typescript.js
```diff
@@ -1917,7 +1925,9 @@ export default (superClass: Class<Parser>): Class<Parser> =>
     }
 
     isClassProperty(): boolean {
-      return this.match(tt.colon) || super.isClassProperty();
+      return (
+        this.match(tt.bang) || this.match(tt.colon) || super.isClassProperty()
```
DanielRosenwasser: Weird, how does `?` get parsed from here?

packages/babylon/src/plugins/typescript.js
```diff
@@ -1704,6 +1708,10 @@ export default (superClass: Class<Parser>): Class<Parser> =>
     // `let x: number;`
     parseVarHead(decl: N.VariableDeclarator): void {
       super.parseVarHead(decl);
+      if (this.eat(tt.bang)) {
```
DanielRosenwasser: Like @nicolo-ribaudo, it probably doesn't make sense to parse a `!` here unless there's a way to issue an error later.

packages/babylon/src/plugins/typescript.js
nicolo-ribaudo: I think that it is done in parsePostMemberNameModifiers

packages/babel-types/src/definitions/core.js
sgoll: In theory, I'd stick to the syntactic characteristics of a token to ease later semantic changes to its meaning, i.e. name it `exclamation[Token]` instead of `definite` since token definitions may later be broadened to include more than the original intentions—and the parser should add as little semantic overlay/interpretation as possible, IMHO.

However, since there is precedent for using semantic descriptions even when the TS AST uses a syntactic naming scheme (`questionToken` → `optional`), following that same path again may be the better way with regard to internal consistency.

In any case, my feelings are not very strong either way, so I'd be fine with both.

DanielRosenwasser:
@nicolo-ribaudo I would give it another shot. Several of those should now parse correctly. I think in general, we should just have the *parsing* behavior match what TypeScript does (i.e. certain things still parse, but don't necessarily lead to errors). Basically, the behavior for variable declarations is just to allow a `!` after identifier names (i.e. not binding patterns). @andy-ms will send an update to restrict the current behavior to only identifier declarations.

chrisgervang:
2.7 was released today, in case `typescript@next` was causing issues

packages/babel-types/src/definitions/core.js
```diff
@@ -836,6 +836,10 @@ defineType("VariableDeclarator", {
     id: {
       validate: assertNodeType("LVal"),
     },
+    definiteAssignment: {
```
j-f1: How about `definite`?

nicolo-ribaudo:
`var x!: number = 2;` is an error in TS, but is it a parser error or a type error? In the first case, it should be disallowed by Babylon.

andy-ms:
TypeScript does not really distinguish between error kinds. It's a "grammar error", meaning that parsing completes without adding any diagnostics, but the TypeChecker will add a diagnostic on the node. But that's more an implementation detail than anything.

nicolo-ribaudo:
Ok. WDYT that Babylon should do then? I think that it should throw, but I don't know TS much.

andy-ms:
I don't see any benefit to throwing here -- TypeScript users already need to run tsc to get all errors, so there's never really a time when we need to throw in the parser other than for convenience of implementation.

Issue #7128, opened 2017-12-29
Inner-references in TypeScript enums should resolve appropriately


DanielRosenwasser:
**Bug**

### Input Code

```ts
enum E {
    a = 10,
    b = a
}
```

### Babel/Babylon Configuration (.babelrc, package.json, cli command)

```json
{
    "presets": [
        "@babel/env",
        "@babel/typescript"
    ],
    "plugins": [
        "@babel/proposal-class-properties",
        "@babel/proposal-object-rest-spread"
    ]
}
```

```sh
babel src --out-dir lib --extensions ".ts,.tsx"
```

### Expected Behavior

Within the enums, we should resolve `a` to `E.a` and appropriately emit something like `E.a`

```js
var E;

(function (E) {
    E[E["a"] = 10] = "a";
    E[E["b"] = 10] = "b";
})(E || (E = {}));
```

### Current Behavior

```js
var E;
(function (E) 

  E[E["a"] = 10] = "a";
  E[E["b"] = a] = "b";
})(E || (E = {}));
```

### Your Environment

Babel 7

maaz93:
I'd like to take this up, if that's ok. I think I know the fix. The code dealing with evaluating the constant value has a mechanism by which it tracks the `seen` identifiers. This `seen` map is never updated. Hence the value of `b` in the above example just remains `a` and doesn't get the already evaluated value of `a`, i.e. `10`.

I think the fix is to add `seen[name] = constValue;` just above the below piece of code.

https://github.com/babel/babel/blob/f3410e761ceff73cf7c2c45bb79fff6479a28d5b/packages/babel-plugin-transform-typescript/src/enum.js#L104

gs-akhan:
@maaz93 Just a side question, How do you debug this TS parser ? You add break points into the source and run it for a simple script ?

maaz93:
@gs-akhan Yes. But I don't know if  you can debug without running tests (the [contribution guidelines](https://github.com/babel/babel/blob/master/CONTRIBUTING.md#debugging-code) talk about using tests). So, I run the tests related to this particular plugin, using `TEST_DEBUG=true TEST_ONLY=transform-typescript make test-only`) and open `chrome://inspect` in  Chrome. There, you'll find a **Inspect** button. Clicking on it will start the Chrome DevTools debugger and normal debugging can continue. If you want to stay for long in the debugger and have the tests run fully, you can change the `timeout` [here](https://github.com/babel/babel/blob/master/test/mocha.opts).

[...]

Issue #7126, opened 2017-12-28
TypeScript string enum members should not have a reverse mapping


DanielRosenwasser:
**Bug**

### Input Code

```ts
enum E {
    a = "hello",
    b = a,
}
```

### Babel/Babylon Configuration (.babelrc, package.json, cli command)

```json
{
    "presets": [
        "@babel/env",
        "@babel/typescript"
    ],
    "plugins": [
        "@babel/proposal-class-properties",
        "@babel/proposal-object-rest-spread"
    ]
}
```

```sh
babel src --out-dir lib --extensions ".ts,.tsx"
```

### Expected Behavior

```js
var E;

(function (E) {
    E["a"] = "hello";
    E["b"] = "hello";
})(E || (E = {}));
```

or

```js
var E;

(function (E) {
    E["a"] = "hello";
    E["b"] = E.b;
})(E || (E = {}));
```

http://www.typescriptlang.org/play/index.html#src=enum%20E%20%7B%0D%0A%20%20%20%20a%20%3D%20%22hello%22%2C%0D%0A%20%20%20%20b%20%3D%20a%2C%0D%0A%7D

### Current Behavior

```js
var E;

(function (E) {
  E[E["a"] = "hello"] = "a";
  E[E["b"] = a] = "b";
})(E || (E = {}));
```

### Your Environment

Babel 7

existentialism:
Fixed by https://github.com/babel/babel/pull/7160

PR #7160, opened 2018-01-05
typescript: Fix enum emit when values are strings

andy-ms:
| Fixed Issues?            | #7126, #7128
| Documentation PR         |

When the value is a string, emit `E["B"] = "";` and not `E[E[""] = "B"] = ""`.
Also starts handling strings in the set of previously-seen values.
CC @DanielRosenwasser for review.


packages/babel-plugin-transform-typescript/src/enum.js
```diff
@@ -88,21 +90,27 @@ function enumFill(path, t, id) {
...
  *   }
  */
-type PreviousEnumMembers = { [name: string]: number | typeof undefined };
+type PreviousEnumMembers = { [name: string]: number | string | undefined };
 
 function translateEnumValues(path, t) {
   const seen: PreviousEnumMembers = Object.create(null);
   // Start at -1 so the first enum member is its increment, 0.
-  let prev: number | typeof undefined = -1;
+  let prev: number | undefined = -1;
```
nicolo-ribaudo: `undefined` isn't  flow type. It should either be `typeof undefined` or `void`.

packages/babel-plugin-transform-typescript/src/enum.js
```diff
@@ -88,21 +90,27 @@ function enumFill(path, t, id) {
  *     Z = X | Y,
  *   }
  */
-type PreviousEnumMembers = { [name: string]: number | typeof undefined };
+type PreviousEnumMembers = { [name: string]: number | string | undefined };
```
nicolo-ribaudo: Can this actually be`undefined`? Looking at the `translateEnumValues` it seems to me that values are only set when they are a number.

packages/babel-plugin-transform-typescript/src/enum.js
```diff
@@ -88,21 +90,27 @@ function enumFill(path, t, id) {
...
       const constValue = evaluate(initializer, seen);
       if (constValue !== undefined) {
-        value = t.numericLiteral(constValue);
-        prev = constValue;
+        seen[name] = constValue;
+        if (typeof constValue === "number") {
+          value = t.numericLiteral(constValue);
+          prev = constValue;
+        } else {
+          value = t.stringLiteral(constValue);
```
nicolo-ribaudo: Can you add an assertion here? Something like
```js
else if (typeof constValue === "string") {
  value = t.stringLiteral(constValue);
  prev = undefined;
} else {
  throw path.buildCodeFrameError("Invalid enum value");
}
```

To disallow things like
```ts
enum Enum {
  baz = true
}
```

andy-ms: For `baz = true` we will simply not have a `constValue` -- it won't be set to anything other than a number or string. Still, added `assert(typeof constValue === "string");`.

DanielRosenwasser:
Looks like the test needs to be run again, but otherwise seems good to go.

existentialism:
@DanielRosenwasser restarted!

PR #7164, opened 2018-01-06
Fixes React isCompatTag validator accepting leading dash character

### LINKED ISSUE #7163: Reac isCompatTag validator accepts leading dash '-'
Choose one: is this a bug report or feature request? Bug report

The React `isCompatTag` accepts a leading dash `'-'` in tag names. This means it will accept JSX tag names like `<-div>`.

### Input Code

Adding a test suite for the React `isCompatTag` validator:

```js
  suite("isCompatTag", function() {
    it("rejects leading dash '-'", function() {
      assert(t.react.isCompatTag("-div") === false);
      assert(t.react.isCompatTag("-span") === false);
      assert(t.react.isCompatTag("-a") === false);
      assert(t.react.isCompatTag("-input") === false);
    });
  });
```

Produces this error:

```
  7752 passing (2m)
  94 pending
  1 failing

  1)  rejects leading dash '-':

      AssertionError [ERR_ASSERTION]: false == true
      + expected - actual

      -false
      +true

      at Context.<anonymous> (packages/babel-types/test/validators.js:58:7)
```

N/A, reproduced in unit tests

### Expected Behavior

Tag names with leading dash characters `'-'` are not valid

### Current Behavior

Tag names with leading dash characters `'-'` are accepted

### Possible Solution

Fix this regexp:

https://github.com/babel/babel/blob/master/packages/babel-types/src/validators/react/isCompatTag.js#L3

### Context

Spotted in @hzoo's tweet: https://twitter.com/left_pad/status/949301094903099393

### Your Environment

| ---------------- | -------
| Babel            |  
| Babylon          | 
| node             | 
| npm              | 
| Operating System |

claudiopro:
| Fixed Issues?            | Fixes #7163

Fixed regexp test in `isCompatTag` React validator allowing leading dash `'-'` characters.

Was:
```
/^[a-z]|-/.test(tagName)
```

Is:
```
// Must start with a lowercase ASCII letter
/^[a-z]/.test(tagName)
```

### DIFF
--- a/packages/babel-types/src/validators/react/isCompatTag.js
+++ b/packages/babel-types/src/validators/react/isCompatTag.js
@@ -1,4 +1,5 @@
 // @flow
 export default function isCompatTag(tagName?: string): boolean {
-  return !!tagName && /^[a-z]|-/.test(tagName);
+  // Must start with a lowercase ASCII letter
+  return !!tagName && /^[a-z]/.test(tagName);
 }
--- a/packages/babel-types/test/validators.js
+++ b/packages/babel-types/test/validators.js
@@ -32,6 +32,52 @@ suite("validators", function() {
     });
   });
 
+  suite("isCompatTag", function() {
+    it("should handle lowercase tag names", function() {
+      assert(t.react.isCompatTag("div"));
+      assert(t.react.isCompatTag("a")); // one letter
+      assert(t.react.isCompatTag("h3")); // letters and numbers
+    });
+
+    it("should handle custom element tag names", function() {
+      assert(t.react.isCompatTag("plastic-button")); // ascii letters
+      assert(t.react.isCompatTag("math-α")); // non-latin chars
+      assert(t.react.isCompatTag("img-viewer2")); // numbers
+      assert(t.react.isCompatTag("emotion-😍")); // emoji
+    });
+
+    it("accepts trailing dash '-' in custom element tag names", function() {
+      assert(t.react.isCompatTag("div-"));
+      assert(t.react.isCompatTag("a-"));
+      assert(t.react.isCompatTag("h3-"));
+    });
+
+    it("rejects empty or null tag names", function() {
+      assert(t.react.isCompatTag(null) === false);
+      assert(t.react.isCompatTag() === false);
+      assert(t.react.isCompatTag(undefined) === false);
+      assert(t.react.isCompatTag("") === false);
+    });
+
+    it("rejects tag names starting with an uppercase letter", function() {
+      assert(t.react.isCompatTag("Div") === false);
+      assert(t.react.isCompatTag("A") === false);
+      assert(t.react.isCompatTag("H3") === false);
+    });
+
+    it("rejects all uppercase tag names", function() {
+      assert(t.react.isCompatTag("DIV") === false);
+      assert(t.react.isCompatTag("A") === false);
+      assert(t.react.isCompatTag("H3") === false);
+    });
+
+    it("rejects leading dash '-'", function() {
+      assert(t.react.isCompatTag("-div") === false);
+      assert(t.react.isCompatTag("-a") === false);
+      assert(t.react.isCompatTag("-h3") === false);
+    });
+  });
+
   suite("patterns", function() {
     it("allows nested pattern structures", function() {
       const pattern = t.objectPattern([


nicolo-ribaudo:
I think that the regexp should be `/^[a-z][a-z-]*$/`, so that it handles multiple dashes. Also, can a tag name end with `-`?

nicolo-ribaudo:
Quoting the react docs: [[1](https://reactjs.org/docs/jsx-in-depth.html#user-defined-components-must-be-capitalized)]
> When an element type starts with a lowercase letter, it refers to a built-in component like `<div>` or 
 `<span>` and results in a string `'div'` or `'span'` passed to `React.createElement`. Types that start with a capital letter like `<Foo />` compile to `React.createElement(Foo)` and correspond to a component defined or imported in your JavaScript file.

So I think the correct regular expression is just `/^[a-z]/`


[1] https://reactjs.org/docs/jsx-in-depth.html#user-defined-components-must-be-capitalized

packages/babel-types/test/validators.js
```diff
@@ -32,6 +32,43 @@ suite("validators", function() {
...
+      assert(t.react.isCompatTag("Span") === false);
+      assert(t.react.isCompatTag("A") === false);
+      assert(t.react.isCompatTag("Input") === false);
+    });
+
+    it("rejects leading dash '-'", function() {
+      assert(t.react.isCompatTag("-div") === false);
+      // assert(t.react.isCompatTag("-span") === false);
+      // assert(t.react.isCompatTag("-a") === false);
+      // assert(t.react.isCompatTag("-input") === false);
```
bvaughn: Why are these commented out?

packages/babel-types/test/validators.js
```diff
@@ -32,6 +32,43 @@ suite("validators", function() {
...
+    });
+  });
```
bvaughn: Test for trailing dash might be nice. That's a use case that fails currently (but would be fixed by this PR) I think.

packages/babel-types/src/validators/react/isCompatTag.js
```diff
@@ -1,4 +1,5 @@
 // @flow
 export default function isCompatTag(tagName?: string): boolean {
-  return !!tagName && /^[a-z]|-/.test(tagName);
+  // Must start with 'a'-'z', can have a dash '-', but must finish with 'a'-'z'
+  return !!tagName && /^[a-z]+(-?[a-z]+)?$/.test(tagName);
```
bvaughn: This pattern would reject custom element names with more than one hyphen (eg `my-custom-tag`).

I think this would work though?
`/^[^-]([\-a-z]*[^-])*$/`

bvaughn: By the way, glancing at [the spec](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name) it seems like a much broader range of characters than just a-z are allowed for custom elements (eg `<emotion-😍>`) 😆 Not sure if we want to try to support them all. Probably not?

claudiopro: Yep the [custom elements spec](https://www.w3.org/TR/custom-elements/#valid-custom-element-name) says:

```
PotentialCustomElementName ::=
[a-z] (PCENChar)* '-' (PCENChar)*

PCENChar ::=
"-" | "." | [0-9] | "_" | [a-z] | #xB7 | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D] | [#x203F-#x2040] | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF] | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]
```

Do we want to get this far?

packages/babel-types/test/validators.js
claudiopro: Ha, my mistake! I forgot to uncomment before committing 😅

packages/babel-types/test/validators.js
claudiopro: Thanks for taking the time to review on a Saturday @bvaughn 😄  Let me update the PR.

packages/babel-types/src/validators/react/isCompatTag.js
bvaughn: I don't have enough context to answer that. 😄 My guess would be "probably not".

packages/babel-types/test/validators.js
bvaughn: When @hzoo tags me, I come running. (In case there's Mario Kart.)

claudiopro:
I decided to back out changes to validate the element name and only fix the validator not to accept a leading dash, and add unit tests 😄  The regex can be further complicated if needs to be to match [the spec](https://www.w3.org/TR/custom-elements/#valid-custom-element-name).

claudiopro:
@bvaughn +1 on perf concerns. Also, it seems there are other checks happening before the validator as malformed JSX tags like `<-div>` are rejected by the REPL: https://tinyurl.com/y7wfomel

nicolo-ribaudo:
They are rejected by Babylon (the parser). Anyway, I'd still accept this PR if it just remove the `-` from the test for the first character.

bvaughn:
> Also, it seems there are other checks happening before the validator as malformed JSX tags like <-div> are rejected by the REPL: https://tinyurl.com/y7wfomel

Agreed. Leading "-" fails in master now. Trailing "-" doesn't (but still won't if this change lands).

hzoo:
Yeah I didn't think this change was necessary and was never reported before (forgot to mention), just wanted to ping someone on the react team for review

Ya if it errors in the parser already, I think https://github.com/babel/babel/blob/b93184e430c5785dad197f53f73904a2a06fc1ed/packages/babylon/src/plugins/jsx/index.js#L517 then this check is more for babel to pick a string or identifier

bvaughn:
react-dom` will fail at runtime if `ownerDocument.createElement(type)` fails, so I think it's kind of up to Babel how it wants to do compile-time tag name verification.

IMO it's nice if JSX compilation fails earlier for things that are obviously wrong (like leading or trailing slashes) because it's a better developer experience.

bvaughn:
What about using `/^[^-](\-?[a-z]+)*-?$/`?

```js
function test(tag) {
  if (/^[^-](\-?[a-z]+)*-?$/.test(tag)) {
    console.log(tag);
    document.createElement(tag);
  }
}

(['a', 'ab', 'abc', 'a-b', 'a-b-c', 'a-', 'a-b-']).forEach(test);
(['-a', '-', '-a-', 'a--b']).forEach(test);

// a
// ab
// abc
// a-b
// a-b-c
// a-
// a-b-
```

nicolo-ribaudo:
Uppercase letters and numbers are also allowed. e.g. `spaN`, or `h3`.
The `isCompatTag` function should just be "Given this valid* JSX tag, should it become a string or an identifier?". To answer to this question, checking wether the first character is lowercase or not is enough.

* Babel plugins assume that the AST given to them is valid. The check against invalid ASTs should either be in Babylon or in babel-types.

claudiopro:
@nicolo-ribaudo my thoughts exactly, but I relate with @bvaughn, I'm falling prey to nerd sniping too 😅

bvaughn:
> Uppercase letters and numbers are also allowed. e.g.

In JSX, where all tag names are strings, host components (eg div, span) and
custom components (eg MyButton) are distinguished by the casing of the
first letter.

I assume the intent here is to give helpful warnings as early as possible
rather than to be 100% correct or cover all cases.

That being said, I've weighed in enough here. I have only partial contenxt
and no strong opinions :smile: so I'm bowing out now. :wave:

packages/babel-types/test/validators.js
```diff
@@ -32,6 +32,52 @@ suite("validators", function() {
...
+      assert(t.react.isCompatTag("div"));
+      assert(t.react.isCompatTag("a")); // one letter
+      assert(t.react.isCompatTag("h3")); // letters and numbers
+    });
+
+    it("should handle custom element tag names", function() {
+      assert(t.react.isCompatTag("plastic-button")); // ascii letters
+      assert(t.react.isCompatTag("math-α")); // non-latin chars
+      assert(t.react.isCompatTag("img-viewer2")); // numbers
+      assert(t.react.isCompatTag("emotion-😍")); // emoji
```
Kovensky: Not sure if anything besides `/^[a-z][a-z-]*$/` is supposed to work, but I don't see why ban it unless the JSX spec forbids it.

Kovensky: Checked the spec; the spec actually says nothing about compat tags at all; just that the identifier (`JSXIdentifier`) is basically the same thing as [`IdentifierName`](https://tc39.github.io/ecma262/#prod-IdentifierName) but it also allows `-`.

I assume that that part is enforced by babylon instead of `isCompatTag` so this seems sufficient.

nicolo-ribaudo: Emojis in source code are cool :laughing:

Kovensky:
Test failure is unrelated (yarnpkg timeout?)

nicolo-ribaudo:
Yes, it's failing often today (and yesterday). I restarted it but it failed again. I think we can marge this PR anyway since it doesn't affect Babylon?

claudiopro:
Thank you all 😊

PR #7166, opened 2018-01-07
Bail out on JSX fragments instead of throwing

### LINKED ISSUE #7165: transform-react-inline-elements throws on JSX fragments
Choose one: is this a bug report or feature request? **bug report**

When using the `@babel/transform-react-inline-elements` plugin, it throws an exception when it encounters a JSX fragment (i.e. `<>...</>`).

### Input Code

```js
export default class MyFragComp extends React.Component {
  render() {
    return <><div /><span /></>;
  }
}
```

```js
{
  // ...
  "plugins": [ "@babel/transform-react-inline-elements" ]
}
```

### Expected Behavior

Transpilation should succeed with no errors.

### Current Behavior

Transpilation fails on JSX fragments.

### Possible Solution

Quick solution: Bail out on JSX fragments (will send a PR for this now).
Better solution: Optimize fragments the same way we optimize normal JSX tags.

### Context

I use this plugin for production builds. Suddenly, builds started failing and I found out that a jsx fragment was causing it.

### Your Environment

| ---------------- | -------
| Babylon          | 
| node             | 8.7.0
| npm              | 5.4.2
| Operating System | macOS Sierra 10.12.6

mdebbar:
The `transform-react-inline-elements` plugin doesn't handle JSX fragments. It throws an exception because `node.openingElement` is undefined.

| Fixed Issues?            | Fixes #7165
| Documentation PR         | 


packages/babel-plugin-transform-react-inline-elements/src/index.js
```diff
@@ -19,7 +19,9 @@ export default function() {
 
   const visitor = helper({
     filter(node) {
-      return !hasRefOrSpread(node.openingElement.attributes);
+      return (
+        node.openingElement && !hasRefOrSpread(node.openingElement.attributes)
```
jorrit: Maybe explain in a comment in which situation `node.openingElement` is not set.

loganfsmyth:
Thanks!

PR #7169, opened 2018-01-07
renamed test files actual->input | expected->output

### LINKED ISSUE #7063: Test fixtures: rename actual/expected to input/output
Ref https://github.com/babel/babel/pull/7059#issuecomment-352766948

`actual.js` and `expected.js` isn't as intuitive as `input.js` and `output.js`. If someone wants to change this, that would be great.

Going to have to change the references (more than just what I linked) in at least these 2 packages:

https://github.com/babel/babel/blob/68476b6dddfdfc0941142cc45687c5b33c0cfe37/packages/babel-helper-fixtures/src/index.js#L87-L88

https://github.com/babel/babel/blob/68476b6dddfdfc0941142cc45687c5b33c0cfe37/packages/babel-helper-transform-fixture-test-runner/src/index.js#L134

Then all the tests need to be renamed. So this is a lot of work but relatively simple.

---

We might want to consider making https://github.com/babel-utils/babel-plugin-tester the official package or something, or merging that with our current setup eventually. Then we can move toward recommendation + standardization of quality of plugins via tests + use it as a way to upgrade plugins better. I should probably make a separate issue for this @kentcdodds

rajzshkr:
| Fixed Issues?            | Fixes #7063
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |


nicolo-ribaudo:
@rajzshkr Could you split this changes in two commits to make it possible to review it using github's online interface?
e.g.
1. Rename files
2. Changes needed to handle renamed files

rajzshkr:
@nicolo-ribaudo I changed only one file as @hzoo mentioned in the issue description. 

https://github.com/babel/babel/pull/7169/commits/69da49c3d62a3841fceda98a5288f2464330fbe6#diff-5f9c10cb2fef3e48cfb08f9a52d71332

rajzshkr:
@nicolo-ribaudo Done!!

rajzshkr:
@Kovensky Actually I wrote a script, which rename `actual.js->input.js` and `expected.js -> output.js`. But in babylon output is in form of json (not js). So the script didn't replaced the `expected.json -> output.json `, instead when I ran `make test`, It generated output.json, after making PR, I've noticed all the test cases inside babylon has both output.json and expected.json. So I deleted all the expected.json inside the test folders of babylon as a seperate commit.

nicolo-ribaudo:
Moved to https://github.com/babel/babel/pull/7181

PR #7170, opened 2018-01-07
babel-helpers: prevent object shape change in jsx

jorrit:
| Fixed Issues?            |  none
| Any Dependency Changes?  |no

Keeping the shape of an object constant is a performance improvement for modern javascript engines. At the point of the code change it is certain that the `children` property will be set later, so the property can already be set to `undefined`.


Andarist:
cc @bmeurer

bmeurer:
Good catch! Do you see concrete performance improvements somewhere?

sophiebits:
This means that the children property is added before defaultProps instead of after. Not sure how that affects things. It might be worse? I would love to figure out a better way to pull this off and share more hidden classes for React elements.

jorrit:
I encountered this code while researching a different issue and thought this was an easy opportunity for a small performance improvement. I have not tested the performance gains but they will only be there when there are no defaultProps, as it will change the shape again.

I don't see a problem with the defaultProps. The prop is compared to `void 0` which is undefined. Undefined is strictly equal to itself. The loop should execute the same. If the defaultProps is only contains a children property there will be a performance improvement and otherwise the performance will be the same. If you want me to change undefined in my change to void 0 that's fine.

This change is intended to be a small and easy improvement. If you are looking for something more fundamental that should be a separate issue in my opinion. I can understand it when you don't want to apply small changes based on theoretical performancr gains.

Andarist:
@jorrit you'd have to rebase this PR to get this merge - we've recently renamed our fixtures files, so edited `expected.js` should be named `output.js` now.

jorrit:
@Andarist done. I have changed `undefined` to `void 0` because that seems more appropriate in the context of this piece of code.

Andarist:
Merged. Thanks! 🎉

jorrit:
Thank you for merging!

PR #7179, opened 2018-01-09
Issue #6691 - Fix: unicode characters not allowed in regexes

### LINKED ISSUE #6691: Regex unicode escape flags
> Issue originally reported by @nicolo-ribaudo in babel/babylon#582

This code should **not** be considered valid:
```js
/[P QR]/\u0067
```

### Spec
https://tc39.github.io/ecma262/#prod-RegularExpressionFlags
```
RegularExpressionFlags::
    [empty]
    RegularExpressionFlags IdentifierPart 
```

https://tc39.github.io/ecma262/#sec-literals-regular-expression-literals-static-semantics-early-errors
```
11.8.5.1 Static Semantics: Early Errors
    RegularExpressionFlags :: RegularExpressionFlags IdentifierPart
        It is a Syntax Error if IdentifierPart contains a Unicode escape sequence.
```

I also looked at Annex B and Annex E, but I couldn't find anything that makes that code valid.

### 😱 
There is a test that ensures that babylon parses it correcly:
https://github.com/babel/babylon/blob/master/test/fixtures/core/uncategorised/108/actual.js

| software         | version
| ---------------- | -------
| Babylon          |  master

perinikhil:
| Fixed Issues?            | Fixes #6691 
| Major: Breaking Change?  | Not sure?


packages/babylon/src/tokenizer/index.js
```diff
@@ -827,13 +827,19 @@ export default class Tokenizer extends LocationParser {
...
+
+    const validFlags = /^[gmsiyu]*$/;
+    let mods = "";
+    while (this.state.pos < this.input.length) {
+      const char = this.input[this.state.pos];
+      const charCode = this.fullCharCodeAtPos();
+      if (validFlags.test(char)) {
+        ++this.state.pos;
+        mods += char;
+      } else if (!validFlags.test(mods) || charCode === charCodes.backslash) {
```
nicolo-ribaudo: `validFlags.test(mods)`  is always `true`, because we add a character to `mods` only if it is valid.
I think that this condition should be
```js
else if (isIdentifierChar(char)) {
```
so that it throws `Invalid regular expession flag` also for things like `/foo/w` (Instead of `Unexpected token, expected ";"`).

packages/babylon/src/tokenizer/index.js
```diff
@@ -827,13 +827,19 @@ export default class Tokenizer extends LocationParser {
...
     ++this.state.pos;
-    // Need to use `readWord1` because '\uXXXX' sequences are allowed
-    // here (don't ask).
-    const mods = this.readWord1();
-    if (mods) {
-      const validFlags = /^[gmsiyu]*$/;
-      if (!validFlags.test(mods)) {
-        this.raise(start, "Invalid regular expression flag");
+
+    const validFlags = /^[gmsiyu]*$/;
```
nicolo-ribaudo: You can remove the `*`, since you check the flags one by one.

rajzshkr:
actual/expected -> input/output in tests

packages/babylon/src/tokenizer/index.js
perinikhil: RIght, so that means `core/uncategorized/108` and `core/uncategorized/366' should also throw `Invalid regular expression flags` yeah ?

nicolo-ribaudo: Yes :+1:

nicolo-ribaudo:
Thank you!

PR #7183, opened 2018-01-09
Remove excess check for hidden files.

yavorsky:
Make `debug-fixtures` less assertive about hidden files 🙂


existentialism:
guess we wanted to make _really_ sure :)

PR #7187, opened 2018-01-09
Remove old expected.{js,json} files

nicolo-ribaudo:
I didn't staged the deleted files before committing 🤦:sweat_smile:


existentialism:
Guess Circle is just having a case of the Tuesdays?

PR #7188, opened 2018-01-09
Wrap wrapNativeSuper helpers in redefining functions for better tree-…

Andarist:
…shakeability

| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -445,46 +445,57 @@ helpers.inheritsLoose = () => template.program.ast`
 
 // Based on https://github.com/WebReflection/babel-plugin-transform-builtin-classes
 helpers.wrapNativeSuper = () => template.program.ast`
-  var _gPO = Object.getPrototypeOf || function _gPO(o) { return o.__proto__ };
-  var _sPO = Object.setPrototypeOf || function _sPO(o, p) { o.__proto__ = p; return o };
-  var _construct = (typeof Reflect === "object" && Reflect.construct) ||
-    function _construct(Parent, args, Class) {
-      var Constructor, a = [null];
-      a.push.apply(a, args);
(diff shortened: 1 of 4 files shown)


packages/babel-helpers/src/helpers.js
```diff
@@ -427,46 +427,59 @@ helpers.inheritsLoose = defineHelper(`
...
+        return _sPO(new Constructor, Class.prototype);
+      };
+    return _construct(Parent, args, Class);
+  }
 
   export default function _wrapNativeSuper(Class) {
-    if (typeof Class !== "function") {
-      throw new TypeError("Super expression must either be null or a function");
-    }
+    var _cache = typeof Map === "function" && new Map();
```
nicolo-ribaudo: Wouldn't three-shakers remove this even if it was outside from the function, since `Map` is known to be pure?

Andarist: not every tool unfortunately have sets of pure builtins in their codebases (from what I know), I've tried to get rid of this statement with UglifyJS and couldnt make it work, cc @kzc

kzc: Uglify regards `Map` as an unresolved global and won't reduce that expression. Even if it did recognize `Map` it would be conservative and assume that it can be overridden in the worst case:
```
$ echo '"use strict"; console.log(Map); Map=1; console.log(Map);' | node
[Function: Map]
1
```
Rollup, on the other hand, might be able to do it as it assumes an ES6 environment and that no builtins can be overridden.

Andarist: Then imho we should leave it in the form I've proposed.

PR #7189, opened 2018-01-10
Bug-fix: export default decorated class parsed as class expression

### LINKED ISSUE #7123: export default @decl class {} is parsed as a class expression
### Input Code

```js
export default @dec class A {}
```

https://astexplorer.net/#/gist/2688ce655edf069660a7286477acd1fe/289582be3068e25daa64d745f6fa71275ffca05c

### Expected Behavior
It should be a classs declaration

nveenjain:
| Fixed Issues?            | Fixes #7123 

Now decorators are checked just after checking for exported classes, then they are processed and then class is processed.


nicolo-ribaudo:
@nveenjain As you can see by travis-ci, the output of a fixture of a babel plugin changed. If you delete the `output.js` file and run `make test-only`, it will be updated automaticaly.

nveenjain:
@nicolo-ribaudo Fixed that. It took some time as i was struggling with rebasing.

packages/babylon/src/parser/statement.js
```diff
@@ -1368,6 +1368,10 @@ export default class StatementParser extends ExpressionParser {
       return this.parseFunction(expr, true, false, true, true);
     } else if (this.match(tt._class)) {
       return this.parseClass(expr, true, true);
+    } else if (this.match(tt.at)) {
+      this.parseDecorators(false);
+      // startNode has changed
+      return this.parseClass(this.startNode(), true, false, false, true);
```
nicolo-ribaudo: Q: Why do we need to `this.startNode()` instead of using `expr`?

nveenjain: i think expr will still point to the node which is of decorators. After parsing decorators, current node will point to class Node which is needed to be parsed.

nicolo-ribaudo: The `parseDecorator` function (called by `parseDecorators`) creates its own node, so `expr` won't be used. If you have  tried and using `expr` breaks something leave it as it is now, otherwise don't create an unnecessary node :wink:

nveenjain: Yes, it creates its own node which according to me, is required.
For example
```js
export default @dec class x{}
```
In the code above `expr` will point to the node having `start` at 15 (when expr is assigned then current node is still `tt.at`) so when `class` is parsed without creating an extra node, it's current `start` will be pointed by 15 which acc. to me should be 20. that's why we need to create extra Node, which will point to current location of class i.e. 20. I've seen that class declaration/expression nodes start at the starting of class keyword. I maybe wrong though.. :grimacing:

nveenjain: @nicolo-ribaudo, i cross-checked by using expr instead of creating a new node, and tested it for following input:-
```js
export default @dec class A {}
```
, i got this output:- 
```js
[ Node {
    type: 'ExportDefaultDeclaration',
    start: 0,
    end: 30,
    loc: SourceLocation { start: [Object], end: [Object] },
    declaration:
     Node {
       type: 'ClassDeclaration',
       start: 15,
       end: 30,
       loc: [Object],
       decorators: [Array],
       id: [Object],
       superClass: null,
       body: [Object] } } ]
```
However the ouptut must be :- 
```js
[ Node {
    type: 'ExportDefaultDeclaration',
    start: 0,
    end: 30,
    loc: SourceLocation { start: [Object], end: [Object] },
    declaration:
     Node {
       type: 'ClassDeclaration',
       start: 20,
       end: 30,
       loc: [Object],
       decorators: [Array],
       id: [Object],
       superClass: null,
       body: [Object] } } ]
```
I think we need to create new node. However, suggestions/corrections are appreciated :slightly_smiling_face:

nveenjain:
Updated the PR :smiley: . Thanks, got to learn something while doing this.

packages/babylon/src/parser/statement.js
```diff
@@ -1368,6 +1368,9 @@ export default class StatementParser extends ExpressionParser {
       return this.parseFunction(expr, true, false, true, true);
     } else if (this.match(tt._class)) {
       return this.parseClass(expr, true, true);
+    } else if (this.match(tt.at)) {
+      this.parseDecorators(false);
+      return this.parseClass(expr, true, false, false, true);
```
nicolo-ribaudo: Last thing, you passed too many parameters here. It should be `expr, true, true` like the line above.

nveenjain:
@nicolo-ribaudo , sorry for bothering, but is anything still wrong in PR?

PR #7191, opened 2018-01-10
Add eslint plugin to disallow `t.clone` and `t.cloneDeep`

nicolo-ribaudo:
Follow up to https://github.com/babel/babel/pull/7149

I made this rule autofixable.

### DIFF
--- a/scripts/eslint_rules/no-deprecated-clone.js
+++ b/scripts/eslint_rules/no-deprecated-clone.js
@@ -1,0 +1,44 @@
+"use strict";
+
+const isInIncludedFolder = filename => {
+  return /(?:packages|codemods|experimental)\/babel-/.test(filename);
+};
+
+const message =
+  "t.clone() and t.cloneDeep() are deprecated. Use t.cloneNode() instead.";
+
+module.exports = {
+  meta: {
+    schema: [],
+  },
+  fixable: "code",
+  create(context) {
+    if (!isInIncludedFolder(context.getFilename())) {
+      return {};
+    }
+
+    return {
+      CallExpression(node) {
+        const callee = node.callee;
+
+        if (callee.type !== "MemberExpression") return;
+        if (callee.computed) return;
+
+        const object = callee.object;
+        if (object.type !== "Identifier" || object.name !== "t") return;
+
+        const property = callee.property;
+        if (property.type !== "Identifier") return;
+        if (property.name !== "clone" && property.name !== "cloneDeep") return;
+
+        context.report({
+          node,
+          message,
+          fix(fixer) {
+            fixer.replaceText(property, "cloneNode");
+          },
+        });
+      },
+    };
+  },
+};
--- a/scripts/eslint_rules/no-deprecated-clone.js
+++ b/scripts/eslint_rules/no-deprecated-clone.js
@@ -6,84 +6,173 @@ function getVariableDefinition(name /*: string */, scope /*: Scope */) {
   let currentScope = scope;
   do {
     const variable = currentScope.set.get(name);
-    if (variable) return variable.defs[0];
+    if (variable) return { scope: currentScope, definition: variable.defs[0] };
   } while ((currentScope = currentScope.upper));
 }
 
-function isImportedFromTypes(
-  context /*: Context */,
+/*::
+type ReferenceOriginImport = { kind: "import", source: string, name: string };
+type ReferenceOriginParam = {
+  kind: "export param",
+  exportName: string,
+  index: number,
+};
+
+type ReferenceOrigin =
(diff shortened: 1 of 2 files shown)


scripts/eslint_rules/no-deprecated-clone.js
```diff
@@ -0,0 +1,44 @@
...
+
+    return {
+      CallExpression(node) {
+        const callee = node.callee;
+
+        if (callee.type !== "MemberExpression") return;
+        if (callee.computed) return;
+
+        const object = callee.object;
+        if (object.type !== "Identifier" || object.name !== "t") return;
```
Andarist: i know that this is a convention of ours, but could be more bullet proof with some [`referencesImport`](https://github.com/babel/babel/blob/7064b298d70490b8c129e736e322e0b356e8fb24/packages/babel-traverse/src/path/introspection.js#L160) logic, but probably had to check also for  a destructured argument 🤔

Kovensky: Yeah, lots of plugins also take the `types` key from the first argument to the plugin factory

xtuc: Currently https://github.com/babel/babel/pull/7191/files#diff-a22e9508727437cec502b0b6edaaeaf3R16 ensure that it's not used outside of Babel, I think it would be good to allow it.

nicolo-ribaudo:
Now it reports an error in these cases:

```js
import * as foo from "@babel/types";

foo.clone({});
```

```js
import { clone } from "@babel/types";

clone({});
```

But not in these cases:

```js
import { clone } from "lodash";

clone({});
```
```js
import * as _ from "lodash";

_.clone({});
```
```js
function clone() {}

clone();
```

Andarist:
What about ?
```js
export default function somePlugin ({ types: t }) {
  // ...
}
```

xtuc:
yes, good point @Andarist

nicolo-ribaudo:
@Andarist Good point, I will work on it

nicolo-ribaudo:
Error:
```js
export default function({ types: t }) {
  t.clone({});
}
```

```js
export default function({ types }) {
  types.clone({});
}
```

```js
module.exports = function ({ types: t }) {
  t.clone({});
}
```

nicolo-ribaudo:
Btw, all the functions I introduced in this PR can be reused in the other eslint plugin we have (it could be a good first contribution maybe)

xtuc:
Let's hope we don't use something like:

```js
module.exports = function (opts) {
 const t = opts.types;
  t.clone({});
}
```

in our codebase (I don't think so).

nicolo-ribaudo:
I was rewriting this plugin (since I don't like how I coded it, I had to special case everything). I'm rewriting it in a way which should handle that case. Not in our repository, but I recall having seen that snippet in some third-party plugins (it's the only possible way of doing it if you want your code to work on node 4 without transpiling it).

nicolo-ribaudo:
I have rewritten the plugin, it was fun :smile: 

Here are some examples of what it can done:
```js
import * as types from "@babel/types";

types.clone(); // error, autofixable

const { clone } = types;
clone(); // error

import { cloneDeep } from "@babel/types";
cloneDeep(); // error

const cloneDeep2 = cloneDeep;
cloneDeep2(); // error
```
```js
import * as babel from "@babel/core";

babel.types.clone(); // error, autofixable

const { types: tt } = babel;
tt.clone(); // error, autofixable

const { types: { clone } } = babel;
clone(); // error

const cloneDeep = babel.types.cloneDeep;
cloneDeep() // error

t.clone(); // error, autofixable
```
```js
export default function (babel) {
  babel.types.clone(); // error, autofixable

  const t = babel.types;
  t.clone(); // error, autofixable
}
```

```js
module.exports = function ({ types: t }) {
  t.clone(); // error, autofixable
};
```

scripts/eslint_rules/no-deprecated-clone.js
```diff
@@ -10,30 +10,65 @@ function getVariableDefinition(name /*: string */, scope /*: Scope */) {
...
-  scope /*: Scope */,
-  module /*: string */,
   kind /*: "named" | "namespace" */
 ) {
-  const definition = getVariableDefinition(node.name, scope);
+  const definition = getVariableDefinition(node.name, context.getScope());
+
   if (!definition) return false;
+  if (definition.type !== "ImportBinding") return false;
+  if (definition.parent.source.value !== "@@babel/types") return false;
```
Andarist: double `@`?

xtuc: I was also wondering, I assumed it was some sort of escape.

PR #7192, opened 2018-01-10
Initial Roadmap Suggestions

hzoo:
> Rendered markdown link: https://github.com/babel/babel/blob/add-roadmap/ROADMAP.md

## I'm interested in questions/feedback and but more importantly people to help!

This isn't just to *explain* to you the work *we* are going to do but to ask you to get involved in some ideas that aren't fully thought out yet or started

In this PR we can add more points and details if anyone has questions, or we can create a new issue (if there isn't one already) for further discussion.

> Probably can add some more ideas from https://babeljs.io/blog/2016/12/07/the-state-of-babel

Looking forward to some new contributors 🙌!

EDIT: dang if I edit people's comments become "outdated" in the diff 😢 

---

What I'd like to see soon (all related to v6 to v7 support):
- v7 codemod
- versioned/translatable docs
- babel-plugin package that helps with testing/ecosystem health
- REPL that can run any npm plugin, import/export .babelrc, hook into astexplorer/codesandbox

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -23,8 +23,9 @@
 
 ### Build/publish workflow
 
-  - Guide on compiling/publishing ES2015+, .mjs, etc: https://twitter.com/philwalton/status/908082461799616512
-  - Support multi-build/folder outputs?
+- Guide on compiling/publishing ES2015+, .mjs, etc: https://twitter.com/philwalton/status/908082461799616512
+- Support multi-build/folder outputs based on ES version/browser/engines?
+- Includes making sure you can run Babel safely over node_modules [like with #6280](https://github.com/babel/babel/pull/6280). [create-react-app has an issue for this](https://github.com/facebookincubator/create-react-app/issues/3777)
 
 ### Codemods for TC39 Proposals
 
@@ -145,10 +146,12 @@ Can run these benchmarks for perf PRs, should track some over time.
 #### Better REPL
 
 - Dropdown examples/examples of syntax from github?
-- Import any package from npm
+- Import any package from npm (can give test examples for 3rd party plugins, debugging issues)
 - Run any plugin from npm
-- Create a plugin from the repl (merge into/with ASTExplorer?)
+- Create a plugin from the repl (can we merge it with ASTExplorer/codesandbox?), even publish, run from URL?
 - Import/Export a `.babelrc` file
+- Combine ^ with the ability to run the version of Babel in a PR/master.
+- Use plugin's tests in the repo as "examples" for docs.
 
 ### `babel-bot`
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -2,6 +2,10 @@
 
 > Not everything is set in stone or has an issue for it! Looking to post these to get more people involved or integrate with other projects.
 
+## Babel 7
+
+Mentioned these in [babel/notes](https://github.com/babel/notes/blob/master/2017/2017-12/dec-21.md#before-the-v7-final) and some of the [high priority items](https://github.com/babel/babel/labels/Priority%3A%20High).
+
 ## Ecosystem
 
 ### Remake `compat-table` used in preset-env
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -172,9 +172,16 @@ References: https://github.com/eslint/eslint-github-bot, https://github.com/open
 > Better onboarding/contributing guide
 - Guide to different aspects of contributing with real examples to issues/PRs
 
+### Maintainers/Sustainability
+
+- Promote Open Collective, talking with companies about office hours, sponsorship, contracting?
+- Mentoring: Google Summer of Code/Rails Girls Summer of Code were great but was hard to keep up with volunteeers and I felt like we could be doing a lot more with full time help.
+- Maybe doing local meetups on contributing, or livestreaming development/maintenance work?
+- Should focus on bringing in maintainers that will lower the burden, not increase it even if there is upfront work
+
 ---
 
-## Wishlist (might be out of scope/complex)
+## Big Wishlist (might be out of scope/complex/ecosystem)
 
 - Should Babel operate multi-file/take in a dep graph?
 - Should Babel use type info (from other things like ts/flow/runtime info)
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,171 @@
+# Babel Roadmap
+
+> Not everything is set in stone or has an issue for it! Looking to post these to get more people involved or integrate with other projects.
+
+## Ecosystem
+
+### Remake `compat-table` used in preset-env
+
+> https://github.com/kangax/compat-table could use a remake, ideally work with browser vendors on this
+    
+- There is also https://github.com/mdn/browser-compat-data
+- Also use data from test262?
+- Run tests against real browsers
+- have a data-only format
+- Need continued maintainence
+
+### Polyfill behavior
+
+> This is regarding https://github.com/babel/babel/tree/master/packages/babel-preset-env#usebuiltins-usage
+
+- Allow any substitute polyfill instead of `core-js`. You should be able to override anything (custom `Promise`, etc)
+- Make `"usage"` option the default after it is stable.
+
+### Build/publish workflow
+  - Guide on compiling/publishing ES2015+, .mjs, etc: https://twitter.com/philwalton/status/908082461799616512
+  - Support multi-build/folder outputs?
+
+### Codemods for TC39 Proposals
+
+> Lebab/others are already used to convert from ES5 -> ESNext, so incorporate it into Babel itself.
+
+- Refactor [Lebab](https://github.com/lebab/lebab/issues/138) as Babel transforms (can keep the cli since it's a separate tool)
+  - Usecase: ES3 -> ES6+ (on source code)
+  - Usecase: Remove usage of dropped proposals
+  - Usecase: Auto upgrade to the latest version of a proposal spec (if possible)
+  - Can we somehow combine forces in: babel-codemod/jscodeshift/lebab, prettier/recast/babel-generator? I really don't want to update all of these: new syntax equals re-writing the printer in all of these places separately/out of sync.
+
+## Increasing the quality of community plugins
+- Work with the community to create guides on how to write plugins or understand ASTs, etc
+- Analysis of API's/syntax used (Google BigQuery)
+- Have #blessed/sanctioned/curated packages according to some standard
+  - Can use for smoke tests
+  - Official testing package
+  - Certain level of coverage, downloads, etc
+  - Create a scoped namespace on github/npm for these? like webpack-contrib
+  - Can enforce linting rules on apis?
+  - Makes ecosystem changes easier if can notify and upgrade these plugins beforehand
+  - Create a set of standard tests to verify against (handles all syntax)
+  - Documented/tested set of options
+
+### ASTExplorer
+- allow custom version of babel-standalone (same as REPL to allow per PR tests)
+- integrate with repl? (both are in react)
+- auto publish a plugin to npm?
+- create tests?
+- hook into codesandbox?
+
+---
+
+## Feature
+
+### Performance table for ES6+ (used in babel-preset-env under new option e.g `perf`)
+
+> Add a new option in `preset-env`: will continue to compile if the native support is *slower*. Can setup a threshold, may need to compare against the size difference.
+
+- Use [six-speed](https://github.com/kpdecker/six-speed) repo as a base, needs to apply for ES6 and proposals
+- Need continued maintainence
+
+### Compiled Output Stats 
+
+> [#5340](https://github.com/babel/babel/issues/5340) Can show the code size before and after compiling to give a sense of compiled output. Could create suggestions like using "loose" mode or not compiling, etc.
+
+- The [REPL](https://twitter.com/existentialism/status/948940160653291520) just added a before/after code-size
+- Maybe difficult to do per transform
+
+### Async Transforms
+
+Support having async plugins. Will require it to be opt-in and for all other plugins to be async.
+
+### Plugin Ordering
+
+- Add `before`/`after` keys for plugins so they can specify order.
+- Possibily implement related plugins in the same "plugin" but just expose a flag out to the end-user.
+
+### Babel Helpers (shared code)
+- loader should handle these automatically like rollup
+- allow 3rd party helpers?  https://github.com/babel/babel/issues/6487
+- allow compilation of helpers (write in esnext?) https://github.com/babel/babel/issues/5876
+
+## Repurpose `babel` / create `babel-init`
+
+> We can re-use the `babel` package for a more simplified *0 config™* version
+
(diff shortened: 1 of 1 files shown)


ROADMAP.md
```diff
@@ -0,0 +1,171 @@
...
+- Also use data from test262?
+- Run tests against real browsers
+- have a data-only format
+- Need continued maintainence
+
+### Polyfill behavior
+
+> This is regarding https://github.com/babel/babel/tree/master/packages/babel-preset-env#usebuiltins-usage
+
+- Allow any substitute polyfill instead of `core-js`. You should be able to override anything (custom `Promise`, etc)
```
Andarist: I'd like to work on this soon-ish

ROADMAP.md
```diff
@@ -0,0 +1,171 @@
...
+- Make `"usage"` option the default after it is stable.
+
+### Build/publish workflow
+  - Guide on compiling/publishing ES2015+, .mjs, etc: https://twitter.com/philwalton/status/908082461799616512
```
Andarist: And would love to discuss this one, it's the area I'm the most interested in.

ROADMAP.md
```diff
@@ -0,0 +1,171 @@
...
+  - Support multi-build/folder outputs?
```
Andarist: That's interesting and probably useful for some, but I kinda feel that i.e. `rollup` has this area covered for most of use cases - especially that it understand the dep graph, so it can just nicely output flat bundles and all that jazz.

ROADMAP.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Codemods for TC39 Proposals
+
+> Lebab/others are already used to convert from ES5 -> ESNext, so incorporate it into Babel itself.
+
+- Refactor [Lebab](https://github.com/lebab/lebab/issues/138) as Babel transforms (can keep the cli since it's a separate tool)
+  - Usecase: ES3 -> ES6+ (on source code)
+  - Usecase: Remove usage of dropped proposals
```
Andarist: wouldnt this just require something like `--write` to have your files updated in place?

ROADMAP.md
```diff
@@ -0,0 +1,171 @@
...
+  - Usecase: Auto upgrade to the latest version of a proposal spec (if possible)
+  - Can we somehow combine forces in: babel-codemod/jscodeshift/lebab, prettier/recast/babel-generator? I really don't want to update all of these: new syntax equals re-writing the printer in all of these places separately/out of sync.
```
Andarist: that would be sweeeeet, would require major cooperation between projects though

ROADMAP.md
xtuc: I agree, it's not so common in JS but CMake is a super tool for emitting multiple binaries from a monorepo-ish project. 

One issue is that npm/node doesn't support super good a flat bundle. React-dom for example has to put a lot of files in the root of the project https://unpkg.com/react-dom@16.2.0/.

ROADMAP.md
```diff
@@ -0,0 +1,171 @@
...
+
+Support having async plugins. Will require it to be opt-in and for all other plugins to be async.
+
+### Plugin Ordering
+
+- Add `before`/`after` keys for plugins so they can specify order.
+- Possibily implement related plugins in the same "plugin" but just expose a flag out to the end-user.
+
+### Babel Helpers (shared code)
+- loader should handle these automatically like rollup
```
Andarist: maybe actually the easiest way for this is just to publish helpers as a package (they are somewhat already published within `@babel/runtime`) and encourage transpiling libraries (making it default maybe?) with imports being added to that package instead of inlining them into the source code like now

ROADMAP.md
xtuc: Maybe just a standalone tool IMO

ROADMAP.md
```diff
@@ -0,0 +1,178 @@
...
+  - Create a scoped namespace on github/npm for these? like webpack-contrib
+  - Can enforce linting rules on apis?
+  - Makes ecosystem changes easier if can notify and upgrade these plugins beforehand
+  - Create a set of standard tests to verify against (handles all syntax)
+  - Documented/tested set of options
+
+### ASTExplorer
+
+- allow custom version of babel-standalone (same as REPL to allow per PR tests)
+- integrate with repl? (both are in react)
```
yavorsky: Would be good to merge:
https://github.com/babel/website/pull/1421

ROADMAP.md
Andarist: Yeah, sure - encapsulating this in other tool will be much safer for the users. I meant this just as an example and to point out that this one is probably quite a low hanging fruit, beside creating a separate CLI tool actual algorithms etc are pretty much done.

hzoo: Kind of, you still have to run Babel via recast so that it retains the previous formatting which means just run the plugin via babel-codemod (or just run prettier right after which works too). This also requires that those tools are updated to the latest syntax's which they aren't since they are not sync'd with this repo.

Also it maybe easier to write a separate transform because the output from Babel may not be what you are looking for in the source code (or you need to output the loose mode version).

ROADMAP.md
Andarist: In this regard flat bundles and multi-file directories doesnt really differ that much - if one needs multiple entry points for the package.

But actually many use cases are covered automatically - you just have to output multiple flat bundles and make use of `main`, `module`, `browser`, `react-native` fields in `package.json`.

ROADMAP.md
hzoo: I mean to say that `lebab` already is the standalone + cli tool. It's basically the same thing as babel but with the printer being recast instead of babel-generator. This is also related to the idea of combining recast/prettier/babel-generator somehow

ROADMAP.md
```diff
@@ -0,0 +1,171 @@
...
+- Libraries (`(p)react`, `redux`, `vue`)
+- Tools that use individual packages (`debugger.html`, `prepack`)
+- OSS Apps (`nylas-mail`)
+
+### Better Debuggability
+
+- Query config for data for other tools
+- `babel --settings`
+- Validate config better
+- Create/expand on new tools like https://github.com/boopathi/babel-time-travel
```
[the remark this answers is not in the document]
yavorsky: We need to add this to awesome-babel :)

ROADMAP.md
```diff
@@ -0,0 +1,178 @@
...
+
+- Dropdown examples/examples of syntax from github?
+- Import any package from npm
+- Run any plugin from npm
+- Create a plugin from the repl (merge into/with ASTExplorer?)
+- Import/Export a `.babelrc` file
+
+### `babel-bot`
+
+> [#43](https://github.com/babel/babel-bot/issues/43) Rewrite it with [probot](https://github.com/probot/probot)
```
rajzshkr: I liked to work on this,

rajzshkr:
Detailed Team page.

ROADMAP.md
hzoo: Yeah it can stay it's own repo but if it uses `@babel/core` it can stay a cli tool and can be versioned separately. cc @nene @mohebifar on this again for more thoughts?

ROADMAP.md
```diff
@@ -0,0 +1,171 @@
+# Babel Roadmap
+
+> Not everything is set in stone or has an issue for it! Looking to post these to get more people involved or integrate with other projects.
+
+## Ecosystem
+
+### Remake `compat-table` used in preset-env
+
+> https://github.com/kangax/compat-table could use a remake, ideally work with browser vendors on this
```
addyosmani: Is there specific help that would be useful for browser vendors to assist with here? I'd love to see if there's anything we can support to make preset-env + the compat tables easier to rely on over time.

hzoo: I know there is https://github.com/mdn/browser-compat-data which has a javascript section but haven't looked into it that much? I guess it would be nice to be able to query some API from the browser's own status pages to generate data instead of waiting for someone to make a PR? Or if we had a service that ran the test262 tests against browsers we could do it too but that would be a pretty big effort? And maybe we don't need to remake it just change the process of submitting/extraction?

addyosmani: Thanks for those details. Vendors maintain their own status pages for features (e.g ChromeStatus) which might get updated more often than the different compat tables. I like the idea of periodically pulling from there in a more automated way, trying to get the MDN tables kept up to date or explore if test262 might be a better path forward. Will have a chat with my team.

cc @nolanlawson @apple-web-evangelist @AdaRoseCannon in case you have ideas too.

ROADMAP.md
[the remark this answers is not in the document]
rajzshkr: @yavorsky Added in awesome-babel https://github.com/babel/awesome-babel/pull/22

ROADMAP.md
```diff
@@ -0,0 +1,185 @@
...
+### Polyfill behavior
+
+> This is regarding https://github.com/babel/babel/tree/master/packages/babel-preset-env#usebuiltins-usage
+
+- Allow any substitute polyfill instead of `core-js`. You should be able to override anything (custom `Promise`, etc)
+- Make `"usage"` option the default after it is stable.
+
+### Build/publish workflow
+
+- Guide on compiling/publishing ES2015+, .mjs, etc: https://twitter.com/philwalton/status/908082461799616512
```
jdalton: Would this guide include a you might not need section. With Node having 99% ES2015+ it might be nice thing to avoid unneeded build steps. Maybe something in `@babel/env`.

ROADMAP.md
```diff
@@ -0,0 +1,185 @@
...
+- create tests?
+- hook into codesandbox?
+
+---
+
+## Feature
+
+### Performance table for ES6+ (used in babel-preset-env under new option e.g `perf`)
+
+> Add a new option in `preset-env`: will continue to compile if the native support is *slower*. Can setup a threshold, may need to compare against the size difference.
```
taion: How will this interact with `loose` or `spec`?

`loose` is pretty common in the wild: https://github.com/facebook/react/blob/145986ec17cbd136849d8c9ef1157830233e00f0/.babelrc#L12

taion: (Granted, that's not using env, but people do use env mode)

ROADMAP.md
```diff
@@ -0,0 +1,185 @@
...
+
+---
+
+## Wishlist (might be out of scope/complex)
+
+- Should Babel operate multi-file/take in a dep graph?
+- Should Babel use type info (from other things like ts/flow/runtime info)
+- AST
+  - Can we just move back to acorn + estree?
+  - Should we switch to shift?
```
thejameskyle: I spent some time documenting what Shift + JSX/Flow/TypeScript would look like https://github.com/thejameskyle/shift-extensions-spec/blob/master/extensions-spec.idl

ROADMAP.md
IgorMinar: I'd love to see a strong push to make the ES Module output the default via presents as well as documentation. Most developers don't know to opt-in to it, which causes lots of deoptimizations in webpack builds and significantly larger payload sizes.

ROADMAP.md
```diff
@@ -0,0 +1,185 @@
...
+
+### `babel-bot`
+
+> [#43](https://github.com/babel/babel-bot/issues/43) Rewrite it with [probot](https://github.com/probot/probot)
+
+A bot is really useful to do github/maintainer activities automatically.
+
+We haven't updated this in a long time due to Andrew being busy and not setting up the automatic infra on AWS. Switching will make updating actually real so we can add some new features which would save some headache.
+
+References: https://github.com/eslint/eslint-github-bot, https://github.com/open-bot/open-bot
```
orta: [Peril](https://github.com/danger/peril) is also an option we've deprecated a bunch of CocoaPods/Fastlane micro-servers to just a few lines of JS

ROADMAP.md
```diff
@@ -0,0 +1,192 @@
...
+- The [REPL](https://twitter.com/existentialism/status/948940160653291520) just added a before/after code-size
+- Maybe difficult to do per transform
+
+### Async Transforms
+
+Support having async plugins. Will require it to be opt-in and for all other plugins to be async.
+
+### Plugin Ordering
+
+- Add `before`/`after` keys for plugins so they can specify order.
```
stefanpenner: Encoding before/after within a plugin is insufficient, as a user can compose two plugins entirely unaware of each other. In addition to a plugin being able to specific it's own dependencies, it is inevitable that the user will also need to modify/augment those dependencies.

For example:

A user may have two plugins unaware of each other, but with an implicit order dependency. 

To address this, the user must also have control over the order dependency. One solution is to allow the user to impose an additional edge, to inform the DAG's topsort output.

I have used https://github.com/krisselden/dag-map to solve this problem in the past, and it worked like a charm.

cc @rwjblue  / @krisselden

stefanpenner: More Thoughts on this:

We may be able to identify some common milestones during the transpilation process, these Milestones can serve as mechanism of grouping plugins together without tightly coupling them.

For example, one may want X category of work to always occur before Y category of work.

Specifically, one typically wants to strip assertions/feature flags, before module transpilation, and never after. This should continue to work regardless of the module transpilation strategy  be it CJS/AMD/magic bundler or even disabled entirely.

In this case, the feature-flag/assert stripping plugin, shouldn't need to be aware of ALL module transpilation plugins, rather it should be able to simply say. I should be able to add an edge, informing the DAG that it must be run before the module transpilation miletones.

hzoo: Sorry I didn't link to the older prs maybe but I was suggesting allowing a plugin to opt into a "capability" so it wouldn't be hardcoded to a plugin name - like for modules or anything else, I believe that's the same thought you had with a category

hzoo: Sorry I didn't link to the older prs maybe but I was suggesting allowing a plugin to opt into a "capability" so it wouldn't be hardcoded to a plugin name - like for modules or anything else, I believe that's the same thought you had with a category

tbranyen:
Not sure if this feedback is useful, but I'm working on a tool that consumes `.babelrc`, gets the AST, and then further modifies the AST using bablyon. There isn't a good way for this to happen. Looking at tools like parcel, which have the exact same aim, I think it'd be great to have babel and babylon take an AST (optionally a map), and then allow further modifications (since both tools already output code/map).

Andarist:
@tbranyen babylon is a parser, it takes your code and outputs the AST, did you mean that you want to use babel (not babylon) to modify the AST further? Or did I not understand something correctly?

tbranyen:
@Andarist You are understanding correctly. I think this is an issue for babel-generator. I'll ping ya on Twitter to take it off this chat.

hzoo:
I'm going to move this to the website repo, and then link to issues if we create them. https://github.com/babel/website/pull/1657

ROADMAP.md
jedwards1211: @stefanpenner maybe babel could at least log warnings for plugins that know they're in the wrong order relative to each other though?  For instance it sounds doubtful that there will ever be a valid use case for putting `proposal-decorators` after `proposal-class-properties`, so a warning there would help a lot of people, especially since it's not at all obvious what's wrong from the errors that can result.

PR #7201, opened 2018-01-11
babel --settings || babel -S implementation

### LINKED ISSUE #2960: feature request: `babel --settings` for bug reports
Having a command line option to dump all the presets and plugins (with versions) being used by babel would be useful. 

This would help to avoid the bug report back and forth: "provide your .babelrc or what plugins/presets you are using", and would be less error prone. Also would be a good way to verify your own babel set up.

The format of the settings dump would be canonical and unambiguous whether .babelrc or command line flags are used. Perhaps some basic sanity checking could be done for the presets and plugins specified as well.

rajasekarm:
| Fixed Issues?            | Fixes #2960
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Tests Added + Pass?      | 
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-cli/src/babel/index.js
+++ b/packages/babel-cli/src/babel/index.js
@@ -8,6 +8,7 @@ import glob from "glob";
 
 import dirCommand from "./dir";
 import fileCommand from "./file";
+import { printSettings } from "./util";
 
 import pkg from "../../package.json";
 
@@ -168,6 +169,9 @@ commander.option(
   "--delete-dir-on-start",
   "Delete the out directory before compilation",
 );
+
+commander.option("-S, --settings", "Show configuration settings");
+
 /* eslint-enable max-len */
 
 commander.version(pkg.version + " (@babel/core " + version + ")");
@@ -221,6 +225,11 @@ if (commander.deleteDirOnStart && !commander.outDir) {
   errors.push("--delete-dir-on-start requires --out-dir");
 }
 
+if (commander.settings) {
+  printSettings();
+  process.exit(2);
+}
+
 if (errors.length) {
   console.error(errors.join(". "));
   process.exit(2);
@@ -248,6 +257,7 @@ delete opts.configFile;
 delete opts.deleteDirOnStart;
 delete opts.keepFileExtension;
 delete opts.relative;
+delete opts.settings;
 
 // Commander will default the "--no-" arguments to true, but we want to leave them undefined so that
 // @babel/core can handle the default-assignment logic on its own.
@@ -256,4 +266,7 @@ if (opts.comments === true) opts.comments = undefined;
 if (opts.highlightCode === true) opts.highlightCode = undefined;
 
 const fn = commander.outDir ? dirCommand : fileCommand;
+
+console.log(fn, commander, filenames, opts);
+
 fn(commander, filenames, opts);
--- a/packages/babel-cli/src/babel/index.js
+++ b/packages/babel-cli/src/babel/index.js
@@ -227,7 +227,7 @@ if (commander.deleteDirOnStart && !commander.outDir) {
 
 if (commander.settings) {
   printSettings();
-  process.exit(2);
+  process.exit();
 }
 
 if (errors.length) {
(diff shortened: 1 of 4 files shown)


rajzshkr:
```
-------Babel Settings--------
Node version:
 8.9.1
Npm version:
 5.6.0
Yarn version:
 1.3.2
-------Babel packages--------
├─┬ @babel/cli@7.0.0-beta.37
├─┬ @babel/core@7.0.0-beta.37
├─┬ @babel/preset-env@7.0.0-beta.37
├─┬ @babel/preset-flow@7.0.0-beta.37
├─┬ @babel/preset-stage-0@7.0.0-beta.37
├─┬ @babel/register@7.0.0-beta.37
├─┬ babel-eslint@8.0.1
├─┬ babel-loader@8.0.0-beta.0
├─┬ babel-plugin-istanbul@4.1.5
├── eslint-config-babel@7.0.2
├─┬ gulp-babel@8.0.0-beta.0
-------Babel presets--------
-------Babel plugins--------
```

packages/babel-cli/src/babel/util.js
```diff
@@ -111,3 +112,34 @@ export function adjustRelative(relative, keepFileExtension) {
   }
   return relative.replace(/\.(\w*?)$/, "") + ".js";
 }
+
+export function printSettings() {
+  const hasYarn = cwd =>
```
Andarist: should babel be responsible for checking yarn / npm / node versions? Seems a little bit odd, would ease copying "complete" environment setup to the issues though, so i dont know 🤷

packages/babel-cli/src/babel/util.js
```diff
@@ -111,3 +112,34 @@ export function adjustRelative(relative, keepFileExtension) {
...
+  log.push("-------Babel Settings--------");
+  log.push(`Node version: ${indent(process.versions.node)}`);
+  log.push(`Npm version: ${indent(child.execSync("npm -v").toString())}`);
+
+  if (hasYarn()) {
+    log.push(`Yarn version: ${indent(child.execSync("yarn -v").toString())}`);
+  }
+  log.push("-------Babel packages--------");
+  const packages = child
+    .execSync("npm list --silent | grep babel")
```
Andarist: that will catch also not-official babel packages, but actually I think that its fine

xtuc: We talked off but relevant here: I was wondering if npm could be called programmatically? Spawning a new process has some disadvantages like the one you mentioned.

xtuc: + ouput formating looks a bit strange to me.

packages/babel-cli/src/babel/index.js
```diff
@@ -168,6 +169,9 @@ commander.option(
   "--delete-dir-on-start",
   "Delete the out directory before compilation",
 );
+
+commander.option("-S, --settings", "Show configuration settings");
```
xtuc: I'm not sure we should add a shorthand for this. S could be something else in the future. 

Note that this could also replace Babel --version?

packages/babel-cli/src/babel/index.js
```diff
@@ -221,6 +225,11 @@ if (commander.deleteDirOnStart && !commander.outDir) {
   errors.push("--delete-dir-on-start requires --out-dir");
 }
 
+if (commander.settings) {
+  printSettings();
+  process.exit(2);
```
xtuc: Why exit code 2, do we have a manual specifying this? Or conventional?

packages/babel-cli/src/babel/index.js
Andarist: I'd prefer to keep `--version`, it still might be useful.

xtuc: Sorry I meant to say merging them not replacing.

xtuc: That was my point :)

Andarist: I had to misunderstand your intent then 😅

packages/babel-cli/src/babel/util.js
rajzshkr: we can always change the output format.

packages/babel-cli/src/babel/index.js
rajzshkr: it should be empty,  

> Or conventional?

yes, https://nodejs.org/api/process.html#process_exit_codes

PR #7205, opened 2018-01-12
[babel-generator] Add check for undefined endToken in whitespace

MatthieuLemoine:

Hi ! 👋 

In some weird cases, in the following method in ```src/whitespace.js``` : 

```javascript
  getNewlinesAfter(node) {
    let startToken;
    let endToken;
    const tokens = this.tokens;

    let index = this._findToken((token) => token.end - node.end, 0, tokens.length);
    if (index >= 0) {
      while (index && node.end === tokens[index - 1].end) --index;
      startToken = tokens[index];
      endToken = tokens[index + 1];
     // endToken can be undefined if startToken is EOF
      if (endToken.type.label === ",") endToken = tokens[index + 2];
    }

    if (endToken && endToken.type.label === "eof") {
      return 1;
    } else {
      return this._getNewlinesBetween(startToken, endToken);
    }
  }
```

`startToken` can be an `EOF` which leads to `endToken` being `undefined` and the following error : 

```
error: bundling failed: TypeError: Cannot read property 'type' of undefined
    at Whitespace.getNewlinesAfter (/home/user/app/node_modules/babel-generator/lib/whitespace.js:59:20)
    at Generator._printNewline (/home/user/app/node_modules/babel-generator/lib/printer.js:459:34)
    at Generator.printJoin (/home/user/app/node_modules/babel-generator/lib/printer.js:376:32)
    at Generator.printList (/home/user/app/node_modules/babel-generator/lib/printer.js:430:17)
    at Generator.ObjectExpression (/home/user/app/node_modules/babel-generator/lib/generators/types.js:57:10)
    at /home/user/app/node_modules/babel-generator/lib/printer.js:298:23
    at Buffer.withSource (/home/user/app/node_modules/babel-generator/lib/buffer.js:168:5)
    at Generator.withSource (/home/user/app/node_modules/babel-generator/lib/printer.js:189:15)
    at Generator.print (/home/user/app/node_modules/babel-generator/lib/printer.js:297:10)
```

To avoid this issue, this PR adds an undefined check.

You can look into this issue for reference : https://github.com/detrohutt/babel-plugin-inline-import-graphql-ast/issues/2.

You can find a small [reproduction](https://github.com/real34/tmp-babel-inline-gql-issue4) and a [video](https://www.dropbox.com/s/dwcce8skh5eafus/screencast_00000.mp4?dl=0
) showing this weird behavior.

### DIFF
--- a/packages/babel-generator/src/whitespace.js
+++ b/packages/babel-generator/src/whitespace.js
@@ -41,7 +41,7 @@ export default class Whitespace {
       while (index && node.end === tokens[index - 1].end) --index;
       startToken = tokens[index];
       endToken = tokens[index + 1];
-      if (endToken.type.label === ",") endToken = tokens[index + 2];
+      if (endToken && endToken.type.label === ",") endToken = tokens[index + 2];
     }
 
     if (endToken && endToken.type.label === "eof") {
--- a/packages/babel-generator/test/index.js
+++ b/packages/babel-generator/test/index.js
@@ -291,6 +291,14 @@ describe("whitespace", function () {
     const w = new Whitespace([]);
     assert.equal(w.getNewlinesBefore(t.stringLiteral("1")), 0);
   });
+  it("handles getNewlinesAfter EOF node", function () {
+    const w = new Whitespace([{
+      type : { label : "eof" },
+      start: 446,
+      end: 446,
+    }]);
+    assert.equal(w.getNewlinesAfter({ end : 446 }), 0);
+  });
 });
 
 const suites = fixtures(`${__dirname}/fixtures`);


rajzshkr:
Can you add test case for the same?

MatthieuLemoine:
@rajzshkr Yes but I don't know what kind of tests I can add as there's no unit test for this method and I don't have a deep understanding of babel-generator's internals.

rajzshkr:
check this test case, https://github.com/babel/babel/blob/master/packages/babel-generator/test/fixtures/auto-indentation/hard-tab. Make your failing code in input.js and make sure you get the requires output.

nicolo-ribaudo:
@MatthieuLemoine Does this bug also affect Babel 7? If yes, please make your PR target that branch. In case it has alread been fixed, we probably won't release a new Babel 6 version.

MatthieuLemoine:
@nicolo-ribaudo I don't think that this bug affects Babel 7 as `getNewlinesAfter` has been removed from the codebase.

MatthieuLemoine:
Upgrading to Babel 7 is not straightforward as a lot of plugins has not been updated to work with Babel 7. So I hope that a new bug fix release will be made.

MatthieuLemoine:
Test case added.

MatthieuLemoine:
@nicolo-ribaudo As you can see, a lot of people is experiencing this issue.

This PR only adds a missing undefined check. It looks like an oversight as there's already an undefined check two lines below : 

```javascript
if (endToken && endToken.type.label === "eof")
```

If Babel 6 is still officially supported, a bug fix release should be possible.

gihrig:
I have been running into this issue on a daily basis - very annoying!

A Babel 6 release would be awesome! ✨

Andarist:
@MatthieuLemoine if we can make CircleCI to pass (I've restarted the build now to see if that helps) - then we should be able to prepare a bug fix release.

hzoo:
Not going to be doing this often as the last 6.x release was 6 months ago, and probably won't be doing it at all once 7.x is out unless it's sponsored. We have been working on 7.x for probably a year now and would like to focus on releasing that and fixing any issues people have with upgrading that instead of backporting forever, especially with the limited time people have. No one is really funding the development, especially for something like a backport which is a time consuming thing to do for a volunteer team.

https://medium.com/@mikeal/stop-supporting-old-releases-70cfa0e04b0c

I am going to update the readme's so we can finally switch our docs over and provide an upgrade experience for 7.x while allowing for switching to the 6.x docs https://github.com/babel/website/issues/1544 so might as well add this in since it's unlikely this will cause any issues unlike a new feature request or port on 6.x

klujanrosas:
I'm sorry I must be doing something wrong but I can't get 
`yarn add babel-core@6.26.1` to work, it shows me a list to pick a version but 6.26.1 is missing there

PR #7213, opened 2018-01-14
Fixed _containerInsertAfter setting path key as stringified index

### LINKED ISSUE #7178: babel-transform-react-inline-elements broken since beta35 due to #6984
bug: when transpiling the following JSX code in beta35 the generated code contains untranspiled JSX.

This happens only when @babel/transform-react-inline-elements is present and I think #6984 broke it.

The code is based on [kriasoft/react-starter-kit](https://github.com/kriasoft/react-starter-kit).

### Input Code

```javascript
import React from 'react';
import Layout from '../../components/Layout';
import Contact from './Contact';
import graphqlclient from '../../core/graphqlclient';

const title = 'Neem contact op';

async function action({ data, req }) {
  const { isloaded } = data;
  let { user } = data;

  if (!isloaded) {
    const newdata = (await graphqlclient(`
      query ContactPage {
        me { displayname }
      }`, null, req));
    user = newdata.me;
  }

  return {
    title,
    chunks: ['contact'],
    component: <Layout user={user}><Contact title={title} /></Layout>,
    data: { user, isloaded: true },
  };
}

export default action;
```

```javascript
          presets: [
            [
              '@babel/preset-env',
              {
                targets: {
                  browsers: [">1%", "last 4 versions", "Firefox ESR", "not ie < 9"],
                  forceAllTransforms: true,
                },
                modules: false,
                useBuiltIns: false,
                debug: false,
              },
            ],
            '@babel/preset-stage-2',
            ['@babel/preset-react'],
          ],
          plugins: [
            '@babel/transform-react-constant-elements',
            '@babel/transform-react-inline-elements',
            'transform-react-remove-prop-types',
          ],
```

### Expected Behavior

In beta34 this was the output:

```javascript
var action =
/*#__PURE__*/
function () {
  var _ref2 = _asyncToGenerator(
  /*#__PURE__*/
  regeneratorRuntime.mark(function _callee(_ref) {
    var data, req, isloaded, user, newdata;
    return regeneratorRuntime.wrap(function _callee$(_context) {
      while (1) {
          case 0:
            data = _ref.data, req = _ref.req;
            isloaded = data.isloaded;
            user = data.user;

            if (isloaded) {
              _context.next = 8;
              break;
            }

            _context.next = 6;
            return graphqlclient("\n      query ContactPage {\n        me { displayname }\n      }", null, req);

          case 6:
            newdata = _context.sent;
            user = newdata.me;

          case 8:
            return _context.abrupt("return", {
              title: title,
              chunks: ['contact'],
              component: _jsx(Layout, {
                user: user
              }, void 0, _ref3),
              data: {
                user: user,
                isloaded: true
              }
            });

          case 9:
          case "end":
            return _context.stop();
        }
      }
    }, _callee, this);
  }));

  return function action(_x) {
    return _ref2.apply(this, arguments);
  };
}();

var REACT_ELEMENT_TYPE;

function _jsx(type, props, key, children) { if (!REACT_ELEMENT_TYPE) { REACT_ELEMENT_TYPE = typeof Symbol === "function" && Symbol.for && Symbol.for("react.element") || 0xeac7; } var defaultProps = type && type.defaultProps; var childrenLength = arguments.length - 3; if (!props && childrenLength !== 0) { props = {}; } if (props && defaultProps) { for (var propName in defaultProps) { if (props[propName] === void 0) { props[propName] = defaultProps[propName]; } } } else if (!props) { props = defaultProps || {}; } if (childrenLength === 1) { props.children = children; } else if (childrenLength > 1) { var childArray = new Array(childrenLength); for (var i = 0; i < childrenLength; i++) { childArray[i] = arguments[i + 3]; } props.children = childArray; } return { $$typeof: REACT_ELEMENT_TYPE, type: type, key: key === undefined ? null : '' + key, ref: null, props: props, _owner: null }; }

function _asyncToGenerator(fn) { return function () { var self = this, args = arguments; return new Promise(function (resolve, reject) { var gen = fn.apply(self, args); function step(key, arg) { try { var info = gen[key](arg); var value = info.value; } catch (error) { reject(error); return; } if (info.done) { resolve(value); } else { Promise.resolve(value).then(_next, _throw); } } function _next(value) { step("next", value); } function _throw(err) { step("throw", err); } _next(); }); }; }

import React from 'react';
import Layout from '../../components/Layout';
import Contact from './Contact';
import graphqlclient from '../../core/graphqlclient';
var title = 'Neem contact op';

var _ref3 = _jsx(Contact, {
  title: title
});

export default action;
```

### Current Behavior

In beta 35 this is the output:
Note that there is still JSX in this code.

```javascript
var REACT_ELEMENT_TYPE;

function _jsx(type, props, key, children) { if (!REACT_ELEMENT_TYPE) { REACT_ELEMENT_TYPE = typeof Symbol === "function" && Symbol.for && Symbol.for("react.element") || 0xeac7; } var defaultProps = type && type.defaultProps; var childrenLength = arguments.length - 3; if (!props && childrenLength !== 0) { props = {}; } if (props && defaultProps) { for (var propName in defaultProps) { if (props[propName] === void 0) { props[propName] = defaultProps[propName]; } } } else if (!props) { props = defaultProps || {}; } if (childrenLength === 1) { props.children = children; } else if (childrenLength > 1) { var childArray = new Array(childrenLength); for (var i = 0; i < childrenLength; i++) { childArray[i] = arguments[i + 3]; } props.children = childArray; } return { $$typeof: REACT_ELEMENT_TYPE, type: type, key: key === undefined ? null : '' + key, ref: null, props: props, _owner: null }; }

function _asyncToGenerator(fn) { return function () { var self = this, args = arguments; return new Promise(function (resolve, reject) { var gen = fn.apply(self, args); function step(key, arg) { try { var info = gen[key](arg); var value = info.value; } catch (error) { reject(error); return; } if (info.done) { resolve(value); } else { Promise.resolve(value).then(_next, _throw); } } function _next(value) { step("next", value); } function _throw(err) { step("throw", err); } _next(); }); }; }

import React from 'react';
import Layout from '../../components/Layout';
import Contact from './Contact';
import graphqlclient from '../../core/graphqlclient';
var title = 'Neem contact op';

function action(_x) {
  return _action.apply(this, arguments);
}

var _ref2 = <Contact title={title} />;

function _action() {
  _action = _asyncToGenerator(
  /*#__PURE__*/
  regeneratorRuntime.mark(function _callee(_ref) {
    var data, req, isloaded, user, newdata;
    return regeneratorRuntime.wrap(function _callee$(_context) {
      while (1) {
          case 0:
            data = _ref.data, req = _ref.req;
            isloaded = data.isloaded;
            user = data.user;

            if (isloaded) {
              _context.next = 8;
              break;
            }

            _context.next = 6;
            return graphqlclient("\n      query ContactPage {\n        me { displayname }\n      }", null, req);

          case 6:
            newdata = _context.sent;
            user = newdata.me;

          case 8:
            return _context.abrupt("return", {
              title: title,
              chunks: ['contact'],
              component: _jsx(Layout, {
                user: user
              }, void 0, _ref2),
              data: {
                user: user,
                isloaded: true
              }
            });

          case 9:
          case "end":
            return _context.stop();
        }
      }
    }, _callee, this);
  }));
  return _action.apply(this, arguments);
}

export default action;
```

I captured both outputs by console.log-ing `code` in `transpile` in babel-loader.

### Possible Solution

No clue.

### Context

I disabled babel-transform-react-inline-elements for now.

### Your Environment

| ---------------- | -------
| Babel            |  7.0.0-beta.35
| Babylon          | 
| node             | 9.3.0
| npm              | 5.6.0
| Operating System | Windows 10 1709

Andarist:
| Fixed Issues?            | Fixes #7178
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

This caused paths being not requeued later, when something was i.e. insertedBefore a damaged path - while the insertion worked, the inserted path was not requeued and missed by subsequent traversals.

cc @jorrit @Imundy

### DIFF
--- a/packages/babel-plugin-transform-async-to-generator/test/fixtures/regression/7178/input.js
+++ b/packages/babel-plugin-transform-async-to-generator/test/fixtures/regression/7178/input.js
@@ -1,0 +1,5 @@
+const title = "Neem contact op";
+
+async function action() {
+  return <Contact title={title} />;
+}
--- a/packages/babel-plugin-transform-async-to-generator/test/fixtures/regression/7178/options.json
+++ b/packages/babel-plugin-transform-async-to-generator/test/fixtures/regression/7178/options.json
@@ -1,0 +1,8 @@
+{
+  "plugins": [
+    "external-helpers",
+    "transform-async-to-generator",
+    "transform-react-jsx",
+    "transform-react-constant-elements"
+  ]
+}
--- a/packages/babel-plugin-transform-async-to-generator/test/fixtures/regression/7178/output.js
+++ b/packages/babel-plugin-transform-async-to-generator/test/fixtures/regression/7178/output.js
@@ -1,0 +1,16 @@
+const title = "Neem contact op";
+
+function action() {
+  return _action.apply(this, arguments);
+}
+
+var _ref = React.createElement(Contact, {
+  title: title
+});
+
+function _action() {
+  _action = babelHelpers.asyncToGenerator(function* () {
+    return _ref;
+  });
+  return _action.apply(this, arguments);
+}
--- a/packages/babel-traverse/src/path/modification.js
+++ b/packages/babel-traverse/src/path/modification.js
@@ -56,7 +56,7 @@ export function _containerInsert(from, nodes) {
   this.container.splice(from, 0, ...nodes);
   for (let i = 0; i < nodes.length; i++) {
     const to = from + i;
-    const path = this.getSibling(`${to}`);
+    const path = this.getSibling(to);
     paths.push(path);
 
     if (this.context && this.context.queue) {


nicolo-ribaudo:
Can we make the `getSibling` throw if the given argument is a string which can be converted to a number? (Or automatically convert it)

jorrit:
Perhaps the text `Neem contact op` from my original code sample could be changed to English `Contact us` or some example text to keep the language consistent.

nicolo-ribaudo:
@jorrit I often type random characters, so having a string in a test in another language is ok :slightly_smiling_face:

Andarist:
> Can we make the getSibling throw if the given argument is a string which can be converted to a number? (Or automatically convert it)

Thought about adding those, but Im a little bit unsure how much runtime checking is too much for a project like babel (perf-sensitive) - I have no idea how many times those functions are called over course of the program.

> Perhaps the text Neem contact op from my original code sample could be changed to English Contact us or some example text to keep the language consistent.

I even wanted to keep that untouched in a form of a "tribute" to the original issue 😉 

> Or probably this check should be in the functions which use this.key +/- 1

This would make it harder to find them at the moment, considering that this can get passed as argument, destructured etc. Also probably its better to have them as ints right away - let's convert less and keep the shapes consistent when possible.

jorrit:
Thanks!

PR #7216, opened 2018-01-14
bugfix: set color level when color is forced

Timer:
This fixes a bug introduced when the package was upgraded to `chalk@^2`.

Chalk 2 detects color support and level separately, so you must set enabled and level to gain color in a non-terminal environment.

### DIFF
--- a/packages/babel-code-frame/src/index.js
+++ b/packages/babel-code-frame/src/index.js
@@ -190,7 +190,7 @@ export function codeFrameColumns(
     (opts.highlightCode && Chalk.supportsColor) || opts.forceColor;
   let chalk = Chalk;
   if (opts.forceColor) {
-    chalk = new Chalk.constructor({ enabled: true });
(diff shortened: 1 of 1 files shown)


packages/babel-code-frame/src/index.js
```diff
@@ -190,7 +190,7 @@ export function codeFrameColumns(
     (opts.highlightCode && Chalk.supportsColor) || opts.forceColor;
   let chalk = Chalk;
```
wtgtybhertgeghgtwtg: Totally not in the scope of this PR, but wouldn't this work as a ternary?
```
const chalk = opts.forceColor ? new Chalk.constructor({ enabled: true, level: 1 }) : Chalk;
```

Timer: Sure.

existentialism:
@Timer thanks :P

Timer:
No, thank you! 😄

PR #7217, opened 2018-01-15
[TS] Transform multiple assert and assign correctly

### LINKED ISSUE #7193: [TS] Multiple "as" in assignment expression don't get transformed
This is a **bug report**.

### Input Code

```ts
let a;
(a as number as any) = 42;
```

```js
{
  "plugins": ["transform-typescript"]
}
```

### Expected Behavior

```js
var a;
a = 42;
```
TS Playground: [Link](http://www.typescriptlang.org/play/#src=let%20a%3B%0D%0A(a%20as%20number%20as%20any)%20%3D%2042%3B)

### Current Behavior

Throws an error
> TypeError: Property left of AssignmentExpression expected node to be of a type ["LVal"] but instead got "TSAsExpression"

### Context

Was trying out the transforms after #7098. I seem to remember that this input worked before, because I think I tried this out in the REPL. But I don't remember too well. I can see that the parser results in the right AST, as per the tests in the PR and my knowledge.
### Your Environment

| ---------------- | -------
| Babel            |  7.0.0-beta.37 (latest master)
| Babylon          | 
| node             | 8.9.3
| npm              | 5.5.1
| Operating System | Windows 10

maaz93:
| Fixed Issues?            | Fixes #7193 

The fix helps transform multiple `as` in the LHS of an assignment expression in TypeScript

## Input
```ts
let a;
(a as number as any) = 42;
```

## Output
```js
let a;
a = 42;
```

Also,  ran `node scripts/generate-babel-types-docs.js` to generate the `@babel/types` README.


simonbuchan:
It theoretically would be cleaner to have an explicit `lval: boolean` field on expression syntax nodes, but that's a decent breaking change.
Could instead add `TSLValAsExpression` that aliases `TSAsExpression` and `LVal`, if that's any better? ie. pay when parsing, not when visiting.
Presumably there's a similar issue for the other two type assert/casts: TS `<type> expr`, Flow `(expr: type)`.

simonbuchan:
It looks like the parser verifies lval-ness, but not very well:
- [flow `(x: number) = 123;` works correctly](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=BQDwXABAdgrgtgIwKYCcCUEC8ECMAmAZgG4AoIA&debug=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=false&presets=flow&prettier=false&targets=&version=7.0.0-beta.37)
- [but `((x: number): number) = 123;` doesn't](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=BTAeC4AIDsFcFsBGBTATgSinJb2QLyQCMATAMwDcAUEA&debug=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=false&presets=flow&prettier=false&targets=&version=7.0.0-beta.37)
- [and TS `(x as number) = 123` doesn't either](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=BQHgdgrgtgRgpgJwHwAIAeBKFBeFBGAJgGYBuAKCA&debug=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=false&presets=typescript&prettier=false&targets=&version=7.0.0-beta.35)

It looks like this is happening here:
- https://github.com/babel/babel/blob/master/packages/babylon/src/parser/expression.js#L154-L162
- https://github.com/babel/babel/blob/master/packages/babylon/src/parser/lval.js#L44

But I can't see how the working flow check is happening? Is anyone even using `LVal`?

existentialism:
@simonbuchan re: the beta version repl issue, see https://github.com/babel/babel/issues/6174

simonbuchan:
Makes sense to me to be closer to Flow, but I've no experience with babel code here (for some reason, `make bootstrap` fails under WSL for me, I haven't gotten around to investigating).

Looking at the flow transform code, in TS watch out for `((<U> (x as T)) as V) = 123`, that is nested `TSTypeAssertion` and `TSAsExpression`. Maybe adding an alias so both can be treated the same would be handy?

simonbuchan:
Or as mentioned in the handbook:

> If necessary, you can also apply the same function for multiple visitor nodes by separating them with a | in the method name as a string like Identifier|MemberExpression.

> Example usage in the flow-comments plugin

```js
const MyVisitor = {
  "ExportNamedDeclaration|Flow"(path) {}
};
```

maaz93:
@simonbuchan Thanks. I think parsing `TSTypeAssertion` is not yet supported, so I might not be able to test if it works in the transform or not. Anyway, I'll get back with my investigation.

simonbuchan:
It works fine if JSX syntax is not also enabled:
- [just typescript, works](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBAnhC8EAoA8A7ArgWwEYFMwD4IAPASgG4AoIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=false&presets=typescript&prettier=false&targets=&version=7.0.0-beta.37&envVersion=)
- [with react preset, fails](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBAnhC8EAoA8A7ArgWwEYFMwD4IAPASgG4AoIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=false&presets=react%2Ctypescript&prettier=false&targets=&version=7.0.0-beta.37&envVersion=)

With #7091 you can get the same behavior as `tsc` even:

```
overrides: [
  { test: /\.ts$/, presets: ['@babel/env', '@babel/typescript'] },
  { test: /\.tsx$/, presets: ['@babel/env', '@babel/typescript', '@babel/react'] },
],
```

simonbuchan:
Oops, missed

> in an assignment expression

sorry! Yeah, you're right; forgot about that.

PR #7218, opened 2018-01-15
Added custom NodePath.prototype.toString method as debug utility

Andarist:
| Fixed Issues?            | 
| Patch: Bug Fix?          | 
| Major: Breaking Change?  | 
| Tests Added + Pass?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Just a debug utility, I've found the pattern useful when debugging. Continuing inspecting ASTs is quite mind-bending over course of longer debugs.

### DIFF
--- a/packages/babel-traverse/src/path/index.js
+++ b/packages/babel-traverse/src/path/index.js
@@ -7,6 +7,7 @@ import traverse from "../index";
 import Scope from "../scope";
 import * as t from "@babel/types";
 import { path as pathCache } from "../cache";
+import generator from "@babel/generator";
 
 // NodePath is split across many files.
 import * as NodePath_ancestry from "./ancestry";
@@ -145,6 +146,10 @@ export default class NodePath {
   debug(message) {
     if (!debug.enabled) return;
     debug(`${this.getPathLocation()} ${this.type}: ${message}`);
+  }
+
+  toString() {
+    return generator(this.node).code;
   }
 }
 


nicolo-ribaudo:
It looks like ci failures are not related?

Andarist:
They shouldnt be related I think, I've restarted CI to see what happens.

Andarist:
CI is failing because of new `browserslist` version with a bug (it's getting fixed right now). How did it get installed though? We have a lock file after all... 😢

nicolo-ribaudo:
> We have a lock file after all... 

I think that lockfiles are ignored by default, unless you specify a flag I don't remember right now

packages/babel-traverse/src/path/index.js
```diff
@@ -7,6 +7,7 @@ import traverse from "../index";
 import Scope from "../scope";
 import * as t from "@babel/types";
 import { path as pathCache } from "../cache";
+import generator from "@babel/generator";
```
loganfsmyth: This is just a devDependency at the moment, so this is going to throw when installed in production.

hzoo: Also is this going to be a circular dep if generator uses traverse?

packages/babel-traverse/src/path/index.js
```diff
@@ -146,6 +147,10 @@ export default class NodePath {
     if (!debug.enabled) return;
     debug(`${this.getPathLocation()} ${this.type}: ${message}`);
   }
+
+  toString() {
+    return generator(this.node).code;
```
hzoo: Or could we somehow make this change in babel-core which does have it as a dep?

packages/babel-traverse/src/path/index.js
loganfsmyth: It doesn't, so I think we're fine for the dependency-cycle case. I'm happy with just adding it as a dependency for now.

Andarist: oh, my bad, ive checked `package.json` too quickly, didnt notice its just a devDep - fix is on the way

packages/babel-traverse/src/path/index.js
Andarist: any ideas on how that would look like? at the moment it's as easy as doing:
```js
CallExpression(path) {
  console.log(path + '')
}
```

packages/babel-traverse/src/path/index.js
loganfsmyth: No problem.

jridgewell:
<h1>THIS IS THE BEST THING EVER</h1>

Why did we never think of this before? How many hours have I wasted stepping into object literals to try an figure out what AST tree I was mutating? ❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️

Issue #2931, opened 2015-11-09
Option to exclude babel-polyfill for babel-node


deadbeef84:
I'm getting the `"only one instance of babel-polyfill is allowed"` error when running babel-node on my application that has a `require('babel-polyfill')`. I know this has been reported several times, and I understand that this is the expected behavior, but I would expect `babel-node index.js` to have the same effect as `babel index.js | node`.

With the current behaviour it is not possible to use babel-node on code that is intended to be compiled with babel and at the same time uses babel-polyfill. Well, I guess it is possible with the following hack:

```
if (!global._babelPolyfill) {
  require('babel-polyfill')
}
```

**Feature request:**
Either add a parameter to babel-node to skip automatic inclusion of babel-polyfill **OR** have the user include the babel-polyfill manually even when using babel-node.

hzoo:
It's possible to do via adding a cli param + doing an if check and then requiring it in https://github.com/babel/babel/blob/68148b8a05a562740712dcddb483747480748c50/packages/babel-cli/src/_babel-node.js#L11-L27.

We don't recommend using babel-node for prod apps anyway though (the recommendation is to compile your code and then run it via `babel src -d dist` and then running that.

hzoo:
If you want to make a pr for this (maybe `--no-polyfill`) or something then I think we would accept this. Closing for inactivity

PR #7226, opened 2018-01-16
Add —nopolyfill to babel-node cli args

sfarthin:
*The Problem*
When working on an app in development, it is common to use babel-node which automatically adds babel-polyfill. When packaging the app for production with babel, babel-polyfill is _not_ added which may cause the app to fail only in production. This especially is surprising to the developer when the app uses babel-runtime, but some dependency from another author expects babel-polyfill globals. It is best to catch these cases in development

*Feature*
Following @hzoo advice in https://github.com/babel/babel/issues/2931, This PR adds a command-line argument to use babel-node without babel-polyfill (`--nopolyfill`).

This can also be used to fix the issue "only one instance of babel-polyfill is allowed" as reported in the issue above. I did not use `--no-polyfill` because commander confusingly renames the option to `program.polyfill`

### DIFF
--- a/packages/babel-node/src/_babel-node.js
+++ b/packages/babel-node/src/_babel-node.js
@@ -5,7 +5,6 @@ import path from "path";
 import repl from "repl";
 import * as babel from "@babel/core";
 import vm from "vm";
-import "@babel/polyfill";
 import register from "@babel/register";
 
 import pkg from "../package.json";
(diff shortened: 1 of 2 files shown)


sfarthin:
Setting the default value of the argument to `false` makes `program.polyfill` true. This seemed confusing at first, but it makes sense now. I updated to conform to that format.

loganfsmyth:
Could you switch to using `babel-register` directly instead of using `babel-node` if you need this behavior?

sfarthin:
@loganfsmyth Do you mean create a new file like this (or do you suggest another way):

Example dev.js:
```javascript
require("babel-register")();
require('./app');
```

Then in development run `node dev.js` ? I supposed this accomplishes the same thing, but is not quite as nice (since we use babel-node religiously for running everything else in development). Running `babel-node dev.js` seems like an easy accident.

Are there any side-effects (other than babel-polyfill) to running this way as opposed to `babel-node app` ?

PR #7227, opened 2018-01-17
Fix up flow errors

existentialism:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Pretty straightforward, but couple notes:

* The remaining flow error related to the variance node is fixed in https://github.com/babel/babel/pull/7101

* We should really rethink all the is/has state booleans in tokenzier/state

* Since we run flow from root, it couldn't resolve charcodes in `babylon`. I originally used `--hoist` with lerna, but @hzoo mentioned it was buggy, and `--hoist charcodes` wouldn't build (likely related to https://github.com/lerna/lerna/issues/1028)... so for now, added it as a top level dep, not great, but couldn't get anything else to work that wasn't also just as hacky. Open to suggestions!

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -13,8 +13,8 @@ codemods/*/src
 [libs]
 lib/file.js
 lib/parser.js
-lib/packages/babel-types/lib/index.js.flow
 lib/third-party-libs.js.flow
+packages/babel-types/lib/index.js.flow
 
 [options]
 include_warnings=true
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "bundle-collapser": "^1.2.1",
     "chai": "^4.1.0",
     "chalk": "^2.0.0",
-    "charcodes": "0.0.10",
+    "charcodes": "0.1.0",
     "derequire": "^2.0.2",
     "eslint": "^4.5.0",
     "eslint-config-babel": "^7.0.2",
--- a/packages/babel-template/src/formatters.js
+++ b/packages/babel-template/src/formatters.js
@@ -12,7 +12,7 @@ function makeStatementFormatter<T>(
   return {
     // We need to prepend a ";" to force statement parsing so that
     // ExpressionStatement strings won't be parsed as directives.
-    // Alonside that, we also prepend a comment so that when a syntax error
+    // Alongside that, we also prepend a comment so that when a syntax error
     // is encountered, the user will be less likely to get confused about
     // where the random semicolon came from.
     code: str => `/* @babel/template */;\n${str}`,
@@ -54,16 +54,17 @@ export const statement: Formatter<BabelNodeStatement> = makeStatementFormatter(
 
 export const expression: Formatter<BabelNodeExpression> = {
   code: str => `(\n${str}\n)`,
-  validate: (ast: BabelNodeFile) => {
-    const { program } = ast;
+  validate: ({ program }) => {
     if (program.body.length > 1) {
       throw new Error("Found multiple statements but wanted one");
     }
+    // $FlowFixMe
     const expression = program.body[0].expression;
     if (expression.start === 0) {
       throw new Error("Parse result included parens.");
     }
   },
+  // $FlowFixMe
   unwrap: ast => ast.program.body[0].expression,
 };
 
--- a/packages/babylon/src/tokenizer/index.js
+++ b/packages/babylon/src/tokenizer/index.js
@@ -467,6 +467,7 @@ export default class Tokenizer extends LocationParser {
       const assign =
         this.input.charCodeAt(this.state.pos + 2) === charCodes.equalsTo;
       if (assign) {
+        // $FlowFixMe
         this.expectPlugin("logicalAssignment");
       }
       this.finishOp(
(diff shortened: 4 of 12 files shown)


packages/babel-template/src/formatters.js
```diff
@@ -54,16 +54,16 @@ export const statement: Formatter<BabelNodeStatement> = makeStatementFormatter(
...
       throw new Error("Found multiple statements but wanted one");
     }
-    const expression = program.body[0].expression;
-    if (expression.start === 0) {
+
+    if (program.body[0].expression && program.body[0].expression.start === 0) {
       throw new Error("Parse result included parens.");
     }
   },
+  // $FlowFixMe
```
existentialism: Def a todo... couldn't find a solution that made Flow happy :(

.flowconfig
```diff
@@ -13,8 +13,8 @@ codemods/*/src
 [libs]
 lib/file.js
 lib/parser.js
-lib/packages/babel-types/lib/index.js.flow
 lib/third-party-libs.js.flow
+packages/babel-types/lib/index.js.flow
```
existentialism: 😭😭😭😭😭

existentialism:
@nicolo-ribaudo @yavorsky added a few updates. still looking into why flow reports no errors on travis, since... it kinda of defeats the entire purpose of this :)

packages/babel-template/src/formatters.js
```diff
@@ -54,16 +54,16 @@ export const statement: Formatter<BabelNodeStatement> = makeStatementFormatter(
...
-  validate: (ast: BabelNodeFile) => {
-    const { program } = ast;
+  validate: ({ program }) => {
     if (program.body.length > 1) {
       throw new Error("Found multiple statements but wanted one");
     }
-    const expression = program.body[0].expression;
-    if (expression.start === 0) {
+
+    if (program.body[0].expression && program.body[0].expression.start === 0) {
```
Andarist: want previous version more readable? would ofc need to add `expression &&` to the if's condition

existentialism: For sure it was, this is just not great workaround for flow :(

Similarly to unwrap below it, `const expression = program.body[0].expression;` trips up a bunch of `Property not found` errors due to `program.body[0]` being `Array<BabelNodeStatement>`.

Andarist: Not sure if I understand the problem here, I guess I'd have to check out produced errors but it seems that those 2 should be equivalent and that they shouldnt trip flow:
```js
if (program.body[0].expression && program.body[0].expression.start === 0)
```
vs
```js
const expression = program.body[0].expression;
if (expression && expression.start === 0)
```

existentialism: @Andarist I don't disagree :)

existentialism: Might be better to just revert this change and just mark this as a FIXME as well (like unwrap)?

Andarist: i'd say it's better to keep it "fixed" (like it is now) - Im only confused why those 2 are not equivalent

existentialism: Yeah, I'll try to do more digging later today.

package.json
```diff
@@ -25,6 +25,7 @@
     "bundle-collapser": "^1.2.1",
     "chai": "^4.1.0",
     "chalk": "^2.0.0",
+    "charcodes": "0.1.0",
```
loganfsmyth: Is this being used somewhere? I'd have thought the import in Babylon would be enough

packages/babel-template/src/formatters.js
loganfsmyth: There's a lot of other stuff in this PR that would be good to land. I'm fine with `$FlowFixMe` if it'll let us land this.

package.json
existentialism: > Since we run flow from root, it couldn't resolve charcodes in babylon. I originally used --hoist with lerna, but @hzoo mentioned it was buggy, and --hoist charcodes wouldn't build (likely related to lerna/lerna#1028)... so for now, added it as a top level dep, not great, but couldn't get anything else to work that wasn't also just as hacky. Open to suggestions!

I tried a number of things to get it to resolve, but couldn't. I'll revisit it after digging more into why Travis is reporting zero errors.

loganfsmyth: Sounds good. Flow is definitely smart enough to use the definitions relative to the module using the package, so that's very weird.

packages/babylon/src/tokenizer/index.js
```diff
@@ -467,6 +467,7 @@ export default class Tokenizer extends LocationParser {
       const assign =
         this.input.charCodeAt(this.state.pos + 2) === charCodes.equalsTo;
       if (assign) {
+        // $FlowFixMe
```
existentialism: Flow doesn't like that the inheritance chain goes:

`BaseParser (which has hasPlugin) -> CommentsParser -> LocationParser -> Tokenizer -> UtilParser (which has expectPlugin)`

PR #7229, opened 2018-01-17
Fix typo [skip ci]

LucaT1:
Fixed a small typo in a `@babel/traverse` error message. Just noticed when playing around with visitors.

OT: Is there currently a way to use an async visitor?


hzoo:
We added this message because hasn't ever been supported but we'd like to look into it at some point (not soon).

LucaT1:
Thanks so much, you're always so polite and friendly. Remember to take care of yourself and keep up the amazing work for the community!:heart:

PR #7239, opened 2018-01-18
TypeScript: Support parsing 'unique' type operator

andy-ms:
| Fixed Issues?            |
| Documentation PR         |

Microsoft/TypeScript#15473 added the `unique` keyword to the language. This adds support in babylon to parse it. There were no necessary changes to the generator, which already handles any keyword, or to the plugin, which erases all types without needing to look at their contents.
@rbuckton and @DanielRosenwasser for review.
See also eslint/typescript-eslint-parser#433, which does the same thing in their parser. @azz @JamesHenry

### DIFF
--- a/packages/babel-generator/test/fixtures/typescript/types-type-operator/input.js
+++ b/packages/babel-generator/test/fixtures/typescript/types-type-operator/input.js
@@ -1,1 +1,2 @@
 let x: keyof T;
+let y: unique symbol;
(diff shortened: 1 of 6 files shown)


nicolo-ribaudo:
Q: Is it `unique [any type]` or only `unique symbol`?

andy-ms:
We parse `unique` as a type operator, so any type will *parse* after it. But only `unique symbol` will type-check without error. If you write `unique number` there will be a grammar diagnostic.

azz:
Is it worth setting `prefix: true` on the TSTypeOperator node? If not I can remove it from the tsep PR.

andy-ms:
There's no such property in the TypeScript implementation, so I don't think it's worth adding yet. `ArrayTypeNode` is a separate node kind, so there aren't any postfix type operators.

hzoo:
I guess we technically would want a test to see that it's stripped out but AST structure is the same so it's covered.

PR #7240, opened 2018-01-18
Compile Babylon with the normal workflow and use "overrides" in our c…

nicolo-ribaudo:
Now `make watch` also works for Babylon.

This PR doesn't introduces breaking changes, because rollup's output format was "cjs" and so it would still have needed to be `webpack`ed to work in a browser.

I had moved the `babel-plugin-transform-charcodes` and `charcodes` packages to the root; if you prefer I can keep them in Babylon and use the relatve path for `babel-plugin-transform-charcodes`.


packages/babylon/package.json
```diff
@@ -24,12 +24,6 @@
   },
   "devDependencies": {
     "@babel/helper-fixtures": "7.0.0-beta.38",
-    "babel-plugin-transform-charcodes": "0.0.10",
```
Andarist: cant we have those (`charcodes`) in 2 places and let lerna link them? I mean - it only makes sense to me to have them here, it will be super weird to see uknown dep in the source files and not see it in package.json at all

xtuc: That would simplify the process but the usage is very specific to Babylon and char codes.

packages/babylon/rollup.config.js
```diff
@@ -1,34 +0,0 @@
-import babel from "rollup-plugin-babel";
-import nodeResolve from "rollup-plugin-node-resolve";
-
-export default {
-  input: "src/index.js",
-  output: {
-    file: "lib/index.js",
-    format: "cjs",
```
hzoo: there was no deps though so I don't think it was any different for the format?

hzoo:
This is cool but then this just means we don't have a single file for babylon anymore? I think that's still worth doing for publish and this for dev.

nicolo-ribaudo:
I have an implementation locally that compiles to different files in `watch` mode but to a single file in `build` mode (also because, as Brian pointed out on slack, `require` calls make Babylon much slower).

packages/babylon/rollup.config.js
nicolo-ribaudo: It still has exports.

hzoo: ah yeah I just meant no requires

nicolo-ribaudo:
Now `make build` compiles Babylon to a single file, while `make watch` generates multiple files. This behavior can be easily extended in the future to other packages split in many files (like `babel-traverse`).

.babelrc.js
```diff
@@ -30,6 +32,11 @@ function istanbulHacks() {
   };
 }
 
+const babylonTransformCharcodes = relative.resolve(
+  "babel-plugin-transform-charcodes",
+  "./packages/babylon"
+);
```
xtuc: Since it's an external package, couldn't we just require it and pass it as a function? Might be easier, right?

.babelrc.js
```diff
@@ -1,5 +1,7 @@
 "use strict";
 
+const relative = require("require-relative");
+
 // Blame Logan for this.
```
xtuc: > Blame Logan for this.

Let's remove this?

.babelrc.js
nicolo-ribaudo: Since `"babel-plugin-transform-charcodes"` is in a subpackage, we would still need to
```js
relative(
  "babel-plugin-transform-charcodes",
  "./packages/babylon"
);
```

.babelrc.js
nicolo-ribaudo: https://github.com/istanbuljs/istanbuljs/issues/92 is still open, so I think that that function is still needed.

.babelrc.js
hzoo: Why don't we just add it to the top level?

nicolo-ribaudo:
Thank you :heart:

.babelrc.js
xtuc: This plugins isn't a subpackage from our repo but a external one. We can add it to the top level one.

.babelrc.js
xtuc: I meant the comment. We can't blame Logan for this

.babelrc.js
nicolo-ribaudo: I initially added it to the top level but, [as pointed out](https://github.com/babel/babel/pull/7240#pullrequestreview-89976862), it would be confusing to see `import "charcodes"` in Babylon without it beeing in Babylon's `package.json`.

hzoo: makes sense, I guess we could put it in both places too

PR #7242, opened 2018-01-20
Add regexp support to include/exclude

### LINKED ISSUE #6636: Easier way to exclude/include (especially builtIns)
> Issue originally reported by @hzoo in babel/babel-preset-env#132

Context:

Docs for the include/exclude options: https://github.com/babel/babel-preset-env#include
Example: https://github.com/babel/babel-preset-env#example-with-includeexclude

---

This is a bit more involved, but still basically the gist is changing indexOf to handle regex.

This is mostly for built-ins.

There's a lot of them: https://github.com/babel/babel-preset-env/blob/master/test/fixtures/preset-options/use-builtins-all/expected.js

If we don't use `reflect` or `number` or `math` you currently have to list each one individually since we do an `indexOf` check

I think we can change it to accept a regex or glob (I think regex) to match multiple of them.

So instead of:

```json
{
  "exclude": [
    "es6.math.acosh",
    "es6.math.asinh",
    "es6.math.atanh",
    "es6.math.cbrt",
    "es6.math.clz32",
    "es6.math.cosh",
    "...etc"
  ]
}
```

be able to handle something like

```json
{
  "exclude: [/^es6.math.+/]
}
```

The current logic is https://github.com/babel/babel-preset-env/blob/8eb0c0799cefe3548216a3b2bc372e21dcd3b065/src/index.js#L250-L257
and https://github.com/babel/babel-preset-env/blob/8eb0c0799cefe3548216a3b2bc372e21dcd3b065/src/index.js#L277-L279

```js
    polyfills = Object.keys(builtInsList)
      .filter((builtInName) => isPluginRequired(targets, builtInsList[builtInName]))
      .concat(defaultInclude)
      .filter((plugin) => exclude.builtIns.indexOf(plugin) === -1) // here
      .concat(include.builtIns); // here
```

```js
  const allTransformations = transformations
  .filter((plugin) => exclude.plugins.indexOf(plugin) === -1) // here
  .concat(include.plugins); // here
```

- [ ] change indexOf check to handle regex
- [ ] tests

aminmarashi:
| Fixed Issues?            | Fixes #6636  
| Patch: Bug Fix?          |
| Major: Breaking Change?  | 
| Documentation PR         | 
| Any Dependency Changes?  | 

This PR adds support for regular expressions to includes/excludes list of plugins/builtins. Regular expressions can be passed as normal `RegExp` objects, or in `string` format.
We expand the regular expressions to get the list of plugins and then pass them around, therefore virtually none of the previous functions are affected.
Plugins can still be passed with full name or a partial name.

Examples:

- Full name: `"es6.math.sign"`
- Partial name: `"es6.math.*"`
- RegExp in string: `"es6.math.sin.?"`
- `RegExp` Object: `/^transform-.*$/` or `new RegExp("^transform-modules-.*")`

We still normalize the plugin names _before turning them into `RegExp`_, therefore patterns like `"^babel-plugin-transform-.*"` still work (, but `/^babel-plugin-transform-.*/`  doesn't).

**Note:** In all of the above examples `.` represents _a match for any character_ in a regular expression, and not the _dot_ character so for example  `"es6.math.sign"` would match  `es6-math-sign`.

### DIFF
--- a/packages/babel-preset-env/src/normalize-options.js
+++ b/packages/babel-preset-env/src/normalize-options.js
@@ -26,8 +26,10 @@ const pluginToRegExp = (plugin: any): RegExp => {
 
 const selectPlugins = (regexp: RegExp): Array<string> =>
   Array.from(validIncludesAndExcludes).filter(
-    item => regexp instanceof RegExp && item.match(regexp),
+    item => regexp instanceof RegExp && regexp.test(item),
   );
+
+const flatten = array => [].concat(...array);
 
 const expandIncludesAndExcludes = (
   plugins: Array<string | RegExp> = [],
@@ -50,7 +52,7 @@ const expandIncludesAndExcludes = (
     valid. Please check data/[plugin-features|built-in-features].js in babel-preset-env`,
   );
 
-  return [].concat(...selectedPlugins);
+  return flatten(selectedPlugins);
 };
 
 const validBrowserslistTargets = [
@@ -58,8 +60,8 @@ const validBrowserslistTargets = [
   ...Object.keys(browserslist.aliases),
 ];
 
-const normalizePluginName = (plugin: string): string =>
-  plugin.replace("babel-plugin-", "");
+export const normalizePluginName = (plugin: string): string =>
+  plugin.replace(/^babel-plugin-/, "");
 
 export const checkDuplicateIncludeExcludes = (
   include: Array<string> = [],
--- a/packages/babel-preset-env/src/normalize-options.js
+++ b/packages/babel-preset-env/src/normalize-options.js
@@ -30,7 +30,7 @@ const selectPlugins = (regexp: RegExp): Array<string> =>
 const pluginToRegExp = (plugin: any): RegExp => {
   return plugin instanceof RegExp
     ? plugin
-    : new RegExp(normalizePluginName(plugin));
+    : new RegExp(`^${normalizePluginName(plugin)}$`);
 };
 
 const populatePlugins = (
@@ -39,7 +39,8 @@ const populatePlugins = (
 ): Array<string> => pluginList.concat(selectPlugins(regexp));
 
 const isValidPlugin = (plugin: any): boolean =>
-  validRegExp(plugin) && selectPlugins(pluginToRegExp(plugin)).length > 0;
+  validRegExp(plugin) &&
+  selectPlugins(`^${normalizePluginName(plugin)}$`).length > 0;
 
 const expandIncludesAndExcludes = (
   plugins: Array<string>,
--- a/packages/babel-preset-env/src/normalize-options.js
+++ b/packages/babel-preset-env/src/normalize-options.js
@@ -85,7 +85,7 @@ const validBrowserslistTargets = [
 ];
 
 const normalizePluginName = (plugin: string): string =>
-  plugin.replace(/^babel-plugin-/, "");
+  plugin.replace(/babel-plugin-/, "");
 
 export const checkDuplicateIncludeExcludes = (
   include: Array<string> = [],
--- a/packages/babel-preset-env/src/normalize-options.js
+++ b/packages/babel-preset-env/src/normalize-options.js
@@ -24,11 +24,28 @@ const validRegExp = (regexp: string): boolean => {
   }
 };
 
+const selectPlugins = (regexp: RegExp): Array<string> =>
+  Array.from(validIncludesAndExcludes).filter(item => item.match(regexp));
+
+const pluginToRegExp = (plugin: any): RegExp => {
+  return plugin instanceof RegExp
+    ? plugin
+    : new RegExp(normalizePluginName(plugin));
+};
+
+const populatePlugins = (
+  pluginList: Array<RegExp>,
+  regexp: RegExp,
+): Array<string> => pluginList.concat(selectPlugins(regexp));
+
+const isValidPlugin = (plugin: any): boolean =>
+  validRegExp(plugin) && selectPlugins(pluginToRegExp(plugin)).length > 0;
+
 const expandIncludesAndExcludes = (
   plugins: Array<string>,
   type: string,
 ): Array<string> => {
-  const invalidRegExp = plugins.filter(plugin => !validRegExp(plugin));
+  const invalidRegExp = plugins.filter(plugin => !isValidPlugin(plugin));
 
   invariant(
     invalidRegExp.length === 0,
@@ -37,17 +54,7 @@ const expandIncludesAndExcludes = (
     )}' passed to the '${type}' option are not valid.`,
   );
 
-  const validIncludesAndExcludesArray = Array.from(validIncludesAndExcludes);
-
-  const pluginRegExpList = plugins.map(plugin => new RegExp(plugin));
-
-  const selectPlugins = regexp =>
-    validIncludesAndExcludesArray.filter(item => item.match(regexp));
-
-  const populatePlugins = (pluginList, regexp) =>
-    pluginList.concat(selectPlugins(regexp));
-
-  return pluginRegExpList.reduce(populatePlugins, []);
+  return plugins.map(pluginToRegExp).reduce(populatePlugins, []);
 };
 
 export const validateIncludesAndExcludes = (
@@ -77,11 +84,8 @@ const validBrowserslistTargets = [
   ...Object.keys(browserslist.aliases),
 ];
 
-export const normalizePluginName = (plugin: string): string =>
+const normalizePluginName = (plugin: string): string =>
   plugin.replace(/^babel-plugin-/, "");
-
-export const normalizePluginNames = (plugins: Array<string>): Array<string> =>
-  plugins.map(normalizePluginName);
 
 export const checkDuplicateIncludeExcludes = (
   include: Array<string> = [],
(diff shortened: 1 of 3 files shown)


packages/babel-preset-env/src/normalize-options.js
```diff
@@ -15,6 +15,48 @@ const validIncludesAndExcludes = new Set([
...
+  const invalidRegExp = plugins.filter(plugin => !isValidPlugin(plugin));
+
+  invariant(
+    invalidRegExp.length === 0,
+    `Invalid RegExp: The given regular expression '${invalidRegExp.join(
+      ", ",
+    )}' passed to the '${type}' option are not valid.`,
+  );
+
+  return plugins.map(pluginToRegExp).reduce(populatePlugins, []);
```
yavorsky: And if the value is a `string` and it's a valid regExp and could be converted with `pluginToRegExp` it's still a `string`, so there will be the 3-rd `new RegExp(plugin)` call.

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -15,6 +15,48 @@ const validIncludesAndExcludes = new Set([
...
+  try {
+    new RegExp(regexp);
+    return true;
+  } catch (e) {
+    return false;
+  }
+};
+
+const selectPlugins = (regexp: RegExp): Array<string> =>
+  Array.from(validIncludesAndExcludes).filter(item => item.match(regexp));
```
yavorsky: Notice that with `new RegExp(value)`, for case like `es6.reflect.set`, it will also match the `es6.reflect.set-prototype-of`, which is a different built-in.

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -15,6 +15,48 @@ const validIncludesAndExcludes = new Set([
...
+
+const pluginToRegExp = (plugin: any): RegExp => {
+  return plugin instanceof RegExp
+    ? plugin
+    : new RegExp(normalizePluginName(plugin));
```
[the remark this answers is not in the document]
aminmarashi: True. I'll look into optimizing the `RegExp` creation.

nicolo-ribaudo:
1) Do you think that uses would expect strings to work as regexs or as glob pattern?
2) I think that we should avoid converting the string to a regex, since in my experience writing regexs in strings lead to strange escapes. (e.g. `\\[` to mean `[`). If a user wants to use a regular expression, there is `.babelrc.js` for that.

aminmarashi:
@yavorsky Thank you. I think I've addressed all of the issues.

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -15,38 +15,52 @@ const validIncludesAndExcludes = new Set([
   ...defaultWebIncludes,
 ]);
 
-export const validateIncludesAndExcludes = (
-  opts: Array<string> = [],
+const pluginToRegExp = (plugin: any): RegExp => {
+  try {
+    return plugin instanceof RegExp
```
Qantas94Heavy: Nitpick: Is there a chance of `plugin instanceof RegExp` throwing? If not, moving this out of the try-catch block may be a bit more readable. Not so important though.

aminmarashi: This `try/catch` is actually for `new RegExp`.

Qantas94Heavy: I was thinking of something like this, but it doesn't really matter anyway. Don't worry about it.

```js
if (plugin instanceof RegExp) {
  return plugin;
}

try {
  return new RegExp(`^${normalizePluginName(plugin)}$`);
} catch (e) {
  return null;
}
```

aminmarashi: That's actually a good idea. Will change this and the doc and ask for a review again. Thank you. 👍

aminmarashi:
@Qantas94Heavy Applied your comments 👍

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -15,38 +15,51 @@ const validIncludesAndExcludes = new Set([
   ...defaultWebIncludes,
 ]);
 
-export const validateIncludesAndExcludes = (
-  opts: Array<string> = [],
+const pluginToRegExp = (plugin: any): RegExp => {
+  if (plugin instanceof RegExp) return plugin;
+  try {
+    return  new RegExp(`^${normalizePluginName(plugin)}$`);
```
Qantas94Heavy: Typo, remove double space.

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -15,38 +15,51 @@ const validIncludesAndExcludes = new Set([
...
   ...Object.keys(browserslist.aliases),
 ];
 
-export const normalizePluginName = (plugin: string): string =>
-  plugin.replace(/^babel-plugin-/, "");
-
-export const normalizePluginNames = (plugins: Array<string>): Array<string> =>
-  plugins.map(normalizePluginName);
+const normalizePluginName = (plugin: string): string =>
+  plugin.replace("babel-plugin-", "");
```
Andarist: why the change? it can replace too much now

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -15,38 +15,51 @@ const validIncludesAndExcludes = new Set([
...
+  try {
+    return new RegExp(`^${normalizePluginName(plugin)}$`);
+  } catch (e) {
+    return null;
+  }
+};
+
+const selectPlugins = (regexp: RegExp): Array<string> =>
+  Array.from(validIncludesAndExcludes).filter(
+    item => regexp instanceof RegExp && item.match(regexp),
```
Andarist: `regexp.test(item)` would be IMHO more appropriate as it returns boolean instead of maybe array

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -15,38 +15,51 @@ const validIncludesAndExcludes = new Set([
...
-    `Invalid Option: The plugins/built-ins '${unknownOpts.join(
+    invalidRegExpList.length === 0,
+    `Invalid Option: The plugins/built-ins '${invalidRegExpList.join(
       ", ",
     )}' passed to the '${type}' option are not
     valid. Please check data/[plugin-features|built-in-features].js in babel-preset-env`,
   );
 
-  return opts;
+  return [].concat(...selectedPlugins);
```
Andarist: I guess this is suppose to make a `flatten` operation, it would be good to introduce this utility and use it here
```
const flatten = arr => [].concat(...arr)
```

packages/babel-preset-env/src/normalize-options.js
aminmarashi: We already wrap it in `^$` at [line 21](https://github.com/babel/babel/pull/7242/files#diff-1558508a41a0cf543273e188fe7bbf85R21)

packages/babel-preset-env/src/normalize-options.js
aminmarashi: Yeah, good idea 👍

aminmarashi: Do we have any special place for keeping these kinds of utils?

Andarist: not rly (at least i dont think so), I would just make it a local function in this file for now

packages/babel-preset-env/src/normalize-options.js
aminmarashi: yes, sounds good.

packages/babel-preset-env/src/normalize-options.js
Andarist: isnt it wrapped AFTER normalization? with ur approach we can get a final regexp of `/^smth-foo$/` for `smth-babel-plugin-foo` input, which was not the case before the PR, while we are at it maybe you could also add a test case for this?

aminmarashi: Yes, I made this change to still normalize cases like `"^babel-plugin-foo"`, but I don't think that's a good idea either. I'll add the test and fix this, nice catch 👍

packages/babel-preset-env/test/normalize-options.spec.js
```diff
@@ -23,6 +21,11 @@ describe("normalize-options", () => {
       ]);
     });
 
+    it("should not normalize babel-plugin with prefix", () => {
+      const normalized = normalizePluginName("prefix-babel-plugin-postfix");
```
Andarist: Imho this is kinda testing an implementation detail, could you rewrite it to test an actual effect rather than a single normalization step?

aminmarashi: It's a little bit tricky to write a test for the `include`/`exclude` field, because as far as I know there is no `prefix-babel-plugin-*`, therefore any test will end up throwing an error (invalid plugin) regardless of this behavior.

Andarist: makes sense, u'd have to mock something or create fake `node_modules` in test dir, not worth it then at this point

hzoo:
Thanks for the PR!

After looking at this again (I suggested this change originally), it feels like I did this as more of a workaround. Ideally you wouldn't have to do anything like this because Babel would handle it for you, but before that this should work for people that need it.

PR #7245, opened 2018-01-20
Tweak and add tests to babel-helper-annotate-as-pure

existentialism:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #7026
| Patch: Bug Fix?          | y
| Major: Breaking Change?  | n
| Minor: New Feature?      | n
| Tests Added + Pass?      | y/y
| Documentation PR         |
| Any Dependency Changes?  |
| License                  | MIT

Trying to clear through a big backlog of PRs I've been too busy to get done :(

This was originally part of a larger PR that normalizes comment props on nodes, but figured we could land this while bikeshedding that.

PR #7246, opened 2018-01-20
Make comment props more consistent

existentialism:
| Fixed Issues?            | 
| Major: Breaking Change?  | maybe?
| Tests Added + Pass?      | n/y
| Documentation PR         |
| Any Dependency Changes?  |

As mentioned in https://github.com/babel/babel/pull/7245 and some slack convos, this PR makes `leadingComments`, `trailingComments` and `innerComments` more consistent by not allowing them to be set to `null`.

I ran this through https://github.com/babel/babylon_performance a number of times and noticed no real impact.

Conditionally setting them seemed better than instantiating them as empty arrays, but def open to discussing alternatives as well.

### DIFF
--- a/packages/babylon/src/parser/comments.js
+++ b/packages/babylon/src/parser/comments.js
@@ -64,16 +64,14 @@ export default class CommentsParser extends BaseParser {
         // later.
         this.state.trailingComments.length = 0;
       }
-    } else {
-      if (stack.length > 0) {
-        const lastInStack = last(stack);
-        if (
-          lastInStack.trailingComments &&
-          lastInStack.trailingComments[0].start >= node.end
-        ) {
-          trailingComments = lastInStack.trailingComments;
-          lastInStack.trailingComments = null;
-        }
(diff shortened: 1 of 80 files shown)


Andarist:
Any particular reason why deleting the property instead of having an empty array there?

Those cases than you have refactored seems to be some special cases when comments "migrate" from one place to another, are copied from a node to node or smth and thats why they have to be cleared out from the previous node. I think in most cases nodes just have empty arrays as those properties.

existentialism:
> Any particular reason why deleting the property instead of having an empty array there?

None other than originally thinking it was more consistent for it to be not set after it was moved (and fwiw, it matches the [code this is based on](https://github.com/eslint/espree/blob/master/lib/comment-attachment.js).)

I don't really have strong thoughts about it, and will be fine with whatever the rest of the team thinks.

hzoo:
Don't really care if it's not slower, and if there's less in the AST maybe that's good

PR #6332, opened 2017-09-28
Adding parserOptions to babel-core transform result

kesne:
| Q                        | A
| ------------------------ | ---
| Patch: Bug Fix?          | No
| Major: Breaking Change?  | No
| Minor: New Feature?      | Yes
| Tests Added/Pass?        | Not yet
| Spec Compliancy?         | N/A
| License                  | MIT
| Doc PR                   | No
| Any Dependency Changes?  | No

This adds `parserOptions` to the transform result from `babel-core`. The intent here is to be able to run a no-op through `babel.transform()` and get the resulting babylon options. These could then be used in something like `babel-eslint` to automatically have it configured based on the result of your babel plugins.

PR #7248, opened 2018-01-21
Expose parse (babylon) for plugins

xtuc:
| Fixed Issues?            | no (probably unused) 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

See the `transform-eval` plugin for use case:

https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-eval/src/index.js#L15

I like this idea because it tigh alone with traverse the parse with the current version of Babel.

### DIFF
--- a/packages/babel-core/src/index.js
+++ b/packages/babel-core/src/index.js
@@ -12,6 +12,7 @@ export { getEnv } from "./config/helpers/environment";
 export * as types from "@babel/types";
 export { default as traverse } from "@babel/traverse";
 export { default as template } from "@babel/template";
+export { parse } from "babylon";
 
 import loadConfig from "./config";
 


loganfsmyth:
We've talked about this a bit for https://github.com/babel/babel-eslint/issues/573#issuecomment-356352031 and https://github.com/babel/babel/pull/6332#issuecomment-333399009. Personally I think we should to expose a parsing function, but it should be more than just a pass-through to Babylon. We'd probably want it to essentially do the whole first phase of config loading and just skip transformation in favor of parsing and returning the AST.

xtuc:
@loganfsmyth I agree, but i'm not sure to understand what do you mean with the config? Probably the Babylon configuration because right now it's not configured correctly, I will change it.

kaicataldo:
@xtuc I believe this is to ensure that Babylon parses with the correct plugins given a user's .babelrc. This would be extremely useful for downstream projects like babel-eslint.

xtuc:
@kaicataldo yes I understand, I will change it to take the current configuration into account.

kaicataldo:
@xtuc Cool - I was going to work on this, but it looks like you've already got it!

xtuc:
@kaicataldo this is a naive implementation, Babylon is not configured like Babel is (like the parser plugins).

xtuc:
Closing this in favor of https://github.com/babel/babel/pull/7291

PR #7255, opened 2018-01-22
linting passing

### LINKED ISSUE #7035: Regression babel-template with uppercase variables not meant to be replaced

bug

### Input Code

```js
const unsafeWrapTemplate = template(`
  if (process.env.NODE_ENV !== "production") {
    NODE;
  }
`);

unsafeWrapTemplate({
  NODE: path.node,
}),
```

### Expected Behavior
returns the template with NODE replaced

### Current Behavior

```
Error: No substitution given for "NODE_ENV"
    at /Users/danieltschinder/Documents/Github/react-fast-highlight/node_modules/@babel/template/lib/populate.js:16:15
    at populatePlaceholders (/Users/danieltschinder/Documents/Github/react-fast-highlight/node_modules/@babel/template/lib/populate.js:14:27)
    at /Users/danieltschinder/Documents/Github/react-fast-highlight/node_modules/@babel/template/lib/string.js:20:51
    at /Users/danieltschinder/Documents/Github/react-fast-highlight/node_modules/@babel/template/lib/builder.js:83:14
    at remove (/Users/danieltschinder/Documents/Github/react-fast-highlight/node_modules/babel-plugin-transform-react-remove-prop-types/lib/remove.js:88:28)
```
### Context
Found in `babel-plugin-remove-prop-types`

### Your Environment

| ---------------- | -------
| Babel            |  beta 35
| node             | 8.9.3
| npm              |  5.5
| Operating System | mac

thescientist13:
| Fixed Issues?            | helps resolve #7035 
| Documentation PR         | 

1. Updated babel-template placeholder text for missing substitution per suggestion [here](https://github.com/babel/babel/issues/7035#issuecomment-359587117)

### DIFF
--- a/packages/babel-template/src/populate.js
+++ b/packages/babel-template/src/populate.js
@@ -15,7 +15,10 @@ export default function populatePlaceholders(
       if (
         !Object.prototype.hasOwnProperty.call(replacements, placeholder.name)
       ) {
-        throw new Error(`No substitution given for "${placeholder.name}"`);
+        throw new Error(`Error: No substitution given for "NODE_ENV". If this is not meant to be a
+          placeholder you may want to consider passing one of the following options to babel-template:
+          - { placeholderPattern: false, placeholderWhitelist: new Set(['NODE']) }
+          - { placeholderPattern: /^NODE$/ }`);
       }
     });
     Object.keys(replacements).forEach(key => {


packages/babel-template/src/populate.js
```diff
@@ -15,7 +15,10 @@ export default function populatePlaceholders(
       if (
         !Object.prototype.hasOwnProperty.call(replacements, placeholder.name)
       ) {
-        throw new Error(`No substitution given for "${placeholder.name}"`);
+        throw new Error(`Error: No substitution given for "NODE_ENV". If this is not meant to be a
+          placeholder you may want to consider passing one of the following options to babel-template:
+          - { placeholderPattern: false, placeholderWhitelist: new Set(['NODE']) }
```
thescientist13: Not sure if 
`new Set(['NODE']`

should be this ?
`new Set(['NODE_ENV']`

thescientist13:
Working on fixing the test

maaz93:
Try running `make fix` to fix all the problems that can be fixed automatically. All errors reported by `prettier` are auto-fixable.

abiduzz420:
@thescientist13 You need to modify the expected error message here https://github.com/babel/babel/blob/master/packages/babel-template/test/index.js#L130
Hope this helps.

thescientist13:
thank you for that tip @maaz93 !!  The console output showed `--fix` so I was sad that it wasn't working for me
> `3 errors potentially fixable with the `--fix` option.`

Now it's working!  though not fixable it seems : ( 

Will keep plugging away

thescientist13:
thanks @abiduzz420, that's what I'm working on now 👍

packages/babel-template/src/populate.js
existentialism: The error message should continue using `placeholder.name`, don't think we want to say `NODE_ENV` for every error :)

thescientist13: @existentialism 
Ah... of course!

thescientist13:
think I got it now 🤞

packages/babel-template/src/populate.js
```diff
@@ -15,7 +15,14 @@ export default function populatePlaceholders(
       if (
         !Object.prototype.hasOwnProperty.call(replacements, placeholder.name)
       ) {
-        throw new Error(`No substitution given for "${placeholder.name}"`);
+        throw new Error(`Error: No substitution given for "${
+          placeholder.name
+        }". If this is not meant to be a
+          placeholder you may want to consider passing one of the following options to babel-template:
```
thescientist13: this formatting was put in by prettier.  maybe it expects `{` to always be the start of an object and thus need a line break? 

let me know if this 🆗

maaz93: Template literals, the ``, preserve spaces and new lines. You'll need to construct your string something like the way it's done here https://github.com/babel/babel/blob/180eda3211faa855654757bc44aeaf3961a8be05/packages/babel-core/src/transformation/normalize-file.js#L82

Otherwise, the error message will have unnecessary indentation.

thescientist13: @maaz93 
It seems to keep complaining about that `{`
```shell
error: Replace `··`Error:·No·substitution·given·for·"${placeholder.name` with ``Error:·No·substitution·given·for·"${⏎············placeholder.name⏎··········` (prettier/prettier) at packages/babel-template/src/populate.js:19:11:
  17 |       ) {
  18 |         throw new Error(
> 19 |             `Error: No substitution given for "${placeholder.name}". If this is not meant ` +
     |           ^
  20 |               `to be a placeholder you may want to consider passing one of the following options to babel-template:` +
  21 |               `- { placeholderPattern: false, placeholderWhitelist: new Set(['${placeholder.name}'])}` +
  22 |               `- { placeholderPattern: /^${placeholder.name}$/ }`

```

maaz93: Alright. This seems a little tricky. I formatted it here to work fine:

```js
        const placeholderName = placeholder.name;
        throw new Error(
          `Error: No substitution given for "${placeholderName}". If this is not meant to be a placeholder ` +
            `you may want to consider passing one of the following options to babel-template:` +
            `- { placeholderPattern: false, placeholderWhitelist: new Set(['${placeholderName}'])}` +
            `- { placeholderPattern: /^${placeholderName}$/ }`,
        );
```
It was complaining about `printWidth` and the `{` placement. I've overcome the `{` by not having a property access, rather storing it in `placeholderName`. Also, I changed number of characters per line in one of the lines to meet the allowed `printWidth`. Let me know if this works

maaz93:
It's generally recommended to run `make test` once locally, after you make a local commit. This will give you an opportunity to see if any tests are failing or not on your system, instead of waiting for the CI. Were you able to figure out why tests in your last commit failed?

thescientist13:
@maaz93 
It looks like it's because of the formatting still (but because the output differs because of how the test / `populatePlaceholders` are constructing the string.  I am sure I will get it, should be able to fix it up in a little bit.

packages/babel-template/test/index.js
```diff
@@ -133,7 +133,13 @@ describe("babel-template", function() {
           ID;
           ANOTHER_ID;
         `)({ ID: t.identifier("someIdent") });
-      }).to.throw(Error, 'No substitution given for "ANOTHER_ID"');
+      }).to.throw(
+        Error,
+        `Error: No substitution given for "ANOTHER_ID". If this is not meant to be a
```
nveenjain: ES6 template literals take raw string, so if you have multi-line string, then it will automatically add new Line character there. However, when you use + (with strings), you are actually doing concatenation and not putting any new Line character there. I would suggest you to convert both of them into template literals, that would automatically add new line character in both. Hence resolving test!

thescientist13: thanks @nveenjain 
I forgot to include it [in my comment](https://github.com/babel/babel/pull/7255#issuecomment-359852487), but locally both implementations are seemingly matching to me in terms of implementation, yet still getting a test failure on string construction.  (I've since updated my comment to show everything)

I can push my changes if it helps.

Andarist: this one is still constructed with multiline template string, you just have to adjust this one so it is constructed like the other one (the other one is correct)

thescientist13: @Andarist 
Do you mean in this PR?  Yes I understand they are different and I know why they are incompatible, but if you see the updated link from [my comment above](https://github.com/babel/babel/pull/7255#discussion_r163566474), locally you will see I have them identical.  Still struggling with the difference between this
```
        const placeholderName = placeholder.name;

        throw new Error(
          `Error: No substitution given for "${placeholderName}". If this is not meant to be a 
            placeholder you may want to consider passing one of the following options to babel-template:
            - { placeholderPattern: false, placeholderWhitelist: new Set(['${placeholderName}'])}
            - { placeholderPattern: /^${placeholderName}$/ }`,
        );
```

and this

```
      }).to.throw(
        Error,
        `Error: No substitution given for "ANOTHER_ID". If this is not meant to be a
          placeholder you may want to consider passing one of the following options to babel-template:
          - { placeholderPattern: false, placeholderWhitelist: new Set(['ANOTHER_ID']) }
          - { placeholderPattern: /^ANOTHER_ID$/ }`,
      );
```

🤒

Andarist: notice that in the source file (when throwing) the string is created (in the PR) with concatenation (`+`) and in tests it's defined with multiline template literal

This:
```js
		const str1 = `first line` +
												`second line`
```
and this:
```js
		const str2 = `first line
												second line`
```
are not equivalents.
```js
str1 === 'first linesecond line'
str2 === first line\n\t\t\t\t\t\t\t\t\t\t\t\tsecond line
```

thescientist13: @Andarist 
Understood that is what is causing the failure in this PR specifically, but my comment was just trying to demonstrate what I was trying locally to fix the failure here but also avoiding pushing intentionally failing commits since I'm not getting it work in the way my comment illustrates.

I suppose that keeping my local work local is causing more confusion, so I will just push my latest changes.

Appreciate you helping out, sorry for any confusion, hopefully the issue I'm seeing will be better highlighted in a few moments.

existentialism: > Error: No substitution given for "${placeholderName}". If this is not meant to be a 

@thescientist13 the difference is an extra space after "a" in your input vs output.

thescientist13: Thanks @existentialism !  I think part of the issue is that what I see in Travis is different output than what I see locally.  Will need to see if I need to align my local environment
- OSX: 10.13.2
- node: 6.9.1
- npm: 5.6.0

existentialism: @thescientist13 i pushed a change to your branch that should fix it

packages/babel-template/src/populate.js
```diff
@@ -15,7 +15,14 @@ export default function populatePlaceholders(
       if (
         !Object.prototype.hasOwnProperty.call(replacements, placeholder.name)
       ) {
-        throw new Error(`No substitution given for "${placeholder.name}"`);
+        const placeholderName = placeholder.name;
+
+        throw new Error(
+          `Error: No substitution given for "${placeholderName}". If this is not meant to be a
```
Andarist: multiline string like this will look weirdly in the console, personally I would prefer using regular concatenation for this

packages/babel-template/test/index.js
thescientist13: wow, thanks @existentialism !!

I'll certainly never underestimate a string again... 😌

packages/babel-template/src/populate.js
thescientist13: fair point.  was too focused on the struggle to step back and consider the alternatives.

hzoo:
Thanks for sticking with this - we should use `@babel/template` instead of `babel-template` now but I'l just fix it up in another commit

Issue #7256, opened 2018-01-23
Optional Chaining Bugs Checklist


jridgewell:
- [x] Delete on optional chain should work `delete a?.b`: https://github.com/babel/babel/pull/7257
- [x] Optional should be removed from `NewExpression` AST
- [x] Optional `MemberExpressions` need to become a new AST node `OptionalChain`, to fix issues with should-not-parse grammars
- [x] Ensure OptionalChain cannot be the LHS of Assignment/Update expressions, tag of a TaggedTemplateLiteral, have a Super object, nor the callee of a NewExpresion

nveenjain:
I would love to work on this issue if available.

jridgewell:
Hi @nveenjain!

Part 1 is already done.

Part 2/3 are closely intertwined. Unfortunately, they're rather difficult to accomplish. It'll take creating a brand new node type, called `OptionalMemberExpression`, which propogates upwards, and can be broken by a non-optional member or call expression (or a parenthesis, because syntax). It'll also likely break things inside the babel transforms, since they now have to handle a new member syntax.

nveenjain:
Well, if the tasks are difficult, i will try to implement them. Incase I'm not able to, i would surely like to review the PR, to watch out for implementation.

nveenjain:
I was going through proposed spec. Don't we need the name of new node be `OptionalExpression` and it's expression property be one of `MemberExpression` or `CallExpression` and an `OptionalChain` node?

jridgewell:
We can name it whatever. Leaving it `OptionalMemberExpression` will likely be easiest for transforms.

`OptionalChainingPunctuator` is just the `??.` sigil.

jridgewell:
Regarding the `OptionalMemberExpression` AST node, it should be:

```js
interface OptionalMemberExpression <: MemberExpression {
  type: "OptionalMemberExpression";
  object: MemberExpression | CallExpression;
  property: Identifier | Expression;
  computed: boolean;
}
```

nveenjain:
@jridgewell what should be `OptionalChain` AST node Interface?

jridgewell:
That is `OptionalChain`. 😉

Spec names don't matter for us. What it's trying to define is two separate member expressions:

```js
obj?.a
```

Is an `OptionalMemberExpression`, with object `Identifier("obj")`, property `Identifier("a")`, computed `false`, and optional `true`.

[...]

PR #7257, opened 2018-01-23
Re-add optional chaining delete

jridgewell:
Fixes part 1 of https://github.com/babel/babel/issues/7256.


packages/babel-plugin-proposal-optional-chaining/test/fixtures/execute/delete.js
```diff
@@ -0,0 +1,22 @@
+"use strict";
```
Andarist: those added tests files are using the same source code, maybe we could put this file as `fixtures/general/delete/exec.js` so if in the future any of those need an update it will be easier to locate others that need an update too

if you decide to do this, I would prefer using the very same code in `input` and `exec` (except asserts of course), that way 2 files can be easily compared when viewed alongside

jridgewell: Done.

PR #7259, opened 2018-01-23
Expose findbabelrc method

rajzshkr:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | 
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-core/src/index.js
+++ b/packages/babel-core/src/index.js
@@ -4,7 +4,7 @@ export { default as File } from "./transformation/file/file";
 export {
   default as buildExternalHelpers,
 } from "./tools/build-external-helpers";
-export { resolvePlugin, resolvePreset } from "./config/files";
+export { resolvePlugin, resolvePreset, findBabelrc } from "./config/files";
 
 export { version } from "../package.json";
 export { getEnv } from "./config/helpers/environment";


Andarist:
Im all in for this change - it would be good though to document **why** this is needed. Most times finding this should be left to babel

nicolo-ribaudo:
I think that we should expose a method which loads the config (merges the applicable `.babelrc`s, handles `overrides`...)

packages/babel-core/src/config/files/configuration.js
```diff
@@ -30,7 +30,7 @@ export function findBabelrc(
   filepath: string,
   envName: string,
 ): ConfigFile | null {
-  const dirname = path.dirname(filepath);
+  const dirname = filepath ? path.dirname(filepath) : process.cwd();
```
loganfsmyth: We could also consider changing this function to accept a `dirname` as the argument. Then you can call

loganfsmyth: That got posted before I intended, sorry.

loganfsmyth: One concern I have about exposing this is that we probably want to change the signature of this function more in the future, and it will be easy to miss that we're breaking a public API. I'm wondering if it would be better to  make a new function right in `src/index.js` that calls through to this. For instance we may add flags to tell a the `.babelrc.js` file whether Babel allows a promise to be returned, and we'll probably add another parameter for passing a `root` folder to stop searching at.

Andarist:
> I think that we should expose a method which loads the config (merges the applicable .babelrcs, handles overrides...)

Yeah, this seems more useful :) still I'd like to discuss exact use cases (we could even list them in the docs and advise people not to use it for regular use cases)

loganfsmyth:
I'm going to close this for now since I don't think it's what we want, but I do think exposing _something_ will be good, and we can continue this discussion in Slack.

PR #7274, opened 2018-01-25
add guard for cannot find decorator of undefined

ajaymathur:
| Documentation PR         | 
| Any Dependency Changes?  |

could get error `cannot find decorators of undefined` in case node is undefined. Could happen if node is deleted, therefore added guard for that.

### DIFF
--- a/packages/babel-plugin-transform-classes/src/vanilla.js
+++ b/packages/babel-plugin-transform-classes/src/vanilla.js
@@ -255,7 +255,7 @@ export default class ClassTransformer {
         throw path.buildCodeFrameError("Missing class properties transform.");
       }
 
-      if (node.decorators) {
+      if (node && node.decorators) {
         throw path.buildCodeFrameError(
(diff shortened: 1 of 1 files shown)


Andarist:
Why would it be deleted? Could you paste in example code and babel config which has resulted in an unexpected error being thrown?

ajaymathur:
this case comes in this scenario, https://github.com/babel/babel/pull/7038

nicolo-ribaudo:
I suspect that it is either a bug with that PR or with babel-traverse, since a NodePath should always have a node.

ajaymathur:
So, In that PR. I am trying to running a loop over all the container nodes transforming them and then removing them from container. Like I don't want to transform the next nodes in container.

loganfsmyth:
I'm triaging old PR on Babel. Since this has been inactive for ages, I'm going to close it. 

As mentioned, the node should be there, so if it is missing, It would be better to file a new bug with reproduction steps for that behavior so we can fix whatever is causing the node to be missing.

PR #7279, opened 2018-01-26
Add notice on constant checks to documentation

maurobringolf:
| Fixed Issues?            |  
| Documentation PR         | yes
| Any Dependency Changes?  |

Document changes from https://github.com/babel/babel/pull/6987 in the `transform-block-scoping` readme.

### DIFF
--- a/packages/babel-plugin-transform-block-scoping/README.md
+++ b/packages/babel-plugin-transform-block-scoping/README.md
@@ -24,6 +24,11 @@ let a = 3
 var a = 3;
 ```
 
+## Constant checks
+
+This plugin also validates all `const` variables.
+Reassignment of constants is a runtime error and it will insert the necessary error code for those.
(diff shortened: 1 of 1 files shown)


packages/babel-plugin-transform-block-scoping/README.md
```diff
@@ -24,6 +24,11 @@ let a = 3
 var a = 3;
 ```
 
+## Constant checks
+
+This plugin also validates all `const` variables.
+Reassignment of constants is a runtime error and it will insert the necessary error code for those.
```
xtuc: Would be cool to add an example IMO: 

```js
const a = true;
if (false) {
    a = "nope";
}
```

hzoo: for sure, always good to have

PR #7288, opened 2018-01-29
[BugFix] : OptionalChaining Bug fixes

### LINKED ISSUE #7256: Optional Chaining Bugs Checklist

nveenjain:
| Fixed Issues?            | Fixes #7256 

Created new node types on babel-types and babylon. OptionalMemberExpressions and OptionalCallExpressions, which while transforming are converted to Normal MemberExpression and CallExpressions (to satisfy transformed AST). Added some tests to babylon and transforms. This will be a breaking change 
1. for those plugins who are using MemberExpression as of now as they'll have to change visitor's entry point
2. for those plugin who convert to OptionalChain right now, as after this commit, MemberExpression will no longer have `optional` property.

Suggestions/edits are welcome.


packages/babylon/src/parser/expression.js
```diff
@@ -454,7 +454,7 @@ export default class ExpressionParser extends LValParser {
       );
     } else if (this.match(tt.questionDot)) {
       this.expectPlugin("optionalChaining");
-
+      state.optionalChainMember = true;
```
Qantas94Heavy: Not sure if this is relevant to this case, but is there any issue with `state` possibly being reused and `optionalChainMember` being true when it shouldn't?

nveenjain: `optinalChainMember` is `true` only after encountering `?.` so i don't think there might be any issue. Once we encounter `?.` then we are sure that the rest of chain will be `optionalChain Member` with optional value = `false` ( meaning that the node is in `optionalChain` but it is not optional(root of optional chain) ). When we encounter ')' like in `(a?.b).c` then the chain is no longer optional and so state doesn't have `optionalChainMember` property for `(a?.b)` object. Should i also add `(a?.b).c` in test cases to make it more clear or can we make it a good first issue for first timers looking for contribution?

nveenjain: @Qantas94Heavy , i added additional test cases, please take a look. Maybe that clear the confusion or is there still some doubt?

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -128,8 +140,12 @@ export default function(api, options) {
     inherits: syntaxOptionalChaining,
 
     visitor: {
-      "MemberExpression|CallExpression"(path) {
+      "OptionalCallExpression|OptionalMemberExpression"(path) {
         if (!path.node.optional) {
+          path.node.type =
```
jridgewell: We can simplify the `findReplacementPath` by not changing the node type here.

packages/babylon/src/parser/expression.js
```diff
@@ -454,7 +454,7 @@ export default class ExpressionParser extends LValParser {
...
       if (noCalls && this.lookahead().type == tt.parenL) {
```
jridgewell: I think we might need to throw here to prevent `new fn?.()` and `new obj?.fn()`

packages/babylon/test/fixtures/experimental/optional-chaining/class-contructor-call/input.js
```diff
@@ -1,9 +1 @@
-new C?.()
-
-new C?.(a, b)
-
-new B?.C?.()
-
-new B?.C?.(a, b)
-
-new B?.C
+new C?.b.d()
```
jridgewell: We also need a test for `new C?.()`

packages/babylon/test/fixtures/experimental/optional-chaining/optional-constructor/input.js
```diff
@@ -0,0 +1 @@
+new a?.();
```
nveenjain: @jridgewell, Is this test case OK?

nveenjain: @jridgewell , is this test case OK?

nveenjain:
Also @jridgewell , since issue #7256 didn't have super/tagged template literal as it's key points, i've not implemented them in this PR. Do you mind if I open another issue stating them, and then make a PR??

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -128,7 +122,8 @@ export default function(api, options) {
     inherits: syntaxOptionalChaining,
 
     visitor: {
-      "MemberExpression|CallExpression"(path) {
+      "OptionalCallExpression|OptionalMemberExpression"(path) {
+        debugger;//eslint-disable-line
```
jridgewell: Remove.

packages/babylon/test/fixtures/experimental/optional-chaining/optional-constructor/input.js
jridgewell: Yup.

jridgewell:
> Do you mind if I open another issue stating them (super?.key), and then make a PR??

Works for me.

packages/babel-plugin-proposal-optional-chaining/src/index.js
nveenjain: Sorry for that :sweat_smile: , test cases were passing and i completely forgot that lived here.

PR #7291, opened 2018-01-29
babel-core: Add parse method

kaicataldo:
| Fixed Issues?            | Fixes https://github.com/babel/babel-eslint/issues/573
| Documentation PR         |  Not yet, but will add before this is merged.
| Any Dependency Changes?  | No,

This is an initial implementation so we can hammer out the details. I'll be more than happy to add docs once we've finalized what this should look like. Thanks!

### DIFF
--- a/packages/babel-core/src/parse.js
+++ b/packages/babel-core/src/parse.js
@@ -1,0 +1,22 @@
+// @flow
+
+import loadConfig, { InputOptions } from "./config";
+import normalizeFile from "./transformation/normalize-file";
+import normalizeOptions from "./transformation/normalize-opts";
+
+type AstRoot = BabelNodeFile | BabelNodeProgram;
+
+export default function parse(
+  code: string,
+  opts: InputOptions,
+): AstRoot | null {
+  const config = loadConfig(opts);
+
+  if (config === null) {
+    return null;
+  }
+
+  const file = normalizeFile(config.passes, normalizeOptions(config), code);
+
+  return file.ast;
+}
--- a/packages/babel-core/src/index.js
+++ b/packages/babel-core/src/index.js
@@ -43,6 +43,8 @@ export { default as transformFileSync } from "./transform-file-sync";
 export { default as transformFromAst } from "./transform-ast";
 export { default as transformFromAstSync } from "./transform-ast-sync";
 
+export { default as parse } from "./parse";
+
 /**
  * Recommended set of compilable extensions. Not used in @babel/core directly, but meant as
  * as an easy source for tooling making use of @babel/core.
--- a/packages/babel-core/test/fixtures/parse/.babelrc
+++ b/packages/babel-core/test/fixtures/parse/.babelrc
@@ -1,0 +1,3 @@
+{
+  "plugins": ["@babel/syntax-decorators"]
+}
--- a/packages/babel-core/test/fixtures/parse/input.js
+++ b/packages/babel-core/test/fixtures/parse/input.js
@@ -1,0 +1,6 @@
+@annotation
+class MyClass { }
+
+function annotation(target) {
+   target.annotated = true;
+}
--- a/packages/babel-core/test/fixtures/parse/output.json
+++ b/packages/babel-core/test/fixtures/parse/output.json
@@ -1,0 +1,286 @@
+{
+  "type": "File",
+  "start": 0,
+  "end": 90,
+  "loc": {
+    "start": {
+      "line": 1,
+      "column": 0
+    },
+    "end": {
+      "line": 6,
+      "column": 1
+    }
+  },
+  "program": {
+    "type": "Program",
+    "start": 0,
+    "end": 90,
+    "loc": {
+      "start": {
+        "line": 1,
+        "column": 0
+      },
+      "end": {
+        "line": 6,
+        "column": 1
+      }
+    },
+    "sourceType": "module",
+    "body": [
+      {
+        "type": "ClassDeclaration",
+        "start": 0,
+        "end": 29,
+        "loc": {
+          "start": {
+            "line": 1,
+            "column": 0
+          },
+          "end": {
+            "line": 2,
+            "column": 17
+          }
+        },
+        "decorators": [
+          {
+            "type": "Decorator",
+            "start": 0,
+            "end": 11,
+            "loc": {
+              "start": {
+                "line": 1,
+                "column": 0
+              },
+              "end": {
+                "line": 1,
+                "column": 11
+              }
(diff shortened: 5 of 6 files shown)


packages/babel-core/src/parse.js
```diff
@@ -0,0 +1,22 @@
...
+
+import loadConfig, { InputOptions } from "./config";
+import normalizeFile from "./transformation/normalize-file";
+import normalizeOptions from "./transformation/normalize-opts";
+
+type AstRoot = BabelNodeFile | BabelNodeProgram;
+
+export default function parse(
+  code: string,
+  opts?: InputOptions,
```
kaicataldo: Is this the API that we want? Technically, this should require either a `filename` property or the desired configuration, so I suppose it's not really optional.

packages/babel-core/src/parse.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+
+import loadConfig, { InputOptions } from "./config";
+import normalizeFile from "./transformation/normalize-file";
```
kaicataldo: I reused `normalizeFile` and `normalizeOptions` from the transformation utils to avoid duplication of code and to make sure we're processing everything the same way. Does this make sense?

packages/babel-core/src/parse.js
kaicataldo: Maybe we should add some warnings at the top of this function to help a user debug if the above requirements are not met.

loganfsmyth: Personally I'd love to change

https://github.com/babel/babel/blob/252ea5a966c1968d8aac21a1a81c6d45173e57dd/packages/babel-core/src/config/config-chain.js#L126

to
```
  if (opts.babelrc !== false) {
    if (context.filename === null) {
      throw new Error("Babel was not told to skip babelrc files, but was given no filename to search from");
    const filename = context.filename;
```

loganfsmyth: On the fence about if that would be too aggressive, but it seems like a case where it would help users, so they'd have to explicitly pass `babelrc: false` if they didn't pass in a filename.

kaicataldo: Would you like that change to happen in this PR?

packages/babel-core/src/parse.js
```diff
@@ -0,0 +1,22 @@
...
+  code: string,
+  opts?: InputOptions,
+): AstRoot | null {
+  const config = loadConfig(opts);
+
+  if (config === null) {
+    return null;
+  }
+
+  const file = normalizeFile(config.passes, normalizeOptions(config), code);
```
loganfsmyth: Is there any other logic in these that we'd want to avoid? I don't know off the top of my head.

packages/babel-core/src/parse.js
```diff
@@ -0,0 +1,22 @@
...
+
+type AstRoot = BabelNodeFile | BabelNodeProgram;
+
+export default function parse(
+  code: string,
+  opts?: InputOptions,
+): AstRoot | null {
+  const config = loadConfig(opts);
+
+  if (config === null) {
```
loganfsmyth: This was the main question I had when thinking through this originally. I wonder if it might be better do do `config || {}` so if a file was ignore, it is just like no config was given, instead of returning `null` since that's not that useful from what I can see. Tempted to do the same for the rest of the Babel functions.

packages/babel-core/src/parse.js
loganfsmyth: I'm up for it if it sounds good to you. The only thing I'll say is I think we should be moving toward throwing clearer errors, so I'd vote for probably like

```
  if (context.filename === null) {
    throw Object.assign(new Error("Babel was not told to skip babelrc files, but was given no filename to search from");
  }
```

packages/babel-core/src/parse.js
kaicataldo: Good call - it might be better not to do all this extra stuff. My only question is if [this](https://github.com/babel/babel/blob/master/packages/babel-core/src/transformation/normalize-file.js#L64-L89) is needed. Other than that, I should be to just call `babylon.parse` directly with the normalized config's `parserOpts`.

packages/babel-core/src/parse.js
kaicataldo: So would this then try to parse without a config, potentially throwing errors if it encounters syntax that it needs a plugin for?

loganfsmyth: Yeah IDK, I've been thinking about this more and I think I'm changing back to this being fine as-is. If Babel returns `null` them maybe ESLint can just use its default parser and we go from there?

packages/babel-core/src/parse.js
loganfsmyth: Hmm, I feel like that code probably should run so that we avoid differences between the `parse` and `transform`? Then we can aim for having `babel.parse()` basically be a nicer, faster version of `babel.transform` with no plugins?

kaicataldo: Sure, makes sense! This could always be optimized in the future.

packages/babel-core/src/parse.js
kaicataldo: We can figure that out on the babel-eslint side. If this is the desired API for Babel, I think babel-eslint has a few options:

- Throw an error and show a helpful warning about needing a .babelrc or needing to define parserOptions in the ESLint config or to be passed on the command line.
- Fallback to parsing with Babylon with the current [default list of plugins](https://github.com/babel/babel-eslint/blob/master/lib/parse.js#L18-L45) (may not be up to date). This would make the fallback be essentially the behavior it has now.
- Fallback to trying to parse with Espree (ESLint's default parser). Not sure how I feel about this, because a lot of things might work/not work by accident, and I think this will end up being really confusing for users.

hzoo: yeah 1 or 3 makes sense. Not sure what case you would use it without a config (I guess you might if you have a preset and use the cli directly like `babel src -d lib --preset your-preset`), so maybe 3 but then ESLint wouldn't pick up any syntax lol.

packages/babel-core/src/parse.js
kaicataldo: I'm having trouble getting Flow to cooperate here - will leave it for now and try to pick it up in a future PR.

packages/babel-core/src/parse.js
existentialism: I'm initially leaning towards number 1

kaicataldo:
Yes, happy to make a PR for docs!

PR #7295, opened 2018-01-30
"babel-standalone": broadcast "BabelScriptsRun" event

warren-bank:
..after all script tags are transformed and run

* the 1st commit implements the feature
* the 2nd commit adds:
  * a simplified build script to generate a browser-build for "babel-standalone"
  * a [webpage](https://cdn.rawgit.com/warren-bank/babel/PR-02-standalone-BabelScriptsRun-event/scripts/browser-build/test/index.html
) that uses the newly built version of "babel-standalone" to:
    * transpile the unit tests for "babel-standalone" in Mocha w/ Chai
    * wait for the event to fire
    * update the status in the DOM and begin the Mocha test runner


warren-bank:
I'm not sure if this PR will get any traction, but it seems that when transpiling code in the browser.. there really should be an event fired to signal when the translation is complete and all the ES5 code has been run.

Using the unit tests that already exist for "babel-standalone" serves as an excellent example for this need. Mocha needs to wait until the `describe()` blocks have run.

warren-bank:
I'm far from an expert on Mocha or Chai, but:

* I just tested: `npm install --production mocha chai`
  * which resulted in a `node_modules` directory containing 32 modules, 313 files, and no apparent (webpack-generated) browser builds

Even if it were possible to obtain the vendor bundles from a package management system, I included them directly into my repo for the purpose that the test webpage is completely off-line and standalone. This also enables it to be hosted directly from the repo, and executed directly in the web browser.. without requiring any additional steps to build it.

That was my reason for including it. You may choose the entire 2nd commit is unnecessary. Personally, I find all of the contents in "scripts/browser-bundle" to be useful.. though `scripts/browser-bundle/dist/babel.js` (ie: the non-minified version of "babel-standalone" that is not needed by the SPA) could be added to `scripts/browser-build/.gitignore` to save a little space.

warren-bank:
That being said.. all of the scripts in `scripts/browser-build/test/vendor/*` could be obtained from a public CDN, rather than being mirrored in the repo. It would make the SPA require internet access and be dependent upon the CDN.. but that would remove the "vendor" files from the repo.

warren-bank:
I just had a thought.. unrelated to the above..

in `packages/babel-standalone/src/transformScriptTags.js`
* would it be a good idea to wrap `window.document.dispatchEvent(event)` in a `setTimeout(0)` so it occurs on the next iteration of the event loop?
* I was thinking it may be a good idea to give the browser time to parse and execute the <script> tag that was added to the DOM
* I assume that occurs in a different thread than the javascript engine.. so one tick may not even be enough of a guarantee

Thoughts?

warren-bank:
anecdotally.. the SPA test runner doesn't show any sign of it being a problem

warren-bank:
which actually raises the question..

is it better to inject the transpiled ES5 javascript into a `<script>` DOM element and then appending that element to `<head>`, or..

would it be more efficient (and synchronous) to execute it directly (via either `exec()` or `new Function()` or similar?

Issue #7304, opened 2018-01-31
[preset-env] Duplicate declaration "object"


julien-f:
### Is this a bug report or feature request?

Bug

[Repro repo](https://github.com/julien-f/issue-babel-env-duplicate).

### Input Code

```js
export default class {
  method ({ ...object }) {}
}
```

### Babel/Babylon Configuration (.babelrc, package.json, cli command)

```js
{
  "presets": [
    [
      "@babel/env",
      {
        "shippedProposals": true,
        "targets": {
          "node": 8
        },
        "useBuiltIns": "usage"
      }
    ]
  ]
}
```

### Expected Behavior

Should compile.

### Current Behavior

```
TypeError: index.js: Duplicate declaration "object"
  1 | export default class {
> 2 |   method ({ ...object }) {}
    |                ^
  3 | }
  4 |
    at File.buildCodeFrameError (issue-babel-env-duplicate/node_modules/@babel/core/lib/transformation/file/file.js:209:12)
    at Scope.checkBlockScopedCollisions (issue-babel-env-duplicate/node_modules/@babel/traverse/lib/scope/index.js:303:27)
    at Scope.registerBinding (issue-babel-env-duplicate/node_modules/@babel/traverse/lib/scope/index.js:485:16)
    at Scope.registerDeclaration (issue-babel-env-duplicate/node_modules/@babel/traverse/lib/scope/index.js:404:14)
    at Object.BlockScoped (issue-babel-env-duplicate/node_modules/@babel/traverse/lib/scope/index.js:146:28)
    at Object.newFn (issue-babel-env-duplicate/node_modules/@babel/traverse/lib/visitors.js:266:17)
    at NodePath._call (issue-babel-env-duplicate/node_modules/@babel/traverse/lib/path/context.js:64:19)
    at NodePath.call (issue-babel-env-duplicate/node_modules/@babel/traverse/lib/path/context.js:34:14)
    at NodePath.visit (issue-babel-env-duplicate/node_modules/@babel/traverse/lib/path/context.js:99:12)
    at TraversalContext.visitQueue (issue-babel-env-duplicate/node_modules/@babel/traverse/lib/context.js:139:18)
```

### Your Environment

| software         | version(s)
| ---------------- | -------
| Babel            |  7.0.0-beta.39
| Babylon          | 
| node             | 9.4.0
| npm              | 
| Operating System |

rajzshkr:
[...]

PR #7305, opened 2018-01-31
Failing test involving object rest/spread and clearScope().

benjamn:
| Fixed Issues?            | Possibly #7304, once this test passes
| Patch: Bug Fix?          | Not yet; I hope to push a fix if I can figure it out
| Tests Added + Pass?      | Added a failing test

This failing test case demonstrates a regression between `7.0.0-beta.38` and `7.0.0-beta.39` in the [`@babel/plugin-proposal-object-rest-spread`](https://github.com/babel/babel/tree/eb38ea2b10d07b3451e4f68b6505cb4730aa5311/packages/babel-plugin-proposal-object-rest-spread) package.

I distilled this test case from a larger configuration of plugins in my application, one of which calls `api.traverse.cache.clearScope()`. Although calling `clearScope()` is an uncommon thing for a plugin to do, it was a reliable way to reproduce the problem. If I can find other reliable reproductions, I'll push some additional failing tests to this PR.

Regardless of how common it is, clearing the scope cache should be a safe operation that only slows down the transform (because scopes have to be recreated and re-crawled). Crashing due to a spurious duplicate declaration seems like a bug worth fixing.

My hunch is that [these two lines](https://github.com/babel/babel/blob/eb38ea2b10d07b3451e4f68b6505cb4730aa5311/packages/babel-plugin-proposal-object-rest-spread/src/index.js#L75-L76) (which were changed in `7.0.0-beta.39`) are not actually removing the original rest element as a binding from the enclosing `Scope`, in certain circumstances, so the new variable declaration ends up colliding with the old (removed) binding.

cc @nicolo-ribaudo & @Andarist who worked on #7149

Possibly related: #7304 (reported by @julien-f)

Here's the error:
```
% TEST_GREP=object-rest-spread make test-only
./scripts/test.sh


  ․․!․․․․․․․․․․․․․․․․․․․․․․

  24 passing (312ms)
  1 failing

  1) babel-plugin-proposal-object-rest-spread/object rest duplicate decl bug:
     TypeError: /Users/ben/dev/babel/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/duplicate-decl-bug/input.js: Duplicate declaration "copy"
  1 | it("es7.objectRestSpread", () => {
  2 |   let original = { a: 1, b: 2 };
> 3 |   let { ...copy } = original;
    |            ^
  4 | });
      at File.buildCodeFrameError (packages/babel-core/lib/transformation/file/file.js:207:12)
      at Scope.checkBlockScopedCollisions (packages/babel-traverse/lib/scope/index.js:303:27)
      at Scope.registerBinding (packages/babel-traverse/lib/scope/index.js:485:16)
      at Scope.registerDeclaration (packages/babel-traverse/lib/scope/index.js:404:14)
      at Object.BlockScoped (packages/babel-traverse/lib/scope/index.js:146:28)
      at Object.newFn (packages/babel-traverse/lib/visitors.js:266:17)
      at NodePath._call (packages/babel-traverse/lib/path/context.js:64:19)
      at NodePath.call (packages/babel-traverse/lib/path/context.js:34:14)
      at NodePath.visit (packages/babel-traverse/lib/path/context.js:99:12)
      ...
```

### DIFF
--- a/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/duplicate-decl-bug/input.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/duplicate-decl-bug/input.js
@@ -1,0 +1,4 @@
+it("es7.objectRestSpread", () => {
+  let original = { a: 1, b: 2 };
+  let { ...copy } = original;
+});
--- a/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/duplicate-decl-bug/options.json
+++ b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/duplicate-decl-bug/options.json
@@ -1,0 +1,7 @@
+{
+  "plugins": [
+    "./plugin-clear-scope",
+    "proposal-object-rest-spread",
+    "external-helpers"
+  ]
+}
--- a/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/duplicate-decl-bug/output.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/duplicate-decl-bug/output.js
@@ -1,0 +1,7 @@
+it("es7.objectRestSpread", () => {
+  let original = {
+    a: 1,
+    b: 2
+  };
+  let copy = babelHelpers.objectWithoutProperties(original, []);
+});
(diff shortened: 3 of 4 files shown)


nicolo-ribaudo:
@benjamn I included your commit in https://github.com/babel/babel/pull/7326.

PR #7308, opened 2018-02-01
Added license files to packages

Bamieh:

Babel packages do not have a LICENSE file on them. It is a good practice to have one per package. I added a bash script to copy the main license file into the packages.

```
make clone-license
```

Dependency watchers like [versioneye](https://www.versioneye.com/) show UNKNOWN license to the new babel, this will probably solve such issues as well.

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,6 +110,8 @@ test-test262-ci:
 test-test262-update-whitelist:
 	node scripts/tests/test262/run_babylon_test262.js --update-whitelist
 
+clone-license:
+	./scripts/clone-license.sh
 publish:
 	git pull --rebase
 	make clean-lib
--- a/packages/babel-cli/LICENSE
+++ b/packages/babel-cli/LICENSE
@@ -1,0 +1,22 @@
+MIT License
+
+Copyright (c) 2014-2018 Sebastian McKenzie <sebmck@gmail.com>
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
+without limitation the rights to use, copy, modify, merge, publish,
+distribute, sublicense, and/or sell copies of the Software, and to
+permit persons to whom the Software is furnished to do so, subject to
+the following conditions:
+
+The above copyright notice and this permission notice shall be
+included in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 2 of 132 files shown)


lukescott:
I'm not sure a LICENSE file should be committed to each and every package. The root LICENSE is enough for the project repo.

However, you could copy each LICENSE file just before the publish step and then remove them afterward. That way each published npm package has one. You should probably also have `/packages/*/LICENSE` in the gitignore to make sure they aren't accidentally committed.

There is a lerna issue that mentions this problem https://github.com/lerna/lerna/issues/1213.

hzoo:
Is it necessary to add the actual files? it's also in the package.json already

Bamieh:
@lukescott no need to remove after adding them if they are not committed. However I'm not sure if we should not commit them to github, since some might have different things (some LICENSE files are already in some packages already).

@hzoo the `package.json` file does have the name of the license only. It's usually a better idea to include the LICENSE file for whatever code you download (via npm or github)

lukescott:
I think it's going to depend on a companies legal team.

Bamieh:
@lukescott legally, it is not satisfactory to have the license name alone in the package.

This is also explicitly required by some license in order for it to take effect, such as `ICS` or `Apache 2.0`, all `GNU` licenses and **`MIT`**.

> The MIT license requires that "the above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software." By including the license in the source file, you have met this obligation.

Not specifying the license text in each package restrictedly means that no one may reproduce, distribute, or create derivative works from your work. **The name of the license alone does not "apply" the license, hence it retains the default license instead.**

> without a license, the default copyright laws apply, meaning that you retain all rights to your source code and no one may reproduce, distribute, or create derivative works from your work.

hzoo:
Ok I think we should just do this at publish time then, this seems to be what other projects do like React https://github.com/facebook/react/tree/master/packages/react

> However, you could copy each LICENSE file just before the publish step and then remove them afterward. That way each published npm package has one. You should probably also have /packages/*/LICENSE in the gitignore to make sure they aren't accidentally committed.

This sounds good to me

Bamieh:
@hzoo cool. i'll update the code then

klimashkin:
Hi @Bamieh, any updates on that?
I would be really appreciate for that PR, since tools like license-checker still [can't pick up](https://github.com/davglass/license-checker/issues/134#issuecomment-358654097) packages licenses automatically until they are shipped

Bamieh:
@klimashkin wow I completely forgot about this, I'll have it ready today

opichals:
@Bamieh Gave the script-only solution a shot in https://github.com/babel/babel/pull/8409

hzoo:
Done in https://github.com/babel/babel/pull/8409

PR #7309, opened 2018-02-01
[BugFix] Compile export default from

### LINKED ISSUE #7293: 'export default from' can not be correctly compiled
Choose one: bug

### Input Code

```js
var from = 'test';
export default from;
```

```js
{
  presets: [
        [
          '@babel/preset-env', {
            targets: { browsers: ['android>=4', 'ios>=7'] },
            modules: false,
          },
        ],
      ],
      plugins: [
        '@babel/plugin-transform-runtime',
      ],
}
```

### Expected Behavior
Compiled correctly

### Current Behavior
Throw an error
```bash
[!] (babel plugin) SyntaxError: /xxx/node_modules/@babel/runtime/core-js/array/from.js: Support for the experimental syntax 'exportDefaultFrom' isn't currently enabled (6:8):

  4 | var from = require$$0;
  5 |
> 6 | export default from;
    |        ^
  7 | export { from as __moduleExports };

```

| ---------------- | -------
| Babel            |  7.0.0-beta.38|
| Babylon          | 
| node             | 9.4 |
| npm              | 
| Operating System | mac os|

nveenjain:
| Fixed Issues?            | Fixes #7293 
| Tests Added + Pass?      | NO
| Documentation PR         | 

Now we check in export default statements that if from is either an assignment or  from is exported then valid AST are generated, which doesn't require any sort of plugin.

### DIFF
--- a/packages/babylon/src/parser/statement.js
+++ b/packages/babylon/src/parser/statement.js
@@ -1310,21 +1310,44 @@ export default class StatementParser extends ExpressionParser {
       this.parseExportStar(node);
       if (node.type === "ExportAllDeclaration") return node;
     } else if (this.isExportDefaultSpecifier()) {
-      this.expectPlugin("exportDefaultFrom");
       const specifier = this.startNode();
       specifier.exported = this.parseIdentifier(true);
+      let isDefaultfrom = false;
+      if (specifier.exported.name === "default" && this.isContextual("from")) {
+        isDefaultfrom = true;
+        const lookaheadState = this.lookahead();
+        if (
+          lookaheadState.value === "=" ||
+          lookaheadState.type === tt.semi ||
+          lookaheadState.type === tt.eof ||
+          lookaheadState.type === tt.braceR ||
+          lineBreak.test(
+            this.input.slice(lookaheadState.lastTokEnd, lookaheadState.start),
+          )
+        ) {
+          node.declaration = this.parseMaybeAssign();
+          this.checkExport(node, true, true);
+          return this.finishNode(node, "ExportDefaultDeclaration");
+        }
+      }
+      this.expectPlugin("exportDefaultFrom");
       const specifiers = [this.finishNode(specifier, "ExportDefaultSpecifier")];
       node.specifiers = specifiers;
-      if (this.match(tt.comma) && this.lookahead().type === tt.star) {
-        this.expect(tt.comma);
-        const specifier = this.startNode();
-        this.expect(tt.star);
-        this.expectContextual("as");
-        specifier.exported = this.parseIdentifier();
-        specifiers.push(this.finishNode(specifier, "ExportNamespaceSpecifier"));
-      } else {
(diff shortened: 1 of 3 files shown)


packages/babylon/src/parser/statement.js
```diff
@@ -1310,21 +1310,44 @@ export default class StatementParser extends ExpressionParser {
...
     } else if (this.isExportDefaultSpecifier()) {
-      this.expectPlugin("exportDefaultFrom");
       const specifier = this.startNode();
       specifier.exported = this.parseIdentifier(true);
+      let isDefaultfrom = false;
+      if (specifier.exported.name === "default" && this.isContextual("from")) {
+        isDefaultfrom = true;
+        const lookaheadState = this.lookahead();
+        if (
+          lookaheadState.value === "=" ||
```
nicolo-ribaudo: What if I have another token? e.g. `export default from + 2;`

nveenjain: Yeah Sorry, didn't consider that.. Will update the PR

nicolo-ribaudo:
I just read the spec and it looks like this is disallowed? :thinking:
cc @leaves4j

https://tc39.github.io/proposal-export-default-from/#prod-ExportDeclaration
```
export default [lookahead ∉ { function, class, from }] AssignmentExpression[In];


> NOTE 1
The addition of from to the lookahead after export default allows for the disambiguation
of export default from "mod", which is the proposed syntactic form for exporting
"mod"'s default export as this module's default export, from export default from, which
is the potentially confusing export of a local named variable from.
```

nveenjain:
but since it is still in proposal, should i work on this PR or not? like first we should check whether `export-default-from` plugin is used or not.
Shouldn't export default from work without `export-default-from` plugin and don't work with `export-default-from` plugin enabled?

nicolo-ribaudo:
For now I think we should wait: even if it is currently allowed, we should prevent users from depending on `export default from;`.

Usually the spec doesn't includes breaking changes, but since people who are using modules are also using Babel (because native esm support is really new), that change wouldn't break anyone's existing code and thus I think tc39 folks could decide do make it.

cc @leebyron I'd like to know your opinion, since you are the author of the proposal 🙂

stephenfriedman:
Any status update with this merge request? I believe that this issue may arise for folks who are attempting to run the [codemod script ](https://github.com/mui-org/material-ui/tree/master/packages/material-ui-codemod) when updating to material-ui newest 1.x.x version from the previous 0.x.x version (which is presumably a lot of projects that will undergo this upgrade inevitably)

leebyron:
I believe @jdalton has taken over for these. I’m still in support of them, but have not been attending TC39 meetings and don’t know their status

vzaidman:
bump

JLHwung:
@nveenjain Although this PR is closed by #11676, I want to thank you for your inspiring work here!

PR #7313, opened 2018-02-02
Added babel-helper-split-export-declaration

Andarist:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR         | 
| Any Dependency Changes?  | internal

Extracted from - https://github.com/babel/babel/pull/6963

cc @nicolo-ribaudo

### DIFF
--- a/packages/babel-helper-module-transforms/src/rewrite-live-references.js
+++ b/packages/babel-helper-module-transforms/src/rewrite-live-references.js
@@ -112,6 +112,7 @@ const rewriteBindingInitVisitor = {
 
     Object.keys(path.getOuterBindingIdentifiers()).forEach(localName => {
       const exportNames = exported.get(localName) || [];
+
       if (exportNames.length > 0) {
         const statement = t.expressionStatement(
           buildBindingExportAssignmentExpression(
--- a/packages/babel-helper-split-export-declaration/README.md
+++ b/packages/babel-helper-split-export-declaration/README.md
@@ -1,0 +1,23 @@
+# @babel/helper-split-export-declaration
+
+## API
+
+```js
+declare export default splitExportDeclaration(path: NodePath);
+```
+
+## Usage
+
+```js
+import traverse from "@babel/traverse";
+import splitExportDeclaration from "@babel/helper-split-export-declaration";
+
+// ...
+
+traverse(file, {
+  ExportDefaultDeclaration(path) {
+    if (!path.get("declaration").isClassDeclaration()) return;
+    splitExportDeclaration(path);
+  },
+});
+```
--- a/packages/babel-helper-split-export-declaration/src/index.js
+++ b/packages/babel-helper-split-export-declaration/src/index.js
@@ -1,0 +1,76 @@
+import * as t from "@babel/types";
+
+export default function splitExportDeclaration(exportDeclaration) {
+  if (!exportDeclaration.isExportDeclaration()) {
+    throw new Error("Only export declarations can be splitted.");
+  }
+
+  // build specifiers that point back to this export declaration
+  const isDefault = exportDeclaration.isExportDefaultDeclaration();
+  const declaration = exportDeclaration.get("declaration");
+  const isClassDeclaration = declaration.isClassDeclaration();
+
+  if (isDefault) {
+    const standaloneDeclaration =
+      declaration.isFunctionDeclaration() || isClassDeclaration;
+
+    const scope = declaration.isScope()
+      ? declaration.scope.parent
+      : declaration.scope;
+
+    let id = declaration.node.id;
+    let needBindingRegistration = false;
+
+    if (!id) {
+      needBindingRegistration = true;
+
+      id = scope.generateUidIdentifier("default");
+
+      if (
+        standaloneDeclaration ||
+        declaration.isFunctionExpression() ||
+        declaration.isClassExpression()
+      ) {
+        declaration.node.id = t.cloneNode(id);
+      }
+    }
+
+    const updatedDeclaration = standaloneDeclaration
+      ? declaration
+      : t.variableDeclaration("var", [
+          t.variableDeclarator(t.cloneNode(id), declaration.node),
+        ]);
+
+    const updatedExportDeclaration = t.exportNamedDeclaration(null, [
+      t.exportSpecifier(t.cloneNode(id), t.identifier("default")),
+    ]);
+
+    exportDeclaration.insertAfter(updatedExportDeclaration);
+    exportDeclaration.replaceWith(updatedDeclaration);
+
+    if (needBindingRegistration) {
+      scope.registerBinding(
+        isClassDeclaration ? "let" : "var",
+        exportDeclaration,
+      );
+    }
+
+    return exportDeclaration;
+  }
+
+  if (exportDeclaration.get("specifiers").length > 0) {
+    throw new Error("It doesn't make sense to split exported specifiers.");
+  }
+
+  const bindingIdentifiers = declaration.getOuterBindingIdentifiers();
+
+  const specifiers = Object.keys(bindingIdentifiers).map(name => {
+    return t.exportSpecifier(t.identifier(name), t.identifier(name));
+  });
+
+  const aliasDeclar = t.exportNamedDeclaration(null, specifiers);
+
+  exportDeclaration.insertAfter(aliasDeclar);
+  exportDeclaration.replaceWith(declaration.node);
+  return exportDeclaration;
+}
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -210,7 +210,7 @@ export default class Scope {
    * Generate a unique identifier and add it to the current scope.
    */
 
-  generateDeclaredUidIdentifier(name: string = "temp") {
+  generateDeclaredUidIdentifier(name?: string) {
     const id = this.generateUidIdentifier(name);
     this.push({ id });
     return t.cloneNode(id);
@@ -220,7 +220,7 @@ export default class Scope {
    * Generate a unique identifier.
    */
 
-  generateUidIdentifier(name: string = "temp") {
+  generateUidIdentifier(name?: string) {
     return t.identifier(this.generateUid(name));
   }
 
--- a/packages/babel-helper-module-transforms/package.json
+++ b/packages/babel-helper-module-transforms/package.json
@@ -10,6 +10,7 @@
   "dependencies": {
     "@babel/helper-module-imports": "7.0.0-beta.40",
     "@babel/helper-simple-access": "7.0.0-beta.40",
+    "@babel/helper-split-export-declaration": "7.0.0-beta.40",
     "@babel/template": "7.0.0-beta.40",
     "@babel/types": "7.0.0-beta.40",
     "lodash": "^4.2.0"
--- a/packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js
+++ b/packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js
@@ -1,6 +1,6 @@
 import { basename, extname } from "path";
 
-import * as t from "@babel/types";
+import splitExportDeclaration from "@babel/helper-split-export-declaration";
 
 export type ModuleMetadata = {
   exportName: string,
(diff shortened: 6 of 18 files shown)


packages/babel-helper-split-export-declaration/src/index.js
```diff
@@ -0,0 +1,65 @@
...
+      ? declaration
+      : t.variableDeclaration("var", [
+          t.variableDeclarator(t.cloneNode(id), declaration.node),
+        ]);
+
+    const updatedExportDeclaration = liveBinding
+      ? t.exportNamedDeclaration(null, [
+          t.exportSpecifier(t.cloneNode(id), t.identifier("default")),
+        ])
+      : t.exportDefaultDeclaration(t.cloneNode(id));
```
Andarist: @loganfsmyth  [commented](https://github.com/babel/babel/pull/6963#discussion_r164818661) on this:
> Even if it doesn't have an ID, cyclic dependencies can still rely on it being hoisted. It's never safe to do export default foo; in this case.

Andarist: I'm not sure when that could have an impact? is it about smth like this?
```js
export default (() => console.log('side effect'))()
```

nicolo-ribaudo: It might be about something like https://github.com/rollup/rollup/issues/1784

loganfsmyth:
```

Comment (loganfsmyth) [reply packages/babel-helper-split-export-declaration/src/index.js]:
```
// index.js
import "./a";

export default function(){
    return 4;
}


// a.js
import getFour from "./index";

getFour();
```

File: packages/babel-helper-split-export-declaration/src/index.js [hunk 2]
```diff
@@ -0,0 +1,65 @@
...
+    throw new Error("It doesn't make sense to split exported specifiers.");
+  }
+
+  const bindingIdentifiers = declaration.getOuterBindingIdentifiers();
+
+  const specifiers = Object.keys(bindingIdentifiers).map(name => {
+    return t.exportSpecifier(t.identifier(name), t.identifier(name));
+  });
+
+  if (specifiers.length) {
```
loganfsmyth: If there aren't specificers, should we throw an exception? Seems like there should always be some. If this is indented to handle `export { foo } from "bar";` we should probably bail out if there is a `source` instead.

packages/babel-helper-split-export-declaration/README.md
```diff
@@ -0,0 +1,23 @@
+# @babel/helper-split-export-declaration
+
+## API
+
+```js
+declare export default splitExportDeclaration(path: NodePath);
```
loganfsmyth: `declare`?

packages/babel-helper-split-export-declaration/src/index.js
```diff
@@ -0,0 +1,65 @@
...
+    throw new Error("Only export declarations can be splitted.");
+  }
+
+  // build specifiers that point back to this export declaration
+  const isDefault = exportDeclaration.isExportDefaultDeclaration();
+  const declaration = exportDeclaration.get("declaration");
+
+  if (isDefault) {
+    if (declaration.isIdentifier()) {
+      throw new Error("It doesn't make sense to split exported identifier.");
```
loganfsmyth: Is this strictly necessary? It seems like we'd be able to change the logic for generating CmmonJS separately so

```
class Foo {}
export { Foo };
```

still transforms to something readable, even when it is split.

packages/babel-helper-module-transforms/src/rewrite-live-references.js
```diff
@@ -29,13 +30,24 @@ export default function rewriteLiveReferences(
   }
 
   for (const [local, data] of metadata.local) {
+    const names = data.bindings.map(({ name }) => name);
```
loganfsmyth: Could we 
Could we avoid all of this if we adjusted `rewriteBindingInitVisitor` instead? e.g.

```
export default foo;
```

to
```
var _tmp = foo;
export { _tmp as default };
```

then in `rewriteBindingInitVisitor` we could say if it's a var that is never reassigned, instead of
```
var _tmp = foo;
exports.default = _tmp;
```
do

```
exports.default = foo;
```

packages/babel-helper-split-export-declaration/src/index.js
Andarist: ok, from what I understand this can affect anonymous functions and classes, right? import/exports are hoisted so with the given example we end up with
```
import "./a";

function _default() {
  return 4;
}

export default _default;
```
which is later transformed to by cjs transform
```
"use strict";

exports.__esModule = true;
exports.default = _default;

require("./a");

function _default() {
  return 4;
}
```

everything seems to be ok, `a.js` should be able to import/require it - as those anonymous default exports are splitted into an identifier + named function/class declaration

I suspect that there might be problems with non-hoisted by default things being splitted, right? Like in
```
// index.js
import "./a";

export default 4;

// a.js
import four from "./index";
console.log(four)
```

this would actually end up as
```js
// index.js
exports.__esModule = true;
exports.default = void 0;

require("./a");

var _default = 4;
exports.default = _default;

// a.js
var _index = _interopRequireDefault(require("./index"));

console.log(_index.default);
```

and break - I'll work on fixing this later today. Thanks for catching this!

packages/babel-helper-split-export-declaration/README.md
Andarist: this was intended as API doc - I've followed what was already done in exmaple [here](https://github.com/babel/babel/blob/master/packages/babel-helper-bindify-decorators/README.md#api)

packages/babel-helper-split-export-declaration/src/index.js
loganfsmyth: I can't tell from your comment if I was actually clear enough, so just to clarify:

loganfsmyth: I can't tell from your comment if I was actually clear enough, so just to clarify:

The issue is that
```
import "./a";

function _default() {
  return 4;
}

export default _default;
```
exports the default as an expression, meaning the `export default _default` line must have been evaluated before the imported default function will be available, whereas
```
export { _default as default };
```
explicitly re-exports the `_default` binding, which means it is exported up front before the imports are processed.

Mostly It doesn't seem like there's anything to gain from this optimization. 

```
export { _default as default };
```
is universally more consistent than
```
export default _default;
```
so essentially all we are gaining is a couple bytes.

packages/babel-helper-split-export-declaration/README.md
loganfsmyth: Alright, I wondered if it was done somewhere else. I guess it's fine, I don't feel strongly. Just potentially distracting since not everyone knows Flowtype.

packages/babel-helper-split-export-declaration/src/index.js
Andarist: I've just extracted this functionality for reuse in other places, so I'm not sure what this if was supposed to handle.

I've added a `throw` in there and tests did pass, so I'm going to remove this `if` - this helper is intended to actually split things and is not supposed to do `maybeSplit*`, so I actually prefer not to swallow any cases here.

As to `export { foo } from "bar";` - this case is handled few lines above with:
```js
if (exportDeclaration.get("specifiers").length > 0) {
  throw new Error("It doesn't make sense to split exported specifiers.");
}
```

packages/babel-helper-split-export-declaration/src/index.js
Andarist: What would be the value of splitting this? It would end up being something like
```js
class Foo {}
let Foo2 = Foo
export { Foo2 as Foo };
```

I've explicitly forbidden those cases as I'm throwing for them here, both for `ExportDefaultDeclaration` and `ExportNamedDeclaration`.

If you feel like those should be allowed, I'm really curious about when it could be useful.

packages/babel-helper-split-export-declaration/src/index.js
loganfsmyth: > this case is handled few lines above with:

oh so it is, good call

packages/babel-helper-split-export-declaration/src/index.js
loganfsmyth: Yeah sorry, not sure what I was thinking when I wrote this. I do have a comment here, but I don't think this comment was actually it haha.

My question here is, what do we gain by throwing? There's nothing inherently wrong with splitting

```
export default foo;
```
into
```
var _default = foo;
export default _default;
```
and if the user doesn't want that they can already special-cause it before calling split.

packages/babel-helper-split-export-declaration/src/index.js
```diff
@@ -0,0 +1,65 @@
+import * as t from "@babel/types";
+
+const splitExportDeclaration = exportDeclaration => {
```
loganfsmyth: Style-wise, I find splitting up the `export default splitExportDeclaration;` and this harder to follow. Any reason not to use `export default function splitExportDeclaration(exportDeclaration) {` instead?

packages/babel-helper-split-export-declaration/src/index.js
Andarist: I don't think there is anything inherently wrong with this, I just don't see any gains in splitting this. I've added error to guard users against blindly using this helper and ending up with weirdo outputs - especially that helpers are mainly used as internal packages from what I understand.

I also don't actually feel about this really strongly, so if you feel that this error is unnecessary let me know and I'll remove it.

packages/babel-helper-split-export-declaration/src/index.js
Andarist: Agreed, no idea why I didn't use a one-liner here. Gonna change it soon.

packages/babel-helper-split-export-declaration/src/index.js
Andarist: It seems that we are still not on the same page, probably because I don't know spec that well as you do. I've tried to look into it right now to clarify things, but man, this thing is still too hard for me to comprehend.

I realize that I do not understand how import/export hoisting works **exactly**.

Let's look on the given example of:
```js
import "./a";

function _default() {
  return 4;
}

export default _default;
```

gets transpiled by the **master** branch to 
```js
"use strict";

  value: true
});
exports.default = void 0;

require("./a");

function _default() {
  return 4;
}

var _default2 = _default;
exports.default = _default2;
```

and on my PR to 
```js
"use strict";

  value: true
});
exports.default = _default;

require("./a");

function _default() {
  return 4;
}
```

Which is actually incorrect, do I get it correct? Because we are exporting an expression it shouldnt be hoisted right away and assigned to `exports.default`.

It is correct by accident in tests etc because this is an output of splitting the:
```js
export default function () {
  return 4;
}
```
which should get hoisted and all, but after we do the split the intermediate output is wrong (or rather the intermediate output is wrongfully transformed into "correct" output).

I believe transforming into `export { _default as default };` is correct approach IF `_default` should get hoisted. If we split `export default 4` I'd argue that the current way (transforming into `export default _default`) is more correct, because it doesn't "allow" for changing value of `_default` binding afterwards and thus changing the exported value. Such export is not live.

packages/babel-helper-split-export-declaration/src/index.js
loganfsmyth: I see the goal of splitting as "give me a standalone declaration that isn't tied to `export`. I'm curious what you see the goal as, if not that. Given that goal, whether the exported thing is an identifier or function/class declaration isn't really a special  case.

packages/babel-helper-split-export-declaration/src/index.js
loganfsmyth: > Which is actually incorrect, do I get it correct? Because we are exporting an expression it shouldnt be hoisted right away and assigned to exports.default.

Yep, that's right.

> I'd argue that the current way (transforming into export default _default) is more correct, because it doesn't "allow" for changing value of _default binding afterwards and thus changing the exported value. Such export is not live.

It seems like an overoptimization to me. If some other plugin is mutating bindings locally, they already run the risk of mutating a live binding of something they weren't supposed to if the user wrote their code without ever reassigning an exported variable. Also it seems counter to what I mentioned in my other comment for what the overall goal of this module should be. I view the goal of splitting specifically to separate an export into a standard JS statement with no export info, and an export statement without runtime behavior. By splitting it into `export default foo;`, we've split one statement with an expression into two separate statements with expressions.

packages/babel-helper-split-export-declaration/src/index.js
Andarist: I've focused more on the reasons why the splitting was already used - to simplify rest of the transform. Will change it to allow identifiers.

Andarist:
@loganfsmyth @nicolo-ribaudo would love your review on the updated version.

We still can end up with some unnecessary temp variables when using modules transforms - ie:
```js
var _default = [];
exports.default = _default;"
```
or 
```js
const foo = 'foo';
var _default = foo;
exports.default = _default;
```
but this PR hasn't changed anything regarding this. I've tried to use @loganfsmyth suggestion from [here](https://github.com/babel/babel/pull/7313#discussion_r165759173) but because of scope bindings being WAY out of sync after multiple transformations, I've ditched it for now.

packages/babel-traverse/src/scope/index.js
```diff
@@ -210,7 +210,7 @@ export default class Scope {
    * Generate a unique identifier and add it to the current scope.
    */
 
-  generateDeclaredUidIdentifier(name: string = "temp") {
```
loganfsmyth: Should we mark these `name?: string` so it is clear they are still optional? At a glance, this kind of looks like you've made them required, which isn't the case.

loganfsmyth:
> I've tried to use @loganfsmyth suggestion from here but

I'm 100% fine with that being some later optimization.

packages/babel-traverse/src/scope/index.js
Andarist: Sure thing, kinda forgot this part of the codebase is typed. Gonna fix soon.

PR #7314, opened 2018-02-02
Add location information to parsing errors

kaicataldo:
| Fixed Issues?            | refs https://github.com/babel/babel-eslint/issues/573#issuecomment-362142343
| Documentation PR         | 

As discussed in [Slack](https://babeljs.slack.com/archives/C062RC35M/p1517507859000347), this PR retains location information for errors thrown by the parser and adds error codes so that the consumer can distinguish between errors easily.

A few questions I have:
1. Did I miss any spots?
2. Should we be exposing these error codes as constants that a consumer can import so they don't have to care about Babel changing these strings? With only two errors codes, maybe this isn't necessary. They're also shared across packages, so I'm not sure how the team would want to go about that.
3. I didn't update any tests - should I have?

### DIFF
--- a/packages/babel-core/src/transformation/normalize-file.js
+++ b/packages/babel-core/src/transformation/normalize-file.js
@@ -91,7 +91,6 @@ function parser(pluginPasses, options, code) {
   } catch (err) {
     const { loc, missingPlugin } = err;
     if (loc) {
-      err.loc = null;
       const codeFrame = codeFrameColumns(
         code,
         {
@@ -110,6 +109,7 @@ function parser(pluginPasses, options, code) {
         err.message =
           `${options.filename || "unknown"}: ${err.message}\n\n` + codeFrame;
       }
+      err.code = "BABEL_PARSE_ERROR";
     }
     throw err;
   }
--- a/packages/babel-template/src/parse.js
+++ b/packages/babel-template/src/parse.js
@@ -148,8 +148,8 @@ function parseWithCodeFrame(code: string, parserOpts: {}): BabelNodeFile {
   } catch (err) {
     const loc = err.loc;
     if (loc) {
-      err.loc = null;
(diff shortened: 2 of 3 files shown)


packages/babel-core/src/transformation/normalize-file.js
```diff
@@ -110,6 +109,7 @@ function parser(pluginPasses, options, code) {
         err.message =
           `${options.filename || "unknown"}: ${err.message}\n\n` + codeFrame;
       }
+      err.code = "BABEL_PARSE_ERROR";
```
hzoo: I don't know if this is the first time we started doing this but are we going to have some kind of convention for this, docs? Or is this just a one-off @loganfsmyth

loganfsmyth: I think error codes are good. Probably in the long run anywhere we throw an error in babel-core should have its own code. Even better would be if Babylon already included a code and this code checked
```
if (err.code === "BABYLON_PARSE_ERROR")
```

hzoo: Ok yeah ^ was what I was thinking. maybe @kaicataldo can you make an issue about that? I think a task could be to look through all the places we throw and to think about if we want to make an errors file kind of deal

kaicataldo: Would be great to make this consistent! Would this be just for Babylon and babel-core? Or all packages?

hzoo: It's funny also because I'm reminded we have a package called `babel-messages` that we sometimes use?

kaicataldo:
Updated!

kaicataldo:
Whoops - was trying to delete the branch, not revert the merge!

PR #7315, opened 2018-02-02
Add core-js as valid polyfill source

danez:
| Fixed Issues?            | 
| Patch: Bug Fix?          | ?
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

I brought that up a while ago, but wanted to finally start the discussion. I'm not sure why it was removed in v7.x. I found it very useful to be able to use `import "core-js"` with babel-preset-env builtIn option. Especially to mitigate the problem that the wrong version of core-js could be installed on toplevel, so that the generated imports from core-js do fail.

Would also be interested why it was removed?

### DIFF
--- a/packages/babel-preset-env/README.md
+++ b/packages/babel-preset-env/README.md
@@ -342,6 +342,8 @@ import "core-js/modules/es7.string.pad-start";
 import "core-js/modules/es7.string.pad-end";
 ```
 
+This will also work for `core-js` directly (`import "core-js";` or `require('core-js');`)
+
 #### `useBuiltIns: false`
 
 Don't add polyfills automatically per file, or transform `import "@babel/polyfill"` to individual polyfills.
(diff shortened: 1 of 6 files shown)


hzoo:
It was confusing on npm 2 (babel-polyfill -> core-js import) so we did this whole thing about creating these stub files but it was too unwieldy and it sucked on our end to maintain it. Yeah we reverted that so this is ok now. And also in addition we wanted to either kill babel-polyfill entirely given it's literally 2 imports and/or allow preset-env to map against any polyfill you want

xtuc:
@kaicataldo Why this has been reverted?

Andarist:
@xtuc I think I've seen a comment that this was a mistake and the intention was to delete the branch

kaicataldo:
Uhhh, I have no idea. I accidentally clicked the revert button on [this PR](https://github.com/babel/babel/pull/7314) last night while trying to delete my branch, but I've never looked at this PR before...

PR #7341, opened 2018-02-06
Allow falsey, yet valid options for codeFrameColumns()

### LINKED ISSUE #7340: Allow for overriding default linesAbove/linesBelow options.
Choose one: is this a bug report or feature request? **Bug**

The `linesAbove` & `linesBelow` options for `@babel/code-frames`'s `codeFrameColumns` method are defaulting to `2`/`3` when using valid values which are consequently falsey.

### Input Code
```js
const { codeFrameColumns } = require('@babel/code-frame');

const str = `{
  "a": 1,
  "b": 2
  "c": 3
}`;

console.log(
  codeFrameColumns(
    str,
    { start: { line: 2 } },
    { .linesAbove: 0, linesBelow: 0 }
  )
);
```

N/A

### Expected Behavior
Per the code above, the expected output is:
```bash
> 2 |   "a": 1,
```

### Current Behavior
Because `linesAbove` defaults to 2, and `linesBelow` defaults to 3, the output is as follows:
```bash
  1 | {
> 2 |   "a": 1,
  3 |   "b": 2
  4 |   "c": 3
  5 | }
```

### Possible Solution
Improve option defaulting logic.

### Context
I'd like to use this tool for some cases where I only want to show a single line of a code frame.

### Your Environment
N/A

hulkish:
Allow for overriding default linesAbove/linesBelow values.

| Fixed Issues?            | Fixes #7340

Improved option defaulting logic, small change.

### DIFF
--- a/packages/babel-code-frame/src/index.js
+++ b/packages/babel-code-frame/src/index.js
@@ -127,10 +127,7 @@ function getMarkerLines(
     loc.start,
   );
   const endLoc: Location = Object.assign({}, startLoc, loc.end);
-  const {
-    linesAbove = 2,
-    linesBelow = 3,
-  } = opts || {};
+  const { linesAbove = 2, linesBelow = 3 } = opts || {};
   const startLine = startLoc.line;
   const startColumn = startLoc.column;
   const endLine = endLoc.line;
--- a/packages/babel-code-frame/src/index.js
+++ b/packages/babel-code-frame/src/index.js
@@ -127,9 +127,10 @@ function getMarkerLines(
     loc.start,
   );
   const endLoc: Location = Object.assign({}, startLoc, loc.end);
-  const linesAbove = opts.linesAbove || 2;
-  const linesBelow = opts.linesBelow || 3;
-
+  const {
+    linesAbove = 2,
+    linesBelow = 3,
+  } = opts || {};
   const startLine = startLoc.line;
   const startColumn = startLoc.column;
   const endLine = endLoc.line;
--- a/packages/babel-code-frame/test/index.js
+++ b/packages/babel-code-frame/test/index.js
@@ -184,6 +184,57 @@ describe("@babel/code-frame", function() {
     );
   });
 
+  it("opts.linesAbove no lines above", function() {
+    const rawLines = [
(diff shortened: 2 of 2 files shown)


packages/babel-code-frame/src/index.js
```diff
@@ -127,8 +127,8 @@ function getMarkerLines(
     loc.start,
   );
   const endLoc: Location = Object.assign({}, startLoc, loc.end);
-  const linesAbove = opts.linesAbove || 2;
-  const linesBelow = opts.linesBelow || 3;
+  const linesAbove = isNaN(opts.linesAbove) ? 2 : opts.linesAbove;
+  const linesBelow = isNaN(opts.linesBelow) ? 3 : opts.linesBelow;
```
Andarist: `isNaN` is imho a legacy API and should not be used nowadays, not sure what the intention was but I guess this should be fine, right?
```
const linesAbove = typeof opts.linesAbove !== 'number' ? 2 : opts.linesAbove;
const linesBelow = typeof opts.linesBelow !== 'number' ? 3 : opts.linesBelow;
```

hzoo: seems like it's just to allow 0

hulkish: yes, to allow 0. I agree about the isNaN usage - i would have preferred to use destructuring with defaults. But, guess I was going for lowest impacting change lol.

hulkish:
@Andarist better?

Andarist:
@hulkish Yeah, that works too. `opts` was not guarded before, so I'm not sure if `|| {}` is needed here though.

hzoo:
Do we just need to rebase this? Since the other prs are passing

nicolo-ribaudo:
There is no need to rebase: it will work when merged, since the failing test has been fixed in code not touched by this PR.

hzoo:
thanks @hulkish! good find

apepper:
When will there be a release of `babel-code-frame` containing this bugfix?

nicolo-ribaudo:
You can try it in Babel 7 beta (`@babel/code-frame`)

PR #7345, opened 2018-02-07
Small tweaks to prep for coming .babelrc lookup work

loganfsmyth:
| Fixed Issues?            | 
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

See individual commits if you want details. Nothing crazy, just a little work to stabilize things, with the main change being re-merging `.babelrc` and `.babelignore` searching into a single pass. Wanted to make it less work to bail out of _both_ once some end-goal in the hierarchy is found.

### DIFF
--- a/packages/babel-core/src/config/files/index-browser.js
+++ b/packages/babel-core/src/config/files/index-browser.js
@@ -12,16 +12,16 @@ export type IgnoreFile = {
   ignore: Array<string>,
 };
 
-export function findBabelrc(
+export type RelativeConfig = {
+  config: ConfigFile | null,
+  ignore: IgnoreFile | null,
+};
+
+export function findRelativeConfig(
   filepath: string,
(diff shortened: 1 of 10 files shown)


packages/babel-core/src/config/files/index-browser.js
```diff
@@ -12,16 +12,16 @@ export type IgnoreFile = {
   ignore: Array<string>,
 };
 
-export function findBabelrc(
+export type RelativeConfig = {
+  config: ConfigFile | null,
+  ignore: IgnoreFile | null,
+};
```
nicolo-ribaudo: Can this be reexported from `configuration.js`?

loganfsmyth: Good call! I feel like there was a reason for duplicating these types like this originally, but I don't know what it was.

PR #7351, opened 2018-02-08
Extract @babel/highlight package from @babel/code-frame

suchipi:

This creates a new package called `@babel/highlight` which syntax highlights
JavaScript code for terminal output. This functionality was already
present in `@babel/code-frame`, but exposing it as a separate package lets
other projects leverage it. `@babel/code-frame` has been refactored
slightly so that it uses `@babel/highlight` to do its syntax highlighting.

### DIFF
--- a/packages/babel-code-frame/package.json
+++ b/packages/babel-code-frame/package.json
@@ -8,11 +8,10 @@
   "repository": "https://github.com/babel/babel/tree/master/packages/babel-code-frame",
   "main": "lib/index.js",
   "dependencies": {
-    "chalk": "^2.0.0",
-    "esutils": "^2.0.2",
-    "js-tokens": "^3.0.0"
+    "@babel/highlight": "7.0.0-beta.39"
   },
   "devDependencies": {
+    "chalk": "^2.0.0",
     "strip-ansi": "^4.0.0"
   }
 }
--- a/packages/babel-highlight/package.json
+++ b/packages/babel-highlight/package.json
@@ -1,0 +1,18 @@
+{
+  "name": "@babel/highlight",
+  "version": "7.0.0-beta.39",
+  "description": "Syntax highlight JavaScript strings for output in terminals.",
+  "author": "suchipi <me@suchipi.com>",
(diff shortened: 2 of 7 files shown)


packages/babel-code-frame/package.json
```diff
@@ -8,11 +8,10 @@
   "repository": "https://github.com/babel/babel/tree/master/packages/babel-code-frame",
   "main": "lib/index.js",
   "dependencies": {
-    "chalk": "^2.0.0",
-    "esutils": "^2.0.2",
-    "js-tokens": "^3.0.0"
+    "@babel/highlight": "7.0.0-beta.39"
   },
   "devDependencies": {
+    "chalk": "^2.0.0",
```
rajzshkr: why we need chalk as a devdep here?

suchipi: It's used in the tests.

packages/babel-highlight/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@babel/highlight",
+  "version": "7.0.0-beta.39",
+  "description": "Syntax highlight JavaScript strings for output in terminals.",
+  "author": "Sebastian McKenzie <sebmck@gmail.com>",
```
xtuc: Shouldn't we change the author for new packages? Maybe @hzoo?

suchipi: I wasn't sure what to put. Since he was listed as the author on code-frame, I assumed most of this code was written by Sebastian, so I kept it the same since I didn't add much.

xtuc: Yes good point ok

hzoo: I guess it could even be you @suchipi? It's a new package name, not really based on who wrote the most, although even for new packages we have been doing that which I'm not sure about either?

suchipi: I'm fine changing it to my name 🤷‍♀️

suchipi:
🎉 Thanks everyone!

PR #7353, opened 2018-02-08
build types as part of ci, to re-enable flow checks (also see #7227)

bcherny:
See https://github.com/babel/babel/pull/7101#issuecomment-363682345


bcherny:
https://github.com/babel/babel/pull/7227

PR #7356, opened 2018-02-08
Fixes issues regarding super in optionalChain

### LINKED ISSUE #7256: Optional Chaining Bugs Checklist

nveenjain:
| Fixed Issues?            | Fixes #7256 
| Documentation PR         | 

Fixes issues regarding super in optionalChain, Now can't use super object/method with optionalChain.

### DIFF
--- a/packages/babylon/test/fixtures/experimental/optional-chaining/optional-super-property-class/input.js
+++ b/packages/babylon/test/fixtures/experimental/optional-chaining/optional-super-property-class/input.js
@@ -1,0 +1,5 @@
+class A{
+    b(){
+        return super?.b;
+    }
+}
--- a/packages/babylon/test/fixtures/experimental/optional-chaining/optional-super-property-class/options.json
+++ b/packages/babylon/test/fixtures/experimental/optional-chaining/optional-super-property-class/options.json
@@ -1,0 +1,4 @@
+{
+    "plugins":["optionalChaining"],
+    "throws": "Unexpected token (3:20)"
+}
--- a/packages/babylon/test/fixtures/experimental/optional-chaining/optional-super-property/input.js
+++ b/packages/babylon/test/fixtures/experimental/optional-chaining/optional-super-property/input.js
@@ -1,0 +1,5 @@
+const a = {
+    b(){
+        return super?.c;
+    }
+}
--- a/packages/babylon/test/fixtures/experimental/optional-chaining/optional-super-property/options.json
+++ b/packages/babylon/test/fixtures/experimental/optional-chaining/optional-super-property/options.json
@@ -1,0 +1,4 @@
+{
+    "plugins":["optionalChaining"],
+    "throws": "Unexpected token (3:20)"
+}
--- a/packages/babylon/test/fixtures/experimental/optional-chaining/super-method-class/input.js
+++ b/packages/babylon/test/fixtures/experimental/optional-chaining/super-method-class/input.js
@@ -1,0 +1,5 @@
+class A{
+    constructor(){
+        super()?.b;
+    }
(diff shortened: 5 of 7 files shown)


packages/babylon/test/fixtures/experimental/optional-chaining/optional-super-method/input.js
```diff
@@ -0,0 +1,5 @@
+class a{
+    constructor(){
+        super()?.b;
```
jridgewell: This is actually legal. 

```
OptionalExpression[Yield, Await]:
    CallExpression[?Yield, ?Await] OptionalChain[?Yield, ?Await]
    ...elided...

CallExpression[Yield, Await]:
    SuperCall[?Yield, ?Await]
    ...elided...

SuperCall[Yield, Await]:
    super Arguments[?Yield, ?Await]
```

So a `SuperCall` is a `CallExpression`, and `CallExpression` be used as the base of an `OptionalChain`.

nveenjain: So, only super properties are not supported, right?

jridgewell: Correct.

nveenjain: Just had a look at spec, I'm having doubts:- Acc. to spec:- 
```js
OptionalExpression[Yield, Await]:
   MemberExpression[?Yield, ?Await]OptionalChain[?Yield, ?Await]
   CallExpression[?Yield, ?Await]OptionalChain[?Yield, ?Await]
   OptionalExpression[?Yield, ?Await]OptionalChain[?Yield, ?Await]
```
and 
```js
MemberExpression[Yield, Await]:
   PrimaryExpression[?Yield, ?Await]
   MemberExpression[?Yield, ?Await][Expression[+In, ?Yield, ?Await]]
   MemberExpression[?Yield, ?Await].IdentifierName
   MemberExpression[?Yield, ?Await]TemplateLiteral[?Yield, ?Await]
   SuperProperty[?Yield, ?Await]
```

nveenjain: So I think we don't need to throw errors in super property/method, am i right?

jridgewell: A `SuperProperty` is `super.prop`, which may be the base of an optional chain: `super.prop?.p`. But, the optional chain syntax cannot appear between the `super` the `prop`, only after.

```js
super.prop?.p // legal, base is a `object` is a MemberExpression.

super?.prop // illegal, super is a keyword and may not appear as the `object`

nveenjain: This is already implemented. ( though test cases were not provided).. Should i add test cases only in this PR?

jridgewell: If that's the case, then just tests to verify.

jridgewell:
Awesome work!

PR #6133, opened 2017-08-18
add --config-file option to CLI to pass in .babelrc location

bdwain:
7.0 version of #6125 

| Q                        | A 
| ------------------------ | ---
| Fixed Issues             | Fixes #5689 
| Patch: Bug Fix?          | 
| Major: Breaking Change?  | 
| Minor: New Feature?      | 👍
| Tests Added/Pass?        | 👍
| Spec Compliancy?         | 
| License                  | MIT
| Doc PR                   | 
| Any Dependency Changes?  | 

This Fixes #5689  by adding the --config-file option in the CLI. Now you can point babel to any .babelrc file when running it with the CLI.

PR #7358, opened 2018-02-08
Allow more flexible file-based configuration while preventing .babelrcs from breaking things

### LINKED ISSUE #6766: How do we want to handle `.babelrc` lookup in 7.x
### Edit 2
Most recent status: https://github.com/babel/babel/issues/6766#issuecomment-352225586

### Edit 1

All below still applies, but also see more explanation in https://github.com/babel/babel/issues/6766#issuecomment-347966349

### Original:

Not a fleshed out proposal, but filing for tracking and discussion.

Currently if you have:

```
.babelrc
src/
  index.js
  utils/
    .babelrc
    other.js
```
compiling `index.js` will use the root-level `.babelrc`, and `other.js` will use the `src/utils/.babelrc`. I assume this pattern was pulled from other utilities like ESLint, but not 100% sure.

We're at the point now where this approach has become a problem for us because:

1. Very few people seem to expect this behavior to begin with.
2. People expect their project-level `.babelrc` file to also apply to things that live in `node_modules`. Babel 6 was terrible at figuring out how to handle `node_modules` because it always assumes things are strict mode, and always rewrites `this`, but a lot of those issues aren't as big in Babel 7, or at least ideally won't be, because we plan to have better handling of the `sourceType: script` vs `sourceType: module` distinction. This _kind_ of works now, but see point (3).
3. If people publish packages to npm and include `.babelrc` files in their packages, or more troublingly, use the `babel` key in `package.json` (which you can't really get around publishing), users trying to use Webpack or anything like it to compile node_modules get extremely confusing errors about not being able to find plugins and presets referenced in those files. See https://github.com/facebookincubator/create-react-app/issues/1125 for a massive list of users that want to compile `node_modules`, but would _absolutely_ be bitten by this issue.

So where does that leave us? The simplest thing I can think for us to do would be to change up Babel's API to, by default, accept a root directory as an argument where it would search for the config, defaulting to the working directory.

We're already at a point where I've thought about adding an argument for `root directory` to resolve the `filename` relative to, so that should be uncontroversial. Actually changing config resolution is a bigger question.

It would be _pretty_ easy for us to allow users to opt for the old behavior, so I think my proposal would probably be to add this new root-dir config resolution logic and make it the default, but keep the old behavior around as an opt in. So where now we have

```
babelrc: boolean
```
to toggle config resolution off and on, I could imagine us doing
```
babelrc: true | "root" => Look for config in root directory
babelrc: "relative" => Look for config, starting from compiled file, working upward
babelrc: false => Skip search entirely.
```
instead.

The primary loss of functionality here is that if users _do_ want to use `.babelrc` files inside nested folders to change Babel configs based on path, they don't have an easy way to do it anymore. I don't however have a good grasp of how many users that might be. It seems like we may want to consider https://github.com/babel/babel/issues/5451 as something to be included in this effort if we do want to change, which might end up blocked on https://github.com/babel/babel/issues/6765.

loganfsmyth:
| Fixed Issues?            | Fixes #6766
| Patch: Bug Fix?          |
| Tests Added + Pass?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Alright! Let's do this. This is my proposal for fixing the issues spelled out in #6766. This PR does the following:

### When searching for `.babelrc` files for a given file, stop at package.json

Currently, Babel will look all the way up the directory hierarchy. This change prevents config files outside of a given package from potentially breaking your project's builds, which has been a problem occasionally. Also potentially frustrating for projects that integrate Babel, because they often allow users can _optionally_ create .babelrc files, and if they haven't created one, but have one in a parent folder, it might find a bad config.

The downside here is that monorepos like

```
.babelrc
packages/
  one/
    package.json
    index.js
  two/
    package.json
    index.js
```
will fail to find the `.babelrc`. I think that is acceptable because there are alternatives:

1. Have each module have their own `.babelrc` doing `{ extends: "../../.babelrc" }` to explicitly load the config from a parent directory.
2. See the next point.

### Adds support for a `babel.config.js` file along the lines of what Webpack does

This PR adds a `root` option (that defaults to the working directory), that defines the conceptual "root" of your project. This location is used to automatically search for the config file `babel.config.js`. Optionally instead the user can specify `configFile: "./foo.js"` or whatever they'd like.

The file itself behaves just like a `.babelrc.js` file would, but is not loaded relative to the individual file being compiled, and thus has more flexibility in what files it applies to. For instance, if users are symlinking in `node_modules` or something, it is currently impossible for a `.babelrc` to configure their builds any way other than by hard-coding their config into webpack.config.js, which is not great.

With our new support for `overrides` it's now easy for users to have a single `babel.config.js` which could define the full build process for their project, when used alongside `env`.

### Only search for a `.babelrc` for files inside the `root` (which defaults to cwd) (unless the user opts out)

Specifically excluding `node_modules` folders. Given the path of the file being compiled, if the path relative to the root contains `node_modules`, do not attempt to search for `.babelrc` files. This essentially has the effect of making `.babelrc` files only useful by default in a "root" package, which is essentially how most people used them.

This (potentially combined with the previous `babel.config.js`) allows users to _safely_ do
```
test: /\.js$/,
loader: "babel-loader",
```
without risking loading a `.babelrc` from a node_modules that was not intented to apply to the current project. There _are_ still other edge cases to compiling `node_modules`, but this was a particularly dangerous edge case, especially for things like `babel-register` that don't have a nice `rules` list like Webpack does.

To allow people to opt into `.babelrc` usage, potentially for local development, this PR allows Babel's `babelrc` option to be an array, so users could do

```
test: /\.js$/,
loader: "babel-loader",
options: {
  babelrc: [
    ".",
    "../some-linked-package"
  ]
}
```
which would also attempt to load `.babelrc` files for anything within that package. Alternatively

```
options: {
  babelrc: [
    ".",
    "node_modules/foo"
  ]
}
```
if there was a specific package you wanted to load a config for, and were confident in its installation location.

### What this PR does _not_ do

Part of the reason I personally had so much back and forth on this issue is because it wasn't necessarily obvious what `root` should actually be responsible for. For instance, if Babel is passed a file that is not in the `root`, should it compile it at all? As is hopefully clear, my answer there was yes.

One core thing that this PR does is that it leaves it up to the individual _callers_ of Babel to decide what files should actually be passed to Babel in the first place. Alongside that, if you configure Babel with a custom `root` or `configFile` or `babelrc` array, the user will have to decide if those need to be replicated between Babel callers, because they are top-level Babel options, not ones that can be included inside a config file (for obvious reasons).

### What is left

This PR has no tests, and an unoptimized implementation of a lot of stuff. It will need more work before it can land, but I want to post it now to get a feel for what everyone thinks.

### DIFF
--- a/packages/babel-cli/src/babel/index.js
+++ b/packages/babel-cli/src/babel/index.js
@@ -238,10 +238,6 @@ if (errors.length) {
 //
 
 const opts = commander.opts();
-//the configFile CLI option maps to the extends option in the node API
-if (opts.configFile) {
-  opts.extends = opts.configFile;
-}
 
 // Delete options that are specific to @babel/cli and shouldn't be passed to @babel/core.
 delete opts.version;
@@ -253,7 +249,6 @@ delete opts.outDir;
 delete opts.copyFiles;
 delete opts.includeDotfiles;
 delete opts.verbose;
-delete opts.configFile;
 delete opts.deleteDirOnStart;
 delete opts.keepFileExtension;
 delete opts.relative;
--- a/packages/babel-core/README.md
+++ b/packages/babel-core/README.md
@@ -218,7 +218,8 @@ Following is a table of the options you can use:
 | `auxiliaryCommentBefore` | `null`               | Attach a comment before all non-user injected code |
 | `root`                   | `"."`                | Specify the "root" folder that defines the location to search for "babel.config.js", and the default folder to allow `.babelrc` files inside of.|
 | `configFile`             | `undefined`          | The config file to load Babel's config from. Defaults to searching for "babel.config.js" inside the "root" folder. `false` will disable searching for config files.|
-| `babelrc`                | `(root)`             | Specify whether or not to use .babelrc and .babelignore files. Not available when using the CLI, [use `--no-babelrc` instead](https://babeljs.io/docs/usage/cli/#babel-ignoring-babelrc). `false` to disable searching, and `true` to always search, a string path of the package to search inside of, or an array of paths to packages to search inside of. |
+| `babelrc`                | `true`               | Specify whether or not to use .babelrc and .babelignore files. Not available when using the CLI, [use `--no-babelrc` instead](https://babeljs.io/docs/usage/cli/#babel-ignoring-babelrc) |
+| `babelrcRoots`           | `(root)`             | Specify which packages should be search for .babelrc files when they are being compiled. `true` to _always_ search, or a path string or an array of paths to packages to search inside of. Defaults to only searching the "root" package. |
 | `envName`                | env vars             | Defaults to environment variable `BABEL_ENV` if set, or else `NODE_ENV` if set, or else it defaults to `"development"` |
 | `code`                   | `true`               | Enable code generation |
 | `comments`               | `true`               | Output comments in generated output |
--- a/packages/babel-core/README.md
+++ b/packages/babel-core/README.md
@@ -216,7 +216,9 @@ Following is a table of the options you can use:
 | `ast`                    | `false`              | Include the AST in the returned object |
 | `auxiliaryCommentAfter`  | `null`               | Attach a comment after all non-user injected code |
 | `auxiliaryCommentBefore` | `null`               | Attach a comment before all non-user injected code |
-| `babelrc`                | `true`               | Specify whether or not to use .babelrc and .babelignore files. Not available when using the CLI, [use `--no-babelrc` instead](https://babeljs.io/docs/usage/cli/#babel-ignoring-babelrc) |
+| `root`                   | `"."`                | Specify the "root" folder that defines the location to search for "babel.config.js", and the default folder to allow `.babelrc` files inside of.|
+| `configFile`             | `undefined`          | The config file to load Babel's config from. Defaults to searching for "babel.config.js" inside the "root" folder. `false` will disable searching for config files.|
+| `babelrc`                | `(root)`             | Specify whether or not to use .babelrc and .babelignore files. Not available when using the CLI, [use `--no-babelrc` instead](https://babeljs.io/docs/usage/cli/#babel-ignoring-babelrc). `false` to disable searching, and `true` to always search, a string path of the package to search inside of, or an array of paths to packages to search inside of. |
 | `envName`                | env vars             | Defaults to environment variable `BABEL_ENV` if set, or else `NODE_ENV` if set, or else it defaults to `"development"` |
 | `code`                   | `true`               | Enable code generation |
 | `comments`               | `true`               | Output comments in generated output |
--- a/packages/babel-core/src/config/config-chain.js
+++ b/packages/babel-core/src/config/config-chain.js
@@ -13,6 +13,7 @@ import {
 const debug = buildDebug("babel:config:config-chain");
 
 import {
+  findPackageData,
   findRelativeConfig,
   loadConfig,
   type ConfigFile,
@@ -125,15 +126,17 @@ export function buildRootChain(
   );
   if (!programmaticChain) return null;
 
-  let ignore, babelrc;
+  const pkgData =
+    typeof context.filename === "string"
+      ? findPackageData(context.filename)
+      : null;
 
+  let ignore, babelrc;
   const fileChain = emptyChain();
   // resolve all .babelrc files
-  if (opts.babelrc !== false && context.filename !== null) {
-    const filename = context.filename;
-
+  if (opts.babelrc !== false && pkgData) {
     ({ ignore, config: babelrc } = findRelativeConfig(
-      filename,
+      pkgData,
       context.envName,
     ));
 
--- a/packages/babel-core/src/config/config-chain.js
+++ b/packages/babel-core/src/config/config-chain.js
@@ -15,6 +15,7 @@ const debug = buildDebug("babel:config:config-chain");
 import {
   findPackageData,
   findRelativeConfig,
+  findRootConfig,
   loadConfig,
   type ConfigFile,
   type IgnoreFile,
@@ -107,6 +108,7 @@ const loadPresetOverridesEnvDescriptors = makeWeakCache(
 
 export type RootConfigChain = ConfigChain & {
   babelrc: ConfigFile | void,
+  config: ConfigFile | void,
   ignore: IgnoreFile | void,
 };
 
@@ -125,6 +127,26 @@ export function buildRootChain(
     context,
   );
   if (!programmaticChain) return null;
+
+  const { root: rootDir = ".", configFile: configFileName } = opts;
+
+  let configFile;
+  if (typeof configFileName === "string") {
+    configFile = loadConfig(configFileName, context.cwd, context.envName);
+  } else if (configFileName === undefined || configFileName === true) {
+    configFile = findRootConfig(
+      path.resolve(context.cwd, rootDir),
+      context.envName,
+    );
+  }
+
+  const configFileChain = emptyChain();
+  if (configFile) {
+    const result = loadFileChain(configFile, context);
+    if (!result) return null;
+
+    mergeChain(configFileChain, result);
+  }
 
   const pkgData =
     typeof context.filename === "string"
@@ -155,7 +177,7 @@ export function buildRootChain(
   // Insert file chain in front so programmatic options have priority
   // over configuration file chain items.
   const chain = mergeChain(
-    mergeChain(emptyChain(), fileChain),
+    mergeChain(mergeChain(emptyChain(), configFileChain), fileChain),
     programmaticChain,
   );
 
@@ -165,6 +187,7 @@ export function buildRootChain(
     options: chain.options.map(o => normalizeOptions(o)),
     ignore: ignore || undefined,
     babelrc: babelrc || undefined,
+    config: configFile || undefined,
   };
 }
 
--- a/packages/babel-core/src/config/config-chain.js
+++ b/packages/babel-core/src/config/config-chain.js
@@ -8,6 +8,7 @@ import {
   type ValidatedOptions,
   type IgnoreList,
   type ConfigApplicableTest,
+  type BabelrcSearch,
 } from "./validation/options";
 
 const debug = buildDebug("babel:config:config-chain");
@@ -19,6 +20,7 @@ import {
   loadConfig,
   type ConfigFile,
   type IgnoreFile,
+  type FilePackageData,
 } from "./files";
 
 import { makeWeakCache, makeStrongCache } from "./caching";
@@ -128,12 +130,16 @@ export function buildRootChain(
   );
   if (!programmaticChain) return null;
 
-  const { root: rootDir = ".", configFile: configFileName } = opts;
+  const {
+    root: rootDir = ".",
+    babelrc = undefined,
+    configFile: configFileName = true,
+  } = opts;
 
   let configFile;
   if (typeof configFileName === "string") {
     configFile = loadConfig(configFileName, context.cwd, context.envName);
-  } else if (configFileName === undefined || configFileName === true) {
+  } else if (configFileName === true) {
     configFile = findRootConfig(
       path.resolve(context.cwd, rootDir),
       context.envName,
@@ -153,21 +159,24 @@ export function buildRootChain(
       ? findPackageData(context.filename)
       : null;
 
-  let ignore, babelrc;
+  let ignoreFile, babelrcFile;
   const fileChain = emptyChain();
   // resolve all .babelrc files
-  if (opts.babelrc !== false && pkgData) {
-    ({ ignore, config: babelrc } = findRelativeConfig(
+  if (pkgData && babelrcLoadEnabled(context, pkgData, babelrc, rootDir)) {
+    ({ ignore: ignoreFile, config: babelrcFile } = findRelativeConfig(
       pkgData,
       context.envName,
     ));
 
-    if (ignore && shouldIgnore(context, ignore.ignore, null, ignore.dirname)) {
+    if (
+      ignoreFile &&
+      shouldIgnore(context, ignoreFile.ignore, null, ignoreFile.dirname)
+    ) {
       return null;
     }
 
-    if (babelrc) {
-      const result = loadFileChain(babelrc, context);
+    if (babelrcFile) {
+      const result = loadFileChain(babelrcFile, context);
       if (!result) return null;
 
       mergeChain(fileChain, result);
@@ -185,10 +194,37 @@ export function buildRootChain(
     plugins: dedupDescriptors(chain.plugins),
     presets: dedupDescriptors(chain.presets),
     options: chain.options.map(o => normalizeOptions(o)),
-    ignore: ignore || undefined,
-    babelrc: babelrc || undefined,
+    ignore: ignoreFile || undefined,
+    babelrc: babelrcFile || undefined,
     config: configFile || undefined,
   };
+}
+
+function babelrcLoadEnabled(
+  context: ConfigContext,
+  pkgData: FilePackageData,
+  babelrc: BabelrcSearch | void,
+  rootDir: string,
+): boolean {
+  if (typeof babelrc === "boolean") return babelrc;
+
+  const absoluteRoot = path.resolve(context.cwd, rootDir);
+
+  // Fast path to avoid having to load micromatch if the babelrc is just
+  // loading in the standard root directory.
+  if (
+    babelrc === undefined ||
+    babelrc === rootDir ||
+    (Array.isArray(babelrc) && babelrc.length === 1 && babelrc[0] === rootDir)
+  ) {
+    return pkgData.directories.indexOf(absoluteRoot) !== -1;
+  }
+
+  const babelrcRoots = (Array.isArray(babelrc) ? babelrc : [babelrc]).map(pat =>
+    path.resolve(context.cwd, pat),
+  );
+
+  return micromatch(pkgData.directories, babelrcRoots).length > 0;
 }
 
 /**
--- a/packages/babel-core/src/config/config-chain.js
+++ b/packages/babel-core/src/config/config-chain.js
@@ -132,18 +132,18 @@ export function buildRootChain(
 
   const {
     root: rootDir = ".",
-    babelrc = undefined,
+    babelrc = true,
+    babelrcRoots,
     configFile: configFileName = true,
   } = opts;
+
+  const absoluteRoot = path.resolve(context.cwd, rootDir);
 
   let configFile;
   if (typeof configFileName === "string") {
     configFile = loadConfig(configFileName, context.cwd, context.envName);
   } else if (configFileName === true) {
-    configFile = findRootConfig(
-      path.resolve(context.cwd, rootDir),
-      context.envName,
-    );
+    configFile = findRootConfig(absoluteRoot, context.envName);
   }
 
   const configFileChain = emptyChain();
@@ -162,7 +162,11 @@ export function buildRootChain(
   let ignoreFile, babelrcFile;
   const fileChain = emptyChain();
   // resolve all .babelrc files
-  if (pkgData && babelrcLoadEnabled(context, pkgData, babelrc, rootDir)) {
+  if (
+    babelrc &&
+    pkgData &&
+    babelrcLoadEnabled(context, pkgData, babelrcRoots, absoluteRoot)
+  ) {
     ({ ignore: ignoreFile, config: babelrcFile } = findRelativeConfig(
       pkgData,
       context.envName,
@@ -203,28 +207,28 @@ export function buildRootChain(
 function babelrcLoadEnabled(
   context: ConfigContext,
   pkgData: FilePackageData,
-  babelrc: BabelrcSearch | void,
-  rootDir: string,
+  babelrcRoots: BabelrcSearch | void,
+  absoluteRoot: string,
 ): boolean {
-  if (typeof babelrc === "boolean") return babelrc;
-
-  const absoluteRoot = path.resolve(context.cwd, rootDir);
+  if (typeof babelrcRoots === "boolean") return babelrcRoots;
 
   // Fast path to avoid having to load micromatch if the babelrc is just
   // loading in the standard root directory.
-  if (
-    babelrc === undefined ||
-    babelrc === rootDir ||
-    (Array.isArray(babelrc) && babelrc.length === 1 && babelrc[0] === rootDir)
-  ) {
+  if (babelrcRoots === undefined) {
     return pkgData.directories.indexOf(absoluteRoot) !== -1;
   }
 
-  const babelrcRoots = (Array.isArray(babelrc) ? babelrc : [babelrc]).map(pat =>
-    path.resolve(context.cwd, pat),
-  );
+  let babelrcPatterns = babelrcRoots;
+  if (!Array.isArray(babelrcPatterns)) babelrcPatterns = [babelrcPatterns];
+  babelrcPatterns = babelrcPatterns.map(pat => path.resolve(context.cwd, pat));
 
-  return micromatch(pkgData.directories, babelrcRoots).length > 0;
+  // Fast path to avoid having to load micromatch if the babelrc is just
+  // loading in the standard root directory.
+  if (babelrcPatterns.length === 1 && babelrcPatterns[0] === absoluteRoot) {
+    return pkgData.directories.indexOf(absoluteRoot) !== -1;
+  }
+
+  return micromatch(pkgData.directories, babelrcPatterns).length > 0;
 }
 
 /**
--- a/packages/babel-core/src/config/files/configuration.js
+++ b/packages/babel-core/src/config/files/configuration.js
@@ -5,45 +5,33 @@ import path from "path";
 import fs from "fs";
 import json5 from "json5";
 import resolve from "resolve";
-import { makeStrongCache, type CacheConfigurator } from "../caching";
+import {
+  makeStrongCache,
+  makeWeakCache,
+  type CacheConfigurator,
+} from "../caching";
 import makeAPI from "../helpers/config-api";
+import { makeStaticFileCache } from "./utils";
+import type { FilePackageData, RelativeConfig, ConfigFile } from "./types";
 
 const debug = buildDebug("babel:config:loading:files:configuration");
 
-export type ConfigFile = {
-  filepath: string,
-  dirname: string,
-  options: {},
-};
-
-export type IgnoreFile = {
-  filepath: string,
-  dirname: string,
-  ignore: Array<string>,
-};
-
-export type RelativeConfig = {
-  config: ConfigFile | null,
-  ignore: IgnoreFile | null,
-};
-
 const BABELRC_FILENAME = ".babelrc";
 const BABELRC_JS_FILENAME = ".babelrc.js";
-const PACKAGE_FILENAME = "package.json";
 const BABELIGNORE_FILENAME = ".babelignore";
 
 export function findRelativeConfig(
-  filepath: string,
+  packageData: FilePackageData,
   envName: string,
 ): RelativeConfig {
   let config = null;
   let ignore = null;
 
-  const dirname = path.dirname(filepath);
-  let loc = dirname;
-  while (true) {
+  const dirname = path.dirname(packageData.filepath);
+
+  for (const loc of packageData.directories) {
     if (!config) {
-      config = [BABELRC_FILENAME, BABELRC_JS_FILENAME, PACKAGE_FILENAME].reduce(
+      config = [BABELRC_FILENAME, BABELRC_JS_FILENAME].reduce(
         (previousConfig: ConfigFile | null, name) => {
           const filepath = path.join(loc, name);
           const config = readConfig(filepath, envName);
@@ -62,6 +50,23 @@ export function findRelativeConfig(
         null,
       );
 
+      const pkgConfig =
+        packageData.pkg && packageData.pkg.dirname === loc
+          ? packageToBabelConfig(packageData.pkg)
+          : null;
+
+      if (pkgConfig) {
+        if (config) {
+          throw new Error(
+            `Multiple configuration files found. Please remove one:\n` +
+              ` - ${path.basename(pkgConfig.filepath)}#babel\n` +
+              ` - ${path.basename(config.filepath)}\n` +
+              `from ${loc}`,
+          );
+        }
+        config = pkgConfig;
+      }
+
       if (config) {
         debug("Found configuration %o from %o.", config.filepath, dirname);
       }
@@ -75,10 +80,6 @@ export function findRelativeConfig(
         debug("Found ignore %o from %o.", ignore.filepath, dirname);
       }
     }
-
-    const nextLoc = path.dirname(loc);
-    if (loc === nextLoc) break;
-    loc = nextLoc;
   }
 
   return { config, ignore };
@@ -107,7 +108,7 @@ export function loadConfig(
 function readConfig(filepath, envName): ConfigFile | null {
   return path.extname(filepath) === ".js"
     ? readConfigJS(filepath, { envName })
-    : readConfigFile(filepath);
+    : readConfigJSON5(filepath);
 }
 
 const LOADING_CONFIGS = new Set();
@@ -180,26 +181,33 @@ const readConfigJS = makeStrongCache(
   },
 );
 
-const readConfigFile = makeStaticFileCache((filepath, content) => {
-  let options;
-  if (path.basename(filepath) === PACKAGE_FILENAME) {
-    try {
-      options = JSON.parse(content).babel;
-    } catch (err) {
-      err.message = `${filepath}: Error while parsing JSON - ${err.message}`;
-      throw err;
-    }
-    if (!options) return null;
-  } else {
-    try {
-      options = json5.parse(content);
-    } catch (err) {
-      err.message = `${filepath}: Error while parsing config - ${err.message}`;
-      throw err;
+const packageToBabelConfig = makeWeakCache(
+  (file: ConfigFile): ConfigFile | null => {
+    if (typeof file.options.babel === "undefined") return null;
+    const babel = file.options.babel;
+
+    if (typeof babel !== "object" || Array.isArray(babel) || babel === null) {
+      throw new Error(`${file.filepath}: .babel property must be an object`);
     }
 
-    if (!options) throw new Error(`${filepath}: No config detected`);
+    return {
+      filepath: file.filepath,
+      dirname: file.dirname,
+      options: babel,
+    };
+  },
+);
+
+const readConfigJSON5 = makeStaticFileCache((filepath, content) => {
+  let options;
+  try {
+    options = json5.parse(content);
+  } catch (err) {
+    err.message = `${filepath}: Error while parsing config - ${err.message}`;
+    throw err;
   }
+
+  if (!options) throw new Error(`${filepath}: No config detected`);
 
   if (typeof options !== "object") {
     throw new Error(`${filepath}: Config returned typeof ${typeof options}`);
@@ -227,27 +235,6 @@ const readIgnoreConfig = makeStaticFileCache((filepath, content) => {
     ignore,
   };
 });
-
-function makeStaticFileCache<T>(fn: (string, string) => T): string => T | null {
-  return makeStrongCache((filepath, cache) => {
-    if (cache.invalidate(() => fileMtime(filepath)) === null) {
-      cache.forever();
-      return null;
-    }
-
-    return fn(filepath, fs.readFileSync(filepath, "utf8"));
-  });
-}
-
-function fileMtime(filepath: string): number | null {
-  try {
-    return +fs.statSync(filepath).mtime;
-  } catch (e) {
-    if (e.code !== "ENOENT") throw e;
-  }
-
-  return null;
-}
 
 function throwConfigError() {
   throw new Error(`\
--- a/packages/babel-core/src/config/files/configuration.js
+++ b/packages/babel-core/src/config/files/configuration.js
@@ -16,6 +16,8 @@ import type { FilePackageData, RelativeConfig, ConfigFile } from "./types";
 
 const debug = buildDebug("babel:config:loading:files:configuration");
 
+const BABEL_CONFIG_JS_FILENAME = "babel.config.js";
+
 const BABELRC_FILENAME = ".babelrc";
 const BABELRC_JS_FILENAME = ".babelrc.js";
 const BABELIGNORE_FILENAME = ".babelignore";
@@ -83,6 +85,19 @@ export function findRelativeConfig(
   }
 
   return { config, ignore };
+}
+
+export function findRootConfig(
+  dirname: string,
+  envName: string,
+): ConfigFile | null {
+  const filepath = path.resolve(dirname, BABEL_CONFIG_JS_FILENAME);
+
+  const conf = readConfig(filepath, envName);
+  if (conf) {
+    debug("Found root config %o in $o.", BABEL_CONFIG_JS_FILENAME, dirname);
+  }
+  return conf;
 }
 
 export function loadConfig(
--- a/packages/babel-core/src/config/files/index-browser.js
+++ b/packages/babel-core/src/config/files/index-browser.js
@@ -1,17 +1,35 @@
 // @flow
 
-import type { ConfigFile, IgnoreFile, RelativeConfig } from "./configuration";
+import type {
+  ConfigFile,
+  IgnoreFile,
+  RelativeConfig,
+  FilePackageData,
+} from "./types";
 
-export type { ConfigFile, IgnoreFile, RelativeConfig };
+export type { ConfigFile, IgnoreFile, RelativeConfig, FilePackageData };
 
-export function findRelativeConfig(
-  filepath: string,
-  envName: string, // eslint-disable-line no-unused-vars
-): RelativeConfig {
-  return { config: null, ignore: null };
+export function findPackageData(filepath: string): FilePackageData {
+  return {
+    filepath,
+    directories: [],
+    pkg: null,
+    isPackage: false,
+  };
 }
 
-export function loadConfig(name: string, dirname: string): ConfigFile {
+export function findRelativeConfig(
+  pkgData: FilePackageData,
+  envName: string, // eslint-disable-line no-unused-vars
+): RelativeConfig {
+  return { pkg: null, config: null, ignore: null };
+}
+
+export function loadConfig(
+  name: string,
+  dirname: string,
+  envName: string, // eslint-disable-line no-unused-vars
+): ConfigFile {
   throw new Error(`Cannot load ${name} relative to ${dirname} in a browser`);
 }
 
--- a/packages/babel-core/src/config/files/index-browser.js
+++ b/packages/babel-core/src/config/files/index-browser.js
@@ -25,6 +25,13 @@ export function findRelativeConfig(
   return { pkg: null, config: null, ignore: null };
 }
 
+export function findRootConfig(
+  dirname: string,
+  envName: string, // eslint-disable-line no-unused-vars
+): ConfigFile | null {
+  return null;
+}
+
 export function loadConfig(
   name: string,
   dirname: string,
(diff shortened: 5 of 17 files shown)


nicolo-ribaudo:
I like this implementation. Just one thought: should Babel accept `babel.config.json`? Most of the babelrcs I use are just JSON files, so from a user's point of view it seems unnecessary to require to use JavaScript.

billyjanitsch:
It might be confusing for `babel.config.js` to behave differently than `.babelrc.js`, since the widely-used [cosmiconfig](https://github.com/davidtheclark/cosmiconfig#cosmiconfig) treats `foorc.js` and `foo.config.js` identically. For example, postcss (which is often used alongside Babel) allows `postcss.config.js` and `.postcssrc.js` to be used interchangeably, with no difference in lookup semantics.

Maybe there's a better name for this file?

loganfsmyth:
> should Babel accept `babel.config.json`? Most of the babelrcs I use are just JSON files, so from a user's point of view it seems unnecessary to require to use JavaScript.

That's a fair point. I didn't bother because while I'm happy to run JSON5 on `.babelrc`, I don't love the idea of running it on a real `.json` file, and I know people like using comments and such, but we might be able to find a middle-ground. The main downside is that it's one more file that Babel has to search for, which means more calls to fs.stat and slightly slower config loading times.

> It might be confusing for `babel.config.js` to behave differently than `.babelrc.js`

Maybe, but at the end of the day I feel like there are only names that we'd want to encourage. In this case I was definitely mimicing Webpack, which I don't believe does hierarchical lookups. This PR also makes `.babelrc` stop at the package boundary, which that also doesn't appear to do by default. Not sure if postcss adds that though. If people do have name suggestions we can explore it. My goal is very explicitly to make this file _not_ hierarchical though, so I'm not sure anything but documentations can really make that clear.

hzoo:
Is there a way to limit the number of config files we support? Or is that better than having another option in configs?

Options:
- cli for basic stuff like `babel --presets env`
- `.babelrc` as majority
- `package.json` "babel" key
- pass config babel-core/loader directly
- new `.babelrc.js`
- new `babel.config.js` - I think it's going to be confusing some of these act differently unless its like `babel.root.js` or something but that's probably not great either? Or can we remove the new `.babelrc.js`

loganfsmyth:
> Is there a way to limit the number of config files we support?

Good question. At the end of the day, the hierarchical behavior of `.babelrc` is the problem here, so we at least need some way around that. That said, we've also heard from plenty of people over the years that aren't a fan of using dotfiles for build configuration, which I think is a fair complaint.

> .babelrc as majority

Personally I think the majority of people don't need or care about configuring via hierarchy, so it'd be just as easy to recommend `babel.config.js` as the default.

> Or can we remove the new .babelrc.js

I don't know, I feel like the logic there is identical to `.babelrc`, it just happens to be a `.js` file, so I think most people will get it.

I see Babel's config as having a few pieces

* Config passed programmatically (wether babel-loader, or CLI or whatever, they are all the same)
* Config from a `babel.config.js` file that works pretty much like programmatic args, but allows you to load it from a file centrally.
* Config passed based on directory hierarchy, like `.babelrc`/`.babelrc.js`/`package.json#babel` (now limited to within a given package only)

danez:
I'm not sure exactly but this might break react native, because they transpile node_modules, although I'm not sure how they handle the babelconfig for this packages. We should maybe ask someone who is more into react native.

loganfsmyth:
@danez Fair point. The only question would be stopping `.babelrc` resolution at package boundaries. This PR still allows people to do `babelrc: true` to enable resolving babelrc config for every single file (rather than those in the root package), it's just that now project-level babelrcs won't be affect things in node_modules.

hzoo:
cc @zertosh to ping someone on react-native then?

zertosh:
@vjeux who has thoughts on this on the RN side? @hzoo, internally we tend not to use babelrc configs (the ones we do have are really basic and self contained), we tend to use transpire scripts that wire up the config

vjeux:
@yungsters spent some time in that area

hulkish:
i feel like `enhanced-resolve` would really be ideal for this. Would need to write some resolver plugins to facilitate the read-up-but-stop-at logic

michael-ciniawsky:
Why not just using cosmiconfig ? It has a defined set of config (names) which are supported, including `package.json` and it also stops at the first 'valid' config found, which is quite convenient if config lookup starts relative to the file being processed by default (Config Cascading). While specifying a custom path is easy (to start from/or exact) if needed. Stoping at a`package.json` is not always really 1:1 to where a `.babelrc` maybe located and tangles to 'independent' files together e.g the monorepo example with cosmiconfig could work like this (via Config Cascading)

```
|– packages
| | –1
| | |– package.json (with 'babel' field => 'local' config for this package (if needed, stops here))
| |– 2
| | |–package.json (no 'babel' field => use 'global' .babelrc)
|– .babelrc ('Global')
```

> ℹ️ Cascading works with any config 'format'

loganfsmyth:
@michael-ciniawsky 

> Stoping at apackage.json is not always really 1:1 to where a .babelrc maybe located and tangles to 'independent' files together

I agree. That's essentially the downside I called out in the PR description. The issue is that if we don't stop there, where do we stop? We could had yet another option to Babel, but that starts getting extreme. I explicitly want to stop doing "search all the way up" because that has caused users pain in the past, and the monorepo case seems to be the only one where it's potentially painful, and this PR presents an alternative.

> Why not just using cosmiconfig?

That's essentially what we've already had until now, just implemented internally. I'm trying here to add a config file that _doesn't_ cascade because cascading breaks for anyone who wants to symlink packages during local development, and for anyone looking to specific compilation configuration for packages that live inside `node_modules`.

michael-ciniawsky:
I agree. That's essentially the downside I called out in the PR description. The issue is that if we don't stop there, where do we stop?

`process.cwd() > ... > file.dirname` (default lookup range) or an absolute config path (no lookup)

I explicitly want to stop doing "search all the way up" because that has caused users pain in the past, 

What pains ? :)

> That's essentially what we've already had until now, just implemented internally.

`postcss`, `posthtml` `eslint` etc already use `cosmiconfig` => only one package dependency, no direct maintenance, common config format(s)

> I'm trying here to add a config file that doesn't cascade because cascading breaks for anyone who wants to symlink packages during local development, and for anyone looking to specific compilation configuration for packages that live inside node_modules

symlink => specify the absolute path to the config you intend to use
`node_modules` (not sure if I completely follow, but in case it goes in the direction of a 'shareable' config (e.g `eslint-config-*`), =>`require` it in the projects `.babelrc.js` or set the absolute config path ?

loganfsmyth:
> process.cwd() > ... > file.dirname (default lookup range) or an absolute config path (no lookup)

What if the file being compiled isn't even in the working directory? e.g.

```
my-projects/
  some-library/
    index.js
  some-app/
    node_modules/
      some-library -> ../../some-library/
    index.js
```
assuming `some-app` is the working directory. Even if you exposed an option to pass Babel the root, you could have multiple libraries and each one would have its own root, making the ergonomics extremely painful.

> What pains ? :)

Most often its new/inexperienced users who create a `.babelrc` in their home directory or something. Then some time later they create a new project and get babel errors because it is reading the config from outside the project (since they either haven't created one in their new project yet, or are passing the options programmatically instead).

It's note insanely common, but it exposes what I consider to be a pretty big flaw in hierarchical configs.

> postcss, posthtml eslint etc already use cosmiconfig => only one package dependency, no direct maintenance, common config format(s)

What I'm saying with that comment is that cosmicconfig doesn't add anything _new_ functionality-wise. Adding it would not materially affect the objectives that this PR is trying to address, from what I can tell anyway. I'm also not familiar with how any of those projects handle cross-package compilation. 

> symlink => specify the absolute path to the config you intend to use

That's pretty much what this PR is adding with `babel.config.js` except that it loads from the working directory automatically and lets you pass an explicit name as an override.


> node_modules (not sure if I completely follow, but in case it goes in the direction of a 'shareable' config (e.g eslint-config-*), =>require it in the projects .babelrc.js or set the absolute config path

To clarify, for ESLint a configuration is by-definition associated with a package itself. A package defines its own linting rules. Where that breaks down is that for Babel, in the case of node_modules, it is often some root application that wishes to configure the compilation rules. The fact that a given `node_modules` has a `.babelrc` with a set of rules isn't relevant for the root-level application.

Say the user has

```
project/
  .babelrc
  src/
    index.js
  node_modules/
    random-third-party-module/
      .babelrc // random config that was published because the author of the module didn't think to put it in npmignore
      index.js
```
Because the node_modules has a `.babelrc`, hierarchical resolution will not only block the correct config (the `project/.babelrc` one) from loading, it will load a `.babelrc` that is totally unrelated to the project being compiled and will probably cause Babel to throw an exception.

hulkish:
I sort of feel like the find-up and hierarchical babelrc logic may be the source of this challenge.

> What if the file being compiled isn't even in the working directory? e.g.

Couldn't it be argued that this is not a case we can reliably support with auto find up behavior? Maybe, instead of a hierarchical babelrc we allow assignment of paths/patterns to a known babelrc location? Something like webpack's "rules" config structure:

```js
{
  
  configs: [
    { 
       include: [path.resolve('src')],
       use: path.resolve('.babelrc') // or inline config
     }
  ]
}
```

loganfsmyth:
> I sort of feel like the find-up and hierarchical babelrc logic may be the source of this challenge.

Yeah, to me it feels like find-up only makes sense on an individual-package basis.

> Maybe, instead of a hierarchical babelrc we allow assignment of paths/patterns to a known babelrc location? Something like webpack's rules config structure:

With this PR, you can get that by putting those rules into an `overrides` block inside the `babel.config.js` or passed programmatically to Babel.

> Then, for something like babel cli - we would throw an error to use this strategy if we failed to find a babelrc up to or a particular level?

I'm not sure I follow what you mean by that.

michael-ciniawsky:
> What if the file being compiled isn't even in the working directory? e.g.

```
my-projects/
  some-library/
    index.js
  some-app/
    node_modules/
      some-library -> ../../some-library/
    index.js
```

Dependens where the `.babelrc` is located in this example :)

```
my-projects/
  .babelrc (works for some-library && some-app)
  some-library/
    .bablerc (works for some-library only, if not set as absolute path)
    index.js
  some-app/
    .babelrc (works for some-app only, if not set as absolute path)
    node_modules/
      some-library -> ../../some-library/
    index.js
```

> assuming some-app is the working directory. Even if you exposed an option to pass Babel the root, you could have multiple libraries and each one would have its own root, making the ergonomics extremely painful.

If I would need to work with the file tree given in the example I would assume `process.cwd()` (Root) is `my-projects` as the stop-dir. If `some-library` is needed in `some-app` => `node_modules` (as dependency) and otherwise each 'package/lib/app' needs their own `.babelrc`

Where would the `package.json`'s be located in this example ? 

> Most often its new/inexperienced users who create a .babelrc in their home directory or something. Then some time later they create a new project and get babel errors because it is reading the config from outside the project (since they either haven't created one in their new project yet, or are passing the options programmatically instead).

Yeah, it's something one needs to learn and it can be annoying at first, but that's something one at some point definitely needs to understand imho :). Is `babel` merging `.babelrc`'s or looking up the configs when options are already applied programmatically ? Why ?  `programmatically (no lookup) > .babelrc (look up) > Error` ?

> That's pretty much what this PR is adding with babel.config.js except that it loads from the working directory automatically and lets you pass an explicit name as an override.

👍 That's what I meant 

> To clarify, for ESLint a configuration is by-definition associated with a package itself. A package defines its own linting rules.

This is imho also true for transpiling in case the package author uses a `prepare` (prebuild) hook/script and 

> Where that breaks down is that for Babel, in the case of node_modules, it is often some root application that wishes to configure the compilation rules.

1. An author should publish the packages already transpiled (especially if additional stuff is used (plugins, marcos etc))
2. Untranspiled and `babel-preset-env` (`presets`), without an `.babelrc` published by the package itself (without additional plugins, marcos etc, 'just' ES201X => ES201Y)
3. `'root'` and package `.babelrc` use `babel-preset-env` => `.babelrc` (bottom-up lookup) && `babel-preset-env` (`.browserlistrc`) (top-down lookup), while `babel` is also able to (auto) install eventual missing additional `plugins/marcos` for dependencies, if missing in the `root` applications `package.json`
4. Your/other 'unique/exotic' combos might simply break at some point :) => No Support => Refactor
```
|– src
| |– index.js
|
|– node_modules
| |– dep
| | |– .babelrc (preset-env + additional plugins/marcos*)
| | |– .browserslistrc (=> I'm not used in this example :) )
| | |– package.json
| | |– index.js
|
|– .babelrc (preset-env)
|– .browserslistrc
|– package.json
```

```js
{
  'src/index.js': {
     babelrc: '../'
     browserslist: '../'
  },
  'node_modules/dep/index.js'. {
     babelrc: './node_modules/dep'
     browserslist: '../../'
  }
}
```

### `*`

```js
try {
  plugin = require('babel-plugin-*')
} catch (err) {
  const installer = cp.spawn('npm', [ 'i', '-D', 'babel-plugin-*' ])

  installer.on('close', (code) => {
     if (code !== 0)  throw new Error(`... ${code}`)
    
     plugin = require('babel-plugin-*')
  })
}
```

But yeah... 😛  the latter is 💯 learn how, follow the convention, but be done and it's idealistic/naive in the current ecosystem shape, but ¯\_(ツ)_/¯

loganfsmyth:
@michael-ciniawsky 

> An author should publish the packages already transpiled (especially if additional stuff is used (plugins, marcos etc))

I think we all agree that non-standard language functionality should be compiled before publishing, but it is quite clear at this point that the community as a whole is going to adopt language features as they land in Node (LTS anyway). That means over time users looking to compile for browsers are going to need to run `babel-preset-env` on things in `node_modules` more often.

> Untranspiled and babel-preset-env (presets), without an .babelrc published by the package itself (without additional plugins, marcos etc, 'just' ES201X => ES201Y)

You mean the package in `node_modules` should be published with a `.babelrc`? To me that would be a massive foot-gun for users. Most package developers are never going to stop and think "oh I shouldn't publish this .babelrc to npm unless others can use it". A `.babelrc` is an extremely common file and if we went this route and it is published accidentially, it would break the build of every user using that package or that had it as a subdependency. Not to mention, a `.babelrc` and the plugins within it are highly tied to the version of Babel that is trying to use it. Existing packages that have published `.babelrc` files are all going to reference Babel 6 plugins that may well not work on 7.x. If we consider published `.babelrc` files as part of the public API of a module from Babel's perspective, it would essentially require us to never make a breaking change to Babel's APIs ever again, and Babel's API is _huge_.

> 'root' and package .babelrc use babel-preset-env => .babelrc (bottom-up lookup) && babel-preset-env (.browserlistrc) (top-down lookup), while babel is also able to (auto) install eventual missing additional plugins/marcos for dependencies, if missing in the root applications package.json

Automatic installation of packages isn't something I can see us ever wanting. If a config is so complicated that a user can't install it themselves, to me that points to an architectural problem in the config loading system itself. Avoiding that is a big part of why I'm doing this.

Doing `.browserlistrc` in this way would also break once again in the symlink case because the `.browserlistrc` wouldn't be in the parent folder of `dep` at all.


Going back to the example structure

```
my-projects/
  .babelrc
  some-library/
    .bablerc
    package.json
    index.js
  some-app/
    .babelrc
    webpack.config.js
    package.json
    index.js
    node_modules/
      some-library -> ../../some-library/
      other-library/
        .babelrc
        package.json
        index.js
```
First off, to clarify, what I'm talking about here is a user configuration where `some-app` has a webpack.config.js with something like

```
rules: [{
    test: /\.js$/,
    loader: 'babel-loader',
}]
```

Webpack will attempt to pass _all_ JS files in a project. In this world, a user is conceptually trying to compile "some-app" with a set of dependencies. As far as they are concerned, they don't particularly care what configuration is in individual node_modules. They just want to say "compile all standards-compliant code to ESX" based on their own config in `some-app/`. Because of hierarchical lookup, there is literally no way for config options inside `my-projects/some-app/.babelrc` to affect `my-projects/some-library/index.js`. It is entirely unrealistic to expect users to create a `my-projects/.babelrc` file since that would likely be outside of a version controlled folder, making it a maintenance nightmare. This PR allows creating `my-projects/some-app/babel.config.js` that would be loaded and apply to _all_ files passed to babel-loader.

The core tenants of `.babelrc` files after this PR are that

* Users generally don't want individual `node_modules` dependencies to be able to configure their own compilation process in the first place.
* When they do, they can opt into it specifically, and accept that they are then required to install whatever plugins that `.babelrc` reference (which may or may not even be possible depending on directory structure).
* If they _do_ opt into it and the module doesn't have a `.babelrc`, picking one up from a random parent directory is never what they'd expect.


> But yeah... 😛 the latter is 💯 learn how, follow the convention, but be done and it's idealistic/naive in the current ecosystem shape, but ¯_(ツ)_/¯

I'm not sure I follow 100%, but I'd say being idealistic/naive is just what Babel has been doing up until now, and this is my attempt to fix that so our users have options for usecases that are quite painful currently, but are also becoming more and more common.

loganfsmyth:
Woops, missed these:



> Yeah, it's something one needs to learn and it can be annoying at first, but that's something one at some point definitely needs to understand imho :). Is babel merging .babelrc's or looking up the configs when options are already applied programmatically ? Why ? programmatically (no lookup) > .babelrc (look up) > Error ?

Babel does not merge multiple `.babelrc`s but it does merge programmatic options with `.babelrc` options. Users can opt for only programmatic args by passing `babelrc: false`.

> That's pretty much what this PR is adding with babel.config.js except that it loads from the working directory automatically and lets you pass an explicit name as an override.

👍 That's what I meant

I thought you wanted `babel.config.js` to be search for hierarchically too? If we did that, it would require users to pass an explicit path for the config to apply consistently to dependencies, which doesn't seem ideal to me.

michael-ciniawsky:
Could `node_modules/` not better be skipped anyways (by default), instead of two different configs and match common naming conventions `--config`, `pkg.babel`, `.babelrc.json`, `babelrc.js`, `babel.config.js` (similiar to `cosmiconfig`) but besides `JSON/JS` behave the same ? Just read everything again, makes perfectly sense now. Final questions :)

1. Two different config loading systems ? (This is really confusing)
2. Merging programmatically and config file by default ?
3. Still support config lookup (excluding `node_modules`) relative to the file, if nothing else is setup (stops at `root` (`process.cwd()`) by default) ?

hulkish:
@loganfsmyth 

> Then, for something like babel cli - we would throw an error to use this strategy if we failed to find a babelrc up to or a particular level?

> I'm not sure I follow what you mean by that.

Meaning; if babel cli fails to find `.babelrc` in the cwd/wherever (assuming you intend to keep that logic) - then we throw an error recommending dev user explicitly defines where config file is located.

loganfsmyth:
@michael-ciniawsky 

> Could node_modules/ not better be skipped anyways (by default)

In the cases where compilation of `node_modules` might come up, like in `babel-loader`, Babel isn't generally in control of what files are passed to it. 

> Two different config loading systems ?

I guess I don't really think of them as two different systems. They are all the same config format, it's just varies in how they are loaded.

> Merging programmatically and config file by default ?

I'm not really looking to change that, since it is existing. We've made improvements to _how_ the merging works in Babel 7 at least, so it should be more user-friendly than Babel 6.

> Still support config lookup (excluding node_modules) relative to the file, if nothing else is setup (stops at root (process.cwd()) by default) ?

I'm not sure I understand what you mean by `if nothing else is setup` here.

loganfsmyth:
@hulkish 

> Meaning; if babel cli fails to find .babelrc in the cwd/wherever (assuming you intend to keep that logic) - then we throw an error recommending dev user explicitly defines where config file is located.

I'm not really looking to change the existing behavior of `.babelrc` files beyond limiting their scope. Doing what you ask isn't really workable because the `babel` command lets you compile a whole directory, and it's allowed that some files in a directory could find a `.babelrc` and some might not, depending on how the directory is structured. And even if we could, it could only be babel-cli specific, which likely isn't that useful for users.

> eslint approaches this with a concept of "overrides"

Support for this has landed in Babel as well: https://github.com/babel/babel/pull/7091

> They also have a concept of "root"

Babel already stops searching once it finds a `.babelrc`, so every `.babelrc` essentially acts as if it contained `root: true` in the ESLint sense.

michael-ciniawsky:
In the cases where compilation of node_modules might come up, like in babel-loader, Babel isn't generally in control of what files are passed to it.

This could be optional in the way it's implemented in this PR (Array of module paths). (Whitelist `node_modules/path/to/module` paths, but `node_modules/*` is excluded by default (from config lookup perspective)). 

> I'm not sure I understand what you mean by if nothing else is setup here.

```js
{
   test: /\.js$/
   // no options ('inline' or a config path) passed (nothing setup => parse => stringify (useless))
   // => config lookup from file.dirname
   // => if file.dirname.includes('node_modules'), start from node_modules upwards
   // if any options are passed, there wouldn't be any lookup at all
   // that's how e.g `(posthtml|postcss)-loader` handles it atm 
   // (but without any special node_modules handling)
   use: [ 'babel-loader' ]
}
```

But anyways... :) the solution proposed in this PR is reasonable and good as it is, it's already just nitpicking about eventual details without much additional benefit.

I personally wouldn't introduce two different config loader modes if avoidable, but this maybe required by `babel`.

@loganfsmyth Thx for your time and patience :)

packages/babel-core/src/config/config-chain.js
```diff
@@ -148,6 +172,43 @@ export function buildRootChain(
...
+      const inDependency =
+        match
+          .pop()
+          .split(path.sep)
+          .indexOf("node_modules") !== -1;
+
+      if (!inDependency) return true;
+    }
+
+    return !!micromatch.capture(path.pattern, filename);
```
nicolo-ribaudo: What is `path.pattern`?

loganfsmyth: Looks like just a bug. Should be `pattern`.

packages/babel-core/src/config/config-chain.js
```diff
@@ -115,11 +116,34 @@ export function buildRootChain(
...
+    configFileObj = loadConfig(configFile, context.cwd, context.envName);
+  } else {
+    try {
+      configFileObj = loadConfig(
+        "./babel.config.js",
+        path.resolve(context.cwd, root),
+        context.envName,
+      );
+    } catch (e) {
+      if (e.code !== "MODULE_NOT_FOUND") throw e;
```
Kovensky: Annoyingly, this may actually eat resolving errors that come from dependencies of the config file itself (e.g. if it tries to require a plugin to pass by value instead of passing it by name).

We probably need to do a `require.resolve` check first to ensure any exception we ignore is because the config isn't present.

loganfsmyth: Good call. When I switch this up to be real production-ready code, I'll see what I can do.

loganfsmyth:
I think the main place I'd like feedback on this is the `babelrc` option. This PR makes it a string by default and that string is a pattern (or array of string patterns). If the pattern matches the any parent folder inside the package of the file being compiled, it will search for a `.babelrc` relative to that.

I think it makes sense, but if someone things of something clearer I'd love to hear it.

packages/babel-cli/src/babel/index.js
```diff
@@ -240,10 +240,6 @@ if (errors.length) {
 //
 
 const opts = commander.opts();
-//the configFile CLI option maps to the extends option in the node API
-if (opts.configFile) {
```
hzoo: didn't know this was an option before but I see from the comment it was just extends

loganfsmyth: Yeah it landed back in https://github.com/babel/babel/pull/6133 but I've always wanted to kill it and make us have a proper config file instead :)

packages/babel-core/README.md
```diff
@@ -216,7 +216,9 @@ Following is a table of the options you can use:
 | `ast`                    | `true`               | Include the AST in the returned object |
 | `auxiliaryCommentAfter`  | `null`               | Attach a comment after all non-user injected code |
 | `auxiliaryCommentBefore` | `null`               | Attach a comment before all non-user injected code |
-| `babelrc`                | `true`               | Specify whether or not to use .babelrc and .babelignore files. Not available when using the CLI, [use `--no-babelrc` instead](https://babeljs.io/docs/usage/cli/#babel-ignoring-babelrc) |
+| `root`                   | `"."`                | Specifiy the "root" folder that defines the location to search for "babel.config.js", and the default folder to allow `.babelrc` files inside of.|
```
novemberborn: Small typo here in `Specifiy`.

danez:
Instead of having `babelrc` being a string, why not using extends? If a user knows which directory to search in, they probably know which file it is.

In general I do not really like the idea of having to configuration files which just by having a different name have a different behaviour. I think this is super confusing even if we document this well. I would rather have an option inside the configuration file that changes the behaviour, similar to eslints `root: true` option. Also I have never seen this before in any other tool.

Why don't we instead of stopping at package.json add a option `root: true` that needs to be added to the .babelrc file in the root directory and we stop there. If not found then we warn a user and fallback to the current behaviour?


My intention is that it should be easy to setup for example a monorepo where the `babelrc` file is in the repo-root. This will not work out of the box without lookking up how config resolution works in babel and which config files you have to place where but imho it should just work.

loganfsmyth:
@novemberborn 

> perhaps I missed some discussion, but could you elaborate on what this does? "Parent folder inside the package"? Is there an assumption of a root value here?

Currently the root defaults to the working directory, so we'd only search for `.babelrc` files inside whatever package the working directory points at.

loganfsmyth:
@danez 

First, right now `.babelrc` files are strictly resolved relative to the file being compiled. This means they have 2 limitations:

* A `.babelrc` at the root of your project isn't guaranteed to apply to things in `node_modules`, e.g.

    ```
    project/
      .babelrc
      node_modules/some-package/
        .babelrc
        index.js
    ```
    In essentially every case we'd want to support, reading the inner `.babelrc` file would be incorrect as the default behavior, but because `.babelrc` files are strictly hierarchical, it would always find the inner `.babelrc`.

* A `.babelrc` at the root of your project cannot apply to linked packages, e.g.

    ```
    some-package/
      index.js
    project/
      .babelrc
      node_modules/some-package -> ../../some-package
    ```

> Why does the config in the root has to be babel.config.js why can't this file be .babelrc

For this to work, we'd essentially be redefining the behavior of `.babelrc` files in the project root to no longer be hierarchical, because we'd have to load it up front without considering the path of the file being compiled.

I personally don't like the idea of overloading the behavior of `.babelrc` to have two different resolution mechanisms depending on what directory the `.babelrc` is inside of. It would also mean for instance if you did

```
project
  .babelrc (1)
  src/
    .babelrc (2)
    client/
      .babelrc (3)
      index.js
```
It is easy to say "deeper `.babelrc` files take precedence". If we redefine the behavior of root-level `.babelrc` files to always load, it also means that a case like the above example would now load _both_ `.babelrc` (1) _and_ (3), which I think is harder to explain.

> This option needs to be set programatically in the case of transpiling node_modules anyway

I don't follow this point. Since the option defaults to the working directory, `.babelrc` files in `node_modules` would be ignored by default, so most times configuration shouldn't be needed.

> in my opinion is confusing as then you probably end up having a .babelrc file and a babel.config.js

We could certainly have it error if you have both a `.babelrc` and a `babel.config.js` in your root. I'd support that.

danez:
As far as I understand the concept here is: 
Search for `.babelrc` hierarchical from the sourcefile and stop at package.json if not found. Then if not found search for babel.config.js in root directory which defaults to cwd. 

Is that correct?

So all that changes is that the hierarchical search ends earlier and additionally searches in the root for `babel.config.js` and that is why I don't see why it has to be exactly this file and not a `.babelrc`. Because then this `.babelrc` file would be picked up hierarchical for my files in the `src` folder and also for transpilation of `node_modules` (although I guess I don't want babel to search in that case anyway as I want to use my config not the the one from the package in node_modules.).

loganfsmyth:
> Search for .babelrc hierarchical from the sourcefile and stop at package.json if not found. Then if not found search for babel.config.js in root directory which defaults to cwd.

Not quite. It will load the `babel.config.js`, or whatever config file the user set, either way. The search for `.babelrc` files will happen only if the file being compiled is inside of the `root` package.

> In my view there is nothing that changes for .babelrc files lookup besides stoping earlier and an additional lookup in the root. I don't understand what you mean with overload.

If we go that route, the additional lookup at root doesn't happen hierarchically, which means searching for `.babelrc` now has two different behaviors. This PR splits that into two separate filenames with a clear definition of how  each type is resolved, instead of overloading `.babelrc` to serve both cases.

If we have both types of searching either way, to me  having them both use the same filename seems super confusing.

packages/babel-core/src/config/files/configuration.js
```diff
@@ -5,45 +5,35 @@ import path from "path";
...
-  filepath: string,
-  dirname: string,
-  ignore: Array<string>,
-};
-
-export type RelativeConfig = {
-  config: ConfigFile | null,
-  ignore: IgnoreFile | null,
-};
+const BABEL_CONFIG_FS_FILENAME = "babel.config.js";
```
Andarist: why the `FS` in the constant's name? is it for file system? seems weird that only this one has anything other than default name + filename suffix in its identifier

loganfsmyth: That's definitely a typo of `JS` hahah

novemberborn:
> It will load the babel.config.js, or whatever config file the user set, either way. The search for .babelrc files will happen only if the file being compiled is inside of the root package.

@loganfsmyth is this an accurate summary?

* Attempt to resolve `babel.config.js` in the `root`
* For files inside the `root`, but not in any way inside `node_modules`, hierarchically search for `.babelrc` until the `root` is encountered

I'm confused as to what you're proposing the `babelrc` option will do.

simonpai:
Just my 2 cents:
Perhaps it will also be helpful if you export `findRelativeConfig` as an API, so several dependent package (like babel-loader, babel-plugin-module-resolver, etc) don't need to reimplement and keep up with the same logic?

Some references:
[babel-loader implementation](https://github.com/babel/babel-loader/blob/master/src/config.js)
[babel-plugin-module-resolver implementation](https://github.com/tleunen/find-babel-config)

loganfsmyth:
Sorry, been pretty busy with other tasks.

@novemberborn 

> I'm confused as to what you're proposing the babelrc option will do.

Yeah I've gone back and forth on whether to have that honestly. Essentially what this PR allows is for `babelrc` to be a string (defaulting to the value of `root`), _or_ an array. So for your comment

> For files inside the `root`, but not in any way inside `node_modules`, hierarchically search for .babelrc until the `root` is encountered

It is totally accurate, but it would be "for files inside any of the `babelrc` values, if it is a single string, or an array of strings.

loganfsmyth:
@simonpai 

In fact we have! Check out #7472. I will say I think the usage in `babel-plugin-module-resolver` can be entirely removed in Babel 7 anyway, because we now tell plugins the path of the config file that loaded them.

novemberborn:
@loganfsmyth 

> Essentially what this PR allows is for babelrc to be a string (defaulting to the value of root), or an array. So for your comment
> 
>> For files inside the root, but not in any way inside node_modules, hierarchically search for .babelrc until the root is encountered
> 
> It is totally accurate, but it would be "for files inside any of the babelrc values, if it is a single string, or an array of strings.

Ah so Babel will hierarchically search for `.babelrc` files until any of the paths in the `babelrc` option is encountered. So you could segment the search boundaries inside your package on a per-directory basis. Am I getting that right?

loganfsmyth:
Yeah. I was trying to address the usecase of potentially explicitly _wanting_ to use `.babelrc` files from other projects, like if you have a bunch of your own modules that you've connected with `npm link` or something, I could imagine wanting each of them to optionally be able to have `.babelrc` files to configure themselves, or something. So the main caller of `babel-loader` or something could say "these 4 linked modules should also be considered when searching for `.babelrc` files".

novemberborn:
> So the main caller of babel-loader or something could say "these 4 linked modules should also be considered when searching for .babelrc files".

So somehow Babel determines that it won't search for `.babelrc` files for those linked modules. How does that happen? Is it because they're outside of the `root`?

And then `babelrc: []` is a way of allowing the `.babelrc` search to happen for files inside the matching directories?

How is this different from an `overrides` with a custom `root` option?

hzoo:
We can do another beta with this change in?

packages/babel-core/src/config/files/configuration.js
danez: Let's fix that before merging.

loganfsmyth:
Yeah, I think I'll have time this afternoon to do a pass over this to fix the outstanding issues and write up some documentation about how config loading works in Babel 7.

packages/babel-core/README.md
```diff
@@ -216,7 +216,9 @@ Following is a table of the options you can use:
...
+| `configFile`             | `undefined`          | The config file to load Babel's config from. Defaults to searching for "babel.config.js" inside the "root" folder. `false` will disable searching for config files.|
+| `babelrc`                | `(root)`             | Specify whether or not to use .babelrc and .babelignore files. Not available when using the CLI, [use `--no-babelrc` instead](https://babeljs.io/docs/usage/cli/#babel-ignoring-babelrc). `false` to disable searching, and `true` to always search, a string path of the package to search inside of, or an array of paths to packages to search inside of. |
```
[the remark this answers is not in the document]
loganfsmyth: Maybe we could add that validation in `babel-loader`? I don't think it should be the responsibility of core itself since it would be extra work every time Babel runs, and wouldn't be applicable for any other callers.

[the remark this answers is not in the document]
danez: Oh yeah sorry I was only talking about the loader and did not mean to change anything here.

loganfsmyth:
We could also consider leaving `babelrc` as a boolean and making `babelRoots` a separate argument, if people wanted to. @danez Thoughts? That would avoid the conflict with babel-loader and might be clearer for users anyway.

novemberborn:
@loganfsmyth I think that'd be a lot clearer, yea.

danez:
@loganfsmyth That sounds good and I also think the name explain the new option better.

lukescott:
This has broken out build process, and I have spent 2 days trying to figure it out. Still haven't found a solution. Our projects are monorepos, and we use Yarn workspaces. It's generally load out like this:

```
node_modules/ <--- common packages are hoisted here
packages/one/node_modules
packages/one/.babelrc
packages/one/.babelrc
packages/two/node_modules
packages/two/.babelrc
```

There is no root .babelrc. Each package has their own .babelrc. When package `one` depends on package `two`, yarn sym links to node_modules.

Before this would just work. Now package `two` does not load `.babelrc`. I've tried doing a babel.config.js file in the root, but it doesn't seem to pick that up. If I put it in the `one` package (which is our working directory) it complains that babelrcRoots is `is only allowed in root programmatic options`. So I wrap that in a function, and it wants a whole mess of other options.

What's frustrating about this is it took something really simple to configure and made it a heck of a lot more complicated. I still haven't figured out how to fix this, and I have lost a lot of time trying to figure it out.

loganfsmyth:
> What's frustrating about this is it took something really simple to configure and made it a heck of a lot more complicated. I still haven't figured out how to fix this, and I have lost a lot of time trying to figure it out.

I understand that it is frustrating, but you've chosen to opt into our beta version, and the issues this PR is attempting to address are in fact one of the primary blockers that were making us stay beta, because we knew they would break people one way or the other.

With all that said, I'm more than happy to help you out and also see what we can do to make things clearer, or if logic can be changed in Babel itself to be less painful for users, which may well be the case. I see you posted a bit on Slack yesterday, but at that time, especially since it is a weekend, no-one appears to have been around to help. Let's continue the discussion there, if that's alright with you.

lukescott:
Logan was able to help me figure this out. Thank you so much! Just thought I would post the solution here, in case anyone else runs into the same issue.

My setup is we have a monorepo with a couple main packages. We use yarn workspaces. Other than the main two, we have shared packages the two utilize. Each of the main packages has a `webpack.config.js` file. Each package has a `.babelrc` file. We run our build within each main package, not in the root. We also have no config files in the root.

The solution is put this in `babel-loader`'s options:

```
babelrcRoots: ["../*"], // which is ../../packages/*
```

If you try to use babelrcRoots in a .babelrc or even in babel.config.js as a plain object (not returned in a function) it will tell you that it `is only allowed in root programmatic options`. This basically means, in at least my situation, put it in the webpack config under the loader options. You'll also need this in each separate build process (like for tests).

loganfsmyth:
It would probably not be too hard to allow that option in the root of a `babel.config.js` file so that you could share it across different build tools. I'll file an issue for that.

salper:
Let's say I have the following project structure:
```
foo/
  .babelrc
  index.js
  package.json
bar/
  .babelrc
  index.js
  package.json
```
Here's the code in `foo/index.js`:
```javascript
import bar from '../bar'

console.log(bar())
```
Then, what I used to do was:
```
cd foo && yarn babel-node index.js
```
As I understand, it does not work due to the previous considerations, as the `bar/.babelrc` cannot be loaded. The thing is that the `babelrcRoots` can only be used programmatically. So I can make it work for example by doing this:
```javascript
// foo/my-babel-node
#!/usr/bin/env node

require('@babel/register')({ babelrcRoots: true, ignore: [/node_modules/] });
require(`${__dirname}/${process.argv[2]}`)
```
But I suppose you may have think of something smarter than this, for example using `babel.config.js` and the `overrides`. Or by the previous comment concerning `babelrcRoots` in the `babel.config.js`. But that would mean that I should run the `yarn babel-node` from the `/` and not from `/foo`, or having a 'foo/babel.config.js' no ?

salper:
Soooo... I have found a way to do as wanted without a programmatic approach:
```
foo/
  babel.config.js
  index.js
  package.json
bar/
  babel.config.js
  index.js
  package.json
```
In `foo/babel.config.js`:
```javascript
module.exports = {
  overrides: [
    {
      test: '*',
      presets: [
        ['@babel/preset-env', {'node': 'current'}],
      ],
    },
    {
      test: '../modules/*',
      extends: '../bar/babel.config.js'
    },
  ]
}
```
In `bar/babel.config.js`:
```javascript
module.exports = {
  overrides: [
    {
      test: '*',
      presets: [
        ['@babel/preset-env', {'node': 'current'}],
      ],
    },
    {
      test: '../modules/*',
      extends: '../foo/babel.config.js'
    },
  ]
}
```
Then I can:
```
cd foo && yarn babel-node --ignore node_modules index.js
```
The `ignore` option is needed otherwise the default programmatic ones are used (`only`). I have also tried a more centralized approach:
```
foo/
  babel.config.js
  index.js
  package.json
bar/
  babel.config.js
  index.js
  package.json
babel.config.js
package.json
```
In `foo/babel.config.js`:
```javascript
  extends: '../babel.config.js'
```
In `bar/babel.config.js`:
```javascript
module.exports = {
    extends: '../babel.config.js'
}
```
In `babel.config.js`:
```javascript
module.exports = {
  overrides: [
    {
      test: 'foo/*',
      presets: [
        ['@babel/preset-env', {'node': 'current'}],
      ],
    },
    {
      test: 'bar/*',
      extends: '../foo/babel.config.js'
    },
  ]
}
```

CH-RhyMoore:
I'm not opted into the beta version, this has broken my build, I don't find the documentation of very helpful so far (particularly not in terms of what I need to change in my monorepo to get it working the way it did yesterday), and I find the behavior thoroughly undesirable and would vastly prefer cascading config that applies to sibling files like eslint's.

lukescott:
@CH-RhyMoore add a `babel.config.js` file to the root of the monorepo with the following:
```js
module.exports = {
	babelrcRoots: ["./packages/*"],
}
```
Assuming you run commands from the root. If you run a command from within a package, add this to each package:
```js
module.exports = {
	babelrcRoots: ["../*"],
}
```
`babel.config.js` is picked up in the current working directory. It also gets picked up by build tools like webpack.

CH-RhyMoore:
Okay, that is simple enough. Thank you very, very much! 

Still, my 2 cents--eslint really has always had my favorite configuration experience. Personal preference plays a part here, but I've always found it very intuitive.

CH-RhyMoore:
I might have another issue going on at the same time, potentially with Webpack 4? Because that alone isn't fixing it. Scenario:

```
|_ project/
   |- babel.config.js
   |- package.json
   |_ packages/
      |_ package/
         |- .babelrc
         |- babel.config.js
         |- webpack.config.babel.js
```

```
/project/packages/dls-falcor/webpack.config.babel.js:1
(function (exports, require, module, __filename, __dirname) { import path from 'path';
                                                                     ^^^^
SyntaxError: Unexpected identifier
```

But searching that sent me back here. Oh well, I'll keep looking.

lukescott:
It may have something to do with the `modules` on the`@babel/env` preset. Webpack 4 can handle ES6 module syntax. Hard to say what you need to do exactly, but that's likely where the problem is.

CH-RhyMoore:
Thing is, we use `es2015` for some of the ones that are failing. Our monorepo also currently mixes Babel 6 and Babel 7.

#### project root
```
|_ project/
   |- babel.config.js
```
```
const lernaConfig = require('./lerna.json');

module.exports = {
  babelrcRoots: ['.'].concat(lernaConfig.packages),
  rootMode: 'upward-optional'
};
```

#### package with build script

##### `.babelrc`

```
|_ project/
   |_ packages/
      |_ package-a/
         |- .babelrc
```
```
{
  "presets": ["es2015"]
}
```

```
|_ project/
   |_ packages/
      |_ package-b/
         |- .babelrc
```
```
{
  "presets": [["env"]],
  "plugins": ["lodash", "transform-runtime"]
}
```

##### `babel.config.js`

```
|_ project/
   |_ packages/
      |_ package-a/
         |- babel.config.js
      |_ package-b/
         |- babel.config.js
```
```
const lernaConfig = require('./../../lerna.json');

module.exports = {
  babelrcRoots: ['.', '../*'].concat(lernaConfig.packages),
  rootMode: 'upward-optional'
};
```

Tried a few different combinations of this. With and without `'.'` in `babelrcRoots`. With and without the project root `babel.config.js`. With and without the package-level `babel.config.js`. With and without `'rootMode'`. With and without `.concat(lernaConfig.packages)` in the package-level `babel.config.js`.

The symptoms started yesterday in Jest with our source code and now are happening on the Webpack configuration file (though I didn't initially connect that error to this one). Everything worked last week and we didn't actually update anything on purpose in the interim. Same `.babelrc` file position and contents, same version of Webpack, same version of Babels...within the constraints of Lerna's use of `package-lock.json` with hoisting, which I believe I've previously observed as being looser than one might prefer.

CH-RhyMoore:
FYI, getting a 523 at that Slack link. Was trying to peek at it in case I felt brave.

lukescott:
The `es2015` preset is deprecated. I think even in Babel 6. As far as Jest is concerned, I'm not sure if they've updated to Babel 7 or not, but if they are pointing to the old `babel-core` package, it's just a peer dependency so you can install `7.0.0-bridge.0`. I'm not sure you really can mix babel 6 and 7.

The Slack link: Looks like the host that does the redirect is down. Was working earlier today. I don't have the direct link. I copied it from the project README.

CH-RhyMoore:
In a monorepo where each package manages its own build, it should absolutely be possible. Why wouldn't it be? npm is quite capable of installing multiple versions of the same package when their semver ranges aren't compatible, Lerna is quite capable of not getting in the way of npm doing so, and Node.js is quite capable of resolving the version needed for a files in a particular package. 

(We've been doing so without any issues for ~2 months. Our packages produce sufficiently different things that their builds aren't consistent enough to raise up to the project level at this point. We'll do that at some later point when we see which consistent patterns emerge.)

My biggest issue wasn't Babel related directly at all, I don't believe. For some reason, the update to Webpack 4 we did 15 days ago suddenly began requiring us to install `@babel/register`|`babel-register` and pass it as an argument (`webpack -r @babel/register`|`webpack -r babel-register`) in order to use ES `import`/`export` in our `webpack.config.babel.js` file. Previously, simply using the file extension `babel.js` was sufficient. This was also necessary for `webpack-dev-server`. While the `-r` option is documented, I couldn't find any documentation of this change in functionality.

I think that issue is entirely about how Webpack does or doesn't run Babel on its own configuration file. So FYI for the future, if someone mentions the errors coming _from their Webpack config file_, it _could_ be this issue.

But by the time I finally found that, I'd already upgraded a half dozen or more dependencies in each package (times 20) to use Babel 7 in everything (and update plugins and then the tools that use the plugins to versions capable of handling it). We have updating on our roadmap for this month, so I here I am working late, because I thought I might as well continue, and I hope to god I'm close to finishing.

I've had to change a multitude of other things to get this far in migrating from Babel 6 to 7 today, but, strangely, `babel.config.js`-related configuration was not one. I deleted all of them and everything that's not broken for another reason is still working. 

So it seems if you set up your monorepo like ours, you _may not need to do anything about the new configuration behavior_. Though it's possible it may yet arise as an issue. I haven't even gotten to Jest yet today since I'm still trying to get everything in the project building.

Issue #7210, opened 2018-01-13
Duplicate declaration error for exported functions with object destructuring arguments 


MrOrz:
I have a function that takes a complicated options object as its only argument.
Also, I am using [object destructuring with default values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Default_values_2) to set default values, which are actually callbacks with similar signatures.

The parameters of the callbacks should be independent of each other. However, babel 6.26.0 is still giving a "Duplicate declaration" error.

### Input Code
```js
export function createConnectionType({
  resolveTotalCount = ({ first, ...searchContext }) => {},
  resolveEdges = ({ first, ...searchContext }) => {},
}) {}
```
REPL: http://bit.ly/2CVCWuB

### Babel/Babylon Configuration (.babelrc, package.json, cli command)

```js
# .babelrc
{
  "presets": [
    ["env", {
      "targets": {
        "node": "current"
      }
    }],
    "stage-3"
  ]
}
```

```
# Versions
$ node_modules/.bin/babel --version
6.26.0 (babel-core 6.26.0)
```

### Expected Behavior

It should compile to something like:

```js
export function createConnectionType({
  resolveTotalCount,
  resolveEdges,
}) {
  if(!resolveTotalCount) resolveTotalCount = (_ref) => {const {first} = _ref; const searchContext = _objectWithoutProperties(_ref)};
  if(!resolveEdges) resolveEdges = (_ref) => {const {first} = _ref; const searchContext = _objectWithoutProperties(_ref)};
}
```

### Current Behavior

When babel-preset-env is enabled, it gives compile error
```
repl: Duplicate declaration "searchContext"
  1 | export function createConnectionType({
  2 |   resolveTotalCount = ({ first, ...searchContext }) => {},
> 3 |   resolveEdges = ({ first, ...searchContext }) => {},
    |                               ^
  4 | }) {}
```

When babel-preset-env is disabled, it gives:
```
/* (_objectWithoutProperties omitted) */
export function createConnectionType(_ref) {
  let {
    resolveTotalCount = ({ first }) => {},
    resolveEdges = ({ first }) => {}
  } = _ref,
      searchContext = _objectWithoutProperties(_ref.resolveTotalCount, ["first"]),
      searchContext = _objectWithoutProperties(_ref.resolveEdges, ["first"]);
}
```

Note that searchContext in the two default values are totally messed up.

[...]

PR #7364, opened 2018-02-10
Don't extract rest elements from nested expressions

nicolo-ribaudo:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/7210
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

The problem was that the plugin didn't only traverse the destructuring pattern when searching for rest elements, but also default values.

I suggest to review this PR ignoring whitespaces (https://github.com/babel/babel/pull/7364/files?w=1)


packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -9,15 +9,37 @@ export default function(api, opts) {
...
-        path.stop();
-      },
+    visitRestElements(path, () => {
+      foundRestElement = true;
+      return /* stop */ true;
     });
     return foundRestElement;
   }
 
+  function visitRestElements(path, visitor) {
```
Andarist: I realize that this is probably more performant solution, but the code itself is imho a little bit convoluted, hard to read and to reason about. It has a lot of indirection in it. Would it be possible to rephrase it into a simpler implementation?

packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -11,33 +11,26 @@ export default function(api, opts) {
...
-        if (type === "AssignmentPattern") next([path.get("left")]);
-        else if (type === "ArrayPattern") next(path.get("elements"));
-        else if (type === "ObjectPattern") next(path.get("properties"));
-        else if (type === "ObjectProperty" && !path.node.shorthand) {
-          next([path.get("value")]);
-        } else if (type === "RestElement") {
-          const stop = visitor(path);
-          if (stop) return;
+    path.traverse({
+      Expression(path) {
```
Andarist: this is way easier to track! ❤️

Andarist:
@nicolo-ribaudo I think this caused the merge conflict now for you - #7388

PR #7370, opened 2018-02-12
Added support for passing-through modules transform plugin options

### LINKED ISSUE #6624: Support `noInterop` option for modules
> Issue originally reported by @avaly in babel/babel-preset-env#297

Version `6.24.0` shipped a new `noInterop` option for `babel-plugin-transform-es2015-modules-amd` and `babel-plugin-transform-es2015-modules-commonjs`: https://github.com/babel/babel/pull/5427

Currently, the env preset does not forward this option to the modules plugins.

hulkish:
| Fixed Issues?            | Fixes #6624 
| Documentation PR         | No, but includes docs

Added support to allow passing through modules transform plugin options via the `@babel/preset-env` `modules` option.

### DIFF
--- a/packages/babel-preset-env/README.md
+++ b/packages/babel-preset-env/README.md
@@ -231,7 +231,7 @@ Enable ["loose" transformations](http://2ality.com/2015/12/babel6-loose-mode.htm
 
 ### `modules`
 
-`"amd" | "umd" | "systemjs" | "commonjs" | "cjs" | false | [<module type>, <plugin options>]`, defaults to `"commonjs"`.
+`"amd" | "umd" | "systemjs" | "commonjs" | "cjs" | false | [<module type>, <plugin options>]`, defaults to `"commonjs" (which resolves to ["commonjs", { "loose": false }])`.
 
 Enable transformation of ES6 module syntax to another module type.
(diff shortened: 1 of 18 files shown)


Andarist:
While this is certainly useful, I feel like we should establish how to pass options to any underlaying plugin.

hulkish:
@Andarist I agree, while working through this one - I kind of established that `@babel/preset-env` could probably use some more thinking through on how to tackle this in a way that has a lower likelihood in the future for constant change to this config structure.

For example, In a way this sort of begs the question "if dev user needs this, then why not force them to do so by adding the plugin manually themselves and using `modules: false` for this preset?"

Andarist:
> For example, In a way this sort of begs the question "if dev user needs this, then why not force them to do so by adding the plugin manually themselves and using modules: false for this preset?".

That's actually what I've been doing recently in my projects, because I do not want `loose` for this particular single transform (because it makes `__esModule` enumerable in cjs).

> Maybe remove modules transforms all together from this preset? Is the only reason it's there is to forward the loose option?

preset-env has all plugins necessary to transpile down to es5 environments and that requires modules transform.

PR #7372, opened 2018-02-12
Mark hoisted react constant elements as #__PURE__

Andarist:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-transform-react-constant-elements/src/index.js
+++ b/packages/babel-plugin-transform-react-constant-elements/src/index.js
@@ -1,4 +1,5 @@
 import { types as t } from "@babel/core";
+import annotateAsPure from "@babel/helper-annotate-as-pure";
 
 export default function transformReactConstantElement(api, options) {
   const { allowMutablePropsOnTags } = options;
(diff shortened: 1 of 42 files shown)


packages/babel-plugin-transform-react-constant-elements/src/index.js
```diff
@@ -101,7 +102,11 @@ export default function transformReactConstantElement(api, options) {
         path.traverse(immutabilityVisitor, state);
 
         if (state.isImmutable) {
-          path.hoist();
+          const hoisted = path.hoist();
+          // instanceof NodePath
+          if (hoisted instanceof path.constructor) {
```
jridgewell: Why not `NodePath`? Is the `instanceof` even necessary?

Andarist: just lazy before hitting the bed and I didn't search from where I could import this

`path.hoist()` can return undefined, that is a guard against that - I guess I could reverse the condition and remove 'dep' on the `instanceof` altogether

jridgewell: If it's only `undefined` or a path, why not just truthy check `hoisted`?

```js
if (hoisted) {
  //...
}

Andarist: Fixed

PR #7375, opened 2018-02-13
Empty src dir error fix

### LINKED ISSUE #7337: Babel CLI returning non-zero if directory is empty?
From https://github.com/gatsbyjs/gatsby/pull/3785#discussion_r164922293. Haven't investigated but assuming it is accurate, it doesn't seem ideal to me. If I do
```
babel src -d lib
```
I'd say as long as the `src` directory exists, the fact that it has no files in it shouldn't be considered an error.

rajzshkr:
| Fixed Issues?            |Fixes #7337 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | 
| Documentation PR         | 
| Any Dependency Changes?  |


xtuc:
Could you please add a test case for that?

packages/babel-cli/src/babel/dir.js
```diff
@@ -59,6 +59,13 @@ export default function(commander, filenames, opts) {
     return path.join(commander.outDir, filename);
   }
 
+  function outputDestFolder(commander) {
```
Andarist: this function probably doesnt need `commander` as its argument, just `outDir`

loganfsmyth:
Made a tweak to fix the tests, but looks good otherwise!

PR #7379, opened 2018-02-14
Solves Tagged template literal size optimization

### LINKED ISSUE #7352: Tagged Template Literal: Size Optimization
Just noticed with https://github.com/babel/babel/issues/7350, we're making a bunch of very long string arrays. What's more, they're (mostly) duplicated between the `strings` and `raw`.

```js
tag`some template`;
// "some template" "some template"
_taggedTemplateLiteral(["some template"], ["some template"]);
```

The only (I think) time that string arrays aren't duplicated is when an "invalid" escape sequence comes up (which are now valid, due to template literal revision).

```js
tag`test${1}\xhello`;
// "invalid" string becomes undefined
_taggedTemplateLiteral(["test", void 0], ["test", "\\xhello"]);
```

At worst, we could optimize the "all valid strings" case using slice:

```js
// default raw value to a slice of strings
function _taggedTemplateLiteral(strings, raw = strings.slice()) {
  return Object.freeze(Object.defineProperties(strings, {
    raw: { value: Object.freeze(raw) }
  }));
}

// all valid
tag`some template`;
_taggedTemplateLiteral(["some template"]);

// some invalid
tag`\xhello`;
// "invalid" string becomes undefined
_taggedTemplateLiteral([void 0], ["\\xhello"]);
```

But, realistically, we could just reuse the same string array (it's being frozen, anyways):

```js
function _taggedTemplateLiteral(strings, raw) {
  return Object.freeze(Object.defineProperties(strings, {
    raw: { value: raw ? Object.freeze(raw) : strings }
  }));
}
```

debugpai2:
| Fixed Issues? | Fixes #7352 

1. Adds a check to see if strings and raw strings are equal
2. If they are equal doesn't pass raws to call expression
3. Modifies taggedTemplateHelpers to slice strings if raws is not passed

### DIFF
--- a/packages/babel-plugin-transform-template-literals/src/index.js
+++ b/packages/babel-plugin-transform-template-literals/src/index.js
@@ -54,6 +54,9 @@ export default function(api, options) {
         const strings = [];
         const raws = [];
 
+        // Flag variable to check if contents of strings and raw are equal
+        let isStringsRawEqual = true;
+
         for (const elem of (quasi.quasis: Array)) {
           const { raw, cooked } = elem.value;
           const value =
@@ -63,6 +66,11 @@ export default function(api, options) {
 
           strings.push(value);
           raws.push(t.stringLiteral(raw));
+
+          if (raw !== cooked) {
+            // false even if one of raw and cooked are not equal
+            isStringsRawEqual = false;
+          }
         }
 
         // Generate a unique name based on the string literals so we dedupe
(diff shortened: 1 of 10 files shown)


packages/babel-plugin-transform-template-literals/src/index.js
```diff
@@ -81,10 +89,15 @@ export default function(api, options) {
...
-            t.arrayExpression(raws),
-          ]);
+
+          // only add raw arrayExpression if there is any difference between raws and strings
+          const init = t.callExpression(
+            helperId,
+            [
+              t.arrayExpression(strings),
+              !isStringsRawEqual ? t.arrayExpression(raws) : null,
+            ].filter(x => x),
```
jridgewell: Let’s hoist out the array and push `raws` in inside an if-statement.

packages/babel-plugin-transform-template-literals/test/fixtures/default/simple-tag/input.js
```diff
@@ -0,0 +1 @@
+var foo = tag`wow`;
```
jridgewell: Let’s add a multi quasi case:

```js
var bar = tag`first${1}second`;

debugpai2: made this change

packages/babel-plugin-transform-template-literals/src/index.js
debugpai2: Made this change

PR #7383, opened 2018-02-15
TypeScript: support mapped type modifiers syntax

andy-ms:
| Fixed Issues?            |
| Documentation PR         |
| Any Dependency Changes?  |

Supports the syntax from Microsoft/TypeScript#21919. Now you can write `let x: { +readonly [P in string]-?: number };`.

### DIFF
--- a/packages/babel-generator/src/generators/typescript.js
+++ b/packages/babel-generator/src/generators/typescript.js
@@ -280,6 +280,7 @@ export function TSMappedType(node) {
   this.token("{");
   this.space();
   if (readonly) {
+    tokenIfPlusMinus(this, readonly);
     this.word("readonly");
     this.space();
   }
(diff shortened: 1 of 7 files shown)


packages/babylon/src/types.js
```diff
@@ -1161,9 +1161,9 @@ export type TsIndexedAccessType = TsTypeBase & {
 
 export type TsMappedType = TsTypeBase & {
   type: "TSMappedType",
-  readonly?: true,
+  readonly?: true | "+" | "-",
   typeParameter: TsTypeParameter,
-  optional?: true,
+  optional?: true | "+" | "-",
```
DanielRosenwasser: Funky but functional.

existentialism: Funky indeed 😆

packages/babel-generator/src/generators/typescript.js
```diff
@@ -302,6 +304,12 @@ export function TSMappedType(node) {
   this.token("}");
 }
 
+function tokenIfPlusMinus(self, tok) {
+  if (tok !== true) {
+    self.token(tok);
```
xtuc: Sorry, what do you mean by `tok`? Token?

PR #516, opened 2015-01-16
Playground Proposal: Mallet operator

jridgewell:
The mallet operator is similar to the current memoization operator,
except it can be used outside of just objects.

In Ruby, it’s almost the same as `a = a || b`. Note that only `nil` and
`false` are falsey in Ruby. I’ve defined it as `== null`, though that
could be changed to any JS falsey value.

PR #7385, opened 2018-02-16
Proposal: Logical Assignment Operators

jridgewell:
https://github.com/jridgewell/proposal-logical-assignment, which I'll hopefully be [presenting in March](https://github.com/tc39/agendas/pull/318).

I'm bringing it [back](https://github.com/babel/babel/pull/516). 😉


packages/babylon/src/tokenizer/index.js
```diff
@@ -464,9 +464,14 @@ export default class Tokenizer extends LocationParser {
     const next = this.input.charCodeAt(this.state.pos + 1);
 
     if (next === code) {
+      const assign =
+        this.hasPlugin("logicalAssignment") &&
+        this.input.charCodeAt(this.state.pos + 2) === charCodes.equalsTo;
```
nicolo-ribaudo: Can you use `this.expectPlugin`?

jridgewell: Done.

hzoo:
I want to be able to test this in the REPL haha, should we add it to stage 0 or make a -1 lol..

jridgewell:
I don't think stage 0 can be blocked, so my proposal is de-facto stage 0. I'll add it to the preset.

hzoo:
Yeah -1 only relevant in that it hasn't been presented, but doesn't matter too much if it will be soon. Either way :+1:, I guess we should re-think how we signal stage x presets (given there's been a lot of talk about removing them entirely)

nicolo-ribaudo:
Oh, I forgot to ask you to add this plug-in to https://github.com/babel/babel/blob/master/packages/babel-core/src/transformation/util/missing-plugin-helper.js

jridgewell:
Done.

PR #7388, opened 2018-02-16
Fix over-zealous traversal by object-rest-spread

jamesreggio:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Prior to this change, we'd conduct an open-ended traversal on the 'id' of any VariableDeclarator to find a RestElement. The 'id' of a VariableDeclarator can contain an AssignmentPattern (to supply
a default value), and if the right-hand side of the AssignmentPattern contained a RestElement, we'd transform it.

The problem here is that the right-hand side of an AssignmentPattern can be *any* Expression. If the right-hand side is a function body, we'd traverse the entire function body, and if a RestElement occurred anywhere in that function body, we'd transform it and emit the transformations wherever we began the traversal (at least one scope outside its usage).

The fix is to stop the inner traversal if we encounter an AssignmentPattern. The outer traversal will still visit the AssignmentPattern, so RestElements within the right-hand side of an AssignmentPattern will be properly transformed at that time.

**This is my first time ever working with the Babel/Bablyon codebase, so I'd appreciate some extra scrutiny on my logic here. I'm afraid there may be some edge-case I've wrongly short-circuited.**

### Repro

**In**

```js
function fn0(obj0) {
  const {
    fn1 = (obj1 = {}) => {
      const {
        fn2 = (obj2 = {}) => {
          const {a, ...rest} = obj2;
          console.log(rest);
        }
      } = obj1;
    }
  } = obj0;
}
```

**Broken output**

Notice that the `rest = babelHelpers.objectWithoutProperties(obj0, ["a"]);` call is on the outermost scope, where the inner traversal began.

```js
function fn0(obj0) {
  const {
    fn1 = (obj1 = {}) => {
      const {
        fn2 = (obj2 = {}) => {
          const {
            a
          } = obj2;
          console.log(rest);
        }
      } = obj1;
    }
  } = obj0,
        rest = babelHelpers.objectWithoutProperties(obj0, ["a"]);
}
```

**Fixed output**

```js
function fn0(obj0) {
  const {
    fn1 = (obj1 = {}) => {
      const {
        fn2 = (obj2 = {}) => {
          const {
            a
          } = obj2,
                rest = babelHelpers.objectWithoutProperties(obj2, ["a"]);
          console.log(rest);
        }
      } = obj1;
    }
  } = obj0;
}
```


nicolo-ribaudo:
This PR introduces a regression:
```js
const { x: { ...a } = {} } = {};
```

Also, it doesn't account for rest elements in computed properties name:
```js
const { [{ ...foo} = {}]: x } = {};
```

They are both fixed by #7364

jamesreggio:
Sorry about that, @nicolo-ribaudo. This was my first contribution to Babel and I explicitly asked for additional scrutiny because I was afraid that I may have been over-zealous in my traversal skipping.

I appreciate you fixing this quickly.

nicolo-ribaudo:
No worries 😉

PR #7401, opened 2018-02-20
Bump compat-table and regen preset-env data

existentialism:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Precursor to updating preset-env with es2018 stuff.

~~~Note, this actually changes `esmodules: true` a bit, since `caniuse` [updated `es6.modules`](https://github.com/Fyrd/caniuse/blob/master/features-json/es6-module.json), unless we don't care about the `nomodules` attribute?~~~

/cc: @kristoferbaxter

EDIT: Adds a workaround in `scripts/build-modules-support` to allow Safari 10.1/iOS 10.3

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+package.json
--- a/packages/babel-preset-env/data/built-ins.json
+++ b/packages/babel-preset-env/data/built-ins.json
@@ -1149,7 +1149,7 @@
   "es7.promise.finally": {
     "chrome": "63",
     "firefox": "58",
-    "safari": "tp",
+    "safari": "11.1",
     "opera": "50"
   }
 }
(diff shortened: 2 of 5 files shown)


kristoferbaxter:
Will take a look today. 

First thought: Concerned that for Safari 10.1 `<script type=module>` could contain JS that it cannot parse with this change.

existentialism:
Yeah, we can just as easily allow `a #4` for safari/ios_saf

kristoferbaxter:
I'd call that the best work around for now. Perhaps a note too explaining why "a #4" counts as "y" for safari.

kristoferbaxter:
LGTM

PR #7406, opened 2018-02-21
WIP: rename proposal-object-rest-spread to transform-object-rest-spread

dnalborczyk:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | 👍
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  | 👍

object rest spread is part of the official ECMA syntax now. this is a breaking change.

_not sure what the babel/semver policy is around renaming plugins regarding breaking changes in a beta release. I kinda got the feeling reading through some github issues/PRs that it should be admissible._

I got a couple questions on how to progress:

I'm not sure about how to handle dependencies, e.g.: https://github.com/babel/babel/blob/master/packages/babel-standalone/package.json#L21 or https://github.com/babel/babel/blob/master/packages/babel-preset-env/package.json#L15, since the change hasn't been published yet. if I change the reference, _bootstrapping_ the project would not work, hence the build breaks.

_edited: removed question regarding yearly preset._


dnalborczyk:
follow up from https://github.com/babel/babel/pull/7409

regarding deprecated yearly stage presets, I suppose we would still need to do any of the following to avoid confusion:

applies to object-rest-spread and async generator functions:

1. leave them in stage-3 preset, but rename both (proposal -> transform)
2. remove both

hzoo:
Cool!

Ok a few things: We haven't really decided on the rename policy yet but my suggestion was that we probably need to rename it once it's Stage 4.

Can you run the tests locally before you push? The tests take a while to run and it blocks other PRs as well. As for the packages, you just need to rename everything and run `make bootstrap` and it will work. Part of that command links the packages so it's unrelated to if it's published because it uses the package in the repo.

dnalborczyk:
@hzoo  cool, thanks, will do!

Kovensky:
Object Rest/Spread has been Stage-4 since before this PR was made https://github.com/tc39/ecma262/pull/1048 🤔

Anyway, we need to figure out how to do this rename in practice. Would need someone familiar with the build system...

nicolo-ribaudo:
After some discussion (https://github.com/babel/babel/pull/9106#issuecomment-443351884), we decided that it is not wort renaming plugins, since people should use `@babel/preset-env` anyway. We might reconsider it in Babel 8. Thank you for your help anyway!

PR #7408, opened 2018-02-21
WIP: rename proposal-async-generator-functions to transform-async-generator-functions

dnalborczyk:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | 👍
| Minor: New Feature?      |
| Tests Added + Pass?      |
| Documentation PR         | 
| Any Dependency Changes?  |
| License                  | MIT

waiting for feedback on https://github.com/babel/babel/pull/7406

PR #7409, opened 2018-02-21
WIP: stage 2018 preset

dnalborczyk:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | 👍 `-> we're removing object rest/spread and async generator functions from stage 3 preset`
| Minor: New Feature?      |
| Tests Added + Pass?      |
| Documentation PR         | 
| Any Dependency Changes?  |

based on https://github.com/babel/babel/pull/7406 and https://github.com/babel/babel/pull/7408

### DIFF
--- a/packages/babel-preset-es2018/.npmignore
+++ b/packages/babel-preset-es2018/.npmignore
@@ -1,0 +1,3 @@
+src
+test
+*.log
--- a/packages/babel-preset-es2018/README.md
+++ b/packages/babel-preset-es2018/README.md
@@ -1,0 +1,35 @@
+# @babel/preset-es2018
+
+> Babel preset for all es2018 plugins.
+
+## Install
+
+```sh
+npm install --save-dev @babel/preset-es2018
+```
+
+## Usage
+
+### Via `.babelrc` (Recommended)
+
+**.babelrc**
+
+```json
+{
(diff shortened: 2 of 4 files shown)


TrySound:
@dnalborczyk I thought this kind of modules are deprecated in favor of `babel-preset-env`. Do we need new deprecated package?

dnalborczyk:
@TrySound  good point, if that is the case this can be closed.

hzoo:
Thanks for the PR!

>  I suppose we would still need to rename both in the stage-3-package, or, better, remove them entirely. it would be quite confusing otherwise - as well as adjust the preset-env package.

I'm confused? Yeah we are planning on making any yearly presets as I have described in https://babeljs.io/blog/2017/12/27/nearing-the-7.0-release + deprecated yearly presets

hzoo:
Going to close

dnalborczyk:
@hzoo  as, you and @TrySound mentioned, yearly presets are deprecated.

I suppose we would still need to do any of the following to avoid confusion:

_(same applies to async generator functions)_

1. leave the v7 beta.40 in stage-3-preset package as is (proposal)
2. leave them in stage-3, but rename both (proposal -> transform)
3. remove both

PR #7411, opened 2018-02-22
Add "use strict" directive

### LINKED ISSUE #7349: Ensure that class bodies have 'use strict' if the parent scope wasn't already
ES6 class scopes are automatically strict-mode. In Babel 6 the module transform pretty much just _always_ made modules strict, but now that people are more likely to be using `sourceType: "unambiguous"`, we'll want to make sure the class transform injects `use strict` if the file isn't a module and doesn't have a `use strict` directive in a parent scope.

### Input Code

```js
class Foo {
  method(){

  }
}
```

### Expected Behavior
```
var Foo = function () {
  "use strict";
  function Foo() {
    _classCallCheck(this, Foo);
  }

  _createClass(Foo, [{
    key: "method",
    value: function method() {}
  }]);

  return Foo;
}();
```

### Possible Solution

We likely want to walk up the parent paths, and any time a BlockStatement that is a function body is passed, we can check if `path.node.directives` has a `use strict` or not. If we get all the way to the top and didn't find one, and the root `Program` node doesn't have `sourceType === "module"`, we should add a Directive node to the classes wrapper function expression.

MarkusToe:
| Fixed Issues?            | Fixes #7349  
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

I added the "use strict" directive in the function body BlockStatement.


xtuc:
Thanks for your pr!

Could you please add a test case where the parent already has the `"use strict";`?

```js
"use strict";

class Foo {
  method(){

  }
}
```

should be 

```js
"use strict";

var Foo = function () {
  function Foo() {
    _classCallCheck(this, Foo);
  }

  _createClass(Foo, [{
    key: "method",
    value: function method() {}
  }]);

  return Foo;
}();
```

MarkusToe:
@xtuc I guess I would have to check if the program already has a `"use strict";` directive? But I'm not sure where to look for it. I was trying to check `this.node.directives` around L165 but it is `undefined`. Do you have any tips for where to check if the program already has a `"use strict";` directive?

xtuc:
I think we could do it with `path.findParent(node => void)`, my understanding is that you need to check if one of your parent nodes already has a `use strict` directive (not only in the top-level scope).

packages/babel-plugin-transform-classes/src/vanilla.js
```diff
@@ -163,10 +163,27 @@ export default class ClassTransformer {
...
+      container = t.arrowFunctionExpression(
+        closureParams,
+        t.blockStatement(body),
+      );
+    } else {
+      container = t.arrowFunctionExpression(
+        closureParams,
+        t.blockStatement(body, [t.directive(t.directiveLiteral("use strict"))]),
+      );
+    }
```
xtuc: These two blocks looks really similar, maybe we could add on a condition for the injection of the directive.

Something like:

```js
const directives = [];

if (parentHasUseStrict) {
    directives push t.directive(t.directiveLiteral("use strict"))
}

t.blockStatement(body, directives)
```

packages/babel-plugin-transform-classes/test/fixtures/loose/program-with-use-strict-directive/input.js
```diff
@@ -0,0 +1,7 @@
+"use strict";
+
+class Foo {
+    method() {
+        
+    }
+}
```
xtuc: Test showing the case where the directive is already preset.+1: 

Could you also please add a test where the directive isn't in the program but in a parent node? like that:

```js
function t() {
  "use strict";
  class Foo {
    method() {}
  }
}
```

packages/babel-plugin-transform-classes/src/vanilla.js
```diff
@@ -163,10 +163,24 @@ export default class ClassTransformer {
...
+    );
+    const parentHasUseStrict = parent.node.directives.some(
+      directive => directive.value.value === "use strict",
+    );
+
+    const directives = [];
+
+    if (!parentHasUseStrict) {
+      directives.push(t.directive(t.directiveLiteral("use strict")));
+    }
```
xtuc: Nit: we are matching two times the `use strict` string, we could construct the node `t.directiveLiteral("use strict")` before. Just an idea.

packages/babel-plugin-transform-classes/src/vanilla.js
```diff
@@ -163,10 +163,23 @@ export default class ClassTransformer {
     //
     body.push(t.returnStatement(t.cloneNode(this.classRef)));
 
-    const container = t.arrowFunctionExpression(
-      closureParams,
-      t.blockStatement(body, [t.directive(t.directiveLiteral("use strict"))]),
-    );
+    const parent = this.path.findParent((path) => path.isProgram() || path.isBlockStatement());
```
Andarist: wont it check only the nearest parent? we can have really deep tree and I guess we should check for something like:
```js
const strictParent = this.path.findParent((path) => {
  if (!path.isProgram() && !path.isBlockStatement()) {
    return false
  }

  return path.node.directives.some(directive => directive.value.value === 'use strict')
});
```

packages/babel-plugin-transform-classes/src/vanilla.js
```diff
@@ -163,10 +163,24 @@ export default class ClassTransformer {
     //
     body.push(t.returnStatement(t.cloneNode(this.classRef)));
 
+    const parent = this.path.findParent(
```
Andarist: ive left comment regarding this [here](https://github.com/babel/babel/pull/7411#discussion_r171531339), it got hidden by github though (i think i might have commented on some older commit)

packages/babel-plugin-transform-classes/src/vanilla.js
xtuc: Yes that absolutely right! Thanks @Andarist

packages/babel-plugin-transform-classes/src/vanilla.js
```diff
@@ -163,10 +163,27 @@ export default class ClassTransformer {
     //
     body.push(t.returnStatement(t.cloneNode(this.classRef)));
 
+    const strictParent = this.path.findParent(path => {
```
Andarist: looks better now 👍 would be great to add tests to cover this

also would be great to skip adding those directives in modules, but im not entirely sure how to check that, maybe you can check some property on the program or file? there are things like `sourceType` which could hint you this info, im not sure though where this info resides

xtuc: I'm not sure this is actually testable. Having the directive in the program or in a parent function produces the same result.

Kovensky: You can test by writing a random non-simple arguments list (destructuring, defaults) in the IIFE the transform generates and seeing if it compiles.

`'use strict'` inside a function is too late to disable sloppy mode in the arguments list.

Andarist: Not sure if we have understood each other - I was asking if there is a possibility to skip adding this for such input
```js
```
import a from './a'
class B {
  constructor() { this.c = a() }
}
```

Modules are already strict be definition and they do not have to add use strict directive explicitly.

Comment (loganfsmyth) [reply packages/babel-plugin-transform-classes/src/vanilla.js]: Agreed. We'll want to make sure to avoid adding it to module. It should be safe to do 

```
if (path.isProgram() && path.node.sourceType === "module") {
  return true;
}
```
to this to cover the module case.

**Comment by Kovensky** [comment]
We definitely need to always create an IIFE then. Otherwise, it won't be possible to have argument defaults or destructuring in the constructor's arguments.

**Comment by Andarist** [comment]
@Kovensky could u give example of problematic input? I've looked into constructor with default argument case and I don't see why it wouldn't be possible to skip the IIFE in such case.

**Comment by loganfsmyth** [comment]
@Andarist For instance of we converted

```
class Foo {
  constructor(arg = {}) {}
}
```
to
```
let Foo = function(arg = {}) {
  "use strict";
};
```
that in fact throws an error because non-simple parameters need to _already_ be strict by the time the parser reaches them, if you're going to try to make them strict.

We could likely skip the IIFE and keeping the simple function, as long as we special-case the non-simple function params to still create the IIFE.

**Comment by Andarist** [comment]
Ok, that's an interesting case. As you said - we could try to continue optimizing this and de-opt on non-simple params **when** strict directive has to be added to the class (which in reality should be in minority of cases). I would be for optimizing it as it shouldn't be too complicated code-wise and I'm a little weirdo who likes optimized outputs 😅 dunno how you & other feel about such cases though.

Also I'm wondering why this:
```js
let Foo = function(arg = {}) {};
```
doesn't throw? 🤔 Why it starts to throw once we add a strict directive in the function's body?

loganfsmyth:
> Also I'm wondering why this: ... doesn't throw?

There's nothing wrong with having non-simple parameters when non-strict. The problem is that `use strict` in the body of a function affects the _whole_ function, including the parameters. But because `use strict` is in the _body_ of the function, there is a whole ton of work done by the parser before it has reached that point.

The issue essentially is that if your code is non-strict and you're parsing a whole function and then suddenly hit a `use strict` in the body, you'd have to rewind and redo all of the processing for the function a second time. For example

```
function fn(
  arg = function() {
    var let;
  }
) {
  "use strict";

}
```
when the parser encounters `var let`, should it error, or not? If `use strict` were allowed in the function body like this, it's impossible to say when the statement was parsed, because even if it isn't strict yet, it would be possible that later on it would suddenly _become_ strict.

Andarist:
@loganfsmyth thanks for clarifying, makes sense 😄

MarkusToe:
@loganfsmyth I am not sure how to handle the constructor only cases - could you please give me some guidance on where to start and what to check for?

So I guess at first we have to add the "use strict" directive to constructor-only function bodies, and then we should check for non-simple arguments and not add the "use strict" directive?

Andarist:
I think there are such cases:
- already in strict mode - just create "only constructor"
- not in strict, we need to add it
a) simple arguments - add it inside the "only constructor"
b) non-simple arguments - fallback, create IIFE with strict directive and return "only constructor"

loganfsmyth:
How can I help us get this across the finish line? You've done some great work and I feel bad that it keep getting conflicts.

jridgewell:
Pushed a rebase, and included the constructor-only and non-simple parameter edge case.

nicolo-ribaudo:
@jridgewell  Can you move the `strictParent` function to be reusable in the other PR? e.g. Maybe `path.scope.isStrict()`.
Also, you should revert the changes to `lerna.json`

jridgewell:
Thanks for the core of the work @MarkusToe!

loganfsmyth:
@MarkusToe No problem at all. Thanks for getting the work on this moving in the first place!

PR #7413, opened 2018-02-22
Update babel to beta.40

existentialism:


hzoo:
lol we could use `babel-upgrade` for these?

PR #7417, opened 2018-02-23
Rely entirely on sourceType for module vs script differentiation.

loganfsmyth:
| Fixed Issues?            | Fixes #7416
| Patch: Bug Fix?          | 
| Major: Breaking Change?  | Yes (Within 7 yes, from 6 it depends on if we disable `allowCommonJSExports` by default)
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Per #7416, we're in a weird place right now in Babel 7 because we've landed some logic that tries to be smart about whether something is a module or a script. The two points that affect this issue are:

* If something is a ES6 module, we automatically add `use strict` and convert `this` to `undefined`.
* In `transform-runtime` (and `transform-async-to-generator` when using `module` and `method` options), we try to insert either `require` or `import` based on the type of file.

So this puts us in a weird place. 

In Babel 6, we just assumed it was _always_ an ES6 module, which leads to breakage with Webpack because if `transform-runtime` inserts an import, but the user's file was CommonJS, Webpack will consider the whole file an ES6 module and the CommonJS stuff will break.

In Babel 7 we rectified that in the short term by basically saying it's an ES6 module if __both__:

* It has `sourceType: module`
* It has at least one `import` or `export`.

Unfortunately this means that if you want something to be an ES6 module and it _doesn't_ have an import or export, it's actually impossible, hence #7416.

To resolve this, we've introduced `sourceType: "unambiguous"` that performs the above logic at the _parser_ level, so instead of checking those 2 states during transformation, the parser produces either a `script` or `module`, which means that Babel itself can respect the `sourceType` as a source of truth.

With that work landed in the parser, this PR removes the logic from the transformation system. The downside here being that we don't want to default to `sourceType: unambiguous` because it isn't a formal specification and isn't backed by anything in particular. Instead we'd like to leave that up to the user to opt into.

This means that with this PR, Babel will revert to it's Babel 6 behavior as mentioned above for `use strict` and `this` rewriting.

I personally think that is fine. The main question I have is, should we try to make it more painful to treat a CommonJS module this way (auto-injecting use-strict and this->undefined) to encourage people to explicitly set `sourceType` based on what they actually want? 

This could easily amount to us injecting code to throw a nice error saying "please set `sourceType:script` or `sourceType:unambiguous` in your Babel config". We already have an `allowCommonJSExports` option on the CommonJS transform that we could easily disable by default, though the message will need updating to be more helpful. If we _don't_, it'll just be like Babel 6 if someone runs `transform-runtime` on a CommonJS module, with `modules:false` for Webpack they'll get unhelpful error messages about `module` and `exports` being undeclared.

### DIFF
--- a/packages/babel-core/test/api.js
+++ b/packages/babel-core/test/api.js
@@ -206,6 +206,7 @@ describe("api", function() {
 
     function execTest(passPerPreset) {
       return babel.transform("type Foo = number; let x = (y): Foo => y;", {
+        sourceType: "script",
         passPerPreset: passPerPreset,
         presets: [
           // First preset with our plugin, "before"
--- a/packages/babel-helper-module-imports/src/import-injector.js
+++ b/packages/babel-helper-module-imports/src/import-injector.js
@@ -209,7 +209,7 @@ export default class ImportInjector {
     // to a variable.
     let name = nameHint || importName;
 
-    const isMod = isModule(this._programPath, true);
+    const isMod = isModule(this._programPath);
     const isModuleForNode = isMod && importingInterop === "node";
     const isModuleForBabel = isMod && importingInterop === "babel";
 
--- a/packages/babel-helper-module-imports/src/is-module.js
+++ b/packages/babel-helper-module-imports/src/is-module.js
@@ -1,11 +1,7 @@
 /**
- * A small utility to check if a file qualifies as a module, based on a few
- * possible conditions.
+ * A small utility to check if a file qualifies as a module.
  */
-export default function isModule(
-  path: NodePath,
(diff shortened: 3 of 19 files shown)


nicolo-ribaudo:
> We already have an allowCommonJSExports option on the CommonJS transform that we could easily enable by default, though the message will need updating to be more helpful

Isn't it already enabled by default? I think that it shouldn't, since it is better to throw soon than making webpack throw later with a less descriptive error.

loganfsmyth:
@nicolo-ribaudo Good catch, that's just a typo, definitely meant disable by  default.

PR #7418, opened 2018-02-23
Avoid re-traversing inserted references to the namespace binding.

### LINKED ISSUE #7176: 'lazy: true' on commonjs transform fails when namespace and named both in use
Due to the option added in https://github.com/babel/babel/pull/6952:

### Input Code

```js
import * as mod from "mod";
import { named } from "mod";

named;
mod;
```

```js
{
  "plugins": [
    ["transform-modules-commonjs", {"lazy": true}],
    "external-helpers"
  ]
}

```

### Expected Behavior
```
"use strict";

function mod() {
  const data = babelHelpers.interopRequireWildcard(require("mod"));

  mod = function () {
    return data;
  };

  return data;
}

mod().named;
mod();
```

### Current Behavior
The final two lines of the result are instead
```
mod()().named; // incorrectly double-wrapped
mod(); // correct
```

### Context

This is because, when importing a namespace, we try to re-use the name of the namespace, e.g. `mod` in this case from the `* as mod`, as the name of the local variable, so things are easier when debugging.

Unfortunately this is causing issues during traversal because it converts `named;` to `mod().named`, and then the traversal continues, and it seems `mod` used as a function and converts it to `(0, mod())()` a second time.

### Solution

Not quite settled on the right fix for this. Really this is another example where traversal just isn't a very nice way to mutate the tree. Ideally we'd traverse for all the paths to change, and then change them afterward so we don't risk re-traversing a node that the traversal itself has inserted.

loganfsmyth:
| Fixed Issues?            | Fixes #7176
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Nice simple fix for https://github.com/babel/babel/issues/7176

### DIFF
--- a/packages/babel-helper-module-transforms/src/rewrite-live-references.js
+++ b/packages/babel-helper-module-transforms/src/rewrite-live-references.js
@@ -202,6 +202,10 @@ const rewriteReferencesVisitor = {
       }
 
       requeueInParent(path);
+
+      // The path could have been replaced with an identifier that would
+      // otherwise be re-visited, so we skip processing its children.
+      path.skip();
     }
   },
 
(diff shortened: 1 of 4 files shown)


loganfsmyth:
This is super tiny, so just landing it :P

PR #7428, opened 2018-02-24
Prevent undefined function name for template literal computed property

### LINKED ISSUE #7199: undefined is compiled as function name by using template literal as method name.
## BUG REPORT

As metioned on the title I give an example:

```javascript
var methodName = 'test';
var json = {
  [`${methodName}`]: function (){
    
  }
}
```
is compiled to following code:
```javascript
var methodName = 'test';
var json = _defineProperty({}, '' + methodName, function undefined() {}); // function name is undefined

```

Commonly it's ok, but if I use object deconstruction in method, the compiling result will cause undefined overwriting. For example:

### Current Behavior
```javascript
var methodName = 'test';
var json = {
  [`${methodName}`]: function (){
    let o = {};let {a='2'}=o
    console.log( a )
  }
}
```
is compiled to following code:
```javascript
var methodName = 'test';
var json = _defineProperty({}, '' + methodName, function undefined() {
  var o = {};var _o$a = o.a,
      a = _o$a === undefined ? '2' : _o$a; // undefined in this line is referred to current function

  console.log(a);
});

```

In above code, variable "a" will not be assigned to 2, because _o$a is not equal to "undefined", which is referred to current function; I think following Expected Behavior is proper behavior.

### Expected Behavior
```javascript
var methodName = 'test';
var json = _defineProperty({}, '' + methodName, function () { // no undefined name
  var o = {};var _o$a = o.a,
      a = _o$a === undefined ? '2' : _o$a; 

  console.log(a);
});

```

### Babel Configuration (.babelrc, package.json)
```js
{
  "presets": ["env","react"],
  "plugins": ["syntax-dynamic-import"]
}
```

```js
{
  "name": "dynamic-import",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "build": "webpack",
    "start": "webpack-dev-server --open"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "babel-cli": "^6.26.0",
    "babel-core": "^6.26.0",
    "babel-loader": "^7.1.2",
    "babel-polyfill": "^6.26.0",
    "babel-preset-env": "^1.6.1",
    "babel-preset-react": "^6.24.1",
    "clean-webpack-plugin": "^0.1.17",
    "html-webpack-plugin": "^2.30.1",
    "webpack": "^3.10.0",
    "webpack-dev-server": "^2.9.7"
  },
  "dependencies": {
    "react": "^16.2.0",
    "react-dom": "^16.2.0"
  }
}

```

### Environment
| ---------------- | -------
| Babel            | see above Babel Configuration part ,
| node             | v8.9.1
| npm              | 5.5.1
| Operating System | macOS Sierra v10.12.6

Axnyff:
| Fixed Issues?            | Fixes #7199 

This PR ensures that when inferring a function name from a literal, it will make sure that the literal has a truthy value to prevent having undefined as a function name.

I'm not sure this is the best way to address it, I added my thoughts on the issue.
https://github.com/babel/babel/issues/7199#issuecomment-368229625

### DIFF
--- a/packages/babel-plugin-transform-function-name/test/fixtures/issues/7199/output.js
+++ b/packages/babel-plugin-transform-function-name/test/fixtures/issues/7199/output.js
@@ -1,0 +1,3 @@
+const x = {
+  [`y`]: function () {}
+};
--- a/packages/babel-helper-function-name/src/index.js
+++ b/packages/babel-helper-function-name/src/index.js
@@ -167,7 +167,7 @@ export default function({ node, parent, scope, id }, localBinding = false) {
   }
 
   let name;
-  if (id && t.isLiteral(id)) {
+  if (id && t.isLiteral(id) && id.value) {
     name = id.value;
   } else if (id && t.isIdentifier(id)) {
     name = id.name;
--- a/packages/babel-plugin-transform-function-name/test/fixtures/issues/7199/input.js
+++ b/packages/babel-plugin-transform-function-name/test/fixtures/issues/7199/input.js
@@ -1,0 +1,3 @@
+const x = {
+  [`y`]: function () {}
+};
--- a/packages/babel-plugin-transform-function-name/test/fixtures/issues/7199/options.json
+++ b/packages/babel-plugin-transform-function-name/test/fixtures/issues/7199/options.json
@@ -1,0 +1,3 @@
+{
+  "plugins": ["external-helpers", "transform-function-name"]
+}


packages/babel-plugin-transform-function-name/test/fixtures/issues/7199/output.js
```diff
@@ -0,0 +1,3 @@
+const x = {
+  [`y`]: function () {}
```
Andarist: actually i think we could evaluate simple expressions such as this to know the inferred function name

Axnyff:
I played a little with my build and I noticed a couple things:
* There's also the same bug for a computed property for null which will generate a function name of undefined
* I break better behavior for other falsy values ( 0, false and "")
Maybe I should make sure the value is neither undefined nor null.

Regarding converting simple expressions, I'm not sure it is such an use-case as the use of a computed property is useless in those case. 
What would be the best way to check if it's doable, should I check that the expression array is empty

PR #7429, opened 2018-02-25
Fix default descriptor setting for class properties with decorators

### LINKED ISSUE #5519: babel-plugin-transform-decorators: Property is readonly if no value set
Original issue here https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy/issues/57

It's still not solved on `7.0.0-alpha.1`

### LINKED ISSUE #7391: Class property is non-writable when used with decorator without initializer
Choose one: is this a bug report or feature request? bug report

### Input Code

```js
function Dec(target, key, descriptor) {
    // do nothing
    return descriptor;
}

class Test {
    @Dec
    a;
}

let t = new Test();

console.log(t.a); // prints 'undefined'

console.log(Object.getOwnPropertyDescriptor(t, 'a')); // writable: false (should be true)
/* prints:
{ value: undefined,
  writable: false,
  enumerable: true,
  configurable: false }
*/

t.a = 2;

console.log(t.a); // prints 'undefined' (should be 2)

```

```js
{
  "presets": [
    ["@babel/preset-env", {
      "targets": {
        "node": "current"
      }
    }],
    "@babel/preset-stage-2"
  ],
  "plugins": [
    [
      { "loose": true }
    ]
  ]
}

```

### Expected Behavior

The property `a` should be `writable`.

### Current Behavior

`writable` in `a`'s descriptor is set to false.

### Possible Solution

Currently Babel transpiles the code with `_applyDecoratedDescriptor` function declaration as following:
```js
function _applyDecoratedDescriptor(target, property, decorators, descriptor, context) { var desc = {}; Object['ke' + 'ys'](descriptor).forEach(function (key) { desc[key] = descriptor[key]; }); desc.enumerable = !!desc.enumerable; desc.configurable = !!desc.configurable; if ('value' in desc || desc.initializer) { desc.writable = true; } desc = decorators.slice().reverse().reduce(function (desc, decorator) { return decorator(target, property, desc) || desc; }, desc); if (context && desc.initializer !== void 0) { desc.value = desc.initializer ? desc.initializer.call(context) : void 0; desc.initializer = undefined; } if (desc.initializer === void 0) { Object['define' + 'Property'](target, property, desc); desc = null; } return desc; }
```
which sets `desc.writable` true in the following condition:
`if ('value' in desc || desc.initializer) { desc.writable = true; }`
I don't understand this part. Should this be changed somehow?

### Context

This behavior occurs **only** when both of the following conditions are met:

- A decorator is present on the class property.
- No initializer is defined in the class property declaration.

For example, if there's no decorator:
```js
class Test {
    a;
}
```
`a` is writable.
<details>
<summary>Transpiled code:</summary>

```js
function Dec(target, key, descriptor) {
  // do nothing
  return descriptor;
}

class Test {
  constructor() {
    this.a = void 0;
  }

}

let t = new Test();
console.log(t.a); // prints 'undefined'

console.log(Object.getOwnPropertyDescriptor(t, 'a')); // writable: false (should be true)

t.a = 2;
console.log(t.a); // prints 'undefined' (should be 2)
```

</details>

Also, if I set an initializer:
```js
class Test {
    @Dec
    a = 1;
}
```
`a` is writable.
<details>
<summary>Transpiled code:</summary>

```js
var _class, _descriptor;

function _initializerDefineProperty(target, property, descriptor, context) { if (!descriptor) return; Object.defineProperty(target, property, { enumerable: descriptor.enumerable, configurable: descriptor.configurable, writable: descriptor.writable, value: descriptor.initializer ? descriptor.initializer.call(context) : void 0 }); }

function _applyDecoratedDescriptor(target, property, decorators, descriptor, context) { var desc = {}; Object['ke' + 'ys'](descriptor).forEach(function (key) { desc[key] = descriptor[key]; }); desc.enumerable = !!desc.enumerable; desc.configurable = !!desc.configurable; if ('value' in desc || desc.initializer) { desc.writable = true; } desc = decorators.slice().reverse().reduce(function (desc, decorator) { return decorator(target, property, desc) || desc; }, desc); if (context && desc.initializer !== void 0) { desc.value = desc.initializer ? desc.initializer.call(context) : void 0; desc.initializer = undefined; } if (desc.initializer === void 0) { Object['define' + 'Property'](target, property, desc); desc = null; } return desc; }

function _initializerWarningHelper(descriptor, context) { throw new Error('Decorating class property failed. Please ensure that ' + 'proposal-class-properties is enabled and set to use loose mode. ' + 'To use proposal-class-properties in spec mode with decorators, wait for ' + 'the next major version of decorators in stage 2.'); }

function Dec(target, key, descriptor) {
  // do nothing
  return descriptor;
}

let Test = (_class = class Test {
  constructor() {
    _initializerDefineProperty(this, "a", _descriptor, this);
  }

}, (_descriptor = _applyDecoratedDescriptor(_class.prototype, "a", [Dec], {
  enumerable: true,
  initializer: function () {
    return 1;
  }
})), _class);
let t = new Test();
console.log(t.a); // prints 'undefined'

console.log(Object.getOwnPropertyDescriptor(t, 'a')); // writable: false (should be true)

t.a = 2;
console.log(t.a); // prints 'undefined' (should be 2)
```

</details>

### Your Environment

| ---------------- | -------
| Babel            |  see below
| Babylon          | 
| node             | 
| npm              | 
| Operating System | 

```
    "@babel/cli": "^7.0.0-beta.40",
    "@babel/core": "^7.0.0-beta.40",
    "@babel/plugin-proposal-class-properties": "^7.0.0-beta.40",
    "@babel/plugin-proposal-decorators": "^7.0.0-beta.39",
    "@babel/preset-env": "^7.0.0-beta.39",
    "@babel/preset-stage-2": "^7.0.0-beta.39"
```

yhpark:
| Fixed Issues?            | Fixes #7391, Fixes #5519, Fixes https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy/issues/57, Fixes https://github.com/typeorm/typeorm/issues/1458 and probably more 
| Major: Breaking Change?  |
| Minor: New Feature?      |

As described in #7391, class properties **with decorators** are given a descriptor whose `writable` value is false by default, which created an issue that the value of a class property with decorators is **fixed as undefined value** when it doesn't have an initializer. This patch changes the behavior to be consistent with the class properties without decorators as handled by `proposal-class-properties` plugin, where a class property gets a base descriptor with `configurable: true, enumerable: true, writable: true` (https://github.com/babel/babel/blob/master/packages/babel-plugin-proposal-class-properties/src/index.js#L51-L53).

I also made a small change in `applyDecoratedDescriptor` helper that determines `desc.writable` value, since `desc.writable` is always set to either true or false after the above change is made.

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -953,9 +953,7 @@ helpers.applyDecoratedDescriptor = () => template.program.ast`
         });
         desc.enumerable = !!desc.enumerable;
         desc.configurable = !!desc.configurable;
-        if ('value' in desc || desc.initializer){
-            desc.writable = true;
-        }
+        desc.writable = !!desc.writable;
 
         desc = decorators.slice().reverse().reduce(function(desc, decorator){
             return decorator(target, property, desc) || desc;
--- a/packages/babel-plugin-proposal-decorators/src/transformer-legacy.js
+++ b/packages/babel-plugin-proposal-decorators/src/transformer-legacy.js
@@ -169,7 +169,15 @@ function applyTargetDecorators(path, state, decoratedProps) {
             ),
             t.objectExpression([
               t.objectProperty(
+                t.identifier("configurable"),
+                t.booleanLiteral(true),
+              ),
+              t.objectProperty(
                 t.identifier("enumerable"),
+                t.booleanLiteral(true),
+              ),
+              t.objectProperty(
+                t.identifier("writable"),
                 t.booleanLiteral(true),
               ),
               t.objectProperty(t.identifier("initializer"), initializer),
--- a/packages/babel-plugin-proposal-decorators/test/fixtures/legacy-class-prototype-properties/mutate-descriptor/exec.js
+++ b/packages/babel-plugin-proposal-decorators/test/fixtures/legacy-class-prototype-properties/mutate-descriptor/exec.js
@@ -16,6 +16,16 @@ function dec(target, name, descriptor) {
   });
 }
 
+function plainDec(target, name, descriptor) {
+  expect(target).toBeTruthy();
+  expect(typeof name).toBe("string");
+  expect(typeof descriptor).toBe("object");
+
+  target.decoratedProps = (target.decoratedProps || []).concat([name]);
+
+  return descriptor;
+}
+
 class Example {
   @dec
   enumconfwrite = 1;
@@ -40,6 +50,9 @@ class Example {
 
   @dec
   _ = 8;
+
+  @plainDec
+  plain = 9;
 }
 
 const inst = new Example();
@@ -54,6 +67,7 @@ expect(inst.decoratedProps).toEqual([
   "conf",
   "write",
   "_",
+  "plain",
 ]);
 
 const descs = Object.getOwnPropertyDescriptors(inst);
@@ -97,3 +111,8 @@ expect(descs._.enumerable).toBe(false);
 expect(descs._.writable).toBe(false);
 expect(descs._.configurable).toBe(false);
 expect(inst._).toBe("__8__");
+
+expect(descs.plain.enumerable).toBeTruthy();
+expect(descs.plain.writable).toBeTruthy();
+expect(descs.plain.configurable).toBeTruthy();
+expect(inst.plain).toBe(9);
--- a/packages/babel-plugin-proposal-decorators/test/fixtures/legacy-class-prototype-properties/properties-without-initializer/exec.js
+++ b/packages/babel-plugin-proposal-decorators/test/fixtures/legacy-class-prototype-properties/properties-without-initializer/exec.js
@@ -1,5 +1,11 @@
 function dec(target, name, descriptor) {
+  expect(target).toBeTruthy();
+  expect(typeof name).toBe("string");
+  expect(typeof descriptor).toBe("object");
 
+  target.decoratedProps = (target.decoratedProps || []).concat([name]);
+
+  return descriptor;
 }
 
 class Example {
@@ -7,5 +13,17 @@ class Example {
 }
 
 let inst = new Example();
+
+expect(Example.prototype).toHaveProperty("decoratedProps");
+expect(inst.decoratedProps).toEqual([
+  "prop",
+]);
+
 expect(inst).toHaveProperty("prop");
 expect(inst.prop).toBeUndefined();
+
+const descs = Object.getOwnPropertyDescriptors(inst);
+
+expect(descs.prop.enumerable).toBeTruthy();
+expect(descs.prop.writable).toBeTruthy();
+expect(descs.prop.configurable).toBeTruthy();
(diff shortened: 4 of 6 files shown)


packages/babel-plugin-proposal-decorators/src/index.js
```diff
@@ -172,10 +172,18 @@ export default function() {
                 decorators.map(dec => t.cloneNode(dec.expression)),
               ),
               t.objectExpression([
+                t.objectProperty(
```
Andarist: i'd suspect this change touch some `output.js` test file, but this PR has no such changes, do you know why?

yhpark: Sorry for the late reply. There are only 4 fixture-based tests in this package and one of them is affected by this change. I've already included the change in this commit (`test/fixtures/decl-to-expression/method-decorators/output.js`). Since there's no plugin depending on this package, no tests outside this package would be affected by this change.

yhpark:
Can someone please review this PR? This issue makes class property decorators practically unusable.

jdolle:
I've also run in to this bug and second @yhpark in that it makes class decorators unusable. I'd be willing to help resolve the existing conflicts if someone is interested in reviewing.

marano:
Please merge this <3 As @jdolle and @yhpark have stated, in the current situation the decorators are unusable in many cases.

yhpark:
@Andarist (cc. @nicolo-ribaudo)
I have rebased the patch onto the latest master branch. This is a small fix that would affect many people who are trying to use class property decorators. I really hope someone would review and merge this PR. Thanks

eduard-malakhov:
Hi guys, does anyone know whether there is a way to use property decorators today? As far as I can see, the change proposed in this issue hasn't been integrated into the master. Is a new implementation of decorators available?

marano:
@eduard-malakhov You need to assign an initial value to the property like this:
```
@decorator()
property = undefined
```

eduard-malakhov:
@marano I've tried assigning a default value, but this still results in a property that is `configurable: false` which I guess is the reason why the `lazyInject` decorator from inversify-inject-decorators which I apply fails to redefine the property. However, I am not sure about the root cause, because it is hard to tell at which stage this happens - typescript, babel, or the JS interpreter.

nicolo-ribaudo:
Merged. Sorry if it took so long; the fix will be released in Babel 7.

eduard-malakhov:
@nicolo-ribaudo Thank you! I am afraid this patch alone is not enough to make decorators work properly in all cases. I've been playing with this [sample project](https://github.com/eduard-malakhov/inversify-inject-decorators-test) which makes use of inversify-inject-decorators and to make it function properly besides the changes in property descriptor I needed to explicitly `delete` the property on the object in order to get access to the decorated property from the prototype. You can see this in the repo.

When the object is created, it already contains an own property and therefore the property created by the decorator on the prototype is not accessible unless one deletes the object's property. I am not sure, where this property comes from and whether this is a new behavior because if it has always been this way then it is not clear how inversify-inject-decorators and similar decorators could work in principle.

nicolo-ribaudo:
The problem is that Babel doesn't allow changing the placement of a field (the decorators you use changes it from being defined on the instance to being defined on the prototype). If anyone wants to open a PR (legacy spec: https://github.com/wycats/javascript-decorators) I will be happy to review it.

PR #7404, opened 2018-02-20
TypeScript: Support conditional types syntax

andy-ms:
| Q                        | A
| ------------------------ | ---
| Fixed Issues?            | #7389
| Patch: Bug Fix?          | No
| Major: Breaking Change?  | No
| Minor: New Feature?      | Yes
| Tests Added + Pass?      | Yes
| Documentation PR         |
| Any Dependency Changes?  | No
| License                  | MIT

@JamesHenry This changes the AST format. CC @DanielRosenwasser for review.
Supports the syntax from Microsoft/TypeScript#21316 and Microsoft/TypeScript#21496. Now you can write `type Element<T> = T extends (infer U)[] ? U : T;`.

PR #7431, opened 2018-02-26
Upgrade flow to 0.66 and fix a few minor errors.

loganfsmyth:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Looks like a few minor errors on the new version, and one regression in checking from #7404.

### DIFF
--- a/packages/babylon/src/types.js
+++ b/packages/babylon/src/types.js
@@ -1068,6 +1068,8 @@ export type TsType =
   | TsArrayType
   | TsTupleType
   | TsUnionOrIntersectionType
+  | TsConditionalType
+  | TsInferType
   | TsParenthesizedType
   | TsTypeOperator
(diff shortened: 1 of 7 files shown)


packages/babylon/src/types.js
```diff
@@ -1068,6 +1068,8 @@ export type TsType =
   | TsArrayType
   | TsTupleType
   | TsUnionOrIntersectionType
+  | TsConditionalType
+  | TsInferType
```
loganfsmyth: cc @andy-ms I assume this was just missed in your PR?

andy-ms: @loganfsmyth Yep, thanks

PR #7433, opened 2018-02-26
eslint: Rename to .json, mark as root config

mc10:
This renames the file to have a .json suffix, as [.eslintrc is deprecated](https://eslint.org/docs/user-guide/configuring#configuration-file-formats), and [marks the file as the root config](https://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy) to prevent ESLint from looking for config files in parent directories.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,4 +1,5 @@
 {
+  "root": true,
   "extends": "babel",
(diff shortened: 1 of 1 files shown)


loganfsmyth:
Thanks!

PR #7435, opened 2018-02-26
Fix function name computation for literal values

### LINKED ISSUE #7199: undefined is compiled as function name by using template literal as method name.

Axnyff:
| Fixed Issues?            | Fixes #7199 
| Documentation PR         | 
| Any Dependency Changes?  |

I made a first PR [here](https://github.com/babel/babel/pull/7428) but I realized it broke some existing functionalities and I also discovered some other bugs in the way.
Currently,
``    const x = {
        [null]: function(){},
        [/regex/]: function(){},
        [`test`]: function(){}  
};``  

will all create a function with a name of undefined.

My PR correctly creates for null: _null, /regex/: _regex_ and for simple template literals ( which don't have any expressions).
It will not generate a name for more complicated template literals.

### DIFF
--- a/packages/babel-helper-function-name/src/index.js
+++ b/packages/babel-helper-function-name/src/index.js
@@ -47,13 +47,15 @@ const visitor = {
 
 function getNameFromLiteralId(id) {
   if (t.isNullLiteral(id)) {
-    return null;
+    return "null";
   } else if (t.isRegExpLiteral(id)) {
     return `_${id.pattern}_${id.flags}`;
-  } else if (t.isTemplateLiteral(id) && id.expressions.length === 0) {
-    return id.quasis[0].value.raw;
+  } else if (t.isTemplateLiteral(id)) {
+    return id.quasis.map(quasi => quasi.value.raw).join("");
+  } else if (id.value !== undefined) {
+    return id.value + "";
   } else {
-    return id.value;
+    return "";
   }
 }
 
--- a/packages/babel-helper-function-name/src/index.js
+++ b/packages/babel-helper-function-name/src/index.js
@@ -45,6 +45,18 @@ const visitor = {
   },
 };
 
+function getNameFromLiteralId(id) {
+  if (t.isNullLiteral(id)) {
+    return null;
+  } else if (t.isRegExpLiteral(id)) {
+    return `_${id.pattern}_${id.flags}`;
+  } else if (t.isTemplateLiteral(id) && id.expressions.length === 0) {
+    return id.quasis[0].value.raw;
+  } else {
+    return id.value;
+  }
+}
+
 function wrap(state, method, id, scope) {
   if (state.selfReference) {
     if (scope.hasBinding(id.name) && !scope.hasGlobal(id.name)) {
@@ -168,10 +180,12 @@ export default function({ node, parent, scope, id }, localBinding = false) {
 
   let name;
   if (id && t.isLiteral(id)) {
-    name = id.value;
+    name = getNameFromLiteralId(id);
   } else if (id && t.isIdentifier(id)) {
     name = id.name;
-  } else {
+  }
+
+  if (name === undefined) {
     return;
   }
 
--- a/packages/babel-helper-function-name/src/index.js
+++ b/packages/babel-helper-function-name/src/index.js
@@ -48,15 +48,21 @@ const visitor = {
 function getNameFromLiteralId(id) {
   if (t.isNullLiteral(id)) {
     return "null";
-  } else if (t.isRegExpLiteral(id)) {
-    return `_${id.pattern}_${id.flags}`;
-  } else if (t.isTemplateLiteral(id)) {
-    return id.quasis.map(quasi => quasi.value.raw).join("");
-  } else if (id.value !== undefined) {
-    return id.value + "";
-  } else {
-    return "";
   }
+
+  if (t.isRegExpLiteral(id)) {
+    return `_${id.pattern}_${id.flags}`;
+  }
+
+  if (t.isTemplateLiteral(id)) {
+    return id.quasis.map(quasi => quasi.value.raw).join("");
+  }
+
+  if (id.value !== undefined) {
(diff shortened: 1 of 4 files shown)


packages/babel-helper-function-name/src/index.js
```diff
@@ -45,6 +45,18 @@ const visitor = {
   },
 };
 
+function getNameFromLiteralId(id) {
+  if (t.isNullLiteral(id)) {
+    return null;
+  } else if (t.isRegExpLiteral(id)) {
+    return `_${id.pattern}_${id.flags}`;
+  } else if (t.isTemplateLiteral(id) && id.expressions.length === 0) {
```
loganfsmyth: It seems like it would be best to concat the quasis all together for this value.

packages/babel-helper-function-name/src/index.js
```diff
@@ -45,6 +45,18 @@ const visitor = {
...
+    return id.quasis[0].value.raw;
+  } else {
+    return id.value;
```
loganfsmyth: Given the name of this function, I think we should make sure that this is converted to a string before returning it.

packages/babel-helper-function-name/src/index.js
```diff
@@ -45,6 +45,18 @@ const visitor = {
   },
 };
 
+function getNameFromLiteralId(id) {
+  if (t.isNullLiteral(id)) {
+    return null;
```
loganfsmyth: Same here, probably want `"null"`.

packages/babel-plugin-transform-function-name/test/fixtures/issues/7199/output.js
```diff
@@ -2,7 +2,7 @@ const x = {
   [null]: function _null() {},
   [/regex/gi]: function _regex_gi() {},
   [`y`]: function y() {},
-  [`${y}`]: function () {},
+  [`${y}`]: function _() {},
```
nicolo-ribaudo: Can you use something like `abc ${y} de` in this test?

packages/babel-plugin-transform-function-name/test/fixtures/issues/7199/input.js
```diff
@@ -0,0 +1,8 @@
+const x = {
+  [null]: function () {},
+  [/regex/gi]: function () {},
+  [`y`]: function () {},
+  [`abc${y}def`]: function () {},
```
danez: This could produce duplicate names if there is another template string with for example `abc${other}def`, but not sure if this is critical or will produce problems.

nicolo-ribaudo: Since you can't reference to those functions using their name anyway (the name is only used for debuging), it shouldn't create problems.

packages/babel-helper-function-name/src/index.js
```diff
@@ -45,6 +45,20 @@ const visitor = {
   },
 };
 
+function getNameFromLiteralId(id) {
+  if (t.isNullLiteral(id)) {
+    return "null";
+  } else if (t.isRegExpLiteral(id)) {
```
jghinestrosa: Since you are using the `return` statement, the `else if` statement becomes unnecessary. You could use `if` instead.

Axnyff: That's true, i'll update

Axnyff:
Awesome, thanks everybody for your help!

PR #7436, opened 2018-02-27
Default to `ast:false` and do less work when loading core

loganfsmyth:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Y, `ast` now defaults to `false`
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-core/src/config/index.js
+++ b/packages/babel-core/src/config/index.js
@@ -3,7 +3,6 @@
 import path from "path";
 import * as context from "../index";
 import Plugin from "./plugin";
-import merge from "lodash/merge";
 import {
   buildRootChain,
   buildPresetChain,
@@ -13,10 +12,9 @@ import {
 } from "./config-chain";
 import type { UnloadedDescriptor } from "./config-descriptors";
 import traverse from "@babel/traverse";
-import clone from "lodash/clone";
 import { makeWeakCache, type CacheConfigurator } from "./caching";
 import { getEnv } from "./helpers/environment";
-import { validate } from "./validation/options";
+import { validate, type ValidatedOptions } from "./validation/options";
 import { validatePluginObject } from "./validation/plugins";
 
 type LoadedDescriptor = {
@@ -141,7 +139,8 @@ export default function loadConfig(inputOpts: mixed): ResolvedConfig | null {
     throw e;
   }
 
-  const opts: Object = merge(optionDefaults, options);
+  const opts: Object = optionDefaults;
+  merge(opts, options);
 
   // Tack the passes onto the object itself so that, if this object is passed back to Babel a second time,
   // it will be in the right structure to not change behavior.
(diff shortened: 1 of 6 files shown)


packages/babel-core/src/config/index.js
```diff
@@ -3,7 +3,6 @@
 import path from "path";
 import * as context from "../index";
 import Plugin from "./plugin";
-import merge from "lodash/merge";
```
loganfsmyth: Once we make dependencies lazy-load, not loading lodash during anywhere in `src/config` makes a big difference.

packages/babel-core/src/transformation/index.js
```diff
@@ -64,7 +64,7 @@ export function runSync(
   return {
     metadata: file.metadata,
     options: opts,
-    ast: opts.ast !== false ? file.ast : null,
```
loganfsmyth: Most callers of Babel don't use this and it is a huge structure. Not including it in the output means when we land caching logic in core one day, we'll have less data to cache by default.

packages/babel-helpers/src/helpers.js
```diff
@@ -23,9 +19,9 @@ helpers.typeof = defineHelper(`
 
     return _typeof(obj);
   }
-`);
+`;
 
-helpers.jsx = defineHelper(`
+helpers.jsx = () => template.program.ast`
```
loganfsmyth: This just avoids parsing the helper snippets until the first time they are used.

packages/babel-core/src/config/index.js
```diff
@@ -161,6 +160,30 @@ export default function loadConfig(inputOpts: mixed): ResolvedConfig | null {
...
+      const targetObj = (target.generatorOpts = target.generatorOpts || {});
+      mergeObject(targetObj, generatorOpts);
+    } else {
+      const val = source[k];
+      if (val !== undefined) target[k] = (val: any);
+    }
+  }
+}
+
+function mergeObject<T: {}>(target: T, source: T) {
```
Andarist: maybe we could figure out some better name for this one? i had to look into the implementation to see how it is different from `Object.assign`, maybe just smth like `safeMerge`?

packages/babel-helpers/src/helpers.js
```diff
@@ -5,11 +5,7 @@ import template from "@babel/template";
 const helpers = {};
 export default helpers;
 
-// Helpers never include placeholders, so we disable placeholder pattern
-// matching to allow us to use pattern-like variable names.
-const defineHelper = template.program({ placeholderPattern: false });
-
-helpers.typeof = defineHelper(`
+helpers.typeof = () => template.program.ast`
```
Andarist: was removing `placeholderPattern` for those helper templates deliberate?

loganfsmyth: Yeah, the `.ast` variant already returns the AST unchanged, so there aren't any placeholders to bother searching for.

packages/babel-core/src/config/index.js
loganfsmyth: How about `mergeDefaults` since that seems to be the common name for this operation?

Andarist: Sounds good to me 😃

packages/babel-helpers/src/helpers.js
Andarist: Cool, good to know. Thanks for clarifying

packages/babel-core/src/transformation/index.js
Sunbridger: o

PR #7438, opened 2018-02-27
Ensure babel-preset-env targets input object is not mutated

guybedford:
Currently the `opts.target` object is being passed through and mutated. This caught me up in a workflow where I was passing `{ esmodules: true }` which was being mutated into `{ esmodules: true, browsers: ... browser list... }` and then giving a warning for future transforms using the same input object that both browsers and esmodules were being provided.

| Fixed Issues?            | N-->
| Tests Added + Pass?      | N


existentialism:
@guybedford thanks!

PR #7439, opened 2018-02-27
feat(babel-cli): add a brief summary to build output

### LINKED ISSUE #6303: In babel-cli add a short log option when compiling
Choose one: is this a bug report or feature request? feature request

### Input Code

N/A

```
{
  "babel-cli": "^6.24.1"
}
```

`node_modules/.bin/babel ./app/src -w -p -d ./app/scripts --modules amd --extensions '.es6,.jsx' --source-maps`

### Expected Behavior

When compiling a project with a large number of files there can be an excessive amount of logs produced.

The `--quiet` option removes all logs, however something in between may be of use. Perhaps a `x of x files compiled` or a percentage based option would be useful?

screen shot 2017-09-25 at 16 31 34

### Current Behavior

For large projects either a large amount of logs are shown, or none at all with the `--quiet` option.

### Possible Solution

Add another method of logging the progress of the compiler to show percentage complete or `x of x compiled` message.

### Context

Build step of my project, I can choose between 500+ lines of output or 0.

### Your Environment

N/A

thymikee:
| Fixed Issues?            | Fixes #6303

Adds a brief summary when running `babel-cli`. It displays even when `--quiet` flag is passed.
screen shot 2018-02-27 at 12 38 49

cc @hzoo @Andarist


packages/babel-cli/src/babel/dir.js
```diff
@@ -31,7 +36,10 @@ export default function(commander, filenames, opts) {
         opts,
       ),
       function(err, res) {
-        if (err) return callback(err);
+        if (err) {
+          result.failure += 1;
```
thymikee: For now I gather failures count, but don't use them anywhere. Happy to remove this.

packages/babel-cli/src/babel/dir.js
```diff
@@ -125,10 +135,17 @@ export default function(commander, filenames, opts) {
     const filename = filenames[index];
 
     handle(filename, function(err) {
-      if (err) throw err;
+      if (err) throw err.toString();
```
thymikee: Since the stack trace of build errors isn't very helpful (lots of internal Babel stuff not necessarily helpful to the end user), I've decided to use only the stringified message. It works well on Node >=8. Node <=6 produces a totally unhelpful message anyway 🤷‍♂️

packages/babel-cli/test/fixtures/babel/--copy-files --include-dotfiles with ignore/stdout.txt
```diff
@@ -1 +1,2 @@
 src/index.js -> lib/index.js
+Successfully compiled 1 file.
```
thymikee: I thought about putting this into `stderr` instead of `stdout` (I wouldn't need to adjust all tests, just add a new one), but based on experience in Jest CLI, I'd rather avoid that, as users tend to abuse this stream for catching their own errors or such.

packages/babel-cli/src/babel/dir.js
Andarist: what happens when error occurs here? i suspect it crash whole process, right? in other words - can `failure` increase beyond 1 here?

packages/babel-cli/src/babel/dir.js
Andarist: Is it a common practice in other CLI tools? is this error here crashing the process directly & going right into stderr?

packages/babel-cli/src/babel/dir.js
thymikee: It forwards the error object and crashes [here](https://github.com/babel/babel/pull/7439/files/ed6291a6ee05d73fb4577fbe989be9e103461e1f#diff-f93fbf6dce06b5cbf6cc65f5f7dac101L128) with non-zero exit code. I don't use `failures` anywhere now (I used to show number of failed files but decided it's not necessary, as usually babel just crash the whole thing)

packages/babel-cli/src/babel/util.js
```diff
@@ -46,8 +46,8 @@ export function addSourceMappingUrl(code, loc) {
   return code + "\n//# sourceMappingURL=" + path.basename(loc);
 }
 
-export function log(msg) {
-  if (!commander.quiet) console.log(msg);
+export function log(msg, alwaysLog) {
```
Andarist: nit picking: i think `always` would suffice here

packages/babel-cli/src/babel/dir.js
thymikee: Yup, crashes entire thing. Throwing `err.toString()` is definitely not something common, usually you end up creating your own error extending from `Error` and override stacktrace if necessary, for better error messages.

packages/babel-cli/test/fixtures/babel/--copy-files --include-dotfiles with ignore/stdout.txt
Andarist: stdout is imho a perfect place for this, stderr implies some kind of error, warning at least, and this is a regular info message

packages/babel-cli/src/babel/dir.js
Andarist: we could remove counting `failure` then and convert `result` to a "flat" counter of successes, would give future readers a clear purpose for this, with unused failures it might be a little bit unclear

packages/babel-cli/src/babel/dir.js
Andarist: so if this is direct exit point of the program & to avoid throwing non error it could be converted to smth like:
```js
console.error(err.toString())
process.exit(1)
```

although im not entirely convinced that this is a good change, logged value is still logged currently, right? Only with less-readable for most users stack trace. Personally I often take such stack traces and go into my node_modules to get better understanding of what happened if the error message is not descriptive enough (which is most of the times 😉)

thymikee: I just noticed we throw raw error without wrapping it with an Error object. I've pushed a change to adjust that

Here's how it looks now:
screen shot 2018-02-27 at 23 41 15

imho better than:
screen shot 2018-02-27 at 23 42 23

Andarist: Definitely better!

packages/babel-cli/src/babel/dir.js
```diff
@@ -125,10 +129,17 @@ export default function(commander, filenames, opts) {
     const filename = filenames[index];
 
     handle(filename, function(err) {
-      if (err) throw err;
+      if (err) throw Error(err);
```
Andarist: this probably should be a `new Error`

thymikee: they're equivalent https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

Andarist: While they might be equivalent, it's most commonly used as a constructor - I'd prefer to `new` it for consistency.

existentialism:
Just spitballing, but any thoughts on prefixing the message (and maybe all other logs/errors from Babel)?

```sh
Babel: 🎉 Successfully compiled 1,492 files.
```

Might help with searching/parsing logs, or helping users understand where errors are coming from? Just thinking aloud :)

thymikee:
I was thinking of adding an emoji too! But it looked weird with Yarn output (unprefixed), because they use 2 spaces:
```
🎉 Successfully compiled 1,492 files.
✨  Done in 0.72s.
```
Not sure about prefixing though, it's not something I've seen in the wild tbh.

How about this (also aligning with Yarn):
```
🎉  Successfully compiled 1,492 files with Babel.
✨  Done in 0.72s.
```

packages/babel-preset-env/test/debug-fixtures/usage/stdout.txt
```diff
@@ -43,4 +43,6 @@ src/in.js -> lib/in.js
 [src/in2.js] Added following polyfills:
   regenerator-runtime { "chrome":"52", "firefox":"50", "ie":"11" }
   web.dom.iterable { "chrome":"52", "firefox":"50", "ie":"11" }
-src/in2.js -> lib/in2.js
\ No newline at end of file
+src/in2.js -> lib/in2.js
+🎉  successfully compil
```
Andarist: this seems off, the line got splitted into 2, some other `stdout.txt` suffer from the same thing, any idea why this happens?

packages/babel-cli/src/babel/index.js
```diff
@@ -243,7 +243,7 @@ delete opts.outFile;
 delete opts.outDir;
 delete opts.copyFiles;
 delete opts.includeDotfiles;
-delete opts.quiet;
+delete opts.verbose;
```
Andarist: was this change requested somewhere? whats the motivation behind it?

thymikee: discussed here: https://github.com/babel/babel/issues/6303#issuecomment-369064626

packages/babel-preset-env/test/debug-fixtures/usage/stdout.txt
thymikee: huh, that's just my regex not playing the way I liked. Weird that test didn't caught it (at least locally)?

thymikee: Oh, I was scoping my testing to `babel-cli` only, I'll update this shortly

packages/babel-cli/src/babel/index.js
Andarist: 👍

thymikee:
`npm`:
screen shot 2018-03-01 at 22 45 24

`yarn`:
screen shot 2018-03-01 at 22 45 36

Andarist:
The only thing missing (I think) is like at least a single test covering default mode, right? (i dont see any test converted from previously quiet to now default)

packages/babel-cli/test/fixtures/babel/dir --out-dir/stdout.txt
```diff
@@ -1,2 +1 @@
-src/bar/bar.js -> lib/bar/bar.js
-src/foo.js -> lib/foo.js
+🎉  Successfully compiled 2 files with Babel.
```
thymikee: @Andarist, here is the default without `--verbose` flag

packages/babel-preset-env/test/debug-fixtures/android/stdout.txt
```diff
@@ -133,4 +133,4 @@ Using polyfills with `entry` option:
   web.timers { "android":"4" }
   web.immediate { "android":"4" }
   web.dom.iterable { "android":"4" }
-src/in.js -> lib/in.js
\ No newline at end of file
+🎉  Successfully compiled 1 file with Babel.
```
thymikee: Also, all tests in `babel-preset-env` cover default quiet mode.

packages/babel-cli/test/fixtures/babel/dir --out-dir/stdout.txt
Andarist: 👍 disregard my comment then

packages/babel-preset-env/test/debug-fixtures/android/stdout.txt
existentialism: 👍

thymikee:
Travis failure _seems_ unrelated.

existentialism:
@thymikee it almost certainly is, those particular tests have been finicky lately, haven't had a chance to look into it much. restarted it tho!

packages/babel-cli/src/babel/index.js
```diff
@@ -163,7 +163,7 @@ commander.option(
   "--include-dotfiles",
   "Include dotfiles when compiling and copying non-compilable files",
 );
-commander.option("-q, --quiet", "Don't log anything");
+commander.option("-v, --verbose", "Log everything");
```
thymikee: Also, what do you think about the alias – should it be `-v` or maybe no alias at all? At least it doesn't collide with anything right now.

Andarist: Oh, havent spotted that yesterday - I think `verbose` is not like the most wanted/common thing and it might not deserve alias just yet, especially that `-v` is most commonly used for `version` in other tools.

thymikee: removed it

xtuc:
Sorry I missed this PR and it's a while ago. I noticed that the flag `--quiet` has been renamed, people that relied on that flag will now get an error? It's a breaking change.

Also we'd like to remove the emojis https://github.com/babel/babel/pull/8250.

thymikee:
>I noticed that the flag --quiet has been renamed, people that relied on that flag will now get an error? It's a breaking change.

Not sure about how CLI handles unknown flags passed to it. If it errors, then it's a breaking change – to unbreak it, we could add the `--quiet` flag back which would do nothing at the moment, and be there for future updates.

Andarist:
As removal of `--quiet` was merged only to beta line, I would assume it's OK for it to be a breaking change. If we do not need it right now then it doesn't seem necessary to add it only to avoid a breaking change in this situation. We might want to document this change though as breaking one.

xtuc:
I tried using `7.0.0-beta.51 (@babel/core 7.0.0-beta.51)`, passing an unknown flag returns an error and an exit code 1. It's a breaking change, even if it's since Babel 7 we should communicate accordingly.

IMO we should rather not introduce unnecessary breaking changes.

PR #7442, opened 2018-02-27
Adds function injectExternalHelpers to inject babelHelpers in global

### LINKED ISSUE #7357: injecting external-helpers in a node app
Choose one: is this a bug report or feature request? (docs?) bug report

I'm trying to use a package that assumes that the external-helpers are available as a global. From the [docs](https://babeljs.io/docs/plugins/external-helpers/#injecting-the-external-helpers), I should be able to inject them to `global` in my node app by using `require("babel-core").buildExternalHelpers();`. However, use of that still results in the following error: `ReferenceError: babelHelpers is not defined`

Since the `buildExternalHelpers()` function needs to run before the package is imported and my app uses es module imports, I'm using a bootstrap file as an entry point that is ignored from transpilation and just tries to inject the helpers before loading the actual app:

```js
require("babel-core").buildExternalHelpers();
const app = require('./app');
```

### Expected Behavior

`babelHelpers` should be added to `global` so that it is available for the package that assumes it is available there.

from the docs:
> This injects the external helpers into `global`.

### Current Behavior

`babelHelpers` is not made available on `global`, resulting in `ReferenceError: babelHelpers is not defined`

### Possible Solution

The docs also [mention](https://babeljs.io/docs/plugins/external-helpers/#getting-the-external-helpers) generating a helpers file with `./node_modules/.bin/babel-external-helpers [options] > helpers.js`. It wasn't obvious to me that this file could be imported to accomplish the same goal as `buildExternalHelpers()` until I started reading the source of that file. Importing that file instead does work for my app. I'll need this file elsewhere, so I'll likely just continue importing that instead, even if there is a way to use `buildExternalHelpers()`.

With that approach, my bootstrap file has the following contents instead:

```js
require('../../vendor/babel-helpers');
const app = require('./app');
```

### Your Environment

| ---------------- | -------
| Babel            | 6.26.0
| node             | 8.9.4
| npm              | 5.6.0
| Operating System | macOS High Sierra 

### Forum

While I was still trying to find a working solution, I was trying to find the best place to ask questions. The website still links to a discourse forum that no longer seems to exist. It'd be a good idea to either remove the link from the site or at least have it link to somewhere that describes the currently recommended approach for getting that type of help.

debugpai2:
global context
Solves #7357

| Fixed Issues?            | `Fixes #7357` 

Adds a new function `injectExternalHelpers ` in `build-external-helpers` which automatically inject `buildHelpers` global context when called.

### DIFF
--- a/packages/babel-core/test/api.js
+++ b/packages/babel-core/test/api.js
@@ -574,11 +574,6 @@ describe("api", function() {
       it("builds external helpers in var output type", function() {
         babel.buildExternalHelpers(null, "var");
       });
-
-      it("injects external helpers in global context", function() {
-        babel.injectExternalHelpers();
-        assert.ok(global.babelHelpers);
-      });
     });
 
     it("all", function() {
--- a/packages/babel-core/test/api.js
+++ b/packages/babel-core/test/api.js
@@ -574,6 +574,11 @@ describe("api", function() {
       it("builds external helpers in var output type", function() {
         babel.buildExternalHelpers(null, "var");
       });
+
+      it("injects external helpers in global context", function() {
+        babel.injectExternalHelpers();
+        assert.ok(global.babelHelpers);
+      });
     });
 
     it("all", function() {
--- a/packages/babel-core/src/index.js
+++ b/packages/babel-core/src/index.js
@@ -3,6 +3,7 @@
 export { default as File } from "./transformation/file/file";
 export {
   default as buildExternalHelpers,
+  injectExternalHelpers,
 } from "./tools/build-external-helpers";
 export { resolvePlugin, resolvePreset } from "./config/files";
 
--- a/packages/babel-core/src/tools/build-external-helpers.js
+++ b/packages/babel-core/src/tools/build-external-helpers.js
@@ -139,7 +139,10 @@ function buildHelpers(body, namespace, whitelist) {
   });
   return refs;
 }
-export default function(
+export function injectExternalHelpers() {
+  eval(buildExternalHelpers());
+}
+export default function buildExternalHelpers(
   whitelist?: Array<string>,
   outputType: "global" | "module" | "umd" | "var" = "global",
 ) {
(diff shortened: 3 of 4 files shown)


packages/babel-core/test/api.js
```diff
@@ -574,6 +574,11 @@ describe("api", function() {
       it("builds external helpers in var output type", function() {
         babel.buildExternalHelpers(null, "var");
       });
+
+      it("injects external helpers in global context", function() {
+        babel.injectExternalHelpers();
+        assert.ok(global.babelHelpers);
```
nicolo-ribaudo: We should run this test in a new vm context, to avoid polluting the global namespace.

debugpai2: can you guide me on how on how I should go about doing that?

debugpai2: Moved the test to a different file. Now I think it should be using a new vm

nicolo-ribaudo: I don't know exactly how our test runner works, but I think that `global` is the same for every file. I think we should execute this test in a new `global` context, using the [vm](https://nodejs.org/api/vm.html) module.

debugpai2: I asked in the jest discord channel. They confirmed that they use a new vm for each file. 
Also I tried using the vm module. Its not possible to test it that way because we can't import babel inside the sandbox so we have to pass babel from outside the sandbox. If we do it this way the babel that runs inside the sandbox will create the global outside. I tried this code
```js
const sandbox = { babel }; // this is babel imported at top of the file
const code = `
  babel.injectExternalHelpers();
`;
vm.createContext(sandbox);
vm.runInNewContext(code, sandbox);
assert.ok(sandbox.babelHelpers); //false
assert.ok(babelHelpers); // true
```

nicolo-ribaudo: 👍

nicolo-ribaudo: 👍

debugpai2:
Closing this PR as @loganfsmyth clarified in the original issue why adding `babelExternalHelpers` in the global context is a bad idea.

PR #7443, opened 2018-02-27
Update README for Babel TypeScript Plugin

orta:
Provides additional docs on what the trade-offs are for moving to Babel 7, and a few work-arounds.

The one that I'm not 100% on is the `const enum`s. Would love some help on what the recommendation for replacing that could look like?

/cc @andy-ms @existentialism @DanielRosenwasser for ideas on that

### DIFF
--- a/packages/babel-plugin-transform-typescript/README.md
+++ b/packages/babel-plugin-transform-typescript/README.md
@@ -4,8 +4,17 @@
 
 Does not type-check its input. For that, you will need to install and set up TypeScript.
 
-Does not support `namespace`s or `const enum`s because those require type information to transpile.
-Also does not support `export =` and `import =`, because those cannot be transpiled to ES.next.
+## Caveats
+
+* Does not support [`namespace`][namespace]s.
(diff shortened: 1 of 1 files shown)


packages/babel-plugin-transform-typescript/README.md
```diff
@@ -4,8 +4,17 @@
...
+## Caveats
+
+* Does not support [`namespace`][namespace]s.
+* Does not support [`const enum`][const_enum]s because those require type information to transpile.
+* Does not support [`export =`][exin] and [`import =`][exin], because those cannot be transpiled to ES.next.
+
+## Workarounds
+
+* `namespace`: Migrate to defining modules instead.
+* `const enum`: Avoid to using enums outside of current file
```
andy-ms: You can just change to `enum` (without `const`), which is available as a value at runtime and can be transpiled without type information.

orta: Thanks @andy-ms - updated 👍

existentialism:
@orta thanks!

PR #7444, opened 2018-02-27
Refactor inheritance in babel-plugin-transform-classes

### LINKED ISSUE #7202: Refactor transform-classes
At the moment transform-classes has 2 modes - regular one and loose. The loose mode is extending VanillaTransformer **class**. It seems to me that this hierarchical structure here is clunky and harder to maintain.

Don't know how others feel, but imho it would be nice to refactor this into a flat structure and reuse code between 2 transforms if possible, but more in a composition manner than with inheritance.

I'd like to have at least 2-3 upvotes from other core team members for this, before anyone starts to work on this.

thymikee:
| Fixed Issues?            | Fixes #7202
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Implementing `loose` type of classes transform as a class extending from vanilla transform seems a bit indirect, cause confusion and artificial constructs like [here](https://github.com/babel/babel/pull/7038/files#r160910562). To simplify things I've decided to merge them into one class and expose as a function to not promote further inheritance.

By the way I've added `// @flow` pragma to the files and adjusted some typings (but not all of them, I'm not sure how to type Babel so help appreciated)

@Andarist let me know if that's what you were thinking about, or maybe I got it wrong.
cc @xtuc


Andarist:
Im wondering if we could just iterate through given class to collect "meta" object describing the class, smth like:
```
{ methods, constructor, super, }
```
(not that i havent analyzed the source to see what properties we would have to gather).

And use that collected meta (mostly consisted of PathNodes) to construct conditionally (based on the meta & loose) the final es5 class. Would that make sense to you? Would it make it more readable in your eyes?

I kinda feel that having a class here (in its current form at least) is kinda clunky and the same output could be achieved by "just functions".

xtuc:
> I kinda feel that having a class here (in its current form at least) is kinda clunky and the same output could be achieved by "just functions".

I agree with that.

Looking at the code and your change it feels like we have removed the inheritance in favor of conditions, which is probably inevitable but would be clearer with just functions.

thymikee:
Yep, I'm planning to move this to functions 👍. 
Some conditions were already there, so the separation was artificial (and undone). Let's treat it as a in-progress state, I'll try to update this today.

thymikee:
My battery is running low, so I'll just commit some wip stuff. I've moved class to a function with a simple object as a state. I also found pretty flexible to update the state with a function `setState()` (can be refactored to other data structure, or debugged easily).
Let me know what do you think of this direction.

thymikee:
I think this is ready to review. Some things got simplified throughout the process, but not much. I recommend reviewing with `?w=1` query param :)

thymikee:
FYI, just rebased to master

xtuc:
Thanks, I'll review your work shortly.

packages/babel-plugin-transform-classes/src/vanilla.js
```diff
@@ -200,22 +144,20 @@ export default class ClassTransformer {
   }
 
   /**
-   * [Please add a description.]
-   * https://www.youtube.com/watch?v=fWNaR-rxAic
+   * Creates a class constructor if there is none
    */
-
-  constructorMeMaybe() {
+  function createConstructor() {
```
Andarist: nitpick - from what I observed `maybe*` is a convention used in babel's codebase to indicate that a function attempts to do something, but may bail out - imho it might be good to preserve it here as this function won't always "create constructor" 😉

thymikee: Renamed it to `maybeCreateConstructor`. See last 2 commits.

thymikee:
Thanks @Andarist & @xtuc! :)

xtuc:
Thanks for your contribution!

packages/babel-plugin-transform-classes/src/transformClass.js
```diff
@@ -0,0 +1,718 @@
+// @flow
```
loganfsmyth: It looks like this file was given Flow annotations, but fails a few checks. Would someone be up for taking a look? Looks like just a couple failures.

thymikee: Oh, sorry about that! I wasn't sure how to type it in a usable way (without blindly patching) 😬

loganfsmyth: No problem. I'd say we should just remove the `@flow` annotation if it isn't actually intended to typecheck them. The flow type definitions are still useful for documentation, so those are fine though.

PR #7450, opened 2018-02-28
Allow plugins to assert that a specific babel version has loaded the plugin.

### LINKED ISSUE #7265: Ensure that plugins throw helpful errors if used on versions of Babel that they do not support
Feedback on various options would be appreciated!

Issues like https://github.com/babel/babel/issues/7110 are tripping up users pretty often, so making the transition from 6 to 7 smoother for users should be considered a blocker for the release of 7.x.

I think there are a bunch of ways to approach this. The simplest being that we literally go through each plugin and change

```
export default function(api, options){
  const { loose } = options; // or whatever options the plugin has
```
to

```
export default function(api, options){
  if (!/^7\./.test(api.version))) {
    throw new Error(
      `This plugin requires Babel 7 and appears to be running on version ${api.version}`,
    );
  }
  const { loose } = options; // or whatever options the plugin has
```

Obviously the downside here would be having to copy-paste this between modules. It also means we'll still throw when `options` is `undefined` on early versions of the 7.x beta, though that may be acceptable.

Alternatively a more complex solution might be more like

```
import declare from "babel-helper-plugin-utils";

  const { loose } = options; // or whatever options the plugin has
```
with `declare` wrapping the logic encapsulated above like as
```
function declare(handler) {
  return (api, options, dirname) => {
    if (!/^7\./.test(api.version))) {
      throw new Error(
        `This plugin requires Babel 7 and appears to be running on version ${api.version}`,
      );
    }
    return handler(api, options, dirname);
  };
}
```

The downsides here being that if we publicize this pattern, then `babel-helper-plugin-utils` ends up hard-coding the current stable version of Babel, and third-party plugins using this would potentially be locking themselves to version 7 if we eventually release a version 8.x that still supports plugins that work on 6 and 7, which seems likely to happen.

So then my question is, should we consider something like

```
import declare from "babel-helper-plugin-utils";

  api.assertVersion("^7.0.0-0");
  const { loose } = options; // or whatever options the plugin has
```
with `declare` wrapping the logic encapsulated above like as
```
function declare(handler) {
  return (api, options, dirname) => {
    if (!api.assertVersion)
      api = {
        ...api,
        assertVersion(version) {
          throw new Error(
            `This plugin requires at least Babel ${version} and appears to be running on version ${api.version}.`,
          );
        },
      };
    }
    return handler(api, options, dirname);
  };
}
```
and `babel-core` itself exposing an `assertVersion` function that uses semver to check a value vs its own version. I propose that as opposed to having `semver` a dependency of the helper itself so that if people don't want to use the helper, they still have access to the API, just with fewer guarantees and helpful warnings. It also means `npm` will have to do less work deduplicating packages, and ideally we have a goal of keeping the helper utils module as small as possible.

Personally I like the approach the most because it gives us a base to build on if we want to add functionality to the API, while still offering useful errors when plugins access things that they don't support.

loganfsmyth:
| Fixed Issues?            | Fixes #7265
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Not really
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

This allows plugins to use `api.assertVersion(7)` as a general version, or any semver string to assert that the current version of Babel is one that is supported.

Trying to load `@babel/preset-env` for instance with Babel 6's CLI results in

```
Error: Requires Babel "^7.0.0-0", but was loaded with "6.26.0". If you are sure you have a compatible version of @babel/core, it is likely that something in your build process is loading the wrong version. Inspect the stack trace of this error to look for the first entry that doesn't mention "@babel/core" or "babel-core" to see what is calling Babel. (While processing preset: "/Users/<my stuff>/babel/packages/babel-preset-env/lib/index.js")
    at throwVersionError (/Users/<my stuff>/babel/packages/babel-helper-plugin-utils/lib/index.js:45:11)
    at Object.assertVersion (/Users/<my stuff>/babel/packages/babel-helper-plugin-utils/lib/index.js:11:11)
    at _default (/Users/<my stuff>/babel/packages/babel-preset-env/lib/index.js:132:7)
    at /Users/<my stuff>/babel/packages/babel-helper-plugin-utils/lib/index.js:17:12
    at /private/tmp/node_modules/babel-core/lib/transformation/file/options/option-manager.js:317:46
    at Array.map (<anonymous>)
    at OptionManager.resolvePresets (/private/tmp/node_modules/babel-core/lib/transformation/file/options/option-manager.js:275:20)
    at OptionManager.mergePresets (/private/tmp/node_modules/babel-core/lib/transformation/file/options/option-manager.js:264:10)
    at OptionManager.mergeOptions (/private/tmp/node_modules/babel-core/lib/transformation/file/options/option-manager.js:249:14)
    at OptionManager.init (/private/tmp/node_modules/babel-core/lib/transformation/file/options/option-manager.js:368:12)
    at File.initOptions (/private/tmp/node_modules/babel-core/lib/transformation/file/index.js:212:65)
    at new File (/private/tmp/node_modules/babel-core/lib/transformation/file/index.js:135:24)
    at Pipeline.transform (/private/tmp/node_modules/babel-core/lib/transformation/pipeline.js:46:16)
    at transform (/private/tmp/node_modules/babel-cli/lib/babel/util.js:50:22)
    at Object.compile (/private/tmp/node_modules/babel-cli/lib/babel/util.js:59:12)
    at /private/tmp/node_modules/babel-cli/lib/babel/file.js:142:23
    at walk (/private/tmp/node_modules/babel-cli/lib/babel/file.js:133:16)
    at files (/private/tmp/node_modules/babel-cli/lib/babel/file.js:156:7)
    at module.exports (/private/tmp/node_modules/babel-cli/lib/babel/file.js:184:5)
    at Object.<anonymous> (/private/tmp/node_modules/babel-cli/lib/babel/index.js:129:1)
    at Module._compile (module.js:660:30)
    at Object.Module._extensions..js (module.js:671:10)
    at Module.load (module.js:573:32)
    at tryModuleLoad (module.js:513:12)
```

where before it would throw about accessing `loose` on `undefined` or something along those lines because the `options` object isn't passed to plugins in Babel 6.x.

### DIFF
--- a/packages/babel-plugin-external-helpers/src/index.js
+++ b/packages/babel-plugin-external-helpers/src/index.js
@@ -1,9 +1,12 @@
+import { declare } from "@babel/helper-plugin-utils";
 import { types as t } from "@babel/core";
 
-export default function() {
+export default declare(api => {
+  api.assertVersion(7);
+
   return {
     pre(file) {
       file.set("helpersNamespace", t.identifier("babelHelpers"));
     },
   };
-}
+});
--- a/packages/babel-core/package.json
+++ b/packages/babel-core/package.json
@@ -41,6 +41,7 @@
     "lodash": "^4.2.0",
     "micromatch": "^2.3.11",
     "resolve": "^1.3.2",
+    "semver": "^5.4.1",
     "source-map": "^0.5.0"
   },
   "devDependencies": {
--- a/packages/babel-core/src/config/helpers/config-api.js
+++ b/packages/babel-core/src/config/helpers/config-api.js
@@ -1,4 +1,7 @@
 // @flow
+
+import semver from "semver";
+import { version as coreVersion } from "../../";
 import type { CacheConfigurator, SimpleCacheConfigurator } from "../caching";
 
 type EnvFunction = {
@@ -9,9 +12,11 @@ type EnvFunction = {
 };
 
 export type PluginAPI = {
+  version: string,
   cache: SimpleCacheConfigurator,
   env: EnvFunction,
   async: () => boolean,
+  assertVersion: typeof assertVersion,
 };
 
 export default function makeAPI(
@@ -32,9 +37,55 @@ export default function makeAPI(
     });
 
   return {
+    version: coreVersion,
     cache: cache.simple(),
     // Expose ".env()" so people can easily get the same env that we expose using the "env" key.
     env,
     async: () => false,
+    assertVersion,
   };
+}
+
+function assertVersion(range: string | number): void {
+  if (typeof range === "number") {
+    if (!Number.isInteger(range)) {
+      throw new Error("Expected string or integer value.");
+    }
+    range = `^${range}.0.0-0`;
+  }
+  if (typeof range !== "string") {
+    throw new Error("Expected string or integer value.");
+  }
+
+  if (semver.satisfies(coreVersion, range)) return;
+
+  const limit = Error.stackTraceLimit;
+
+  if (typeof limit === "number" && limit < 25) {
+    // Bump up the limit if needed so that users are more likely
+    // to be able to see what is calling Babel.
+    Error.stackTraceLimit = 25;
+  }
+
+  const err = new Error(
+    `Requires Babel "${range}", but was loaded with "${coreVersion}". ` +
+      `If you are sure you have a compatible version of @babel/core, ` +
+      `it is likely that something in your build process is loading the ` +
+      `wrong version. Inspect the stack trace of this error to look for ` +
+      `the first entry that doesn't mention "@babel/core" or "babel-core" ` +
+      `to see what is calling Babel.`,
+  );
+
+  if (typeof limit === "number") {
+    Error.stackTraceLimit = limit;
+  }
+
+  throw Object.assign(
+    err,
+    ({
+      code: "BABEL_VERSION_UNSUPPORTED",
+      version: coreVersion,
+      range,
+    }: any),
+  );
 }
--- a/packages/babel-helper-plugin-utils/package.json
+++ b/packages/babel-helper-plugin-utils/package.json
@@ -1,0 +1,10 @@
+{
+  "name": "@babel/helper-plugin-utils",
+  "version": "7.0.0-beta.40",
+  "description": "General utilities for plugins to use",
+  "author": "Logan Smyth <loganfsmyth@gmail.com>",
+  "homepage": "https://babeljs.io/",
+  "license": "MIT",
+  "repository": "https://github.com/babel/babel/tree/master/packages/babel-helper-plugin-utils",
+  "main": "lib/index.js"
+}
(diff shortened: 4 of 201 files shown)


packages/babel-plugin-external-helpers/src/index.js
```diff
@@ -1,9 +1,12 @@
+import { declare } from "@babel/helper-plugin-utils";
 import { types as t } from "@babel/core";
 
-export default function() {
+export default declare(api => {
+  api.assertVersion(7);
```
loganfsmyth: Definitely curious for feedback on this approach. It's a little verbose, but it means that since this is the function that throws, the preset/plugin doing the validation is the thing that throws, so the name of it shows up in the stack trace making debugging a lot nicer.

probablyup: Would it make sense to convert this to "assertSemverVersioning" or something like that? Presumably it will be possible for plugins to be compatible with multiple major versions of Babel.

loganfsmyth: Hmm, I worry about that getting pretty verbose. Is the concern that people will assume that `assertVersion` requires a single version?

`api.assertVersion(">=7.x < 9.x")` I at least think is acceptable.

evan-scott-zocdoc: >  Is the concern that people will assume that assertVersion requires a single version?

Yeah exactly. When I read `assertVersion` and saw your example it wasn't immediately obvious that a semver range could be given.

probablyup: > Is the concern that people will assume that assertVersion requires a single version? Yeah exactly. When I read `assertVersion` and saw your example it wasn't immediately obvious that a semver range could be given.

loganfsmyth: I feel like documentation would probably be enough to address that, but I do see where you're coming from. I was thinking of the function like `assertVersionSatisfies(...)` so the `version` is the Babel core version, not argument string, so the fact that the string can be semver seemed not strongly tied to the mention of version in the name. 

At the end of the day, most people  bothering to put this assertion in in the first place will probably look at our docs, or else copy paste and not give it a second thought. Since this is only an issue once we start releasing future versions, it seems expected that once it becomes important, people will wonder what specifically it supports as an argument, and at that point the docs are enough.

danez:
I at first thought maybe it would be nicer if modules that contain plugins simple export a second constant which is the version of babel-core that this plugin supports.
But that fails as soon as the plugin is required outside of babel-core which is not an uncommon case I guess. So I think it has to be inside the plugin itself.

I don't think it is too verbose, but if so then we could maybe include the assert in the helper?
```
export default declare('>=7.2.3', () => plugin);
```
But I also like how it is now.

loganfsmyth:
@danez Yeah, I was on the fence about that approach and mostly I opted to go with this PR's approach instead because having it be a programmatic API like `.assertVersion` means that:

1. The API is easy to use, even if you're not using `@babel/helper-plugin-utils`.
2. Because it is called from inside the plugin, the plugin filename shows up in the stack trace, which helps with debugging.

`declare(range, fn)` is also an API that could be built on top `.assertVersion`, but not the other way around, so I figured it would be a better low-level API.

PR #7452, opened 2018-02-28
test(plugin-proposal-class-properties): add failing test for decorator interop

buschtoens:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes, No
| Documentation PR         | 
| Any Dependency Changes?  |

This PR adds a failing test for the `proposal-class-properties` plugin, that shows an incompatibility with decorators in the TDZ check.

Specifically, there seems to be a problem with the scope. `classRef` is `undefined` here

https://github.com/babel/babel/blob/a07f96ce3f28ed053c7e5e061dae3627ff451bc5/packages/babel-plugin-proposal-class-properties/src/index.js#L123

and `path.scope.getBinding(path.node.name)` is `undefined` here

https://github.com/babel/babel/blob/a07f96ce3f28ed053c7e5e061dae3627ff451bc5/packages/babel-plugin-proposal-class-properties/src/index.js#L36

Therefore the above `if` condition evaluates to `true` and errorneously adds the `classNameTDZError`.


buschtoens:
Referencing https://github.com/babel/babel/pull/6855, pinging @maurobringolf :smile:

buschtoens:
There may be an edge case without the decorators transform that triggers this as well, but I could not come up with one.

maurobringolf:
Thank you for providing the test case. I'm looking into it but I am not familiar with the decorators plugin yet.

nicolo-ribaudo:
Merged in #8463. Thank you!

PR #6179, opened 2017-08-31
WIP: Migrate to jest

danez:
| Q                        | A 
| ------------------------ | ---
| Patch: Bug Fix?          | No
| Major: Breaking Change?  |  No
| Minor: New Feature?      | Yes
| Deprecations?            | No
| Spec Compliancy?         | Yes
| Tests Added/Pass?        | Yes
| Fixed Tickets            | Fixes #5197, Fixes #5359 
| License                  | MIT
| Doc PR                   | No
| Dependency Changes       | Yes

Rebase and squash of #5359 

This does not work correctly so far because facebook/jest#4393 is not released yet.

Some tests are not working and I haven't found a solution or did not look at them yet.

* All babel-register tests fail, because we cannot really use babel-register inside the jest-node-environment and we do not have access to a real require inside the test. Not sure what we should do here. Even if i try to write our own environment that runs in the same context, require would still be hijacked by jest and not do the real internal require cycle I think?
* babel-plugin-transform-es2015-block-scoping/general › label fails because babylon cannot parse the source.
* packages/babel-core/test/api.js - complex plugin and preset ordering is failing because the order does not match. Not sure why.

//cc @cpojer

PR #7455, opened 2018-02-28
Use jest

danez:
| Fixed Issues?            | #6179 #6442
| Patch: Bug Fix?          |n
| Documentation PR         | 
| Any Dependency Changes?  |

Rebase of #6179. Seems i removed my fork repo. :)

All tests work, babel-register is skipped for now. But I have an idea how to fix it, by mocking pirates and just checking what babel-register is doing, pirates is covered with tests anyway so we don't have to test it.


test/testSetupFile.js
```diff
@@ -0,0 +1,2 @@
+/* eslint-env jasmine */
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 10000;
```
SimenB: you can use [`jest.setTimeout`](https://facebook.github.io/jest/docs/en/jest-object.html#jestsettimeouttimeout)

.babelrc.js
```diff
@@ -32,11 +43,4 @@ if (process.env.BABEL_ENV === "cov") {
   config.plugins.push("babel-plugin-istanbul");
```
SimenB: should this still be here? Jest adds it itself (and you removed it from package.json)

danez: good catch

test/testSetupFile.js
danez: Fixed

.babelrc.js
```diff
@@ -40,7 +40,6 @@ const config = {
 
 if (process.env.BABEL_ENV === "cov") {
   config.auxiliaryCommentBefore = "istanbul ignore next";
```
SimenB: jest also does this

danez: :) even better

SimenB: https://github.com/facebook/jest/blob/a88d3c553d94b220f9dae47f64983d9d7b3baed8/packages/jest-runtime/src/script_transformer.js#L153-L169

hzoo:
@danez maybe you didn't see https://github.com/babel/babel/pull/6442? we were going to merge that but not sure the relation @rogeliog

danez:
I think there is no need to use the mocha runner, because just using the default jest runner works fine.

.travis.yml
```diff
@@ -7,9 +7,9 @@ cache:
   directories:
   - node_modules
 node_js:
+  - '9'
   - '8'
   - '6'
-  - '4'
```
hzoo: This seems unrelated? We are we dropping node 4 testing already?

packages/babel-code-frame/src/index.js
```diff
@@ -180,15 +180,17 @@ export default function(
   if (!deprecationWarningShown) {
     deprecationWarningShown = true;
 
-    const deprecationError = new Error(
```
hzoo: I'm ok with just not doing the warning anymore and making this a breaking change. We can just fix this automatically with https://github.com/babel/babel-upgrade

.travis.yml
existentialism: Can use `node` environment and keep 4? (like in babel-upgrade)

hzoo: If it's because of jest, it still works on node 4 if you use the node environment - https://github.com/babel/babel-upgrade works fine with it.

danez: ok cool, I revert then.

packages/babel-code-frame/src/index.js
danez: We can do this in a separate PR.

rogeliog:
I agree with @danez! The Mocha runner is a good option when the complexity of the codebase makes it hard to fully migrate to Jest or as an intermediate step. Seem that I'll close it in favor of this, I can reopen it if needed.

hzoo:
Ok well if this works I'm for it too 👍 , thanks for your efforts @rogeliog, sorry we didn't actually end up merging the PR 😅

danez:
Ok I hope the coverage is now correct. I had to readd the babel-plugin-istanbul as some of our tests test the code in `lib` and not `src`. Wherever possible I changed to test `src` directly, but the fixture runner which runs stuff in separate envs couldn't be changed easily.

SimenB:
Anything jest can do to help, or is it just a quirk (feature) of babel's tests?

danez:
I think that we in the past implemented some features that jest might already do, like running each test in a spearate vm (https://github.com/babel/babel/blob/master/packages/babel-helper-transform-fixture-test-runner/src/index.js#L87). 
The other thing is that we bundle some packages with rollup/webpack and we want to test the bundle rather than the source in this cases to ensure that these works.
We might be able to get rid of some of this now, but I want to make this PR as small as possible.

Andarist:
Super cool 👍I was wondering - it should be possible to migrate to builtin snapshots instead of custom setup in the future, right? This would allow easier test updates

hzoo:
I actually like the individual files lol compared to reading the snapshot output now lol

Andarist:
What's the difference though? A single snapshot format? Maybe it's possible to integrate with snapshot system to output whatever we want in whatever directory we want, just have it backed by jest's built-in system for this. `expected` files are a form of snapshots anyway

hzoo:
Oh ok if you can change the format then that's cool. I just mean https://github.com/babel/babel-upgrade/blob/51534a4bd3f742d284d42e231ae36e71c83bc93d/__tests__/__snapshots__/packageData.test.js.snap#L5-L12 vs the actual js

Andarist:
Actually I have no idea if its possible to change the whole format, it's for sure possible to write custom serializer, but not sure if a snapshot file can be an actual js file, not assigned to exports etc (jest snapshots are actually requireable). Worth investigating in the future

PR #7466, opened 2018-03-02
Fix for running babel in a symlinked directory

AlastairTaft:
For my use case, I'm using yarn workspaces, however I want to be able to symlink a package under the workspace file structure. Due to the way yarn only installs babel packages in the root `node_modules` directory, without this change the directory resolves to the real location which then can't resolve any of the babel preset and plugin packages when it runs babel as part of the `install` and I end up with errors similar to the below.

```
Error: Couldn't find preset "env" relative to directory "src"
```

| Fixed Issues?            | 
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-core/src/transformation/file/options/build-config-chain.js
+++ b/packages/babel-core/src/transformation/file/options/build-config-chain.js
@@ -34,7 +34,7 @@ export default function buildConfigChain(opts: Object = {}, log?: Logger) {
   builder.mergeConfig({
     options: opts,
     alias: "base",
-    dirname: filename && path.dirname(filename)
+    dirname: filename && path.resolve(__dirname, path.dirname(filename))
   });
 
   return builder.configs;


loganfsmyth:
We're not going to change Babel 6 for this either way. This fix would also break in any case since it does not make sense to resolve the filename relative to this location.

We are doing work in 7.x to making efforts to allow symlink workflows to be easier.

AlastairTaft:
Okay, thanks for the note on better symlink support in 7.

AlastairTaft:
If it's helpful for anyone else my final monkey patch solution to get this working with yarn workspaces was to update the `buildConfigChain()` function in the `babel-core/lib/transformation/file/options/build-config-chain.js` file to the following.

This assumes you have a mono repo depending on another mono repo with the file structure
`sameParent/root` (root yarn workspaces repo)
`sameParent/reformed/packages` (Mono repo with other dependent packages)

```es6
function buildConfigChain() {
  var opts = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : {};
  var log = arguments[1];

  var filename = opts.filename;
  var builder = new ConfigChainBuilder(log);

  if (opts.babelrc !== false) {
    builder.findConfigs(filename);
  }

  builder.mergeConfig({
    options: opts,
    alias: "base",
    dirname: filename && _path2.default.dirname(filename)
  });

  builder.configs[0].alias = maybeMonkeyPatchPath(builder.configs[0].alias)
  builder.configs[0].loc = maybeMonkeyPatchPath(builder.configs[0].loc)
  builder.configs[0].dirname = maybeMonkeyPatchPath(builder.configs[0].dirname)

  /**
   * This adds in 'root' to the directory path used so that it follows the
   * symlinked directory structure rather than the actual file system structure.
   * This makes babel work when ran from the parent mono-repo.
   * Hopefully this will not be needed in babel 7 as they are improving symlink
   * support.
   */
  function maybeMonkeyPatchPath(path){
    var index = path.indexOf('/reformed/packages')
    if (index == -1) // Pass through legit paths unchanged
      return path
    return path.slice(0, index) + '/root' + path.slice(index)
  }

  return builder.configs;
}
```

loganfsmyth:
Are you compiling with Webpack, or what? It is hard to give alternate suggestions without more information.

AlastairTaft:
I'm running it like this, in my `root` folder I have an `employee-portal` package that depends on a `portal-shared` package. But the `portal-shared` package lives at `sameParent/reformed/packages/portal-shared`, however its symlinked under `sameParent/root/reformed/packages/portal-shared`. 

The `portal-shared` package has a `start` script that runs `babel src --out-dir lib --watch` which I run from the `root` folder with this command `npm start --prefix reformed/packages/portal-shared`.

Originally the `portal-shared` package ran babel as an `install` script but due to that not working on `yarn install`, due to above issue, I removed all the `install` scripts and checked in the generated `lib` folder so it didn't need to do any building as part of the install, a bit crap but I just want it to work so I can get on with developing.

However to make updates during development I still need to run the `start` script hence the post install monkey patching to babel core.

Yeah I use webpack to run `employee-portal` but I don't think it effects any of the above?

That said, at the moment it all works with the monkey patch and I can get on with development, I'm holding out that babel 7 will make the issue go away.

loganfsmyth:
> Yeah I use webpack to run employee-portal but I don't think it effects any of the above?

Mostly I was wondering if you were using babel-cli, or babel-loader to compile. So I understand, you're using the CLI to compile, rather than using babel-loader?

It would also help to know where your `.babelrc` is, and where the `node_modules` containing the presets are actually installed.

AlastairTaft:
Yeah, using the cli to compile, the issue is with compiling the `portal-shared` package because yarn workspaces puts the babel dependencies under `root/node_modules`. So when it uses the real path `sameParent/reformed/package/portal-shared` it can't find those dependencies. My hack changes the path to `sameParent/root/reformed/package/portal-shared` so that it can find `sameParent/root/node_modules`.

loganfsmyth:
Do you have a `.babelrc` in `employee-portal`, or `portal-shared`, or are you passing the plugin/preset names directly to the `babel` command?

AlastairTaft:
My `.babelrc` in `employee-portal`

```
{
  "presets": ["es2015", "stage-0", "react"]
}
```

Which runs the webpack rule 
```
{
  test: /\.js$/,
  loader: 'babel-loader',
  include: path.join(__dirname, 'src'),
}
```
My `.babelrc` in `portal-shared`

```
{
  "presets": [ "env" ]
}
```

loganfsmyth:
So to repeat that back to you, that'd be like
```
sameParent/
  root/
    node_modules/
      // workspaces puts plugins/presets here

    employee-portal/
      reformed/packages/portal-shared/ -> ../../../../../reformed/packages/portal-shared
      .babelrc => { "presets": ["es2015", "stage-0", "react"] }

      // In here, runs "npm start --prefix reformed/packages/portal-shared"
      // _and_ loads an default-configured babel-loader

  reformed/
    packages/
      portal-shared/
        package.json#scripts
          start -> "babel src --out-dir lib --watch"
        .babelrc => { "presets": [ "env" ] }
```

Assuming I have that all right, my expectations would be that there's also be a separate `sameParent/reformed/portal-shared/node_modules/` that would have `babel-preset-env`, and the rest of the presets are for the other config are in `sameParent/root/node_modules/`.

Am I understanding right that you're saying that `babel-preset-env` doesn't exist in `sameParent/reformed/portal-shared/node_modules/` and instead is placed in `sameParent/root/node_modules/`?

If so, that seems like a bug in workspaces.

AlastairTaft:
Not quite right

```
sameParent/
  root/
    node_modules/
      // workspaces puts plugins/presets here
    employee-portal/
      .babelrc => { "presets": ["es2015", "stage-0", "react"] }
      // In here is a webpack config, that has a babel-loader
      // It runs a server using web pack dev middleware
    reformed/packages/portal-shared/ -> ../reformed/packages/portal-shared
  reformed/
    packages/
      portal-shared/
        package.json#scripts
          start -> "babel src --out-dir lib --watch"
        .babelrc => { "presets": [ "env" ] }
```

yarn workspaces puts any duplicated dependencies in `sameParent/root/node_modules`, there's other dependencies than the ones listed, but the result is all presets end up in `sameParent/root/node_modules`

> Am I understanding right that you're saying that babel-preset-env doesn't exist in sameParent/reformed/portal-shared/node_modules/ and instead is placed in sameParent/root/node_modules/?

Yeah, but that's correct behavior based on the symlinked path its putting duplicate dependencies in the root `node_modules` directory, which is a good thing. I guess the problem lies in it using the physical path rather than the symlinked one, perhaps that's a yarn issue? When I delved into babel core, it looked like it was using `process.cwd()`, which would bring back the physical path, but depending on how yarn called it I'm not sure if that's correct or not.

AlastairTaft:
On second thought, I don't think it's a `yarn` problem because I can repo it when running `npm start --prefix reformed/packages/portal-shared`

loganfsmyth:
`portal-shared`'s `package.json` has `babel-preset-env` in its dependencies, right? Given that, if the dependency gets installed anywhere outside that specific directory hierarchy of

```
sameParent/reformed/packages/portal-shared`
```
that seems like something has gone wrong with the installation. If it is in a `package.json` in that folder, it should absolutely be installed into a `node_modules` folder in one of those.

AlastairTaft:
Yeah it's in there as a dependency in `portal-shared`'s `package.json`. And it ends up living in `sameParent/root/node_modules` as part of the `yarn install` for workspaces. Because the path `yarn` sees it at is `sameParent/root/reformed/packages/portal-shared`.

Issue #7465, opened 2018-03-02
babel-node


jdalton:
babel-node doesn't currently support Node's `--require` and `-r` flag. It would great if it did.

hzoo:
Fixed by https://github.com/babel/babel/pull/7471

PR #7471, opened 2018-03-02
added support Node's --require and -r flags in babel-node

yakotika:
| Fixed Issues?            |  #7465  
| Tests Added + Pass?      | n
| Documentation PR         | 

### DIFF
--- a/packages/babel-node/src/babel-node.js
+++ b/packages/babel-node/src/babel-node.js
@@ -52,8 +52,8 @@ getV8Flags(function(err, v8Flags) {
 
       case "-r":
       case "--require":
-        args.push(babelArgs[index + 1]);
         args.push(flag);
+        args.push(babelArgs[index + 1]);
         delete babelArgs[index + 1];
         break;
 
--- a/packages/babel-node/src/_babel-node.js
+++ b/packages/babel-node/src/_babel-node.js
@@ -23,6 +23,7 @@ function collect(value, previousValue): Array<string> {
 
 /* eslint-disable max-len */
 program.option("-e, --eval [script]", "Evaluate script");
+program.option("-r, --require [module]", "Require module");
 program.option("-p, --print [code]", "Evaluate script and print result");
 program.option(
   "-o, --only [globs]",
@@ -132,7 +133,11 @@ if (program.eval || program.print) {
 
       if (arg[0] === "-") {
         const parsedArg = program[arg.slice(2)];
-        if (parsedArg && parsedArg !== true) {
+        if (
+          arg === "-r" ||
+          arg === "--require" ||
+          (parsedArg && parsedArg !== true)
+        ) {
           ignoreNext = true;
         }
       } else {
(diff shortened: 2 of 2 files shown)


packages/babel-node/src/babel-node.js
```diff
@@ -50,6 +50,11 @@ getV8Flags(function(err, v8Flags) {
         args.unshift(arg);
         break;
 
+      case "-r":
+      case "--require":
+        args.unshift(flag);
```
danez: `--require` has an option according to the docu: https://nodejs.org/api/cli.html#cli_r_require_module
So using `flag` here is probably wrong and should be `arg` instead, but I haven't tested it, so correct me if I'm wrong.

danez: Although it seems I missed that the module is not separated by equal sign but with space in this case, so `flag` and `arg` are the same. Still I think this will not work as the preceding module name will be pushed to args whereas this here gets unshifted.

yakotika: Yes, `flag` and `arg` are equal. You propose to change the `unshift` on that ? `push` ?

yakotika:
@danez check plz

yakotika:
@danez ckeck plz

mkovel:
Hi guys, can you push forward a little with this task? I am very looking forward it. 
Thanks.

danez:
Okay I checked it and it wasn't working. I pushed a commit now it is working

mkovel:
Guys, when do you plan, include this awesome feature in a release?

jdalton:
Hi all!

Thanks for adding this feature. At the moment though it looks like the `-r` option doesn't resolve modules like the traditional `node -r` by that I mean `babel-node -r esm` will result in an error
```
Cannot find module '/Users/jdalton/projects/bar/esm'
```
because it's not resolving `esm` to the node_modules folder of
```
'/Users/jdalton/projects/bar/node_modules/esm'
```

danez:
I see so we should change this line https://github.com/babel/babel/pull/7471/files#diff-bd092c1aa1f13cc3b620707dc4fc8a1eR153 to check if the path starts with dot and only then append cwd. In the other cases (absolute paths starting with / or modules starting with [a-zA-Z]) we can simply require them.

jdalton:
@danez Node has an API, `require.resolve(path, { paths:['.'] })`], [in Node 8+](https://nodejs.org/dist/latest-v10.x/docs/api/all.html#modules_require_resolve_request_options) that allows using the `require.resolve(path, ["."])` which would use Node's own path resolution to resolve `path` to the current working directly. This would handle resolving absolute, local, and package names. There may be a package on npm to backport the API for Node 6 (the actual bit of code internally is like 5 lines).

```js
const Module = require('module')
const fakeParent = new Module('', null)

fakeParent.paths = Module._nodeModulePaths('.')
const lookupPaths = ['.', ...Module._resolveLookupPaths(request, fakeParent, true)]
const filename = Module._findPath(request, paths, false)
```

loganfsmyth:
My vote would be to use [`resolve`](https://www.npmjs.com/package/resolve) rather than rely on internals of Node, but otherwise woulds like a good plan.

PR #6107, opened 2017-08-15
Decorators 2 Transform [WIP]

peey:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues             | Progress on https://github.com/babel/proposals/issues/11
| Patch: Bug Fix?          | No
| Major: Breaking Change?  | No
| Minor: New Feature?      | Yes
| Tests Added/Pass?        | Yes
| Spec Compliancy?         | Yes
| License                  | MIT
| Doc PR                   | No
| Any Dependency Changes?  | Yes, on decorators2 syntax plugin

Note: I'll use the terms *evaluation* to mean evaluation of `@dec(args)` and *decoration* to mean actual execution of the decorators

Also, note that the decorator proposal introduces the concept of an [*elementDescriptor*](https://tc39.github.io/proposal-decorators/#sec-to-element-descriptor), which is distinct from the [*propertyDescriptor*](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor) we all know about.

To understand the code, you need to know that an *elementDescriptor* is an object with the shape:
```
{
  kind: "property", // I don't know what other values kind can take
  key: <the name of the method>,
  isStatic: true|false,
  descriptor: <the property descriptor of the method>
}
```

I've implemented the methods `decorate`, `decorateElement`, `decorateClass`, `makeElementDescriptor`, and `mergeDuplicateElements` in `helpers.js`. Out of these, `decorateElement` and `decorateClass` correspond to the spec operations [*DecorateElement*](https://tc39.github.io/proposal-decorators/#sec-decorate-element), and [*DecorateClass*](https://tc39.github.io/proposal-decorators/#sec-decorate-class) respectively. `mergeDuplicateElements` corrseponds to the operation used in step 6 of both *DecorateClass* and *DecorateElement*. 

The method `decorate` runs the whole thing. It takes care of the evaluation order (see below) and corrsponds to [Runtime Semantics: ClassDefinitionEvaluation](https://tc39.github.io/proposal-decorators/#sec-runtime-semantics-classdefinitionevaluation) step 21 onwards.

The changes in babel-core are from https://github.com/babel/babel/pull/6058

## Assumptions Made: (Please feel free to correct)
1. evaluation order is as the following: ([relevant spec operation](https://tc39.github.io/proposal-decorators/#sec-runtime-semantics-classdefinitionevaluation))
[...]

PR #7475, opened 2018-03-03
Fix class properties tdz decorator interop

### LINKED ISSUE #7705: Symbol.iterator fails when decorators and class-properties are used in Babel 7
**Bug report**

A class that implements `Symbol.iterator` when a decorator is used on a class property produces the following error when running the transpiled code:

> Error: Class "Symbol" cannot be referenced in computed property keys.

[Gist to demonstrate the issue](https://gist.github.com/AlexanderOMara/58ab0e943a441120758dbe472fcb47db)

### Input Code

```js
function property(enumerable = true, configurable = true, writable = true) {
	return (target, key, descriptor) => {
		descriptor.enumerable = enumerable;
		descriptor.configurable = configurable;
		descriptor.writable = writable;
	};
}

class Example extends Object {

	@property()
	test;

	constructor() {
		super();
	}

	[Symbol.iterator]() {
		return {
			next: function() {
				return {done: true};
			}
		};
	}
}
```

```js
{
	"presets": [],
	"plugins": [
		["@babel/plugin-proposal-decorators"],
		["@babel/plugin-proposal-class-properties", {
			"loose" : true
		}]
	]
}
```

[See transpiled code here](https://gist.github.com/AlexanderOMara/58ab0e943a441120758dbe472fcb47db#file-index-js), in particular this part:

```js
_iterator = (_classNameTDZError("Symbol"), Symbol).iterator
```

### Expected Behavior

Expected `Symbol.iterator` to function normally, as it does when there are no decorated class-properties.

### Current Behavior

Transpiled code unconditionally throws an error when these conditions are met.

### Context

Just trying to use `Symbol.iterator`, class `@babel/plugin-proposal-decorators`, and `@babel/plugin-proposal-class-properties` at the same time.

It's possible to work around by aliasing `Symbol.iterator` with a local variable, but that's very unpleasant, especially for TypeScript code (other TypeScript parsing code cannot infer the type properly then).

### Your Environment

| ---------------- | -------
| Babel            |  7.0.0-beta.41
| node             | v9.11.1
| npm              | 5.6.0
| Operating System | macOS 10.13.4

maurobringolf:
| Fixed Issues?            | Closes #7452, fixes #7705
| Documentation PR         | 
| Any Dependency Changes?  |

The decorators plugin does some renaming and rearranging of classes which can break my code from #6855 . I figured out that I am unnecessarily recomputing the class binding via `ref.name` when I can get it directly via `path.scope.getBinding(path.node.id.name)`. I left `ref` untouched and renamed the class binding used in the TDZ check to `classBinding`.

I am not sure what's the best Git way of doing this. I added @buschtoens commit with the test case to my fork, so this PR contains his too.


packages/babel-plugin-proposal-class-properties/test/fixtures/property-tdz-decorator-interop/exec.js
```diff
@@ -0,0 +1,7 @@
+function dec() {}
```
Andarist: could we test something here? it seems that this show only the output, so technically could be just a fixture test, u dont actually evaluate the output right now so i think it doesnt justify using `exec.js`

buschtoens: Before the fix `exec.js` incorrectly threw a runtime error. There was no way to determine the correct `output.js` fixture back then. Personally, I would add the (currently defunct) [`proposal-decoratoras` transform](https://github.com/babel/babel/tree/master/packages/babel-plugin-proposal-decorators) to make sure that once https://github.com/babel/babel/pull/6107 lands, independently of how both transforms interplay, they still yield valid, executable code.

maurobringolf: Converting to a fixture makes sense to me. I renamed to `input.js` but was unable to get an output generated locally. Committed to see what the CI does.

Also I noticed that some other tests are called `input.js` but executed? For example https://github.com/babel/babel/tree/master/packages/babel-plugin-proposal-class-properties/test/fixtures/static-property-tdz-edgest-case.

existentialism:
Can we remove the yarn.lock?

maurobringolf:
@existentialism Sure, I removed it

maurobringolf:
Do you need anything from me for this one? I'd really like to get this in, since there are a lot of potential bugs with renamed classes with the current code.

jsg2021:
Is there a plan for this?

nicolo-ribaudo:
Yeah sorry, currently bugs related to the old decorators specification aren't getting much attention. We are working on implementing the new decorators proposal (https://github.com/babel/babel/pull/7976) which shouldn't be affected by many of these bugs.

jsg2021:
Thats understandable. Thanks for the reply.

JSteunou:
@nicolo-ribaudo but the issue is still present with `@babel/plugin-proposal-decorators` which is the new babel decorators plugin, isnt it?

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -33,7 +33,7 @@ export default function(api, options) {
     },
 
     ReferencedIdentifier(path) {
-      if (this.classRef === path.scope.getBinding(path.node.name)) {
+      if (this.classBinding === path.scope.getBinding(path.node.name)) {
```
nicolo-ribaudo: Instead of using `false` as a special value to indicate tha the class doen't have an `id`, I think that we should let it be undefined (as it was) and check `this.classBinding && this.classBinding === ...` here.

nicolo-ribaudo:
Merged in #8463. Thank you!

PR #7484, opened 2018-03-04
Require tests to use input.mjs for modules, and output.js/.mjs based on active transforms

loganfsmyth:
| Fixed Issues?            | 
| Minor: New Feature?      | Y (barely, `result.sourceType`)
| Documentation PR         | 
| Any Dependency Changes?  |

Give meaning to test fixture input and output file extensions so that we can validate that transforms actually manipulate the `sourceType` properly.

* To parse input as `sourceType: 'module'` the name must be `input.mjs`
* Output will be `.js` or `.mjs` based on if the sourceType was changed.


packages/babel-helper-transform-fixture-test-runner/src/index.js
```diff
@@ -386,11 +397,28 @@ function run(task) {
...
+        /\.m?js$/,
+        result.sourceType === "module" ? ".mjs" : ".js",
+      );
+
+      console.log(`New test file created: ${expectedFile}`);
+      fs.writeFileSync(expectedFile, `${result.code}\n`);
+
+      if (expected.loc !== expectedFile) {
+        try {
+          fs.unlinkSync(expected.loc);
```
nicolo-ribaudo: I don't understend when this code is executed. If `!expected.code` (line 394), shouldn't `output.m?js` not exist?

loganfsmyth: It's not super likely, so that's a fair question. Technically the file could have been emptied instead of deleted, so I figured it wouldn't hurt to add some logic to try to avoid creating sibling files.

Issue #7476, opened 2018-03-03
Migrate chai and assert usages to jest expect


danez:
We currently - now after jest has been landed - have a mix of different assertion methods in tests.

- There is `chai` in use.
- There is `assert` in use.
- There is jests `expect` in use.

The first two should be removed and rewritten to make use of `expect` which comes with jest.

so for example

```diff
-import chai from "chai";
-import assert from "assert";
-
-chai.expect(actualCode).to.be.equal("var fn;");
+expect(actualCode).toBe("var fn;")
-assert.strictEqual(this.constructor, Bar);
+expect(this.constructor).toBe(Bar)
```

The jest matchers should be used as good as possible, so for example not `expect(x > 3).toBe(true)` but instead `expect(x).toBeGreaterThanOrEqual(3)`

As we have a lot of tests split up in a lot of packages this task can also be split up and only one/some package(s) at a time be migrated.

Andarist:
> The jest matchers should be used as good as possible, so for example not expect(x >= 3).toBe(true) but instead expect(x).toBeGreaterThanOrEqual(3)

Should they really? API surface of those matchers is pretty big and I dont think that they actually bring any value to the code. Personally I prefer to stick to some basic matchers only if they can achieve the same thing and I wish jest would have something like `expect.ok(x >= 3)`.

danez:
I just prefer to do this because usually the reported error is better, but we don't have to.

devenbansod:
I'd love to take this up for my first PR. Shall I?

nicolo-ribaudo:
Yes ☺️

@Andarist I agree with Daniel: using `expect(x >= 3)` can only report something like `Expected false to be true`, which is totally unuseful.

Andarist:
Sure, error reports are totally valid point and after @danez has raised it, I'm totally on board with using those more 'complicated' matchers :)

devenbansod:
Thanks @nicolo-ribaudo for the confirmation. 

I was thinking that I'll start with migrating one of the small packages: `babel-code-frame`, make a WIP PR, request feedback and then expand it over other packages. Do you have any suggestions over the initial choice?

SimenB:
I'd recommend checking out [`jest-codemods](https://github.com/skovhus/jest-codemods). Even if you're unable to run it, the transforms in the package should tell you what assertions to switch to

devenbansod:
[...]

PR #7485, opened 2018-03-04
Migrate `babel-code-frame` tests to use `expect`

devenbansod:
| Any Dependency Changes?  | Might ultimately enable us to remove `assert` dependency

* Replace `assert.equal` calls with appropriate `expect` calls
* *Work-in-progress*: Currently targeting only `babel-code-frame` tests


Andarist:
Thanks!

PR #7487, opened 2018-03-04
Enable and fix babel-register tests

### LINKED ISSUE #7477: Fix and reenable babel-register tests
Instead of doing end2end tests which do not work with jests way of handling requires, we should mock `pirates` and instead only test the code that is actually in babel-register.

danez:
| Fixed Issues?            | Fixes #7477 
| Documentation PR         | 
| Any Dependency Changes?  |

I changed the base register test to mock `pirates and source-map support` and just test it calls these libs correctly.
The cache tests worked after fixing the cache invalidation with jest.

### DIFF
--- a/packages/babel-register/test/cache.js
+++ b/packages/babel-register/test/cache.js
@@ -28,13 +28,13 @@ function resetCache() {
   process.env.BABEL_DISABLE_CACHE = oldBabelDisableCacheValue;
 }
 
-describe.skip("@babel/register - caching", () => {
+describe("@babel/register - caching", () => {
   describe("cache", () => {
     let load, get, save;
 
     beforeEach(() => {
       // Since lib/cache is a singleton we need to fully reload it
-      delete require.cache[require.resolve("../lib/cache")];
+      jest.resetModuleRegistry();
       const cache = require("../lib/cache");
 
       load = cache.load;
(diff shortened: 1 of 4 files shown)


packages/babel-register/test/cache.js
```diff
@@ -28,13 +28,13 @@ function resetCache() {
...
 
-describe.skip("@babel/register - caching", () => {
+describe("@babel/register - caching", () => {
   describe("cache", () => {
     let load, get, save;
 
     beforeEach(() => {
       // Since lib/cache is a singleton we need to fully reload it
-      delete require.cache[require.resolve("../lib/cache")];
+      jest.resetModuleRegistry();
```
nicolo-ribaudo: Isn't this run automatically by jest before each test? (https://github.com/facebook/jest/issues/587#issuecomment-153263356)

danez: Maybe it changed? Or it is only run once for each test file? At least the tests weren't working without.

SimenB: Should be https://facebook.github.io/jest/docs/en/jest-object.html#jestresetmodules. `resetModuleRegistry` is a deprecated alias and not publicly documented.

https://github.com/facebook/jest/blob/0b076cdfd6020810fc05fb0f26cdca30d688ead1/packages/jest-runtime/src/index.js#L829-L830

It is not run automatically between tests in a single suite, but each suite (file) is run in isolation

danez: Ah thanks, I somewhere found it documented, but now I can't find it anymore :) Will fix it.

PR #7490, opened 2018-03-04
Give helpful errors if the wrong sourceType is detected

loganfsmyth:
| Fixed Issues?            | 
| Documentation PR         | 
| Any Dependency Changes?  |

Recommend either sourceType: unambiguous, script or module when a file is detected as having the wrong sourceType.

### DIFF
--- a/packages/babel-core/src/transformation/normalize-file.js
+++ b/packages/babel-core/src/transformation/normalize-file.js
@@ -89,6 +89,12 @@ function parser(pluginPasses, options, code) {
     }
     throw new Error("More than one plugin attempted to override parsing.");
   } catch (err) {
+    if (err.code === "BABEL_PARSER_SOURCETYPE_MODULE_REQUIRED") {
+      err.message +=
+        "\nConsider renaming the file to '.mjs', or setting sourceType:module " +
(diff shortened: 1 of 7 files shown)


packages/babel-core/src/transformation/normalize-file.js
```diff
@@ -89,6 +89,12 @@ function parser(pluginPasses, options, code) {
     }
     throw new Error("More than one plugin attempted to override parsing.");
   } catch (err) {
+    if (err.code === "BABEL_PARSER_SOURCETYPE_MODULE_REQUIRED") {
+      err.message +=
+        "\nConsider renaming the file to '.mjs', or setting sourceType:module " +
+        "or sourceType:unambiguous in your Babel config for this file.";
```
nicolo-ribaudo: Can we move this hint to Babylon, since it accepts the `sourceType` option?

loganfsmyth: The message in Babylon itself already says

> 'import' and 'export' may appear only with 'sourceType: "module"'

So this is mostly adding additional reminder about `.mjs` and Babel's config, since Babylon doesn't know anything about them.

Issue #7258, opened 2018-01-23
Incorrect compilation for empty array


kevinptt0323:
### Input Code
```js
const assert = require('assert')

let arr = new Array(3);
/*
 * or
 * let arr = [,,,];
 */
let arr2 = [...arr];

assert(arr2.hasOwnProperty(0));
```

### .babelrc
```js
{
  "presets": ["@babel/preset-env"]
}
```

### Expected Behavior
No error will be thrown

### Current Behavior
```js
var assert = require('assert');

var arr = new Array(3);
var arr2 = arr.concat();
assert(arr2.hasOwnProperty(0));
```

```
AssertionError [ERR_ASSERTION]: false == true
```

### Possible Solution
Since empty array (e.g., array with `new`) has difference (weird?) structure.

Add `_toConsumableArray` and check properties when compiling rest operator.

### Context
The following code will be incorrect.
```js
let tmp = new Array(100);
let series = [...tmp].map((_,i) => i);
// [0, 1, 2, 3, ..., 99]
```

### Your Environment

| software         | version(s)
| ---------------- | -------
| Babel            |  7.0.0-beta.38
| node             | 8.8.1
| npm              | -
| Operating System | -

Andarist:
@kevinptt0323 It seems that you are using a `loose` mode in which we do not use `_toConsumableArray` helper

kevinptt0323:
@Andarist 
No, the following code will use `_toConsumableArray`.
```js
let arr = [...new Array(3)];
```

```js
function _toConsumableArray(arr) { /* ... */ }

var arr = _toConsumableArray(new Array(3));
```

Andarist:
You are right, this needs a fix in default mode.

Andarist:
**Additional note**

This happens also in the destructuring transform for an example input of:
```js
var arr = new Array(3);
var [a, ...b] = arr;
```

Faulty code is in `scope.toArray` detection - identifiers gets passed in and it's checked if identifiers refers to a generic Array - it does, but it doesn't mean we shouldn't convert those arrays with `toConsumableArray`.

Need to dig further into the best way to fix this, but at least we know what's going on already.

PR #7491, opened 2018-03-05
Explicitly throw if the array rest/spread items are not iterable.

### LINKED ISSUE #2160: No runtime error for spreading non-iterables
`console.log([...2]);` produces code which outputs an empty array, instead of throwing a runtime error. (Whether or not this is capable of being fixed is unknown to me.)

loganfsmyth:
| Fixed Issues?            | Fixes #2160
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

`Array.from` will just silently create an empty array for non-iterable, non-arraylike objects, so we want to make sure we're actually show an error if a bad this is iterated.

### DIFF
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -401,7 +401,7 @@ export default class Scope {
     console.log(sep);
   }
 
-  toArray(node: Object, i?: number) {
+  toArray(node: Object, i?: number | boolean) {
     const file = this.hub.file;
 
     if (t.isIdentifier(node)) {
@@ -431,14 +431,20 @@ export default class Scope {
       );
     }
 
-    let helperName = "toArray";
+    let helperName;
     const args = [node];
     if (i === true) {
+      // Used in array-spread to create an array.
       helperName = "toConsumableArray";
     } else if (i) {
       args.push(t.numericLiteral(i));
+
+      // Used in array-rest to create an array from a subset of an iterable.
       helperName = "slicedToArray";
       // TODO if (this.hub.file.isLoose("es6.forOf")) helperName += "-loose";
+    } else {
+      // Used in array-rest to create an array
+      helperName = "toArray";
     }
     return t.callExpression(file.addHelper(helperName), args);
   }
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -687,17 +687,21 @@ helpers.slicedToArrayLoose = () => template.program.ast`
 
 helpers.toArray = () => template.program.ast`
   import arrayWithHoles from "arrayWithHoles";
+  import iterableToArray from "iterableToArray";
+  import nonIterableRest from "nonIterableRest";
 
   export default function _toConsumableArray(arr) {
-    return arrayWithHoles(arr) || Array.from(arr);
+    return arrayWithHoles(arr) || iterableToArray(arr) || nonIterableRest();
   }
 `;
 
 helpers.toConsumableArray = () => template.program.ast`
   import arrayWithoutHoles from "arrayWithoutHoles";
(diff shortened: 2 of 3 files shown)


nicolo-ribaudo:
What about `arguments`? In old browsers it doesn't have `Symbol.iterator`.

Kovensky:
That's a though one. Probably the only thing we can do is detect the `arguments` pseudo-keyword itself and replace it with `Array.from(arguments)`, because there's nothing else detectable about `arguments`.

nicolo-ribaudo:
Well, we can stringify it (I can't test right know, but if I recall correctly `Object.prototype.toString.call(arguments)` is `[object Arguments]`)

loganfsmyth:
I think I'm up for adding a case to handle arguments. Will give it a shot in a bit.

loganfsmyth:
Alright, this PR got bigger. Might be easiest to follow if you look at the commits individually.

packages/babel-traverse/src/scope/index.js
```diff
@@ -431,14 +431,20 @@ export default class Scope {
       );
     }
 
-    let helperName = "toArray";
+    let helperName;
```
Andarist: while you are tweaking this - any ideas how [this one](https://github.com/babel/babel/issues/7258) could be solved? The only thing I've came up with was to track holey-ness on bindings for simple cases & bail out for more complex ones.

loganfsmyth: I don't think the helpers themselves would need to change for that. Looks like the core issue is that we call `.concat` on the original. Maybe we just should call `toConsumableArray` on the original data too before `.concat`?

Andarist: It's indeed unrelated to this. I've only reminded myself about the other issue while looking at this one, so thought I'd ask about it.  

> Maybe we just should call toConsumableArray on the original data too before .concat?

That's ofc a possibility - would be good to optimize it though, as i.e. this `.concat` call gets generated often for arguments rest and we could analyze there that it's safe to call `.concat` in such situation. Going to look into this issue again, hopefully soon, gonna try to figure it out to be both safe & optimized if possible.

laurre:
How does this change play with the plugin proposal-object-rest-spread ? I can't get it to work, and I always get the exception mentioned in this issue

laurre:
This change broke my code that was working, no problem with that, I use create-react-app in the latest version that also brings in the plugin proposal-object-rest-spread, it works when I try it with simple objects.

My use case is a bit different. I was using the spread operator with a variable coming from an import, like this (simplified):


```
import { functions as configFunctions } from "../config";

const tab = [ ...configFunctions ];
```

The idea is to get all the functions declared in various modules in a single array. I get the 'Invalid attempt to spread non-iterable instance' error, but I don't get why, isn't the 'functions' variable a normal object ?

In my case the compiled code looks like this:

`var tab = Object(path_to_my_project_node_modules_babel_runtime_helpers_esm_toConsumableArray__WEBPACK_IMPORTED_MODULE_0__["default"])(_location__WEBPACK_IMPORTED_MODULE_8__["actions"]);`

Maybe the way I am doing this is not the most appropriate, but isn't is supposed to work as it was previously ?

laurre:
I have a file '../config/functions.js', in which at the bottom I have:

```
export default {
  function1,
  function2,
  function3
};
```

nicolo-ribaudo:
You can't spread an object into an array[¹]. You should use `var tab = Array.from(Object.values(configFunctions))`. I am surprised that it worked :thinking: 	

[1] You can try this code in any modern browser:
```js
[ ...{ function1() {}, function2() {}, function3() {} } ]
```

laurre:
It works now, thanks, I think it was working before because Babel was to permissive

PR #7498, opened 2018-03-05
Disallow setters to have RestElement

danez:
| Fixed Issues?            | 
| Major: Breaking Change?  | y
| Documentation PR         | 
| Any Dependency Changes?  |

```js
class X {
  set fun(...a) {}
}

const x = {
  set fun(...a) {}
};
```

This is not allowed according to the spec. https://www.ecma-international.org/ecma-262/6.0/#sec-method-definitions

V8/FF/acorn/flow/espree throw on this.

Although this is a breaking change, I think the impact will be minimal, as all other parsers already do the right thing, so people would have run into a problem with eslint/webpack/flow etc before anyway.

Found this while looking at https://kangax.github.io/compat-table/es6/


danez:
I just made more changes as i noticed I missed some things. I renamed the method and also made the change for flow.

PR #7499, opened 2018-03-05
Wmertens add prettier config

danez:
| Fixed Issues?            | Fixes #6860
| Documentation PR         | 
| Any Dependency Changes?  |

Rebased #6860 + updated dependencies + ignoring package.json


Makefile
```diff
@@ -53,10 +53,11 @@ flow:
 	./node_modules/.bin/flow check --strip-root
 
 lint:
-	./node_modules/.bin/eslint scripts $(SOURCES) *.js --format=codeframe --rulesdir="./scripts/eslint_rules"
+	./node_modules/.bin/eslint scripts $(SOURCES) '*.js' '**/.*.js' --format=codeframe --rulesdir="./scripts/eslint_rules"
 
 fix:
-	./node_modules/.bin/eslint scripts $(SOURCES) *.js --format=codeframe --fix --rulesdir="./scripts/eslint_rules"
+	./node_modules/.bin/prettier --config .prettierrc --write --ignore-path .eslintignore '**/*.json'
```
loganfsmyth: Is the config file not found automatically?

danez: It is, but we have invalid package.json files in the repo which prettier wants to format and while searching for the config for this invalid file it searches in the file itself and then crashes because the json is invalid.

loganfsmyth: Gotcha. Maybe worth a comment here so it is clear?

danez: added

danez:
Thank you @wmertens

Makefile
```diff
@@ -53,10 +53,12 @@ flow:
 	./node_modules/.bin/flow check --strip-root
 
 lint:
-	./node_modules/.bin/eslint scripts $(SOURCES) *.js --format=codeframe --rulesdir="./scripts/eslint_rules"
+	./node_modules/.bin/eslint scripts $(SOURCES) '*.js' '**/.*.js' --format=codeframe --rulesdir="./scripts/eslint_rules"
 
 fix:
-	./node_modules/.bin/eslint scripts $(SOURCES) *.js --format=codeframe --fix --rulesdir="./scripts/eslint_rules"
+    # The config is hardcoded because otherwise prettier searches for it and also picks up some broken package.json files from tests
```
Kovensky: I'm surprised this worked 🤔

I assume make is parsing it as a comment at the end of the dependency list.

Makefiles must only ever be indented with tabs to avoid confusion...

danez: Oh didn't know that. will fix it

Kovensky: Yeah, tabs are actually significant in Makefiles; very annoying :(

PR #7500, opened 2018-03-05
Remove the sourceMapTarget option from core and implement it in babel-cli.

loganfsmyth:
| Fixed Issues?            | 
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Ok, background first, the `sourceMapTarget` option was passed straight through `babel-core` into the `file` field of the sourcemap created for the output. The `file` field is optional in sourcemaps, but up until now we've actually had it default to the _input_ filename, but per the spec it should really be the output file name.

The issue is for most cases, you can't know whether the filename should end with .js or .mjs. That leaves us a few options:

1. Allow it to be a function that returns the output filename
2. Make it default to undefined, so users can choose to set it if they want, but otherwise it defaults to empty.
3. Remove it entirely and let the caller set it if they want once they get the sourcemap back.

(1) sounds like a painful API, so I vetoed that.
(2) is an option, but I started to question the necessity for this option to begin with. If a sourcemap is inlined into the output as a comment, including the filename doesn't really make sense. If it _isn't_ inlined into the file, then it is trivial for the tooling that calls Babel to assign `.file` on the map itself afterward, rendering the option useless.

So essentially all that this PR removes is the ability to set the `file` field on inline sourcemap JSON.

If anyone feel that matters, I'd be happy to go with option (2) instead.


packages/babel-cli/src/babel/file.js
```diff
@@ -16,7 +16,10 @@ export default function(commander, filenames, opts) {
 
   const buildResult = function() {
     const map = new sourceMap.SourceMapGenerator({
-      file: path.basename(commander.outFile || "") || "stdout",
+      file:
+        commander.sourceMapTarget ||
```
loganfsmyth: We also had weird cases like this, where if the option was in babel-core, then every file has their own target, making it unclear what the `file` field should be when using babel-cli to concatenate a bunch of files together. By implementing it in CLI itself, we know that the option is meant for the final output file.

This is the only place the CLI option is actually used now, which I think makes sense.

packages/babel-cli/src/babel/index.js
```diff
@@ -259,6 +259,7 @@ delete opts.configFile;
 delete opts.deleteDirOnStart;
 delete opts.keepFileExtension;
 delete opts.relative;
+delete opts.sourceMapTarget;
```
loganfsmyth: There is still a `--source-map-target` CLI option that sets the `file` field above now when concatenating things into a single file.

loganfsmyth: If you're using Babel 7, you absolutely need to use a lockfile. `grunt-babel` has not yet been updated to support Babel 7's beta. For now I would recommend locking your version of Babel and all the plugins to `beta.40`. `beta.41` is not compatible yet.

danez: `grunt-babel@8.0.0-beta.0` was released to support the scoped versions of babel including the latest version.

Issue #7457, opened 2018-02-28
Env 7 / Node 4 useBuiltIns usage transpiling polyfill call with "import" not "require"


reggi:
Here's my config:

```
{
  "babel": {
    "presets": [
      [
        "@babel/preset-env",
        {
          "targets": {
            "node": "4.0.0"
          },
          "useBuiltIns": "usage",
          "debug": true,
          "shippedProposals": true
        }
      ]
    ]
  }
  "devDependencies": {
    "npm-run-all": "^4.1.2",
    "@babel/cli": "^7.0.0-beta.40",
    "@babel/core": "^7.0.0-beta.40",
    "@babel/preset-env": "^7.0.0-beta.40",
    "babel-core": "^7.0.0-bridge.0",
    "babel-jest": "^22.4.1",
    "jest": "^22.4.2"
  }
}
```

Here's debug output:

```

@babel/preset-env: `DEBUG` option

Using targets:
{
  "node": "4"
}

Using modules transform: commonjs

Using plugins:
  transform-arrow-functions { "node":"4" }
  transform-block-scoping { "node":"4" }
@babel/preset-env: `DEBUG` option
  transform-classes { "node":"4" }
  transform-destructuring { "node":"4" }
  transform-for-of { "node":"4" }
  transform-function-name { "node":"4" }
  transform-object-super { "node":"4" }
  transform-parameters { "node":"4" }
  transform-spread { "node":"4" }
  transform-sticky-regex { "node":"4" }
  transform-unicode-regex { "node":"4" }
  transform-new-target { "node":"4" }
  transform-regenerator { "node":"4" }
  transform-exponentiation-operator { "node":"4" }
  transform-async-to-generator { "node":"4" }
  transform-dotall-regex { "node":"4" }
  proposal-async-generator-functions { "node":"4" }

Using targets:
  proposal-object-rest-spread { "node":"4" }
  proposal-optional-catch-binding { "node":"4" }
  proposal-unicode-property-regex { "node":"4" }

Using polyfills with `usage` option:
{
  "node": "4"
}

Using modules transform: commonjs

Using plugins:
  transform-arrow-functions { "node":"4" }
  transform-block-scoping { "node":"4" }
  transform-classes { "node":"4" }
  transform-destructuring { "node":"4" }
  transform-for-of { "node":"4" }
  transform-function-name { "node":"4" }
  transform-object-super { "node":"4" }
  transform-parameters { "node":"4" }
  transform-spread { "node":"4" }
[...]

PR #7501, opened 2018-03-05
Default to sourceType:script when parsing files.

### LINKED ISSUE #6242: Gain consensus on path forward for "sourceType" for parsing modules vs scripts in 7.x
Rather than have this on Slack, I thought it would be useful to have this as an issue so we can refer to it in the future. So, the question:

> Should we switch from `sourceType: "module"` for `.js` files, to `sourceType: "script"`.

What this means in practice is:

* Users using ES6 module syntax will have to rename their files to `.mjs` _or_ put `sourceType: "module"` in their `.babelrc` or whatever Babel config they have.
* Users who were relying on Babel automatically adding `use strict` to their CommonJS files will have to add `transform-strict-mode` to their Babel configs.

Because Babel has parsed as `module` by default for a long time, it has logic that essentially says "if there are no imports/exports in this `module`, it's actually a `script`". This adds a lot of complexity to Babel's module processing because this logic is not standard in any way, and it means that there is no central way for code to ask "is this a module".

The reason "is this a module" is such an important question is that because of this, currently it is valid to have _both_ 

1. ES6 `import` statements _and_ `require()` usage
2. ES6 `export` statements _and_ `module.exports`/`exports.foo` usage

Webpack now disallows the `#2` case, while allowing `#1`. Ideally Babel should also adopt that behavior, but where it's not too bad for Webpack, Babel's iterative transformational nature makes it a pain to constantly not know what type of `sourceType` the file should be considered as.

### My proposal, then is the following:

* Change `sourceType` to default to `script` for `.js`, and force to `module` for `.mjs`
* Disallow use of CommonJS's `module.exports`/`exports` when `sourceType: module` (potentially with a flag to enable this again to ease transition)
* Update the parts of Babel that inject `import` statements to be smart enough to inject both `import` and `require` based on the `.sourceType`

To ease the transition, I'd propose a few more possibilities.
* Land [sourceType: "unambiguous" parsing](https://github.com/babel/babylon/pull/449) so that users with codebases that aggressively mix ES6 modules and CommonJS modules don't pull their hair out. This should essentially make parsing act like it did before, minus the `use strict` injection.
* Fix [Ability to output mjs extension](https://github.com/babel/babel/issues/6222) by having `babel-cli` output files with a `.mjs` extension if they have `sourceType: 'module'`
* Support [Configuration based on file path](https://github.com/babel/babel/issues/5451) so that if people don't want to use `sourceType: 'unambiguous'`, they can at least explicitly opt specific files into a `sourceType`

What do people think?

loganfsmyth:
| Fixed Issues?            | Fixes #6242
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Alright, let's have this discussion since #6242 never really took off, and most of my TODOs in there have landed.

I am of the opinion that Babel 7 should parse `.js` files using the `script` parse goal, and `.mjs` files with the `module` parse goal. In Babel 6, and 7 up to present, the default has been `module`. I also want to make clear that you can still _easily_ set `sourceType:module` in your Babel config to keep the current behavior, but the important question here is the default, because it influences many other parts of Babel's transformations.

There are a few primary issues surrounding using `module` as the default:

* Many perfectly valid Script files will fail to parse. This means that users attempting to run Babel on things in node_modules for instance will fail to parse.
* Defaulting to `module` means Babel converts top-level `this` to `undefined` and injects `use strict`, which can break when run against arbitrary files
* When performing transformations, Babel needs to know whether to inject `import` or `require`, and it does so based on the `sourceType`. Defaulting to `module` means we may inject `import` into valid `CommonJS` modules, breaking them.
* When writing files to disk with `babel-cli` we'd ideally add `.mjs` to files that are modules, and `.js` to scripts. If we default to `module` then we'd potentially output `.mjs` extensions on CommonJS files.

These issues become more problematic when you consider that people are more frequently running Babel across arbitrary `node_modules` dependencies. Realistically this pattern is only going to increase in frequency for forseeable future.

Node now follows this same pattern, with `.js` meaning CommonJS and `.mjs` meaning ESM. There is potential for a field to land in `package.json` that would allow toggling this behavior, and once that lands we could for instance consider a `sourceType: packageMode` option or something, but even with that the default in Node will remain `.js` for CommonJS.

Given all of this, it seems like it would be in the best interest of Babel to follow that same pattern, by requiring users to either use `.mjs` or opt into specific ESM behavior in their Babel configuration.

### DIFF
--- a/packages/babel-node/src/_babel-node.js
+++ b/packages/babel-node/src/_babel-node.js
@@ -39,6 +39,7 @@ program.option(
   "List of extensions to hook into [.es6,.js,.es,.jsx,.mjs]",
   collect,
 );
+program.option("--source-type [script|module|unambiguous]", "");
 program.option("-w, --plugins [string]", "", collect);
 program.option("-b, --presets [string]", "", collect);
 /* eslint-enable max-len */
@@ -53,6 +54,7 @@ register({
   only: program.only,
   plugins: program.plugins,
   presets: program.presets,
+  sourceType: program.sourceType,
 });
 
 const replPlugin = ({ types: t }) => ({
@@ -90,6 +92,7 @@ const _eval = function(code, filename) {
     filename: filename,
     presets: program.presets,
     plugins: (program.plugins || []).concat([replPlugin]),
+    sourceType: program.sourceType,
   }).code;
 
   return vm.runInThisContext(code, {
@@ -131,7 +134,10 @@ if (program.eval || program.print) {
       }
 
       if (arg[0] === "-") {
-        const parsedArg = program[arg.slice(2)];
+        const name = arg.slice(2);
+        const parsedArg =
+          name === "source-type" ? program.sourceType : program[arg.slice(2)];
+
         if (parsedArg && parsedArg !== true) {
           ignoreNext = true;
         }
--- a/packages/babel-core/src/transformation/normalize-opts.js
+++ b/packages/babel-core/src/transformation/normalize-opts.js
@@ -7,7 +7,7 @@ export default function normalizeOptions(config: ResolvedConfig): {} {
   const {
     filename,
     filenameRelative = filename || "unknown",
-    sourceType = "module",
+    sourceType = "script",
     inputSourceMap,
     sourceMaps = !!inputSourceMap,
 
--- a/packages/babel-node/test/index.js
+++ b/packages/babel-node/test/index.js
@@ -123,6 +123,7 @@ const buildTest = function(binName, testName, opts) {
 
     let args = [binLoc];
 
+    args.push("--source-type", "module");
     args.push("--presets", presetLocs, "--plugins", pluginLocs);
     args.push("--only", "../../../../packages/*/test");
 
--- a/packages/babel-plugin-transform-modules-commonjs/src/index.js
+++ b/packages/babel-plugin-transform-modules-commonjs/src/index.js
@@ -20,8 +20,7 @@ export default declare((api, options) => {
     strictMode,
     noInterop,
     lazy = false,
-    // Defaulting to 'true' for now. May change before 7.x major.
-    allowCommonJSExports = true,
+    allowCommonJSExports = false,
   } = options;
 
   if (
--- a/packages/babel-plugin-transform-modules-commonjs/test/copied-nodes.js
+++ b/packages/babel-plugin-transform-modules-commonjs/test/copied-nodes.js
@@ -5,6 +5,7 @@ test("Doesn't use the same object for two different nodes in the AST", function(
   const code = 'import Foo from "bar"; Foo; Foo;';
 
   const ast = babel.transform(code, {
+    sourceType: "module",
     ast: true,
     plugins: [[require("../"), { loose: true }]],
   }).ast;
--- a/packages/babel-plugin-transform-modules-commonjs/test/esmodule-flag.js
+++ b/packages/babel-plugin-transform-modules-commonjs/test/esmodule-flag.js
@@ -20,6 +20,7 @@ test("Re-export doesn't overwrite __esModule flag", function() {
   context.exports = context.module.exports;
 
   code = babel.transform(code, {
+    sourceType: "module",
     plugins: [[require("../"), { loose: true }]],
     ast: false,
   }).code;
--- a/packages/babel-plugin-transform-modules-commonjs/test/fixtures/source-map/exec.js
+++ b/packages/babel-plugin-transform-modules-commonjs/test/fixtures/source-map/exec.js
@@ -23,6 +23,7 @@ var tests = [
 
 tests.forEach(function (code) {
   var res = transform(code, {
+    sourceType: "module",
     sourceMap: true,
     plugins: opts.plugins
   });
--- a/packages/babel-preset-env/test/debug-fixtures.js
+++ b/packages/babel-preset-env/test/debug-fixtures.js
@@ -51,7 +51,7 @@ const buildTest = opts => {
     saveInFiles(opts.inFiles);
 
     let args = [binLoc];
-    args = args.concat(opts.args);
+    args = args.concat(opts.args, "--source-type", "module");
 
     const spawn = child.spawn(process.execPath, args);
 
--- a/packages/babel-register/test/index.js
+++ b/packages/babel-register/test/index.js
@@ -105,6 +105,7 @@ describe("@babel/register", function() {
 
   test("hook transpiles with config", () => {
     setupRegister({
+      sourceType: "module",
       babelrc: false,
       sourceMaps: false,
       plugins: ["@babel/transform-modules-commonjs"],
@@ -117,6 +118,7 @@ describe("@babel/register", function() {
 
   test("hook transpiles with babelrc", () => {
     setupRegister({
+      sourceType: "module",
       babelrc: true,
       sourceMaps: false,
     });
--- a/packages/babel-standalone/src/transformScriptTags.js
+++ b/packages/babel-standalone/src/transformScriptTags.js
@@ -134,6 +134,7 @@ function loadScripts(transformFn, scripts) {
       executed: false,
       plugins: getPluginsOrPresetsFromScript(script, "data-plugins"),
       presets: getPluginsOrPresetsFromScript(script, "data-presets"),
+      sourceType: script.getAttribute("data-source-type") || undefined,
     };
 
     if (script.src) {
--- a/packages/babel-standalone/test/babel.js
+++ b/packages/babel-standalone/test/babel.js
@@ -78,6 +78,7 @@ const assert = require("assert");
 
     it("handles presets with options", () => {
       const output = Babel.transform("export let x", {
+        sourceType: "module",
         presets: [["es2015", { modules: false }]],
       }).code;
       assert.equal(output, "export var x;");


packages/babel-node/src/_babel-node.js
```diff
@@ -39,6 +39,7 @@ program.option(
   "List of extensions to hook into [.es6,.js,.es,.jsx,.mjs]",
   collect,
 );
+program.option("--source-type [script|module|unambiguous]", "");
```
ljharb: can this be made to assume "module" for `.mjs` files?

loganfsmyth: `.mjs` files are hard-coded to `module` currently, so this would only effect other types.

ljharb: Thanks, makes sense.

loganfsmyth:
Also curious to hear from @gaearon if you have thoughts. This would certainly affect CRA though realistically you may _already_ just want to set `sourceType: unambiguous` since that is what Webpack does.

hulkish:
If this lands as currently proposed, I suggest reporting a warning for when `sourceType` is not defined in `.babelrc`. 

Or, we could just keep `scriptType: module` as default and warn for the opposite when `.mjs` extension is detected?

loganfsmyth:
Oh that's something I can clarify. The error right now is quite helpful. If Babel is parsing as `script` and hits an `import` or `export`, the message is:

> SyntaxError: the-file-path.js: 'import' and 'export' may appear only with 'sourceType: "module"'
> Consider renaming the file to '.mjs', or setting sourceType:module or sourceType:unambiguous in your Babel config for this file.

Which I think is helps. We can certainly expand that too if we find it is needed.

loganfsmyth:
@jdalton I know this is something you have thoughts on. Where do you stand on this these days?

Kovensky:
How are TypeScript files going to be handled?

I _assume_ it's safe to hardcode those to modules too.

loganfsmyth:
@Kovensky Fair, I'd be fine with hard-coding `.ts` and `.tsx` to be `sourceType:module` too. It would be a question of if that would ever need overriding, and I hon't know the answer to that.

Kovensky:
It would need overriding if babel supports `import Identifier =` and `export =` syntax. The only way for an `--isolatedModules` typescript file to not be a `module` is if they only have assign-style import/exports.

loganfsmyth:
@Kovensky How does TS itself differentiate these cases? I assume there is a flag in their config file?

Kovensky:
TS specifies itself as unambiguous grammar. `--isolatedModules` (which is a prerequisite of the babel transform) makes it so that no global scripts are allowed; every file must be either a commonjs (with `import Identifier =` or `export =` syntax) or ES-like module, but they're still differentiated based on whether ES `import` or `export` are present 😞

jdalton:
\cc @weswigham @DanielRosenwasser

loganfsmyth:
Maybe we should consider having Babylon just ignore the `sourceType` open when the `typescript` Babylon plugin is enabled, and have it handle the unambiguous logic itself? Then we could make the necessary changes to babel-core to allow Babel's typescript plugin to automatically only activate itself when the file extension is `.ts`/`.tsx`? We've talked about that type of toggling already so that you could safely enable the Flow and Typescript in the same project without them conflicting or requiring the users to use `overrides` themselves.

weswigham:
In the browser, `.js` is king (and the default), and in node, tons of effort is (actively) being poured into making `.js` as close to the default as possible (mode flags, loaders, etc - and still not finalized). I believe it's premature to call `.mjs` acceptable as the default mode of operation for any transpiler (I know we're holding off on it at typescript because 1. we really dislike `.mjs` since it'll cause our resolver complexity to bloom considerably and seriously damage the ecosystem, and 2. there's the possibility that its usage ends up really low and isn't worth supporting if it's not implemented well, given the large existing cjs ecosystem) prior to any implementation that actually uses `.mjs` shipping without an `experimental` flag. Besides, in the unlikely event that someone _actually_ loads things they didn't write (ie, from `node_modules`) _and_ had problems loading some files, they could override the default parse goal at that point, no? Why would you want the default mode to be the old unfancy cjs-y mode? I don't think I've ever seen or used a `babel`-using project that just transplied things inside commonjs modules or script files, myself. Babel's tagline used to be along the lines of "The JavaScript of tomorrow, today!", no? Is ESM no longer an accepted part of "tomorrow"?

Y'all should be tracking the [node modules WG](https://github.com/nodejs/modules) if you're not already, and use the progress you see there help inform where you invest your time right now, but at least on our end we're planning on sitting on it until `node`'s plan is more complete. Like, if `"mode": "esm"` merges and stays in, it'd be silly not to continue to consider `js` as a `module` by default, since both old users would expect it, and most new users would.

ljharb:
No mater what ships in node, CJS will remain the default, and babel will absolutely need to support mjs when it ships unflagged in node.

I often use babel without using ESM; converting syntax is nice but having to remember not to pollute the ecosystem with a .default is less so.

weswigham:
@ljharb What about all the browser people who just wanna transpile their async/await down to normal esm modules for browser use by default? Node's default isn't the browser's default. Your default isn't theirs.

bmeck:
@weswigham even with `"mode": "esm"` the default is CJS and `.mjs` would be supported. I'm not sure I understand your statement that `.js` should be a `module` by default in that situation. `babel` would still need to read the `package.json` field to find the mode which would be CJS by default and  still support `.mjs`.

ljharb:
@weswigham the browser’s default is Script, unless you specify type=module. I agree with you - Babel’s default should match the browser’s.

weswigham:
@ljharb And that file is a `.js` file.

I'm saying that without knowing if the code is targeted for `node` or a browser, you can't actually know if a `.js` file is intended as a `module` or a `script` without inspecting it and making assumptions.

Changing the default behavior here just changes the default platform assumption.

IMO, the current default is _usable_ and it is _configurable_ so that a given audience can achieve their desired behavior, anyway; so why would you sacrifice that usability? Like, webpack's last release's push was for zero-config starts for webdev projects; this pushes babel in the opposite direction for web projects.

ljharb:
No, browsers ignore the extension - there’s no such thing as dot js in the browser. Browsers use MIME types.

Babel’s output is always targeted for a server environment first - that’s how it’ll end up in a browser - and servers need a way to tell the browser what type of script tag to use.

weswigham:
@ljharb That's how the implementation works, but that's not how authors think about it.

bmeck:
@weswigham that ambiguity isn't solved by inspecting the file itself. You need out of band data. This is not about zero-config, the 3 different modes of JS all occupying a single `.js` file extension cannot be disambiguated. We tried to disambiguate them at TC39 with the Unambiguous Grammar proposals, "use module", etc. but web browsers saw the disambiguation within a source text as unacceptable.

This would make a default but keep it configurable. Node cannot change their default behavior and relies on file extensions. Browsers cannot change their behavior but are not using file extensions. I think it is a reasonable decision to make the default fit with Node.js in this situation, since it is not actively doing any problematic behavior to browsers. The idea of defaulting to `module` is however problematic behavior for Node.js which cannot change the default and requires out of band configuration like `"mode"` / `.mjs` / etc.

ljharb:
Unfortunately, the language having two parse goals means that authors will likely need to start thinking about things differently (perhaps, correctly?).

I’m not stoked about the increase in complexity, but that ship sailed with the advent of ESM many years ago.

weswigham:
> The idea of defaulting to module is however problematic behavior

The last few years seem to indicate otherwise. 

I'm just saying that transpilers continuing to use anambiguous-type behaviors (since, honestly, we're more flexible than runtimes) is fine. A transpiler's _job_ is to paper over runtime differences, and this is one of them. If it prevents me form needing to think about all the nonsense going on in the esm-extension-world 95% of the time, than it's done it's job of hiding platform differences.

Kovensky:
The way everyone implemented modules **is wrong**. There is _nothing_ relevant at all about the `.js` extension. Knowing whether the input is a module or a script _is_ relevant as, even for things completely unrelated to whether you can write `import` or `export` keywords, modules have stricter syntax than strict mode scripts.

I've said it back when you implementd `--esModuleInterop` in typescript, and you still insist in pushing that the wrong way people have been doing it for years because of mistakes done in Babel 5 and Babel 6 is the way that should be kept going forward.

bmeck:
@weswigham 

I think there is a very clear point that we need to bring up here. I want to focus on 

> A transpiler's job is to paper over runtime differences

Unfortunately, modes are not purely a runtime difference. This is about parsing and loading semantics. It does have runtime differences though with how syntax works but I'd still lean onto that being related to the parsing goal considering how it functions in the JS spec and Node's CJS implementation.

The ability to preprocess files leads to almost infinite ability to say that something is not a problem, because you can preprocess a file outside of the runtime and do any sort of transformation on the source text. I equate arguments using the logic of preprocessing into a valid output from ambiguous input as dangerous. It attempts to separate the logic of the input from the output in a non-trivial way. In particular it seeks to treat the output as the only important part of the tooling. If the input of a tool is always invalid and requires preprocessing I would not state something as being the same language as the output. Continuing down this logic I would be very hesitant to even call CommonJS a JS dialect since it always requires some preprocessing logic such as using a bundler in order to run in a JS environment per specification.

I would be far more swayed by some argument that I don't see as trying to diverge how various platforms work at runtime and what input should be fed into preprocessing tools like `babel`.

weswigham:
> The way everyone implemented modules is wrong

Consider, for a moment, that you're not an almighty all-knowing specialist. You've read about this hip new tool called _babel_ (or _tsc_). You've got a token personal webpage setup with all the jquery you could ever want, but a friend recently tipped you off to the wonders of the `npm` ecosystem. You write add your `.js` file, and you include the `dist` folder version it via a `script module` tag like that blog post said to. Then you run `babel`. What do you, as the new programmer, expect?

1. The tool to work flawlessly and put a file in the output, so I can look at my buttons.
2. The tool to break and complain that I didn't pass in an arcane configuration spell to specify that I'm targeting browsers and not `node`.

This argument can then be repeated for a fledgling `node` developer. The only way to satisfy _both_ scenarios is to have an unambiguous mode as the default.

The spec is an ideal, and tools like `babel` and `tsc` are around to map people's expectations to that ideal, not to _be_ that ideal. 

The _default_ is really that there _is no default_, so a transpiler choosing whatever gives the best UX is within very much reason.

> If the input of a tool is always invalid and requires preprocessing I would not state something as being the same language as the output

"Invalid" and "valid with assumptions" and "invalid with assumptions" are meaningfully different.

bmeck:
> "Invalid" and "valid with assumptions" and "invalid with assumptions" are meaningfully different.

I would agree, but still see both as problematic since "valid with assumptions" means invalid with the wrong ones?

Kovensky:
That is an argument _against_ tools like Babel because it causes people to think the language work a certain way when it doesn't.

weswigham:
> I would agree, but still see both as problematic since "valid with assumptions" means invalid with the wrong ones?

Under some edge cases, maybe. Tons of code can execute happily in both `module` and `script` contexts, though. And yet more can be unambiguously determined to only execute correctly in just one of them (we know this from the history the last few years has taught us). The overlap where they can both be "correct" but behave differently is so small that most real people (ie, not people working on js transpilers or runtimes) just never notice the differences (and those who do notice even still have an out - they can configure for what they need). Removing configuration for _the most users_ is, IMO, should always be a priority.

But hey, if babel sees itself moving to a library-type tool, and letting someone else take over the `cli` and new dev ux story, go for it (I'm sure `webpack` will happily assimilate the role like they assimilate anything else they like). After all, anyone can always just layer saner defaults on top of insane ones and redistribute it, right? Why even let UX be your job when you can punt it to someone else?

ljharb:
Then it sounds like perhaps there should be no default, so it throws when nothing is configured (with a helpful message), so that fledgling developers can learn how things actually work?

Good UX also includes teaching users what they need to know when it can be hidden from them completely - the difference between Script and Module can not, and more importantly should not, be hidden.

weswigham:
> Then it sounds like perhaps there should be no default, so it throws when nothing is configured (with a helpful message), so that fledgling developers can learn how things actually work?

TBQH, I remember a (recent!) era of "it just works" mantras and am sad to see it go to the wayside.

bmeck:
> Under some edge cases, maybe. Tons of code can execute happily in both module and script contexts, though. And yet more can be unambiguously determined to only execute correctly in just one of them (we know this from the history the last few years has taught us). The overlap where they can both be "correct" but behave differently is so small that most real people (ie, not people working on js transpilers or runtimes) just never notice the differences (and those who do notice even still have an out - they can configure for what they need). Removing configuration for the most users, IMO, should always be a priority.

This is somewhat true, but it doesn't consider that the grammars can become increasingly ambiguous and Script can gain features that make it look like a Module. The most glaring example of this is `import()` allowing Script to `import` other Module Records (not necessarily ESM). If things like the multiple-module proposal require a new mode, and REPL as well we also need to contend with an increasing difficulty in how similar these 4 grammars may be. Removing configuration seems fine, but I would just like to be sure that the defaults don't cause conflicts with the platform expectations.

weswigham:
> I would just like to be sure that the defaults don't cause conflicts with the platform expectations

The requires assuming a platform in the first place 😄 Each platform already has conflicting expectations in the absence of complete side-channel data 😄 

> Good UX also includes teaching users what they need to know when it can be hidden from them completely - the difference between Script and Module can not, and more importantly should not, be hidden.

In the extreme I agree, that a strict mode where you can opt in to being on your best behavior can be nice (occasionally it can find a problem in a large project) - and that definitely exists, but broader usability by default with stricter opt-ins is my stance. Like, really, if I was using `babel-preset-env`, and I said my targets were both `chrome` and a hypothetical `node 11`, what even is, reasonably, the expected input for a `js` file (seriously, go poll some people and be _amazed_ at how ingrained the current status quo is)? (Or the output for that matter, ffs - are you expecting borderline crazy things like `node 11-mode/esm` or something) Copping out and saying "we don't know unless you tell us" is bad, and swaying to either platform's preference is to the detriment of the other.

loganfsmyth:
> I don't think I've ever seen or used a babel-using project that just transplied things inside commonjs modules or script files, myself.

This is happening more and more as packages in the npm ecosystem are published to support specific Node versions. ES6 within node packages is quite common and will only become more-so.

> The spec is an ideal, and tools like babel and tsc are around to map people's expectations to that ideal, not to be that ideal.

This may be the central disagreement from my standpoint. If user can be given clear direction saying "hey we expect X and Y" to explain what they've misunderstood, that seems like the most appropriate approach. The ECMA spec has clearly defined separate grammars, and given that, it isn't something that should be hidden away. I respect that you disagree, but to me it seems like treating your users like they are unable to learn.

I could see that it could be confusing for users if it was something they'd run into and get stuck on, but given [a clear error message](https://github.com/babel/babel/pull/7501#issuecomment-370607245), I have a hard time following the reasoning behind trying to hide it from users entirely.

> TBQH, I remember a (recent!) era of "it just works" mantras

The entire goal of this change, from my standpoint, is to _allow_ things to "just work" in the Node/`npm` ecosystem. Automatically injecting `use strict` into any and all files given to Babel usually breaks things. Given that, default to `module` is absolutely the wrong approach. That leaves us with either `unambiguous`, or `script`.

From my standpoint, `unambiguous` is immediately a no-go as the default because it would be inconsistent Node. It's there, and if people want to use it, they certainly can, but it seems like you'd just be lying to users, which seems counterproductive.

> I'm saying that without knowing if the code is targeted for node or a browser, you can't actually know if a .js file is intended as a module or a script without inspecting it and making assumptions.

Babel's most common usecase is either pre-building an npm package before publishing, or compiling an application during bundling in Webpack or some other bundler. Given that, iterop with the node/npm ecosystem is the most important factor for us.

I agree with your point here that you can't know, but assuming `script` and expecting users to tell us otherwise, is the safest option in that case, not guessing.

> But hey, if babel sees itself moving to a library-type tool, and letting someone else take over the cli and new dev ux story, go for it (I'm sure webpack will happily assimilate the role like they assimilate anything else they like). After all, anyone can always just layer saner defaults on top of insane ones and redistribute it, right? Why even let UX be your job when you can punt it to someone else?

At the end of the day, we can always change again in the future. I'd much rather choose the safely of well-defined behavior that can be clearly messaged to users than tell them that we've decided to guess randomly and hope we chose the right type of file to consider their code.

> The requires assuming a platform in the first place 😄 Each platform already has conflicting expectations in the absence of complete side-channel data 😄

I feel like it is relatively clear from current use in the community, that Babel is used 9 times out of 10 to compile things in the node/npm/webpack ecosystem. We can always reevaluate as things progress over time, but it seems like `script` is the safest default for users in that case.

hulkish:
Could solve this by having no defaults and require users to use maybe `babel-preset-preferred-defaults` which then covers the particular "common preference"

weswigham:
> I respect that you disagree, but to me it seems like treating your users like they are unable to learn.

Even if it's ultimately part if what keeps me employed, I wholeheartedly wish that I never had to learn any of the nonsense around this module madness in the JS ecosystem. Anecdotally, so many friends refuse to _touch_ js dev because of the crazy tooling environment (that I contribute to so shouldn't really talk).

But alright, if the core coceit is that `webpack` has taken over all the transpiler ux for browser-focused people, I don't see why you don't make the default an error, then, since it's the only really correct thing to assume without any explicit data. Since making script the default is probably going to confuse someone down the line, and has the potential to cause subtle breaks on upgrade.

Kovensky:
This module madness is our fault as early incorrect implementers. I believe we should take steps to correct it, not perpetuate it.

Also, `script` is only the default for `.js` files, exactly the same way `node --experimental-modules` sees them. `.mjs` files will be `module`.

loganfsmyth:
> I don't see why you don't make the default an error, then, since it's the only really correct thing to assume without any explicit data.

I don't understand this point. Our options as far as I understand your argument are

1. Default to `module`, and which will trigger random runtime errors when users process a file that isn't a module.
2. Default to `unambiguous`, which will lead to modules sometimes silently not actually running as modules.
3. Default to` script`, which throws a clear helpful message for users about what steps they should take
4. Default to nothing and require the user to choose, showing a helpful error message.

Both options 3 and 4 will end up requiring the user to set a flag in their config if they want a `.js` file to be treated as a module and don't want to use `.mjs`, and both can provide clear descriptions of why the error is occuring. Given that, why would it be desirable to error in all cases instead of passing on the common non-module case, and giving a clear message in the module case?

> Since making script the default is probably going to confuse someone down the line, and has the potential to cause subtle breaks on upgrade.

I'm willing to accept confusion for users as long as we have a way to direct them to the information they need to fix their config. I don't see how defaulting to `script`, at this moment, has more of a chance to cause breakage than `unambiguous` would. Defaulting to `module` will constantly break our users as compilation of `node_modules` becomes more common and it will happen at runtime, leaving them with unhelpful error messages and no clear recourse on how to resolve their issues.

appsforartists:
So what happens if library authors continue to decide that `.mjs` is gross and publish ESM in a `.js` module? (I expect there are presently more modules published as `.js` than as `.mjs`.)  If you wanted to presume that everything that ended in `.js` on npm was in CommonJS, you probably should have done that about 4 years ago.  You'll have the same problems presuming all `.js` is CommonJS as you would presuming all `.js` is a module: there will be libraries that use `.js` and defy your presumption.

I know it sounds irrational to some, but the `.js` extension is ingrained in both the identity and habit of many JavaScript developers who see `.mjs` both ugly and [lazy](https://github.com/dherman/defense-of-dot-js).  Those who don't use `node` for their own projects may not care how `node` decides to parse modules, and will continue to ship modules as `.js` files.

loganfsmyth:
@appsforartists Great question. Honestly, if that happens I'd expect Babel to adapt and adjust over time. I'm not talking about permanent decisions that we can never be adapted as new information comes up. My goal with this PR specifically is to make Babel less likely to silently cause breakage for users given the present state of the Node/npm-packaged community. 

> So what happens if library authors continue to decide that .mjs is gross and publish ESM in a .js module?

At this point it seems _extremely_ unlikely that Node will get to a place any time soon where `.js` is the default for ESM. I won't say never because who knows. If that's the case, and packages are being published as `.js` for ESM, there's 2 ways it seems they could have gotten there:

* They are simply publishing a module that is incompatible with Node entirely.
* They are publishing a module that uses a flag like [the one being planned for `package.json`](https://github.com/nodejs/node/pull/18392) which tells node to treat `.js` files as ESM.

Given that, if Node lands that flag I'd likely be for making _that_ behavior the default for Babel instead of `script`, which would address your concern I think?

> You'll have the same problems presuming all .js is CommonJS as you would presuming all .js is a module: there will be libraries that use .js and defy your presumption.

Do you see these libraries as also being incompatible with Node? That may be true in the short term, but I'm not convinced that will be the case once Node ships non-experimental modules.

> Those who don't use node for their own projects may not care how node decides to parse modules, and will continue to ship modules as .js files.

That's totally fair and I'm 100% behind making that as easy path too, which is setting `sourceType: "module"` is still super easy. I'm not interested in preventing that. What I am interested in is making sure Babel is able to participate _safely_ in the Node/npm-packaged ecosystem as it stands today. 

By indiscriminantly defaulting to `module` for `.js` files, we silently break packages for anyone who tries to compile `node_modules` with Babel (which is an increasingly-important thing that people are doing) and we encourage people to write packages that will be incompatible with Node. Many people see Babel's role in the community as a tool to transform code until you don't need it anymore. By encouraging the use of `.js` for _all_ modules, not just when opted into somewhere, we're shirking that responsibility by locking people into a file extension that will not work without Babel's intervention. To me that seems like a dangerous precedent.

not-an-aardvark:
From the peanut gallery: The following may not be feasible due to existing code and strongly-held opinions about file extensions, but I think the ideal scenario would be for babel to parse `.mjs` as a module and `.js` as a script, with no option to switch.

One of the reasons that newcomers to the JS ecosystem sometimes feel overwhelmed is that there are many different mutually-incompatible ways that projects write JS-like code (scripts, modules, JSX, Vue, etc.) and every tool generally needs to be configured separately to work with a project's choice of syntax. This leads to a large amount of configuration complexity. Of course, there are advantages of using different JS variants and extensions, and so in many cases the configuration complexity is worthwhile.

But I think this is a case where the tooling ecosystem can eliminate that complexity for everyone, by standardizing on `.mjs` as the *only* way to distinguish scripts and module files, without needing to independently configure the source type for every tool. If we can accomplish that, we can eliminate a bit of complexity from a huge number of future projects and config files.

appsforartists:
@loganfsmyth I hear you, and I appreciate the thoughtful response.  

To be clear, I'm speaking about library authors (not just app developers).  Imagine that someone cares more about using a familiar file extension than being importable by node.  (Like I said, not necessarily rational, but it seems to be a common opinion.)  That person publishes a module to npm.  Another person installs that module in a project that uses babel.  The same failure mode you're concerned about (making the wrong presumption about file format and using the wrong parse mode) can occur no matter which one you default to - you can't know which parse mode file expects just by checking the extension.

Moreover, Node and `<script type="module">` are already slightly incompatible because the browser uses relative paths and complete file names; whereas Node's resolution algorithm appends suffixes and indices.  This is another point of friction that may incite authors to follow browser conventions rather than node ones in their published modules.

Thanks for sharing that node PR - glad to see package.json is being considered again.  It seems like the most robust (and least incendiary) proposal, but I was afraid it died when the [In Defense of .JS](https://github.com/dherman/defense-of-dot-js) folks gave up.

loganfsmyth:
> The same failure mode you're concerned about (making the wrong presumption about file format and using the wrong parse mode) can occur no matter which one you default to - you can't know which parse mode file expects just by checking the extension.

You're absolutely right there. Both approaches have failure modes that cause issues. To me, it seems that the failure mode for defaulting to `module` is the one that is more likely to cause problems for Babel's users though.

1. If this .js file had an `import` or `export`, we can give a clear error to the user telling them what they need to do to tell Babel that something has gone wrong at build time.
2. If it doesn't `import` or `export`, then the runtime difference is essentially the lack of strict mode for that file.

Given a quick overview of the compatibility

|                               | Goal:Script         | Goal:Module        |
| -----------                   | ---                 | ----               |
| Is ESM with import/export     | Clear error         |  Success           |
| Is ESM without import/export  | Silently non-strict |  Success           |
| Is Script                     | Success             |  Silently strict   |

Silent strict is much more likely to cause breakage when doing broad codebase transformations.

> Moreover, Node and <script type="module"> are already slightly incompatible because the browser uses relative paths and complete file names; whereas Node's resolution algorithm appends suffixes and indices. This is another point of friction that may incite authors to follow browser conventions rather than node ones in their published modules.

Agreed there. I think it will take time for the community to settle on an approach for that. I certainly don't have an answer.

Kovensky:
There are also some identifiers that, even if they're not keywords, are only reserved inside modules; the easiest example is `await`. `"use strict"; let await;` is a valid script but invalid module.

DanielRosenwasser:
I guess I'm just really confused over this. Most of your users have come to expect `module` as the default behavior. Node.js *might* pick a separate file-extension because historically `.js` has implied CommonJS for lack of a better alternative; but Babel already picked "module" as its default parse goal.

What are the compelling scenarios you want to support by switching to `script` as the default? Especially given that this is something which Babel users can already ostensibly opt into. You could even imagine emitting from `.js` to `.mjs` for users who care to.

A lot of people in this thread have banged their heads against the JS module predicament for a while now, so I understand there's a desire to just pick a common strategy to simplify things. If that's the motivation here, I think you'd be better off waiting for Node's Modules Working Group to figure out if the behavior you're anticipating is the one that gets settled on.

weswigham:
> You're absolutely right there. Both approaches have failure modes that cause issues. To me, it seems that the failure mode for defaulting to module is the one that is more likely to cause problems for Babel's users though.

Is it? By your own table, the column on the right had two "Success" entries and one maybe, while the column on the left is one "Success", one fail, and _still_ a maybe. And the assertion that "accidentally loose" is inherently better than "accidentally strict" seems like a tenuous one...? Especially since the more code is written with modules (which is what people here are gunning for), the less and less it's true.

weswigham:
Why change the default for an unstable feature that's in a state of flux, only to change it again later. You could argue that implementing too early is exactly what's fragmented the module ecosystem to where it is today.

Andarist:
Just chiming in - if we go with defaulting to "script", I think it's in best interest of the community to warn when using the default option, therefore "force" people to use pass value for this option explicitly.

Many people still misconfigure their webpack builds by transpiling without `modules: false` simply because there is no warning - i.e. rollup has made way better job by trying to detect this and warn people about misconfiguration.

Also I strongly believe that this approach is just saner for both us, maintainers, and the community. By educating people fast about the need to chose we should end up with way less issues from surprised users that got unexpected results.

wmhilton:
I'm for this. I've only held off on using .mjs file extensions because I was afraid frontend tooling wouldn't support it. If Babel 7 establishes that ".js" is for scripts by default and ".mjs" is for modules, that will make it much easier going forward.

iarna:
I see other folks from the Node.js modules committee here saying that mjs will be inevitable and that is super not-good-faith imo. The reason that committee exists is due to conflict over this issue. I for one am convinced that while we may end up with mjs allowed, we will not end up with it mandated. 

Let's not litigate the issue here, that's the whole point of the committee. Nonetheless, making this change in Babel is extremely premature. To say again: **There's no reason to think that it is where Node.js will land at this point.**

For anyone who says that "but mjs was the consensus" you should be able to see how little consensus you actually had, given the amount of push back there still is every time it comes up.

So please, don't preumaturely and possibly pointlessly break your users.

bmeck:
@iarna We are talking about existing implementations and proposals on those. While I agree that we may not end up in a given direction we also have some things that are outside of the Modules group in Node. In particular with https://tools.ietf.org/html/draft-ietf-dispatch-javascript-mjs-01 queued up for vote soon that would make `.mjs` part of the MIME for all JS.

iarna:
@bmeck That does not seem relevant to the discussion. It certainly does not relate to Babel needing to mandate an extension change.

ljharb:
@iarna Babel isn't mandating anything of the sort; all it's doing is setting a default of "sourceType" to "script". You can change it to "module" and it will work fine with any file extension (except mjs, which is hardcoded to ESM, apparently)

iarna:
@ljharb Escape hatches don't make it less a mandate.

bmeck:
@iarna it was relevant to your statement about discussing `.mjs` being something that you see as bad faith actions on members of Module working group. I think it is very hard to make any progress if we cannot discuss ideas about proposals nor if we can reference things such as the MIME update when discussing such ideas. I have little desire to halt all progress as I have stated before in other areas, but I am open to reverting changes or even removing implementations. In fact, if halting progress is desired I generally prefer removing implementation since it basically means we need to start from scratch. Introducing a level playing field of proposals that do not account for existing usage of implementations/proposals while existing implementation/proposals are still in use and not halted seems impossible to me without removing those implementations from the ecosystem. I much prefer to keep existing implementations and to iterate personally.

@ljharb `.mjs` is required to be Module per the MIME update in IESG/IANA so under that flag it must be "module". That hasn't gone through yet, but there hasn't been any objections so far and it is mostly pending a vote.

iarna:
@bmeck What's in bad faith is talking like this is under discussion in the committee and then going out and acting like its settled elsewhere. I'm _well aware_ of what your preferences are, believe me. =p

bmeck:
@iarna my intention with all of the above comments was framed against the existing PR for `"mode"` in Node core. I do not see it as bad faith to continue discussions about proposals and their intricacies nor have I stated that things are frozen here. If such comments need revision above I am happy to do so. I do think we disagree on what is appropriate or in bad faith to discuss in public though. We could talk to the Modules group about a gag on topics at next meeting perhaps if you think we need to avoid discussing anything outside of that group?

weswigham:
@loganfsmyth Hopefully it should be at least slightly obvious by the discussion here why at least our team doesn't consider the current implementation stable enough to build on yet. As much as something has "shipped" (under a flag with a guarantee to change), it is also guaranteed not to be final. It's excellent that babel _can_ support the current flag already, but it's definitely not ready for default-behavior-informing spotlights (it doesn't even have a _candidate_ to ship unflagged yet), especially when the change _will_ be breaky.

ljharb:
This thread is full of hyperbole and emotion, and I think we've all gotten very far off topic.

Here are the salient points:
 1. whatever mechanism(s) the node modules working group ends up deciding on to indicate that a file is a Module, *babel will support it*. Here ends the relevance/involvement of the node modules working group to this PR; there's no "bad faith" as long as Modules Working Group members aren't arguing that consensus opinions *not* be supported. Nothing node or the Modules WG decides, will change the default sourceType *in the absence of a disambiguation mechanism* for babel.
 1. Since "`.mjs` is a Module" is behind a flag in node, perhaps babel indeed shouldn't support that yet - that's off topic for this PR, but if you believe that babel shouldn't support `.mjs` at all, please do file a separate issue about that!
 1. The default for `.js` files - in browsers, and in node - has always been, is, and probably always will be, Script. Module requires `type="module"` in browsers, `.mjs` in node with the experimental flag, potentially a "mode" flag in package.json - but it always requires an explicit opt in. As such, defaulting to "module" is incorrect, and has always been incorrect, and this change fixes that error for babel. In no way does it block any individual project's ability to determine sourceType for itself.

If you'd like to see the default for `.js` files change, the route to doing that is to convince browser representatives to allow unambiguous parsing to proceed in TC39 - barring that, while "unambiguous" is a fine option, it's an incorrect default.

weswigham:
> The default for .js files - in browsers, and in node - has always been, is, and probably always will be, Script.

False. The "default" is contextual by entrypoint. If the file is `import`'ed, or `script module`'d, it is a module, otherwise if it is included via a script tag it is a script. The browser sees no ambiguous scenarios as it always has sidechannel information to determine correct behavior as far as it is concerned - there is no "default". The default for an unconfigured transpiler is about what is correct when there is _no_ entrypoint and _no_ side channel data from which to know what the goal is, which there is no correct answer for. Saying "script should be default because script tags are like a default" is as much opinion as "module should be default because imports are like the modern default". Hence why in the absence of actual input you should either be lenient and do work that is likely to be the same for both or be strict and demand more information.

loganfsmyth:
Before I dive in more, I really do want to thank everyone for participating in this discussion so far. I know this is a tough subject that a lot of people have strong feelings about, and I appreciate that we've kept things level-headed in the face of that.

loganfsmyth:
@DanielRosenwasser

> Most of your users have come to expect module as the default behavior. Node.js might pick a separate file-extension because historically .js has implied CommonJS for lack of a better alternative; but Babel already picked "module" as its default parse goal.

I think that characterization of how users think about this is inaccurate, because realistically must users don't think about it in the first place until it bites them, since the majority of code will parse with either goal. Eventually they'll get a random runtime error that is almost certainy non-obvious to them. To me that speaks to a problem with defaulting to module and making code auto-strict. If something does eventually break, it does so in a way that is unhelpful for users and doesn't provide them with anything beyond google to figure out their issues, and generally what they'd be googling for would be case-specific and thus hard to find results for.


> What are the compelling scenarios you want to support by switching to script as the default?

My primary concern here is that compiling `node_modules` is becoming more and more common as more packages are published to the npm registry with ES6 features. By defaulting to `module`, this frequently means that users end up injecting `use strict` into random third-party code, or even just getting syntax errors about unexpected keyword, most of which is super hard to understand for people who aren't super knowledgeable.

By encouraging the use of `.js` for both Script and Module, Babel simply has no way to disambiguate within the guidelines of the ECMA spec because there is no additional side channel beyond our config file and the filename. Because of that, it means we'd have to either leave it to users to decide on an individual basis, which of their dependencies should be parsed with which goal, which seems like a recipe for disaster, or encourage people to use `unambiguous`, which isn't a part of the spec at at this point seems very unlikely to be.

By encouraging the use of `.mjs` and `.js`, we'd be helping the ecosystem have a common source of side-channel data for parsing goal, which seems like a big win to me. Like I've mentioned, I'm also happy to expand to allow other side channels as they arise, but the fact of the matter is that we need _something_, and `.mjs` is the best something we currently have. I'm 100% open to that evolving over time though.


> A lot of people in this thread have banged their heads against the JS module predicament for a while now, so I understand there's a desire to just pick a common strategy to simplify things. If that's the motivation here, I think you'd be better off waiting for Node's Modules Working Group to figure out if the behavior you're anticipating is the one that gets settled on.

I'll fully admit to not following everything happening in that group. I guess my big question there is, where can I look to see the various options that seem likely? From what I've followed from the outside anyway, it's seemed very much like no matter what, `.mjs` will at least by _an_ option, whether or not it's the recommended approach. Given that, it seems like it's a safe default in the short term and leaves us open to evolving more as the community does.

loganfsmyth:
@andarist

> Just chiming in - if we go with defaulting to "script", I think it's in best interest of the community to warn when using the default option, therefore "force" people to use pass value for this option explicitly.

To me, it seems like defaulting to `script` without config will either

* immediately trigger a error when an `import` or `export` are encountered
* Silently parse a file as non-strict

It seems to me that the probability of an ESM file _not_ using at least one `import` or `export` is so low that I'm willing to assume that the 99% case will get the error immediately. Given that, it seems overly cautious to require the users to choose up front.

loganfsmyth:
@novemberborn 

> Presumably one would use the overrides option with a node_modules/package-name/**/*.js pattern to adjust the source type for dependency files if necessary? That might be fragile if the dependency tree changes though.

Potentially. It seems like realistically in the short term people might opt to just say `sourceType:unambiguous` for the time being when parsing `node_modules`, but that's also the case _currently_ so it doesn't seem like it will have a short-term effect there. In the long term either people will transition to `.mjs` or Babel can change to default to whatever other side-channel Node has settled on (though I know you already have how Babel searches for .babelrc, so I bet you also don't love the idea of a flag in `package.json`?)

> To clarify something else, with these changes, transform-modules-commonjs will throw if used on a Script? And would Babel preserve require, module, exports, __filename and __dirname even in a Module? Perhaps with a configuration option?

Currently `transform-modules-commonjs` bails entirely if the file has not `sourceType:module`, so by defaulting to `script` it should be easier to run CommonJS code through Babel. 

We don't _currently_ error if we detect `module` and `exports` inside an ESM module, but that is something I'd very much like to do, no matter the decision here. If we _do_ leave the default as `module`, I'd expect Babel to start throwing exceptions whenever it is passed a CommonJS file, since the alternative is auto-strict unexpectedly. Even with that, it's not ideal to default to `module` since we'd still be unable to throw helpful error messages if the input code relies on annex B functionality or other things disallowed in strict mode.

loganfsmyth:
@iarna

> I for one am convinced that while we may end up with mjs allowed, we will not end up with it mandated.

I want to be clear, I'm not at all looking to influence that. If Node decides it's only one of multiple suggested side-channels, I'd be more than happy to adopt those alternatives into Babel. My goal here is entirely to make sure that Babel's default at least has _some_ side-channel. Babel's current approach has essentially been to hope for the best and push the pain onto users so when something breaks, they have to figure it out, which is something I find super undesirable.

> So please, don't preumaturely and possibly pointlessly break your users.

It's not clear to me why this attempt is seen as pointless. This is a problem users will be facing consistently as more packages are published as ES6-only and requires compilation for older environments.

It seems quite that `.mjs` will _one_ of several possible ways to mark a file as a module. Choosing it over nothing at all, and letting people settle into alternatives over time seems like the best approach.

> Escape hatches don't make it less a mandate.

I'm not particularly excited about breaking users either way, but the point remains that Babel needs to know what type of file it is working with to reliably transform code. Either we give a recommended like .mjs that is likely to at least be _one_ possible option, or we stick with `.js` and show anyone using CommonJS errors all over the place, sometimes helpful and sometimes not.

I think the big question for me is what we do actually gain by _not_ doing this? By doing nothing, we're actively encouraging the community to continue with no particular choice of side-channel for the time being, which seems harmful to the future of the community moreso than suggesting `.mjs` for now.

loganfsmyth:
@weswigham

> the assertion that "accidentally loose" is inherently better than "accidentally strict" seems like a tenuous one...? Especially since the more code is written with modules (which is what people here are gunning for), the less and less it's true.

Accidentally module leads to random runtime or parsing errors with unclear messaging on how users are supposed to fix it. It also means that Babel's transformations may inject `import` statements into code that is not an ES6 module, parsing errors in code that the user didn't even write and causing confusion.

Accidentally script means that code may throw _less_ which is certainly not ideal, but I'll also add that I'd expect the frequence of this to be drammatically less than the frequency with which auto-strict will break things because you're much more likely to have a `.js` file that is a script, than you are to have a `.js` file that is a module that never uses `import` or `export`.

> Why change the default for an unstable feature that's in a state of flux, only to change it again later. You could argue that implementing too early is exactly what's fragmented the module ecosystem to where it is today.

That's totally fair. For me at least is for Babel to function well in a world where users are broadly compiling third-party code, this side-channel needs to exist. If someone can point to information saying `.mjs` is legitimately something node _won't_ use, rather than something that that might be optional alongside other side-channels, it seems like `.mjs` is a reasonable choice to make until those other side-channels exist.


> Hopefully it should be at least slightly obvious by the discussion here why at least our team doesn't consider the current implementation stable enough to build on yet.

I promise I never had any doubt that this would blow up as a PR, but I still think it's an important discussion to have. I'm reading every single response carefully.

> As much as something has "shipped" (under a flag with a guarantee to change), it is also guaranteed not to be final. It's excellent that babel can support the current flag already, but it's definitely not ready for default-behavior-informing spotlights

I still feel like no-one here has been able to state that `.mjs` will not happen (and obviously proving a negative is not exactly easy, but still :P). If it's not the primary suggestion Node settles on, we can certainly adjust over time, but it seems extremely like to at least be one of several options for a side-channel. Given that, it doesn't actually seem important to me that Node has or has not shipped it.

Can anyone here say that they realistically expect `.mjs` to be entirely unsupported?


> especially when the change will be breaky.

I anticipate, if we _don't_ land this, then Babel 100% needs to add a loud breaking change to throw if users try to parse a CommonJS file as a module, since many Babel transforms now rely critically on being able to know what kind of file it is working with. Given that we're already going to piss off users with that, the fact that it is breaking, on its own, isn't super compelling for me.

TheLarkInn:
> I think the big question for me is what we do actually gain by not doing this? By doing nothing, we're actively encouraging the community to continue with no particular choice of side-channel for the time being, which seems harmful to the future of the community moreso than suggesting .mjs for now.

I couldn't agree more here. This is the exact same reason webpack supports .mjs as a "module type". 

One thing I'd like clarified: does this mean that babel will not transpile module statements and leave as is if this PR lands? (Unless otherwise configured to do so). This in combination with the million who are using webpack will see some incredibly beneficial size reductions in their linked bundles because they are explicitly opting (in one way or another) to using ESM (and it will not be down converted to cjs)

appsforartists:
My hunch is that people just don't like `mjs`, and they're afraid that Babel is popular enough that its decisions will add legitimacy to a solution that they don't want to see adopted at all.

I looked at the [Unambiguous Grammar](https://github.com/bmeck/UnambiguousJavaScriptGrammar) spec, and while I laud its aims, I also understand why browser vendors may be unmotivated to push it forward.  If you're importing a script in the browser, you're either using `<script type="module">` or `import {} from "";`, so the browser is never ambiguous.  I also expect they'd be hesitant to add anything to the critical path for parsing/execution.

That said, node and Babel could certainly follow Unambiguous Grammar spec without waiting for browsers to endorse it.

ljharb:
@TheLarkInn this PR only deals with sourceType; transforming import/export is a separate issue (albeit, one that only applies when the sourceType is module).

@appsforartists if node and babel are doing things "without waiting for browsers", then any argument that relates to browser interop imo vanishes; I don't think that'd be a good idea for any of the "camps" here.

appsforartists:
@ljharb Unambiguous Grammar feels orthogonal to browsers.  Imports are never ambiguous for browsers in the first place.

loganfsmyth:
@TheLarkInn

> One thing I'd like clarified: does this mean that babel will not transpile module statements and leave as is if this PR lands? (Unless otherwise configured to do so).

I hadn't thought about it, but that does seem like a conversation we should have pre-RC. I'd probably be open to that, since it's always easier to add a plugin than remove it in Babel. Maybe continue that discussion in another issue?

loganfsmyth:
@appsforartists

> My hunch is that people just don't like mjs, and they're afraid that Babel is popular enough that its decisions will add legitimacy to a solution that they don't want to see adopted at all.

It does seem that way, and I do understand the concern there because they are right that people associate very strongly with `.js`. Ultimately, I have a hard time seeing any way for the community as a whole to adopt a side-channel that will satisfy _every_ usecase that isn't either in the filename, or in the file content. There are _so_ many competing groups trying to decide what is best for their specific case. 

To me, a file extension is the least offensive way to handle that without typing the solution to one specific ecosystem. I can easily see individual communities adopting other options, or simply assuming that code is ESM, and that may well be the right choice for that specific community. Babel cannot hope to have a default that keeps every one of those happy with their individual ecosystem choices, so we're essentially left with the choice of choosing the least offensive option (in my opinion `.mjs`), or letting our users continue to get random errors and have a painful experience in some cases to avoid making a decision.

> That said, node and Babel could certainly follow Unambiguous Grammar spec

Unless there's real indication of Node pursuing that approach and it having a real standard, I don't think I'd be comfortable to have Babel default to that.

Kovensky:
Whatever side channel the working group decides on, it will strictly be a way for _opting in_ to modules instead of opting out. Opting out is guaranteed to break a ton of real world code.

If the default is script, we can just add support to whatever side-channels are decided on (such as `.mjs`, even if it eventually turns out to not be the recommended one) as they are decided on by the WG, in any future minor update to babel-core, without compatibility risk.

If we keep module as the default, we'll just keep on silently doing the wrong thing for the vast majority of 3rd party code, and then when the WG gets their ducks in order we're going to have to make another breaking change to implement their algorithm.

As for changing the default, a lot of code happens to run fine in both strict and sloppy mode, so you won't notice anything if all that babel ends up doing is writing a "use strict".

But combine that with transform-runtime, preset-env with useBuiltIns: "usage", or any plugin that needs to add an import to a file, and you easily create invalid "mixed" modules, that have require, static imports and exports/module.exports assignment all in the same file. If that output were to then be treated as a module (which it would in unambiguous grammar too, it has a static import), you then get a ReferenceError, but only at runtime. Babel will silently take your valid script (even strict mode script) and write an invalid module.

If we know the input is not a module, then the aforementioned plugins know to write commonjs requires instead (well, env will once it uses the correct helper module). If we think the input is not a module but it has module syntax, it's easy to raise an error telling you what to do in your .babelrc. If we think the input is a module but it's not, we corrupt your module and there's nothing you can do to detect other than seeing if it crashes at runtime.

weswigham:
> Can anyone here say that they realistically expect .mjs to be entirely unsupported?

I think the conceit is that everyone knows `.mjs` is _likely_ to be available, but that it's _also_ likely that a another solution with some set of tradeoffs is _also_ implemented, since on it's own it's very not-friendly-to-use alongside the existing cjs package ecosystem 😄 

The _core_ of the issue is that changing babel's default to be `script` for any `js` is that it _insists_ that, by default, only `.mjs` files can be modules. This is false for a browser environment, and very likely to be generally false in `node`, in one way or another. That's a really bad understanding of the world, by default; being just a _little_ bit smarter would allows you to be way less naggy in this space. But since y'all have no problem breaking all your consumers every 3 months, nbd; there's no brakes on this train, and I guess you _just have_ to implement to whatever partially done work is there _right now_, since you can just break everyone _again_ and redo your work later anyway. TBQH, at TS since we power editor services for a bunch of editors, we'll probably continue to be using an `unambiguous`-style heuristic for almost forever because it generally _works_, with an opt-in to whatever node's _final_ behavior ends up as it'd be crazy for an editor to ask "hey, I'm not sure how to handle this .js file that's clearly a module, you wanna tell me?" before providing useful tools for a given file (and there's no way we'd break everyone using all the editors we power, because that'd be extreme), and all of our command line tools generally fall out from the target editor experience.

Kovensky:
Editor services are a completely different world from compilers. Editor services will only be of use on files that a user actually wants to look at and edit, have to be able to work with incomplete code, and even if the user opens a file that the editor services guesses wrong, the worst thing that can happen is bad syntax highlighting and code completion.

Babel is a compiler. If we guess wrong, _we write broken code_. Even if TS guesses wrong, it can throw an error under `--isolatedModules` that your code is not actually unambiguous enough to tell, and that's something that TS can afford to do as it's a language _based_ on JS, it's not JS.

TS is also not used to compile third party code, but the ecosystem is moving in such a way that babel, with preset-env, will be used to compile all third party code as well. Third party code that, in a lot of cases, has already been transformed from a module into a script. Third party code that we can correctly parse as module, because its original source code was a module, and so it doesn't have sloppy mode syntax. Third party code that we will _break_ by then emitting extra module-related garbage.

weswigham:
> Editor services are a completely different world from compilers. 

Many modern compilers are written as services, as service behavior is a superset of the analysis a compiler needs to perform (and is quite useful to have). A compiler pretty much falls out for free from a language service, while building a service on top of a compiler is really difficult.

> it's not JS

And I'm not talking about typescript. I'm talking about JS. I'm talking about when `vscode` opens a `.js` file what's going to happen. What's even reasonable. Just because TS is marketed as a seperate language doesn't mean that's all its' tools handle, ffs - the TS service does the whole intellisense gig for all kinds of vanilla JS (and TS and JS with closure-compiler type annotations because we like to be a magical as possible). And I'm saying this will never be the default mode of operation an editor takes when handling JS because it'd just be silly to ask a user for so much before they even start reading code.

> TS is also not used to compile third party code

It is. All the time. Especially within editors. We probably consume more of it than _almost_ anything else. That's why a decent heuristic is so important because there's so much _varied_ code out there. And way more old code than new code. So much old code. Nobody updates old code unless they're forced to, really.

> Third party code that we can correctly parse as module, because its original source code was a module, and so it doesn't have sloppy mode syntax. Third party code that we will break by then emitting extra module-related garbage.

Do you have real examples? Because TBQH most of the ones I can think of are caught at parse time inside babel/ts, such as `with` or block-scoped function declarations. Top-level `this` being loose by default instead of strict by default is worse, since a loose mode `this` leaks environment information that was supposed to be removed.

Kovensky:
```js
// src/index.js
import 'react' // or any commonjs module without a modules build that uses ES6+ libraries
```

```js
// webpack 4 config
module.exports = {
  mode: 'development',
  // note: does not exclude node_modules
  modules: { rules: [
    { test: /\.js$/, exclude: /core-js/, loader: 'babel-loader' }
  ] }
}
```

```json
{
  "presets": [[ "@babel/env", { "useBuiltIns": "usage", "modules": false } ]]
}
```

This will write `import` for a bunch of `core-js` dependencies on top of all your CJS dependencies, which will cause webpack to either fail to parse your script, or to incorrectly parse it as a module with no exports.

```
webpack:///./node_modules/object-assign/index.js?:82
module.exports = shouldUseNative() ? Object.assign : function (target, source) {
               ^

TypeError: Cannot assign to read only property 'exports' of object '#<Object>'
```

(`module.exports` is readonly in webpack because it still lets you look at everything else in `modules`; in actual modules it'd throw a `ReferenceError`)

Note, there is actually a bug with preset-env in that it does not respect the `sourceType` (https://github.com/babel/babel/issues/7457), so even setting `sourceType` to "script" won't help here; but this is exactly the behavior it would have with `sourceType` defaulting to "module" anyway.

weswigham:
@Kovensky `modules: false` effectively means input module kind should equal output module kind, no? Since it's disabling the transform, not asserting that you have no modules? - it's then up to webpack to determine the correct module kind to work with for each file in the build is, no?

Kovensky:
@weswigham the problem here is not our modules transform (which would, at worst, just take sloppy code and accidentally make it strict); the problem is when the modules transform is disabled (which it should! you should never use transform-modules-commonjs with webpack).

If the modules transform is disabled, then the output module kind should equal the input module kind. The input module kind is `"module"`, because that's our current default. So we would write a module in the output, with `import 'core-js/modules/es6.foo.bar'` for the `useBuiltIns` polyfills, even after that bug I mentioned is fixed.

weswigham:
> The input module kind is "module", because that's our current default.

Y'all should default to something like `unambiguous` and not `script`, then, cause you're not doing a great job of preserving the intent of the code as written by passing it on while assuming it's one or the other... You could get the exact same issues but with cjs stuff instead of module stuff if you just swap the default...

loganfsmyth:
@weswigham 

> You could get the exact same kind of issues but with cjs stuff instead of module stuff if you just swap the default...

The fact that it's not exactly the same is an important factor for me. If we default to `module` and it was actually cjs, we've injected an `import` and you'll get a runtime error in your code with no clear reason why it has happened.

If we default to `script` and it was actually supposed to be a module, 99% of the time the file will have an `import` or `export` statement that causes parsing to fail up front, and users will get a super clear message saying "hey it looks like this is a module, is that right? If so please set `.js` files to be modules in your config, or rename it to be `.mjs`.

Defaulting to `script` means we can give super useful feedback to our users right up front. Defaulting to `module` makes that much harder.

> Y'all should default to something like unambiguous and not script, then, cause you're not doing a great job of preserving the intent of the code as written by passing it on while assuming it's one or the other...

But `unambiguous` is in no way standards track and attempts to make it so failed. It's also _trivially_ easy to either opt into it if you want, and if it does eventually become part of the spec it is easy to make it the default instead of `script` without breaking anyone's setup. 

My concern with `unambiguous`, _at this point in time_, is that it encourages people to publish code that relies on unambiguous to be present in all tooling that processes those files, not just now, but for the existence of those module's lifetimes. Given that there is no current plans for standardization on it, this seems like a massive problem to me. If you're random tooling that parses a file for metadata where the specifics of module vs script don't matter as much, all the more power to you, `unambiguous` makes sense.

If we're trying to build a package repository like the npm registry, it makes zero sense for people to publish ESM as `.js` because the answer is that we just haven't settled as a community on an answer yet. If we encourage people _now_ to publish and assume `.js` as unambiguous is a set thing, we're still setting a mandate for the community, but we're setting one that is _way_ more likely to change than the existence of `.mjs`.

* If we encourage `.mjs` for now and next year Node also handles something, it's an easy change to _also_ support that.
* If we encourage `.js` as ESM now, and next year Node settles on something without unambiguous, we've now got a registry full of packages that no longer work right. More importantly, we've gone an additional year training users to expect ESM to work in `.js` files automatically, and it'll be that much harder to transition them to another mental model.

`.mjs` is the safest option for ESM until the community settles on a common solution broadly.

appsforartists:
> If we encourage .js as ESM now, and next year Node settles on something without unambiguous, we've now got a registry full of packages that no longer work right.

Aren't we already a few years down this path?

novemberborn:
@loganfsmyth 

> (though I know you already have how Babel searches for .babelrc, so I bet you also don't love the idea of a flag in package.json?)

Apologies everyone for the inside baseball — @loganfsmyth  I'm sure this'll be fine.

> Currently transform-modules-commonjs bails entirely if the file has not sourceType:module, so by defaulting to script it should be easier to run CommonJS code through Babel.

👍 

> We don't currently error if we detect module and exports inside an ESM module, but that is something I'd very much like to do, no matter the decision here.

If you could loop me in at that stage that'd be appreciated. I'll be advocating for an option to disable that behavior 😄

loganfsmyth:
@appsforartists

> Aren't we already a few years down this path, with authors publishing ESM as .js?

Some have, but to me this point seems like the sunk cost fallacy. We're talking about the  an unspecificed amount of time into the future where we'll need to maintain this ecosystem. I don't think locking the ecosystem to unambiguous based on reliance of undefined behavior at the beginning makes the decision the right one for the future of the community. And again, if it turns out `.mjs` means module and `.js` means unambiguous in the future, we've lost nothing, whereas if unambiguous is not adopted, we've now spent even more time creating an incompatible ecosystem.

@novemberborn 

> If you could loop me in at that stage that'd be appreciated. I'll be advocating for an option to disable that behavior 

Fair. Let's continue that discussion on Slack so as not to diverse this thread.

loganfsmyth:
Hey all, we've talked about this a bunch both here and in Babel's chat. I really appreciate all the discussions we've had.

We've settled on postponing this change for the 7.x release because it will be such a large breaking change. The behavior will essentially be remaining like it is now, which is roughly how it was in Babel 6, assuming everything is a module and hoping for the best, and relying on users to configure Babel if things go wrong with that. It's important to not discard the good will of the community itself, _especially_ given that we've been in beta for quite a while and people are unlikely to be expecting such large changes at this phase.

I'd still like to seriously consider this as an early change for the next major release of Babel, which would allow us time to test it out and see how it feels to people, but we'll cross that bridge when we come to it.

PR #7502, opened 2018-03-05
babel-cli: Output files with .mjs extensions when sourceType:module 

### LINKED ISSUE #6222: Ability to output mjs extension when -d is specified
Feature request:

[Recommendation](https://twitter.com/bradleymeck/status/906572163519578112) from Node core is to output mjs alongside js files so that the package.json `main` field can be ambiguous. Since babel only outputs `.js` files when used with the `-d` directory flag, it becomes complicated to make this work. I'd like to see an option available to specify the extension the same as with parsing/consuming mjs.

### Current Behavior

This is the most elegant solution I've found so far, which involves building all files into a temporary directory, renaming all the files to mjs, copying them into the `dist/node` destination, and then removing the temp directory.

``` sh
"build-node-esm": "babel lib -d dist/node-es && rename -v .js .mjs dist/node-es/*.js && rename -v .js .mjs dist/node-es/**/*.js && cp -R dist/node-es/* dist/node && rm -rf dist/node-es",
```

### Possible Solution

My suggestion would be to add a new CLI flag option to specify the output extension. This will allow developers to easily support both Node-based ESM/CJS.

``` sh
babel lib -d dist/node --ext ".mjs"
```

### Context

I'm trying to figure out how to build a library that will be ready to take advantage of the new Node ESM semantics.

### Your Environment

| ---------------- | -------
| node             | v9.0.0-pre

loganfsmyth:
| Fixed Issues?            | Fixes #6222
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

This changes the CLI so that if the file being output is a module, and the sourceType wasn't rewritten to `script` by a transform, the file will be output with a `.mjs` extension. This feature is desirable because Node has made it clear that `.mjs` is the primary means to mark a file as a module. It may not end up being the _only_ way, but certainly for users wishing to create dual-mode packages, it will be necessary. That means I'd expect many packages written with ES6 module may want to do:

```
{
  env: {
    commonjs: {
      plugins: ["@babel/transform-modules-commonjs"]
    },
  }
}
```
with
```
babel src -d lib --env-name commonjs
babel src -d lib
```
where the first would compile all module files in `src` into `.js` files in `lib`, with the second command compiling all the files into `.mjs` files in `lib`. There could certainly be other plugins and presets still activated in there.

I don't know that I want to land this as-is, but I'm posting it as motivation. The downside _currently_ is that because we default to `sourceType: module`, this would make _every_ file output as `.mjs`. On the other hand, if we _don't_ do this, it will be painful for users to perform this type of compilation, since we'd have to instead have an option like `--use-mjs` or something, but it's also possible that some files in a project would be CommonJS still, even if most of them are ESM, so a blanket flag like `--use-mjs` would then introduce the opposite problem.

I think we really need to consider defaulting to `sourceType: script` for #6242 and pending #7501.

### DIFF
--- a/packages/babel-cli/package.json
+++ b/packages/babel-cli/package.json
@@ -33,7 +33,8 @@
   },
   "devDependencies": {
     "@babel/core": "7.0.0-beta.40",
-    "@babel/helper-fixtures": "7.0.0-beta.40"
+    "@babel/helper-fixtures": "7.0.0-beta.40",
+    "@babel/plugin-transform-modules-commonjs": "7.0.0-beta.40"
   },
   "bin": {
     "babel": "./bin/babel.js",
--- a/packages/babel-cli/src/babel/dir.js
+++ b/packages/babel-cli/src/babel/dir.js
@@ -18,22 +18,26 @@ export default function(commander, filenames, opts) {
       return process.nextTick(callback);
     }
 
-    // remove extension and then append back on .js
-    relative = util.adjustRelative(relative, commander.keepFileExtension);
-
-    const dest = getDest(commander, relative, base);
-
     util.compile(
       src,
       defaults(
         {
-          sourceFileName: slash(path.relative(dest + "/..", src)),
+          sourceFileName: slash(
+            path.relative(getDest(commander, relative, base) + "/..", src),
+          ),
         },
         opts,
       ),
       function(err, res) {
         if (err) return callback(err);
         if (!res) return callback();
+
+        // remove extension and then append back on .js
+        if (!commander.keepFileExtension) {
+          relative = util.adjustRelative(relative, res.sourceType === "module");
+        }
+
+        const dest = getDest(commander, relative, base);
 
         // we've requested explicit sourcemaps to be written to disk
         if (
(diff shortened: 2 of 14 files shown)


xtuc:
I think that this is a big breaking change. 

I think that this should be in babel's configuration (and not by default for now). 

That would enable more scenarios where you can emit mjs or js depending on the context/dir/config etc

nicolo-ribaudo:
I think that the default output extension for modules should be `.mjs`. This change doesn't affect users which use babel-loader, gulp-babel or other bundlers/task runners. It will only affect people which use nodejs or `<script type="module" src="...">` in the browser.
Node only supports `mjs` modules, while browsers don't care about the extension. `mjs` works in both, `js` doesn't.

I *think* that most people areusing preset-es2015 or preset-env, so even if Babel defaults to `soureType: module` it will be changed to `script` by the commonjs/umd transformer. If someone doesn't transpile modules because doesn't use them, they can add `sourceType: script` to their config.

loganfsmyth:
Per https://github.com/babel/babel/pull/7501#issuecomment-391232964, I'm going to close this since I'll need to rework this logic to be opt-in.

PR #7507, opened 2018-03-06
Use $ReadOnlyArray in generated babel-types type defs

sophiebits:


sophiebits:
Not 100% sure this is right but I think it is. It generates like:

```
  declare function callExpression(callee: BabelNodeExpression, _arguments: $ReadOnlyArray<BabelNodeExpression | BabelNodeSpreadElement | BabelNodeJSXNamespacedName>, optional?: true | false, typeParameters?: BabelNodeTypeParameterInstantiation | BabelNodeTSTypeParameterInstantiation): BabelNodeCallExpression;
```

but also makes $ReadOnlyArray inside each class's type.

nicolo-ribaudo:
I like the idea of an immutable ast (which shouldn't appy only to arrays, but also to nodes), but this PR would make it an error to use functions like
```js
objectNode.properties.push(t.objectProperty(key, value));
```
because `objectNode.properties` would be a read-only array.

We might try to create a `$DeepReadOnly` type helper in the future, so that consumers could choose to freeze their ast.

PR #7513, opened 2018-03-06
Migrate babel-core tests to use jest-expect

devenbansod:

* Replace `assert.*equal` calls with appropriate `expect` calls

* Used codemods at: https://gist.github.com/devenbansod/03c5cff857661e076cbec72fcb2e7eb3 (along with some manual intervention and review)

### DIFF
--- a/packages/babel-core/test/api.js
+++ b/packages/babel-core/test/api.js
@@ -1,16 +1,15 @@
 import * as babel from "../lib/index";
 import sourceMap from "source-map";
-import assert from "assert";
 import path from "path";
 import Plugin from "../lib/config/plugin";
 import generator from "@babel/generator";
 
 function assertIgnored(result) {
-  assert.ok(!result);
+  expect(result).toBeFalsy();
 }
 
 function assertNotIgnored(result) {
-  assert.ok(!result.ignored);
+  expect(result.ignored).toBeFalsy();
 }
 
 // shim
@@ -48,18 +47,16 @@ describe("parser and generator options", function() {
 
   it("options", function() {
     const string = "original;";
-    assert.deepEqual(
-      newTransform(string).ast,
+    expect(newTransform(string).ast).toEqual(
       babel.transform(string, { ast: true }).ast,
     );
-    assert.equal(newTransform(string).code, string);
+    expect(newTransform(string).code).toBe(string);
   });
 
   it("experimental syntax", function() {
     const experimental = "var a: number = 1;";
 
-    assert.deepEqual(
-      newTransform(experimental).ast,
+    expect(newTransform(experimental).ast).toEqual(
       babel.transform(experimental, {
         ast: true,
         parserOpts: {
@@ -67,7 +64,7 @@ describe("parser and generator options", function() {
         },
       }).ast,
     );
-    assert.equal(newTransform(experimental).code, experimental);
+    expect(newTransform(experimental).code).toBe(experimental);
 
     function newTransformWithPlugins(string) {
       return babel.transform(string, {
@@ -82,8 +79,7 @@ describe("parser and generator options", function() {
       });
     }
 
-    assert.deepEqual(
-      newTransformWithPlugins(experimental).ast,
+    expect(newTransformWithPlugins(experimental).ast).toEqual(
       babel.transform(experimental, {
         ast: true,
         parserOpts: {
@@ -91,14 +87,13 @@ describe("parser and generator options", function() {
         },
       }).ast,
     );
-    assert.equal(newTransformWithPlugins(experimental).code, experimental);
+    expect(newTransformWithPlugins(experimental).code).toBe(experimental);
   });
 
   it("other options", function() {
     const experimental = "if (true) {\n  import a from 'a';\n}";
 
-    assert.notEqual(
-      newTransform(experimental).ast,
+    expect(newTransform(experimental).ast).not.toBe(
       babel.transform(experimental, {
         ast: true,
         parserOpts: {
@@ -106,21 +101,19 @@ describe("parser and generator options", function() {
         },
       }).ast,
     );
-    assert.equal(newTransform(experimental).code, experimental);
+    expect(newTransform(experimental).code).toBe(experimental);
   });
 });
 
 describe("api", function() {
   it("exposes the resolvePlugin method", function() {
-    assert.throws(
-      () => babel.resolvePlugin("nonexistent-plugin"),
+    expect(() => babel.resolvePlugin("nonexistent-plugin")).toThrow(
       /Cannot find module 'babel-plugin-nonexistent-plugin'/,
     );
   });
 
   it("exposes the resolvePreset method", function() {
-    assert.throws(
-      () => babel.resolvePreset("nonexistent-preset"),
+    expect(() => babel.resolvePreset("nonexistent-preset")).toThrow(
       /Cannot find module 'babel-preset-nonexistent-preset'/,
     );
   });
@@ -135,9 +128,9 @@ describe("api", function() {
       res,
     ) {
       if (err) return done(err);
-      assert.equal(res.code, "foo();");
+      expect(res.code).toBe("foo();");
       // keep user options untouched
-      assert.deepEqual(options, { babelrc: false });
+      expect(options).toEqual({ babelrc: false });
       done();
     });
   });
@@ -147,20 +140,19 @@ describe("api", function() {
       babelrc: false,
     };
     Object.freeze(options);
-    assert.equal(
+    expect(
       babel.transformFileSync(__dirname + "/fixtures/api/file.js", options)
         .code,
-      "foo();",
-    );
-    assert.deepEqual(options, { babelrc: false });
+    ).toBe("foo();");
+    expect(options).toEqual({ babelrc: false });
   });
 
   it("options throw on falsy true", function() {
-    return assert.throws(function() {
+    return expect(function() {
       babel.transform("", {
         plugins: [__dirname + "/../../babel-plugin-syntax-jsx", false],
       });
-    }, /.plugins\[1\] must be a string, object, function/);
+    }).toThrow(/.plugins\[1\] must be a string, object, function/);
   });
 
   it("options merge backwards", function() {
@@ -168,9 +160,8 @@ describe("api", function() {
       presets: [__dirname + "/../../babel-preset-es2015"],
       plugins: [__dirname + "/../../babel-plugin-syntax-jsx"],
     }).then(function(result) {
-      assert.ok(
-        result.options.plugins[0].manipulateOptions.toString().indexOf("jsx") >=
-          0,
+      expect(result.options.plugins[0].manipulateOptions.toString()).toEqual(
+        expect.stringContaining("jsx"),
       );
     });
   });
@@ -185,7 +176,7 @@ describe("api", function() {
           return callback;
         }
 
-        assert.equal(visitorType, "enter");
+        expect(visitorType).toBe("enter");
 
         return function() {
           calledIntercept++;
@@ -205,8 +196,8 @@ describe("api", function() {
       ],
     });
 
-    assert.equal(calledRaw, 4);
-    assert.equal(calledIntercept, 4);
+    expect(calledRaw).toBe(4);
+    expect(calledIntercept).toBe(4);
   });
 
   it("pass per preset", function() {
@@ -266,9 +257,9 @@ describe("api", function() {
 
     let result = execTest(true);
 
-    assert.equal(aliasBaseType, "NumberTypeAnnotation");
+    expect(aliasBaseType).toBe("NumberTypeAnnotation");
 
-    assert.deepEqual(result.code, "var x = function x(y) {\n  return y;\n};");
+    expect(result.code).toEqual("var x = function x(y) {\n  return y;\n};");
 
     // 2. passPerPreset: false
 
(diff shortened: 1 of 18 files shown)


packages/babel-core/test/api.js
```diff
@@ -1,16 +1,15 @@
 import * as babel from "../lib/index";
 import sourceMap from "source-map";
-import assert from "assert";
 import path from "path";
 import Plugin from "../lib/config/plugin";
 import generator from "@babel/generator";
 
 function assertIgnored(result) {
-  assert.ok(!result);
+  expect(!result).toBeTruthy();
```
Andarist: this probably should at least get reversed to
```js
expect(result).toBeFalsy();
```

packages/babel-core/test/api.js
```diff
@@ -1,16 +1,15 @@
...
 }
 
 function assertNotIgnored(result) {
-  assert.ok(!result.ignored);
+  expect(!result.ignored).toBeTruthy();
```
Andarist: same here

packages/babel-core/test/api.js
```diff
@@ -48,26 +47,24 @@ describe("parser and generator options", function() {
 
   it("options", function() {
     const string = "original;";
-    assert.deepEqual(
-      newTransform(string).ast,
+    expect(newTransform(string).ast).toEqual(
       babel.transform(string, { ast: true }).ast,
     );
-    assert.equal(newTransform(string).code, string);
+    expect(newTransform(string).code).toEqual(string);
```
Andarist: `.code` is a string so imho `.toBe` would be more appropriate here

packages/babel-core/test/api.js
```diff
@@ -608,25 +596,25 @@ describe("api", function() {
 
     it("all", function() {
       const script = babel.buildExternalHelpers();
-      assert.ok(script.indexOf("classCallCheck") >= -1);
-      assert.ok(script.indexOf("inherits") >= 0);
+      expect(script.indexOf("classCallCheck") >= -1).toBeTruthy();
```
Andarist: you could use `arrayContaining` matcher here

packages/babel-core/test/api.js
devenbansod: I agree. :-) Will make the change in sometime.

packages/babel-core/test/api.js
devenbansod: I believe I messed up in applying the codemod (meaning I applied the code-mod to an already modified file. These were supposed to be correctly changed to `toBe` according to the code-mod (if I had applied it correctly).
Changing now.

packages/babel-core/test/api.js
devenbansod: I am not sure if we can use it here, since `typeof script` returns `string`.

Kovensky: You can if you use `.split(/\b/)` first :trollface: 

Or use the `stringContaining` matcher 🤷‍♀️ 

Replacing `indexOf`s would need to be a case-by-case thing, not really something a codemod can do without type information 😞

devenbansod: Agreed. Thanks @Kovensky for the suggestion. 
Now, changed to using `stringContaining`. :-)

danez:
In fixtures tests, we could make both available `expect` and `assert`, so we don't have to migrate everything at once.
I guess it should work if we forward expect here, but haven't tested: 

https://github.com/babel/babel/blob/master/packages/babel-helper-transform-fixture-test-runner/src/index.js#L19..L25

packages/babel-core/test/api.js
```diff
@@ -386,48 +377,46 @@ describe("api", function() {
...
+      column: 6,
+    });
   });
 
   it("code option false", function() {
     return transformAsync("foo('bar');", { code: false }).then(function(
       result,
     ) {
-      assert.ok(!result.code);
+      expect(!result.code).toBeTruthy();
```
danez: Could be `expect(result.code).toBeFalsy();`

packages/babel-core/test/api.js
```diff
@@ -386,48 +377,46 @@ describe("api", function() {
...
     });
   });
 
   it("ast option false", function() {
     return transformAsync("foo('bar');", { ast: false }).then(function(result) {
-      assert.ok(!result.ast);
+      expect(!result.ast).toBeTruthy();
```
danez: Could be `expect(result.ast).toBeFalsy();`

packages/babel-core/test/api.js
```diff
@@ -386,48 +377,46 @@ describe("api", function() {
...
     return transformAsync("foo('bar');", { ast: true }).then(function(result) {
-      assert.ok(result.ast);
+      expect(result.ast).toBeTruthy();
     });
   });
 
   it("ast option default", function() {
     return transformAsync("foo('bar');").then(function(result) {
-      assert.ok(!result.ast);
+      expect(!result.ast).toBeTruthy();
```
danez: Could be `expect(result.ast).toBeFalsy();`

devenbansod:
Thanks @danez for the review. Adding `expect` at your suggested location did help. Have added those changes with the latest squashed commit.

Only thing missing to be migrated in `babel-core`'s test is left is migrating the tests in [packages/babel-core/test/parse.js](https://github.com/babel/babel/blob/master/packages/babel-core/test/parse.js). It seems to have been added in https://github.com/babel/babel/commit/22c8f6376c622bd589b828d66d88377b9080101a#diff-a9a30dd3ba728e2fa262f32730241c5f but I am not sure what is the test meant to assert [here](https://github.com/babel/babel/blob/master/packages/babel-core/test/parse.js#L14). Any pointers?

mallikarjuna91:
@devenbansod The test is meant to assert the Parsed input file matches the output file ?

danez:
I guess these should be `toBe` or `toEqual`. They just compare if the result of parse is equal to the content of `output.json`.

devenbansod:
Because [here](https://nodejs.org/api/assert.html#assert_assert_value_message) it says it's an alias of `assert.ok()` which just takes in value and message and tests whether the value is truthy. It did not look like that file assertion. May be the author meant to add `assert.equal` or `assert.deepEqual` but forgot?

According to your suggestion, I have changed to:
```
function fixture(...args) {
  return path.join(__dirname, "fixtures", "parse", ...args);
}

describe("parse", function() {
  it("should parse using configuration from .babelrc when a filename is provided", function() {
    const input = fs.readFileSync(fixture("input.js"), "utf8");
    const output = require(fixture("output"));
    expect(
      parse(input, { filename: fixture("input.js"), cwd: fixture() }),
    ).toEqual(output);
  });

  it("should parse using passed in configuration", function() {
    const input = fs.readFileSync(fixture("input.js"), "utf8");
    const output = require(fixture("output"));
    expect(parse(input, { parserOpts: { plugins: ["decorators"] } })).toEqual(
      output,
    );
  });
});
```

I seem to be getting errors mentioned in this [gist](https://gist.github.com/devenbansod/2d70d53e9d920ad77462a867c6937c83). Is there something that I am missing here? Apologies for repeated questions and queries.

danez:
Okay seems the test never worked as intended. 
I think it could work when you change it to:

```js
...
const result = parse(...);
expect(JSON.parse(JSON.stringify(result))).toEqual(output);
...
```

If it doesn't work this way, then you can also simply check for `truthy` now but leave a comment that says it should be fixed. We can then create a separate ticket to fix the tests.

@kaicataldo We might need to do something more clever here anyway unless we really want to snapshot the order of the fields. Babylon has this compare function maybe we should share it and use it here too.

devenbansod:
Thanks for the suggestion @danez! 

The difference seemed to have reduced to just this:
```
Difference:
    
    - Expected
    + Received
    
    @@ -212,11 +212,10 @@
                },
                "start": 59,
                "type": "BlockStatement",
              },
              "end": 90,
    -         "expression": false,
              "generator": false,
              "id": Object {
                "end": 50,
                "loc": Object {
                  "end": Object {

      24 | 
      25 |     const result = parse(input, { parserOpts: { plugins: ["decorators"] } });
    > 26 |     expect(JSON.parse(JSON.stringify(result))).toEqual(output);
      27 |   });
      28 | });
      29 | 
      
      at Object.<anonymous> (packages/babel-core/test/parse.js:26:48)
```

I believe this actually requires us to update the `output.json` and is not actually a failure.

devenbansod:
Updated the tests. 
I believe all tests assertions in `babel-core` have been moved to jest expect. (Please point out if some have been left still)

devenbansod:
Not that I want to force this through (I understand maintainers are generally over-crowded with tasks). Please let me know if some action item is still pending and needs to be addressed to get this through.

packages/babel-core/test/fixtures/transformation/regenerator/default-parameters/exec.js
```diff
@@ -2,5 +2,5 @@ function* foo(bar = "bar") {
   return bar;
 }
 
-assert.deepEqual(foo().next().value, "bar");
-assert.deepEqual(foo("foo").next().value, "foo");
+expect(foo().next().value).toEqual("bar");
```
nicolo-ribaudo: `.toBe`?

packages/babel-core/test/parse.js
```diff
@@ -10,16 +9,20 @@ function fixture(...args) {
...
-      parse(input, { filename: fixture("input.js"), cwd: fixture() }),
-      output,
-    );
+    const output = require(fixture("output"));
+
+    const result = parse(input, {
+      filename: fixture("input.js"),
+      cwd: fixture(),
+    });
+    expect(JSON.parse(JSON.stringify(result))).toEqual(output);
```
nicolo-ribaudo: Why is `JSON.parse(JSON.stringify())` needed?

packages/babel-helper-transform-fixture-test-runner/src/index.js
```diff
@@ -22,6 +22,7 @@ const testContext = vm.createContext({
   transform: babel.transform,
   setTimeout: setTimeout,
   setImmediate: setImmediate,
+  expect,
```
nicolo-ribaudo: Can you also use `expect` in `./helpers`?

packages/babel-core/test/parse.js
danez: Because the returned AST returns classes, whereas the json file just has objects.

devenbansod: @danez had suggested [it](https://github.com/babel/babel/pull/7513#issuecomment-371194571) to fix the errors described [here](https://gist.github.com/devenbansod/2d70d53e9d920ad77462a867c6937c83).

packages/babel-helper-transform-fixture-test-runner/src/index.js
devenbansod: I had left it to be migrated when we are in a stage to completely migrate away from `assert`. Anyway, at most what I can change *directly* is `assertNoOwnProperties`. We can't migrate the other helper `assertArrayEquals` without running the code-mod at related places (I plan to do it a separate PR).

packages/babel-core/test/config-chain.js
```diff
@@ -384,7 +383,7 @@ describe("buildConfigChain", function() {
           comments: true,
         });
 
-        assert.equal(opts.comments, undefined);
```
danez: `toBe(undefined)` could be `toBeUndefined()`
also in other places

packages/babel-core/test/config-chain.js
```diff
@@ -533,7 +532,7 @@ describe("buildConfigChain", function() {
           fixture("nonexistant-fake", "other.js"),
         ],
       });
-      assert.equal(opts1, null);
```
danez: `toBe(null)` could be `toBeNull()`
also in other places

packages/babel-core/test/config-chain.js
devenbansod: Changed. (Y)

packages/babel-core/test/config-chain.js
devenbansod: Changed. Thanks for pointing it out.

packages/babel-core/test/api.js
```diff
@@ -266,19 +257,19 @@ describe("api", function() {
...
+    expect(result.code).toEqual("var x = function x(y) {\n  return y;\n};");
 
     // 2. passPerPreset: false
 
     aliasBaseType = null;
 
     result = execTest(false);
 
-    assert.equal(aliasBaseType, null);
+    expect(aliasBaseType).toBe(null);
```
devenbansod: Sorry, missed it here. Changing.

devenbansod:
I am not sure if the failure at https://travis-ci.org/babel/babel/jobs/351350222 is related. Can someone re-trigger it?

packages/babel-core/test/option-manager.js
```diff
@@ -160,23 +160,21 @@ describe("option-manager", () => {
           ],
         });
 
-        assert.equal(true, Array.isArray(options.plugins));
-        assert.equal(1, options.plugins.length);
-        assert.equal(0, options.presets.length);
+        expect(true).toBe(Array.isArray(options.plugins));
+        expect(1).toBe(options.plugins.length);
+        expect(0).toBe(options.presets.length);
```
nicolo-ribaudo: Can you reverse those conditions? (And use the appropriate `expect().*` functions)

Sorry for not spotting this in my earlier review.

devenbansod: Added these changes. Thanks for pointing out.

devenbansod:
Thanks a lot for your patience and support throughout this PR @danez @nicolo-ribaudo :-)

danez:
Thank you. Looking forward to the next PRs :)

PR #7516, opened 2018-03-07
when using regenerator in the es2015 preset transform async and asyncGenerators

pravdomil:
otherwise it will not produce es2015 compatible code?


mAAdhaTTah:
`async/await` is not in es2015. The es2015 preset transforms everything in es2015 to es5-compatible code.

pravdomil:
is there any preset for es2015+ to es5?

mAAdhaTTah:
You're looking for `babel-preset-env`.

PR #7531, opened 2018-03-08
Fix flowtype errors introduced in #7503.

loganfsmyth:
There was a new function param added in #7503 but it wasn't added in the plugins that override it.


existentialism:
👍

PR #7533, opened 2018-03-09
Fix bugs in the _wrapNativeSuper helper

### LINKED ISSUE #7506: Extending Builtins (HTMLElement) in @babel/core@7.0.0-beta.40 is not working
Bug Report

### Input Code
https://babeljs.io/repl/build/6204/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAEAqCmEAuB1AlgEwObyQYXCmngA8l4A7DGACVgFkAZAURHgFtKloBvAKD4BfPsACuyAPbtWHLhAB0GeADM0FeAAoAROWQBaAO6YcSLQBo4iVMdwFIEAJQBuPm24HoAXmgYJYzhRI8sAATvBg5DIBSNq6SIY2ps58QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=true&presets=es2015%2Cenv&prettier=false&targets=&version=7.0.0-beta.35%2Bpr.7020&envVersion=7.0.0-beta.35%2Bpr.7020

```js
class TestWidgetClass extends HTMLElement {

}
customElements.define("test-widget", TestWidgetClass);
let w = document.createElement("test-widget");
```

```js
//Config 
{
                    sourceMap: true,
                        "targets": {
                            "browsers": ["ie >= 11", "chrome >= 58", "safari >= 9", "firefox >= 47", "edge >= 38"]
                        }
                    }]]
           }
```

### Expected Behavior
I would expect this Custom Element to be created without an error.  

### Current Behavior
Instead I see an error in the console saying:
The result of constructing a custom element must be a HTMLElement (Safari)
OR 
Failed to construct 'CustomElement': The result must implement HTMLElement interface (Chrome)

### Possible Solution
It looks like _wrapNativeSuper has a bug.  It was adopted from code in another project, and it appears that code expected that the class it was wrapping had already been converted to ES5  by Babel.  Ultimately what looks like is happening is Reflect.construct is never called.

```js
// Two possible suggested fixes (not sure which is better):
function _wrapNativeSuper(Class) {
    if (typeof Class !== "function") {
        throw new TypeError("Super expression must either be null or a function");
    }
    if (typeof _cache !== "undefined") {
        if (_cache.has(Class))
            return _cache.get(Class);
        _cache.set(Class, Wrapper);
    }
    function Wrapper() {
      return _construct(Class, arguments, _gPO(this).constructor);
    }
    Wrapper.prototype = Object.create(Class.prototype, {
        constructor: {
            value: Wrapper,
            enumerable: false,
            writeable: true,
            configurable: true
        }
    });
    return _sPO(Wrapper, Class);
}

// OR 

function _wrapNativeSuper(Class) {
    if (typeof Class !== "function") {
        throw new TypeError("Super expression must either be null or a function");
    }
    if (typeof _cache !== "undefined") {
        if (_cache.has(Class))
            return _cache.get(Class);
        var Wrapper =  _sPO(function Super() {
          return _construct(Class, arguments, _gPO(this).constructor);
         }, Class);
        _cache.set(Class, Wrapper);
        return Wrapper;
    }
}
```

### Context
This prevents using Custom Elements in Babel 7.

### Your Environment
@babel/core": "^7.0.0-beta.40"
@babel/preset-env": "^7.0.0-beta.40"

### LINKED ISSUE #7527: Bug extending builtin object
bug Report

### Input Code
```js
class MyDate extends Date {
    constructor(time) {
        super(time);
    }
}
let myDate = new Date();
console.log(myDate.getTime());
```

### Output Code
```js
var _Symbol$iterator = require("@babel/runtime/core-js/symbol/iterator");

var _Symbol = require("@babel/runtime/core-js/symbol");

var _Object$create = require("@babel/runtime/core-js/object/create");

var _Map = require("@babel/runtime/core-js/map");

var _Reflect$construct = require("@babel/runtime/core-js/reflect/construct");

var _Object$setPrototypeOf = require("@babel/runtime/core-js/object/set-prototype-of");

var _Object$getPrototypeOf = require("@babel/runtime/core-js/object/get-prototype-of");

function _typeof(obj) { if (typeof _Symbol === "function" && typeof _Symbol$iterator === "symbol") { _typeof = function _typeof(obj) { return typeof obj; }; } else { _typeof = function _typeof(obj) { return obj && typeof _Symbol === "function" && obj.constructor === _Symbol && obj !== _Symbol.prototype ? "symbol" : typeof obj; }; } return _typeof(obj); }

function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }

function _possibleConstructorReturn(self, call) { if (call && (_typeof(call) === "object" || typeof call === "function")) { return call; } if (self === void 0) { throw new ReferenceError("this hasn't been initialised - super() hasn't been called"); } return self; }

function _inherits(subClass, superClass) { if (typeof superClass !== "function" && superClass !== null) { throw new TypeError("Super expression must either be null or a function"); } subClass.prototype = _Object$create(superClass && superClass.prototype, { constructor: { value: subClass, enumerable: false, writable: true, configurable: true } }); if (superClass) _Object$setPrototypeOf ? _Object$setPrototypeOf(subClass, superClass) : subClass.__proto__ = superClass; }

var _gPO = _Object$getPrototypeOf || function _gPO(o) { return o.__proto__; };

var _sPO = _Object$setPrototypeOf || function _sPO(o, p) { o.__proto__ = p; return o; };

var _construct = _typeof(Reflect) === "object" && _Reflect$construct || function _construct(Parent, args, Class) { var Constructor, a = [null]; a.push.apply(a, args); Constructor = Parent.bind.apply(Parent, a); return _sPO(new Constructor(), Class.prototype); };

var _cache = typeof _Map === "function" && new _Map();

function _wrapNativeSuper(Class) { if (typeof Class !== "function") { throw new TypeError("Super expression must either be null or a function"); } if (typeof _cache !== "undefined") { if (_cache.has(Class)) return _cache.get(Class); _cache.set(Class, Wrapper); } function Wrapper() {} Wrapper.prototype = _Object$create(Class.prototype, { constructor: { value: Wrapper, enumerable: false, writeable: true, configurable: true } }); return _sPO(Wrapper, _sPO(function Super() { return _construct(Class, arguments, _gPO(this).constructor); }, Class)); }

var MyDate =
/*#__PURE__*/
function (_Date) {
  _inherits(MyDate, _Date);

  function MyDate(time) {
    _classCallCheck(this, MyDate);

    return _possibleConstructorReturn(this, (MyDate.__proto__ || _Object$getPrototypeOf(MyDate)).call(this, time));
  }

  return MyDate;
}(_wrapNativeSuper(Date));

var myDate = new Date();
console.log(myDate.getTime());
```

```js
{
    "presets": [
        ["@babel/preset-env", {
            "targets": {
                "browsers": [
                    "iOS >= 9", "Android >= 4", "last 2 ChromeAndroid versions"
                ]
            },
            "useBuiltIns": false,
            "modules": false,
            "debug": false
        }],
        "@babel/preset-stage-0"
    ],
    "plugins": [
        [
            "@babel/plugin-transform-runtime", {
                "helpers": false,
                "polyfill": true,
                "regenerator": true
            }
        ]
    ]
}
```

### Expected Behavior
The code runs successfully and prints out the result without an error.

### Current Behavior
an error occurred while running in `ios<=9` and `android <= 4.x` browser environment：

```
ReferenceError: Can't find variable: Reflect
```

### Possible Solution
Use `typeof Reflect! == 'undefined'` to determine if `Reflect` is present.

### Your Environment

| ---------------- | -------
| Babel            |  7.0.0-beta.40
| node             | 8.9.4
| npm              | 5.6.0
| Operating System | mac os 10.13.3(17D102)

nicolo-ribaudo:
| Fixed Issues?            | Fixes #7506 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |


packages/babel-helpers/src/helpers.js
```diff
@@ -472,7 +472,7 @@ helpers.wrapNativeSuper = () => template.program.ast`
       constructor: {
         value: Wrapper,
         enumerable: false,
-        writeable: true,
```
existentialism: 😢

packages/babel-preset-env/test/fixtures/plugins-integration/issue-7527/options.json
```diff
@@ -0,0 +1,4 @@
+{
+  "plugins": [],
+  "presets": ["es2015"]
```
existentialism: Should this use `"../../../../lib"`?

nicolo-ribaudo: Right!

packages/babel-preset-env/test/fixtures/plugins-integration/issue-7527/options.json
```diff
@@ -0,0 +1,4 @@
+{
+  "plugins": [],
+  "presets": ["../../../../lib"]
```
Andarist: shouldnt it be possible to specify preset name here instead of the hardcoded path?

nicolo-ribaudo: I don't know, in the other tests of this package it is done like this.
We should modify our tests to accept _real_ plugin and preset names anyway (with `@babel/plugin-*`).

jridgewell:
While we're here, want to fix the `extends null` bug on [Line 461](https://github.com/babel/babel/pull/7533/files#diff-ba75bd4d116cc90ad8c8002e3010e821L461)?

jridgewell:
Test Case:

```js
let called = false;
window.Array = function() {
  called = true;
};

class List extends Array {
  push(value) {
    super.push(value);
    return this;
  }
}

assert.equal(called, true);
```

The issue is our `wrapper`, we've got too many.

```js
    function Wrapper() {
      // Inline Super's code here
      return _construct(Class, arguments, _gPO(this).constructor);
    }
    Wrapper.prototype = Object.create(Class.prototype, {
      constructor: {
        value: Wrapper,
        enumerable: false,
        writable: true,
        configurable: true,
      }
    });
    // Get rid of that Super function
    return _sPO(Wrapper, Class);
```

nicolo-ribaudo:
> While we're here, want to fix the extends null bug on Line 461?

This helper is only used when Babel knows that the extended class is a built-in class. It won't be used in any of these cases, so the helper doesn't need to be updated:
```js
class A extends null {}

var Array = null;
class B extends Array {}

class C extends (() => null) {}
```

jridgewell:
> This helper is only used when Babel knows that the extended class is a built-in class.

I can overwrite the Builtins. It's definitely an edge case.

```js
window.Array = null;
class Test extends Array {
}

File: packages/babel-preset-env/test/fixtures/plugins-integration/issue-7527/exec.js [hunk 4]
```diff
@@ -0,0 +1,6 @@
+class MyDate extends Date {
+  constructor(time) {
+      super(time);
+  }
+}
+let myDate = new MyDate();
```
jridgewell: Can we also add an assertion here? To make sure this is a real date instance.

Something like `assert.equal(typeof myDate.toString(), 'string');` should work.

nicolo-ribaudo:
I need some help with the failing tests... If I use node 4 and run `make test-only` I can see them fail, but If I transpile them and then I manually run the output file it works :thinking:

packages/babel-helpers/src/helpers.js
```diff
@@ -447,8 +447,15 @@ helpers.inheritsLoose = () => template.program.ast`
...
-    function _construct(Parent, args, Class) {
+  
+  // Here we don't use (Reflect.construct || function _construct(...) {})
+  // because, when running the exec tests in node 4, there is a buggy
+  // Reflect.construct implementation which ignores the third parameter.
+  // I don't know why, since Reflect shouldn't exists on node 4 (and it
+  // doesn't exist if you try to run the compiled code, it only exists in exec
+  // tests).
+  // @nicolo-ribaudo
+  function _construct(Parent, args, Class) {
```
jridgewell: Whoa, can’t do this. `Reflect.construct` cannot be properly polyfilled, but it’s necessary for non-constructable constructors (like `HTMLElement`).

packages/babel-preset-env/test/fixtures/plugins-integration/issue-7527/exec.js
```diff
@@ -0,0 +1,8 @@
...
+
+assert.equal(myDate.toString, Date.prototype.toString);
```
jridgewell: This is not equivalent to my proposed test. The method has to be executed, because their an internal branding check (that’s the actual test, not the `typeof`).

packages/babel-helpers/src/helpers.js
nicolo-ribaudo: I think that it won't work anyway:
```js
> function Wrapper() {
    return Reflect.construct(HTMLElement, [], Wrapper);
  }
  Wrapper.__proto__ = HTMLElement;
  Wrapper.prototype = Object.create(HTMLElement.prototype, { constructor: { value: Wrapper, configurable: true } })

-> HTMLElement {constructor: ƒ}

> new Wrapper()

-> VM195:2 Uncaught TypeError: Illegal constructor
     at new Wrapper (<anonymous>:2:18)
     at <anonymous>:1:1
   Wrapper @ VM195:2
   (anonymous) @ VM199:1
```

nicolo-ribaudo:
(https://github.com/babel/babel/pull/7533#discussion_r173670341 hasn't been resolved yet)

packages/babel-helpers/src/helpers.js
jridgewell: It's because you can't directly construct an `HTMLElement`, it has to go through the custom elements APIs.

```js
function Wrapper() {
  return Reflect.construct(HTMLElement, [], Wrapper);
}
Wrapper.__proto__ = HTMLElement;
Wrapper.prototype = Object.create(HTMLElement.prototype, { constructor: { value: Wrapper, configurable: true } })

customElements.define('x-w', Wrapper)

// no errors should be raised
w = document.createElement('x-w');
```

As opposed to using a `bind` constructor:

```js
function Binder() {
    var Constructor, a = [null];
    a.push.apply(a, arguments);
    Constructor = HTMLElement.bind.apply(HTMLElement, a);
    return Object.setPrototypeOf(new Constructor, Binder.prototype);
}
Binder.__proto__ = HTMLElement;
Binder.prototype = Object.create(HTMLElement.prototype, { constructor: { value: Binder, configurable: true } })

customElements.define('x-b', Binder)

// TypeError: Illegal constructor
b = document.createElement('x-b');
```

Kovensky: Back when this was first implemented, I had actually checked if node 4 had a Reflect.construct (because I know node 4 has buggy classes) and it did not have Reflect at all.

Is core-js maybe providing a buggy one? Or is the one they provide doing approximately the same thing as the helper, and thus would not work just like the helper wouldn't?

packages/babel-preset-env/test/fixtures/plugins-integration/issue-7527/exec.js
Kovensky: I think you need both.

packages/babel-helpers/src/helpers.js
jridgewell: It's [doing](https://github.com/zloirock/core-js/blob/master/modules/es6.reflect.construct.js) roughly what we are.

nicolo-ribaudo:
@jridgewell The test you proposed (which tests if the internal slots are correct) is failing on node 4. I think that it can't not fail :thinking: 

The code I tested in this screenshot is the `Reflect.construct` polyfill + `setPrototypeOf` hack to make `instanceof` work.

reflect-polyfill

packages/babel-helpers/src/helpers.js
```diff
@@ -448,14 +448,20 @@ helpers.wrapNativeSuper = () => template.program.ast`
...
-    a.push.apply(a, args);
-    Constructor = Parent.bind.apply(Parent, a);
-    return _sPO(new Constructor, Class.prototype);
-  }
+  var _construct = typeof Reflect === "object" && Reflect.construct
+    ? function _construct(Parent, args, Class) {
+        // This wrapper is needed because Reflect.construct can't be properly
+        // polyfilled, thus core-js doesn't set the correct __proto__.
+        var result = Reflect.construct(Parent, args, Class);
+        if (!result instanceof Class) result = _sPO(result, Class.prototype);
```
jridgewell: This seems like a core-js bug, not ours.

jridgewell:
Core-js's polyfill is buggy then. You can't construct proper date objects using `Function#call` (or `#apply`), it has to be a `new Ctor.bind.apply(Ctor, [null].concat(args))`.

/cc @zloirock to see if core-js can update.

packages/babel-helpers/src/helpers.js
```diff
@@ -447,17 +447,27 @@ helpers.inheritsLoose = () => template.program.ast`
...
-      Constructor = Parent.bind.apply(Parent, a);
-      return _sPO(new Constructor, Class.prototype);
-    };
+
+  var _construct = typeof Reflect === "object" && Reflect.construct
+    ? function _construct(Parent, args, Class) {
+        // This wrapper is needed because Reflect.construct can't be properly
+        // polyfilled, thus core-js doesn't set the correct __proto__.
+        var result = Reflect.construct(Parent, args, Class);
+        if (!(result instanceof Class)) result = _sPO(result, Class.prototype);
```
nicolo-ribaudo: > (By @jridgewell )
>
> This seems like a core-js bug, not ours.

It's something that `core-js` can't safely polyfill, but we can because we know that the super constructor is a built-in constructor, so its return value hasn't been overridden.

e.g.
```
function A() {
  return {};
}

function Wrapper() {
  return _construct(A, [], Wrapper);
}
Wrapper.prototype = Object.create(...);

// This should be false, but if core-js always did setPrototypeOf it would be true.
new Wrapper() instanceof Wrapper;
```

zloirock:
@jridgewell it's not a bug, it just can't be completely polyfilled, see [this logic](https://github.com/zloirock/core-js/blob/v3/packages/core-js/modules/es.reflect.construct.js). If you have some ideas how to improve this polyfill - feel free to propose it. `.bind` + `.setPrototypeOf` is not a solution and can't be used **inside** the polyfill because it will break `this` prototype chain in constructors.

jridgewell:
> .bind + .setPrototypeOf is not a solution and can't be used inside the polyfill because it will break this prototype chain in constructors.

Is this https://github.com/babel/babel/pull/7533#discussion_r174247273?

It can be solved with `instanceof`, no? Which reminds me, we should probably be doing this ourselves to guard against overridden Builtins.

```js
function _construct(Parent, args, Class) {
  var Constructor, a = [null];
  a.push.apply(a, args);
  Constructor = Parent.bind.apply(Parent, a);
  var c = new Constructor;
  return c instanceof Class ? _sPO(c, Class.prototype) : c;
}
```

zloirock:
@jridgewell I mean that it's not a solution for this polyfill, it can be used in helpers. About your code, try:
```js
class A {
  constructor() {
    this.foo(42);
  }
  foo() {
    console.log('A');
  }
}

class B extends A {
  foo() {
    console.log('B');
  }
}

Reflect.construct(A, [], B);

function _construct(Parent, args, Class) {
  var Constructor, a = [null];
  a.push.apply(a, args);
  Constructor = Parent.bind.apply(Parent, a);
  var c = new Constructor;
  return c instanceof Class ? _sPO(c, Class.prototype) : c;
}

_construct(A, [], B);
```

nicolo-ribaudo:
> ```js
> c instanceof Class ? _sPO(c, Class.prototype) : c
> ```

If `c` is an instance of `Class`, it's `__proto__` will already be `Class.prototype`.

jridgewell:
> `c instanceof Class ? _sPO(c, Class.prototype) : c`
> If c is an instance of Class, it's __proto__ will already be Class.prototype.

Sorry, that should have been `c instanceof Parent ? _sPO(c, Class.prototype) : c`

> _construct(A, [], B); // => 'A'

Given the corrected code (above), the dev console will still think this an `A`, but prototype inheritance makes it a `B`. I actually don't know why console thinks it's still an `A`, but I can't think of any observable side-effects.

zloirock:
> I actually don't know why console thinks it's still an `A`

Because constructor called before `setPrototypeOf`.

jridgewell:
But is it an observable difference anywhere other than dev console (where it has access to the internal slots of the object)?

As far as I know, the only way to properly create these non-callable Builtin constructors is to create the Parent, and run the Child's constructor on the result.

zloirock:
@jridgewell it's observable in many real use cases.

jridgewell:
Where, other than during `Parent`'s constructor?

```js
class Date {
  constructor() {
    // `this` is a Date, not a Sub
  }
}

class Sub extends Date {
  constructor() {
    super();
    // Now, it's a Sub
  }
}
```

Everything else should be the same.

Since it's impossible to properly polyfill `Reflect.construct`, it's a trade off of where we want it to break. For this PR, I'm only concerned with creating Builtins (specifically `HTMLElement` subclasses), so core-js is actually preventing the necessary changes.

zloirock:
So you propose to break `Reflect.construct` instead of improving helper?

zloirock:
In `core-js@3`, you will be able to detect incomplete implementation as `Reflect.construct.sham` and use a fallback like in `_construct`.

jridgewell:
They're both broken, though.

Native `Reflect.construct` is the only thing that'll accomplish extending `HTMLElement`, and any polyfill will fail to do it. Here, core-js attempting to polyfill actually prevents us from other static/runtime fixes (below the fold).

For other non-callable native constructors, only the bind approach will work. The babel helper can't fix it after the fact, because of brand checking during construction. So this one's requires a change to core-js.

For callable constructors, core-js' current approach is fine, and the helper doesn't need to change.

- - -

We could actually fix the bind polyfill using a combination static/runtime code. For instance:

```js
// Imagine Parent is not a builtin, just some class
class Parent {
  constructor() {
    // `this` is a Date, not a Sub
    this.overridableInit();
  }

  overridableInit() {
    console.log('Parent');
  }
}

class Sub extends Date {
  constructor() {
    super();
    // Now, it's a Sub
  }

  overridableInit() {
    console.log('Sub');
  }
}

// - - - - - -

// Global accessible "constructor"
Reflect.constructor = undefined;

function Parent() {
  // `this` is a Date, not a Sub
  const _this = this;
  const proto = Reflect.constructor && Reflect.constructor.prototype;
  (proto || this).overridableInit.call(this);
}

Parent.p.overridableInit = function() {
  console.log('Parent');
}

function Sub() {
  const old = Reflect.constructor;
  Reflect.constructor = Sub;
  const _this = _construct(Date, [], Sub);
  Reflect.constructor = old;
}

Sub.p.overridableInit() {
  console.log('Sub');
}
```

There really are no perfect solutions here.

nicolo-ribaudo:
Can I split this PR into two? (One with `writeable`->`writable` and the `typeof` fix; one with the `_wrapNativeSuper`) So that the second part doesn't block the first one.

zloirock:
@jridgewell you always can detect this behaviour at runtime. Anyway, what you propose is a breaking change, so it can't be added before `core-js@3`. As I wrote, in `core-js@3` you will be able to detect incomplete implementation like `Reflect.construct.sham`, so helper just can be changed to
```js
(typeof Reflect === "object" && Reflect.construct && !Reflect.construct.sham) ? Reflect.construct : _construct
```

jridgewell:
> Can I split this PR into two?

Sure.

zloirock:
cc @ljharb just interesting your opinion about it because `es6-shim` `Reflect.construct` implementation does not work with built-ins even with 2 arguments.

ljharb:
I don't really have the context here; but since Reflect.construct can't be shimmed, only shammed, it would be ideal if babel actually errored out on any cases where the behavior wouldn't end up correct (with or without a custom Reflect.construct implementation).

Kovensky:
Since `Reflect` is an API that's, in a way, meant to be used from proxies, a crazy idea could be to check for `typeof Proxy === 'function'` instead, and surely nobody shams `Proxy`...... right?

nicolo-ribaudo:
I tested the output of this code on node 4:
```js
class MyDate extends Date {
  constructor(time) {
      super(time);
  }
}
let myDate = new MyDate();

expect(myDate.toString).toBe(Date.prototype.toString);
expect(typeof myDate.toString()).toBe("string");
```

and the output of this code on Chrome (both wirh `Reflect.construct` and `Reflect.construct = undefined`):

```js
class CoolElement extends HTMLElement {
  connectedCallback () {
    this.innerText = 'Hello World';
  }
}

customElements.define('cool-element', CoolElement);
const c = document.createElement('cool-element');
```

They all seem to work, so this PR should be ok now (I hope :laughing: ). We should have a way of running exec tests on older platforms.

nicolo-ribaudo:
CI passes :heart_eyes:

packages/babel-helpers/src/helpers.js
```diff
@@ -434,16 +434,23 @@ helpers.construct = () => template.program.ast`
...
 
   export default function _construct(Parent, args, Class) {
-    if (typeof Reflect !== "undefined" && Reflect.construct) {
-      _construct = Reflect.construct;
+    if (
+      typeof Reflect !== "undefined" && Reflect.construct
+    ) {
+      _construct = function _construct(Parent, args, Class) {
+        // This wrapper is needed because Reflect.construct can't be properly
+        // polyfilled, thus core-js doesn't set the correct __proto__.
```
jridgewell: This fails to fix `Date` construction when Reflect.construct is polyfilled. We have to test for sham implementations, fall back to the bind `_construct` below when it's a sham.

https://babeljs.io/repl/build/7946/#?babili=false&browsers=&build=&builtIns=false&code_lz=PTAEBUAsEsGdTqAprWSB2AXaBDANngJ6gDGA9gE5IC0AVrAOSgBKSAZnkiZgHTnqxMFAK7dQABzJE20AgChWHLr36CRYgLyg2w9N2hl0ACnA4KAcySYANKDPnYt9EgDupi1YCUoAN5zQAaAAbmYSFGSYZKBazm72Vjzi4ZGYhOJIANz-gSEUCAKYOHpI0aAA8gBGtMp8VDiYSEZJEWSeWYHBoVSwwniYpQBiuvqGPDji4kR8-Hgm8Tb5gkUkSLb2sG3ZAVSYwhTooN29_QBkJ6Cp6WRshyjH0RpaDGRVykwA_Lc9faAAXIuFYpZAC-WTkJDwOFQoAAsoQACL1EpIAAeDXQABN4IiGr5sqohKJIhQjNgALZIbx-DoBHrpEnkyntUDAuSszj9MkIpGlWKw7kNIybcGGWBSJBjVBICiYIxcnESyIAZSE0HQ5geWgViWSETSirIKooavMm1U4slaBlpP111A8qRPGVqvVQs1oAARGoTR62kA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=true&presets=es2015&prettier=false&targets=&version=7.0.0-beta.46%2Bpr.7533&envVersion=

packages/babel-helpers/src/helpers.js
```diff
@@ -434,16 +434,23 @@ helpers.construct = () => template.program.ast`
...
       _construct = function _construct(Parent, args, Class) {
-        var a = [null];
+        var Constructor, a = [null];
         a.push.apply(a, args);
-        var Constructor = Parent.bind.apply(Parent, a);
-        var instance = new Constructor();
-        if (Class) setPrototypeOf(instance, Class.prototype);
-        return instance;
+        Constructor = Parent.bind.apply(Parent, a);
+        return setPrototypeOf(new Constructor, Class.prototype);
```
jridgewell: Where were these reverted? This supports `null` as a super class (other transforms use this helper).

packages/babel-helpers/src/helpers.js
nicolo-ribaudo: Is there another way other than try/catch?

jridgewell: [Something like this](https://babeljs.io/repl/build/7946/#?babili=false&browsers=&build=&builtIns=false&code_lz=PTAEBUAsEsGdTqAprWSB2AXaBDANngJ6gDGA9gE5IC0AVrAOSgBKSAZnkiZgHTnqxMFAK7dQABzJE20AgChWHLr36CRYgLyg2w9N2hl0ACnA4KAcySYANKDPnYt9EgDupi1YCUoAN5zQAaAAbmYSFGSYZKBazm72Vjzi4ZGYhOJIANz-gSEUCAKYOHpI0aAA8gBGtMp8VDiYSEZJEWSeWYHBoVSwwniYpQBiuvqGPDji4kR8-Hgm8Tb5gkUkSLb2sG3ZAVSYwhTooN29_QBkJ6Cp6WRshyjH0RpaDGRVykwA_Lc9faAAXIuFYpZAC-WTknH6sEgOAAtqU2Pg0FkhMQ_AEACL1JCJZIRNLYyIAZSE0HQ5mmBCMWxY7E43D4hjUokwRkxDVsAG0ALq2HR6bCGIzeHzAzzZTbA0j1EiQUBGJDC7JQ2GlITCTJyYFyVRSbF4MjmIzKmFtIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=true&presets=es2015&prettier=false&targets=&version=7.0.0-beta.46%2Bpr.7533&envVersion=) is the only foolproof way to detect if they're not using a `Binder` polyfill like we are.

core-js@3 added a `sham` property to the method, but that won't help anyone using an older version.

trusktr:
@nicolo-ribaudo Can you point me to the change that makes your test work?

trusktr:
I see, it uses `Reflect.construct` when available. So otherwise it is impossible to fix (f.e. the Date toString test)?

packages/babel-helpers/src/helpers.js
```diff
@@ -462,14 +487,17 @@ helpers.wrapNativeSuper = () => template.program.ast`
...
       if (typeof Class !== "function") {
         throw new TypeError("Super expression must either be null or a function");
       }
       if (typeof _cache !== "undefined") {
         if (_cache.has(Class)) return _cache.get(Class);
         _cache.set(Class, Wrapper);
       }
-      function Wrapper() {}
+      function Wrapper() {
+        return _construct(Class, arguments, _gPO(this).constructor)
```
nicolo-ribaudo: @trusktr The main bugfix is this line, which calls the super constructor.

packages/babel-helpers/src/helpers.js
```diff
@@ -433,14 +433,39 @@ helpers.setPrototypeOf = () => template.program.ast`
...
+      return false;
+    }
+  }
+
   export default function _construct(Parent, args, Class) {
-    if (typeof Reflect !== "undefined" && Reflect.construct) {
+    if (isNativeReflectConstruct()) {
       _construct = Reflect.construct;
     } else {
       _construct = function _construct(Parent, args, Class) {
```
nicolo-ribaudo: @trusktr If `Reflect.construct` doesn't exist we use this polyfill, which isn't 100% correct but works well for extending builtins.

PR #7535, opened 2018-03-09
Add `exact` Param To ObjectTypeAnnotations

wmonk:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | 
| Documentation PR         | 
| Any Dependency Changes?  |

Currently if you try to use `t.objectTypeAnnotation` it fails, because while there is documentation for the param `exact` it is not added in the `visitor` entry. This PR adds `exact` and an optionally validated param for `ObjectTypeAnnotation`.

I'm not sure if removing this was intentional, but the documentation is quite confusing if it was. I think it was probably just a mistake.

### DIFF
--- a/packages/babel-types/src/definitions/flow.js
+++ b/packages/babel-types/src/definitions/flow.js
@@ -7,6 +7,7 @@ import defineType, {
   validateOptional,
   validateOptionalType,
   validateType,
+  validateDefault,
 } from "./utils";
 
 defineType("AnyTypeAnnotation", {
@@ -250,15 +251,16 @@ defineType("NumberTypeAnnotation", {
 });
 
 defineType("ObjectTypeAnnotation", {
-  visitor: ["properties", "indexers", "callProperties"],
+  visitor: ["properties", "indexers", "callProperties", "builder"],
   aliases: ["Flow", "FlowType"],
+  builder: ["exact"],
   fields: {
     properties: validate(
       arrayOfType(["ObjectTypeProperty", "ObjectTypeSpreadProperty"]),
     ),
     indexers: validateOptional(arrayOfType("ObjectTypeIndexer")),
     callProperties: validateOptional(arrayOfType("ObjectTypeCallProperty")),
-    exact: validate(assertValueType("boolean")),
+    exact: validateDefault(assertValueType("boolean"), false),
   },
 });
 
--- a/packages/babel-types/src/definitions/flow.js
+++ b/packages/babel-types/src/definitions/flow.js
@@ -7,7 +7,6 @@ import defineType, {
   validateOptional,
   validateOptionalType,
   validateType,
-  validateDefault,
 } from "./utils";
 
 defineType("AnyTypeAnnotation", {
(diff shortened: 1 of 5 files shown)


packages/babel-types/src/definitions/flow.js
```diff
@@ -250,15 +250,15 @@ defineType("NumberTypeAnnotation", {
 });
 
 defineType("ObjectTypeAnnotation", {
-  visitor: ["properties", "indexers", "callProperties"],
+  visitor: ["properties", "indexers", "callProperties", "exact"],
```
nicolo-ribaudo: `visitors` should only contain `Node` properties, since it represents the properties which should be visited by babel-traverse.
What you need is `builder`, which represents the parameters passed to the `t.objectTypeAnnotation` function.

packages/babel-types/src/definitions/flow.js
```diff
@@ -250,15 +250,15 @@ defineType("NumberTypeAnnotation", {
...
   aliases: ["Flow", "FlowType"],
   fields: {
     properties: validate(
       arrayOfType(["ObjectTypeProperty", "ObjectTypeSpreadProperty"]),
     ),
     indexers: validateOptional(arrayOfType("ObjectTypeIndexer")),
     callProperties: validateOptional(arrayOfType("ObjectTypeCallProperty")),
-    exact: validate(assertValueType("boolean")),
+    exact: validateOptional(assertValueType("boolean")),
```
nicolo-ribaudo: Why is this optional? It should be required but defaulting to `false` (I don't remember if it is actually possible to set default properties value)

wmonk:
I've replaced this with builder, which i'm not sure is actually right when this get used in the `@babel/generator` module. Would be good to get some guidance 😃 

I've added a test now in `@babel/generator` that shows the correct outputted exact type.

nicolo-ribaudo:
You can see what I meant by "builder" here: https://github.com/wmonk/babel/blob/4fc40bdc21b96265c724daa50bb072316739ce58/packages/babel-types/src/definitions/core.js#L48-L49.

babel-generator was correct, you should revert its changes.

packages/babel-types/src/definitions/flow.js
```diff
@@ -250,15 +251,16 @@ defineType("NumberTypeAnnotation", {
...
   aliases: ["Flow", "FlowType"],
+  builder: ["exact"],
   fields: {
     properties: validate(
       arrayOfType(["ObjectTypeProperty", "ObjectTypeSpreadProperty"]),
     ),
     indexers: validateOptional(arrayOfType("ObjectTypeIndexer")),
     callProperties: validateOptional(arrayOfType("ObjectTypeCallProperty")),
-    exact: validate(assertValueType("boolean")),
+    exact: validateDefault(assertValueType("boolean"), false),
```
nicolo-ribaudo: I'd prefer this to be written as an object instead of using the `validateDefault` function, since `validator` and `default` have very little in common.

wmonk:
@nicolo-ribaudo thanks for pointing me in the right direction, makes sense now I think! The latest change should be correct now. I added a test to `babel/generator` anyway - think that's right?

wmonk:
😄 thanks!

Issue #7294, opened 2018-01-30
ESM that imports CJS can't be transpiled for both CJS and native ESM environments


jaydenseric:
Bug…

[This issue arises](https://github.com/jaydenseric/apollo-upload-server/issues/40#issuecomment-361271674) when attempting to create a package that:

- Transpiles both ESM (`.mjs`) for recent versions of Node.js run with `--experimental-modules` enabled, and CJS (`.js`) as a fallback.
- Imports a CJS package that contains `exports.__esModule`.

When imported CJS in `.mjs`, [Node.js provides a single default export representing the value of `module.exports`](https://nodejs.org/api/esm.html#esm_interop_with_existing_modules), so the syntax to use in source is:

```js
import graphql from 'graphql'
console.log(graphql.GraphQLScalarType) // Class
```

This syntax survives transpilation when [using `preset-env` with `modules: false`](https://github.com/jaydenseric/apollo-upload-server/blob/v4.0.2/.babelrc.js#L13) for [the `.mjs` target](https://github.com/jaydenseric/apollo-upload-server/blob/v4.0.2/package.json#L64) and works ok when running `--experimental-modules`. When transpiled with `modules: 'commonjs'` for the `.js` target, it creates syntax like this:

```js
'use strict'

var _graphql = _interopRequireDefault(require('graphql'))

function _interopRequireDefault(obj) {
  return obj && obj.__esModule ? obj : { default: obj }
}

console.log(_graphql.default.GraphQLScalarType) // Undefined
```

`_graphql.default.GraphQLScalarType` becomes causes an error as `_graphql.default` is undefined, because `_interopRequireDefault` found `__esModule` in the imported module and didn't wrap the exports under `default`.

Trying either of these approaches syntax in the source results in a functional `.js` target:

```js
import { GraphQLScalarType } from 'graphql'
console.log(GraphQLScalarType)
```

```js
import * as graphql from 'graphql'
console.log(graphql.GraphQLScalarType)
```

But breaks the `.mjs` target, because with `--experimental-modules` CJS can only be imported using default syntax.

The only workaround that seems to work in the source for both CJS and ESM targets:

```js
import graphqlDefault, * as graphqlExports from 'graphql'

const graphql = graphqlDefault || graphqlExports

console.log(graphql.GraphQLScalarType)
```

### Babel/Babylon Configuration (.babelrc, package.json, cli command)

`.babelrc.js`:

```js
module.exports = {
  presets: [
    [
      '@babel/env',
      {
[...]

PR #7545, opened 2018-03-10
Make `import`s in `.mjs` files use node-like behavior where 'exports' is '.default' only. 

### LINKED ISSUE #7518: Add an option for strict CJS interop (no destructuring-like behavior, just .default)
Seems like we should add an option for this to the CommonJS transform. We'd likely also want to enable it by default for `.mjs` files since that is what Webpack does AFAIK.

loganfsmyth:
| Fixed Issues?            | Fixes #7518
| Major: Breaking Change?  | Y, if people happen to already be using .mjs
| Documentation PR         | 
| Any Dependency Changes?  |

e.g.

```
// data.js - a standard CommonJS dfile
exports.named = "named";
```
then
```
import def, * as namespace from "./data";

console.log(def); // { named: "named" }
console.log(namespace); // { default: { named: "named" } }
```
whereas in `.js` files and in Babel 6, it was
```
import def, * as namespace from "./data";

console.log(def); // { named: "named" }
console.log(namespace); // { default: { named: "named" }, named: "named" }
```
so that
```
import { named } from "./data";
```
will work in Babel 6, but will not work in `.mjs` files in Babel 7.


packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js
```diff
@@ -107,6 +112,16 @@ export default function normalizeModuleAndLoadMetadata(
     }
 
     if (noInterop) metadata.interop = "none";
+    else if (esNamespaceOnly) {
+      // Both the default and namespace interops pass through __esModule
+      // objects, but the namespace interop is used to enable Babel's
+      // destructuring-like interop behavior for normal CommonJS.
+      // Since some tooling has started to remove that behavior, we expose
```
Kovensky: :tada:

Some people have actually _published_ `.mjs` with this behavior too.

jaydenseric:
Does this relate to, or fix https://github.com/babel/babel/issues/7294?

loganfsmyth:
@jaydenseric Tangentially in that it related to node ESM stuff, but I don't think this would fix that specifically.

PR #7547, opened 2018-03-11
Fix "ran" to "run" for monorepo.md

ngohoangthang:
I just want to improve the document. The verb should be "run" not ran because "run" has past participle is "run"


nicolo-ribaudo:
Thank you!

PR #7550, opened 2018-03-11
Replace lodash/map with array equivalent

danez:
| Fixed Issues?            | 
| Documentation PR         | 

I#m working on improving babel-standalone and noticed that this is totally unnecessary as `arr.map()` should be supported everywhere.

### DIFF
--- a/packages/babel-generator/src/node/whitespace.js
+++ b/packages/babel-generator/src/node/whitespace.js
@@ -1,4 +1,3 @@
-import map from "lodash/map";
 import * as t from "@babel/types";
 
 type WhitespaceObject = {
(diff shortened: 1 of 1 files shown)


packages/babel-generator/src/node/whitespace.js
```diff
@@ -220,7 +219,7 @@ export const list = {
    */
 
   VariableDeclaration(node: Object): Array<Object> {
-    return map(node.declarations, "init");
+    return node.declarations ? node.declarations.map(decl => decl.init) : [];
```
wtgtybhertgeghgtwtg: Is `declarations` nullable?

danez:
Looking at other parts of the code we already used node.declarations without checking null, so it probably isn't. Also codecoverage says that the else branch here is never triggered. I change it.

PR #3536, opened 2016-06-20
Correct super property access is loose mode

kriszyp:
In loose mode super property access is wrong, accessing and changing properties on the prototype instance instead of `this`:
https://phabricator.babeljs.io/T7389
This fixes super property access, by using fast/loose lookups of methods on calls, but reverts to spec handling for property access/assignments, where it is necessary to call the setter/getter on the correct `this` instance.

EDIT @aaronang:
Fixes #4312

PR #7553, opened 2018-03-12
Correctly assign super properties

### LINKED ISSUE #4312: In loose mode super setters/getters output is incorrect (T7389)
> Issue originally made by Kris Zyp (kriszyp)
### Bug information
- **Babel version:** 6.8.0
- **Node version:** 6.1.0
- **npm version:** 3.8.9
### Options

```
presets: ['es2015-loose']
```
### Input code

``` js
class Foo extends Bar {
    set foo(value) {
      super.foo = value
    }
  }
```
### Description

The setter produces:

`Bar.prototype.foo = value`

I know that loose mode isn't precisely spec equivalent, but the resulting code _never_ behaves even close to a true super assignment. A super assignment isn't supposed to assign to the proto object, it is supposed to do assignment to `this`, using a super setter function if available. I would recommend just leaving loose mode super property handling the same as spec mode, and calling `getSuperProperty` here (and getting rid of `getLooseSuperProperty`):
https://github.com/babel/babel/blob/master/packages/babel-helper-replace-supers/src/index.js#L189

### LINKED ISSUE #5769: class inheritance: uses `super` to access the properties of the parent class
The following code:

    class Point {
      getX() {
        console.log(this.x); // C
      }
    }
    
    class ColorPoint extends Point {
      constructor() {
        super();
        this.x = 2;
        super.x = 3;
        console.log(this.x)   // A
        console.log(super.x)  // B
      }
      
      m() {
        this.getX()
      }
    }
    
    const cp = new ColorPoint();
    cp.m();

On chrome 58.0.3029.110 64bit (V8 5.8.283.38), the result is: `3`, `undefined`, `3`.

But on babel 6.24.2, the result is: `2`, `undefined`, `2`.

[![][1]][1]

- Chrome 58.0.3029.110 64bit (V8 5.8.283.38)
- Babel Repl 6.24.2

links:

- https://stackoverflow.com/questions/44149248/es6-inheritance-uses-super-to-access-the-properties-of-the-parent-class
- [gist](https://gist.github.com/justjavac/82494287bb25cd671c781242bce975e1)
- [babel](http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015&targets=&browsers=&builtIns=false&debug=false&code=class%20Point%20%7B%0D%0A%20%20getX()%20%7B%0D%0A%20%20%20%20console.log(this.x)%3B%20%2F%2F%20C%0D%0A%20%20%7D%0D%0A%7D%0D%0A%0D%0Aclass%20ColorPoint%20extends%20Point%20%7B%0D%0A%20%20constructor()%20%7B%0D%0A%20%20%20%20super()%3B%0D%0A%20%20%20%20this.x%20%3D%202%3B%0D%0A%20%20%20%20super.x%20%3D%203%3B%0D%0A%20%20%20%20console.log(this.x)%20%20%20%2F%2F%20A%0D%0A%20%20%20%20console.log(super.x)%20%20%2F%2F%20B%0D%0A%20%20%7D%0D%0A%20%20%0D%0A%20%20m()%20%7B%0D%0A%20%20%20%20this.getX()%0D%0A%20%20%7D%0D%0A%7D%0D%0A%0D%0Aconst%20cp%20%3D%20new%20ColorPoint()%3B%0D%0Acp.m()%3B%0D%0A)

  [1]: https://i.stack.imgur.com/Y5gkp.png

danez:
| Fixed Issues?            | Fixes #3536 Fixes #4312 Fixes #5769 Fixes #5774
| Documentation PR         | 
| Any Dependency Changes?  |

This is a rebase of #3536 with additional fixes and tests. It ensures that we never assign anything to the superclass prototype and instead use the helper to do so.

### DIFF
--- a/packages/babel-helper-replace-supers/src/index.js
+++ b/packages/babel-helper-replace-supers/src/index.js
@@ -158,9 +158,12 @@ export default class ReplaceSupers {
     property: Object,
     value: Object,
     isComputed: boolean,
+    loose: boolean = false,
   ): Object {
     return t.callExpression(this.file.addHelper("set"), [
-      getPrototypeOfExpression(this.getObjectRef(), this.isStatic),
+      loose
+        ? t.memberExpression(this.superRef, t.identifier("prototype"))
+        : getPrototypeOfExpression(this.getObjectRef(), this.isStatic),
       isComputed ? property : t.stringLiteral(property.name),
       value,
       t.thisExpression(),
@@ -176,9 +179,13 @@ export default class ReplaceSupers {
    *
    */
 
-  getSuperProperty(property: Object, isComputed: boolean): Object {
+  getSuperProperty(
+    property: Object,
+    isComputed: boolean,
+    loose: boolean = false,
+  ): Object {
     let thisExpr = t.thisExpression();
-    if (this.inConstructor) {
+    if (this.inConstructor && !loose) {
       thisExpr = t.callExpression(
         this.file.addHelper("assertThisInitialized"),
         [thisExpr],
@@ -186,7 +193,9 @@ export default class ReplaceSupers {
     }
 
     return t.callExpression(this.file.addHelper("get"), [
-      getPrototypeOfExpression(this.getObjectRef(), this.isStatic),
+      loose
+        ? t.memberExpression(this.superRef, t.identifier("prototype"))
+        : getPrototypeOfExpression(this.getObjectRef(), this.isStatic),
       isComputed ? property : t.stringLiteral(property.name),
       thisExpr,
     ]);
@@ -206,6 +215,7 @@ export default class ReplaceSupers {
       return;
     } else if (t.isMemberExpression(parent) && !methodNode.static) {
       // super.test -> objectRef.prototype.test
+
       return t.memberExpression(
         t.cloneNode(superRef),
         t.identifier("prototype"),
@@ -221,19 +231,73 @@ export default class ReplaceSupers {
       return this.getLooseSuperProperty(node, path.parent);
     } else if (path.isCallExpression()) {
       const callee = node.callee;
-      if (!t.isMemberExpression(callee)) return;
-      if (!t.isSuper(callee.object)) return;
+      if (!isMemberExpressionSuper(callee)) return;
 
       // super.test(); -> objectRef.prototype.MethodName.call(this);
       t.appendToMemberExpression(callee, t.identifier("call"));
       node.arguments.unshift(t.thisExpression());
       return true;
+    } else if (
+      t.isUpdateExpression(node) &&
+      isMemberExpressionSuper(node.argument)
+    ) {
+      const binary = t.assignmentExpression(
+        node.operator[0] + "=",
+        node.argument,
+        t.numericLiteral(1),
+      );
+      if (node.prefix) {
+        // ++super.foo;
+        // to
+        // super.foo += 1;
+        return this.looseHandleAssignmentExpression(null, path, binary);
+      } else {
+        // super.foo++;
+        // to
+        // let _ref = super.foo; super.foo = _ref + 1;
+        const ref = path.scope.generateUidIdentifier("ref");
+        return this.looseHandleAssignmentExpression(ref, path, binary).concat(
+          // concat the reference in case the whole statement is within a returnStatement
+          t.expressionStatement(ref),
+        );
+      }
+    } else if (
+      t.isAssignmentExpression(node) &&
+      isMemberExpressionSuper(node.left)
+    ) {
+      return this.looseHandleAssignmentExpression(null, path, node);
+    } else if (isMemberExpressionSuper(node)) {
+      const isWritable = path.findParent(({ node }) => {
+        if (t.isAssignmentExpression(node) || t.isUpdateExpression(node)) {
+          return true;
+        }
+      });
+
+      // if any parent is a assignment or update convert to
+      // super.a.b = v; -> _get(objectRef.prottype, "a", this).b = v;
+      if (isWritable) {
+        return this.getSuperProperty(node.property, node.isComputed, true);
+      }
     }
+  }
+
+  looseHandleAssignmentExpression(ref, path, node) {
+    if (node.operator === "=") {
+      // super.name = "val"; -> _set(Object.getPrototypeOf(objectRef.prototype), "name", "val", this);
+      return this.setSuperProperty(
+        node.left.property,
+        node.right,
+        node.left.computed,
+        true,
+      );
+    }
+
+    return this.specHandleAssignmentExpression(ref, path, node);
   }
 
   specHandleAssignmentExpression(ref, path, node) {
     if (node.operator === "=") {
-      // super.name = "val"; -> _set(Object.getPrototypeOf(objectRef.prototype), "name", this);
+      // super.name = "val"; -> _set(Object.getPrototypeOf(objectRef.prototype), "name", "val", this);
       return this.setSuperProperty(
         node.left.property,
         node.right,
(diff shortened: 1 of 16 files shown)


danez:
I added 4 exec tests, which to my understanding should work. The work in chrome, but not with babel (loose and spec). Not sure what to do about it.

ljharb:
In that example, it should be hardcoded to Foo.prototype.x = 1, iirc.

danez:
Okay I found the issue. The set helper we are using was not doing anything when the property was not defined yet on one of the prototypes. If it is defined, then it would have properly set it. (https://github.com/babel/babel/blob/master/packages/babel-helpers/src/helpers.js#L602) So we should probably handle the case when the property was not found in the prototype chain. Chrome and node seem to call setters on the superclass, which is also what we are doing.

But now I have a different problem. It seems that update operations are not allowed?
```
class Foo {
}

class Test extends Foo {
  constructor() {
    super();
    super.test = 0;
    super.test++;
  }
}

const obj = new Test();
```

here `obj.test` is `NaN`, and when I log what `super.test` is after setting it in constructor it is undefined. (tested in node 8) It seems to me that accessing properties with `super.xxx` is always `undefined`.

danez:
I fixed the remaining tests and also added the test which checks that `super.x++` sets the property to NaN.

packages/babel-helpers/src/helpers.js
```diff
@@ -601,6 +601,15 @@ helpers.set = () => template.program.ast`
 
       if (parent !== null) {
         _set(parent, property, value, receiver);
+      } else if (Object.defineProperty) {
+        Object.defineProperty(receiver, property, {
+          value: value,
+          enumerable: true,
+          configurable: true,
+          writable: true
+        });
```
littledan: I don't understand why you're calling Object.defineProperty here. If the superclass prototype has a setter, it should be invoked.

packages/babel-helper-replace-supers/src/index.js
```diff
@@ -158,9 +158,12 @@ export default class ReplaceSupers {
     property: Object,
     value: Object,
     isComputed: boolean,
+    loose: boolean = false,
   ): Object {
     return t.callExpression(this.file.addHelper("set"), [
-      getPrototypeOfExpression(this.getObjectRef(), this.isStatic),
+      loose
+        ? t.memberExpression(this.superRef, t.identifier("prototype"))
+        : getPrototypeOfExpression(this.getObjectRef(), this.isStatic),
```
littledan: Seems like much of this PR is about making loose mode even looser. I'm wondering what the motivation is here--do you have any benchmarks showing that this version is faster? Or is this about consistency with other places where loose mode takes shortcuts?

littledan:
I'm not sure what @ljharb means, but super property access is based on statically looking up `Foo.prototype` and then *dynamically* finding the `__proto__` of that object. I agree that a setter in `Foo.prototype` should not be invoked, but I'm not sure if this patch is getting at that behavior in the right way--superclass setters should still be invoked, I believe.

There are tests in test262 (cc @leobalter) about super property access. See [my instructions](https://github.com/babel/babel/pull/7555#issuecomment-372633951) for how to run test262 on Babel.

packages/babel-helpers/src/helpers.js
danez: As far as I understand this code we are looking through the `prototype` chain and if we reach the end of that chain, then it means we haven't found the property nowhere so we are creating it.

danez:
I'm not quite sure what you are saying regarding the `getPrototypeOfExpression`. We should always use `Foo` instead of `Foo.prototype`? Or what is the "original value"?

packages/babel-helpers/src/helpers.js
littledan: Oh sorry, I missed the recursive loop. Seems reasonable then.

littledan:
For static methods, use `Foo`. For instance methods, use `Foo.prototype` (`getPrototypeOfExpression` already gets this right). However, `Foo.prototype` is actually a property access that may change over time. The ES spec is very specific that we're referring to the *initial* value of `Foo.prototype`, not the current value which may have changed.

danez:
After #7691 this completelly conflicts on every single line. I leave it open, but it has to be done from scratch.

jridgewell:
Oh man, I didn't realize there was already a PR working on it. I'll review this one to see what can be incorporated.

jridgewell:
The core functionality of this change is done in #7691. Both this and #7691 match on spec behavior (though #7691 includes many more tests around it).

There are a few remaining edge cases with loose mode:

- What is the expected behavior of getters in the proto chain?
  - Spec says to call them with the `receiver` instance as `this`.
  - Loose mode now just does `const value = Super.prototype.field`, which will call the getter with `Super.prototype` as `this`.
  - See [babel-plugin-transform-classes/test/fixtures/get-set/get-semantics-getter-defined-on-parent/exec.js](https://github.com/babel/babel/blob/5166eef103fd18bbb3f498cbf7065866d9c1ea83/packages/babel-plugin-transform-classes/test/fixtures/get-set/get-semantics-getter-defined-on-parent/exec.js)
- What is the expected behavior of setters in the proto chain?
  - Spec says to call them with the `receiver` instance as `this`.
  - Loose mode now just does `Super.prototype.field = value`, which will call with `Super.prototype` as the `this.
  - See [babel-plugin-transform-classes/test/fixtures/get-set/set-semantics-setter-defined-on-parent/exec.js](https://github.com/babel/babel/blob/5166eef103fd18bbb3f498cbf7065866d9c1ea83/packages/babel-plugin-transform-classes/test/fixtures/get-set/set-semantics-setter-defined-on-parent/exec.js)
- Where do we want the written value to live?
  - Spec says it gets defined on the `receiver`.
  - Loose mode now sets it on `Super.prototype`.
  - See [babel-plugin-transform-classes/test/fixtures/get-set/set-semantics-data-defined-on-parent/exec.js](https://github.com/babel/babel/blob/5166eef103fd18bbb3f498cbf7065866d9c1ea83/packages/babel-plugin-transform-classes/test/fixtures/get-set/set-semantics-data-defined-on-parent/exec.js)
- How do we want to handle a setter/non-writable defined on the instance?
  - Spec says to throw, since it's not a writable property
  - Loose mode now doesn't even encounter it, since it'll set on the `Super.prototype`
  - setter: [babel-plugin-transform-object-super/test/fixtures/get-set/set-semantics-not-defined-on-parent-setter-on-obj/exec.js](https://github.com/babel/babel/blob/5166eef103fd18bbb3f498cbf7065866d9c1ea83/packages/babel-plugin-transform-object-super/test/fixtures/get-set/set-semantics-not-defined-on-parent-setter-on-obj/exec.js)
  - non-writable: [babel-plugin-transform-object-super/test/fixtures/get-set/set-semantics-not-defined-on-parent-data-non-writable-on-obj/exec.js](https://github.com/babel/babel/blob/5166eef103fd18bbb3f498cbf7065866d9c1ea83/packages/babel-plugin-transform-object-super/test/fixtures/get-set/set-semantics-not-defined-on-parent-data-non-writable-on-obj/exec.js)


I propose the following:

- Gets use simple `Super.prototype.field` lookup
  - including the non-spec behavior of getters being called with `Super.prototype`.
- Sets be rewritten to:
  - `super.field = value` -> `this.field = value`
  - `super.field += 1` `ref = Number(Super.prototype.field), this.field = ref + 1`
  - `super.field++` `ref = Number(Super.prototype.field), this.field = ref + 1, ref`
  - `++super.field` `ref = Number(Super.prototype.field), this.field = ref + 1`

danez:
I guess one problem is nested objects, so for example `super.field.subfield++` in that case i did in this PR `helpers.get(... 'field').subfield++`, not sure if that would work with this.field when the parent class has the field?

All the tests in this PR have exec parts which should at least work, I checked all of them with chrome/v8.

And this are the issues i tried to solve: #3536 #4312 #5769 #5774

PR #7556, opened 2018-03-12
Re-add TEST_ONLY and use Jest's -t for TEST_GREP.

loganfsmyth:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

In #7455 we lost TEST_ONLY, but some of the docs still reference it.

I also changed TEST_GREP to use Jest's `-t`, which I think is more consistent with the behavior before #7455.

I actually successfully did something with Bash. Amazing.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -1,20 +1,26 @@
 #!/bin/bash
 set -e
 
-if [ -z "$TEST_GREP" ]; then
-   TEST_GREP=""
-fi
-
 node="node"
-jestArgs=""
+jestArgs=()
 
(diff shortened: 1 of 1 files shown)


danez:
But this also means TEST_ONLY needs to match the complete package name and cannot be a simple subset. 
But that's okay with me.

loganfsmyth:
Oh I missed that TEST_ONLY could actually match a part of the package name, woops!

PR #7571, opened 2018-03-15
Remove outdated spec deviation note

benwiley4000:
`JSXText` was added to the JSX spec [here](https://github.com/facebook/jsx/pull/80).


hzoo:
Cool, thanks for your contribution!

PR #7578, opened 2018-03-15
Rename actual/expected test files to input/output

CodingItWrong:
| Fixed Issues?            |
| Patch: Bug Fix?          | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         |
| Any Dependency Changes?  |

Several test files appear to have been missed in the update of test file naming from actual/expected.js to input/output.js. As a result, they were silently not being run. I've confirmed that they were not running, updated the names, then confirmed that they are running now.


nicolo-ribaudo:
Thank you!

PR #7549, opened 2018-03-11
Migrate `babel-cli` and `babel-generator` tests to use jest-expect

devenbansod:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Working towards #7476 
| Patch: Bug Fix?          | NA
| Major: Breaking Change?  | NA
| Minor: New Feature?      | NA
| Tests Added + Pass?      | 👍
| Documentation PR         | NA
| Any Dependency Changes?  | NA
| License                  | MIT

* Migrate babel-cli tests to use jest-expect
* Migrate babel-generator tests to use jest-expect

PR #7579, opened 2018-03-15
Migrate a few packages' tests to use Jest Expect (see below)

devenbansod:
| Fixed Issues?            | Works towards #7476  
| Major: Breaking Change?  | NA
| Minor: New Feature?      | NA

* Migrate the following packages' tests:
    * babel-helper-annotate-as-pure
    * babel-helper-module-imports
    * babel-helper-transform-fixture-test-runner
    * babel-highlight
    * babel-node
    * babel-plugin-transform-modules-commonjs
    * babel-preset-env-standalone
    * babel-preset-env
    * babel-preset-es2015
    * babel-preset-react
    * babel-standalone
    * babel-template
    * babel-traverse
    * babel-types

* Some of the fixtures still need to be updated, I'll update them soon


packages/babel-highlight/test/index.js
```diff
@@ -57,20 +56,20 @@ describe("@babel/highlight", function() {
...
+        expect(shouldHighlight({})).toBeTruthy();
       });
     });
 
     describe("when colors are not supported", function() {
       stubColorSupport(false);
 
       it("returns false", function() {
-        assert.ok(!shouldHighlight({}));
+        expect(!shouldHighlight({})).toBeTruthy();
```
nicolo-ribaudo: Does something like `.toBeFalsy` exist?

packages/babel-node/test/index.js
```diff
@@ -73,15 +65,9 @@ const assertTest = function(stdout, stderr, opts) {
...
   if (opts.stdout) {
     if (opts.stdoutContains) {
-      assert.ok(
-        includes(stdout, expectStdout),
-        "stdout " +
-          JSON.stringify(stdout) +
-          " didn't contain " +
-          JSON.stringify(expectStdout),
-      );
+      expect(includes(stdout, expectStdout)).toBeTruthy();
```
nicolo-ribaudo: Can we somehow keep the old error message?

packages/babel-traverse/test/inference.js
```diff
@@ -48,7 +47,7 @@ describe("inference", function() {
 
       const strictMatch = left.baseTypeStrictlyMatches(right);
 
-      assert.ok(!strictMatch, "type might change in if statement");
+      expect(!strictMatch).toBeTruthy();
```
nicolo-ribaudo: `toBeFalsy()`

packages/babel-traverse/test/inference.js
```diff
@@ -59,255 +58,202 @@ describe("inference", function() {
...
       const path = getPath("new T")
         .get("body")[0]
         .get("expression");
       const type = path.getTypeAnnotation();
-      assert.ok(
+      expect(
         t.isGenericTypeAnnotation(type) && type.id.name === "T",
-        "should be T",
-      );
+      ).toBeTruthy();
```
nicolo-ribaudo: Can you split this in two assertions?

packages/babel-traverse/test/inference.js
```diff
@@ -59,255 +58,202 @@ describe("inference", function() {
...
     });
     it("should infer number from ++x", function() {
       const path = getPath("++x")
         .get("body")[0]
         .get("expression");
-      assert.ok(
-        t.isNumberTypeAnnotation(path.getTypeAnnotation()),
-        "should be number",
-      );
+      expect(t.isNumberTypeAnnotation(path.getTypeAnnotation())).toBeTruthy();
```
nicolo-ribaudo: Instead of `expect(t.isFooBar(node)).toBeTruthy()`, can we use `expect(node.type).toBe("FooBar")`? (Also in other places)

packages/babel-traverse/test/scope.js
```diff
@@ -30,175 +29,161 @@ function getIdentifierPath(code) {
...
-      );
+      ).toBeUndefined();
     });
 
     it("variable constantness", function() {
-      assert.ok(getPath("var a = 1;").scope.getBinding("a").constant === true);
-      assert.ok(
+      expect(
+        getPath("var a = 1;").scope.getBinding("a").constant === true,
+      ).toBeTruthy();
```
nicolo-ribaudo: `.toBe(true)` instead of `=== true`

packages/babel-traverse/test/scope.js
```diff
@@ -30,175 +29,161 @@ function getIdentifierPath(code) {
...
-      assert.ok(
+      ).toBeTruthy();
+      expect(
         !getPath("let a = 1; `${a++}`")
           .get("body")[1]
           .get("expression")
           .isPure(),
-      );
-      assert.ok(
+      ).toBeTruthy();
```
nicolo-ribaudo: `.toBeFalsy()`

packages/babel-traverse/test/scope.js
```diff
@@ -30,175 +29,161 @@ function getIdentifierPath(code) {
...
+      expect(
         !getPath("tagged`foo`")
           .get("body")[0]
           .get("expression")
           .isPure(),
-      );
-      assert.ok(
+      ).toBeTruthy();
```
nicolo-ribaudo: `.toBeFalsy()`

packages/babel-traverse/test/scope.js
```diff
@@ -30,175 +29,161 @@ function getIdentifierPath(code) {
...
     });
 
     test("label", function() {
-      assert.strictEqual(
-        getPath("foo: { }").scope.getBinding("foo"),
-        undefined,
-      );
-      assert.strictEqual(
-        getPath("foo: { }").scope.getLabel("foo").type,
+      expect(getPath("foo: { }").scope.getBinding("foo")).toBe(undefined);
```
nicolo-ribaudo: `.toBeUndefined()`

packages/babel-traverse/test/scope.js
```diff
@@ -30,175 +29,161 @@ function getIdentifierPath(code) {
...
+      expect(getPath("foo: { }").scope.getLabel("foo").type).toBe(
         "LabeledStatement",
       );
-      assert.strictEqual(
-        getPath("foo: { }").scope.getLabel("toString"),
-        undefined,
-      );
+      expect(getPath("foo: { }").scope.getLabel("toString")).toBe(undefined);
```
nicolo-ribaudo: Also here

packages/babel-traverse/test/traverse.js
```diff
@@ -81,22 +80,24 @@ describe("traverse", function() {
...
 
-    assert.ok(!traverse.hasType(ast, "ThisExpression", ["MemberExpression"]));
-    assert.ok(!traverse.hasType(ast, "ThisExpression", ["Program"]));
+    expect(
+      !traverse.hasType(ast, "ThisExpression", ["MemberExpression"]),
+    ).toBeTruthy();
+    expect(!traverse.hasType(ast, "ThisExpression", ["Program"])).toBeTruthy();
 
-    assert.ok(!traverse.hasType(ast, "ArrowFunctionExpression"));
+    expect(!traverse.hasType(ast, "ArrowFunctionExpression")).toBeTruthy();
```
nicolo-ribaudo: `.toBeFalsy()`

packages/babel-types/test/asserts.js
```diff
@@ -17,18 +16,9 @@ describe("asserts", () => {
...
-            return (
-              err instanceof Error &&
-              err.message ===
-                `Expected type "${nodeType}" with option {}, but instead got "FlavorTownDeclaration".`
-            );
-          },
-        );
+        expect(() => {
+          t[k]({ type: "FlavorTownDeclaration" }, {});
+        }).toThrow();
```
nicolo-ribaudo: Can we test the error message?

packages/babel-highlight/test/index.js
Kovensky: it does

packages/babel-node/test/index.js
Kovensky: `expect(stdout).stringContaining(expectStdout)`

packages/babel-traverse/test/inference.js
Kovensky: Maybe we should add a custom expectation for babel nodes.

`expect(node).nodeOfType("FooBar")`? https://facebook.github.io/jest/docs/en/expect.html#expectextendmatchers

packages/babel-types/test/asserts.js
Kovensky: Yes; you can give a constructor to get an `instanceof` check, a regexp to test against the message, or a string to `===` the message (without name).

packages/babel-node/test/index.js
devenbansod: I don't know if `stringContaining` will work out of the box here, since I believe we are using `lodash.includes` because stdout is a String in some cases and an object in some.

packages/babel-traverse/test/inference.js
devenbansod: I tried doing [this](https://gist.github.com/devenbansod/e519bde4b2e58bbc74ea4ace732c8d22), but it seems Jest has some problem running with Node4 (works perfectly in Node6+).

We are facing a similar problem with Node4 in adding a new extension to Jest in #7549 too. Any help in this regard would be appreciated.

devenbansod:
Can some one please restart the job at https://travis-ci.org/babel/babel/builds/354156525?utm_source=github_status&utm_medium=notification? This seems to be an unrelated timeout error, but please correct me if I am wrong here.

devenbansod:
Hi, I'm struggling to figure out the cause of the failure at https://travis-ci.org/babel/babel/jobs/357726707#L1144

Moreover, I seem to be getting the same error on re-cloning the project and running the tests on master, but the CI on master seems to be green somehow.

Any help/pointers would be really appreciated.

packages/babel-helper-transform-fixture-test-runner/src/helpers.js
```diff
@@ -11,5 +11,3 @@ export function assertLacksOwnProperty() {}
 export function multiline(arr) {
   return arr.join("\n");
 }
-
-export const assertArrayEquals = assert.deepEqual;
```
hzoo: technically a breaking change but we didn't intend for anyone to use this package

nicolo-ribaudo: Since this PR will land in Babel 7, I think we could even remove also the empty functions from this file.

@devenbansod You missed another usage of `assert` in this file :wink:

xtuc:
Lgtm so far. That looks like some good hard work there. Thanks.

packages/babel-helper-transform-fixture-test-runner/src/helpers.js
devenbansod: The methods `assertHasOwnProperty` and `assertLacksOwnProperty` seem to be used in `babel-preset-es2015/test/fixtures/traceur/Classes/InheritanceFromMemberExpression.js` and `babel-preset-es2015/test/fixtures/traceur/Classes/MethodInheritance.js`, though actually since they are empty so are really not testing anything.

Can we replace with following:
```
export function assertHasOwnProperty() {}
```
by 
```
export function assertHasOwnProperty(obj, propertyName) {
     expect(obj).toHaveProperty(propertyName);
}
```

I can't think of functionally equivalent of `assertHasNoOwnProperty` though. Any ideas?
OR should we just remove the usages of these functions?

@nicolo-ribaudo thanks for pointing that out. Just replaced that usage with expect.

packages/babel-preset-es2015/test/fixtures/traceur/NumericLiteral/Simple.js
```diff
@@ -84,18 +84,18 @@
...
-  expect(C.hasOwnProperty('5')).toBe(true);;
-  expect(C.hasOwnProperty('9')).toBe(true);;
-  expect(C.hasOwnProperty('10')).toBe(true);;
-  expect(C.hasOwnProperty('11')).toBe(true);;
+  expect(C.hasOwnProperty('3')).toBe(true);
+  expect(C.hasOwnProperty('4')).toBe(true);
+  expect(C.hasOwnProperty('5')).toBe(true);
+  expect(C.hasOwnProperty('9')).toBe(true);
+  expect(C.hasOwnProperty('10')).toBe(true);
+  expect(C.hasOwnProperty('11')).toBe(true);
```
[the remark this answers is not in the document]
devenbansod: I had tried to implement one of our own extensions as mentioned in https://github.com/babel/babel/pull/7579#discussion_r175000256 which seems to fail on Node v4. I'm not sure if we can find a workaround (I haven't, in the last few days that I tried to search :-( )

packages/babel-helper-transform-fixture-test-runner/src/helpers.js
nicolo-ribaudo: > ```
> export function assertHasOwnProperty() {}
> ```
>
> by
>
> ```
> export function assertHasOwnProperty(obj, propertyName) {
>      expect(obj).toHaveProperty(propertyName);
> }
> ```

You can directly use expect().toHaveProperty() in those tests, without the `assertHasOwnProperty` function.

> I can't think of functionally equivalent of assertHasNoOwnProperty though. Any ideas?
OR should we just remove the usages of these functions?

You can use something like this:
```js
expect(Object.getOwnPropertyNames(obj)).toHaveLength(0);
```

packages/babel-preset-es2015/test/fixtures/traceur/NumericLiteral/Simple.js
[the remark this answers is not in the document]
nicolo-ribaudo: I just noticed that `expect().toHaveProperty` only checks own properties, so it can be used here.

packages/babel-plugin-transform-classes/test/fixtures/extend-builtins/imported_babel-plugin-transform-builtin-classes/exec.js
```diff
@@ -11,24 +11,30 @@ class List extends Array {
...
 class SecondLevel extends List {
   method() {
     return this;
   }
 }
 
-assert.ok(new SecondLevel(1) instanceof SecondLevel, 'new SecondLevel is an instanceof SecondLevel');
-assert.ok(new SecondLevel(2) instanceof List, 'new SecondLevel is an instanceof List');
-assert.ok(new SecondLevel(3) instanceof Array, 'new SecondLevel is an instanceof Array');
+expect(new SecondLevel(1) instanceof SecondLevel).toBe(true);
```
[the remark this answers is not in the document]
devenbansod: I just realized we can use this instead: https://facebook.github.io/jest/docs/en/expect.html#tobeinstanceofclass

devenbansod:
So, again Jest throws up a different behavior for `not.instanceOf` on Node v4 *only*, while `expect(b instanceof Greeting).toBe(false);` passes.

nicolo-ribaudo:
Ok, we can change it when we won't support node 4 anymore (I guess by the end of april, since it won't be maintained anymore).

devenbansod:
Thanks everyone for the patience and support throughout this PR. :-)

nicolo-ribaudo:
The yarn.lock it's ok. Thank you for your hard work! :)

PR #7581, opened 2018-03-15
WIP: Bundle @babel/standalone with rollup

danez:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  | rollup stuff

This is still very early and super messy right now. Also it does not yet work completely as long as rollup/rollup#2059 isn't released. Locally I got the tests already passing.

So far it looks like the bundle size will decrease (I might have some more ideas for example excluding `debug` in the bundle):
**babel.js** `2.7MB` -> `1.5MB`
**babel.min.js** `1.4MB` -> `0.8MB`

Depends on :
- ~~rollup/rollup#2059~~
- https://github.com/rollup/rollup-pluginutils/pull/28


Gulpfile.js
```diff
@@ -92,33 +94,182 @@ function buildBabel(exclude) {
...
+                          path.join(packageFolder, browserFile),
+                          "UTF-8"
+                        );
+                      }
+                    }
+                  }
+                }
+              },
+              resolveId(importee) {
+                const matches = importee.match(/^@babel\/([^/]+)$/);
```
Andarist: I've created https://github.com/Andarist/lerna-alias pretty much for this exact thing - aliasing to respective `src`.

Gulpfile.js
```diff
@@ -92,33 +94,182 @@ function buildBabel(exclude) {
...
+                    __dirname,
+                    "packages",
+                    `babel-${matches[1]}`
+                  );
+                  const packageJson = require(path.join(
+                    packageFolder,
+                    "package.json"
+                  ));
+
+                  const filename =
```
Andarist: does any of our packages use `browser` field that should be resolved here? `rollup-plugin-node-resolve` handles `browser` field if u pass `browser: true` as an option to it

danez: The problem is that the browser field is using /lib/ and not /src/ so the node plugin won't pick it up.

Gulpfile.js
danez: But it does not replace lib with src in theain entry?

Gulpfile.js
Andarist: Oh, ok - I've searched for a simple "browser" fields with string value, I see we have some object maps there in i.e. `@babel/core`.

I think it might be possible to workaround this easier - rename our browser files to have `.browser.js` extension and make `rollup-plugin-node-resolve` to pick it up instead of the original one by passing `extensions: ['.browser.js', '.js']` option

Gulpfile.js
Andarist: It does exactly this (or I've misunderstood your question) - it searches for packages managed by lerna (using lerna's algo - although it's copied) and creates an alias map from `PKG_NAME` to `PKG_PATH/src`

Gulpfile.js
loganfsmyth: I'd assume it is fine for the `browser` field map both `src` and `lib` for Babel at least. Would that work? e.g.
```
  "browser": {
    "./lib/config/files/index.js": "./lib/config/files/index-browser.js",
    "./lib/transform-file.js": "./lib/transform-file-browser.js",
    "./lib/transform-file-sync.js": "./lib/transform-file-sync-browser.js"
    "./src/config/files/index.js": "./src/config/files/index-browser.js",
    "./src/transform-file.js": "./src/transform-file-browser.js",
    "./src/transform-file-sync.js": "./src/transform-file-sync-browser.js"
  },
```

Andarist: yeah, this definitely should work too 👍

Andarist:
What's the status of this PR? Is this still something we'd like to pursue? I could work on rebasing this and taking this through the finish line.

danez:
It is finally working.
The savings for babel-standalone is 497kB vs 1.0 MB
babel-preset-env-standalone is 302kB vs 320kB

I wonder why babel-standalone is 50% smaller. In the repl everything works correctly, besides typescript (but typescript is also not working without this PR for me.).

I will now cleanup the PR and do some more stats.

Andarist:
Do we intend to support IE11 in the repl? If yes then would be great to check if its working there.

> I wonder why babel-standalone is 50% smaller. 

It depends on the previous setup, I've seen such reductions before when migrating from webpack to rollup (well, maybe not that big ones, but I've seen 30-40% reductions). If you are curious about it then the easiest way to find out would be to compare what was in the final bundle before and what is now - source map explorer could come in handy for that.

JLHwung:
Closing it in favour of #10779.

@danez Thanks for your previous work on Rollup bundling, from which I learned and benefited a lot when working on #10779.

PR #7583, opened 2018-03-16
Allow placeholders in JSXElements when parsing templates

Andarist:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR         | 
| Any Dependency Changes?  |


xtuc:
Ah sorry, could we update the doc? that's an interesting feature.

xtuc:
Yes, you are right, it's a pretty obvious feature. I was just thinking about adding an example but i'm fine either ways.

Issue #7208, opened 2018-01-12
Babel 7 "SyntaxError: Unexpected token import" in lerna environment


Raigen:
We use lerna to seperate concernes into multiple packages. One of them is depending on two others to combine them in the final build. For production we build them independently and then put the dependencies into the `node_modules` directory of the first package.
In development we do nothing special because babel-node from the first package seems to take care of the `import` statements also from the other packages.
After I updated to babel 7 for testing purpose babel does not transform the other packages anymore.
I created two repositories with a minimal reproducer of the environment, one with babel6 where the `import` works and one with babel 7 dependencies where it does not work anymore.

Using `@babel/register` here https://github.com/Raigen/babel7-tmp/blob/master/packages/pck2/index.js#L4 does not work either.

I am out of ideas right now. I expect this to work because babel itself is a monorepo with lerna and `import`-dependencies inside `./packages`. But I can not get it to work even in the simple reproducer repos I created.

### Input Code
https://github.com/Raigen/babel6-tmp
https://github.com/Raigen/babel7-tmp

### Expected Behavior
https://github.com/Raigen/babel6-tmp

### Current Behavior
https://github.com/Raigen/babel7-tmp

### Context
We plan to refactor one of our packages with typescript. For this purpose we need to figure out how the development environment can work with one package on typescript and the other packages still on babel. And the best idea until now is to keep babel everywhere and move to babel 7 with the typescript preset. This does work for all packages running independently and in the build, but not when I try to run the one package that `import`s from the others.
Also we of course want to keep our dependencies updated, so even without typescript we would want to upgrade to babel 7 when it is final.

### Your Environment

| software         | version(s)
| ---------------- | -------
| Babel            |  7.0.0-beta.37
| node             | 8.9.4
| npm              | 5.6.0
| Operating System | tested on Windows 10 and Mac OS High Sierra

Andarist:
You need to tweak your `--ignore` arg, babel@7 by default ignores things in `node_modules` and your lerna packages are linked inside it.

Raigen:
First of all, thank you for the fast answer!

I tweaked around with `--ignore` (more like trial and error) and found out some things.
First of all, it seems that it does **not** compiles `pck2` from the `node_modules` but from `../` as we can see, when we have a closer look into the error message
```
C:\Users\foellerich\Git\babel7-tmp\packages\pck2\server.js:1
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #6134, opened 2017-08-18
Infra: build/test master before publish


hzoo:
Similar to smoke tests: https://github.com/babel/babel/issues/1837

Basically we should setup a test such that Babel itself would be able to compile itself with the latest built version in master (currently 7.0). Then we can find basic issues without having to publish first. Basic idea would be to just write so some temporary folders and then link something?

It would need to be able to revert back if something fails since otherwise it wouldn't even be able to compile anymore.

---

Context: Babel uses Babel to compile itself (we write using ES6+ and compile to ES5 for npm). However do that, we download the last version of Babel on npm. This also means that we have to publish to npm before testing that last published version on Babel itself. Ideally we would have a setup where before we publish (or on every PR if we wanted), we could switch out compiling Babel with the version of itself on master. Maybe it means doing an npm pack/link etc, but we should also check that it works with the monorepo/yarn/lerna setup that we have?

ljharb:
You should definitely be able to use `npm link`; or even `file://./` or similar. Not sure about working with lerna, but it should work.

Worst case, you could set up a separate package folder like `babel-test`, that on `postinstall` does the link/copy, and imports and re-exports babel.

jkarttunen:
Test scripts maybe need to edit package.json?

PR #7588, opened 2018-03-17
Have @babel/core lazy-load all dependencies and make @babel/register not explode because of that

### LINKED ISSUE #5702: It takes 380ms to print the version string
I'm running Babel 6.24.1 on a Retina MacBook Pro, with a 2.9 GHz Intel Core i5 processor and 16 GB of RAM.

It takes 380ms to print the version string:

```
$ time ./node_modules/.bin/babel --version
6.24.1 (babel-core 6.24.1)
./node_modules/.bin/babel --version  0.34s user 0.05s system 102% cpu 0.381 total
```

"Print the version string" is probably the simplest thing you can ask Babel to do and serves as a good proxy for start up cost. This seems a little excessive. Roughly, this is the same amount of time it takes to bounce a packet [from us-east-1 to Turkmenistan and back](https://datapath.io/resources/blog/aws-network-latency-map/).

Node takes about 13ms.

```
$ time node --version
v7.9.0
node --version  0.00s user 0.00s system 45% cpu 0.013 total
```

I'm wondering what steps could be taken to optimize the start up time as Babel is usually not a long-running application.

loganfsmyth:
| Fixed Issues?            | Fixes #5702 (but not primary goal)
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

There is some serious craziness in here for `@babel/register` sooooo sorry 😬 Essentially we want to be able to load `babel/core` with lazy requires safely in `@babel/register`, and that's dangerous/broken if register were left as-is because calling `babel.transform` could then actually trigger the require hook again causing a cycle. 

Triggering a cycle isn't _guaranteed_ to be an issue, since you could imagine a world where someone does:

```
// .babelrc
{
  presets: ['@babel/env'],
  overrides: [{
    exclude: "./my-custom-plugin",
    plugins: [
      "./my-custom-plugin",
    ],
  }],
}
```
to compile a project with some custom plugin, while _also_ compiling that plugin itself with a different config.

To achieve all that while still having `@babel/core` kick off `require` calls at random times during transformation, this PR makes `@babel/register` much smarter about how it loads `@babel/core`. You can almost think of it like it loads it with its own loader context that explicitly tracks while files it has loaded. By tracking all that, we can explicitly skip those files in the require hook, avoiding trying to compile files that are necessary for the compilation itself.

There is then a small hole exposed in that loader logic so that when `@babel/core` loads `.babelrc.js`s, plugins, and presets, it will _not_ track those, so that they actually do get passed to the require hook. This means it is up to the user to ensure that their Babel config does not introduce dependencies. We also have existing logic in `@babel/core` to show helpful error messages when there are cycles detected in config files/plugins/presets.

The end result of this PR is that `@babel/core` can lazy-load everything and all the rest of babel-register pretty much ends up behaving like it did before.

## Why!

The end goal of this PR is the lazy-loading. With PR's like #7472 encouraging tooling to load `@babel/core` to check for config, this means we need to make loading core as fast as possible. Currently, it is the exact opposite of that. A single call to `require("@babel/core")` on my machine for instance takes `~250-350ms`. With this PR, it now takes `~25-30ms` on my machine.

This brings down the total time for a call to `require("@babel/core").loadPartialConfig()` dramatically, which should make it easier for people to adopt over trying to implement the logic themselves.


packages/babel-register/src/babel-core/track-deps.js
```diff
@@ -0,0 +1,94 @@
...
+    );
+    try {
+      return req(id);
+    } finally {
+      removeHook();
+    }
+  }
+
+  // Ensure that require.resolve and other properties are carried over.
+  return Object.assign(babelRegisterRequire, req);
```
Kovensky: They're enumerable? :o

packages/babel-register/src/babel-core/track-deps.js
```diff
@@ -0,0 +1,94 @@
...
+
+// For any file load kicked off by @babel/core or its dependencies has
+// a snippet injected into it to wrap the 'require' hook so that we
+// can avoid Babel triggering cyclical dependencies.
+function injectWrap(code, uid, inRoot) {
+  const override = `require(${JSON.stringify(
+    __filename,
+  )}).wrap(require, __filename, ${uid}, ${inRoot})`;
+
+  return `return (function(require){${code}\n}).call(this, ${override});`;
```
Kovensky: top level return is icky 😅

packages/babel-register/src/babel-core/track-deps.js
```diff
@@ -0,0 +1,94 @@
...
+    id: string,
+    // The name of the file that is loading the id.
+    filename: string,
+    // 'true' if the 'filename' is known to be in the same package that
+    // was originally passed to 'load'. 'true' as long as the file was
+    // reached using only relative paths within the original module.
+    inRoot: boolean,
+  ) => { exports: mixed } | void,
+) {
+  const files = new Set();
```
Kovensky: The set seems to be write-only?

packages/babel-register/src/babel-core/track-deps.js
loganfsmyth: Yeah for some reason I was thinking it was needed, but it isn't is it.

packages/babel-register/src/babel-core/track-deps.js
loganfsmyth: Seems like!

packages/babel-register/src/babel-core/track-deps.js
loganfsmyth: It ends up exported and used in `/node.js` to skip them.

packages/babel-register/src/babel-core/track-deps.js
hzoo: wonder how this affects the error messages (although that might just be messed up already)

hzoo:
Were you able to test this? should probably run locally before next release then like https://github.com/babel/babel/issues/6134. For compiling with lazy and also new register although I guess this is a fix for the previous time we tried this.

loganfsmyth:
The tests for `babel-node` fail without this logic in babel-register so I'm relatively confident that it works well.

The main question for me right now, which I'm going back and forth on, is whether we care about the case of re-entrant configs. And if we do (which I think I do?), should it be something we require users to opt into?

For instance right now people often do `ignore: []` to have `babel-register` process every file, which because we allow re-entrant plugins, means that this triggers a cyclical error because loading the preset itself will trigger an error.

I think I'm personally leaning toward us making that behavior opt-in instead. We'd then instead require users to do 
```
require("@babel/register")({
  reentrant: true
});
```
if they wanted `register` to compile a plugin/preset itself, before using that plugin/preset somewhere else in their build process.

It'd be a lot safer for the average user since we get issues like https://github.com/babel/babel/issues/7208 reasonably often.

Doing that would still require the logic in this PR, but I'd remove the `overrideCallback` entirely and instead we'd have to pass the actual plugin/config `require` callback to Babel itself during the call to `.transform`.

packages/babel-register/src/babel-core/track-deps.js
loganfsmyth: They'll be an extra scope in the debugger if you pause, but otherwise I don't think it will affect anything.

hzoo:
Ok, just needs docs on that and maybe think if that's the best word for it. I guess not sure what else it would be called but the intention is to compile configs themselves with register which affect itself? Or could we error with a nicer message when it happens without?

loganfsmyth:
> Or could we error with a nicer message when it happens without?

I don't follow. If we error without the flag on you mean? Without the flag I don't think I'd expect it to error, and with the flag we already give some errors about reentrant code.

hzoo:
just wondering how someone would know to use the flag, if it's clear then never mind

loganfsmyth:
I think what I might do is tweak this PR to disallow reentrant plugins/presets entirely for now, and land that, and then do a second PR to add the `reentrant` option and the code that is necessary in babel-core to make it work.

loganfsmyth:
> just wondering how someone would know to use the flag, if it's clear then never mind

Gotcha, yeah it's not super clear, but then I don't think it'll be super common anyway. It's not something that'll work with any of the other Babel integrations anyway since they don't do require hooks, so I think making it a special opt-in feature of babel-register seems fine. We'll definitely have to document it clearly though.

loganfsmyth:
Alright I decided to remove most of the complex logic that I had originally here for `register` in favor of totally disallowing re-entrant compilation. I added a note to the README on the recommended approach of you do want to compile a plugin up front, it should still cover 99% of the cases where people would actually want to live-compile a plugin/preset.

PR #7605, opened 2018-03-20
Disable flow on transformClass, fix preset-env errors

thymikee:
Disabled Flow from `transformClass.js` (see #7444) which caused a cascade of errors, as per @loganfsmyth [request](https://github.com/babel/babel/pull/7444#discussion_r175861060). Also fixed small Flow issues I found in other files so now `yarn flow` pass.

| Fixed Issues?            | 
| Patch: Bug Fix?          | Fix
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 

### DIFF
--- a/packages/babel-preset-env/src/normalize-options.js
+++ b/packages/babel-preset-env/src/normalize-options.js
@@ -15,7 +15,7 @@ const validIncludesAndExcludes = new Set([
   ...defaultWebIncludes,
 ]);
 
-const pluginToRegExp = (plugin: any): RegExp => {
+const pluginToRegExp = (plugin: any): ?RegExp => {
   if (plugin instanceof RegExp) return plugin;
   try {
     return new RegExp(`^${normalizePluginName(plugin)}$`);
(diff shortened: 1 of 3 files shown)


packages/babel-preset-env/src/normalize-options.js
```diff
@@ -24,15 +24,15 @@ const pluginToRegExp = (plugin: any): RegExp => {
...
+const selectPlugins = (regexp: ?RegExp): Array<string> =>
   Array.from(validIncludesAndExcludes).filter(
     item => regexp instanceof RegExp && regexp.test(item),
   );
 
 const flatten = array => [].concat(...array);
 
 const expandIncludesAndExcludes = (
-  plugins: Array<string | RegExp> = [],
+  plugins: Array<string> = [],
```
loganfsmyth: @existentialism Is this right? I thought we supported regexes now?

existentialism: Ack, you're right.

@thymikee can you revert please? Support for this landed in https://github.com/babel/babel/pull/7242

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -35,7 +35,7 @@ const expandIncludesAndExcludes = (
   plugins: Array<string | RegExp> = [],
   type: string,
 ): Array<string> => {
-  if (plugins.length === 0) return plugins;
+  if (plugins.length === 0) return [];
```
thymikee: Couldn't return `plugins` here because the output is different type than the input. I assumed that we're safe to return just an empty array here, because technically we don't accept any other types.

PR #7615, opened 2018-03-22
clean makefile a bit

xtuc:
| Tests Added + Pass?      | no/Yes
| Documentation PR         | 

This works with GNU Make, but I have no way to test if it works on Windows.


Andarist:
If we care about Windows maybe it would be a good thing to use npm scripts instead of make?

xtuc:
@Andarist Make should work with the Ubuntu subsystem and other terminal emulator, but I can't confirm. 

Our makefile is pretty big and would be hard to maintain using only npm-scripts. I think that Make is the good solution here.

xtuc:
I'll go ahead and merge this.

PR #7622, opened 2018-03-23
Check MemberExpression nodes in `isPure`

jlongster:
| Fixed Issues?            | 
| Major: Breaking Change?  | x
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

I noticed that `isPure` is failing in case where it shouldn't. Currently is doesn't check MemberExpression's at all. Originally, I was going to copy `isPure` and use my own version, with the added ability to return `true` if the binding is from the top-level scope (so it's safe to hoist there). I was happy to find that babel's `isPure` already does this! Let's say I was trying to implement CSS-in-JS hoisting. The scope check allows this:

```js
import React from 'react';
import { color } from 'colors';

class App extends React.Component {
  render() {
    return <View style={{ backgroundColor: 'red', color: color }}>hi</View>;
  }
}
```

to be transformed into this:

```js
'use strict';

var _react = require('react');

var _react2 = _interopRequireDefault(_react);

var _colors = require('colors');

var _ref = { backgroundColor: 'red', color: _colors.color };

class App extends _react2.default.Component {
  render() {
    return _react2.default.createElement(
      View,
      { style: _ref },
      'hi'
    );
  }
}
```

Assuming we are detecting `style` props and checking if the ObjectExpression inside of them is pure. That's sweet! But let's say I make a simple change and instead use `colors.color` instead of `color` inside the render function:

```js
import React from 'react';
import * as colors from 'colors';

class App extends React.Component {
  render() {
    return (
      <View style={{ backgroundColor: 'red', color: colors.color }}>hi</View>
    );
  }
}
```

Using the current babel `isPure` would fail in this case. Here's the output:

```js
'use strict';

var _react = require('react');

var _react2 = _interopRequireDefault(_react);

var _colors = require('colors');

var colors = _interopRequireWildcard(_colors);

// boilerplate snipped

class App extends _react2.default.Component {
  render() {
    return _react2.default.createElement(
      View,
      { style: { backgroundColor: 'red', color: colors.color } },
      'hi'
    );
  }
}
```

The problem is that we don't traverse into MemberExpression nodes, when there's no reason we shouldn't. Now that we traverse, we eventually hit `colors` and notice that's it's a top-level binding, so the same logic as before marks this is safe to hoist. The new behavior is now:

```js
var _ref = { backgroundColor: 'red', color: colors.color };

class App extends _react2.default.Component {
  render() {
    return _react2.default.createElement(
      View,
      { style: _ref },
      'hi'
    );
  }
}
```

It must be common in React-land at least to use this syntax (I do it all the time to pull colors from my style guide). Adding this might allow apps to take advantage of hoisting a lot more.

To be honest, I don't have time to figure all of babel's setup and thoroughly add tests and such. This PR is a starting point for somebody else to finish.

### DIFF
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -619,7 +619,10 @@ export default class Scope {
       }
       return true;
     } else if (t.isMemberExpression(node)) {
-      return this.isPure(node.object, constantsOnly);
+      return (
+        this.isPure(node.object, constantsOnly) &&
+        this.isPure(node.property, constantsOnly)
+      );
     } else if (t.isClassMethod(node)) {
       if (node.computed && !this.isPure(node.key, constantsOnly)) return false;
       if (node.kind === "get" || node.kind === "set") return false;
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -618,6 +618,8 @@ export default class Scope {
         if (!this.isPure(prop, constantsOnly)) return false;
       }
       return true;
+    } else if (t.isMemberExpression(node)) {
+      return this.isPure(node.object, constantsOnly);
     } else if (t.isClassMethod(node)) {
       if (node.computed && !this.isPure(node.key, constantsOnly)) return false;
       if (node.kind === "get" || node.kind === "set") return false;


jlongster:
Actually we also need to traverse into the property because you could do something like `foo[bar]`. I ignore the `computed` property, but that should be OK if we also traverse in to the property.

existentialism:
👋 @jlongster!

We should add some test cases here:
https://github.com/babel/babel/blob/master/packages/babel-traverse/test/scope.js#L108

jlongster:
Ok. Somebody did bring up that member expressions aren't *technically* pure because of stuff like getters. But it seems like `isPure` is pretty relaxed about that kind of stuff? What is the stance on assuming more esoteric  JS features aren't used?

jlongster:
Turns out I don't need `isPure`. What I really wanted to check is if an expression references any non-toplevel variables. Unless somebody else thinks this is a good idea, I'll close this PR in a few days.

hzoo:
You probably saw this already but just leaving a link to the react constant transform https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-react-constant-elements/src/index.js

Yeah since it would affect getters this kind of behavior would be better opt-in (for these kinds of optimizations)? Otherwise would be a breaking change for something relying on that.

jlongster:
I actually hadn't looked at its code yet, but it looks relatively straight-forward.

It seems like you all want `isPure` to be strict and try to really check if there are any side effects. That's a hard ask for something like JS, when something like `value` can have side effects, and `isPure` returns true for identifiers:

```js
let x = 1;

const y = {
  valueOf() {
    return x++;
  }
}

console.log(y);
console.log(y);
console.log(y);
console.log(x);

// Output:
//
// 1
// 2
// 3
// 4
```

`+y` is also not side-effect free, but is considered pure by `isPure`. I think almost any expression in JS can never be statically determined to be pure. It's only a useful heuristic to give the user something to use that says "yeah, unless you are doing something esoteric, this is pure". Member expressions would be a part of that. Other expressions can have side effects just as much as them.

But I don't need this change anymore, I'm taking an alternative approach. In my opinion though, you are "opting in" to this behavior by adding the specific babel plugins that perform these kinds of optimizations, and accepting the fact that expressions handling by those plugins must only use a subset of JS.

nicolo-ribaudo:
`+y` is also not side-effect free, but is considered pure by `isPure`. If we can't prove that `+y` is pure (e.g. `y` is a constant variable whose value is not an object), it should return false.
We should make the `isPure` function check for purity better, not worse.

jlongster:
The fact is almost all JS expressions are not side-effect free.

nicolo-ribaudo:
I know, for example in this code:
```js
var foo = a;
```
`a` is not reported as pure.

The `isPure` should be strict, but if needed we could have an `isPureLoose` function.

hzoo:
> The isPure should be strict, but if needed we could have an isPureLoose function.

Yeah that was what I was thinking

> In my opinion though, you are "opting in" to this behavior by adding the specific babel plugins that perform these kinds of optimizations, and accepting the fact that expressions handling by those plugins must only use a subset of JS.

Yeah I totally agree, but we are talking about the API for Babel so I meant that for that it's about the opt-in for which method you use for those plugins and I didn't want to break what people are using with isPure now if that makes sense?

jlongster:
Yeah, it is a breaking change so it's fine if you don't think it's worth it. Any change to something like `isPure` is breaking. It could be confusing that multiple `isPure` implementations exist but it's totally up to you all.

jlongster:
Closing unless someone else wants to pick this up and make further progress on `isPureLoose` or something like that.

PR #7623, opened 2018-03-24
Add ??= to Logical Assignment Operators

jridgewell:
`??=` is being merged into the Logical Assignment Operator proposal, and the overall proposal will wait until nullish coalescing is finalized.


loganfsmyth:
@jridgewell This PR seems to be causing Flow errors. 

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ packages/babylon/src/tokenizer/index.js:610:12

Cannot call this.expectPlugin because property expectPlugin is missing in Tokenizer [1].

 [1] 125│ export default class Tokenizer extends LocationParser {
        :
     607│     const next = this.input.charCodeAt(this.state.pos + 1);
     608│     const next2 = this.input.charCodeAt(this.state.pos + 2);
     609│     if (next === charCodes.questionMark) {
     610│       this.expectPlugin("nullishCoalescingOperator");
     611│
     612│       if (next2 === charCodes.equalsTo) {
     613│         // '??='


Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ packages/babylon/src/tokenizer/index.js:614:14

Cannot call this.expectPlugin because property expectPlugin is missing in Tokenizer [1].

 [1] 125│ export default class Tokenizer extends LocationParser {
        :
     611│
     612│       if (next2 === charCodes.equalsTo) {
     613│         // '??='
     614│         this.expectPlugin("logicalAssignment");
     615│         this.finishOp(tt.assign, 3);
     616│       } else {
     617│         // '??'



Found 2 errors
```

jridgewell:
Does this need `$FlowFixMe`s? Also, why the hell do we need `$FlowFixMe`?

loganfsmyth:
No, it's a legit error. You moved the calls into the `Tokenizer` class, but those helpers are defined in a subclass, so from a type standpoint it is possible that there'd be an instance of `Tokenizer` where `expectPlugin` doesn't exist.

loganfsmyth:
I think generally I'd expect checks like this to live in the parser, with the tokenizer being fine to tokenize it, even if it isn't allowed. Then the parser can just throw on the unexpected token.

jridgewell:
Ohhh, I'll get this fixed.

PR #7629, opened 2018-03-25
Fix flow errors with Logical Assignment Operators

jridgewell:
https://github.com/babel/babel/pull/7623#issuecomment-375998350


loganfsmyth:
I don't know anything about the actual plugins you're checking, but the fix itself seems reasonable to me.

Issue #7634, opened 2018-03-27
allowAwaitOutsideFunction parser option


jdalton:
Just as there's a loose mode to allow top-level `return` it would be nice to have some kind of loose mode, `allowAwaitEverywhere`, or `allowAwaitOutsideFunction` option for top-level `await`.

nicolo-ribaudo:
What would it be used to? I'd suggest wrapping your code in an async function before parsing it.

jdalton:
There is now environments like REPLs which support top-level await. Being able to parse it would be great.

hzoo:
Sounds fine to me, if someone wants to PR it

jdalton:
🤘 Do you have an option name preference?

hzoo:
Any of those work, open to ideas:

`allowAwaitEverywhere`
`allowAwaitOutsideFunction`
`allowTopLevelAwait`?

jdalton:
Ok.

jdalton:
Ok. I can prep a PR.

jdalton:
WIP PR opened at #7637.

hzoo:
Found some related issues as reference

https://github.com/babel/babel/issues/6703
https://github.com/babel/babel/issues/6720

nicolo-ribaudo:
Implemented in #7637

PR #7637, opened 2018-03-27
Add options.allowAwaitOutsideFunction.

jdalton:
This PR addresses #7634.

~~It's missing tests as the moment. *(I'll rebase them in soon)*~~

**Update:**

✅ Added tests.


nicolo-ribaudo:
WDYT of naming it `allowTopLevelAwait`, since it is how it's usually referred to?

jdalton:
Fixed lint errors, added tests, and ensured an error is thrown for the case raised by @nicolo-ribaudo.
I kept the name as `allowAwaitOutsideFunction` since it's technically more correct and aligns with existing and similar options *(feel free to tweak to taste however)*.

jdalton:
> Could we maybe rename the tests?

I wasn't sure about that but noticed a large collection of tests incremented as numbers. Is there a particular place or naming scheme you had in mind?

nicolo-ribaudo:
You could put it in `fixtures/es2017/async-functions/allowAwaitOutsideFunction`.

> I kept the name as allowAwaitOutsideFunction since it's technically more correct and aligns with existing and similar options

I still prefer the `allowTopLevelAwait` name, but I'm ok with this since it is how `allowReturnOutsideFunction` is named.

nicolo-ribaudo:
I just learned that this is actually a new proposal (https://github.com/MylesBorins/proposal-top-level-await). Can you define this feture it using a new Babylon plugin (`topLevelAwait`) instead of an option?

jdalton:
I rather not since the proposals are much more involved. I think later, if someone wants to add a plugin to support those proposals more in-depth then they could build on top of `allowAwaitOutsideFunction`

jdalton:
Moved tests to a more readable place.

hzoo:
>  I think later, if someone wants to add a plugin to support the proposal (and variations) more in-depth, then they could build on top of allowAwaitOutsideFunction

that's true

PR #7639, opened 2018-03-28
Allow StringLiteral and NumericLiteral as ObjectTypeProperty.key

### LINKED ISSUE #7635: babel-types validation only accepts Identifiers for ObjectTypeProperty.key
This is a bug report.

Say, we want to build type annotation like shown below with `@babel/types` 

```jsx
type Locales = {
  en: Object;
  "en-US": Object;
}
```

this is what it would look like:

```js

t.typeAlias(
  t.identifier('Locales'),
  null,
  t.objectTypeAnnotation([
    t.objectTypeProperty(
        t.identifier('en'),
        t.genericTypeAnnotation(t.identifier('Object'))
    ),
    t.objectTypeProperty(
        t.stringLiteral('en-US'), // spoiler: results in TypeError
        t.genericTypeAnnotation(t.identifier('Object'))
    )
  ]),
);
```

[This code on RunKit](https://runkit.com/unconfident/babel-types-7.0.0-beta.42-bug) 

And this is the error we get on version version 7.0.0-beta.42:

```
TypeError: Property key of ObjectTypeProperty expected node to be of a type ["Identifier"] but instead got "StringLiteral"
```

Not very confident in what I'm saying now but `StringLiteral` and `NumericLiteral` are perfectly valid keys afaik and they should be allowed there. Could be wrong

Relevant code: https://github.com/babel/babel/blob/a36525be45b905f8830a215f277c72f21481623a/packages/babel-types/src/definitions/flow.js#L290-L301

Slightly related issue: #6507

unconfident:
| Fixed Issues?            | Fixes #7635
| Documentation PR         | 
| Any Dependency Changes?  |

Previously `ObjectTypeProperty` validator only accepted `Identifier` as a key what made impossible to describe some object types using babel-types.

Validation rule changed to also accept `StringLiteral` ~~and `NumericLiteral`~~

### DIFF
--- a/packages/babel-types/README.md
+++ b/packages/babel-types/README.md
@@ -1553,7 +1553,7 @@ See also `t.isObjectTypeProperty(node, opts)` and `t.assertObjectTypeProperty(no
 
 Aliases: `Flow`, `UserWhitespacable`
 
- - `key`: `Identifier` (required)
+ - `key`: `Identifier | StringLiteral` (required)
  - `value`: `FlowType` (required)
  - `variance`: `Variance` (default: `null`)
  - `kind`: `"init" | "get" | "set"` (default: `null`)
--- a/packages/babel-types/src/definitions/flow.js
+++ b/packages/babel-types/src/definitions/flow.js
@@ -291,7 +291,7 @@ defineType("ObjectTypeProperty", {
   visitor: ["key", "value", "variance"],
   aliases: ["Flow", "UserWhitespacable"],
   fields: {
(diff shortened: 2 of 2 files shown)


danez:
I think I would play safe and only allow strings for now as long as flow does not support numbers.

unconfident:
Sounds reasonable, I'll update it in a second

unconfident:
force-pushed

existentialism:
@unconfident thanks!

Issue #7402, opened 2018-02-20
Bug Report: babel-preset-env feature-detection error with dynamic import() statements.


jgerigmeyer:
## Bug Report:

When using babel-preset-env with `useBuiltIns: 'usage'` and [webpack's dynamic imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports) (e.g. `import('my-module').then(...)`) -- along with `@babel/plugin-syntax-dynamic-import` -- feature-detection fails to include `es6.promise`, which is required for the code to work on IE11.

### Input Code

```js
import('my-module');
```

### Compiled input code:

```js
__webpack_require__.e/* import() */(2).then(__webpack_require__.bind(null, 155));
```

Elsewhere in the Webpack runtime:

```js
/******/ 	__webpack_require__.e = function requireEnsure(chunkId) {
/******/ 		...
/******/ 		// setup Promise in chunk cache
/******/ 		var promise = new Promise(function(resolve, reject) {
/******/ 			installedChunkData = installedChunks[chunkId] = [resolve, reject];
/******/ 		});
/******/ 		installedChunkData[2] = promise;
```

### Babel/Babylon Configuration (.babelrc, package.json, cli command)

```js
{
  plugins: ['@babel/plugin-syntax-dynamic-import'],
  presets: [
    [
      '@babel/preset-env',
      {
        modules: false,
        useBuiltIns: 'usage',
      },
    ],
  ],
}
```

### Expected Behavior

```bash
Using targets:
{
  "ie": "11"
}

Using modules transform: false

Using polyfills with `usage` option:
[/path/to/my/file.js] Added following polyfill:
  es6.promise { "ie":"11" }
```

No errors in IE 11.

### Current Behavior

```bash
Using targets:
{
  "ie": "11"
}

Using modules transform: false

Using polyfills with `usage` option:
[/path/to/my/file.js] Based on your code and targets, none were added.
```

IE11 error: `'Promise' is undefined`

### Possible Solution

Not sure the ideal fix -- is this a bug in babel-loader, or @babel/preset-env, or @babel/plugin-syntax-dynamic-import, or even webpack? In the meantime, a workaround is to add either of the following in the input JS file:

```js
const p = new Promise();
```

or

```js
import 'core-js/modules/es6.promise';
```

### Your Environment

| software         | version(s)
| ---------------- | -------
| Babel            |  7.0.0-beta.40
| webpack      |  3.11.0
| node             |  8.9.4
| npm              |  5.6.0
[...]

PR #7646, opened 2018-03-31
Update to `core-js@3`

### LINKED ISSUE #2500: When using babel-runtime, Symbol.iterator is different in user code & in polyfills
The following code:

``` js
const o = {0: 'a', 1: 'b', 2: 'c', length: 3};
o[Symbol.iterator] = Array.prototype[Symbol.iterator];

let counter = 0;
for (const div of o) {
    counter++;
}
console.log('counter', 3);
console.assert(counter === 3, 'Iterability doesn\'t work!');
```

is transformed by:

```
babel --optional runtime iterable-test.js > iterable-test-transformed.js
```

to:

``` js
'use strict';

var _Symbol$iterator = require('babel-runtime/core-js/symbol/iterator')['default'];

var _getIterator = require('babel-runtime/core-js/get-iterator')['default'];

var o = { 0: 'a', 1: 'b', 2: 'c', length: 3 };
o[_Symbol$iterator] = Array.prototype[_Symbol$iterator];

var counter = 0;
var _iteratorNormalCompletion = true;
var _didIteratorError = false;
var _iteratorError = undefined;

try {
    for (var _iterator = _getIterator(o), _step; !(_iteratorNormalCompletion = (_step = _iterator.next()).done); _iteratorNormalCompletion = true) {
        var div = _step.value;

        counter++;
    }
} catch (err) {
    _didIteratorError = true;
    _iteratorError = err;
} finally {
    try {
        if (!_iteratorNormalCompletion && _iterator['return']) {
            _iterator['return']();
        }
    } finally {
        if (_didIteratorError) {
            throw _iteratorError;
        }
    }
}

console.log('counter', 3);
console.assert(counter === 3, 'Iterability doesn\'t work!');
```

After invoking:

```
npm i babel-runtime
```

this transformed script prints `3` correctly in Node 4.1.2 & 0.12.7 but in 0.10.40 it throws:

```
/Users/mgol/_/iterable-test-transformed.js:31
            throw _iteratorError;
                  ^
TypeError: [object Object] is not iterable!
    at TypeError (<anonymous>)
    at module.exports.require.getIterator (/Users/mgol/_/node_modules/core-js/library/modules/core.get-iterator.js:5:40)
    at Object.<anonymous> (/Users/mgol/_/iterable-test-transformed.js:16:23)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:935:3
```

If you don't use babel-runtime but include core-js in a separate script, it works fine.

babel-runtime 5.8.25.

@zloirock presents a solution to the problem in [his comment under the core-js issue](https://github.com/zloirock/core-js/issues/118#event-429294244).

### LINKED ISSUE #6602: Add support for Samsung Internet browser
> Issue originally reported by @philipwalton in babel/babel-preset-env#428

Samsumg Internet uses Chromium under the hood, but it doesn't always stay up-to-date with the latest version. We discovered this recently due to a bug report from a Samsung Internet user.

To fix it, we had to investigate the version of Chromium they use and [then manually add that to the browser list](https://github.com/GoogleChrome/workbox/pull/815/files#diff-7b9b61c220cc1842c38185dad092626bR162):

```js
const babelPlugin = babel({
  presets: [['env', {
    targets: {
      browsers: ['chrome >= 51'],
    },
    // ...
});
```

But this requires a bit or work on the user's side. Since browserlist support Samsung Internet, ideally we'd be able to do this:

```js
const babelPlugin = babel({
  presets: [['env', {
    targets: {
      browsers: ['last 2 Samsung versions'],
    },
    // ...
});
```

I haven't looked into whether this can be done solely on the `babel-preset-env` side, or if it would require changes to http://kangax.github.io/compat-table. If the latter is the case, maybe we can loop in @poshaughnessy for help making that happen?

### LINKED ISSUE #8928: Make all the polyfills available without global namespace pollution
## Feature Request

This is a draft proposal that aims to make all the polyfills available without global namespace pollution.

Make all the polyfills available on the runtime to make development easier. Development of libraries is possible but some of de features (polyfills) are not available when using **babel-runtime**. These are the instance methods (polyfills) like *.includes()*, *.repeat()*... that can pollute our global scope.

Include all the missing instance methods like (.includes, .repeat ...) and without polluting the global scope.
To to so, babel would have to determine the type of the instance at runtime to transfer the call to a static one.

```ts
 console.log(['a', 'b', 'c'].includes('a'));
```
would be transformed to

```js
var _arrayIncludes = require("core-js-pure/features/array/includes");
var _stringIncludes = require("core-js-pure/features/string/includes");

// Here is decided what built-in method to call based on instance type
var _polyfillsCalls = (instance, methodName, methodParams) => {
     // ... Use methodName to determine that it can only be 'String' or 'Array' (because of includes)
     // This feature has to be generic and return the maybe-type of the instance calling the method
     // This can be possibly done by mapping function names to their possible instance types
     // For example .includes() calls can be mapped to types => ['string', 'Array'] !

     // Then we proceed with the execution with type detection #map

     if (Array.isArray(instance))
          return _arrayIncludes(instance, ...methodParams);
     
     if (typeof instance === 'string')
          return _stringIncludes(instance, ...methodParams);

     // No polyfill could be mached, default call
     return instance[methodName](methodParams);
};

console.log(_polyfillsCalls(['a', 'b', 'c'], 'includes', ['a']);
```

For this, core-js features could be used directly. Not sure that all polyfills can be covered this way, but it should cover most of them...

### LINKED ISSUE #9250: @babel/preset-env not always polyfilling Promise
## Bug Report

**Current Behavior**
When working with `async/await` and setting IE11 as a target, it does not always include `es6.promise`.

**Input Code**
- https://github.com/rbnlffl/mehrsprachig/blob/master/src/mehrsprachig.js

```js
async getLocale() {
    const req = await fetch(this.sources[this.language]);
    const res = await req.json();

    this.locales[this.language] = res;
}
```

```bash
[/Users/robin/Sites/mehrsprachig/src/mehrsprachig.js] Added following polyfills:
  regenerator-runtime { "android":"4.4.3", "chrome":"69", "edge":"17", "firefox":"62", "ie":"11", "ios":"11.3", "opera":"56", "safari":"11.1" }
  web.dom.iterable { "android":"4.4.3", "chrome":"69", "edge":"17", "firefox":"62", "ie":"11", "ios":"11.3", "opera":"56", "safari":"11.1" }
  es6.array.iterator { "android":"4.4.3", "ie":"11" }
  es6.symbol { "android":"4.4.3", "edge":"17", "ie":"11" }
  es7.symbol.async-iterator { "android":"4.4.3", "edge":"17", "ie":"11", "ios":"11.3", "safari":"11.1" }
  es6.regexp.split { "android":"4.4.3", "edge":"17", "ie":"11" }
  es6.regexp.replace { "android":"4.4.3", "edge":"17", "ie":"11" }
  es6.string.includes { "android":"4.4.3", "ie":"11" }
  es7.array.includes { "android":"4.4.3", "ie":"11" }
```

**Expected behavior/code**
I expect a polyfill for IE11 for Promises to be included.

**Babel Configuration (babel.config.js)**

```js
module.exports = {
    sourceMaps: true,
    presets: [
        ['@babel/preset-env', {
            targets: {
                ie: 11,
                browsers: 'last 2 versions'
            },
            useBuiltIns: 'usage',
            debug: true
        }]
    ],
    ignore: [ 'node_modules' ]
};
```

**Environment**
- Babel version(s): 7.2.2
- Node/npm version: 10.14.2/6.4.1
- OS: macOS 10.14.2
- Monorepo: no
- How you are using Babel: `rollup-plugin-babel`

I can fix the issue for now by importing `whatwg-fetch` and `promise-polyfill/src/polyfill` manually. However, this is not really elegant, as I'm working with a polyfilling package, anyway, and this always includes these polyfills, even when not transforming anything with babel.

Another thing to mention: `fetch` does not get polyfilled, either. I've read somewhere that you guys think about including it in the future—is this true? If yes, is there any planned eta?

zloirock:
#### `@babel/runtime`
- Added `@babel/runtime-corejs3` package and `corejs: 3` options to `@babel/plugin-transform-runtime`.
- Added support of instance methods, fixes #8928.
- Added flag `proposals` (in `corejs: { version: 3, proposals: true }` format) for support all proposals polyfills from `core-js`.
- Used separate directories in runtime for `core-js` entry points with proposals and without.
- Used `get-iterator-method` helper for getting iterators, fixes #2500.
- As a cheap bonus, added support of IE8- (except some cases of `regenerator`).

#### `@babel/polyfill`
- Should be deprecated in favor of separate usage required features from `core-js` and `regenerator-runtime` with an informative message.

#### `@babel/preset-env`
- Uses for built-ins data from [`core-js-compat`](https://github.com/zloirock/core-js/tree/master/packages/core-js-compat) instead of `compat-table` since information from `compat-table` [is not enough](https://github.com/zloirock/core-js/tree/master/packages/core-js-compat).
- `useBuilIns` now requires direct setting of `corejs` version option, without it will be used `2` by default and shown deprecation warning.
- Added support of minor `core-js` versions for simplify updating in the future.
- For preventing some order-related problems, polyfills in the both `core-js@3` plugins added on `post` stage in the order of `core-js-compat` data.
- Divided plugins and polyfills parts of `preset-env`, instead of 2 internal plugins for adding polyfills, we have 6: usage and entry versions of plugins for `core-js@2`, `core-js@3`, `regenerator-runtime`.
- Added support `samsung` target (for Samsung Internet) since `core-js-compat` and `compat-table` now contains mapping for this, fixes #6602.

#### `useBuilIns: entry` with `corejs: 3`
- No longer transforms `@babel/polyfill`.
- Transforms **all possible** `core-js` entry points to import of related modules (based on data from [`core-js-compat`](https://unpkg.com/core-js-compat@3.0.0-beta.15/entries.json)).
- Since of this, we no longer need `shippedProposals` / `proposals` flags with `useBuilIns: entry`.
- Removes `regenerator-runtime/runtime` import where it's not required.

#### `useBuilIns: usage` with `corejs: 3`
- In addition to `shippedProposals`, added flag `proposals`  (in `corejs: { version: 3, proposals: true }` format) for polyfill all proposals from `core-js`.
- Fixed list of dependencies in built-in definitions.
- Improved the way of determination method / built-in name and source of this method.
- Adds import of required polyfills on `MemberExpression`, `ObjectPattern`, `in` operator.
- Adds import of required polyfills on access to global object properties.
- Adds import of all required common iterators on all syntax features which use iterators protocol (`for-of`, destructuring, spread, `yield` delegation, etc.).
- Adds import of promises on syntax features which use promises (async functions/generators, dynamic import, etc.), fixes #9250, #7402, etc.

#### `core-js@2` stuff
I didn't want to tough `core-js@2`-related stuff, however
- Fixed some serious errors in definitions which breaks `Object.getOwnPropertySymbols`, `Symbol.toStringTag` logic, `Promise#finally`, `Array#forEach`, etc.
- `Array#flatMap` and trim methods moved to stable features as a part of ES2019 and loaded by deprecated `@babel/polyfill` and `@babel/preset-env` with `corejs: 2` option.

#### `@babel/runtime`

Here we have no serious problems.
- Added `@babel/runtime-corejs3` package and `corejs: 3` options to `@babel/plugin-transform-runtime`.
- Added support of instance methods.
- Used `get-iterator-method` helper for getting iterators, fixes #2500.
- As a cheap bonus, added support of IE8- (except some cases of `regenerator`).

#### Breaking changes
`@babel/runtime` can be updated without any problems by adding `@babel/runtime-corejs3` package. But not `@babel/polyfill` and `@babel/preset-env`.

Updating `@babel/polyfill` to `core-js@3` will break `@babel/preset-env`. `@babel/preset-env` with `useBuiltIns` transforms import of `@babel/polyfill` to import of required `core-js` modules. Most users don't add `core-js` dependency directly. In case of a minor update, even if `@babel/polyfill` dependency will be updated with `@babel/preset-env`, in the top level of `node_modules` we will have `core-js@2`, not `core-js@3` because a serious part of dependencies depends on packages with `core-js@2` in dependencies.

Updating `@babel/preset-env` will cause the same problem. More other, `@babel/preset-env` transforms also imports of `core-js`. If someone has `core-js@2` in dependencies, a minor update of `@babel/preset-env` will break it.

Before `babel@7` release, I proposed to deprecate `@babel/polyfill` in favor of separate inclusion of `core-js` and `regenerator-runtime`, but it was not done.

So,

Options for `@babel/polyfill`:
- A major bump of `@babel/polyfill`.
- Deprecation of `@babel/polyfill` package in favor recommendation of separate inclusion of required parts of `core-js` and `babel-runtime`.

Options for `@babel/preset-env`:
- A major bump of `@babel/preset-env`.
- Adding an option with a version of `core-js` like in `@babel/plugin-transform-runtime`. I'm not a fan of this way, because in this case, we will use a deprecated version of `core-js` by default, will cause spaghetti in configs and the code base and will cause problems for other mentioned here changes, at least, see `polyfilling of proposals` section.
- Deprecation of `useBuiltIns` option of `@babel/preset-env` in favor of creation plugin(s) / preset in the scope of `core-js` for polyfills only.

#### `@babel/preset-env`
Some of the improvements already added, some should be added.
- `compat-table` is not enough for correct detection of environments where `core-js` modules required. More info about it you could find [here](https://github.com/zloirock/core-js/pull/466). So I made [`core-js-compat`](https://github.com/zloirock/core-js/tree/master/packages/core-js-compat) package with required data which should be used here.
- Built-ins definitions should be updated for include correct dependencies. Some parts of that were added before `babel@7` release, some still not.
- All syntax features that work with iterators should add common iterators, async functions / generators should add `Promise` with dependencies (#9250), etc.

#### Polyfilling of proposals

In the final `babel@7` versions, from all polyfilling-related stuff was removed polyfilling of proposals. Seems that it was a mistake.

New babel plugins for proposals **will** rely on new standard library features. And, with the current approach to polyfilling of proposals, it's too problematic.

I can understand why polyfilling of proposals was removed from the default `@babel/polyfill`. But in this case, we should have added entry points which include proposals.

Adding polyfills of proposals from `core-js` when the main part of polyfills loaded from `@babel/polyfill` is not a good solution because they could be loaded from another instance of `core-js` and it will cause problems.

Removing proposals from `runtime` will kill usage of new proposal plugins which depends on new standard library features with the `runtime`. They should be available in helpers. Even import of required part of `core-js` without global namespace pollution is rocket science for a serious part of developers. Also, `@babel/runtime-corejsx` is not `@babel/runtime-stable-corejsx-features`.

The same for `@babel/preset-env`, but not so critical.

`@babel/preset-env` transforms `core-js` only to import of stable ES features. But the main `core-js` entry point also contains proposals and someone can expect some of them.

`runtime` and `preset-env` inject polyfills based on feature detection, only when they are used, so I don't see any arguments why polyfilling of proposals was removed from those tools.

The `shippedProposals` option of `@babel/preset-env` looks like something strange. We haven't any equal in `core-js` entry points, so I don't see what should be transformed. Implementing in browser definitely is not an indicator that it should be polyfilled, but non-implemented proposals should not. I don't think that it's something that should be supported.

So, by my vision, the best approach is:
- Deprecate `@babel/polyfill` in favor recommendation of separate inclusion of required parts of `core-js` and `babel-runtime`. If someone needs polyfills of proposals - he will use polyfills of proposals, not - not.
- Deprecate transform of `@babel/polyfill` in `@babel/preset-env`.
- Add to `preset-env` with `useBuiltIns: entry` (or a successor for polyfilling) transforms: `core-js` to all modules, `core-js/es`, `core-js/proposals` and `core-js/web` to related parts of `core-js`.
- Allow `runtime` and `preset-env` inject proposals polyfills. Maybe - with an option for that, but option will make it harder.

#### --------------------

Before the next step, we should decide what should we do with `@babel/polyfill`, `@babel/preset-env` breaking changes and polyfilling of proposals. Thoughts?


packages/babel-preset-env/src/use-built-ins-plugin.js
```diff
@@ -119,22 +119,28 @@ export default function({ types: t }: { types: Object }): Plugin {
         return;
       }
 
-      addImport(path, "web.dom.iterable", this.builtIns);
+      addImport(path, "es.array.iterator", this.builtIns);
+      addImport(path, "es.string.iterator", this.builtIns);
+      addImport(path, "web.dom-collections.iterator", this.builtIns);
```
zloirock: Is this really required if we have those dependencies for `iterator` static property of `Symbol` in `built-in-definitions.js`?

packages/babel-preset-env/src/use-built-ins-plugin.js
```diff
@@ -119,22 +119,28 @@ export default function({ types: t }: { types: Object }): Plugin {
...
     },
 
     // yield*
     YieldExpression(path) {
       if (!path.node.delegate) return;
 
-      addImport(path, "web.dom.iterable", this.builtIns);
+      addImport(path, "es.array.iterator", this.builtIns);
+      addImport(path, "es.string.iterator", this.builtIns);
+      addImport(path, "web.dom-collections.iterator", this.builtIns);
```
zloirock: Should it also be added to `ForOfStatement`and other features which works with iterators?

packages/babel-preset-env/src/use-built-ins-plugin.js
existentialism: I don't think it is... iirc, @hzoo added this as just a general safety

packages/babel-plugin-transform-runtime/src/definitions.js
```diff
@@ -109,32 +106,19 @@ export default {
...
-      trimRight: "string/trim-right",
-      trimStart: "string/trim-start",
-      trimEnd: "string/trim-end",
     },
 
     Number: {
       EPSILON: "number/epsilon",
+      MAX_SAFE_INTEGER: "number/max-safe-integer",
+      MIN_SAFE_INTEGER: "number/min-safe-integer",
+      formString: "number/from-string",
```
Kovensky: typo; `fromString`

packages/babel-preset-env/src/built-in-definitions.js
```diff
@@ -1,140 +1,259 @@
+const ArrayNatureIterators = [
+  "es.array.iterator",
+  "web.dom-collections.iterator",
```
Kovensky: I assume this correctly bails out if not on a browser?

Would be nice to be able to omit it if `targets` did not include a browser, but the <1KB you'd save is probably not worth the effort.

packages/babel-preset-env/src/built-in-definitions.js
```diff
@@ -1,140 +1,259 @@
...
+    __lookupGetter__: ["es.object.lookup-getter"],
+    __lookupSetter__: ["es.object.lookup-setter"],
   },
 
   staticMethods: {
     Array: {
-      from: "es6.array.from",
-      isArray: "es6.array.is-array",
-      of: "es6.array.of",
+      from: ["es.array.from", "es.string.iterator"],
```
Kovensky: Nice catch; I remember having encountered this problem before but I never remembered to report it... 😇

packages/babel-preset-env/src/built-in-definitions.js
```diff
@@ -1,140 +1,259 @@
...
-      assign: "es6.object.assign",
-      is: "es6.object.is",
-      getOwnPropertySymbols: "es6.object.get-own-property-symbols",
-      setPrototypeOf: "es6.object.set-prototype-of",
-      values: "es7.object.values",
-      entries: "es7.object.entries",
-      getOwnPropertyDescriptors: "es7.object.get-own-property-descriptors",
+      assign: "es.object.assign",
+      create: "es.object.create",
+      defineProperty: "es.object.define-property",
```
Kovensky: I wonder; will, in core-js 3, it be possible to omit including the polyfill for this if your target does not include a browser that needs it?

Right now, in core-js 2, even if you don't include it via preset-env, it'll be overwhelmingly likely that whatever feature you do include the polyfill of will indirectly require the defineProperty polyfill.

It's true that it's safer that way, but it's unfortunate that you can't get significantly smaller code size even when you're specifically being unsafe by using this instead of just loading all of `core-js`.

packages/babel-preset-env/src/built-in-definitions.js
```diff
@@ -1,140 +1,259 @@
...
+      isNaN: "es.number.is-nan",
+      parseFloat: "es.number.parse-float",
+      parseInt: "es.number.parse-int",
+      EPSILON: "es.number.epsilon",
+      MIN_SAFE_INTEGER: "es.number.min-safe-integer",
+      MAX_SAFE_INTEGER: "es.number.max-safe-integer",
     },
 
     Promise: {
-      all: ["es6.string.iterator", "web.dom.iterable"],
```
Kovensky: I never thought of giving a string to `Promise.all` 🤣 

```js
await Promise.all("a string")
// => Array(8) [ "a", " ", "s", "t", "r", "i", "n", "g" ]
```

existentialism:
@zloirock this is looking great, mind if I regen and push the test fixtures so we can see the effects?

packages/babel-preset-env/src/built-in-definitions.js
zloirock: I hope that it should be removed for a non-browser target, otherwise we should add logic for removing it.

zloirock:
@existentialism makes sense, I wanted to add some more changes and after that work on stabilisation and tests.

packages/babel-preset-env/src/built-in-definitions.js
zloirock: No, if any polyfill requires this logic - it will be added as an internal module. And it's about almost all fundamental ES5 features.

It's related [this issue](https://github.com/zloirock/core-js/issues/388). You talk about the second option from [this comment](https://github.com/zloirock/core-js/issues/388#issuecomment-377818011). It's not hard for implementation, but it will cause too many problems.

For example, `Array#fill` from the issue. It requires not just `Object.defineProperty` for adding this polyfill, but, for example, `Object.create` for adding `Array#[@@unscopables]` where should be the name of this method. `Object.create` depends on `Object.defineProperties`, `Object.defineProperties` depends on `Object.getOwnPropertyNames`, etc.

So for working `es.array.fill` in IE8 we should add all of those dependencies. It's critically hard to determinate this list for manual adding. It's not so critical for `babel-preset-env`, but the list of internal dependencies could be changed in `core-js`, for example, after fixing any engine bug and it could be a breaking change for `babel-preset-env`.

I wanted to propose this issue to discuss before `core-js@4`. And it's the main problem which can be some time solved by dropping IE8 support in `core-js`.

packages/babel-helpers/src/helpers.js
```diff
@@ -943,9 +943,7 @@ helpers.applyDecoratedDescriptor = () => template.program.ast`
         }
 
         if (desc.initializer === void 0){
-            // This is a hack to avoid this being processed by 'transform-runtime'.
-            // See issue #9.
-            Object['define' + 'Property'](target, property, desc);
+            Object.defineProperty(target, property, desc);
```
zloirock: @loganfsmyth I hope that it should not cause a problem now?

zloirock: (related https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy/issues/9)

loganfsmyth: Yeah I think this is fine now.

sujiangyin:
I meet the same error like：
`if('get'` in Attributes || 'set' in Attributes)throw TypeError('Accessors not supported!');`
hen I load up the page in IE8,and I use webpack3.xx and babel. any ways for me to handle this?thx

zloirock:
@suarasaur support of getters and setters can't be polyfilled or transpiled in IE8-, so you can't use it or use transforms which use it. Webpack use it for modules, so, in this case, I recomment to use `babel` `modules-commonjs` transform with loose mode option.

dan-kez:
Would love to see this in babel 7! In core v3 they remove the dependency on `global.core` which is a fantastic addition

zloirock:
In work again. Because some problems were not fixed before `babel@7` release, now we have some serious problems. I'll add a full explanation of the current situation after some time.

zloirock:
To the head of this thread added explainer of the current situation with updating `babel` to `core-js@3` and we should decide what should we do.

nicolo-ribaudo:
@zloirock WDYT about keeping the discussions about "How should we integrate core-js@3" and "Should we polyfill proposals?" separate?

zloirock:
@nicolo-ribaudo these two topics are closely related to each other. I wanted to continue the discussion about both topics at the next meeting.

tstirrat15:
What's the status on this PR? I'd love to see this go through, specifically to get access to the `URL` polyfill in my project. Is there anything I can do to nudge the process along?

tstirrat15:
Will that take advantage of the way that babel-preset-env includes or excludes core-js imports based on whether they're used?

nicolo-ribaudo:
Until this PR is merged, unfortunatly it won't :(

zloirock:
It's not completely finished but works as I'd expected and ready for review.

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -209,30 +236,59 @@ export default declare((api, options, dirname) => {
...
+        const { node } = path;
+        const { callee } = node;
 
-        const callee = path.node.callee;
         if (!t.isMemberExpression(callee)) return;
+
+        const { object, property } = callee;
+
+        // transform array.includes()
+        if (!injectCoreJS2 && !hasStaticMapping(object.name, property.name)) {
```
nicolo-ribaudo: Nit: it would be more clear to use `injectCoreJS3` instead of `!injectCoreJS2`

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -209,30 +236,59 @@ export default declare((api, options, dirname) => {
...
+          if (hasMapping(definitions.instanceMethods, property.name)) {
+            let context1, context2;
+            if (object.type === "Identifier") {
+              context1 = context2 = object;
```
nicolo-ribaudo: We might need to use `t.cloneNode` here, I'm not sure. You can tell it if some tests are failing complaining about duplicated nodes.

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -209,30 +236,59 @@ export default declare((api, options, dirname) => {
...
-        if (path.node.arguments.length) return;
+        const { node } = path;
+        const { callee } = node;
 
-        const callee = path.node.callee;
         if (!t.isMemberExpression(callee)) return;
+
+        const { object, property } = callee;
+
+        // transform array.includes()
```
nicolo-ribaudo: Nit: These comments (also the one above about `Symbol` and `new Promise` and the one about `Array.from`) can be made more clear, like `// Transform instance methods, e.g. [].includes()`.

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -103,7 +111,7 @@ export default declare((api, options, dirname) => {
     } else {
       throw new Error(
         "The 'polyfill' option has been removed. Use the 'corejs'" +
-          "option with value '2' to polyfill with CoreJS 2.x via @babel/runtime.",
+          "option to polyfill with CoreJS via @babel/runtime.",
```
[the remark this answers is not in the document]
zloirock: Yep, I just didn't change what's already was here.

packages/babel-plugin-transform-runtime/src/index.js
zloirock: In the future, we can have, for example, `injectCoreJS4` -)

packages/babel-plugin-transform-runtime/test/fixtures/runtime-corejs3/math/options.json
```diff
@@ -0,0 +1,12 @@
+{
+  "plugins": [
+    [
+      "transform-runtime",
+      {
+        "corejs": 3,
+        "version": "^7.0.2"
```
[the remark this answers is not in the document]
zloirock: It's just a copy of `@babel/runtime-corejs2` tests, need to remove it.

packages/babel-plugin-transform-runtime/src/index.js
nicolo-ribaudo: I think that at that time it would be better to have `injectCoreJS >= 3` 😛

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -39,10 +40,9 @@ export default declare((api, options, dirname) => {
     useESModules = false,
     version: runtimeVersion = "7.0.0-beta.0",
     absoluteRuntime = false,
+    proposals = false,
```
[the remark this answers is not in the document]
zloirock: > I'd like to see something like this in the future

Yep, something like that could be a great addition in the future.

> Maybe we could rename `proposals` to `experimentalProposals`, to make it even more clear that those things aren't stable yet?

`proposals` already mean that they are not stable -) I think to rename it to something like `builInsProposals` for by the reason of consistency with `preset-env` where the same option works only with polyfills.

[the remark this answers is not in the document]
nicolo-ribaudo: Maybe `corejsProposals`? So that it is obvious that it only affects the `corejs` option.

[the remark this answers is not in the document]
zloirock: It's not a proposal to or from `core-js`, so I'm not sure.

packages/babel-preset-env/CONTRIBUTING.md
```diff
@@ -1,6 +1,6 @@
 # Contributing
 
-## Adding a new plugin to support (when approved in the next ECMAScript version)
+## Adding a new plugin of polyfill to support (when approved in the next ECMAScript version)
```
nicolo-ribaudo:
```suggestion
## Adding a new plugin or polyfill to support (when approved in the next ECMAScript version)
```

packages/babel-preset-env/CONTRIBUTING.md
```diff
@@ -38,19 +38,31 @@ and split it with `/`:
 
 `Object static methods / Object.values`
 
-Find the corresponding module on [core-js](https://github.com/zloirock/core-js/tree/master/modules):
+Find the corresponding module on [`core-js@2`](https://github.com/zloirock/core-js/tree/v2/modules):
 
-`es7.object.values.js`
+`es.object.values.js`
```
nicolo-ribaudo: Doesn't `core-js@2` still use `esNUMBER` prefixes?

packages/babel-preset-env/CONTRIBUTING.md
```diff
@@ -38,19 +38,31 @@ and split it with `/`:
...
 
-Find required ES version in [`built-in-features.js`](https://github.com/babel/babel/blob/master/packages/babel-preset-env/data/built-in-features.js) and add the new feature:
+Find required ES version in [`corejs2-built-in-features.js`](https://github.com/babel/babel/blob/master/packages/babel-preset-env/data/corejs2-built-in-features.js) and add the new feature:
 
 ```js
-const es2017 = {
+const es = {
   //...
-  "es7.object.values": "Object static methods / Object.values"
+  "es6.object.values": "Object static methods / Object.values"
```
nicolo-ribaudo: This is es7: https://github.com/zloirock/core-js/blob/e778e8026aed8a58f93f1ee4e3192cd1a7d7bdf5/modules/es7.object.values.js

packages/babel-preset-env/src/debug.js
```diff
@@ -43,50 +37,56 @@ export const logPlugin = (item, targetVersions, list, context) => {
...
     return;
   }
   console.log(
-    `
-[${filename}] Added following polyfill${wordEnds(polyfills.size)}:`,
+    `\n[${filename}] Added following core-js polyfill${wordEnds(
+      polyfills.size,
+    )}:`,
   );
-  onDebug(polyfills);
```
nicolo-ribaudo: This wasn't part of the public API, right?

packages/babel-preset-env/src/filter-items.js
```diff
@@ -0,0 +1,78 @@
+import semver from "semver";
```
nicolo-ribaudo: (NOTE for other reviewers)

This file isn't new, it has just been extracted from `babel-preset-env/src/index.js`

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -1,17 +1,15 @@
-//@flow
```
nicolo-ribaudo: Why this change?

packages/babel-preset-env/CONTRIBUTING.md
zloirock: Hmm... Seems I forgot to revert it.

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -154,19 +146,7 @@ export const validateModulesOption = (
   return modulesOpt;
 };
 
-export const objectToBrowserslist = (object: Targets): Array<string> => {
```
nicolo-ribaudo: (NOTE for reviewers)

This has just been moved to `babel-preset-env/src/targets-parser.js`

packages/babel-preset-env/CONTRIBUTING.md
zloirock: Yep, seems I made some typos when I revert it to `core-js@2`.

packages/babel-preset-env/src/polyfills/corejs2/usage-plugin.js
```diff
@@ -0,0 +1,224 @@
...
+        path.node.specifiers.length === 0 &&
+        isPolyfillSource(path.node.source.value)
+      ) {
+        console.warn(NO_DIRECT_POLYFILL_IMPORT);
+        path.remove();
+      }
+    },
+
+    Program: {
+      enter(path) {
```
nicolo-ribaudo: Nit: You can use `Program(path)`, which is the same as `Program: { enter(path)`

packages/babel-preset-env/src/polyfills/corejs2/usage-plugin.js
```diff
@@ -0,0 +1,224 @@
...
+    getPlatformSpecificDefaultFor(polyfillTargets),
+  );
+
+  const addAndRemovePolyfillImports = {
+    ImportDeclaration(path) {
+      if (
+        path.node.specifiers.length === 0 &&
+        isPolyfillSource(path.node.source.value)
+      ) {
+        console.warn(NO_DIRECT_POLYFILL_IMPORT);
```
nicolo-ribaudo: I think that this would warn providing the context, but it needs to be checked.

```js
console.warn(
  path.buildCodeFrameError(NO_DIRECT_POLYFILL_IMPORT)
)
```

packages/babel-preset-env/src/polyfills/corejs2/usage-plugin.js
```diff
@@ -0,0 +1,224 @@
...
+        path.remove();
+      }
+    },
+
+    Program: {
+      enter(path) {
+        path.get("body").forEach(bodyPath => {
```
nicolo-ribaudo: Instead of looping through Program's body, what about visiting `CallExpression`s?

packages/babel-preset-env/src/debug.js
zloirock: If internal `preset-env` plugins are not public API - not.

packages/babel-preset-env/src/polyfills/corejs2/usage-plugin.js
```diff
@@ -0,0 +1,224 @@
...
+            InstancePropertyDependencies = InstancePropertyDependencies.filter(
+              module => module.includes(instanceType),
+            );
+          }
+          this.addUnsupported(InstancePropertyDependencies);
+        }
+      },
+
+      // Symbol.match
+      exit(path) {
```
nicolo-ribaudo: Why do we need to do this on `exit`?

packages/babel-preset-env/src/polyfills/corejs2/usage-plugin.js
```diff
@@ -0,0 +1,224 @@
...
+        if (path.scope.getBindingIdentifier(name)) return;
+
+        const BuiltInDependencies = BuiltIns[name];
+        this.addUnsupported(BuiltInDependencies);
+      },
+    },
+
+    // var { repeat, startsWith } = String
+    VariableDeclarator(path) {
+      if (!path.isReferenced()) return;
```
nicolo-ribaudo: Also this, when can it happen?

packages/babel-preset-env/src/normalize-options.js
zloirock: In one moment of refactoring, support of `flow` created some serious problems and I just disabled it. Maybe in the final version, I'll revert it.

packages/babel-preset-env/src/polyfills/corejs2/usage-plugin.js
```diff
@@ -0,0 +1,224 @@
...
+
+      const { node } = path;
+      const { id, init } = node;
+
+      if (!t.isObjectPattern(id)) return;
+      if (!t.isReferenced(init, node)) return;
```
nicolo-ribaudo: This is always true, the `isReferenced` function checks if `node.init === init` in this case.

packages/babel-preset-env/src/polyfills/corejs2/usage-plugin.js
zloirock: This and 2 above comments above - it's a code from current `preset-env` code, I didn't change it.

About of looping through Program's body - it's related the only top-level `require`, so I think it can be better than visiting `CallExpression`.

packages/babel-preset-env/src/polyfills/corejs2/usage-plugin.js
zloirock: The same and previous comments - it from the current version of `preset-env`.

packages/babel-preset-env/src/utils.js
```diff
@@ -77,56 +95,97 @@ export const prettifyTargets = (targets: Targets): Object => {
...
   }, {});
-};
+}
+
+export function isPolyfillSource(source) {
+  return source === "@babel/polyfill" || source === "core-js";
+}
+
+export function isBabelPolyfillSource(source) {
+  return source === "@babel/polyfill" || source === "babel-polyfill";
```
nicolo-ribaudo: I don't think that we should support `babel-polyfill` in v7

zloirock: This helper used only for deprecation message on `babel-polyfill` with `corejs: 3`.

nicolo-ribaudo:
Please ignore the overall failing status of this PR, but check travis/circleci statuses. I'm not sure why GitHub is marking it as such (and only for this PR), but it is related to the old action I tried to use for publishing.

schermata da 2019-02-18 09-26-26

packages/babel-preset-env/src/polyfills/corejs3/usage-plugin.js
```diff
@@ -84,33 +86,56 @@ export default function(
...
     // for-of loop
     ForOfStatement() {
       this.addUnsupported(CommonIterators);
     },
 
-    // spread
-    ArrayExpression({ node }) {
-      if (node.elements.some(el => t.isSpreadElement(el))) {
-        this.addUnsupported(CommonIterators);
+    ArrayPattern() {
```
nicolo-ribaudo: Nit: you can use `"ArrayPattern|SpreadElement"` and have a single visitor.

packages/babel-preset-env/src/polyfills/corejs3/usage-plugin.js
```diff
@@ -84,33 +86,56 @@ export default function(
...
+    },
+
+    ObjectPattern({ node, parent, scope }) {
+      let canBeStatic, builtIn, right;
+      if (t.isVariableDeclarator(parent)) {
+        right = parent.init;
+      } else if (t.isAssignmentExpression(parent)) {
+        right = parent.right;
       }
+      if (right) {
```
nicolo-ribaudo: We should check if `right && t.isIdentifier(right)`

packages/babel-preset-env/src/polyfills/corejs3/usage-plugin.js
```diff
@@ -84,33 +86,56 @@ export default function(
...
+          has(StaticProperties, builtIn) &&
+          !scope.getBindingIdentifier(builtIn);
+      }
+
+      for (const { key } of node.properties) {
+        const name = t.isIdentifier(key)
+          ? key.name
+          : t.isStringLiteral(key)
+          ? key.value
+          : null;
```
nicolo-ribaudo: I think that this might have problems with things like
```js
let trim = "toString";
const { [trim]: toString } = ""; 
```
Since the key is an `Identifier`, it will add the polyfill for `"".trim`.

Maybe something like this?
```js
for (const { key, computed } of node.properties) {
  const name = !computed && t.isIdentifier(key)
      	         ? key.name
                 : t.isStringLiteral(key)
                 ? key.value
                 : null;
}
```

zloirock: I rewrote it in another way yesterday and will push it in an hour. Ralated the above comment too.

packages/babel-preset-env/src/polyfills/corejs3/usage-plugin.js
zloirock: It can be not just an identifier.

packages/babel-preset-env/test/fixtures/corejs3/usage-in/output.mjs
```diff
@@ -0,0 +1,10 @@
+import "core-js/modules/es.array.includes";
+import "core-js/modules/es.array.iterator";
+import "core-js/modules/es.object.entries";
+import "core-js/modules/es.object.from-entries";
+import "core-js/modules/es.object.values";
+'entries' in Object;
```
nicolo-ribaudo: I'm not sure about this. I think that if someone uses
```js
if ("entries" in Object)
```
or
```js
if (typeof Symbol.iterator !== "undefined")
```
they don't want to load the polyfill: they are explicitly checking if that feature is supported.

Ref: https://github.com/babel/babel/pull/8947

zloirock: However, they should be loaded for preventing conflicts. As you can see in #8947, they should be resolved as `true`.

Jessidhia:
If, by accident, one attempts to load something from `core-js@2` after `core-js@3` has evaluated, what would happen? Would `core-js@2` detect the `core-js@3` version as compliant?

Mainly wondering about potential dependencies that were compiled with babel 6 transform-runtime, or that specifically used things from `core-js/library/...`.

zloirock:
@Jessidhia 
- `core-js@2` and `core-js@3` API are incompatible. And they can't be compatible for `runtime` case since a version without global namespace pollution moved to a separate `core-js-pure` package.
- This PR should not break anything. `@babel/polyfill` will be deprecated and left on `core-js@2`. `@babel/preset-env` by default will use `core-js@2` and will show a message that the user should specify `core-js` version directly.
- `@babel/runtime` and `@babel/preset-env` will allow using required version of `core-js`, 2 or 3.
- Only one problem case - if someone has `@babel/polyfill` instead of `core-js` dependency, but use `core-js@2` entry points (for example, because of transpiled by `@babel/preset-env` code) and somehow his top-level dependency will be resolved as `core-js@3`. It's a mistake of this user and we have no way to prevent it and it will happen even without this PR.

Jessidhia:
I'm specifically thinking of cases where a transitive dependency happens.

For example, this is how the folder structure would look like as installed by yarn or modern npm:

```sh
node_modules/@babel/runtime # core-js@3 version
node_modules/core-js # core-js@3
node_modules/core-js-pure # also @3
node_modules/some-old-package # was compiled with babel 6 transform-runtime
node_modules/babel-runtime # transitive dependency of some-old-package
node_modules/babel-runtime/node_modules/core-js # core-js@2
node_modules/less-old-package # was compiled with babel 7 transform-runtime, ok
node_modules/another-old-package # directly uses core-js/library
node_modules/another-old-package/node_modules/core-js # core-js@2
```

You'd of course end up with multiple copies of core-js, which is bad, but at least if `core-js@3` (non-pure) loads first, hopefully the copies of `core-js@2/library` would just detect the polyfills as valid implementations and not polyfill them.

zloirock:
It will require 1-2 thousands of additional entry points in `core-js`, so it's problematically. Agree, it's a problem, but just in the possible increasing size of the bundle - it should work fine. I don't see any way to prevent it. However, this problem can speed up updating.

Jessidhia:
I had actually looked into writing a webpack plugin to alias `core-js@2` paths into `core-js@3` but indeed it is a big undertaking ~~and you keep doing updates~~ :)

The specific compatibility worry here is not the bundle size (that'd be what an alias plugin could solve) but on whether loading a copy of core-js@2 after another copy of core-js (2 or 3) would cause runtime conflicts.

zloirock:
It should not cause conflicts.

nicolo-ribaudo:
Does this PR inject `Symbol.asyncIterator` when it finds an async generator or a `for await` loop?

zloirock:
@existentialism `usage` plugin works after all of the rest plugins. If native async generators available and this syntax was not transformed before - `Symbol.asyncIterator` is not required, if not - `Symbol.asyncIterator` will be be used in helpers and should be injected without any special cases.

nicolo-ribaudo:
Is this PR ready to be reviewed? We would like to release a minor version the next week or maybe the following one.

zloirock:
@nicolo-ribaudo sure, it's ready for review. Only one change required - renaming `proposals` option of `runtime` and `preset-env`. However, need to prepare documentation for all changes. I have time to prepare it for `core-js` release, but what about `babel`?

nicolo-ribaudo:
I can do it

nicolo-ribaudo:
What do you think about merging the `corejs` and `proposal` options?

i.e.
```js
["@babel/plugin-transform-runtime", {
  corejs: { version: 2 }
}]

["@babel/plugin-transform-runtime", {
  corejs: { version: 3, proposals: false /* default */ }
}]

["@babel/plugin-transform-runtime", {
  corejs: { version: 3, proposals: true }
}]
```

otherwise I'm ok with the option as currently implemented in this PR

zloirock:
@nicolo-ribaudo it's one of the options which I'm considering. At least for consistency with `runtime`, `corejs` option should also pass string or number with a version, not only object. Why I still didn't merge it - it looks more verbose than just 2 options. However, it's more logical.

packages/babel-preset-env/scripts/build-data.js
```diff
@@ -186,20 +181,7 @@ const getLowestImplementedVersion = ({ features }, env) => {
     }, []);
 
   const unreleasedLabelForEnv = unreleasedLabels[env];
-  const envTests = tests.map(({ res: test, isBuiltIn }, i) => {
-    // Babel itself doesn't implement the feature correctly,
-    // don't count against it
-    // only doing this for built-ins atm
-    //
-    // NOTE: when/if compat-table adds a babel7 key, we'll want to update this
-    if (!test.babel6corejs2 && isBuiltIn) {
```
nicolo-ribaudo: Why is this check removed?

zloirock: The main reason for removing this check was that this script will not generate correct output before adding unstable `core-js` version results to `compat-table`. Since now it's used only for generating data for `core-js@2`, it's not so actual. However, some features can't be fully implemented and with this check, polyfills for them would be ignored. So I don't see any reason leave this additional check for built-ins which anyway ignored for syntax.

nicolo-ribaudo: Ok :+1:

nicolo-ribaudo:
Does this fixes https://github.com/babel/babel/issues/6870?

dan-kez:
I don't want to derail conversation here but will this PR enable the ability to set the name of the global object used by core js? IIRC this is a new feature in core-js 3 and would love to be able to use it via babel once this is released.

zloirock:
@nicolo-ribaudo yes.

zloirock:
@dan-kez which object do you mean? if something like `window.core` - `core-js@3` no longer set it. `window['__core-js_shared__']` - `core-js@3` still use it since we have no more way for communication in some cases.

dan-kez:
> @dan-kez which object do you mean? if something like `window.core` - `core-js@3` no longer set it. `window['__core-js_shared__']` - `core-js@3` still use it since we have no more way for communication in some cases.

That's perfect - I am most concerned about the `window.core` variable. Thank you!

zloirock:
I'm concerned about an issue related latest `regenerator-runtime` which no longer set global `regeneratorRuntime` variable. The actual version of this PR works just with `import 'regenerator-runtime/runtime'` for setting global variable...

nicolo-ribaudo:
cc @benjamn How do you suggest handling regenerator-runtime?

danez:
Do we need to update regeneratorRuntime in this PR? We also have a separate PR for it #9442

zloirock:
@danez see https://github.com/babel/babel/pull/9442#issuecomment-468928043. It's too problematic.

nicolo-ribaudo:
Currently `preset-env` doesn't remove `regenerator-runtime`: can't we keep this behabior?

(On master, not in this PR):
```js
const code = `
import 'regenerator-runtime/runtime';
import 'core-js';


function* gen() {}
`;

const out = babel.transformSync(code, {
  configFile: false,
  presets: [
    [
      load("preset-env"),
      {
        useBuiltIns: "entry",
        modules: false,
        targets: { chrome: 65 },
      },
    ],
  ],
});

console.log(out.code);
```
This code outputs
```js
import "core-js/modules/web.timers";
import "core-js/modules/web.immediate";
import "core-js/modules/web.dom.iterable";
import 'regenerator-runtime/runtime';

function* gen() {}
```

Regenerator isn't needed, but as you can see it isn't removed.

zloirock:
> Currently `preset-env` doesn't remove `regenerator-runtime`: can't we keep this behavior?

Since we deprecate `@babel/polyfill`, we should somehow remove `regenerator-runtime` for targets where `regenerator` is not used, otherwise, we will have a useless heavy dependency in bundles.

> Anyway, I think that if `regenerator-runtime` exposed something like `import "regenerator-runtime/global"` it would be way easier.

Agree.

nicolo-ribaudo:
If @benjamn doesn't want to add `regenerator-runtime/global` to his package (that functinality was removed because of some problem when detecting the global object), we could use a "fake" `regenerator-runtime/preset-env/entry` which is replaced by preset-env with
```js
import globalThis from 'core-js-pure/features/global-this';
import regeneratorRuntime from 'regenerator-runtime';
globalThis.regeneratorRuntime = regeneratorRuntime;
```
or removed when not needed.

zloirock:
@nicolo-ribaudo I don't think that a fake entry point is a solution. The user should be able to use `preset-env` without changing the code, just by changing the config. And it will not solve a problem replacement for `@babel/polyfill` without `preset-env`.

benjamn:
@nicolo-ribaudo @zloirock What if we changed [this line](https://github.com/facebook/regenerator/blob/70cde2f55d82c9ba62f5860182e2b93f90922358/packages/regenerator-runtime/runtime.js#L8):
```js
var regeneratorRuntime = (function (exports) {
```
to an unguarded global assignment, by simply removing the `var`:
```js
regeneratorRuntime = (function (exports) {
```
The `runtime.js` script isn't in strict mode at the top level, so this seems like the easiest place to create the global variable without having to import a polyfill for `globalThis`, or use the `Function` constructor, or `typeof`-detect a bunch of possible identifiers (`window`, `global`, `self`, etc.).

Also, when `runtime.js` is used as a module, it will be evaluated only once, so I don't think we have to worry about it assigning to `regeneratorRuntime` multiple times.

benjamn:
Alternatively:
```diff
diff --git a/packages/regenerator-runtime/runtime.js b/packages/regenerator-runtime/runtime.js
index 9cec7aec6..16ebd93c9 100644
--- a/packages/regenerator-runtime/runtime.js
+++ b/packages/regenerator-runtime/runtime.js
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-var regeneratorRuntime = (function (exports) {
+var runtime = (function (exports) {
   "use strict";
 
   var Op = Object.prototype;
@@ -709,3 +709,11 @@ var regeneratorRuntime = (function (exports) {
   // the regeneratorRuntime variable at the top of this file.
   typeof module === "object" ? module.exports : {}
 ));
+
+try {
+  regeneratorRuntime = runtime;
+} catch (ignored) {
+  // If we can't assign the global regeneratorRuntime variable for any
+  // reason (such as a bundling strategy that mistakenly puts this module
+  // in strict mode), that's fine. At least we tried.
+}
```

zloirock:
@benjamn someone definitely will apply babel modules transform to dependencies, it will add `'use strict'` to this code and it will break this hack.

After [some explorations](https://github.com/zloirock/core-js/issues/86#issuecomment-115759028), I use [this global object detection](https://github.com/zloirock/core-js/blob/master/packages/core-js/internals/global.js).

As an option, you can import `globalThis` polyfill from `core-js` and it will be removed for targets which supports it if you'll apply `preset-env` by technics from this PR.

nicolo-ribaudo:
What if instead of ignoring the exception when assigning to `regeneratorRuntime` we bring throw an helpful error like

```
It looks like you are running `regenerator-runtime/runtime` in strict mode,
which is not supported. This is probably caused by your bundler or compiler considering this module as an ECMAScript module instead of a CommonJS one.
If you can't configure it not to do so, you can use `[entry point which does
not throw]` which exports `regeneratorRuntime` instead of adding it to
the global object.
```

Jessidhia:
Ideally users should be using the non-global version of `regenerator-runtime`. Babel already does this with `transform-runtime`, but maybe it should just be done by `transform-regenerator` or by a `preset-env`-specific mechanism instead?

An alternative is to recommend `new webpack.ProvidePlugin({ regeneratorRuntime: 'regenerator-runtime' })`, but that only solves the problem for webpack.

nicolo-ribaudo:
Not everyone uses a bundler or runs their code on node: we can't inject imports by default.

zloirock:
@Jessidhia it could work only for `useBuiltIns: usage`, but not for `useBuiltIns: entry` and it does not solve `@babel/polyfill` replacement issue. Now you can combine `preset-env` with `useBuiltIns: usage` and `runtime` and have the same result.

benjamn:
As the maintainer of `regenerator-runtime`, I think I just need to make a call here.

If you're using `@babel/transform-runtime` as recommended, you won't be relying on a global `regeneratorRuntime`, so there won't be any problem. In other words, this problem only exists for folks using `@babel/polyfill` with the runtime transform, so that's the only use case we have to consider (as long as we don't break any other use cases).

My proposed solution fixes that problem, and the only objection is that strict mode might accidentally be turned on, but in that case my solution will throw, and (as @nicolo-ribaudo [suggests](https://github.com/babel/babel/pull/7646#issuecomment-470344435)) we can provide helpful instructions about how to fix the mistake. I'm not interested in supporting the use case of someone forcing my non-strict code to run in strict mode. That's a bundling bug. I would very much like to rely on simple global assignment, without having to use heuristics to detect the global object, and non-strict mode gives me that ability.

At the end of the day, we're trying to fix people's existing code without making them think about the details if possible. We know how to do that, and I'm happy to implement what I've suggested. We can continue this discussion in that PR thread.

benjamn:
Alright, please take a look! https://github.com/facebook/regenerator/pull/369

packages/babel-polyfill/package.json
```diff
@@ -12,6 +12,6 @@
   "main": "lib/index.js",
   "dependencies": {
     "core-js": "^2.6.5",
-    "regenerator-runtime": "^0.12.0"
+    "regenerator-runtime": "^0.13.2"
```
nicolo-ribaudo: I'd like to do this in #9442 to give credits for this update to @MattiasBuelens.
`^0.12.0` still loads the old regenerator version anyway, which added `regeneratorRuntime` to the global object.

zloirock: I don't care. Anyway, it's also required here since `@babel/runtime-corejs3` should be updated too.

danez:
The size increase in `preset-env-standalone` is probably because we bundle core-js 2 and 3 now?

packages/babel-polyfill/package.json
danez: If it is required here, I would do it here and give credit in changelog.

zloirock:
@danez mainly because we bundle `core-js-compat` (info about compatibility and entry points).

nicolo-ribaudo:
While writing docs, I realized that we should throw if `corejs` is set but `useBuiltIns` is false.

zloirock:
@nicolo-ribaudo I think that just a warning would be better.

packages/babel-preset-env/src/utils.js
```diff
@@ -77,56 +93,62 @@ export const prettifyTargets = (targets: Targets): Object => {
...
   const hasUnreleased = [a, b].some(item => item === unreleasedLabel);
   if (hasUnreleased) {
     return a === hasUnreleased ? b : a || b;
   }
   return semverMin(a, b);
-};
+}
 
-export const filterStageFromList = (list: any, stageList: any) => {
+export function filterStageFromList(list, stageList) {
```
danez: We should keep the flow types. What is the reason you removed all of them?

zloirock: @danez  see https://github.com/babel/babel/pull/7646#discussion_r257186983

danez: I saw that, but disabling a file is something different than removing all existent type declarations. I'm fine to disable a file for now, but removing all types seems unnecessary.

nicolo-ribaudo: We should check if now that everything is finished flow types can be re-introduced without type warnings.
@zloirock If you prefer, I can do that.

existentialism: I can also help with restoring flow types, if needed!

zloirock: @nicolo-ribaudo that could be great -)

nicolo-ribaudo:
3b706ad7d2c30c08aaf199750bea968258e28306 Is a breaking change.

zloirock:
@nicolo-ribaudo It should not be a breaking change. I have no ideas, why `web.dom.iterable` was added on `forEach` - it contains just iterators.

nicolo-ribaudo:
Ah ok

PR #7652, opened 2018-04-02
revert prev pr to temp fix osx

hzoo:
will just wait on CI and then do a release


hzoo:
All good, you would have no idea since no mac testing haha 😁

xtuc:
It would be worth adding it right? Something optional; I will PR something.

hzoo:
not a big deal since we would be a user anyway

PR #7653, opened 2018-04-02
add OSX on travis

xtuc:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

I guess that most of our user use Mac OS X? (excluding CIs)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,10 @@ matrix:
       env: JOB=babylon-flow-tests
     - node_js: "node"
       env: JOB=babylon-test262-tests
+    - os: osx
+      node_js: "node"
+  allow_failures:
+    - os: osx
 
 notifications:
   slack:


.travis.yml
```diff
@@ -41,6 +41,10 @@ matrix:
       env: JOB=babylon-flow-tests
     - node_js: "node"
       env: JOB=babylon-test262-tests
+    - os: osx
+      node_js: "node"
+  allow_failures:
+    - os: osx
```
xtuc: I made it optional so it doesn't impact the build time, we will an email for failure anyway.

loganfsmyth:
Is this still something we want to do? Seems reasonable to me as long as it doesn't slow things down in CI.

xtuc:
We could add windows/OS X in the future when we'll notice specific platform issues

PR #7655, opened 2018-04-02
Remove setting BABEL_ENV to test by default in makefile

### LINKED ISSUE #7654: @babel/traverse/lib/path/evaluation.js shipped with `**` use causing test fails in Node 6
It looks like the line [here](https://github.com/babel/babel/blob/v7.0.0-beta.43/packages/babel-traverse/src/path/evaluation.js#L309) is tripping up Node 6 because it failed to translate its use of `**` to something Node 6 can understand.

In [`@babel/traverse@7.0.0-beta.42`](https://unpkg.com/@babel/traverse@7.0.0-beta.42/lib/path/evaluation.js) it was
```js
return Math.pow(_left, _right);
```
In [`@babel/traverse@7.0.0-beta.43`](https://unpkg.com/@babel/traverse@7.0.0-beta.43/lib/path/evaluation.js) it is
```js
return left ** right;
```

existentialism:
| Fixed Issues?            | Fixes #7654
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

This was causing `BABEL_ENV=production make build-dist` to run with `test` env, which passed `node: current` as target to preset-env during the prod build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,6 @@ MAKEFLAGS = -j1
 FLOW_COMMIT = 622bbc4f07acb77eb1109830c70815f827401d90
 TEST262_COMMIT = 52f70e2f637731aae92a9c9a2d831310c3ab2e1e
 
-export BABEL_ENV = test
-
 # Fix color output until TravisCI fixes https://github.com/travis-ci/travis-ci/issues/7967
 export FORCE_COLOR = true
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -2,8 +2,6 @@ MAKEFLAGS = -j1
 FLOW_COMMIT = 622bbc4f07acb77eb1109830c70815f827401d90
 TEST262_COMMIT = 52f70e2f637731aae92a9c9a2d831310c3ab2e1e
 
-export BABEL_ENV = test
```
loganfsmyth: I think we'll want to set this down on line 72 and 82?

existentialism: Yep, caught that too.. just updated!

PR #7666, opened 2018-04-04
Private Properties phase 1

jridgewell:
Co-authored-by: CodingItWrong

We keep getting stalled on Class Private Properties. This is an attempt to take all the small parts into a separate PR, so we can make some progress.


packages/babel-generator/src/generators/classes.js
```diff
@@ -117,6 +117,21 @@ export function ClassProperty(node: Object) {
   this.semicolon();
 }
 
+export function ClassPrivateProperty(node: Object) {
```
nicolo-ribaudo: This should also print the decorators. Also, do Flow and Ts support type annotations?

packages/babel-types/src/definitions/experimental.js
```diff
@@ -115,6 +115,20 @@ defineType("OptionalCallExpression", {
...
+defineType("ClassPrivateProperty", {
+  visitor: ["key", "value"],
+  builder: ["key", "value"],
+  aliases: ["Property", "Private"],
+  fields: {
+    key: {
+      validate: assertNodeType("PrivateName"),
+    },
+    value: {
+      validate: assertNodeType("Expression"),
```
nicolo-ribaudo: `optional: true`?

PR #7669, opened 2018-04-05
Cleanup isReferenced a bit

jridgewell:
Noticed this a bit when making #7668. Combines a few cases, and fixes a "bug" in `ObjectProperty` which fell through to `VariableDeclarator`.

### DIFF
--- a/packages/babel-types/src/validators/isReferenced.js
+++ b/packages/babel-types/src/validators/isReferenced.js
@@ -7,7 +7,7 @@ export default function isReferenced(node: Object, parent: Object): boolean {
     // yes: object::NODE
     // yes: NODE::callee
     case "BindExpression":
-      return parent.object === node || parent.callee === node;
+      return true;
 
     // yes: PARENT[NODE]
     // yes: NODE.child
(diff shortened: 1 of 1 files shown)


packages/babel-types/src/validators/isReferenced.js
```diff
@@ -7,60 +7,50 @@ export default function isReferenced(node: Object, parent: Object): boolean {
...
     case "FunctionExpression":
+      if (parent.id === node) {
+        return false;
+      }
       for (const param of (parent.params: Array<any>)) {
         if (param === node) return false;
       }
-
-      return parent.id !== node;
+      return true;
```
nicolo-ribaudo: Can you add a `yes` comment which shows when this is true?

jridgewell: I actually don't think there is a case.... This can all be simplified down to `return false`.

packages/babel-types/src/validators/isReferenced.js
```diff
@@ -7,60 +7,50 @@ export default function isReferenced(node: Object, parent: Object): boolean {
     // yes: object::NODE
     // yes: NODE::callee
```
xtuc: Mostly nit but I noticed that sometimes we are grouping `return false` and some are not. Maybe could you regroup them?

packages/babel-types/src/validators/isReferenced.js
jridgewell: Stand corrected, `ArrowFunctionExpression`s can have a referenced body identifier: `() => NODE`

packages/babel-types/src/validators/isReferenced.js
jridgewell: Done.

PR #7671, opened 2018-04-05
Fix class properties after nested class' bare super

### LINKED ISSUE #7371: class property initializer of extended class is duplicated in nested extended class constructor
## This is a bug report

With the `@babel/plugin-proposal-class-properties` plugin, it appears that property initializers get duplicated in the nested class' constructor when both classes are extended classes that call super.

### Input Code
Minimum repro:
```js
class A extends C {
  B;
  constructor() {
    super();
    this.B = class B extends C {
      constructor() {
        super();
      }
    };
  }
}

```

```js
module.exports = {
  plugins: [
    ['@babel/plugin-proposal-class-properties', { loose: true }],
  ],
  presets: [
    ['@babel/preset-env', { targets: { node: 'current' } }],
  ],
};
```

### Expected Behavior
```js
class A extends C {
  constructor() {
    super();
    this.B = void 0;
    this.B = class B extends C {
      constructor() {
        super();
      }
    };
  }
}
```

### Current Behavior
```js
class A extends C {
  constructor() {
    super();
    this.B = void 0;
    this.B = class B extends C {
      constructor() {
        super();
        this.B = void 0; // THIS LINE SHOULDN'T BE HERE!
      }
    };
  }
}
```

example (same deal, but with `B = 1;`): https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAECC0FMAeAXBA7AJjAwtA3gFDTQBC0AvNAIwDcx0wA9uhCgE4CuwKT7AFAEoCDEhE4AHBAMH0SJFAAsAlhAB05KqEgxyyNFlwj585qw7deM4yfnipMubYC-o6M6evnQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=false&presets=stage-2&prettier=false&targets=&version=6.26.0&envVersion=

There is an extra `this.B = void 0;` which is problematic because it will make `this.B = undefined` if the inner class is instantiated at any point, overwriting `this.B = class ...`. 

### Possible Solution
There's probably some logic that just naively inserts the initializer based on the constructor pattern without accounting for context?

### Context
I am implementing some React components, where a nested component is defined during the parent's constructor call because it will need to read props from the parent. The reason why there's a property initializer that's set to `undefined` in the first place is for flow types.

### Your Environment

| ---------------- | -------
| Babel            |  7.0.0-beta.39
| Babylon          | 
| node             |  8.9.0
| npm              | 5.5.1
| Operating System | macos

jridgewell:
Fixes #7371.

Depends on #7691.


packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -10,10 +10,15 @@ export default declare((api, options) => {
...
     Super(path) {
-      if (path.parentPath.isCallExpression({ callee: path.node })) {
-        this.push(path.parentPath);
+      const { node, parentPath } = path;
+      if (parentPath.isCallExpression({ callee: node })) {
+        this.push(parentPath);
       }
     },
+
+    ClassBody(path) {
```
jridgewell: This should skip methods, not the entire body.

nicolo-ribaudo:
How should this code behave?

```js
class A extends B {
  x = 2;
  constructor() {
    class C extends D {
      y = super();
    }
  }
}
```

jridgewell:
It's an early error.

packages/babel-helper-replace-supers/src/index.js
```diff
@@ -39,7 +40,7 @@ const visitor = {
     // We do have to traverse the key, since it's evaluated in the outer class
     // context.
     if (path.node.computed) {
-      path.get("key").traverse(visitor, state);
+      path.get("key").traverse(path.context.opts, state);
```
jridgewell: I'm trying to create a base visitor that will be merged into several other complete visitors. This one defines the class environment traversal behavior (skip instance fields, class methods, but traverse computed keys).

But, it needs to be recursive. This is the best way I've found to reference the complete visitor from the sub-visitor.

/cc @loganfsmyth

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -122,7 +125,7 @@ export default declare((api, options) => {
           // Make sure computed property names are only evaluated once (upon class definition)
           // and in the right order in combination with static properties
           if (!computedPath.get("key").isConstantExpression()) {
-            computedPath.traverse(ClassFieldDefinitionEvaluationTDZVisitor, {
+            computedPath.traverse(classFieldDefinitionEvaluationTDZVisitor, {
               classRef: path.scope.getBinding(ref.name),
               file: this.file,
               shouldSkip: computedPath.get("value"),
```
nicolo-ribaudo: You can probably remove this.

packages/babel-helper-replace-supers/src/index.js
jridgewell: Curious. It's been in there since v5, but doesn't seem to be used anymore...

jridgewell: Done.

packages/babel-plugin-proposal-class-properties/src/index.js
jridgewell: Done.

nicolo-ribaudo:
Travis is failing because of `assert` instead of `expect`

PR #7678, opened 2018-04-06
Default derived constructor uses Rest and Spread argument

### LINKED ISSUE #5147: default derived constructor doesn't spread/rest the args in super call
### Input Code

```js
Array.prototype[Symbol.iterator] = function() {
  throw new Error();
}
class c {} ;
class b extends c {}
new b();
```

presets -latest

### Expected Behavior
```
→ cat test2.js
Array.prototype[Symbol.iterator] = function() {
  throw new Error();
}
class c {} ;
class b extends c {}
var y = new b();

→ ./out.gn/x64.debug/d8 test2.js
test2.js:2: Error
  throw new Error();
  ^
Error
    at Array.(anonymous function) (test2.js:2:9)
    at b (test2.js:5:1)
    at test2.js:6:9
```

### Current Behavior
No error

### Your Environment

| software         | version
| ---------------- | -------
| Babel            |  
| node             | 
| npm              | 
| Operating System |

jridgewell:
Fixes #5147.
Depends on #7675.

### DIFF
--- a/packages/babel-plugin-transform-classes/src/transformClass.js
+++ b/packages/babel-plugin-transform-classes/src/transformClass.js
@@ -146,20 +146,17 @@ export default function transformClass(
    * Creates a class constructor or bail out if there is none
    */
   function maybeCreateConstructor() {
-    let hasConstructor = false;
     const paths = classState.path.get("body.body");
     for (const path of paths) {
-      hasConstructor = path.equals("kind", "constructor");
-      if (hasConstructor) break;
+      if (path.equals("kind", "constructor")) return;
     }
-    if (hasConstructor) return;
 
     let params, body;
 
     if (classState.isDerived) {
       const constructor = template.expression.ast`
-        (function () {
-          super(...arguments);
+        (function (...args) {
+          super(...args);
         })
       `;
       params = constructor.params;
--- a/packages/babel-plugin-transform-classes/src/transformClass.js
+++ b/packages/babel-plugin-transform-classes/src/transformClass.js
@@ -349,20 +349,9 @@ export default function transformClass(
       call = t.logicalExpression("||", bareSuperNode, t.thisExpression());
     } else {
       bareSuperNode = optimiseCall(
-        t.logicalExpression(
-          "||",
-          t.memberExpression(
-            t.cloneNode(classState.classRef),
-            t.identifier("__proto__"),
-          ),
-          t.callExpression(
-            t.memberExpression(
-              t.identifier("Object"),
-              t.identifier("getPrototypeOf"),
-            ),
-            [t.cloneNode(classState.classRef)],
-          ),
-        ),
+        t.callExpression(classState.file.addHelper("getPrototypeOf"), [
+          t.cloneNode(classState.classRef),
+        ]),
         t.thisExpression(),
         bareSuperNode.arguments,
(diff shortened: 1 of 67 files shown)


packages/babel-plugin-transform-classes/src/transformClass.js
```diff
@@ -146,20 +146,17 @@ export default function transformClass(
...
     }
-    if (hasConstructor) return;
 
     let params, body;
 
     if (classState.isDerived) {
       const constructor = template.expression.ast`
-        (function () {
-          super(...arguments);
+        (function (...args) {
```
loganfsmyth: An issue description would be helpful. What's the motivation for this change?

jridgewell: Whoops, fixes #5147.

packages/babel-plugin-transform-classes/src/transformClass.js
```diff
@@ -349,20 +346,9 @@ export default function transformClass(
       call = t.logicalExpression("||", bareSuperNode, t.thisExpression());
     } else {
       bareSuperNode = optimiseCall(
```
Kovensky: Looks like the `helper-optimise-call-expression` needs to be updated to support the `super(...args)` case; right now the `optimizeCall` is essentially a noop (or, well, it writes a `.call`, but not an "optimized" one)

packages/babel-plugin-transform-classes/test/fixtures/loose/super-class-id-member-expression/output.js
```diff
@@ -15,8 +15,8 @@ var BaseController2 =
 function (_Chaplin$Controller$A) {
   babelHelpers.inheritsLoose(BaseController2, _Chaplin$Controller$A);
 
-  function BaseController2() {
-    return _Chaplin$Controller$A.apply(this, arguments) || this;
+  function BaseController2(...args) {
+    return _Chaplin$Controller$A.call.apply(_Chaplin$Controller$A, [this].concat(args)) || this;
```
loganfsmyth: I wonder how this might affect class instantiation performance? My main concern with this change is that it increases output complexity for an extremely edge-casey issue that doesn't seem like anyone would rely on.

jridgewell: I'm working on a total refactor here, it's just really difficult because the code's a bit low quality at the moment. This will be optimized into a straight `#apply` when I'm done.

Kovensky: See https://github.com/babel/babel/pull/7678#discussion_r179643407; `helper-optimise-call-expression` just doesn't deal with this corner case and is inserting a `.call` instead of an `.apply`

packages/babel-plugin-transform-classes/src/transformClass.js
nicolo-ribaudo: I think that in loose mode we should keep using `arguments`, since it produces a smaller (and probably faster) output.

nicolo-ribaudo:
Closing since tc39/ecma262#2216 got consensus during the last TC39 meeting.

PR #7680, opened 2018-04-06
Cleanup object super traversal

jridgewell:
Also fixes a few bugs with super traversal in general (needs to run on computed keys, but not initializers, unless it's static).


jridgewell:
Need to split this into two.

PR #7687, opened 2018-04-07
Get set helpers

jridgewell:
Fixes https://github.com/babel/babel/issues/5769.

Wow, this had all kinds of bug. For instance, `super.set = 1` attempted to do `descriptor.value = 1` thinking did anything. Whoops.

I kinda went overboard matching the spec. We can probably prune it down a bit.


packages/babel-helpers/src/helpers.js
```diff
@@ -633,28 +603,93 @@ helpers.possibleConstructorReturn = () => template.program.ast`
   }
 `;
 
+helpers.superPropBase = () => template.program.ast`
+  import getPrototypeOf from "getPrototypeOf";
+
+  export default function _superPropReceiver(object, property) {
```
loganfsmyth: Looks like this needs a rename to match the helper name.

zloirock:
I think that we should add use `Reflect.{get, set}` where it's possible.

packages/babel-helpers/src/helpers.js
jridgewell: Done.

packages/babel-helpers/src/helpers.js
```diff
@@ -633,28 +603,91 @@ helpers.possibleConstructorReturn = () => template.program.ast`
...
+    }
 
-      if (setter !== undefined) {
-        setter.call(receiver, value);
+    // Without a super that defines the property, spec boils down to "define on
+    // receiver" for some reason.
+    desc = Object.getOwnPropertyDescriptor(receiver, property);
+    if (desc) {
+      if (desc.set) {
+        if (isStrict()) {
```
loganfsmyth: Hmm, I wonder if it'd be better to actually pass this as a true/false value to the helper? Since it is the caller's location, not the actual helper that we want to know the strictness of. Especially with https://github.com/babel/babel/pull/7411 and if we do `sourceType:script` I'd expect a lot more code to be strict only inside class bodies for instance, so this helper would return the wrong value.

jridgewell: That'd be great. How do I get that value? The same traversal?

zloirock:
@jridgewell it can be polyfilled without any problems. And it's definitely better if native methods are available.

jridgewell:
It'll use the `Reflect` helpers now.

packages/babel-helpers/src/helpers.js
jridgewell: Done.

packages/babel-helper-replace-supers/src/index.js
```diff
@@ -225,14 +227,33 @@ export default class ReplaceSupers {
 
   specHandleAssignmentExpression(ref, path, node) {
     if (node.operator === "=") {
-      // super.name = "val"; -> _set(Object.getPrototypeOf(objectRef.prototype), "name", this);
+      const strictParent = path.findParent(path => {
```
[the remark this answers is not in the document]
jridgewell: Good catch.

packages/babel-helpers/src/helpers.js
```diff
@@ -470,7 +440,7 @@ helpers.construct = () => template.program.ast`
   import setPrototypeOf from "setPrototypeOf";
 
   export default function _construct(Parent, args, Class) {
-    if (typeof Reflect === "object" && Reflect.construct) {
+    if (typeof Reflect !== "undefined" && Reflect.construct) {
```
xtuc: Isn't "object" better? What if Reflect is a number? that will probably never happen tho.

jridgewell: `typeof X === "object"` and `=== "symbol"` requires a typeof helper to compensate for polyfilled `Symbol` instances.

PR #7691, opened 2018-04-08
Fix super nested class bugs

### LINKED ISSUE #7682: Super in object method key has wrong HomeObject

### Input Code

```js
var get_super_foo, get_super_bar;

var outer = {
  method() {
    var inner = {
      [(() => { get_super_foo = () => super.foo; return "fn" })()]() {},
      [(() => { get_super_bar = () => super.bar; return "fn" })()]: 0,
      __proto__: { foo: "inner_foo", bar: "inner_bar" }
    };
    return inner;
  },
  __proto__: { foo: "outer_foo", bar: "outer_bar" }
}

outer.method();

console.log(get_super_foo() + " " + get_super_bar());
```

es2015 preset

### Expected Behavior
`outer_foo outer_bar`

### Current Behavior
`inner_foo outer_bar`

### Your Environment

| ---------------- | -------
| Babel            |  master (7.0.0-beta.44)
| Babylon          | 
| node             | 
| npm              | 
| Operating System |

jridgewell:
The computed keys, and static field initializers (for classes) are evaluated in the outer scope's home environment, not the newly created environment.

Fixes #7682.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -16,10 +16,19 @@
   "cacheDir": ".changelog",
   "commands": {
     "publish": {
-      "ignore": ["*.md", "test/**", "codemods/**"]
+      "ignore": [
+        "*.md",
+        "test/**",
+        "codemods/**"
+      ]
     }
   },
-  "packages": ["packages/*", "codemods/*"],
+  "packages": [
+    "packages/*",
+    "codemods/*"
+  ],
   "npmClient": "yarn",
-  "npmClientArgs": ["--no-lockfile"]
+  "npmClientArgs": [
+    "--no-lockfile"
+  ]
 }
--- a/lerna.json
+++ b/lerna.json
@@ -16,19 +16,10 @@
   "cacheDir": ".changelog",
   "commands": {
     "publish": {
-      "ignore": [
-        "*.md",
-        "test/**",
-        "codemods/**"
-      ]
+      "ignore": ["*.md", "test/**", "codemods/**"]
     }
   },
-  "packages": [
-    "packages/*",
-    "codemods/*"
-  ],
+  "packages": ["packages/*", "codemods/*"],
   "npmClient": "yarn",
-  "npmClientArgs": [
-    "--no-lockfile"
-  ]
+  "npmClientArgs": ["--no-lockfile"]
 }
--- a/packages/babel-helper-replace-supers/src/index.js
+++ b/packages/babel-helper-replace-supers/src/index.js
@@ -5,37 +5,24 @@ import * as t from "@babel/types";
 // ✌️
 const HARDCORE_THIS_REF = new WeakSet();
 
-function isIllegalBareSuper(node, parent) {
-  if (!t.isSuper(node)) return false;
-  if (t.isMemberExpression(parent, { computed: false })) return false;
-  if (t.isCallExpression(parent, { callee: node })) return false;
-  return true;
-}
-
-function isMemberExpressionSuper(node) {
-  return t.isMemberExpression(node) && t.isSuper(node.object);
-}
-
 /**
  * Creates an expression which result is the proto of objectRef.
- * Uses CLASS.__proto__ first for InternetExplorer <= 10 support
  *
  * @example <caption>isStatic === true</caption>
  *
- *   CLASS.__proto__ || Object.getPrototypeOf(CLASS)
+ *   helpers.getPrototypeOf(CLASS)
  *
  * @example <caption>isStatic === false</caption>
  *
- *   CLASS.prototype.__proto__ || Object.getPrototypeOf(CLASS.prototype)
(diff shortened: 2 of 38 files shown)


jridgewell:
/to @nicolo-ribaudo, since you filled the bug. 😉

lerna.json
```diff
@@ -16,19 +16,10 @@
   "cacheDir": ".changelog",
```
nicolo-ribaudo: (Not related to this PR)

I think that we should exclude this file from prettier, since it clashes with the how lerna generates it (I suppose so because [it has been prettified and reverted two times](https://github.com/jridgewell/babel/commits/80e502905bc80de0074d823d711b1d49bde43b65/lerna.json))

packages/babel-helper-replace-supers/src/index.js
```diff
@@ -164,7 +152,7 @@ export default class ReplaceSupers {
    *
    * @example
    *
-   *   _get(CLASS.prototype.__proto__ || Object.getPrototypeOf(CLASS.prototype), "METHOD", this)
+   *   _get(Object.getPrototypeOf(CLASS.prototype), "METHOD", this)
```
nicolo-ribaudo: This should be `_getPrototypeOf`

packages/babel-helper-replace-supers/src/index.js
```diff
@@ -188,139 +176,143 @@ export default class ReplaceSupers {
...
+      t.binaryExpression(operator.slice(0, -1), t.cloneNode(ref), node.right),
+      computed,
+    );
+    return [
+      t.assignmentExpression(
+        "=",
+        t.cloneNode(ref),
+        this.getSuperProperty(property, computed),
+      ),
+      setter,
```
nicolo-ribaudo: Why do we need to store the `_get` result?

packages/babel-helper-replace-supers/src/index.js
```diff
@@ -188,139 +176,143 @@ export default class ReplaceSupers {
...
+      let nodes;
+
+      // TODO this needs some more cleanup.
+      if (prefix) {
         // ++super.foo;
         // to
-        // super.foo += 1;
-        return this.specHandleAssignmentExpression(null, path, binary);
+        // _ref = super.foo, super.foo = _ref + 1
+        nodes = this.specHandleAssignmentExpression(grandParentPath);
```
nicolo-ribaudo: I think that we should convert `super.foo` to a number, like we do for `super.foo++`?

packages/babel-helper-replace-supers/src/index.js
jridgewell: Illustrative, since `Object.getPrototypeOf` is a known algorithm.

packages/babel-helper-replace-supers/src/index.js
jridgewell: We don't yet have a set algorithm that will do the get, too. So, we have to do the get, then pass it to the set.

And, given the codesize for the get/set algorithms, I didn't want in include yet another for this case.

packages/babel-helper-replace-supers/src/index.js
jridgewell: Yup, good catch.

lerna.json
jridgewell: Reverted.

packages/babel-plugin-transform-classes/src/transformClass.js
```diff
@@ -397,7 +397,7 @@ export default function transformClass(
       return ref;
     };
 
-    for (const bareSuper of classState.bareSupers) {
+    for (const bareSuper of new Set(classState.bareSupers)) {
```
nicolo-ribaudo: Is this `new Set` to avoid duplicated paths? When can it happen? I think that you should add a comment.

jridgewell:
```
super[super().name]();
```

We find the original `super`, and transform it. But the MemberExpression `super[NEXT]` is still queued, so we find `super()` and add it. Then the transformed `_get(PROTO, super().name)` is traversed, finding `super()` again.

jridgewell: Maybe the set would be better inside ReplaceSupers?

nicolo-ribaudo: Yeah, state.bareSupers should be a set since the beginning if possible.

jridgewell: Done.

nicolo-ribaudo: I think that you forgot to commit the file? In the last commit only some tests and comments are changed.

jridgewell: Whoops.

PR #7706, opened 2018-04-10
Fix literal type annotation argument number

### LINKED ISSUE #7697: stringLiteralTypeAnnotation node creation function accepts no arguments but should accept 1

Bug!

### Input Code

```js
import * as T from '@babel/types';

T.stringLiteralTypeAnnotation('literal');
```

### Expected Behavior

This should create a new node of type string literal. String literals are literal types and need a value. In code this would look like this:

```js
const literal: 'literal' = 'literal';

// This is different from StringTypeAnnotation which needs no args:
const literal: string = 'literal';
```

This should create a node of this shape:

```json
{
  "type": "StringLiteralTypeAnnotation",
  "start": 15,
  "end": 24,
  "loc": null,
  "extra": {
    "rawValue": "literal",
    "raw": "'literal'"
  },
  "value": "literal"
}
```

### Current Behavior
Exception is thrown:
```
StringLiteralTypeAnnotation: Too many arguments passed. Received 1 but can receive no more than 0
```

### Possible Solution

I have no idea what I am doing but when I add the visitor `"value"` to the definition in `packages/babel-types/src/definitions/flow.js` it works as expected. Maybe I can prepare a PR but I really don't know if this is the actual fix. There are some more types that have the same mismatch of fields and visitors.

```diff
 defineType("StringLiteralTypeAnnotation", {
+  visitor: ["value"],
   aliases: ["Flow", "FlowType"],
   fields: {
     value: validate(assertValueType("string")),
```

### Context

I am building a type generator to create Flow type definitions for GraphQL queries. I use `@babel/types` to build ASTs and then `@babel/generator` to create the code. This is working so well so far, awesome work on babel 7. I absolutely love it ❤️ 

### Your Environment

| ---------------- | -------
| Babel            | @babel/types@7.0.0-beta.44

hendrikniemann:
| Fixed Issues?            | Fixes #7697
| Patch: Bug Fix?          | 🐛 
| Major: Breaking Change?  | Maybe (see below)
| Minor: New Feature?      |
| Tests Added + Pass?      | No tests added

The flow literal type creation functions in the `@babel/types` package did not take any arguments. As discussed in the linked issue they will need to have the single required  `value` argument listed in the definitions file. I ran the build and also updated the README.
Technically this is a breaking change since the functions now require an argument but they did not work before so I don't think anybody was able to use them without arguments.

There are more functions that have a mismatch of arguments in the fields definition and visitors/builders. I can maybe work on that in the future. Is there someone particularly responsible/invested in the _types_ package?

### DIFF
--- a/packages/babel-types/README.md
+++ b/packages/babel-types/README.md
@@ -183,14 +183,14 @@ Aliases: `Expression`, `Pureish`, `Literal`, `Immutable`
 
 ### booleanLiteralTypeAnnotation
 ```javascript
-t.booleanLiteralTypeAnnotation()
+t.booleanLiteralTypeAnnotation(value)
 ```
 
 See also `t.isBooleanLiteralTypeAnnotation(node, opts)` and `t.assertBooleanLiteralTypeAnnotation(node, opts)`.
 
 Aliases: `Flow`, `FlowType`
 
- - `value`: `boolean` (default: `null`)
+ - `value`: `boolean` (required)
 
 ---
 
@@ -1408,14 +1408,14 @@ Aliases: `Flow`, `FlowType`
 
 ### numberLiteralTypeAnnotation
 ```javascript
-t.numberLiteralTypeAnnotation()
+t.numberLiteralTypeAnnotation(value)
 ```
 
 See also `t.isNumberLiteralTypeAnnotation(node, opts)` and `t.assertNumberLiteralTypeAnnotation(node, opts)`.
 
 Aliases: `Flow`, `FlowType`
 
- - `value`: `number` (default: `null`)
+ - `value`: `number` (required)
 
 ---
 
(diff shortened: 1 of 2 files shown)


hzoo:
Hey Hendrik! Yeah I don't think there's anyone in particular invested in in `@babel/types` so if you'd like to take that on that would be great! It's not really something that is thought about much even though obviously used everywhere. Usually we add new types for proposals or things like typescript and then maybe it's a perf fix like in https://github.com/babel/babel/pull/7685.

I know we moved to generating out the types with code instead of dynamically for perf earlier too

hzoo:
Nice first pr! Yeah it's probably that it has been rare for someone to use the type annotation stuff until recently.

Maybe should think about a way to test the validity of these definitions

PR #7708, opened 2018-04-10
ESLint: Ignore lerna.json and .git

jridgewell:


.eslintignore
```diff
@@ -1,6 +1,8 @@
 /lib
 /build
+/.git
```
xtuc: that should be by default tho

PR #7712, opened 2018-04-10
Add Path#isInStrictMode

jridgewell:
Adds `path#isInStrictMode` to determine if a parent set strict mode.

### DIFF
--- a/packages/babel-traverse/src/path/introspection.js
+++ b/packages/babel-traverse/src/path/introspection.js
@@ -450,11 +450,15 @@ export function isInStrictMode() {
 
     if (!path.isProgram() && !path.isFunction()) return false;
 
+    if (
+      path.isArrowFunctionExpression() &&
+      !path.get("body").isBlockStatement()
+    ) {
+      return false;
+    }
+
     let { node } = path;
     if (path.isFunction()) node = node.body;
-
-    const { directives } = node;
-    if (!directives) return false;
 
     for (const directive of node.directives) {
       if (directive.value.value === "use strict") {
--- a/packages/babel-traverse/src/path/introspection.js
+++ b/packages/babel-traverse/src/path/introspection.js
@@ -444,22 +444,17 @@ export function isInStrictMode() {
   const start = this.isProgram() ? this : this.parentPath;
 
   const strictParent = start.find(path => {
-    if (path.isProgram({ sourceType: "module" })) {
-      return true;
-    }
+    if (path.isProgram({ sourceType: "module" })) return true;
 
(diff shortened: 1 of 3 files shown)


packages/babel-traverse/src/path/introspection.js
```diff
@@ -439,3 +439,29 @@ export function isConstantExpression() {
...
+
+  const strictParent = start.find(path => {
+    if (path.isProgram({ sourceType: "module" })) return true;
+
+    if (path.isClass()) return true;
+
+    if (!path.isProgram() && !path.isFunction()) return false;
+
+    let { node } = path;
+    if (path.isFunction()) node = node.body;
```
loganfsmyth: Would it be worth handling cases where the `path` is the function name, or a function parameter? Like technically it could be

```
function NODE(){"use strict";}
// OR
function fn(NODE){"use strict";}
```
too. It isn't valid to having function initializers or patterns at least, so we wouldn't have to check too many cases.

jridgewell: Already handled.

> `function NODE(){"use strict";}`

https://github.com/babel/babel/pull/7712/files#diff-e0bc96e619aa8318bff982e8def5245fR82

> `function fn(NODE){"use strict";}`

https://github.com/babel/babel/pull/7712/files#diff-e0bc96e619aa8318bff982e8def5245fR92

jridgewell: Note: this is directly related to https://github.com/rwaldron/tc39-notes/blob/master/es9/2018-03/mar-22.md#8ic-expand-text-included-in-function-code which we just talked about in last month's meeting. Consensus reached was the `id` and `params` are strict if the function body is strict.

loganfsmyth: Oh of course, that's what I get for reviewing while sleepy :P

packages/babel-traverse/src/path/introspection.js
```diff
@@ -439,3 +439,29 @@ export function isConstantExpression() {
...
+
+    const { directives } = node;
+    if (!directives) return false;
```
loganfsmyth: I think in theory this is supposed to always exist unless `node` isn't `BlockStatement` (because it is an arrow body expression)? Might be clearer to do

```
if (!path.isProgram() && !path.isFunction()) return false;

if (path.isArrowFunction() && !path.get("body").isBlockStatement()) return false;
```
then remove this check?

jridgewell: Yup, that was it. Done.

PR #7713, opened 2018-04-11
Add slack links to CONTRIBUTING.md

nicolo-ribaudo:


hzoo:
Nice! I think it's weird if it's 2 links to the same place though right?

nicolo-ribaudo:
Whoops I copy-pasted wrong. I will update it later.

PR #7717, opened 2018-04-12
Provide better error message for invalid default export declaration

dczombera:
| Fixed Issues?            | Fixes ##6708 
| Documentation PR         | 
| Any Dependency Changes?  |

I was not sure if an error message like the one suggested in the issue is consistent with the other error messages that babylon provides (is consistency even an issue when it comes to error messages?).
Apart from that, it's my first contribution to the project. I welcome any type of feedback. Thanks! :)

[REPL from build](https://babeljs.io/repl/build/7567/#?babili=false&browsers=&build=&builtIns=false&code_lz=KYDwDg9gTgLgBAE2AMwIYFcA28DGEB2AzvALICeAwhALaT7D7wC8cAFAJRxMB8cA3gCg4cKMBjoo-OAAMAPAgCWANzg5MqQoSYAiZBAjbuACQCSsgPSKl3aQG4BAX1tA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=true&presets=&prettier=true&targets=&version=7.0.0-beta.44%2Bpr.7717&envVersion=)

### DIFF
--- a/packages/babylon/src/parser/statement.js
+++ b/packages/babylon/src/parser/statement.js
@@ -1408,6 +1408,15 @@ export default class StatementParser extends ExpressionParser {
     } else if (this.match(tt.at)) {
       this.parseDecorators(false);
       return this.parseClass(expr, true, true);
+    } else if (
+      this.match(tt._let) ||
+      this.match(tt._const) ||
+      this.match(tt._var)
+    ) {
+      this.raise(
+        this.state.start,
+        "Only expressions, functions or classes are allowed as the `default` export.",
+      );
     } else {
       const res = this.parseMaybeAssign();
       this.semicolon();
--- a/packages/babylon/test/fixtures/es2015/modules/export-default-class-declaration/input.js
+++ b/packages/babylon/test/fixtures/es2015/modules/export-default-class-declaration/input.js
@@ -1,0 +1,1 @@
+export default class Foo {}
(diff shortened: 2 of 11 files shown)


packages/babylon/src/parser/statement.js
```diff
@@ -1408,6 +1408,16 @@ export default class StatementParser extends ExpressionParser {
...
       this.parseDecorators(false);
       return this.parseClass(expr, true, true);
+    } else if (
+      this.match(tt._let) ||
+      this.match(tt._const) ||
+      this.match(tt._var)
+    ) {
+      this.raise(
+        this.state.start,
+        "The `default` export can only take an expression;" +
```
nicolo-ribaudo: It can also take function and class declarations.

dczombera: So would you suggest to just add both to the error message?

existentialism: Maybe just:

> Only expressions, functions or class declarations are allowed as the default export.

(Let's add tests for those too).

dczombera:
Alright, I changed the error message as discussed and added tests for class declarations/expressions. :)

hzoo:
Nice work, a great first PR!

existentialism:
@dczombera thanks!

nicolo-ribaudo:
An easy way to think about it is that named exports must have a name, while default exports not.

The real reason is that named exports export _bindings_, or variables (infact, if you modify an already exported variable the modules which import it will see the changes), while default exports export _values_.

Kovensky:
But there's an asterisk... there's always an asterisk 😇 

If your default export is a `function` or `class` declaration (no parens wrapping them), then those will be mutable bindings, and reassigning the `function` or `class` name will update the default export.

It's also possible to make anything you want the default export by using `export { foo as default }`.

PR #7718, opened 2018-04-12
[babylon] Fix arguments & super in class properties

nicolo-ribaudo:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

> 6d6d9cb Disallow super in functions in class properties
> d3ca51d Disallow arguments in class properties

### DIFF
--- a/packages/babel-core/test/fixtures/transformation/misc/class-property-initializer-blocks-shadow/input.js
+++ b/packages/babel-core/test/fixtures/transformation/misc/class-property-initializer-blocks-shadow/input.js
@@ -1,10 +1,6 @@
 class A {
   prop1 = () => this;
   static prop2 = () => this;
-  prop3 = () => arguments;
-  static prop4 = () => arguments;
   prop5 = this;
   static prop6 = this;
-  prop7 = arguments;
-  static prop8 = arguments;
 }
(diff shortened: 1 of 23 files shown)


jridgewell:
> Disallow super in functions in class properties

The only thing disallowed is `super()`. Can we get a test for that?

nicolo-ribaudo:
Isn't `foo = function () { super.x }` disallowed?

jridgewell:
That's disallow because `super` (both call and member expression) is only allowed inside methods (class or object):

```js
const foo = function () {
  super.x // syntax error
}

function foo() {
  super.x // syntax error
}
```

But that's not a class properties restriction. The only class props ones are `arguments` (unless inside a function) and `super()` (unless in a nested class)

```js
class Outer extends Object {
  class Foo {
    x = arguments // syntax error
    x = () => arguments // syntax error
    x = super() // syntax error
    x = () => super() // syntax error
  }
}

nicolo-ribaudo:
> But that's not a class properties restriction.

Yeah, that commit fixes leakage of the `inClassProperty` flag to nested functions.

I will check the `foo = super()` case in a few hours.

PR #7542, opened 2018-03-10
Decorators

nicolo-ribaudo:
EDIT by @hzoo: REPL link https://babeljs.io/repl/build/8074/ - need to turn on Stage 2

| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Yes
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes
| Documentation PR         | 
| Any Dependency Changes?  |
| License                  | MIT

I'm working on https://github.com/babel/babel/pull/6107. Since that PR has become quite old (and there are different things that needs to be done) I'm not rebasing it, but I will just copy-paste the needed code.
Then I will set both me and @peey as the PR authors (I have no idea of how to do it, but I'm sure that it is possible :rofl:).

---

TODO:
- [ ] Private elements (I will work on this after #7555 is merged, I can do it in another PR).
- [x] Tests

---

NOTE: The `legacy-transformer.js` file contains the code that was peviously in `index.js`, I didn't change it.

---

### Breaking change!

I added a new option to keep both the new and legacy decorators in the same plugin. By default, the new proposal is used but it can be disabled with the `legacy: true` option (accepted by `-syntax-` and `-proposal-`). In the `stage-0,1,2` presets this option is called `decoratorsLegacy`.

PR #7719, opened 2018-04-12
Update decorators parsing

nicolo-ribaudo:
| Fixed Issues?            |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

This PR is part of #7542; I'm splitting it in smaller PRs.

> This commit introduces three changes:
> 1) Class properties can be decorated
> 2) Decorators can contain arbitrary expressions, using `@(...)`
> 3) The Decorator node type has a new property, "arguments". This
      makes it possible do distinguish `@dec()` and `@(dec())`, which have
      different behaviors because `@(dec())` is equivalent to `@(dec())()`.


packages/babel-types/src/definitions/experimental.js
```diff
@@ -135,11 +135,18 @@ defineType("Import", {
 });
 
 defineType("Decorator", {
-  visitor: ["expression"],
+  visitor: ["expression", "arguments"],
```
jridgewell: `callee`?

jridgewell:
The code changes themselves look fine to me.

A few edge cases:

```js
@foo().bar
@foo()()()
class Foo {}
```

I don't think either of those decorators syntaxes are legal according to the [syntax](https://tc39.github.io/proposal-decorators/#sec-new-syntax). `@foo()` certainly is, but there's no way to compose a `DecoratorMemberExpression` with a `DecoratorCallExpression` as the base.

And there's no way to continue compose a a `DecoratorCallExpression` into a another `DecoratorCallExpression`...

benjamn:
The parsing changes here make sense, but the `@babel/plugin-proposal-decorators` transform doesn't seem to respect the new `.arguments` property [here](https://github.com/babel/babel/blob/dccaec7691dd3a5217e8464ed3acd91370668f4a/packages/babel-plugin-proposal-decorators/src/transformer-legacy.js#L72) or [here](https://github.com/babel/babel/blob/dccaec7691dd3a5217e8464ed3acd91370668f4a/packages/babel-plugin-proposal-decorators/src/transformer-legacy.js#L49), which is leading to bugs such as https://github.com/meteor/meteor/issues/9895. I will try to submit a PR to fix this, but I wanted to flag it as a problem here, in case I'm missing something, or anyone else has thoughts.

PR #7720, opened 2018-04-12
Migrate more packages' tests to use jest expect assertions

devenbansod:
| Major: Breaking Change?  | Yes, removed `chai.assert` from the context in which test fixtures are run
| Minor: New Feature?      | NA

* Migrate -optional-catch-binding, -block-scoping to use jest expect assertions
* Migrate -transform-classes tests to use jest expect assertions
* Migrate -transform-object-super tests to use jest expect assertions
* Migrate -computed-properties and -jscript tests to use jest expect assertions

### DIFF
--- a/packages/babel-helper-transform-fixture-test-runner/src/index.js
+++ b/packages/babel-helper-transform-fixture-test-runner/src/index.js
@@ -28,17 +28,6 @@ const testContext = vm.createContext({
 });
 testContext.global = testContext;
 
-// Add chai's assert to the global context
-// It has to be required inside the testContext as otherwise some assertions do not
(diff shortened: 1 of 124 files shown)


nicolo-ribaudo:
LGTM :+1: 

It just needs to be rebased after https://github.com/babel/babel/pull/7721

devenbansod:
@nicolo-ribaudo thanks for the previous review.

Have removed the injection of `assert` into the global context in latest commits (also made me realize some missing instances to be migrated, when those tests failed as I removed `assert` from the context).

jridgewell:
👍

PR #7722, opened 2018-04-13
Remove tagged template literal global caching

### LINKED ISSUE #7350: Spec change: Tagged template literals are no longer cached globally
Spec PR: https://github.com/tc39/ecma262/pull/890

Babel has never been able to implement a true global cache anyway, but now we can remove the file-level caching. e.g.

Currently say we have [this example](https://babeljs.io/repl/#?babili=false&browsers=chrome%20%3E%2055%20%20&build=&builtIns=false&code_lz=MYewdgzgLgBFCGBzGBeGA3VA-DBuAUPgBQBmArmMFAJbhECUA3vjHEgAYQgC2ApnL24AHADbwovdgVYJEnHvwnCxEqfgC-9BriA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=false&presets=es2015%2Cenv&prettier=false&targets=&version=6.26.0&envVersion=1.6.2)
```
const tag = v => v;

(function(){
  tag`some template`;
  tag`some template`;
})();
```
we generate
```
var _templateObject = _taggedTemplateLiteral(["some template"], ["some template"]);

function _taggedTemplateLiteral(strings, raw) { return Object.freeze(Object.defineProperties(strings, { raw: { value: Object.freeze(raw) } })); }

var tag = function tag(v) {
  return v;
};

(function () {
  tag(_templateObject);
  tag(_templateObject);
})();
```
Note how the `_taggedTemplateLiteral` call is hoisted to the top of the function, and _both_ tagged examples reference the same `_templateObject` object.

With the current spec changes we'd at least want to change this logic to be
```
var _templateObject = _taggedTemplateLiteral(["some template"], ["some template"]);
var _templateObject2 = _taggedTemplateLiteral(["some template"], ["some template"]);

function _taggedTemplateLiteral(strings, raw) { return Object.freeze(Object.defineProperties(strings, { raw: { value: Object.freeze(raw) } })); }

var tag = function tag(v) {
  return v;
};

(function () {
  tag(_templateObject);
  tag(_templateObject2);
})();
```
so that each template literal has its own `_templateObject`.

But now that that things aren't shared anymore, there's really no reason to hoist the initialization at all, so I'd like to see us update this to
```
var _templateObject, _templateObject2;

function _taggedTemplateLiteral(strings, raw) { return Object.freeze(Object.defineProperties(strings, { raw: { value: Object.freeze(raw) } })); }

var tag = function tag(v) {
  return v;
};

(function () {
  tag(_templateObject || (_templateObject = _taggedTemplateLiteral(["some template"], ["some template"])));
  tag(_templateObject2 || (_templateObject2 = _taggedTemplateLiteral(["some template"], ["some template"])));
})();
```
so we only initialize the template literal object when it is needed.

All of this should be doable in https://github.com/babel/babel/blob/dd0337cc85c341b36b0657e05964532a650fd088/packages/babel-plugin-transform-template-literals/src/index.js#L73

Marking as good first issues because it is relatively self-contained, but it will likely require some time to understand what needs to change in the AST.

jridgewell:
Fixes #7350.

### DIFF
--- a/packages/babel-plugin-transform-template-literals/src/index.js
+++ b/packages/babel-plugin-transform-template-literals/src/index.js
@@ -45,9 +45,6 @@ export default declare((api, options) => {
   }
 
   return {
-    pre() {
-      this.templates = new Map();
-    },
     visitor: {
       TaggedTemplateExpression(path) {
         const { node } = path;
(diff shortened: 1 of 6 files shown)


xtuc:
I'm wondering why that didn't break other tests? Was the global caching untested?

packages/babel-plugin-transform-template-literals/test/fixtures/default/cache-revision/input.js
```diff
@@ -0,0 +1,11 @@
...
+
+function foo() {
+  return tag`some template`;
+}
+function bar() {
+  return tag`some template`;
+}
+assert.equal(foo(), foo());
+assert.equal(bar(), bar());
+assert.notEqual(foo(), bar());
```
xtuc: Could you please test the output? A bug could cause `assert.equal(0, 0);` for ex.

packages/babel-plugin-transform-template-literals/src/index.js
```diff
@@ -75,41 +72,29 @@ export default declare((api, options) => {
...
-          this.templates.set(name, templateObject);
-
-          const helperId = this.addHelper(helperName);
-          const callExpressionInput = [];
-          callExpressionInput.push(t.arrayExpression(strings));
-
-          if (!isStringsRawEqual) {
-            callExpressionInput.push(t.arrayExpression(raws));
-          }
+        const programPath = path.find(p => p.isProgram());
```
nicolo-ribaudo: You can use `path.scope.getProgramParent()`

jridgewell: Done.

packages/babel-plugin-transform-template-literals/test/fixtures/default/cache-revision/input.js
jridgewell: Done.

PR #7734, opened 2018-04-15
Decorators legacy option

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

This PR contains the only breaking change introduced by #7542 which directly affects users.
I added a new option to keep both the new and legacy decorators in the same plugin. By default, the new proposal will be used but it can be disabled with the `legacy: true` option (accepted by `-syntax-` and `-proposal-`). In the `stage-0,1,2` presets this option is called `decoratorsLegacy`.
This PR doesn't contain the new decorators transformer, so if `"legacy": true` isn't passed it throws an error.

The `transformer-legacy.js` file's content was copied from the `index.js` file (actually from #7719's one, since it changed a bit), so it doesn't need to be reviewed. I suggest reviewing the new `index.js` not from the diff but from https://github.com/nicolo-ribaudo/babel/blob/6940427d4048046fb2bee1a9f5b531bd16c5b194/packages/babel-plugin-proposal-decorators/src/index.js, since I rewrote almost all of it.

I will open a PR to `babel-upgrade` to reflect this change.

Depends on #7719.


packages/babel-plugin-proposal-decorators/README.md
```diff
@@ -62,11 +62,33 @@ Add the following line to your .babelrc file:
...
+});
+```
+
+## Options
+
+### `legacy`
+
+`boolean`, defaults to `false`.
+
+Use the legacy (stage 1) decorators syntax and behavior.
```
hzoo: can do this later or skip ci commit - can mention this replaces the legacy plugin and link to it? (We should probably make a similar change in that readme about using this plugin if you are using v7)

nicolo-ribaudo:
It should be fixed by `vue-cli`. This option is needed to provide a smooth transition from the old decorators proposal to the new one.

dmk255:
I'm a little confused by this change, is there even a way to use this transform without setting `legacy: true`?

I've attempted to follow the documentation but am still seeing the same error. Reading the code, I don't see how index.js will accept anything other than `legacy: true`

nicolo-ribaudo:
Modern decorators are not implemented yet. We added that option so that we can implement the new decorators proposal and enable it by default without breaking changes.

PR #7745, opened 2018-04-17
Improve asyncIterator error

jquense:
When an object is has neither asyncIterator or iterator defined, throw the "not an async iterable" error

| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -76,11 +76,11 @@ helpers.asyncIterator = () => template.program.ast`
     if (typeof Symbol === "function") {
       if (Symbol.asyncIterator) {
         method = iterable[Symbol.asyncIterator]
-        if (method != null) return method();
+        if (method != null) return method.call(iterable);
       }
       if (Symbol.iterator) {
         method = iterable[Symbol.iterator]
-        if (method != null) return method();
+        if (method != null) return method.call(iterable);
       }
     }
     throw new TypeError("Object is not async iterable");
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -73,11 +73,12 @@ helpers.jsx = () => template.program.ast`
 helpers.asyncIterator = () => template.program.ast`
   export default function _asyncIterator(iterable) {
     if (typeof Symbol === "function") {
-      var method;
-      if (Symbol.asyncIterator) method = iterable[Symbol.asyncIterator];
-      else if (Symbol.iterator) method = iterable[Symbol.iterator];
-
-      if (method != null) return method.call(iterable);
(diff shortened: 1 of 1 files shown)


jquense:
cc @yodaiken

packages/babel-helpers/src/helpers.js
```diff
@@ -73,11 +73,10 @@ helpers.jsx = () => template.program.ast`
 helpers.asyncIterator = () => template.program.ast`
   export default function _asyncIterator(iterable) {
     if (typeof Symbol === "function") {
-      if (Symbol.asyncIterator) {
-        var method = iterable[Symbol.asyncIterator];
-        if (method != null) return method.call(iterable);
+      if (Symbol.asyncIterator && iterable[Symbol.asyncIterator] != null) {
+        return iterable[Symbol.asyncIterator]();
```
nicolo-ribaudo: After reading this again, it has a little bug: it gets `iterable[Symbol.asyncIterator]` and `iterable[Symbol.iterator]` twice.
It should be something like this:
```js
if (Symbol.asyncIterator && (method = iterable[Symbol.asyncIterator]) != null) return method.call(iterable);
```

jquense: whats wrong with accessing it twice?

nicolo-ribaudo:
```js
var i = 0;

var obj = {
  get [Symbol.asyncIterator]() {
    i++;
    return async function* () { yield 1; yield 2; };
  }
};

(async () => {
  for await (const x of obj) {}
  alert(i); // Should alert 1, not 2
})();
```

nicolo-ribaudo: ^ Actually this should be a test, but it can be added in a different PR.

jquense: that seems like a bit of a contrived example, but sure

nicolo-ribaudo: Usually if it doesn't add too much complexity and it isn't slower, we try to be as much spec compliant as possible.

packages/babel-helpers/src/helpers.js
```diff
@@ -72,12 +72,15 @@ helpers.jsx = () => template.program.ast`
 
 helpers.asyncIterator = () => template.program.ast`
   export default function _asyncIterator(iterable) {
+    var method
     if (typeof Symbol === "function") {
-      if (Symbol.asyncIterator && iterable[Symbol.asyncIterator] != null) {
-        return iterable[Symbol.asyncIterator]();
+      if (Symbol.asyncIterator) {
+        method = iterable[Symbol.asyncIterator]
+        if (method != null) return method();
```
nicolo-ribaudo: `method.call(iterable)`, since the `this` context should be preserved.

Issue #7733, opened 2018-04-14
Property to Function Call Visitor


jridgewell:
Part of the class Private Properties transform is the terribly complicated ["property" to function call visitor](https://github.com/CodingItWrong/babel/blob/ccd941057ab297c0f80cacb7c89e892f58e92b00/packages/babel-plugin-proposal-class-properties/src/index.js#L114-L217). It's strikingly similar to the replace supers ([1](https://github.com/CodingItWrong/babel/blob/ccd941057ab297c0f80cacb7c89e892f58e92b00/packages/babel-helper-replace-supers/src/index.js#L147-L193), [2](https://github.com/CodingItWrong/babel/blob/ccd941057ab297c0f80cacb7c89e892f58e92b00/packages/babel-helper-replace-supers/src/index.js#L199-L333)) transform.

That's because they're the same thing, just with different functions to call. This needs to be extracted into a new helper visitor, so we can cut down the duplication.

PR #7750, opened 2018-04-18
Move more class state out of replaceSupers

jridgewell:
Yes, the output is uglier. But, this is necessary for me to refactor
`replaceSupers` for #7733, which is necessary for both #7555 and
https://github.com/babel/babel/pull/7553#issuecomment-381434519.

I'm still in the middle of cleaning up all this code. Don't expect
`transformClass` to survive much longer as it's written currently.

### DIFF
--- a/packages/babel-helper-replace-supers/src/index.js
+++ b/packages/babel-helper-replace-supers/src/index.js
@@ -1,4 +1,4 @@
-import type { NodePath, Scope } from "@babel/traverse";
+import type { NodePath } from "@babel/traverse";
 import traverse from "@babel/traverse";
 import memberExpressionToFunctions from "@babel/helper-member-expression-to-functions";
 import optimiseCall from "@babel/helper-optimise-call-expression";
@@ -62,19 +62,9 @@ export const environmentVisitor = {
 const visitor = traverse.visitors.merge([
   environmentVisitor,
   {
-    ReturnStatement(path, state) {
-      // TODO get this shit out of here
-      if (!path.getFunctionParent().isArrowFunctionExpression()) {
-        state.returns.push(path);
-      }
-    },
-
     Super(path, state) {
       const { node, parentPath } = path;
-      if (parentPath.isCallExpression({ callee: node })) {
-        state.bareSupers.add(parentPath);
-        return;
-      }
+      if (!parentPath.isMemberExpression({ object: node })) return;
       state.handle(parentPath);
     },
   },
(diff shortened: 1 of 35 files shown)


packages/babel-helper-replace-supers/src/index.js
```diff
@@ -61,19 +61,9 @@ export const environmentVisitor = {
 const visitor = traverse.visitors.merge([
   environmentVisitor,
   {
-    ReturnStatement(path, state) {
-      // TODO get this shit out of here
```
hzoo: lol 🙌 ?

packages/babel-plugin-transform-classes/test/fixtures/loose/accessing-super-class/output.js
```diff
@@ -13,14 +13,14 @@ function (_Foo) {
     woops.super.test();
     _this = _Foo.call(this) || this;
 
-    _Foo.prototype.test.call(_this);
```
hzoo: I would think loose mode doesn't need this assertion right?

packages/babel-plugin-proposal-class-properties/test/fixtures/nested-class/super-property-in-key/output.js
```diff
@@ -27,7 +27,7 @@ function (_Hello) {
     babelHelpers.classCallCheck(this, Outer);
     _this = babelHelpers.possibleConstructorReturn(this, babelHelpers.getPrototypeOf(Outer).call(this));
 
-    var _babelHelpers$get$cal = babelHelpers.get(babelHelpers.getPrototypeOf(Outer.prototype), "toString", babelHelpers.assertThisInitialized(_this)).call(_this);
```
hzoo: Didn't really look into it, but is it supposed to call `assertThisInitialized` multiple twice like this?

packages/babel-helper-replace-supers/src/index.js
jridgewell: https://github.com/babel/babel/pull/7737/files#diff-74b3eb66512735b6b5013e5d41e33642R65. I've been planing this removal for the issues linked in OP.

packages/babel-plugin-proposal-class-properties/test/fixtures/nested-class/super-property-in-key/output.js
jridgewell: No, but it won't hurt. Like I said in the OP, this output is worse because I'm not optimizing any special cases yet, I just need to refactor code out of here so I can work on access-to-function-calls so we can get Private props. Then I can start work on refactoring all of the classes transform.

packages/babel-plugin-transform-classes/test/fixtures/loose/accessing-super-class/output.js
jridgewell: It's closer to spec code, but this is only here because I haven't optimized any special cases yet.

PR #7755, opened 2018-04-18
drop support for Node.js v4

### LINKED ISSUE #7753: drop support for Node.js v4
It's about that time... 

In a little under two weeks (April 30, 2018), Node.js v4 will be End-of-Life'd.

I see that v7 has already dropped support for v0.10 and v0.12.  I wanted to encourage the Babel team to drop support for v4.x as soon as is feasible. Mocha will also be dropping support ASAP.

boneskull:
| Fixed Issues?            | Fixes #7753
| Patch: Bug Fix?          | n/a
| Minor: New Feature?      | n/a
| Documentation PR         | n/a
| Any Dependency Changes?  | n/a

- remove version 4 from Travis build matrix
- update environment support doc
- update `CONTRIBUTING.md`
- update `engines` field of root `package.json`

If I need to make changes elsewhere, please let me know.  I didn't see that the `engines` field was present (nor getting copied) into packages' `package.json`; maybe it should be.

(I'll also look to see if the website needs a PR for this.)


boneskull:
I don't see the minimum Node.js version mentioned anywhere on the site; not sure where exactly that information should live.  Any suggestions?

anajavi:
Babylon's package.json has engines field:
https://github.com/babel/babel/blob/master/packages/babylon/package.json

xtuc:
We have some engines restriction here https://github.com/babel/babel/blob/master/package.json#L70-L73.

nicolo-ribaudo:
There are some tests which use `"minNodeVersion": "6.0.0"`. That option can be removed if we don't run tests on Node 4 anymore.

jridgewell:
> That option can be removed if we don't run tests on Node 4 anymore.

Let's do that as a follow up?

danez:
I would have thought we should also change the node-version for preset-env in our own babelrc file, but as it turns out we are not even setting it. So we always apply all transforms for publish-builds. Is that on purpose?

hzoo:
Yeah it was, it didn't seem any faster at the time and it was for babel-standalone but that is compiled separately now (and needs transform-runtime or polyfill).

boneskull:
Is there anything actionable here for me to do?

jridgewell:
Good enough for now.

If want to do a bit more, we can now remove every `minNodeVersion` that's <= to 6 in the code base, since that's now the default minimum version.

boneskull:
GUESS NOT

PR #7763, opened 2018-04-20
Implement MemberExpressionToFunctions helper

### LINKED ISSUE #7733: Property to Function Call Visitor

jridgewell:
Fixes #7733.

This will also be used to simplify the Private Fields transform, which had [almost the same code](https://github.com/CodingItWrong/babel/blob/ccd941057ab297c0f80cacb7c89e892f58e92b00/packages/babel-plugin-proposal-class-properties/src/index.js#L114-L217) hand written.

### DIFF
--- a/packages/babel-helper-member-expression-to-functions/README.md
+++ b/packages/babel-helper-member-expression-to-functions/README.md
@@ -1,5 +1,66 @@
 # @babel/helper-member-expression-to-functions
 
+Helper function to replace certain member expressions with function calls
+
 ## Usage
 
-TODO
+> Designed for internal Babel use.
+
+Traverses the `path` using the supplied `visitor` and an augmented `state`.
+
+```js
+const visitor = {
+  MemberExpression(memberPath, state) {
+
+    if (someCondition(memberPath)) {
+
+      // The handle method is supplied by memberExpressionToFunctions.
+      // It should be called whenever a MemberExpression should be
+      // converted into the proper function calls.
+      state.handle(memberPath);
+
+    }
+
+  },
+};
+
+// The helper requires three special methods on state: `get`, `set`, and
+// `call`.
+// Everything else will be passed through as normal.
+const state = {
+  get(memberPath) {
+    // Return some AST that will get the member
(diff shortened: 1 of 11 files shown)


packages/babel-helper-member-expression-to-functions/README.md
```diff
@@ -0,0 +1,5 @@
+# @babel/helper-member-expression-to-functions
+
+## Usage
+
+TODO
```
xtuc: I don't think we are actually going to document it. We could just mention: internal usage only?

packages/babel-helper-member-expression-to-functions/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@babel/helper-member-expression-to-functions",
```
xtuc: Author is missing. I guess you can put your name into it?

packages/babel-helper-member-expression-to-functions/README.md
jridgewell: Just did. 😛

packages/babel-helper-member-expression-to-functions/package.json
jridgewell: Done.

PR #7766, opened 2018-04-21
Correct update expression Number coercion

jridgewell:
You have to `ToNumber` the `UpdateExpression`'s `argument`.

### DIFF
--- a/packages/babel-plugin-transform-modules-systemjs/test/fixtures/systemjs/hoisting-bindings/output.mjs
+++ b/packages/babel-plugin-transform-modules-systemjs/test/fixtures/systemjs/hoisting-bindings/output.mjs
@@ -4,8 +4,10 @@ System.register([], function (_export, _context) {
   var c;
 
   function a() {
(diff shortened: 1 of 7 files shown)


packages/babel-plugin-transform-modules-systemjs/test/fixtures/systemjs/hoisting-bindings/output.mjs
```diff
@@ -4,8 +4,10 @@ System.register([], function (_export, _context) {
   var c;
 
   function a() {
+    var _c;
+
     alert("a");
-    _export("c", c + 1), c++;
+    _export("c", _c = c++), _c;
```
nicolo-ribaudo: This is wrong, it should export c's new value, not the old one.

jridgewell: Whoops.

PR #7777, opened 2018-04-22
Use Object Spread Syntax

jridgewell:


hzoo:
I had a link to an ASTExplorer version http://astexplorer.net/#/gist/d5bc2800efc9397ff9d2646feee07297/c55f0c1318edd40bb853599afd16175608557dc6

hzoo:
Also is it reasonable that we would use `loose` mode and `useBuiltIns` then in the Babel config?

codemods/babel-plugin-codemod-object-assign-to-object-spread/src/index.js
```diff
@@ -0,0 +1,34 @@
+import syntaxObjectRestSpread from "@babel/plugin-syntax-object-rest-spread";
+import { types as t } from "@babel/core";
+
+export default function() {
```
existentialism: Nit: can avoid importing types from core, with:

```
export default function({ types: t })
```

PR #7780, opened 2018-04-22
Test source before geting the key [preventing Object.keys(false)]

namiltd:
| Fixed Issues?            | Object.keys(false) fix


loganfsmyth:
A lot of the time Babel assumes that you've loaded a polyfill. For instance core-js polyfills `Object.keys` to handle things like this.

namiltd:
Optimized

packages/babel-helpers/src/helpers.js
```diff
@@ -377,7 +377,7 @@ helpers.objectSpread = () => template.program.ast`
 
   export default function _objectSpread(target) {
     for (var i = 1; i < arguments.length; i++) {
-      var source = (arguments[i] != null) ? arguments[i] : {};
+      var source = arguments[i] != null && arguments[i] !== false ? arguments[i] : {};
```
jridgewell: We can fix this and speed up all non-objets:

```js
var source = arguments[i];
if (source !== Object(source)) continue;

//...

nicolo-ribaudo: I believe that strings should be spread.
The check could be `if (!source) continue;`

loganfsmyth: If we do want to do this, which I'm not convinced we do, it seems like we should approach it as including our own polyfill for `Object.keys`, but really it seems not too hard to tell people to load their own polyfills?

namiltd:
Implemented filtering of non-objects.

namiltd:
I have tested, the 'source !== Object (source)' works correctly

packages/babel-helpers/src/helpers.js
namiltd: I discovered the problem in the Bootstrap 4 framework that uses babel.

namiltd:
I understand, therefore, that this code has little chance of being introduced, which is a pity.

namiltd:
The problem has disappeared, probably https://github.com/babel/babel/pull/7777 blocking this type of calls.

PR #7784, opened 2018-04-23
Upgrade Babel to self-host with beta.46

loganfsmyth:
| Fixed Issues?            | N
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

As expected, definitely some work to fix this one. Nothing unexpected though. Changes are:

* `.babelrc` search stops at package boundary, so monorepos like ours need to use a `babel.config.js` or have a `.babelrc` per-package that `extends` some other central config. We went the `babel.config.js` route
* Since it finds `babel.config.js` in the working directory, our root config started applying to test files too, which is a problem unique to us, and not really an issue for others. Fixed by passing a non-root `cwd` to all the Babel calls that failed.
* Our bundle builds used to explicitly pass `babelrc: false` to avoid config files in `node_modules`, but that's no longer an issue!


loganfsmyth:
Broke this again haha, I'll fix it eventually, don't have time right now 😬

marcinczenko:
Is there any documentation on how to properly discover .babelrc with monorepos? I just upgraded from `7.0.0-beta.44` to `7.0.0-beta.46` and all my jest tests are now failing. I have `.babelrc` in every package but they are not respected any more...

hzoo:
This is our own PR to upgrade so I think that can tell you what to do with `babel.config.js`. I would look at https://github.com/babel/babel/pull/7358 for an explanation in the meantime while we add the docs.

marcinczenko:
@hzoo Thank, I will check it out.

terion-name:
How to use babelrc now with yarn runs?

I have in package.json:
```
"scripts": {
    "build": "./node_modules/.bin/babel-node src/build.js",
  }
```
If i tun `yarn run build` it fails to allocate .babelrc in root

If I run `./node_modules/.bin/babel-node src/build.js` directly — it works.

PR #7785, opened 2018-04-23
Change `const` to `var` inside helper

micmcg:
One of the helpers has a `const` rather than a `var` in its body, which causes our minifier (closure compiler) to throw an error as it is set to ES5. 

| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 


loganfsmyth:
Thanks!

Issue #7789, opened 2018-04-23
We should discourage the use of "last 2 versions" in babel-preset-env


jamiebuilds:
"last 2 versions" is a really bad way to determine browser support and means that babel-preset-env will forever compile everything.

This is because "last 2 versions" includes browsers which:

- Are no longer are maintained 
- Will not be releasing new versions
- Have 0% market share

Which includes:

- Internet Explorer (not including Edge)
- Internet Explorer Mobile (0.23% market share)
- Opera Mobile (0.01% market share)
- Blackberry Browser (0.07% market share)
- QQ Browser (0% market share)
- Baidu Browser (0% market share)

This means that anything those browsers don't support will end up getting compiled by Babel. Which translates to literally every single feature that Babel supports.

In effect, by telling people to use "last 2 versions" we're making `babel-preset-env` work exactly the same as `babel-preset-es2015`. If people are using that as part of their browser list, there is zero benefit to using `babel-preset-env`.

I think when most people read "last 2 versions" they think the last two versions of IE/Edge (as a continuation of the same browser), Safari, Firefox, Chrome, and other maintained browsers.

I think we should actively discourage people from using "last X versions" and encourage them to use purely market percentages and manual browser whitelists instead.

I think that this should be our new default:

```
>1%, not op_mini
```

But this should be what we use within documentation:

```
>1%, not ie 11, not op_mini
```

I don't think it would be a good idea for us to drop IE 11 by default for everyone. But I think we should actively encourage people towards dropping it. If people want to support it, they know exactly what is going on

Offirmo:
Excellent suggestion. Questions:
- do we have a way to target "evergreen" browsers, for ex. "evergreen last 2 versions"?
- why not opera mini?

existentialism:
I'm for this :+1:

jamiebuilds:
https://jamie.build/last-2-versions

hzoo:
Ok I'll send a PR

hzoo:
I think the issue with %/evergreen is that when you update dependency it may break given an update of browserlist % will drop more browsers (hopefully no one is surprised at what that really means, it should only be a positive 😄).

PR #7790, opened 2018-04-23
update targets [skip ci]

hzoo:
Fixes https://github.com/babel/babel/issues/7789

Wondering now if we should warn the user if they don't pass an option to preset-env or have a browserslist?


jkup:
I think this is a really cool thing to be tackling! I worry a bit about the `>0.25%` because it's not super clear what that's referring to.

hzoo:
Yeah it says it's using browserslist above but you didn't see that in the diff; I realize the description is wrong anyway though

jkup:
😀 looks like i commented too quickly!

hzoo:
Also made this change to the other repo https://github.com/babel/babel-preset-env (archived but people might still go there) and changed that to 1.x

PR #7806, opened 2018-04-25
Update variableDeclaration documentation [skip ci]

jamime:
| Fixed Issues?            | :x: 
| Patch: Bug Fix?          |:x:
| Major: Breaking Change?  |:x:
| Minor: New Feature?      |:x:
| Tests Added + Pass?      | :x:
| Documentation PR         | :heavy_check_mark:
| Any Dependency Changes?  |:x:

The **babel-types** documentation incorrectly stated that the `variableDeclaration` function accepts a `declare` argument. I have removed this.

### DIFF
--- a/packages/babel-types/README.md
+++ b/packages/babel-types/README.md
@@ -2870,7 +2870,6 @@ Aliases: `Statement`, `Declaration`
 
  - `kind`: `"var" | "let" | "const"` (required)
  - `declarations`: `Array<VariableDeclarator>` (required)
- - `declare`: `boolean` (default: `null`)
 
 ---
 


existentialism:
@jamime thanks for the PR! The babel-types documentation is actually [generated automatically](https://github.com/babel/babel/blob/master/scripts/generators/docs.js).

The reason this shows up is that `declarations` is in the `builder`:
https://github.com/babel/babel/blob/master/packages/babel-types/src/definitions/core.js#L810

jamime:
@existentialism I did not realise this, although it is the `declare` param which is not in the `builder`.

existentialism:
@jamime sorry, that's what I get for not having coffee this morning :)

`declare` needs to be in the `builder` array for it to work.

The README represents the current state of `master`, while babeljs.io represents 6.x since v7 is still in beta. (we definitely agree this is causing confusion, and are actively working on getting versioned docs up asap)

jamime:
@existentialism Thanks for the explanation!

PR #7809, opened 2018-04-25
Clean up and add some additional polyfill mappings in preset-env

existentialism:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #5978
| Patch: Bug Fix?          | y
| Major: Breaking Change?  | n
| Minor: New Feature?      | n
| Tests Added + Pass?      | y/y
| Documentation PR         | 
| Any Dependency Changes?  |
| License                  | MIT

Depends on https://github.com/babel/babel/pull/7808.

This PR tweaks/fixes/adds a number of polyfill-to-compat-table mappings and merges them into a single collection.

This should make the eventual upgrade to core-js@3 a bit easier to read (https://github.com/babel/babel/pull/7646) as it should be a lot of `es7.* -> es.*`.

All the credit goes to @zloirock.

PR #7808, opened 2018-04-25
Bump compat-table for preset-env

existentialism:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | 
| Patch: Bug Fix?          | n
| Major: Breaking Change?  | n
| Minor: New Feature?      | n
| Tests Added + Pass?      | Yes
| Documentation PR         | 
| Any Dependency Changes?  | y
| License                  | MIT

Staging a few small-ish PRs to hopefully help make landing https://github.com/babel/babel/pull/7646 easier.

This adds mappings for ios 11.3, and limits the scope of the `es6.set` mapping as some stage-1 proposals were added to compat-table.

PR #7810, opened 2018-04-25
Improve useBuiltIns: usage mappins in preset-env

existentialism:
| Fixed Issues?            | 
| Tests Added + Pass?      | n/y
| Documentation PR         | 
| Any Dependency Changes?  |

Similar to https://github.com/babel/babel/pull/7809, this depends on https://github.com/babel/babel/pull/7808.

This PR tweaks/fixes/adds a number of the `useBuiltIns: usage` mappings. Which should make the eventual upgrade to core-js@3 a bit easier to read (https://github.com/babel/babel/pull/7646), as it should be a lot of es7.* -> es.*.

All the credit goes to @zloirock.


packages/babel-preset-env/src/built-in-definitions.js
```diff
@@ -1,67 +1,97 @@
...
     Int32Array: "es6.typed.int32-array",
-    Uint32Array: "es6.typed.uint32-array",
-    Float32Array: "es6.typed.float32-array",
-    Float64Array: "es6.typed.float64-array",
-    Map: "es6.map",
-    Set: "es6.set",
-    WeakMap: "es6.weak-map",
-    WeakSet: "es6.weak-set",
+    Map: ["es6.map"].concat(CommonIterators),
+    Number: "6es.number.constructor",
```
zloirock: typo

packages/babel-preset-env/data/built-in-features.js
```diff
@@ -79,7 +79,33 @@ const es2015 = {
   },
 
   "es6.map": "Map",
-  "es6.set": "Set",
+
+  // This is explicit due to prevent the stage-1 Set proposals under the
+  // category "Set methods" from being included.
```
zloirock: Interesting. I didn't think that it could cause any problems.

existentialism: Yeah... currently, the the mapping name (in this case "Set") gets matched via `indexOf`, so the es6 compat-table data `Set` and the esnext `Set methods` both match.

We can improve this for sure, but I wanted to keep the [upgrade PR](https://github.com/babel/babel/pull/7808) small.

PR #6656, opened 2017-11-01
Optimize class properties output

Andarist:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | n/a
| Patch: Bug Fix?          | no
| Major: Breaking Change?  | no
| Minor: New Feature?      | no
| Tests Added + Pass?      | Yes
| Documentation PR         | no
| Any Dependency Changes?  | no

Similar to change in https://github.com/babel/babel/pull/6652 . Additionally I switch to using `defineProperty` helper instead of using bare `Object.defineProperty` call - which should have benefit of not having to hit slow `Object.defineProperty` path in most cases.

PR #6652, opened 2017-10-31
Optimize computed properties output (byte-wise)

Andarist:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | n/a
| Patch: Bug Fix?          | no
| Major: Breaking Change?  | no
| Minor: New Feature?      | no
| Tests Added + Pass?      | Yes
| Documentation PR         | no
| Any Dependency Changes?  | no

It is a small "optimization" of the computed properties output. The change basically wraps (if possible) previous element - `initProps` object or previous `defineProperty` call when "pushing" new `defineProperty` calls.

It leverages the fact that `defineProperty` returns the passed object, therefore in most cases we do not need to create temporary references and stuff.

[babel-plugin-transform-computed-properties/test/fixtures/spec/multiple](https://github.com/babel/babel/compare/master...Andarist:polish/computed-properties?expand=1#diff-226ea89fff707b39966cfbb41e57eaa8) illustrates this nicely.

**Before**
```js
var _obj;
var obj = (_obj = {}, babelHelpers.defineProperty(_obj, "x" + foo, "heh"), babelHelpers.defineProperty(_obj, "y" + bar, "noo"), _obj);
```

**After**
```js
var obj = babelHelpers.defineProperty(babelHelpers.defineProperty({}, "x" + foo, "heh"), "y" + bar, "noo");
```

Not only it saves some bytes in the output, it is probably more friendly for the minifiers too.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/6652"></a>

PR #7814, opened 2018-04-26
Don't fold class property calls

jridgewell:
This undoes the property call folding from #6656.

It complicates the private property transforms, since they boil down to `map.set(this, value)` and we definitely don't want the next call define a property on the map.

/cc @Andarist

### DIFF
--- a/packages/babel-plugin-proposal-class-properties/src/index.js
+++ b/packages/babel-plugin-proposal-class-properties/src/index.js
@@ -51,26 +51,19 @@ export default declare((api, options) => {
     environmentVisitor,
   ]);
 
-  const foldDefinePropertyCalls = nodes =>
-    t.expressionStatement(
-      nodes.reduce((folded, node) => {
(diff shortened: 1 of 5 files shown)


Andarist:
Huh, I havent realized my other PR got merged in 😉 Could you post a simple input & output of the problematic code? I remember implementing a possible [deopt in other similar PR](https://github.com/babel/babel/pull/6652/files#diff-1d9e9f3e4f77952b0bacc8ff79b20d1bR140). Admittedly it has complicated the code (what doesn't? 😉) but produced more optimized code in majority of scenarios. Let me know what do you think.

jridgewell:
```js
class Test {
  foo = 1;
  #bar = 2;
  baz = 3;
}

// Desugars
function Test() {
  define(
    _bar.set(
      define(this, 'foo', 1),
      2
    ),
    'baz',
    3
  );
}
```

We can look into re-optimizing after I land private props.

PR #7818, opened 2018-04-26
Bring `pipelineOperator` flag in line with minimal

mAAdhaTTah:

The minimal proposal requires parentheses around arrow functions and bans await from the pipeline. This brings babylon & the pipeline proposal plugin into alignment w/ the current spec.


mAAdhaTTah:
cc @littledan @js-choi Upstreaming these changes from our fork.

packages/babylon/src/parser/expression.js
```diff
@@ -310,8 +310,25 @@ export default class ExpressionParser extends LValParser {
         const startLoc = this.state.startLoc;
 
         if (op === tt.pipeline) {
-          // Support syntax such as 10 |> x => x + 1
-          this.state.potentialArrowAt = startPos;
+          const lookahead = this.lookahead();
+
+          if (lookahead.type === tt.arrow) {
```
nicolo-ribaudo: Can we handle `|> ArrowFuncition` like we handle `|| ArrowFunction` and avoid using lookahead?

mAAdhaTTah: 1. TIL `const val = false || x => x` is a SyntaxError.
2. Where is that logic implemented?

nicolo-ribaudo: I *think* that removing line 314 (which you already did) should be enough, because it was used to "hack" Babylon's default precedence rules.

You only need to remove the custom error message (it would produce a less useful error message, but it would avoid using `lookahead` which does a deep copy of the parser state).

Also, can you add a test for `a |> (b) => b`? Babylon handles arrow functions with and without parentheses differently.

nicolo-ribaudo:
In https://github.com/babel/babel/pull/7818#discussion_r184537096 I meant a failing test 🙂

mAAdhaTTah:
@nicolo-ribaudo Whoops, done!

PR #7821, opened 2018-04-26
Rename decorators&decorators2 plugins to decorators-legacy&decorators.

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Reasons:
1) Naming consistency with Babel plugins
2) Sooner or later the decorators2 plugin should become decorators


nicolo-ribaudo:
@diervo They will stay separate, until we can remove the legacy version (probably in Babel 8?)

hzoo:
This is the versioning issue we have been discussing. I think if we drop the stage 0-2 presets we should think about independent versioning for proposals and maybe keep them at 0.x.

PR #6238, opened 2017-09-13
Named exports should always take priority over star exports

loganfsmyth:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues             | Fixes #2438, Fixes #3468 
| Patch: Bug Fix?          | 
| Major: Breaking Change?  | 
| Minor: New Feature?      | 
| Tests Added/Pass?        | 
| Spec Compliancy?         | 
| License                  | MIT
| Doc PR                   | 
| Any Dependency Changes?  | 

Just ported https://github.com/babel/babel/pull/3468 onto my new modules refactor. Left the author info on the commits as @arcanis since he did all the work, I just don't want to force you to do yet another pass on your PR because I introduced massive conflicts :P

I made one change, which is that it uses `hasOwnProperty` instead of `.indexOf`.

@arcanis, sound good?

PR #3468, opened 2016-04-13
Named exports should always take priority over star exports

arcanis:
This commit fixes https://phabricator.babeljs.io/T2438 by allowing imported symbols to be overriden.

There's still a flag that has to be raised: there might be an ES3 compatibility issue (check the following). However, I feel like an ES3 transform could be made to convert these constructs (ie. defineProperty calls that only use a constexpr getter) into proper ones (ie. defineProperty calls that use values instead of getters and are thus easily polyfillable).

I'm using this code on the development branch of a real-size application, and it works fine.

PR #7822, opened 2018-04-26
Backport: Named exports should always take priority over star exports

### LINKED ISSUE #2438: Named exports should always take priority over star exports.
Also related to the export priority in https://github.com/babel/babel/issues/2437, since this works if the `export var bar` comes after `export * from` and pre-assigning `undefined` https://github.com/babel/babel/issues/2421.

Given a named export `bar`, and another module `foo` that _also_ has an export named `bar`.

```
export var bar;
export * from 'foo';

function fn(){
  foo = 4;
}
```

Currently the `_defaults` helper assigns variables essentially checks if `exports[key] === undefined` for each exported name from `foo`, and overwrites it, but in an instance like this, the export hasn't been assigned yet, so the `export *` will assign `bar` as the value from `foo`, instead of properly preserving the value of `undefined`.

Once https://github.com/babel/babel/issues/2421 lands, seems like that check should be changed to use `hasOwnProperty`.

A secondary question on this:

`_defaults` also does `Object.getOwnPropertyDescriptor(defaults, key)`, shouldn't that be `Object.getOwnPropertyDescriptor(obj, key)`?

asapach:
| Fixed Issues?            | Fixes #2438
| Major: Breaking Change?  | 
| Minor: New Feature?      | 
| Documentation PR         | 
| Any Dependency Changes?  | 

This PR back-ports #6238 and #3468 for 6.x. The fix for #2438 has landed in 7.x, but is not available in 6.x

//CC @loganfsmyth

### DIFF
--- a/packages/babel-plugin-transform-es2015-modules-amd/test/fixtures/amd/export-from/expected.js
+++ b/packages/babel-plugin-transform-es2015-modules-amd/test/fixtures/amd/export-from/expected.js
@@ -4,8 +4,10 @@ define(["exports", "foo"], function (exports, _foo) {
   Object.defineProperty(exports, "__esModule", {
     value: true
   });
+  var _exportNames = {};
   Object.keys(_foo).forEach(function (key) {
     if (key === "default" || key === "__esModule") return;
+    if (Object.prototype.hasOwnProperty.call(_exportNames, key)) return;
     Object.defineProperty(exports, key, {
       enumerable: true,
       get: function () {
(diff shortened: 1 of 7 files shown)


xtuc:
We are not planning to release new Babel 6.x version. Where did you see that we wanted to backport it?

asapach:
I've noticed that there were several fixes backported in `v6.26.2` and `v6.26.3` and thought this one might be useful as well. I've run into this issue on one of my projects and verified that this fix works. It might benefit the community if it's in the official repo.

loganfsmyth:
I think at this point we're focused mostly on 7.x and anything backported needs to be extremely minimal and unlikely to cause issues. In my opinion that limits it to strictly small bug fixes in the transforms themselves, rather than new features like this.

asapach:
Feel free to close this PR.

loganfsmyth:
Okie doke, sorry it didn't work out :(

asapach:
For future generations: I've published the fixed version as [@asapach/babel-plugin-transform-es2015-modules-commonjs](https://www.npmjs.com/package/@asapach/babel-plugin-transform-es2015-modules-commonjs) on npm.

PR #7830, opened 2018-04-27
fix(babel-plugin-transform-typescript): do not remove unused imports

### LINKED ISSUE #7592: extremely odd interaction of scope, filename and preset-typescript
### Input Code

Repro available: https://github.com/jeysal/babel-scope-filename-typescript-repro

```js
const { transform } = require("@babel/core");

const plugin = ({ types: t }) => ({
  visitor: {
    Program(path) {
      const [importPath] = path.unshiftContainer(
        "body",
        t.importDeclaration(
          [t.importDefaultSpecifier(t.identifier("local"))],
          t.stringLiteral("source")
        )
      );
      path.scope.registerDeclaration(importPath);
      //path.scope.crawl() // same result
    }
  }
});

console.log(
  transform(`a();`, {
    presets: ["@babel/preset-env"],
    plugins: [plugin],
    filename: "test.js"
  }).code
);
```

```json
{
  "devDependencies": {
    "@babel/core": "^7.0.0-beta.42",
    "@babel/preset-env": "^7.0.0-beta.42",
    "@babel/preset-typescript": "^7.0.0-beta.42"
  }
}
```

```json
{
  "presets": [
    [
      "@babel/preset-env",
      {
        "targets": {
          "node": 6
        }
      }
    ],
    "@babel/preset-typescript"
  ]
}
```

### Expected Behavior
Generated code has `_interopRequireDefault` boilerplate that `require`s from `'source'`

### Current Behavior
```js
"use strict";

a();
```

### Context

Here comes the weird part:
The import can be brought back by changing any one of the following things:
* remove the `filename` in the transform call
* change `a()` to `local()`
* disable `@babel/preset-typescript` in .babelrc
* do not `registerDeclaration` / `crawl`

### Your Environment

| ---------------- | -------
| Babel            | 7.0.0-beta.42
| Babylon          | 
| node             | 8.10.0
| yarn             | 1.5.1
| Operating System | Linux version 4.15.9-1-ARCH

jeysal:
| Fixed Issues?            | Fixes #7592 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

See #7592 for an example of what this fixes.
The added `test` passes *only* with the `src` changes ✓
The other tests still pass ✓ but if anyone with more experience thinks the new behavior is not desired / could break stuff, please come forward :)

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -260,6 +260,10 @@ export default declare(api => {
   }
 
   function isImportTypeOnly(binding, programPath) {
+    if (!binding.referencePaths.length) {
+      return false;
+    }
+
     for (const path of binding.referencePaths) {
       if (!isInType(path)) {
         return false;
--- a/packages/babel-plugin-transform-typescript/test/fixtures/imports/keep-unused/input.mjs
+++ b/packages/babel-plugin-transform-typescript/test/fixtures/imports/keep-unused/input.mjs
@@ -1,0 +1,1 @@
+import a from "lib";
--- a/packages/babel-plugin-transform-typescript/test/fixtures/imports/keep-unused/output.mjs
+++ b/packages/babel-plugin-transform-typescript/test/fixtures/imports/keep-unused/output.mjs
@@ -1,0 +1,1 @@
+import a from "lib";


jeysal:
What's awkward about keeping it the way it is:
```javascript
const { transform } = require("@babel/core");

const plugin = ({ types: t }) => ({
  visitor: {
    CallExpression(path) {
      if (path.node.callee.name === "a") {
        path.scope
          .getProgramParent()
          .path.unshiftContainer(
            "body",
            t.importDeclaration(
              [t.importDefaultSpecifier(t.identifier("local"))],
              t.stringLiteral("source")
            )
          );

        path.node.callee = t.identifier("local");
        path.scope.crawl();
      }
    }
  }
});

console.log(
  transform(`a();`, {
    presets: ["@babel/preset-typescript"],
    plugins: [plugin]
  }).code
);
```
works (does not remove the import),
while `scope.crawl`ing one line earlier instead removes it.

So if I'm correct about this, after adding an import, you must not crawl the scope until you've add a reference to that import, otherwise the import will be removed.
That doesn't seem intuitive to me, especially since crawling after adding an import is generally a good idea to avoid incorrect behavior of subsequent `scope.bindingIdentifierEquals` etc.
Is there anything other than simply not removing unused imports we could do to improve this behavior?

loganfsmyth:
Perhaps we should consider having the TS plugin remove them in the `Program` visitor so it doesn't visit newly-injected import declarations.

jeysal:
I've created #7833 for the solution @loganfsmyth proposed. Closing this PR.

PR #7833, opened 2018-04-28
fix(transform-typescript): do not elide injected imports

jeysal:
| Fixed Issues?            | #7592 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

see https://github.com/babel/babel/pull/7830#issuecomment-385189226


jeysal:
will take a look at failing tests asap

jeysal:
Fixed the tests on node 6 by removing ES2017 trailing commas from function calls in the test's `plugin.js`.
`make lint` does not catch this, but the file is now technically not formatted correctly according to the `overrides` in `.prettierrc`. Do we do anything about this?

jeysal:
Anyone else? @nicolo-ribaudo?

PR #7835, opened 2018-04-29
Change babel-preset-env docs according Browserslist best practices

### LINKED ISSUE #7829: Promoting Browserslist in babel-preset-env
Having common config with target browsers instead of separate `browsers` option in every tool will be better for our community.

1. The single place with config to Autoprefixer, babel and many other tools will reduce mistakes when you developer change config in only one place
2. The single config will be useful for new developers. Instead of asking the same “do we support X browser”, a new developer will just look into Browserslist config
3. Common config will force developers to create more tool based on target browsers. Like [postcss-normalize](https://github.com/jonathantneal/postcss-normalize), which insert `normalize.css` rules needed only for target browsers.

Autoprefixer, PostCSS and CRA force only Browserslist config, instead of per-tool `browsers` option.

If you agree with it, I can send PR to `babel-preset-env`

/cc @hzoo @yavorsky

ai:
| Fixed Issues?            | Fixes #7829
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | 
| Any Dependency Changes?  |

Having common config with target browsers instead of separate `browsers` option in every tool will be better for our community.

1. The single place with config to Autoprefixer, babel and many other tools will reduce mistakes when you developer change config in only one place
2. The single config will be useful for new developers. Instead of asking the same “do we support X browser”, a new developer will just look into Browserslist config
3. Common config will force developers to create more tool based on target browsers. Like [postcss-normalize](https://github.com/jonathantneal/postcss-normalize), which insert `normalize.css` rules needed only for target browsers.

Autoprefixer, PostCSS and CRA force only Browserslist config, instead of per-tool `browsers` option.

/cc @yavorsky

### DIFF
--- a/packages/babel-preset-env/README.md
+++ b/packages/babel-preset-env/README.md
@@ -29,7 +29,7 @@ not dead
 
 The full list of queries could be found in [Browserslist docs](https://github.com/browserslist/browserslist#queries).
 
-If you need to use different browsers for Autoprefixer and Babel, you can use `browsers` options of each tool.
+If you need to use different browsers for Babel, you can specify Browserslist queries in `targets.browsers` option.
 
 You may also target browsers supporting ES Modules (https://www.ecma-international.org/ecma-262/6.0/#sec-modules). When specifying this option, the browsers field will be ignored. You can use this approach in combination with `<script type="module"></script>` to conditionally serve smaller scripts to users (https://jakearchibald.com/2017/es-modules-in-browsers/#nomodule-for-backwards-compatibility).
 
@@ -47,7 +47,7 @@ You may also target browsers supporting ES Modules (https://www.ecma-internation
 }
 ```
 
-Similarly, if you're targeting Node.js instead of the browser, you can configure @babel/preset-env to only include the polyfills and transforms necessary for a particular version:
+Similarly, if you're targeting Node.js instead of the browser, you can configure `@babel/preset-env` to only include the polyfills and transforms necessary for a particular version:
 
 ```json
 {
--- a/packages/babel-preset-env/README.md
+++ b/packages/babel-preset-env/README.md
@@ -31,7 +31,7 @@ not dead
 
 The full list of queries could be found in [Browserslist docs](https://github.com/browserslist/browserslist#queries).
 
-If you need to use different browssers for Autoprefixer and Babel, you can use `browsers` options of each tool.
+If you need to use different browsers for Autoprefixer and Babel, you can use `browsers` options of each tool.
 
 You may also target browsers supporting ES Modules (https://www.ecma-international.org/ecma-262/6.0/#sec-modules). When specifying this option, the browsers field will be ignored. You can use this approach in combination with `<script type="module"></script>` to conditionally serve smaller scripts to users (https://jakearchibald.com/2017/es-modules-in-browsers/#nomodule-for-backwards-compatibility).
 
--- a/packages/babel-preset-env/README.md
+++ b/packages/babel-preset-env/README.md
@@ -18,22 +18,20 @@ Without any configuration options, @babel/preset-env behaves exactly the same as
 
 You can also configure it to only include the polyfills and transforms needed for the browsers you support. Compiling only what's needed can make your bundles smaller and your life easier.
 
-This example only includes the polyfills and code transforms needed for coverage of users > 0.25%, ignoring Internet Explorer 11 and Opera Mini. We use [browserslist](https://github.com/ai/browserslist) to parse this information, so you can use [any valid query format supported by browserslist](https://github.com/ai/browserslist#queries).
+By default, it includes the polyfills and code transforms needed for [this modern browsers](http://browserl.ist/?q=defaults).
 
-```js
-{
-  "presets": [
-    ["@babel/preset-env", {
-      "targets": {
-        // The % refers to the global coverage of users from browserslist
-        "browsers": [ ">0.25%", "not dead"]
-      }
-    }]
-  ]
-}
+If you want to specify own target browsers, we recommend to use [`.browserslistrc`](https://github.com/browserslist/browserslist) config. This config is used by many tools including Autoprefixer.
+
+This example only includes the polyfills and code transforms needed for coverage of users > 0.25%, ignoring browsers without security updates like IE Internet Explorer 10 and BlackBerry.
+
+```
+> 0.25%
+not dead
 ```
 
-> You can also target individual versions of browsers instead of using a query with `"targets": { "chrome": "66" }`.
+The full list of queries could be found in [Browserslist docs](https://github.com/browserslist/browserslist#queries).
+
+If you need to use different browssers for Autoprefixer and Babel, you can use `browsers` options of each tool.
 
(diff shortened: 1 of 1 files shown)


packages/babel-preset-env/README.md
```diff
@@ -18,22 +18,20 @@ Without any configuration options, @babel/preset-env behaves exactly the same as
 
 You can also configure it to only include the polyfills and transforms needed for the browsers you support. Compiling only what's needed can make your bundles smaller and your life easier.
 
-This example only includes the polyfills and code transforms needed for coverage of users > 0.25%, ignoring Internet Explorer 11 and Opera Mini. We use [browserslist](https://github.com/ai/browserslist) to parse this information, so you can use [any valid query format supported by browserslist](https://github.com/ai/browserslist#queries).
+By default, it includes the polyfills and code transforms needed for [this modern browsers](http://browserl.ist/?q=defaults).
```
yavorsky: Actually, it's not by default for preset-env. [We are rewriting](https://github.com/babel/babel/blob/master/packages/babel-preset-env/src/targets-parser.js#L126) defaults to targets specified by the user. It's because w/o any configuration we want all transofrm/polyfills to be applied. It's safe and easier to migrate from ES2015 preset.

So by default, it includes the polyfills and code transform needed for **all browsers**.

packages/babel-preset-env/README.md
```diff
@@ -18,22 +18,20 @@ Without any configuration options, @babel/preset-env behaves exactly the same as
...
+
+```
+> 0.25%
+not dead
 ```
 
-> You can also target individual versions of browsers instead of using a query with `"targets": { "chrome": "66" }`.
+The full list of queries could be found in [Browserslist docs](https://github.com/browserslist/browserslist#queries).
+
+If you need to use different browssers for Autoprefixer and Babel, you can use `browsers` options of each tool.
```
yavorsky:
```diff
- browssers
+ browsers
```

yavorsky: Also, after `you can use browser options of each tool`: we could have [removed JSON `targets.browsers` configuration example](https://github.com/babel/babel/pull/7835/files#diff-da1de098f27287f389b3b8f85303e4f1L23) from above.

packages/babel-preset-env/README.md
ai: Is a having different behavior with Autoprefixer and other tools is a great idea for community and long-term?

Also Browserslist `defaults` has a lot of browsers. Most of the browsers which are really in use by someone. http://browserl.ist/?q=defaults

Note, that if you will take *all* browsers, it means Chrome 3, old Opera and many browsers which you never tested. Are you sure, that you truly support them?

packages/babel-preset-env/README.md
ai: I remove it by design. Most users don’t read the text. They just quick look for example and copy-paste them to try, rather than read all warnings and explanation.

There is a good rule for frameworks. Dangerous things must be harder to implement, rather than safer things. Remember how React is using long options (`dangerouslySetInnerHTML`) to force people to avoid them.

Same logic here. If you really want to use `browsers` (there are only a few cases, like bundlers) you are capable enough to understand docs (not everyone creates bundlers). For others, it will be a good limit to use best practices.

ai:
@yavorsky I removed text about defaults. We can discuss it later (it is not part of current PR). https://github.com/ai/babel/commit/b9b383a9e2d7f107f3b2165dbf1075eb8b613e07

packages/babel-preset-env/README.md
```diff
@@ -18,22 +18,18 @@ Without any configuration options, @babel/preset-env behaves exactly the same as
...
+
+```
+> 0.25%
+not dead
 ```
 
-> You can also target individual versions of browsers instead of using a query with `"targets": { "chrome": "66" }`.
+The full list of queries could be found in [Browserslist docs](https://github.com/browserslist/browserslist#queries).
+
+If you need to use different browsers for Autoprefixer and Babel, you can use `browsers` options of each tool.
```
nicolo-ribaudo: Maybe it should be specified that it is `targets.browsers`?

ai: Good idea. Fixed.

packages/babel-preset-env/README.md
```diff
@@ -18,22 +18,18 @@ Without any configuration options, @babel/preset-env behaves exactly the same as
 
 You can also configure it to only include the polyfills and transforms needed for the browsers you support. Compiling only what's needed can make your bundles smaller and your life easier.
 
-This example only includes the polyfills and code transforms needed for coverage of users > 0.25%, ignoring Internet Explorer 11 and Opera Mini. We use [browserslist](https://github.com/ai/browserslist) to parse this information, so you can use [any valid query format supported by browserslist](https://github.com/ai/browserslist#queries).
+If you want to specify own target browsers, we recommend to use [`.browserslistrc`](https://github.com/browserslist/browserslist) config. This config is used by many tools including Autoprefixer.
```
existentialism: Some edits:

> To target specific browsers, we recommend using a [`.browserslistrc`](https://github.com/browserslist/browserslist) config, which is also used by many tools including Autoprefixer.

packages/babel-preset-env/README.md
```diff
@@ -18,22 +18,18 @@ Without any configuration options, @babel/preset-env behaves exactly the same as
...
-  "presets": [
-    ["@babel/preset-env", {
-      "targets": {
-        // The % refers to the global coverage of users from browserslist
-        "browsers": [ ">0.25%", "not dead"]
-      }
-    }]
-  ]
-}
+This example only includes the polyfills and code transforms needed for coverage of users > 0.25%, ignoring browsers without security updates like IE Internet Explorer 10 and BlackBerry.
```
existentialism: Some edits:

> For example, to only include polyfills and code transforms needed for users whose browsers have >0.25% market share (ignoring browsers without security updates like IE 10 and BlackBerry):
> 
> ```sh
> > 0.25%
> not dead
> ```

packages/babel-preset-env/README.md
```diff
@@ -18,22 +18,18 @@ Without any configuration options, @babel/preset-env behaves exactly the same as
...
+
+```
+> 0.25%
+not dead
 ```
 
-> You can also target individual versions of browsers instead of using a query with `"targets": { "chrome": "66" }`.
+The full list of queries could be found in [Browserslist docs](https://github.com/browserslist/browserslist#queries).
+
+If you need to use different browsers for Babel, you can specify Browserslist queries in `targets.browsers` option.
```
existentialism: Small edits in bold (also think it would be helpful to link to the target.browsers option):

> If you need to use different browsers for Babel, you can **also** specify Browserslist queries in **the** [`targets.browsers` option](https://github.com/babel/babel/tree/master/packages/babel-preset-env#targetsbrowsers) in your Babel config.

ai:
@existentialism done, thanks for fixes

existentialism:
@ai thanks!

ai:
@existentialism oops, mistake in `master`

```diff
- If you want to specify own target browsers, we recommend to target specific browsers, we recommend using a [`.browserslistrc`](https://github.com/browserslist/browserslist) config, which is also used by many tools including Autoprefixer.
+ If you want to specify own target browsers, we recommend using a [`.browserslistrc`](https://github.com/browserslist/browserslist) config, which is also used by many tools including Autoprefixer.
```

Can you fix it quickly in the master?

existentialism:
@ai yep!

PR #7854, opened 2018-05-02
remove @babel/preset-latest from docs

pravdomil:
because it's deprecated

[see preview](https://github.com/pravdomil/babel/blob/patch-2/packages/babel-preset-env/README.md)

### DIFF
--- a/packages/babel-preset-env/README.md
+++ b/packages/babel-preset-env/README.md
@@ -6,7 +6,7 @@
 npm install @babel/preset-env --save-dev
 ```
 
-Without any configuration options, `@babel/preset-env` behaves exactly the same as [`@babel/preset-es2015`](https://github.com/babel/babel/tree/master/packages/babel-es2015), [`@babel/preset-es2016`](https://github.com/babel/babel/tree/master/packages/babel-es2016) and [`@babel/preset-es2017`](https://github.com/babel/babel/tree/master/packages/babel-es2017) together (or the deprecated `babel-preset-lastest`).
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
WDYT about this wording? It might help people migrating from Babel 6 to Babel 7 which use the old preset (it still has ~100.000 downloads/week).

> Without any configuration options, `@babel/preset-env` behaves exactly the same as `@babel/preset-es2015`, `@babel/preset-es2016` and `@babel/preset-es2017` together (or the deprecated `babel-preset-lastest`).

Also, can you write the preset names in ``?

existentialism:
I've been in the midst of a big rewrite of the preset-env docs, but I think this could be reworded as:

> Without any configuration options, `@babel/preset-env` will transpile any ES2015+ code (similar to adding _all_ of the yearly presets (@babel/preset-es2015, etc.) individually, or the deprecated `babel-preset-latest`).

pravdomil:
I prefer something like

> The idea of `@babel/preset-env` is write JavaScript with all stage-4 features and earlier and transpile it so that it is an exact fit for your target platform(s).

taken from http://2ality.com/2017/06/pkg-esnext.html

pravdomil:
true, I would like to create a link in docs where developers can check what presets is being used but in https://github.com/pravdomil/babel/blob/patch-2/packages/babel-preset-env/package.json see no presets? I would expect all presets being listed as dependency?

nicolo-ribaudo:
The presets are listed in https://github.com/pravdomil/babel/blob/patch-2/package.json. Anyway, I don't think that it is needed since there already are examples.

pravdomil:
that are presets for all `babel` package I meant for `@babel/preset-env`

nicolo-ribaudo:
They are all compiled from the root folder.

pravdomil:
so is there list of presets that `babel-env` is using?

nicolo-ribaudo:
https://github.com/babel/babel/blob/d6085357194cf6e0a5025d9477a427027cae601a/babel.config.js#L54-L69

PR #7855, opened 2018-05-02
Lazy load tagged template literal strings

### LINKED ISSUE #7723: Lazy Init Tagged Template Literal Strings
Background
----------

The `strings` of a Tagged Template Literal (``` tagger`some${value}` ```) are cached, so that repeated invocations of the same Tagged Template Literal return the same `strings` instance:

```js
// Using tagger as a Tagged Template Literal will return the
// underlying Strings object that the JS language generates.
const tagger = (strings) => strings;

function text(value) {
  return tagger`some ${value} strings`;
}

const foo = text(); // => ["some ", " strings"]
const bar = text('hello'); // => ["some ", " strings"]

assert.ok(Array.isArray(foo));
assert.ok(Array.isArray(bar));

// Yes, even though bar was passed "hello".
// The expressions passed to the template literal (inside the `${}`)
// are not part of the Strings array. Only the literal strings are in it.
assert.equal(foo, bar);
```

The way Babel currently handles this Strings caching is by hoisting the array to the top of the file:

```js
// input
function text(value) {
  return tag`some ${value} strings`;
}

// output
var _templateObject = _helper_function_taggedTemplateLiteral(["some ", " strings"]);

function text(value) {
  // Notice, _templateObject.
  // That's the Strings
  return tag(_templateObject, value);
}
```

Notice how `_templateObject` is an array of strings that's hoisted to the top of the file? That means they're parsed and initialized long before they're actually used.

For very long strings, this can take up a lot of memory. And it's it's not used until much later in the application's lifecycle, it's just blocking initial renders.

Proposal
--------

We should make `_templateObject` (the Strings) lazy load:

```js
// input
function text(value) {
  return tag`some ${value} strings`;
}

// output
// No longer parsed or initialized
var _templateObject;

function text(value) {
  return tag(
    // Now that it's needed, we'll init
    // Next time it's called, it'll reuse the same Strings
    (_templateObject = _templateObject || _helper_function_taggedTemplateLiteral(["some ", " strings"])),
    value
  );
}

### PR DESCRIPTION
**Pull request by dczombera**
| Fixed Issues?            | Fixes #7723
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         
| Any Dependency Changes?  |

This change makes `_templateObject` lazy loaded. While working on it, I removed the pure annotation because I was not sure if it is still useful when used like this (please correct me if wrong). 
I was also not sure if it is recommended to clone `templateObject` every time it is used as an argument to create a new AST node. In this case, I used the existing code in the file as kind of a guidance and cloned the node each time. I appreciate every bit of feedback! :)

### DIFF
--- a/packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop-loose/imports-hoisting/output.js
+++ b/packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop-loose/imports-hoisting/output.js
@@ -6,4 +6,4 @@ var _taggedTemplateLiteral2 = _interopRequireDefault(require("@babel/runtime/hel
 
 var _templateObject;
 
-tag((_templateObject = _templateObject || (0, _taggedTemplateLiteral2.default)(["foo"])));
+tag(_templateObject = _templateObject || /*#__PURE__*/ (0, _taggedTemplateLiteral2.default)(["foo"]));
--- a/packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop-loose/imports-hoisting/output.js
+++ b/packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop-loose/imports-hoisting/output.js
@@ -6,4 +6,4 @@ var _taggedTemplateLiteral2 = _interopRequireDefault(require("@babel/runtime/hel
 
 var _templateObject;
 
-tag(_templateObject = _templateObject || /*#__PURE__*/ (0, _taggedTemplateLiteral2.default)(["foo"]));
+tag(_templateObject || (_templateObject = /*#__PURE__*/ (0, _taggedTemplateLiteral2.default)(["foo"])));
--- a/packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop-loose/imports-hoisting/output.js
+++ b/packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop-loose/imports-hoisting/output.js
@@ -4,6 +4,14 @@ var _interopRequireDefault = require("@babel/runtime/helpers/interopRequireDefau
 
 var _taggedTemplateLiteral2 = _interopRequireDefault(require("@babel/runtime/helpers/taggedTemplateLiteral"));
 
-var _templateObject;
+function _templateObject() {
+  const data = /*#__PURE__*/ (0, _taggedTemplateLiteral2.default)(["foo"]);
 
-tag(_templateObject || (_templateObject = /*#__PURE__*/ (0, _taggedTemplateLiteral2.default)(["foo"])));
+  _templateObject = function () {
+    return data;
+  };
+
+  return data;
+}
+
+tag(_templateObject());
--- a/packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop-loose/imports-hoisting/output.js
+++ b/packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop-loose/imports-hoisting/output.js
@@ -4,6 +4,6 @@ var _interopRequireDefault = require("@babel/runtime/helpers/interopRequireDefau
 
 var _taggedTemplateLiteral2 = _interopRequireDefault(require("@babel/runtime/helpers/taggedTemplateLiteral"));
 
-var _templateObject = /*#__PURE__*/ (0, _taggedTemplateLiteral2.default)(["foo"]);
+var _templateObject;
 
-tag(_templateObject);
+tag((_templateObject = _templateObject || (0, _taggedTemplateLiteral2.default)(["foo"])));
--- a/packages/babel-plugin-transform-template-literals/src/index.js
+++ b/packages/babel-plugin-transform-template-literals/src/index.js
@@ -90,22 +90,17 @@ export default declare((api, options) => {
         const lazyLoad = t.logicalExpression(
           "||",
           t.cloneNode(templateObject),
-          callExpression,
-        );
-
-        const conditionalAssignment = t.assignmentExpression(
-          "=",
-          t.cloneNode(templateObject),
-          lazyLoad,
+          t.assignmentExpression(
+            "=",
+            t.cloneNode(templateObject),
+            callExpression,
+          ),
         );
 
         scope.push({ id: templateObject });
 
         path.replaceWith(
-          t.callExpression(node.tag, [
-            conditionalAssignment,
-            ...quasi.expressions,
-          ]),
+          t.callExpression(node.tag, [lazyLoad, ...quasi.expressions]),
         );
       },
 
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: packages/babel-plugin-transform-template-literals/src/index.js [hunk 1]
```diff
@@ -83,19 +82,27 @@ export default declare((api, options) => {
...
+        const conditionalAssignment = t.assignmentExpression(
+          "=",
+          t.cloneNode(templateObject),
+          lazyLoad,
+        );
+
         path.replaceWith(
           t.callExpression(node.tag, [
-            t.cloneNode(templateObject),
+            t.parenthesizedExpression(conditionalAssignment),
```
nicolo-ribaudo: Why is this `ParenthesizedExpression` needed?

packages/babel-plugin-transform-template-literals/src/index.js
```diff
@@ -83,19 +82,27 @@ export default declare((api, options) => {
...
         }
 
-        const init = t.callExpression(helperId, callExpressionInput);
-        annotateAsPure(init);
-        init._compact = true;
         scope.push({
           id: templateObject,
-          init,
           // This ensures that we don't fail if not using function expression helpers
           _blockHoist: 1.9,
```
nicolo-ribaudo: I think that we can remove this "private" parameter now, because it was needed to initialize the template in the right place (while the `var templateObject;` statement can be anywhere).

packages/babel-plugin-transform-template-literals/src/index.js
dczombera: Actually, it is not.:D  I didn't realize that this is an assignment expression that returns a value. I will remove it. Thanks for the feedback. 👍

packages/babel-plugin-transform-template-literals/src/index.js
dczombera: I had no idea what this was good for that's why I left it. In this case I will remove it.

Do you know if there is anything written down that explains this kind of private parameters?

dczombera:
@nicolo-ribaudo All right, I added the discussed changes.

dczombera:
Thanks for your feedback!

packages/babel-plugin-transform-template-literals/src/index.js
```diff
@@ -83,19 +83,27 @@ export default declare((api, options) => {
...
+          "||",
+          t.cloneNode(templateObject),
+          callExpression,
+        );
+
+        const conditionalAssignment = t.assignmentExpression(
+          "=",
+          t.cloneNode(templateObject),
+          lazyLoad,
+        );
```
jridgewell: Nit: we can optimize a little bit more by doing the assignment inside the logical expression:

```js
tag(_templateObject || (_templateObject = _taggedTemplateLiteral(...)))

File: packages/babel-plugin-transform-template-literals/src/index.js [hunk 4]
```diff
@@ -83,19 +83,27 @@ export default declare((api, options) => {
...
-          id: templateObject,
-          init,
-          // This ensures that we don't fail if not using function expression helpers
-          _blockHoist: 1.9,
-        });
+        const callExpression = t.callExpression(helperId, callExpressionInput);
+        annotateAsPure(callExpression);
+        callExpression._compact = true;
+
+        const lazyLoad = t.logicalExpression(
```
jridgewell: I wonder if we'll hit function size deopts with this? Maybe a redefining function?

```js
function _templateObject() {
  const result = _taggedTemplateLiteral(...);
  _templateObject = function() { return result; }
  return result;
}
```

We use the same pattern for [lazy-load imports](https://github.com/babel/babel/pull/6952/files#diff-ebca4dc345d93ae3ea13a65000a5d5acR3). /cc @loganfsmyth

dczombera: Can you elaborate a bit more on this? I don't understand what you mean by "function size deopts". :sob:

jridgewell: > The redefining function for modules is needed because it is used in multiple places

I understand why we do it in imports. This is a different optimization using the same code pattern.

Old V8 (Crankshaft) has a character limit for inlining functions in JIT (even comments count towards this limit!):

```js
// I get called from a lot of places
function test() {
  // My function body is short, so I'll get inlined
  return tag(_templateObject, 1, 2, 3);
}

// But now, we lazy-load:
function test() {
  // We just inlined a really big template string expression,
  // so I cross the max_inlined_source_size limit
  // Guess this function won't be inlined anymore...
  return tag(_templateObject || (_templateObject = _taggedTemplateLiteral(['some really long string'])), 1, 2, 3);
}
```

Now, V8 switched to Turbofan in 5.9, so it's no big deal in recent releases. But, `Chrome < 59` and `Node < 8.3` have to worry about this. So, my idea is to reuse the same re-define function pattern to move the long template string out of the user's function, and get us back down to small function bodies.

dczombera: Thanks a lot for this explanation. 👍 I will work on implementing the redefining function over the weekend.

dczombera:
I finally got to add the redefining function pattern. Feedback welcome!:grinning:

packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop-loose/imports-hoisting/output.js
```diff
@@ -4,6 +4,14 @@ var _interopRequireDefault = require("@babel/runtime/helpers/interopRequireDefau
 
 var _taggedTemplateLiteral2 = _interopRequireDefault(require("@babel/runtime/helpers/taggedTemplateLiteral"));
 
-var _templateObject = /*#__PURE__*/ (0, _taggedTemplateLiteral2.default)(["foo"]);
+function _templateObject() {
+  const data = /*#__PURE__*/ (0, _taggedTemplateLiteral2.default)(["foo"]);
```
jridgewell: Pure shouldn't be necessary anymore, since the wrapper itself is statically pure.

packages/babel-plugin-transform-template-literals/src/index.js
```diff
@@ -83,19 +83,22 @@ export default declare((api, options) => {
...
 
+        const lazyLoad = template.ast`
+          function ${templateObject}() {
+            const data = ${callExpression};
+            ${templateObject} = function() { return data };
+            return data;
+          } 
+        `;
+
+        scope.path.unshiftContainer("body", lazyLoad);
```
jridgewell: Actually, do we want to push into the scope and using hoisting?

loganfsmyth: I think unshift is fine, since we're usually pretty consistent about helpers being at the top of each file, and this is helper-like.

jridgewell:
We can remove the `PURE` comment in a followup.

PR #7857, opened 2018-05-02
Fix: README.md sponsor icons [skip ci]

ZeroCho:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      |
| Documentation PR         | Yes
| Any Dependency Changes?  |

I found out that some icons of sponsors don't match real icons of sponsors.
They should be treated better! 😄 
I think it's due to cache of opencollective.com. I'm not sure this is the best way to reload svg icons, but still it works.


ZeroCho:
@xtuc It's strange. Yesterday, some icons were wrong, but today everything is fixed! haha

PR #7869, opened 2018-05-04
Add an option to Babylon to have decorators before export

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR         | 
| Any Dependency Changes?  |

- 8f76ff3 adds support for options to Babylon plugins.

- 3c5c47f adds support for decorators before export (via `["decorators2", { "decoratorsBeforeExport": true }]`). I think that this should be an option because the discussion about it is still open; this would allow us to experiment both syntaxes. We can discuss about what the default value is.

### DIFF
--- a/packages/babylon/src/index.js
+++ b/packages/babylon/src/index.js
@@ -71,7 +71,7 @@ function getParserClass(
     hasPlugin(pluginsFromOptions, "decorators-legacy")
   ) {
     throw new Error(
-      "Cannot use decorators and decorators-legacy plugin together",
+      "Cannot use the decorators and decorators-legacy plugin together",
     );
   }
 
--- a/packages/babylon/src/index.js
+++ b/packages/babylon/src/index.js
@@ -67,39 +67,39 @@ function getParserClass(
   pluginsFromOptions: $ReadOnlyArray<string>,
 ): Class<Parser> {
   if (
-    pluginsFromOptions.indexOf("decorators-legacy") >= 0 &&
-    pluginsFromOptions.indexOf("decorators") >= 0
+    hasPlugin(pluginsFromOptions, "decorators") &&
+    hasPlugin(pluginsFromOptions, "decorators-legacy")
   ) {
-    throw new Error("Cannot use decorators and decorators2 plugin together");
+    throw new Error(
+      "Cannot use decorators and decorators-legacy plugin together",
+    );
   }
 
   // Filter out just the plugins that have an actual mixin associated with them.
-  let pluginList = pluginsFromOptions.filter(
-    p => p === "estree" || p === "flow" || p === "jsx" || p === "typescript",
-  );
+  let pluginList = pluginsFromOptions.filter(plugin => {
+    const p = getPluginName(plugin);
+    return p === "estree" || p === "flow" || p === "jsx" || p === "typescript";
+  });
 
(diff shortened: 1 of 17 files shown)


nicolo-ribaudo:
I think that after this PR I will need to open only one more. If we can finish the work on decorators before the TC39 meeting it would be super!

packages/babylon/src/index.js
```diff
@@ -67,39 +67,39 @@ function getParserClass(
...
 ): Class<Parser> {
   if (
-    pluginsFromOptions.indexOf("decorators-legacy") >= 0 &&
-    pluginsFromOptions.indexOf("decorators") >= 0
+    hasPlugin(pluginsFromOptions, "decorators") &&
+    hasPlugin(pluginsFromOptions, "decorators-legacy")
   ) {
-    throw new Error("Cannot use decorators and decorators2 plugin together");
+    throw new Error(
+      "Cannot use decorators and decorators-legacy plugin together",
```
existentialism: Nit:

> Cannot use the `decorators` and `decorators-legacy` plugins together.

packages/babylon/src/index.js
```diff
@@ -67,39 +67,39 @@ function getParserClass(
...
+      "Cannot use the decorators and decorators-legacy plugin together",
+    );
   }
 
   // Filter out just the plugins that have an actual mixin associated with them.
-  let pluginList = pluginsFromOptions.filter(
-    p => p === "estree" || p === "flow" || p === "jsx" || p === "typescript",
-  );
+  let pluginList = pluginsFromOptions.filter(plugin => {
+    const p = getPluginName(plugin);
```
loganfsmyth: Hey all, I'm concerned about this. Now that we've defined the ability to pass options to Babylon plugins, ordering becomes critical. What if some other piece of tooling injected a syntax plugin into the plugin list after this one, or before? Which one should take priority? For instance for https://github.com/babel/babel/pull/7934/files#diff-3a8233bcd2766d2c7d87f23f944f7726R12, say the user's framework auto-enables `flow`, but the user wants to enable `all`, which comes first?

So far we've avoided this problem by not having options, and not worrying about duplicates. Now it will matter very much both what order your Babel plugins are in, but also what order they enable Babylon plugins in.

At a minimum, have we verified that the ordering on these is consistent with the ordering of visitors? e.g. If two plugins are enabled with different options, whichever one's visitors run first, is also the one who's Babylon configuration takes affect?

packages/babylon/src/parser/statement.js
```diff
@@ -1422,6 +1425,12 @@ export default class StatementParser extends ExpressionParser {
     } else if (this.match(tt._class)) {
       return this.parseClass(expr, true, true);
     } else if (this.match(tt.at)) {
+      if (
+        this.hasPlugin("decorators") &&
+        this.getPluginOption("decorators", "decoratorsBeforeExport")
+      ) {
+        this.unexpected();
```
loganfsmyth: It would be good to have these throw more helpful errors about what the user is supposed to do to get their code to parse.

packages/babylon/src/index.js
loganfsmyth: Let's continue the discussion over in https://github.com/babel/babel/issues/7984 instead of here.

PR #7878, opened 2018-05-07
Fix handling of different JSX pragmas in Typescript

### LINKED ISSUE #7631: [Typescript] es6 import of JSX pragma is being removed
Bug Report

### Input Code

```js
import { h } from 'preact'
import Button from '.'

const foo = <Button type="submit" />
```

```js
module.exports = {
  plugins: [
    '@babel/plugin-transform-modules-commonjs',
    ['@babel/plugin-transform-react-jsx', { pragma: 'h' }],
    '@babel/plugin-transform-typescript'
  ]
}
```

### Expected Behavior
I expect Babel to transform the es6 import of 'preact' into a require() call

### Current Behavior
Right now Babel removes the preact import because it sees it as unused, even though it is the jsx pragma. When I remove the typescript plugin, it works fine, except then it is unable to parse/transform typescript.

Here is what it does right now:
```js
"use strict";

var _ = _interopRequireDefault(require("."));

const foo = h(_.default, {
  type: "submit"
});
```
As you can see, it removes the import of {h} from preact, instead of transforming it to a require

### Possible Solution
I'm not 100% sure about this, but it seems like the typescript plugin is doing some react-specific things for jsx, which makes it not work with preact and other jsx pragmas.
https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-typescript/src/index.js#L269

### Context
I want to use these things together:
- Typescript
- JSX
- ES6 imports
- Preact

### Your Environment

| ---------------- | -------
| Babel            | ^7.0.0-beta.42
| node             | v9.9.0
| npm              | 5.7.1
| Operating System | Arch Linux

calebeby:
| Fixed Issues?            | Fixes #7631

Before, if a JSX pragma was imported in typescript, the import was removed because it did not think it was used as a value. This changes it so it knows that the import is being used as a JSX pragma


calebeby:
Yeah, like @hzoo mentioned in #7631, we could figure out some way for the ts plugin to read the config from the jsx plugin, but at this point, I think that adding an option to the ts plugin is the best solution, since babel doesn't have anything in place afaik to retrieve configuration from other plugins.

@hzoo maybe in ~a month when I'm out of school I can help with making a ts preset that has a single jsx pragma option that it passes to both the ts plugin and the jsx plugin

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -266,11 +266,14 @@ export default declare(api => {
       }
     }
 
-    if (binding.identifier.name != "React") {
+    if (
+      binding.identifier.name !== "React" &&
+      binding.identifier.name !== options.jsxPragma
```
xtuc: what about using "React" as default in `options.jsxPragam`, that would simplify the logic a bit.

calebeby: ✔️

xtuc:
That's a good point, how could we keep both in sync?

At the moment it's an option but that's probably not the easiest way to configure it. I mean the pragma will be better.

weswigham:
IMO, extract inline pragma detection out of the `jsx` plugin, then call it in both plugins.

calebeby:
@weswigham can you elaborate a bit? I can understand extracting the logic that finds the jsx pragma from the @jsx comment, but do you think we can somehow extract the option for the pragma? How?

calebeby:
Also, how does Babel plugin ordering work? I really don't understand this, because if we could just get the jsx plugin to run before this, we wouldn't have to deal with this.

calebeby:
Also we should deal with fragments 😞

calebeby:
```
Import referenced as value -> keep
Import is jsx pragma by options or by @jsx comment -> keep
Import is jsx fragment pragma by options or by @jsxFragment comment -> keep
Otherwise, remove the import
```

I don't think it is necessary to check if the pragma is used, because if an import is the jsx pragma, it must not be a type, it is a value.

I think we need to come up with a way for the typescript plugin to talk to the jsx plugin, and it should ask what the jsx pragma and jsx fragment pragma are.

Is there a way we can attach additional data along with the ast, which the jsx plugin can write to and the ts plugin can read from?

hzoo:
We can just land this as is since it fixes the bug

> IMO, extract inline pragma detection out of the jsx plugin, then call it in both plugins.

But yeah we should probably do this or figure out a better fix

PR #7906, opened 2018-05-11
Update @babel/plugin-codemod-object-assign-to-object-spread version

nicolo-ribaudo:
Lerna didn't update it?


nicolo-ribaudo:
I'm waiting for Henry or someone else which is familiar with the release process to merge this, since I don't want to break things stuck_out_tongue

hzoo:
That is weird.. it's in the config https://github.com/babel/babel/blob/master/lerna.json

PR #7909, opened 2018-05-11
Inherit parens in replaceWith

arv:
If the original expression had parens then add parens around the new
expression too,

Based on  #7015

Co-authored-by: Toru Kobayashi <koba0004@gmail.com>

| Fixed Issues?            | 
| Patch: Bug Fix?          | 
| Major: Breaking Change?  | 
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-traverse/src/path/replacement.js
+++ b/packages/babel-traverse/src/path/replacement.js
@@ -166,6 +166,7 @@ export function replaceWith(replacement) {
   if (oldNode) {
     t.inheritsComments(replacement, oldNode);
     t.removeComments(oldNode);
+    t.inheritParens(replacement, oldNode);
   }
 
   // replace the node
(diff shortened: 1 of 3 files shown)


arv:
@hzoo I'm not really sure if there is any action I need to take here? I looked at inheritComments/inherits and there is not enough overlap there.

loganfsmyth:
My concern with this approach is that `.extra` isn't a field most people know about, and could easily be lost in any number of transforms, especially in cases where we create replacement nodes from scratch. If our goal is the make things work consistently for ClosureCompiler, it seems like using a full AST node would be the safest.

arv:
@loganfsmyth I have another branch with a ParenTree AST node but that seems much more disruptive. It means that the AST now may change when exiting code did not expect it.

```js
(function() { ... })();
```

Is now

- Program
  - Script
    - ExpressionStatement
      - ParenExpression
        - FunctionExpression
  
when it used to be

- Program
  - Script
    - ExpressionStatement
       - FunctionExpression

I'm concerned that it will break things

arv:
Ok. I'll close this PR and work on cleaning up my other branch.

PR #7913, opened 2018-05-12
Update README.md

houshuang:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Minor: New Feature?      |No
| Tests Added + Pass?      |No
| Documentation PR         |Yes

[skip ci]

Because of the breaking change in 7-beta.46 regarding to monorepos and .babelrc, I needed to switch to babel.config.js, and somehow it could not find it unless I specified its path on the command line. 

I spent more than 45 minutes not getting it to work and even digging into babel core code, because I followed instructions on the babel-core README exactly, meaning I provided the parameter --configFile==... I realize I could have written babel --help and seen the correct parameter, but by choosing a parameter that is multi-word for the example, we show people right away that they need to switch from camelCase to kebab-case when supplying parameters on the command line.

### DIFF
--- a/packages/babel-core/README.md
+++ b/packages/babel-core/README.md
@@ -205,7 +205,7 @@ Each `ConfigItem` exposes all of the information Babel knows. The fields are:
     You can pass these options from the Babel CLI like so:
   </p>
   <p>
-    <code>babel --name<span class="o">=</span>value</code>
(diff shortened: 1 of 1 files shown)


houshuang:
Done

PR #7928, opened 2018-05-14
Create InterpreterDirective AST node type and use to replace babel/core File's 'shebang' handling

### LINKED ISSUE #6586: Generated sourcemaps don't include shebangs
### Input Code

```js
#!/usr/bin/env node

console.log("Hello, world!");
```

```sh
$ babel --source-maps true --out-file shebang.compiled.js shebang.js
```

### Expected Behavior

Generated sourcemaps should include the shebang in the `mappings` and `sourcesContent` sections, but they do not.

### Current Behavior

Generated sourcemaps do not include the shebang. This can be seen most obviously in this scenario;

###### shebang.js

```js
#!/usr/bin/env node

console.log("Hello, world!");
```

###### no-shebang.js

```js
// remove this comment

console.log("Hello, world!");
```

> I've had to add a comment to this one to prevent GitHub from eating the whitespace - it doesn't exist in the actual code.

```sh
$ babel --source-maps true --out-file shebang.compiled.js shebang.js
$ babel --source-maps true --out-file no-shebang.compiled.js no-shebang.js
# Using sed to make the filenames the same.
$ diff -u <(cat shebang.compiled.js.map | sed 's/shebang/file/g') <(cat no-shebang.compiled.js.map | sed 's/no-shebang/file/g')
```

### Possible Solution

Take shebangs into account when generating sourcemaps.

### Your Environment

| ---------------- | -------
| Babel            | `6.26.0 (babel-core 6.26.0)`
| node             | `v8.4.0`
| npm              | `5.5.1`
| Operating System | `Ubuntu 17.04`

### LINKED ISSUE #7905: retainLines does not work with shebang
## Bug Report

**Current Behavior**
A line numbers retention does not work properly with a shebang in a JS file.

**Input Code**

The following code with `shebang`

```js
#!/usr/bin/env node
import { readdirSync, lstatSync, watchFile, unwatchFile } from 'fs'
import { join, resolve } from 'path'
```

will have incorrect lines

```js
#!/usr/bin/env node
"use strict";
var _fs = require("fs");
```

The code without `shebang` is fine

```js
import { readdirSync, lstatSync, watchFile, unwatchFile } from 'fs'
import { join, resolve } from 'path'
```

```js
"use strict";var _fs = require("fs");
var _path = require("path");
```

**Expected behavior/code**

Produces code should respect `shebang` comment.

```js
#!/usr/bin/env node
"use strict"; var _fs = require("fs");
var _path = require("path");
```

```json
{
  "plugins": [
    "@babel/plugin-syntax-object-rest-spread",
    "@babel/plugin-transform-modules-commonjs"
  ],
  "retainLines": true
}
```

**Environment**
- Babel version(s): 7.0.0-beta.46
- OS: OSX 10.13.1
- Monorepo no
- How you are using Babel: cli, register

**Possible Solution**

I need to debug the code using VS Code however if lines are not retained it's impossible.

loganfsmyth:
| Fixed Issues?            | Fixes #6586, Fixes #7905 
| Documentation PR         | 
| Any Dependency Changes?  |

Both of these bugs were essentially because of the way we removed the shebang up front, meaning it was totally separated from the normal babel-generator codegen process.

This is another divergence from ESTree, but we've also got https://github.com/bmeck/proposal-hashbang as an active proposal anyway, so it seems like adding an AST node for this isn't too unreasonable.

I was considering adding a Babylon plugin and auto-enabling it in babel-core, but then I realized Babylon was _already_ parsing shebangs and sticking it in a comment, which I think is even weirder. So this essentially means that any third-parties using Babylon directly and not through Babel will see changed behavior, but no-one using Babel.


packages/babel-types/src/definitions/core.js
```diff
@@ -525,8 +534,8 @@ defineType("MemberExpression", {
 defineType("NewExpression", { inherits: "CallExpression" });
 
 defineType("Program", {
-  visitor: ["directives", "body"],
-  builder: ["body", "directives", "sourceType"],
+  visitor: ["interpreter", "directives", "body"],
```
loganfsmyth: One comment here from Slack is we might not want to visit it, since it is comment-like and we don't visit other comment types. I don't feel strongly since it is trivial to reach from `program`

packages/babylon/src/tokenizer/index.js
```diff
@@ -437,6 +437,32 @@ export default class Tokenizer extends LocationParser {
...
+
+    while (
+      ch !== charCodes.lineFeed &&
+      ch !== charCodes.carriageReturn &&
+      ch !== charCodes.lineSeparator &&
+      ch !== charCodes.paragraphSeparator &&
+      ++this.state.pos < this.input.length
+    ) {
+      ch = this.input.charCodeAt(this.state.pos);
+    }
```
nicolo-ribaudo: Since this code is equal to the one used to parse single-line comments, maybe there should be a `skipLine` method?

loganfsmyth: We could, but I also don't feel that strongly about factoring out every piece of sharable logic.

xtuc:
I think the `interpreter` key should be optional in the AST. It doesn't really make sense for scripts/modules meant for the web?

I like the way you introduced that because we could have an option/plugin that automatically adds a shebang to the generated file.

PR #7934, opened 2018-05-15
Add support for explicit type arguments in new and call expressions

samwgoldman:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Yes†
| Documentation PR         | 
| Any Dependency Changes?  |

Flow 0.72 added support for passing explicit type arguments to polymorphic functions and constructors. Parsing support was added in https://github.com/facebook/flow/commit/40fbcdd77c0c1d0d7ef169535499020f70335624, for reference.

Note that this is a breaking change, because some valid programs will now have different parse trees. For example, the program `f<T>(e)` previously parsed as a nested binary expression, `((f < T) > e)`.

Since this is a Flow-only feature, the parsing rules are conditional. We will only attempt to parse type arguments if either (a) the `all` option, added in 1c359ad9fcf84f7db98232fe0f0f2a3b31db0947 is provided to babel-plugin-syntax-flow or (b) the file starts with a docblock comment containing the `@flow` or `@noflow` pragma.

† While this is a breaking change, I have taken pains to ensure this only affects Flow users, who are already affected by the change if they update Flow. Based on this, I am hoping that it's OK to push this through without a major version bump.


samwgoldman:
cc @babel/flow

nicolo-ribaudo:
Oh right, it's ok to have a single PR then.

Also, I will soon merge https://github.com/babel/babel/pull/7869, which adds support for plugin options to Babylon. Can you then use `plugins: [ ["flow", { all: true }] ]` instead of `plugins: [ "flow" ], flowAll: true`?

samwgoldman:
> I will soon merge #7869, which adds support for plugin options to Babylon

That sounds great! Will happily update this when that lands.

existentialism:
@samwgoldman just landed #7869!

packages/babel-generator/test/fixtures/flow/typeapp-call/input.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+f<T>();
+f<T><U></U>;
```
nicolo-ribaudo: This looks orrible... What does it even mean?

nicolo-ribaudo: Oh nvm, I got it

packages/babylon/src/plugins/flow.js
```diff
@@ -2340,6 +2378,103 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+          return this.finishNode(node, "OptionalCallExpression");
+        } catch (e) {
+          if (e instanceof SyntaxError) {
+            this.state = state;
+            const node: N.OptionalMemberExpression = this.startNodeAt(
+              startPos,
+              startLoc,
+            );
+            node.object = base;
+            node.property = this.parseIdentifier(true);
```
nicolo-ribaudo: We already know that after `?.` there is `<` because we lookead ahead... When can this be an identifier?

samwgoldman: You're right. The idea isn't to successfully parse an identifier, but to error as if an identifier were expected if we failed to parse as a call with type args. This is probably overcomplicated, since we could also just call `this.unexpected`. Alternatively, since `o.?<` is not ambiguous, we could avoid the `state.clone` and fallback code entirely. I slightly prefer the latter course. What do you think?

samwgoldman: OK, pushed an update with this change.

samwgoldman:
Thanks for the reviews so far! Is there anything I can do to move this along?

loganfsmyth:
@samwgoldman It looks like this PR introduced a few Flow errors in the parser codebase, would you be able to look into those at some point?

packages/babel-plugin-syntax-flow/src/index.js
```diff
@@ -1,11 +1,15 @@
 import { declare } from "@babel/helper-plugin-utils";
 
-export default declare(api => {
+export default declare((api, options) => {
   api.assertVersion(7);
 
+  // When enabled and plugins includes flow, all files should be parsed as if
+  // the @flow pragma was provided.
+  const { all } = options;
```
loganfsmyth: @samwgoldman In our Babel team meeting today, we talked about how it would be nice to set this automatically based on the user's `.flowconfig` to avoid having to duplicate the value in two places. I have a few questions related to that:

1. Does flow allow multiple `.flowconfig` files, or can we just read the one in the project root? And along those lines, do you expect published npm libraries to then include a `.flowconfig` if they are using `all`, since otherwise it would publish without the flow pragma comments?
2. Is there a JS library we can use for parsing Flow's config?

existentialism:
@loganfsmyth @samwgoldman I had already started working on fixing the flow errors, pushed up https://github.com/babel/babel/pull/8002!

nmote:
@samwgoldman is on vacation so I don't think he'll see this for a little while. @existentialism, thanks for stepping in with the fix!

packages/babel-plugin-syntax-flow/src/index.js
loganfsmyth: Alright, if there's no parser then we don't really have any options for now, but it does seem unfortunate that Babel needs to be told separately what type of file it is parsing.

nmote: It's a pretty simple format. If you just want to check for the `all` option I imagine it wouldn't be hard to reimplement. I would think that the other problems are bigger roadblocks.

PR #7937, opened 2018-05-15
Rename "Babylon" to "@babel/parser"

### LINKED ISSUE #7901: Rename "Babylon" to "@babel/parser"
## Feature Request

It's cool to have fun names, but in the end new people don't know what they do, and if even if you are new you get confused if you aren't in the weeds doing all this. I suggested we do this a while ago and just didn't get around to doing it yet.

image

I made a poll and it's 2/3 for yes: https://twitter.com/left_pad/status/959918310451277831

> I didn't know babylon is the parser until now (sorry never check it before) 😅
> The name "parser" tells us what it does. IMO it will be helpful for novice

The name says what it does so it should be better in the long run.

Docs, talks, etc. Still not enough. Same problem as `babili` -> `babel-minify` which I changed earlier.

This is just annoying because all of our docs and other people's docs/articles are going to change.
- We should probably mention the name change in the readme and other places, not just switch it.
- And also PRs are going to have merge conflicts as well. I'm not sure if we want to just make a new folder and have both to land some PRs or just simply rename it? Maybe I'm thinking too much about it.
- This could be help-wanted but might be easier for someone more attuned to the codebase to do it (or maybe just review). I just know it will need to be rebased a lot unless we review it asap, depends.

kamthamc:
| Fixed Issues?            | Fixes #7901
| Tests Added + Pass?      | No + Yes
| Documentation PR         | 

Renamed all `babylon` references to 'babel-parser', updated package name to `@babel/parser` from `babylon`


packages/babel-core/README.md
```diff
@@ -99,7 +99,7 @@ Given an [AST](https://astexplorer.net/), transform it.
 
 ```js
 const sourceCode = "if (true) return;";
-const parsedAst = babylon.parse(sourceCode, { allowReturnOutsideFunction: true });
+const parsedAst = babelParser.parse(sourceCode, { allowReturnOutsideFunction: true });
```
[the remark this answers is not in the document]
kamthamc: if i use `babel.parser`, it might confusing as i might think babel has a method parser??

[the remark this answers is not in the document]
loganfsmyth: Indeed it does now! It does more than `babylon` does, but it's probably more what user's would want in the context of these examples anyway, so I think the examples should use it instead.

kamthamc:
@hzoo @loganfsmyth addressed the comments, if someone can help me update the image for babylon it would be great

packages/babel-parser/CHANGELOG.md
```diff
@@ -57,7 +57,7 @@ See the [Babel Changelog](https://github.com/babel/babel/blob/master/CHANGELOG.m
 ## 7.0.0-beta.7 (2017-03-22)
 
 ### Spec Compliancy
-* Remove babylon plugin for template revision since it's stage-4 (#426) (Henry Zhu)
+* Remove babel-parser plugin for template revision since it's stage-4 (#426) (Henry Zhu)
```
loganfsmyth: I'm sorry, I think we need to clarify there this terminology is important. The only places we should be using `babel-parser` instead of `@babel/parser` are when we're talking about the directory paths in the repository, I'd expect the vast majority of the time, references to `babylon` would be talking about the package, not about the monorepo folder, so the majority of references should be `@babel/parser`.

kamthamc:
@loganfsmyth updated it.

packages/babel-parser/CHANGELOG.md
kamthamc: yes you are right i missed it i guess

packages/babel-parser/README.md
```diff
@@ -18,9 +18,9 @@ thanks to the awesome work of [@RReverser](https://github.com/RReverser) and [@m
 
 ## API
 
-### `babylon.parse(code, [options])`
+### `babel.parse(code, [options])`
```
loganfsmyth: Since these references are in the Babylon readme itself, this documentation is specifically about the `.parse` function on the parser package, not the `.parse` function on the `babel-core` package. In this case we should document these as `babelParser.parse`.

kamthamc:
@loganfsmyth you right but how i can I change without publishing it?

loganfsmyth:
@kamthamc If you rename it in the `package.json`, and re-run `make bootstrap`, it will be symlinked to the package in the monorepo.

packages/babel-template/src/parse.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 import * as t from "@babel/types";
 import type { TraversalAncestors, TraversalHandler } from "@babel/types";
-import { parse } from "babylon";
+import { parse } from "@babel/parser";
```
loganfsmyth: Looks like the `package.json` for this one was missed.

packages/babel-traverse/src/path/replacement.js
```diff
@@ -3,7 +3,7 @@
 import { codeFrameColumns } from "@babel/code-frame";
 import traverse from "../index";
 import NodePath from "./index";
-import { parse } from "babylon";
+import { parse } from "@babel/parser";
```
loganfsmyth: This one also needs a `package.json` update.

packages/babel-types/src/validators/generated/index.js
```diff
@@ -1387,6 +1387,23 @@ export function isInterfaceDeclaration(node: Object, opts?: Object): boolean {
 
   return false;
 }
+export function isInterfaceTypeAnnotation(
```
loganfsmyth: FYI, these changes landed on `master` so if you want to rebase they'll be gone.

package.json
```diff
@@ -22,7 +22,7 @@
     "babel-jest": "^22.4.1",
     "babel-loader": "8.0.0-beta.0",
     "babel-plugin-transform-charcodes": "^0.1.0",
-    "babylon": "7.0.0-beta.46",
+    "@babel/parser": "7.0.0-beta.46",
```
loganfsmyth: We shouldn't be changing the root monorepo packages. There is no `beta.46` of `@babel/parser`.

jridgewell:
We need to be very careful not to lose an changes done to `packages/babylon` that were merged after this PR was created.

hzoo:
Ok just need to rebase and merge this before anything else?

kamthamc:
the re are no changes to rebase. `everything is up to date`

hzoo:
Thanks @kamthamc 👌, I know this was a big change (also hard to review 😂)

kamthamc:
👍 Thank you all for taking time to review.

PR #7944, opened 2018-05-16
plugin-transform-typescript: Extend caveats with known issues.

apepper:
This PR add two known issues of @babel/plugin-transform-typescript to the caveats list.

| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      |
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-transform-typescript/README.md
+++ b/packages/babel-plugin-transform-typescript/README.md
@@ -10,8 +10,8 @@ Does not type-check its input. For that, you will need to install and set up Typ
 * Does not support [`const enum`][const_enum]s because those require type information to compile.
 **Workaround**: Remove the `const`, which makes it available at runtime.
 * Does not support [`export =`][exin] and [`import =`][exin], because those cannot be compile to ES.next. **Workaround**: Convert to using `export default` and `export const`, and `import x, {y} from "z"`.
-* Does not (yet?) support binding patterns in signature. So `export type T = ({x, y}: any) => void;` does currently not work. See [babel issue #6667](https://github.com/babel/babel/issues/6667) and [babylon issue #768](https://github.com/babel/babylon/issues/768) for details.
-* Does not (yet?) support default export of an interface. So `export default interface A {}` does currently not work. See [babel issue #7118](https://github.com/babel/babel/issues/7118) and [babel issue #7640](https://github.com/babel/babel/issues/7640) for details.
+* Does not support binding patterns in signature. So `export type T = ({x, y}: any) => void;` does currently not work. See [babel issue #6667](https://github.com/babel/babel/issues/6667) and [babylon issue #768](https://github.com/babel/babylon/issues/768) for details.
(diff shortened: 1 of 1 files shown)


packages/babel-plugin-transform-typescript/README.md
```diff
@@ -10,6 +10,8 @@ Does not type-check its input. For that, you will need to install and set up Typ
 * Does not support [`const enum`][const_enum]s because those require type information to compile.
 **Workaround**: Remove the `const`, which makes it available at runtime.
 * Does not support [`export =`][exin] and [`import =`][exin], because those cannot be compile to ES.next. **Workaround**: Convert to using `export default` and `export const`, and `import x, {y} from "z"`.
+* Does not (yet?) support binding patterns in signature. So `export type T = ({x, y}: any) => void;` does currently not work. See [babel issue #6667](https://github.com/babel/babel/issues/6667) and [babylon issue #768](https://github.com/babel/babylon/issues/768) for details.
+* Does not (yet?) support default export of an interface. So `export default interface A {}` does currently not work. See [babel issue #7118](https://github.com/babel/babel/issues/7118) and [babel issue #7640](https://github.com/babel/babel/issues/7640) for details.
```
xtuc: I think we can remove `(yet)` here, it seems clear to me

apepper:
REVIEW (f93eec0) /cc @xtuc 

* Removed the ` (yet?)` part from the README.

xtuc:
cc @babel/typescript

hzoo:
I don't really see us writing down every thing that isn't supported in the readme like this? I would just think to the typescript label, lest we get another PR that adds all the rest? We added the caveats as big features that won't be supported where as the rest are just bugs if that makes sense?

apepper:
> Thanks for the PR! Not opposed entirely but I don't really see us writing down every thing that isn't supported in the readme like this? I would just think to the typescript label, lest we get another PR that adds all the rest? We added the caveats as big features that won't be supported where as the rest are just bugs if that makes sense?

In an ideal world, where these bugs would be fixed in a timely manner, I would agree. But since both bugs are now open for over half a year I think it is best to give people as much information as possible up front (in my view "fail early").

In my specific case I was investigating, who to replace `awesome-typescript-loader` with babel. I stumbled upon these issues during that investigation and had to dig into the issues to find them. So at least myself a couple of weeks before this PR would have appreciated the information provided by the PR.

And one more word about context: Thank you so much for all the work had has been put into babel and into this plugin. It is much appreciated 👏 !

xtuc:
I've changed my mind on this, we recently merged https://github.com/babel/babel/pull/8108 which adds a link to issues related to the preset. Issues are maintenable (and actually fixable).

loganfsmyth:
Since the docs have moved to the website repo, I'm going to close this.

These docs moved to https://github.com/babel/website/blob/master/docs/plugin-transform-typescript.md and it also looks like we need to fix the link from preset -> plugin in https://github.com/babel/website/blob/master/docs/preset-typescript.md

PR #7946, opened 2018-05-16
update core-js in babel-polyfill to 2.5.6

kytart:
Update core-js in babel-polyfill from 2.5.3 to 2.5.6 which has a few bug fixes, that I need in my project.


existentialism:
Should we make babel-runtime match?

kytart:
@nicolo-ribaudo sure I can, but in babel-runtime the jump is pretty big, because the current version there is 2.4.0. Since I'm not using babel-runtime, someone else who has experience with it should make the update and test it. I will update babel-register, since the current version there is the same as it was here.

PR #7947, opened 2018-05-16
Internal slot properties

samwgoldman:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Yes†‡
| Documentation PR         | 
| Any Dependency Changes?  |

Flow 0.72 added support for so-called "internal slot" properties. Parsing support was added in https://github.com/facebook/flow/commit/732eae55e102cdb7dfa7b6a85f0147d48c3afed7, for reference and more information about the feature itself.

† This is a minor breaking change in the parser itself. Flow also has an indexer property, which looks like `{ [name: K]: V }`, where `K` can be any type. The name is actually optional, so `{ [K]: V }` is also valid. Currently, `{ [[foo]]: V }` is parsed as an indexer property where `K = [foo]`, i.e., a tuple type of a single value with the type `foo`. It's very unlikely that a tuple type would be used in this position, and the change can be worked around by adding an explicit name or using a type alias. Because this change only affects Flow users, who are already affected by this breaking change if they update to Flow 0.72, I am hoping that it's OK to push this through without a major version bump.

‡ This is also a breaking change to babel-types, because it adds an additional parameter to the `t.objectTypeAnnotation` builder function. I added the new parameter before the existing `exact` parameter, which might cause breaking behavior. I'm happy to move the new parameter to the end of the list, but it seemed messy to me. Let me know what you prefer.


samwgoldman:
cc @babel/flow

packages/babel-generator/test/fixtures/flow/internal-slot/output.js
```diff
@@ -0,0 +1,12 @@
+declare class C { static [[foo]]: T
```
nicolo-ribaudo: Nit: Can this go on a new line?

samwgoldman: Sure! The logic around inserting newlines wasn't immediately clear to me so I glossed over that part. Happy to make this change.

PR #7949, opened 2018-05-16
update core-js to 2.5.6 in babel-register

kytart:
Update core-js to 2.5.6 in babel-register as well, the same as #7946.

### DIFF
--- a/packages/babel-register/package.json
+++ b/packages/babel-register/package.json
@@ -10,7 +10,7 @@
   "dependencies": {
     "babel-core": "^6.26.0",
     "babel-runtime": "^6.26.0",
-    "core-js": "^2.5.3",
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
https://github.com/babel/babel/pull/7946#issuecomment-389554495

> @nicolo-ribaudo sure I can, but in babel-runtime the jump is pretty big, because the current version there is 2.4.0. Since I'm not using babel-runtime, someone else who has experience with it should make the update and test it. I will update babel-register, since the current version there is the same as it was here.

Isn't it 2.5.3? https://github.com/babel/babel/blob/e45d5c3b6524a8842ac44f60f86b2d204c74e2f2/packages/babel-runtime/package.json#L9

PR #7955, opened 2018-05-17
Verify that files are .ts/.tsx before treating as Typescript files.

loganfsmyth:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

I think this is mergeable, but I'm also very interested to hear what people would think of this approach.

This PR makes means:

* For `preset-typescript`, it needs to be used in a context where Babel is given a filename, so it will error if called without a filename.
  * BUT if there is no filename, the user could set `allExtensions: true` to force-enable TS parsing without extensions taken into account. Then they could set `isTSX: true` to parse with JSX.
* For `plugin-typescript`, when enabled things will automatically be parsed as `.ts`, with the `isTSX: true` option to enable parsing as `.tsx`.

This essentially means that the preset tries to auto-configure the plugin based on the file path.

We could also consider making the preset always expect a filename (delete `allExtensions`), and just tell people to use the plugin if there is no filename.


packages/babel-preset-typescript/src/index.js
```diff
@@ -1,10 +1,42 @@
...
 
-  return {
-    plugins: [[transformTypeScript, { jsxPragma }]],
-  };
-});
+    if (typeof allExtensions !== "boolean") {
+      throw new Error(".allExtensions must be a boolean, or undefined");
+    }
+    if (typeof isTSX !== "boolean") {
+      throw new Error(".allExtensions must be a boolean, or undefined");
```
Kovensky: typo 🙊

xtuc: Has been changed https://github.com/babel/babel/commit/06e5c46c8df784ab8f6e7713ada68bbb011dbb68

PR #7956, opened 2018-05-17
Make the filename option, as exposed to the plugins, consistently relative to the working directory

loganfsmyth:
| Fixed Issues?            | 
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Currently the `opts.filename` value exposed to plugins is just whatever the user passed. While it _could_ be relative to the working directory, if Babel was passed an absolute URL, it'll be absolute.

This PR explicitly ensures the filename is a relative path based on the working directory. This also exposes an officially endorsed API for reading the working directory path.

### DIFF
--- a/packages/babel-cli/test/index.js
+++ b/packages/babel-cli/test/index.js
@@ -1,4 +1,3 @@
-const includes = require("lodash/includes");
 const readdir = require("fs-readdir-recursive");
 const helper = require("@babel/helper-fixtures");
 const rimraf = require("rimraf");
@@ -51,7 +50,7 @@ const assertTest = function(stdout, stderr, opts) {
 
   if (opts.stderr) {
     if (opts.stderrContains) {
-      expect(includes(stderr, expectStderr)).toBe(true);
+      expect(stderr).toContain(expectStderr);
     } else {
       expect(stderr).toBe(expectStderr);
     }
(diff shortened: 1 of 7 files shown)


rtsao:
I think it's worth noting that [`babel-loader` seems to always pass an absolute path to the`filename` option](https://github.com/babel/babel-loader/blob/1a7a4400da73d503166ca976da6eecd0cefd4774/src/index.js#L34) (I think [`this.resourcePath` in the webpack loader API](https://webpack.js.org/api/loaders/#this-resourcepath) yields an absolute path).

This change may cause specialized plugins that only run in the context of `babel-loader` might break if they had assumptions about this. Of course, this wasn't part of the official public API, but was an implicit invariant for this use case.

loganfsmyth:
I went back and forth about whether it made sense to default to relative or absolute and I think both are entirely possible.

I ended up going with relative, but in all honesty I'd be happy to change it if people feel it'd be easier to manage. I just want it to be consistent.

KevinGrandon:
Consistency is awesome, so thanks for working on this. I think changing it to absolute would fix our use-case, as well as other folks who use babel-loader. It would be awesome if you're up for changing it to that. Happy to test out a beta/alpha version if it would help.

PR #7960, opened 2018-05-17
[skip ci] Add badges to display install size

styfle:
| Fixed Issues?            | 👎 
| Tests Added + Pass?      | 👎
| Documentation PR         | ✅ 

This adds a column to the `README.md` that displays the npm install size of each package.

It looks like this:

screenshot


Andarist:
Is this really an important information? Would you base any technical decision based on it - like using babel or not based on that size?

styfle:
> Would you base any technical decision based on it - like using babel or not based on that size?

@Andarist Yes, that's why I made Package Phobia...because no one is monitoring dev tools bloat. If babel were 100 MB I would not use it, I would find an alternative.

Right now, babel is actually on the smaller side compared to a lot of dev tools and it would be good to keep it that way.

See [this README](https://github.com/styfle/packagephobia#readme) for some more info and some real comparisons of existing packages.

Thanks 👍

styfle:
Also note that some prominent projects have added this badge.

Just to name a few:

- [webpack](https://github.com/webpack/webpack#readme)
- [webpack-command](https://github.com/webpack-contrib/webpack-command#readme)
- [preact](https://github.com/developit/preact#readme)
- [ow](https://github.com/sindresorhus/ow#readme)
- [isomorphic-git](https://github.com/isomorphic-git/isomorphic-git#readme)
- [greenlet](https://github.com/developit/greenlet#readme)

xtuc:
This is not an important information IMO.

styfle:
The purpose of Package Phobia is for tools that are not used on the web.

Package Phobia was used to identify a bug in TypeScript (a built tool that's not used on the web): https://github.com/Microsoft/TypeScript/issues/23339#issuecomment-390365339

No one is monitoring these stats as far as I know so that's why I created Package Phobia...it's about visibility.

nicolo-ribaudo:
During the last meeting we decided not to merge this PR, because 99% of our users want to know the size of Babel's _output_, not of Babel itself.
On the other hand, we enabled `buildsize` https://github.com/babel/babel/pull/9364#issuecomment-467178710 to prevent size regressions in `@babel/standalone`, since it is loaded in the browser.

PR #7966, opened 2018-05-17
Add CLI defaults for input and output directory

### LINKED ISSUE #7237: @babel/cli: default the src/out directory
https://babeljs.io/docs/usage/cli/#compile-directories

I think we can default `babel` without any arguments to be `babel src --out-dir lib` (default to the src/ directory and output to lib), I guess we may need to check existence and all that?

This is just so if you want to do something in package.json like `"build": "babel src -d lib"`, you can do ``"build": "babel"`` instead.

Can look to other tools that do this kind of thing as well.

> aside: we should add some better error messages around this kind of thing, like if a config isn't found, etc.

iansu:
| Fixed Issues?            | Fixes #7237

When running `babel` with no arguments input directory will default to `src` and output directory will default to:

1. Directory specified in `main` field in `package.json`
2. `lib` if it exists
3. `dist` if it exists
4. `lib`

Work in progress. To do:
- [ ] Add and fix tests
- [ ] Add documentation
- [ ] Update CLI help

### DIFF
--- a/packages/babel-cli/src/babel/options.js
+++ b/packages/babel-cli/src/babel/options.js
@@ -1,4 +1,5 @@
 import fs from "fs";
+import path from "path";
 
 import commander from "commander";
 import { version } from "@babel/core";
@@ -124,6 +125,7 @@ commander.option(
 commander.option(
   "-d, --out-dir [out]",
   "Compile an input directory of modules into an output directory",
+  getOutDir(),
 );
 commander.option(
   "--relative",
@@ -151,6 +153,10 @@ export default function parseArgv(args: Array<string>) {
   commander.parse(args);
 
   const errors = [];
+
+  if (commander.args.length === 0) {
+    commander.args.push("src");
+  }
 
   let filenames = commander.args.reduce(function(globbed, input) {
     let files = glob.sync(input);
@@ -280,4 +286,25 @@ function collect(value, previousValue): Array<string> {
   const values = value.split(",");
 
   return previousValue ? previousValue.concat(values) : values;
+}
+
+function getOutDir(): string {
+  const outDirOptions = ["lib", "dist"];
+
+  // check package.json main field if it exists
+  if (fs.existsSync("package.json")) {
+    const packageJson = JSON.parse(fs.readFileSync("package.json"));
+
+    if (packageJson && packageJson.main) {
+      return path.dirname(packageJson.main);
+    }
+  }
+
+  outDirOptions.some(function(outDir) {
+    if (fs.existsSync(outDir)) {
+      return outDir;
+    }
+  });
+
+  return "lib";
 }


xtuc:
I saw on some projects that `lib` is used for vendored libraries and that would be not good to add our js files in it. 

I think the rules should be: if directory x exists pick another one. Note that the build destination directories are usually not committed and thus not exisinting.

iansu:
That's a good point about `lib` existing but being used for another purpose.

I don't think the rule you're proposing will work either though. The first time I build `lib` won't exist so it will be used. If I build again `lib` will exist so `dist` will be used. Then if I build again both will exist and the build will fail?

Maybe we should just use the path from the `main` field in `package.json` and if that field doesn't exist show an error and exit?

packages/babel-cli/src/babel/options.js
```diff
@@ -281,3 +287,24 @@ function collect(value, previousValue): Array<string> {
...
+
+function getOutDir(): string {
+  const outDirOptions = ["lib", "dist"];
+
+  // check package.json main field if it exists
+  if (fs.existsSync("package.json")) {
+    const packageJson = JSON.parse(fs.readFileSync("package.json"));
+
+    if (packageJson && packageJson.main) {
+      return path.dirname(packageJson.main);
```
loganfsmyth: It's valid for a `package.json#main` to actually be a folder, so this will have issues, I'd suggest that we do

    const basedir = process.cwd();
    const filepath = require.resolve(path.resolve(basedir, packageJson.main));
    return path.dirname(filepath);

so that we resolve the specific file first.

packages/babel-cli/src/babel/options.js
```diff
@@ -124,6 +125,7 @@ commander.option(
 commander.option(
   "-d, --out-dir [out]",
   "Compile an input directory of modules into an output directory",
+  getOutDir(),
```
loganfsmyth: This is going to break the check at https://github.com/iansu/babel/blob/3ec44323aa3477de09dc5b836a0f7233fc6b2967/packages/babel-cli/src/babel/index.js#L9 because there will always be a dir. We should invert that so that if the user specifies `--out-file`, it uses the file approach.

This is going to break Babel's ability to write to stdout: https://github.com/babel/babel/blob/6b91d6434d2682ef6b0bf4ff5c9540b8ddbcd7f9/packages/babel-cli/src/babel/file.js#L90 because there will always be a file. We'll want to allow `--out-file -` to say "output to stdout".

packages/babel-cli/src/babel/options.js
```diff
@@ -152,6 +154,10 @@ export default function parseArgv(args: Array<string>) {
 
   const errors = [];
 
+  if (commander.args.length === 0) {
+    commander.args.push("src");
```
loganfsmyth: This is going to break https://github.com/babel/babel/blob/6b91d6434d2682ef6b0bf4ff5c9540b8ddbcd7f9/packages/babel-cli/src/babel/file.js#L228 because there will always be a filename. We'll want to allow

    babel - ...

where the `-` in place of a filename means it will read from stdin.

packages/babel-cli/src/babel/options.js
loganfsmyth: Also we may want to skip here if the dirname that is returned is the root of the package and not a subfolder.

packages/babel-cli/src/babel/options.js
```diff
@@ -281,3 +287,24 @@ function collect(value, previousValue): Array<string> {
...
+  // check package.json main field if it exists
+  if (fs.existsSync("package.json")) {
+    const packageJson = JSON.parse(fs.readFileSync("package.json"));
+
+    if (packageJson && packageJson.main) {
+      return path.dirname(packageJson.main);
+    }
+  }
+
+  outDirOptions.some(function(outDir) {
```
loganfsmyth: I think skipping the `lib` and `dist` and only using `package.json` might be reasonable, given the concern about `lib`, but it definitely has risks either way.

iansu:
@loganfsmyth Thanks for the review. I will try to address those comments as well as some of the previous discussion later today. The stdin/stdout thing is definitely an issue. That's why a bunch of tests are currently failing. Using `-` to indicate stdin/stdout seems like a good solution to me.

packages/babel-cli/src/babel/options.js
iansu: For this second comment are you saying if `dirname` is the same as the `basepath` then we should just default to something like `lib` or `dist` (instead of writing the output to the root of the package)?

billyjanitsch:
Using `main` from package.json wouldn't work well for packages using [esm](https://github.com/standard-things/esm), wherein `main` points to a single file which activates the loader and points to the ESM entry point. Something like:

```js
// index.js
module.exports = require('esm')(module)('./lib')
```

+1 for @loganfsmyth's suggestion of skipping `main` if it resolves to the root package directory, which would alleviate this problem.

dhruvdutt:
@iansu Any update on this?

iansu:
Yes! I made the suggested changes but haven't had a chance to test them yet. It kind of fell off my radar. Thanks for the reminder. I will try to push an update soon.

dhruvdutt:
@iansu Let me know if I can lend a hand. Still looking for my first contribution to babel.

PR #7967, opened 2018-05-18
TypeScript: Fix TSInferType .typeParameter type.

benjamn:
| Fixed Issues?            | Various `ast-types` test failures, e.g. https://travis-ci.org/benjamn/ast-types/jobs/369634972
| Tests Added + Pass?      | Fixed existing tests that expected `TypeParameter`

Follow-up to https://github.com/babel/babel/pull/7404

There is no `TypeParameter` type in the Babylon TypeScript AST hierarchy. Flow does have a `TypeParameter` type, but it should not be confused with the TypeScript `TSTypeParameter`, since Flow !== TypeScript, and the types have totally different fields. Instead, the `.typeParameter.type` of a Babylon-parsed `TSInferType` node should be `TSTypeParameter`.

It would probably be fine to leave the declared type of the `TSInferType` `.typeParameter` field as `TSType` instead of `TSTypeParameter`, but the more specific `TSTypeParameter` type should be safe/correct, since the TypeScript [`parseInferType` function](https://github.com/Microsoft/TypeScript/blob/66d6e5e6e007ce3ac70b7371c3e7ac46a99a0fcd/src/compiler/parser.ts#L3006) always uses `SyntaxKind.TypeParameter`.

I noticed this typo because it has been causing `ast-types` test failures lately, e.g. https://travis-ci.org/benjamn/ast-types/jobs/369634972

### DIFF
--- a/packages/babel-types/README.md
+++ b/packages/babel-types/README.md
@@ -2131,7 +2131,7 @@ See also `t.isTSInferType(node, opts)` and `t.assertTSInferType(node, opts)`.
 
 Aliases: `TSType`
 
- - `typeParameter`: `TSType` (required)
+ - `typeParameter`: `TSTypeParameter` (required)
 
 ---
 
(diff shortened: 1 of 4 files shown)


benjamn:
@andy-ms does this seem reasonable?

Issue #7771, opened 2018-04-21
Getters/Setters in Superclass called when loose defining Subclass


jridgewell:
[Given](http://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=usage&code_lz=MYGwhgzhAEBCkFNoG8BQ1oHMEBdo4QhwAoBKFdaAX1RtVEhgHkAjAK2gQA8CA7AExjwISNBgD046ACUE_BADMAlrxWZoAW1wALAPb9o2hACcElAkTIoaNIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=true&presets=es2015-loose&prettier=false&targets=Node-4&version=7.0.0-beta.44&envVersion=7.0.0-beta.44):

```js
class Base {
  get test() {
  }
}

class Obj extends Base {
  // Redefining method here
  test() { }
}
```

This outputs:

```js
var Base = function () {
  function Base() {}

  _createClass(Base, [{
    key: "test",
    get: function get() {}
  }]);

  return Base;
}();

var Obj = function (_Base) {
  _inheritsLoose(Obj, _Base);

  function Obj() {
    return _Base.apply(this, arguments) || this;
  }

  var _proto = Obj.prototype;

  // Redefining method here
  _proto.test = function test() {};

  return Obj;
}(Base);
```

Notice we do `_inheritsLoose(Obj, _Base)` immediately, then do `Obj.prototype.test = function() {}`. This tries to set to a inherited property that's only a getter, so it throws in strict mode.

We could just do the `_inheritsLoose` after defining the prototype methods?

nicolo-ribaudo:
> We could just do the _inheritsLoose after defining the prototype methods?

If it would work, yes. If we would need to use `Object.defineProperty` I think that we could keep this bug as a "loose mode thing", since it isn't very common to overwrite a setter with a method.

jridgewell:
I can't think of any behavior that requires the prototype chain to be set so soon.

In strict mode, we have to use `defineProperty` to set non-enumerable methods, so they don't run into edge cases like this. But in loose, we don't care about enumerability. We could define the inheritance as the last step and everything would still be the same.

Even static class properties are defined after the class closure, so they would still get the same error treatment as now.

But we can get closer to spec behavior by just moving the inherit call later.

Issue #11, opened 2014-10-08
Support `u` flag for regular expressions


mathiasbynens:
See https://mathiasbynens.be/notes/es6-unicode-regex for more info on this feature.

Here’s a module you can use for the transpilation: https://github.com/mathiasbynens/regexpu Traceur depends on regexpu as well.

PR #7969, opened 2018-05-18
Fix class inheritance in IE10

### LINKED ISSUE #7892: Class inherit prototype uncorrect in IE10 using @babel/runtime@7.0.0-beta.46
## Bug Report

**Current Behavior**

The output is `undefined` in IE10.

**Input Code**

```js
function A() {}
A.prototype.a = 'aaa';

class B extends A {}
console.log(B.prototype.a);
```

**Expected behavior/code**

Expect output `aaa`, but got `undefined` in IE10. I also tested this in `@vue/cli@3.0.0-beta.9`, it have the same problem.

**Environment**
- Babel version(s): v7.0.0-beta.36
- OS: OSX 10.13.4

**Possible Solution**

### LINKED ISSUE #7965: Verify and fix broken subclassing in IE10 due to change in getPrototypeOf usage
I'm 99% sure this is an issue, but I haven't tested myself.

Looks like the point was raised in https://github.com/babel/babel/pull/7020/files#r156839425, but at the time it was brushed off as not a blocker because the code was specifically for subclassing builtins, which won't work on older browsers anyway.

Since then however, in @jridgewell's https://github.com/babel/babel/pull/7674, this code has moved into more general use, meaning that it will affect any code performing subclassing.

We currently have the helper
```
helpers.getPrototypeOf = () => template.program.ast`
  export default function _getPrototypeOf(o) {
    _getPrototypeOf = Object.getPrototypeOf || function _getPrototypeOf(o) {
      return o.__proto__;
    };
    return _getPrototypeOf(o);
  }
`;
```
but on IE10, it already has a `getPrototypeOf`, but since it doesn't have a `setPrototypeOf`, IE10 will simply assign `.__proto__`, which means that the `getPrototypeOf` helper will not return the prototype assigned by the `setPrototypeOf` helper.

To clarify, this is all about our helpers. We use `helpers.getPrototypeOf(Foo).apply()` for `super()` for instance, so even in cases where `Object.setPrototypeOf` is unsupported, we still need `helpers.setPrototypeOf()` to do _something_ so that asking for the prototype back with `helpers.getPrototypeOf()` return the correct value, even if the prototype chain itself isn't actually manipulated. In Babel 6 this is accomplished by _always_ reading `.__proto__` first, so if we do `foo.__proto__` in `helpers.setPrototypeOf`, `helpers.getPrototypeOf` should read `.__proto__` even though `Object.getPrototypeOf` returns something different.

jridgewell:
This reverts commit f8ab9466d331871a90f458af40b14e8d831e0c29, and unfixes #7771, because of IE 10. 😩

Also updates the `getPrototypeOf` helper according to https://github.com/babel/babel/issues/7965#issuecomment-390046440.

Fixes #7892.
Fixes #7965.


packages/babel-helpers/src/helpers.js
```diff
@@ -399,23 +399,31 @@ helpers.inherits = () => template.program.ast`
     if (typeof superClass !== "function" && superClass !== null) {
       throw new TypeError("Super expression must either be null or a function");
     }
-    setPrototypeOf(subClass.prototype, superClass && superClass.prototype);
+    subClass.prototype = Object.create(superClass && superClass.prototype, {
+      constructor: {
+        value: subClass,
+        enumerable: false,
```
Kovensky: It could be better to omit `enumerable: false` just to not force sham implementations to throw, but this would only happen on IE<9 anyway.

packages/babel-plugin-transform-classes/test/fixtures/get-set-loose/call-semantics-getter-defined-on-parent/exec.js
```diff
@@ -20,11 +20,14 @@ class Obj extends Base {
     super.test(...[1, 2, 3]);
     return super.test(...arguments);
   }
-
-  test() {
+}
+Object.defineProperty(Obj.prototype, 'test', {
```
loganfsmyth: I don't understand what caused this change. Any idea? It doesn't seem like this should be split off like this.

packages/babel-helpers/src/helpers.js
```diff
@@ -399,23 +399,31 @@ helpers.inherits = () => template.program.ast`
...
 `;
 
 helpers.getPrototypeOf = () => template.program.ast`
   export default function _getPrototypeOf(o) {
-    _getPrototypeOf = Object.getPrototypeOf || function _getPrototypeOf(o) {
-      return o.__proto__;
-    };
+    _getPrototypeOf = Object.setPrototypeOf
+      ? Object.getPrototypeOf
+      : function _getPrototypeOf(o) { return o.__proto__; };
```
loganfsmyth: I think I'd still want to fall back to `Object.getPrototypeOf` if `__proto__` falsy, since that was the old behavior.

packages/babel-helpers/src/helpers.js
jridgewell: This is the spec inherits. I think changing it to be enumerable would break more people who use a for-in.

You can always use loose mode to support these IE9?

packages/babel-helpers/src/helpers.js
jridgewell: Done.

packages/babel-helpers/src/helpers.js
loganfsmyth: I think they just mean that the default is already `false` anyway, so leaving it out doesn't make anything fail, but allows polyfills to be less agressive in requiring enumerability on older platforms.

packages/babel-plugin-transform-classes/test/fixtures/get-set-loose/call-semantics-getter-defined-on-parent/exec.js
jridgewell: This is the reason I filled #7771. https://github.com/jridgewell/babel/commit/ab403f3d6d768a98130ef3588feed71b941beaac#diff-0cd85dd4813df0687b8938547953d585L24

`test` here will be set on the subclass, but the super defines a `get test`, so it'll throw.

packages/babel-helpers/src/helpers.js
jridgewell: Ahh, ok.

packages/babel-plugin-transform-classes/test/fixtures/get-set-loose/call-semantics-getter-defined-on-parent/exec.js
loganfsmyth: Ahh got it.

packages/babel-helpers/src/helpers.js
```diff
@@ -399,23 +399,34 @@ helpers.inherits = () => template.program.ast`
...
 helpers.getPrototypeOf = () => template.program.ast`
   export default function _getPrototypeOf(o) {
-    _getPrototypeOf = Object.getPrototypeOf || function _getPrototypeOf(o) {
-      return o.__proto__;
-    };
+    if (Object.setPrototypeOf && Object.getPrototypeOf) {
+      _getPrototypeOf = Object.getPrototypeOf;
+    } else {
+      _getPrototypeOf = function _getPrototypeOf(o) {
+        return o.__proto__;
```
loganfsmyth: I think my comment here was misinterpreted. My concern was that this does not preserve the behavior changed in https://github.com/babel/babel/pull/7675/files#diff-a6fddd06921a2b0da9eb1fb68cd794b9L14

Old super calls used to be `(Foo.__proto__ || Object.getPrototypeOf(Foo)).call`, but here we assume that `__proto__` is the source of truth and will always have a value, instead of falling back to `Object.getPrototypeOf` of `__proto__` is falsy.

jridgewell: Wow, that was dyslexic of me. Fixed.

benjamn:
How soon can this be released? Meteor implemented [a pretty nasty workaround](https://github.com/meteor/meteor/blob/e3bac92c82c56acb65c42f8e4bfa7b244666f107/packages/babel-runtime/legacy.js#L17-L38) for this, which I would love to revert before shipping Meteor 1.7.

loganfsmyth:
@benjamn Likely tomorrow. What's your timetable for 1.7?

benjamn:
Tomorrow is great. We're on RC #11 right now, but I think this is a good reason to wait. I'm giving a talk at [Meteor Night](https://blog.meteor.com/come-to-meteor-night-on-may-30-in-san-francisco-a7472b81d147) next Wednesday, but we can just ship the workaround if the next Babel beta somehow doesn't get released by then.

PR #7973, opened 2018-05-18
Support Flow's inline interface syntax

samwgoldman:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Documentation PR         | 
| Any Dependency Changes?  |

Flow added support for interface type annotations, which can be used inline like object type annotations, in https://github.com/facebook/flow/commit/4f93f3adb7a5fedd3d572d62d718cf48c15a08ce.

This adds the matching support to Babylon. Note that `interface` was already disallowed as a type identifier (i.e., `type interface = string`), so this syntax does not break any existing programs.

### DIFF
--- a/packages/babel-generator/src/generators/flow.js
+++ b/packages/babel-generator/src/generators/flow.js
@@ -262,6 +262,19 @@ function andSeparator() {
   this.space();
 }
 
+export function InterfaceTypeAnnotation(node: Object) {
+  this.word("interface");
+  this.space();
+  if (node.extends.length) {
+    this.space();
+    this.word("extends");
+    this.space();
+    this.printList(node.extends, node);
+  }
+  this.space();
+  this.print(node.body, node);
+}
+
 export function IntersectionTypeAnnotation(node: Object) {
   this.printJoin(node.types, node, { separator: andSeparator });
 }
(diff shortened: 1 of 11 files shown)


samwgoldman:
cc @babel/flow

packages/babel-generator/src/generators/flow.js
```diff
@@ -262,6 +262,19 @@ function andSeparator() {
   this.space();
 }
 
+export function InterfaceTypeAnnotation(node: Object) {
+  this.word("interface");
+  this.space();
```
nicolo-ribaudo: I think that this can be removed, since it is already called at lines 269 and 274.

packages/babel-types/src/definitions/flow.js
```diff
@@ -211,6 +211,15 @@ defineType("InterfaceExtends", {
 
 defineInterfaceishType("InterfaceDeclaration");
 
+defineType("InterfaceTypeAnnotation", {
+  visitor: ["extends", "body"],
+  aliases: ["Flow", "FlowType"],
+  fields: {
+    extends: validateOptional(arrayOfType("InterfaceExtends")),
```
nicolo-ribaudo: If this is optional, in the generator you should check `if (node.extends && node.extends.length)` instead of `if (node.extends.length)`.

PR #7977, opened 2018-05-18
Update babel-types (flow internal slots)

nicolo-ribaudo:
We should teach @babel-bot how to do this.

#7947


existentialism:
Yeah, or (in the short term) we can lint check or just regen on precommit lol.

PR #7978, opened 2018-05-18
Support Flow's proto modifier syntax for declared classes

samwgoldman:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Documentation PR         | 
| Any Dependency Changes?  |

See https://github.com/facebook/flow/commit/eb815be907eeae9f4851ff154087ebf079510d18 for more information about this feature.
    
The proto modifier indicates that a property declared using `x: T` syntax is actually present on the prototype object of the class, rather than an own property.
    
The proto and static modifiers are mutually exclusive, as class declarations don't simultaneously define the static prototype object, as they do the instance prototype.
    
This syntax is only supported on declared classes, not object types, interfaces, or runtime class declarations, and as such should only appear in library definitions.

### DIFF
--- a/packages/babel-generator/src/generators/flow.js
+++ b/packages/babel-generator/src/generators/flow.js
@@ -472,6 +472,10 @@ export function ObjectTypeIndexer(node: Object) {
 }
 
 export function ObjectTypeProperty(node: Object) {
+  if (node.proto) {
+    this.word("proto");
+    this.space();
+  }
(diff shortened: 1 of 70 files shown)


existentialism:
Rebased, and updated to assert `proto` is boolean.

@nicolo-ribaudo we can probably just follow up with both an exclusive assertion util and updating the builder?

PR #7981, opened 2018-05-18
Remove parenthesizedArgument

### LINKED ISSUE #7957: parenthesizedArgument is never set to true
## Bug Report

The code that sets `parenthesizedArgument` can never set it to true.

https://github.com/babel/babel/blob/ac13c302f74e58179698bfb95055c9abda632b75/packages/babylon/src/parser/expression.js#L370-L375

Because if there is a `tt.parenL` then `node.argument` will have `extra.parenthesized` set to `true`.

The code was added by @danez in https://github.com/babel/babel/commit/97325592fa0a0ed36389ad80e8e74aa38d875868

What was the intent of `parenthesizedArgument`? Am I missing something?

arv:
| Fixed Issues?            | Fixes #7957
| Patch: Bug Fix?          | Cleanup

The extra field parenthesizedArgument was never set to true. It was always set to false and the code that read this was incorrectly checking if extra was present.

### DIFF
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -287,9 +287,7 @@ export default class ExpressionParser extends LValParser {
         if (
           operator === "**" &&
           left.type === "UnaryExpression" &&
-          left.extra &&
-          !left.extra.parenthesizedArgument &&
-          !left.extra.parenthesized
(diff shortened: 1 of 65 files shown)


danez:
If the tests still work  👍

loganfsmyth:
Thanks!

PR #7986, opened 2018-05-19
Better error message for invalid decorators syntax

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

https://github.com/babel/babel/pull/7869#discussion_r189399454 @loganfsmyth

### DIFF
--- a/packages/babel-parser/src/parser/statement.js
+++ b/packages/babel-parser/src/parser/statement.js
@@ -230,13 +230,6 @@ export default class StatementParser extends ExpressionParser {
   }
 
   parseDecorators(allowExport?: boolean): void {
-    if (
-      this.hasPlugin("decorators") &&
-      !this.getPluginOption("decorators", "decoratorsBeforeExport")
-    ) {
-      allowExport = false;
-    }
-
     const currentContextDecorators = this.state.decoratorStack[
       this.state.decoratorStack.length - 1
     ];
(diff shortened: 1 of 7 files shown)


packages/babel-parser/src/parser/statement.js
```diff
@@ -246,18 +239,22 @@ export default class StatementParser extends ExpressionParser {
...
     if (this.match(tt._export)) {
-      if (allowExport) {
-        return;
-      } else {
+      if (!allowExport) {
+        this.unexpected();
+      }
+
+      if (
+        this.hasPlugin("decorators") &&
```
loganfsmyth: Not a blocker for this PR, but it would be great if we could integrate this with the plugin handling logic in https://github.com/babel/babel/blob/6b91d6434d2682ef6b0bf4ff5c9540b8ddbcd7f9/packages/babel-core/src/transformation/normalize-file.js#L129 so that the message gave useful feedback about the Babel plugin instead of the parser plugin.

packages/babel-parser/src/parser/statement.js
```diff
@@ -1531,7 +1533,12 @@ export default class StatementParser extends ExpressionParser {
       this.expectOnePlugin(["decorators", "decorators-legacy"]);
       if (this.hasPlugin("decorators")) {
         if (this.getPluginOption("decorators", "decoratorsBeforeExport")) {
-          this.unexpected();
+          this.unexpected(
+            this.state.start,
+            "Decorators must be placed *before* the 'export' keyword." +
+              " You can set the 'decoratorsBeforeExport' option to false to use" +
+              " the 'export @decorator class {}' syntax",
```
xtuc: Maybe we could add where to set the configuration?

> You can set the option on the decorator plugin in your Babel configuration

hzoo: Yeah similar to Logan's question ^ about using the similar error feedback. This error is for Babylon itself which also has plugin options (first time we are doing this), and we'd probably want to do a similar wrapping of the error

PR #7988, opened 2018-05-19
Clone id before mutating in babel-helpers-function-name

devongovett:
This appears to already be resolved in 7, but I'd like to get it back ported to 6.x if possible. Without this, both the variable declaration and function name identifiers will get `NOT_LOCAL_BINDING` set on them instead of just the function name since they reference the same object when processing code like below. That causes subsequent renaming of the variable to also affect the function name, which messes with scope analysis as well.

```javascript
let foo = function () {};
```

| Fixed Issues?            | 


loganfsmyth:
I've merged this, but I don't know when we'll next publish 6.x at this point.

PR #7990, opened 2018-05-21
passes the jsxPragma options from preset-typescript to plugin-transform-typescript

### LINKED ISSUE #7979: TypeScript JSX Fragment removes React import
## Bug Report

**Current Behavior**
Currently whenever I'm using "@babel/preset-typescript" with JSX fragments and no other JSX, the React import is removed from the compiled output. This only happens with JSX fragments. If you use `React.Fragment` instead, the compiled output is correct. Also worth noting that any other non-fragment jsx in the source will cause the React import to be maintained.

**Input Code**

```tsx
import React, { SFC } from "react";

export const BreadcrumbItem: SFC = ({ children }) => <>{children}</>;
```

**Expected code**

```js
import React from "react";
export var BreadcrumbItem = function BreadcrumbItem(_ref) {
  var children = _ref.children;
  return React.createElement(React.Fragment, null, children);
};
```

**Actual code**

```js
export var BreadcrumbItem = function BreadcrumbItem(_ref) {
  var children = _ref.children;
  return React.createElement(React.Fragment, null, children);
};
```

```js
{
  "presets": [
    "@babel/preset-typescript",
    "@babel/preset-react",
    [
      "@babel/preset-env",
      {
        "modules": false,
        "targets": {
          "browsers": "ie 11"
        },
        "useBuiltIns": "entry"
      }
    ]
  ],
  "plugins": [
    ["emotion", { "sourceMap": true, "autoLabel": true }],
  ]
}
```

**Environment**
- Babel version(s): v^7.0.0-beta.47
- Node/npm version: v8.9.4
- OS: Windows 10
- Monorepo no
- How you are using Babel: cli & loader

I have determined this issue to only happen when when "@babel/preset-typescript" is included. Removing any number of other plugins and presets does not fix issue. Additionally I've tried the following variations of the source and they all had the same incorrect output.

```tsx
import React from "react";
import { SFC } from "react";

export const BreadcrumbItem: SFC = ({ children }) => <>{children}</>;
```

```tsx
import * as React from "react";
import { SFC } from "react";

export const BreadcrumbItem: SFC = ({ children }) => <>{children}</>;
```

```tsx
import React from "react";

export const BreadcrumbItem = ({ children }) => <>{children}</>;
```

### LINKED ISSUE #7989: option `jsxPragma` is not working in preset/typescript
## Bug Report

**Current Behavior**
Jsx pragma is being removed from the source, having preset-typescript configured properly.

**Input Code**

```js
render(<App />);
```

**Expected behavior/code**
It should be transpiled to this:

```js
import { h } from "preact";

render(h(App));
```

**Actual behavior/code**
It is transpiled to this:

```js
render(h(App));
```

**Babel Configuration**

```json
{
  "presets": [
    ["@babel/preset-env", { "targets": { "node": "current" } }],
    ["@babel/preset-typescript", { "jsxPragma": "h" }],
    ["@babel/preset-react", { "pragma": "h" }]
  ]
}
```

**Environment**
- Babel: v7.0.0-beta.47
- Node: v9.11.1
- OS: MacOs 10.13.4
- Monorepo: Lerna
- How you are using Babel: register

**Possible Solution**

Probably the options are not being passed down, since plugin-transform-typescript accepts the `jsxPragma` and using plugin-transform-typescript actually work.

maybe this would do the trick:

```js
/* babel-preset-typescript/src/index.js */

export default declare((api, { jsxPragma = "React" }) => {
  api.assertVersion(7);

  return {
    plugins: [[transformTypeScript, { jsxPragma }]],
  };
});
```

malbernaz:
| Fixed Issues?            |  Fixes #7989, Fixes #7979
| Patch: Bug Fix?          |  yes
| Tests Added + Pass?      | yes

Jsx pragmas were being ignored by preset-typescript. I've simply passed the jsxPragma option to the plugin-transform-typescript.

### DIFF
--- a/packages/babel-preset-typescript/src/index.js
+++ b/packages/babel-preset-typescript/src/index.js
@@ -1,7 +1,7 @@
 import { declare } from "@babel/helper-plugin-utils";
 import transformTypeScript from "@babel/plugin-transform-typescript";
 
-export default declare((api, { jsxPragma = "React" }) => {
+export default declare((api, { jsxPragma }) => {
   api.assertVersion(7);
 
   return {
--- a/packages/babel-preset-typescript/src/index.js
+++ b/packages/babel-preset-typescript/src/index.js
@@ -1,10 +1,10 @@
 import { declare } from "@babel/helper-plugin-utils";
 import transformTypeScript from "@babel/plugin-transform-typescript";
 
-export default declare(api => {
+export default declare((api, { jsxPragma = "React" }) => {
   api.assertVersion(7);
 
   return {
-    plugins: [transformTypeScript],
+    plugins: [[transformTypeScript, { jsxPragma }]],
   };
 });
--- a/packages/babel-preset-typescript/README.md
+++ b/packages/babel-preset-typescript/README.md
@@ -54,6 +54,8 @@ require("@babel/core").transform("code", {
 });
 ```
 
+## Options
+
 ### `jsxPragma`
 
 `string`
--- a/packages/babel-preset-typescript/README.md
+++ b/packages/babel-preset-typescript/README.md
@@ -53,3 +53,11 @@ require("@babel/core").transform("code", {
   presets: ["@babel/preset-typescript"]
 });
 ```
+
+### `jsxPragma`
+
+`string`
+
+Replace the function used when compiling JSX expressions.
+
+This is so that we know that the import is not a type import, and should not be removed


packages/babel-preset-typescript/src/index.js
```diff
@@ -1,10 +1,10 @@
 import { declare } from "@babel/helper-plugin-utils";
 import transformTypeScript from "@babel/plugin-transform-typescript";
 
-export default declare(api => {
+export default declare((api, { jsxPragma = "React" }) => {
```
loganfsmyth: This default is already set for the transform itself, so it would better not to repeat it again.

hzoo:
Awesome nice work @malbernaz! Both for the issue and PR!!

malbernaz:
Yo! Thanks! Loved to help

Issue #7984, opened 2018-05-18
Babylon plugin option ordering is not well-defined, and potentially inconsistent with Babel


loganfsmyth:
See: https://github.com/babel/babel/pull/7869/files#r189395861

> Hey all, I'm concerned about this. Now that we've defined the ability to pass options to Babylon plugins, ordering becomes critical. What if some other piece of tooling injected a syntax plugin into the plugin list after this one, or before? Which one should take priority? For instance for https://github.com/babel/babel/pull/7934/files#diff-3a8233bcd2766d2c7d87f23f944f7726R12, say the user's framework auto-enables flow, but the user wants to enable all, which comes first?
>
> So far we've avoided this problem by not having options, and not worrying about duplicates. Now it will matter very much both what order your Babel plugins are in, but also what order they enable Babylon plugins in.
>
> At a minimum, have we verified that the ordering on these is consistent with the ordering of visitors? e.g. If two plugins are enabled with different options, whichever one's visitors run first, is also the one who's Babylon configuration takes affect?

The plugin map logic in https://github.com/babel/babel/pull/7869/files#diff-8b7138068ab1dfcbfafe8422f16505cbR45 will always take the last set of options. If I'm piecing things together in my head correctly, that'll be a problem, because the _first_ plugin in the list will run its visitors first, but the _last_ plugin in the list will be the thing specifying the options to Babylon.

hulkish:
@loganfsmyth I think we should throw a config validation error or at least report a warning for any plugins which are not intended to be used multiple times. Then, for plugins that permit it - we force them to specify via plugin api.

loganfsmyth:
> Throw if a plugin is specified twice with different options.

My concerns with this approach are:

1. We don't have any means to provide useful feedback to the user. If we throw in the parser, the user doesn't know how that translates back to their Babel plugins.
2. There's no safe definition of "different options". If a plugin is enabled as `["plugin"]` and then enabled again as `[["plugin", {feature: false}]]`, those _could_ be identical behavior, but we don't know, and it would complicate things a lot to try to handle that.
3. Along those same lines, it also means if someone published a new version of a preset or something with `flow` or whatever included, it has to qualify as a breaking change because they might have already enabled `flow` in their own plugin list.

> Merge the options, but throw if the same option is specified twice with different values. 

I don't think merging is the way to go, because we don't really know how the options will evolve over time, so it's just as possible that merging would cause us to end up with a invalid options, e.g.

```
[
[...]

PR #7994, opened 2018-05-21
[Babylon] Take the first set of options for plugins

nicolo-ribaudo:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/7984
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

(Discussed during https://github.com/babel/notes/pull/64)

### DIFF
--- a/packages/babel-parser/test/plugin-options.js
+++ b/packages/babel-parser/test/plugin-options.js
@@ -1,37 +1,31 @@
-import Parser from "../lib/parser";
+import { parse } from "../lib";
+
+function getParser(code, plugins) {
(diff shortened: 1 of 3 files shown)


packages/babel-parser/test/plugin-options.js
```diff
@@ -0,0 +1,37 @@
+import Parser from "../lib/parser";
```
loganfsmyth: These tests are failing because the parser is bundled, so only the index is accessible consistently. Looks like these need to be tweaked to test the actual public interface of the parser.

PR #7995, opened 2018-05-21
fix typo in error message

nason:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | 

Fixing a typo in an error message. Thanks so much!


loganfsmyth:
Thanks! Out of curiosity, where you actually using the same plugin twice, or did this catch a mistake you made in your config? When I added this error, I wasn't sure how many people would actually be using a plugin twice in real life.

nason:
> Out of curiosity, where you actually using the same plugin twice, or did this catch a mistake you made in your config?

It caught a mistake in my config 🌴

PR #7996, opened 2018-05-21
Adds JSXFragment to plugin-transform-typescript check for the presence of jsx

### LINKED ISSUE #7979: TypeScript JSX Fragment removes React import

malbernaz:
| Fixed Issues?            | Fixes #7979

This PR Adds `JSXFragment` to plugin-transform-typescript check for type imports.

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -276,7 +276,10 @@ export default declare((api, { jsxPragma = "React" }) => {
     // "React" or the JSX pragma is referenced as a value if there are any JSX elements in the code.
     let sourceFileHasJsx = false;
     programPath.traverse({
-      JSX() {
+      JSXElement() {
+        sourceFileHasJsx = true;
+      },
+      JSXFragment() {
         sourceFileHasJsx = true;
       },
     });
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -276,10 +276,7 @@ export default declare((api, { jsxPragma = "React" }) => {
     // "React" or the JSX pragma is referenced as a value if there are any JSX elements in the code.
     let sourceFileHasJsx = false;
     programPath.traverse({
-      JSXElement() {
-        sourceFileHasJsx = true;
-      },
-      JSXFragment() {
+      JSX() {
         sourceFileHasJsx = true;
       },
     });
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -279,6 +279,9 @@ export default declare((api, { jsxPragma = "React" }) => {
       JSXElement() {
         sourceFileHasJsx = true;
       },
+      JSXFragment() {
+        sourceFileHasJsx = true;
+      },
     });
     return !sourceFileHasJsx;
   }
--- a/packages/babel-plugin-transform-typescript/test/fixtures/imports/elide-react-no-3/input.mjs
+++ b/packages/babel-plugin-transform-typescript/test/fixtures/imports/elide-react-no-3/input.mjs
@@ -1,3 +1,3 @@
 // Don't elide React if a JSX fragment appears somewhere.
-import React from "react";
-<></>
+import * as React from "react";
+<></>;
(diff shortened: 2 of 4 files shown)


hzoo:
Nice, Can you rebase on top of master? the other code changes shouldn't be necessary because of the PR I merged earlier I believe?

malbernaz:
Thanks! Had some trouble to rebase it, but there you go. Think it's ok now

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -279,6 +279,9 @@ export default declare((api, { jsxPragma = "React" }) => {
       JSXElement() {
         sourceFileHasJsx = true;
       },
+      JSXFragment() {
+        sourceFileHasJsx = true;
+      },
```
xtuc: Actually we could just use `JSX() { sourceFileHasJsx = true }` which covers both

hzoo: Yeah or "JSXElement|JSXFragment(){}", but not sure I care

xtuc: I don't really care either but I have a preference for the jsx since it's smaller.

xtuc:
That's unexpected, now the tests are failing. The JSX syntax isn't parsed anymore for some reason.

malbernaz:
should I regress?

xtuc:
Well yes but I don't understand why it fails.

malbernaz:
Yes, very strange.. No clue at all of whats happening?

xtuc:
I couldn't reproduce locally so it's hard to debug anyway.

malbernaz:
and it failed again...

xtuc:
I'm looking into it, in previous commit the tests passed.

I also pinged you on Slack, if you want to chat over there.

malbernaz:
ok, thanks

xtuc:
Sorry but i'm not able to reproduce it locally.

jquense:
Just ran across a version of this bug, if you specify the `jsx` pragma in a comment that differs from the Fragment pragma it's still removed.

```js
/** @jsx jsx **/
/** @jsxFrag React.Fragment **/

import React from 'react'

<><p/></>
```
Output will be:

```js
/** @jsx jsx **/
/** @jsxFrag React.Fragment **/

jsx(React.Fragment, null, jsx('p))
```

PR #7999, opened 2018-05-22
[babylon] Refactor mixin plugins handling & validation

nicolo-ribaudo:
| Fixed Issues?            | (internal flow errors)
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

- I moved the plugin validation from src/index.js to src/plugins-utils.js, so now the `getParserClass` function is easier to undarstand.
- I fixed some flow errors introduced https://github.com/babel/babel/pull/7869 (but master still has some flow errors)

`make build` fails for me, so I didn't run `yarn jest`.

### DIFF
--- a/packages/babel-parser/src/plugin-utils.js
+++ b/packages/babel-parser/src/plugin-utils.js
@@ -1,0 +1,50 @@
+// @flow
+
+import type Parser from "./parser";
+
+export type Plugin = string | [string, Object];
+
+export type PluginList = $ReadOnlyArray<Plugin>;
+
+export type MixinPlugin = (superClass: Class<Parser>) => Class<Parser>;
+
+export function hasPlugin(plugins: PluginList, name: string): boolean {
+  return plugins.some(plugin => {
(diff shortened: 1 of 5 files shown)


packages/babel-parser/src/plugin-utils.js
```diff
@@ -0,0 +1,48 @@
+import type { Parser } from "./parser";
```
existentialism: Should add a `// @flow` atop this file.

Also, I don't think there's a `Parser` type export in "./parser"?

nicolo-ribaudo: > Should add a // @flow atop this file.

Fixing flow errors by not enabling flow is easier :rofl: 

> Also, I don't think there's a Parser type export in "./parser"?

Yeah, it should have been a default import. Classes can be imported as types, so flow strips the import.

PR #8002, opened 2018-05-22
Fix some flow issues in @babel/parser flow plugin

existentialism:


packages/babel-parser/src/plugins/flow.js
```diff
@@ -2455,6 +2456,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         node.callee = base;
         node.typeArguments = this.flowParseTypeParameterInstantiation();
         this.expect(tt.parenL);
+        // $FlowFixMe
```
existentialism: Not great, but related to the [TODO](https://github.com/babel/babel/pull/8002/files#diff-b8d40ce8ecac1a4432171ce346567e3dR542) around converting arguments into `$ReadOnlyArray`, which requires quite a bit more changes.

packages/babel-parser/src/types.js
```diff
@@ -541,6 +540,8 @@ export type ConditionalExpression = NodeBase & {
 export type CallOrNewBase = NodeBase & {
   callee: Expression | Super | Import,
   arguments: Array<Expression | SpreadElement>, // TODO: $ReadOnlyArray
+  optional?: boolean, // For OptionalCallExpression only
```
loganfsmyth: Is moving this something we can avoid if we refactored the code to verify the type before setting the flag? Even if it's a pain, it seems like ideally we'd keep the types as specific as possible.

existentialism: Agreed on keeping types specific.

I went down a bit of a rabbit hold with refactoring last night... but actually on second look, seems like simply removing the CallExpression typecast makes flow happy.

samwgoldman:
Thanks so much @existentialism! I went on vacation right after pushing up those PRs, so I really appreciate your cleaning up my mess. I will be sure to keep Flow clean in the future.

PR #8003, opened 2018-05-22
fix proposal-async-generator-functions version for node

jquense:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      | 
| Documentation PR         | 
| Any Dependency Changes?  |

Node 10 supports async generators, this (should?) adds it to the support matrix


existentialism:
@jquense we'll need compat-table to get updated with node 10 support, since data/*.json are generated directly from it.

jquense:
boo i was worried that was the case. Any idea how to PR correctly upstream?

PR #8005, opened 2018-05-22
Handle Infinity, -Infinity, NaN, and -0 in t.valueToNode().

### LINKED ISSUE #8001: -0 gets printed as 0 in Babel
## Bug Report

**Current Behavior**
`t.numericLiteral` does not support `-0`.

**Input Code**
http://astexplorer.net/#/gist/c5f83ae7704f4089946202031abba218/3c1d2aa4249f46f18aa4b58c0b995d353d0d5186

**Possible Solution**
Check that `0` is `-0` with `Object.is` before printing it.

loganfsmyth:
| Fixed Issues?            | Fixes #8001
| Major: Breaking Change?  | Not unless things were relying on a very specific AST structure as output
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

The filed bug was for `-0`, but I also realize that `numericLiteral` also shouldn't be getting `-Infinity`/`Infinity` or `NaN` since those just get stringified with the JS-standard stringification logic which means we get a reference to `NaN` or `Infinity` identifiers, which could have been shadowed in a local scope and thus not be the right value. I've avoided that by generating `NaN` as `0/0` and the infinity values as `1/0` and `-1/0`.

### DIFF
--- a/packages/babel-types/src/converters/valueToNode.js
+++ b/packages/babel-types/src/converters/valueToNode.js
@@ -12,6 +12,8 @@ import {
   arrayExpression,
   objectProperty,
   objectExpression,
+  unaryExpression,
+  binaryExpression,
 } from "../builders/generated";
 
 export default function valueToNode(value: any): Object {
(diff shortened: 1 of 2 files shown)


packages/babel-types/src/converters/valueToNode.js
```diff
@@ -37,7 +39,25 @@ export default function valueToNode(value: any): Object {
...
+      numerator = numericLiteral(0);
+    } else if (value > 0) {
+      // Infinity
+      numerator = numericLiteral(1);
+    } else {
+      // -Infinity
+      numerator = unaryExpression("-", numericLiteral(1));
+    }
+
+    return binaryExpression("/", numerator, numericLiteral(0));
```
nicolo-ribaudo: Why do we need a fraction instead of just `NaN/Infinity/unary(-, Infinity)`?

loganfsmyth:
@nicolo-ribaudo Great point

loganfsmyth:
@nicolo-ribaudo Yeah, because it'll probably break third-party stuff I'm not a huge fan at this point, might be better as the a very early change in 8.x so people have time to fix things.

PR #8007, opened 2018-05-22
Fix 'Missing class properties transform' error when parsing class properties with Typescript syntax

pterolex:
Fixes issue in https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAEBiD29oG8BQ1oCMwCcBc0EALjgJYB2A5qgL5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=false&presets=es2015%2Ctypescript&prettier=false&targets=&version=7.0.0-beta.33&envVersion=7.0.0-beta.36


loganfsmyth:
The way to fix this in line with the Flow strip plugin would be for the Typescript plugin to remove the class properties in the `Class` visitor. That's how we handle this edge case for Flowtype: https://github.com/babel/babel/blob/981bff08e4acebac32369af54c5d8fa3f2793ad4/packages/babel-plugin-transform-flow-strip-types/src/index.js#L75 so all the logic in https://github.com/babel/babel/blob/981bff08e4acebac32369af54c5d8fa3f2793ad4/packages/babel-plugin-transform-typescript/src/index.js#L158 should also be run for all properties inside https://github.com/babel/babel/blob/981bff08e4acebac32369af54c5d8fa3f2793ad4/packages/babel-plugin-transform-typescript/src/index.js#L186

pterolex:
@loganfsmyth Thanks, I've changed the implementation as you suggested.

Now, there's is another issue:  the `babel-plugin-proposal-class-properties` tests broke.
For example, according to the test, the following code 
```js
class C {
    // Output should not use `_initialiseProps`
    x: T;
    y = 0;
    constructor(T) {}
}

```
should be transpiled to 

```js
class C {
  // Output should not use `_initialiseProps`
  constructor(T) {
    this.x = void 0;
    this.y = 0;
  }

}
```
So, the `x` class property is initialized with undefined, but my change removes `x` entirely.

What should be the correct behavior in this case?

P.S. The CI process hung, so these particular tests are not shown

pterolex:
TypeScript compiler removes the unused type declarations inside a class.

https://www.typescriptlang.org/play/#src=class%20C%20%7B%0D%0A%20%20%20%20x%3A%20T%3B%0D%0A%20%20%20%20y%20%3D%200%3B%0D%0A%20%20%20%20constructor(T)%20%7B%7D%0D%0A%7D

hzoo:
Thanks for the PR!

PR #8009, opened 2018-05-23
update AST spec - interpreter

xtuc:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Update our AST spec after https://github.com/babel/babel/pull/7928

### DIFF
--- a/packages/babel-parser/ast/spec.md
+++ b/packages/babel-parser/ast/spec.md
@@ -239,6 +239,7 @@ interface NumericLiteral <: Literal {
 ```js
 interface Program <: Node {
   type: "Program";
+  interpreter: string | null;
   sourceType: "script" | "module";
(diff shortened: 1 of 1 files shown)


packages/babel-parser/ast/spec.md
```diff
@@ -239,6 +239,7 @@ interface NumericLiteral <: Literal {
 ```js
 interface Program <: Node {
   type: "Program";
+  interpreter: string | null;
```
loganfsmyth: The current implementation isn't a string, it's an `InterpreterDirective` node with a `value` string.

xtuc: true, thanks for pointing that out

PR #8016, opened 2018-05-23
Expose the Flow 'all' option on the preset too.

loganfsmyth:
| Fixed Issues?            | 
| Patch: Bug Fix?          | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

The `all` option landed in https://github.com/babel/babel/pull/7934/files#diff-3a8233bcd2766d2c7d87f23f944f7726R3 but it is only exposed from the plugin, not the preset, so this exposes it there too since the flow preset is what we want people to use.

### DIFF
--- a/packages/babel-plugin-syntax-flow/src/index.js
+++ b/packages/babel-plugin-syntax-flow/src/index.js
@@ -7,6 +7,10 @@ export default declare((api, options) => {
   // the @flow pragma was provided.
   const { all } = options;
 
+  if (typeof all !== "boolean" && typeof all !== "undefined") {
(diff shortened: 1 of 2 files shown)


loganfsmyth:
@hzoo Personally I think the fact that it is something we have to do manually is a good thing.

hzoo:
Ok I agree on the manual otherwise we'd pass all options down as you mentioned a while ago; was thinking more of a linting thing, or we should document that we do this somewhere

PR #8021, opened 2018-05-24
Preserve arguments for @callable(...)-style decorators.

benjamn:
| Fixed Issues?            | https://github.com/meteor/meteor/issues/9895

Reported [here](https://github.com/babel/babel/pull/7719#issuecomment-391551435).

Although #7719 updated the parser to produce `decorator.arguments` when a decorator is called with arguments, the legacy `@babel/plugin-proposal-decorators` implementation was not updated to respect this new property.

This fix is loosely modeled after [similar functionality](https://github.com/nicolo-ribaudo/babel/blob/f6239a6c8f1cf3eaf76a66c40de08aa3305d1d1a/packages/babel-plugin-proposal-decorators/src/transformer.js#L33-L41) in the [non-legacy decorators PR](https://github.com/babel/babel/pull/7976) by @nicolo-ribaudo.

### DIFF
--- a/packages/babel-plugin-proposal-decorators/src/transformer-legacy.js
+++ b/packages/babel-plugin-proposal-decorators/src/transformer-legacy.js
@@ -38,13 +38,13 @@ function applyEnsureOrdering(path) {
     : path.get("properties")
   ).reduce((acc, prop) => acc.concat(prop.node.decorators || []), []);
 
-  const identDecorators = decorators.filter(
+  const nonIdentDecorators = decorators.filter(
     decorator => !t.isIdentifier(decorator.callee),
   );
-  if (identDecorators.length === 0) return;
+  if (nonIdentDecorators.length === 0) return;
 
   return t.sequenceExpression(
-    identDecorators
+    nonIdentDecorators
       .map(decorator => {
         const callee = decorator.callee;
         const id = (decorator.callee = path.scope.generateDeclaredUidIdentifier(
@@ -69,7 +69,11 @@ function applyClassDecorators(classPath) {
   const name = classPath.scope.generateDeclaredUidIdentifier("class");
 
   return decorators
-    .map(dec => dec.callee)
+    .map(decorator => {
+      return decorator.arguments
+        ? t.callExpression(decorator.callee, decorator.arguments)
+        : decorator.callee;
+    })
     .reverse()
     .reduce(function(acc, decorator) {
       return buildClassDecorator({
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-proposal-decorators/src/transformer-legacy.js
```diff
@@ -38,13 +38,13 @@ function applyEnsureOrdering(path) {
...
-  const identDecorators = decorators.filter(
+  const nonIdentDecorators = decorators.filter(
     decorator => !t.isIdentifier(decorator.callee),
   );
-  if (identDecorators.length === 0) return;
+  if (nonIdentDecorators.length === 0) return;
 
   return t.sequenceExpression(
-    identDecorators
+    nonIdentDecorators
```
benjamn: This renaming is not a functional change, but it seemed appropriate because this array contains only decorators whose `callee` properties are _not_ `Identifier`s.

benjamn:
I see where you're coming from, and I agree this PR should not be merged until the tests actually fail without the other changes. However, I would note that this change does the right thing when `decorator.arguments` is undefined, by falling back to using `decorator.callee`. Given the potential for parser misconfiguration during this weird temporary transition period, wouldn't it be better if this code worked either way?

Since the `decorator.arguments` property was introduced to support automatic insertion of parentheses (by disambiguating between `@dec()` and `@(dec())`, it may be relevant to note that the TC39 consensus has shifted yet again: per https://github.com/tc39/proposal-decorators/issues/81#issuecomment-384775533, @littledan and others have decided that parentheses should *not* be automatically inserted, and the corresponding spec language has been [reverted](https://github.com/tc39/proposal-decorators/pull/96). In other words, I think the `automaticParentheses` option can be removed from #7976, and the new parser behavior of adding a separate `decorators.arguments` property could perhaps be simplified/removed.

These changes would be safe with or without `decorator.arguments`, and I still think they have short-term value, but of course they will not be necessary if/when `decorator.arguments` goes away.

nicolo-ribaudo:
> Given the potential for parser misconfiguration during this weird temporary transition period

The only possible problem is when someone is using an old version of `@babel/plugin-proposal-decorators`, they only need to update it :wink: 

Anyway, I have opened #8037 to remove `decorator.arguments`.

benjamn:
@nicolo-ribaudo Ok! Sounds like you're on top of this, and this PR won't be useful for very long. I'll find another way to work around the short-term problem folks are having in https://github.com/meteor/meteor/issues/9895.

littledan:
I haven't seen a proposal for decorators with the semantics in this patch. Closing in favor of the no-autoinsertion semantics, as we came to consensus on in the May 2018 TC39 meeting, seems like a good call.

PR #8023, opened 2018-05-24
Add a promise-returning *Async version of the transform and parse fns

loganfsmyth:
| Fixed Issues?            | 
| Documentation PR         | 
| Any Dependency Changes?  |

We've already got `transformFile` and `transformFileSync` for instance, so this adds `transformFileAsync` as a promise-returing version.


packages/babel-core/src/parse.js
```diff
@@ -6,17 +6,79 @@ import normalizeOptions from "./transformation/normalize-opts";
...
+
+export const parse: Parse = (function parse(code, opts, callback) {
+  if (typeof opts === "function") {
+    opts = undefined;
+    callback = opts;
+  }
+
+  // For backward-compat with Babel 7's early betas, we allow sync parsing when
+  // no callback is given. Will be dropped in some future Babel major version.
+  if (callback === undefined) return parseSync(code, opts);
```
nicolo-ribaudo: Q: If we wanted to introduce a warning, would it be a breaking change? Or we could do it in a minor/patch version?

loganfsmyth: I mostly added this to make lives easier for people, I actually thought about removing it in this PR, but since it's so trivial leaving it to 8.x isn't really that much of a problem.

PR #8024, opened 2018-05-24
Run transform-runtime on the standalone bundle so it stays ES5-compatible.

### LINKED ISSUE #7607: standalone does not work in ES5
I'm trying to run @babel/standalone in an ES5 environment. I'm not sure this is target platform or not?

Anyway, one way to try this out is to use a really old version of node (0.10 for example)

### Input Code

```sh
yarn add @babel/standalone
nvm install 0.10
nvm use 0.10
node node_modules/@babel/standalone/babel.js
```

### Expected Behavior

I was hoping this would work. It works on the 6.x branch

### Current Behavior

```sh
$ node node_modules/\@babel/standalone/babel.js

/src/test-babel/node_modules/@babel/standalone/babel.js:6578
var BLOCK_SCOPED_SYMBOL = Symbol.for("var used to be block scoped");
                          ^
ReferenceError: Symbol is not defined
    at Object.<anonymous> (/src/test-babel/node_modules/@babel/standalone/babel.js:6578:27)
    at __webpack_require__ (/src/test-babel/node_modules/@babel/standalone/babel.js:30:30)
    at Object.<anonymous> (/src/test-babel/node_modules/@babel/standalone/babel.js:9308:18)
    at __webpack_require__ (/src/test-babel/node_modules/@babel/standalone/babel.js:30:30)
    at Object.<anonymous> (/src/test-babel/node_modules/@babel/standalone/babel.js:5258:1)
    at __webpack_require__ (/src/test-babel/node_modules/@babel/standalone/babel.js:30:30)
    at Object.exports.__esModule (/src/test-babel/node_modules/@babel/standalone/babel.js:9266:20)
    at __webpack_require__ (/src/test-babel/node_modules/@babel/standalone/babel.js:30:30)
    at Object.<anonymous> (/src/test-babel/node_modules/@babel/standalone/babel.js:8151:38)
    at __webpack_require__ (/src/test-babel/node_modules/@babel/standalone/babel.js:30:30)
```

### Possible Solution
Are we not including the ES6 polyfills?

### Context
Trying to run @babel/standalone in an ES5 engine (nashorn)

loganfsmyth:
| Fixed Issues?            | Refs #7607
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -13,13 +13,18 @@ module.exports = function(api) {
 
   let convertESM = true;
   let ignoreLib = true;
+  let includeRuntime = false;
 
   switch (env) {
     // Configs used during bundling builds.
     case "babel-parser":
+      convertESM = false;
+      ignoreLib = false;
+      break;
     case "standalone":
       convertESM = false;
       ignoreLib = false;
+      includeRuntime = true;
       break;
     case "production":
       // Config during builds before publish.
@@ -41,6 +46,10 @@ module.exports = function(api) {
   }
 
   const config = {
+    // Our dependencies are all standard CommonJS, along with all sorts of
+    // other random files in Babel's codebase, so we use script as the default,
+    // and then mark actual modules as modules farther down.
+    sourceType: "script",
     comments: false,
     ignore: [
       // These may not be strictly necessary with the newly-limited scope of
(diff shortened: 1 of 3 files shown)


babel.config.js
```diff
@@ -84,7 +93,29 @@ module.exports = function(api) {
...
+          "codemods/*/src",
+          "codemods/*/test",
+        ],
+        sourceType: "unambiguous",
+      },
+      {
+        // The runtime transform shouldn't process its own runtime or core-js.
+        exclude: [
+          "packages/babel-runtime",
+          /[\\/]node_modules[\\/](?:@babel\/runtime|babel-runtime|core-js)[\\/]/,
```
nicolo-ribaudo: Aren't paths normalized to always use `/` instead of `\\`? If not `[\\/]` should also be used in `@babel\/runtime`.

loganfsmyth: Good question, maybe I just shouldn't worry about it for now then. And you're right that I missed the slash in the middle anyway.

arv:
I finally had time to test this patch. 

It almost works :-) The only remaining issue is a usage of `String.prototype.repeat` which comes from 
https://github.com/mathiasbynens/jsesc/blob/09de8e3fc85fb3ec77a88376895081cb13a604ed/src/jsesc.js#L80

hzoo:
That one not sure how we would fix since it's a instance method (also a dependency), I think that would be just using babel-polyfill instead of transform-runtime.

loganfsmyth:
I think we should land this as-is, and continue exploring what we're missing for ES5 support. The PR for the corejs 3.x runtime has a bunch of work to handle instance methods if I remember right, so hopefully we can consider switching to that once it lands.

babel.config.js
loganfsmyth: Having looked into this now, the input filename is normalized before validation, so on Windows it would be `\\`. Node does not normalize paths to always be `/`.

PR #8025, opened 2018-05-24
Parenthesized expressions

### LINKED ISSUE #7661: Proposal: Add Closure Compiler Syntax Plugin
Choose one: Feature request

### Input Code

```js
// Closure Compiler input/output
/** @return {T} */
function f() {
   return /** @type {T} */ (someExpression);
}
```

The parentheses are important to Closure Compiler and it defines a Cast expression

https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System#type-casts

https://github.com/google/closure-compiler/blob/79deb15554ae0903b21fa9dc94746a879ae8ef12/src/com/google/javascript/jscomp/parsing/IRFactory.java#L783-L788

### Expected Behavior

Keep the parens on a paren expression after a JSDoc comment.

### Current Behavior

The ParenExpression is "transformed" to an Expression. (Not really transformed but Babel does not output unnecessary parens).

### Possible Solution

Introduce a syntax plugin that detects the case where we have a paren expression preceded by a JSDoc comment with `@type` in it and create a `CcCastExpression` instead. Have the generator for the `CcCastExpression` output the comment, parens and the expression.

### Context

Cannot use Babel to do JSX transformation because we use Closure Compiler for type checking and minimization.

At this time I'm not suggesting adding any other Closure Compiler features. I was not planning on parsing the actual type in the JSDoc but future improvements to this plugin could do that.

arv:
| Fixed Issues?            | Fixes #7661, closes #7015
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | New feature
| Documentation PR         | 

This adds a parser option called `createParenthesizedExpressions`, which when enabled creates `ParenthesizedExpression` nodes instead of setting `extra.parenthesized`.


nicolo-ribaudo:
1) I think that this shouldn't be a Babylon plugin (since plugins are meant to enable new syntax), but it should just be an option (which you already implemented).

2) Can you add a test about `(foo) => {}` works with this new option enabled?

arv:
All done!

1. Removed the parser plugin
2. Added more tests and fixed/improved issues related to lval.

packages/babel-generator/test/fixtures/comments/create-parenthesized-expressions/options.json
```diff
@@ -0,0 +1,3 @@
+{
+  "parserOpts": {"createParenthesizedExpressions": true}
```
arv: This works because I updated the runner in packages/babel-generator/test/index.js to spread the parserOpts into the parser options.

arv:
Rebased on top of https://github.com/babel/babel/commit/2a8ebbe7aeb6e5cd834584d0d83884624afbbebc

arv:
I'm still interested in getting this feature in. We are using Babel more and more in our tools and using a fork is getting old. Any chance of this getting another review?

@hzoo @nicolo-ribaudo

arv:
This is still blocking us. Any suggestions on how to proceed?

CC @rafaelweinstein

jridgewell:
AMP Project also needed this. We solved it by running a custom plugin as the first step in our babel config:

```js
// https://github.com/ampproject/amphtml/blob/2dc43d8df8a899f95348c2149aceddc9ca96b5a0/build-system/babel-plugins/babel-plugin-transform-parenthesize-expression/index.js
module.exports = function(babel) {
  const {types: t} = babel;
  return {
    visitor: {
      Expression: {
        exit(path) {
          const {node} = path;
          const {parenthesized} = node.extra || {};
          if (!parenthesized) {
            return;
          }

          path.replaceWith(t.parenthesizedExpression(node));
          path.skip();
        },
      },
    },
  };
};
```

A babel-parser config option would be welcome.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1166,10 +1178,21 @@ export default class ExpressionParser extends LValParser {
...
+    if (!this.options.createParenthesizedExpressions) {
+      this.addExtra(val, "parenthesized", true);
+      this.addExtra(val, "parenStart", startPos);
+      return val;
+    }
 
-    return val;
+    const parenExpression = this.startNodeAt(startPos, startLoc);
+    parenExpression.expression = val;
+    this.finishNodeAt(
```
nicolo-ribaudo: Isn't using `this.finishNode` enough? The end position should be the correct one at this point I think.

packages/babel-parser/test/fixtures/core/uncategorised/556/input.js
```diff
@@ -0,0 +1 @@
+({x}) = {x: 1};
```
nicolo-ribaudo: Could you also test `([ a ]) = []`?

packages/babel-parser/src/parser/statement.js
```diff
@@ -366,13 +366,20 @@ export default class StatementParser extends ExpressionParser {
     return this.finishNode(node, "DebuggerStatement");
   }
 
+  parseHeaderExpression(): N.Expression {
+    this.expect(tt.parenL);
+    const val = this.parseExpression();
+    this.expect(tt.parenR);
+    return val;
+  }
```
danez: What is the purpose of this? This is an exact duplicate of `parseParenExpression()`

packages/babel-parser/test/fixtures/core/uncategorised/556/input.js
arv: Done

packages/babel-parser/src/parser/statement.js
danez: `parse*Expression()` should not be in `statement.js`, as it is about expressions.

I also cannot fully understand the idea behind that name, what does `Header` mean in this context?
`parseParenExpression`, was pretty clear to me -> an expression inside of parens.

nicolo-ribaudo: It is only used to parse statements headers.
I like the new name, since with the old name it can be confused with something which produces a `ParenthesizedExpression`.

danez: Ok I see.

packages/babel-parser/src/options.js
```diff
@@ -19,6 +19,7 @@ export type Options = {
   strictMode: ?boolean,
   ranges: boolean,
   tokens: boolean,
+  createParenthesizedExpressions: boolean,
```
nicolo-ribaudo: This also needs to be added to `babel-parser/typings/index.d.ts`

packages/babel-parser/src/parser/statement.js
arv: That was exactly my point. When I first started looking into this I was expecting parseParenExpression to return a ParenExpression expression node. Later I learned that it is called ParenthesizedExpression and that the parser never produced these. This was my attempt at cheating this.

Another name could be parseControlFlowHeaderParenthesizedExpression or some combination of that.

packages/babel-parser/src/options.js
arv: I assume you mean [packages/babel-parser/typings/babel-parser.d.ts](https://github.com/babel/babel/blob/master/packages/babel-parser/typings/babel-parser.d.ts)

arv:
Thanks @danez

nicolo-ribaudo:
We were waiting to decide to release v7.4 😛

jridgewell:
Whoops, didn't see the milestone.

nicolo-ribaudo:
Nvm, it will probbaly be the next release anyway

PR #8031, opened 2018-05-24
Validate @babel/preset-env options

### LINKED ISSUE #7445: Usability issue: @babel/preset-env ignores errors in targets spec
@babel/preset-env silently ignores errors in the `targets` spec in several places ([sample repo to run this](https://github.com/joliss/babel-preset-env-issue)):

```js
        targets: {
          browsers: 59, // silently ignored
          // browsers: "invalid123", // this fails however
          invalid123: "invalid123" // silently ignored
        }
```

```js
        targets: "invalid123" // silently ignored
```

When it encounters such invalid targets, babel-preset-env seems to fall back to some ES3-type baseline target.

Ignoring `browsers: 59` seems to be intentional ([test case](https://github.com/babel/babel/blob/f0049726253178d7d97c6790f03b35628eee86f4/packages/babel-preset-env/test/targets-parser.spec.js#L67)), though I'm not sure why.

This seems like a usability issue to me: When I was trying to figure out on my project whether preset-env did anything at all, my first instinct was to put some invalid strings in there, and I was very confused when nothing changed. I even accidentally misspelled "browsers" at one point. 

I think preset-env should throw an error (ideally) or a warning (if throwing an error is not possible for some reason).

| ---------------- | -------
| Babel            | 7.0.0-beta.40
| node             | 9.5.0
| npm              | 5.6.0
| Operating System | Linux

### LINKED ISSUE #7541: env preset accepts invalid config

```js
{
    "presets": [
      [
        "env",
        {
          "node": "current"
        }
      ]
    ]
}
```

### Expected Behavior
The env preset should validate the config and throw, because "node" is not a valid top-level property.

### Current Behavior
The invalid config is silently ignored and default config is used.

### Context
https://github.com/exercism/ecmascript/issues/400

### Your Environment

babel-preset-env 1.4.0

sun1x:
| Fixed Issues?            | Fixes #7445, Fixes #7541

This PR adds validation for [@babel/preset-env](https://github.com/babel/babel/tree/master/packages/babel-preset-env) options.
Also I've added [js-levenshtein](https://www.npmjs.com/package/js-levenshtein) to display suggestions in case of typo / error for some options.

### DIFF
--- a/packages/babel-preset-env/src/types.js
+++ b/packages/babel-preset-env/src/types.js
@@ -1,27 +1,17 @@
 //@flow
 
+import { TargetNames, ModulesOption, UseBuiltInsOption } from "./options";
+
 // Targets
-export type Target =
-  | "esmodules"
-  | "node"
-  | "browsers"
-  | "chrome"
-  | "opera"
-  | "edge"
-  | "firefox"
-  | "safari"
-  | "ie"
-  | "ios"
-  | "android"
-  | "electron";
+export type Target = $Keys<typeof TargetNames>;
 export type Targets = {
   [target: Target]: string,
 };
 
 // Options
 // Use explicit modules to prevent typo errors.
-export type ModuleOption = false | "amd" | "commonjs" | "systemjs" | "umd";
-export type BuiltInsOption = false | "entry" | "usage";
+export type ModuleOption = $Values<typeof ModulesOption>;
+export type BuiltInsOption = $Values<typeof UseBuiltInsOption>;
 
 export type Options = {
   configPath: string,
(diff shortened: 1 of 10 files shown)


packages/babel-preset-env/src/types.js
```diff
@@ -1,9 +1,21 @@
 //@flow
 
 // Targets
-export type Target = string;
+export type Target =
```
existentialism: Wonder if it'd be better if we turned `validTargetNames` into an "enum":

```js
const ValidTargetNames = {
  esmodules: "esmodules",
  // ...etc,
};

// Simplifies Target type into:
export type Target = $Keys<typeof ValidTargetNames>;
```

Also avoids scanning array during validation too?

PR #8032, opened 2018-05-24
Bump flow version

existentialism:
| Fixed Issues?            | 
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-core/src/config/files/configuration.js
+++ b/packages/babel-core/src/config/files/configuration.js
@@ -198,8 +198,9 @@ const readConfigJS = makeStrongCache(
 
 const packageToBabelConfig = makeWeakCache(
   (file: ConfigFile): ConfigFile | null => {
-    if (typeof file.options.babel === "undefined") return null;
-    const babel = file.options.babel;
+    const babel = file.options[("babel": string)];
+
+    if (typeof babel === "undefined") return null;
 
     if (typeof babel !== "object" || Array.isArray(babel) || babel === null) {
       throw new Error(`${file.filepath}: .babel property must be an object`);
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "eslint-config-babel": "^7.0.2",
     "eslint-plugin-flowtype": "^2.20.0",
     "eslint-plugin-prettier": "^2.5.0",
-    "flow-bin": "^0.66.0",
+    "flow-bin": "^0.73.0",
     "graceful-fs": "^4.1.11",
     "gulp": "^4.0.0",
     "gulp-babel": "^8.0.0-beta.2",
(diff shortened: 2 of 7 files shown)


packages/babel-core/src/config/files/types.js
```diff
@@ -3,7 +3,7 @@
 export type ConfigFile = {
   filepath: string,
   dirname: string,
-  options: {},
+  options: Object,
```
existentialism: Still playing around with improving this... but everything so far is quite nasty (json5 returning mixed =/)

loganfsmyth: Does this just start failing with the new Flow version then? I was definitely under the impression everything should refine fully to `{}` with the existing logic.

existentialism: Yeah new from the version bump

loganfsmyth: Hmm, let me look quickly, I'd rather keep these types accurate.

loganfsmyth: The failures is
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ packages/babel-core/src/config/files/configuration.js:202:16

Property babel is missing in object type [1].

     199│ const packageToBabelConfig = makeWeakCache(
     200│   (file: ConfigFile): ConfigFile | null => {
 [1] 201│     const options: {} = file.options;
     202│     if (typeof options.babel === "undefined") return null;
     203│     const babel = options.babel;
     204│
     205│     if (typeof babel !== "object" || Array.isArray(babel) || babel === null) {



Found 1 error
```
it looks like we can change that code to
```
const babel = file.options[("babel": string)];
if (typeof babel === "undefined") return null;
```
to get it to typecheck safely.

packages/babel-core/src/config/files/configuration.js
```diff
@@ -198,8 +198,9 @@ const readConfigJS = makeStrongCache(
 
 const packageToBabelConfig = makeWeakCache(
   (file: ConfigFile): ConfigFile | null => {
-    if (typeof file.options.babel === "undefined") return null;
-    const babel = file.options.babel;
+    const babel = file.options[("babel": string)];
```
nicolo-ribaudo: Does `const babel: string | void = file.options.babel` work? (If it doesn't it's ok - it's just that this seems too hacky to me)

existentialism: Nope :(

```sh
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ packages/babel-core/src/config/files/configuration.js:201:47

Cannot get file.options.babel because property babel is missing in object type [1].

     packages/babel-core/src/config/files/configuration.js
     198│
     199│ const packageToBabelConfig = makeWeakCache(
     200│   (file: ConfigFile): ConfigFile | null => {
     201│     const babel: string | void = file.options.babel;
```

nicolo-ribaudo: Oh nevermind

PR #8036, opened 2018-05-24
Compile against beta 48

hzoo:


package.json
```diff
@@ -10,19 +10,18 @@
...
+    "@babel/preset-env": "7.0.0-beta.48",
+    "@babel/preset-flow": "7.0.0-beta.48",
+    "@babel/preset-stage-0": "7.0.0-beta.48",
+    "@babel/register": "7.0.0-beta.48",
     "babel-core": "^7.0.0-0",
     "babel-eslint": "^8.0.1",
     "babel-jest": "^22.4.1",
     "babel-loader": "8.0.0-beta.0",
     "babel-plugin-transform-charcodes": "^0.1.0",
-    "babylon": "7.0.0-beta.46",
```
hzoo: Not sure this is necessary, don't see the use either atm (and with the rename it's not needed)?

nicolo-ribaudo: I can't find where it was used :woman_shrugging:

PR #8045, opened 2018-05-25
Do not quote JSX attribute keys for IdentifierName

arv:
| Fixed Issues?            |
| Patch: Bug Fix?          | Bug fix
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 

Given the following

```js
a = <F new/>
```

We used to generate:

```js
a = React.createElement(F, {"new": true})
```

but now we generate

```js
a = React.createElement(F, {new: true})
```

If you need to quote these (for ES3  for example, you can use transform-property-literals)

### DIFF
--- a/packages/babel-helper-builder-react-jsx/src/index.js
+++ b/packages/babel-helper-builder-react-jsx/src/index.js
@@ -89,14 +89,14 @@ You can turn on the 'throwIfNamespace' flag to bypass this warning.`,
       }
     }
 
-    if (t.isValidIdentifier(node.name.name)) {
+    if (t.isJSXNamespacedName(node.name)) {
+      node.name = t.stringLiteral(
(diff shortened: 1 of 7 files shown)


nicolo-ribaudo:
Two approved reviews

xtuc:
I'm just wondering why we don't quote everything?

arv:
@xtuc Do you quote all your keys in object literals? In member expressions?

xtuc:
@arv quoting every key would simplify the logic here and it's in the generated code.

arv:
@xtuc The code here is pretty simple and the gain in readability in the generated code makes it worth it.

Plus, quoting keys had some bad side effects for people using Closure Compiler further down their build pipeline.

arv:
Thanks, @jridgewell

PR #8080, opened 2018-05-30
fix/Incorrect parse error thrown for unexpected Flow code

### LINKED ISSUE #7954: Incorrect parse error thrown for unexpected Flow code
## Bug Report

**Current Behavior**
An incorrect parse error is thrown when unexpected Flow code is present.

For the below, I see:

```
SyntaxError: node_modules/some-package/test.js: Support for the experimental syntax 'exportDefaultFrom' isn't currently enabled (3:8):

  1 | /* @flow */
  2 |
> 3 | export type Foo = number;
    |        ^
  4 |

```

**Input Code**
```js
/* @flow */

export type Foo = number;
```

**Expected behavior/code**

N/A (I think?)

**Environment**
- Babel version(s): `7.0.0-beta.46 (@babel/core 7.0.0-beta.46)`
- Node/npm version: v10.0.0
- OS: OSX 10.12.6
- Monorepo: N/A

HassanAli321:
| Fixed Issues?            | Fixes #7954
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | 
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/src/parser/statement.js
+++ b/packages/babel-parser/src/parser/statement.js
@@ -1410,6 +1410,10 @@ export default class StatementParser extends ExpressionParser {
       node.specifiers = [];
       node.source = null;
       node.declaration = this.parseExportDeclaration(node);
+    } else if (this.isExportFlowTypes()) {
+      // export type foo = bar || export type { foo, bar }
+      this.expectPlugin("flow");
+      node.declaration = this.parseExportDeclaration(node);
     } else {
       // export { x, y as z } [from '...']
       node.declaration = null;
@@ -1457,6 +1461,13 @@ export default class StatementParser extends ExpressionParser {
         this.state.start,
         "Only expressions, functions or classes are allowed as the `default` export.",
       );
+    } else if (this.isExportFlowTypes()) {
+      // export default type foo = 2;
+      this.expectPlugin("flow");
+      return this.raise(
+        this.state.start,
+        "Only expressions, functions or classes are allowed as the `default` export.",
+      );
     } else {
       const res = this.parseMaybeAssign();
       this.semicolon();
@@ -1470,7 +1481,12 @@ export default class StatementParser extends ExpressionParser {
   }
 
   isExportDefaultSpecifier(): boolean {
+    const lookahead = this.lookahead();
+
     if (this.match(tt.name)) {
+      if (this.state.value === "type" || this.state.value === "interface") {
+        return lookahead.type === tt.name && lookahead.value === "from";
+      }
       return this.state.value !== "async";
     }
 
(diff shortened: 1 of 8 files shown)


nicolo-ribaudo:
Does this code still work?

```js
var type = 2;
export default type;
```

HassanAli321:
yes the above code "export default type" is allowed and all i need to do is throw an exception that indicates flow is required for the code "export default type Foo = 2" right ?

packages/babel-parser/src/parser/statement.js
```diff
@@ -1471,7 +1482,11 @@ export default class StatementParser extends ExpressionParser {
 
   isExportDefaultSpecifier(): boolean {
     if (this.match(tt.name)) {
-      return this.state.value !== "async";
+      return (
+        this.state.value !== "async" &&
+        this.state.value !== "type" &&
+        this.state.value !== "interface"
+      );
```
nicolo-ribaudo: Here you need some lookahead, because `export type from "foo";` should still work.

packages/babel-parser/test/fixtures/experimental/export-extensions/default-type-without-flow/options.json
```diff
@@ -1,4 +1,5 @@
 {
   "plugins": ["exportDefaultFrom"],
-  "sourceType": "module"
-}
+  "sourceType": "module",
+  "throws": "This experimental syntax requires enabling the parser plugin: 'flow' (1:7)"
```
existentialism: There are three cases we need to make sure we have tests for,

1. `plugins: []`, this should return a syntax error and suggest turning on `exportDefaultFrom`.

1. `plugins: ['exportDefaultFrom']`, this should parse correctly.

1. `plugins: ['flow']`, this should return a syntax error.

HassanAli321: got it

packages/babel-parser/src/parser/statement.js
```diff
@@ -1470,21 +1481,33 @@ export default class StatementParser extends ExpressionParser {
...
-    const lookahead = this.lookahead();
     return (
       lookahead.type === tt.comma ||
       (lookahead.type === tt.name && lookahead.value === "from")
     );
   }
 
+  isExportFlowTypes(): boolean {
+    const next = this.lookahead();
+    if (this.match(tt.name) && next.type === tt.name) {
```
existentialism: Let's perform the lookahead _only_ on a `tt.name` match

PR #8083, opened 2018-05-31
Fixes for the wrapNativeSuper helper

### LINKED ISSUE #8061: Error extending builtin: "ReferenceError: _construct is not defined"
### Current Behavior

A typo was introduced in `@babel/runtime/helpers/wrapNativeSuper.js` in #7533 which results in an error when trying to extend a builtin type: `ReferenceError: _construct is not defined`.

Interestingly, the error only occurs if the external helpers are used (i.e. via `@babel/plugin-transform-runtime`), which suggests that there's a discrepancy between the inlined and external implementations of `wrapNativeSuper` (and other helpers?).

#### Input Code

```js
class Test extends Array {
    foo () {
        return 42
    }
}

const test = new Test()

console.log(test.foo())
```

#### Expected output

```
42
```

#### Actual output

```
./node_modules/@babel/runtime/helpers/wrapNativeSuper.js:28
      return _construct(Class, arguments, getPrototypeOf(this).constructor);
      ^

ReferenceError: _construct is not defined
    at Test.Wrapper (./node_modules/@babel/runtime/helpers/wrapNativeSuper.js:28:7)
    at new Test (./test.js:24:95)
    at Object.<anonymous> (./test.js:7:14)
    at Module._compile (internal/modules/cjs/loader.js:678:30)
    at Module._compile (./node_modules/pirates/lib/index.js:91:24)
    at Module._extensions..js (internal/modules/cjs/loader.js:689:10)
    at Object.newLoader [as .js] (./node_modules/pirates/lib/index.js:96:7)
    at Module.load (internal/modules/cjs/loader.js:589:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:528:12)
    at Function.Module._load (internal/modules/cjs/loader.js:520:3)
```

#### package.json

```json
{
  "devDependencies": {
    "@babel/core": "7.0.0-beta.49",
    "@babel/node": "7.0.0-beta.49",
    "@babel/plugin-transform-runtime": "7.0.0-beta.49",
    "@babel/preset-es2015": "7.0.0-beta.49",
    "@babel/runtime": "7.0.0-beta.49"
  }
}
```

#### .babelrc

```js
{
    presets: ['@babel/preset-es2015'],
    plugins: ['@babel/plugin-transform-runtime'],
}
```

#### Test

```
$ ./node_modules/.bin/babel-node ./test.js
```

#### Environment
- Babel version(s): v7.0.0-beta.48, v7.0.0-beta.49
- Node/npm version: v10.1.0/npm 5.6.0
- OS: Linux (Arch)
- How you are using Babel: `cli`, `loader`

**Possible Solution**

```diff
--- node_modules/@babel/runtime/helpers/wrapNativeSuper.js
+++ node_modules/@babel/runtime/helpers/wrapNativeSuper.js.new
@@ -25,7 +25,7 @@
     }
 
     function Wrapper() {
-      return _construct(Class, arguments, getPrototypeOf(this).constructor);
+      return construct(Class, arguments, getPrototypeOf(this).constructor);
     }
 
     Wrapper.prototype = _Object$create(Class.prototype, {
```

chocolateboy:
| Fixed Issues?            | Fixes #8061, reverts #7188
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

This PR:
  
1\) fixes a ReferenceError caused by a typo (`_construct` instead of `construct`) in the external `wrapNativeSuper` helper. (The typo doesn't usually cause an error in inline helpers because `_construct` happens to be the default name given to the `construct` helper function.)

2\) fixes caching of the shim/wrapper class in external helpers

Extending builtins in ES5 requires the use of a shim class between the subclass and the builtin. This class can be (and should be) reused for each builtin e.g. extensions of Array should all extend the same "ArrayShim" class:

    MyArray1 < ArrayShim < Array < Object
    MyArray2 < ArrayShim < Array < Object

The reuse of this class is handled by a cache, but the current implementation restricts the scope of this cache to the body of the `wrapNativeSuper` function for ["tree-shakeability"](https://github.com/babel/babel/pull/7188). However, in addition to making the code less clear, this also breaks caching because it relies on the redefinition of the `wrapNativeSuper` function, which (by default) [doesn't work with external helpers](https://github.com/babel/babel/issues/8087#issuecomment-393622260).

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -449,7 +449,7 @@ helpers.construct = () => template.program.ast`
     if (Reflect.construct.sham) return false;
 
     // Proxy can't be polyfilled. Every browser implemented
-    // proxies before or at the same time of Reflect.construct,
+    // proxies before or at the same time as Reflect.construct,
     // so if they support Proxy they also support Reflect.construct.
     if (typeof Proxy === "function") return true;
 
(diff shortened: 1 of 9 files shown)


packages/babel-helpers/src/helpers.js
```diff
@@ -449,7 +449,7 @@ helpers.construct = () => template.program.ast`
     if (Reflect.construct.sham) return false;
 
     // Proxy can't be polyfilled. Every browser implemented
-    // proxies before or at the same time of Reflect.construct,
+    // proxies before or at the same time as Reflect.construct,
     // so if they support Proxy they also support Reflect.construct.
     if (typeof Proxy === "function") return true;
```
zloirock: > Every browser implemented proxies before or at the same time as `Reflect.construct`, so if they support `Proxy` they also support `Reflect.construct`.

Definitely not. But it's not related this PR.

nicolo-ribaudo: Which browsers support Proxy but not Reflect.construct?

zloirock: For example, FF18?

nicolo-ribaudo: Oh ok, it's the other way. If a browser supports Reflect.construct, it also supports Proxy.

chocolateboy:
@nicolo-ribaudo [Done](https://github.com/babel/babel/pull/8100), thanks.

PR #8113, opened 2018-06-04
Change decoratorsBeforeExport default to false

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Ref: https://github.com/babel/babel/pull/7976#discussion_r192599893 (cc @diervo)

I also asked to Daniel, and he said that this option could be removed but

> I am not 100% sure, we will have to confirm, but it could always be added back

Instead of removing the option, this PR changes the default value to false.

### DIFF
--- a/packages/babel-parser/src/parser/statement.js
+++ b/packages/babel-parser/src/parser/statement.js
@@ -615,7 +615,9 @@ export default class StatementParser extends ExpressionParser {
 
     const kind = this.state.type.isLoop
       ? "loop"
-      : this.match(tt._switch) ? "switch" : null;
+      : this.match(tt._switch)
+        ? "switch"
+        : null;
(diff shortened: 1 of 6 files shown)


packages/babel-parser/src/parser/statement.js
```diff
@@ -263,8 +263,7 @@ export default class StatementParser extends ExpressionParser {
         this.raise(
           this.state.start,
           "Using the export keyword between a decorator and a class is not allowed. " +
-            "Please use `export @dec class` instead, or set the " +
-            "'decoratorsBeforeExport' option to true.",
+            "Please use `export @dec class` instead.",
```
nicolo-ribaudo: Since it looks like that the propoal is going to change, I removed the suggestion of using that option.

PR #8122, opened 2018-06-06
 isConstantExpression should return true for immuable bindings

### LINKED ISSUE #8110: [plugin-proposal-class-properties] Unnecessary variable is injected for defined Symbol property name
## Bug Report

**Current Behavior**
I'm working with code where is a lot of private variables are defined with Symbols. Transpiling them with Babel I found the problem that `@babel/plugin-proposal-class-properties` for some reason creates another variables for every symbol I use to define my fields.

**Input Code**
[Link](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYewdgzgLgBAZgSwKYBsAmMC8MDKBPAWwCMQUAKAckVTQoEoBuAKCeBQEMIIYBBGAbyYBIANrV0AXSwwAjAAZmAXyA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=es2015%2Ces2016%2Ces2017%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3&prettier=false&targets=&version=7.0.0-beta.49&envVersion=7.0.0-beta.49)
```js
const field = Symbol('field');

class A {
  [field] = 10;
}
```

**Expected behavior/code**
Expect output to not have additional `var _field = field` line because it is unnecessary.

```js
{
  "plugins": [
    ["@babel/plugin-proposal-class-properties", {"loose": true}],
  ],
  "presets": [
    ["@babel/preset-env", {
      "modules": false,
      "shippedProposals": true,
      "targets": {
        "browsers": "last 1 chrome version",
      },
    }],
  ],
}
```

**Environment**
- Node/npm version: node v10.3.0/ npm v6.0.1
- OS: Windows 10
- Monorepo: Lerna 
- How you are using Babel: `rollup-plugin-babel@0.59.1`

**Possible Solution**

Output code:
```js
"use strict";

function _instanceof(left, right) { if (right != null && typeof Symbol !== "undefined" && right[Symbol.hasInstance]) { return right[Symbol.hasInstance](left); } else { return left instanceof right; } }

function _classCallCheck(instance, Constructor) { if (!_instanceof(instance, Constructor)) { throw new TypeError("Cannot call a class as a function"); } }

var field = Symbol('field');
var _field = field; // the unnecessary line 

var A = function A() {
  _classCallCheck(this, A);

  _defineProperty(this, _field, 10);
};
```

I'm not sure if this issue is connected with #6916

nicolo-ribaudo:
| Fixed Issues?            | Fixes #8110
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

The tests in the class-properties plugin were not run because they weren't in a subfolder. They look deleted and created just because I both renamed and regenerated them.


packages/babel-traverse/src/path/introspection.js
```diff
@@ -402,10 +402,7 @@ export function _resolve(dangerous?, resolved?): ?NodePath {
 export function isConstantExpression() {
   if (this.isIdentifier()) {
     const binding = this.scope.getBinding(this.node.name);
-    if (!binding) {
-      return false;
-    }
-    return binding.constant && binding.path.get("init").isConstantExpression();
+    return binding && binding.constant;
```
xtuc: I'm personally not a fan of that, I would prefer something like:

```js
if (!binding) { return false }
return binding.constant;
```

xtuc: @nicolo-ribaudo the styling is wrong here, prettier will probably fix it for you?

nicolo-ribaudo: Actually prettier accepts it. I think that it is ok to have a single statement on the sam line of the if.

PR #8130, opened 2018-06-08
Merge class features plugins

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

cc @loganfsmyth @jridgewell I tried to move plugin-proposal-class-properties to the new "big" plugin. All the original tests are still passing :tada: 

I split the class properties transformation logic in different functions, so that it can be reused. For example, to transform decorators we will need to reuse `buildPrivateNamesMap`, `buildPrivateNamesNodes`, `transformPrivateNamesUsage` and `injectInitialization`.

---

**Questions:** (next meeting?)
- [x] Should we also merge `@babel/plugin-transform-classes`?
  - **Pro**: It makes it a lot easier to reuse the class IIFE. It can be accomplished anyway, but in an uglier way ([#6993](https://github.com/babel/babel/pull/6963/files#diff-94bf3502cb4eafb89afa7ae0ae328994)).
  - **Con**: What will we do if/when we will implement the [classes-1.1 proposal](https://github.com/zenparsing/js-classes-1.1)? It doesn't fit with this `@babel/plugin-proposal-enhanced-classes` plugin because every feature is not compatible, so we would probably need to create another plugin which depends on this one with only class transpilation enabled.
  - NOTE: This can be done later.
- [x] What will we do when a proposal implemented by this plugin becomes part of preset-env?
  - Should we just remove `-proposal-` from this plugin's name now? e.g. `@babel/plugin-enhanced-classes` or `@babel/plugin-transform-enhanced-classes`.
- [x] Should it be possible to enable `loose` for the different features separately?
  - This would probably make sense only if we merge the class transform plugin.
  - NOTE: This can be added later if we don't merge the classes transform. Otherwise, it will be a breaking chance because both `transform-classes` and `proposal-class-properties` can be loose separately (this can be probably worked-around though).
- [x] How much specific the options should be?
  - One option per proposal? (`instanceFields`, `staticFieldsAndMethods`, `instancePrivateMethods`?
  - Make a "grid"? (`instance`, `static` - `public`, `private` - `fields`, `methods`, `accessors`)
  - Very specific? (`publicInstanceFields`, `privateStaticMethods`, ...)
  - NOTE: Options can be added/divided/merged later.
- [x] Name bikeshedding: `@babel/plugin-proposal-enhanced-classes`? `@babel/plugin-proposal-class-features`?

---

**NOTE**: This PR only "merges" the class properties plugin, but it creates the whole infrastructure.


packages/babel-plugin-proposal-enhanced-classes/README.md
```diff
@@ -0,0 +1,19 @@
+# @babel/plugin-proposal-enhanced-classes
```
Andarist: it's not obvious for me how this is supposed to be used (by looking into README), could you add some simple example here?

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -1,436 +1,22 @@
...
-          );
-        } else if (!path.node.id) {
-          // Anonymous class declaration
-          path.node.id = ref;
-        }
-
-        path.insertBefore(computedNodes);
-        path.insertAfter(staticNodes);
-      },
+    inherits(babel) {
```
Andarist: can this be used in combination with enhanced classes? or is it using enhanced-classes or class-properties (where latter is ofc just a thin config wrapper over former)? could we validate misusage somehow?

Andarist:
Whats the story now behind enhanced-classes and transform-classes? Have u considered merging them too? Atm it's hard to do some transformations which requires coordination between transform-classes & class-properties - would love to solve this.

Andarist:
Also - what will happen once i.e. class properties get standardised? Will it stay as part of `enhanced-classes`?

Kovensky:
I suspect they have to stay because of decorators. Though it could be possible to add a mode to make the plugin not transform properties if the class isn't decorated.

Andarist:
> I suspect they have to stay transformed because of decorators.

I would suggest to merge this with `transform-classes` to avoid such ambiguity altogether.

packages/babel-plugin-proposal-enhanced-classes/README.md
nicolo-ribaudo: I can't :stuck_out_tongue: (https://github.com/babel/babel/pull/8108)

I will open a new PR containing the docs when this PR is more ready

packages/babel-plugin-proposal-class-properties/src/index.js
nicolo-ribaudo: `class-properties` is meant to be a "switch" which turns on cass fields in the `enhanced-classes` plugin. If `enhanced-classes` hasn't already been added by the user, `class-properties` will do it.

nicolo-ribaudo:
> Whats the story now behind enhanced-classes and transform-classes? Have u considered merging them too? Atm it's hard to do some transformations which requires coordination between transform-classes & class-properties - would love to solve this.

I didn't do it yet because classes have been part of JS for a long time, while in this plugin I wanted to merge the proposals about class fields, private methods, static class features and decorators (which are considered like a "family" of proposal, since they depend on each other). We could do it if it simplifies classes transformation.

> Also - what will happen once i.e. class properties get standardised? Will it stay as part of enhanced-classes?

This plugin would stay as-is, maybe we will remove the `classFields` option and it will only be possible to enable it using the `@babel/plugin-transform-class-fields` switcher.

Andarist:
Either way is fine to me (although I believe it would make certain transformations easier if we would just merge all class-related transforms into a single plugin). I feel like there should be a clear path defined what will happen in the future regarding those. I think we might be pretty certain that all of those (class fields, private methods, decorators) will get to the language - so it's only a matter of time. And knowing that let us define now what should be done in the future in regards of those transforms. We kinda have an opportunity here to avoid users being confused in the future by moved parts etc.

packages/babel-plugin-proposal-class-properties/src/index.js
Andarist: But can they be used side by side? Would `enhanced-classes` turned on by `class-properties` be "merged" with "standalone" `enhanced-classes`? In other words - is `enhanced-classes`' instance shareable among such invocations?

Andarist:
Im going to discuss further here - because it's imho annoying to reply under "hidden" section of comments.

From what I understand - `loose` is shared between all of the options here, right? Would be cooler to let users specify it for each feature separately.

nicolo-ribaudo:
> From what I understand - loose is shared between all of the options here, right? Would be cooler to let users specify it for each feature separately.

Yeah, we can do something like this:
```js
["@babel/plugin-proposal-enhanced-classes", {
  "loose": true, // Applies to all the features, but it can be overwritten
  "privateMethods": true, // loose, as specified before
  "instanceFields": { "loose": false } // spec compliant
}]
```

Andarist:
If it complicates things too much, then I guess this is not a blocker of any kind. I was thinking about the idea of merging this with regular class transform and in such situation it's easier to imagine a situation where 1 would like to use spec classes but i.e. loose class fields. 

What do u think anyway about the idea of merging them?

Andarist:
>  Should we also merge @babel/plugin-transform-classes?

👍 IMHO it should be done, it simplifies transformations around classes.

> Con: What will we do if/when we will implement the classes-1.1 proposal? It doesn't fit with this @babel/plugin-proposal-enhanced-classes plugin because every feature is not compatible, so we would probably need to create another plugin which depends on this one with only class transpilation enabled.

Couldnt it be just another option of the plugin that couldnt be used in combination with other options?

>  Should it be possible to enable loose for the different features separately?

Would be a good feature in places where it makes sense.

>  How much specific the options should be?

It seems sane to me to just go with 1 option per proposal.

nicolo-ribaudo:
Rebased to include #8463 #8205, #8614 and #8620

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/native-classes/output.js
```diff
@@ -18,13 +18,12 @@ class Foo {
...
 var _foo = babelHelpers.classPrivateFieldLooseKey("foo");
 
+var _bar = babelHelpers.classPrivateFieldLooseKey("bar");
+
 Object.defineProperty(Foo, _foo, {
   writable: true,
   value: "foo"
 });
-
-var _bar = babelHelpers.classPrivateFieldLooseKey("bar");
```
nicolo-ribaudo: This is the only change in the output, because I merged the logic for static and instance properties, but it shouldn't be observable.

packages/babel-plugin-proposal-enhanced-classes/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@babel/plugin-proposal-enhanced-classes",
+  "version": "7.0.0",
+  "author": "The Babel Team (https://babeljs.io/team)",
```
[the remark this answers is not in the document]
nicolo-ribaudo: Well, there is the `"contributors"` field for that (https://docs.npmjs.com/files/package.json#people-fields-author-contributors). I think that `"author"` shouldn't be whoever wrote some parts of the code, but who is responsible for it.

packages/babel-plugin-class-features/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@babel/plugin-class-features",
+  "version": "7.0.0",
+  "author": "The Babel Team (https://babeljs.io/team)",
+  "license": "MIT",
+  "description": "Compile class public and private fields, private methods and decorators to ES6",
+  "repository": "https://github.com/babel/babel/tree/master/packages/babel-plugin-enhanced-classes",
```
Andarist: `enhanced-classes` -> `classes-features` everywhere

packages/babel-plugin-class-features/src/decorators.js
```diff
@@ -0,0 +1,3 @@
+export function hasDecorators(path) {
+  return path.node.decorators && path.node.decorators.length;
```
Andarist: you could cast it to boolean here

packages/babel-plugin-class-features/src/fields.js
```diff
@@ -0,0 +1,315 @@
...
+
+export function buildPrivateNamesNodes(privateNamesMap, loose, state) {
+  const initNodes = [];
+
+  const looseHelper = () => state.addHelper("classPrivateFieldLooseKey");
+
+  for (const [name, { id, static: isStatic }] of privateNamesMap) {
+    if (loose) {
+      initNodes.push(
+        template.statement.ast`var ${id} = ${looseHelper()}("${name}")`,
```
Andarist: for my taste it would be better to inline `state.addHelper("classPrivateFieldLooseKey")` here, but do whatever u see fit

packages/babel-plugin-class-features/src/fields.js
```diff
@@ -0,0 +1,315 @@
...
+  return privateNamesMap;
+}
+
+export function buildPrivateNamesNodes(privateNamesMap, loose, state) {
+  const initNodes = [];
+
+  const looseHelper = () => state.addHelper("classPrivateFieldLooseKey");
+
+  for (const [name, { id, static: isStatic }] of privateNamesMap) {
+    if (loose) {
```
Andarist: i suspect this is not even part of this PR and those parts are just moved from elsewhere, but maybe u could clarify this for me - why static nodes are not handled here in default mode but are handled here in loose mode? maybe adding a comment with some clarification would be helpful

packages/babel-plugin-class-features/src/fields.js
```diff
@@ -0,0 +1,315 @@
...
+) {
+  const staticNodes = [];
+  const instanceNodes = [];
+
+  for (const prop of props) {
+    const isStatic = prop.node.static;
+    const isPrivate = prop.isPrivate();
+
+    // Pattern matching please
+    switch (true) {
```
Andarist: lol, TIL

Andarist: is this really more readable than cascading if/else chain though?

packages/babel-plugin-class-features/test/fixtures/plugin-proposal-class-properties/loose-not-overwritten/options.json
```diff
@@ -0,0 +1,6 @@
+{
+  "plugins": [
+    ["proposal-class-properties", { "loose": true }, "name 1"],
```
Andarist: what is supposed to happen here? I admit I don't quite get it

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -1,524 +1,25 @@
...
-          );
-        } else if (!path.node.id) {
-          // Anonymous class declaration
-          path.node.id = ref;
-        }
-
-        path.insertBefore(computedNodes);
-        path.insertAfter(staticNodes);
-      },
+    manipulateOptions(opts, parserOpts) {
```
Andarist: hm, it's already done in `classFeatures`, so I would expect this to either override classFeatures hook, or duplicate those entries in plugins array, which one is it? does this hook need to be even here if its already specified by classFeatures?

packages/babel-plugin-class-features/src/fields.js
nicolo-ribaudo: I find `switch` more readable than long if/else chains with small bodies, but I don't have anything against using if/else here.

<details>
<summary>switch(true)</summary>

```js
    switch (true) {
      case isStatic && isPrivate && loose:
        staticNodes.push(
          buildPrivateFieldInitLoose(t.cloneNode(ref), prop, privateNamesMap),
        );
        break;
      case isStatic && isPrivate && !loose:
        staticNodes.push(
          buildPrivateStaticFieldInitSpec(prop, privateNamesMap),
        );
        break;
      case isStatic && !isPrivate && loose:
        staticNodes.push(buildPublicFieldInitLoose(t.cloneNode(ref), prop));
        break;
      case isStatic && !isPrivate && !loose:
        staticNodes.push(
          buildPublicFieldInitSpec(t.cloneNode(ref), prop, state),
        );
        break;
      case !isStatic && isPrivate && loose:
        instanceNodes.push(
          buildPrivateFieldInitLoose(t.thisExpression(), prop, privateNamesMap),
        );
        break;
      case !isStatic && isPrivate && !loose:
        instanceNodes.push(
          buildPrivateInstanceFieldInitSpec(
            t.thisExpression(),
            prop,
            privateNamesMap,
          ),
        );
        break;
      case !isStatic && !isPrivate && loose:
        instanceNodes.push(buildPublicFieldInitLoose(t.thisExpression(), prop));
        break;
      case !isStatic && !isPrivate && !loose:
        instanceNodes.push(
          buildPublicFieldInitSpec(t.thisExpression(), prop, state),
        );
        break;
      default:
        throw new Error("Unreachable.");
    }
```

</details>

<details>
<summary>If/else</summary>

```js
    if (isStatic && isPrivate && loose) {
      staticNodes.push(
        buildPrivateFieldInitLoose(t.cloneNode(ref), prop, privateNamesMap),
      );
    } else if (isStatic && isPrivate && !loose) {
      staticNodes.push(buildPrivateStaticFieldInitSpec(prop, privateNamesMap));
    } else if (isStatic && !isPrivate && loose) {
      staticNodes.push(buildPublicFieldInitLoose(t.cloneNode(ref), prop));
    } else if (isStatic && !isPrivate && !loose) {
      staticNodes.push(buildPublicFieldInitSpec(t.cloneNode(ref), prop, state));
    } else if (!isStatic && isPrivate && loose) {
      instanceNodes.push(
        buildPrivateFieldInitLoose(t.thisExpression(), prop, privateNamesMap),
      );
    } else if (!isStatic && isPrivate && !loose) {
      instanceNodes.push(
        buildPrivateInstanceFieldInitSpec(
          t.thisExpression(),
          prop,
          privateNamesMap,
        ),
      );
    } else if (!isStatic && !isPrivate && loose) {
      instanceNodes.push(buildPublicFieldInitLoose(t.thisExpression(), prop));
    } else if (!isStatic && !isPrivate && !loose) {
      instanceNodes.push(
        buildPublicFieldInitSpec(t.thisExpression(), prop, state),
      );
    } else {
      throw new Error("Unreachable.");
    }
```

</details>

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -1,524 +1,25 @@
...
+      parserOpts.plugins.push("classProperties", "classPrivateProperties");
```
Andarist: sideways question - isnt this a single proposal now? shouldnt it be a single parser plugin?

packages/babel-plugin-class-features/test/fixtures/plugin-proposal-class-properties/loose-not-overwritten/options.json
nicolo-ribaudo: When you specify the same plugin two times with different options, only the first set of options is considered because that instance of the plugin runs first.
This test ensures that this behvior is kept even with the merged plugin.

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -1,524 +1,25 @@
...
+    },
 
-      PrivateName(path) {
-        throw path.buildCodeFrameError(`Unknown PrivateName "${path}"`);
-      },
+    pre() {
+      setLoose(this.file, loose);
```
Andarist: and last question - how `loose` is going to work across merged packages? that is if we decide (yes please!) es2015 classes into this one? to keep backwards compatibility they have to work independently

packages/babel-plugin-proposal-class-properties/src/index.js
nicolo-ribaudo: Yeah, we could create a `classFields` plugin which enables both.

packages/babel-plugin-proposal-class-properties/src/index.js
nicolo-ribaudo: When we are using the `@babel/plugin-propsal-class-properties` plugins, `@babel/plugin-class-features` is called without the `fields` option. Thus, it doesn' enable any parser plugin and it needs to be done here.

packages/babel-plugin-class-features/test/fixtures/plugin-proposal-class-properties/loose-not-overwritten/options.json
Andarist: oh, that's why those have ids (name 1/2) on them? otherwise the second one would just override first completely, right?

packages/babel-plugin-class-features/src/fields.js
Andarist: i guess the switch version is more readable indeed, putting aside the weird (to me) shape of this switch/case

packages/babel-plugin-class-features/src/index.js
```diff
@@ -18,11 +18,34 @@ import {
...
+
+  if (value === undefined || value === false) return { enabled: false };
+  if (value === true) return { enabled: true, loose: false };
+
+  if (typeof value === "object") {
+    if (
+      typeof value.loose !== "undefined" &&
+      typeof value.loose !== "boolean"
+    ) {
+      throw new Error(`.${name}.loose must be a boolean.`);
```
Andarist: technically "or undefined", not sure if mentioning it isnt confusing though

packages/babel-plugin-class-features/test/fixtures/plugin-proposal-class-properties/loose-not-overwritten/options.json
nicolo-ribaudo: IIRC Babel 7 throws when a plugin is duplicated in the same config file, unless you give them two different names.

nicolo-ribaudo:
This PR can probably make it easier to fix https://github.com/babel/babel/issues/8750

Andarist:
what's the status of this PR? is this something that rest of the team wants too? or is it still under consideration?

nicolo-ribaudo:
Yeah, we want this PR but it still needs to be reviewed.

Andarist:
Is there anything I could help with?

nicolo-ribaudo:
> Should it be possible to enable loose for the different features separately?

I think that I will move the loose option to `fields: { loose: true }`, so that it specific for each feature. Otherwise we couldn't have, for example, loose class properties with spec-compliant decorators.

@Andarist
> Is there anything I could help with?

You could review this PR. (or, if you already did it, leave a ✔️)

pabloalmunia:
> This PR can probably make it easier to fix #8750

The workaround don't work in my cases. 

Any progress about the support of HTMLElements decoration in this PR?

packages/babel-plugin-class-features/src/features.js
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export function isLoose(file, feature) {
+  return !!(file.get(looseKey) & feature);
+}
+
+export function verifyUsedFeatures(path, file) {
+  if (hasFeature(file, FEATURES.decorators)) {
+    throw new Error(
+      "@babel/plugin-class-features doesn't suport decorators yet.",
```
hzoo:
```suggestion
      "@babel/plugin-class-features doesn't support decorators yet.",
```

and below

packages/babel-plugin-class-features/src/index.js
```diff
@@ -0,0 +1,189 @@
...
+  const decorators = getFeatureOptions(options, "decorators");
+
+  return {
+    manipulateOptions(opts, parserOpts) {
+      if (fields) {
+        parserOpts.plugins.push("classProperties", "classPrivateProperties");
+      }
+    },
+
+    pre() {
```
loganfsmyth: I'm a bit concerned around how some of this will work if there are multiple versions of things. If we eventually add decorators with their own `inherits`, we'll have two separate copies of this plugin running. Would we want to allow both to run? We could potentially have one detect that another instance has already loaded or something, but then we still have the question of _which_ one we end up choosing.

nicolo-ribaudo: For now, only the first plug-in runs (actually both runs, but it is a noop because everything is already transformed).
Probably I should `require("package.json"). version` and only run the most recent version? Or is it overkill?

nicolo-ribaudo: Btw, I will merge the decorators transform before the next release, to check that everything works.

loganfsmyth: The big question for me is if we _want_ plugins like `proposal-class-properties` and `proposal-decorators` at the same time as `babel-plugin-class-features` (which I also notice we may want to be `babel-plugin-proposal-class-features`)?

For instance, we could also deprecate those two plugins, and instead instruct people to use

```
plugins: [
  ["@babel/proposal-class-features", {
    decorators: true,
    publicFields: true,
    privateFields: true,
  }]
]
```
which would more nicely bundle up everything more consistently with our current plugin design.

It's more work for users though, which is why I'm on the fence, but it would mean things are less confusingly side-effectful.

nicolo-ribaudo: My initial idea was to create a single plugin, but different people advocated for keeping the interface as separate plugins.

> The big question for me is if we want plugins like proposal-class-properties and proposal-decorators at the same time as babel-plugin-class-features

Unless people rely on exact versions of the class-features plugin (without `^`) and, at the same time, npm/yarn doesn't dedupe this package, I don't think that using both `class-features` and `class-properties` at the same time would cause confusing effects.
Anyway, we could disallow using them together if you think that it would be better.

> which I also notice we may want to be babel-plugin-proposal-class-features

There will be a time when this plugin will transform standard features and proposals (e.g. when class fields become standard). This was one of the reason that the multi-plugins interface was preferred.

I'm going to split up the actual transform into another PR, so we can keep discussing the interface without blocking https://github.com/babel/babel/issues/8654 and https://github.com/babel/babel/pull/8248

loganfsmyth: > different people advocated for keeping the interface as separate plugins.

Do you have a link to those discussions? I'm concerned that the approach here will make it really hard to maintain and predict how changes might affect users.

> Unless people rely on exact versions of the class-features plugin (without ^) and, at the same time, npm/yarn doesn't dedupe this package, I don't think that using both class-features and class-properties at the same time would cause confusing effects.

Sorry, I should clarify, I don't mean whether they should work together, I mean whether we should maintain both. I'd be tempted to have `babel-plugin-proposal-class-features` be an entire replacement for `proposal-decorators` and `proposal-class-properties`, with those being removed from the monorepo.

> There will be a time when this plugin will transform standard features and proposals (e.g. when class fields become standard). This was one of the reason that the multi-plugins interface was preferred.

There hard part there is that in many cases it's not possible to have one without the other. If we keep things bundled in one plugin, at least it is clear to users that this plugin of class props + decorators + private would likely take recedence, so if say class fields stabilized first, preset-env could default-include

```
['transform-class-features', { publicFields: true }]
```
Which could throw if it encounters classes that have public fields that are decorated, or private fields.

Then if the user adds
```
['proposal-class-features', { decorators: true }]
```
this would essentially override the previous config because it would run first, so it could loudly throw if it encounters public/private fields because they haven't been enabled on that instance of the plugin. Potentially we'd only have to technically error here if a class contained _both_ a decorator and a public field.

nicolo-ribaudo: > Do you have a link to those discussions?

Unfortunately there isn't very much in the meeting notes - the older reference I could find is https://github.com/babel/notes/blob/f58a330d02a2ba7eb29f85dfb88f3c95174d077b/2018/2018-09/sept-10.md#new-class-proposals-decorators-private-fields-etc, which is three months after this PR was opened.

> Sorry, I should clarify, I don't mean whether they should work together, I mean whether we should maintain both.

`class-properties` is only a switch for the `class-features` plugin, it's not something on itself (https://github.com/nicolo-ribaudo/babel/blob/e70d0f03543058ae804bb7814dd7936ae9e0ace7/packages/babel-plugin-proposal-class-properties/src/index.js).

> Then if the user adds
> ```
> ['proposal-class-features', { decorators: true }]
> ```
> this would essentially override the previous config because it would run first, so it could loudly throw if it encounters public/private fields because they haven't been enabled on that instance of the plugin.

I don't think that this is what a user expects: decorators and class fields are different features (even if they need to be implemented together). Why would enabling decorators disable class fields?
Having the separate "switchers" plugin avoids this confusing behavior: if I enable decorators, it doesn't affect class fields being enabled/disabled.

Btw, the fact that you used both `transform` and `proposal` shows that this package's name shouldn't contain either (or should contain both), unless you meant that they should be two separated plugins.

nicolo-ribaudo: If I had to choose, I would disable using the `class-features` plugin directly, since I agree that this configuration would be confusing:

```json
plugins: [
  ["@babel/plugin-class-features", { "classFields": true }],
  ["@babel/plugin-class-features", { "decorators": true }],
]
```

would be surprising if none of them are hidden inside a preset.

packages/babel-plugin-class-features/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+```sh
+npm install --save-dev @babel/plugin-class-features
+```
+
+or using yarn:
+
+```sh
+yarn add @babel/plugin-class-features --dev
+```
```
xtuc: This is auto generated btw, see the script in `scripts`.

packages/babel-plugin-class-features/src/index.js
```diff
@@ -0,0 +1,204 @@
...
+      }
+      if (decorators.enabled) {
+        throw new Error("Decorators are not supported yet");
+        enableFeature(this.file, FEATURES.decorators);
+      }
+    },
+
+    visitor: {
+      Class(path, state) {
+        if (this.file.get(versionKey) !== version) return;
```
loganfsmyth: I think we'll want this check on the `PrivateName` visitor below too.

existentialism:
Great work @nicolo-ribaudo!

nicolo-ribaudo:
Thank you! Merging this PR is so relieving :slightly_smiling_face:

PR #8132, opened 2018-06-08
Fix Safari TP and regular versions comparison

### LINKED ISSUE #8129: Safari TP has lower version then Safari 11 according to getLowestVersion
Safari TP has lower version then Safari 11 according to getLowestVersion
https://github.com/babel/babel/blob/master/packages/babel-preset-env/src/targets-parser.js#L37-L68

All unreleased and released version should be compared together and released version should be < the unreleased. Instead versions are compared with semver that throws when comparing tp and 11 that cause the issue.

yuri-karadzhov:
| Fixed Issues?            | Fixes #8129
| Documentation PR         | 

Fix browser versions comparison issue (Safari TP was < then 11)


packages/babel-preset-env/test/targets-parser.spec.js
```diff
@@ -84,6 +94,41 @@ describe("getTargets", () => {
...
+        }),
+      ).toEqual({
+        android: "4.0.0",
+      });
+    });
+
+    it("ignores invalid", () => {
+      expect(
+        getTargets({
+          browsers: 59,
```
hzoo: I merged this and it's failing on this?

yuri-karadzhov: I will fix it.

yuri-karadzhov: @hzoo The test was removed in previous commits (https://github.com/babel/babel/commit/3de053cc6c94be9411deb213cfc53c1392b5f6eb) and was changed to "throws on invalid browsers target". I will prepare PR that will remove this test.

yuri-karadzhov: https://github.com/babel/babel/pull/8149

PR #8143, opened 2018-06-11
add jsesc options support

### LINKED ISSUE #8124: Feature Request: jsesc option for babel-generator
## Feature Request

I am using babel-generator to generate some codes for ASTs. But I want to customize the `jsesc` option. Currently the option is mostly fixed:

```javascript
// @babel/generator
// src/generators/types.js

// ensure the output is ASCII-safe
const opts = {
  quotes: "double",
  wrap: true,
};
if (this.format.jsonCompatibleStrings) {
   opts.json = true;
}
const val = jsesc(node.value, opts);
```
I want to specific the `opts` variable in the global option of `generate` function. **I will submit a PR if you think it's OK.**

The usage of `generate` function:
```javascript
import {parse} from '@babel/parser';
import generate from '@babel/generator';

const code = 'class Example {}';
const ast = parse(code);

const output = generate(ast, {
  jsescOption: {/* option */},
}, code);
```
And the exist jsesc related option `jsonCompatibleStrings` would be reserved. And if the jsescOption is `null` or `undefined`, it will use default option. And the option object will also be merged to default option.

None.

None.

vincentdchan:
| Fixed Issues?            | Fixes #8124  

### DIFF
--- a/packages/babel-generator/src/index.js
+++ b/packages/babel-generator/src/index.js
@@ -56,7 +56,7 @@ function normalizeOptions(code, opts): Format {
     jsescOption: {
       quotes: "double",
       wrap: true,
-      ...(opts.jsescOption || {}),
+      ...opts.jsescOption,
     },
   };
 
--- a/packages/babel-generator/src/index.js
+++ b/packages/babel-generator/src/index.js
@@ -53,6 +53,11 @@ function normalizeOptions(code, opts): Format {
       base: 0,
     },
     decoratorsBeforeExport: !!opts.decoratorsBeforeExport,
+    jsescOption: {
+      quotes: "double",
+      wrap: true,
+      ...(opts.jsescOption || {}),
+    },
   };
 
   if (format.minified) {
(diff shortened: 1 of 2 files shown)


packages/babel-generator/src/index.js
```diff
@@ -53,6 +53,11 @@ function normalizeOptions(code, opts): Format {
       base: 0,
     },
     decoratorsBeforeExport: !!opts.decoratorsBeforeExport,
+    jsescOption: {
+      quotes: "double",
+      wrap: true,
+      ...(opts.jsescOption || {}),
```
nicolo-ribaudo: `|| {}` isn't needed; object spread works with `null`/`undefined`/`false`

vincentdchan: fixed

hzoo:
I realize that with this change we can just remove the `jsonCompatibleStrings` option? 

Instead users can do `{ jsescOption: { json: true } }`?

vincentdchan:
@hzoo Of cause it can, but it would be a breaking change, I don't know if babel can accept it?

PR #8146, opened 2018-06-11
patch: import Parser as type

yanbingbing:
import Parser as type avoid circular dependency


nicolo-ribaudo:
Is the circular dependency a problem?

yanbingbing:
@nicolo-ribaudo A warning on building

PR #8152, opened 2018-06-12
Use older semver range since lerna update does not change peerDeps on…

hzoo:


codemods/babel-plugin-codemod-object-assign-to-object-spread/package.json
```diff
@@ -13,7 +13,7 @@
     "@babel/plugin-syntax-object-rest-spread": "7.0.0-beta.50"
   },
   "peerDependencies": {
-    "@babel/core": "7.0.0-beta.49"
+    "@babel/core": "^7.0.0-0"
```
loganfsmyth: Would it be worth doing
```
">=7.0.0-beta.0 <7.0.0-rc.0"
```
I realize that's what we were doing before with https://github.com/babel/babel/commit/948f7410677c4ef0561b3c7dcbc71aecd8dca708 before Lerna force-overrode everything.

PR #8169, opened 2018-06-14
babel-plugin-transform-react-jsx: supoprt @jsxLit annotation

mikesamuel:
| Fixed Issues?            | 
| Patch: Bug Fix?          |

This change adds an option to *babel-plugin-transform-react-jsx* with a default that is semantics neutral for existing code.

When `options.pragmaLit` or the `@jsxLit` annotation specifies a function name, the plugin now wraps literal attribute and text node values in a call like `functionName("text", "context hint")`.

For background, ["Options for Hardening React &| JSX"](https://gist.github.com/mikesamuel/094d3fe4b1b2f702f543fa0c263ca8ff) discusses ways to address XSS and crafted intents in JSX Frameworks like React and React native.

> Desugar string literals in JSXAttributeValue and JSXText nodes so that they are clearly marked as specified by a trusted developer.

This should allow `ReactDOM` to prevent `javascript:` URLs that reach

```jsx
(<a href={url}>Link</a>)
```

without preventing developers from doing something like

```jsx
(<a href="javascript:doSomethingAwesome()">Link</a>)
```

This is meant to interoperate with the [trusted-types](https://github.com/WICG/trusted-types) polyfill and other schemes for distinguishing strings from trusted sources at runtime.

### DIFF
--- a/packages/babel-plugin-transform-react-jsx/src/index.js
+++ b/packages/babel-plugin-transform-react-jsx/src/index.js
@@ -11,9 +11,11 @@ export default declare((api, options) => {
 
   const PRAGMA_DEFAULT = options.pragma || "React.createElement";
   const PRAGMA_FRAG_DEFAULT = options.pragmaFrag || "React.Fragment";
+  const PRAGMA_LIT_DEFAULT = options.pragmaLit || null;
 
   const JSX_ANNOTATION_REGEX = /\*?\s*@jsx\s+([^\s]+)/;
   const JSX_FRAG_ANNOTATION_REGEX = /\*?\s*@jsxFrag\s+([^\s]+)/;
+  const JSX_LIT_ANNOTATION_REGEX = /\*?\s*@jsxLit\s+((?:[*]*[^\s*])+)/;
 
   // returns a closure that returns an identifier or memberExpression node
   // based on the given id
@@ -48,8 +50,10 @@ export default declare((api, options) => {
 
       let pragma = PRAGMA_DEFAULT;
       let pragmaFrag = PRAGMA_FRAG_DEFAULT;
+      let pragmaLit = PRAGMA_LIT_DEFAULT;
       let pragmaSet = !!options.pragma;
       let pragmaFragSet = !!options.pragmaFrag;
+      let pragmaLitSet = !!options.pragmaLit;
 
       for (const comment of (file.ast.comments: Array<Object>)) {
         const jsxMatches = JSX_ANNOTATION_REGEX.exec(comment.value);
@@ -62,13 +66,22 @@ export default declare((api, options) => {
           pragmaFrag = jsxFragMatches[1];
           pragmaFragSet = true;
         }
+        const jsxLitMatches = JSX_LIT_ANNOTATION_REGEX.exec(comment.value);
+        if (jsxLitMatches) {
+          pragmaLit = jsxLitMatches[1];
+          pragmaLitSet = true;
+        }
       }
 
       state.set("jsxIdentifier", createIdentifierParser(pragma));
       state.set("jsxFragIdentifier", createIdentifierParser(pragmaFrag));
+      if (pragmaLitSet) {
+        state.set("jsxLitIdentifier", createIdentifierParser(pragmaLit));
+      }
       state.set("usedFragment", false);
       state.set("pragmaSet", pragmaSet);
       state.set("pragmaFragSet", pragmaFragSet);
+      state.set("pragmaLitSet", pragmaLitSet);
     },
     exit(path, state) {
       if (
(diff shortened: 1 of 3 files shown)


mikesamuel:
@sophiebits This follows up on our discussion about mitigating quoting confusion in React* from a few weeks ago.  (Sorry for the delay, was traveling).

mikesamuel:
The Travis failure seems to be unrelated to this PR.  For node@6:

> The command "curl -o- -L https://yarnpkg.com/install.sh | bash" failed and exited with 1 during .

nicolo-ribaudo:
I restarted travis

mikesamuel:
@gaearon  @sebmarkbage  Any idea who might be a good reviewer for this?

sophiebits:
@mikesamuel Sorry, missed this thread. Is there a related proposal to annotate all literal strings in a program? It seems like it would be no less useful for JS in general than for React specifically in JSX.

When we last talked we discussed using https://github.com/WICG/trusted-types as a way of hardening `<a href>` – namely, requiring the `href` value to be `TrustedScriptURL.unsafelyCreate(url)` in cases where a javascript: URL is allowable. (I think I would tend to prefer consistency in this instead of having code break from adding indirection from where a plan string literal previously was.) But I see the proposal has been revised since and it's not as simple as before with the addition of policies. I'm struggling to wrap my head around it fully because it's unclear to me where I'm expected to get a policy. For first-party code do you expect to have a single shared module by convention that exposes a policy (ex: `require('./utils/TrustedTypesPolicy.js')` or similar)? I can see that the idea is you may be able to do DI for different policies when pulling in swaths of less-trusted code but don't understand exactly how those pieces' APIs are imagined to work.

One concrete question stemming from that: what stops me from writing `TrustedTypes.createPolicy('foo', () => {}).createScriptURL()`? Only the CSP header? Is the intention that I'm supposed to generate a random policy name for each one which is put in the CSP header and send them down via a secure channel? Without that it seems like whether the code works or not is heavily dependent on initialization order of your application (in the case that two pieces of code attempt to use the same name).

mikesamuel:
Responding inline and out of order.

> When we last talked we discussed using https://github.com/WICG/trusted-types as a way of hardening `<a href>` – namely, requiring the `href` value to be `TrustedScriptURL.unsafelyCreate(url)` in cases where a `javascript:` URL is allowable.

That matches my recollection, and I'd like to do that too.  I've got approval to sign the FB CLA, and am prepping a patch for ReactDOM that does that.

This feature should be orthogonal to that.

Recognizing literal strings in templates made it much easier for me to rollout contextual autoescaping to large codebases when I did Closure Templates and Go html/template.
I was thinking that React.literal might do the same for existing React projects.

> I'm struggling to wrap my head around it fully because it's unclear to me where I'm expected to get a policy. 

@koto can correct me if I'm wrong.

I believe the intent was to separate two concerns:
*  A library, like a template language, HTML sanitizer, or markdown2html converter, might want to get a string->TrustedType converter, so looking up a policy by name allows that.
*  A product team might trust some libraries but not others or trust with conditions like chain the markdown2html converter through a sanitizer.  Letting early running code define policies so that they also transform provides that.

> One concrete question stemming from that: what stops me from writing `TrustedTypes.createPolicy('foo', () => {}).createScriptURL()`? Only the CSP header? Is the intention that I'm supposed to generate a random policy name for each one which is put in the CSP header and send them down via a secure channel? Without that it seems like whether the code works or not is heavily dependent on initialization order of your application (in the case that two pieces of code attempt to use the same name).

@koto again.  I believe that initialization order does matter.  And name collisions are possible.  Our model assumes good faith among developers, so if both *node_modules/foo* and *node_modules/bar* request a policy named `'foo'`, then developers might become suspicious that *node_modules/bar* is intentionally trying to work around a policy.

https://www.npmjs.com/package/module-keys aims to provide reliable channels between modules but trusted types is not waiting on that.


>  Is there a related proposal to annotate all literal strings in a program? It seems like it would be no less useful for JS in general than for React specifically in JSX.

Like [`@CompileTimeConstant` from Java](https://github.com/google/error-prone/blob/127fbe1a1ccf53cddda693b5c1a68fd388ff3851/annotations/src/main/java/com/google/errorprone/annotations/CompileTimeConstant.java#L71-L73)?

>  Compile-time constant values are implicitly under the control of the trust domain of the application whose source code they are part of. Hence, this annotation is useful to constrain the use of APIs that may only be safely called with values that are under application control.

If so, I know of no current proposals that would provide that kind of guarantee.

IIUC, preventing non-constant values from being used as inputs to certain functions that way requires type system support, so absent a type system, the JavaScript analogue would probably be akin to reverse taint tracking ([paper on tainting across WASM/JS boundaries](https://arxiv.org/pdf/1802.01050.pdf)) or dynamic flow ([multiexecution to prevent leaks](http://oakland10.cs.virginia.edu/slides/multi-execution.pdf#page=12)).

I think there have been flow control proposals for JavaScript but none have made it.

koto:
> I'm struggling to wrap my head around it fully because it's unclear to me where I'm expected to get a policy.

You can create a policy on your own, or accept a preconfigured policy object from the callers when initializing your library. The first approach requires the callers to whitelist the canonical policy you'd create  in CSP (hopefully after they conclude that it's a secure one). The second approach puts the callers completely in control as to how the library may interact with the DOM. 

> For first-party code do you expect to have a single shared module by convention that exposes a policy (ex: require('./utils/TrustedTypesPolicy.js') or similar)? 

A safe policy (i.e. one that sanitizes data) can be exposed like that globally, yes. Then it's accessible by name (`TrustedTypes.getExposedPolicy('default')`), via wrappers `import {p} from 'policy-init.js'` or even called implicitly as a [fallback](https://github.com/WICG/trusted-types/issues/63). 

But we expect the dependencies, especially 3rd party ones to create their own policies, that later on get whitelisted in CSP. Sometimes this is even unavoidable - e.g. an HTML sanitizer library needs to create a allow-all policy to write unsanitized strings to an inert DOM (to later sanitize it). In general, the convention should be:
  * dangerous policies like `(s) => s` are created close to where they're used (e.g. in the same module), and are not exposed in a global namespace (`expose = false`). 
  * sanitizing policies can be created early (e.g. in a bootstrap code), be exposed (`expose = true`) and then may be passed around to chosen module / libraries - either by DI or relying on name convention.

For both of those methods, the application authors are in total control of what policies they allow to be created (and any name collisions are also functional bugs).

mikesamuel:
@sophiebits 
I put up a patch to vet attribute values that this would feed into.

[Proof of concept](https://github.com/facebook/react/pull/13199)
[RFC](https://github.com/reactjs/rfcs/pull/55/files)

PR #8170, opened 2018-06-14
@babel/parser: expose a TypeScript definition file from package

AviVahl:
This allows the new parser package to be used in TypeScript projects, with static type checking and auto-completions. IDEs like VSCode will use this to show auto completions even in JavaScript projects, which is a huge benefit to users wanting to use the parser directly.

This PR follows the auto-generated .d.ts that was added to @babel/types in #7101

- took outdated DefinitelyTyped babylon .d.ts and updated it using new site's documentation.
- added "typings" folder to package, with the new .d.ts
- connected new .d.ts via package.json's "types" field. also adjusted "files" so folder will be published.
- definition file uses TypeScript 2.9's new `import('package').Type` syntax.

### DIFF
--- a/packages/babel-parser/package.json
+++ b/packages/babel-parser/package.json
@@ -15,9 +15,11 @@
   ],
   "repository": "https://github.com/babel/babel/tree/master/packages/babel-parser",
   "main": "lib/index.js",
+  "types": "typings/babel-parser.d.ts",
   "files": [
     "bin",
-    "lib"
+    "lib",
+    "typings"
   ],
   "engines": {
     "node": ">=6.0.0"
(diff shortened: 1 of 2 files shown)


AviVahl:
Hey babel team,
Could anybody check this out?
Would be great to get built-in type checking for the next 7 beta :)

xtuc:
A Flow to TypeScript transpiler would be very usefull, since Babel has the support for both, it should be pretty easy :)

AviVahl:
Thanks for merging!

mgroenhoff:
Is anyone already removing the typings for  `babel__parser` on DefinitelyTyped?
https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/babel__parser

PR #8178, opened 2018-06-15
babel/parser: test helpers: write expected throw message to options.json ~ complete `jest -u` functionality

GerHobbelt:
| Fixed Issues?            | -
| Minor: New Feature?      | 👍 (parser test rig)
| Documentation PR         | -

babel/parser: test helpers: when test fails with an exception and both reference output file and options.json file do not exist, write a simple options.json with the error message to expect when such a file doesn't exist yet. This completes the `jest -u` capability for new tests which are expected to *fail*.

(In my own fork I added tests which are expected to fail and `jest -u` updates the snapshot files quite nicely, yet the error-throwing test cases still had to be updated by hand, which was a chore. This code checks if the options.json exists and it not, writes the exception message in there. This code *only does that when there's no output reference file present*, i.e. this code does *not* patch/write the options.json file when the output reference file (`expects.code`) already tells us this test is **not expected to fail**. This prevents the test runner from acting up when tests actually *fail* *unexpectedly*.

### DIFF
--- a/packages/babel-parser/test/helpers/runFixtureTests.js
+++ b/packages/babel-parser/test/helpers/runFixtureTests.js
@@ -7,11 +7,25 @@ export function runFixtureTests(fixturesPath, parseFunction) {
     fixtures[name].forEach(function(testSuite) {
       testSuite.tests.forEach(function(task) {
         const testFn = task.disabled ? it.skip : it;
+        const fs = require("fs");
+        const path = require("path");
 
         testFn(name + "/" + testSuite.title + "/" + task.title, function() {
           try {
             runTest(task, parseFunction);
           } catch (err) {
+            if (!task.expect.code && !process.env.CI) {
+              const fn = path.dirname(task.expect.loc) + "/options.json";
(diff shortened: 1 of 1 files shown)


packages/babel-parser/test/helpers/runFixtureTests.js
```diff
@@ -7,11 +7,25 @@ export function runFixtureTests(fixturesPath, parseFunction) {
     fixtures[name].forEach(function(testSuite) {
       testSuite.tests.forEach(function(task) {
         const testFn = task.disabled ? it.skip : it;
+        const fs = require("fs");
```
Andarist: would prefer having those on top, mixing cjs/esm requires/imports is imho a bad practice in general (unless ofc rly necessary in some cases)

I think (although not 100% sure) most of babel packages are compiled now with CJS lazy mode, so moving it to top shouldnt even cost initialization time.

hzoo:
Greattt, ditto to what @Andarist said - great workflow prs!

PR #8179, opened 2018-06-15
jest binary is a bash shell script, not a nodeJS file.

GerHobbelt:
| Fixed Issues?            | 👍 (Windows: fail to run jest as bash is *not* default SHELL; invoking jest via `node` should be handled by the active shell, wich *is* bash.)
| Documentation PR         | -
| Any Dependency Changes?  |

Fixes failure to run tests on Windows platform as there bash is not the default shell, hence invoking shell scripts via node is not going to work as expected.

node_modules/.bin/jest *is* a sh/bash shell script, hence invoking that script via the `node` executable is is done without this patch, is begging for trouble on Windows boxes at least and superfluous on unix boxes.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -23,4 +23,4 @@ if [ -n "$TEST_ONLY" ]; then
   jestArgs+=("(packages|codemods)/.*$TEST_ONLY.*/test")
 fi
 
-$node node_modules/.bin/jest "${jestArgs[@]}"
+node_modules/.bin/jest "${jestArgs[@]}"


AviVahl:
@Andarist you are looking at the package itself, which does contain the `.js` entry point (with shebang comment at the top).
`yarn` and `npm` each create a platform-specific executable file in `node_modules/.bin`. On Windows, they generate a `cmd` script, which can't be launched using node.

hzoo:
since it has the shebang yea it doesn't need the `node` call, great!

PR #8183, opened 2018-06-16
[1/2] preset-env: rename useBuiltins to injectPolyfills

### LINKED ISSUE #6830: "useBuiltIns" option in env preset is confusing compared to spread plugin
For context, this is how the object spread plugin currently works:

https://babeljs.io/docs/plugins/transform-object-rest-spread/#optionsuse-built-ins

| useBuiltIns | Result                        |
|-------------|-------------------------------|
| false       | `extends` helper (don't use the built-in)   |
| true        | `Object.assign` (use the built-in) |

This makes sense. Setting `useBuiltIns: true` means we turn off adding extra code in favor of the built-in functionality.

### Expected Behavior
**How the env preset should work**
I think using a different name for this option might lead to less confusion, for example `polyfills`

| polyfills (or some other name) | Result                                                |
|----------|-------------------------------------------------------|
| usage    | Add individual polyfills in every file based on usage |
| entry    | Replace top-level polyfill with individual polyfills  |
| false    | Never apply polyfills                                 |

### Current Behavior

**How the env preset currently works**

| useBuiltIns | Result                                                |
|----------|-------------------------------------------------------|
| usage    | Add individual polyfills in every file based on usage |
| entry    | Replace top-level polyfill with individual polyfills  |
| false    | Never apply polyfills                                 |

In a sense, this is the opposite of how `useBuiltIns` works in the spread plugin, where `false` means adding the `extend` ponyfill.

### Possible Solution
- Use a different name for `useBuiltIns` option in the env preset?

ishitatsuyuki:
| Fixed Issues?            | Fix #6830
| Minor: New Feature?      |
| Tests Added + Pass?      | Not applicable
| Documentation PR         | https://github.com/babel/website/pull/1698

See the linked issue for why this should be renamed.


packages/babel-preset-env/src/index.js
```diff
@@ -230,7 +235,6 @@ export default declare((api, opts) => {
   }
 
   const plugins = [];
-  const pluginUseBuiltIns = useBuiltIns !== false;
```
ishitatsuyuki: TODO: add a default value (determined by injectPolyfills)

xtuc:
We need to update the documentation of the website too

ishitatsuyuki:
@xtuc Done.

ishitatsuyuki:
@hzoo Can you review?

existentialism:
Two thoughts:

1. I'd prefer it if instead of a straight breaking change, we add the new name and add a warning/deprecate the old one.

1. I prefer `polyfills` as suggested in https://github.com/babel/babel/issues/8195

ishitatsuyuki:
@existentialism

1. This does issue a warning when a string is passed into useBuiltins and forwards it to injectPolyfills so technically this does not charge any behaviour.

2. I don't have a strong opinion on this. cc @rtsao what do you think?

existentialism:
@ishitatsuyuki cool, sorry I missed that! can we add a test for that, while we 🚲🏠 on the name?

rtsao:
IMHO `polyfills` is a more name if it supports non-boolean values.

Any thoughts on removing support for the`{useBuiltIns: "entry"}` behavior as proposed in https://github.com/babel/babel/issues/8195? This is probably one of the more controversial changes so I'd love to get feedback.

I think arguably that behavior is bad practice, and furthermore this allows for a nicer `polyfills` option if custom polyfill support is added. Otherwise, I think a 1) separate option will need be required, or 2) it needs to be a complex object along the lines of:

```
{
  polyfills: {mode: "entry", overrides: {"es6.promise": "my-custom-promise-polyfill"}}
}
```

existentialism:
I'd prefer to _not_ remove "entry" in the short-term, as the implementation of "usage" is still pretty naive and hasn't really been worked on much outside of its initial implementation.

I don't think number 2 is terrible, nor is a separate `polyfillOverrides` option.

rtsao:
In that case, I guess 1/2 are basically the same, but 1 is probably easier in terms of migration. It also unblocks this PR.

ishitatsuyuki:
I've renamed the option to `polyfills`; is there anything I should also do?

ishitatsuyuki:
Bump; it has been a while since I last heard from you. Is there anything we need to discuss before merging?

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -150,13 +154,13 @@ export const objectToBrowserslist = (object: Targets): Array<string> => {
   }, []);
 };
 
-export const validateUseBuiltInsOption = (
+export const validateInjectPolyfillsOption = (
```
Andarist: should this be called `validatePolyfillsOption` now?

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -179,6 +183,15 @@ export default function normalizeOptions(opts: Options) {
 
   checkDuplicateIncludeExcludes(include, exclude);
 
+  let polyfills = validateInjectPolyfillsOption(opts.polyfills);
+  let useBuiltIns;
+  if (typeof opts.useBuiltIns === "string") {
+    console.warn("'useBuiltIns' has been renamed to 'polyfills'.");
+    polyfills = validateInjectPolyfillsOption(opts.useBuiltIns);
+  } else {
+    useBuiltIns = opts.useBuiltIns;
```
Andarist: whats the use case for passing `useBuiltIns` further down? Shouldnt it just get converted here to `polyfills`?

packages/babel-preset-env/src/options.js
```diff
@@ -22,7 +23,7 @@ export const ModulesOption = {
   umd: "umd",
 };
 
-export const UseBuiltInsOption = {
+export const InjectPolyfillsOption = {
```
Andarist: this should also be renamed to `PolyfillsOption`

packages/babel-preset-env/src/normalize-options.js
ishitatsuyuki: Pardon me? `useBulitIns` is a new option (has a different functionality) now?

Andarist: My bad. I was under the impression that this PR was only about changing `useBuiltIns` to `polyfills`. Disregard my comment, you are right 👍

ishitatsuyuki:
This is ready again.

loganfsmyth:
I'm not sure where this stands now that 7.x has been released. If we did it, we'd have to do it in a backward-compatible way, at least.

ishitatsuyuki:
It is. Please see the previous discussion.

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -179,6 +179,15 @@ export default function normalizeOptions(opts: Options) {
 
   checkDuplicateIncludeExcludes(include, exclude);
 
+  let polyfills = validatePolyfillsOption(opts.polyfills);
+  let useBuiltIns;
+  if (typeof opts.useBuiltIns === "string") {
+    console.warn("'useBuiltIns' has been renamed to 'polyfills'.");
```
loganfsmyth: I don't think a warning is workable if we're landing this during the lifetime of 7.x. If it were still before the 7.x release maybe, but not during.

loganfsmyth:
Ah sorry, my mistake then.

drinksong:
#8249

nicolo-ribaudo:
In Babel 8 we'll remove `useBuiltIns` in favor of https://github.com/babel/babel-polyfills. Thanks anyway for the PR!

Issue #8182, opened 2018-06-16
Recent betas don't create output full directory hierarchy with --out-dir


loganfsmyth:
## Bug Report / Regression

Reported via https://github.com/twbs/bootstrap/pull/26658#issuecomment-397739198

**Current Behavior**
```
babel src --out-dir other/nested/dir
```
will attempt to create `dir` and throw if `other/nested` doesn't already exist.

**Expected behavior/code**

It should create all parent directories of the output directory. We should probably just replace `outputDestFolder` with `mkdirp`.

**Where did this come from**

This is a regression from the refactoring in https://github.com/babel/babel/pull/7929/files#diff-f93fbf6dce06b5cbf6cc65f5f7dac101L86 that essentially exposed an edge case that was already present but unlikely.

In that PR, I made the code _always_ call `outputDestFolder`, where before it was only called if there were `0` files found to write to the output.

The issue is that where the logic for writing _files_ uses `mkdirp` semantics, `outputDestFolder` uses `mkdir` semantics (no `-p`).

So essentially that PR takes a bug that would previously only have happened if there were no files found, and makes it happen all the time.

himanish-star:
Hey @loganfsmyth, I would like to take up this issue. May I?

loganfsmyth:
@himanish-star Go for it!

PR #8189, opened 2018-06-17
remove regression to support output of full directory hierarchy with --out-dir

himanish-star:
fixes: #8182 

@loganfsmyth, I have made the replacement of `outputDestFolder` with `mkdirp`. However, I have no idea whatsoever of how to try out whether these changes do the job.

One way I had tried to check whether the code changes worked or not was by trying to install the local copy of the newly built `babel-cli` plugin and then type in the command in the shell to see whether it works or not. But I have no idea of how to install this newly built local copy of the `babel-cli` plugin so that I can use it.

P.S. by the local copy of `babel-cli` I mean the one generated after running `make build`

Also the `make build` cmd takes too much of time. Is there any other way to only build the specific files

### DIFF
--- a/packages/babel-cli/src/babel/dir.js
+++ b/packages/babel-cli/src/babel/dir.js
@@ -1,5 +1,6 @@
 import defaults from "lodash/defaults";
 import outputFileSync from "output-file-sync";
+import { sync as mkdirpSync } from "mkdirp";
 import slash from "slash";
 import path from "path";
 import fs from "fs";
@@ -71,19 +72,6 @@ export default async function({ cliOptions, babelOptions }) {
     return path.join(cliOptions.outDir, filename);
   }
 
-  function outputDestFolder(outDir) {
-    const outDirPath = path.resolve(outDir);
-
-    try {
-      fs.mkdirSync(outDirPath);
-    } catch (err) {
-      // Testing for the directory and then creating it can lead to race
-      // conditions if there are multiple processes, so we try to create it
-      // and bail if it already exists.
-      if (err.code !== "EEXIST") throw err;
-    }
-  }
-
   async function handleFile(src, base) {
     const written = await write(src, base);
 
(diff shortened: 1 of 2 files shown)


hzoo:
Yeah good question. At the very least you can use `make build` instead so it only rebuilds the files that you changed https://github.com/babel/babel/blob/master/CONTRIBUTING.md#setup 👍

himanish-star:
@hzoo, sorry to bother you again but I just wanted to share with you the way I tested this feature out so that I could get a confirmation that I am doing things the right way.

1. I used `make build`
2. Then the `lib` folder was created inside the `babel-cli package`.
3. Then I typed
```Shell
$ cd lib/babel
```
4. Then I created a `temp.js` file in the same directory and wrote dummy code inside it and ran the following cmd in the shell
```Shell  
$ node index.js temp.js --out-dir other/nested
```
and it worked :tada: 
image

So is this the right way :sweat_smile: ? Could you also guide me regarding the failing tests of Travis-CI

packages/babel-cli/src/babel/dir.js
```diff
@@ -1,5 +1,6 @@
 import defaults from "lodash/defaults";
 import outputFileSync from "output-file-sync";
+import { sync as mkdirpSync } from "mkdirp";
```
loganfsmyth: Looks like this needs to be added to the `package.json`.

loganfsmyth:
I have no idea why the tests are failing on this. Maybe adding the `package.json` will help.

packages/babel-cli/src/babel/dir.js
himanish-star: Sure I'll do this.

himanish-star:
image
Travis CI error

loganfsmyth:
Thanks!

XhmikosR:
Shouldn't there be a test for this?

loganfsmyth:
@XhmikosR I'd be happy to accept a PR for a test, but it didn't feel necessary to block merging based on that.

XhmikosR:
NP, it's just that personally I'd feel much safer if there was a test :)

himanish-star:
@XhmikosR, if you want I can write a test for this too! Shall I?

PR #8194, opened 2018-06-18
Add missing parameters to ast creators

hendrikniemann:
| Fixed Issues?            | Sorry, didn't create an additional issue
| Patch: Bug Fix?          |
| Major: Breaking Change?  | No _dramatic_ breaking changes (see below)
| Minor: New Feature?      | This will allow users of the package to create nodes with all allowed kinds of fields! ✨ 
| Tests Added + Pass?      | No new tests but I fixed the old ones (see below)

Hey Babel team,

first, thanks for the awesome package _@babel/types_. The package has been making so much progress in the last 6 month. It is so useful when generating code and I rely heavily on it in my project.

A while ago I already fixed a small problem regarding arguments of creator functions (#7697). Some of the creator functions in @babel/types have less arguments than they have fields. This makes it possible to verify nodes of that type but not possible to create new nodes with these properties using the functions from the package.
I added the missing arguments to a bunch of functions by either introducing the `builder` property or adding more values to it. Furthermore I made sure that all additional parameters that are now available for these functions are optional (see exception below). This should prevent them from breaking for users that are relying on these functions.

#### Scope of the change:

I went through the documentation and changed all the functions that have less arguments than fields to match the arguments with the fields. I did not do that for the typescript functions (I tried to keep the scope size comfortable for me) but could maybe do this in another PR later.

~I also removed an optimisation in the `cloneNode` function (first change in the diff). It was only copying the `name` property but _ Identifiers_ can have all sorts of fields these days. My change broke the test because it did not copy over all of the new fields (which are initialised as null when not provided).~ _Edit: Removed from this PR_

#### Breaking changes

As I said I tried to not make any breaking changes. I think two things are technically breaking:
- I added the `name` argument to `TypeParameter` as the first argument (and therefore changed the arguments) but I am not sure how a type parameter without a name would work since the field is required.
- ~In `ArrayPattern` I inserted the `decorators` argument in between the existing arguments to make it consistent in ordering with all the other functions expecting `decorators` and `typeAnnotation`.~ _Edit: Removed from this PR_

### DIFF
--- a/packages/babel-types/src/definitions/core.js
+++ b/packages/babel-types/src/definitions/core.js
@@ -383,6 +383,7 @@ export const patternLikeCommon = {
       assertValueType("array"),
       assertEach(assertNodeType("Decorator")),
     ),
+    optional: true,
   },
 };
 
@@ -538,10 +539,12 @@ defineType("Program", {
   // Note: We explicitly leave 'interpreter' out here because it is
   // conceptually comment-like, and Babel does not traverse comments either.
   visitor: ["directives", "body"],
-  builder: ["body", "directives", "sourceType", "interpreter"],
+  builder: ["body", "directives", "sourceType", "interpreter", "sourceFile"],
   fields: {
     sourceFile: {
       validate: assertValueType("string"),
+      default: null,
+      optional: true,
     },
     sourceType: {
       validate: assertOneOf("script", "module"),
@@ -585,7 +588,7 @@ defineType("ObjectExpression", {
 });
 
 defineType("ObjectMethod", {
-  builder: ["kind", "key", "params", "body", "computed"],
+  builder: ["kind", "key", "params", "body", "computed", "async", "generator"],
   fields: {
     ...functionCommon,
     ...functionTypeAnnotationCommon,
@@ -620,6 +623,7 @@ defineType("ObjectMethod", {
         assertValueType("array"),
         assertEach(assertNodeType("Decorator")),
       ),
+      optional: true,
     },
     body: {
       validate: assertNodeType("BlockStatement"),
--- a/packages/babel-types/src/definitions/es2015.js
+++ b/packages/babel-types/src/definitions/es2015.js
@@ -13,7 +13,7 @@ import {
 } from "./core";
 
 defineType("AssignmentPattern", {
-  visitor: ["left", "right"],
+  visitor: ["left", "right", "decorators", "typeAnnotation"],
   builder: ["left", "right"],
   aliases: ["Pattern", "PatternLike", "LVal"],
   fields: {
@@ -24,17 +24,11 @@ defineType("AssignmentPattern", {
     right: {
       validate: assertNodeType("Expression"),
     },
-    decorators: {
-      validate: chain(
-        assertValueType("array"),
-        assertEach(assertNodeType("Decorator")),
-      ),
-    },
   },
 });
 
 defineType("ArrayPattern", {
-  visitor: ["elements", "typeAnnotation"],
+  visitor: ["elements", "decorators", "typeAnnotation"],
   builder: ["elements"],
   aliases: ["Pattern", "PatternLike", "LVal"],
   fields: {
@@ -45,17 +39,11 @@ defineType("ArrayPattern", {
         assertEach(assertNodeType("PatternLike")),
       ),
     },
-    decorators: {
-      validate: chain(
-        assertValueType("array"),
-        assertEach(assertNodeType("Decorator")),
-      ),
-    },
   },
 });
 
 defineType("ArrowFunctionExpression", {
-  builder: ["params", "body", "async"],
+  builder: ["params", "body", "async", "generator", "expression"],
   visitor: ["params", "body", "returnType", "typeParameters"],
   aliases: [
     "Scopable",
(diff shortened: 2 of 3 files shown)


packages/babel-types/src/definitions/core.js
```diff
@@ -329,7 +329,16 @@ export const functionDeclarationCommon = {
...
 
 defineType("FunctionDeclaration", {
-  builder: ["id", "params", "body", "generator", "async"],
+  builder: [
+    "id",
+    "params",
+    "body",
+    "generator",
+    "async",
+    "declare",
```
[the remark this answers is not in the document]
hendrikniemann: I see your point, some functions' number of arguments really gets out of hand. Another idea could be to include all important (required) fields as arguments and then have another argument where you can have an object that takes the remaining optional fields:

```js
t.identifier("foo", { typeAnnotation: t.anyTypeAnnotation() });
```

This would give us more of an immutable looking interface. I understand that for example in transformers you want to optimise for speed, on the other hand in my current project I use babel to generate complex type definitions and its complexity reduces significantly with a declarative immutable style.

I can maybe reduce this PR to the fixes (e.g. `name` argument of _TypeParameter_) and open a separate issue for the proposal mentioned above where we can discuss. What do you think?

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -34,10 +34,7 @@ export default function cloneNode<T: Object>(node: T, deep: boolean = true): T {
   const { type } = node;
   const newNode = (({ type }: any): T);
 
-  // Special-case identifiers since they are the most cloned nodes.
-  if (type === "Identifier") {
```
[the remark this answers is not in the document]
hendrikniemann: Same here: I will maybe simply create a bug report that this function does not clone all fields of an Identifier node correctly and we discuss there what to do about it. This really just came up because the test case that is testing this function is testing the optimised case and fails if `t.identifier` takes additional (nullable) arguments.

packages/babel-types/src/definitions/core.js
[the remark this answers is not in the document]
existentialism: Separate issue sounds good for the proposal!

hendrikniemann:
I now removed the Flow and Typescript related fields from the builders in `core.js` and `es2015.js`. In `flow.js` they are of course still in place and added fully to the functions when there was a mismatch. Some fields on these flow definitions are a bit weird though, for example I don't know how an `ObjectTypeCallProperty` can be static.

The optimisation for identifier cloning is back in place, I will put this in a separate issue.

packages/babel-types/src/definitions/core.js
```diff
@@ -538,10 +539,12 @@ defineType("Program", {
   // Note: We explicitly leave 'interpreter' out here because it is
   // conceptually comment-like, and Babel does not traverse comments either.
   visitor: ["directives", "body"],
-  builder: ["body", "directives", "sourceType", "interpreter"],
+  builder: ["body", "directives", "sourceType", "interpreter", "sourceFile"],
```
loganfsmyth: I don't think anyone knows this field is even there and I'm not even sure off the top of my head what it is for, so I'd rather leave this out.

packages/babel-types/src/definitions/es2015.js
```diff
@@ -69,8 +57,9 @@ defineType("ArrowFunctionExpression", {
     ...functionCommon,
     ...functionTypeAnnotationCommon,
     expression: {
```
loganfsmyth: This is actually stale code. We removed this field in #6773, we should drop this too.

packages/babel-types/src/definitions/es2015.js
```diff
@@ -13,7 +13,7 @@ import {
 } from "./core";
 
 defineType("AssignmentPattern", {
-  visitor: ["left", "right"],
+  visitor: ["left", "right", "decorators", "typeAnnotation"],
```
loganfsmyth: Usually visitors try to run in code order, and I _think_ in this context, `decorators` is essentially a prefix on the assignment. Same for type annotations being attached to the left item, so this should probably be

    ["decorators", "left", "typeAnnotation", "right"],

packages/babel-types/src/definitions/es2015.js
```diff
@@ -24,17 +24,11 @@ defineType("AssignmentPattern", {
...
-        assertValueType("array"),
-        assertEach(assertNodeType("Decorator")),
-      ),
-    },
   },
 });
 
 defineType("ArrayPattern", {
-  visitor: ["elements", "typeAnnotation"],
+  visitor: ["elements", "decorators", "typeAnnotation"],
```
loganfsmyth: Same here for decorators being first.

packages/babel-types/src/definitions/es2015.js
```diff
@@ -471,7 +471,7 @@ defineType("ClassMethod", {
 });
 
 defineType("ObjectPattern", {
-  visitor: ["properties", "typeAnnotation"],
+  visitor: ["properties", "decorators", "typeAnnotation"],
```
loganfsmyth: Same here about being first.

loganfsmyth:
I'm triaging old PR on Babel. Since this has been sitting for ages, I'm going to close it. Feel free to re-open though, if you have a chance to address my comments.

Issue #7770, opened 2018-04-21
Should we remove Stage 0 - 2 presets, leaving only stage 3?


loganfsmyth:
EDIT: PR at https://github.com/babel/babel/pull/8293
Stage-0 Readme: https://github.com/babel/babel/blob/48fe688a10d83486defdda223ac8d8d88057975d/packages/babel-preset-stage-0/README.md#babelpreset-stage-0

I think there's a reasonable argument to be made for stage 3 features being easily toggle-able, and by that point they are less likely to have major breaking changes, so I'm not as against it. We've had occasional conversations, but I think we should decide.

The stage < 3 presets are likely to change over time, making versioning difficult for them, and the community has grown to overactively enable stage-0 which I see as a dangerous precedent. To me it seems like we should remove these presets and instead use documentation to encourage people to enable experimental plugins on a case-by-case basis for the ones they _actually_ want, rather than just enabling everything for the sake of it.

I just want to make sure we have this conversation while we have the easy opportunity of dropping them before `7.x`.

loganfsmyth:
I'll add that the new `decoratorsLegacy` option for `stage-2` to toggle decorator behavior is a good example of why they are annoying right now. If people aren't even using decorators in the first place, we still currently force them to choose which version of decorators they will want.

nicolo-ribaudo:
Also, people shouldn't be required to know what stages are and which stage a particular feature is.

ljharb:
Nobody should be using anything prior to stage 3 in production, so i think removing them would be excellent.

Those who want to help proposals advance, and experiment, can handle whatever configuration burden ends up being necessary.

ljharb:
The year-specific presets match what's actually in the spec, and are still useful (altho using `env` is better).

jridgewell:
We definitely should keep yearly presets, though you should be using `env`.

👍to killing Stage 0-2.

NullDivision:
I understand the motivation behind this discussion, however I think what most of the people who are calling to remove stages 0 through 2 (most likely because of an inability to manage project specs) fail to realize is that stages are not code switches.

Stage-X is a preset. A group of plugins. They are not a part of Babel but of TC39 specs. All Babel is doing is creating a copy of plugins for that spec.

Removing the presets doesn't prevent usage of the underlying plugins. It prevents visibility. Say you remove stages 0-2. A startup comes along and feels like using non-standards plugins is fine. They will use it.

[...]

PR #8196, opened 2018-06-19
Require proposal flag for pipeline plugin

mAAdhaTTah:
| Fixed Issues?            | https://github.com/babel/proposals/issues/29

I'm opening this now to get feedback early. I'm not sure if the logic should live in the proposal plugin, the syntax plugin, or the parser? (or all three?)

Is `proposal` an acceptable flag name?

Also, this breaks the current stage-0 plugin as well, which I don't know if you're going to keep. Is that going to be an issue?

* [x] Implement validation logic in all required places
* [x] Update README

### DIFF
--- a/packages/babel-plugin-proposal-class-properties/test/fixtures/public-loose/foobar/options.json
+++ b/packages/babel-plugin-proposal-class-properties/test/fixtures/public-loose/foobar/options.json
@@ -1,7 +1,7 @@
 {
   "plugins": ["external-helpers", ["proposal-class-properties", {"loose": true}]],
   "presets": [
-    ["stage-0", { "decoratorsLegacy": true }],
+    ["stage-0", { "decoratorsLegacy": true, "pipelineProposal": "minimal" }],
     "es2015"
   ]
 }
--- a/packages/babel-plugin-syntax-pipeline-operator/src/index.js
+++ b/packages/babel-plugin-syntax-pipeline-operator/src/index.js
@@ -1,11 +1,17 @@
 import { declare } from "@babel/helper-plugin-utils";
 
-export default declare(api => {
+const proposals = ["minimal"];
+
+export default declare((api, { proposal }) => {
   api.assertVersion(7);
+
+  if (typeof proposal !== "string" || !proposals.includes(proposal)) {
+    throw new Error("'proposal' must be one of: " + proposals.join(", "));
+  }
 
   return {
     manipulateOptions(opts, parserOpts) {
-      parserOpts.plugins.push("pipelineOperator");
+      parserOpts.plugins.push(["pipelineOperator", { proposal }]);
     },
   };
 });
--- a/packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/foobar/options.json
+++ b/packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/foobar/options.json
@@ -1,3 +1,6 @@
 {
-  "presets": [["stage-0", { "decoratorsLegacy": true }], "es2015"]
+  "presets": [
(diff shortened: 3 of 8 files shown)


loganfsmyth:
I think this is a reasonable approach.

packages/babel-plugin-syntax-pipeline-operator/src/index.js
```diff
@@ -1,11 +1,17 @@
 import { declare } from "@babel/helper-plugin-utils";
 
-export default declare(api => {
+const proposals = ["minimal"];
+
+export default declare((api, { proposal }) => {
   api.assertVersion(7);
 
+  if (typeof proposal !== "string" || !proposals.includes(proposal)) {
+    throw new Error("'proposal' must be one of: " + proposals.join(", "));
```
hzoo: Yep looks good, wonder if we should introduce a url to show what the options are and why etc.

mAAdhaTTah: I also plan on updating the README. Is there any other place where we should include that documentation?

hzoo: I think the readme is good. We should introduce a new general doc on the page explaining how we want to think about proposals moving forward and how they are implemented though. Kinda like https://github.com/babel/proposals/pull/26 but I didn't get to finishing it, not sure.

mAAdhaTTah:
@loganfsmyth @hzoo This also breaks the stage-0 preset, which would need this configuration. Is that acceptable? If so, I can update the tests to take the config.

hzoo:
Yeah we are going to remove the stage presets because of this issue https://github.com/babel/babel/issues/7770. Same issues we had with decorators actually. Not really worth the trouble anymore so people should opt-into what they want or make their own preset.

mAAdhaTTah:
Ok cool, I'll make those changes this evening.

nicolo-ribaudo:
Even if we are going to deprecate the stage-0 preset, I'd prefer to add a flag which is passed down to this plugin, similarly to what I did with the `decoratorsLegacy` option.

mAAdhaTTah:
@nicolo-ribaudo Yeah, that the test that's failing, so I'll make sure that works when I fix it.

mAAdhaTTah:
Documentation update: https://github.com/babel/website/pull/1700

mAAdhaTTah:
I think all the open question have been addressed, so this should be ready to go.

mAAdhaTTah:
Bumping this PR. Would like to get this landed so we can refresh our branch with the incoming pipeline changes.

packages/babel-plugin-proposal-class-properties/test/fixtures/public-loose/foobar/options.json
```diff
@@ -1,7 +1,7 @@
 {
   "plugins": ["external-helpers", ["proposal-class-properties", {"loose": true}]],
   "presets": [
-    ["stage-0", { "decoratorsLegacy": true }],
+    ["stage-0", { "decoratorsLegacy": true, "pipelineProposal": "minimal" }],
```
hzoo: Unrelated to the PR but I wonder why the stage-0 preset is even necessary to use in this test?

mAAdhaTTah: I have no idea, but it did catch that I broke state-0 so maybe that's ok?

nicolo-ribaudo: Yeah, it is ok. I often use `stage-0` in tests for lazyness 😛

nicolo-ribaudo:
(please update the docs at https://github.com/babel/website/tree/master/docs)

PR #5547, opened 2017-03-26
[7.0] Require babel-(preset|plugin) or module: on plugins/presets

loganfsmyth:
| Q                        | A 
| ------------------------ | ---
| Patch: Bug Fix?          | N
| Major: Breaking Change?  | Y
| Minor: New Feature?      | N
| Deprecations?            | 
| Spec Compliancy?         | 
| Tests Added/Pass?        | 
| Fixed Tickets            | 
| License                  | MIT
| Doc PR                   | 
| Dependency Changes       | 

This is my proposed fix for the issue described in https://github.com/babel/babel/issues/5006. Essentially because we try to resolve _both_ "mod" and "babel-plugin-mod", it means that you can get the wrong module if you forget to install something, rather than getting a nice error message, you get a totally weird one.

I propose changing resolution to be deterministic, rather than based on what exists in the filesystem. So

| Input | Resolved to | Behavior
| -------- | --- | ------
| `/dir/plugin.js` | `/dir/plugin.js` | Absolute paths pass through
| `./dir/plugin.js` | `./dir/plugin.js` | Relative paths pass through
| `mod/plugin` | `mod/plugin` | Files nested inside modules pass through
| `@scope/mod/plugin` | `@scope/mod/plugin` | Files nested inside modules pass through
| `@babel/mod/plugin` | `@babel/mod/plugin` | Files nested inside modules pass through
| `mod` | `babel-plugin-mod` | Standalone module references get a prefix
| `babel-plugin-mod` | `babel-plugin-mod` | Modules with the prefix already pass through unchanged
| `@babel/mod` | `@babel/plugin-mod` | @babel-scoped packages get a `plugin-` prefix without `babel-`
| `@babel/plugin-mod` | `@babel/plugin-mod` | @babel-scoped packages with a `plugin-` pass through
| `@scope/mod` | `@scope/babel-plugin-mod` | Scoped packages get a `babel-plugin-` prefix
| `@scope/babel-plugin-mod` | `@scope/babel-plugin-mod` | Scoped packages with a `babel-plugin-` prefix pass through unchanged.
| `module:foo` | `foo` | Names prefixed with `module:` resolve to just the name

So the downside is that if modules do not begin with the prefix, they will have to prefix the name with `module:`.

Only real downside I see here is if we do start using the `@babel` namespace, people will be required to include the `@babel` in the name, e.g. `@babel/es2015`

Other bonuses in this PR:
* If you pass a plugin/preset and forget the `module:`, throw a helpful error saying you forgot it.
[...]

PR #8197, opened 2018-06-19
Allow @foo/babel-plugin as an unexpanded plugin name, and @foo as a shorthand for it.

### LINKED ISSUE #8091: Make the plugin naming less strict
I would like to publish a babel plugin named following the pattern `@scope/babel-plugin`. Currently this is not possible. My opinion is that this could be accepted.

loganfsmyth:
| Fixed Issues?            | Fixes #8091
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

In #5547 we hard-coded the logic to automatically inject `babel-plugin-` prefix, but it's a little aggressive in this case. This PR expands the behavior added in that PR to handle

| Input | Previous behavior | Resolved to
| -------- | --- | ------
| `@foo/babel-plugin` | `@foo/babel-plugin-babel-plugin` | `@foo/babel-plugin`
| `@foo` | `@foo` | `@foo/babel-plugin`


xtuc:
I have mixed feelings on this, while it looks really magic/implicit, `@foo` itself is not a valid package name right? We want to avoid having the react bugs again. 

Should it expand to a preset instead? Or both?

hzoo:
As long as the error message prints out what expanded path it tried to load we are ok?

xtuc:
@hzoo yes I agree

xtuc:
I think we can merge this now

PR #8198, opened 2018-06-19
Prefer explicit object maps, and properly load relative maps.

### LINKED ISSUE #8070: Input Sourcemap Annotation gets removed
## Bug Report

**Current Behavior**
if my Input Sourcecode has a sourcemap annotation, babel will remove it (if it cant find/parse the sourcemap annotation), even if i pass an input sourcemap object to babel.

**Input Code**

```js
const foo = "bar"
//# sourceMappingURL=/some/path/that/not/exists/yet/myfile.js.map
```
**Babel Call**
```js
const babel = util.promisify(require("@babel/core").transform);

let { code, map } = await babel(inputCode, { ...babelConfig,
    inputSourceMap: inputMap ? JSON.parse(inputMap) : undefined
});
```

**Output Code** 

```js
var foo = "bar"
```

**Expected behavior/code**
the sourcemap annotation should remain in the file. it gets removed [here](https://github.com/babel/babel/blob/master/packages/babel-core/src/transformation/normalize-file.js#L20)

**Expected Code**
```js
var foo = "bar"
//# sourceMappingURL=/some/path/that/not/exists/yet/myfile.js.map
```
> and transformed sourcemap object

**Babel Configuration**

```js
 { extends: 'base/.babelrc',
  babelrc: false,
  compact: false,
  plugins: [],
  presets:
   [ [ '@babel/preset-env/lib/index.js',
       [...] ] ],
  sourceType: 'script' }
```

**Environment**
- Babel version(s): [v7.0.0-beta.49]
- Node/npm version: [Node 8.11]
- OS: [OSX]
- Monorepo [e.g. yes]
- How you are using Babel: [programmatically]

**Possible Solution**

* it worked with v7.0.0-beta.46
* dont fiddle around with the sourcemap annotation if the inputSourcemap is already a sourcemap object

loganfsmyth:
| Fixed Issues?            | Fixes #8070
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-core/src/transformation/normalize-file.js
+++ b/packages/babel-core/src/transformation/normalize-file.js
@@ -27,15 +27,23 @@ export default function normalizeFile(
 
   let inputMap = null;
   if (options.inputSourceMap !== false) {
-    try {
-      inputMap = convertSourceMap.fromSource(code);
+    // If an explicit object is passed in, it overrides the processing of
+    // source maps that may be in the file itself.
+    if (typeof options.inputSourceMap === "object") {
+      inputMap = convertSourceMap.fromObject(options.inputSourceMap);
+    }
 
-      if (inputMap) {
+    if (!inputMap) {
+      try {
+        inputMap = convertSourceMap.fromSource(code);
+
+        if (inputMap) {
+          code = convertSourceMap.removeComments(code);
+        }
+      } catch (err) {
+        debug("discarding unknown inline input sourcemap", err);
         code = convertSourceMap.removeComments(code);
       }
-    } catch (err) {
-      debug("discarding unknown inline input sourcemap", err);
-      code = convertSourceMap.removeComments(code);
     }
 
     if (!inputMap) {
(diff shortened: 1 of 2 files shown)


packages/babel-core/src/transformation/normalize-file.js
```diff
@@ -27,32 +28,44 @@ export default function normalizeFile(
...
     }
 
-    if (!inputMap && typeof options.inputSourceMap === "object") {
-      inputMap = convertSourceMap.fromObject(options.inputSourceMap);
+    if (!inputMap) {
+      if (typeof options.filename === "string") {
+        try {
+          inputMap = convertSourceMap.fromMapFileSource(
+            code,
+            path.dirname(options.filename),
```
loganfsmyth: This logic landed recently, and I realized this argument was missing.

PR #8200, opened 2018-06-19
Update readme in plugin-unicode-property-regex [skip ci]

Bartuz:
yarn doesn't have `--save` flag (save is default functionality)

| Documentation PR         |


existentialism:
@Bartuz thanks for the PR, but our READMEs are now generated via:
https://github.com/babel/babel/blob/master/scripts/generators/readmes.js

Mind fixing and regenerating them?

xtuc:
Unfortunately the generation script doesn't support adding that. We should instead add it on our website.

nicolo-ribaudo:
This has been fixed in the generator script:

https://github.com/babel/babel/blob/master/packages/babel-plugin-proposal-unicode-property-regex/README.md

Issue #8052, opened 2018-05-25
Static Class Features: Stage 3


rricard:
# Static Class Features: Stage 3

> Advance Static Class Features for Stage 3 (Shu-yu Guo @syg) (Daniel Ehrenberg @littledan)

Proposal co-written with @tim-mc and @robpalme. We also intend to follow up and implement it with @rpamely,  @tim-mc, @mkubilayk .

## Info

Proposed at TC39 Meeting: [May 2018](https://github.com/tc39/agendas/blob/master/2018/05.md)

Slides at the meeting: https://docs.google.com/presentation/d/1YzFr7EIGiX2YagfFMjkI-lVR6ouoRfPbTNLY--NGbC4/edit?usp=sharing

Proposal Repo: https://github.com/tc39/proposal-static-class-features/

Complete article on the subject by Shu-yu Guo: https://rfrn.org/%7Eshu/2018/05/02/the-semantics-of-all-js-class-elements.html

Babel contribution recommended by @jridgewell to @robpalme

The work to be done towards Stage 3 will be added in [`plugin-proposal-class-properties`](https://github.com/babel/babel/tree/master/packages/babel-plugin-proposal-class-properties).

### Static Public Fields

```javascript
class Account {
  static routingNumber = 1017;
  get transactionIdPrefix() {
    return `${Account.routingNumber}-${this.uuid}`;
  }
}
```

### Static Public Methods

```javascript
class Account {
  static convertToEuro(dollars) {
     return util.USDToEuro(dollars);
  }

  get euroConversionDisplayAmt () {
    return `Current Balance (EUR): ${Account.convertToEuro(this.accountBalance)}`
  }
```

### Static private fields

```js
class Account {
  static #transactions = [];
  static get nbTransactions() {
    return Account.#transactions.length;
  }
  getDollars() {
    return Account.#transactions.reduce(/* ... */);
  }
}
```

### Static private methods

```js
class Account {
   // ...
  static #makeTransaction(dollars, from, to) {
    Account.#transactions = this.#transactions.concat(/* ... */);
  }
  
  transfer(dollars, targetAccount) {
    return Account.#makeTransaction(dollars, this, targetAccount);
  }
}
```

## Basic Rules

- Check that the `<Class Name>.#<static private field/method>` notation is within the lexical scope of `<Class Name>` - Checked during transform
[...]

PR #8205, opened 2018-06-20
Private Static Fields Features: Stage 3

rricard:
| Fixed Issues?            | #8052 (except accessors)
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | Add private static fields support
| Tests Added + Pass?      | Tests Updated
| Documentation PR         | 
| Any Dependency Changes?  |
| Sponsor                 | @bloomberg

- No changes in parser
- Added a new helper to check private static field access provenance: `classStaticPrivateFieldBase`
- While visiting a class:
    1. Check for duplicate private static declarations
    2. For every private static declaration add it to a container object in the classe's scope and remove it from the class itself
        ```js
        class C {
          static #foo = "bar";
          /*...*/
        }
        // will become (in loose mode):
        class C {
          /*...*/
        }
        var _CStatics = {};
        _CStatics._foo = "bar";
        ```
    3. For every private static declaration, start visiting the class to explore each member method in order to wrap the access in the helper:
        ```js
        myMethod() {
          C.#foo = this.#foo + "baz";
        }
        // Becomes
        myMethod() {
          babelHelper.classStaticPrivateFieldBase(C, C, _CStatics)._foo = 
            babelHelper.classStaticPrivateFieldBase(this, C, _CStatics)._foo + "baz";
        }
        ```

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -1041,3 +1041,35 @@ helpers.classPrivateFieldSet = helper("7.0.0-beta.0")`
     return value;
   }
 `;
+
+helpers.classStaticPrivateFieldLooseBase = helper("7.0.0-beta.0")`
+  export default function _classStaticPrivateFieldLooseBase(receiver, classConstructor) {
+    if (receiver !== classConstructor) {
+      throw new TypeError("Private static access of wrong provenance");
+    }
+    return classConstructor;
+  }
+`;
+
+helpers.classStaticPrivateFieldSpecGet = helper("7.0.0-beta.0")`
+  export default function _classStaticPrivateFieldSpecGet(
+    receiver, classConstructor, privateClass, privateId
+  ) {
+    if (receiver !== classConstructor) {
+      throw new TypeError("Private static access of wrong provenance");
+    }
+    return privateClass[privateId];
+  }
+`;
+
+helpers.classStaticPrivateFieldSpecSet = helper("7.0.0-beta.0")`
+  export default function _classStaticPrivateFieldSpecSet(
+    receiver, classConstructor, privateClass, privateId, value
+  ) {
+    if (receiver !== classConstructor) {
+      throw new TypeError("Private static access of wrong provenance");
+    }
+    privateClass[privateId] = value;
+    return value;
+  }
+`;
--- a/packages/babel-plugin-proposal-class-properties/src/index.js
+++ b/packages/babel-plugin-proposal-class-properties/src/index.js
@@ -160,6 +160,60 @@ export default declare((api, options) => {
     },
   };
 
+  const staticPrivatePropertyHandlerSpec = {
+    ...privateNameHandlerSpec,
+
+    get(member) {
+      const { file, name, privateClassId, classRef } = this;
+
+      return t.callExpression(
+        file.addHelper("classStaticPrivateFieldSpecGet"),
+        [
+          this.receiver(member),
+          classRef,
+          privateClassId,
+          t.stringLiteral(name),
+        ],
+      );
+    },
+
+    set(member, value) {
+      const { file, name, privateClassId, classRef } = this;
+
+      return t.callExpression(
+        file.addHelper("classStaticPrivateFieldSpecSet"),
+        [
+          this.receiver(member),
+          classRef,
+          privateClassId,
+          t.stringLiteral(name),
+          value,
+        ],
+      );
+    },
+
+    call(member, args) {
+      // The first access (the get) should do the memo assignment.
+      this.memoise(member, 1);
+
+      return optimiseCall(this.get(member), this.receiver(member), args);
+    },
+  };
+
+  const staticPrivatePropertyHandlerLoose = {
+    handle(member) {
+      const { file, privateId, classRef } = this;
+      member.replaceWith(
+        template.expression`BASE(RECEIVER, CLASS).PRIVATE_ID`({
+          BASE: file.addHelper("classStaticPrivateFieldLooseBase"),
+          RECEIVER: member.node.object,
+          CLASS: classRef,
+          PRIVATE_ID: privateId,
+        }),
+      );
+    },
+  };
+
   function buildClassPropertySpec(ref, path, state) {
     const { scope } = path;
     const { key, value, computed } = path.node;
@@ -255,6 +309,75 @@ export default declare((api, options) => {
       });
   }
 
+  function buildClassStaticPrivatePropertySpec(
+    ref,
+    path,
+    state,
+    privateClassId,
+  ) {
+    const { scope, parentPath } = path;
+    const { key, value } = path.node;
+    const { name } = key.id;
+    const staticNodesToAdd = [];
+
+    if (!privateClassId) {
+      // Create a private static "host" object if it does not exist
+      privateClassId = path.scope.generateUidIdentifier(ref.name + "Statics");
+      staticNodesToAdd.push(
+        template.statement`const PRIVATE_CLASS_ID = Object.create(null);`({
+          PRIVATE_CLASS_ID: privateClassId,
+        }),
+      );
+    }
+
+    memberExpressionToFunctions(parentPath, privateNameVisitor, {
+      name,
+      privateClassId,
+      classRef: ref,
+      file: state,
+      ...staticPrivatePropertyHandlerSpec,
+    });
+
+    staticNodesToAdd.push(
+      t.expressionStatement(
+        t.callExpression(state.addHelper("defineProperty"), [
+          privateClassId,
+          t.stringLiteral(name),
+          value || scope.buildUndefinedNode(),
+        ]),
+      ),
+    );
+
+    return [staticNodesToAdd, privateClassId];
+  }
+
+  function buildClassStaticPrivatePropertyLoose(ref, path, state) {
+    const { scope, parentPath } = path;
+    const { key, value } = path.node;
+    const { name } = key.id;
+    const privateId = scope.generateUidIdentifier(name);
+
+    parentPath.traverse(privateNameVisitor, {
+      name,
+      privateId,
+      classRef: ref,
+      file: state,
+      ...staticPrivatePropertyHandlerLoose,
+    });
+
+    const staticNodesToAdd = [
+      t.expressionStatement(
+        t.assignmentExpression(
+          "=",
+          t.memberExpression(ref, privateId),
+          value || scope.buildUndefinedNode(),
+        ),
+      ),
+    ];
+
+    return [staticNodesToAdd];
+  }
+
   const buildClassProperty = loose
     ? buildClassPropertyLoose
     : buildClassPropertySpec;
@@ -262,6 +385,10 @@ export default declare((api, options) => {
   const buildClassPrivateProperty = loose
     ? buildClassPrivatePropertyLoose
     : buildClassPrivatePropertySpec;
+
+  const buildClassStaticPrivateProperty = loose
+    ? buildClassStaticPrivatePropertyLoose
+    : buildClassStaticPrivatePropertySpec;
 
   return {
     inherits: syntaxClassProperties,
@@ -288,17 +415,11 @@ export default declare((api, options) => {
 
           if (path.isClassPrivateProperty()) {
             const {
-              static: isStatic,
               key: {
                 id: { name },
               },
             } = path.node;
 
-            if (isStatic) {
-              throw path.buildCodeFrameError(
-                "Static class fields are not spec'ed yet.",
-              );
-            }
             if (privateNames.has(name)) {
               throw path.buildCodeFrameError("Duplicate private field");
             }
@@ -354,7 +475,7 @@ export default declare((api, options) => {
         const privateMaps = [];
         const privateMapInits = [];
         for (const prop of props) {
-          if (prop.isPrivate()) {
+          if (prop.isPrivate() && !prop.node.static) {
             const inits = [];
             privateMapInits.push(inits);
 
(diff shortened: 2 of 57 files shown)


rricard:
To do list:

- [ ] Review all current test outputs one by one, make sure we run `exec.js` versions when possible
- [ ] Private static setters/getters
- [ ] Account for a spec version

rricard:
We only handled private static fields for now, @tim-mc is going to help bring private static methods now. From there we'll be able to think about accessors...

rricard:
It feels like this PR is ready for review.

I also would like some help for making that Travis build pass...

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/static-undefined/exec.js
```diff
@@ -0,0 +1,15 @@
...
+  }
+
+  test() {
+    return Foo.#bar;
+  }
+}
+
+expect("bar" in Foo).toBe(false)
+expect(Foo.test()).toBe(undefined)
+expect(Foo.test()).toBe(undefined)
```
hzoo: Some of these duplicated tests are weird to me `expect(Foo.test()).toBe(undefined)` - I know they were just copied over but is it supposed to be testing the static and instance methods or just remove it?

nicolo-ribaudo: Maybe it should be `expect(() => Foo.test()).not.toThrow()`?

rricard: @hzoo, you're right, I'll go ahead and do a second review pass on those tests (was only checking for the output I was seeing). I'm also trying to run test-262 right now against it (having `npm link` issues) and I might add some relevant tests then

packages/babel-helpers/src/helpers.js
```diff
@@ -1011,3 +1011,12 @@ helpers.classPrivateFieldSet = () => template.program.ast`
     return value;
   }
 `;
+
+helpers.classStaticPrivateFieldBase = () => template.program.ast`
+  export default function _classStaticPrivateFieldBase(receiver, classConstructor, privateClass) {
+    if (receiver !== classConstructor && receiver.constructor !== classConstructor) {
```
jridgewell: The `receiver.constructor !== classConstructor` shouldn't be necessary.

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -285,14 +377,18 @@ export default declare((api, options) => {
             } = path.node;
 
             if (isStatic) {
-              throw path.buildCodeFrameError(
-                "Static class fields are not spec'ed yet.",
-              );
+              if (privateStaticNames.has(name)) {
```
jridgewell: Private names are shared between instance and static, ie, the following is an error:

```js
class Example {
  static #x;
  #x;
}
```

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/native-classes/output.js
```diff
@@ -1,18 +1,37 @@
-var _foo, _bar;
+var Foo =
+/*#__PURE__*/
+function () {
+  "use strict";
```
jridgewell: This is my mistake. This test case was meant to output `class` expressions, not compile classes to functions.

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -83,6 +83,18 @@ export default declare((api, options) => {
     },
   };
 
+  // Traverses the class scope, handling private static name references.
+  const staticPrivatePropertyVisitor = {
```
jridgewell: We should be able to reuse `privateNameVisitor`. This is necessary because we have to use its inner traverser, too.

packages/babel-helpers/src/helpers.js
```diff
@@ -1011,3 +1011,12 @@ helpers.classPrivateFieldSet = () => template.program.ast`
     return value;
   }
 `;
+
+helpers.classStaticPrivateFieldBase = () => template.program.ast`
+  export default function _classStaticPrivateFieldBase(receiver, classConstructor, privateClass) {
```
jridgewell: So there are a few issues here, and we can choose either of two ways to handle it:

- We can remove this `privateClass` holder object for `loose` mode.
- We can handle `get`, `set`, and `call` for both `loose` and `spec` mode.

The first option would allow us to reuse the same code we already have for instance private fields.

If we continue to use a holder object, we have to handle `get`, `set`, and `call` operations on the holder, and use the `receiver` as the calling context for all those cases. This complicates the runtime for `loose` mode. But, we're gonna need to do that anyways for spec mode.

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -157,6 +169,23 @@ export default declare((api, options) => {
     },
   };
 
+  const staticPrivatePropertyHandler = {
+    handle(member) {
```
jridgewell: So this doesn't handle `get`, `set`, and `call` operations. Given that all this is stored on a holder object, anything like `ClassName.#x()` will really have to be transformed to `base(ClassName, ClassName, holder)['x'].call(ClassName)`. This currently just outputs `base(ClassName, ClassName, holder)['x']()`.

rricard:
Thanks @jridgewell, I'm sorry I have a lot of work on the side of this but I'll try to get back to it this week.

packages/babel-plugin-proposal-class-properties/src/index.js
rricard: Wasn't sure about this one, I'll make the change.

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/native-classes/output.js
rricard: That makes more sense indeed, I'll redo the whole test dir with that then!

packages/babel-plugin-proposal-class-properties/src/index.js
rricard: Ok, I kept them separate to make sure I don't mess with your logic but indeed, it's better to change just a few things in your logic than to branch out an entirely new one!

packages/babel-helpers/src/helpers.js
rricard: This is super helpful, I had a hard time understanding the line between loose and spec.

packages/babel-plugin-proposal-class-properties/src/index.js
rricard: I think I just missed something about how the handlers are working and that made it way clearer. Thanks!

rricard:
Todo:

- [ ] Rework the helper(s)
- [ ] Redo the tests
- [ ] get/set/call handler
    - [ ] spec
    - [ ] loose

packages/babel-helpers/src/helpers.js
rricard: Unfortunately, this seems necessary to be spec compliant. We should discuss that on slack.

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/native-classes/output.js
rricard: @jridgewell what should I do to convert tests to output class expressions?

rricard:
Hi @jridgewell, I could use some help changing the tests to not transform the classes. Otherwise, I think I'm almost done.

packages/babel-helpers/src/helpers.js
```diff
@@ -1029,3 +1029,34 @@ helpers.classPrivateFieldSet = () => template.program.ast`
     return value;
   }
 `;
+
+helpers.classStaticPrivateFieldLooseBase = () => template.program.ast`
+  export default function _classStaticPrivateFieldLooseBase(receiver, classConstructor) {
+    if (receiver !== classConstructor && receiver.constructor !== classConstructor) {
```
jridgewell: What's the `receiver.constructor !== classConstructor` for?

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -245,6 +299,77 @@ export default declare((api, options) => {
...
+    const { key, value } = path.node;
+    const { name } = key.id;
+    const privateId = scope.generateUidIdentifier(name);
+    const staticNodesToAdd = [];
+
+    if (!privateClassId) {
+      // Create a private static "host" object if it does not exist
+      privateClassId = path.scope.generateUidIdentifier(ref.name + "Statics");
+      staticNodesToAdd.push(
+        template.statement`const PRIVATE_CLASS_ID = {};`({
```
jridgewell: Nit: Should probably use a prototype-less object here.

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -278,19 +407,13 @@ export default declare((api, options) => {
...
             } = path.node;
 
-            if (isStatic) {
-              throw path.buildCodeFrameError(
-                "Static class fields are not spec'ed yet.",
-              );
-            }
             if (privateNames.has(name)) {
-              throw path.buildCodeFrameError("Duplicate private field");
+              throw path.buildCodeFrameError("Duplicate static private field");
```
jridgewell: Let's leave this as the old message.

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -353,11 +476,27 @@ export default declare((api, options) => {
...
+              [
+                staticNodesToAdd,
+                privateClassId,
+              ] = buildClassStaticPrivateProperty(
+                t.cloneNode(ref),
+                prop,
+                state,
+                privateClassId,
+              );
+              staticNodesToAdd.forEach(node => staticNodes.push(node));
```
jridgewell: Can just `staticNodes.push(...staticNodesToAdd)`

packages/babel-helpers/src/helpers.js
rricard: Rereading the whole thing, this logic is busted. That's why I want to make sure the tests work as well

rricard: That's for when the static field gets called from `this`. In that case I'll probably have an instance of the class instead of the constructor itself. I do think that's what we expect from the spec but it's been a while since the last time I dived in the proposal.

rricard:
@jridgewell don't worry, we're in no rush. I'll just need some help setting up the tests correctly.

packages/babel-helpers/src/helpers.js
jridgewell: It has to throw a type error if it's called on an instance of the class. This check should just be the `receiver !== classConstructor`.

jridgewell:
You can probably duplicate some of the tests from the instance private fields, we just need to change the `input.js` code to use statics.

rricard:
I was trying to prevent the class transform itself but I managed to get that right in the plane

rricard:
It should be good for a second review

packages/babel-helpers/src/helpers.js
rricard: I'm not sure, I think that if the instance is of the class type, access should work through `this`. I can check that again. However, in the meantime, we can restrict the access so it does not work on the instance while we figure that out.

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -245,6 +299,77 @@ export default declare((api, options) => {
...
+  function buildClassStaticPrivatePropertySpec(
+    ref,
+    path,
+    state,
+    privateClassId,
+  ) {
+    const { scope, parentPath } = path;
+    const { key, value } = path.node;
+    const { name } = key.id;
+    const privateId = scope.generateUidIdentifier(name);
```
jridgewell: Nit: So we actually don't need to generate a unique id here, we can just use `name`.

rricard:
Loose tests have been removed while I changed my test methodology, they're back now. I also removed the privateId for spec mode but not loose mode (where the unique id prevents test breakage in loose mode)

rricard:
Rebased the whole thing

rricard:
@jridgewell I think this covers all of the requested changes. Thanks!

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/native-classes/output.js
```diff
@@ -0,0 +1,27 @@
...
+    return babelHelpers.classStaticPrivateFieldLooseBase(Foo, Foo)._foo;
+  }
+
+  test() {
+    return babelHelpers.classPrivateFieldLooseBase(this, _bar)[_bar];
+  }
+
+}
+
+Foo._foo = "foo";
```
jridgewell: My last nit: This should be using `Object.defineProperty` with `{ enumerable: false, configurable: false }`. We can do this in a follow PR if you'd like.

rricard: That is a good point, I'll try to put that here before we merge

rricard:
@jridgewell I made the change, would you like to have it in a separate helper though?

jridgewell:
Eh, let's just go for it.

nicolo-ribaudo:
I'm sorry, I forgot to review this PR :sweat_smile:
I'll open a few PR with some small tweaks.

PR #8208, opened 2018-06-22
Ensure that we don't get unexpected output files for tests that throw.

loganfsmyth:
| Fixed Issues?            | 
| Patch: Bug Fix?          | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Just noticed that when I removed all the output files to regenerate them, some of them didn't actually regenerate. These tests all throw exceptions, but we still had some output files for them anyway.

cc @jridgewell


loganfsmyth:
I'm so confused about these failures. The file that is erroring is deleted in this PR...

packages/babel-plugin-proposal-pipeline-operator/test/fixtures/pipeline-operator/async-arrow/output.js
```diff
@@ -1,12 +0,0 @@
-var _ref, _;
```
hzoo: It looks like for this one, it doesn't throw so the input file is still there and thus expects an output file? https://github.com/babel/babel/tree/9283efaeb7a8c5b505b19332c3407dc44c668777/packages/babel-plugin-proposal-pipeline-operator/test/fixtures/pipeline-operator/async-arrow

loganfsmyth: Ugh yeah I just missed that the test didn't re-run because it was skipped on Node 6.

PR #8210, opened 2018-06-22
Fix optional chaining bug regarding spread in function calls

### LINKED ISSUE #8136: spread-operator doesn't transform to ES5
## Bug Report

**Current Behavior**
spread-operator doesn't transform to ES5 

```js
"use strict";

var _arr;

_arr = [1, 2, 3].map(function (e) {
  return Math.pow(e, 2);
}), Math === null || Math === void 0 ? void 0 : Math.max(..._arr);
```
**Input Code**
```js
[1, 2, 3].map(e => e ** 2) |> (arr => Math?.max(...arr))
```

**Expected behavior/code**
```js
"use strict";

var _arr;

_arr = [1, 2, 3].map(function (e) {
  return Math.pow(e, 2);
}), Math === null || Math === void 0 ? void 0 : Math.max.apply(Math, _arr);
```

```js
const presetEnv = require('@babel/preset-env');
const presetStage0 = require('@babel/preset-stage-0');
{
  presets: [
    [presetEnv, { targets: { browsers: ['ie >= 9'] } }],
    [presetStage0, { decoratorsLegacy: true }],
  ],
}
```

**Environment**
- Node/npm version: Node 10.0.0/npm 5.6.0
- OS: Windows 7
- Monorepo yes
- How you are using Babel: `babel.transform`

**Possible Solution**

microbouji:
| Fixed Issues?            | Fixes #8136  
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Currently when `proposal-optional-chaining` transforms

```
a?.b()
```

into

```
(_a = a) === null || _a === void 0 ? void 0 : _a.b()
```

that last `_a.b()` is technically an `OptionalCallExpression` with no `optional` field set.

This PR changes it to be just a `CallExpression`. I believe this is how it should be, and this also enables other plugins (like `transform-spread`) who use `CallExpression` in their visitors to keep working.


packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -83,6 +83,10 @@ export default declare((api, options) => {
         }
       }
 
+      if (replacementPath.isOptionalCallExpression()) {
+        replacementPath.node.type = "CallExpression";
```
hzoo: Looks good, the type should be changed in the transform for sure? This logic is similar to what was done in https://github.com/babel/babel/pull/8210/files#diff-accf2589ebb96f2b32131e0a1f8a08eaR24

```
if (objectPath.isOptionalMemberExpression()) {
  objectPath.node.type = "MemberExpression";
}
```

dackmin:
@hzoo Do you have an idea of when this PR will be merged & released ? I came across this issue using a spread operator into an optional chaining method call and the spread operator is invisible for `@babel/plugin-transform-spread`.

microbouji:
@jridgewell thank you!

PR #8238, opened 2018-06-29
Typescript: Avoid stripping class properties when a decorator is set

pmdartus:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

**Current behavior:**
The `babel-plugin-transform-typescript` removes class properties without value regardless if decorators are assigned to it or not.

```ts
// Original
class C {
  @foo
  d;
  @foo
  e = 3;
}

// Transformed
class C {
  // <-- Missing `d` field
  @foo
  e = 3;
}
```

**Expected behavior:**
The `babel-plugin-transform-typescript` should only remove a class property if it has no value and no decorators.

```ts
// Original
class C {
  @foo
  d;
  @foo
  e = 3;
}

// Transformed
class C {
  @foo
  d;
  @foo
  e = 3;
}
```

[REPL example](http://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEASCmIQHtoG8BQ1pgNxegCNoBeaARn2wAEAzZVYK6Oh6AE1IvwF8g&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&sourceType=module&lineWrap=false&presets=typescript&prettier=true&targets=Node-8&version=7.0.0-beta.35&envVersion=)

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -157,10 +157,6 @@ export default declare((api, { jsxPragma = "React" }) => {
 
       ClassProperty(path) {
         const { node } = path;
-        if (!node.value) {
-          path.remove();
-          return;
-        }
 
         if (node.accessibility) node.accessibility = null;
         if (node.abstract) node.abstract = null;
@@ -196,7 +192,10 @@ export default declare((api, { jsxPragma = "React" }) => {
         path.get("body.body").forEach(child => {
           if (child.isClassProperty()) {
             child.node.typeAnnotation = null;
-            if (!child.node.value) child.remove();
+
+            if (!child.node.value && !child.node.decorators) {
+              child.remove();
+            }
           }
         });
       },
--- a/packages/babel-plugin-transform-typescript/test/fixtures/class/properties/input.js
+++ b/packages/babel-plugin-transform-typescript/test/fixtures/class/properties/input.js
@@ -2,4 +2,6 @@ class C {
     public a?: number;
     private b: number = 0;
     readonly c!: number = 1;
+    @foo d: number;
+    @foo e: number = 3;
 }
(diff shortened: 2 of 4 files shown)


packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -157,10 +157,6 @@ export default declare((api, { jsxPragma = "React" }) => {
 
       ClassProperty(path) {
         const { node } = path;
-        if (!node.value) {
```
pmdartus: This code is a duplicate of the one in the `Class` visitor.

xtuc:
I don't have enough context here, sorry. Why do we want to emit decorators syntax? Shoudln't we strip out unsupported syntax?

nicolo-ribaudo:
The typescript plugin should leave decorators there: they will be handled by the decorators plugin.

diervo:
@nicolo-ribaudo are you good with this changes? If so, shall bribe @hzoo or @loganfsmyth to merge this? :)

xtuc:
I'm :+1: only if we can make sure the user is aware of this behavior:
- we should warn if decorator are not transformed (because no plugin loaded)

nicolo-ribaudo:
I don't think that we should, the same way we don't warn if class properties or other proposals supported by typescript are not transpiled.
Users will see very easily that their code is not transformed, since it would cerash at the line containing the decorators as soon as it is loaded in a browser.

xtuc:
That's right @nicolo-ribaudo, I was wondering if the "why it didn't transform it" would be confusing for users or not.

nicolo-ribaudo:
If the user just adds the typescript plugin, Babylon throws and suggests the user to add the `transform-decorators` plugin.
The only problem would be if the users only enables the syntax plugin, but it is known that syntax plugins leave the input as-is.

hzoo:
This is fine, it's expected behavior 👍.

xtuc:
Ah yes Nicolo, perfect!

PR #8215, opened 2018-06-22
Update README.md

techmexdev:
`❯ yarn add --save @babel/preset-env`
yields error
```
yarn add v1.3.2
error Missing list of packages to add to your project.
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
```

`yarn add @babel/preset-env --dev` Successfully installs the package as a dev dependency.

| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | `Fixes #1, Fixes #2` 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes
| Documentation PR         | 
| Any Dependency Changes?  |
| License                  | MIT

PR #8255, opened 2018-07-03
Fix yarn installation command

vfonic:
The one proposed throws error


hzoo:
Thanks for this, we have an existing PR https://github.com/babel/babel/pull/8215 that fixes the script that generates the readme's but didn't land it yet actually!

PR #8259, opened 2018-07-04
upgrades eslint v5 (major), babel-eslint, eslint-plugin-flowtype, eslint-plugin-prettier

dnalborczyk:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  | **eslint** v5.0.1, **babel-eslint** v8.25, **eslint-plugin-flowtype** v2.49.3, **eslint-plugin-prettier** v2.6.1

`make lint` and `make fix` pass

`migration guide:`
https://eslint.org/docs/5.0.0/user-guide/migrating-to-5.0.0

breaking change, v5 throws for non-existing files:`
https://eslint.org/docs/5.0.0/user-guide/migrating-to-5.0.0#-linting-nonexistent-files-from-the-command-line-is-now-a-fatal-error
-> removed non-existing glob, in addition removed default file .js extension glob

`breaking change, v5 no-unused-vars reports all after-used params`
https://eslint.org/blog/2018/04/eslint-v5.0.0-alpha.1-released#breaking-changes
-> added  // eslint-disable-line no-unused-vars


Makefile
```diff
@@ -47,10 +47,10 @@ flow:
 	./node_modules/.bin/flow check --strip-root
 
 lint:
-	./node_modules/.bin/eslint scripts $(SOURCES) '*.js' '**/.*.js' --format=codeframe --rulesdir="./scripts/eslint_rules"
```
nicolo-ribaudo: Why where these removed? I think that it was used to lint things like `babel.config.js`

dnalborczyk: I tried to explain it above:
- removed non-existing glob:  '**/.*.js'   https://eslint.org/docs/5.0.0/user-guide/migrating-to-5.0.0#-linting-nonexistent-files-from-the-command-line-is-now-a-fatal-error
- in addition removed default file .js extension glob: '*.js'

dnalborczyk: oh I see. just looking at it a little closer:  `'**/.*.js'`

I suppose there are no files with .something.js anymore. I guess babel.config had a dot prior.

eslint doesn't find anything, and throws.

dnalborczyk: I think we could take out $(SOURCES), and replace it with '**/*.js' (everything), and .eslintignore excludes everything which needs to be excluded.

nicolo-ribaudo: Oh, i didn't notice the `.` in `.*.js`. It is ok to have sources instead of `**/*.js` (we would lint too much otherwise), but `*.js` should still be there. I tested it locally ant with this patch `babel.config.js` isn't linted anymore.

dnalborczyk: I didn't really see it either :smiley: I only saw *eslint* complaining, and removed it.

alrighty, I'll put `'*.js'` back shortly and leave $(SOURCES) as is.

PR #8261, opened 2018-07-04
Introduce objectWithoutPropertiesLoose helper

Andarist:
| Fixed Issues?            | 
| Patch: Bug Fix?          | 
| Major: Breaking Change?  |
| Documentation PR         | 
| Any Dependency Changes?  |

Just a shorter version (without symbol handling) of `objectWithoutProperties` helper - for `loose` mode


Andarist:
Yeah, it could be - don't see any downside of this. That would have to be done in a separate PR though.

Andarist:
@nicolo-ribaudo good idea 👍 done

Andarist:
@TrySound actually I've realized now that this particular helper (`extends`) is used in other contexts too, so renaming it to `objectSpreadLoose` would actually make it confusing.

Issue #8155, opened 2018-06-13
Split out helpers from core-js/built-ins


hzoo:
Having the version of babel-runtime be tied to core-js has caused us issues so Logan has suggested we split it out. Just writing down in an issue what I think we can do? (There's plenty of stuff to fix but we can always move things till later since we just want an RC too).

- `@babel/plugin-transform-runtime` should only transform the babel helpers (devDep) instead of also doing it for `core-js` and `regenerator`. The other two would be in another package like `babel-runtime-corejs2`? (not sure if regenerator has to be it's own thing again or not)
  - it currently has a few options that we can change.
  - `helpers`: true: this is the whole point now so we can just remove this option.
  - `polyfill` false: removing this option
  - `regenerator` false: removing this option (maybe)?
  - `moduleName` false: not even sure who is using this anymore maybe CRA?
  - `useBuiltIns` true: by default?
  - `useESModules` - true: or remove the option?
- [`@babel/runtime`](https://github.com/babel/babel/tree/master/packages/babel-runtime) should only contain helpers code (dependency)?
  - I think we have 4 copies of helpers: useBuiltIns + useESModules combo.

Yes this is more configuration but you don't have to use it. Ideally with this we would just include  `@babel/plugin-transform-runtime` as part of preset-env by default? and we can combine the logic with preset-env's useBuiltIns.

Hopefully this makes sense, feel free to edit cc @loganfsmyth

Andarist:
Do we have to support helpers referencing core-js as part of this module? Seems like those are completely different things and if people want to polyfill some stuff in helpers modules then they should have an easy way (~) of compiling helpers with core-js v2, v3 or any others. Either as standalone file that could be referenced by `@babel/plugin-transform-runtime` or as part of bundlers loaders/plugins as "in memory" module.

Issue #8087, opened 2018-05-31
De-pessimize external helpers


chocolateboy:
## Feature Request

Several helper functions redefine themselves (i.e. replace themselves with a more optimized version) at runtime e.g.:

* `construct`
* `extends`
* `get`
* `getPrototypeOf`
* `set`
* `setPrototypeOf`
* `typeof`
* `wrapNativeSuper`

This has the advantage of scoping the optimization to the helper, which [reportedly](https://github.com/babel/babel/pull/7188) makes inline helpers more amenable to minification/tree-shaking in some cases.

The downside is that it a) makes the code less clear, which makes it harder to spot [bugs](https://github.com/babel/babel/pull/8083), and b) doesn't work with (and *slows down*) the (default, non-ES6) external helpers.

If minification/tree-shaking is a concern, *external helpers should be used*, rather than making the code slower for external helpers and less maintainable for everyone.

I propose that these optimizations should be reimplemented without redefining the helper e.g.:

## Before

```js
export default function _setPrototypeOf (o, p) {
    _setPrototypeOf = Object.setPrototypeOf || function _setPrototypeOf (o, p) {
        o.__proto__ = p;
        return o;
    }

    return _setPrototypeOf(o, p)
}
```

## After

```js
function _objectSetPrototypeOf (o, p) {
    o.__proto__ = p
    return o
}

var __setPrototypeOf = Object.setPrototypeOf || _objectSetPrototypeOf

export default function _setPrototypeOf (o, p) {
    return __setPrototypeOf(o, p)
}
```

loganfsmyth:
> doesn't work with (and slows down) the (default, non-ES6) external helpers.

Could you clarify what you mean by that? What doesn't work?

>     var __setPrototypeOf = Object.setPrototypeOf || _objectSetPrototypeOf

The reason these helpers tend to put this stuff inside the helper is that they often rely on function declaration hoisting. For example if we inject a helper

```
function _interopRequireDefault(){}
```
it's entirely possible that the module transform will then do

```
var _mod = _interopRequireDefault(require("mod"));

function _interopRequireDefault(){}
```

Things get especially complex once you consider cyclic dependencies in modules. For instance we could output something like

```
import Other from "./other";

export default function fn() {
  return class Foo extends Other {}
}
```
which could end up as
```
exports.default = fn;
var _mod = require("./other");
function _setPrototypeOf(){ ... }

function fn() {
  function Foo(){}
  _setPrototypeOf(Foo, Other);
[...]

PR #8266, opened 2018-07-05
Split @babel/runtime into 2 modules via @babel/runtime-corejs2

loganfsmyth:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/8155
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | EDIT by @hzoo: https://github.com/babel/website/pull/1714
| Any Dependency Changes?  |

In Babel 6, we had `@babel/runtime` which _always_ uses `core-js` in its helpers, meaning that anyone who wants to use our helpers in a centralized modular way had to also use core-js, which isn't always ideal.

Early in the Babel 7 alpha, we added a `useBuiltins: true` option that would use separate copies of the helpers that would skip loading `core-js`, but this _still_ requires users to have an indirect npm dependency on `core-js`.

This PR splits our setup into two separate runtimes:

* ```
  'plugins': [
    ['transform-runtime'],
  ]
  ```
  with `npm install --save @babel/runtime` that just includes our helpers, and relies on users to globally polyfill any APIs they need.
* ```
  'plugins': [
    ['transform-runtime', { corejsVersion: 2 }],
  ]
  ```
  with `npm install --save @babel/runtime-corejs2` that includes everything `@babel/runtime` does, but also includes a passthrough API for `corejs@2.x` and rewrites all of the helper modules to reference `core-js`.


packages/babel-plugin-transform-runtime/scripts/build-dist.js
```diff
@@ -0,0 +1,158 @@
...
+  const tree = t.program([], [], esm ? "module" : "script");
+  const dependencies = {};
+  let bindings = null;
+
+  if (!esm) {
+    bindings = [];
+    for (const dep of helpers.getDependencies(helperName)) {
+      const id = (dependencies[dep] = t.identifier(t.toIdentifier(dep)));
+      tree.body.push(template.statement.ast`
+        var ${id} = require("${`./${dep}`}");
```
nicolo-ribaudo: Some helpers redefine the exported function. Does this work in that case?

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -8,28 +8,83 @@ export default declare((api, options) => {
   api.assertVersion(7);
 
   const {
-    helpers,
-    moduleName = "@babel/runtime",
-    polyfill,
-    regenerator,
-    useBuiltIns,
-    useESModules,
+    corejsVersion = false,
```
nicolo-ribaudo: Maybe this option should be named just `corejs`? `corejs: 2` looks as good as `corejsVersion: 2` to me, but `corejs: false` looks better than `corejsVersion: false`.

zloirock:
- Please, use kebab-case for `core-js` where it's possible for consistency with package / repo name and simplify of searching.
- I haven't any plans about long time support of `core-js@2` after `core-js@3` release and creating runtime package for `babel@7` which use  `core-js@2` could cause problems in the future.

packages/babel-plugin-transform-runtime/scripts/build-dist.js
loganfsmyth: Yeah, it's a bug, but it's existing and tracked in https://github.com/babel/babel/issues/8087

packages/babel-plugin-transform-runtime/src/index.js
loganfsmyth: Sounds good. I was on the fence.

loganfsmyth:
@zloirock

> use kebab-case for core-js where it's possible

Do you have strong feelings about specific cases in this PR? The primary issue is that I think the extra dash makes things harder in some cases here, like `babel-runtime-core-js-2` is definitely harder to read than `-corejs2`. Same if we rename the `corejsVersion` option to `corejs`, using the dash would be frustrating since users would have to quote the option name.

> I haven't any plans about long time support of core-js@2 after core-js@3 release

I think that's fine. I thought about making it just `babel-runtime-corejs`, but my concern was that people may want control over it because they may have other packages built with Babel 6's `babel-runtime`, and if we don't provide a runtime that uses `core-js@2` then they'll be forced to use two parallel versions of `core-js` at the same time, which would increase their bundle sizes a lot. I'd still like to get `core-js@3` landed and recommend it over the `2.x` version.

packages/babel-plugin-transform-runtime/scripts/build-dist.js
nicolo-ribaudo: Oh ok, I didn't realize you just moved around this code

.gitignore
```diff
@@ -12,24 +12,27 @@ coverage
...
+/packages/babel-runtime-corejs2/helpers/*.js
+!/packages/babel-runtime-corejs2/helpers/toArray.js
+!/packages/babel-runtime-corejs2/helpers/iterableToArray.js
+!/packages/babel-runtime-corejs2/helpers/temporalRef.js
+/packages/babel-runtime-corejs2/helpers/es6/*.js
+!/packages/babel-runtime-corejs2/helpers/es6/toArray.js
+!/packages/babel-runtime-corejs2/helpers/es6/iterableToArray.js
+!/packages/babel-runtime-corejs2/helpers/es6/temporalRef.js
+/packages/babel-runtime-corejs2/core-js/**/*.js
+!/packages/babel-runtime-corejs2/core-js/map.js
```
hzoo: For anyone else reviewing, the reason why I added these `!` gitignore's earlier is so that we are able to view the changes in output for the runtime since we would change the build script and it had a bug before and we never viewed the diff. But maybe there is a better way to do this since it's hardcoded. Basically we want to catch issues with the script

packages/babel-plugin-transform-runtime/test/fixtures/runtime-corejs2/aliased-constructors/output.js
```diff
@@ -0,0 +1,14 @@
+var _Map = require("@babel/runtime-corejs2/core-js/map");
```
hzoo: This is just a nit and maybe not worth looking into, but do we want to just make the path `"@babel/runtime-corejs2/map"`?

hzoo:
Was looking at https://github.com/loganfsmyth/babel/blob/runtime-refactor/packages/babel-runtime and I believe you can delete https://github.com/loganfsmyth/babel/blob/runtime-refactor/packages/babel-runtime/core-js.js?

packages/babel-plugin-transform-runtime/scripts/build-dist.js
```diff
@@ -0,0 +1,158 @@
...
+}
+
+function writeHelperFiles(runtimeName, { esm, corejsVersion }) {
+  const pkgDirname = getRuntimeRoot(runtimeName);
+
+  for (const helperName of helpers.list) {
+    const helperFilename = path.join(
+      pkgDirname,
+      "helpers",
+      esm ? "es6" : "",
```
Andarist: if the option is named `esm` then this directory should be IMHO named `esm` too, `es6` makes it more confusing because othen than module format those files wont be written using es6 syntax

Andarist:
Could u explain how `/es6/*` helpers are supposed to be used in comparison to `.mjs` ones? I also can't see `.mjs` being created with the script in this PR - when they are created?

packages/babel-plugin-transform-runtime/test/fixtures/runtime-corejs2/aliased-constructors/output.js
loganfsmyth: Hmm, I don't love having `core-js` repeated twice, but I do think having the core-js stuff grouped into a folder is nice. Do you just want to make it shorter, or is the repetition the part you want to avoid? Thoughts on other names?

packages/babel-plugin-transform-runtime/scripts/build-dist.js
loganfsmyth: Sounds good to me. This was just copied over as it was from `babel/runtime`

packages/babel-plugin-transform-runtime/scripts/build-dist.js
```diff
@@ -0,0 +1,158 @@
+"use strict";
```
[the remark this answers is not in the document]
loganfsmyth: The transform and the build script share the same source of data, which was my motivation. If we split it out, the build script will still have to reach into the transform to get that data, and doing that felt equally awkward.

loganfsmyth:
@Andarist This is just moving the existing logic. We don't currently offer `.mjs` helpers. I considered adding them, but was concerned about it potentially being more painful for users. It's also not too bad to add `.mjs` versions later, I think.

packages/babel-plugin-transform-runtime/scripts/build-dist.js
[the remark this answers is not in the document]
Andarist: In that case - feel free to go with whatever u prefer, I would still prefer to have this top-level, but I'm not super attached to it now (because of the reasons u gave)

Andarist:
[This](https://github.com/babel/babel/pull/8266/files#diff-831c9f9a0aadcb6feca4a45e1817dde4) gave me the conclusion that `.mjs` is somehow provided - I guess this is some kind of smoke test checked into the repo?

> We don't currently offer .mjs helpers. I considered adding them, but was concerned about it potentially being more painful for users

Why?

> It's also not too bad to add .mjs versions later, I think.

Agreed.

loganfsmyth:
> This gave me the conclusion

Oh shoot, that's left over from when I was experimenting with support for it. It shouldn't have been committed.

> Why?

Mostly that the semantics of `.mjs` files still aren't official anywhere, especially around how they are loaded, and which one is loaded when, for instance if you don't provide a file extension.

packages/babel-plugin-transform-runtime/scripts/build-dist.js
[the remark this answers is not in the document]
hzoo: I guess we should explain somewhere about running this

loganfsmyth:
Alright I have:

* Renamed the `es6` folder to `esm`
* Renamed the `corejsVersion` option to `corejs`
* Removed accidentally-included `.mjs` files
* Deleted `babel-runtime/core-js.js` file (and not moved it to `babel-runtime-corejs2` since it isn't used as far as I can see.

@hzoo You had comments about the `core-js/` folder making `core-js` be repeated twice, is that still a concern you have?

@zloirock Any thoughts on my responses to your thoughts above?

hzoo:
LGTM

>  core-js/ folder making core-js be repeated twice, is that still a concern you have?

I'm ok if we don't change it, I understand it's nice to have it sorted by folder. Not sure I have a better suggestion

zloirock:
@loganfsmyth sorry for delayed reply. My opinion about kebab-case for `core-js` enough strong by the reason of searching. I don't think that it could make reading more difficult. I could change my opinion if a serious part of maintainers against kebab-case here.

loganfsmyth:
@zloirock Are there specific instances that you see in this PR that you disagree with? It would make discussions easier to have specific cases to discuss.

zloirock:
@loganfsmyth 
- `@babel/runtime-corejs2` package name
- `corejs` option of `@babel/plugin-transform-runtime`

nicolo-ribaudo:
I'm neutral to the package name, but I strongly oppose to the `core-js` option name. Our options (and the majority of the options used by JS packages) are all CamelCase; also, it would force users to use quotes instead of `babel.config.js`.

loganfsmyth:
@zloirock Perfect, my thoughts are:

* `@babel/runtime-corejs2`: 

  If `-` is the goal, the alternatives I see are
  * `@babel/runtime-core-js2`
  * `@babel/runtime-core-js-2`

  and my primary concern is that, to me at least, these seem uglier and harder to type for users. The first case makes it seem like `js2` is a standalone item, and the second is harder to type/remember due to the extra `-`s.

* `corejs` option
  My concern there is that it would require users to do quotes, e.g.

  ```
  plugins: [
    ['@babel/transform-runtime', { 'core-js': 2 }]
  ]
  ```
  which isn't something we've done before in Babel's options. So far I don't think any of them require quoting. We could also consider `coreJS: 2`. That at least avoids the need for quotes, and we could always add validation to look for `corejs` and tell people they forgot to captitalize it. I still don't love it, but at least there's more flexibility there.

Andarist:
> and my primary concern is that, to me at least, these seem uglier and harder to type for users. The first case makes it seem like js2 is a standalone item, and the second is harder to type/remember due to the extra -s.

I wouldnt be concerned about it that much, primary usage of this package will be through AST transformations of other plugins (like transform-runtime). People won't use it directly - they only have to install it which they most likely will do by copy-pasted from somewhere anyway.

> My concern there is that it would require users to do quotes, e.g.

This is imho a real concern, not sure what correct answer is but would be cool to align package name to whatever has been chosen for this one.

diervo:
@loganfsmyth I think as part of this refactor (although somewhat orthogonal) we should make sure all of the helpers can be imported via the right source. For example `AwaitValue` is just a magic helper we is used inside another helper via `import AwaitValue from "AwaitValue` which makes for some explicit hardcoding on bundling tools.

loganfsmyth:
@diervo Could you elaborate on that? I'm not sure I follow what you mean.

loganfsmyth:
> they only have to install it which they most likely will do by copy-pasted from somewhere anyway.

Installing was my main concern, but much less confident about users copy-pasting than you are. I guess for me I also just like that `runtime` and `corejs2` are clearly two separate items, whereas `runtime-core-js-2` makes it much harder to differentiate.

hzoo:
Ok it's been a while now and I believe we should move forward with what we have currently. We seem to be stalled on just the naming of the package/option name? cc @existentialism 

Personally I think the argument for searching isn't as important, as we don't reference `core-js` anywhere in babel before? It was an implementation detail that 
we are exposing now (unless that is actually a reason for doing it that way).

```
@babel/runtime-corejs2 // current
@babel/runtime-core-js2
@babel/runtime-core-js-2
@babel/runtime-core-js-v2
```

```js
plugins: [
  ['@babel/transform-runtime', { 'corejs': 2 }], // current
  ['@babel/transform-runtime', { 'core-js': 2 }]
]
```

I think given a few people (@nicolo-ribaudo, @Andarist) don't want to enforce the quotes usage and to keep `corejs`, the package name can follow and be the same, which suggests the current state of PR?

existentialism:
FWIW, I'm personally fine with `@babel/runtime-corejs2` and `corejs` as the option name.

rtsao:
This might be more long-term, but assuming the helpers/runtime package has a particular API (i.e. the names of helpers and how they are exported) maybe it would make sense for this option to simply be the name of the package containing the helpers (i.e. `"@babel/runtime"` or `"@babel/runtime-corejs2"`). I think this config API might have some a few possible advantages in that:
- It would allow for totally custom helper implementations for power users
- It avoids some of the awkwardness related to dashes and quoting
- It leaves out implementation details (i.e. no hard-coded corejs references).

loganfsmyth:
@rtsao 

> maybe it would make sense for this option to simply be the name of the package containing the helpers

I'd be open to that too. I had considered it, but it seemed a little verbose so I wasn't sure. I'm fine with `name: "@babel/runtime-corejs2"` instead of `corejs: 2` as the option if others are.

> assuming the helpers/runtime package has a particular API

Unfortunately, while the helpers do, the runtimes also for instance contain polyfill items that are specific to the given polyfill. Currently we rely on a 1-1 match between the logic in the transform, and the logic in the runtime module itself.

@diervo 

My concern with your goal is that I don't consider the specific structure of the `@babel/runtime` module to be a public API in the sense that you should rely on it in any of your own logic. We can't change it, since it'd be a breaking change, but it's not an API designed for developer use.

If your goal is to have a customized runtime, it seems like there isn't a ton to gain by layering it on top of Babel's own suggested runtime. Nothing that the transform/runtime modules do relys on private APIs, so you could take what we have and make your own runtime module and transform pair. It's not clear to me that making our own runtime transform fully user-configurable is a good thing though.

> However if we now go to the implementation of the object spread helper, you will see it depends on import defineProperty from "defineProperty";, which means I have to have some magical hardcoding list where "defineProperty" is special.

Could you elaborate here? The intention is that any `import` within a helper references some other helper, so by definition the `defineProperty` name string in the import references another property on the `helpers`. I wouldn't expect you to have to hard-code a list because the `helpers` object is already available.

> Helpers should be like any other APIs where developers can build on top.

This may be where we disagree. I'd consider the programmatic API exposed by `@babel/helpers` to be the API that developers can build around, but I would not consider the file structure of `@babel/runtime` to be something that would interact with. It is meant as a static dump of of `@babel/helpers` essentially.

diervo:
I agree with you on that any import brought from a 3party we can't treat as an "official" API since we don't own the content (however it will be nice if they would document it regardless).

My only ask is for all `@babel/helpers/...` to have the same prefix so we can resolve them in a predictable fashion.

`import x from "defineProperty"` => `import x from "babel/helpers/defineProperty"`

loganfsmyth:
@diervo 

`@babel/helpers` is a programmatic API. It does not expose the helpers as static files. That's a purposeful decision because `@babel/helpers` is meant as a runtime dependency and is primarily used via `@babel/core`.

`@babel/runtime` is specifically meant as a runtime-only dependency and this has minimal dependencies of its own, whereas `@babel/helpers` depends on `@babel/types` and such.

diervo:
Sorry maybe thats the confusion, I'm not talking about the module being an API, I'm talking about the helpers themselves being preditable.

In other words when babel brings a helper of the form `import foo from "@babel/helper/foo` the source of the import should follow a pattern, that is predictable.

"Any helper that babel brings must be of the form @babel/helper/{id}"
That invariant is not correct today (ex. defineProperties) and thats what I would love to standarize on.

If this is still unclear, we can do a quick call so I can show you the exact code

loganfsmyth:
> In other words when babel brings a helper of the form import foo from "@babel/helper/foo the source of the import should follow a pattern, that is predictable.

I do not think that `@babel/helper` should expose static file versions of the helpers. That is what `@babel/runtime` is for.

diervo:
As a developer I want to be able to control all the generated dependencies for a module (I believe that's why the runtime plugin today has the moduleName option).

Since I'm configuring those babel options, I should be able to predict all the set of helpers that might be used so I can resolved them accordingly.

After this thread I think this idea can be explored as a separated PR.

hzoo:
I think we can move forward on the current? If there's reason a reason to do otherwise we can republish as a different name, deprecate the old name and it won't need to be breaking.

```
@babel/runtime-corejs2 // current
@babel/runtime-core-js2
@babel/runtime-core-js-2
@babel/runtime-core-js-v2
```

```
plugins: [
  ['@babel/transform-runtime', { corejs: 2 }], // current
  ['@babel/transform-runtime', { 'core-js': 2 }]
  ['@babel/transform-runtime', { name: '@babel/runtime-corejs2' }]
]
```

hzoo:
Whoo, also merged https://github.com/babel/website/pull/1714. We should add to the migration guide too

benjamn:
I fully support this change and I'm glad it happened, but folks seem to be having trouble because
1. they're not using the latest version of `@babel/plugin-transform-runtime` but accidentally upgraded to `@babel/runtime@7.0.0-beta.56`, so the `helpers/builtin` directory no longer exists, or
2. they've upgraded to the latest `@babel/plugin-transform-runtime` but haven't upgraded `@babel/runtime` yet, so modules in the `helpers` directory still import `core-js` polyfills, or
3. they've upgraded `@babel/plugin-transform-runtime` but haven't yet installed `@babel/runtime-corejs2`, if that's necessary.

Some of this is just how it goes with breaking changes, but I have an idea that might help with case 1: what if the latest `@babel/runtime` package included a symlink from `@babel/runtime/helpers/builtin` to the parent directory, `@babel/runtime/helpers`, so that you could import built-in helpers from either directory? That makes sense because all the helpers are now built-in (in the sense that they do not import `core-js` polyfills), and because the previous contents of `@babel/runtime/helpers/builtin` were exactly the same as the list of modules in `@babel/runtime/helpers`.

If we did this, then `@babel/runtime` would continue to work both for code generated by the latest `@babel/plugin-transform-runtime` and for previously generated code, and we could remove the symlink in the future, once we're confident most folks have updated `@babel/plugin-transform-runtime`.

Thoughts?

nicolo-ribaudo:
We have been asking users not to use `^` in their `@babel/*` dependencies for ages, specifically to avoid problems like the one you are mentioning. I think that we can keep asking users to check the version of the installed packages, especially because we are close to the first rc so we are trying to avoid introducing things that will soon lead to breaking changes (e.g. removing the symlink).

benjamn:
Excited for the RC and the final release! These versioning puzzles will all be much easier to manage once Babel 7 is officially released, so we can go back to using semantic versioning to flag breaking changes like the removal of the `helpers/builtin` directory. It's just hard to know right now how significant each increment of `-beta.n` number is.

benjamn:
That said, if you've been asking people to do something for a long time and they still haven't done it, perhaps it's time to consider other ways of helping them fall into the Pit of Success™?

hzoo:
> That said, if you've been asking people to do something for a long time and they still haven't done it, perhaps it's time to consider other ways of helping them fall into the Pit of Success™?

Wanted to do this since the beginning but didn't know how and believe everyone else was ok with just telling people. Do know a good way of preventing it because other projects/people are going to whatever and it's hard to figure out how to know how to do in that case other than not releasing until the end. And even versioning it differently like `beta.n` to `c.n` doesn't fix the `^` issue. This is why I spent some time to submit https://github.com/npm/rfcs/pull/14 as well. In the end it's our fault, but it's also still beta and that comes with using beta which we have said there are intentional breaking changes. Believe me, I can't wait to finally release it either (still get people everyday asking me to release/when and to be honest it's not helping me mentally)

I think we don't have anything left so this shouldn't be more of an issue at least for this release moving forward..

insin:
> I don't think anybody really uses the current `moduleName` option now

nwb [depended on this](https://github.com/insin/nwb/issues/403#issuecomment-411065335) to have the transform-runtime plugin generate imports which pointed to the absolute path to babel-runtime in nwb's own dependencies, allowing it to serve and build stuff with all devDependencies managed in a global install - is there an alternative way to get Babel to resolve the runtime module when it's not going to be available in `require()` scope?

loganfsmyth:
@insin The original intention of `moduleName` was to allow custom runtimes, which isn't something we're looking to support in 7.x, but the usecase of passing an absolute path for the runtime itself seems entirely reasonable. 

I'd be happy to have an option of some kind to have Babel inject absolute paths. How would you feel about a `useAbsoluteRuntime` option? My preference would probably be to allow both:
* `useAbsoluteRuntime: true`
  * If the plugin is passed in the programmatic args, it resolves relative to Babel's `cwd`
  * If the plugin is passed via a config file, it resolve relative to that file
* `useAbsoluteRuntime: "/some/abs/dir"`
  * Resolve the runtime relative to some specific directory, which for `nwb` could just be `__dirname`

Thoughts?

insin:
@loganfsmyth Those both look good to me 👍

trusktr:
Hello awesome Babel devs, it'd be awesome to avoid in-range breaking changes, and it would be _highly_ valued and appreciated!

benjamn:
As disruptive as these changes might be, I think they're totally fair game for a beta release. Being more careful about breaking changes during beta just means it takes longer for Babel 7 to be officially released, which is the goal we're all aiming for. Once we drop the `-beta.n` suffix, we can go back to normal semantic versioning, which probably would have saved you from this particular pain.

loganfsmyth:
@insin 

> Would that also support passing the absolute path, as in path.dirname(require.resolve('@babel/runtime/package'))?

It would not, and I don't know that it's something I'd want to allow, because it should be the runtime transform itself deciding the actual name of the runtime, at least in my opinion. In the context of `nwb` though, since `require.resolve` resolves a package relative to `__dirname`, you'd get exactly the behavior of your current implementation by doing

```
plugins: [
  ['@babel/plugin-transform-runtime', { absoluteRuntime: __dirname }],
]
```
to resolve the runtime relative to `__dirname`.

loganfsmyth:
For the absolute-path discussion, I created https://github.com/babel/babel/pull/8435, so let's continue that over there.

nicolo-ribaudo:
If you (or a dependency who uses directly uses Babel) require a beta version, all your babel packages _and_ the packages used by your deps need to match that beta version.
I strongly suggest you to update your dependencies.

hijiangtao:
> If you (or a dependency who uses directly uses Babel) require a beta version, all your babel packages _and_ the packages used by your deps need to match that beta version.
> I strongly suggest you to update your dependencies.

Thanks for your reminding, I tried upgraded @babel/runtime's version in my repo, however the same problem reproduced in some dependencies inside the dependencies of my repo, which has got out of my control (I guess those dependencies require an old version of @babel/runtime and use the grammar like `.../es6/...`, a little award to express this situation, kind of `recall hell` in some aspect).

I'll look and see if there's some method to solve it, thanks.

PR #8267, opened 2018-07-05
Optimize async to generator

jridgewell:
We can reuse the step function, avoiding a closure on every invocation.

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -238,7 +238,7 @@ helpers.asyncGeneratorDelegate = () => template.program.ast`
 `;
 
 helpers.asyncToGenerator = () => template.program.ast`
-  function step(gen, resolve, reject, _next, _throw, key, arg) {
+  function asyncGeneratorStep(gen, resolve, reject, _next, _throw, key, arg) {
     try {
       var info = gen[key](arg);
       var value = info.value;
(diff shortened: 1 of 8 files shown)


packages/babel-helpers/src/helpers.js
```diff
@@ -238,30 +238,35 @@ helpers.asyncGeneratorDelegate = () => template.program.ast`
 `;
 
 helpers.asyncToGenerator = () => template.program.ast`
+  function step(gen, resolve, reject, _next, _throw, key, arg) {
```
loganfsmyth: I think it'd be worth giving this a more verbose name if we're splitting it out. Thoughts?

jridgewell: Sure.

packages/babel-helpers/src/helpers.js
```diff
@@ -238,30 +238,35 @@ helpers.asyncGeneratorDelegate = () => template.program.ast`
...
+          asyncGeneratorStep(gen, resolve, reject, _next, _throw, "next", value);
+        }
+        function _throw(err) {
+          asyncGeneratorStep(gen, resolve, reject, _next, _throw, "throw", err);
         }
-        function _next(value) { step("next", value); }
-        function _throw(err) { step("throw", err); }
 
-        _next();
+        _next(undefined);
```
Andarist: any particular reason for this explicit undefined?

jridgewell: Monomorphic calls. Probably negligible.

Andarist: is arg length important for monomorphism? always thought more about shapes in this regard

PR #8268, opened 2018-07-05
Refactor async-to-generator transform

jridgewell:
This uses the memoizable function declaration pattern to wrap the call
to `asyncToGenerator` factory. Then, it hoists that function declaration
as high as it'll go.

This should help with performance, and memory pressure, since we won't
be redefining that asyncGenerator wrapper every time we call the async
function.

https://jsbench.github.io/#fe2924043feb9085a27a2edb671ecec0

### DIFF
--- a/packages/babel-plugin-proposal-async-generator-functions/test/fixtures/nested/generator-in-async/output.js
+++ b/packages/babel-plugin-proposal-async-generator-functions/test/fixtures/nested/generator-in-async/output.js
@@ -1,22 +1,22 @@
-function f() {
-  return _f.apply(this, arguments);
+function _g() {
+  _g = babelHelpers.wrapAsyncGenerator(function* g() {
+    yield babelHelpers.awaitAsyncGenerator(2);
+    yield 3;
+  });
+  return _g.apply(this, arguments);
 }
 
 function _f() {
-  _f = babelHelpers.asyncToGenerator(function* () {
+  _f = babelHelpers.asyncToGenerator(function* f() {
     yield 1;
 
     function g() {
       return _g.apply(this, arguments);
     }
-
-    function _g() {
-      _g = babelHelpers.wrapAsyncGenerator(function* () {
-        yield babelHelpers.awaitAsyncGenerator(2);
-        yield 3;
-      });
-      return _g.apply(this, arguments);
-    }
   });
+  return _f.apply(this, arguments);
+}
+
+function f() {
   return _f.apply(this, arguments);
 }
--- a/packages/babel-plugin-transform-async-to-generator/test/fixtures/async-to-generator/expression/output.js
+++ b/packages/babel-plugin-transform-async-to-generator/test/fixtures/async-to-generator/expression/output.js
@@ -1,34 +1,29 @@
-var foo =
-/*#__PURE__*/
-function () {
-  var _ref = babelHelpers.asyncToGenerator(function* () {
-    var wat = yield bar();
-  });
+var foo = function () {
+  return _wrapped.apply(this, arguments);
+};
 
-  return function foo() {
-    return _ref.apply(this, arguments);
-  };
-}();
-
-var foo2 =
-/*#__PURE__*/
-function () {
-  var _ref2 = babelHelpers.asyncToGenerator(function* () {
-    var wat = yield bar();
-  });
-
-  return function foo2() {
-    return _ref2.apply(this, arguments);
-  };
-}(),
-    bar =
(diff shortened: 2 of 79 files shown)


packages/babel-traverse/src/path/lib/hoister.js
```diff
@@ -204,29 +206,38 @@ export default class PathHoister {
...
-    ]);
-
-    const parent = this.path.parentPath;
-    if (parent.isJSXElement() && this.path.container === parent.node.children) {
-      // turning the `span` in `<div><span /></div>` to an expression so we need to wrap it with
-      // an expression container
-      uid = t.JSXExpressionContainer(uid);
-    }
+    if (this.path.isDeclaration()) {
+      const [attached] = attachTo[insertFn](t.cloneNode(this.path.node));
```
Andarist: if we are going to remove this node from the old place right away (line below - `this.path.remove()`) cloning here seems excessive

jridgewell: I was worried about reusing the same nodes causing weird scoping issues. Like, the object identity in some scope map would cause it to reference the old path, or something.

jridgewell:
To guide the review a bit, here are 3 selected outputs:

1. [Function Declarations](https://github.com/babel/babel/pull/8268/files#diff-4a67664c7ffe3dd3fa3b69dc0107febf)
   - Minimal change, really just a reordering
2. [Function Expressions](https://github.com/babel/babel/pull/8268/files#diff-48d37dfb26fac628f22b04671ff526e8)
   - Removes an IIFE, removing need for `PURE`
3. [Methods](https://github.com/babel/babel/pull/8268/files#diff-c7dffcce27cb56d996e9d1268086cd61)
   - Wrapper is now extracted out of the function
   - This was the usecase I was targeting

packages/babel-plugin-transform-async-to-generator/test/fixtures/async-to-generator/object-method-with-super/output.js
```diff
@@ -1,12 +1,15 @@
-class Foo extends class {} {
-  method() {
-    var _superprop_callMethod = (..._args) => super.method(..._args);
+function _wrapped() {
+  _wrapped = babelHelpers.asyncToGenerator(function* () {
+    super.method();
```
jridgewell: This is a bug.

packages/babel-traverse/src/path/lib/hoister.js
Andarist: gotcha 👍

packages/babel-plugin-proposal-async-generator-functions/test/fixtures/nested/generator-in-async/output.js
```diff
@@ -1,22 +1,22 @@
...
+  _g = babelHelpers.wrapAsyncGenerator(function* g() {
+    yield babelHelpers.awaitAsyncGenerator(2);
+    yield 3;
+  });
+  return _g.apply(this, arguments);
 }
 
 function _f() {
-  _f = babelHelpers.asyncToGenerator(function* () {
+  _f = babelHelpers.asyncToGenerator(function* f() {
```
nicolo-ribaudo: Does it work if the asynchronous function refers to itself? This new `f` binding shadows the correct one.

packages/babel-plugin-transform-async-to-generator/test/fixtures/async-to-generator/expression/output.js
```diff
@@ -1,34 +1,29 @@
...
-/*#__PURE__*/
-function () {
-  var _ref2 = babelHelpers.asyncToGenerator(function* () {
+var foo2 = function () {
+  return _wrapped2.apply(this, arguments);
+},
+    bar = function () {
+  return _wrapped3.apply(this, arguments);
+},
+    function _wrapped2() {
```
nicolo-ribaudo: This doesn’t look correct to me. "_wrapped2" is a function expression whose value is discarded (it is not assigned to anything, only "_wrapped" is assigned to "bar". "_wrapped2" is not accessible from the function assigned to "foo2".

Andarist: good catch, it seems that `_wrapped2` and `_wrapped` were supposed to be declarations - like `_wrapped3`

wa-Nadoo:
image
It seems that Edge doesn't improve perf with this patch

PR #8274, opened 2018-07-05
Remove yearly presets from repo

hzoo:
These were already deprecated in v6, just didn't remove the folders?

Can skip the 2nd commit since it just deletes the folders. Also moved the folders to https://github.com/babel/babel-archive via https://github.com/babel/babel-archive/commit/5396cc3db08b7891a75e77c89e8f80a87279c446

- [x] need to update fixtures


packages/babel-cli/test/index.js
```diff
@@ -14,12 +14,10 @@ const fileFilter = function(x) {
   return x !== ".DS_Store";
 };
 
-const presetLocs = [
-  path.join(__dirname, "../../babel-preset-es2015"),
```
hzoo: Was going to switch to env but I don't think we really need to test for it if it's just the cli?

chicoxyzzy:
I see that yearly presets still exist in `packages/` directory and contain `package.json` file

hzoo:
Dang it 😅.. ok that is weird and I realize I had that commit in the Stage preset removal. So that last version included it but oh well.

PR #8286, opened 2018-07-07
Move v4-v6 changelog to another file and all prerelease 7.0 logs [ski…

hzoo:
changelog is rather long but yeah.. or I could try to regenerate all of it for a 7.0


hzoo:
Yeah I like that 👍, I'l do that and merge

PR #8289, opened 2018-07-08
Implement Smart Pipeline proposal in @babel/parser

mAAdhaTTah:
| Fixed Issues?            | Related to https://github.com/babel/proposals/issues/29
| Major: Breaking Change?  | Maybe? (`pipelineOperator` in @babel/parser now requires proposal option)
| Documentation PR Link    | N?A
| Any Dependency Changes?  |

Most of this work was implemented by the illustrious @js-choi, which some adjustments and refreshments from myself to use the new plugin options API. He's been fairly busy so I'm going to be working through the suggestions / changes here.

This is a big PR, but it enables `@babel/parser` to parse the Smart Pipeline proposal's topic reference. As discussed in https://github.com/babel/proposals/issues/29, this is the first step towards enabling the pipeline plugin to parse and emit all three proposals.

Next, I'll be implementing F#-style parsing, then we'll work on the pipeline plugin itself. I want to get them in piece by piece to avoid merge conflicts. This work had drifted quit a bit, and took me a few days to refresh @js-choi's work with the latest from `master`, so I'd like to land each piece one by one.

This is a pretty major chunk of work, so definitely interested in getting as much feedback as possible.

🎉 One step closer to the landing the pipeline proposals! 🎉


nicolo-ribaudo:
> Major: Breaking Change?	Maybe? (pipelineOperator in @babel/parser now requires proposal option)

Could you move it to a separate PR? Just in case this one doesn't get merged before the Babel 7 RC because it is a huge PR.

mAAdhaTTah:
@nicolo-ribaudo Done! #8291

mAAdhaTTah:
Refreshed from #8291, so this is ready to review again.

packages/babel-parser/src/tokenizer/types.js
```diff
@@ -86,6 +86,7 @@ export const types: { [name: string]: TokenType } = {
   regexp: new TokenType("regexp", { startsExpr }),
   string: new TokenType("string", { startsExpr }),
   name: new TokenType("name", { startsExpr }),
+  primaryTopicReference: new TokenType("#", { startsExpr }),
```
nicolo-ribaudo: Why is this token needed? Can’t we just reuse tt.hash?

packages/babel-parser/test/fixtures/experimental/pipeline-operator/invalid-proposal/options.json
```diff
@@ -2,5 +2,5 @@
   "plugins": [
     ["pipelineOperator", { "proposal": "invalid" }]
   ],
-  "throws": "'pipelineOperator' requires 'proposal' option whose value should be one of: minimal"
+  "throws": "'pipelineOperator' requires 'proposal' option whose value should be one of: minimal, smart"
```
nicolo-ribaudo: Nit: the proposal names should be quoted.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1920,4 +1988,206 @@ export default class ExpressionParser extends LValParser {
...
+    return this.finishNode(bodyNode, pipelineStyle);
+  }
+
+  checkSmartPipelineBodyStyle(
+    expression: N.Expression,
+  ):
+    | "PipelineBareFunction"
+    | "PipelineBareConstructor"
+    | "PipelineBareAwaitedFunction"
+    | "PipelineTopicExpression" {
```
nicolo-ribaudo: You can reuse the PipelineStyle (defined later) here.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1920,4 +1988,206 @@ export default class ExpressionParser extends LValParser {
...
+
+    const callbackResult = callback();
+
+    this.state.topicContextState = outerContextTopicState;
+    return callbackResult;
+  }
+
+  readTopicContextState(): tt.TopicContextState {
+    return this.state.topicContextState;
+  }
```
nicolo-ribaudo: Is this function really needed?

packages/babel-parser/src/parser/expression.js
```diff
@@ -1920,4 +1988,206 @@ export default class ExpressionParser extends LValParser {
...
+
+  checkSmartPipelineBodyEarlyErrors(
+    childExpression: N.Expression,
+    pipelineStyle: PipelineStyle,
+    startPos: number,
+  ): void {
+    if (this.match(tt.arrow)) {
+      // If the following token is invalidly `=>`, then throw a human-friendly error
+      // instead of something like 'Unexpected token, expected ";"'.
```
nicolo-ribaudo: ❤️

packages/babel-parser/src/tokenizer/state.js
```diff
@@ -33,6 +33,12 @@ export default class State {
     this.hasFlowComment = false;
     this.isIterator = false;
 
+    // Used by smartPipelines.
+    this.topicContextState = {
```
nicolo-ribaudo: Nit: this can be renamed to "topicContext", since "State" is already part of the class name.

packages/babel-parser/src/types.js
```diff
@@ -565,6 +565,32 @@ export type SequenceExpression = NodeBase & {
   expressions: $ReadOnlyArray<Expression>,
 };
 
+// Pipelines
+
+export type PipelineBody = NodeBase & {
+  type: "PipelineBody",
+};
+
+export type PipelineBareFunctionBody = PipelineBody & {
```
nicolo-ribaudo: I think that in every other place these node types don’t contain "Body"?

nicolo-ribaudo: Also, they can’t extend PipelineBody because the "type" properties are not compatible

mAAdhaTTah:
Thanks for the review! As for the question

> Question: are `x |> await foo` and `x |> (await foo)` the same thing?

No; `x |> await foo` is banned in every proposal, as its semantics are prone to confusion by users. `x |> (await foo)` desugars to:

```js
const _temp = await foo;
_temp(x);
```

in every proposal. Smart pipelines enables `x |> await foo(#)`, whereas F# enables `x |> foo |> await`. Both desugar to `await foo(x)`.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -404,6 +404,43 @@ export default class Tokenizer extends LocationParser {
...
+      "smart" === this.getPluginOption("pipelineOperator", "proposal")
+    ) {
+      if (next >= charCodes.digit0 && next <= charCodes.digit9) {
+        this.raise(
+          this.state.pos,
+          `Unexpected digit after topic reference: '#${String.fromCodePoint(
+            next,
+          )}'`,
+        );
+      } else {
```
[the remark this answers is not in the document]
mAAdhaTTah: Is this a better error message than the "unexpected digit" message? If you prefer it, we could lift that up so all `#4`-type combinations are invalid.

packages/babel-parser/src/tokenizer/types.js
mAAdhaTTah: On first blush, it conflicts with the private properties plugin, but I'll try to reuse that.

packages/babel-parser/src/parser/expression.js
mAAdhaTTah: I can inline this with the property assignment.

mAAdhaTTah:
Just pushed fixes or responses for everything except fixing the Pipeline types, which I have to dig into.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1920,4 +1988,192 @@ export default class ExpressionParser extends LValParser {
...
+    this.state.topicContext = {
+      // Enable the use of the primary topic reference.
+      maxNumOfResolvableTopics: 1,
+      // Hide the use of any topic references from outer contexts.
+      maxTopicIndex: null,
+    };
+
+    const callbackResult = callback();
+
+    this.state.topicContext = outerContextTopicState;
```
nicolo-ribaudo: Lines 2129-2131 should be wrapped in a `try/finally` block. The typescript plugin throws and catches exceptions to diambiguate some tokens, so the `callback` here could throw an error.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1920,4 +1988,192 @@ export default class ExpressionParser extends LValParser {
...
+    this.state.topicContext = {
+      // Disable the use of the primary topic reference.
+      maxNumOfResolvableTopics: 0,
+      // Hide the use of any topic references from outer contexts.
+      maxTopicIndex: null,
+    };
+
+    const callbackResult = callback();
+
+    this.state.topicContext = outerContextTopicState;
```
nicolo-ribaudo: Also here

mAAdhaTTah:
I addressed the comments so far, so this is ready for review again.

packages/babel-parser/src/parser/expression.js
```diff
@@ -886,6 +932,33 @@ export default class ExpressionParser extends LValParser {
         }
       }
 
+      case tt.hash: {
+        if (this.state.inPipeline) {
+          this.expectPlugin("pipelineOperator");
+          node = this.startNode();
+
+          if (
+            "smart" !== this.getPluginOption("pipelineOperator", "proposal")
```
nicolo-ribaudo: Please swap these operands.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1920,4 +1993,194 @@ export default class ExpressionParser extends LValParser {
...
+        return this.isSimpleReference(expression)
+          ? "PipelineBareFunction"
+          : "PipelineTopicExpression";
+    }
+  }
+
+  isSimpleReference(expression: N.Expression): boolean {
+    switch (expression.type) {
+      case "MemberExpression":
+        return this.isSimpleReference(expression.object);
```
nicolo-ribaudo: This should also check that `expression` isn’t a computed property access (e.g. `foo[bar]`).

packages/babel-parser/src/parser/expression.js
```diff
@@ -1920,4 +1993,194 @@ export default class ExpressionParser extends LValParser {
...
+        }
+        bodyNode.expression = childExpression;
+        break;
+      default:
+        throw this.raise(startPos, `Unknown pipeline style ${pipelineStyle}`);
+    }
+    return this.finishNode(bodyNode, pipelineStyle);
+  }
+
+  checkSmartPipelineBodyStyle(expression: N.Expression): PipelineStyle {
```
nicolo-ribaudo: By reading the proposal specification (https://jschoi.org/18/es-smart-pipelines/spec#sec-pipe-operator, “Supplemental Syntax” section) it looks like only `foo.bar.baz` are considered simple expressions, everything else (e.g. `new Foo` and `await foo`) are parsed as `PipelineTopicStep`.

`new` and `await` are introduced, respectively, by the BC and BA additional features. I think that this PR should only focus on the core smart pipelines proposal, and any addition should be introduced by a separate flag (maybe an array of feature). We can do it in a separate PR to keep this simpler for other reviewers.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -403,7 +403,44 @@ export default class Tokenizer extends LocationParser {
...
+    const nextPos = this.state.pos + 1;
+    const next = this.input.charCodeAt(nextPos);
+    if (
+      (this.hasPlugin("classPrivateProperties") ||
+        this.hasPlugin("classPrivateMethods")) &&
+      this.state.classLevel > 0
+    ) {
+      ++this.state.pos;
+      this.finishToken(tt.hash);
+      return;
```
nicolo-ribaudo: I think that we can make this function simpler by handling private names and topic references the same way (always throwing if there is a digit).
As it is written now, I think that the nice error thrown in the next `if` won’t be thrown for `#1` references inside classes when the private fields plugin is enabled.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -651,25 +688,12 @@ export default class Tokenizer extends LocationParser {
 
   getTokenFromCode(code: number): void {
     switch (code) {
-      case charCodes.numberSign:
-        if (this.state.pos === 0 && this.readToken_interpreter()) {
-          return;
-        }
+      // The interpretation of a number sign "#" depends on whether it is
+      // followed by an identifier or not.
```
nicolo-ribaudo: I think that we can remove this comment, since the `readToken_numberSign` always return `tt.hash` now.

packages/babel-parser/src/tokenizer/index.js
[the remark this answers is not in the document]
nicolo-ribaudo: You are right, the current error message is better.

packages/babel-parser/src/tokenizer/index.js
mAAdhaTTah: The way it's currently set up, the error is thrown elsewhere. I moved that `if` block up before the `classLevel` check, which resulted in this test failure:

> SyntaxError: experimental/class-private-properties/failure-numeric-start-identifier/input.js: Expected error message: Identifier directly after number (2:4). Got error message: Unexpected digit after topic reference: '#2' (2:2)

Itself, not a really big deal, but I realized that putting the `if` upfront doesn't distinguish between class properties & topic tokens. Should we simplify the message so it applies generically to both (e.g. "Unexpected digit after hash token") or fork the logic there based on whether we're in a class, or something else entirely?

packages/babel-parser/src/parser/expression.js
mAAdhaTTah: True, the smaller the PR, the easier it'll be to land. I'll leave that function there so I can put the logic in a separate PR easily (as opposed to inlining the `switch` statement.

packages/babel-parser/src/parser/expression.js
mAAdhaTTah: I added a test for this, but it parsed as a member expression. That seems... wrong? But I'm having trouble visualizing the AST, so I can't tell how they differ. Thoughts?

mAAdhaTTah:
@nicolo-ribaudo Some questions in comments; I pushed the latest commits + associated code.

packages/babel-parser/src/parser/expression.js
nicolo-ribaudo: `foo[bar]` has a `computed: true` property that `foo.bar` doesn't have.
You can test it on astexplorer.net: 
https://astexplorer.net/#/gist/5a1b8e1eb1075f95587c222b320285cf/ea034bcff7ccb53ab81fa5bd1fc8b44328fe3911

packages/babel-parser/src/tokenizer/index.js
nicolo-ribaudo: +1 for `Unexpected digit after hash token`: users only using private properties may not know what a "topic reference" is.

packages/babel-parser/src/parser/expression.js
mAAdhaTTah: Fixed!

packages/babel-parser/src/tokenizer/index.js
mAAdhaTTah: Updated the error message.

mAAdhaTTah:
PR is updated & ready for review again.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1920,4 +1993,188 @@ export default class ExpressionParser extends LValParser {
...
+
+  topicReferenceWasUsedInCurrentTopicContext(): boolean {
+    return (
+      this.state.topicContext.maxTopicIndex != null &&
+      this.state.topicContext.maxTopicIndex >= 0
+    );
+  }
 }
+
+type PipelineStyle =
```
xtuc: I'm not sure that's the correct place to put type definitions?

mAAdhaTTah: Where should this go?

packages/babel-parser/src/parser/expression.js
```diff
@@ -296,18 +296,20 @@ export default class ExpressionParser extends LValParser {
...
 
         this.next();
 
-        const startPos = this.state.start;
-        const startLoc = this.state.startLoc;
-
-        if (op === tt.pipeline) {
+        if (
+          op === tt.pipeline &&
+          this.getPluginOption("pipelineOperator", "proposal") === "minimal"
```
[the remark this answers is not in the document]
mAAdhaTTah: Constants would be nice, but the proposals are used both by `@babel/parser` & the plugin itself. Where should we put them so they can be shared?

packages/babel-parser/src/parser/expression.js
thiagoarrais: Just a hunch here: should it go to `parser/types.js`? The structure of the declaration looks a lot like `AssignmentOperator` and `BinaryOperator`.

littledan:
It'd be great to have a review of this PR. Many people are interested in the pipeline proposal progressing, and although we're not sure if this alternative will be the selected one, we're trying to go for an approach of proposing multiple alternatives and getting developer feedback about them. Thanks!

packages/babel-parser/src/parser/expression.js
mAAdhaTTah: Done!

mAAdhaTTah:
This PR has been refreshed with the latest from master, and I addressed the comment regarding where to put the `PipelineStyle` type. I know this is a large PR, but I believe it's ready to go and would love to get this landed ASAP so we can start moving forward on experimenting with the proposal.

packages/babel-parser/src/parser/expression.js
```diff
@@ -910,6 +956,33 @@ export default class ExpressionParser extends LValParser {
...
+            this.raise(
+              node.start,
+              "Primary Topic Reference found but pipelineOperator not passed 'smart' for 'proposal' option.",
+            );
+          }
+
+          this.next();
+          if (this.primaryTopicReferenceIsAllowedInCurrentTopicContext()) {
+            this.registerTopicReference();
+            return this.finishNode(node, "PrimaryTopicReference");
```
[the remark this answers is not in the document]
xtuc: I mean only proposal's mode specific nodes.

xtuc:
We can do it in a follow up PR, but we should make sure that https://github.com/babel/babel/blob/master/packages/babel-parser/ast/spec.md keeps updated, we already have the flow definition.

packages/babel-parser/test/fixtures/experimental/pipeline-operator/proposal-smart-bare-style,-function-calls,-chained,-multiline/input.js
```diff
@@ -0,0 +1,4 @@
+let result = "hello"
+|> doubleSay
+|> text.capitalize
+|> a.b.exclaim;
```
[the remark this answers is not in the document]
igabesz: I see no problem on Win10

cmd.exe:
```
w:\tmp>mkdir "hello,world"
w:\tmp>cd "hello,world"
w:\tmp\hello,world>
```

powershell:
```
PS W:\tmp> mkdir "hello,again"
PS W:\tmp> cd ".\hello,again\"
PS W:\tmp\hello,again>
```

packages/babel-parser/src/parser/expression.js
```diff
@@ -910,6 +956,33 @@ export default class ExpressionParser extends LValParser {
         }
       }
 
+      case tt.hash: {
+        if (this.state.inPipeline) {
+          this.expectPlugin("pipelineOperator");
```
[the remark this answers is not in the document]
mAAdhaTTah: Done.

packages/babel-parser/src/parser/expression.js
```diff
@@ -296,18 +296,20 @@ export default class ExpressionParser extends LValParser {
         }
 
         const op = this.state.type;
-        if (op === tt.nullishCoalescing) {
+
+        if (op === tt.pipeline) {
+          this.state.inPipeline = true;
```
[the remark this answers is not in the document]
mAAdhaTTah: True, done.

packages/babel-parser/src/parser/expression.js
```diff
@@ -348,6 +345,55 @@ export default class ExpressionParser extends LValParser {
...
+          const startLoc = this.state.startLoc;
+          return this.withTopicPermittingContext(() => {
+            return this.parseSmartPipelineBody(
+              this.parseExprOpBaseRightExpr(op, prec, noIn),
+              startPos,
+              startLoc,
+            );
+          });
+        } else {
+          return this.parseExprOpBaseRightExpr(op, prec, noIn);
```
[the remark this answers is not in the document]
mAAdhaTTah: I think letting it fall through would make eslint, although I'm checking it now. Generally not a fan of that kind of implicit behavior.

packages/babel-parser/src/parser/expression.js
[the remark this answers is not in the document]
mAAdhaTTah: There are several other new AST nodes that are prefixed with `Pipeline`. Some of those nodes are shared between the proposals. I can change this one to be `PipelinePrimaryTopicReference` but I don't know as I would subdivide them by proposal.

packages/babel-parser/src/parser/expression.js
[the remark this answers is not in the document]
mAAdhaTTah: ESLint doesn't complain (maybe it should?), so I can do this if you want.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1956,4 +2029,182 @@ export default class ExpressionParser extends LValParser {
...
+    return this.finishNode(bodyNode, pipelineStyle);
+  }
+
+  checkSmartPipelineBodyStyle(expression: N.Expression): N.PipelineStyle {
+    switch (expression.type) {
+      default:
+        return this.isSimpleReference(expression)
+          ? "PipelineBareFunction"
+          : "PipelineTopicExpression";
+    }
```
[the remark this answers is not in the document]
mAAdhaTTah: This originally had some additional logic for some of Smart's follow-on proposals, but we decided to remove them and leave the `switch` in place to make adding those follow-on proposals easier. See here: https://github.com/babel/babel/pull/8289#discussion_r207909339

If you disagree, I can remove, but that's why at least.

packages/babel-parser/src/parser/expression.js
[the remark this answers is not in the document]
danez: I would prefer fall-through too, but with comment saying that we fall through. like here: https://github.com/babel/babel/blob/a2afb974be2c9aba0bf0bc092ffb08e172f7b847/packages/babel-parser/src/plugins/typescript.js#L1171

mAAdhaTTah:
@xtuc Updated as per comments, responded to some of your comments inline. Made a note to update the spec.md.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -374,7 +374,39 @@ export default class Tokenizer extends LocationParser {
...
+      this.finishToken(tt.hash);
+      return;
+    } else if (
+      this.getPluginOption("pipelineOperator", "proposal") === "smart"
+    ) {
+      this.finishOp(tt.hash, 1);
+    } else {
+      this.raise(
+        this.state.pos,
+        `Unexpected character '${String.fromCodePoint(code)}'`,
```
nicolo-ribaudo: Nit: we already know that `String.fromCodePoint` will return `#`

packages/babel-parser/src/types.js
```diff
@@ -566,6 +566,38 @@ export type SequenceExpression = NodeBase & {
...
+  callee: Expression,
+};
+
+export type PipelineTopicBody = NodeBase & {
+  type: "PipelineTopicBody",
+  expression: Expression,
+};
+
+export type PipelineStyle =
+  | "PipelineBareFunction"
```
nicolo-ribaudo: Is this the only non-node export of this file? I'm not sure that this is the right place where it should be put, but also I can't think of a better place :confused:

packages/babel-parser/test/fixtures/experimental/pipeline-operator/proposal-smart-bare-style,-function-calls,-chained,-multiline/input.js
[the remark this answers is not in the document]
nicolo-ribaudo: Yeah, we are already using commas in babel-cli tests

packages/babel-parser/src/parser/expression.js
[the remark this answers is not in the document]
mAAdhaTTah: Done.

packages/babel-parser/src/tokenizer/index.js
mAAdhaTTah: Done.

packages/babel-parser/src/types.js
mAAdhaTTah: I mean, the file is called `types.js` so maybe it's ok?

packages/babel-parser/src/parser/expression.js
```diff
@@ -1956,4 +2029,182 @@ export default class ExpressionParser extends LValParser {
...
+  checkPipelineAtInfixOperator(left: N.Expression, leftStartPos: number) {
+    this.expectPlugin("pipelineOperator");
+
+    if (this.getPluginOption("pipelineOperator", "proposal") === "smart") {
+      if (left.type === "SequenceExpression") {
+        // Ensure that the pipeline head is not a comma-delimited
+        // sequence expression.
+        throw this.raise(
+          leftStartPos,
+          `Pipeline head may not be a comma-separated sequence expression`,
```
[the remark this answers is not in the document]
mAAdhaTTah: Done.

thiagoarrais:
@mAAdhaTTah: 🍾🍾🍾

mAAdhaTTah:
Thanks so much! I think @xtuc is the last holdout. Any remaining changes you need?

xtuc:
Also thanks for your contribution and sorry if that took so long on our side (or at least mine)!

thiagoarrais:
That's awesome, y'all! I can now get the PR for the transform going.

js-choi:
Thanks everyone for the hard work with the update and review!

PR #8291, opened 2018-07-09
Support pipeline proposal flag in  `@babel/parser`

mAAdhaTTah:

Split from #8289; implements the flag to babylon so babel can move to an RC and still land that PR without shipping breaking changes. Should be easy to rebase this back into the PR.


nicolo-ribaudo:
I'd prefer the validation to happen at https://github.com/babel/babel/blob/dd6da3b3af7a09833ef5d90079ca833985eed7dc/packages/babel-parser/src/plugin-utils.js#L21, so that it is checked even if the pipeline operator isn't actually used.

mAAdhaTTah:
@nicolo-ribaudo Fixed.

PR #8293, opened 2018-07-09
Remove Stage presets

hzoo:
Sad times 😭, but for the best? Fun because I added them back in originally 😛 

Fixes https://github.com/babel/babel/issues/7770

- [x] Deprecated use of these presets in the last release (beta.52), for an initial warning.
- [x] Throw error with the use in next beta, show the correct config to use otherwise. Optional: Add autofix with https://github.com/babel/babel-upgrade/issues/67
- [x] Update the readme to explain better.
- [x] copy old code to babel-archive: https://github.com/babel/babel-archive/commit/c36fc7ddade1b82c86dba900289c7656c3491c13
- [x] test the Stage options in the REPL works?
- Check the readme and src/index.js of the Stages https://github.com/babel/babel/blob/b2f42e8cb4889e6fbff97e12d3db3dd0be0bfbd5/packages/babel-preset-stage-0/README.md
- [ ] writing a post to explain the changes

We should prob make a better guide on making your own preset

### DIFF
--- a/packages/babel-preset-stage-0/README.md
+++ b/packages/babel-preset-stage-0/README.md
@@ -1,19 +1,33 @@
 # @babel/preset-stage-0
 
-> Babel preset for stage 0 plugins
+As of v7.0.0-beta.52, we've decided to remove
+the official Babel Stage presets. You can find more information
+at issue #7770: https://github.com/babel/babel/issues/7770, but
+the TL;DR is that it's causing more harm than convenience in that
+the preset is always out of date, each change is usually going to
+require a major version bump and thus people will be behind,
+and it encouraged too many people to opt-in to too many proposals
+that they didn't intend to.
+If you want the same configuration as before, you can use this configuration,
+although keep in mind that Stage 0 contains Stage 1 which is also deprecated.
 
-See our website [@babel/preset-stage-0](https://babeljs.io/docs/en/next/babel-preset-stage-0.html) for more information.
-
-## Install
-
-Using npm:
-
-```sh
-npm install --save-dev @babel/preset-stage-0
+```json
+{
+  "presets": [
+    [
+      "@babel/preset-stage-1", { 
+        "loose": false,
+        "useBuiltIns": false,
+        "decoratorsLegacy": true,
+        "pipelineProposal": "minimal"
+      }
+    ]
+  ],
+  "plugins": ["@babel/plugin-proposal-function-bind"]
+}
 ```
 
-or using yarn:
-
-```sh
-yarn add @babel/preset-stage-0 --dev
-```
+This will be the last publish of "@babel/preset-stage-0", and it won't be
+in the final release.
+If it's a hassle to maintain, you can certainly make your own preset to use
+across projects, or there might be one in the community to use.
--- a/packages/babel-preset-stage-0/README.md
+++ b/packages/babel-preset-stage-0/README.md
@@ -1,15 +1,17 @@
 # @babel/preset-stage-0
 
-As of v7.0.0-beta.52, we've decided to remove
+> As of v7.0.0-beta.52, we've decided to remove
 the official Babel Stage presets. You can find more information
-at issue #7770: https://github.com/babel/babel/issues/7770, but
+at issue [#7770](https://github.com/babel/babel/issues/7770), but
 the TL;DR is that it's causing more harm than convenience in that
 the preset is always out of date, each change is usually going to
 require a major version bump and thus people will be behind,
 and it encouraged too many people to opt-in to too many proposals
 that they didn't intend to.
-If you want the same configuration as before, you can use this configuration,
-although keep in mind that Stage 0 contains Stage 1 which is also deprecated.
+
+---
+
+If you want the same configuration as before, you can use this configuration, although keep in mind that Stage 0 contains Stage 1 which is also deprecated.
 
 ```json
 {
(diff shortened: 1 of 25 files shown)


hzoo:
> I think we can add list of plugins down to current preset instead of higher stage preset + list of plugins for current preset.

Yeah I was thinking that too.

packages/babel-preset-stage-0/README.md
```diff
@@ -1,15 +1,17 @@
...
 the preset is always out of date, each change is usually going to
 require a major version bump and thus people will be behind,
 and it encouraged too many people to opt-in to too many proposals
 that they didn't intend to.
-If you want the same configuration as before, you can use this configuration,
-although keep in mind that Stage 0 contains Stage 1 which is also deprecated.
+
+---
+
+If you want the same configuration as before, you can use this configuration, although keep in mind that Stage 0 contains Stage 1 which is also deprecated.
```
chicoxyzzy: This lines in readmes should be reworded somehow (also probably same for deprecation messages) =)

packages/babel-standalone/src/preset-stage-3.js
```diff
@@ -0,0 +1,21 @@
...
+import syntaxImportMeta from "@babel/plugin-syntax-import-meta";
+import transformClassProperties from "@babel/plugin-proposal-class-properties";
+import transformJsonStrings from "@babel/plugin-proposal-json-strings";
+
+export default (_, opts) => {
+  let loose = false;
+
+  if (opts !== undefined) {
+    if (opts.loose !== undefined) loose = opts.loose;
+  }
```
nicolo-ribaudo: Nit: We can just use `{ loose = false } = {}` instead of the `opts` parameter to avoid lines 7-11.

packages/babel-standalone/src/preset-stage-2.js
```diff
@@ -0,0 +1,22 @@
...
+import transformFunctionSent from "@babel/plugin-proposal-function-sent";
+import transformExportNamespaceFrom from "@babel/plugin-proposal-export-namespace-from";
+import transformNumericSeparator from "@babel/plugin-proposal-numeric-separator";
+import transformThrowExpressions from "@babel/plugin-proposal-throw-expressions";
+
+export default (_, opts = {}) => {
+  const { loose = false, useBuiltIns = false, decoratorsLegacy = false } = opts;
+
+  return {
+    presets: [[presetStage3, { loose, useBuiltIns }]],
```
nicolo-ribaudo: `useBuiltIns` can be removed, since it was only used by object-rest-spread.
It can also be removed in the stage-0 and stage-1 presets.

packages/babel-preset-stage-0/src/index.js
```diff
@@ -1,57 +1,54 @@
...
 
-  if (typeof pipelineProposal !== "string") {
-    throw new Error(
-      "The pipeline operator requires a proposal set." +
-        " You must pass 'pipelineProposal' option to" +
-        " @babel/preset-stage-0 whose value must be one of: " +
-        proposals.join(", "),
-    );
-  }
+We recommend that make your own presets to use across projects for
```
existentialism: Nit:

We recommend that _you_ make your own presets...

packages/babel-preset-stage-0/README.md
```diff
@@ -1,19 +1,58 @@
...
+    "@babel/plugin-syntax-dynamic-import",
+    "@babel/plugin-syntax-import-meta",
+    ["@babel/plugin-proposal-class-properties", { "loose": false }],
+    "@babel/plugin-proposal-json-strings"
+  ]
+}
 ```
 
-or using yarn:
+We recommend that make your own presets to use across projects for
```
loganfsmyth: Is this something we want to recommend? I'm not sure how I feel about this being in here as a recommendation for good.

packages/babel-standalone/src/preset-es2015.js
```diff
@@ -35,19 +34,6 @@ export default (_, opts) => {
     if (opts.spec !== undefined) spec = opts.spec;
   }
 
-  if (typeof loose !== "boolean") {
```
loganfsmyth: What happened with this?

packages/babel-preset-stage-0/README.md
hzoo: Hmm yeah, I guess it shouldn't have to be a recommendation but more that people tend to forget you can make your own

packages/babel-standalone/src/preset-es2015.js
hzoo: Oh hmm I think I ended up removing this since it's for standalone and it would be a UI option in the repl, but we can add it back too.

packages/babel-preset-stage-0/README.md
loganfsmyth: Yeah, I'm fine if we want to say that creating a custom preset is an option, I think the wording right now just makes it sound like what we're expecting everyone to do, which isn't good.

Maybe something like

> If you're using the same configuration across many separate projects, keep in mind that you can also create your own custom presets with whichever plugins and presets you're looking to use.

packages/babel-preset-stage-0/README.md
```diff
@@ -1,19 +1,58 @@
...
+reusability. This can be as simple as exporting a function that returns your config/array of plugins.
 
-```sh
-yarn add @babel/preset-stage-0 --dev
+```js
+module.exports = function() {
+  return {
+    plugins: [
+      // ...
```
loganfsmyth: If we want this example, we should probably also clarify that doing `require()` on these is recommended. If they just do `"@babel/plugin-proposal-function-bind"` as a string, the plugins will probably not be found properly. I'd actually thought about making core error when that happens, but I haven't yet. Maybe I should?

packages/babel-preset-stage-0/README.md
```diff
@@ -1,19 +1,64 @@
...
+---
 
-## Install
+If you want the same configuration as before:
 
-Using npm:
+```json
+{
+  "plugins": [
+    // Stage 0
```
xtuc: Comments in JSON are not valid, could we remove them? (because user will need to) or can we switch to js?

packages/babel-preset-stage-0/src/index.js
```diff
@@ -1,39 +1,60 @@
...
+of "@babel/preset-stage-0"
 
-import transformFunctionBind from "@babel/plugin-proposal-function-bind";
+If you want the same configuration as before:
 
-export default declare((api, opts = {}) => {
-  api.assertVersion(7);
+{
+  "plugins": [
+    // Stage 0
```
xtuc: Same comment about JS comments here

xtuc:
Could we also remove the stage plugins from the REPL? I think that the possibility of loading plugins from npm would suffice? IRC parser flags were not pushed, we might need to fix that first.

packages/babel-preset-stage-0/README.md
nicolo-ribaudo: `.babelrc` uses Json5, so comments work.

hzoo:
Removing from repl isn't part of this PR since I left it in babel-standalone so we are leaving it for now? Yeah would like to figure out how we want to go about doing that. Don't have a good solution and is simple to do atm. I think the use case for the repl is different since it's fine to just allow any syntax mostly and we can easily keep it up to date since it's just for testing

mAAdhaTTah:
It might be confusing for people to find presets on the repl that don't exist / shouldn't be used anymore.

scripts/generators/readmes.js
```diff
@@ -66,6 +66,7 @@ yarn add ${name} --dev
 
 packages
   .filter(x => x !== "README.md") // ignore root readme
+  .filter(x => x.indexOf("babel-preset-stage-") === -1) // ignore stages
```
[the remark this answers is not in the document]
hzoo: That's fine but just FYI this is a temporary change because we are going to not publish any more of the package anyway so I would remove this code the next release after.

hzoo:
I think that's fine, this is only for v7 while we are figuring out what to do.

Issue #8195, opened 2018-06-18
RFC: Better babel-preset-env API


rtsao:
## Feature Request

**Is your feature request related to a problem? Please describe.**

First off, major props for creating such an incredibly awesome feature 😄. I think it's fair to say the `env` preset is truly game-changing and should be regarded as a huge milestone in advancing the JS ecosystem forward.

However, I think at times the existing configuration and mechanics of the `env` preset is sometimes unintuitive, suboptimal, and/or lacking for certain common use cases.

Various aspects of this have already been discussed in several related issues:

- https://github.com/babel/babel/issues/6830
- https://github.com/babel/babel/issues/8155
- https://github.com/babel/babel/issues/5699
- https://github.com/babel/babel/issues/6628
- https://github.com/babel/babel/issues/7267
- https://github.com/babel/babel/issues/6626
- https://github.com/babel/babel/issues/6613
  - https://github.com/babel/babel/pull/6731

**Describe the solution you'd like**

To address many of these issues, I propose the following configuration interface for the `env` preset:

**`options.targets`**

- `targets["chrome" | "firefox" | "edge" | ...]` *(unchanged)*
- `targets.browsers` *(unchanged)*
- `targets.node` *(unchanged)*
- `targets.esmodules` *(unchanged)*
- ❌ ~~`targets.uglify`~~ Note: already deprecated
- 🆕 `targets.spec`

This option is used to target a specific minimum ES specification, with support for the following values:
 - `"es5"`
 - `"es2015"`
 - `"es2016"`
 - `"es2017"`
 - `"es2018"`

This option is useful when targeting various tools using a parser supporting a particular ECMAScript specification (e.g. Uglify, webpack, etc.).

This option is also especially useful for library authors, allowing them to explicitly specify a minimum supported version of ECMAScript for their library, or even publish multiple variants of their code.

❌ ~~**`options.forceAllTransforms`**~~

Replaced by `options.target.spec`. In my opinion, `{forceAllTransforms: true}` is really just a more amiguous way of saying `{targets: {spec: "es5"}}`, which also reduces the configuration surface area by elegantly fitting into the existing "targets" paradigm.

**`options.modules`**

*(unchanged)*

🆕 **`options.defaultTransformMode`**

This option replaces the exising top-level `spec` and `loose` options for the preset and supports the following values:

- `"performance"` - When possible, bias plugin configuration towards high performance, (i.e. `spec: false`, `loose: true` for transforms that support these options)
[...]

PR #8296, opened 2018-07-09
Add transformMode option to env preset

rtsao:
| Fixed Issues?            | Partially addresses https://github.com/babel/babel/issues/8195
| Patch: Bug Fix?          |
| Documentation PR Link    | https://github.com/babel/website/pull/1718

See https://github.com/babel/babel/issues/8195


packages/babel-preset-env/src/normalize-options.js
```diff
@@ -165,6 +176,71 @@ export const validateUseBuiltInsOption = (
...
+  spec?: boolean,
+  loose?: boolean,
+}) => {
+  const specIsUndefined = typeof specOption === "undefined";
+  const looseIsUndefined = typeof looseOption === "undefined";
+  const transformModeIsUndefined = typeof transformMode === "undefined";
+
+  if (!specIsUndefined) {
+    console.log("");
+    console.log("The top-level `spec` option has been deprecated.");
```
Andarist: IMHO we shouldn't deprecate spec/loose - they map nicely to options accepted by plugins, if we introduce in the future more things that `transformMode` would account for it would still be useful to use spec/loose appropriately (maybe in combination with `transformMode`, maybe not - not sure if it would be possible to use them together at this point)

that's my opinion though, `transformMode` seems like something targeting newcomers, atm hiding spec/loose under more descriptive option, but for more experienced users loose/spec options are good because we can easily relate them to plugins' options

hzoo: Yeah I think it would be rather weird if we left the plugin options to use spec/loose still but the preset uses something different?

Andarist: Yeah, that's what I'm thinking. I see a value in `transformMode` though as some kind of options preset for the future use, although not sure how many more granular (than loose/spec) options we will truly have.

Also there is a question that would have to be IMHO answered here - what should `transformMode === 'performance'` do [here](https://github.com/babel/babel/blob/19a170529350216989384330988560b14eed6ae8/packages/babel-plugin-transform-for-of/src/index.js#L11)? How one can use preset-env with `assumeArray` nowadays? I suspect this might not be possible, but aint sure

rtsao: I think it might be helpful to frame the discussion in terms of the two overarching types of preset users:

**1. Newcomers, general usage**

I think the beauty of the env preset is it is a unifying abstraction that makes the Babel ecosystem extremely easy to use (by providing a declarative, high-level interface that corresponds closely to user intent). So just as inclusion/exclusion of individual plugins is abstracted over with the `targets` option, so should configuration of plugins. I think there are some issues with `spec`/`loose` in that regard:
- They only "loosely" correspond to user intent. `loose` is rather vague and doesn't imply the the actual desired behavior, which is performance.
- It's unclear what happens when `{loose: true, spec: true}` is set. Without knowledge of the underlying plugins (which is very possible given the nature of this preset), this is quite confusing.
- It's unclear what happens when `{loose: false, spec: false}` is set. Again, understanding what this does requires familiarity of conventions surrounding plugin implementation and configuration. Babel experts of course know that there's some default which strikes as balance between perf and spec compliance, but this is totally implicit (in terms of the config `{loose: false, spec: false}`). Instead, I think an explicit option (e.g. `{transformMode: "normal"}` or `{transformMode: "balanced"}` would be much more approachable and intuitive.

I think exposing `spec` and `loose` as the primary means of configuration leaks what should be considered implementation details at this level of abstraction. I think generally, people don't want have to care about configuration of individual plugins. Which brings me to the next use case:

**2. Power users**

I think it is dubious whether top-level `spec`/`loose` options are even sufficient for power users. I think in practice, usually there's one or more specific transformations in specific plugins that a power user wants to configure. For this, top-level `spec`/`loose` are far too broad, instead a more precise configuration mechanism is needed. 

This is why in https://github.com/babel/babel/issues/8195 I propose `options.configOverrides`, which allows for direct configuration of individual plugins. I think `assumeArray` is the perfect use case for this.

hzoo: Ok I generally agree with that, although you think that we shouldn't change the name of the options for individual plugins?

I never really thought the top level options would be sufficient for power users either. We tend to want to include multiple options as it increases in complexity, with the possibility one might want an option that includes them all under an umbrella "loose" mode. This was a suggestion but not sure how it plays out.

Although it's out of scope but related, I'm curious on your thoughts for https://github.com/babel/babel/issues/7770 (and my PR https://github.com/babel/babel/pull/8293). One of the reasons for removing them is the fact we have to pass the options down all the presets (not including updating them all the time, making breaking changes, and people opt-ing into too much or not knowing what's in it). If we remove the yearly/stage presets, preset-env will be the main way people interact with Babel. Should we have a preset for any proposals, (basically decided no now)?

rtsao: I was hesitant to also change the configuration for each plugin in this PR (simply because it would increase the scope of this PR), but I agree with you that it would be better to change the plugin config, if not simply to colocate/encapsulate the "transform mode" behavior/logic within each plugin rather than in the env preset. I think it would be probably best if each plugin decides on its own what the reasonable behavior is for "normal", "performance", and "spec" modes (used by the env preset), but have more granular options that can be turned on/off (when configuring directly). I can update this PR to update the plugin config as well.

Personally, I'm definitely on board with #7770 and agree that the env preset should probably supplant the use of yearly/stage presets. I think not including proposals (at least initially) probably makes sense to me 👍

nicolo-ribaudo:
Hey, thanks for the PR and sorry for leaving it around for more than two years :sweat_smile:

We have now decided to let users tune their output with a new `assumptions` option: https://github.com/babel/rfcs/pull/5. Our goal is to make Babel 100% spec compliant by default, but users can enable assumptions one by one to let Babel generate faster and smaller code.

We think that this better describes the different trade-offs of what "fast" and "small" mean, making it less likely that users accidentally rely on non-spec compliant behavior.

Thanks for the PR anyway, and for starting the discussion about how to overcome the `loose` and `spec` options, which have always been confusing!

PR #8299, opened 2018-07-10
Make env preset build-data scripts reproducible

rtsao:
| Fixed Issues?            | 
| Patch: Bug Fix?          | Yes, only internal change
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  | Removed `request` devDep, added devDep on `caniuse-db`, pinned `electron-to-chromium` version

The previous script had a few problems, which means builds are not reproducible:
1. It required internet access
2. It fetched remote data without any versioning

This PR addresses these issues by:
1. Using an explicit version of the `caniuse-db` which is published to npm.
2. Pinning `electron-to-chromium` to the version used the last time the script was run.

### DIFF
--- a/packages/babel-preset-env/package.json
+++ b/packages/babel-preset-env/package.json
@@ -60,8 +60,8 @@
     "@babel/core": "7.0.0-beta.52",
     "@babel/helper-fixtures": "7.0.0-beta.52",
     "@babel/helper-plugin-test-runner": "7.0.0-beta.52",
+    "caniuse-db": "1.0.30000851",
     "compat-table": "kangax/compat-table#90d02e486227d179d2ce9b850dbb3f9846443cab",
-    "electron-to-chromium": "^1.3.27",
-    "request": "^2.83.0"
+    "electron-to-chromium": "1.3.48"
   }
 }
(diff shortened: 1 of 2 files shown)


packages/babel-preset-env/package.json
```diff
@@ -60,8 +60,8 @@
     "@babel/core": "7.0.0-beta.52",
     "@babel/helper-fixtures": "7.0.0-beta.52",
     "@babel/helper-plugin-test-runner": "7.0.0-beta.52",
+    "caniuse-db": "1.0.30000851",
```
Andarist: could we maybe try some auto-updates for `caniuse-db` & `electron-to-chromium`? maybe just a greenkeeper integration (as fully automatic probably would have numerous flaws)?

rtsao: I've had good success with using https://renovatebot.com, which is very customizable and supports monorepos.

hzoo: Yeah we pinned the version of compat-table as well so this is good to me, unless we want to allow people to update before we do via peerDep or override if we are slow to update.

rtsao: Although we'd need to get renovate to update dependencies *and* run the update script. I'm not sure if that'd be possible or not.

PR #8316, opened 2018-07-13
Ensure that the wildcard interop is used with re-export + default.

### LINKED ISSUE #8306: Babel 7: Re-exporting both default and named symbols fails to export named symbols
## Bug Report

**Current Behavior**
A module is importing both the default and named (via wildcard) symbols from a non-ES module, and re-exporting them. We use this in various scenarios either to have internal packages that can synchronise versions of 3rd parties, or for platform selection with React Native.

Since `import { Foo } from "react"` works, I would expect `export * from "react"` to work the same way—and it does, *unless* the default is also imported. Once there’s a default import, Babel uses `_interopRequireDefault()` to import the package, which means that any properties on the object become unavailable on the imported object.

*Note: Although `babel-preset-react-native` forces all `@babel` packages to `7.0.0-beta.47`, I tried forcing `7.0.0-beta.53` (latest) with the same results.*

**Input Code**
- REPL or Repo link if applicable: https://github.com/haggholm/babel-re-export-failure

```js
import _default from 'react';
export default _default;
export * from 'react';
```

**Expected behavior/code**
The output should export the default `react` import *and* the named exports.

```js
{
  "presets": ["babel-preset-react-native"]
}
```

**Environment**
- Babel version(s): 7.0.0-beta.47, 7.0.0-beta.53
- Node/npm version: Node 8.11.1/npm 6.1.0
- OS: Ubuntu 18.04 LTS
- Monorepo: no

**Possible Solution**

Currently, the import interop helper looks like this:

```js
```
Maybe it could do something along these lines?
```js
function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : Object.assign({ default: obj }, obj); }
```
This is just a conceptual suggestion, I have not (yet) tested it.

See https://github.com/facebook/react-native/issues/20083

loganfsmyth:
| Fixed Issues?            | Fixes #8306
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

One missed edge case!

### DIFF
--- a/packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop/export-all/input.js
+++ b/packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop/export-all/input.js
@@ -1,0 +1,6 @@
+// The fact that this exports both a normal default, and all of the names via
+// re-export is an edge case that is important not to miss. See
+// https://github.com/babel/babel/issues/8306 as an example.
+import _default from 'react';
+export default _default;
+export * from 'react';
--- a/packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js
+++ b/packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js
@@ -246,12 +246,18 @@ function getModuleMetadata(
   });
 
   for (const metadata of sourceData.values()) {
-    if (metadata.importsNamespace.size > 0) {
-      metadata.interop = "namespace";
-      continue;
-    }
     let needsDefault = false;
     let needsNamed = false;
+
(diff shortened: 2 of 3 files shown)


packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop/export-all/input.js
```diff
@@ -0,0 +1,5 @@
+// The fact that it exports both a default, _and_ all of the names via re-export
```
Andarist: would be good to actually describe those edge cases (or link to some issue describing them)

PR #8317, opened 2018-07-13
Fix stage-0/1 import of pipeline proposals array

### LINKED ISSUE #8307: Cannot read property 'join' of undefined ( preset-stage-0 )
## Bug Report

**Current Behavior**
When i try to run webpack I am getting the following error

```
Module build failed (from ./node_modules/babel-loader/lib/index.js):
  TypeError: [BABEL] SomeFile.js: Cannot read property 'join' of undefined (While processing: "informed/node_modules/@babel/preset-stage-0/lib/index.js")
      at "informed/node_modules/@babel/helper-plugin-utils/lib/index.js:19:12
```

**Input Code**
- https://github.com/joepuzzo/informed

```js
// webpack config snippit
  rules: [
      {
        test: /\.(js)$/,
        exclude: /node_modules/,
        use: {
          loader: 'babel-loader',
          options: {
            presets: [
              "@babel/preset-env",
              ["@babel/preset-stage-0", { "decoratorsLegacy": true }],
              "@babel/preset-react"
            ],
          }
        }
      },
```

**Environment**
- Node/npm version: Node: 9.5.0 NPM: 5.8.0
- OS: OSX 10.13.5
- Monorepo NO
- How you are using Babel: Webpack 

My project uses storybook and after updating to the latest beta release i started to have issues with babel. I tracked it back to an issue where there latest version started to use `@babel/core` instead of the old V6. I have a seprate webpack config for my tests and for distribution/production and they are failing with the error above.

mAAdhaTTah:
| Fixed Issues?            | Fixes #8307

The proposals array is exported from the syntax plugin, so it must be imported from there as well.

### DIFF
--- a/packages/babel-preset-stage-0/src/index.js
+++ b/packages/babel-preset-stage-0/src/index.js
@@ -9,8 +9,8 @@ export default declare((api, opts = {}) => {
   const {
     loose = false,
     useBuiltIns = false,
-    decoratorsLegacy = false,
-    pipelineProposal,
+    decoratorsLegacy = true,
+    pipelineProposal = "minimal",
   } = opts;
 
   if (typeof loose !== "boolean") {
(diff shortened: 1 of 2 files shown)


nicolo-ribaudo:
I don't know if we are going to release this before removing the stage presets.

mAAdhaTTah:
Yeah, makes sense. I PR'd the fix because my initial thought was that they should work as expected before we deprecate them. In retrospect, it might have been better to deprecate & remove the stage presets before we introduced the `pipelineProposal` flag, but that's past now.

Up to you if you want to merge & deploy before we kill off the presets, but it's here if you want it.

packages/babel-preset-stage-0/src/index.js
```diff
@@ -2,7 +2,7 @@ import { declare } from "@babel/helper-plugin-utils";
 import presetStage1 from "@babel/preset-stage-1";
 
 import transformFunctionBind from "@babel/plugin-proposal-function-bind";
-import { proposals } from "@babel/plugin-proposal-pipeline-operator";
+import { proposals } from "@babel/plugin-syntax-pipeline-operator";
```
loganfsmyth: FYI, not that it matters since we'll land #8293 soon, but I noticed that this is also missing from https://github.com/valtech-nyc/babel/blob/0256b8739954f803959d45c298cf79163e4a68f5/packages/babel-preset-stage-0/package.json

PR #8318, opened 2018-07-13
Save full descriptor instead of only value for private fields.

nicolo-ribaudo:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Decorators can make private fields non-writable, so we need to store this information smewhere.
The descriptor can also be used to implement private accessors.

### DIFF
--- a/packages/babel-plugin-proposal-class-properties/src/index.js
+++ b/packages/babel-plugin-proposal-class-properties/src/index.js
@@ -201,7 +201,14 @@ export default declare((api, options) => {
 
     // Must be late evaluated in case it references another private field.
     return () =>
-      template.statement`MAP.set(REF, VALUE);`({
+      template.statement`
+        MAP.set(REF, {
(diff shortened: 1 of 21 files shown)


packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -201,7 +201,14 @@ export default declare((api, options) => {
 
     // Must be late evaluated in case it references another private field.
     return () =>
-      template.statement`MAP.set(REF, VALUE);`({
+      template.statement`
+        MAP.set(REF, {
+          // configurable is always false for private elements
+          // enumerable is always false for private elements
+          writable: true,
```
jridgewell: We can save bytes be doing an explicit check for `=== false`. Then, we don't have to always include the `writable: true`.

Issue #8140, opened 2018-06-10
Parse private class properties with Flow types


jamiebuilds:
Flow already supports this syntax:

```js
class Thing {
  #prop1: string;
  #prop2: string = value;
}
```

cc @jridgewell since you implemented private properties

wtgtybhertgeghgtwtg:
I've been getting by with
```
class Thing {
  #prop1;
  #prop2 = value;
}
```
and then having the types inferred from assignment, since `prettier` and `eslint` error out otherwise, but, yeah, it's sub-optimal.  Would this be a `babylon` issue, then?

jamiebuilds:
babylon has been moved from a separate repo into the babel monorepo.

C4Phone:
Hi, I'm new to this community so please correct me if I'm wrong. I tried to reproduce the issue but cannot see any actual problem when trying to use babel to transpile your typescript code.

I tried compiling your script using this babel start kit: clarkdave/babel-typescript-starter@3f16d3186953b181f9f0215ace5cc9a35cf9dd89. And babel was able to:

1. Compile  ``` class Thing {   abc: string;   def: string="ggg"; } ```  To  ``` let Thing = class Thing {   def = "ggg"; }; //# sourceMappingURL=test.js.map ```
2. Find a type error if I try to compile:   ``` class Thing {   abc: string;   def: string=123; } ```  Below is babel's output:  ``` lib/test.ts(3,3): error TS2322: Type '123' is not assignable to type 'string'.  3   def: string=123;     ~~~~~~~~~~~~~~~~ error Command failed with exit code 1. ```

So it appears to me that babel is now able to parse the type information you wanted. Could you be more specific on what your expected output from babel is and how the actual output differs from your expectation? 

Thank you!

C4Phone:
I did find the following issue when using the flow version of babel starter kit from omimo/es6-flow-starter@edc8cae70a63ea2405d60316ef34bf5e3eeb43d4.

Babel compiled the following code without finding out about the type error of assigning 123 to a string.

```
class Thing {
  abc: string;
  def: string = 123;
}

console.log('hello');
```

Here is the resulting code:

```
'use strict';

function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }

var Thing = function Thing() {
  _classCallCheck(this, Thing);

  this.def = 123;
};

console.log('hello');
```

However, I am not sure what the expected output is.

[...]

PR #8340, opened 2018-07-18
Fix private property parsing in Flow

kalenikalexander:
| Fixed Issues?            | [Parse private class properties with Flow types](https://github.com/babel/babel/issues/8140) 
| Patch: Bug Fix?          |  Yes


packages/babel-parser/test/fixtures/flow/type-annotations/140/input.js
```diff
@@ -0,0 +1,4 @@
+class A {
```
existentialism: I know we do a really bad job of it, but can we give this folder a name? Maybe `class-private-property` or something?

existentialism:
We might also want to make a note to bump https://github.com/babel/babel/blob/master/Makefile#L2 to make sure we're syncing with flow's parsing of private props.

kalenikalexander:
I add `private_class_properties/multiple.js` and `private_class_properties/super.js` to whitelist because parser doesn't do semantic analysis of private properties.

radex:
Hey, what's the status here?  This seems to be the last bit necessary to make private properties work nicely when using Flow.

packages/babel-parser/src/types.js
```diff
@@ -716,6 +716,7 @@ export type ClassPrivateProperty = NodeBase & {
   value: ?Expression, // TODO: Not in spec that this is nullable.
   static: boolean,
   computed: false,
+  typeAnnotation?: ?TypeAnnotationBase, // TODO: Not in spec
```
nicolo-ribaudo: Shouldn't this type be ` TypeAnnotation`?

existentialism:
Parsing looks good, should we also add tests to `generator` and `flow-strip-types` to make sure all is well?

kalenikalexander:
I should put `flow-strip-types` tests in `fixtures/regression` or `fixtures/strip-types` directory?

nicolo-ribaudo:
Thank you!

PR #8342, opened 2018-07-19
Do not mutate ast

### LINKED ISSUE #7987: 'transformFromAst' should avoid mutating the AST passed into it
## Bug Report

**Current Behavior**
The description in the documentation for `transformFromAst` in `@babel/core` does not make it obvious that the input AST object is transformed *in-place*. A cursory reading would imply that the input is readonly and only under closer inspection hints that the input might be changed.

**Possible Solution**
Add a simple blurb that says 

> The AST object passed in is transformed in-place. To preserve the original object, make a deep copy before passing it in.

thiagoarrais:
| Fixed Issues?            | Fixes #7987  
| Major: Breaking Change?  | :-1:
| Minor: New Feature?      | :-1:
| Documentation PR Link    | 
| Any Dependency Changes?  | :-1:

This clones the given ast prior to working on it to avoid mutating the AST given to `transformFromAST`.

This is work in progress and intended as input for review. Some questions I have:

- Is this the intended use for `t.cloneNode`?
- `parse.js` was the easiest place to insert the test. I'm not sure where it should go

Please check inline review comments for context.

### DIFF
--- a/packages/babel-core/src/transformation/normalize-file.js
+++ b/packages/babel-core/src/transformation/normalize-file.js
@@ -2,6 +2,7 @@
 
 import path from "path";
 import buildDebug from "debug";
+import cloneDeep from "lodash/cloneDeep";
 import * as t from "@babel/types";
 import type { PluginPasses } from "../config";
 import convertSourceMap, { typeof Converter } from "convert-source-map";
@@ -75,6 +76,7 @@ export default function normalizeFile(
     } else if (ast.type !== "File") {
       throw new Error("AST root must be a Program or File node");
     }
+    ast = cloneDeep(ast);
   } else {
     // The parser's AST types aren't fully compatible with the types generated
     // by the logic in babel-types.
--- a/packages/babel-core/test/api.js
+++ b/packages/babel-core/test/api.js
@@ -12,6 +12,13 @@ function assertNotIgnored(result) {
   expect(result).not.toBeNull();
 }
 
+function parse(code, opts) {
+  return babel.parse(code, {
+    cwd: __dirname,
+    ...opts,
+  });
+}
+
 function transform(code, opts) {
   return babel.transform(code, {
     cwd: __dirname,
@@ -38,6 +45,13 @@ function transformFileSync(filename, opts) {
 
 function transformAsync(code, opts) {
   return babel.transformAsync(code, {
+    cwd: __dirname,
+    ...opts,
+  });
+}
+
+function transformFromAst(ast, code, opts) {
+  return babel.transformFromAst(ast, code, {
     cwd: __dirname,
     ...opts,
   });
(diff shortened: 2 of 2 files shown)


packages/babel-core/src/transformation/normalize-file.js
```diff
@@ -69,12 +69,14 @@ export default function normalizeFile(
...
   }
 
+  debugger; // eslint-disable-line no-debugger
   if (ast) {
     if (ast.type === "Program") {
       ast = t.file(ast, [], []);
     } else if (ast.type !== "File") {
       throw new Error("AST root must be a Program or File node");
     }
+    ast = t.cloneNode(ast);
```
thiagoarrais: Is this the intended use for t.cloneNode?

packages/babel-core/test/parse.js
```diff
@@ -28,4 +28,32 @@ describe("parse", function() {
...
+    const node = parse(program, {
+      filename: fixture("input.js"),
+      cwd: fixture(),
+    });
+    expect(node.program.body[0].declarations[0].id.name).toBe("identifier");
+    expect(node).not.toBeNull();
+    transformFromAst(node, program, {
+      filename: fixture("input.js"),
+      cwd: fixture(),
+      plugins: [
```
thiagoarrais: this inline plugin forces the AST to change. any ideas on how to write it more succinctly?

packages/babel-core/test/parse.js
```diff
@@ -28,4 +28,32 @@ describe("parse", function() {
     });
     expect(JSON.parse(JSON.stringify(result))).toEqual(output);
   });
+
```
thiagoarrais: parse.js was the easiest place to insert this test. I'm not sure where it should go. My first instinct was to add it to `test/transformation.js`, but there weren't any describe'd tests there...

packages/babel-core/test/parse.js
nicolo-ribaudo: You can replace `function() { return` with `() => (`, but it is not really important

packages/babel-core/src/transformation/normalize-file.js
nicolo-ribaudo: The problem is that `cloneNode` only clones properties which represents other nodes. It doesn't copy every property (e.g. position, extras...).

I would use lodash's cloneDeep function.

thiagoarrais: Good idea. Will take a look at that.

packages/babel-core/test/parse.js
thiagoarrais: Yup. I was wondering if you folks already have some established helper or something for this kind of stuff.

thiagoarrais:
I think this is good to go now. I'm removing the WIP marker.

thiagoarrais:
Closing and reopening to try and trigger a Travis rebuild.

packages/babel-core/test/api.js
```diff
@@ -168,6 +182,30 @@ describe("api", function() {
...
+                path.node.name = "replaced";
+              },
+            },
+          };
+        },
+      ],
+    });
+
+    expect(code).toBe("const replaced = 1;");
+    expect(node.program.body[0].declarations[0].id.name).toBe(
```
hzoo: Since there was a discussion on deep cloning the test (or a new test) should probably reflect that, like check the the other metadata fields aren't changed?

thiagoarrais: That would be nice, yes. Any ideas on how to force a metadata change? I'm outputing the AST for this transform and it doesn't seem to change any metadata fields...

thiagoarrais: I can add something like that if you folks think it would be useful. I am not really in a good position to evaluate that, though. Do people usually sprinkle such metadata fields into their ASTs?

hzoo: Actually this works fine! thanks!

PR #8345, opened 2018-07-19
throwOnUninitializedRead option for babel-plugin-transform-modules-common-js

adamjernst:
If you use ES6 modules with `require` and trigger a circular dependency, what should happen? Currently you receive `undefined`; this adds an off-by-default option to throw instead.

| Fixed Issues?            | 
| Documentation PR Link    | 

### DIFF
--- a/packages/babel-helper-module-transforms/src/index.js
+++ b/packages/babel-helper-module-transforms/src/index.js
@@ -30,6 +30,7 @@ export function rewriteModuleStatementsAndPrepareHeader(
     loose,
     noInterop,
     lazy,
+    throwOnUninitializedRead,
     esNamespaceOnly,
   },
 ) {
@@ -74,7 +75,14 @@ export function rewriteModuleStatementsAndPrepareHeader(
   }
 
   // Create all of the statically known named exports.
-  headers.push(...buildExportInitializationStatements(path, meta, loose));
+  headers.push(
+    ...buildExportInitializationStatements(
+      path,
+      meta,
+      loose,
+      throwOnUninitializedRead,
+    ),
+  );
 
   return { meta, headers };
 }
@@ -311,6 +319,7 @@ function buildExportInitializationStatements(
   programPath: NodePath,
   metadata: ModuleMetadata,
   loose: boolean = false,
+  throwOnUninitializedRead: boolean = false,
 ) {
   const initStatements = [];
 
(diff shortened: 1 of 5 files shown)


adamjernst:
@loganfsmyth Got a second to review this little change? It leaves the default behavior unchanged, only introducing new behavior behind an off-by-default switch. It would be a lifesaver for my project, where some developers are severely confused by the combination of export and require cycles.

Seems the failed tests are just flaky responses from the registry:

> Received malformed response from registry for undefined. The registry may be down

loganfsmyth:
My general thoughts on this are that it's something I'm in favor of adding, but I don't think the approach in this PR is the one I'd like. If we're already making use of getters, it'd likely be easier to do:
```
"use strict";
 
  value: true
});
exports.bar = bar;
Object.defineProperty(exports, "x", {
  get: function () {
    return x;
  },
  set: function () {
    throw new Error('Cannot reassign exported value.');
  },
  enumerable: true,
});
 
const foo = require('foo');
 
const x = "Hello world";
 
function bar() {
  return 42;
}
```
and then we avoid needing the `delete` usage entirely.

adamjernst:
@loganfsmyth Indeed, that works great if your VM has implemented the temporal dead zone… which mine has not. I didn't want to add the overhead of another separate variable to track set-vs-unset.

I am not enough of a Babel expert to know if there's e.g. a transform that emulates temporal dead zone support and runs after this transform. Sorry about that.

Anyway, let me know what I should do. I suppose I could implement this in an entirely separate babel transform that is private to my use case if you wish.

adamjernst:
@loganfsmyth The most obvious option I see for me here is to introduce two new locals for each exported value — one to hold the value, one to indicate if it is set or unset. Implement the setter and getter to operate on these values. That neatly removes the need for `delete`, which I agree is unsightly.

The downside is that this is a fair bit less efficient, but I plan on using this only for development purposes anyway.

loganfsmyth:
The TDZ point is a tough one. We do have an experimental implementation of TDZ in the block-scoping transform, it's just off by default. Even if it's on though, I think it would have trouble here because it would run before the actual module transform does, with out current system.

My preference would probably be to split this into a few parts:
* Make the exports getters like my example above.
* See if we can adjust the ordering of the module transform so that it runs before the block scoping statements are converted so block scoping transform is can add TDZ checks inside the actual getter fn for us.
* Fix bugs in the TDZ implementation.

adamjernst:
@loganfsmyth OK, good advice! This implements step 1 of your suggested split. This also means there are no changes to rewrite-live-references for now which is pretty nice.

(I implemented behavior to emulate TDZ, along with a lengthy comment explaining how it goes away once step 2 & 3 are complete.)

adamjernst:
@loganfsmyth  as an open source maintainer myself I know how annoying pings are… but mind giving this a thumbs-up or thumbs-down? 🙏

PR #8348, opened 2018-07-20
Fix es2015-module-commonjs implementation of loose option as README

### LINKED ISSUE #8347: 6.x transform-es2015-modules-commonjs loose option doesn't work as expected
## Bug Report

**Current Behavior**
babel-plugin-transform-es2015-modules-commonjs
`loose` option doesn't work as [README](https://www.npmjs.com/package/babel-plugin-transform-es2015-modules-commonjs).

With the `loose` option on, babel 6 is still producing code using Object.defineProperty. And would cause re-export to fail due to the use of `Object.defineProperty`

Also related to #6795

**Expected behavior/code**
As the README

**Environment**

**Possible Solution**
Make the `loose` option work as it suggested in [README](https://www.npmjs.com/package/babel-plugin-transform-es2015-modules-commonjs)

Add `loose` option handling for `buildExportAll` at
https://github.com/babel/babel/blob/6.x/packages/babel-plugin-transform-es2015-modules-commonjs/src/index.js

yshing:
| Fixed Issues?            |  Fixes #8347
| Patch: Bug Fix?          | YES
| Major: Breaking Change?  | NO
| Minor: New Feature?      | NO
| Tests Added + Pass?      | Not sure
| Documentation PR Link    | 
| Any Dependency Changes?  | NO

### DIFF
--- a/packages/babel-plugin-transform-es2015-modules-commonjs/src/index.js
+++ b/packages/babel-plugin-transform-es2015-modules-commonjs/src/index.js
@@ -23,6 +23,10 @@ const buildExportsFrom = template(`
   });
 `);
 
+const buildLooseExportsFrom = template(`
+  Object.defineProperty(exports, $0, {})
+`);
+
 const buildLooseExportsModuleDeclaration = template(`
   exports.__esModule = true;
 `);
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
Tests?

nicolo-ribaudo:
Does it work as expected in Babel 7? Since this issue isn't critial, this PR is highly unlikely to get merged to the 6.x branch.

xtuc:
Since we release Babel 7 is this still needed?

loganfsmyth:
I think at this point this would be a dangerous change to make in `6.x`. This essentially disables live bindings, and I could easily see users having `loose: true` in their config while also relying on live binding and not caring about `Object.defineProperty`. We want to explore more fine-grained options eventually, but we haven't gotten that done yet.

PR #8376, opened 2018-07-24
Fix order of optional argument reordering

Qix-:
| Patch: Bug Fix?          | Yes. |
| Major: Breaking Change?  | No. |
| Minor: New Feature?      | No. |
| Tests Added + Pass?      | We'll see (minor enough to warrant waiting for CI) |
| Documentation PR Link    | N/A |
| Any Dependency Changes?  | No. ||
| License                  | MIT |

Caught this looking into an unrelated issue with AST transformations. Previously, if the optional `opts` parameter wasn't passed, the _intent_ was to move the function it held into the `callback` parameter and null out the `opts` param - but instead, it was nulling both.

### DIFF
--- a/packages/babel-core/src/transform-ast.js
+++ b/packages/babel-core/src/transform-ast.js
@@ -31,8 +31,8 @@ export const transformFromAst: TransformFromAst = (function transformFromAst(
   callback,
 ) {
   if (typeof opts === "function") {
-    opts = undefined;
     callback = opts;
+    opts = undefined;
   }
 
   // For backward-compat with Babel 6, we allow sync transformation when


nicolo-ribaudo:
:+1: for the additional validation. Can you also add a test which shows how the old behavior was wrong?

loganfsmyth:
Woops, great catch. As @nicolo-ribaudo mentioned, we've got the same issue in https://github.com/babel/babel/blob/a811cf78e19442bbc10b05be2076c330fb702087/packages/babel-core/src/transform-file.js#L24 and https://github.com/babel/babel/blob/a811cf78e19442bbc10b05be2076c330fb702087/packages/babel-core/src/transform-ast.js#L34, so if you could fix those it'd be great.

xtuc:
Good catch!

After a quick search I found another one:
https://github.com/babel/babel/blob/98ff2ce87747df3a62fff97a508af5451ba12eae/packages/babel-core/src/parse.js#L27

hzoo:
Added those ^, so we never added tests for async versions of those methods in https://github.com/babel/babel/pull/6780?

hzoo:
Can merge now as well to fix the issue, but yeah we should figure out the tests too

PR #8395, opened 2018-07-27
optionally prevent double compile of amd define

kellyselden:
This is an unfortunate pull request, but here it goes.

In the Ember.js world, we are dealing with legacy compatiblility issues that are preventing us from moving forward. Our current issue is that we can be in a state where there can be some files that are ES6 and some that are AMD. In the past, it was either all ES6 where we would run this plugin, or all AMD where we would skip this plugin.

We have attempted some interesting hacks to detect AMD files and filter them out before compiling (https://github.com/ember-cli/ember-cli/pull/7935), but this solution feels wrong and can be quite slow.

My alternate solution is to prevent the double AMD right before compilation in this plugin. Since the file is already read at this point, we no longer have to read it twice.

I think we can make it so that this option dies in babel 6, it doesn't need to be ported into babel 7. Correct me if I'm wrong @rwjblue.

cc @rwjblue @twokul


loganfsmyth:
Tweaked the validation logic a bit, but otherwise looks fine to me.

kellyselden:
@loganfsmyth Thank you. Do you mind releasing this, or are you waiting to batch other things?

loganfsmyth:
@kellyselden There's at least one other thing waiting to be backported, but some time in the next week or so it will probably be available. Is that acceptable?

kellyselden:
I was hoping sooner that a week+ if possible.

kellyselden:
Would it be possible to release this? I would love to start using it.

PR #8397, opened 2018-07-27
Remove our own use of stage presets

hzoo:


package.json
```diff
@@ -12,10 +12,13 @@
   "devDependencies": {
     "@babel/cli": "7.0.0-beta.52",
     "@babel/core": "7.0.0-beta.52",
+    "@babel/plugin-proposal-class-properties": "7.0.0-beta.52",
+    "@babel/plugin-proposal-export-namespace-from": "7.0.0-beta.52",
+    "@babel/plugin-proposal-numeric-separator": "7.0.0-beta.52",
+    "@babel/plugin-proposal-object-rest-spread": "7.0.0-beta.52",
```
hzoo: Actually we can probably remove since it's part of preset-env?

existentialism: Yep

PR #8398, opened 2018-07-29
Only reference helpers from external/runtime helpers if they are known to be available.

loganfsmyth:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Documentation PR Link    | 
| Any Dependency Changes?  |

## Why

This PR aims to resolve a specific problem. Babel's helpers live as part of `@babel/core` conceptually, or indirectly from `@babel/helpers`, but when users use `@babel/plugin-transform-runtime` or `@babel/plugin-external-helpers`, the helpers actually load from a static file like `@babe/runtime/helpers/classCallCheck`. 

While very handy, this has the downside that we don't actually know what version of a helper we're going to get, and even more frustratingly, we don't actually know if the helper even exists in the runtime, for instance if we've recently added new helpers. In 6.x, this meant that we could add a new transform for instance, along with a helper for it, but users using `@babel/runtime` would get the extremely unhelpful error that their import didn't find the file. 

It also meant that there was literally no way for a plugin to check if a helper was available. If `transform-runtime` or similar plugin was enabled, _any_ attempt to load a helper would insert a reference, even if it was totally broken, or you'd just typoed the name of a helper or something.

This PR aims to resolve that problem by changing the expectations around intercepting runtime helper loading. If an override function is given, it is that function's responsibility to only insert a reference to a helper if it knows the helper exists. That means that plugins themselves can more easily evolve over time. For instance, a plugin can attempt to add a helper, and if it doesn't exist, it could fall back to inserting a different helper instead, or it could provide useful feedback to users saying that they'll need to update their version of `@babel/core` in order to use this new helper.

## How

This PR accomplishes it's goal by having `@babel/core` itself be a source of truth for what helpers exist, and what helpers became available when. When a user does

```
plugins: [
  ['@babel/transform-runtime']
]
```
they are declaring "load all helpers from the runtime, BUT only if they were available at the release of whichever version this PR lands in.

What this means is that, in the future if we make a new plugin with some new helper, it will _not_ be loaded via the runtime at all. It would be up to the user to explicitly do

```
plugins: [
  ['@babel/transform-runtime', { version: "7.4.2"}] // or a range
]
```
to essentially say "I promise that I have version 7.4.2 of `@babel/runtime` available, so any helpers that existed in `7.4.2` are free to be loaded from this runtime.

## Downsides

The main downside of this approach is that it's more work for users to have to manually provide the version. In a perfect world, I'd love for `@babel/transform-runtime` to actually try to look at the user's `package.json` or something, and I thought about doing that for this PR, but I ended up deciding that it complicated things too much for now. We can always explore adding it as an opt-in feature during 7.x's lifetime.


packages/babel-helpers/src/helpers.js
```diff
@@ -1,9 +1,11 @@
 import template from "@babel/template";
 
-const helpers = {};
+const helpers = Object.create(null);
 export default helpers;
 
 helpers.typeof = () => template.program.ast`
+  "minVersion: 7.0.0-beta.0";
```
Andarist: imho it's slightly weird to use directive-like for this, making it a property of the helper (or wrapping helpers in objects with this and smth like `create` method) would make it less magical

packages/babel-core/src/transformation/file/file.js
```diff
@@ -64,6 +65,14 @@ export default class File {
   }
 
   set(key: mixed, val: mixed) {
+    if (key === "helpersNamespace") {
+      throw new Error(
+        "Babel 7.0.0-beta.56 has dropped support for the 'helpersNamespace' utility." +
+          "If you are using @babel/plugin-external-helpers you will need to use a newer " +
+          "version than the one you currently have installed.",
```
nicolo-ribaudo: This error message should suggest plugin authors to use `helperGenerator` instead.

packages/babel-core/src/transformation/file/file.js
```diff
@@ -136,17 +145,38 @@ export default class File {
...
+    } catch (err) {
+      if (err.code !== "BABEL_HELPER_UNKNOWN") throw err;
+
+      return false;
+    }
+
+    return (
+      typeof versionRange !== "string" ||
+      (!semver.intersects(`<${minVersion}`, versionRange) &&
+        !semver.intersects(`>=8.0.0`, versionRange))
```
nicolo-ribaudo: Why is this second check needed? Some helpers almost never change (e.g. the `typeof` one), so I would expect it to be possible that some helpers are compatible even between major versions.

packages/babel-core/src/transformation/file/file.js
```diff
@@ -136,17 +145,38 @@ export default class File {
...
+    try {
+      minVersion = helpers.minVersion(name);
+    } catch (err) {
+      if (err.code !== "BABEL_HELPER_UNKNOWN") throw err;
+
+      return false;
+    }
+
+    return (
+      typeof versionRange !== "string" ||
```
nicolo-ribaudo: Maybe this should also work with integers, like the `api.assertVersion` function does?

packages/babel-helpers/src/helpers.js
xtuc: I agree, what about `helpers.typeof.minVersion = 7.0.0-beta.0`?

packages/babel-helpers/src/index.js
```diff
@@ -247,6 +252,15 @@ function loadHelper(name) {
           globals: metadata.globals,
         };
       },
+      minVersion() {
+        return fn()
```
xtuc: And following my previous comment. Here: `fn.minVersion`.

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -14,6 +14,7 @@ export default declare((api, options) => {
     regenerator,
     useBuiltIns,
     useESModules,
+    version: runtimeVersion = "7.0.0-beta.0",
```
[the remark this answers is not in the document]
loganfsmyth: In `availableHelper`, if you omit the version, it just checks if the helper exists in the current version of core, rather than if it was historically available at a particular version. If we left this blank, we'd essentially always insert the helper imports, even when it was invalid.

packages/babel-helpers/src/helpers.js
loganfsmyth: I was being lazy :P Since none of this is actually part of the public API, I figured refactoring it later would be trivial. I'll see if I can make it better now though.

loganfsmyth:
@nicolo-ribaudo 

> Will this PR allow us to introduce breaking change to the helpers, by incrementing the minVersion?

No, that's not my intention with this PR, but what it will allow us to do is add a _new_ helper with a new API, and use that new helper as a replacement for the old one in a safe way.

DreRandaci:
@loganfsmyth
 
I am implementing `babel-plugin-external-helpers` into a Rollup build process and am getting the `Babel 7.0.0-beta.56 has dropped support for the 'helpersNamespace' utility` error when I do a build. I'm using the `rollup-plugin-babel` library and am new to implementing Rollup in general, so I don't doubt I'm doing something wrong. Do I need to specify an `external-helpers` version in my `.babelrc` plugins? I am following the [Rollup docs](https://rollupjs.org/guide/en#babel) for this.

### .babelrc file
```json
{
  "presets": [
    [
      "env",
      {
        "modules": false
      }
    ]
  ],
  "plugins": [
    "external-helpers"
  ]
}
```  

### package.json
```json
"devDependencies": {
    "@babel/core": "^7.0.0",
    "babel-plugin-external-helpers": "^6.22.0",
    "babel-preset-env": "^1.7.0",
    "eslint": "^5.4.0",
    "eslint-config-prettier": "^3.0.1",
    "eslint-config-standard": "^12.0.0",
    "eslint-plugin-import": "^2.14.0",
    "eslint-plugin-node": "^7.0.1",
    "eslint-plugin-prettier": "^2.6.2",
    "eslint-plugin-promise": "^4.0.0",
    "eslint-plugin-standard": "^4.0.0",
    "prettier": "1.14.2",
    "rollup": "^0.65.0",
    "rollup-plugin-babel": "^4.0.2",
    "rollup-plugin-size-snapshot": "^0.6.1"
  }
```

nicolo-ribaudo:
You can use `@babel/transform-runtime` and `@babel/runtime`: the former will inject `import _classCallCheck from "@babel/runtime/helpers/classCallCheck` for every needed helper, and then Rollup will bundle them with your code.

DreRandaci:
@nico29 I changed my `.babelrc` config to use the `@babel/preset-env` preset and got rid of the "external-helpers" plugin, which seems to have resolved the issue (since I'm using babel 7). Does your solution target babel < 6? Or is it for deduping the inserted helpers?

#### Updated .babelrc
```
{
  "presets": [
    [
      "@babel/preset-env",
      {
        "modules": false
      }
    ]
  ],
  "plugins": []
}
```

nicolo-ribaudo:
Babel by default injects the helpers in every file. `@babel/plugin-transform-runtime` avoids it by allowing the helpers to be imported, so that they can be deduped. `babel-plugin-external-helpers` had the same goal, but it relied on a global variable.

Andarist:
Just to add what kinda got already said `@babel/plugin-external-helpers` is no longer needed to dedupe helpers when using babel@7 and corresponding rollup-plugin-babel@4.

Need to update rollup docs (had no idea they have section about this) to explain difference between configuring for babel@6 and babel@7.

PR #8405, opened 2018-07-31
Optim constructor cleanup redundant 'assertThisInitialized`, ensure 'this' aliases (#8335)

### LINKED ISSUE #8335: Redundant call to `assertThisInitialized` for classes in loose mode
## Bug Report

**Current Behavior**

There are duplicated calls to the `assertThisInitialized` helper. 

```js
let Foo =
/*#__PURE__*/
function (_Bar) {
  _inheritsLoose(Foo, _Bar);

  function Foo() {
    var _this;

    _this = _Bar.call(this) || this;
    something(_assertThisInitialized(_assertThisInitialized(_this)));
    return _this;
  }

  return Foo;
}(Bar);
```

**Input Code**

```js
class Foo extends Bar {
  constructor() {
    super();
    something(this);
  }
}
```

**Expected behavior/code**

```js
let Foo =
/*#__PURE__*/
function (_Bar) {
  _inheritsLoose(Foo, _Bar);

  function Foo() {
    var _this;

    _this = _Bar.call(this) || this;
    something(_assertThisInitialized(_this));
    return _this;
  }

  return Foo;
}(Bar);
```

**Environment**
- Babel version(s): v7.0.0-beta.48
- Node/npm version: Node 8
- OS: OS X
- Monorepo: no
- How you are using Babel: core

allex:
| Fixed Issues?            | Fixes #8335
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/declaration-order/output.js
+++ b/packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/declaration-order/output.js
@@ -1,9 +1,11 @@
 var C = function C() {
   "use strict";
 
+  var _this = this;
+
   babelHelpers.classCallCheck(this, C);
-  this.y = babelHelpers.classPrivateFieldLooseBase(this, _x)[_x];
-  Object.defineProperty(this, _x, {
+  _this.y = babelHelpers.classPrivateFieldLooseBase(_this, _x)[_x];
+  Object.defineProperty(_this, _x, {
     writable: true,
     value: void 0
   });
--- a/packages/babel-plugin-transform-classes/src/transformClass.js
+++ b/packages/babel-plugin-transform-classes/src/transformClass.js
@@ -19,6 +19,81 @@ function buildConstructor(classRef, constructorBody, node) {
   return func;
 }
 
+function findNearestBlock(p) {
+  let last = p;
+  return p.find(p => {
+    if (p.isBlockParent() || p.isSequenceExpression()) {
+      // exclude CallExpression as an object key name.
+      if (last && p.isMethod({ key: last.node })) {
+        return false;
+      }
+      return true;
+    }
+    last = p;
+  });
+}
+
+const checkSuperCalleeVisitor = traverse.visitors.merge([
+  environmentVisitor,
+  {
+    Super(path, state) {
+      const { node, parentPath } = path;
+      if (
+        parentPath.isCallExpression({ callee: node }) &&
+        !isInConditional(parentPath, state.root)
+      ) {
+        state.path = path;
+        path.stop();
+      }
+    },
+  },
+]);
+
+function findPrevSiblingSuper(p) {
+  let ret;
+  return (
+    p.getAllPrevSiblings().some(p => {
+      const state = { path: null, root: p };
+      return p.traverse(checkSuperCalleeVisitor, state), (ret = state.path);
+    }) && ret
+  );
+}
+
+function isInConditional(p, ref) {
+  do {
+    if (p.isConditional() || p.isLogicalExpression() || p.isSwitchStatement()) {
+      return true;
+    } else if (p === ref) return false;
+  } while ((p = p.parentPath));
+  return false;
+}
+
+function findUp(p, fn) {
+  do {
+    if (!p || p.isClassBody()) return false;
+    else if (fn(p)) return p;
+  } while ((p = p.parentPath));
+  return false;
+}
+
+// Check block assertion of `assertThisInitialized`
+const isThisAsserted = p =>
+  findUp(p, p => !!p.getData("_assertThisInitialized"));
+
+// Find `super()` in forwards and upwards
+const findUpwardsSuper = p => {
+  let ret;
+  return (
+    findUp(p, p => !!(ret = findPrevSiblingSuper(p.getStatementParent()))), ret
+  );
+};
+
+// Add path with assertThisInitialized flag
+const setThisAssert = p => {
+  const b = findNearestBlock(p);
+  b && b.setData("_assertThisInitialized", 1);
+};
+
 const verifyConstructorVisitor = traverse.visitors.merge([
   environmentVisitor,
   {
@@ -32,23 +107,16 @@ const verifyConstructorVisitor = traverse.visitors.merge([
         );
       }
     },
-
     ThisExpression(path, state) {
-      if (!state.isDerived) return;
+      const { isDerived, superThises, ctorThises } = state;
 
-      const { node, parentPath } = path;
-      if (parentPath.isMemberExpression({ object: node })) {
-        // In cases like this.foo or this[foo], there is no need to add
-        // assertThisInitialized, since they already throw if this is
-        // undefined.
-        return;
+      ctorThises.push(path);
+
+      // keep cache superThises for wrapSuperCall()
+      if (isDerived) {
+        superThises.push(path);
       }
 
-      const assertion = t.callExpression(
-        state.file.addHelper("assertThisInitialized"),
-        [node],
-      );
-      path.replaceWith(assertion);
       path.skip();
     },
   },
@@ -84,7 +152,8 @@ export default function transformClass(
     instancePropRefs: {},
     staticPropBody: [],
     body: [],
-    bareSupers: new Set(),
+    bareSupers: [],
+    ctorThises: [],
     superThises: [],
     pushedConstructor: false,
     pushedInherits: false,
@@ -100,15 +169,6 @@ export default function transformClass(
   const setState = newState => {
     Object.assign(classState, newState);
   };
-
-  const findThisesVisitor = traverse.visitors.merge([
-    environmentVisitor,
-    {
-      ThisExpression(path) {
-        classState.superThises.push(path);
-      },
-    },
-  ]);
 
   function pushToMap(node, enumerable, kind = "value", scope?) {
     let mutatorMap;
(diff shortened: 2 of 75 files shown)


packages/babel-plugin-transform-classes/src/transformClass.js
```diff
@@ -371,21 +338,66 @@ export default function transformClass(
...
-    let guaranteedSuperBeforeFinish = !!classState.bareSupers.size;
-
+    let thisAlias;
     let thisRef = function() {
-      const ref = path.scope.generateDeclaredUidIdentifier("this");
+      const ref = path.scope.generateUidIdentifier("this");
       thisRef = () => t.cloneNode(ref);
+      path.scope.push({
+        id: ref,
+        init: (thisAlias = !isDerived ? t.thisExpression() : null),
```
xtuc: `assertThisInitialized` return `this` again. What about calling it here?

Example:
```js
var _this = _assertThisInitialized(this);
```

So you don't need to keep track of `assertThisInited`

xtuc:
Looks good so far, thanks for your PR!

packages/babel-plugin-transform-classes/src/transformClass.js
```diff
@@ -19,40 +19,23 @@ function buildConstructor(classRef, constructorBody, node) {
...
-        return;
-      }
-
-      const assertion = t.callExpression(
-        state.file.addHelper("assertThisInitialized"),
-        [node],
-      );
-      path.replaceWith(assertion);
-      path.skip();
+const verifyConstructorVisitor = (path, visitors) =>
```
nicolo-ribaudo: What is the `path` argument needed for?

packages/babel-plugin-transform-classes/src/transformClass.js
```diff
@@ -371,21 +338,66 @@ export default function transformClass(
...
+      });
       return ref;
     };
 
+    // prevent `assertThisInitialized(this)` duplicates
+    let assertThisInited = false;
+
+    // Fix issue #8335, traverse all of the `this` expressions with assertion and aliases
+    path.traverse(
+      verifyConstructorVisitor(path, {
```
nicolo-ribaudo: This recreates the visitor every time a class is transformed. Can you hoist it up so that it is created only once?

packages/babel-plugin-proposal-class-properties/test/fixtures/nested-class/super-call-in-key/output.js
```diff
@@ -23,8 +23,10 @@ function (_Hello) {
     var _this2 = _this = babelHelpers.possibleConstructorReturn(this, babelHelpers.getPrototypeOf(Outer).call(this));
 
     let Inner = function Inner() {
+      var _this3 = this;
```
nicolo-ribaudo: Can we somehow avoid always assigning `this`?

packages/babel-plugin-transform-classes/src/transformClass.js
allex: y, no used any more, i've put the visitor as a const.

packages/babel-plugin-transform-classes/src/transformClass.js
allex: y, already hoist it up in the latest commit.

packages/babel-plugin-proposal-class-properties/test/fixtures/nested-class/super-call-in-key/output.js
allex: thanks for your suggestion, theses assigning has cleanup if REFs less than one.

packages/babel-plugin-transform-classes/src/transformClass.js
```diff
@@ -371,22 +362,94 @@ export default function transformClass(
...
-      const ref = path.scope.generateDeclaredUidIdentifier("this");
-      thisRef = () => t.cloneNode(ref);
-      return ref;
+    const thisRef = () => {
+      let ref;
+      if ((ref = thisRef.ref)) {
+        return t.cloneNode(ref);
+      }
+      // avoid always assigning this if ctorThises count more than 1
+      if (isDerived || ctorThises.length > 1) {
```
allex: thanks for @nicolo-ribaudo suggestion, skip assigning if REFs less than one.

allex:
@nicolo-ribaudo could you please review the latest mr changes?

nicolo-ribaudo:
I won't have access to a PC until next week, but two things come to my mind:
1) Why do we need to assign `this` to a variable in non-derived constructors?
2) This code should throw before alerting `1`: https://babeljs.io/repl/build/8754/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEBiD29oFMAeAXZA7AJjUkMA3gL7REBQ0ABcPFhOgE4Cuw68TAFAJTlXUa6ABYBLCAG4BNMCGRN0XAIw8BJCuopZkAdziIps-YoBMPCUA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&sourceType=module&lineWrap=false&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.0.0-beta.55%2Bpr.8405&envVersion=

allex:
hi @nicolo-ribaudo,

Thanks for your test case, just add new test at :`https://github.com/babel/babel/pull/8405/files#diff-40845640a1dc1f671fa4cbb1c4ef637a`

For assign this to a variable in non-derived constructors, in my opinion, first of all we can make the constructor style consistent.  and reassign `this` as a
local variable can reduce the file size quite a bit when been uglified.

allex:
@nicolo-ribaudo, any problem about this optims?

rubennorte:
Thanks for working on this @allex! I haven't looked at the code yet but I see some cases where this might fail. For example, [when we have a `this` that's conditionally evaluated in a logical expression](https://babeljs.io/repl/build/8767/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=true&code_lz=MYGwhgzhAEBCYCdoG8C-AodpIwGIHt9oBTADwBdiA7AExniWXWmmHyonIQFdhz8EACgCUKZi2gAzMCAjFoAMgXRyACwCWEANziWazTpYYMQA&debug=false&forceAllTransforms=true&shippedProposals=true&circleciRepo=&evaluate=true&fileSize=false&sourceType=script&lineWrap=false&presets=es2015%2Ces2015-loose&prettier=true&targets=Node-8.9&version=7.0.0-beta.56%2Bpr.8405&envVersion=2.0.0-beta.2).

It looks like the code path analysis should be better to handle this cases.

allex:
@rubennorte [fixed](https://babeljs.io/repl/build/8840/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=true&code_lz=MYGwhgzhAEBCYCdoG8C-AodpIwGIHt9oBTADwBdiA7AExniWXWmmHyonIQFdhz8EACgCUKZi2gAzMCAjFoAMgXRyACwCWEANziWazTpYYM6KsQDu0AviA&debug=false&forceAllTransforms=true&shippedProposals=true&circleciRepo=&evaluate=true&fileSize=false&sourceType=script&lineWrap=false&presets=es2015%2Ces2015-loose&prettier=true&targets=Node-8.9&version=7.0.0-rc.1%2Bpr.8405&envVersion=2.0.0-beta.2) now.

nicolo-ribaudo:
This throws but it shouldn't:
```js
class Bar {}

class Foo extends Bar {
  constructor() {
    false && this;
    super();
  }
}

new Foo
```

<details>
<summary>Output</summary>

```js
var Foo =
/*#__PURE__*/
function (_Bar) {
  _inherits(Foo, _Bar);

  function Foo() {
    var _this;

    _classCallCheck(this, Foo);

    _assertThisInitialized(_this)

    false && _this;
    return _this = _possibleConstructorReturn(this, _getPrototypeOf(Foo).call(this));
  }

  return Foo;
}(Bar);
```

</details>

---

This PR is revealing to be very complex to get right. I think that (at least for now), se should only avoid `_assertThisInitialized(_assertThisInitialized(_this))`. Maybe we could also optimize the simplest case (`super();` called in an expression statement not inside a loop/if/arrow fn, with no preceding this).

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/declaration-order/output.js
```diff
@@ -1,9 +1,11 @@
 var C = function C() {
   "use strict";
 
+  var _this = this;
+
   babelHelpers.classCallCheck(this, C);
-  this.y = babelHelpers.classPrivateFieldLooseBase(this, _x)[_x];
-  Object.defineProperty(this, _x, {
+  _this.y = babelHelpers.classPrivateFieldLooseBase(_this, _x)[_x];
```
nicolo-ribaudo: (Stylistic opinion)
I highly prefer `this` over `_this` when possible. I find it clearer, and it is gzipped as fine as `_this`.

allex:
sorry for losing  some tests.
```
diff --git a/packages/babel-plugin-transform-classes/test/fixtures/spec/derived-constructor-must-call-super-2/input.js b/packages/babel-plugin-transform-classes/test/fixtures/spec/derived-constructor-must-call-super-2/input.js
index 985f12206..4cdce429f 100644
--- a/packages/babel-plugin-transform-classes/test/fixtures/spec/derived-constructor-must-call-super-2/input.js
+++ b/packages/babel-plugin-transform-classes/test/fixtures/spec/derived-constructor-must-call-super-2/input.js
@@ -3,3 +3,10 @@ class Foo extends Bar {
     if (eval("false")) super();
   }
 }
:
@@ -5,5 +5,27 @@ class Foo extends Bar {
     if (eval("false")) super();
   }
 }
-
 expect(() => new Foo()).toThrow("this hasn't been initialised");
+
+
+class Ok extends Bar {
+  constructor() {
+    false && this;
+    super()
+  }
+}
+expect(() => new Ok()).not.toThrow();
```
this PR mainly do somethings:
1. cleanup some duplicate `assertThisInitialized` calls.
2. removed `assertThisInitialized` if `super()` before all of these `this` at the same statement-scope level. (striped lots of no-used asserts.)

PR #8409, opened 2018-07-31
Add LICENSE file to published npm packages [skip ci]

opichals:
| Fixed Issues?            |  Legal licensing compliance

This PR is a clone of https://github.com/babel/babel/pull/7308 which based on this
https://github.com/babel/babel/pull/7308#issuecomment-367686736 should be acceptable.

It clones the LICENSE file to all the packages before publishing using lerna.


existentialism:
Lerna itself can actually [do this now](https://github.com/lerna/lerna/pull/1465), so it's probably better we spend the effort on upgrading to the latest beta?

opichals:
@existentialism You are right, it would be nice. However till lerna v3.0 reaches GA this could be the interim solution.

This issue is also affecting the previous major versions and it would be nice to have it also backported to at least the v6.x branch.

opichals:
@existentialism Added #8413 just in case it is more likely to get us to legally acceptable babel packages sooner. :)

honzajavorek:
Can we get this or https://github.com/babel/babel/pull/8413 merged? It's very problematic to use @existentialism @xtuc Babel as of now in an environment where people actually do care about licenses. The code distributed in the npm packages is effectively unlicensed, thus copyrighted and proprietary.

Even code distributed on CDNs ([babel-standalone](https://babeljs.io/docs/en/next/babel-standalone.html)) is not licensed in any way. There is currently no existing distribution of Babel, which would be properly licensed.

We are forced to consider migrating to TypeScript, but it seems silly if the only change needed is to copy license files into subdirectories and publish a patch version.

.gitignore
```diff
@@ -1,6 +1,7 @@
 .DS_Store
 /node_modules
 /packages/*/node_modules
+/packages/*/LICENSE
```
loganfsmyth: I think we should explicitly exclude `babel-parser|babel-plugin-transform-object-assign` here, e.g.
```
/packages/*/LICENSE
!/packages/babel-parser/LICENSE
!/packages/babel-plugin-transform-object-assign/LICENSE
```

hzoo:
Thanks for the PR, added gitignore for the files where necessary!

I went with this one since we probably aren't ready to upgrade to Lerna 3 (since possible issues that I don't have time to test atm). Thanks everyone.

honzajavorek:
@hzoo Thank you! :heart:

PR #8410, opened 2018-07-31
Babel standalone type module

lukejacksonn:
| Fixed Issues?            | Fixes https://github.com/babel/babel-standalone/issues/96
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 

I have been building an in browser based dev environment that takes advantage of script `type="module"` with native static `import` statements. Everything worked great until I tried building an app that uses JSX. For that to work in the browser it needs to be transpiled.

I employed `@babel/standalone` to do this for me in the browser. It was at this point that I realised a conflict in script types; I needed `<script type="text/babel">` to flag to babel standalone that the code was to be transpiled but also required `<script type="module">` in order for native imports to work correctly.

I searched the `babel-standalone` repo for similar issues and found this  https://github.com/babel/babel-standalone/issues/96. So I added this comment https://github.com/babel/babel-standalone/issues/96#issuecomment-395994115 questioning wether use of the `type` attribute was required. Some others replied saying they had similar issues but we got no reply for any maintainers and the issue was subsequently archived a week or so later.

So here is my suggestion but in working code. The changes are minimal but have the desired effect; if you mark a script with `type="text/jsx/module"` then babel standalone picks up that the script needs to be transpiled and adds `type="module"` to the resultant script tag which gets appended to the head.

Thanks for your time 🙇

### DIFF
--- a/packages/babel-standalone/examples/scriptTag-module.htm
+++ b/packages/babel-standalone/examples/scriptTag-module.htm
@@ -1,0 +1,22 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8" />
+  <title>babel-standalone example - External script</title>
+</head>
+<body>
+  Using Babel <strong id="version"></strong>:
+  <pre id="output">Loading...</pre>
+
+  <script src="../babel.js"></script>
+  <script type="text/babel/module" data-presets='es2016'>
+    import vuid from 'https://unpkg.com/vuid'
+    const doStuff = () => {
+      const name = vuid();
+      document.getElementById('output').innerHTML = `Hello ${name}`;
+      document.getElementById('version').innerHTML = Babel.version;
+    };
+    doStuff();
+  </script>
+</body>
+</html>
--- a/packages/babel-standalone/src/transformScriptTags.js
+++ b/packages/babel-standalone/src/transformScriptTags.js
@@ -55,6 +55,9 @@ function buildBabelOptions(script) {
 function run(transformFn, script) {
   const scriptEl = document.createElement("script");
   scriptEl.text = transformCode(transformFn, script);
+  if (script.module) {
+    scriptEl.type = "module";
+  }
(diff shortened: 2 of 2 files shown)


packages/babel-standalone/examples/scriptTag-module.htm
```diff
@@ -0,0 +1,22 @@
...
+<head>
+  <meta charset="utf-8" />
+  <title>babel-standalone example - External script</title>
+</head>
+<body>
+  Using Babel <strong id="version"></strong>:
+  <pre id="output">Loading...</pre>
+
+  <script src="../babel.js"></script>
+  <script type="text/babel/module" data-presets='es2016'>
```
loganfsmyth: If we want to do this, we should use a mimetype parameter like `text/babel; goal=module`

lukejacksonn: I'd be happy with that. I used `indexOf` because I saw it used here https://github.com/babel/babel/pull/8410/files#diff-7a71ff75920619cbc0e67883df263ca8L188. Is there anywhere in the codebase (or otherwise) you can point me toward regard parsing something like the mimetype you suggest?

loganfsmyth: Gotcha, yeah that at least does `type.split(";")` so it's not as generic. I'd vote for 
```
script.type.split(';').slice(1).some(meta => /^\s*type\s*=\s*module\s*$/.test(meta))
```
probably.

lukejacksonn: Cool, thanks! I will give it a go 👍

toqueteos:
I just found this PR, any status updates?

nicolo-ribaudo:
We added support for `data-type="module"`, so that you can use native modules. Even if we went with another design, thanks for the PR!

https://github.com/babel/babel/pull/11466

PR #8414, opened 2018-08-01
Allow patterns as argument of RestElement

### LINKED ISSUE #5673: Destructuring rest parameters
Based on ES6 documentation from [here](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/rest_parameters#Destructuring_rest_parameters) it should be possible to destruct rest parameters. For example:

```
function f(...[a, b, c]) {
  return a + b + c;
}
```
or
```
function a (...[, secondParamOnly]) {
  // use second param
}
```

Is there any plugin/version of babel I can use to accomplish this? Thanks.

### LINKED ISSUE #8396: parse error on BindingPattern in BindingRestElement
## Bug Report

**Current Behavior**
Doesn't parse BindingPattern in BindingRestElement

**Input Code**

```js
let [...[a]] = [1];

function foo(...{length}) { console.log(length); }
```

**Expected behavior/code**
BindingRestElement can contain BindingPattern. Therefore the variable declaration should parse without problems.
FunctionRestParameter is basically a BindingRestElement and therefore also allows BindingPattern.

https://tc39.github.io/ecma262/#prod-BindingRestElement
https://tc39.github.io/ecma262/#prod-FunctionRestParameter

Note that BindingPattern is not allowed in BindingRest**Property**. Therefore the following code doesn't need to parse correctly:

```js
let {...{}} = {};
```

whatever the default in the REPL is

**Environment**
- Babel version(s): 6.26.0
- How you are using Babel: repl

**Possible Solution**

microbouji:
| Fixed Issues?            | Fixes #8396, fixes #5673  
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Patterns as arguments to `RestElement`s were apparently removed back in https://github.com/babel/babel/commit/9a97d92217dffcf6478611067c1525fa4004fce4 when, as mentioned there, they were not part of es6. They're part of the standard now, so this PR adds them back in.

```js
function foo(...[a]) {}
```

will now get converted to:

```js
function foo() {
  for (var _len = arguments.length, _ref = new Array(_len), _key = 0; _key < _len; _key++) {
    _ref[_key] = arguments[_key];
  }

  var [a] = _ref;
}

```

letting the destructuring plugin handle it from there (`[a] = _ref` => `a = _ref[0]`).

Note that this does not make any of the optimizations that are otherwise made when the argument is an identifier - e.g. even though `a` is not referenced inside the function it will still be made available. If any optimizations are possible I think they can be added in other PRs?

~~I'll try to make a backport for v6 as well.~~ EDIT: Don't think it's possible to change babylon 6.x anymore.

### DIFF
--- a/packages/babel-plugin-transform-parameters/src/rest.js
+++ b/packages/babel-plugin-transform-parameters/src/rest.js
@@ -228,9 +228,19 @@ export default function convertFunctionRest(path) {
   const { node, scope } = path;
   if (!hasRest(node)) return false;
 
-  const rest = node.params.pop().argument;
+  let rest = node.params.pop().argument;
 
   const argsId = t.identifier("arguments");
+
+  if (t.isPattern(rest)) {
+    const pattern = rest;
+    rest = scope.generateUidIdentifier("ref");
+
+    const declar = t.variableDeclaration("let", [
+      t.variableDeclarator(pattern, rest),
+    ]);
+    node.body.body.unshift(declar);
+  }
 
   // check and optimise for extremely common cases
   const state = {
--- a/packages/babel-plugin-transform-parameters/test/fixtures/parameters/rest-patterns/exec.js
+++ b/packages/babel-plugin-transform-parameters/test/fixtures/parameters/rest-patterns/exec.js
@@ -1,0 +1,5 @@
+function foo(...{ length }) {
+  return length;
+}
+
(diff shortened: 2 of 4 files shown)


nicolo-ribaudo:
Thank you!

PR #8418, opened 2018-08-03
Babel cli fix recursion in file mode

bitcloud:
| Fixed Issues?            | 
| Documentation PR Link    | 

### Why
While using babel-cli in file mode and passing it a folder to recurse throu (e.g. `babel src --out-file test.js`) it just scans for all files in the provided directory but does not recurse into it's subdirectories.

Currently it uses an extension filter that is applied to all folder elements before processing. But the extension filter doesn't exclude directories from filtering.
This results in unexpected behaviour because it would recurse into folders that would have an extension, but not in those who don't. And concerning the documentation and its behaviour in its previous version, it should recurse into all directories.

### How
The PR adds an additional check to the recurse filter and accepts all directories.


TrySound:
@bitcloud You should've upgrade `fs-readdir-recursive` package. 1.0.0 doesn't have api you use in this PR. Because of this change after updating from 55 to 56 beta we get the error like this
https://github.com/babel/babel/issues/8431

bitcloud:
ah, damn, sorry about that. Was working with my current freshly installed env so didn't catch that one. Is there actually a way to install all minimal versions of required packages to test something like that?

PR #8429, opened 2018-08-06
Use preferred property syntax in polyfill

zachmullen:
| Fixed Issues?            | 
| Patch: Bug Fix?          | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -883,7 +883,7 @@ helpers.iterableToArrayLimit = helper("7.0.0-beta.0")`
       _e = err;
     } finally {
       try {
-        if (!_n && _i["return"] != null) _i["return"]();
+        if (!_n && _i.return != null) _i.return();
       } finally {
         if (_d) throw _e;
       }


xtuc:
May I ask why this is needed?

zachmullen:
Was there a reason it was written as `["return"]` instead of the way properties are referenced in the rest of the file? If not, there is value in consistency.

hzoo:
It's probably since `return` is a reserved word, and the helpers themselves are not compiled against what browsers you support so it's matched against the lowest common denominator which is ES3.

zachmullen:
If you look around the same file I edited, you'll notice the property named `return` is referenced with dot notation elsewhere, so that doesn't seem like the explanation.

xtuc:
I would actually prefer to use array access because of ES3 compat (which already caused an issue on one of my projects).

zachmullen:
Ok, in that case I'll close this.

PR #8435, opened 2018-08-08
Allow transform-runtime to insert runtime references with absolute paths.

loganfsmyth:
| Fixed Issues?            | Fixes feature request in https://github.com/babel/babel/pull/8266#issuecomment-411069941 for runtime transform
| Patch: Bug Fix?          |
| Tests Added + Pass?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This allows users to run `transform-runtime` broadly across a whole project. By default, `transform-runtime` imports from `@babel/runtime/foo` directly, but that only works if `@babel/runtime` is in the node modules of the file that is being compiled. This can be problematic for nested node modules, or npm-linked modules, among other cases. To avoid worrying about how the runtime module's location, this allows users to resolve the runtime once up front, and then insert absolute paths to the runtime into the output code.

Using absolute paths is not desirable if files are compiled for use at a later time, but in contexts where a file is compiled and then immediately consumed, they can be quite helpful.

This PR exposes two approaches for resolving the runtime:

* `absoluteRuntime: true` - Resolve the runtime:
  * relative to Babel's `cwd` if the `transform-runtime` plugin is passed as a programmatic arg to Babel
  * relative to the config file, if `transform-runtime` is enabled via a config file
* `absoluteRuntime: '/projects/thing/src'`
  * Resolve the runtime relative to some specific location, for instance if you wanted to pass the `__dirname` of some general build process script. e.g. will look at
    * `/projects/thing/src/node_modules/@babel/runtime/foo`
    * `/projects/thing/node_modules/@babel/runtime/foo`
    * `/projects/node_modules/@babel/runtime/foo`
    * `/node_modules/@babel/runtime/foo`

cc @insin

### DIFF
--- a/packages/babel-plugin-transform-runtime/src/index.js
+++ b/packages/babel-plugin-transform-runtime/src/index.js
@@ -1,10 +1,31 @@
+import path from "path";
+import resolve from "resolve";
 import { declare } from "@babel/helper-plugin-utils";
 import { addDefault, isModule } from "@babel/helper-module-imports";
 import { types as t } from "@babel/core";
 
 import definitions from "./definitions";
 
-export default declare((api, options) => {
+function resolveAbsoluteRuntime(moduleName: string, dirname: string) {
+  try {
+    return path.dirname(
+      resolve.sync(`${moduleName}/package.json`, { basedir: dirname }),
+    );
+  } catch (err) {
+    if (err.code !== "MODULE_NOT_FOUND") throw err;
+
+    throw Object.assign(
+      new Error(`Failed to resolve "${moduleName}" relative to "${dirname}"`),
+      {
+        code: "BABEL_RUNTIME_NOT_FOUND",
+        runtime: moduleName,
+        dirname,
+      },
+    );
+  }
+}
+
+export default declare((api, options, dirname) => {
   api.assertVersion(7);
 
   const {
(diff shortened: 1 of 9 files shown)


packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -1,10 +1,31 @@
...
 import { addDefault, isModule } from "@babel/helper-module-imports";
 import { types as t } from "@babel/core";
 
 import definitions from "./definitions";
 
-export default declare((api, options) => {
+function resolveAbsoluteRuntime(moduleName: string, dirname: string) {
+  try {
+    return path.dirname(
+      resolve.sync(`${moduleName}/package.json`, { basedir: dirname }),
```
nicolo-ribaudo: Why do we need to look for package.json?

loganfsmyth: The runtime doesn't have a `main` or `index` because it's meant to be a collection of independent files. Since we want to resolve the directory location though, we do need a specific file to resolve, and `package.json` fits those requirements well.

nicolo-ribaudo:
Could you add some tests (maybe just for the string value, since `true` depends on where Babel is cloned)?

PR #8461, opened 2018-08-13
Fix _isNativeReflectConstruct helper

### LINKED ISSUE #8419: isNativeReflectConstruct may not be always work
## Bug Report

**Current Behavior**
`isNativeReflectConstruct` helper tests if `Reflect.construct` is legit by using `Date.p.toString`'s brand check. However, `toString` [used to be](https://github.com/tc39/ecma262/issues/1268#issuecomment-410290967) a generic method (it could operate on any object, not just real `Date` instances).

We may need to update this to use another brand checking method, such as `Map.p.get`.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #8419
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

> Date#toString is generic in ES2015 [1] and it doesn't throw, but it returns "Invalid Date"
>
> [1]: tc39/ecma262#1268 (comment)

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -506,12 +506,10 @@ helpers.setPrototypeOf = helper("7.0.0-beta.0")`
 
 helpers.isNativeReflectConstruct = helper("7.9.0")`
   export default function _isNativeReflectConstruct() {
-    if (
-      typeof Reflect === "undefined" ||
-      !Reflect.construct ||
-      // core-js@3
-      Reflect.construct.sham
-    ) return false;
+    if (typeof Reflect === "undefined" || !Reflect.construct) return false;
+
+    // core-js@3
+    if (Reflect.construct.sham) return false;
 
     // Proxy can't be polyfilled. Every browser implemented
     // proxies before or at the same time as Reflect.construct,
@@ -524,13 +522,10 @@ helpers.isNativeReflectConstruct = helper("7.9.0")`
     // use our fallback implementation.
     try {
       // If the internal slots aren't set, this throws an error similar to
-      //   TypeError: this is not a Date object.
-      // Date#toString is generic in ES2015 [1] and it doesn't throw, but it
-      // returns "Invalid Date" if the receiver is not a properly initialized
-      // Date object.
-      // [1]: https://github.com/tc39/ecma262/issues/1268#issuecomment-410104832
+      //   TypeError: this is not a Boolean object.
 
-      return Date.prototype.toString.call(Reflect.construct(Date, [], function() {})) !== "Invalid Date";
+      Boolean.prototype.valueOf.call(Reflect.construct(Boolean, [], function() {}));
+      return true;
     } catch (e) {
       return false;
     }
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -506,10 +506,12 @@ helpers.setPrototypeOf = helper("7.0.0-beta.0")`
 
 helpers.isNativeReflectConstruct = helper("7.9.0")`
   export default function _isNativeReflectConstruct() {
-    if (typeof Reflect === "undefined" || !Reflect.construct) return false;
-
-    // core-js@3
-    if (Reflect.construct.sham) return false;
+    if (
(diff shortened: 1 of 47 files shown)


packages/babel-helpers/src/helpers.js
```diff
@@ -472,7 +472,15 @@ helpers.construct = helper("7.0.0-beta.0")`
     try {
       // If the internal slots aren't set, this throws an error similar to
       //   TypeError: this is not a Date object.
+      // Date#toString is generic in ES2015 [1] and it doesn't throw, so we also
+      // check Map#get (which on the other hand doesn't exist in ES5 browsers).
+      //
+      // [1]: https://github.com/tc39/ecma262/issues/1268#issuecomment-410104832
+
       Date.prototype.toString.call(Reflect.construct(Date, [], function() {}));
```
JLHwung: Per [20.3.4.1](https://www.ecma-international.org/ecma-262/6.0/#sec-date.prototype.tostring), if `Reflect.construct(Date, [], function() {})` is not a Date object, it must returns `"Invalid Date"`, thus it suffices to check if it does _not_ return "Invalid Date".

It helps shrink the code size a bit as it is included in the `wrapNativeClass` helpers.

packages/babel-helpers/src/helpers.js
```diff
@@ -506,10 +506,12 @@ helpers.setPrototypeOf = helper("7.0.0-beta.0")`
...
-    if (typeof Reflect === "undefined" || !Reflect.construct) return false;
-
-    // core-js@3
-    if (Reflect.construct.sham) return false;
+    if (
+      typeof Reflect === "undefined" ||
+      !Reflect.construct ||
+      // core-js@3
+      Reflect.construct.sham
+    ) return false;
```
jridgewell, line 514: Weak Nit: I actually liked the previous style better.

packages/babel-helpers/src/helpers.js
```diff
@@ -523,8 +525,12 @@ helpers.isNativeReflectConstruct = helper("7.9.0")`
...
       // If the internal slots aren't set, this throws an error similar to
       //   TypeError: this is not a Date object.
-      Date.prototype.toString.call(Reflect.construct(Date, [], function() {}));
-      return true;
+      // Date#toString is generic in ES2015 [1] and it doesn't throw, but it
+      // returns "Invalid Date" if the receiver is not a properly initialized
+      // Date object.
+      // [1]: https://github.com/tc39/ecma262/issues/1268#issuecomment-410104832
+
+      return Date.prototype.toString.call(Reflect.construct(Date, [], function() {})) !== "Invalid Date";
```
jridgewell, line 533: We could also change this to `Boolean.prototype.valueOf`:

- ES5: https://262.ecma-international.org/5.1/#sec-15.6.4.3
- ES6: https://262.ecma-international.org/6.0/#sec-boolean.prototype.valueof
- Next: https://tc39.es/ecma262/#sec-boolean.prototype.valueof

jridgewell (approved):

packages/babel-helpers/src/helpers.js
nicolo-ribaudo: Me too, but I noticed that merging them minifies better (terser doesn't merge the two if statements):

jridgewell: It should, I made the PR myself 😉 : https://github.com/terser/terser/pull/555

JLHwung (approved):

packages/babel-helpers/src/helpers.js
nicolo-ribaudo: Not with this code :thinking:

```js
// write or paste code here

export function f() {
  if (!a) return false;
  if (!b) return false;
  try {
    foo();
    return true;
  } catch {
    return false;
  }
}
```

(tested at https://try.terser.org/)

Anyway, I'm happy to revert this change since it will be eventually handled by terser.

jridgewell: Ahh, it's because it can't collapse the try-catch statement into an expression, so the if-statements exist only as a series of if-statements and not a nested ternary. Open an issue?

PR #8465, opened 2018-08-13
Require decoratorsBeforeExport option for decorators

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | yes
| Minor: New Feature?      |
| Tests Added + Pass?      | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

It has not been decided yet where decorators should go.

### DIFF
--- a/packages/babel-parser/test/plugin-options.js
+++ b/packages/babel-parser/test/plugin-options.js
@@ -14,9 +14,10 @@ describe("plugin options", function() {
     const OPT_2 = [NAME, { decoratorsBeforeExport: false }];
     const SYNTAX_1 = "@dec export class C {}";
     const SYNTAX_2 = "export @dec class C {}";
-    const SYNTAX_DEFAULT = "export @dec class C {}";
+    const SYNTAX_DEFAULT = "";
 
-    it("when they aren't specified", function() {
+    // Currently every option is required, so this can't be tested.
+    it.skip("when they aren't specified", function() {
       expect(getParser(SYNTAX_DEFAULT, [NAME, OPT_1])).not.toThrow();
       expect(getParser(SYNTAX_DEFAULT, [NAME, OPT_2])).not.toThrow();
     });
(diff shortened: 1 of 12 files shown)


packages/babel-parser/test/plugin-options.js
```diff
@@ -14,9 +14,10 @@ describe("plugin options", function() {
     const OPT_2 = [NAME, { decoratorsBeforeExport: false }];
     const SYNTAX_1 = "@dec export class C {}";
     const SYNTAX_2 = "export @dec class C {}";
-    const SYNTAX_DEFAULT = "export @dec class C {}";
+    const SYNTAX_DEFAULT = "";
 
-    it("when they aren't specified", function() {
+    // Currently every option is required, so this can't be tested.
+    it.skip("when they aren't specified", function() {
```
xtuc: Maybe we can just remove this dead code?

nicolo-ribaudo: I forgot about flow's `all` option :laughing: I re-enabled the test using it.

wyuenho:
This change makes no sense, `plugin-utils.js` throws as soon as `decoratorsBeforeExport` is undefined, but `babel-plugin-syntax-decorators/src/index.js` suggests that as long as `legacy: true` implies `decoratorsBeforeExport` is optional. Which exactly is the intention of this change?

nicolo-ribaudo:
Are you using Babel or the parser? They have slightly different options:

| *Babel*                                                                    | *@babel/parser*                                       |
|----------------------------------------------------------------------------|-------------------------------------------------------|
| `["@babel/plugin-syntax-decorators", { "legacy": true }]`                  | `"decorators-legacy"`                                 |
| `["@babel/plugin-syntax-decorators", { "decoratorsBeforeExport": true }]`  | `["decorators", { "decoratorsBeforeExport": true }]`  |
| `["@babel/plugin-syntax-decorators", { "decoratorsBeforeExport": false }]` | `["decorators", { "decoratorsBeforeExport": false }]` |

PR #5050, opened 2016-12-28
Rewrite Hub as interface #5047

yongxu:
| Q                        | A 
| ------------------------ | ---
| Patch: Bug Fix?          |  yes
| Major: Breaking Change?  |  no
| Minor: New Feature?      |  yes
| Deprecations?            |  no
| Spec Compliancy?         |  yes
| Tests Added/Pass?        |  yes
| Fixed Tickets            | Fixes #5047, Fixes #4640, Fixes #4413 
| License                  | MIT
| Doc PR                   | no
| Dependency Changes       |  no

This PR implements Rewrite Hub as interface #5047 <--check out more details in this issue
will make `babel-traverse` a standalone package

PR #8469, opened 2018-08-14
Rewrite Hub as interface

### LINKED ISSUE #4640: babel-traverse throws on import and variable collisions
### Input Code

In a new folder run `npm install babel-core babylon`, then put the following code in `traverse-error.js`

``` js
var code = `
import foo from './foo'
const foo = undefined
`

var babelCore = require('babel-core')
var babylon = require('babylon')

var ast = {
    type: 'File',
    start: 0,
    end: code.length,
    program: babylon.parse(code, {
        sourceType: 'module'
    })
}

babelCore.traverse(ast, {})
```
### Expected Behavior

`node traverse-error.js` should run with no output. If you simply rename the `import foo` or `const foo` to avoid the collision, it will run as expected.
### Current Behavior

```
asa@asa:/tmp/babel-issue$ node traverse-error.js 
/tmp/babel-issue/node_modules/babel-traverse/lib/scope/index.js:397
      throw this.hub.file.buildCodeFrameError(id, messages.get("scopeDuplicateDeclaration", name), TypeError);
                    ^

TypeError: Cannot read property 'file' of undefined
    at Scope.checkBlockScopedCollisions (/tmp/babel-issue/node_modules/babel-traverse/lib/scope/index.js:397:21)
    at Scope.registerBinding (/tmp/babel-issue/node_modules/babel-traverse/lib/scope/index.js:579:16)
    at Scope.registerDeclaration (/tmp/babel-issue/node_modules/babel-traverse/lib/scope/index.js:483:14)
    at Object.BlockScoped (/tmp/babel-issue/node_modules/babel-traverse/lib/scope/index.js:185:28)
    at Object.newFn (/tmp/babel-issue/node_modules/babel-traverse/lib/visitors.js:318:17)
    at NodePath._call (/tmp/babel-issue/node_modules/babel-traverse/lib/path/context.js:76:18)
    at NodePath.call (/tmp/babel-issue/node_modules/babel-traverse/lib/path/context.js:44:14)
    at NodePath.visit (/tmp/babel-issue/node_modules/babel-traverse/lib/path/context.js:105:12)
    at TraversalContext.visitQueue (/tmp/babel-issue/node_modules/babel-traverse/lib/context.js:150:16)
    at TraversalContext.visitMultiple (/tmp/babel-issue/node_modules/babel-traverse/lib/context.js:103:17)
```
### Context

I have an Atom plugin called [js-hyperclick](https://atom.io/packages/js-hyperclick). It parses your JS files and allows you to ctrl+click filenames or variables to jump to their definitions. Today I've been working on getting `js-hyperclick` to inform the user when it can't function. If you have a syntax error that causes `babylon.parse` to throw I can help the user:

selection_019

Because this error happens when traversing and the error points to `babel-traverse` code, showing the user doesn't help them:

selection_020

I tend to run into this when I'm trying to extract `foo` into its own file. I'll add `import foo from './new-file'`, then if this were able to parse I could ctrl+click the `./new-file` or `foo` to open a new editor at that path and cut/paste whatever `foo` needs.
### Your Environment

| software | version |
| --- | --- |
| Babel | 5.8.25 and 6.17.0 |
| node | 4 and 6 |

I wouldn't expect it to get fixed back in Babel5, but the code above also demonstrates the error with `npm install babel-core@5` and replace `babylon.parse` with `babelCore.parse`

### LINKED ISSUE #5047: Rewrite Hub as interface
The current `babel-traverse` implementation of `Hub` will cause bugs if `this.hub.file` does not exist, which is only injected by `File` class in `babel-core`. 

This means `babel-traverse` package can not be used without `babel-core` unless we monkey-patch hub with a fake file object. The `Hub` class itself doesn't carry any useful information besides the `file` reference.

Defining hub as an interface instead of an object, will make `babel-traverse` a standalone package and resolve issues #4640 #4413 
### Solution
we can redefine Hub(or a better name maybe) as 
```js
// all the methods should be optional(or make hub optional?), so there will be no exception
declare interface Hub {
  mark(type: string, message: string): void;
  getCode(): string;
  getFileScope(): Scope; // or just getFile, maybe?
  onError(node: Object, msg: string, Error: typeof Error = SyntaxError): void;
}
```
This interface would be sufficient for the current version, and can be extended in the future.
It also can be checked by flowtype

Here are the whys, where `Hub` appears
```js
// this function is only used in debugging
  mark(type: string, message: string) {
    this.hub.file.metadata.marked.push({
      type,
      message,
      loc: this.node.loc
    });
  }

// this can be replaced by getCode()
export function getSource() {
  let node = this.node;
  if (node.end) {
    return this.hub.file.code.slice(node.start, node.end);
  } else {
    return "";
  }
}

// JSX plugin
  let visitor = {
    JSXOpeningElement(path, state) {
//..........................................
          path.hub.file.scope.push({id: fileNameIdentifier, init: t.stringLiteral(fileName)});

// and called buildCodeFrameError in few places
throw this.hub.file.buildCodeFrameError
```

Please let me know if you have any suggestion, or better improvement! I will make a PR if I can get some positive faceback :)

### more about hub
the current implementation of hub itself only consist 6 lines of code, and the `options` has never been used anywhere:
```js
export default class Hub {
  constructor(file, options) {
    this.file = file;
    this.options = options;
  }
}
```

Here are all the places hub appears, when searching keyword `hub` in package/:
image

nicolo-ribaudo:
| Fixed Issues?            | Fixes #5047, Fixes #4640
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Ref: #5050. Check #5047 for more details.

### DIFF
--- a/packages/babel-core/src/transformation/file/file.js
+++ b/packages/babel-core/src/transformation/file/file.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as helpers from "@babel/helpers";
-import { NodePath, Hub, Scope } from "@babel/traverse";
+import { NodePath, Scope, type HubInterface } from "@babel/traverse";
 import { codeFrameColumns } from "@babel/code-frame";
 import traverse from "@babel/traverse";
 import * as t from "@babel/types";
@@ -27,9 +27,17 @@ export default class File {
   ast: Object = {};
   scope: Scope;
   metadata: {} = {};
-  hub: Hub = new Hub(this);
   code: string = "";
   inputMap: Object | null = null;
+
+  hub: HubInterface = {
+    // keep it for the usage in babel-core, ex: path.hub.file.opts.filename
+    file: this,
+    getCode: () => this.code,
+    getScope: () => this.scope,
+    addHelper: this.addHelper.bind(this),
+    buildError: this.buildCodeFrameError.bind(this),
+  };
 
   constructor(options: {}, { code, ast, inputMap }: NormalizedFile) {
     this.opts = options;
--- a/packages/babel-helper-module-imports/src/import-builder.js
+++ b/packages/babel-helper-module-imports/src/import-builder.js
@@ -10,11 +10,11 @@ export default class ImportBuilder {
   _resultName = null;
 
   _scope = null;
-  _file = null;
+  _hub = null;
 
-  constructor(importedSource, scope, file) {
+  constructor(importedSource, scope, hub) {
     this._scope = scope;
-    this._file = file;
+    this._hub = hub;
     this._importedSource = importedSource;
   }
 
@@ -91,10 +91,10 @@ export default class ImportBuilder {
   }
 
   defaultInterop() {
-    return this._interop(this._file.addHelper("interopRequireDefault"));
+    return this._interop(this._hub.addHelper("interopRequireDefault"));
   }
   wildcardInterop() {
-    return this._interop(this._file.addHelper("interopRequireWildcard"));
+    return this._interop(this._hub.addHelper("interopRequireWildcard"));
   }
 
   _interop(callee) {
--- a/packages/babel-helper-module-imports/src/import-injector.js
+++ b/packages/babel-helper-module-imports/src/import-injector.js
@@ -108,7 +108,7 @@ export default class ImportInjector {
   /**
    * The file used to inject helpers and resolve paths.
    */
-  _file;
+  _hub;
 
   /**
    * The default options to use with this instance when imports are added.
@@ -127,7 +127,7 @@ export default class ImportInjector {
 
     this._programPath = programPath;
     this._programScope = programPath.scope;
-    this._file = programPath.hub.file;
+    this._hub = programPath.hub;
 
     this._defaultOpts = this._applyDefaults(importedSource, opts, true);
   }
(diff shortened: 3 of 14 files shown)


packages/babel-core/src/transformation/file/file.js
```diff
@@ -27,10 +27,18 @@ export default class File {
   ast: Object = {};
   scope: Scope;
   metadata: {} = {};
-  hub: Hub = new Hub(this);
   code: string = "";
   inputMap: Object | null = null;
 
+  hub: HubInterface = {
+    // keep it for the usage in babel-core, ex: path.hub.file.opts.filename
+    file: this,
```
nicolo-ribaudo: @loganfsmyth In the other PR you asked to keep `file` here, but I think that it can be removed? (people can already get it in visitors using `this.file`)

loganfsmyth: @nicolo-ribaudo Yeah, my concern was avoiding breaking changes. I think my preference would still be to keep it at this point since we're already pretty close to release. I don't think we'd want it as part of the interface, but leaving it in the object is so easy. Can we leave it with a big comment saying that we should remove it in 8.x but it's being left for backward-compat with 6.x during 7.x's lifetime?

nicolo-ribaudo: Yes, let's avoid breaking things at this point.

Issue #8411, opened 2018-08-01
path.evaluate wrongly retruns with confidence for "block-scoped" var's


mjesun:
## Bug Report

**Current Behavior**
`path.evaluate` is broken for `var` scopes.

**Input Code**
http://astexplorer.net/#/gist/583ea7d0c0501af69a1b224526f7f864

**Expected behavior/code**
Cannot return with confidence that `doesExist` is true.

**Babel Configuration (.babelrc, package.json, cli command)**

N/A

**Environment**
N/A

mjesun:
And, like #8017, [this also affects the minifier](https://babeljs.io/repl/#?babili=true&browsers=ie%206&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABMOcAUAhAhgJwJSIDeAUIojMImlAJ4AOApnJdjogIQC8iA5OACYNgMMA348CJMmQBuuRPzgMAzgFEAHjGVRE3KDhAMA3KUQBfYqYpVFKjVqiTTZHAyggcSVibJnEDABtlBiJnRAgEZTgAhgA6ALgAczQeVgUlZUQwOB0GTW0JH3NiC2IUdGAsIIY8IyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&sourceType=module&lineWrap=false&presets=es2015%2Cstage-0%2Cstage-1%2Ces2015-loose%2Cbabili&prettier=true&targets=&version=6.26.0&envVersion=1.6.2).

Issue #8017, opened 2018-05-23
`path.evaluate` wrongly returning with confidence on && expression


mjesun:
## Bug Report

**Current Behavior**
`z() && undefined` is `path.evaluate`'d to `undefined` with confident set to `true`

**Input Code**
http://astexplorer.net/#/gist/52894896cb2e791ec2f5b4f405175ec1/1cc1f21330b2b8bc0df178bea039da61c9305af6

**Possible Solution**
Don't return `true` when there can be side-effects 😄

mjesun:
I think the root of the issue is [here](https://github.com/babel/babel/blob/master/packages/babel-traverse/src/path/evaluation.js#L258):

> If we are confident that one side of an && is false, or the left side of an || is true, we can be confident about the entire expression

The "one side of an &&" is not correct, it should be "the right side of an &&".

mjesun:
@loganfsmyth Just a friendly ping, this popped up after the `-0` bug 🙂

mjesun:
Btw, [this affects Babel's minifier too](https://babeljs.io/repl/#?babili=true&browsers=ie%206&build=&builtIns=false&code_lz=GYVwdgxgLglg9mABARwBQEpEG8BQjEBOAplCAUsAIYA2AzkQNw4C-OOAbpQYgF6IC8KDIgBkIxOAAmRYDDBFJTHDGCJUffpsRU6RTLnwQEtONSIA6anADmqAMzomzIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=false&presets=es2015%2Cstage-0%2Cstage-1%2Ces2015-loose%2Cbabili&prettier=true&targets=&version=6.26.0&envVersion=1.6.2).

liuxingbaoyu:
Has been fixed.

PR #8478, opened 2018-08-15
Fix path.scope.rename() to not change break clauses

### LINKED ISSUE #8476: `path.scope.rename()` generates invalid JS code when it encounters `break` statements
## Bug Report

**Current Behavior**
`path.scope.rename()` generates invalid JS code when it encounters `break` statements. It detects the identifier of the `break` statement as a standard scope binding and renames it.

**Input Code**

```js
function foo(a) {
  a: {
    break a;
  };
}
```

**Output Code**

```js
function foo(b) {
  a: {
    break b; // <== Invalid!!!
  };
}
```

*([ASTExplorer example](http://astexplorer.net/#/gist/40efa4a3f7ac193d1201e602b7e1408c/13ca714f5bee88c0e8687aecb4420c69e92920a5))*

**Expected behavior/code**

`path.scope.rename()` should not rename the break statement identifier.

**Environment**
- Babel version(s): Tested on v7.0.0-beta.56
- Node/npm version: Node 8
- OS: OSX 10.13.4
- Monorepo: no
- How you are using Babel: Directly

**Possible Solution**

Only potential solution that I can think of is to add a check in the renaming logic to filter out identifiers that have a BreakStatement parent

rafeca:
| Fixed Issues?            | Fixes #8476 

`path.scope.rename()` currently generates invalid JS code when it encounters break statements: It detects the identifier of the break statement as a standard scope binding and renames it, making the `break` statement point to a potentially invalid breakpoint.

In order to fix that, the easiest way that I found is to add an additional check that ensures that we never rename a `break` identifier (which would not never make sense anyways). If there's a preferred alternative solution let me know and I'll update the PR

A test case that demonstrates the behaviour has been added (note that the test fails without the patch on this PR).


mjesun:
@loganfsmyth it would be nice if someone could also get a look at #8017 and #8411; they are also related to breakages in `Scope` traversal.

nicolo-ribaudo:
1) Do `continue` work?
2) Maybe we should just not consider labels `ReferenceIdentifier`s?

mjesun:
I'm all in for no. 2, yes! (Or, if so, labels should also be).

rafeca:
> 1. Do continue work?

Good point, I also need to check for `continue`...

> 2. Maybe we should just not consider labels ReferenceIdentifiers?

This would make the renaming logic cleaner, but this change could have bigger implications (and I'm not familiar enough with the Babel internals to know if this could break other stuff). Plus, it would be a breaking change.

rafeca:
@nicolo-ribaudo : I've updated the PR to not consider labels ReferenceIdentifiers

nicolo-ribaudo:
Could you also add a test for https://github.com/babel/babel/issues/8491?

PR #8479, opened 2018-08-15
Fixes setter paratemer default value

### LINKED ISSUE #7688: Setter default value
Default value in setter argument causes an error:

```js
const obj = {
	set field(num = 1) {this._num = num;},
	get field() {return this._num;}
};
```

Got unhandled error `setter should have exactly one param`, cause setter transpiled as:
```js
set field() {
	var num = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : 1;
	this._num = num;
}
```

And it hasn't any argument.

Perfectly works in chrome without transpiling:

nikolayemrikh:
| Fixed Issues?            | Fixes #7688
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    |
| Any Dependency Changes?  |

Forces setter's default parameter to use loose transform

### DIFF
--- a/packages/babel-plugin-transform-parameters/src/params.js
+++ b/packages/babel-plugin-transform-parameters/src/params.js
@@ -46,7 +46,7 @@ const iifeVisitor = {
 
 export default function convertFunctionParams(path, loose) {
   const { node, scope } = path;
-  loose = node.kind === "set" ? true : loose;
+
   const state = {
     iife: false,
     scope: scope,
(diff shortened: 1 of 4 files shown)


packages/babel-plugin-transform-parameters/src/params.js
```diff
@@ -60,7 +60,8 @@ export default function convertFunctionParams(path, loose) {
   for (let i = 0; i < params.length; i++) {
     const param = params[i];
 
-    if (param.isAssignmentPattern() && loose) {
+    const paramIsAssignmentPattern = param.isAssignmentPattern();
+    if (paramIsAssignmentPattern && loose) {
```
nicolo-ribaudo: Instead of setting `loose` to true for setters, could you update this check to `(loose || node.kind === "set")`?
It would be a little more descriptive and, if we use `loose` for other parts of the transform on the future, they won't be automatically enabled for setters since they might not be needed.

hzoo: Yep I was thinking that as well!

nikolayemrikh: In the beginning i did so. But later..

PR #8485, opened 2018-08-17
Allow preset-env to toggle module handling based on flags from the caller (like babel-loader)

### LINKED ISSUE #7517: Should we leave the module transform enabled by default for `env` and `es2015` in 7.x?
@TheLarkInn asked, and it seems like a discussion worth having. 

Leaving it off by default would lead to people being more likely to get errors in things like Node by default, but it would potentially make the workflow easier for people using Webpack, since it's usually easier to add additional plugins than it is to remove them, since opting out first is more verbose, so people could do

```
{
  presets: ['@babel/env'],
  env: {
    test: {
      plugins: ['@babel/transform-modules-commonjs'],
    }
  }
}
```
which is a lot nicer than 

```
{
  presets: [['@babel/env', { modules: false }]],
  env: {
    test: {
      plugins: ['@babel/transform-modules-commonjs'],
    }
  }
}
```
or even 
```
{
  presets: [['@babel/env', { modules: false }]],
  env: {
    test: {
      presets: ['@babel/env'],
    }
  }
}
```
at least in my view?

loganfsmyth:
| Fixed Issues?            | Fixes #7517
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

What this PR essentially means is that utilities calling Babel can:

* Pass a name to Babel, so that if there are errors, it can give more useful error messages
* Pass misc flags that presets can hook into

For instance, `babel-loader` would be able to do:
```
babel.transform("code;", {
  filename,
  presets: ["@babel/preset-env"],
  caller: {
    name: "babel-loader",
    supportsStaticESM: true,
  },
});
```
and `preset-env` can automatically change its behavior from `modules: "commonjs"` to `modules: false`.


loganfsmyth:
@Andarist Any thoughts on this, since it'll probably fit nicely into your work on the Rollup plugin too.

packages/babel-preset-env/src/index.js
```diff
@@ -230,12 +230,24 @@ export default declare((api, opts) => {
...
-    plugins.push([getPlugin(moduleTransformations[modules]), { loose }]);
+    // NOTE: not giving spec here yet to avoid compatibility issues when
+    // transform-modules-commonjs gets its spec mode
+    const item = [getPlugin(moduleTransformations[modules]), { loose }];
+
+    if (modules === "auto") {
+      // When the `exclude` callback returns a truthy value, Babel not use the
+      // 'item' plugin here.
+      overrides.push({
+        exclude: (filename, { caller }) => caller && caller.supportsStaticESM,
```
nicolo-ribaudo: Since the caller isn't file-specific, can it be passed via the `api` object instead of using the `exclude` option, which is usually used to exclude plugins based on the file path?

nicolo-ribaudo: Since the `caller` isn't file-specific but it is more similar to `env`, can we pass it using the `api` object instead using the `exclude` option, which is usally used to exclude plugins based on the file path?

packages/babel-core/src/config/config-chain.js
```diff
@@ -50,12 +51,26 @@ export type ConfigContext = {
...
 };
 
 /**
  * Build a config chain for a given preset.
  */
-export const buildPresetChain: (
+export function buildPresetChain(
+  arg: PresetInstance,
+  context: *,
+): ConfigChain | null {
```
nicolo-ribaudo: This function never returns null.

loganfsmyth: Oh good catch.

packages/babel-preset-env/src/index.js
loganfsmyth: That's a great question. Let me revisit that. This approach was from an old branch I had, so I hadn't acutally revisited that. My concern would be caching of plugin instances. The issue with caching is that the API would need to be

```
const supportsStaticESM = api.caller(caller => !!(caller && caller.supportsStaticESM));
```
because your build might be used both with and without static ESM support, and Babel needs to know whether or not it should re-evaluate the plugin function, since it does its best to only evaluate the plugin/preset functions a single time unless told otherwise.

I'm actually fine if we want to expose the `api` approach, but I'd still probably be tempted to leave this exposed in `test/include/exclude` too.

nicolo-ribaudo: Caller is an object that can only contain `null`, `undefined`, booleans, strings, and numbers: it can be `JSON.stringify`ed and stored to invalidate the cache when needed.

nicolo-ribaudo:
Does Babel use the same config validation for config files and options directly passed to `babel.transform`? If yes, we could disallow `caller` in `.babelrc`/`babel.config.js` since users should directly configure the preset.

packages/babel-preset-env/src/index.js
loganfsmyth: That's true, I'm personally not a huge fan of that because we could always want to expand that in the future, and having to re-stringify the object all the time just seems undesirable. At the end of the day I don't feel that strongly about it I guess. One other downside being that the plugin cache would get invalidated if _any_ caller flag changed, even if it was something other than `supportsStaticESM`.

loganfsmyth:
> If not, we could disallow caller in .babelrc/babel.config.js since users should directly configure the preset.

Already done! `caller` is only allowed as part of the programmatic call options to Babel.

packages/babel-preset-env/test/normalize-options.spec.js
```diff
@@ -134,16 +134,20 @@ describe("normalize-options", () => {
...
 
   describe("validateModulesOption", () => {
-    it("`undefined` option returns commonjs", () => {
-      expect(validateModulesOption()).toBe("commonjs");
+    it("`undefined` option returns auto", () => {
+      expect(validateModulesOption()).toBe("auto");
     });
 
-    it("`false` option returns commonjs", () => {
+    it("`false` option returns false", () => {
```
existentialism: 🤦‍♂️

hzoo:
This is really great! Something we should document for our integrations to also take advantage of.

Andarist:
🚀 super! Sorry I've missed this PR before, but this is super awesome - preventing the users misconfiguring their tool in clean way, gonna land this in rollup-plugin-babel once it gets released

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -128,13 +128,16 @@ export const validateIgnoreBrowserslistConfig = (
...
   invariant(
     ModulesOption[modulesOpt] ||
       ModulesOption[modulesOpt] === ModulesOption.false,
-    `Invalid Option: The 'modules' option must be either 'false' to indicate no modules, or a
-    module type which can be be one of: 'commonjs' (default), 'amd', 'umd', 'systemjs'.`,
+    `Invalid Option: The 'modules' option must be one of \n` +
+      ` - 'false' to indicate no module processing\n` +
+      ` - a specific module type: 'commonjs', 'amd', 'umd', 'systemjs'` +
+      ` - 'auto' (default) which will automatically select 'false' if the current\n` +
+      `   process is known to support ES module syntax, or "commonjs" otherwise\n`,
```
szarouski: Hi, I have a ❓ about "if the current process is known to support ES module syntax". I'm not sure what that means. Can someone please clarify? Thank you!

nicolo-ribaudo: For example, if you are calling Babel using `babel-loader`, `modules` will be set to false because webpack supports ES modules

szarouski: Thanks @nicolo-ribaudo, I'll link to your reply from https://github.com/babel/babel-loader/issues/521 😃

PR #8487, opened 2018-08-17
Allow helpers to be put in separate files

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This gives a few advantages:
1) They can be linted and prettified
2) It's easier to review them on GitHub, since they are syntax-highlighted.

I had to set up a separate gulp task because, when using `make watch`, changes to any extracted helper need to trigger a recompilation of `babel-helpers/src/helpers.js`.
I only extracted one helper, as an example and to check that everything works. The compiled output looks like this:
```js
helpers.jsx = helper("7.0.0-beta.0")("var REACT_ELEMENT_TYPE;\n\nexport ... ");
```

cc @Kovensky, since you had the original idea (accordingly to what I wrote in https://github.com/babel/babel/pull/7976/commits/bfb359263376d02507e91019d7aff32cae31611e) :smile:


.eslintrc.json
```diff
@@ -28,6 +28,12 @@
       "env": {
         "jest": true
       }
+    },
+    {
+      "files": ["packages/babel-helpers/src/helpers/*.js"],
+      "rules": {
+        "no-var": "off"
```
nicolo-ribaudo: Here probably we will need to add some other rules.

packages/babel-plugin-transform-react-constant-elements/test/fixtures/constant-elements/inline-elements/output.js
```diff
@@ -1,6 +1,6 @@
...
+function _jsx(type, props, key, children) { if (!REACT_ELEMENT_TYPE) { REACT_ELEMENT_TYPE = typeof Symbol === "function" && Symbol.for && Symbol.for("react.element") || 0xeac7; } var defaultProps = type && type.defaultProps; var childrenLength = arguments.length - 3; if (!props && childrenLength !== 0) { props = { children: void 0 }; } if (props && defaultProps) { for (var propName in defaultProps) { if (props[propName] === void 0) { props[propName] = defaultProps[propName]; } } } else if (!props) { props = defaultProps || {}; } if (childrenLength === 1) { props.children = children; } else if (childrenLength > 1) { var childArray = new Array(childrenLength); for (var i = 0; i < childrenLength; i++) { childArray[i] = arguments[i + 3]; } props.children = childArray; } return { $$typeof: REACT_ELEMENT_TYPE, type: type, key: key === undefined ? null : "" + key, ref: null, props: props, _owner: null }; }
```
nicolo-ribaudo: The only change here is that prettier transformed `''` to `""`.

nicolo-ribaudo:
I have some ideas for this, but I'll create a RFC because I don't think that this PR is the best solution.

PR #8501, opened 2018-08-21
[_wrapNativeSuper] Don't wrap non-native constructors

### LINKED ISSUE #8496: `this.constructor` incorrect inside wrapped native constructors.
## Bug Report

**Current Behavior**
`this.constructor` is incorrect inside wrapped native constructors on browsers that don't support `Reflect.construct`.

**Input Code**
https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=PTAEAUCcFMBdoHYBNSwBYENagAYCVoAzAG2gGNYA6MgewQGdZIBXCnUAS3oQHJt7mAB0E1I8JJQBQAdw7Ia0ygRLlsAXlDNkROdCQBuSTLlIFlABIAVALIAZAKKkAtonWhCWihzoAKAJSgAN6SoKAgoObQMAA0uLLyilZ2jtAuCLDsXKhoWR4IXnSxCDTY6NCgohwA5nIYxBE2Ds6uIWFgtAxMrLCilLjoXNR0jCwUouz0aDTMxCgARuUD9O6esN4IqDQ0sXPM2Fy8VK0d9DSklMQ0VT5LQ52jPZCgai-g8aaJjSlp2AD8oAAiQgYDjEAGgABcgMEGHo9ABfkMAF9DJIyMxGDQnN9XPRKEgdAhoD4eE4AJ4AWmgzXSPFiZGIsOW0AAHvBkMskk1Uq4gkjEZJTOiftQYFhoDj0iTyVSabAeIigA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&sourceType=module&lineWrap=true&presets=es2015%2Ces2016%2Ces2017&prettier=false&targets=&version=7.0.0-rc.1&envVersion=

**Expected behavior/code**
`this.constructor` should refer to the original constructor for the object.

Whatever is used by the REPL.

**Environment**
- Babel version(s): v7.0.0-rc.1

**Additional context**
This seems to be a result of the replacement used for `Reflect.construct`: https://github.com/babel/babel/blob/f8b0a5579f9765dc959658759f895e4e01820189/packages/babel-helpers/src/helpers.js#L486-L493
This bug prevents our [custom elements polyfill](https://github.com/webcomponents/custom-elements) from knowing the original constructor being called, which prevents it from looking up the definition of the element and upgrading it properly.

cc @justinfagnani

nicolo-ribaudo:
| Fixed Issues?            | Fixes #8496 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

`_wrapNativeSuper` doesn't set the correct prototype when calling the super constructor if `Reflect.construct` isn't available. This PR doesn't actually fix that (I don't think that it is possible), but makes it non-observable by not wrapping non-native constructors.


TimvdLippe:
@bicknellr could you verify this PR fixes the issue with our custom elements polyfill?

web-padawan:
What steps we do need to take in order to get this merged and shipped with the upcoming stable release? 

I will test the fix with my Webpack repo I have, and confirm if it works.

hzoo:
verifying it fixes the issue is a good step (and review of approach)!

web-padawan:
Verified manually, the fix worked for me in IE11, getting the same output as in Chrome with ES5 transpilation (where `Reflect.construct` is used):

screen shot 2018-08-23 at 15 37 13

jridgewell:
This also prevents us from extending lib provided native classes. I would add a `|| toString.startsWith('class')`, check.

nicolo-ribaudo:
We never supported extending `class` classes. I created https://github.com/babel/babel/pull/7081 a while ago but never finished it.

Kovensky:
IIRC the problem is that it will break if `Reflect.construct` is polyfilled, which core-js@2 does, but it only marks it as polyfilled in core-js@3, which babel doesn't use yet.

nicolo-ribaudo:
I will ensure that this work with es6 classes alongside with #7081 (I'll try to get that PR in v 7.0.x)

web-padawan:
@nicolo-ribaudo updated to `7.0.0-rc.3` and for some reason I'm now getting the error:

```
Uncaught TypeError: r.isNativeFunction is not a function
```

Is something wrong with the helpers change?

nicolo-ribaudo:
Can you delete `node_modules` and reinstall it? Probably you have `@babel/helpers@7.0.0-rc.2` downoaded.

web-padawan:
Figured it out, sorry for false alarm. Had to update my code to include this helper in the build.

Confirmed working with web-padawan/polymer3-webpack-starter@/4775c2d

loganfsmyth:
@web-padawan Hmm, if it's alright I'm gonna look over that commit. If things broke because of this PR, it indicates a bigger issue with your build process. Adding a new helper like this should not break anything, and we'll likely add more over the course of 7.x that could break the same way if this one broke.

loganfsmyth:
@web-padawan Oh or is the overall issue that you'd updated to `rc.3` for everything, but forgotten to update the whitelist? I guess that makes sense then, as long as it only happened.

I do think I'll add something to address this though. Really the `whitelist` you have should also be passed into `@babel/plugin-external-helpers` so it knows to not insert references to helpers that aren't in the whitelist. Let me add an option to the plugin for that.

web-padawan:
Sorry about that, it was my bad as I forget to update the list of the helpers I'm bundling 😕

loganfsmyth:
No problem, that's an understandable edge case. I've posted a PR in https://github.com/babel/babel/pull/8531 that allows you to pass the whitelist to `plugin-external-helpers` too, so things can stay synchronized. Then in your project you can do https://github.com/web-padawan/polymer3-webpack-starter/compare/master...loganfsmyth:helper-whitelist-example in order to share the whitelist between the Babel configuration and the helper generation script.

web-padawan:
Thanks for clarifying and for the help, I appreciate it! This option would be quite reasonable.

PR #8509, opened 2018-08-22
Add browserslist 4 support.

### LINKED ISSUE #8228: Use Browserslist 4
## Feature Request

Browserslist 4 with Node.js versions support was released

1. Use Browserslist 4
2. Use Browserslist to load Node.js versions as well
3. If Babel config has `target`, do not use Browserslist config (even if `target` contains only Node.js)

We already discussed it with @yavorsky and few people in Twitter and GitHub

We need to change `babel-preset-env` docs as well

yavorsky:
| Fixed Issues?            | Fixes #8228
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  | browserslsit > 4

After discussion with @ai and @existentialism and update browserslist to 4 we had an idea to possibly have `targets` as a query.
For now, browserslist 4 supports `node`. So in most cases, we don't need to use a plain object to specify targets directly. It's mostly confusing to have 2 sources of truth and merge this targets. For many people, it works like magic under the hood and it's not everytime clear why `chrome: 55, browsers: 'chrome 40'` will select `chrome 55`.
With browserslist 4 we just can specify every target we need using the query: `targets: "node 8.1, last 2 chrome versions, maintained node versions"`.
Most of the people are using targets as a plain object just to support `node` version. This PR contains `node current` and `current node` versions which are being replaced to `node ${currentNode}` query.

So what was changed?
1. updated `browserslist` to 4 and added node.js mappings.
2. Allow targets to be as a query for browserslsit. (It's supposed that people won't use the object version in future and someday we could remove this functionality. Should be less confusing for users).
3. Don't search for `browserslistconfig` if `targets` field was specified. (Since browserslist support node) (breaking change).
4. support `current node` and `node current` queries, which is replaced with `node ${CURRENT_VERSION}`.

Would be good to merge it before babel 7, since it contains small breaking change (point 3).

cc @ai @existentialism


nicolo-ribaudo:
In case we don't merge this before Friday, would it be possible to remove (or workaround) the breaking change?

ai:
@nicolo-ribaudo it is possible, but will be bad for users. Merging behavior is what is expected (and how other tools do it).

ai:
I added `current node` query support to Browserslist `master` https://github.com/browserslist/browserslist/commit/40b4816badfc476d84375021ab6d0630585cf845

Let’s remove `node current`, since aliases are nott common in Browserslist and we try to keep queries more English-ish.

yavorsky:
Sounds awesome @ai

packages/babel-preset-env/test/fixtures/preset-options-add-used-built-ins/source-type-script/options.json
```diff
@@ -2,9 +2,7 @@
   "presets": [
     ["../../../../lib", {
       "modules": false,
-      "targets": {
-        "node": "4.0.0"
-      },
+      "targets": "node 4.0.0",
```
existentialism: Instead of modifying existing tests, should we just add new ones to make sure both syntaxes work (for now)?

ai:
Browserlist 4.1 was released with `current node` query support for Babel

chicoxyzzy:
Just my 2 cents: it will be breaking change for babel-preset-env, not Babel if we'll land it after Babel 7

chicoxyzzy:
Actually it’s no so important to keep preset version in sync with Babel. More then that, this causes some issues. Also for example babel-loader and some other packages are not in sync with Babel version

TrySound:
All packages in babel monorepo has the same version. babel-loader is part of it.
https://github.com/babel/babel/tree/master/packages

existentialism:
@ai @TrySound bumped to 4.1!

ai:
Do we need something else to merge it?

Or just time for final decision

PR #8514, opened 2018-08-22
fix: object rest with default values bug

### LINKED ISSUE #8323: [Babel 7][plugin-proposal-object-rest-spread] TypeError: Property value of ObjectProperty expected node to be of a type
## Bug Report

**Current Behavior**
With `7.0.0-beta.53`, setting `loose: true` for `@babel/plugin-proposal-object-rest-spread` and compiling some ES2018 Rest Properties code will produce error: 

`TypeError: Property value of ObjectProperty expected node to be of a type ["Expression","PatternLike"] but instead got null`

This does not happen with `7.0.0-beta.51`.

**Input Code**

```js
// test-babel.js

const f = ({ a, b = 0, ...z }) => {} // will produce error
// const f = ({ a, ...z }) => {} // will pass
```
**Babel Configuration**

```js
// babel.config.js

module.exports = {
  presets: [ [ '@babel/env', { targets: { node: '8' } } ] ],
  plugins: [ [ '@babel/plugin-proposal-object-rest-spread', { loose: true } ] ]
}
```

```json
// package.json

{
  "dependencies": {
    "@babel/cli": "7.0.0-beta.53",
    "@babel/core": "7.0.0-beta.53",
    "@babel/preset-env": "7.0.0-beta.53"
  }
}
```

**The Error**

```
> npx babel test-babel.js -o test-babel-o.js
TypeError: Property value of ObjectProperty expected node to be of a type ["Expression","PatternLike"] but instead got null
    at Object.validate (PATH\node_modules\@babel\types\lib\definitions\utils.js:128:13)
    at Object.validate (PATH\node_modules\@babel\types\lib\validators\validate.js:17:9)
    at NodePath._replaceWith (PATH\node_modules\@babel\traverse\lib\path\replacement.js:194:9)
    at NodePath._remove (PATH\node_modules\@babel\traverse\lib\path\removal.js:51:10)
    at NodePath.remove (PATH\node_modules\@babel\traverse\lib\path\removal.js:30:8)
    at Object.keys.forEach.bindingName (PATH\node_modules\@babel\plugin-proposal-object-rest-spread\lib\index.js:136:40)
    at removeUnusedExcludedKeys (PATH\node_modules\@babel\plugin-proposal-object-rest-spread\lib\index.js:131:27)
    at VariableDeclarator.path (PATH\node_modules\@babel\plugin-proposal-object-rest-spread\lib\index.js:247:13)
    at NodePath._call (PATH\node_modules\@babel\traverse\lib\path\context.js:53:20)
```

**Environment**
- Babel version(s): 7.0.0-beta.53
- Node/npm version: Node 10.5/npm 6.1
- OS: Windows 10

jquense:
| Fixed Issues?            | fixes #8323
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-proposal-object-rest-spread/src/index.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/src/index.js
@@ -96,10 +96,14 @@ export default declare((api, opts) => {
     const bindings = path.getOuterBindingIdentifierPaths();
 
     Object.keys(bindings).forEach(bindingName => {
-      if (path.scope.getBinding(bindingName).references > 1) {
+      const binding = bindings[bindingName].parentPath;
+      if (
+        path.scope.getBinding(bindingName).references > 1 ||
+        !binding.isObjectProperty()
+      ) {
         return;
       }
-      bindings[bindingName].parentPath.remove();
+      binding.remove();
     });
   }
 
--- a/packages/babel-plugin-proposal-object-rest-spread/src/index.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/src/index.js
@@ -96,14 +96,14 @@ export default declare((api, opts) => {
     const bindings = path.getOuterBindingIdentifierPaths();
 
     Object.keys(bindings).forEach(bindingName => {
-      const binding = bindings[bindingName].parentPath;
+      const bindingParentPath = bindings[bindingName].parentPath;
       if (
         path.scope.getBinding(bindingName).references > 1 ||
-        !binding.isObjectProperty()
+        !bindingParentPath.isObjectProperty()
       ) {
         return;
       }
-      binding.remove();
+      bindingParentPath.remove();
     });
   }
 
--- a/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/regression/gh-8323/input.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/regression/gh-8323/input.js
@@ -1,0 +1,8 @@
+const get = () => {
+  fireTheMissiles();
+  return 3;
+};
+
+const f = ({ a = get(), b, c, ...z }) => {
+  const v = b + 3;
+};
--- a/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/regression/gh-8323/options.json
+++ b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/regression/gh-8323/options.json
@@ -1,0 +1,4 @@
+{
+  "presets": [ [ "env", { "targets": { "node": "8" } } ] ],
+  "plugins": [ [ "proposal-object-rest-spread", { "loose": true } ] ]
+}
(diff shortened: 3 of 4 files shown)


jquense:
i'm not sure what the circle failure is about...tests pass locally :/

Andarist:
It doesn't seem like you have committed changed source files.

jquense:
>It doesn't seem like you have committed changed source files.

omg haha

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/regression/gh-8323/input.js
```diff
@@ -0,0 +1,3 @@
+const f = ({ a, b = 0, ...z }) => {
```
Andarist: I'm wondering about "reversed" case of:
```js
const f = ({ a = 0, b, ...z }) => {
  const v = b + 3;
}
```

Judging based on the code it will remove `a` and I'm not sure if it should. The whole point of this optimization was to output terser code for spread as "omit" case and removing properties which have explicit defaults doesnt seem falling into that use case - my expectancy would be that either it shouldnt have a default value if it was used merely as part of the omit operation OR that it shouldnt be there at all if the identifier is intentionally unused.

WDYT?

Andarist: > you'd have a much better sense of what the right thing to do here is.

Why would that be? 😉 I'm just thinking out loud to get a second opinion. What I'm trying to say - I'm not any better than you 😃 

> I think what you're saying is that if someone went through the trouble of assigning a default value we probably shouldn't remove it from the output?

Yeah.

> presumably if you aren't using it the defualt value was unnecessary.

I just kinda feel that this maybe is a linter's job to warn against such cases

> fireTheMissiles

I'm not that much concern about this case, because the optimization happens only in loose mode anyway & can also "break" other code such as:
```js
const f = ({ a, b = 0, ...z }) => {
  const v = b + 3;
};

f({
  get a() {
    fireTheMissiles()
    return 3
  }
})
```

although the breakage possibility is much higher with functions as default parameters than with getters.

Personally I would leave defaults out of this optimization but I'm not overly attached to it.

jquense:
updated, this is safer and the logic is simpler anyway, so lets go with this?

packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -96,10 +96,14 @@ export default declare((api, opts) => {
     const bindings = path.getOuterBindingIdentifierPaths();
 
     Object.keys(bindings).forEach(bindingName => {
-      if (path.scope.getBinding(bindingName).references > 1) {
+      const binding = bindings[bindingName].parentPath;
```
Andarist: nit pick, this should be named smth like `bindingParentPath`

Andarist:
This is a really small PR & a fix to that so gonna go ahead and merge it.

PR #8524, opened 2018-08-24
Add LICENSE file to published npm packages (backport for v6)

honzajavorek:
Backporting https://github.com/babel/babel/pull/8409 (a7cc3325cdf4fc5f1c4436171b20ba34fdfad622) to v6, because v6 is still in wide use (`create-react-app`, etc.).

_I did not fill the table as all information is in https://github.com/babel/babel/pull/8409 already._


xtuc:
Now that Babel 7 has bee released is this still needed?

loganfsmyth:
It's fine, it may require extra work to actually get new packages published with the new LICENSE though, so we'll see if they actually get published.

honzajavorek:
> Now that Babel 7 has bee released is this still needed?

Unfortunately it is. This legalizes the past packages everyone is using everywhere, at least for those without strictly pinned versions. Also, we made some preliminary research on how much work would be to upgrade to 7 now and it's just not trivial. For a small team it is a significant effort, which needs time. It would be totally awesome if you could figure out how to release this now.

And you already are totally awesome for merging the PR, thank you 👍

opichals:
@loganfsmyth In https://github.com/babel/babel/pull/8644 I have put together a patch to the 6.x version of the `Makefile` which could be used to make lerna publish all the packages affected by this PR.

Could you please have a look?

PR #8531, opened 2018-08-24
Add a 'whitelist' option for the external-helpers plugin to mirror the helper builder.

loganfsmyth:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The external helper generator in https://github.com/babel/babel/blob/cada040becacb9a1c378763fb2b6e65185fa6d31/packages/babel-core/src/tools/build-external-helpers.js#L146 accepts a whitelist, but the `@babel/plugin-external-helpers` will still insert references to plugins that aren't in the whitelist.

This PR adds a `whitelist` to the plugin as well, so the plugin will fall back to babel-standard behavior of injecting the helper if it isn't included in the whitelist. That way at worse extra code gets injected, instead of having hard-to-diagnose errors about missing helpers.

### DIFF
--- a/packages/babel-plugin-external-helpers/src/index.js
+++ b/packages/babel-plugin-external-helpers/src/index.js
@@ -4,7 +4,18 @@ import { types as t } from "@babel/core";
 export default declare((api, options) => {
   api.assertVersion(7);
 
-  const { helperVersion = "7.0.0-beta.0" } = options;
+  const { helperVersion = "7.0.0-beta.0", whitelist = false } = options;
+
+  if (
+    whitelist !== false &&
+    (!Array.isArray(whitelist) || whitelist.some(w => typeof w !== "string"))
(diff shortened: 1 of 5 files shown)


Andarist:
While the change itself is OK and probably useful, I wouldn't say that it quite solves the issue that it is supposed to solve - the reason is that people won't really know that there is this new option available & that they should synchronize whitelist for both build-external-helpers and external-helpers plugin.

It seems to me that `external-helpers` are quite confusing at the moment anyway (this will be a little bit off topic) - personally I would expect it (and I've seen people expecting it too) to just insert imports to the helpers (what transform-runtime does). Based on what it does it would be better to rename it to `babel-plugin-global-helpers`.

PR #8534, opened 2018-08-25
Fix version badges in monorepo README

Simek:
Badges caching disabled by removing the maxAge parameter from image URLs to ensure that always the latest version of package is displayed.


nicolo-ribaudo:
Maybe we could set it to one day (86400 seconds)?

loganfsmyth:
> Badges caching disabled by removing the maxAge parameter

Is that true? https://shields.io/ makes it seem like it will default to `120` seconds when left off.

Given that `2592000` is in seconds, that does seem extremely high, but 120 seems unnecessarily fast. Maybe just say an hour or something (`3600`)?

Simek:
I have missed the default value of cache time on Shields.io, my bad, sorry.

One hour seems reasonable for me, and sounds better than default `120`s.

PR has been updated, but if you finally decide to extend cache to one day just let me know here.

loganfsmyth:
Thanks!

PR #8535, opened 2018-08-25
 Do not unpack array patterns that update a referenced binding

### LINKED ISSUE #8528: Breaks the swapping of variable values using array destructuring.
## Bug Report

**Current Behavior**
Babel breaks the swapping of two variable values using array destructuring. This only seems to happen if the two values to be swapped are also function arguments.

**Input Code**
Correct Behaviour: [REPL version 7.0.0-beta.39](https://babeljs.io/en/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABDAzgIQKZQO4Y2ACgA8AaRAQzICMBKRAbwChFlhEDzEA-RWxAbUq8AuogC8AqmXLCA3M0QAnLCEVIi3CogBk2xBoA8veQF9GQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.0.0-beta.39&envVersion=7.0.0-beta.39)

Incorrect Behaviour: [REPL version 7.0.0-rc.2](https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABDAzgIQKZQO4Y2ACgA8AaRAQzICMBKRAbwChFlhEDzEA-RWxAbUq8AuogC8AqmXLCA3M0QAnLCEVIi3CogBk2xBoA8veQF8gA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.0.0-rc.2&envVersion=7.0.0-rc.2)

```js
function isBetween(x, a, b) {
  if (a > b) [a, b] = [b, a];
  return x > a && x < b;
}
```

```js
{
  presets: [['@babel/preset-env', {
    targets: {browsers: ['> 1%', 'not ie <= 11', 'not ios < 10']},
    modules: false,
    useBuiltIns: false,
    loose: true
  }]]
};
```

**Environment**
- Babel version: v7.0.0-rc.2

nicolo-ribaudo:
| Fixed Issues?            | Fixes #8528 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

> Do not unpack array patterns that update a referenced binding
>
> We used this.scope.traverse to check if the array initializer referenced
> a binding which is present in the array pattern. It doesn't work because,
> in order to check if an identifier is a reference, we need to check its
> parent.
> this.scope.binding can't set the correct `parentPath`, because the only
> path that it knows is the root scope's (e.g. the function path).
> Use `t.traverse` which gives the correct parent.

This regression has been brought to the surface by 29d249e3b700e9bff9d3b6e47f7a8ba9c69612cc because it made the `isReferenced` check stricter, but it is unrelated.

### DIFF
--- a/packages/babel-plugin-transform-destructuring/src/index.js
+++ b/packages/babel-plugin-transform-destructuring/src/index.js
@@ -44,13 +44,23 @@ export default declare((api, options) => {
     return false;
   }
 
-  const arrayUnpackVisitor = {
-    ReferencedIdentifier(path, state) {
-      if (state.bindings[path.node.name]) {
-        state.deopt = true;
-        path.stop();
-      }
-    },
+  const STOP_TRAVERSAL = {};
+
+  // NOTE: This visitor is meant to be used via t.traverse
+  const arrayUnpackVisitor = (node, ancestors, state) => {
+    if (!ancestors.length) {
+      // Top-level node: this is the array literal.
+      return;
+    }
+
+    if (
+      t.isIdentifier(node) &&
+      t.isReferenced(node, ancestors[ancestors.length - 1]) &&
+      state.bindings[node.name]
+    ) {
+      state.deopt = true;
+      throw STOP_TRAVERSAL;
+    }
   };
 
   class DestructuringTransformer {
@@ -282,7 +292,13 @@ export default declare((api, options) => {
       // deopt on reference to left side identifiers
       const bindings = t.getBindingIdentifiers(pattern);
       const state = { deopt: false, bindings };
-      this.scope.traverse(arr, arrayUnpackVisitor, state);
+
+      try {
+        t.traverse(arr, arrayUnpackVisitor, state);
+      } catch (e) {
+        if (e !== STOP_TRAVERSAL) throw e;
+      }
+
       return !state.deopt;
     }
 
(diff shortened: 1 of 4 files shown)


packages/babel-plugin-transform-destructuring/src/index.js
```diff
@@ -44,13 +44,16 @@ export default declare((api, options) => {
...
-    },
+  // NOTE: This visitor is meant to be used via t.traverseFast
+  const arrayUnpackVisitor = (node, state, parent) => {
+    if (
+      t.isIdentifier(node) &&
+      t.isReferenced(node, parent) &&
+      state.bindings[node.name]
+    ) {
+      state.deopt = true;
+      return true;
```
loganfsmyth: I'm not a huge fan of `return true` in traversal-like APIs because it's not clear if the behavior is `.skip()`-like or `.stop()`-like.

My preference would probably be for it to either:
* Explicitly return a STOP constant
* Skip the return value and throw an exception instead, e.g.
```
const STOP = {};
// ...
throw STOP;
```
and do
```
try {
  t.traverseFast(arr, arrayUnpackVisitor, state);
} catch (err) {
  if (err !== STOP) throw err;
}
```

JSteunou:
Any step missing for this PR to be accepted?

PR #8536, opened 2018-08-25
Fixed install commands

ffxsam:
The official docs say:

>Because this is a polyfill (which will run before your source code), we need it to be a dependency, not a devDependency

| Documentation PR Link    | 


hzoo:
Ah the issue here is that the readme here is generated, so we need to update the script (annoying haha).

Since the next time we run this it will be overwritten again, https://github.com/babel/babel/blob/1ac46a256d5e78ff83e150f46cf05acfb9489967/scripts/generators/readmes.js#L57-L63

nicolo-ribaudo:
This has been fixed in the generator script (56044c7851d583d498f919e9546caddf8f80a72f)

PR #8539, opened 2018-08-25
Update parser whitespace for clarity

jridgewell:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Tests Added + Pass?      |
| Documentation PR Link    |
| Any Dependency Changes?  |

This removes 0x180E MONGOLIAN VOWEL SEPARATOR from valid whitespace characters. Chance of this actually breaking code is minimal. Per https://github.com/tc39/ecma262/pull/1218#issuecomment-395340891, ECMA262 follows whatever the latest Unicode version specifies for Zs Space_Separator category. MONGOLIAN VOWEL SEPARATOR was moved to the Cf Other_Format category in Unicode 8.

Has the nice benefit of not requiring a charCode to String conversion.

### DIFF
--- a/packages/babel-parser/src/util/whitespace.js
+++ b/packages/babel-parser/src/util/whitespace.js
@@ -1,13 +1,53 @@
 // @flow
 
+import * as charCodes from "charcodes";
+
 // Matches a whole line break (where CRLF is considered a single
 // line break). Used to count lines.
-
 export const lineBreak = /\r\n?|\n|\u2028|\u2029/;
 export const lineBreakG = new RegExp(lineBreak.source, "g");
 
+// https://tc39.github.io/ecma262/#sec-line-terminators
 export function isNewLine(code: number): boolean {
-  return code === 10 || code === 13 || code === 0x2028 || code === 0x2029;
+  switch (code) {
+    case charCodes.lineFeed:
+    case charCodes.carriageReturn:
+    case charCodes.lineSeparator:
+    case charCodes.paragraphSeparator:
+      return true;
+
+    default:
+      return false;
+  }
 }
 
-export const nonASCIIwhitespace = /[\u1680\u180e\u2000-\u200a\u202f\u205f\u3000\ufeff]/;
+// https://tc39.github.io/ecma262/#sec-white-space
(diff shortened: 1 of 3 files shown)


packages/babel-parser/src/util/whitespace.js
```diff
@@ -1,13 +1,53 @@
...
 // line break). Used to count lines.
-
 export const lineBreak = /\r\n?|\n|\u2028|\u2029/;
 export const lineBreakG = new RegExp(lineBreak.source, "g");
 
+// https://tc39.github.io/ecma262/#sec-line-terminators
 export function isNewLine(code: number): boolean {
-  return code === 10 || code === 13 || code === 0x2028 || code === 0x2029;
+  switch (code) {
+    case charCodes.lineFeed:
```
loganfsmyth: Out of curiosity, does using properties like this cause V8 to optimize this less? In theory it'd now have to access the property ever time `isNewLine` is called.

jridgewell: It does, but we use https://github.com/xtuc/charcodes/tree/master/packages/babel-plugin-transform-charcodes to inline the code.

loganfsmyth: Ahhhh perfect.

packages/babel-parser/src/util/whitespace.js
```diff
@@ -1,13 +1,53 @@
...
+    case 0x2005: // FOUR-PER-EM SPACE
+    case 0x2006: // SIX-PER-EM SPACE
+    case 0x2007: // FIGURE SPACE
+    case 0x2008: // PUNCTUATION SPACE
+    case 0x2009: // THIN SPACE
+    case 0x200a: // HAIR SPACE
+    case 0x202f: // NARROW NO-BREAK SPACE
+    case 0x205f: // MEDIUM MATHEMATICAL SPACE
+    case 0x3000: // IDEOGRAPHIC SPACE
+    case 0xfeff: // ZERO WIDTH NO-BREAK SPACE
```
xtuc: Should we add those in `xtuc/charcodes` for clarity and consistency?

PR #5195, opened 2017-01-23
Don't transpile ES7 symbol properties

taion:
| Q                        | A 
| ------------------------ | ---
| Patch: Bug Fix?          | Yes
| Major: Breaking Change?  | 
| Minor: New Feature?      | 
| Deprecations?            | 
| Spec Compliancy?         | 
| Tests Added/Pass?        | 
| Fixed Tickets            | Fixes #4783
| License                  | CC0
| Doc PR                   | 
| Dependency Changes       | 

Pending https://github.com/zloirock/core-js/pull/267, this works around the root cause of #4783 in https://github.com/zloirock/core-js/issues/262 by pulling in the full `Symbol` polyfill instead of just the `Symbol.asyncIterator` polyfill.

Generally in the context of using the runtime transform, it'd be almost impossible to avoid pulling in the `Symbol` polyfill when using async iterators anyway, so in practice this is unlikely to make things worse.

More to the point, as the `asyncIterator` helper and the core-js imports are set up, async iteration won't work without pulling in the full `Symbol` polyfill anyway.

It's not clear to me how to add a test case for this, or whether it'd be necessary to do so here.

Issue #4783, opened 2016-10-25
building async iterators with transform-runtime produces wrong build, but babel-node works properly


darsain:
Isolated issue here: https://gist.github.com/anonymous/dcbc9f57057724ee5ffe0650eedc9472

```
git clone https://gist.github.com/dcbc9f57057724ee5ffe0650eedc9472.git example
cd example
npm install
npm run execute
npm run build
```
### Input Code (`index.js`)

``` js
async function * values(iterable) {
    for (const item of iterable) {
        yield item;
    }
}

async function collect(asyncIterator) {
    const collection = [];
    for await (const value of asyncIterator) {
        collection.push(value);
    }
    return collection;
}

collect(values([5, 4, 3, 2, 1])).then(console.log, console.error);
```
### Babel Configuration (`package.json`)

``` js
{
  "scripts": {
    "build": "babel index.js --out-file build.js && node build.js",
    "execute": "babel-node index.js"
  },
  "babel": {
    "plugins": [
      ["transform-runtime", {"polyfill": false, "regenerator": true}]
    ],
    "presets": ["stage-3", "latest"]
  },
  "dependencies": {
    "babel-runtime": "^6.18.0"
  },
  "devDependencies": {
    "babel-cli": "^6.18.0",
    "babel-plugin-transform-runtime": "^6.15.0",
    "babel-preset-latest": "^6.16.0",
    "babel-preset-stage-3": "^6.17.0"
  }
}
```
### Expected Behavior

Both scripts in `package.json:scripts` should log the array `[5, 4, 3, 2, 1]`.
### Current Behavior

The `build` script that relies on `babel` command produces a build file that throws:

```
TypeError: Object is not async iterable
```

On the other hand, the `execute` script relying on `babel-node` works properly.
### Possible Solution

The fact that `babel-node` works might be a clue.

Also @hzoo on slack mentioned it might be related to #4768.
### Your Environment

Module versions are specified in `package.json` above.

| software | version |
| --- | --- |
| node | 6.7.0 |
| npm | 3.10.9 |
| Operating System | Win 10 |

hzoo:
In https://github.com/babel/babel/blob/ec496480381d5c3ce9f14e5df186f2823c50c9b3/packages/babel-helpers/src/helpers.js#L66-L75

`if (typeof Symbol === "function") {` gets `"object"` instead of function

taion:
This is related to https://github.com/babel/babel/issues/4418.

taion:
[...]

PR #8547, opened 2018-08-26
Remove nonstandard functions and fake prototype methods from babel-runtime

loganfsmyth:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Y, if relying on proposals or fake-proto methods, hopefully relatively minimal
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Just like we removed the nonstandard stuff from the polyfill, I'm also removing it from the runtime too.

The runtime also allowed stuff like `Array.fill()` as a fake replacement for `Array.prototype.fill` which is bad form mostly. We can work on improving support for prototype-based runtime functions, but we shouldn't encourage of entirely made-up function placement, if we can avoid it.

Lastly, this adds support for `Symbol.asyncIterator` which I realized was missing.

### DIFF
--- a/packages/babel-plugin-transform-runtime/src/definitions.js
+++ b/packages/babel-plugin-transform-runtime/src/definitions.js
@@ -10,8 +10,8 @@ export default {
     setImmediate: "set-immediate",
     clearImmediate: "clear-immediate",
     asap: "asap",
-    //parseFloat: "parse-float", // temporary disabled
-    //parseInt: "parse-int" // temporary disabled
+    parseFloat: "parse-float",
+    parseInt: "parse-int",
   },
 
   methods: {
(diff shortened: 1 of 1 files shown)


packages/babel-plugin-transform-runtime/src/definitions.js
```diff
@@ -68,31 +46,8 @@ export default {
...
-      tanh: "math/tanh",
-      trunc: "math/trunc",
-      iaddh: "math/iaddh",
-      isubh: "math/isubh",
-      imulh: "math/imulh",
-      umulh: "math/umulh",
-    },
-
     Symbol: {
+      asyncIterator: "symbol/async-iterator",
```
taion: This may have re-broken https://github.com/babel/babel/issues/4783 by rolling back https://github.com/babel/babel/pull/5195.

(Yes, I'm very late on upgrading to Babel 7 in a few places...)

PR #8552, opened 2018-08-27
preset-env: use Chromium versions for Chromium-based browsers

### LINKED ISSUE #6602: Add support for Samsung Internet browser

ylemkimon:
| Fixed Issues?            | Fixes #6602
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 

`@babel/preset-env` only uses `Android`, `Chrome`, `Edge`, `Firefox`, `IE`, `iOS`, `Safari`, `Node` versions for detecting which transforms to include:

https://github.com/babel/babel/blob/e875fc7bc11ffdb8d5176028ae2b336bf7455c6c/packages/babel-preset-env/src/targets-parser.js#L32-L42

and **silently** ignores other browsers:

https://github.com/babel/babel/blob/e875fc7bc11ffdb8d5176028ae2b336bf7455c6c/packages/babel-preset-env/src/targets-parser.js#L72-L76

However, there are local-popular or manufacturer-specific browsers such as Opera Mini, UC browser, QQ browser, Baidu browser, and Samsung Internet. Please see https://github.com/browserslist/browserslist/issues/250 for their importance.

Currently, fortunately, IE 11, Chrome 49 (usage 0.69%) and Android 4.x (4.4 usage 0.58%, 4.4.4 isn't strictly last 2 versions but is in Browserslist default) acts as a stopgap preventing from excluding transform needed for old browsers. However, if user excludes them, they become `dead`, or their usage drops, `preset-env` will recognize as if it targets very modern browsers. Also, user may **explicitly** target them, using Browserslist config.

For Chromium-based browsers, we can use their base Chromium versions to detect features, like [`electron-to-chromium`](https://github.com/Kilian/electron-to-chromium), as a best-effort basis. This PR adds a mapping for Samsung Internet and UC Browser for Android, which is popular in Samsung devices and China, respectively.

### DIFF
--- a/packages/babel-preset-env/data/chromium-browsers.js
+++ b/packages/babel-preset-env/data/chromium-browsers.js
@@ -1,0 +1,14 @@
+module.exports = {
+  // Samsung Internet release notes: https://developer.samsung.com/internet/android/releases
+  // Chrome version page (chrome://version) is accessible
+  // Indicated in user agent string as Chrome/XX
+  "samsung 4": 44,
+  "samsung 5": 51,
+  "samsung 6.2": 56,
+  "samsung 7.2": 59,
+  
+  // UC Browser developer edition: https://plus.ucweb.com/download/
+  // Indicated in user agent string as Chrome/XX
+  "and_uc 11.4": 40,
+  "and_uc 11.8": 57,
+};
--- a/packages/babel-preset-env/data/chromium-browsers.js
+++ b/packages/babel-preset-env/data/chromium-browsers.js
@@ -2,13 +2,13 @@ module.exports = {
   // Samsung Internet release notes: https://developer.samsung.com/internet/android/releases
   // Chrome version page (chrome://version) is accessible
   // Indicated in user agent string as Chrome/XX
-  "samsung 4": 44,
-  "samsung 5": 51,
-  "samsung 6.2": 56,
-  "samsung 7.2": 59,
+  "samsung 4": "44",
+  "samsung 5": "51",
+  "samsung 6.2": "56",
+  "samsung 7.2": "59",
   
   // UC Browser developer edition: https://plus.ucweb.com/download/
   // Indicated in user agent string as Chrome/XX
-  "and_uc 11.4": 40,
-  "and_uc 11.8": 57,
+  "and_uc 11.4": "40",
+  "and_uc 11.8": "57",
 };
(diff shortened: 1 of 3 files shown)


packages/babel-preset-env/data/chromium-browsers.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  // Samsung Internet release notes: https://developer.samsung.com/internet/android/releases
+  // Chrome version page (chrome://version) is accessible
+  // Indicated in user agent string as Chrome/XX
+  "samsung 4": "44",
+  "samsung 5": "51",
+  "samsung 6.2": "56",
+  "samsung 7.2": "59",
```
chicoxyzzy: also `"samsung 3": "38"`

ylemkimon: @chicoxyzzy `caniuse` added Samsung Internet starting with `samsung 4`: https://github.com/Fyrd/caniuse/commit/16d76e4e18dd670dd70b8865d63b8ff85931073e. Therefore `browserslist` never returns `samsung 3`.

zloirock:
Feel free to add a mapping for those browsers to `core-js-compat` and `compat-table` (Samsung Internet already added) and it will be added to `babel`.

PR #8616, opened 2018-09-03
Add static Math methods to plugin-transform-runtime

### LINKED ISSUE #8611: Add static Math methods to plugin-transform-runtime
The transform-runtime plugin does not yet polyfill static methods of the `Math` object, which are not yet supported everywhere, such as `Math.trunc`. I suggest that these are added.

Without transform-runtime covering these, it's necessary to either polyfill manually or to write your own transform plugin which basically does what transform-runtime already does, but at a more limited scope.
The first is a pain - and avoiding it one of the reasons you'd be using the transform-runtime plugin's polyfilling ability to begin with - and the second, while a one-time investment, just feels wrong.

Siddharth11:
| Fixed Issues?            |  Fixes #8611 
| Patch: Bug Fix?          |  Yes
| Major: Breaking Change?  |  No	
| Minor: New Feature?      |  No
| Tests Added + Pass?      |  Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |  Yes


packages/babel-plugin-transform-runtime/src/definitions.js
```diff
@@ -46,6 +46,30 @@ export default {
...
+      log1p: "math/log1p",
+      log2: "math/log2",
+      sign: "math/sign",
+      sinh: "math/sinh",
+      tanh: "math/tanh",
+      trunc: "math/trunc",
+      iaddh: "math/iaddh",
+      isubh: "math/isubh",
+      imulh: "math/imulh",
+      umulh: "math/umulh",
```
nicolo-ribaudo: Some of these (e.g. `umulh`) are not part of the standard (https://tc39.github.io/ecma262/)

Siddharth11:
@nico29 I've removed `iaddh`, `isubh`, `imulh`, and `umulh`

loganfsmyth:
I'll note that, unfortunately, adding this as-is constitutes a breaking change, because users might be using an old version of `@babel/runtime` that doesn't have the `core-js` redirect files that this relies on. We'll need to inject these only when user's set the [`"version"`](https://github.com/babel/babel/blob/4d556f70dc83f4a274544c9390834266853cd2e6/packages/babel-plugin-transform-runtime/src/index.js#L40) option for `transform-runtime` in order to avoid breakage.

Siddharth11:
> I'll note that, unfortunately, adding this as-is constitutes a breaking change, because users might be using an old version of @babel/runtime that doesn't have the core-js redirect files that this relies on. We'll need to inject these only when user's set the "version" option for transform-runtime in order to avoid breakage.

@loganfsmyth can you guide me how to conditionally inject it based on the version type?

loganfsmyth:
We'll have to add a dependency on `semver` and check

```
if (
  !semver.intersects(`<=7.0.0}`, version) &&
  !semver.intersects(`>=8.0.0`, version)
) {
  // only use the `Math` definitions if this condition passes

}

File: packages/babel-plugin-transform-runtime/src/definitions.js [hunk 2]
```diff
@@ -1,3 +1,12 @@
+import semver from "semver";
+import { version } from "@babel/core";
```
loganfsmyth: This isn't the correct version in this context. We'll need to use the `version` passed into the plugin itself in https://github.com/babel/babel/blob/53e4963836a89141845261029d332475668aa26b/packages/babel-plugin-transform-runtime/src/index.js#L40

packages/babel-plugin-transform-runtime/src/definitions.js
```diff
@@ -1,3 +1,12 @@
+import semver from "semver";
```
loganfsmyth: Don't forget to add this to `package.json`

packages/babel-plugin-transform-runtime/src/definitions.js
Siddharth11: Is `runtimeVersion` a global variable?

loganfsmyth: It is not, so either we'd have to do this check in `index.js` instead, or we'd have to change `export default` here to export a function that passes in the version.

Siddharth11:
@loganfsmyth commit updated

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -41,6 +42,16 @@ export default declare((api, options, dirname) => {
...
   } = options;
 
+  // reason for conditionally including this module:
+  // https://github.com/babel/babel/pull/8616#issuecomment-418154753
+  const includeMathModule =
+    !semver.intersects(`<=7.0.0`, runtimeVersion) &&
+    !semver.intersects(`>=8.0.0`, runtimeVersion);
+
+  if (!includeMathModule) {
+    delete definitions.Math;
```
nicolo-ribaudo: Instead of deleting `Math` when it is not supported, could we only define it when it is supported?. e.g.

```js
// definitions.js

export default runtimeVersion => ({
  builtins: { /* ... */ },
  methods: {
    /* ... */
    Math: semver.satisfies(runtimeVersion, "7.0.1 - 8.0.0") ? {
      acosh: "math/acosh",
      /* ... */
    } : {}
  }
});
```

Siddharth11: Yeah, we can do that but then `definitions` will be a function now? All its instances inside `index.js` will need to be invoked (line no. 219, 225, etc). Is that okay? Will there be a significant performance boost if we avoid `delete` operation?

Siddharth11: Okay, I'll make the change

loganfsmyth: One thing to note is, when we make it a function we also need to update https://github.com/babel/babel/blob/bda7cebdb4b5bf0f67022d9739708cb7905e79e1/packages/babel-plugin-transform-runtime/scripts/build-dist.js#L11 and it should just always include all definitions when called in that case.

Siddharth11: @loganfsmyth @nicolo-ribaudo both the issues are addressed

packages/babel-plugin-transform-runtime/src/definitions.js
```diff
@@ -1,104 +1,141 @@
...
-  },
+import semver from "semver";
 
-  methods: {
-    Array: {
-      from: "array/from",
-      isArray: "array/is-array",
-      of: "array/of",
-    },
+const getDefinitions = runtimeVersion => {
```
loganfsmyth: You can do `export default runtimeVersion => {` to avoid splitting up the export and the function.

packages/babel-plugin-transform-runtime/src/definitions.js
```diff
@@ -1,104 +1,141 @@
...
+  let includeMathModule = true;
+  if (runtimeVersion) {
+    // reason for conditionally including this module:
+    // https://github.com/babel/babel/pull/8616#issuecomment-418154753
+    includeMathModule = semver.satisfies(runtimeVersion, "7.0.1 - 8.0.0");
```
loganfsmyth: This logic isn't quite the same as what we had before. `runtimeVersion` can be a range, e.g. `^7.0.0`, and`satisfies` expects a specific version.

loganfsmyth:
Looking good, getting very close!

Siddharth11:
@loganfsmyth fixed

loganfsmyth:
Thanks!

loganfsmyth:
I'm landing this since it looks fine, but if you're interested, it might be good to also add a few tests.

loganfsmyth:
You'd want to copy the this folder: https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-runtime/test/fixtures/runtime-corejs2/symbol-iterator and make a new one called `math`, then edit the `input.js` to add examples of all of the `Math` functions. Then edit `options.json` to pass `"version": "^7.0.1"`, and you can delete `output.js`. When you re-run the tests, Babel will recreate `output.js` with the output that Babel currently creates. Given your fix, the output should include a bunch of `require` calls to load the math functions.

Also FYI, I fixed one small issue I realized this PR had: https://github.com/babel/babel/commit/fc91b94a86a42c8c8c28523ccc612cdeea49ad63

nicolo-ribaudo:
We should test both `version: ^7.0.1` and no version specified.

PR #8620, opened 2018-09-03
[static private] Use explicit descriptors instead of an object

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

> This is similar to ec69b4bb1256c061ac76f53dfed09c4283ec6a31 (#8318), which was about private instance fields.
>
> Private properties can be non-writable (thanks to decorators), or have
get/set accessors. If we stored this information on the `privateClass`
object, we would need to always use `Object.getOwnPropertyDescriptor`
before reading or writing a property because accessors need to be called
with the correct `this` context (it should be the actual class, not the
object hat stores the private properties). This commit simplifies that
operation a bit by removing the container object.
>
> It also have another advantage, which instance fields already have
thanks to the use of separate weakmaps: unused private static fields
can be tree-shaken away or garbage-collected, while properties of an
object can't. Also, they can be easilier minified.

This isn't strictly required, but it makes the transform a little bit simpler.

cc @rricard @jridgewell

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -1055,24 +1055,26 @@ helpers.classPrivateFieldSet = helper("7.0.0-beta.0")`
 `;
 
 helpers.classStaticPrivateFieldSpecGet = helper("7.0.1")`
-  export default function _classStaticPrivateFieldSpecGet(
-    receiver, classConstructor, privateClass, privateId
-  ) {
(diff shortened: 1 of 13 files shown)


nicolo-ribaudo:
I'm going to merge this so I can rebase #8130 to check that everything still works.

PR #8622, opened 2018-09-03
Make dir for babel --out-file

TrySound:
Currently there's unexpected regression after upgrade from babel 6.
On creating file with any depth like dist/index.js the error about
not existing directory is thrown.

In this diff I modified babel-cli to create deep directory for out-file
command.

I also replaced `mkdirp` with more supported `make-dir` package which
also have official promise support.

| Fixed Issues?            | 
| Patch: Bug Fix?          | 6 -> 7 regression fix
| Minor: New Feature?      | maybe
| Documentation PR Link    | 
| Any Dependency Changes?  | yes mkdirp -> make-dir

### DIFF
--- a/packages/babel-cli/package.json
+++ b/packages/babel-cli/package.json
@@ -24,7 +24,7 @@
     "fs-readdir-recursive": "^1.1.0",
     "glob": "^7.0.0",
     "lodash": "^4.17.13",
-    "mkdirp": "^0.5.1",
+    "make-dir": "^2.1.0",
     "output-file-sync": "^2.0.0",
     "slash": "^2.0.0",
     "source-map": "^0.5.0"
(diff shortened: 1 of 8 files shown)


yarn.lock
```diff
@@ -2,9 +2,9 @@
 # yarn lockfile v1
 
 
-"@babel/cli@7.0.0-rc.2":
-  version "7.0.0-rc.2"
-  resolved "https://registry.yarnpkg.com/@babel/cli/-/cli-7.0.0-rc.2.tgz#0c64762f27ccd3fa255111570b6c404f81181c85"
+"@babel/cli@^7.0.0":
```
xtuc: I'm not sure why Babel changed here? make-dir doesn't depends on Babel. Maybe we forgot to update the lock file after our own Babel bump.

PR #8626, opened 2018-09-04
Add tests for static math methods

Siddharth11:
@nicolo-ribaudo @loganfsmyth 

| Tests Added + Pass?      | Yes, PR: #8616 
| Documentation PR Link    | 


packages/babel-plugin-transform-runtime/test/fixtures/runtime-corejs2/math-undefined-version/output.js
```diff
@@ -0,0 +1,51 @@
+var _Math$trunc = require("@babel/runtime-corejs2/core-js/math/trunc");
```
nicolo-ribaudo: I'm a little bit surprised that they are polyfilled even if a version >7.0.0 isn't specified, but I might have misunderstood something.

Siddharth11: https://github.com/babel/babel/pull/8616#discussion_r214744377

https://github.com/babel/babel/pull/8616/files#diff-bb1584ad79ba30b51f843278cc5a4e39R4

Isn't this the expected behaviour?

Siddharth11: I've intentionally kept `includMathModule` as `true` when `runtimeVersion` is `undefined`

loganfsmyth: That does seem surprising for me. It should pick up the default value in https://github.com/babel/babel/blob/d47a97477b86dd91fd8df4ee61ec4a42ee0f5673/packages/babel-plugin-transform-runtime/src/index.js#L40 and wouldn't be `undefined`.

Siddharth11: @loganfsmyth then this default value of runtimeVersion should be mentioned inside [definitions.js](https://github.com/babel/babel/pull/8616/files#diff-bb1584ad79ba30b51f843278cc5a4e39) too, because we don't pass any `runtimeVersion` to `definitions` function inside [build-dist.js](https://github.com/babel/babel/pull/8616/files#diff-26e211fe142ee8fcf4abd6b5cfce4fe9).

Siddharth11: Should I make that change in this PR itself?

loganfsmyth: That part is on purpose because `build-dist.js` should build the files without any version restrictions. The code creating this `output.js` should be running with the default from the transform though.

Siddharth11: Oh, okay. How can we debug this issue?

loganfsmyth: Ugh, this is the `semver` module itself screwing up
```
semver.intersects(`<7.0.0`, "7.0.0-beta.0") // false
semver.intersects(`>=7.0.0`, "7.0.0-beta.0") // false
```
so basically we're gonna be in trouble because we rely on prerelease versioning with `intersects` at the moment. Guess that's something we need to fix.

loganfsmyth: Alright, I've posted #8659 which should address this, so I'm actually going to land this and rebase that PR on top to make sure it fixes this right.

loganfsmyth: Cool, yup that fixes this: https://github.com/babel/babel/pull/8659/files#diff-a7c1c84984aaf6766458087aa80a1a70R1

PR #8628, opened 2018-09-04
add typescript types to generator/traverse

43081j:
| Tests Added + Pass?      | N/Y

---

Recently, #8170 was resolved by adding some type definitions to the parser.

The same issue exists with the traverse and generator packages.

For the generator, I have followed the structure of what was added to the parser.

For the traverse package, things are a little more complex. I have introduced a typescript generator script to the package and added it to the makefile. This produces a set of fixed types along side some dynamically generated types (via the types package)

What are your thoughts on this? Is it fine for us to introduce such a script and does it belong in the makefile?

---

**I think we are in need of a more structured way of maintaining these types.**

Problems we have right now:

* Manually maintain these types (its very easy for them to fall out of sync with the source)
* We are repeating ourselves (both generator scripts do very similar things)
* Because `@babel/types` has its own TS definitions now, _all_ babel packages on definitelytyped are invalid unless you use the old types (those from DT), so we have to support **all or nothing**

Does anyone have a suggestion? Maybe we can get this merged but look into having types generated from the flow annotations?


packages/babel-traverse/scripts/generators/typescript.js
```diff
@@ -0,0 +1,296 @@
...
+  replaceExpressionWithStatements(nodes: t.Node[]): t.Node;
+  replaceInline(nodes: t.Node | t.Node[]): void;
+
+  getTypeAnnotation(): t.TypeAnnotation;
+  isBaseType(baseName: string, soft?: boolean): boolean;
+  couldBeBaseType(name: string): boolean;
+  baseTypeStrictlyMatches(right: NodePath): boolean;
+  isGenericType(genericName: string): boolean;
+${isEntries.join("\n")}
+${assertEntries.join("\n")}
```
alsuren: These generated typings are more complete than the flow type annotations for NodePath. When I did a spike to get the flow type annotations into a useful state, I ran across a bunch of `NodeType.prototype` 
 trickery that flow wasn't clever enough to unpick. I ended up hand-unrolling them so that I could get on with my prototype, but this is not an approach that you would want to merge ( https://github.com/babel/babel/pull/9262/files#r244605487 ).

It may be that the simplest approach would be to generate both the flow and typescript annotations for NodePath from the same source of truth.

liuxingbaoyu:
Thank you for your PR! Since it is now obsolete, I will close it.

PR #8629, opened 2018-09-04
output aliased types in typescript declarations

43081j:
| Tests Added + Pass?      | N/Y

---

We weren't outputting aliased types in our `isXX` assertion methods so we would be losing type information (returning `boolean` instead).


43081j:
Do we have any tests for the generators at all? I'm happy to write some but wasn't sure if we already had any I could extend.

xtuc:
Yes that's right, sorry.

xtuc:
I think it doesn't make sense to add tests on the generator, we could instead just commit the file, but the point is to avoid that.

PR #8654, opened 2018-09-09
Private class methods stage 3

tim-mc:
| Q                       | A                                            |
| :---------------------: | :------------------------------------------: |
| Fixed Issues?           | https://github.com/babel/proposals/issues/22 |
| Patch: Bug Fix?         |                                              |
| Major: Breaking Change? |                                              |
| Minor: New Feature?     | Add private method support                   |
| Tests Added + Pass?     | Yes                                          |
| Documentation PR        |                                              |
| Any Dependency Changes? | No                                           |
| License                 | MIT                                          |
| Sponsor                 | @bloomberg                                   |

cc: @rricard @robpalme @littledan @syg

### DIFF
--- a/packages/babel-generator/test/fixtures/types/ClassBody-MethodDefinition/input.js
+++ b/packages/babel-generator/test/fixtures/types/ClassBody-MethodDefinition/input.js
@@ -5,6 +5,15 @@ class Foo {
   get foo() {}
   set foo(bar) {}
 
+  async #foo() {}
+  #foo() {}
+  get #foo() {}
+  set #foo(bar) {}
+  * #foo() {}
+  async * #foo() {}
+  get #bar() {}
+  set #baz(taz) {}
+
   static async foo() {}
   static foo() {}
   static ["foo"]() {}
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -1739,3 +1739,18 @@ helpers.decorate = helper("7.1.5")`
     return constructor;
   }
 `;
+
+helpers.classPrivateMethodGet = helper("7.1.6")`
+  export default function _classPrivateMethodGet(receiver, privateSet, fn) {
+    if (!privateSet.has(receiver)) {
+      throw new TypeError("attempted to get private field on non-instance");
+    }
+    return fn;
+  }
+`;
+
+helpers.classPrivateMethodSet = helper("7.1.6")`
+  export default function _classPrivateMethodSet() {
+    throw new TypeError("attempted to reassign private method");
+  }
+`;
--- a/packages/babel-plugin-class-features/src/features.js
+++ b/packages/babel-plugin-class-features/src/features.js
@@ -45,17 +45,22 @@ export function verifyUsedFeatures(path, file) {
     );
   }
 
-  if (
-    path.isClassPrivateMethod() &&
-    !hasFeature(file, FEATURES.privateMethods)
-  ) {
-    throw path.buildCodeFrameError("Class private methods are not enabled.");
-  }
+  if (path.isClassPrivateMethod()) {
+    if (!hasFeature(file, FEATURES.privateMethods)) {
+      throw path.buildCodeFrameError("Class private methods are not enabled.");
+    }
 
-  if (path.isClassPrivateMethod() && path.node.static) {
-    throw new Error(
-      "@babel/plugin-class-features doesn't support class static private methods yet.",
-    );
+    if (path.node.static) {
+      throw new Error(
+        "@babel/plugin-class-features doesn't support class static private methods yet.",
+      );
+    }
+
+    if (path.node.kind !== "method") {
+      throw new Error(
+        "@babel/plugin-class-features doesn't support class private accessors yet.",
+      );
+    }
   }
 
   if (hasDecorators(path) && !hasFeature(file, FEATURES.decorators)) {
--- a/packages/babel-plugin-class-features/src/features.js
+++ b/packages/babel-plugin-class-features/src/features.js
@@ -39,6 +39,10 @@ export function isLoose(file, feature) {
 }
 
 export function verifyUsedFeatures(path, file) {
+  if (hasDecorators(path) && !hasFeature(file, FEATURES.decorators)) {
+    throw path.buildCodeFrameError("Decorators are not enabled.");
+  }
+
   if (hasFeature(file, FEATURES.decorators)) {
     throw new Error(
       "@babel/plugin-class-features doesn't support decorators yet.",
@@ -63,8 +67,15 @@ export function verifyUsedFeatures(path, file) {
     }
   }
 
-  if (hasDecorators(path) && !hasFeature(file, FEATURES.decorators)) {
-    throw path.buildCodeFrameError("Decorators are not enabled.");
+  if (
+    hasFeature(file, FEATURES.privateMethods) &&
+    hasFeature(file, FEATURES.fields) &&
+    isLoose(file, FEATURES.privateMethods) !== isLoose(file, FEATURES.fields)
+  ) {
+    throw new Error(
+      "'loose' mode configuration must be the same for both @babel/plugin-proposal-class-properties " +
+        "and @babel/plugin-proposal-private-methods",
+    );
   }
 
   if (path.isProperty()) {
--- a/packages/babel-plugin-class-features/src/features.js
+++ b/packages/babel-plugin-class-features/src/features.js
@@ -55,13 +55,13 @@ export function verifyUsedFeatures(path, file) {
     }
 
     if (path.node.static) {
-      throw new Error(
+      throw path.buildCodeFrameError(
         "@babel/plugin-class-features doesn't support class static private methods yet.",
       );
     }
 
     if (path.node.kind !== "method") {
-      throw new Error(
+      throw path.buildCodeFrameError(
         "@babel/plugin-class-features doesn't support class private accessors yet.",
       );
     }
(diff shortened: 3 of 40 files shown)


packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -187,6 +189,18 @@ export default declare((api, options) => {
     },
   };
 
+  const privateMethodHandlerSpec = {
+    ...privateNameHandlerSpec,
```
littledan: Looks like this will inherit the behavior to use classPrivateFieldSet. I don't think that's what you want here; you probably want to throw a TypeError for methods and call the setter for accessors.

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -187,6 +189,18 @@ export default declare((api, options) => {
...
+    get(member) {
+      const { map, file } = this;
+
+      return t.callExpression(file.addHelper("classPrivateMethodGet"), [
```
littledan: How will this handle getters?

packages/babel-generator/test/fixtures/types/ClassBody-MethodDefinition/input.js
```diff
@@ -3,7 +3,12 @@ class Foo {
   foo() {}
   ["foo"]() {}
   get foo() {}
-  set foo(bar) {}
+  set foo(bar) { }
+
+  async #foo() {}
+  #foo() {}
+  get #foo() {}
+  set #foo(bar) {}
```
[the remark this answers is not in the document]
tim-mc: Added!

packages/babel-types/src/definitions/experimental.js
```diff
@@ -131,6 +134,28 @@ defineType("ClassPrivateProperty", {
   },
 });
 
+defineType("ClassPrivateMethod", {
+  builder: ["kind", "key", "params", "body", "computed", "static"],
```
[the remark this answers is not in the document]
tim-mc: I believe that I put that there because `computed` is a field (from `classMethodOrDeclareMethodCommon`). Should it not be here?

packages/babel-helpers/src/helpers.js
```diff
@@ -1717,3 +1717,12 @@ helpers.decorate = helper("7.0.1")`
     return constructor;
   }
   `;
+
+helpers.classPrivateMethodGet = helper("7.0.0-beta.0")`
+  export default function _classPrivateMethodGet(receiver, privateMap) {
```
nicolo-ribaudo: Can't we reuse the `classPrivateFieldGet` helper? Since with decorators methods could become fields and vice versa, I'd prefer fields/methods/accessors to be handled the same way.

nicolo-ribaudo:
```js
class A {
  #method() {}
  
  getMethod() {
    return this.#method;
  }
}

console.log(new A().getMethod() === new A().getMethod());
```

Is this `true` or `false`?

packages/babel-plugin-proposal-class-properties/src/index.js
```diff
@@ -323,6 +337,51 @@ export default declare((api, options) => {
...
+    const map = scope.generateUidIdentifier(name);
+    memberExpressionToFunctions(parentPath, privateNameVisitor, {
+      name,
+      map,
+      file: state,
+      ...privateMethodHandlerSpec,
+    });
+
+    initNodes.push(
+      template.statement`var MAP = new WeakMap();`({
```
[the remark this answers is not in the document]
tim-mc: Working on implementing it with `WeakSet`s now!

littledan:
@nicolo-ribaudo That should be `true`--they should all have the same function object as the method.

packages/babel-helpers/src/helpers.js
tim-mc: This was recently changed when I switched to using `WeakSet`s.

tim-mc:
I've just pushed up a fix for the name collision issue that I described earlier in https://github.com/babel/babel/pull/8654#issuecomment-430649557

tim-mc:
@nicolo-ribaudo @jridgewell I'm working on the loose implementation currently, but would be good to get some eyes on the re-worked WeakSet implementation I've pushed for spec.

It appears as though the issue I mentioned in my [earlier comment](https://github.com/babel/babel/pull/8654#issuecomment-430649557) re: helpers not getting updated in the REPL has resolved itself somehow. On a related note, I am curious what I should be putting as the version [here](https://github.com/babel/babel/pull/8654/files#diff-ba75bd4d116cc90ad8c8002e3010e821R1743) (and how exactly that works).

Also, in terms of the tests that I've added, is the current set sufficiently exhaustive for the purpose of this PR (given that the additions should also be passing 262 tests)?

tim-mc:
Due to a pending refactor PR (#8130), this PR is currently blocked. A WIP PR that takes the refactor into account has been opened [here](https://github.com/nicolo-ribaudo/babel/pull/1) to garner feedback until the refactor has been merged. Once that happens, I will update this PR with the new changes.

packages/babel-plugin-class-features/src/fields.js
```diff
@@ -20,20 +24,25 @@ export function buildPrivateNamesMap(props) {
...
+    // because static fields are directly assigned to a variable in the
+    // buildPrivateStaticFieldInitSpec function.
 
     if (loose) {
       initNodes.push(
         template.statement.ast`
           var ${id} = ${state.addHelper("classPrivateFieldLooseKey")}("${name}")
         `,
       );
+    } else if (isMethod) {
```
nicolo-ribaudo: Can we add an assertion (it could even be a comment) that `isStatic` is false? It would make it easier to read this code.

packages/babel-plugin-class-features/src/fields.js
```diff
@@ -108,13 +119,24 @@ const privateNameHandlerSpec = {
...
-    const { id, static: isStatic } = privateNamesMap.get(name);
+    const {
+      id,
+      static: isStatic,
+      method: isMethod,
+      methodId,
+    } = privateNamesMap.get(name);
 
     if (isStatic) {
       return t.callExpression(
```
nicolo-ribaudo: Also assert here that `// ASSERT: !isMethod` and in some other places where it is needed.

packages/babel-plugin-class-features/src/fields.js
```diff
@@ -231,6 +257,25 @@ function buildPrivateStaticFieldInitSpec(prop, privateNamesMap) {
...
 }
 
+function buildPrivateMethodInitLoose(ref, prop, privateNamesMap) {
+  const { methodId, id } = privateNamesMap.get(prop.node.key.id.name);
+
+  return template.statement.ast`
+    Object.defineProperty(${ref}, ${id}, {
+      // configurable is false by default
+      // enumerable is false by default
+      writable: false,
```
nicolo-ribaudo: This is false by default

packages/babel-plugin-class-features/src/features.js
```diff
@@ -44,9 +44,17 @@ export function verifyUsedFeatures(path, file) {
...
-  if (hasFeature(file, FEATURES.privateMethods)) {
+
+  if (
+    path.isClassPrivateMethod() &&
+    !hasFeature(file, FEATURES.privateMethods)
+  ) {
+    throw path.buildCodeFrameError("Class private methods are not enabled.");
+  }
+
+  if (path.isClassPrivateMethod() && path.node.static) {
```
nicolo-ribaudo: We should also guard agains private accessors.

packages/babel-plugin-proposal-private-methods/test/fixtures/private-method/exfiltrated/exec.js
```diff
@@ -0,0 +1,13 @@
...
+
+    constructor() {
+        if (exfiltrated === undefined) {
+            exfiltrated = this.#privateMethod;
+        }
+        expect(exfiltrated).toStrictEqual(this.#privateMethod);
+    }
+}
+
+new Foo();
```
nicolo-ribaudo: Can you add another `new Foo()`? So it also checks that the private method is the same accross different instances.

tim-mc: Added to both strict and loose.

packages/babel-plugin-class-features/src/features.js
tim-mc: Added!

tim-mc:
@nicolo-ribaudo Does [this](https://github.com/babel/babel/pull/8654/files#diff-d3e152fc190c42c2a8db2130c28a2ef0R94) need to be changed?

I also added a check for ensuring that the private methods and class fields plugins loose modes are both configured to be the same (if both are enabled).

tim-mc:
@nicolo-ribaudo Any suggestions on how best to handle [that](https://github.com/babel/babel/pull/8654/files#diff-d3e152fc190c42c2a8db2130c28a2ef0R94)? I could change the exported function to check for both class fields and private method loose mode settings. I assume that the change is necessary for this PR?

tim-mc:
I have private method accessors ready in a separate branch - is it preferred that I combine that with this PR? @nicolo-ribaudo

nicolo-ribaudo:
Nope, let's keep this as-is to make it easy to be reviewed.

tim-mc:
Thanks @jridgewell, pushed up changes.

nicolo-ribaudo:
Thank you very much 🎉

deadlyicon:
https://babeljs.io/repl/build/9484/ doesnt seem to work with private methods
image

nicolo-ribaudo:
@deadlyicon The new plugin needs to be added to babel-standalone. Do you want to open a PR?

deadlyicon:
> @deadlyicon The new plugin needs to be added to babel-standalone. Do you want to open a PR?

I'm just suggesting that this comment https://github.com/babel/babel/pull/8654#issuecomment-419682847 be changed so it is more clear on how to play with this new feature. I couldn't get it to work. 

thanks :)

PR #8656, opened 2018-09-09
 Allow extending untranspiled classes

### LINKED ISSUE #7022: How to fix this:Error in mounted hook: "TypeError: Class constructor Base cannot be invoked without 'new'"

nicolo-ribaudo:
| Fixed Issues?            | Fixes #7022
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Yes?[1]
| Documentation PR Link    | 
| Any Dependency Changes?  |

Prev: #7081

This PR allows transpiled classes to extends untranspiled classes. This can happen, for example, when a dependency is transpiled using Babel and extends an user-provided class.
Example: https://github.com/Oblosys/react-lifecycle-visualizer/issues/4

The new helper only allows extending native classes when `Reflect.construct` is natively supported. Some envs (e.g. Node 4) have partial support for classes but don't provide `Reflect.construct`: in this case users need to ensure that every class (also in their dependencies) is transpiled, like it was before.

[1] If someone sets `.constructor` to something else, the new helper won't work. e.g.
```js
class A {}
class B extends A {}

B.prototype.constructor = A;

console.log(new B() instanceof A); // true
console.log(new B() instanceof B); // false, should be true
```
A workaround to avoid the breaking change could be to only use the `Reflect`-based version of the helper only if `Function.toString.call(SuperClass).slice(0, 5) === "class"`, but in those cases `this.constructor` still can't be modified.

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -502,10 +502,8 @@ helpers.setPrototypeOf = helper("7.0.0-beta.0")`
   }
 `;
 
-helpers.construct = helper("7.0.0-beta.0")`
-  import setPrototypeOf from "setPrototypeOf";
-
-  function isNativeReflectConstruct() {
+helpers.isNativeReflectConstruct = helper("7.9.0")`
+  export default function _isNativeReflectConstruct() {
     if (typeof Reflect === "undefined" || !Reflect.construct) return false;
 
     // core-js@3
@@ -529,6 +527,11 @@ helpers.construct = helper("7.0.0-beta.0")`
       return false;
     }
   }
+`;
+
+helpers.construct = helper("7.0.0-beta.0")`
+  import setPrototypeOf from "setPrototypeOf";
+  import isNativeReflectConstruct from "isNativeReflectConstruct";
 
   export default function _construct(Parent, args, Class) {
     if (isNativeReflectConstruct()) {
@@ -731,6 +734,26 @@ helpers.possibleConstructorReturn = helper("7.0.0-beta.0")`
     return assertThisInitialized(self);
   }
 `;
+
+helpers.createSuper = helper("7.9.0")`
+  import getPrototypeOf from "getPrototypeOf";
+  import isNativeReflectConstruct from "isNativeReflectConstruct";
+  import possibleConstructorReturn from "possibleConstructorReturn";
+
+  export default function _createSuper(Derived) {
+    return function () {
+      var Super = getPrototypeOf(Derived), result;
+      if (isNativeReflectConstruct()) {
+        // NOTE: This doesn't work if this.__proto__.constructor has been modified.
+        var NewTarget = getPrototypeOf(this).constructor;
+        result = Reflect.construct(Super, arguments, NewTarget);
+      } else {
+        result = Super.apply(this, arguments);
+      }
+      return possibleConstructorReturn(this, result);
+    }
+  }
+ `;
 
 helpers.superPropBase = helper("7.0.0-beta.0")`
   import getPrototypeOf from "getPrototypeOf";
--- a/packages/babel-core/test/fixtures/transformation/misc/regression-1155/options.json
+++ b/packages/babel-core/test/fixtures/transformation/misc/regression-1155/options.json
@@ -1,7 +1,7 @@
 {
   "plugins": [
     "transform-classes",
-    "external-helpers",
+    ["external-helpers", { "helperVersion": "7.100.0" }],
     "transform-block-scoping"
   ]
 }
--- a/packages/babel-core/test/fixtures/transformation/misc/regression-1155/output.js
+++ b/packages/babel-core/test/fixtures/transformation/misc/regression-1155/output.js
@@ -3,6 +3,8 @@ var Foo = /*#__PURE__*/function (_Bar) {
 
   babelHelpers.inherits(Foo, _Bar);
 
+  var _super = babelHelpers.createSuper(Foo);
+
   function Foo(options) {
     babelHelpers.classCallCheck(this, Foo);
     var parentOptions = {};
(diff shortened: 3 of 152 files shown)


packages/babel-helpers/src/helpers.js
```diff
@@ -478,6 +476,11 @@ helpers.construct = helper("7.0.0-beta.0")`
       return false;
     }
   }
+`;
+
+helpers.construct = helper("7.0.0-beta.0")`
```
xtuc: This one should be `7.0.1` too?

vzaidman:
hey! any update on the status of this pr?

aminsol:
Any update? I have posted a question on StackOverflow which I think is related to this:
[StackOverflow Question](https://stackoverflow.com/questions/56802060/sequelize-five-5-babel-seven-7-typeerror-class-constructor-model-cannot-be-in)

aminsol:
Any update? I have posted a question on StackOverflow which I think is related to this:
[StackOverflow Question](https://stackoverflow.com/questions/56802060/sequelize-five-5-babel-seven-7-typeerror-class-constructor-model-cannot-be-in)

1Jesper1:
Merge please.

vzaidman:
Hey!
Can I help in merging this one somehow?

ntulele:
+1 for getting this merged - thanks!

packages/babel-helpers/src/helpers.js
nicolo-ribaudo: No, because this helper was already available. I only restructured its internals.

vzaidman:
great!

packages/babel-helpers/src/helpers.js
```diff
@@ -730,6 +733,29 @@ helpers.possibleConstructorReturn = helper("7.0.0-beta.0")`
...
+  export default function _createSuper(Derived) {
+    return function () {
+      var Super = getPrototypeOf(Derived), result;
+      if (isNativeReflectConstruct()) {
+        for (var l = arguments.length, args = new Array(l), i = 0; i < l; i++) {
+          args[i] = arguments[i];
+        }
+        // NOTE: This doesn't work if this.__proto__.constructor has been modified.
+        var NewTarget = getPrototypeOf(this).constructor;
+        result = Reflect.construct(Super, args, NewTarget);
```
JLHwung: Why we need shallow copy the `arguments`? `Reflect.construct` accepts array-like objects. https://tc39.es/ecma262/#sec-reflect.construct

nicolo-ribaudo: It's because directly passing `arguments` deoptimizes the function in some engines.

JLHwung: What about using `Array.from`? comparing [`Reflect.construct`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/construct) with [`Array.from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) shows that `Array.from` is supported on every platforms with `Reflect.construct` support.

Also I am not sure if we should still write codes specifically for V8 <= 5.9 (when ignition pipeline is not enabled and when https://jsperf.com/123124235 shows 5x difference), the [usage](https://caniuse.com/usage-table) of `chrome <= 59` is less than 1% now. But since Chrome 49 is 0.52%, guess we should wait for a longer times.

nicolo-ribaudo: Do you mean `Array.from(arguments)`? Then it's the same as directly passing `arguments` to `Reflect.construct` (which I can do).

JLHwung: Well I thought `Array.from` will shallow copy the `arguments` into a new array so to avoid the v8 <= 5.9 performance issue. But 👍 for clearer code.

nicolo-ribaudo: Well, passing `arguments` to `Array.from` doesn't cause a deopt anyway?

PR #8658, opened 2018-09-10
Format fixture options.json with Prettier.

loganfsmyth:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

I don't think we need to run this often, but it seems like a nice-to-have.


Makefile
```diff
@@ -52,6 +52,9 @@ lint:
 fix:
```
xtuc: `fix-json` could be invoked when calling `fix`, i'll do a PR.

xtuc: https://github.com/babel/babel/pull/8661

PR #8659, opened 2018-09-10
Fix version checks in .availableHelper and transform-runtime definitions.

loganfsmyth:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The comments in the code pretty much explain why this change is needed. I _think_ this is enough to fix the issue, but it's actually a little difficult to be 100% sure since I don't entirely trust `semver` anymore. 

As noted by the changes in the tests, this is definitely something that we need to fix in order for things to function properly. On `7.0.0` essentially any helper will show up as available, unfortunately.

An alternative fix here would be to change `transform-runtime` and `external-helpers` to stop passing `7.0.0-beta.0` as their default version, but I think the approach here is better?

Since we've landed so much and might just straight to 7.1.0, I'm actually tempted to branch and release a `7.0.1` with just this applied on top, so people have more of a chance to get updated to a version with a functional `.availableHelper()`. On the fence there though.


hzoo:
Ok, I made https://github.com/babel/babel/tree/7.0.patch branch off of `7.0.0`. We can merge this and cherry pick this and other patches onto that branch and release it as a `7.0.1` (a one-off)? Then land the next minor/releases from master as usual.

PR #8660, opened 2018-09-10
Better support monorepos by allowing users to opt into automatically resolving 'root' with `rootMode: "upward"`.

loganfsmyth:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is an alternative to #8636 that is a bit simpler. It introduces a separate `rootMode: "root" | "upward" | "upward-optional"` option. From [the docs PR](https://github.com/babel/website/pull/1822):

* `"root"` - Passes the `"root"` value through as unchanged.
* `"upward"` - Walks upward from the `"root"` directory, looking
  for a directory containing a `babel.config.js`
  file, and throws an error if a `babel.config.js`
  is not found.
* `"upward-optional"` - Walk upward from the `"root"` directory,
  looking for a directory containing a `babel.config.js`
  file, and falls back to `"root"` if a `babel.config.js`
  is not found.

Otherwise, this is very similar to #8636 and has all of the same requirements of it being an opt-in feature.

### DIFF
--- a/packages/babel-cli/src/babel/options.js
+++ b/packages/babel-cli/src/babel/options.js
@@ -28,6 +28,11 @@ commander.option(
   "The name of the 'env' to use when loading configs and plugins. " +
     "Defaults to the value of BABEL_ENV, or else NODE_ENV, or else 'development'.",
 );
+commander.option(
+  "--root-mode [mode]",
+  "The project-root resolution mode. " +
+    "One of 'root' (the default), 'upward', or 'upward-optional'.",
+);
 
 // Basic file input configuration.
 commander.option("--source-type [script|module]", "");
@@ -220,6 +225,7 @@ export default function parseArgv(args: Array<string>) {
     babelOptions: {
       presets: opts.presets,
       plugins: opts.plugins,
+      rootMode: opts.rootMode,
       configFile: opts.configFile,
       envName: opts.envName,
       sourceType: opts.sourceType,
--- a/packages/babel-core/src/config/files/configuration.js
+++ b/packages/babel-core/src/config/files/configuration.js
@@ -24,6 +24,21 @@ const BABELRC_FILENAME = ".babelrc";
 const BABELRC_JS_FILENAME = ".babelrc.js";
 const BABELIGNORE_FILENAME = ".babelignore";
 
+export function findConfigUpwards(rootDir: string): string | null {
+  let dirname = rootDir;
+  while (true) {
+    if (fs.existsSync(path.join(dirname, BABEL_CONFIG_JS_FILENAME))) {
+      return dirname;
+    }
+
+    const nextDir = path.dirname(dirname);
+    if (dirname === nextDir) break;
+    dirname = nextDir;
+  }
+
+  return null;
+}
+
 export function findRelativeConfig(
   packageData: FilePackageData,
   envName: string,
--- a/packages/babel-core/src/config/partial.js
+++ b/packages/babel-core/src/config/partial.js
@@ -6,9 +6,43 @@ import { mergeOptions } from "./util";
 import { createItemFromDescriptor } from "./item";
 import { buildRootChain, type ConfigContext } from "./config-chain";
 import { getEnv } from "./helpers/environment";
-import { validate, type ValidatedOptions } from "./validation/options";
+import {
+  validate,
+  type ValidatedOptions,
+  type RootMode,
+} from "./validation/options";
 
-import type { ConfigFile, IgnoreFile } from "./files";
+import { findConfigUpwards, type ConfigFile, type IgnoreFile } from "./files";
+
+function resolveRootMode(rootDir: string, rootMode: RootMode): string {
+  switch (rootMode) {
+    case "root":
+      return rootDir;
+
+    case "upward-optional": {
+      const upwardRootDir = findConfigUpwards(rootDir);
+      return upwardRootDir === null ? rootDir : upwardRootDir;
+    }
+
+    case "upward": {
+      const upwardRootDir = findConfigUpwards(rootDir);
+      if (upwardRootDir !== null) return upwardRootDir;
+
+      throw Object.assign(
+        (new Error(
+          `Babel was run with rootMode:"upward" but a root could not ` +
+            `be found when searching upward from "${rootDir}"`,
+        ): any),
+        {
+          code: "BABEL_ROOT_NOT_FOUND",
+          dirname: rootDir,
+        },
+      );
+    }
+    default:
+      throw new Error(`Assertion failure - unknown rootMode value`);
+  }
+}
 
 export default function loadPrivatePartialConfig(
   inputOpts: mixed,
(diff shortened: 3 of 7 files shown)


packages/babel-core/src/config/files/configuration.js
```diff
@@ -24,6 +24,30 @@ const BABELRC_FILENAME = ".babelrc";
...
+
+    const nextDir = path.dirname(dirname);
+    if (dirname === nextDir) break;
+    dirname = nextDir;
+  }
+
+  throw Object.assign(
+    (new Error(
+      `Babel was run with autoRoot:true but a root babel.config.js could not ` +
+        `be found when searching from "${rootDir}"`,
```
xtuc: To help the user troubleshoot its setup, we could log every dir?
> [...] when searching in `/babel/a`, `/babel`, `/`, ...

xtuc:
Note for members: need documentation

packages/babel-core/src/config/files/configuration.js
```diff
@@ -24,6 +24,30 @@ const BABELRC_FILENAME = ".babelrc";
 const BABELRC_JS_FILENAME = ".babelrc.js";
 const BABELIGNORE_FILENAME = ".babelignore";
 
+export function findClosestConfig(rootDir: string): string {
```
xtuc: Instead of `Closes`, which is not very accurate, could we name this: `findConfigUpwards`, `findParentConfig`, etc.

loganfsmyth: Those are good suggestions, thanks. I'll go with `findConfigUpwards`.

packages/babel-core/src/config/files/configuration.js
loganfsmyth: I tweaked the text to mention the upward search, do you think that's enough? I worry that including all the directories would make the error harder to read.

packages/babel-core/src/config/partial.js
```diff
@@ -6,9 +6,43 @@ import { mergeOptions } from "./util";
...
+    }
 
-import type { ConfigFile, IgnoreFile } from "./files";
+    case "upward": {
+      const upwardRootDir = findConfigUpwards(rootDir);
+      if (upwardRootDir !== null) return upwardRootDir;
+
+      throw Object.assign(
+        (new Error(
+          `Babel was run with rootMode:"${rootMode}" but a root could not ` +
```
nicolo-ribaudo: Nit: No need to interpolate, you can just write `rootMode: "upward"`.

loganfsmyth: Thanks! I moved the code around and forgot to update this snippet.

loganfsmyth:
Docs posted in https://github.com/babel/website/pull/1822

xtuc:
I'm wondering about something, why does this even need to be in Babel? User could import a function (from core or a new package) that will search the root directory, like the logic in here.

```js
const { findUpwards } = require("@babel/let-me-configure-it");

module.exports = {
  rootDirectory: findUpwards
}

// or

module.exports = {
  rootDirectory: process.cwd
}
``` 

That would also enable monorepo framework authors (lerna?) to provide their own root resolver.

packages/babel-core/src/config/partial.js
```diff
@@ -6,9 +6,43 @@ import { mergeOptions } from "./util";
...
+    }
 
-import type { ConfigFile, IgnoreFile } from "./files";
+    case "upward": {
+      const upwardRootDir = findConfigUpwards(rootDir);
+      if (upwardRootDir !== null) return upwardRootDir;
+
+      throw Object.assign(
+        (new Error(
+          `Babel was run with rootMode:"upward" but a root could not ` +
```
xtuc: nit: space after the colon

packages/babel-core/src/config/partial.js
```diff
@@ -6,9 +6,43 @@ import { mergeOptions } from "./util";
...
+            `be found when searching upward from "${rootDir}"`,
+        ): any),
+        {
+          code: "BABEL_ROOT_NOT_FOUND",
+          dirname: rootDir,
+        },
+      );
+    }
+    default:
+      throw new Error(`Assertion failure - unknown rootMode value`);
```
xtuc: We could also link to the website to learn more.

packages/babel-cli/src/babel/options.js
```diff
@@ -28,6 +28,11 @@ commander.option(
   "The name of the 'env' to use when loading configs and plugins. " +
     "Defaults to the value of BABEL_ENV, or else NODE_ENV, or else 'development'.",
 );
+commander.option(
+  "--root-mode [mode]",
+  "The project-root resolution mode. " +
+    "One of 'root' (the default), 'upward', or 'upward-optional'.",
```
xtuc: I think that Instead of `root` could we name it `cwd`? or `.`?

packages/babel-core/src/config/files/configuration.js
```diff
@@ -24,6 +24,21 @@ const BABELRC_FILENAME = ".babelrc";
...
+export function findConfigUpwards(rootDir: string): string | null {
+  let dirname = rootDir;
+  while (true) {
+    if (fs.existsSync(path.join(dirname, BABEL_CONFIG_JS_FILENAME))) {
+      return dirname;
+    }
+
+    const nextDir = path.dirname(dirname);
+    if (dirname === nextDir) break;
+    dirname = nextDir;
```
xtuc: Maybe we could set a max limit, we can be almost sure to not find any config after 30 iterations. That will avoid tranversing the whole system in some special setups.

packages/babel-cli/src/babel/options.js
nicolo-ribaudo: `cwd` and `root` can be two different things when running Babel, so it would be confusing.

packages/babel-core/src/config/partial.js
loganfsmyth: I don't feel strongly really, but I left out the space to make it clearer it was kind of a key/value pair.

packages/babel-core/src/config/partial.js
loganfsmyth: I don't generally do that with assertions. This is mostly here to satisfy Flow. The user should never see this and if we got here, it means Flow didn't catch a type error somewhere, since it would mean we got a `RootMode` type that didn't match the type definition.

packages/babel-cli/src/babel/options.js
loganfsmyth: I'm not sure mentioning the working directory in this message will help honestly. Without the docs, I doubt users would know what any of these 3 mean anyway.

packages/babel-core/src/config/files/configuration.js
loganfsmyth: We do this type of look in several places during config resolution. I'm not sure it's worth it to over-optimize up front, especially since you have to opt into the `upward-optional` mode for it to have been wasted time.

packages/babel-core/src/config/partial.js
xtuc: Got you, there's a config validation before that anyway (`assertRootMode`).

Just FYI I used to use https://github.com/xtuc/mamacro#examples for only assertion, they are quite close to C ones.

packages/babel-core/src/config/files/configuration.js
xtuc: Ok that make sense, i'm just worried about some undebuggable issues that it could cause.

packages/babel-core/src/config/partial.js
xtuc: I don't either, let's leave it like that.

loganfsmyth:
@hzoo Good call, done!

PR #5150, opened 2017-01-17
Add code frame when config parse error

xtuc:
| Q                        | A 
| ------------------------ | ---
| Patch: Bug Fix?          | no
| Major: Breaking Change?  | no
| Minor: New Feature?      | yes
| Deprecations?            | no
| Spec Compliancy?         | no
| Tests Added/Pass?        | no
| Fixed Tickets            | no
| License                  | MIT
| Doc PR                   | no
| Dependency Changes       | yes

I would like to add more context to error regarding the configuration. I wanted to know what do you think about displaying the code frame when a parse error occurred.

Example:

```sh
SyntaxError: 
/home/foo/bar/babel/babel/.babelrc:
  1 | {
> 2 |     plugins: [ 'test'' ]
    |                      ^
  3 | }
  4 | 
Error while parsing JSON - Expected ']' instead of ''' at line 2 column 22 of the JSON5 data. Still to read: "' ]\n}\n"
    at buildConfigChain (/home/foo/bar/babel/babel/packages/babel-core/lib/transformation/file/options/build-config-chain.js:2078:13)
```

PR #8665, opened 2018-09-10
Add CodeFrame on JSON error

xtuc:
| Tests Added + Pass?      | No + yes
| Documentation PR Link    | 
| Any Dependency Changes?  |

Port of https://github.com/babel/babel/pull/5150

### DIFF
--- a/packages/babel-core/src/config/files/configuration.js
+++ b/packages/babel-core/src/config/files/configuration.js
@@ -2,6 +2,7 @@
 
 import buildDebug from "debug";
 import path from "path";
+import codeFrame from "babel-code-frame";
 import fs from "fs";
 import json5 from "json5";
 import resolve from "resolve";
@@ -230,7 +231,10 @@ const readConfigJSON5 = makeStaticFileCache((filepath, content) => {
   try {
     options = json5.parse(content);
   } catch (err) {
-    err.message = `${filepath}: Error while parsing config - ${err.message}`;
+    const code = codeFrame(content, err.lineNumber, err.columnNumber);
+    err.message =
+      `${filepath}: Error while parsing JSON (${err.message})\n\n` + code;
+
     throw err;
   }
 
--- a/packages/babel-core/src/config/files/configuration.js
+++ b/packages/babel-core/src/config/files/configuration.js
@@ -2,7 +2,7 @@
 
 import buildDebug from "debug";
 import path from "path";
-import codeFrame from "babel-code-frame";
+import codeFrame from "@babel/code-frame";
 import fs from "fs";
 import json5 from "json5";
 import resolve from "resolve";
--- a/packages/babel-core/src/config/files/configuration.js
+++ b/packages/babel-core/src/config/files/configuration.js
@@ -2,7 +2,7 @@
 
 import buildDebug from "debug";
 import path from "path";
-import codeFrame from "@babel/code-frame";
+import { codeFrameColumns } from "@babel/code-frame";
 import fs from "fs";
 import json5 from "json5";
 import resolve from "resolve";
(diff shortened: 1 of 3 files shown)


packages/babel-core/src/config/files/configuration.js
```diff
@@ -2,6 +2,7 @@
 
 import buildDebug from "debug";
 import path from "path";
+import codeFrame from "babel-code-frame";
```
existentialism: Should this should be `@babel/code-frame`?

xtuc: yes thanks

packages/babel-core/src/config/files/configuration.js
```diff
@@ -2,6 +2,7 @@
 
 import buildDebug from "debug";
 import path from "path";
+import codeFrame from "@babel/code-frame";
```
loganfsmyth: The default import from this is marked deprecated: https://github.com/babel/babel/blob/9ad8b2bb2c239505a874003166fdb1ca9189feee/packages/babel-code-frame/src/index.js#L183

packages/babel-core/src/config/files/configuration.js
```diff
@@ -230,7 +231,10 @@ const readConfigJSON5 = makeStaticFileCache((filepath, content) => {
   try {
     options = json5.parse(content);
   } catch (err) {
-    err.message = `${filepath}: Error while parsing config - ${err.message}`;
+    const code = codeFrame(content, err.lineNumber, err.columnNumber);
```
loganfsmyth: We should check typeof err.lineNumber === "number" when generating the code frame and skip it otherwise, I think, since .lineNumber isn't a speced property.

packages/babel-core/src/config/files/configuration.js
```diff
@@ -230,7 +231,21 @@ const readConfigJSON5 = makeStaticFileCache((filepath, content) => {
   try {
     options = json5.parse(content);
   } catch (err) {
-    err.message = `${filepath}: Error while parsing config - ${err.message}`;
+    // Check if the error contains source location informations
+    if (
+      typeof err.lineNumber === "number" ||
+      typeof err.columnNumber === "number"
+    ) {
+      const loc: NodeLocation = {
```
loganfsmyth: This is a Flow error, since `NodeLocation` isn't declared anywhere. Also looks like code-frame doesn't actually have `//@flow` and doesn't export the type at the moment :(

packages/babel-core/src/config/files/configuration.js
```diff
@@ -230,7 +231,21 @@ const readConfigJSON5 = makeStaticFileCache((filepath, content) => {
   try {
     options = json5.parse(content);
   } catch (err) {
-    err.message = `${filepath}: Error while parsing config - ${err.message}`;
+    // Check if the error contains source location informations
+    if (
+      typeof err.lineNumber === "number" ||
+      typeof err.columnNumber === "number"
```
loganfsmyth: I don't think this is needed, or if we do want it, this should be `&&`ed, not `||`ed, since things could get weird if there's a column and no line.

xtuc: Oh sorry I wanted to ensure that both are present, I was a bit in a rush sorry.

packages/babel-core/src/config/files/configuration.js
xtuc: I added a flow annotation in the file.

xtuc:
Looks like I broke something. @loganfsmyth i'll ping you once my PR is ready, so you don't need to lose your time on intermediate states. Thanks.

PR #8682, opened 2018-09-11
Fix TSParameterProperty getting lost with transform-classes

### LINKED ISSUE #7074: [plugin-transform-typescript] constructor auto-assign not working with plugin-transform-classes
The following doesn't seem to work, but is also not listed in [the readme](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-typescript)

If it's not possible to fix/implement, it would be good to throw an error when used, and mention it there.

### Input Code
```ts
class Foo {
  constructor(public type: string) {
    console.log('type', this.type);
  }
}
new Foo('bar');
```

```json
{
  "presets": [
    ["@babel/env", {
      "targets": {
        "browsers": ["last 2 versions"],
        "node": "8"
      }
    }],
    "@babel/preset-stage-3",
    "@babel/typescript"
  ]
}
```

### Expected Behavior
* have `this.type = type` added as first line in the constructor, or
* throw an error that this syntax is not supported (so it won't be used by accident, resulting in incorrect transpiled code)

TypeScript transpiles to:
```js
var Foo = /** @class */ (function () {
    function Foo(type) {
        this.type = type;
        console.log('type', this.type);
    }
    return Foo;
}());
new Foo('bar');
```

### Current Behavior
Constructor function without any assignment, so the `type` parameter is never set.

### Context
Wanting to make use of TypeScript type checking, but also (more up-to-date) babel transpilation, without having to do a 2-step-compilation pass. Every unsupported feature should at least be detected with throwing an error, and documented. Without that, people could use that unsupported code by accident without knowing, resulting in non-working code.

### Your Environment
| ---------------- | -------
| Babel            | 7.0.0-beta.35
| node             | 8.9.1
| npm              | 5.5.1 (yarn 1.3.2)
| Operating System | OSX

### LINKED ISSUE #8669: Typescript: Constructor property shorthand not perserved with @babel/preset-env
## Bug Report

**Current Behavior**
When `@babel/babel-preset-typescript` is used with `@babel/babel-preset-env`, constructor shorthand properties (e.g. `constructor(public someProp: type)`) are not preserved in the output. 

**Input Code**
- REPL or Repo link if applicable:  (I don't think the repl currently supports 7.0)

```ts
class Person {
    constructor(public name: string) {}
}
```

**Expected behavior/code**

This should compile to something like:

```js
var Person = function Person(name) {
    _classCallCheck(this, Person);
    this.name = name;
}
```

instead, it compiles to 

```js
var Person = function Person(name) {
  _classCallCheck(this, Person);
};
```

```js
{
    "presets": [
        // {modules: false}, only for cleaner output, does not affect bug
        ["@babel/preset-env", {"modules": false}],
        "@babel/preset-typescript"
    ]
}
```

**Environment**
- Babel version(s): 7.0.0
- Node/npm version: Node 8/npm 6
- OS: OSX 10.13.6
- Monorepo: ? (Not sure what this is asking)
- How you are using Babel: `cli` & `loader`

I've noticed this only happens when it's compiled with both the presets at the same time: if I compile it with the typescript preset first, then take the output and compile that with preset-env, I get the correct output.

existentialism:
| Fixed Issues?            | Fixes #8669, Fixes #7074
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -98,10 +98,6 @@ export default declare((api, { jsxPragma = "React" }) => {
         if (node.abstract) node.abstract = null;
         if (node.optional) node.optional = null;
 
-        if (node.kind !== "constructor") {
-          return;
-        }
-
         // Rest handled by Function visitor
       },
 
@@ -140,14 +136,21 @@ export default declare((api, { jsxPragma = "React" }) => {
         if (node.superTypeParameters) node.superTypeParameters = null;
         if (node.implements) node.implements = null;
 
-        // Same logic is used in babel-plugin-transform-flow-strip-types:
-        // We do this here instead of in a `ClassProperty` visitor because the class transform
-        // would transform the class before we reached the class property.
+        // Similar to the logic in `transform-flow-strip-types`, we need to
+        // handle `TSParameterProperty` and `ClassProperty` here because the
+        // class transform would transform the class, causing more specific
+        // visitors to not run.
         path.get("body.body").forEach(child => {
           const childNode = child.node;
 
           if (t.isClassMethod(childNode) && childNode.kind === "constructor") {
-            // Collect parameter properties
+            // Collects parameter properties so that we can add an assignment
+            // for each of them in the constructor body
+            //
+            // We use a WeakSet to ensure an assignment for a parameter
+            // property is only added once. This is necessary for cases like
+            // using `transform-classes`, which causes this visitor to run
+            // twice.
             const parameterProperties = [];
             for (const param of childNode.params) {
               if (
@@ -190,9 +193,9 @@ export default declare((api, { jsxPragma = "React" }) => {
                 t.isCallExpression(first.expression) &&
                 t.isSuper(first.expression.callee);
 
-              // Make sure to put parameter properties *after* the `super` call.
-              // TypeScript will enforce that a 'super()' call is the first statement
-              // when there are parameter properties.
+              // Make sure to put parameter properties *after* the `super`
+              // call. TypeScript will enforce that a 'super()' call is the
+              // first statement when there are parameter properties.
               childNode.body.body = startsWithSuperCall
                 ? [first, ...assigns, ...statements.slice(1)]
                 : [...assigns, ...statements];
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -20,6 +20,8 @@ interface State {
   programPath: any;
 }
 
+const PARSED_PARAMS = new WeakSet();
+
 export default declare((api, { jsxPragma = "React" }) => {
   api.assertVersion(7);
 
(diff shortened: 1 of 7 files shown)


packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -190,10 +144,63 @@ export default declare((api, { jsxPragma = "React" }) => {
         // We do this here instead of in a `ClassProperty` visitor because the class transform
```
Retsam: Nit: this comment is slightly out of date, now that this block contains both `ClassProperty` and `ClassMethod` logic.

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -190,10 +144,63 @@ export default declare((api, { jsxPragma = "React" }) => {
...
+
+          if (t.isClassMethod(childNode) && childNode.kind === "constructor") {
+            // Collect parameter properties
+            const parameterProperties = [];
+            for (const param of childNode.params) {
+              if (
+                param.type === "TSParameterProperty" &&
+                !PARSED_PARAMS.has(param.parameter)
+              ) {
+                PARSED_PARAMS.add(param.parameter);
```
Retsam: Mostly out of curiosity - it looks like this is the only real functional change added to this block, to prevent duplicate parameters from being added to the `parameterProperties` list?  What sort of case is this preventing?  How would two different entries of `childNode.params` have the same `parameter` object?

Retsam:
Thanks for fixing this!  

I believe this fix still depends on `transform-typescript` running before `transform-classes`, right?  I don't see any way to avoid that, but do we know if that caveat is documented anywhere?  If not, it'd be nice to have some warning about that, somewhere.

packages/babel-plugin-transform-typescript/src/index.js
existentialism: It requires a bit of understanding of the class transformation:
https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-classes/src/index.js#L41

Ultimately, it prevents params from getting parsed twice.

MayhemYDG:
@existentialism Yo, just wanted to let you know about a related issue that this PR doesn't seem to fix, judging from the repl build results: https://github.com/babel/babel/issues/8692

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -186,14 +136,74 @@ export default declare((api, { jsxPragma = "React" }) => {
...
+        // Similar to the logic in `transform-flow-strip-types`, we need to
+        // handle `TSParameterProperty` and `ClassProperty` here because the
+        // class transform would transform the class, causing more specific
+        // visitors to not run.
         path.get("body.body").forEach(child => {
-          if (child.isClassProperty()) {
-            child.node.typeAnnotation = null;
+          const childNode = child.node;
+
+          if (t.isClassMethod(childNode) && childNode.kind === "constructor") {
```
nicolo-ribaudo: Nit: you can use `t.isClassMethod(childNode, { kind: "constructor" })`

PR #8693, opened 2018-09-12
Fix es6.string.iterator mapping in babel-preset-env

### LINKED ISSUE #8691: babel/preset-env useBuiltIns: "entry" does not include "es6.string.iterator" for ie11
## Bug Report

**Current Behavior**
When using the useBuiltIns: "entry" options while targeting ie11, the "es6.string.iterator" builtin is not included in the generated imports, even though ie11 does not support it.

**Input Code**

```js
const babel = require('@babel/core')
const preset = require('@babel/preset-env')

let {code} = babel.transformSync(`
  import '@babel/polyfill'

  for (let c of 'abcdefg') {
    console.log(c)
  }
`.trim(), {
    presets: [[preset, {
        targets: [
            "ie 11"
        ],
        useBuiltIns: "entry",
        // include: ["es6.string.iterator"], // include this and the result will work on ie11
        modules: false,
        debug: true
    }]]
})

console.log(code)
```

**Expected behavior/code**
`import "core-js/modules/es6.string.iterator";` should appear in the output.

**Environment**
- Babel version(s): v7.0.0
- Node/npm version: Node 8.11.3/npm 5.10
- OS: OSX 10.13.6

Looking at babel/packages/babel-preset-env/data/built-ins.json, it seems that preset-env assumes a lot of old browsers fully support `String.prototype[@@iterator]()` even though [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/@@iterator#Browser_compatibility) says otherwise. Also, I personally tried running the generated code in ie11 and get the error: 

> Object doesn't support property or method 'Symbol(Symbol.iterator)

But when I include the `include: ["es6.string.iterator"]` option the result runs fine.

existentialism:
| Fixed Issues?            | Fixes #8691 
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-preset-env/data/built-in-features.js
+++ b/packages/babel-preset-env/data/built-in-features.js
@@ -277,7 +277,7 @@ const es = {
   "es6.string.from-code-point": "String static methods / String.fromCodePoint",
   "es6.string.includes": "String.prototype methods / String.prototype.includes",
   "es6.string.italics": "String.prototype HTML methods",
-  "es6.string.iterator": "String properties and methods / Property access on strings",
+  "es6.string.iterator": "String.prototype methods / String.prototype[Symbol.iterator]",
   "es6.string.link": "String.prototype HTML methods",
   // "String.prototype methods / String.prototype.normalize" not implemented
   "es7.string.pad-start": "String padding / String.prototype.padStart",
--- a/packages/babel-preset-env/data/built-ins.json
+++ b/packages/babel-preset-env/data/built-ins.json
@@ -1183,16 +1183,14 @@
     "electron": "1.1"
   },
   "es6.string.iterator": {
-    "chrome": "5",
-    "opera": "10.10",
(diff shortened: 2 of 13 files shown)


packages/babel-preset-env/data/built-in-features.js
```diff
@@ -277,7 +277,7 @@ const es = {
   "es6.string.from-code-point": "String static methods / String.fromCodePoint",
   "es6.string.includes": "String.prototype methods / String.prototype.includes",
   "es6.string.italics": "String.prototype HTML methods",
-  "es6.string.iterator": "String properties and methods / Property access on strings",
+  "es6.string.iterator": "String.prototype methods / String.prototype[Symbol.iterator]",
```
nicolo-ribaudo: (I hate these human-readable features name)

PR #8701, opened 2018-09-13
Fix "TypeError: comments is not iterable"

### LINKED ISSUE #8700: TypeError: comments is not iterable
## Bug Report

**Current Behavior**
When building a React Native application with Metro, `@babel/plugin-transform-flow-strip-types` throws `TypeError: comments is not iterable`. When that is fixed, `@babel/plugin-transform-react-jsx` starts throwing `TypeError: file.ast.comments is not iterable`.

An occurrence of this problem can be found here: https://github.com/react-navigation/react-navigation/issues/4983

**Input Code**
This PNG file: https://github.com/react-navigation/react-navigation/blob/22926c52303da8f329c4b130cb1bbe606eb7d74c/src/views/assets/back-icon.png

**Expected behavior/code**
Expected not to error.

```js
/* @flow */

module.exports = {
  presets: ['module:metro-react-native-babel-preset'],
  plugins: ['@babel/plugin-transform-runtime', 'relay'],
  env: {
    test: {
      plugins: ['transform-es2015-modules-commonjs'],
    },
  },
};

```

**Environment**
- Babel version(s): v7.0.0
- Node/npm version: v8.11.3
- OS: macOS 10.13.6
- Monorepo no
- How you are using Babel: Metro Bundler

**Possible Solution**
PR #8701

AlicanC:
| Fixed Issues?            | Fixes #8700 
| Documentation PR Link    | 

In some cases, `comments` is `null` and `for (const comment of comments)` throws a TypeError. I have simply wrapped `for`s with a check.

### DIFF
--- a/packages/babel-plugin-transform-flow-strip-types/src/index.js
+++ b/packages/babel-plugin-transform-flow-strip-types/src/index.js
@@ -25,15 +25,19 @@ export default declare(api => {
         skipStrip = false;
         let directiveFound = false;
 
-        for (const comment of (comments: Array<Object>)) {
-          if (comment.value.indexOf(FLOW_DIRECTIVE) >= 0) {
-            directiveFound = true;
+        if (comments) {
+          for (const comment of (comments: Array<Object>)) {
+            if (comment.value.indexOf(FLOW_DIRECTIVE) >= 0) {
+              directiveFound = true;
 
-            // remove flow directive
-            comment.value = comment.value.replace(FLOW_DIRECTIVE, "");
+              // remove flow directive
+              comment.value = comment.value.replace(FLOW_DIRECTIVE, "");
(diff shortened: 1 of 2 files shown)


xtuc:
Could you please add a test case that shows the bug?

danez:
I'm on the fence for this. I see that this could fail if comments is not there, which should never happen in the first place, as the parser always sets it to an array. The only option I see is using the Constructor for `File` in babel-types which probably sets it to `null`, but haven't tested. On the other hand I wonder why the transformations run at all if your run babel on png files. I would guess that the js parser should immediately throw as it is not javascript.

But I don't really know how metro and react native works and how it treats imported images.

danez:
Thank you.

Issue #8623, opened 2018-09-04
[TypeScript 3.0]: New tuple type support


DanielRosenwasser:
TypeScript 3.0 eased one restriction and added support for two new features in tuple types:

1. Tuples types are allowed to be empty (already works in Babel), so `[]` is a valid type.
2. Tuples elements are allowed to be optional (so each element in a tuple can have a trailing `?`).
3. Tuples are allowed to have a trailing rest element (the last element can have a leading `...`).

The following example uses all of these constructs:

```ts
declare function foo(ignored: [], ...args: [number?, ...number[]]);
```

PR #8720, opened 2018-09-17
Typescript - Tuple elements can be optional

Retsam:
| Fixed Issues?            |  #8623 (partial)
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Maybe.
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is part of #8623:

> Tuples elements are allowed to be optional (so each element in a tuple can have a trailing ?).

To implement this, I introduced a new layer in the AST: previously `TSTupleType` had an array of `TSType` nodes; now it has an array of `TSTupleElementType` nodes, which are a wrapper around a `TSType`, with information about whether it's optional and (eventually) whether it's a rest element.  So `[string, number?]` is parsed as:

```
{
    type: "TSTupleType",
    elementTypes: [{
        type: "TSTupleElementType",
        isOptional: false,
        elementType: {
            type: "TSStringKeyword"
        }
    }, {
        type: "TSTupleElementType",
        isOptional: true,
        elementType: {
            type: "TSNumberKeyword"
        }
    }]
}
```         

I'm not sure if this change to the AST would be considered a breaking change.  Alternatively, I think that information could be added onto all `TSType` nodes directly, but it seemed a bit ugly to put that sort of context-specific information onto arbitrary type nodes.


g-plane:
Why add a new type `TSTupleElementType`?

And, compared to the AST of TypeScript, I think we should add a new type called `TSOptionalType`, instead of adding property called `isOptional`.

nicolo-ribaudo:
The implementation proposed by this PR is more in line with what Babel already does (except that we use `optional` instead of `isOptional`:

```ts
interface MaybeName {
    name?: string;
}
function f(x?: number) {}
```
```json
{
  "type": "ObjectTypeProperty",
  "key": { "type": "Identifier", "name": "name" },
  "static": false,
  "proto": false,
  "kind": "init",
  "method": false,
  "value": { "type": "StringTypeAnnotation" },
  "variance": null,
  "optional": true
}

{
  "type": "Identifier",
  "name": "x",
  "optional": true,
  "typeAnnotation": {
    "type": "TypeAnnotation",
    "typeAnnotation": { "type": "NumberTypeAnnotation" }
  }
}
```

Retsam:
I've reworked this to be more inline with what @g-plane suggested.  I went down the route of `TSTupleElementType` because this syntax is a special case for tuples, but I suppose it's perhaps best if the AST doesn't reflect that.  

So the parsing logic is now a special case when parsing a tuple, but the AST is just `TSOptionalType` as suggested.

packages/babel-parser/src/types.js
```diff
@@ -1183,6 +1184,11 @@ export type TsTupleType = TsTypeBase & {
   elementTypes: $ReadOnlyArray<TsType>,
 };
 
+export type TsOptionalType = TsTypeBase & {
+  type: "TSOptionalType",
+  innerType: TsType,
```
Retsam: If anyone has a better name than "innerType", I'm all ears.

g-plane: How about just `typeAnnotation`?

Retsam: Not to be too much of a choosy beggar, I hope, but `typeAnnotation` seems odd to me.  I'd consider this whole thing to be the type annotation:

```
function foo(...args: [number, string?]) {}
                      ^^^^^^^^^^^^^^^^^
```

While I'm trying to name just this part:

```
function foo(...args: [number, string?]) {}
                               ^^^^^^
```

g-plane: Both can have the property `typeAnnotation`, and they don't have conflicts.

Retsam: I'm not saying they conflict or that this property couldn't be `typeAnnotation`, but it just doesn't feel like a very clear name, either.  

Having thought on it a bit more, I think I like `optionalType`, at least more than `innerType`.

nicolo-ribaudo:
Since we already use `typeAnnotation` for flow, I guess that it would be better to use it also for typesctipt?
https://github.com/babel/babel/blob/889cdb8da5d882cd2a83cca832f31cd4520400e0/packages/babel-types/src/definitions/flow.js#L239-L245

Retsam:
Okay; still seems like an odd convention to me, but consistency is good.  I switched it to typeAnnotation.

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -501,13 +501,24 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
       );
       return this.finishNode(node, "TSTupleType");
     }
 
+    tsParseTupleElementType(): N.TsType {
+      // Maybe necessary for optional type
+      const wrapperNode = this.startNode();
+      const type = this.tsParseType();
+      if (this.eat(tt.question)) {
+        wrapperNode.typeAnnotation = type;
```
nicolo-ribaudo: We should create `wrapperNode` inside this condition. You can use `this.startNodeAtNode(type)`

Retsam: 👍 Fixed.

Retsam:
@existentialism What tests are you looking for, specifically?  I've already added [a test with optional tuple parameters](https://github.com/babel/babel/pull/8720/files#diff-44e9b658755302d74d8f42cdfe3cbc19) to transform-typescript.  Does this cover it, or should I add another sort of test?

existentialism:
@Retsam don't mind me, somehow missed that, lol.

PR #8742, opened 2018-09-20
[decorators] Support async and generator methods

### LINKED ISSUE #8737: Async keyword removed when decorator is applied to class
## Bug Report

Hello, first let me thank the Babel contributors for your hard work. Unfortunately, I had an issue using the `@babel/plugin-proposal-decorators` plugin.

**Current Behavior**
The async keyword will be stripped of a class' method if the class has a decorator. The erroneous behavior is not reproduced while the plugin is active if the class is not decorated.

**Input Code**
Here is [a repo](https://github.com/VoidOutput/decorator-bug) with a minimal reproducible example.

Summary of the issue follows

Given the following code

```js
function Decorator ({ property } = {}) {
  return function ClassDecorator (classDescriptor) {
    console.log(property)
    return classDescriptor
  }
}

@Decorator({
  property: true
})
class GenericClass {
  async method () {
    await Promise.resolve()

    return true
  }
}
```

This code will then become (when stripped of babel internal functions):

```javascript
function Decorator({
  property
} = {}) {
  return function ClassDecorator(classDescriptor) {
    console.log(property);
    return classDescriptor;
  };
}

let GenericClass = _decorate([Decorator({
  property: true
})], function (_initialize) {
  class GenericClass {
    constructor() {
      _initialize(this);
    }

  }

  return {
    F: GenericClass,
    d: [{
      kind: "method",
      key: "method",

      value() {
        await Promise.resolve();
        return true;
      }

    }]
  };
});
```

Note the lack of an async keyword on the `value` function. This causes browser or linters to correctly point out the syntax error.

**Expected behavior/code**

```javascript
  // ...

  return {
    F: GenericClass,
    d: [{
      kind: "method",
      key: "method",

      async value() {
        await Promise.resolve();
        return true;
      }

    }]
  };

  // ...
```

or maybe

```javascript
  // ...

  return {
    F: GenericClass,
    d: [{
      kind: "method",
      key: "method",

      value() {
        return (async () => {
          await Promise.resolve();
          return true;
        })();
      }

    }]
  };

  // ...
```

```js
{
  "plugins": [
    [ "@babel/plugin-proposal-decorators", { "decoratorsBeforeExport": true } ]
  ]
}
```

**Environment**
- Babel version: v7.1.0
- Node/npm version: Node 10/NPM 6
- OS: Windows 10
- Monorepo: yes
- How you are using Babel: `cli`, `loader`

Note: This issue was first discovered using the loader along with Webpack but was reproduced later with the CLI.

**Additional context**

I have tried to search for similar issues on Google and this repo's issue tracker but could not find someone with the same experience. My apologies if this is already known and/or was already posted.

Please do not hesitate to ask if you need more details.

Thanks in advance for your time. Have a nice day.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #8737
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-proposal-decorators/src/transformer.js
+++ b/packages/babel-plugin-proposal-decorators/src/transformer.js
@@ -7,8 +7,11 @@ function prop(key, value) {
   return t.objectProperty(t.identifier(key), value);
 }
 
-function value(body, params = []) {
-  return t.objectMethod("method", t.identifier("value"), params, body);
+function value(body, params = [], async, generator) {
+  const method = t.objectMethod("method", t.identifier("value"), params, body);
+  method.async = !!async;
+  method.generator = !!generator;
+  return method;
 }
 
 function hasDecorators({ node }) {
@@ -77,7 +80,7 @@ function getSingleElementDefinition(path, superRef, classRef, file) {
     prop("static", node.static && t.booleanLiteral(true)),
     prop("key", getKey(node)),
     isMethod
-      ? value(node.body, node.params)
+      ? value(node.body, node.params, node.async, node.generator)
       : node.value
         ? value(template.ast`{ return ${node.value} }`)
         : prop("value", scope.buildUndefinedNode()),
--- a/packages/babel-plugin-proposal-decorators/test/fixtures/transformation/async-generator-method/input.js
+++ b/packages/babel-plugin-proposal-decorators/test/fixtures/transformation/async-generator-method/input.js
@@ -1,0 +1,8 @@
+@decorator
+class Foo {
+  async f1() {}
+
+  *f2() {}
+
+  async *f3() {}
+}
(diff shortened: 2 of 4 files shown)


packages/babel-plugin-proposal-decorators/src/transformer.js
```diff
@@ -7,8 +7,11 @@ function prop(key, value) {
   return t.objectProperty(t.identifier(key), value);
 }
 
-function value(body, params = []) {
-  return t.objectMethod("method", t.identifier("value"), params, body);
+function value(body, params = [], async, generator) {
+  const method = t.objectMethod("method", t.identifier("value"), params, body);
```
hzoo: I guess this is because we don't require you do to async/generator? A weird bug just since we don't pass it down but not sure what we would want to do in general for it

nicolo-ribaudo: `t.functionDeclaration` accepts async/generator but `t.objectMethod` doesn't.
I'd love to have a function to specify the fields I want, like
```js
t.objectMethod.with({
  kind: "method",
  key: t.identifier("value"),
  params, body, async, generator
});
```

PR #8743, opened 2018-09-20
Use toPropertyKey in the "decorate" helper, delegate to @@toPrimitive and set correct methods namee

### LINKED ISSUE #8740: New decorators: ToElementDescriptor doesn't follow the spec
## Bug Report

**Current Behavior**

`_toElementDescriptor` at [line 1578](https://github.com/babel/babel/blob/v7.1.0/packages/babel-helpers/src/helpers.js#L1578) coerces a value to string using `String(key)`.

**Input Code**

```js
// ToPrimitive should return `Symbol('k')`
const k = {
    [Symbol.toPrimitive]() {
        return Symbol('k');
    },
};

// Example decorator
const d = () => {};

// Decorated class with symbol-coerced method key
@d
class C {
    [k]() {}
}
```

Whoops, `TypeError: Cannot convert a Symbol value to a string`.

**Expected behavior/code**

The `key` should be coerced following the [spec's `ToPrimitive` operation](https://tc39.github.io/ecma262/#sec-toprimitive) (See [`ToElementDescriptor` definition, step 5](https://tc39.github.io/proposal-decorators/#sec-to-element-descriptor)).

**Babel Configuration**

```js
module.exports = {
    presets: [['@babel/env', { targets: { node: true } }]],
    plugins: [
        ['@babel/plugin-proposal-decorators', { decoratorsBeforeExport: true }],
    ],
};
```

**Environment**
- Babel version: v7.1.0.
- Node/npm version: Node v10.10.0, npm v6.4.1, yarn v1.9.4.
- OS: Windows 10.
- Monorepo: no.
- How you are using Babel: `cli`, `babel-jest`.

**Possible Solution**
Add spec-compliant `_toPrimitive` function to the helpers and use it in `_toElementDescriptor` function.

Possible `_toPrimitive` definition:

```js
function _toPrimitive(
  input: any,
  hint: "default" | "string" | "number" = "default"
): string | number | symbol | bigint | boolean | undefined {
  if (typeof input === "object") {
    // ↑ What about `null`?
    if (typeof input[Symbol.toPrimitive] === "function") {
      // ↑ Well, this doesn't follow `GetMethod` spec, but it should work.
      let result = input[Symbol.toPrimitive](hint);

      if (typeof result !== "object") {
        // ↑ What about `null`, again?
        return result;
      } else {
        throw new TypeError(
          "Blah-blah, @@toPrimitive method should return a primitive value."
        );
      }

      hint = hint === "default" ? "number" : hint;

      // OrdinaryToPrimitive
      // ↓ Well, IDK how much `String` and `Number` functions compatible with this spec.
      // ↓ But it should work.
      if (hint === "string") {
        return String(input);
      } else {
        return Number(input);
      }
    }
  }

  return input;
}
```

Then, replace [line 1578](https://github.com/babel/babel/blob/v7.1.0/packages/babel-helpers/src/helpers.js#L1578) with:

```js
    if (typeof key !== "string" && typeof key !== "symbol") key = _toPrimitive (key);
```

### LINKED ISSUE #8746: New decorators: Methods' `name` property is always `value`
## Bug Report

**Current Behavior**

In any decorated class all methods have `name` property equals to `'value'`.

**Input Code**

```js
const decorator = d => d;

@decorator
class A {
    f() {}
}

const a = new A();

console.log(a.f.name); // 'value'
```

**Expected behavior/code**

`a.f.name` must be `'f'`.

**Babel Configuration**

```js
module.exports = {
    presets: [['@babel/env', { targets: { node: true } }]],
    plugins: [
        ['@babel/plugin-proposal-decorators', { decoratorsBeforeExport: true }],
    ],
};
```

**Environment**
- Babel version: v7.1.0.
- Node/npm version: Node v10.10.0, npm v6.4.1, yarn v1.9.4.
- OS: Windows 10.
- Monorepo: no.
- How you are using Babel: `cli`, `babel-jest`.

**Possible Solution**

`_createElementDescriptor` should define `descriptor.value.name` after line [1229](https://github.com/babel/babel/blob/v7.1.0/packages/babel-helpers/src/helpers.js#L1229):

```js
Object.defineProperty(descriptor.value, 'name', {
  writable: false,
  enumerable: false,
  configurable: true,
  value: typeof def.key === 'symbol' ? '' : String(def.key),
});
```

Or maybe at some other place, idk.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #8740, fixes #8746
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR probably also fixes some bugs with object rest and `@@toPrimitive`.

cc @miyaokamarina Thank you for the detailed investigation in the issue! I used `toPropertyKey` instead of `toPrimitive` because I merged steps 5 and 7 of the `ToElementDescriptor` operation (step 6 is not supported yet).

### DIFF
--- a/packages/babel-core/test/fixtures/transformation/misc/regression-4855/options.json
+++ b/packages/babel-core/test/fixtures/transformation/misc/regression-4855/options.json
@@ -1,5 +1,8 @@
 {
   "compact": false,
   "presets": ["env"],
-  "plugins": ["external-helpers", "proposal-object-rest-spread"]
+  "plugins": [
+    ["external-helpers", { "helperVersion": "7.1.5" }],
+    "proposal-object-rest-spread"
+  ]
 }
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -935,13 +935,28 @@ helpers.skipFirstGeneratorNext = helper("7.0.0-beta.0")`
   }
 `;
 
-helpers.toPropertyKey = helper("7.0.0-beta.0")`
-  export default function _toPropertyKey(key) {
-    if (typeof key === "symbol") {
-      return key;
-    } else {
-      return String(key);
+helpers.toPrimitive = helper("7.1.5")`
+  export default function _toPrimitive(
+    input,
+    hint /*: "default" | "string" | "number" | void */
+  ) {
+    if (typeof input !== "object" || input === null) return input;
+    var prim = input[Symbol.toPrimitive];
+    if (prim !== undefined) {
+      var res = prim.call(input, hint || "default");
+      if (typeof res !== "object") return res;
+      throw new TypeError("@@toPrimitive must return a primitive value.");
     }
+    return (hint === "string" ? String : Number)(input);
+  }
+`;
+
+helpers.toPropertyKey = helper("7.1.5")`
+  import toPrimitive from "toPrimitive";
+
+  export default function _toPropertyKey(arg) {
+    var key = toPrimitive(arg, "string");
+    return typeof key === "symbol" ? key : String(key);
   }
 `;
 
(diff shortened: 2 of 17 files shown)


pabloalmunia:
@nicolo-ribaudo congratulations, this fix work perfectly. We have applied this commit to our local copy of Babel and our entire project is working correctly. Before this commit we had 31 errors from 362 test about our decorator library.

PR #8751, opened 2018-09-22
Fix some missing parens cases with OptionalMemberExpression in generator

### LINKED ISSUE #8733: @babel/generator: Incorrect parens in scenario involving OptionalMemberExpression
## Bug Report

**Current Behavior**
As of 7.1.0, `@babel/generator` is omitting parentheses in some situations when generating code involving untransformed `OptionalMemberExpression`s.

**Input Code**
```js
(a ? b : c)?.d
```

**Expected behavior/code**
Should output the same thing:
```js
(a ? b : c)?.d
```
Instead outputs
```js
a ? b : c?.d
```

Optional chaining syntax enabled, but not transformed.

**Environment**
- Babel version(s): 7.1.0
- Node/npm version: 8.12.0
- OS: OS X 10.13.6
- Monorepo no

**Possible Solution**
I don't fully understand the logic of the parenthesis algorithm, but my best guess is that `OptionalMemberExpression` needs to be added to `UnaryLike` here: https://github.com/babel/babel/blob/master/packages/babel-generator/src/node/parentheses.js#L186

existentialism:
| Fixed Issues?            | Fixes #8733
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-generator/src/node/parentheses.js
```diff
@@ -219,6 +220,13 @@ export function ConditionalExpression(node: Object, parent: Object): boolean {
   return UnaryLike(node, parent);
 }
 
+export function OptionalMemberExpression(
+  node: Object,
+  parent: Object,
+): boolean {
+  return t.isMemberExpression(parent);
```
nicolo-ribaudo: Aren't parentheses also needed when the parent is a `CallExpression`? (`(a?.b)()`)

existentialism: Updated!

PR #8755, opened 2018-09-23
TypeScript: reserve `unknown` as TSUnknownKeyword

g-plane:
| Documentation PR Link    | None

`unknown` is reserved as a keyword since TypeScript 3.0, so I add it. The AST node name is `TSUnknownKeyword`.

### DIFF
--- a/packages/babel-generator/src/generators/typescript.js
+++ b/packages/babel-generator/src/generators/typescript.js
@@ -127,6 +127,9 @@ export function TSIndexSignature(node) {
 export function TSAnyKeyword() {
   this.word("any");
 }
+export function TSUnknownKeyword() {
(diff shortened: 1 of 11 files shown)


g-plane:
Previously, the babel parser will treat `unknown` as an identifier whose name is `unknown`, and now it will be `TSUnknownKeyword`. That's, other identifiers won't be affected. IMHO, it won't be a breaking change for end users. Also, it will be OK for most libraries authors, unless they use the identifier `unknown` to do something, because the `unknown` will be *always* parsed as a keyword, not an indentifier.

xtuc:
Doesn't look like a breaking, apart if people use their own `unknown` type I guess? But that's also the case in tsc.

PR #8756, opened 2018-09-23
class private methods and properties: should not allow spaces between # and identifier

### LINKED ISSUE #8225: Parser allows spaces between # and identifier
## Bug Report

**Current Behavior**
First identified in: [Class Fields (Stage 3) #12](https://github.com/babel/proposals/issues/12#issuecomment-398758818)
The parser currently allows spacing between the private name symbol `#` and the identifier. 

However, it seems to violate the [lexical grammar](https://tc39.github.io/proposal-class-fields/#sec-new-syntax) in the spec proposal:
```
PrivateName:: #IdentifierName
```

**Input Code**
```js
class Foo {
    #    p = 0;
    constructor() {
        this.#   p   ++;
    }

    print() {
        console.log(this   .   #    p);
    }
}
```

**Expected behavior/code**
Babel should communicate the parser error, e.g: "Spaces not permitted in private field name"

```js
const presets = [
  ["@babel/env", {
    targets: {
      node: "6"
    },
    useBuiltIns: "usage"
  }]
];

module.exports = { presets,
    "plugins": ["@babel/plugin-proposal-class-properties"]
 };
```

**Environment**
- Babel version(s): [v7.0.0-beta.34]
- Node/npm version: [e.g. Node 6/npm 3.10.10]
- OS: [Windows 7]
- Monorepo [no]
- How you are using Babel: [cli]

CC'ing @rricard @mheiber and @littledan who were part of the conversation regarding this in  [Class Fields (Stage 3) #12](https://github.com/babel/proposals/issues/12#issuecomment-398758818).

I would also like to volunteer to fix this.

macabeus:
Following https://github.com/babel/proposals/issues/12#issuecomment-399756043, we should not allow spaces between `#` and identifier.
But, since we currently allow it, this fix is a breaking change.

| Fixed Issues?            | Fixes #8225
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -923,7 +923,19 @@ export default class ExpressionParser extends LValParser {
     if (isPrivate) {
       this.expectOnePlugin(["classPrivateProperties", "classPrivateMethods"]);
       const node = this.startNode();
+      const columnHashEnd = this.state.end;
       this.next();
+      const columnIdentifierStart = this.state.start;
+
+      const spacesBetweenHashAndIdentifier =
+        columnIdentifierStart - columnHashEnd;
+      if (spacesBetweenHashAndIdentifier != 0) {
+        this.raise(
+          columnIdentifierStart,
+          "Unexpected space between # and identifier",
+        );
+      }
+
       node.id = this.parseIdentifier(true);
       return this.finishNode(node, "PrivateName");
     } else {
--- a/packages/babel-parser/test/fixtures/experimental/class-private-methods/failure-spaces/input.js
+++ b/packages/babel-parser/test/fixtures/experimental/class-private-methods/failure-spaces/input.js
@@ -1,0 +1,5 @@
+class Spaces {
+  #  wrongSpaces() {
+    return fail();
+  }
+}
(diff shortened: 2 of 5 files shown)


macabeus:
@nicolo-ribaudo I made a rebase with a new commit fixing the problem that you said. Thank you.

Since the spaces are skipped at tokenizer, then the identifier doesn't have the spaces, I need to use others informations at node to check it.

loganfsmyth:
@nicolo-ribaudo Do you have a link for that?

packages/babel-parser/src/parser/expression.js
```diff
@@ -925,6 +925,14 @@ export default class ExpressionParser extends LValParser {
       const node = this.startNode();
       this.next();
```
loganfsmyth: I _think_ you can save `this.state.pos` (or something like that) just before this and just after this and compare them, rather than working with the AST nodes.

loganfsmyth: Actually, save `this.state.end` and then check `this.state.start` after calling `next()`.

nicolo-ribaudo: @macabeus Sorry for not giving any information. Logan's suggestion is good.

macabeus:
@loganfsmyth @nicolo-ribaudo Thank you. I made a new rebase following your suggestion, using `this.state.pos` before `next()` and `this.state.start` after it.

packages/babel-parser/src/parser/expression.js
macabeus: @nicolo-ribaudo No problem 😋

hzoo:
awesome work @macabeus!

PR #8768, opened 2018-09-25
Use babel-check-duplicated-nodes

nicolo-ribaudo:
The code that I removed in this PR was extracted by @Andarist so that it can aldo be used by regenerator. `babel-check-duplicated-nodes` contains exactly the same code.


Andarist:
Maybe we should move babel-check-duplicated-nodes to babel org or even to the monorepo? The reason why I have extracted it was that at the time it has allowed duplicated nodes in regenerator output and I was fixing exactly this - duplicated nodes in regenerator, so I have needed a version without regenerator-specific lines.

PR #8769, opened 2018-09-25
Add plugins name

nicolo-ribaudo:
I recall that they where needed for some repl feature (probably time travel).

Do you prefer them to be named `transform-classes` or `@babel/plugin-transform-classes`?

<details>
<summary>I generated them using this script</summary>

```js
"use strict";

const recast = require("recast");
const babel = require("@babel/core");
const fs = require("fs");
const path = require("path");

const dirs = fs
  .readdirSync("./packages/")
  .filter(/./.test.bind(/babel-plugin-(?:proposal|transform|syntax)/));

const notUpdated = [];

for (const dir of dirs) {
  try {
    const name = dir.replace("babel-plugin-", "");
    const filename = path.resolve(`./packages/${dir}/src/index.js`);
    let content = fs.readFileSync(filename, "utf-8");
    content = babelRecast(
      content,
      { filename },
      {
        configFile: false,
        plugins: [[addName, { name }]],
      }
    );
    fs.writeFileSync(filename, content);
  } catch (e) {
    notUpdated.push([dir, e]);
  }
}

console.log(notUpdated);

function addName({ types: t }, { name }) {
  const ok = new WeakSet();

  function hasProp(node, name) {
    return node.properties.some(prop => t.isIdentifier(prop.key, { name }));
  }

  function program(path) {
    return path.find(p => p.isProgram());
  }

  return {
    visitor: {
      ObjectExpression(path) {
        if (
          !hasProp(path.node, "visitor") &&
          !hasProp(path.node, "manipulateOptions")
        )
          return;

        try {
          if (hasProp(path.node, "name") || ok.has(program(path).node)) {
            return;
          }

          path.node.properties.unshift(
            t.objectProperty(t.identifier("name"), t.stringLiteral(name))
          );
        } finally {
          ok.add(program(path).node, true);
        }
      },
      Program: {
        exit(path) {
          if (!ok.has(path.node)) throw "NOT FOUND";
        },
      },
    },
  };
}

function babelRecast(code, parserOpts, transformOpts) {
  const ast = recast.parse(code, {
    parser: {
      parse: source => babel.parse(source, parserOpts),
    },
  });

  const opts = Object.assign(
    {
      ast: true,
      code: false,
    },
    transformOpts,
    {
      plugins: [
        // For some reason, recast doesn't work with transformFromAst.
        // Use this hack instead.
        () => ({
          visitor: {
            Program(path) {
              path.replaceWith(ast.program);
            },
          },
        }),
      ].concat(transformOpts.plugins || []),
    }
  );

  const output = babel.transformSync("", opts);

  return recast.print(output.ast).code + "\n";
}
```

```
babel on  master [⇣$!+] is 📦 ⚠ via ⬢ v10.10.0 took 3s 
➜ node add-names.js && yarn fix
@babel/preset-env: `DEBUG` option

Using targets:
{
  "node": "10.10"
}

Using modules transform: false

Using plugins:
  proposal-async-generator-functions { "node":"10.10" }
  syntax-object-rest-spread { "node":"10.10" }
  proposal-unicode-property-regex { "node":"10.10" }
  proposal-json-strings { "node":"10.10" }
  proposal-optional-catch-binding { "node":"10.10" }

Using polyfills: No polyfills were added, since the `useBuiltIns` option was not set.
[ [ 'babel-plugin-transform-regenerator', 'NOT FOUND' ] ]
yarn run v1.9.4
$ make fix
./node_modules/.bin/prettier "{packages,codemod}/*/test/fixtures/**/options.json" --write --loglevel warn
./node_modules/.bin/eslint scripts packages codemods '*.js' --format=codeframe --fix
Done in 35.58s.
```
</details>


xtuc:
Very nice, maybe we could keep the script arround? Could you please commit it in the scripts dir?

nicolo-ribaudo:
Good idea :+1:

I think that I will implement it as a linting rule, since the tests would have to be manually enabled for any new package (so it is as easy to forget it as it is to add the plugin name)

nicolo-ribaudo:
ESLint plugin added!

nicolo-ribaudo:
Now Time Travel in the repl shows the correct plugins name :smile:

PR #8787, opened 2018-09-30
jsx: fix sequence expression at JSXAttributeValue

### LINKED ISSUE #8604: Comma expression (Sequence expression) incorrectly allowed as JSXAttributeValue
## Bug Report

**Current Behavior**

Code like `<div>{3, 1}</div>` is not treated as a parse error, but it should be.

[REPL link](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=DwEwlgbgfA3gzAGgAQEYC-wD05oG4BQokSAZgPZkC8MKyATGkplAUA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=es2015%2Ces2016%2Ces2017%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3%2Ces2015-loose&prettier=false&targets=&version=6.26.0&envVersion=1.6.2)

**Input Code**
```js
// No error
<div>{3, 1}</div>;
// No error
<div foo={1, 2} />;
```

**Expected behavior/code**

Expected a parse error.

The [JSX Spec](https://facebook.github.io/jsx/) specifies `JSXChildExpression` and `JSXAttributeValue` as being `{` *AssignmentExpression* `}`, but the comma expression is [above `AssignmentExpression` in the grammar](https://www.ecma-international.org/ecma-262/5.1/#sec-11.14)

**Environment**
- Babel version(s): 6.26.0
- Node/npm version: N/A (repl)
- OS: N/A (repl)
- Monorepo N/A (repl)
- How you are using Babel: N/A (repl)

**Possible Solution**

https://stackoverflow.com/questions/52139174/comma-operator-in-typescript-jsx-expressions

macabeus:
Following https://github.com/babel/babel/issues/8604, we should not allow codes like these
```js
<div key={console.log('foo'), JSON.stringify(props)} />
<div>{console.log('foo'), JSON.stringify(props)}</div>
```

But we should allow codes like
```js
<div key={(console.log('foo'), JSON.stringify(props))} />
<div>{(console.log('foo'), JSON.stringify(props))}</div>
```

This PR fixes it, raising an exception with this message: `Sequence of values at JSX must be parenthesized.`
But, since we currently allow the wrong code, this fix is a breaking change.

| Fixed Issues?            | Fixes #8604
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/src/plugins/jsx/index.js
+++ b/packages/babel-parser/src/plugins/jsx/index.js
@@ -327,8 +327,22 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         node.expression = this.jsxParseEmptyExpression();
       } else {
         node.expression = this.parseExpression();
+
+        if (
+          node.expression.type === "SequenceExpression" &&
+          ((node.expression.extra &&
+            node.expression.extra.parenthesized === false) ||
+            node.expression.extra === undefined)
+        ) {
+          this.raise(
+            this.state.start,
+            "Sequence of values at JSX must be parenthesized.",
+          );
+        }
       }
+
       this.expect(tt.braceR);
+
       return this.finishNode(node, "JSXExpressionContainer");
     }
 
--- a/packages/babel-parser/src/plugins/jsx/index.js
+++ b/packages/babel-parser/src/plugins/jsx/index.js
@@ -330,9 +330,8 @@ export default (superClass: Class<Parser>): Class<Parser> =>
 
         if (
           node.expression.type === "SequenceExpression" &&
-          ((node.expression.extra &&
-            node.expression.extra.parenthesized === false) ||
-            node.expression.extra === undefined)
+          (!node.expression.extra ||
+            node.expression.extra.parenthesized !== true)
         ) {
           this.raise(
             this.state.start,
--- a/packages/babel-parser/src/plugins/jsx/index.js
+++ b/packages/babel-parser/src/plugins/jsx/index.js
@@ -326,20 +326,8 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       if (this.match(tt.braceR)) {
         node.expression = this.jsxParseEmptyExpression();
       } else {
-        node.expression = this.parseExpression();
-
-        if (
-          node.expression.type === "SequenceExpression" &&
-          (!node.expression.extra ||
-            node.expression.extra.parenthesized !== true)
-        ) {
-          this.raise(
-            this.state.start,
-            "Sequence of values at JSX must be parenthesized.",
-          );
-        }
+        node.expression = this.parseMaybeAssign();
       }
-
       this.expect(tt.braceR);
 
       return this.finishNode(node, "JSXExpressionContainer");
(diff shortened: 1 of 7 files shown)


nicolo-ribaudo:
Instead of parsing an Expression and then throw if it is a SequenceExpression, we can parse an AssignmentExpression since the beginning.

macabeus:
@nicolo-ribaudo Where is the `AssignmentExpression` at JSX plugin? I tried to find it, but I can't.

nicolo-ribaudo:
You can use `this.parseMaybeAssign` (defined it `src/parser/expression.js`) instead of `this.parseExpression` at line 329.

macabeus:
@nicolo-ribaudo I updated the code 😄

packages/babel-parser/src/plugins/jsx/index.js
```diff
@@ -327,8 +327,22 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         node.expression = this.jsxParseEmptyExpression();
       } else {
         node.expression = this.parseExpression();
+
+        if (
+          node.expression.type === "SequenceExpression" &&
+          ((node.expression.extra &&
+            node.expression.extra.parenthesized === false) ||
+            node.expression.extra === undefined)
```
danez: You could do this here, which is a little bit simpler:

```
node.expression.type === "SequenceExpression" &&
          (!node.expression.extra ||
            node.expression.extra.parenthesized !== true)
```
I also used `!== true`, as extra might be an object but parenthesized could be undefined.

packages/babel-parser/src/plugins/jsx/index.js
```diff
@@ -327,8 +327,21 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         node.expression = this.jsxParseEmptyExpression();
       } else {
         node.expression = this.parseExpression();
```
nicolo-ribaudo: Is there a reason to check if the expression is parenthesized instead of using `parseMaybeAssign` (which already disallow comma expressions) here?

danez: 👍

nicolo-ribaudo:
Since in a past meating we had concerns that this PR could break someone's code, here is a summary of what other parsers do when parsing `<div>{console.log('foo'), JSON.stringify(props)}</div>`:

  | Parser | Parsed? | Error
  |--------|---------|-------
  | Acorn | :heavy_check_mark: 
  | TypeScript | :x: | `'}' expected.`
  | Flow | :heavy_check_mark: |
  | Espree (Uses acorn) | :heavy_check_mark: |
  | Esprima | :x: | `Unexpected token ,`

Bonus: Pretter prints it as `<div>{(console.log('foo'), JSON.stringify(props))}</div>`

macabeus:
Sorry for not being at the meeting. I'm very busy at last days 😞

@xtuc What were the outputs of the meeting?

nicolo-ribaudo:
For now we are not merging this PR since, as shown in https://github.com/babel/babel/pull/8787#issuecomment-439726230, many tools allow sequence expression and thus some people may rely on it.
We will merge this in Babel 8, and in the meantime we will investigate about an opt-in flag to enable warnings about "things that will break in the next major release".

nicolo-ribaudo:

PR #8791, opened 2018-10-01
types: allow jsxEmptyExpression inside jsxExpressionContainer

### LINKED ISSUE #8549: types: JSXExpressionContainer does not allow `expression` to be of type `JSXEmptyExpression`
## Bug Report

**Current Behavior**
When constructing part of an AST that represents a comment inside JSX, the JSXExpressionContainer constructor does not allow the `expression` argument to be of type `JSXEmptyExpression`. This is however, as far as I'm aware, the default way to put comments into JSX (attaching them to a JSXEmptyExpression inside a JSXExpressionContainer).

**Input Code**

```js
const emptyExpression = types.jsxEmptyExpression();
emptyExpression.leadingComments = [{ type: 'BlockComment', value: 'Hello there' }];
types.jsxExpressionContainer(emptyExpression);
```

**Expected behavior/code**
I expect to be able to construct a JSXExpressionContainer holding a JSXEmptyExpression with comments attached. Something like this:

```js
<div>
  {/* Hello there */}
</div>
```

Don't think that's relevant, as this is an issue with @babel/types

**Environment**
- Babel version(s): v7.0.0-rc.3
- Node/npm version: Node 8.11.3, npm 5.6.0, yarn 1.7.0
- OS: macOS 10.12.6
- Monorepo: yes, Lerna
- How you are using Babel: API, @babel/types

It used to work in Babel 6.x

**Possible Solution**
I assume that this commit is causing the issue:
https://github.com/babel/babel/commit/f0e46772a6140ce74fda22fa29fd862ec8060887

I understand the reason for this commit, but I believe that the regression I'm seeing was not intended. A solution could be to change `validate: assertNodeType("Expression"),` to `validate: assertNodeType("Expression", "JSXEmptyExpression"),` in the type definition for `JSXExpressionContainer`.

tvooo:
This fixes #8549.

This now allows to create a JSXEmptyExpression inside a JSXExpressionContainer, which allows you to construct block comments inside JSX:

```js
<div>{ /* I'm a comment */ }</div>
```

The JSXElement child can be created using:

```js
const expr = types.jsxEmptyExpression();
types.addComment(expr, "leading", " I'm a comment ");
types.jsxExpressionContainer(emptyExpression);
```

| Fixed Issues?            | Fixes #8549 
| Tests Added + Pass?      | Nothing added
| Documentation PR Link    | 

### DIFF
--- a/packages/babel-types/src/definitions/jsx.js
+++ b/packages/babel-types/src/definitions/jsx.js
@@ -73,7 +73,7 @@ defineType("JSXExpressionContainer", {
   aliases: ["JSX", "Immutable"],
   fields: {
     expression: {
-      validate: assertNodeType("Expression"),
+      validate: assertNodeType("Expression", "JSXEmptyExpression"),
     },
   },
 });


nicolo-ribaudo:
Why do we need `JSXEmptyExpression`? Can't we just make the `.expression` field optional?

tvooo:
As far as I see it, we nee `JSXEmptyExpression` in order to put a comment _inside_ the `JSXExpressionContainer`. We need something to attach the comment to. Instead, if we call `addComment` with the first argument being the _container_, the comment will end up before or after the container, not inside. Let me know if I missed something here.

tvooo:
I tried running `make build`, but it does not update any files for me. Running this in the repository root. Anything else I'm missing?

hzoo:
awesome, thanks

PR #8792, opened 2018-10-01
Fix perf issue in typescript parser plugin

### LINKED ISSUE #8779: Performance issues with typescript parser
Configuring the the parser with the typescript plugin causes a significant performance regression. i.e 

```
// this
parser.parse('var foo = 1', {plugins: ['typescript']});
// is much, much slower than this:
parser.parse('var foo = 1');
```

I set up a test case that demonstrates the issue in a repo here: https://github.com/matthewrobertson/babel-parser-perf/blob/master/index.js

On my machine (windows server, node v8.11.1, @babel/parser v7.1.0) I see the following results:

```
Normal parse: 203.346ms
Typescript parse: 13404.139ms
```

This indicates the parser is 80x slower when the typescript plugin is added. I don't see a similar perf regression when I add the flow plugin. Any idea what might be causing this?

matthewrobertson:
| Fixed Issues?            | Fixes #8779 

This commit fixes a perf regression introduced by 8ee24fd. Throwing an
exception in `tsTryParseAndCatch` is expensive. We can avoid it in the
common case by moving the conditional outside of the callback passed
to `tsTryParseAndCatch`.

### DIFF
--- a/packages/babel-parser/src/plugins/typescript.js
+++ b/packages/babel-parser/src/plugins/typescript.js
@@ -1368,10 +1368,11 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         return this.finishNode(nonNullExpression, "TSNonNullExpression");
       }
 
-      // There are number of things we are going to "maybe" parse, like type arguments on
-      // tagged template expressions. If any of them fail, walk it back and continue.
-      const result = this.tsTryParseAndCatch(() => {
-        if (this.isRelational("<")) {
+      if (this.isRelational("<")) {
+        // tsTryParseAndCatch is expensive, so avoid if not necessary.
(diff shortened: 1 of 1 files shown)


matthewrobertson:
I set up a little demo to profile the typescript parser here: https://github.com/matthewrobertson/babel-parser-perf

Before this change adding the typescript plugin made the parser much slower:
```
> node ./index.js ./some-code-to-parse.js

Normal parse: 169.068ms
Typescript parse: 10823.573ms
```

After this change parsing takes close to the same amount of time:
```
> node ./index.js ./some-code-to-parse.js

Normal parse: 155.450ms
Typescript parse: 227.296ms
```

existentialism:
@matthewrobertson thanks!

PR #8793, opened 2018-10-01
Ensure destructuring's computed key handling matches object-rest-spread

### LINKED ISSUE #8785: Issue with using numbers as keys; object destructuring, rest properties, computed property names
## Bug Report

**Current Behavior**

This was discovered a few days ago. We were using three “newer” JavaScript features:
- Object destructuring
- Rest properties
- Computed property names

The following line (from the block below):

`const { [person.id]: excludedPerson, ...includedPeople } = samePeople;`

... will pick off the value and place it in `excludedPerson`, but fail to exclude that entry in `includedPeople`. This is only the case if `person.id` is passed in as a `number`, but if it is cast to a `string` then that property is (correctly) omitted in `includedPeople`.

**Input Code**

```js
const people = {
  21: {
    id: 21,
    name: 'Michelle Rodriguez',
  },
  22: {
    id: 22,
    name: 'Heath Ledger',
  },
  23: {
    id: 23,
    name: 'Jack the Ripper',
  },
};

const id = 21;
const person = people[id];
const samePeople = { ...people };

const { [person.id]: excludedPerson, ...includedPeople } = samePeople;

console.log(excludedPerson, includedPeople);
```

**Expected behavior/code**

It would be expected that that the Michelle Rodriguez object would be omitted from `includedPeople` but it is still included in `includedPeople`.

This can be replicated here: https://babeljs.io/repl

**Environment**

This can be replicated here: https://babeljs.io/repl

**Possible Solution**

Babel spits out the following code in order to omit properties from the `rest` object:

```js
function _objectWithoutProperties(obj, keys) {
  var target = {};
  for (var i in obj) {
    if (keys.indexOf(i) >= 0) continue;

    // Excludes properties on the prototype chain.
    if (!Object.prototype.hasOwnProperty.call(obj, i)) continue;
    target[i] = obj[i];
  }
  return target;
}
```

If `person.id` is passed as a number (i.e. just as it is), then the `keys.indexOf(i) >= 0` check fails, since the `for..in` loop will return the keys as `strings`. The `keys` parameter will be an `Array` of just one `number` (as per the above code), e.g. `'21'` will not be found inside `[21]`. Therefore the property does not get excluded from `target` and ends up in the `rest` object.

But if I cast to a `string`, e.g. `` `${person.id}` ``, `String(person.id)`, `person.id + ''`, `Number(person.id).toString()`, then the condition goes through.

We ended up going with Lodash's `omit` in order to omit the chosen property, but a fix that worked was simply casting `person.id` to a `string`, e.g.

```js
const id = 21;
const person = people[id];
const samePeople = { ...people };

const personId = `${person.id}`;

const { [personId]: excludedPerson, ...includedPeople } = samePeople;
```

At that point the omission works correctly.

The code seems to work correctly in the Chrome Console and the Node.js REPL, but fails when Babel transpiles it.

existentialism:
| Fixed Issues?            | Fixes #8785
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-transform-destructuring/src/index.js
+++ b/packages/babel-plugin-transform-destructuring/src/index.js
@@ -170,7 +170,8 @@ export default declare((api, options) => {
     pushObjectRest(pattern, objRef, spreadProp, spreadPropIndex) {
       // get all the keys that appear in this object before the current spread
 
-      let keys = [];
+      const keys = [];
+      let allLiteral = true;
 
       for (let i = 0; i < pattern.properties.length; i++) {
         const prop = pattern.properties[i];
@@ -182,11 +183,15 @@ export default declare((api, options) => {
         // ignore other spread properties
         if (t.isRestElement(prop)) continue;
 
-        let key = prop.key;
+        const key = prop.key;
         if (t.isIdentifier(key) && !prop.computed) {
-          key = t.stringLiteral(prop.key.name);
+          keys.push(t.stringLiteral(key.name));
+        } else if (t.isLiteral(key)) {
+          keys.push(t.stringLiteral(String(key.value)));
+        } else {
+          keys.push(t.cloneNode(key));
+          allLiteral = false;
         }
-        keys.push(t.cloneNode(key));
       }
 
       let value;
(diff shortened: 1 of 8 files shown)


existentialism:
@nicolo-ribaudo updated!

nicolo-ribaudo:
This code on the repl of this PR alerts `{ a: 1 }` instead of `a`:
```js
const { [(() => 1)()]: a, ...rest } = { 1: "a" };

alert(JSON.stringify(rest))
```

As in the object-rest-spread plugin, we should use the `toPropertyKey` helper when needed. Or, better, you could move the duplicated code to a `@babel/helper-` package so that we don't need to fix bugs twice 😆

existentialism:
@nicolo-ribaudo sigh, I'm an idiot. Fixed (avoided adding another helper package for now...)

Issue #7749, opened 2018-04-17
TypeScript import(...) types


DanielRosenwasser:
TypeScript 2.9 will support a new `import("foo")` syntax for referencing types in modules which have not been imported. See the original pull request at https://github.com/Microsoft/TypeScript/issues/14844 for more details.

To support this, Babel will likely need to implement

- [x] Parsing support
- [x] Emit support

sebinsua:
Has any work been done on this?

xtuc:
@sebinsua I think that no progress has been made yet. Do you want to work on it? Feel free to ask questions on our Slack.

sebinsua:
@xtuc Where's the Slack?

nicolo-ribaudo:
https://babeljs.slack.com/ (https://slack.babeljs.io/)

lostfictions:
Confirming that Babel with `@babel/plugin-transform-typescript` throws with a parsing error when attempting to use a `let x: typeof import('foo')` annotation, as described [here](https://blogs.msdn.microsoft.com/typescript/2018/05/31/announcing-typescript-2-9/).

PR #8798, opened 2018-10-02
Add @babel/parser support for typescript type import()

Kovensky:
| Fixed Issues?            | #7749
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This was my first time touching the parser at all so I'm kinda lost as to what to do; but through hacks and ~~`printf`~~ `console.log` debugging I got something that works, even if it's producing invalid AST nodes.

There are various TODOs and a possible problem that I noted in comments in the code.

One of the things I think is important is that this `import()` be its own node type (`TSImport`?) instead of the `CallExpression(Import, AssignmentExpression)` it normally would be if parsed with the `dynamicImport` parser. This also needs to work even when the `dynamicImport` plugin is not enabled, as it does not actually represent a dynamic import, just copies its syntax.

This definitely needs more work, I just want to know what the next steps are 😊

### DIFF
--- a/packages/babel-parser/src/plugins/typescript.js
+++ b/packages/babel-parser/src/plugins/typescript.js
@@ -219,7 +219,25 @@ export default (superClass: Class<Parser>): Class<Parser> =>
     }
 
     tsParseEntityName(allowReservedWords: boolean): N.TsEntityName {
-      let entity: N.TsEntityName = this.parseIdentifier();
+      let entity: N.TsEntityName;
+      // TODO: package this into its own tsParse function
+      if (this.state.type === tt._import) {
+        entity = this.startNode();
+        this.next();
+        if (!this.match(tt.parenL)) {
+          this.unexpected(null, tt.parenL);
+        }
+        this.eat(tt.parenL);
+        const args = this.parseCallExpressionArguments(tt.parenR, false, {
+          start: -1,
+        });
+        // TODO: assert args is length 1
+        entity.argument = args[0];
+        // TODO: make a TS-specific node type
+        entity = this.finishNode(entity, "Import");
+      } else {
+        entity = this.parseIdentifier();
+      }
       while (this.eat(tt.dot)) {
         const node: N.TsQualifiedName = this.startNodeAtNode(entity);
         node.left = entity;
@@ -595,6 +613,11 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         }
         case tt._typeof:
           return this.tsParseTypeQuery();
+        case tt._import:
+          // TODO: should this really be the same as parsing the RHS of typeof?
+          // XXX: TSQualifiedName is not allowed to be a return type of this function but import() references
+          // are allowed to have dots to access exported members
+          return this.tsParseEntityName(true);
         case tt.braceL:
           return this.tsLookAhead(this.tsIsStartOfMappedType.bind(this))
             ? this.tsParseMappedType()
--- a/packages/babel-parser/test/fixtures/typescript/types/import/input.js
+++ b/packages/babel-parser/test/fixtures/typescript/types/import/input.js
@@ -1,0 +1,2 @@
+let x: typeof import('./x');
+let Y: import('./y').Y;
--- a/packages/babel-parser/test/fixtures/typescript/types/import/options.json
+++ b/packages/babel-parser/test/fixtures/typescript/types/import/options.json
@@ -1,4 +1,0 @@
-{
-  "sourceType": "module",
-  "plugins": ["typescript", "dynamicImport"]
-}
--- a/packages/babel-parser/test/fixtures/typescript/types/import/options.json
+++ b/packages/babel-parser/test/fixtures/typescript/types/import/options.json
@@ -1,0 +1,4 @@
+{
+  "sourceType": "module",
+  "plugins": ["typescript", "dynamicImport"]
+}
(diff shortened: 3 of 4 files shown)


packages/babel-parser/test/fixtures/typescript/types/import/input.js
```diff
@@ -0,0 +1,2 @@
+let x: typeof import('./x');
```
nicolo-ribaudo: Q: Is `typeof import()` `TheModule` or `Promise<TheModule>`? In the second case I don't think that we should use a ts-specific node since `import()` is just a es expression.

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -219,7 +219,25 @@ export default (superClass: Class<Parser>): Class<Parser> =>
     }
 
     tsParseEntityName(allowReservedWords: boolean): N.TsEntityName {
-      let entity: N.TsEntityName = this.parseIdentifier();
+      let entity: N.TsEntityName;
+      // TODO: package this into its own tsParse function
+      if (this.state.type === tt._import) {
```
nicolo-ribaudo: Nit: `this.match(tt._import)`

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -219,7 +219,25 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+        entity = this.startNode();
+        this.next();
+        if (!this.match(tt.parenL)) {
```
nicolo-ribaudo: Nit: I think we have `this.expect`?

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -595,6 +613,11 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         }
         case tt._typeof:
           return this.tsParseTypeQuery();
+        case tt._import:
+          // TODO: should this really be the same as parsing the RHS of typeof?
+          // XXX: TSQualifiedName is not allowed to be a return type of this function but import() references
```
nicolo-ribaudo: Yeah, we should parse them using two different functions.

epeli:
Maybe @andy-ms could help here a bit?

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -219,7 +219,25 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+          this.unexpected(null, tt.parenL);
+        }
+        this.eat(tt.parenL);
+        const args = this.parseCallExpressionArguments(tt.parenR, false, {
+          start: -1,
+        });
+        // TODO: assert args is length 1
+        entity.argument = args[0];
+        // TODO: make a TS-specific node type
+        entity = this.finishNode(entity, "Import");
```
g-plane:
```suggestion
        entity = this.finishNode(entity, "TSImportType");
```

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -219,7 +219,25 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+      let entity: N.TsEntityName;
+      // TODO: package this into its own tsParse function
+      if (this.state.type === tt._import) {
+        entity = this.startNode();
+        this.next();
+        if (!this.match(tt.parenL)) {
+          this.unexpected(null, tt.parenL);
+        }
+        this.eat(tt.parenL);
+        const args = this.parseCallExpressionArguments(tt.parenR, false, {
```
g-plane: I think `StringLiteral` is only allowed, because `import()` in type annotation is not a real function.

alangpierce:
Hi! I just implemented this in the Sucrase parser (a fork of babel-parser that doesn't produce an AST). You might find it useful to work off of: https://github.com/alangpierce/sucrase/pull/380

In the TypeScript codebase, the relevant code to work off of is `parseImportType` ( https://github.com/Microsoft/TypeScript/pull/22592/files#diff-ead24f0f0f59c0ea9c1c511052e8884b )

Looks like one case that's missing in this PR is applying type arguments to imported types, e.g. `import('./data-structures').HashSet<string>`.

packages/babel-parser/test/fixtures/typescript/types/import/input.js
armano2: as a small info typeof is not mandatory for `ImportType`
```
type B = import("B").X;
```
is valid to

armano2:
i played around with this a little and it should be something like this:

```js
    tsParseImportType(allowReservedWords: boolean): N.TsImportType {
      const node: N.TsImportType = this.startNode();
      this.expect(tt._import);
      this.expect(tt.parenL);
      if (!this.match(tt.string)) {
        throw this.unexpected();
      }
      node.parameter = this.parseLiteral(this.state.value, "StringLiteral");
      this.expect(tt.parenR);

      if (this.match(tt.dot)) {
        node.qualifier = this.tsParseEntityName(allowReservedWords)
      }
      if (!this.hasPrecedingLineBreak() && this.isRelational("<")) {
        node.typeParameters = this.tsParseTypeArguments();
      }
      return this.finishNode(node, "TSImportType");
    }
```

but we should add test case for:

```
import("foo").foo.bar<string>
```

armano2:
@hzoo #9302 got merged and this PR is not longer relevant.

existentialism:
Landed as https://github.com/babel/babel/pull/9302

PR #8805, opened 2018-10-02
Typescript - Tuples can include rest elements

### LINKED ISSUE #8623: [TypeScript 3.0]: New tuple type support

Retsam:
| Fixed Issues?            | Fixes #8623
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is the rest of #8623, the part that wasn't addressed by #8720.  The changes here are very similar to those in #8720.

### DIFF
--- a/packages/babel-generator/src/generators/typescript.js
+++ b/packages/babel-generator/src/generators/typescript.js
@@ -246,6 +246,11 @@ export function TSOptionalType(node) {
   this.token("?");
 }
 
+export function TSVariadicType(node) {
+  this.token("...");
+  this.print(node.typeAnnotation, node);
+}
+
 export function TSUnionType(node) {
   this.tsPrintUnionOrIntersectionType(node, "|");
 }
--- a/packages/babel-generator/src/generators/typescript.js
+++ b/packages/babel-generator/src/generators/typescript.js
@@ -246,7 +246,7 @@ export function TSOptionalType(node) {
   this.token("?");
 }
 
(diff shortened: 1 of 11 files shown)


Retsam:
Something to note about this change: it doesn't do any checking about the position of the rest elements in the tuple.

So it'll allow for syntactically reasonable, but semantically invalid constructs like:

```ts
const x: [...number[], ...string[]] // Two rest elements
const y: [...number[], string] // Rest element not at the end of the tuple
```

(Similarly the tuples PR allows for stuff like `[number, string?, number]`)

I'm not sure if that sort of validation is in-scope for babel-parser.

nicolo-ribaudo:
I think that it's not too bad to allow invalid stuff (`[...Foo, bar]`) since people will run ts on their code anyway, which will report an error.
I'd like to disallowed it in the future though

existentialism:
> I'd like to disallowed it in the future though

Agreed!

PR #8828, opened 2018-10-08
Typescript: Validate tuple type element positions

Retsam:
| Fixed Issues?            | 
| Patch: Bug Fix?          | 
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Adds additional validation to the parser functionality added by #8720 and #8805.  Per [this comment](https://github.com/babel/babel/pull/8805#issuecomment-427252654), adds a validation step to the parser which raises syntax errors if a rest param is not at the end of a tuple, or if a mandatory param follows an optional parameter.  So the following snippets are correctly considered syntax errors:

```ts
const x: [number?, number]; // can't have mandatory member after optional member
const x: [...number[], string]; // rest element must be the last element of the tuple
```

### DIFF
--- a/packages/babel-parser/src/plugins/typescript.js
+++ b/packages/babel-parser/src/plugins/typescript.js
@@ -513,14 +513,13 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       //   If there's a rest element, it must be at the end of the tuple
       let seenOptionalElement = false;
       node.elementTypes.forEach((elementNode, i) => {
-        if (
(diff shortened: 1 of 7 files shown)


danez:
Are they SyntaxErrors? If I try to parse these in the typescript playground, it all parses fine.

danez:
Ah the typescript parser always tries to output something useful even if there are syntax errors in the code. 👍

danez:
Thanks for your great contributions and helping to improve our typescript support. 🎉

PR #8830, opened 2018-10-08
Correct handling of newline after async with paren-less arrow func

### LINKED ISSUE #8795: Parser: async paren-less arrow with newline should throw
## Bug Report

**Current Behavior**
Parse the input code. The sloppy/strict script/module modes are not relevant.

**Input Code**

Simple case:
```
async 
x => x
```

And slightly more complex:
```js
let f = async
g => g
```

**Expected behavior/code**

Apply ASI when the newline is encountered.

```js
        babelAst = babelParse(code, {
          sourceType: goalMode === GOAL_MODULE ? 'module' : 'script',
          strictMode: startInStrictMode ? true : false,
          plugins: ['estree'],
        }).program;
```

**Environment**
- Babel version(s): 7.0.0 (regular)
- Node/npm version: node 10
- OS: linux
- Monorepo: ok you got me. yarn installed it. what's the answer?
- How you are using Babel: from code

**Possible Solution**

Apply ASI when parsing `async` and the next token is an identifier and there was at least one newline before that identifier. Note that this is a special case within the already special-async-case. The grammar explicitly has a goal for `async \n binding => body`, which is separate from the "group" cover grammar path.

Example AST for the `let f = async↵g => g` case:

Babel AST (partially stripped):
```
{ type: 'Program',
  body:
   [ { type: 'VariableDeclaration',
       declarations:
        [ { type: 'VariableDeclarator',
            id: { type: 'Identifier', name: 'f' },
            init:
             { type: 'ArrowFunctionExpression',
               async: true,
               body: { type: 'Identifier', name: 'g' },
               expression: true,
               generator: false,
               id: null,
               params: [ { type: 'Identifier', name: 'g' } ] } } ],
       kind: 'let' } ] }
```

Expected AST:
```
{ type: 'Program',
  body:
   [ { type: 'VariableDeclaration',
       kind: 'let',
       declarations:
        [ { type: 'VariableDeclarator',
            id: { type: 'Identifier', name: 'f' },
            init: { type: 'Identifier', name: 'async' } } ] },
     { type: 'ExpressionStatement',
       expression:
        { type: 'ArrowFunctionExpression',
          params: [ { type: 'Identifier', name: 'g' } ],
          id: null,
          generator: false,
          async: false,
          expression: true,
          body: { type: 'Identifier', name: 'g' } } } ] }
```

Retsam:
| Fixed Issues?            | Fixes #8795 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Assuming the issue reported in #8795 is correct (I haven't read the spec, but the described behavior seems to match other parsers such as flow or typescript), then 

```js
async
x => x
```

should parse as two expression statements, not a single arrow function like it currently does.


Retsam:
```
async
(x) => x
```

is a parse error, which seems consistent how other parsers treat it.  (e.g. `acorn`)  I'm not good enough at reading the spec to say if that's the correct behavior.

nicolo-ribaudo:
I think that it should be allowed (they are two expression statements), but you can test it in the console of a modern browser.

Retsam:
Yeah, the version with parens is a syntax error in Firefox, Chrome, and Safari.  (And the paren-less version is not)

pvdz:
Yeah the spec is a little difficult to read. Basically you're supposed to parse `async(` as a regular call first, in which case the newline between `async` and `(` is fine. Then if you find an arrow, you're supposed to apply the cover grammar, which disallows the newline. If the cover grammar can't be applied the parser must consider it a syntax error.

For the paren-less case there is no cover grammar to apply as there is no ambiguity. If there's a newline there's an ASI and if there's no newline then `async foo` can only lead to an arrow (well, or a crash :) )

Thanks for fixing :)

PR #8842, opened 2018-10-11
Skip react inline elements for custom pragma

remcohaszing:
The Babel JSX helps for React elements, but not necessarily for other JSX pragma
implementations.

| Fixed Issues?            | Sorry, dove right into the code
| Major: Breaking Change?  | 👍 (also, sort of)
| Minor: New Feature?      | Nope
| Documentation PR Link    | 
| Any Dependency Changes?  | Nope

`@babel/plugin-transform-react-inline-elements` optimizes JSX in a way that works specificly for React. This means it could break implementations that rely on another JSX implementation.

This change skips the transformation if a custom JSX pragma is found. The implementation, including the regular expression and visitor method, is based on [@babel/plugin-transform-react-jsx](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-react-jsx/src/index.js)


danez:
Is it still common to use jsx pragmas? Especially in bigger codebases that seems a lot of work to add the pragma to each and every file and also error prone, if you for example forget it in one case. Or is this added automatically?
I wonder why the plugin is included in the first place if you do not want it? And I guess one codebase usually never has both React and some other jsx implementation?

remcohaszing:
I use a mono repository which includes multiple frontend projects.

One project uses React.

Another project is very small, so I decided React as a dependency would be overkill. Because I do like the JSX syntax, I created a JSX pragma which converts JSX syntax to simple DOM nodes. Except for the `@babel/plugin-transform-react-inline-elements` plugin, this works great.

As a workaround, I have to put `@babel/plugin-transform-react-jsx` (Which does support a custom JSX pragma and is also named `...-react-...`.) in my _babel.config.js_ before `@babel/plugin-transform-react-inline-elements`. This seems a bit silly, as I also use `@babel/preset-react`.

probablyup:
@danez yeah I use one in markdown-to-jsx to enable some interesting override functionality: https://github.com/probablyup/markdown-to-jsx/blob/master/index.js#L1

PR #8866, opened 2018-10-12
Fixes #8865

### LINKED ISSUE #8865: 'jsonStrings' parser plugin does not increment line count
## Bug Report

I realized that https://github.com/babel/babel/pull/7985 should also have a `++this.state.curLine;` in there along with the `++this.state.pos;`, since otherwise the location information isn't quite right.

cc @jridgewell

byronluk:
| Fixed Issues?            | Fixes #8865 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Whenever the json-strings parser plugin is available and the position is incremented -- the currentLine should be incremented as well to make sure the location information is correct.

### DIFF
--- a/packages/babel-parser/src/tokenizer/index.js
+++ b/packages/babel-parser/src/tokenizer/index.js
@@ -1087,6 +1087,7 @@ export default class Tokenizer extends LocationParser {
         (ch === charCodes.lineSeparator || ch === charCodes.paragraphSeparator)
       ) {
         ++this.state.pos;
+        ++this.state.curLine;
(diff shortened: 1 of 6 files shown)


jridgewell:
The test failures are because the `output.json` files weren't updated. Run `rm packages/babel-parser/test/fixtures/experimental/json-strings/*/output.json`, then `make test-only` to regenerate them.

existentialism:
thanks @byronluk!

PR #8868, opened 2018-10-13
fix single-arg async arrows when retainLines=true

### LINKED ISSUE #7275: `retainLines` breaks `no LineTerminator` requirements in single-arg async arrows and decorated methods
```js
var fn = async (
  arg
) => {}
```

with `retainLines: true` becomes

```js
var fn = async 
  arg => {}
```
which is a syntax error since `async` and `arg` need to be on the same line when there aren't parens.

We need to detect that `retainLines` is enabled, and that the identifier will insert new lines, and if so, wrap parens around the argument. We've got some existing logic in place for noLineTerminator but I don't think it would apply very well to this case. :(

Edit: A second example of this issue can also be found in https://github.com/babel/babel/issues/8650, so if we fix this issue, we should fix that one at the same time.

ryanwmarsh:
| Fixed Issues?            | Fixes #7275
| Documentation PR Link    | ?

I came across babel/babel#7275 trying to work around facebook/create-react-app#5319 @roblourens pointed me here.

re: @loganfsmyth 

> 
> ```javascript
> var fn = async (
>   arg
> ) => {}
> ```
> 
> with retainLines: true becomes
> 
> ```javascript
> var fn = async 
>   arg => {}
> ```
> which is a syntax error since async and arg need to be on the same line when there aren't parens.
> 
> We need to detect that retainLines is enabled, and that the identifier will insert new lines, and if so, wrap parens around the argument.
> 

So that's what I did but there's only one new test. This could probably use a few more test cases but I don't know what they would be, I just want the upstream issues (facebook/create-react-app#5319, facebook/jest#5326, Microsoft/vscode#60468, Microsoft/vscode#60187) to be resolved so I can go on my merry way and this seemed to be the root blocker. If someone could point me to a few more cases that this should handle I will gladly add them.

### DIFF
--- a/packages/babel-generator/src/generators/methods.js
+++ b/packages/babel-generator/src/generators/methods.js
@@ -111,7 +111,23 @@ export function ArrowFunctionExpression(node: Object) {
     t.isIdentifier(firstParam) &&
     !hasTypes(node, firstParam)
   ) {
-    this.print(firstParam, node);
+    if (
+      this.format.retainLines &&
+      node.loc.start.line < node.body.loc.start.line
+    ) {
+      this.token("(");
+      if (firstParam.loc.start.line > node.loc.start.line) {
+        this.indent();
+        this.print(firstParam, node);
+        this.dedent();
+        this._catchUp("start", node.body.loc);
+      } else {
+        this.print(firstParam, node);
+      }
+      this.token(")");
+    } else {
+      this.print(firstParam, node);
+    }
   } else {
     this._params(node);
   }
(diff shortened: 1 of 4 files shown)


ryanwmarsh:
@nicolo-ribaudo Sure thing. What should it output with retainLines? The same?

```javascript
async (x) 
=> {};
```

Also, yes I recognize everyone wants to get rid of retain lines but many tools unfortunately still rely on it. 🙁

nicolo-ribaudo:
Ok, I give you a bad example 😅

```js
async (x
) => {};
```

Just to check that it works even if the body is in the first line after the parameters

ryanwmarsh:
Added it anyway. Is this safe to merge?

ryanwmarsh:
Should I abandon this?

nicolo-ribaudo:
Nope 🙂

ryanwmarsh:
My first commit to Babel!!! 🎉

PR #8872, opened 2018-10-14
transform-spread: fix empty array behaviour

### LINKED ISSUE #7258: Incorrect compilation for empty array

macabeus:
| Fixed Issues?            | Fixes #7258
| Patch: Bug Fix?          | Yeah
| Major: Breaking Change?  | Yep (if someone uses the incorrect behaviour)
| Minor: New Feature?      |
| Tests Added + Pass?      | Yep
| Documentation PR Link    | 
| Any Dependency Changes?  |

Babel wasn't compile correctly if array is hole. For example:

Correct behaviour:
```
var array = new Array(3);
[...array].forEach(console.log); // should print undefined three times

var array2 = [1,,,77,,, 99];
[...array2]; // should be [1, undefined, undefined, 77, undefined, undefined, 99]
```

Babel wrong behaviour fixed:
```
var array = new Array(3);
[...array].forEach(console.log); // print undefined only once!!

var array2 = [1,,,77,,, 99];
[...array2] // is [1, 3: 77, 6: 99] !!
```


packages/babel-plugin-transform-spread/src/index.js
```diff
@@ -62,12 +62,26 @@ export default declare((api, options) => {
...
-        path.replaceWith(
-          t.callExpression(
-            t.memberExpression(first, t.identifier("concat")),
-            nodes,
-          ),
-        );
+        this.addHelper("toConsumableArray");
+
+        const toConsumableArrayExpression = obj =>
+          t.callExpression(t.identifier("_toConsumableArray"), obj);
```
loganfsmyth: I don't have time to think about this PR as a whole at the moment, but FYI this identifier is returned from `addHelper`, you shouldn't create it manually.

macabeus:
@loganfsmyth Nice! I updated the code to use the return of `addHelper`, also I made some readable improvements and fix some others edge case

packages/babel-plugin-transform-spread/src/index.js
```diff
@@ -62,12 +62,29 @@ export default declare((api, options) => {
           return;
         }
 
-        path.replaceWith(
+        const toConsumableArray = this.addHelper("toConsumableArray");
```
nicolo-ribaudo: You should move this call inside the `wrapCallConsumableArray` function, otherwise we are going to reuse the same node multiple times.

Issue #8605, opened 2018-09-02
[Feature Request] Integrate filesystem caching into core & plugin system


loganfsmyth:
## Feature Request

Integrations like `babel-loader` & `babel/register` already have their own caching logic, but because it is not integrated into the core, Babel plugins aren't aware of it and don't have any way to force invalidation of their caches.

I implemented some of this in #6786 but it wasn't quite ready yet and other 7.x features took priority.

PR #8877, opened 2018-10-16
cli: Support `--incremental` option

Zarel:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    |
| Any Dependency Changes?  |

This adds a new `--incremental` option to the Babel CLI, which will, in out-dir mode, only build files that have been modified more recently than the corresponding output file in `--out-dir`.

This also changes it so an error in `--watch` mode doesn't copy files in `--copy-files` mode, matching behavior outside of `--watch` mode, which seemed like a bug to me (I can break that out into a separate pull request if that's important).


packages/babel-cli/src/babel/dir.js
```diff
@@ -7,21 +7,31 @@ import fs from "fs";
 
 import * as util from "./util";
 
+const NOT_COMPILABLE = null;
```
nicolo-ribaudo: I'd prefer `NOT_COMPILABLE = {}`, so that its value is unique.

packages/babel-cli/src/babel/dir.js
```diff
@@ -75,11 +85,20 @@ export default async function({ cliOptions, babelOptions }) {
...
+    if (written === NOT_COMPILABLE) {
+      if (!cliOptions.copyFiles) return false;
       const filename = path.relative(base, src);
       const dest = getDest(filename, base);
+      if (cliOptions.incremental) {
+        try {
+          const srcStat = fs.statSync(src);
+          const destStat = fs.statSync(dest);
+          if (srcStat.ctimeMs < destStat.ctimeMs) return false;
+        } catch (e) {}
```
nicolo-ribaudo: We could move this duplicated logic to a `isSrcNewer(src, dest)` helper.

packages/babel-cli/src/babel/dir.js
Zarel: Yeah, I wasn't sure what the best way to do this is. That works.

packages/babel-cli/src/babel/dir.js
Zarel: Sure.

Zarel:
@nicolo-ribaudo Done.

nicolo-ribaudo:
Is it possible to add a test for this option?

Zarel:
@nicolo-ribaudo I noticed `--skip-initial-build` and `--watch` didn't have tests, so I assumed it wasn't too important. I'm willing to write a test for this, but it's a decent amount of work to screw around with manually setting file change times, so I'd like some assurance that this will be merged before I spend that much time on it.

Zarel:
For the record, I've manually tested this, and am using this in production in https://github.com/Zarel/Pokemon-Showdown-Client/commit/f09d91711001def5cd070eae4603b500653df002

nicolo-ribaudo:
Ok, let's wait for other approvals then.

loganfsmyth:
My concern with this is that it's not just the `mtime` of the content that matters, you could also be running with different environment variables, or have updated a config file or installed a new version of a plugin/preset.

I'd rather fall back to something like https://github.com/babel/babel/issues/8605 for performance optimizations like this instead.

Zarel:
Yeah, a unified cache system does sound better, although it does still need a fix for the "changed/updated something" issue.

My main defense of this system is that you can always just recompile once without `--incremental` when you update something, and, like, this is done and I don't know how long we'd have to wait for a unified cache system.

Zarel:
Yeah, a unified cache system does sound better, although it does still need a fix for the "changed/updated something" issue.

My main defense of this system is that you can always just recompile once without `--incremental` when you update something, and, like, this is done and I don't know how long we'd have to wait for a unified cache system.

Zarel:
It's been a year and I see no work being done on any cache system. Are you willing to revisit this?

Zarel:
I don't think there's much chance this'll get merged, but I'm still using this personally, so I thought I'd rebase this on the latest Babel.

(It's a huge difference, build times are 0.36s rather than 6.91s. Some people on the internet called me crazy for maintaining a Babel fork, but it's well worth the fast build times.)

Unrelatedly, I'm glad to see the bug in `--watch` has been fixed.

dac09:
I'd love to see this, or equivalent functionality merged. Incremental builds works really well with `--watch` on the CLI

PR #8883, opened 2018-10-16
[flow] Add support for parsing `_` as implicit instantiation in call/new

jbrown215:
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Yes? (unlikely to cause new parse errors when using flow, but may)
| Documentation PR Link    | 

In https://github.com/facebook/flow/commit/fb23b93750b8b72307c889dce8f6830f6fc0a30e, flow started to treat "_" used as type arguments to calls/new (`test<_>()` or `new test<_>()`) to mean to implicitly instantiate a type variable for that type parameter.

In that change, Flow disallowed the use of `_` as a type identifier and it can _only_ be used as a type argument in call/new. The estree AST flow outputs still prints `_` in call/new type arguments as a `GenericTypeAnnotation` with Identifier `_`, which this change mimics. 

No old tests had to be re-recorded, since the output AST did not change. Any programs using "\_" as a type identifier name will now have a parse error when using the flow parser after this change. As an example, every use of "\_" in the following code would cause a parse error in flow:
```
var x: Generic<_> = 3;
(x: Generic<_>);
var y: _ = 3;
(y: _);
type _ = number;
opaque type _ = number;
type HasUnderscore<_> = _;
```

This feature will be released in flow v0.84.0


packages/babel-parser/src/plugins/flow.js
```diff
@@ -484,7 +484,17 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+      if (word === "_") {
+        throw this.unexpected(
+          null,
+          "`_` is only allowed as a type argument to call or new",
+        );
+      }
+    }
+
     checkReservedType(word: string, startLoc: number) {
+      this.checkNotUnderscore(word);
```
danez: Can we remove this, I covered this case in #8965 by adding _ to the `primitiveTypes` array.

danez:
Thanks. 

The remaining travis failure has nothing to do with this PR.

jbrown215:
Thanks for the help!

PR #8884, opened 2018-10-16
[flow] Explicit inexact objects with `...`

jbrown215:
| Patch: Bug Fix?          |
| Documentation PR Link    | 

In https://github.com/facebook/flow/commit/1ac913040f38309480934ccb6717a3ffc65094a8, flow started parsing `{foo: number, ...}` as an inexact object. This is part of a larger effort to move to [parsing object types as exact by default](https://medium.com/flow-type/on-the-roadmap-exact-objects-by-default-16b72933c5cf). 

`...` can only appear at the end of the props list, and it cannot appear inside `{||}`. Additionally, it can only appear inside object types, and not interfaces, declare classes, or declare interfaces. `...` also supports trailing commas.

For any object type, but not in declare class, declare interface, or interface, the `ObjectTypeAnnotation` representing it now has an `inexact` property. This property is true if and only if `...` was used to indicate the inexactness. To minimize confusion, it is best to read `exact` as `explicitly exact` (containing curlybars) and `inexact` as `explicitly inexact` (containing `...`). This mimics the behavior in Flow's estree output https://github.com/facebook/flow/commit/231256cfe5afc3c7576f05ac9929859dfa813a35. In a few releases, we will get rid of the exact flag entirely, which will be a breaking change to other tools.

Tests had to be re-recorded to include the `inexact` flag. I separated the re-recording of old tests into a separate commit for reviewing ease.

Let me know if there is anything else I need to update in this change.

### DIFF
--- a/packages/babel-parser/src/plugins/flow.js
+++ b/packages/babel-parser/src/plugins/flow.js
@@ -882,8 +882,8 @@ export default (superClass: Class<Parser>): Class<Parser> =>
             allowInexact,
           );
 
-          if (typeof propOrInexact === "boolean") {
-            inexact = inexact || propOrInexact;
+          if (propOrInexact === null) {
+            inexact = true;
           } else {
             nodeStart.properties.push(propOrInexact);
           }
@@ -918,7 +918,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       kind: string,
       allowSpread: boolean,
       allowInexact: boolean,
-    ): (N.FlowObjectTypeProperty | N.FlowObjectTypeSpreadProperty) | boolean {
+    ): (N.FlowObjectTypeProperty | N.FlowObjectTypeSpreadProperty) | null {
       if (this.match(tt.ellipsis)) {
         if (!allowSpread) {
           this.unexpected(
@@ -936,41 +936,31 @@ export default (superClass: Class<Parser>): Class<Parser> =>
           );
         }
         this.expect(tt.ellipsis);
-        switch (this.state.type) {
-          case tt.comma:
-          case tt.semi:
-          case tt.braceR:
-          case tt.braceBarR: {
-            const isInexactToken = this.eat(tt.comma) || this.eat(tt.semi);
+        const isInexactToken = this.eat(tt.comma) || this.eat(tt.semi);
 
-            if (this.match(tt.braceR)) {
-              if (allowInexact) return true;
-              this.unexpected(
-                null,
-                "Explicit inexact syntax is only allowed inside inexact objects",
-              );
-            }
-
-            if (this.match(tt.braceBarR)) {
-              this.unexpected(
-                null,
-                "Explicit inexact syntax cannot appear inside an explicit exact object type",
-              );
-              return false;
-            }
-
-            if (isInexactToken) {
-              this.unexpected(
-                null,
-                "Explicit inexact syntax must appear at the end of an inexact object",
-              );
-              return false;
-            }
-          }
-          default:
-            node.argument = this.flowParseType();
-            return this.finishNode(node, "ObjectTypeSpreadProperty");
+        if (this.match(tt.braceR)) {
+          if (allowInexact) return null;
+          this.unexpected(
+            null,
+            "Explicit inexact syntax is only allowed inside inexact objects",
+          );
         }
+
+        if (this.match(tt.braceBarR)) {
+          this.unexpected(
+            null,
+            "Explicit inexact syntax cannot appear inside an explicit exact object type",
+          );
+        }
+
+        if (isInexactToken) {
+          this.unexpected(
+            null,
+            "Explicit inexact syntax must appear at the end of an inexact object",
+          );
+        }
+        node.argument = this.flowParseType();
+        return this.finishNode(node, "ObjectTypeSpreadProperty");
       } else {
         node.key = this.flowParseObjectPropertyKey();
         node.static = isStatic;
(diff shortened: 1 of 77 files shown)


packages/babel-parser/src/plugins/flow.js
```diff
@@ -759,6 +765,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       allowExact: boolean,
       allowSpread: boolean,
       allowProto: boolean,
+      allowInexact: boolean,
```
nicolo-ribaudo: Aside: we should use an object instead of all these boolean params

packages/babel-parser/src/plugins/flow.js
```diff
@@ -852,23 +860,37 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
       }
 
       this.expect(endDelim);
 
+      /* The inexact flag should only be added on ObjectTypeAnnotations that
+       * are not the body of an interface, declare interface, or declare class.
+       * Since spreads are only allowed in objec types, checking that is
+       * sufficient here.
+       */
+      if (allowSpread) {
```
nicolo-ribaudo: Nit: we should rename this flag, maybe to `isObjectLiteralType`?

packages/babel-parser/src/plugins/flow.js
```diff
@@ -901,9 +924,42 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
         this.expect(tt.ellipsis);
-        node.argument = this.flowParseType();
+        switch (this.state.type) {
+          case tt.comma:
+          case tt.semi:
+          case tt.braceR:
+          case tt.braceBarR:
+            this.flowObjectTypeSemicolon();
+            switch (this.state.type) {
+              case tt.braceR:
```
nicolo-ribaudo: Nit: I think that this code would be more readable without nested switch statemets:

```js
let isInexactToken = this.eat(tt.comma) || this.eat(tt.semi);

if (this.match(tt.braceR)) {
  if (allowInexact) return true;
  this.unexpected(null, "Explicit inexact syntax is only allowed inside inexact objects");
}

if (this.match(tt.braceBarR)) {
  this.unexpected(null, "Explicit inexact syntax cannot appear inside an explicit exact object type");
}

if (isInexactToken) {
  this.unexpected(null, "Explicit inexact syntax must appear at the end of an inexact object");
}

node.argument = this.flowParseType();
return this.finishNode(node, "ObjectTypeSpreadProperty");
```

Do what you prefer though

packages/babel-parser/src/plugins/flow.js
jbrown215: (y) can do!

packages/babel-parser/src/plugins/flow.js
jbrown215: I agree-- the only reason I'd push back here is because this closely mirrors the way it is implemented in flow, which makes it easier for us to maintain.

https://github.com/facebook/flow/blob/master/src/parser/type_parser.ml#L566-L584

packages/babel-parser/src/plugins/flow.js
jbrown215: I like this better, will follow up

packages/babel-parser/src/plugins/flow.js
nicolo-ribaudo: Oh ok. It's not a big deal then

packages/babel-parser/src/plugins/flow.js
```diff
@@ -901,9 +936,41 @@ export default (superClass: Class<Parser>): Class<Parser> =>
           );
         }
         this.expect(tt.ellipsis);
-        node.argument = this.flowParseType();
+        switch (this.state.type) {
+          case tt.comma:
+          case tt.semi:
+          case tt.braceR:
+          case tt.braceBarR: {
```
nicolo-ribaudo: I think that you can also remove this outer `switch`, since the `if`s already handle only these four cases.

packages/babel-parser/src/plugins/flow.js
```diff
@@ -901,9 +936,41 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+                "Explicit inexact syntax is only allowed inside inexact objects",
+              );
+            }
+
+            if (this.match(tt.braceBarR)) {
+              this.unexpected(
+                null,
+                "Explicit inexact syntax cannot appear inside an explicit exact object type",
+              );
+              return false;
```
nicolo-ribaudo: Those `retuen`s aren't needed, since `this.unexpected` already throws. If you want to be explicit, I think that it would be more correct to use `throw this.unexpected`.

jbrown215:
@nicolo-ribaudo, thanks for the feedback!

I'll backport this to 6.xx as well, since I assume there's still a bunch of flow projects out there that haven't upgraded to babel 7 yet.

jbrown215:
Actually, it's not clear to me where to start with backporting this change. The `babylon` repo is read-only and the `6.x` readme/contributing.md doesn't say where to put up new PRs. The 6.x branch here doesn't have a babylon or babel-parser package. Is there any follow up work I need to do here?

nicolo-ribaudo:
We are not going to release new Babel 6 version, especially for new features.

jbrown215:
Any idea when the next babel release is / if this will be in it?

existentialism:
@jbrown215 no exact date... soon though? :)

PR #8885, opened 2018-10-16
Add useAutoDirectory option to transform-runtime plugin

Andarist:
| Fixed Issues?            | 
| Any Dependency Changes?  |

**Why?** 
When building libraries it's annoying to setup different options for `@babel/plugin-transform-runtime` for ESM and CJS builds.

With this the configuration can be unified with the newly introduced option and i.e. when bundling with rollup we could shorten this:
```js
const config = [
  {
    input: 'src/index.js',
    output: { file: 'dist/library.js', output: 'cjs' },
    plugins: [
      babel({
        runtimeHelpers: true,
        plugins: [
          '@babel/transform-runtime'
        ]
      }),  
    ]
  },
  {
    input: 'src/index.js',
    output: { file: 'dist/library.esm.js', output: 'esm' },
    plugins: [
      babel({
        runtimeHelpers: true,
        plugins: [
          [
            '@babel/transform-runtime',
            { useESModules: true }
          ]
        ]
      }),  
    ]
  }
]
```
to this
```js
const config =   {
  input: 'src/index.js',
  output: [
    { file: 'dist/library.js', output: 'cjs' },
    { file: 'dist/library.esm.js', output: 'esm' }
  ],
  plugins: [
    babel({
      runtimeHelpers: true,
      plugins: [
        [
          '@babel/transform-runtime',
          { useAutoDirectory: true }
        ]
      ]
    }),  
  ]
}
```

This technique of creating extra package.json files is not well-known but it's part of the node resolution algorithm - https://nodejs.org/api/modules.html#modules_all_together & resolution algorithms of webpack, rollup and other tools support this. This has an advantage of allowing tools to pick up appropriate files thanks to being able to specify both main & module entries for a single request path.

⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ 
Unfortunately helpers written in CJS format have different shape from the helpers written in ESM. With ESM things are exported as default, with CJS they are assigned directly to `module.exports`.

As we know those things are not really compatible and I have 2 ideas how we could fix this (the issue to be fixed is that package.json in auto directory should point to modules with the same shape, regardless of the format):
- introduce extra files 😭 that would reexport current CJS helpers as `module.exports.default = require('../createClass')` and point to those proxy files from auto package.json files
- as helpers are mostly functions we can assign to them `.default` property that would self-reference to them, I'm not sure if this is completely safe though, could it cause problems with live bindings such as in `extends` helper?

### DIFF
--- a/packages/babel-plugin-transform-runtime/scripts/build-dist.js
+++ b/packages/babel-plugin-transform-runtime/scripts/build-dist.js
@@ -84,8 +84,6 @@ function writeAutoHelperFiles(runtimeName) {
       filename,
       JSON.stringify(
         {
-          name: `${runtimeName}/helpers/auto/${helperName}`,
-          private: true,
           main: `../../${helperName}.js`,
           module: `../../esm/${helperName}.js`,
         },
--- a/packages/babel-plugin-transform-runtime/scripts/build-dist.js
+++ b/packages/babel-plugin-transform-runtime/scripts/build-dist.js
@@ -44,6 +44,7 @@ function writeCoreJS2(runtimeName) {
 function writeHelpers(runtimeName, { corejs } = {}) {
   writeHelperFiles(runtimeName, { corejs, esm: false });
   writeHelperFiles(runtimeName, { corejs, esm: true });
+  writeAutoHelperFiles(runtimeName);
 }
 
 function writeHelperFiles(runtimeName, { esm, corejs }) {
@@ -63,6 +64,34 @@ function writeHelperFiles(runtimeName, { esm, corejs }) {
         esm,
         corejs,
       })
+    );
+  }
+}
+
+function writeAutoHelperFiles(runtimeName) {
+  const pkgDirname = getRuntimeRoot(runtimeName);
+
+  for (const helperName of helpers.list) {
+    const filename = path.join(
+      pkgDirname,
+      "helpers",
+      "auto",
+      helperName,
+      "package.json"
+    );
+
+    outputFile(
+      filename,
+      JSON.stringify(
+        {
+          name: `${runtimeName}/helpers/auto/${helperName}`,
+          private: true,
+          main: `../../${helperName}.js`,
+          module: `../../esm/${helperName}.js`,
+        },
+        null,
+        2
+      ) + "\n"
     );
   }
 }
--- a/packages/babel-plugin-transform-runtime/src/index.js
+++ b/packages/babel-plugin-transform-runtime/src/index.js
@@ -37,6 +37,7 @@ export default declare((api, options, dirname) => {
     helpers: useRuntimeHelpers = true,
     regenerator: useRuntimeRegenerator = true,
     useESModules = false,
+    useAutoDirectory = false,
     version: runtimeVersion = "7.0.0-beta.0",
     absoluteRuntime = false,
   } = options;
@@ -55,6 +56,14 @@ export default declare((api, options, dirname) => {
     throw new Error(
       "The 'useESModules' option must be undefined, or a boolean, or 'auto'.",
     );
+  }
+  if (useESModules && useAutoDirectory) {
+    throw new Error(
+      "'useESModules' & 'useAutoDirectory' options cannot be used at the same time.",
+    );
+  }
+  if (typeof useAutoDirectory !== "boolean") {
+    throw new Error("The 'useAutoDirectory' option must be a boolean.");
   }
   if (
     typeof absoluteRuntime !== "boolean" &&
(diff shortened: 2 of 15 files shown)


packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -56,6 +57,14 @@ export default declare((api, options, dirname) => {
       "The 'useESModules' option must be undefined, or a boolean, or 'auto'.",
     );
   }
+  if (useESModules && useAutoDirectory) {
+    throw new Error("The 'useAutoDirectory' option must be a boolean.");
```
loganfsmyth: Looks like this message got swapped with the other error.

packages/babel-plugin-transform-runtime/scripts/build-dist.js
```diff
@@ -67,6 +68,34 @@ function writeHelperFiles(runtimeName, { esm, corejs }) {
...
+      "auto",
+      helperName,
+      "package.json"
+    );
+
+    outputFile(
+      filename,
+      JSON.stringify(
+        {
+          name: `${runtimeName}/helpers/auto/${helperName}`,
```
loganfsmyth: Are the `name` and `private` fields needed? I'd have expected only `main` and `module`.

Andarist: They are probably not, Im just accustomed to using them. IMHO `private` gives an extra pinch of confidence that this won't be tried to get published (especially in lerna setup in case of some weird bug on their side).

`name` gives some more meaning to those files (kinda same with private, semantically both are ok). So I find adding those fields a nice touch, but as mentioned - they are not probably strictly required, so I can check it out and remove them if you feel they are not needed here.

Andarist:
@loganfsmyth thanks for the comments! WDYT about the issue described in the "warning" section of the PR?

packages/babel-plugin-transform-runtime/scripts/build-dist.js
loganfsmyth: My feeling is that if `name` isn't there, then an accidental publish isn't a concern either way. If we really want `name`, then I agree having the `private: true` is a good choice though. I'd probably rather leave them out and add them if we get reports of it being an issue.

Andarist: Removed them.

loganfsmyth:
Hmm, I'd actually considered the current behavior to be more compatible since logically the default import of a CommonJS file is the `module.exports` object. For instance in Node's experimental mode, and Webpack, `import helper from "@babel/runtime/helpers/classCallCheck";` would work correctly with either type of helper.

Does Rollup have different expectations in this case?

Andarist:
The problem is that webpack became stricter (I think it has happened in v4) about module shapes and cjs/esm interop. 

I've described the issue here - https://github.com/webpack/webpack/issues/7973 .

loganfsmyth:
> Depends on if we consider manual (not generated by @babel/plugin-transform-runtime) imports to @babel/runtime allowed or not.

I don't think I'd consider that valid, but I also wouldn't want to go out of my way to break it.

> If we allow it then people might use non-.default require in their CJS code and thus make it incompatible for module-aware webpack consumers.

That would only break if they are using the `auto` imports though, I think?

Andarist:
> That would only break if they are using the auto imports though, I think?

If you are referring the other auto behaviour (based on `supportsStaticESM`) then - not sure 😅 

When looking into the source of @babel/plugin-transform-runtime & @babel/helper-module-imports right now I see I was wrong - imports to those helper modules are inserted with `importedInterop: "uncompiled"` option, which from what I see means no interop helpers for those modules.

Let's look at possible scenarios - I'm narrowing it down only to libraries providing only CJS entry, because it's the only situation which is problematic.
1. default (`{useESModules: false, useAutoDirectory: false}`)
```js
var _classCallCheck = require("@babel/runtime/helpers/classCallCheck");

let Foo = function Foo() {
  "use strict";

  _classCallCheck(this, Foo);
};
```
Should be OK as `@babel/runtime/helpers/classCallCheck` provides only CJS export and webpack seeing `__esModule` on the exported object should just reach for `.default` which self-reference the helper

2. auto (`{useESModules: "auto", useAutoDirectory: false}`)
Basically the same as previous case, I dont see right now how it would differ - but I could have missed smth.

3. auto directory (`{useESModules: false, useAutoDirectory: true}`)
```js
var _classCallCheck = require("@babel/runtime/helpers/auto/classCallCheck");

let Foo = function Foo() {
  "use strict";

  _classCallCheck(this, Foo);
};
```
I think this won't work because webpack will think that imported module has `{default}` shape and treat `require(...)` as a require to namespace object instead of a require to a single default export.

To sum it up I think the only way to support it at the moment is to introduce a separate `strict` directory which would contain files such as:
```js
"use strict";

  value: true
});
module.exports.default = require('../createClass')
```
and point to that from `auto` directory. Ofc `addDefault` from @babel/helper-module-imports would have to be used with different mode (non-"uncompiled")

Andarist:
this is being superseded by https://github.com/babel/babel/pull/12295

PR #8897, opened 2018-10-18
Allow `defaults` query in preset-env

### LINKED ISSUE #7405: `defaults` browserslist query breaks preset-env
Bug.

`@babel/preset-env` does not work correctly when the browserslist query is `defaults` (a valid query). The result is that most of the expected transforms don't happen. It doesn't matter if the browserslist string is set via `package.json` `browserslist` field, or directly in `targets.browsers`.

### Input Code

```js
const test = () => ''
```

```js
{
  "presets": [
    [
      "@babel/env",
      {
        "targets": {
          "browsers": ["defaults"]
        }
      }
    ]
  ]
}
```

### Expected Behavior

```js
var test = function test() {
  return ''
}
```

### Current Behavior

```js
const test = () => ''
```

### Context

Unfortunately the [default behavior for `@babel/preset-env`](https://github.com/babel/babel/tree/master/packages/babel-preset-env#support-all-plugins-in-babel-that-are-considered-latest) if you do not provide targets or set a browserslist query is to replicate `@babel/preset-latest`. I want to transpile to the browserslist default/recommended level of browser support, without having to manually copy and paste whatever their default query is, which may change.

If you set `targets.browsers` to what `defaults` represents (`['> 0.5%', 'last 2 versions', 'Firefox ESR', 'not dead']`) then the output is correct. This means it is a bug with `@babel/preset-env` using the `defaults` browserslist query; not the actual transpilation.

### Your Environment

| ---------------- | -------
| Babel            |  7.0.0-beta.40
| node             | 9.5.0
| npm              | 5.6.0
| Operating System | macOS

existentialism:
| Fixed Issues?            | Fixes #7405
| Documentation PR Link    | 
| Any Dependency Changes?  |


nicolo-ribaudo:
Is there a feature which shouldn't be transformed when using Browserlist's default target? If so, we could add it to the tests, to test the difference between "default targets" and "no targets"

existentialism:
@nicolo-ribaudo part of the issue is, while we can handle usage of "defaults" when the query is passed directly as an option to the preset, if a user relies on browserslist to resolve the config (like via .browserslistrc or package.json) then our behavior of "no targets means the same as preset-latest" is difficult to do since we wont' be able to know whether the user didn't have targets or _wants_ `defaults`.

For now, seems like just documenting this difference seems okay?

danez:
I guess the best solution would be to check ourself if there is a browserlist config in the fs. We could do this with `browserlist.findConfig(from)`. If we find one we do not change defaults. But that would still be a breaking change so I guess we would need to wait for v8. https://github.com/browserslist/browserslist/blob/master/node.js#L238

nicolo-ribaudo:
I agree, we should do that in the next major release

PR #8910, opened 2018-10-20
Resolve babel.config.js 'babelrcRoots' values relative to the config file.

### LINKED ISSUE #8909: 'babelrcRoots' values passed in babel.config.js are resolved relative to the working directory, not the config file
## Bug Report

If you've got

```
babel.config.js // babelrcRoots: ['./packages/*']
packages/
  mod/
    module.js
    .babelrc.js
    // from here run `babel --rootMode upward module.js
```
it will fail to load the `.babelrc.js` because it will treat the root pattern as

```
packages/mod/packages/*
```

loganfsmyth:
| Fixed Issues?            | Fixes #8909
| Major: Breaking Change?  | Y, if you were relying on the old incorrect behavior :(
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

We were accidentally resolving these paths relative to the working directory, even though they were supposed to be relative to the config file itself.

I think this qualifies as a bug fix, but it does technically break people if they were relying on the `babel.config.js` value being resolved relative to the working directory. That said, our examples make it _relatively_ clear that it's expecting to be relative to the config file, so hopefully that isn't a real-world issue.

### DIFF
--- a/packages/babel-core/src/config/config-chain.js
+++ b/packages/babel-core/src/config/config-chain.js
@@ -155,6 +155,7 @@ export function buildRootChain(
   }
 
   let { babelrc, babelrcRoots } = opts;
+  let babelrcRootsDirectory = context.cwd;
 
   const configFileChain = emptyChain();
   if (configFile) {
@@ -168,6 +169,7 @@ export function buildRootChain(
       babelrc = validatedFile.options.babelrc;
     }
     if (babelrcRoots === undefined) {
+      babelrcRootsDirectory = validatedFile.dirname;
       babelrcRoots = validatedFile.options.babelrcRoots;
     }
 
@@ -185,7 +187,7 @@ export function buildRootChain(
   if (
     (babelrc === true || babelrc === undefined) &&
     pkgData &&
-    babelrcLoadEnabled(context, pkgData, babelrcRoots)
+    babelrcLoadEnabled(context, pkgData, babelrcRoots, babelrcRootsDirectory)
   ) {
     ({ ignore: ignoreFile, config: babelrcFile } = findRelativeConfig(
       pkgData,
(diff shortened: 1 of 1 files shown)


packages/babel-core/src/config/config-chain.js
```diff
@@ -168,6 +169,7 @@ export function buildRootChain(
       babelrc = validatedFile.options.babelrc;
     }
     if (babelrcRoots === undefined) {
+      babelrcRootsDirectory = validatedFile.dirname;
```
nicolo-ribaudo: Could you explain me how this PR fixes the issue?
If I understand it correctly, this line is the only change: if `babelrcRoots`  undefined, we use `validatedFile.dirname` instead of `context.cwd`, which is the default value of `babelrcRootsDirectory` (which is the same as the old behavior).
In the issue you said that it occurs when `babelrcRoots` is set to _something_, so I don't understand how this PR fixes it.

nicolo-ribaudo: > ```
> babelrcRoots: [
>    ".",
>]
> ```
>the expectation is that the . is relative to the config file, but currently we are calculating it relative to the working directory.

But doesn't this PR only changes the resolution behavior if `babelrcRoots === undefined`? If it is an array we still resolve relative to `context.cwd`, because we never go inside that `if`.

loganfsmyth: Oh, the `if (babelrcRoots === undefined) {` check here is if it was `undefined` in the `babel.transform` arguments, not if it was `undefined` in the config file.

With the way things are set up now, we entirely ignore `babelrcRoots` in the config file if it is passed in the programmatic arguments.

nicolo-ribaudo: Oh ok :+1:

sibelius:
does this fixes a possible issue with jest multi runner on lerna when trying to compile all packages?

Issue #8839, opened 2018-10-11
Wrong behaviour in "for...of" statement


macabeus:
## Bug Report

**Current Behavior**
Throws this exception:
```
TypeError: undefined: Duplicate declaration "i"
  1 | 
  2 | for (let i of 'foo') {
> 3 |   let i = 'a';
    |       ^
  4 |   console.log(i);
  5 | }
  6 | 
```

**Input Code**
```js
for (let i of 'foo') {
  let i = 'a';
  console.log(i);
}
```

**Expected behavior/code**
Should prints `'a'` three times instead of throws this exception.
This is the behaviour in Node and Safari, for example.

**Babel Configuration (CLI command)**

```js
const output = require('@babel/core').transform(`
for (let i of 'foo') {
  let i = 'a';
  console.log(i);
}
`, {
  presets: ['@babel/preset-env']
});
```

**Environment**
- Babel version(s): lastest
- Node/npm version: Node 9
- OS: macOS High Sierra
- Monorepo: yes
- How you are using Babel: `cli`

krovc:
This error have no connection with for...of. At least I think it's expected behaviour. So you can remove let from second line.

nicolo-ribaudo:
It's a bug, similar to https://github.com/babel/babel/issues/8498

macabeus:
@nicolo-ribaudo Yeah, I found this new bug when I was trying to resolve this another issue, but I could not resolve 😞

Do you have any idea how to resolve it? I'm not sure, but I think that this bug happens because [this line](https://github.com/babel/babel/blob/6b8261bbb7ebb8d2980cb30c3b3369678fd99835/packages/babel-plugin-transform-block-scoping/src/index.js#L736 ) not to extend correctly when it have two variables with the same name, or because [this function](https://github.com/babel/babel/blob/6b8261bbb7ebb8d2980cb30c3b3369678fd99835/packages/babel-plugin-transform-block-scoping/src/index.js#L691) should also to add declarations at `node.init` if the node is a `ForStatement` and make more changes at function `updateScopeInfo`.

byronluk:
So it seems like TypeError that you're running into is coming from [this line](https://github.com/babel/babel/blob/b5246994b57f06af871be6a63dcc4c6fd41d94d6/packages/babel-traverse/src/scope/index.js#L349).
[...]

Issue #8498, opened 2018-08-21
wrong behavior in for loop


uni-zheng:
## Bug Report

**Current Behavior**
Just print log once.

**Input Code**
- REPL or Repo link if applicable:

```js
for (let i = 0; i < 3; i++) {
  let i = 'abc';
  console.log(i);
}
```

**Output Code**
```js
"use strict";

for (var i = 0; i < 3; i++) {
  var i = "abc";
  console.log(i);
}
```

**Expected behavior/code**
print log 3 times

**Environment**
use https://new.babeljs.io/repl
set presets both es2015 and stage-3
version v6.26.0

**Possible Solution**
Should create temporary variable in body.

byronluk:
Hi I'm going to try to look at this. Thanks!

byronluk:
@nicolo-ribaudo Is it a better choice to add a closure inside the block scope or to rename the block scoped variable that is colliding with the for loop initializer?

nicolo-ribaudo:
Renaming the variable would be better, but if it is too complex we can just wrap the for body in a block, like in https://github.com/babel/babel/pull/8920, since it isn't a common case.

PR #8920, opened 2018-10-24
Create additional scope when redeclaring array variable in for..of loop

### LINKED ISSUE #8913: for...of array optimization can introduce conflicting variables into the scope
## Bug Report

Initially, I reported this bug here – https://github.com/facebook/create-react-app/issues/5387 – because I thought that something is broken in CRA. After further investigation, I found a bug in the Babel so here I am.

**Current Behavior**

The following error.

```js
index.js:5 TypeError: Cannot read property '0' of undefined
    at Object.to (index.js:5)
    at new t (index.js:5)
    at new t (index.js:5)
    at new t (index.js:5)
    at e.t (index.js:5)
    at e (index.js:5)
    at new e (index.js:5)
    at index.js:5
    at new Promise (<anonymous>)
    at Function.value (index.js:5)
(anonymous) @ index.js:5
Promise.catch (async)
(anonymous) @ index.js:5
n @ index.js:1
(anonymous) @ index.js:5
n @ index.js:1
(anonymous) @ index.js:1
(anonymous) @ index.js:1
```

**Input Code**

I prepared a repository that allows reproducing the issue: https://github.com/pomek/react-app/tree/babel

_`#master` branch is reserved for CRA, this bug was committed to the `#babel` branch._

But in order to speed up resolving the issue, I'll describe everything in the issue as well.

Here is my `./src/index.js`:

```js
import ClassicEditor from '@ckeditor/ckeditor5-build-classic';

ClassicEditor.create( document.getElementById( 'editor' ) )
	.then( editor => {
		console.log( 'Ready to use!', editor );
	} )
	.catch( err => {
		console.error( err );
	} );

```

and `./dist/index.html`:

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8">
    <title>CKEditor 5 – classic editor build – development sample</title>
    <style>
        body {
            max-width: 800px;
            margin: 20px auto;
        }
    </style>
</head>
<body>

<h1>CKEditor 5 – classic editor build – development sample</h1>

<div id="editor">
    <h2>Sample</h2>

    <p>This is an instance of the <a href="https://ckeditor.com/docs/ckeditor5/latest/builds/guides/overview.html#classic-editor">classic editor build</a>.</p>

    <p>You can use this sample to validate whether your <a href="https://ckeditor.com/docs/ckeditor5/latest/builds/guides/development/custom-builds.html">custom build</a> works fine.</p>
</div>

<script src="./index.js"></script>

</body>
</html>
```

**Expected behavior/code**

An application should work:

The working editor screen.

I'm using Webpack to build the app:

`./webpack.config.js`

```js
module.exports = {
	context: __dirname,
	entry: [
		'@babel/polyfill',
		'./src/index.js'
	],
	output: {
		filename: './index.js'
	},

	module: {
		rules: [
			{
				test: /\.js/,
				use: {
					loader: 'babel-loader',
					options: {
						presets: [
							[ '@babel/preset-env', {
								targets: {
									ie: 9,
								},
								ignoreBrowserslistConfig: true,
								useBuiltIns: false,
								modules: false,
								exclude: [ 'transform-typeof-symbol' ],
							} ]
						]
					}
				}
			}
		]
	}
};
```

Webpack's config looks similar to config used by CRA: https://github.com/facebook/create-react-app/blob/7ba343c187744d077b49cfd3b68b805e497bf3ac/packages/babel-preset-react-app/dependencies.js#L83-L103

`./package.json`

```json
{
  "name": "react-app",
  "version": "1.0.0",
  "description": "",
  "dependencies": {
    "@ckeditor/ckeditor5-build-classic": "^11.1.1"
  },
  "devDependencies": {
    "@babel/cli": "^7.1.2",
    "@babel/core": "^7.1.2",
    "@babel/polyfill": "^7.0.0",
    "@babel/preset-env": "^7.1.0",
    "@babel/runtime": "^7.1.2",
    "babel-loader": "^8.0.4",
    "webpack": "^4.20.2",
    "webpack-cli": "^3.1.2"
  },
  "scripts": {
    "build": "webpack --mode=production"
  }
}
```

After installing dependencies, calling `npm run build` and opening `./dist/index.html` you should have a working editor. Unfortunately, you'll end with the error.

**Environment**
- Babel version(s): 7.1.2 (?)
- Node/npm version: 10.12.0 / 6.4.1
- OS: `20:13:40Darwin MacBook-Pro.local 18.0.0 Darwin Kernel Version 18.0.0: Wed Aug 22 20:13:40 PDT 2018; root:xnu-4903.201.2~1/RELEASE_X86_64 x86_64`

**Possible Solution**

Is no a solution but I wanted to check whether it will work if `ckeditor5-build-classic` won't be minified so I did steps specified below:

1. Go to `@ckeditor` namespace in dependencies: `cd node_modules/@ckeditor`
2. Remove `ckeditor5-build-classic` installed from npm: `rm -r ckeditor5-build-classic`
3. Clone package from GitHub and check out to the proper version: `git clone -b v11.1.0 git@github.com:ckeditor/ckeditor5-build-classic.git`
4. Go to the package and install dependencies: `cd ckeditor5-build-classic && npm i`
5. Build the package in development mode: `./node_modules/.bin/webpack --mode development`
6. Go back to your application and call `npm run build` again.

It works!

kyle1320:
| Fixed Issues?            | Fixes #8913  
| Documentation PR Link    | 

When transforming a `for..of` loop that iterates through an array, redeclaring the array identifier within the loop would result in an ambiguity when assigning the iteration variable at the top of the transformed loop body, which could result in incorrect behavior.

This fix detects whether the array identifier is redeclared within the loop, and if so, wraps the original loop body in an additional block statement to prevent ambiguities when assigning the iteration variable.

### DIFF
--- a/packages/babel-plugin-transform-for-of/src/index.js
+++ b/packages/babel-plugin-transform-for-of/src/index.js
@@ -44,6 +44,11 @@ export default declare((api, options) => {
           }
 
           const block = t.toBlock(body);
+
+          if (path.get("body").scope.hasOwnBinding(right.name)) {
+            block.body = [t.blockStatement(block.body)];
+          }
+
           block.body.unshift(assignment);
 
           path.replaceWith(
@@ -139,6 +144,10 @@ export default declare((api, options) => {
 
     t.inherits(loop, node);
     t.ensureBlock(loop);
+
+    if (path.get("body").scope.hasOwnBinding(right.name)) {
+      loop.body.body = [t.blockStatement(loop.body.body)];
+    }
 
     const iterationValue = t.memberExpression(
       t.cloneNode(right),
--- a/packages/babel-plugin-transform-for-of/test/fixtures/for-of-as-array/for-of-redeclare/input.js
+++ b/packages/babel-plugin-transform-for-of/test/fixtures/for-of-as-array/for-of-redeclare/input.js
@@ -1,0 +1,3 @@
+for (let o of arr) {
+  const arr = o;
+}
--- a/packages/babel-plugin-transform-for-of/test/fixtures/for-of-as-array/for-of-redeclare/output.js
+++ b/packages/babel-plugin-transform-for-of/test/fixtures/for-of-as-array/for-of-redeclare/output.js
@@ -1,0 +1,6 @@
+for (let _i = 0, _arr = arr; _i < _arr.length; _i++) {
+  let o = _arr[_i];
+  {
+    const arr = o;
+  }
+}
--- a/packages/babel-plugin-transform-for-of/test/fixtures/regression/redeclare-array-8913/input.js
+++ b/packages/babel-plugin-transform-for-of/test/fixtures/regression/redeclare-array-8913/input.js
@@ -1,0 +1,5 @@
+function f(...t) {
+  for (let o of t) {
+    const t = o;
+  }
+}
--- a/packages/babel-plugin-transform-for-of/test/fixtures/regression/redeclare-array-8913/output.js
+++ b/packages/babel-plugin-transform-for-of/test/fixtures/regression/redeclare-array-8913/output.js
@@ -1,0 +1,8 @@
+function f(...t) {
+  for (var _i = 0; _i < t.length; _i++) {
+    let o = t[_i];
+    {
+      const t = o;
+    }
+  }
+}


pomek:
Seems that this PR fixes the issue. But I used a _primitive_ method to check this out:

1. I found REPL that should fail (taken from https://github.com/babel/babel/pull/8926).
2. I Pasted the entire content of the editor's build: https://github.com/ckeditor/ckeditor5-build-classic/blob/cf4681772537874c5143b8148c3ebbab1f46e6b8/build/ckeditor.js
3. Copied build from the right window on REPL.
4. I modified a sample https://github.com/ckeditor/ckeditor5-build-classic/blob/cf4681772537874c5143b8148c3ebbab1f46e6b8/sample/index.html:

    - `<script src="../build/ckeditor.js"></script>` was replaced with the code produced by REPL
    - I needed to import `babel-regenerator-runtime` so I found a CDN with that:

    ```html
    <script src="https://cdn.jsdelivr.net/npm/babel-regenerator-runtime@6.5.0/runtime.min.js"></script>
    ```

5. As I expected, an error mentioned in the issue (#8913) still appears.
6. Now I am going to use REPL from this PR.
7. It just works! :tada: 

I am going to apply changes from this PR to my package inside the `node_modules/` directory and will check whether it works with Webpack.

pomek:
I applied the patch to package installed locally and it works! ❤️

packages/babel-plugin-transform-for-of/src/index.js
```diff
@@ -140,6 +145,10 @@ export default declare((api, options) => {
     t.inherits(loop, node);
     t.ensureBlock(loop);
```
loganfsmyth: This is a great effort, thanks so much, the main issues are

1. Technically there can be multiple names, and we need to check if _any_ of them collide `for (const {a, b} of arr) { const b = 4; }`
2. Rather than assigning to `block.body` it would be better to replace the `path.node.body` node, I think

Something along the lines of
```
const iterationKey = scope.generateUidIdentifier("i");

const body = path.get("body");

const hasKeyConflict = 
  body.isBlockStatement() &&
  path.getBindingIdentifiers().some(id => body.scope.hasBinding(id));

let body = hasKeyConflict 
  ? t.blockStatement([node.body])
  : node.body;

let loop = buildForOfArray({
  BODY: body,
```
I forget the exact return value of `path.getBindingIdentifiers()` so you might need to tweak that.

Also, we should make sure this also adds the block wrapper for code like

```
let a, b;
for ({a, b} of arr) {
  const b = 4;
}
```
since that would also collide. I think this might just work automatically with `getBindingIdentifiers` but please check if you can.

kyle1320: Thanks for the feedback!

As for your first suggestion, I suppose that comes down the scope of the PR. The original issue was caused by a reuse of the array identifier, not the iteration variable. This was causing the array lookup to use the wrong identifier when the inner variable was renamed.

There are a couple of other issues open which specifically deal with the iteration variables being redeclared, namely #8498 and #8839.

I'm rather new to this whole game, so I'm not sure whether it would be preferred to fix those issues in this PR as well, since it is similar, or just leave them to someone else. This PR has already been referenced in #8498 as a related solution. But I'll defer back to you on that.

As for your second point, the reason I used `block.body` was because `block` is guaranteed to be a block statement, whereas `path.node.body` is not. If there is some other reason this is preferred, then I suppose the code could be changed to the following:
```javascript
if (path.get("body").scope.hasOwnBinding(right.name)) {
    path.node.body = t.blockStatement([t.toBlock(body)]);
}
const block = t.toBlock(path.node.body);
```
But this seems a bit underhanded to me.

getnashty:
Any word on merging this?

Timer:
Any update on this? We've had one or two Create React App users hit this. 😄

danez:
This seems it would be fixed by #9697

PR #8937, opened 2018-10-28
rename colliding let bindings with for loop init

### LINKED ISSUE #8498: wrong behavior in for loop

byronluk:
| Fixed Issues?            | Fixes #8498 
| Documentation PR Link    | 

Babel transform block scoping doesn't remap let declarations if there's a collision in the loop body with the loop init variable. OutsideLetReferences only keeps track of the loop's init variable so I'm checking the scope of the block path for any colliding variable and renaming it.

This only checks the loop's init variable against the bindings in the loop's body -- so I believe there won't be any additional/unnecessary renaming happening.

Lmk if there's any additional test cases I should add or if there's anything I could do better. Thanks!


byronluk:
Added your example, will think of other test cases. Thanks! @nicolo-ribaudo

packages/babel-plugin-transform-block-scoping/test/fixtures/general/issue-8498-loop-init-collision-destructuring/output.js
```diff
@@ -0,0 +1,5 @@
+for (let a, _ref = {}, b = _ref.b;;) {
+  let a,
```
nicolo-ribaudo: You should enable the block scoping plug-in, otherwise we are testing the wrong thing 😛

byronluk: Ahh, I thought the package plugin was enabled by default since the test is in that directory... 🤔  I'm guessing it's okay to leave the transform destructuring plugin? w/o the transform destructuring plugin, the destructured variable gets incorrectly renamed regardless.
```
var { i } = {};
{
  var { i: _i } = {};
}
```

nicolo-ribaudo:
Thank you!

PR #8950, opened 2018-10-31
Fix message when plugin of a wrong type is passed

everdimension:

Display a correct message when a plugin is passed to "presets" and when a preset is passed to "plugins".


existentialism:
@everdimension thanks!

PR #8956, opened 2018-10-31
Do not allow TypeCastExpressions w/o parens 

### LINKED ISSUE #6698: Tighten up parsing of Flow TypeCasts
> Issue originally reported by @DrewML in babel/babylon#413

### Input Code

```js
state.set('foo': bar)
```

### Babylon/Babel Configuration

Flow syntax plugin enabled

### Expected Behavior
Should be a syntax error

### Current Behavior
We're currently parsing `bar` as a TypeCastExpression w/ a value of `foo`. Flow requires that you wrap a TypeCastExpression in parens, and we're allowing the parens for the CallExpression to be used right now, which the Flow parser does _not_ allow.

Context: https://twitter.com/schep_/status/841741581589860355

### LINKED ISSUE #6718: Flow typecasts parsed too leniently
> Issue originally reported by @motiz88 in babel/babylon#185

Code like `(A, B: T)` is a [syntax error in Flow](https://astexplorer.net/#/0EUmRybXf9), but Babel happily parses it as a SequenceExpression with a TypeCastExpression as its second element. More generally, within a parenthesized list anything of the form `A: B` tends to be parsed as a cast, which is incorrect.

Reference: [Flow blog post on typecasts](https://flowtype.org/blog/2015/02/18/Typecasts.html).

danez:
| Fixed Issues?            | Fixes #6718 Fixes #6698 
| Patch: Bug Fix?          |👍
| Major: Breaking Change?  |sort of
| Documentation PR Link    | 

This adds an additional check to disallow flow type casts without parens inside of sequence expressions.

It also enables another check which was already present but never executed.

It is technically a breaking change, but as flow does not allow this, it is probably safe to assume nobody is using this. Or are we now being more strict?

I also tried using babel/babylon#446 (which also solves this problems), but the detection of arrow functions had lots of edge cases and ambiguities and overall seemed harder to maintain than the current parsing of arrows.


danez:
I just saw there is also #6698 which is similar but not solved by these. I will look if there is a better solution to get both of these issues sorted out in one PR.

packages/babel-parser/src/plugins/flow.js
```diff
@@ -1932,36 +1961,16 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       for (let i = 0; i < exprList.length; i++) {
```
nicolo-ribaudo: Nit: we can use for/of here, since we compile it using the `assuneArray` option (`for (a of b)` -> `for (let i = 0; i < b.length ..`)

danez: I leave that for now, i didn't want to make the commit diff bigger. Can do that in a separate PR.

danez:
@nicolo-ribaudo Thanks for the very helpful input. I changed it and it is a lot cleaner and easier now, and still works the same. 👍

PR #8966, opened 2018-11-03
Update test262 to latest commit and enable mapping for features

danez:
| Fixed Issues?            |  n
| Documentation PR Link    | 
| Any Dependency Changes?  |

Also added an automated check for new features which are not mapped or ignored and print these at the end of the run. So in the future when new tests with new features are added to test262 we get a warning and can correctly ignore or map it to a parser plugin.

Lots of new failed tests added to whitelist, especially for private class members and also some dynamic imports.


nicolo-ribaudo:
> Also added an automated check for new features which are not mapped or ignored and print these at the end of the run. So in the future when new tests with new features are added to test262 we get a warning and can correctly ignore or map it to a parser plugin.

Thank you, it was easy to forget to add new features.

PR #8986, opened 2018-11-07
Export @babel/parser#tokTypes in @babel/core

kaicataldo:
| Fixed Issues?            | Refs https://github.com/babel/babel-eslint/pull/711/files#r231363345

This exposes `@babel/parser` in `@babel/core`, as proposed [here](https://github.com/babel/babel-eslint/pull/711/files#r231363345). Figured it would be best to just make a PR that we could discuss on :) I wasn't sure if testing deeper than this would be worth it or not, given it would have to change any time `@babel/parser`'s API changes, but happy to update as the team sees fit. Thanks!

### DIFF
--- a/packages/babel-core/src/config/helpers/config-api.js
+++ b/packages/babel-core/src/config/helpers/config-api.js
@@ -55,6 +55,7 @@ export default function makeAPI(
     async: () => false,
     caller,
     assertVersion,
+    tokTypes: undefined,
   };
 }
 
--- a/packages/babel-core/src/index.js
+++ b/packages/babel-core/src/index.js
@@ -10,7 +10,8 @@ export { version } from "../package.json";
 export { getEnv } from "./config/helpers/environment";
 
 export * as types from "@babel/types";
-export * as parser from "@babel/parser";
+export { tokTypes } from "@babel/parser";
+
 export { default as traverse } from "@babel/traverse";
 export { default as template } from "@babel/template";
 
--- a/packages/babel-core/src/index.js
+++ b/packages/babel-core/src/index.js
@@ -10,6 +10,7 @@ export { version } from "../package.json";
 export { getEnv } from "./config/helpers/environment";
 
 export * as types from "@babel/types";
+export * as parser from "@babel/parser";
 export { default as traverse } from "@babel/traverse";
 export { default as template } from "@babel/template";
 
--- a/packages/babel-core/test/api.js
+++ b/packages/babel-core/test/api.js
@@ -154,6 +154,14 @@ describe("api", function() {
     );
   });
 
+  it("exposes the parser", function() {
+    expect(babel.parser).toBeDefined();
+  });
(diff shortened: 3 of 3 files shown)


nicolo-ribaudo:
Doesn't `babelCore.parse` work for your usecase?

nicolo-ribaudo:
Oh nvm, I saw the babel-eslint pr

kaicataldo:
@nicolo-ribaudo For the PR I'm working on, all I actually need is `babelParser.tokTypes`. The idea was first brought up by @loganfsmyth [here](https://github.com/babel/babel-eslint/pull/594#discussion_r169840591), but I know a lot has changed since then.

I do think it's weird/confusing to have `babelCore.parse` and `babelCore.parser.parse` though 😬

loganfsmyth:
Revisiting this after my one-off comment that prompted this, I'm concerned because everything we add to the root by default is exposed in plugins/presets on the `api` object, and this doesn't seem like something we'd actually want there, especially since we already expose `template` as the suggested way to create ASTs from code snippets, and `parse` as a way to parse using the user's config.

I'm not necessarily against this being exported from `@babel/core`, but I'd want to avoid exposing on `api`. I do wonder if we'd be better off exporting `tokTypes` directly instead. I didn't realize until now what `tokTypes` even was. It's really unfortunate that the `tokens` array relies on object identity for checking token types :(

xtuc:
I agree with Logan, and I'm not a big fan of this implicitly exposed API. Would it be possible to change the API to be a fixed object? and then you can expose the parse function.

kaicataldo:
@xtuc For this PR I actually need access to `tokTypes` in `@babel/parser`, but I mainly made this PR because @loganfsmyth had requested it. I'm fine with just closing this and doing what we're doing now. Or would you prefer for me to leave this open as a place of discussion about how we should move forward?

ljharb:
I would greatly appreciate being able to use the parser in a transform without needing an explicit dependency on it; it would be great to find a way to do that.

xtuc:
As @ljharb mentioned, exposing the parser in a transform and correctly configured does solve a problem. 

I would be in favor of leaving this open for discussion, because we should figure a way to do it I guess.

kaicataldo:
Actually, seems like an issue would be a better way to track this discussion. At this point, I'm not going to pursue this change since it's not necessary for the babel-eslint PR.

loganfsmyth:
@kaicataldo Can we just re-export `tokTypes` on the object for now, and add `tokTypes: undefined` to https://github.com/babel/babel/blob/efa571a42cf33d561f00dacc3cb32aba4b2a3edb/packages/babel-core/src/config/helpers/config-api.js#L50 so the value will be overwritten with a blank value? Then we can always go in later to trim down the plugin API.

kaicataldo:
@loganfsmyth Apologies for the delay - PR updated.

PR #8997, opened 2018-11-08
Fix cloneNode with typeAnnotation.

### LINKED ISSUE #8996: types.cloneNode does not preverse type annotations
## Bug Report

**Current Behavior**
Calling `types.cloneNode` on a node containing a `typeAnnotation` does not preserve it.

**Expected behavior/code**
It should preserve `typeAnnotation`.

**Environment**
- Babel version(s): [v7.1.2]

**Possible Solution**
Clone typeAnnotation here: https://github.com/babel/babel/blob/master/packages/babel-types/src/clone/cloneNode.js#L64

Actually it is a problem because `@babel/template` is using `types.cloneNode`, the result is that it is not possible to have a type annotation using Babel template.

The original use-case is to be able to create TypeScript template for SVGR (see https://github.com/smooth-code/svgr/issues/226).

neoziro:
| Fixed Issues?            | `Fixes #8996` 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

See #8996 for context.


packages/babel-types/src/clone/cloneNode.js
```diff
@@ -66,6 +66,9 @@ export default function cloneNode<T: Object>(node: T, deep: boolean = true): T {
       ...node.extra,
     };
   }
+  if (has(node, "typeAnnotation")) {
```
loganfsmyth: This should be up after line `39`, since I believe it would be the `Identifier` optimization there that causes the problem. The field logic on `43` should already handle it for other types.

This will also need to recursively clone when `deep` is set, like the logic on `45` does.

packages/babel-types/test/cloning.js
```diff
@@ -56,4 +56,17 @@ describe("cloneNode", function() {
...
+      t.variableDeclarator({
+        ...t.identifier("value"),
+        typeAnnotation: t.anyTypeAnnotation(),
+      }),
+    ]);
+    const cloned = t.cloneNode(node, /* deep */ true);
+    expect(cloned.declarations[0].id.typeAnnotation).toEqual(
+      node.declarations[0].id.typeAnnotation,
+    );
+  });
```
nicolo-ribaudo: We should also test that `expect(cloned.declarations[0].id.typeAnnotation).not.toBe(node.declarations[0].id.typeAnnotation);
`

neoziro:
All done, ready to merge!

loganfsmyth:
Thanks!

PR #9001, opened 2018-11-09
prettier@1.15.1

existentialism:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-parser/src/tokenizer/index.js
```diff
@@ -881,10 +881,10 @@ export default class Tokenizer extends LocationParser {
       radix === 16
         ? allowedNumericSeparatorSiblings.hex
         : radix === 10
-          ? allowedNumericSeparatorSiblings.dec
-          : radix === 8
-            ? allowedNumericSeparatorSiblings.oct
-            : allowedNumericSeparatorSiblings.bin;
+        ? allowedNumericSeparatorSiblings.dec
+        : radix === 8
+        ? allowedNumericSeparatorSiblings.oct
```
loganfsmyth: Hmm, this seems to make some of these solidly unreadable. Can we see if this is expected, or a bug?

existentialism: Expected: https://prettier.io/blog/2018/11/07/1.15.0.html#flatten-else-branch-for-nested-ternaries-5039-by-suchipi-5272-by-duailibe-5333-by-ikatyang

loganfsmyth: Alright, that's life with Prettier then :P I always thought the excessive nesting was a nice reminder that hey nested ternaries are usually a bad idea. Now they are easy to miss. Alas.

PR #9008, opened 2018-11-09
[decorators] Add support for initializers

nicolo-ribaudo:
tc39/proposal-decorators@cc2db18b5215c91fc86b02a03f20506b7c9c6365

| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |


nicolo-ribaudo:
TODO: I forgot to update the helper version

packages/babel-helpers/src/helpers.js
```diff
@@ -1358,17 +1363,27 @@ helpers.decorate = helper("7.1.5")`
...
+      if (
+        element.kind === "method" &&
+        (placement === "static" || placement === "prototype")
+      ) {
+        var receiver = placement === "static" ? F : proto;
+        _defineClassElement(receiver, element);
+      }
+    });
+
+    elements.forEach(function(element /*: ElementDescriptor */) {
```
danez: Why two loops instead of combining the two?

nicolo-ribaudo: Becose they differ both in the if condition and in the behavior inside the if.

nicolo-ribaudo: I just noticed that I misinterpreted your question 🙃

The proposal says that all methods must be defined before the fields/initializers. If I used a single loop, that order wouldn't be maintained.

rushjoe18:
Thanks

littledan:
The new test of ordering looks good to me. I'd say, this patch is ready to land.

nicolo-ribaudo:
When is the next meeting? I can enable it using a flag in the meantime

nicolo-ribaudo:
We discussed this at the meeting today and decided to wait for the TC39 meeting, since we don't want to introduce a flag which we might deprecate in two weeks.

nicolo-ribaudo:
@littledan 	What is the meeting outcome?

nicolo-ribaudo:
Blocked by tc39/proposal-decorators#182

littledan:
@nicolo-ribaudo We seem to be in agreement about adding the feature, but a name bikeshed remains.

nicolo-ribaudo:
The feature is now enabled behind an option: `["@babel/plugin-proposal-decorators", { "initializers": true }]`, thanks to #9166. That way, if the name will change it will be easy to update Babel :smile:

nicolo-ribaudo:
I will open a different PR for "hooks"

PR #9025, opened 2018-11-15
Make babel-standalone an ESModule and enable flow

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

As a requirement for #7581, but it is also not abad thing to do for webpack so treeshaking and module concatenation can work.
There also should not be any problems with circle dependencies as standalone is the entry point for webpack and is nowhere referenced.


packages/babel-standalone/src/plugins.js
```diff
@@ -0,0 +1,244 @@
+// @flow
+export { default as externalHelpers } from "@babel/plugin-external-helpers";
```
nicolo-ribaudo: Is it possible to auto-generate this file and the list in packages/babel-standalone/src/index.js?

danez: I will do that.

nicolo-ribaudo: Note that we shouldn't generate it from the list of the directories but we still need to maintain a list of included plugins (which wouldn't be duplicated anymore).

jasonLaster:
Looks good. I like the flow types. Out of curiosity could those land separately?

danez:
Okay I finally added the autogeneration.

packages/babel-standalone/scripts/pluginConfig.json
```diff
@@ -0,0 +1,87 @@
+{
+  "all": [
```
nicolo-ribaudo: Can't this file just be the array?

nicolo-ribaudo:
The failing status of this PR is just because the hack I used to create the publish github action without support for "release" events in publis repos has some problems when PRs are opened from branches of this repo.

packages/babel-standalone/scripts/pluginConfig.json
danez: Oh yes. I first had more keys in there, as I wanted to also specify which plugins go to which preset, but that does not work because of plugin options and other business logic.
Will revert.

danez: Done

PR #9035, opened 2018-11-17
Fix parsing typescript function types with destructuring

### LINKED ISSUE #8099: Syntax error with a TS interface using syntax-typescript
## Bug Report

**Current Behavior**
Syntax error with some TS interfaces and `@babel/plugin-syntax-typescript`

**Input Code**
```ts
interface myInterface  {
  method: ({test}: {test:string}) => string;
}
```

```
ERROR
Unexpected token, expected ")" (2:17)

 |   method: ({test}: {test:string}) => string;
 |                  ^
```

**Expected behavior/code**
The above code snippet should compile. The whole `interface` block should be removed by `@babel/plugin-syntax-typescript`.

**Environment**
- Babel version(s): v7.0.0-beta.42
- OS: Ubuntu 18.04

### LINKED ISSUE #8745: [TypeScript] Unexpected token, expected ")"
## Bug Report

**Current Behavior**
Code fails with Babel, works with TSC.

**Input Code**

  - [Babel REPL](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=C4TwDgpgBAsiAq5oF4oAoDeVgAsIFtoBfALii1wIjIGdgAnASwDsBzKIgSimQD4o6TNgChhAYwD2zOlAAeZOIkg9yRIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=typescript&prettier=false&targets=&version=7.0.0-beta.54&envVersion=) (ps: needs to install the `@babel/plugin-transform-typescript` plugin)
  - [TypeScript Playground](http://www.typescriptlang.org/play/#src=type%20MyType%20%3D%20(%7B%20theme%20%7D%3A%20%7B%20theme%3A%20string%20%7D)%20%3D%3E%20string%0D%0A%0D%0Aconst%20x%3A%20MyType%20%3D%20%7B%7D)

This code fails on Babel but not on TSC:
```ts
type MyType = ({ theme }: { theme: string }) => string

const x: MyType = {}
```

This code works on both:
```ts
type MyType = (param: { theme: string }) => string

const x: MyType = {}
```

**Expected behavior/code**

Support this syntax, because tsc supports it.

The one inside Create React App v2, plus typescript preset. ([file](https://github.com/brunolemos/create-react-app/blob/next-typescript/packages/babel-preset-react-app/create.js))

**Environment**
- Babel version(s): 7.1.0
- Node/npm version: 5.6.0
- OS: macOS 10.14
- Monorepo yes (yarn workspace)
- How you are using Babel: CRA

**Additional context**

Possibly related: #8365

danez:
| Fixed Issues?            | Fixes #8745, fixes #8099 
| Documentation PR Link    | 
| Any Dependency Changes?  |

Allows parsing of destructured params in function type declarations.

### DIFF
--- a/packages/babel-parser/src/plugins/typescript.js
+++ b/packages/babel-parser/src/plugins/typescript.js
@@ -314,13 +314,19 @@ export default (superClass: Class<Parser>): Class<Parser> =>
     }
 
     tsParseBindingListForSignature(): $ReadOnlyArray<
-      N.Identifier | N.RestElement,
+      N.Identifier | N.RestElement | N.ObjectPattern,
     > {
       return this.parseBindingList(tt.parenR).map(pattern => {
-        if (pattern.type !== "Identifier" && pattern.type !== "RestElement") {
+        if (
+          pattern.type !== "Identifier" &&
+          pattern.type !== "RestElement" &&
+          pattern.type !== "ObjectPattern"
+        ) {
           throw this.unexpected(
             pattern.start,
-            "Name in a signature must be an Identifier.",
+            `Name in a signature must be an Identifier or ObjectPattern, instead got ${
+              pattern.type
+            }`,
           );
         }
         return pattern;
@@ -747,6 +753,22 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         this.next();
         return true;
       }
+
+      if (this.match(tt.braceL)) {
+        const braceStack = [tt.braceL];
+        this.next();
+
+        while (braceStack.length > 0) {
+          if (this.match(tt.braceL)) {
+            braceStack.push(tt.braceL);
+          } else if (this.match(tt.braceR)) {
+            braceStack.pop();
+          }
+          this.next();
+        }
+        return true;
+      }
+
       return false;
     }
 
--- a/packages/babel-parser/src/plugins/typescript.js
+++ b/packages/babel-parser/src/plugins/typescript.js
@@ -755,14 +755,14 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       }
 
       if (this.match(tt.braceL)) {
-        const braceStack = [tt.braceL];
+        let braceStackCounter = 1;
         this.next();
 
-        while (braceStack.length > 0) {
+        while (braceStackCounter > 0) {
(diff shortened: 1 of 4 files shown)


packages/babel-parser/src/plugins/typescript.js
```diff
@@ -747,6 +753,22 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         this.next();
         return true;
       }
+
+      if (this.match(tt.braceL)) {
+        const braceStack = [tt.braceL];
```
nicolo-ribaudo: This can be an integer instead of an array, since the check only uses it's length.

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -747,6 +753,22 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+        this.next();
+
+        while (braceStack.length > 0) {
```
nicolo-ribaudo: Is it safe to skip the code between braces without parsing it?

e.g., wouldn't this loop accept something like `{bar: {}{foo = 2=>{} }}`?

packages/babel-parser/src/plugins/typescript.js
danez: changed

packages/babel-parser/src/plugins/typescript.js
danez: It does ignore everything, even if its invalid, but this function is only used in a lookahead() and it's job is to try to skip the first param after `(`.

it is used here: https://github.com/babel/babel/blob/be591698cd7238e8e5b8e682c7b82ad740527712/packages/babel-parser/src/plugins/typescript.js#L775

and this in turn is used here:
https://github.com/babel/babel/blob/be591698cd7238e8e5b8e682c7b82ad740527712/packages/babel-parser/src/plugins/typescript.js#L741

nicolo-ribaudo: Oh ok :+1:

PR #9050, opened 2018-11-20
Remove direct options to @babel/plugin-class-features

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | :warning: We should merge this before the next release
| Minor: New Feature?      |
| Tests Added + Pass?      | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR removes the ability of using the class-features plugin like this:

```json
{
  "plugins": [
    ["@babel/plugin-class-features", { "decorators": true }]
  ]
}
```

Since it could cause problems because it behaves differently from how other plugins behave (see https://github.com/babel/babel/pull/8130#discussion_r231804956).
With this PR only the old packages (`@babel/plugin-proposal-class-properties`, etc) can be used to configure `@babel/plugin-class-features`.

### DIFF
--- a/packages/babel-plugin-class-features/src/index.js
+++ b/packages/babel-plugin-class-features/src/index.js
@@ -27,60 +27,15 @@ export { enableFeature, FEATURES, setLoose };
 const version = pkg.version.split(".").reduce((v, x) => v * 1e5 + +x, 0);
 const versionKey = "@babel/plugin-class-features/version";
 
-const getFeatureOptions = (options, name) => {
-  const value = options[name];
-
-  if (value === undefined || value === false) return { enabled: false };
-  if (value === true) return { enabled: true, loose: false };
-
-  if (typeof value === "object") {
-    if (
-      typeof value.loose !== "undefined" &&
-      typeof value.loose !== "boolean"
-    ) {
-      throw new Error(`.${name}.loose must be a boolean or undefined.`);
-    }
-
-    return { enabled: true, loose: !!value.loose };
-  }
-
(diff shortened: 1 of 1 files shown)


Andarist:
What's going to happen when somebody is going to use this plugin directly? As in:
```js
{
  "plugins": [
    "@babel/plugin-class-features"
  ]
}
```

nicolo-ribaudo:
It is a noop. I could throw if you prefer, but I don't think that it is needed.

Andarist:
IMHO would be good to throw or at least warn the user - somebody mind find this plugin and try to use it manually and be surprised that it's not working.

nicolo-ribaudo:
Before that `@babel/plugin-class-features` you had a configuration similar to this:

```json5
{
  plugins: [
    "@babel/plugin-proposal-private-properties",
    "@babel/plugin-proposal-decorators"
  ]
}
```

Now you should have the same identical config:
```json5
{
  plugins: [
    "@babel/plugin-proposal-private-properties",
    "@babel/plugin-proposal-decorators"
  ]
}
```

The two plugins internally use `@babel/plugin-class-features`, but you shouldn't care since it is an implementation detail.

nicolo-ribaudo:
@babel/babel If it won't be possible to use this plugin directly, I think that its exisence only creates confusion. WDYT about renaming it to `@babel/helper-transform-class-features`?

nicolo-ribaudo:
Closing, #9083 is better

PR #9058, opened 2018-11-21
Fix compatibility between typescript and jsx plugins in interface declarations

### LINKED ISSUE #7742: [TypeScript] Generic callable interface is not parsed correctly
In TypeScript, it is possible to have callable interfaces and in particular the call signature can include generic type parameters. This might look like this:

```ts
interface Foo {
    <T>(bar: T): T;
}
```

When using the latest Beta version (v7.0.0-beta.44 at the time of writing) with `preset-typescript` and `preset-react`, this causes the following error:

```
repl: Unexpected token (2:2)

  1 | interface Foo {
> 2 |   <T>(bar: T): T;
    |   ^
  3 | }
```

When using only `preset-typescript`, this works as expected.

Also see this in the REPL [here](https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&code_lz=JYOwLgpgTgZghgYwgAgGIHt3IN4ChnIA8AKgHwAUARnFAFzLECU9xA3LgL5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=true&presets=react%2Ctypescript&prettier=false&targets=&version=7.0.0-beta.44&envVersion=).

Clearly this is an incompatibility between the two presets, I suspect `preset-react` gets to the `<` first, before `preset-typescript` has a chance to properly parse it.  
I can't think of any reason there would be JSX inside an interface definition, so it might be possible to temporarily 'disable' `preset-react` while inside such a scope?

A possible workaround is to not define the interface as callable and then build a union type with a function:

```ts
interface Foo_ {
  // Other properties...
}
type Foo = Foo_ & (<T>(bar: T) => T);
```

danez:
| Fixed Issues?            | Fixes #7742
| Documentation PR Link    | 

Ensure we enable `inType` when parsing interfaces. There might be more cases like this with enums or abstract classes, but i couldn't find any broken cases easily.

### DIFF
--- a/packages/babel-parser/src/plugins/typescript.js
+++ b/packages/babel-parser/src/plugins/typescript.js
@@ -930,7 +930,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         node.extends = this.tsParseHeritageClause();
       }
       const body: N.TSInterfaceBody = this.startNode();
-      body.body = this.tsParseObjectTypeMembers();
+      body.body = this.tsInType(this.tsParseObjectTypeMembers.bind(this));
       node.body = this.finishNode(body, "TSInterfaceBody");
       return this.finishNode(node, "TSInterfaceDeclaration");
     }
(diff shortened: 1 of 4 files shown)


nicolo-ribaudo:
> There might be more cases like this with enums or abstract classes, but i couldn't find any broken cases easily.

If someone finds other cases, we could use this PR as an exampe and mark them as beginner-friendly.

Issue #2279, opened 2015-08-28
`super()` should initialize the `this` binding with whatever value is returned from parent constructor


loganfsmyth:
The current implementation doesn't handle parent classes that return specific objects properly. The value returned from `[[Construct]]` should be used as the `this` binding value in the child constructor.

```
class Other {
    method(){
        console.log('other');
    }
}
class Base {
    constructor(){
        return new Other();
    }
    method(){
        console.log('base');
    }
}
class Sub extends Base {
    constructor(){
        super();
        this.method();
    }
}

var s = new Sub();
console.log('expected true, got ' + (s instanceof Other));
```

Seems like

```
var Sub = (function (_Base) {
    _inherits(Sub, _Base);

    function Sub() {
        _classCallCheck(this, Sub);

        _get(Object.getPrototypeOf(Sub.prototype), 'constructor', this).call(this);
        this.method();
    }

    return Sub;
})(Base);
```

should be more like

```
var Sub = (function (_Base) {
    _inherits(Sub, _Base);

    function Sub() {
        _classCallCheck(this, Sub);

        var _this = _get(Object.getPrototypeOf(Sub.prototype), 'constructor', this).call(this);
        if (_this === undefined) _this = this;
        _this.method();
        return _this;
    }

    return Sub;
})(Base);
```

katrotz:
Might be the same issue. 
I am trying to extend the [SystemJS constructor](https://github.com/systemjs/systemjs/blob/master/docs/system-api.md#systemconstructor) like:

```
class MyFancySubclass extends System.constructor {
  constructor() {
    super();
    ...
  }
}
```
but in the end the `System.constructor` is not getting called when I do `super()`.

I have traced the potential issue in the babels `super` implementation:
The `inherits` babel runtime helper maps the constructor of the extended class on the `__proto__` of the subclass, but when the `super` is being invoked in the subclass constructor I see the following code:
`_get(Object.getPrototypeOf(MyFancySubclass.prototype), 'constructor', this).call(this);
`
which indeed will return the superClasses prototype, but not the superClasses constructor.

katrotz:
I it might be related to https://github.com/babel/babel/issues/2048

loganfsmyth:
[...]

PR #9060, opened 2018-11-22
Not depending on return value of super(). Fixes #9020.

### LINKED ISSUE #9020: Babel rely on value returned by super() but this is not reliable with Edge
## Bug Report

**Current Behavior**

When using some transpiled code, babel rely on `_this` from  `_this = super()` but this bug on Edge is still present with latest version https://github.com/Microsoft/ChakraCore/issues/4663

**Input Code**

```js
export class App extends React.Component {
    state = {
        fruit: 'banana'
    }

    handleClick = (event, ...rest) => {
        this.setState({fruit: 'apple'})
    }

    render() {
        const {fruit} = this.state
        return (
            <h1 onClick={this.handleClick}>I'm a {fruit}</h1>
        )
    }
}
```

become with preset set on `targets: {esmodules: true}`

```jsx
class App extends react__WEBPACK_IMPORTED_MODULE_0___default.a.Component {
  constructor() {
    var _this;

    _this = super(...arguments);
    this.state = {
      fruit: 'banana'
    };

    this.handleClick = function (event) {
      _this.setState({
        fruit: 'apple'
      });
    };
  }

  render() {
    const fruit = this.state.fruit;
    return react__WEBPACK_IMPORTED_MODULE_0___default.a.createElement("h1", {
      onClick: this.handleClick,
      __source: {
        fileName: _jsxFileName,
        lineNumber: 16
      },
      __self: this
    }, "I'm a ", fruit);
  }

}
```

**Expected behavior/code**

The initial behavior is that `_this` should not be `undefined` but I think that is in the hands of ChakraCore team. Is this something babel can do? There is already this issue which is kinda similar https://github.com/babel/babel/issues/8019 but also this one which emphasis on the fact that `_this` should user the value resturned by `super()` #2279

**Environment**
- Babel version(s): v7.1.5
- Node/npm version: Node 8/npm 6
- OS: build Ubuntu 18.04, this test on Windows 10 / Edge 17
- Monorepo no

**Possible Solution**

Another possible solution would be to not transpile on rest/spread & default argument function when `targets: {esmodules: true}` or `targets: {browsers: 'Edge >= 14'}` because its seems they support it

joeldenning:
| Fixed Issues?            | Fixes #9020 
| Major: Breaking Change?  | 
| Minor: New Feature?      | 

See #9020 and [this comment](https://github.com/babel/babel/issues/9020#issuecomment-440185493) where @nicolo-ribaudo suggested the change that I implemented here. This is my first contribution to babel and I'm not very familiar with the codebase -- I welcome any feedback. I consider this to be a best effort implementation, since I probably am missing some things or don't fully understand the implications of my code changes.


packages/babel-traverse/src/path/conversion.js
```diff
@@ -461,8 +461,12 @@ function getThisBinding(thisEnvFn, inConstructor) {
         if (supers.has(child.node)) return;
         supers.add(child.node);
 
-        child.replaceWith(
-          t.assignmentExpression("=", t.identifier(thisBinding), child.node),
+        child.insertAfter(
```
joeldenning: See [this comment](https://github.com/babel/babel/issues/9020#issuecomment-440185493) for why I made this change

packages/babel-plugin-transform-arrow-functions/test/fixtures/arrow-functions/this/output.js
```diff
@@ -15,9 +15,12 @@ class Foo extends function () {} {
     };
 
     if (true) {
-      console.log(_this2 = super(), foo());
+      var _temp;
+
+      console.log((_temp = super(), _this2 = this, _temp), foo());
```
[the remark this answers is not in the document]
joeldenning: 👍 good catch - @nicolo-ribaudo I just updated it to do what you said. Let me know if I misunderstood.

packages/babel-traverse/src/path/conversion.js
```diff
@@ -461,8 +461,13 @@ function getThisBinding(thisEnvFn, inConstructor) {
         if (supers.has(child.node)) return;
         supers.add(child.node);
 
-        child.replaceWith(
-          t.assignmentExpression("=", t.identifier(thisBinding), child.node),
+        child.replaceWithMultiple(
```
nicolo-ribaudo: It takes an array.

joeldenning: @nicolo-ribaudo oops! I just fixed that. I don't fully understand the expected output in some of the tests, so I didn't catch that the second argument was being ignored.

joeldenning:
Yeah I was surprised that there weren't more tests affected, too. Are you the one who will merge this @nicolo-ribaudo? Or is that someone else?

joeldenning:
I think the reason that there aren't more tests affected is that, if i understand the code correctly, this only affects `super()` calls in classes that have arrow functions in the constructor, such as in the case of compiled class properties that use the `fn1 = () => {}` syntax. I'm not 100% sure on that, but that's what I saw in my own testing of it.

JSteunou:
Is there a risk this fix break #2279 ?

nicolo-ribaudo:
It shouldn't; this PR doesn't touch transpilation of `super()`.

adrienharnay:
Hey, any chance this could be released? Thanks 🙂

PR #9061, opened 2018-11-22
Adding ability to use jest's --watch feature.

joeldenning:
| Fixed Issues?            |  No issue for this
| Patch: Bug Fix?          | N/A (dev change, only)
| Major: Breaking Change?  | N/A (dev change, only)
| Minor: New Feature?      | N/A (dev change, only)

Jest's watch feature is nice because then you can use jest's built-in filename and test name regex searchers and switch between tests without having to wait on it when restarted. It also allows you to run only the failed tests, which is nice.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -140,6 +140,11 @@ $ TEST_DEBUG=true make test
 
 You can combine `TEST_DEBUG` with `TEST_GREP` or `TEST_ONLY` to debug a subset of tests. If you plan to stay long in the debugger (which you'll likely do!), you may increase the test timeout by editing [test/mocha.opts](https://github.com/babel/babel/blob/master/test/mocha.opts).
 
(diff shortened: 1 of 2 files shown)


nicolo-ribaudo:
Thanks for the PR.
Instead of adding many flags to out rudimental jest wrapper, I suggest using
```
yarn jest --any-flag --you-want
```
which is way more powerful :wink:

joeldenning:
Thanks for the clarification. Running `yarn jest` is not what is suggested in the contributing guide. And it seems to depend on yarn's behavior of auto-running scripts that are not defined in the package.json scripts. `make test` is the recommended way in the contributing guide. Should we perhaps add to the contributing guide to clarify that you can also do `yarn jest` instead of `make test`?

nicolo-ribaudo:
Yes please!

nicolo-ribaudo:
This is now mentioned in contributing.md

PR #9073, opened 2018-11-24
Microbouji patch/8136

### LINKED ISSUE #8136: spread-operator doesn't transform to ES5

jridgewell:
| Fixed Issues?            | Fixes #8136, closes #8210
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR         | 
| Any Dependency Changes?  |

Currently when `proposal-optional-chaining` transforms

```
a?.b()
```

into

```
(_a = a) === null || _a === void 0 ? void 0 : _a.b()
```

that last `_a.b()` is technically an `OptionalCallExpression` with no `optional` field set.

This PR changes it to be just a `CallExpression`. I believe this is how it should be, and this also enables other plugins (like `transform-spread`) who use `CallExpression` in their visitors to keep working.

- - -

This supersedes #8210, and fixes deeply nested optional call expressions. I've reused the original commits from that PR.


henrikra:
Are guys using VsCode when your code has optional chain syntax? How well does it work in your editor? Does it mess up editor syntax highlighting?

PR #9078, opened 2018-11-25
Pass `rootMode` from `@babel/node`.

### LINKED ISSUE #8968: babel-node can't be configured with babelRcRoots
## Bug Report

To upgrade our monorepo to Babel 7 I had to add the following to each front-end app's Webpack config:

```
rootMode: 'upward',
babelrcRoots: ['.', './*'],
```

I also need to be able to do something similar for the back-end services which use `babel-node` when running locally on developer machines. I can get half way there by adding a `--config-file ../babel.config.js` switch to the `babel-node` command, but I can't find anyway at present to set `babelrcRoots`, and so I see errors within any JavaScript that sits inside a package boundary.

Anybody know whether this is a genuine omission (as it seems to be) or whether there's some way of doing this? Thanks so much 👍

wtgtybhertgeghgtwtg:
| Fixed Issues?            | Fixes #8968 (although it's already closed).
| Patch: Bug Fix?          | No.
| Major: Breaking Change?  | No.
| Minor: New Feature?      | Yes.
| Tests Added + Pass?      | No.
| Documentation PR Link    | 
| Any Dependency Changes?  | No.

#8660 added support for `rootMode` for `@babel/core` and `@babel/cli`, but it can't be touched from `@babel/node`.  This pretty much copypastes what `@babel/cli` got.

### DIFF
--- a/packages/babel-node/src/_babel-node.js
+++ b/packages/babel-node/src/_babel-node.js
@@ -64,7 +64,7 @@ program.version(pkg.version);
 program.usage("[options] [ -e script | script.js ] [arguments]");
 program.parse(process.argv);
 
-register({
+const babelOptions = {
   caller: {
     name: "@babel/node",
   },
@@ -81,7 +81,15 @@ register({
   // leave them undefined so that @babel/core can handle the
   // default-assignment logic on its own.
   babelrc: program.babelrc === true ? undefined : program.babelrc,
-});
+};
+
+for (const key of Object.keys(babelOptions)) {
+  if (babelOptions[key] === undefined) {
+    delete babelOptions[key];
+  }
+}
+
+register(babelOptions);
 
 const replPlugin = ({ types: t }) => ({
   visitor: {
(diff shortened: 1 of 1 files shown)


packages/babel-node/src/_babel-node.js
```diff
@@ -70,6 +75,7 @@ register({
   presets: program.presets,
   configFile: program.configFile,
   envName: program.envName,
+  rootMode: program.rootMode,
```
loganfsmyth: Would you be up for moving this options object into a variable and adding a loop like https://github.com/babel/babel/blob/1b8d664bbe1976ae294a0c74c3a24b7dbb7e33b5/packages/babel-cli/src/babel/options.js#L257 Otherwise this will run into the same issue the CLI did that the comment https://github.com/babel/babel/blob/1b8d664bbe1976ae294a0c74c3a24b7dbb7e33b5/packages/babel-cli/src/babel/options.js#L253 describes.

wtgtybhertgeghgtwtg: Sure.  Do you want me to move it to an `options.js` as with `@babel/cli`?

loganfsmyth: Nah, just moving it above the `register` call is fine.

PR #9083, opened 2018-11-25
Make @babel/plugin-class-features a normal helper package

nicolo-ribaudo:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is a better alternative to https://github.com/babel/babel/pull/9050.

---

This PR removes the `@babel/plugin-class-features` package, which becames `@babel/helper-create-class-features-plugin` and can only be used inside other plugins.

The problem with the `@babel/plugin-class-features` plugin is that, in order to allow using `@babel/plugin-proposal-class-properties`, etc., it behaves differently when declared multiple times:

```json
{
  "plugins": [
    ["@babel/plugin-class-features", { "fields": true }],
    ["@babel/plugin-class-features", { "decorators": true }]
  ]
}
```

For any other plugin, the last specified options override the others, but in this case they are merged.

In #9050 I removed direct options to that plugin, but it doesn't make sense to allow using it as `plugins: ["@babel/plugin-class-features"]` since it does nothing.


packages/babel-helper-class-features-plugin/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@babel/plugin-class-features",
+  "name": "@babel/helper-class-features-plugin",
```
Andarist: maybe `@babel/helper-create-class-feature-plugin` or `@babel/helper-class-feature-factory`?

nicolo-ribaudo: I like the first name :+1:

existentialism: `@babel/helper-create-class-feature-plugin` 👍

packages/babel-helper-create-class-features-plugin/src/index.js
```diff
@@ -27,61 +25,23 @@ export { enableFeature, FEATURES, setLoose };
...
   return {
-    name: "class-features",
+    name: "class-features/" + name,
 
-    manipulateOptions(opts, parserOpts) {
-      if (fields) {
-        parserOpts.plugins.push("classProperties", "classPrivateProperties");
-      }
-    },
+    manipulateOptions: manipulateOptions,
```
hzoo:
```suggestion
    manipulateOptions,
```

packages/babel-helper-create-class-features-plugin/src/index.js
```diff
@@ -27,61 +25,23 @@ export { enableFeature, FEATURES, setLoose };
...
-
+export function createClassFeaturePlugin({
+  name,
+  feature,
+  loose,
+  manipulateOptions,
+}) {
   return {
-    name: "class-features",
+    name: "class-features/" + name,
```
loganfsmyth: I think I'd rather keep the names consistent with the other plugins. Maybe we can delete this line and have the plugin do:
```
return {
  name: "...",
  createClassFeaturePlugin({ /* ... */ })
};
```
?

PR #9095, opened 2018-11-28
Retain JSX pragma if defined as a comment

### LINKED ISSUE #8958: Import used by @jsx pragma is elided by typescript transform
## Bug Report

**Current Behavior**
When an imported value is used only implicitly via JSX, the typescript type import elision is incorrectly removing the "type-only" import.

**Input Code**

```js
/* @jsx jsx */

import ReactDOM from 'react-dom'

ReactDOM.render(<div/>, document.appendChild(document.createElement('div')))
```

**Expected behavior/code**
The "jsx" import should still be present in the output

**Babel Configuration**

Literally `create-react-app --typescript` defaults.

Output from webpack:

```js
"use strict";
__webpack_require__.r(__webpack_exports__);
/* harmony import */ var react_dom__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(/*! react-dom */ "./node_modules/react-dom/index.js");
/* harmony import */ var react_dom__WEBPACK_IMPORTED_MODULE_0___default = /*#__PURE__*/__webpack_require__.n(react_dom__WEBPACK_IMPORTED_MODULE_0__);
var _jsxFileName = "/Users/kovensky/src/statsim/src/index.tsx";

/* @jsx jsx */

react_dom__WEBPACK_IMPORTED_MODULE_0___default.a.render(jsx("div", {
  __source: {
    fileName: _jsxFileName,
    lineNumber: 6
  },
  __self: undefined
}), document.appendChild(document.createElement('div')));
```

Note how the `@emotion/core` import is missing entirely.

Adding an unused `jsx` reference anywhere to the file makes the build correct.

dstaley:
| Fixed Issues?            | Fixes #8958 
| Documentation PR Link    | 

Currently, the TypeScript transform elides JSX pragma imports unless used explicitly. This PR modifies the transform to properly detect the JSX pragma declaration when defined as a comment, and appropriately retains imports if the import matches the pragma. The check was added by copying over the pragma detection from `babel-plugin-transform-react-jsx`.

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -25,6 +25,8 @@ const PARSED_PARAMS = new WeakSet();
 export default declare((api, { jsxPragma = "React" }) => {
   api.assertVersion(7);
 
+  const JSX_ANNOTATION_REGEX = /\*?\s*@jsx\s+([^\s]+)/;
+
   return {
     name: "transform-typescript",
     inherits: syntaxTypeScript,
@@ -37,6 +39,17 @@ export default declare((api, { jsxPragma = "React" }) => {
 
       Program(path, state: State) {
         state.programPath = path;
+
+        const { file } = state;
+
+        if (file.ast.comments) {
+          for (const comment of (file.ast.comments: Array<Object>)) {
+            const jsxMatches = JSX_ANNOTATION_REGEX.exec(comment.value);
+            if (jsxMatches) {
+              jsxPragma = jsxMatches[1];
+            }
+          }
+        }
 
         // remove type imports
         for (const stmt of path.get("body")) {
--- a/packages/babel-plugin-transform-typescript/test/fixtures/imports/elide-jsx-pragma-no/input.mjs
+++ b/packages/babel-plugin-transform-typescript/test/fixtures/imports/elide-jsx-pragma-no/input.mjs
@@ -1,0 +1,4 @@
+/* @jsx htm */
+// Don't elide htm if a JSX element appears somewhere.
+import { htm } from "fake-jsx-package";
+<div></div>;
--- a/packages/babel-plugin-transform-typescript/test/fixtures/imports/elide-jsx-pragma-no/options.json
+++ b/packages/babel-plugin-transform-typescript/test/fixtures/imports/elide-jsx-pragma-no/options.json
@@ -1,0 +1,4 @@
+{
+    "plugins": [["transform-typescript", { "isTSX": true }]]
+}
+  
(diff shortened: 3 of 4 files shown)


nicolo-ribaudo:
Isn't it enough to use the `jsxPragma` option?

chibicode:
@nicolo-ribaudo 

@jfrolich says: https://github.com/babel/babel/issues/8958#issuecomment-441954247

> This can be fixed by setting the jsxPragma option in the typescript transform. However now React is removed when Fragments are used.

nicolo-ribaudo:
With this PR parsing the comment has exactly the same behavior of using the jsxPragma option.

dstaley:
@nicolo-ribaudo The goal here is to allow the use of different pragmas for different files, which mirrors the behavior of Babel in non-TypeScript projects.

kevinmarrec:
@dstaley Is there a way that you update within this PR the `@babel/preset-typescript` to use the last version of `@babel/core` and `@babel/plugin-transform-typescript` ?

The TypeScript preset seems not being maintained (last publish 3 months ago)

kevinmarrec:
(or maybe should I make a side PR, to keep things clear)

dstaley:
@kevinmarrec yeah it's probably best to keep those separate since there might be a reason the presentation hasn't been updated.

meebix:
Any update on getting this merge in? @dstaley - Some of us Emotion users are pretty stuck right now.

aaronjensen:
@meebix I'm not stuck anymore. I'm using https://emotion.sh/docs/@emotion/babel-preset-css-prop and it works just fine. One thing to note is that I had to make sure that `@babel/preset-env` was the first preset in my list. 

I combined it w/ a custom plugin to handle fragments:

```js
// With emotion 10 and @emotion/babel-preset-css-prop, the only reason to import
// React is to get Fragment syntax. This plugin will automatically add a React
// import if a fragment is found so we don't have to remember to import React.
module.exports = function importReactForFragments(babel) {
  const { types: t } = babel

  const getFragmentImport = () =>
    t.importDeclaration(
      [t.importSpecifier(t.identifier('React'), t.identifier('default'))],
      t.stringLiteral('react')
    )

  const detectReactImport = (path, state) => {
    if (path.node.local.name === 'React') {
      state.set('jsxFragmentImportDetected', true)
    }
  }

  return {
    visitor: {
      Program: {
        exit(path, state) {
          if (!state.get('jsxFragmentDetected')) return
          if (state.get('jsxFragmentImportDetected')) return
          path.unshiftContainer('body', getFragmentImport())
        },
      },
      ImportNamespaceSpecifier: detectReactImport,
      ImportDefaultSpecifier: detectReactImport,
      ImportSpecifier: detectReactImport,
      JSXFragment(path, state) {
        state.set('jsxFragmentDetected', true)
      },
    },
  }
}
```

PR #9096, opened 2018-11-28
Add node 11 to CI and remove node 9

danez:
| Fixed Issues?            | Fixes #8871
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Enable node 11 on circleCI.
Remove node 9 on travis and instead use node 10.
Do not cache node_modules as extracting the cache takes significant amount of time, but not sure about that, lets see what happens to build times.
Do not run `yarn install` twice when running tests.

This also incorporates some changes from #8871.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,7 @@ jobs:
   build:
     working_directory: ~/babel
     docker:
-      - image: circleci/node:10
+      - image: circleci/node:11
     steps:
       - checkout
       - restore-cache: *restore-yarn-cache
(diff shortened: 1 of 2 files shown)


danez:
It seems the test builds are faster now between 1 to 3 minutes each. I do not have a good explanation for this even after looking at the detailed logs. The only thing I can think of is that caching node_modules in every build was also caching `node_modules/.cache` which contains babel-loader, webpack, jest and potentially more caches which grow over time and it takes them longer to load. But 🤷‍♀️

xtuc:
Also we could port this https://github.com/webpack/webpack/commit/edcadbb0a107cf73010395b3fdd8ca65e66a11ad to test against any nightly.

PR #9101, opened 2018-11-29
Private Class Methods Stage 3: Private Accessors

tim-mc:
| Q                       | A                                            |
| :---------------------: | :------------------------------------------: |
| Fixed Issues?           | https://github.com/babel/proposals/issues/22 |
| Patch: Bug Fix?         |                                              |
| Major: Breaking Change? |                                              |
| Minor: New Feature?     | Add private accessors support                |
| Tests Added + Pass?     | Yes                                          |
| Documentation PR        |                                              |
| Any Dependency Changes? | No                                           |
| License                 | MIT                                          |
| Sponsor                 | @bloomberg                                   |

This is a follow-up to #8654. It adds support for private accessors within the current `babel-plugin-class-features` + `babel-plugin-proposal-private-methods` plugin combination.

cc: @robpalme @littledan @syg @jridgewell @nicolo-ribaudo


tim-mc:
I'd like to get some feedback on how this should be handling the usage of update expressions and certain assignment expressions. I'm not sure what exactly the expectation is for how to handle these situations:

```js
// update expressions
this.#privateSetter++;
this.#privateSetter--;
++this.#privateSetter;
// etc.

// assignment expressions
this.#privateSetter += 'foo';
this.#privateSetter -= 'foo';
```

I made a [method that checks for these scenarios](https://github.com/babel/babel/blob/d9593cba07a2b9a14eafa9ec971f57ee479a8f04/packages/babel-helper-create-class-features-plugin/src/fields.js#L59-L75) which is used in both the [spec](https://github.com/babel/babel/blob/d9593cba07a2b9a14eafa9ec971f57ee479a8f04/packages/babel-helper-create-class-features-plugin/src/fields.js#L190) and [loose](https://github.com/babel/babel/blob/d9593cba07a2b9a14eafa9ec971f57ee479a8f04/packages/babel-helper-create-class-features-plugin/src/fields.js#L234) private name handlers.

My main question is: **should these usages be throwing errors?** If so,
  + Should they code frame errors?
  + Is this the proper place to throw an error? I considered putting it over [here in `features.js`](https://github.com/babel/babel/blob/d9593cba07a2b9a14eafa9ec971f57ee479a8f04/packages/babel-helper-create-class-features-plugin/src/features.js), but since this isn't really a check for a "feature," that seemed wrong.

**If they should not be throwing**, what would be the proper way to handle these?

packages/babel-plugin-proposal-private-methods/test/fixtures/private-method-loose/accessors/output.js
```diff
@@ -0,0 +1,46 @@
...
+    Object.defineProperty(this, _privateField, {
+      writable: true,
+      value: "top secret string"
+    });
+    Object.defineProperty(this, _get_privateGetter, {
+      value: _get_privateGetter2
+    });
+    Object.defineProperty(this, _set_privateSetter, {
+      value: _set_privateSetter2
+    });
```
nicolo-ribaudo: This should be
```js
Object.defineProperty(this, _privateGetter, {
  get: _get_privateGetter2
});
```
so that we can rely on the native getter/setter implementation. In case of a getter/setter pair, it should just be
```js
Object.defineProperty(this, _foo, {
  get: _get_foo,
  set: _set_foo,
});
```

nicolo-ribaudo:
Can't `++this.#privateSetter;` be desugared to `this.#privateSetter = this.#privateSetter + 1`, which should set it to `NaN`?

packages/babel-plugin-proposal-private-methods/test/fixtures/private-method-loose/accessors/output.js
tim-mc: Ah, good point, will change.

tim-mc:
@nicolo-ribaudo I've made the updates to the PR that we discussed.

packages/babel-helper-create-class-features-plugin/src/index.js
```diff
@@ -75,10 +75,31 @@ export function createClassFeaturePlugin({
 
           if (path.isPrivate()) {
             const { name } = path.node.key.id;
+            const getName = `get_${name}`;
```
nicolo-ribaudo: Could you use a space instead of _? Otherwise this code throws:
```js
class A {
  get #priv() {}
  #get_priv() {}
}
```

nicolo-ribaudo: Also, it should throw for this code:
```js
class A {
  get #priv() {}
  #priv() {}
}
```

tim-mc: Good point, will update.

tim-mc: It appears that `toIdentifier` [replaces whitespaces in names with dashes](https://github.com/babel/babel/blob/3f9a1c08cc70ac5bfc21bff13ef51eea1bd06f27/packages/babel-types/src/converters/toIdentifier.js#L8) and then [strips the dash out altogether](https://github.com/babel/babel/blob/3f9a1c08cc70ac5bfc21bff13ef51eea1bd06f27/packages/babel-types/src/converters/toIdentifier.js#L14), so the end result is `getMethodName` and `setMethodName`...

nicolo-ribaudo: Well, `setName` and `getName` are only used in the `privateNames` set and never passed to `toIdentifier`, right?

Anyway, to fix both the problems I think you could use the `privateNames` set alongside with a `partialAccessors` map. `partialAccessor` would store the private name as a key, and the value would either be `"get"` or `"set"`.

The check to disallow private names would then look like this:
```js
if (privateNames.has(name)) {
  if ((kind === "get" && partialAccessors.get(name) === "set") || (kind === "set" && partialAccessors.get(name) === "get")) {
    partialAccessors.delete(name);
  } else {
    throw;
  }
} else {
  if (isAccessor) partialAccessors.set(name, kind);
  privateNames.add(name);
}
```

If you have a better idea feel free to ignore me!

tim-mc: Alternatively, I could do: 

```js
if (prop.node.kind === "get") {
  update.getId = t.identifier(`get ${prop.scope.generateUid(`${name}`)}`);
} else if (prop.node.kind === "set") {
  update.setId = t.identifier(`set ${prop.scope.generateUid(`${name}`)}`);
}
```

...but that throws since:

```js
/* ... */
get: get _privateFieldValue2,
/* ... */
```

is a `SyntaxError` (unexpected identifier).

nicolo-ribaudo: The problem is only in the `privateNames` map. The `_get_privateField` identifiers are ok since they are already unique.

tim-mc: @nicolo-ribaudo Ok, this should be fixed now!

packages/babel-helpers/src/helpers.js
```diff
@@ -1058,14 +1062,19 @@ helpers.classPrivateFieldSet = helper("7.0.0-beta.0")`
...
+    } else {
+      if (!descriptor.writable) {
+        // This should only throw in strict mode, but class bodies are
+        // always strict and private fields can only be used inside
+        // class bodies.
+        throw new TypeError("attempted to set read only private field");
+      }
+
+      descriptor.value = value;
+      return value;
```
nicolo-ribaudo: `return value;` should be outside the if/else, because we always need to return it.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -6,16 +6,28 @@ import optimiseCall from "@babel/helper-optimise-call-expression";
...
+      const update = privateNamesMap.has(name)
+        ? privateNamesMap.get(name)
+        : {
+            id: prop.scope.generateUidIdentifier(name),
+            static: !isInstance,
+            method: isMethod,
+            methodId:
+              isMethod && isInstance && prop.node.kind === "method"
+                ? prop.scope.generateUidIdentifier(name)
+                : undefined,
```
nicolo-ribaudo: Nit: could you move the `methodId` definition inside an if condition, like you did for `getId` and `setId`?

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -404,7 +535,7 @@ export function buildFieldsInitNodes(
 
   return {
     staticNodes,
-    instanceNodes,
+    instanceNodes: instanceNodes.filter(instanceNode => instanceNode),
```
nicolo-ribaudo: Nit: we usually use `.filter(Boolean)`

packages/babel-helper-create-class-features-plugin/src/index.js
nicolo-ribaudo: It is ok

nicolo-ribaudo:
This PR looks almost ready!

tim-mc:
@nicolo-ribaudo requested changes have been made.

nicolo-ribaudo:
Can you add a test for the helper change? Something like

```js
let foo;
class Cl {
  set #foo(v) { return 1 }
  test() {
    foo = this.#foo = 2;
  }
}

new Cl().test();

expect(foo).toBe(2);
```

Also, it would be good to have the following tests:
- `duplicated-names/get-set` // does not throw
- `duplicated-names/set-get` // does not throw
- `duplicated-names/get-get` // throws at compile time
- `duplicated-names/set-set` // throws at compile time
- `duplicated-names/get-method` // throws at compile time
- `duplicated-names/method-get` // throws at compile time
- `duplicated-names/set-method` // throws at compile time
- `duplicated-names/method-set` // throws at compile time
- `accessors/set-only-getter` // throws at runtime
- `accessors/get-only-setter` // returns undefined
- `accessors-loose/set-only-getter` // throws at runtime
- `accessors-loose/get-only-setter` // returns undefined
- Move the `testUpdates` functions to `accessors/updates` and `accessors-loose/updates`
- Move the remaining code from `private-methods/accessors` and `private-methods-loose/accessors` to `accessors/basic` and `accessors-loose/basic`

tim-mc:
@nicolo-ribaudo Ok, updated with new & modified tests.

packages/babel-plugin-proposal-private-methods/test/fixtures/accessors-loose/options.json
```diff
@@ -0,0 +1,15 @@
+{
+  "plugins": [
+    [
+      "external-helpers",
+      {
+        "helperVersion": "7.1.6"
```
nicolo-ribaudo: Could you use something like `7.1000.0`? So that we don't have to change this file if the helpers change in a futue version.

tim-mc: Sure, done.

packages/babel-plugin-proposal-private-methods/test/fixtures/accessors-loose/options.json
```diff
@@ -0,0 +1,15 @@
...
+      "external-helpers",
+      {
+        "helperVersion": "7.1000.0"
+      }
+    ],
+    ["proposal-private-methods", { "loose": true }],
+    ["proposal-class-properties", { "loose": true }],
+    "transform-classes",
+    "transform-block-scoping",
+    "syntax-class-properties"
```
hzoo: Not specific to this test/PR but was wondering if we actually need to include these plugins?

`"transform-block-scoping", "syntax-class-properties"`

PR #8448, opened 2018-08-10
Remove Babylon plugins for features already merged to the ECMAScript spec

nicolo-ribaudo:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | :+1:
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

These are now enabled by default:
- objectRestSpread (2018)
- asyncGenerators (2018)
- optionalCatchBInding (2019)
- jsonStrings (2019)

TODO (after this PR):
- [ ] Deprecate the `@babel/plugin-syntax-*` packages.
- [ ] Deprecate the `@babel/plugin-proposal-*` packages.
- [ ] Create the `@babel/plugin-transform-*` packages.

PR #9106, opened 2018-11-30
Rename ES2018/ES2019 plugins (proposal -> transform)

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is quite annoying to do, hopefully it needs to be done only once per year :laughing: 	Follows up to https://github.com/babel/babel/pull/8448.

- The first commit is just a find-replace of strings, only subsequent commits need to be reviewed.
- ~Is 0116876 breaking? I don't think so, but I can revert it.~ Yes, it is for preset-env. I removed that commit.
- This is probably the first time I modified preset-env. Do I need to regenerate other files?
- Should I add the ES2018 and ES2019 presets to Babel standalone, so that they can be used in the repl?
- [ ] TODO: Add the removed plugins to https://github.com/babel/babel-archive/tree/master/packages
- [ ] TODO: Deprecate the `-(syntax|proposal)-` packages on npm?


packages/babel-preset-env-standalone/src/available-plugins.js
```diff
@@ -2,7 +2,7 @@ import { availablePlugins, registerPlugin } from "@babel/standalone";
 
 const notIncludedPlugins = {
   "transform-new-target": require("@babel/plugin-transform-new-target"),
-  "proposal-json-strings": require("@babel/plugin-proposal-json-strings"),
+  "transform-json-strings": require("@babel/plugin-transform-json-strings"),
```
nicolo-ribaudo: @existentialism What is this object for?

existentialism: It's a mapping of packages _not_ included in `babel-standalone`.

hzoo:
Thanks for adding the PR although.. so I feel like it's not worth changing the name and going through all this, as once it's standardized you shouldn't need to use the individual plugins alone and rely on preset-env anyway?

nicolo-ribaudo:
I mainly did it for consistency, O didn't have a strong reason.

loganfsmyth:
Hmm, idk, I support the `proposal-` prefix as an indication to users that something was not standardized proposal. If we're not going to rename things, we should go back to `transform-` for everything.

PR #6763, opened 2017-11-07
No unneeded empty arrays in transform spread

apapirovski:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            |  Fixes #6017 
| Patch: Bug Fix?          | Yes?
| Major: Breaking Change?  | No
| Minor: New Feature?      | No
| Tests Added + Pass?      | Yes
| Documentation PR         | No
| Any Dependency Changes?  | No
| License                  | MIT

Since `Array.prototype.concat` creates a new array from inputs, there's no need to call it from a new empty array (`[].concat()`).

This also allowed simplification of cases where a single `arguments` or `babelHelpers.toConsumableArray` is used with a spread operator (no need to call `concat` at all since both already return a fresh array).

First time diving into Babel internals so wasn't sure if my ways of testing in `isConsumableArrayHelper` and `isArrayPrototypeSliceCall` were necessarily optimal.

PR #9108, opened 2018-12-01
Correctly transform spreads to use proper concat method

### LINKED ISSUE #8907: @babel/plugin-transform-spread regression of spread array
# v7 Regression

> First check out: https://new.babeljs.io/docs/en/next/v7-migration.html
> Also a partial upgrade tool: https://github.com/babel/babel-upgrade

If you have time to investigate, what PR/date introduced this issue.

https://github.com/babel/babel/pull/6763

before that PR `[...array]` was transpiled to `[].concat(array)` now it's  `array.concat()`, this change breaks immutable array(seamless-immutable), for example

```
import Immutable from `seamless-immutable`
const array = Immutable([1, 2, 3]);
console.log(Immutable.isImmutable([...array]);
```
it prints `false` in console of latest Chrome and in Node v8.11.3, but it prints true with Babel@7, it should be `false`, but as it's transpiled to `array.concat()`, and an immutable array returns immutable array too with builtin methods(`concat`, `map`, `reduce`...)

**Describe the regression**
A clear and concise description of what the regression is.

**Input Code**

```js
var your => (code) => here;
```

```js
{
  "your": { "config": "here" }
}
```

**Expected behavior/code**

**Environment**

danez:
| Fixed Issues?            | Fixes #8907 
| Documentation PR Link    | 
| Any Dependency Changes?  |

This reverts some of the optimizations of #6763.

Transforming a spread to this

```js
// [...arr, b]
arr.concat(b)
```
is dangerous as we do not know what `arr` is exactly. It might be a subclass of `Array` with changed behavior of its concat function.

Instead I changed it to use

```js
Array.prototype.concat(arr, b);
```

Other option would be to use what was there before #6763

```js
[].concat(arr, b);
```

which is smaller, but needs to create an additional array.

Also added comments to hopefully make the logic understandable.

### DIFF
--- a/packages/babel-plugin-transform-spread/test/fixtures/regression/issue-8907/exec.js
+++ b/packages/babel-plugin-transform-spread/test/fixtures/regression/issue-8907/exec.js
@@ -3,7 +3,9 @@ const arr = [];
 arr.concat = () => {
     throw new Error('Should not be called');
 };
-
+let x;
 expect(() => {
-    const x = [...arr];
+    x = [...arr];
 }).not.toThrow();
+
+expect(x).not.toBe(arr);
--- a/packages/babel-plugin-transform-spread/test/fixtures/spread/arguments-array/exec.js
+++ b/packages/babel-plugin-transform-spread/test/fixtures/spread/arguments-array/exec.js
@@ -1,0 +1,9 @@
+// test that toConsumableArray clones the array.
+
+function foo() {
+  const x = [...arguments];
+
+  expect(x).not.toBe(arguments);
+}
+
+foo(1,2);
--- a/packages/babel-plugin-transform-spread/test/fixtures/spread/single/exec.js
+++ b/packages/babel-plugin-transform-spread/test/fixtures/spread/single/exec.js
@@ -1,0 +1,7 @@
+// test that toConsumableArray clones the array.
+const arr = [];
+const foo = () => arr;
+
+const x = [...foo()];
+
+expect(x).not.toBe(arr);


packages/babel-plugin-transform-destructuring/test/fixtures/destructuring/array-unpack-optimisation/output.js
```diff
@@ -21,7 +21,7 @@ var _ref4 = [a[1], a[0]];
 a[0] = _ref4[0];
 a[1] = _ref4[1];
 
-var _ref5 = babelHelpers.toConsumableArray(foo).concat([bar]),
+var _ref5 = Array.prototype.concat(babelHelpers.toConsumableArray(foo), [bar]),
```
loganfsmyth: I'm not sure I love the idea of relying on the implicit `Array.prototype` as the `this` context for the `concat` call.

danez: Should i change it to `[].concat(..)` ?

Or i guess we could do 
```
var _a = babelHelpers.toConsumableArray(foo);
Array.prototype.concat.call(_a, _a, ..);
```

danez: Okay it is now transforming to `[].concat()` again.

smashercosmo:
One bad thing about this PR is that before it if somebody tries to spread Set in loose mode, clear error is thrown, but now he will get wrong output of ```[Set()]```, which will be harder to spot.

rubennorte:
I agree with @smashercosmo. This also allows to spread non-iterable values again, which is wrong (e.g.: `[...null]`). I know this is for loose mode but IMO having a subset of the semantics (only accepting arrays) is better than changing the semantics.

fdanielsen:
Sorry if this is the wrong place to comment, but this has references to a lot of related history.

In a modern browser that supports the spread operator natively, `[...el.querySelectorAll('.foo')]` correctly converts the iterable `NodeList` to an array of elements. But `[].concat(el.querySelectorAll('.foo')` gives an array of a single item which is the `NodeList`.

This doesn't seem like the intended result with Babels support of the spread operator?

calebeby:
@fdanielsen I'm seeing the same thing as you, with `DOMTokenList` as well (from "spreading" `element.classList`

jridgewell:
The loose transform only handles Arrays. You'll have to use the spec transform (as well as polyfilling `Symbol.iterator` methods for the iterable) if you want this to be done correctly.

calebeby:
I don't see a `spec` option here: https://babeljs.io/docs/en/babel-plugin-transform-spread

Are the docs out of date? Happy to send a PR if that is the case 😃

jridgewell:
Spec in this case is the default. We only add an explicit `spec` option if the doing it by default would cause a large performance regression or something else undesirable.

calebeby:
Got it. Thanks for the clarification 👍

PR #9110, opened 2018-12-01
Added type-level mapping between aliases and nodes that have that alias.

### LINKED ISSUE #9092: [@babel/types]: Export type-level mapping between aliases and the types of nodes that have that alias.
## Feature Request

[Currently, the TypeScript definitions for `Visitor` do not allow aliases as properties](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/46aea293fad10e638e6885b9f1f0d2938a711e16/types/babel__traverse/index.d.ts#L142). There is not enough information available in the auto-generated type definitions in `@babel/types` to support this.

Add an interface that acts as a type-level mapping between aliases and the types of nodes that have those aliases, e.g.

```typescript
export interface Aliases {
  Function: Function;
  Expression: Expression;
  LVal: LVal;
  // ...etc
}
```

An alternative would be to include all node types, not just aliases, in this interface. However, this can be easily created externally to this library by intersecting with the following type:

```typescript
{ [P in Node["type"]]: Extract<Node, { type: P; }>; }
```

If you can, explain how users will be able to use this and possibly write out a version the docs.
Maybe a screenshot or design?

I believe this section is not too relevant in this instance - please correct me if I'm wrong.

cameron-martin:
| Fixed Issues?            | Fixes #9092 
| Patch: Bug Fix?          | 
| Major: Breaking Change?  | 
| Documentation PR Link    | 
| Any Dependency Changes?  | 

## Problem

The [TypeScript types for `Visitor` in `@babel/traverse`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/46aea293fad10e638e6885b9f1f0d2938a711e16/types/babel__traverse/index.d.ts#L142) do not allow node aliases as properties. These cannot be added with the current contents of the auto-generated TypeScript definitions for `@babel/types`.

## Solution

If `@babel/types` had a type-level mapping between aliases and the node types that have those aliases, e.g:

```typescript
export interface Aliases {
  Function: Function;
  Expression: Expression;
  LVal: LVal;
  // ...etc
}
```

Then the `Visitor` type in the TypeScript definitions can be intersected with the following type:

```typescript
{ [K in keyof Aliases]?: VisitNode<S, Aliases[K]> }
```

This would then allow using aliases as `Visitor` properties when using TypeScript.

This PR adds such an `Aliases` type.

## Concerns

* There are no tests for this, because it seems like the infrastructure for testing `d.ts` files is not set up. If tests are required then I can set this infrastructure up.
* This does not add equivalent functionality to the flow types. I believe that if the same thing was added, then it could be used in the same way as the TypeScript types by using `$ObjMap`. Can somebody confirm what the situation with flow definitions for `@babel/traverse` are and whether this would be a useful addition to the flow types?

### DIFF
--- a/scripts/generators/typescript.js
+++ b/scripts/generators/typescript.js
@@ -160,6 +160,13 @@ for (const type in t.FLIPPED_ALIAS_KEYS) {
     .map(type => `${type}`)
     .join(" | ")};\n`;
 }
+code += "\n";
+
+code += "export interface Aliases {\n";
+for (const type in t.FLIPPED_ALIAS_KEYS) {
+  code += `  ${type}: ${type};\n`;
+}
+code += "}\n\n";
 
 code += lines.join("\n") + "\n";
 


cameron-martin:
What's the process for merging PRs? Specifically, how come this has been approved but not yet merged?

PR #9114, opened 2018-12-02
Parse non-octals with leading zeros in non strict mode correctly

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

Floating point numbers which start with a 0 like `09.5` were treated incorrectly and errored even though it should work in non-strict mode.


packages/babel-parser/src/tokenizer/index.js
```diff
@@ -971,14 +971,23 @@ export default class Tokenizer extends LocationParser {
...
     if (!startsWithDot && this.readInt(10) === null) {
       this.raise(start, "Invalid number");
     }
-    if (octal && this.state.pos == start + 1) octal = false; // number === 0
+    let octal =
+      this.state.pos - start >= 2 &&
+      this.input.charCodeAt(start) === charCodes.digit0;
+    if (octal) {
+      if (this.state.strict) {
+        this.raise(start, "Invalid number");
```
nicolo-ribaudo: `Legacy octal literals are not supported in strict mode`?

existentialism: Yeah, this is better.

Is "Legacy" needed?

> Octal literals are not allowed in strict mode.

danez: Makes sense. Changed it.

PR #9116, opened 2018-12-02
Fix bug while running tests when a top directory contains escaped RegExp characters

kdridi:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

When one of the parent directories of your babel working directory contains a special RexExp char (ex. a dot) some tests fail. For example with those tests `TEST_ONLY=babel-plugin-transform-runtime make test`

The problem comes from the missing usage of `new RexExp()` after `escapeRegExp` the working path.

I've also checked if this could happen in other places in the code but I only found and correct this minor bug. #


nicolo-ribaudo:
This seems to have been fixed, thank you anyway! :)

PR #9122, opened 2018-12-04
Types for pipeline operator

thiagoarrais:
| Documentation PR Link    | Not sure if applies

This is the first of two PRs for implementing the smart pipeline transform. It concentrates on the types and generator code associated with the proposal. As soon as this gets merged, I plan on following up with a PR for the transform itself.

It adds three new types that will be needed for writing the visitor for the transform. I've tried to follow existing conventions. But this is my first PR on this area of the code. So please let me know if anything is out of place. I particularly think this needs an accompanying documentation PR, but I'm not really sure where to make the changes.

### DIFF
--- a/packages/babel-types/src/definitions/experimental.js
+++ b/packages/babel-types/src/definitions/experimental.js
@@ -89,6 +89,30 @@ defineType("OptionalMemberExpression", {
   },
 });
 
+defineType("PipelineTopicExpression", {
+  builder: ["expression"],
+  visitor: ["expression"],
+  fields: {
+    expression: {
+      validate: assertNodeType("Expression"),
+    },
+  },
+});
+
+defineType("PipelineBareFunction", {
+  builder: ["callee"],
+  visitor: ["callee"],
+  fields: {
+    callee: {
+      validate: assertNodeType("Expression"),
+    },
+  },
+});
+
+defineType("PipelinePrimaryTopicReference", {
+  aliases: ["Expression"],
+});
+
 defineType("OptionalCallExpression", {
   visitor: ["callee", "arguments", "typeParameters", "typeArguments"],
   builder: ["callee", "arguments", "optional"],
--- a/packages/babel-generator/src/generators/types.js
+++ b/packages/babel-generator/src/generators/types.js
@@ -151,3 +151,15 @@ export function BigIntLiteral(node: Object) {
   }
   this.token(node.value);
 }
+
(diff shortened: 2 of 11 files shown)


thiagoarrais:
@danez @nicolo-ribaudo @xtuc: can you lend an ear here? This is a follow up for #8289.

packages/babel-types/src/definitions/experimental.js
```diff
@@ -89,6 +89,32 @@ defineType("OptionalMemberExpression", {
...
+    expression: {
+      validate: assertNodeType("Expression"),
+    },
+  },
+});
+
+defineType("PipelineBareFunction", {
+  builder: ["callee"],
+  visitor: ["callee"],
+  aliases: ["Expression"],
```
nicolo-ribaudo: I am not sure that we want this to be considered an `Expression`, since it is very context-dependent.

If `path.isExpression()` is true, I would expect to be able to replace it with something without changing its behavior, for example `foo.bar` -> `_x = foo.bar`. Since this isn't true in this case, I don't think that we should alias it but I'm open to discussion.

thiagoarrais: Noted. I'll try to change this.

packages/babel-types/src/definitions/experimental.js
```diff
@@ -89,6 +89,28 @@ defineType("OptionalMemberExpression", {
...
+  builder: ["callee"],
+  visitor: ["callee"],
+  fields: {
+    callee: {
+      validate: assertNodeType("Expression"),
+    },
+  },
+});
+
+defineType("PipelinePrimaryTopicReference", {});
```
nicolo-ribaudo: I think that this should still be an expression, since it can safely* moved around.

* It's safe only if it is in the correct context, like it would be for any other identifier reference.

thiagoarrais:
I think my last change finishes this PR. I am marking this as WIP just so that I have chance to clean up the commit history, though. Please let me know if you need anything else.

thiagoarrais:
Yup. I just want to be nitpicky and remove my temp commits from history. :-)

thiagoarrais:
This is ready to be merged now.

nicolo-ribaudo:
Yes, we usually squash-merge (unless multiple authors worked on the same PR).

> This is ready to be merged now.

Lets just wait for another :heavy_check_mark:

PR #9124, opened 2018-12-04
Deleting compiled files when source files are deleted, renamed or moved, when using --watch

### LINKED ISSUE #5636: Need --delete-dir-on-start or --delete-file-when-src-file-deleted  option for babel --watch flag

```js
{
  "presets": [
    "flow",
    ["env", {
      "targets": {
        "node": true
      }
    }]
  ],
  "retainLines": true
}
```

### Watch Script

```js
"watchToBuild": "babel app_flow --watch --out-dir app"
```

### Expected Behavior
when i delete some files or directories in app_flow directory, i hope these related files or directories would be deleted at same time. But actually these related files and directories still remain in the output directory app!

### Your Environment

| ---------------- | -------
| Babel            |  6.24.1
| node             | 5.12.0
| npm              | 3.8.6
| Operating System | MacOS

lawtontom:
| Fixed Issues?            | Fixes #5636
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 

When using the --watch flag on the Babel CLI, it currently only watches for files being added or changed.  We have run into issues when files have that have been deleted still have their compiled versions available.  This PR ensures that when a file is unlinked it's compiled version (and the .map if available) is also deleted.

For more information on why this is an issue for us (and potentially others), it's worth adding some context:

We're creating a quite a lot of packages within a monorepo (using Lerna and Yarn workspaces).  These packages are consumed by sample clients, which are also in the monorepo.  To do work we first run a scoped lerna exec command that runs Babel watch on all of the packages, creating the code that's then consumed in the sample clients.  We then work on functionality, previewing using one of the sample clients.  Most of the work is done in the packages, and during any refactoring it's easy to get in the state where the client code is still consuming the 'deleted' compiled code, where a file has been renamed or moved.  This issue only reveals itself when performing a fresh compilation.  At the moment we're restarting the babel --watch with the --delete-dir-on-start flag, so this cleans things up, but it's something each developer has to remember to do, and can be complicated when pulling in others code.  I've been working with this fork and it's working nicely for our use case.

In the issue #5636 (https://github.com/babel/babel/issues/5636#issuecomment-294332495) it was mentioned that if this functionality was implemented it'd be the default behaviour so I have made this the case in this fork.  I know the issue was closed but this piece of the functionality is missing and would be very useful for us, and I'm assuming others (mentioned in https://github.com/runvnc/babel-changed/issues/3).  

I'm hoping this helps, and happy for any feedback.  Great work on Babel (and Lerna), we're really making great use of it!

### DIFF
--- a/packages/babel-cli/src/babel/util.js
+++ b/packages/babel-cli/src/babel/util.js
@@ -105,7 +105,6 @@ export function deleteFile(path) {
   if (fs.existsSync(path)) {
     fs.unlink(path, err => {
       if (err) throw err;
-      console.log(path + " was deleted");
     });
   }
 }
--- a/packages/babel-cli/src/babel/util.js
+++ b/packages/babel-cli/src/babel/util.js
@@ -101,6 +101,17 @@ export function deleteDir(path) {
   }
 }
 
+export function deleteFile(path) {
+  if (fs.existsSync(path)) {
+    //fs.unlinkSync(path);
+
+    fs.unlink(path, err => {
+      if (err) throw err;
+      console.log(path + " was deleted");
+    });
+  }
+}
+
 process.on("uncaughtException", function(err) {
   console.error(err);
   process.exit(1);
--- a/packages/babel-cli/src/babel/util.js
+++ b/packages/babel-cli/src/babel/util.js
@@ -103,8 +103,6 @@ export function deleteDir(path) {
 
 export function deleteFile(path) {
   if (fs.existsSync(path)) {
-    //fs.unlinkSync(path);
-
     fs.unlink(path, err => {
       if (err) throw err;
       console.log(path + " was deleted");
--- a/packages/babel-cli/src/babel/dir.js
+++ b/packages/babel-cli/src/babel/dir.js
@@ -84,6 +84,20 @@ export default async function({ cliOptions, babelOptions }) {
     return written;
   }
 
+  async function handleFileDelete(src, base) {
+    const relative = path.relative(base, src);
+    const dest = getDest(relative, base);
+    util.deleteFile(dest);
+    // Delete the source map if they're enabled
+    if (
+      util.isCompilableExtension(relative, cliOptions.extensions) &&
+      babelOptions.sourceMaps &&
+      babelOptions.sourceMaps !== "inline"
+    ) {
+      util.deleteFile(dest + ".map");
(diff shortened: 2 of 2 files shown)


loganfsmyth:
I wrote the original comment and I'm not against this, but do keep in mind that your workflow will still easily have holes in it, because you could still remove a file when the watcher wasn't running, and it would never be removed from your output directories. You should still have build logic to delete the output folder content entirely before you start the watcher.

lawtontom:
I agree, I've been running it with the --delete-dir-on-start flag.  It's been working well, with only a few empty folders remaining after a moving files or renaming folders, which as you say are cleared up by the --delete-dir-on-start whenever we restart.  If you didn't want it always on, I guess we could turn it on when --delete-dir-on-start is selected?

packages/babel-cli/src/babel/util.js
```diff
@@ -101,6 +101,14 @@ export function deleteDir(path) {
   }
 }
 
+export function deleteFile(path) {
+  if (fs.existsSync(path)) {
+    fs.unlink(path, err => {
+      if (err) throw err;
```
nicolo-ribaudo: We should return a rejected promise, not throw an error which can't be handled.
Then, in `handleFileDelete`, we can do something like this:
```js
util.deleteFile(dest).catch(() => console.error("Unable to delete", dest));
```

rladigital:
@nicolo-ribaudo Sorry about the delay, I didn't get any notification about your comment, I've now got them turned on.  I've updated the PR based around your feedback, I hope this is what you were after.  

I think that in the time between this PR being opened and now there's been another PR that looks like it's trying to achieve the same goal (https://github.com/babel/babel/pull/9303).

packages/babel-cli/src/babel/util.js
lawtontom: @nicolo-ribaudo Sorry about the delay, I didn't get any notification about your comment, I've now got them turned on. I've updated the PR based around your feedback, I hope this is what you were after.

I think that in the time between this PR being opened and now there's been another PR that looks like it's trying to achieve the same goal (https://github.com/babel/babel/pull/9303), so it looks like this is a feature that others are interested in. I know that in our situation it'd help eliminate some annoying issues.

PR #9141, opened 2018-12-07
Add "ignoreUninitialized" option to "@babel/proposal-class-properties"

### LINKED ISSUE #7233: Flow property declarations are not removed
### Input Code

https://github.com/babel/babel/blob/master/packages/babylon/src/parser/lval.js

```js
export default class LValParser extends NodeUtils {
  // Forward-declaration: defined in expression.js
  +checkReservedWord: (
    word: string,
    startLoc: number,
    checkKeywords: boolean,
    isBinding: boolean,
  ) => void;
  +parseIdentifier: (liberal?: boolean) => Identifier;
  +parseMaybeAssign: (
    noIn?: ?boolean,
    refShorthandDefaultPos?: ?Pos,
    afterLeftParse?: Function,
    refNeedsArrowPos?: ?Pos,
  ) => Expression;
  +parseObj: <T: ObjectPattern | ObjectExpression>(
    isPattern: boolean,
    refShorthandDefaultPos?: ?Pos,
  ) => T;
  // Forward-declaration: defined in statement.js
  +parseDecorator: () => Decorator;

  // ...
```

https://github.com/babel/babel/blob/master/.babelrc.js

```js
const config = {
  comments: false,
  presets: [
    ["@babel/env", envOpts],
    "@babel/flow"
  ],
  plugins: [
    ["@babel/proposal-class-properties", { loose: true }],
    "@babel/proposal-export-namespace-from",
    "@babel/proposal-numeric-separator",
    ["@babel/proposal-object-rest-spread", { useBuiltIns: true }],
  ]
};
```

### Expected Behavior

```js
function LValParser() {
  return _NodeUtils.apply(this, arguments) || this;
}
```

### Current Behavior

```js
function LValParser() {
  var _temp, _this;

  for (var _len = arguments.length, args = new Array(_len), _key = 0; _key < _len; _key++) {
    args[_key] = arguments[_key];
  }

  return (
    ((_temp = _this =
      _NodeUtils.call.apply(_NodeUtils, [this].concat(args)) || this),
    (_this.checkReservedWord = void 0),
    (_this.parseIdentifier = void 0),
    (_this.parseMaybeAssign = void 0),
    (_this.parseObj = void 0),
    (_this.parseDecorator = void 0),
    _temp) || _assertThisInitialized(_this)
  );
}
```

### Context

I was trying to build Babylon with gulp

### LINKED ISSUE #8417: v7 plugin-proposal-class-properties transform flow type annotation to undefined
## Bug Report

**Current Behavior**

`@babel/plugin-proposal-class-properties` transform flow type annotation to a property assigning undefined.

this may caught statement order unexpected error.

**Input Code**

```js
let code = `
class B {}
class Logger extends B {
  log: {
    error: boolean
  }
  silent: boolean
  log = {
    error: true
  }
  get silent(): boolean {
    return this._silent;
  }
  set silent(val: boolean) {
    this._silent = !!val;
    if (!this.log) {
      throw new Error('#log should not be nil.')
    }
    Object.keys(this.log).forEach(k => { /* do something... */  })
  }
  constructor() {
    super()
  }
}
`

let r = require("@babel/core").transform(code, {
  "presets": [ "@babel/preset-flow" ],
  "plugins": [
    [ "@babel/plugin-proposal-class-properties", { "loose": true } ]
  ]
})
console.log(r.code)
```

ouput =>

```js
class B {}

class Logger extends B {
  get silent() {
    return this._silent;
  }

  set silent(val) {
    this._silent = !!val;

    if (!this.log) {
      throw new Error('#log should not be nil.')   // <----- will throw if the annotation property assigning undefined
    }

    Object.keys(this.log).forEach(k => { /* do something... */ });
    });
  }

  constructor() {
    super();
    this.log = void 0;  <<<<<----
    this.silent = void 0; <<<<<---- with setter defined.
    this.log = {
      error: true
    };
  }
}
```

**Expected behavior/code**

```js
class B {}

class Logger extends B {
  get silent() {
    return this._silent;
  }

  set silent(val) {
    this._silent = !!val;

    if (!this.log) {
      throw new Error('#log should not be nil.')
    }

    Object.keys(this.log).forEach(k => { /* do something... */ });
  }

  constructor() {
    super();
    this.log = {
      error: true
    };
  }

}
```

**Environment**
- Babel version(s): [e.g. v6.0.0, v7.0.0-beta.55]
- OS: [e.g. OSX 10.13.4]
- How you are using Babel: `js`

**Possible Solution**

Should strip the `ClassProperty` type of `TypeAnnotation`.

### LINKED ISSUE #9105: [TypeScript] Typescript plugin overwrites parent class properties when the child narrows that property's type
## Bug Report

**Current Behavior**
When a class (B) extends another class (A) and includes a type definition to narrow the type of something specified in class A, it will set the narrowed property's value to undefined. This diverges from the TSC behavior.
The output from the below code under babel is `a, undefined`.
**Input Code**
```typescript
  class A {
    value: any
    constructor(value: any) {
      this.value = value
    }
  }

  class B extends A {
    value: string
  }

document.write(`${new A('a').value}, ${new B('b').value}`)
```

**Expected behavior/code**
The output from the code should be `a, b`.

.babelrc: 
```js
{
  "presets": [
    "@babel/env",
    "@babel/typescript"
  ],
  "plugins": [
  ]
}
```

**Environment**
- Babel version(s): 7.1.6
- Node/npm version: Node 10 / NPM 6
- OS: Any OS
- Monorepo: doesn't matter

**Possible Solution**
It seems like babel is setting `value = undefined` in the class property - it needs to check if that value is defined in the parent before overwriting it.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #9105, fixes #8417, closes #8584, fixes #7233
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR introduces a new option to the `@babel/plugin-proposal-class-properties` plugin: `ignoreUninitialized`. When it is enabled, `foo;` and `foo: type;` are removed instead of being (correctly) transpiled as if they were `foo = undefined;`.

This is needed to support the non-standard behavior of Flow and TypeScript, which use uninitialized fields as build-time-only type annotations instead of undefined fields (as mandated by the proposal). This bug has already been reported to those projects (Microsoft/TypeScript#12437, facebook/flow#6811).

When documenting this option we must add a warning that if you are using this option, your code will be incompatible with fields natively implemented by JS engines. We should propose this as an option to make the Flow/TypeScript -> Babel migration easer, but we should encurage users to refactor their code to follow the current proposal.

cc @babel/typescript @babel/flow


vbraun:
Is this going to be merged? Whether or not typescript changes their uninitialized field behavior, it will take quite some time for libraries to be updated for the breaking change. So to make @babel/typescript feasible today there needs to be a way to match tsc.

nicolo-ribaudo:
I'm closing this plugin since we are not going to diverge from the spec on pourpouse.
This can be implemented as a separate unofficial plugin that strips away class fields before that `proposal-class-properties` runs.

PR #9168, opened 2018-12-12
[parser] Handle flow comments with leading spaces

### LINKED ISSUE #9167: Parser doesn't recognize flow comments with leading spaces
## Bug Report

**Current Behavior**

Using `@babel/parser` v7.2.0.

Flow comments are recognized as flow typings when they start with `/*::`, `/*:`, and `/*flow-include`, but when there's spaces it sees them as an ordinary comment (i.e. `/*   ::`, `/* :` and `/* flow-include`).

**Input Code**
- [try it on flow.org](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBLAtgBzgJwBcwAlAUwEMBjQgGjAG8wBhOXOAOzI8NDAC5+9QgE8cZMADk4AEwkowAXzBR8bMAHJ8lGhoDc6PmGMnTZ86YCQsRAFoMHKjACuc1KPFgAYnDhgAvGAAzoT4DgDmBihAA)

```js
/* @flow */

import React, { Component/* ::, type Node */ } from 'react';

/*                          	flow-include
type Foo = string;
*/
```

**Expected behavior/code**
Spaces after the start of the comment should be considered the same as no spaces with flow comments.

The behavior for newlines after the start of the comment shouldn't change though, as flow still sees those as comments, i.e.:

```javascript
/*
:: type Foo = string; */ // <- This should be seen as a comment

/*  :: type Foo = string */ // <- This should be seen as a typing
```

vikr01:
| Fixed Issues?            | Fixes #9167  
| Documentation PR Link    | 

Handle flow comments that have spaces or tabs between the end of the comment opening and the start of the flow annotation:

```javascript
/*<spaces and tabs>:: type Foo = string; */
/*<spaces and tabs>flow-include type Bar = string; */
const foo/*<spaces and tabs>: string*/ = 'foo';
```


packages/babel-parser/src/plugins/flow.js
```diff
@@ -10,6 +10,8 @@ import { types as tc } from "../tokenizer/context";
 import * as charCodes from "charcodes";
 import { isIteratorStart } from "../util/identifier";
 
+const charCodeTab = 9;
+
```
danez: Made a PR to add this: https://github.com/xtuc/charcodes/pull/4

packages/babel-parser/test/fixtures/core/regression/octal-float-fail/options.json
```diff
@@ -1,3 +1,3 @@
 {
   "throws": "Unexpected token, expected \";\" (1:2)"
-}
\ No newline at end of file
+}
```
danez: Can you please remove all the unnecessary whitespace changes from this PR?

packages/babel-parser/test/fixtures/flow/comment/01-type-include/output.json
```diff
@@ -64,85 +64,18 @@
...
-                    "end": {
-                      "line": 2,
-                      "column": 19
-                    }
-                  }
-                }
-              },
-              "value": null
-            }
-          ]
```
danez: This is wrong.

danez:
Are you planing on finishing this or otherwise I can have a look and see what is wrong.
The overall idea seems good to me and we should support spaces/tabs there similar to how flow does it.

vikr01:
@danez Yeah if you could take a look I'd appreciate it, thanks.

packages/babel-parser/test/fixtures/core/regression/octal-float-fail/options.json
vikr01: Ah sorry, this got autogenerated during the test script.

packages/babel-parser/test/fixtures/flow/comment/01-type-include/output.json
amilajack: This comment is pretty vague. Can you please be more specific and explain exactly what's wrong with the code

packages/babel-parser/src/plugins/flow.js
```diff
@@ -2721,17 +2723,29 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+      ) {
+        firstNonWhiteSpace++;
+      }
+
+      const ch2 = this.input.charCodeAt(firstNonWhiteSpace);
+      const ch3 = this.input.charCodeAt(firstNonWhiteSpace + 1);
 
       if (ch2 === charCodes.colon && ch3 === charCodes.colon) {
-        return 4; // check for /*::
+        return firstNonWhiteSpace + 2; // check for /*::
```
vikr01: I think this was what the big issue was. `this.state.pos` was already added, so it was returning the index to shift to instead of the amount to shift by.

packages/babel-parser/test/fixtures/flow/comment/01-type-include/output.json
vikr01: Fixed in b0588ae2c9c50ca5502e920a19b41644215d1712

packages/babel-parser/src/plugins/flow.js
```diff
@@ -2721,17 +2723,32 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+      }
+
+      const ch2 = this.input.charCodeAt(shiftToFirstNonWhiteSpace + pos);
+      const ch3 = this.input.charCodeAt(shiftToFirstNonWhiteSpace + pos + 1);
 
       if (ch2 === charCodes.colon && ch3 === charCodes.colon) {
-        return 4; // check for /*::
+        return shiftToFirstNonWhiteSpace + 2; // check for /*::
       }
-      if (this.input.slice(this.state.pos + 2, 14) === "flow-include") {
```
danez: Oh this was wrong, the second argument to slice should be an Index and not the length. So flow-include wasn't working till now. Thanks for fixing this.

danez:
Thanks for fixing all the stuff, I looked into why the flow tests still failed and fixed some more.

PR #9171, opened 2018-12-12
Fix transforming empty export statement

### LINKED ISSUE #9155: Compiling file with `export {}` does not add `__esModule: true` sigil
# v7 Regression

> First check out: https://new.babeljs.io/docs/en/next/v7-migration.html
> Also a partial upgrade tool: https://github.com/babel/babel-upgrade

~If you have time to investigate, what PR/date introduced this issue.~

**Describe the regression**

Compiling a file with an ES modules' `export` declaration with no exported members does not include the `__esModule: true` helper property, resulting in unexpected behaviour when importing such a module.

Requiring such a module causes Babel to unwrap the module into an empty `default: {}` export due to the `_interopRequireDefault()` helper function checking for the presence of `__esModule` sigil, instead of leaving the exported thing `undefined`.

**Input Code**

```js
// source.mjs
export {}
```

**Output code**

```js
"use strict";
```

```js
// babel.config.js
'use strict'

module.exports = {
  sourceMaps: 'inline',
  plugins: [
    '@babel/syntax-object-rest-spread',
    ['@babel/transform-modules-commonjs', {
      allowTopLevelThis: false,
      // noInterop: true,
    }],
    '@babel/proposal-class-properties',
  ],
}
```

**Expected behavior/code**

```js
"use strict";

  value: true
});
```

**Environment**
- Babel version(s): 
   ```
   @babel/core@7.2.0
   @babel/plugin-proposal-class-properties@7.2.1
   @babel/plugin-syntax-object-rest-spread@7.2.0
   @babel/plugin-transform-modules-commonjs@7.2.0
   ```
- Node/npm version: node v11.3.0
- OS: macOS
- Monorepo: no

danez:
| Fixed Issues?            | Fixes #9155
| Documentation PR Link    | 
| Any Dependency Changes?  |

This adds a new flag indicating if the module has exports and removes the current logic to detect that. This fixes the case when an export statement is present but does not really export anything: `export {}`


packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js
```diff
@@ -52,18 +54,7 @@ export type LocalExportMetadata = {
  * Check if the module has any exports that need handling.
  */
 export function hasExports(metadata: ModuleMetadata) {
```
nicolo-ribaudo: Is this function still needed?

danez: No, but as it is a separate package I cannot remove it without breaking backwards compatibility if a mix between old and new version of Babel packages is used. At first I also wanted to use meta.hasExports directly in the transforms, but I ended up not doing it for the same reason.

Issue #9178, opened 2018-12-13
Cannot parse yield expression in pipeline


thiagoarrais:
This is a problem in the current parser (master version) for pipeline operators with yield expressions. In order to reproduce it, the pipeline proposal mode needs to be set to "smart" (see `options.json` below).

**Current Behavior**
Yield expressions inside pipelines do not parse (see input code below). Instead the parser spits the following error message:

```
Unexpected token, expected "," (4:12)

      2 |   const value = -5.9
      3 |   |> Math.abs
    > 4 |   |> (yield #);
        |             ^
      5 | }
```

**Input Code**
```js
function *myGenerator() {
  const value = -5.9
  |> Math.abs
  |> (yield #);
}
```

**Expected behavior/code**
Babel Parser should be able to parse.

**Babel Configuration (.babelrc, package.json, cli command)**

options.json
```js
{
  "plugins": [
    ["proposal-pipeline-operator", { "proposal": "smart" }],
    "syntax-async-generators"
  ]
}
```

**Environment**
- Babel version(s): master
- Node/npm version: Node 11/npm 6 (should reproduce in any version)
- OS: Ubuntu/docker node:11 (should reproduce in any version)
- How you are using Babel: `make test`

**Additional context/Screenshots**

Running tests on [this branch](https://github.com/thiagoarrais/babel/tree/cannot-parse-yield-expression-in-pipeline-9178) should reproduce the error: https://github.com/thiagoarrais/babel/tree/cannot-parse-yield-expression-in-pipeline-9178

js-choi:
I think I isolated the problem. `parseYield` uses token objects’ `beforeExpr` property to determine whether it should attempt to consume a subexpression. (This unique to `yield` because it needs to check for `*`; `parseAwait` for instance needs no such check.) Currently `tokenizer/types`’ `types.hash` lacks `beforeExpr`, which causes the bug. I’ll make a pull request fixing this soon.

PR #9179, opened 2018-12-13
Transform for the smart pipeline operator proposal

thiagoarrais:
| Documentation PR Link    |

This implements a first draft for the smart pipeline proposal. It is based on IIFE and does not support async/await inside the pipelines yet. That will need to be added by a subsequent PR.

Relates to https://github.com/babel/babel/projects/9#card-15461643


mAAdhaTTah:
Couple quick comments:

1. I _think_ we need to be passing the `proposal` option to the parser.
2. It'll be easier to maintain the various proposal if there are clear "branches" for each proposal, and then extracting shared functionality to a module used by all of the branches.
3. I could be wrong, but I'm not sure tests in a subfolder run. They didn't for the parser, as your test setup is pretty similar to what I initially did for the parser work. My first thought looking at the code is whether the minimal still works. Which it might! Just worth double checking.

thiagoarrais:
> 1. I think we need to be passing the proposal option to the parser.

I think this is done by [line 21 in babel-plugin-syntax-pipeline-operator/src/index.js](https://github.com/babel/babel/blob/ab606c74f9503bb37341a77fd82c3a68ecc98d52/packages/babel-plugin-syntax-pipeline-operator/src/index.js#L21) but I'm not sure this is what you're talking about.

> 2. It'll be easier to maintain the various proposal if there are clear "branches" for each proposal,

Good idea. I guess a first step on this direction would be having separate visitors and switching on the parseOpts proposal.

> 3. I could be wrong, but I'm not sure tests in a subfolder run.

Yeah, they do without a problem. Maybe parser tests are different somehow?

mAAdhaTTah:
1. Yes it is, my b.
2. That sounds good.
3. Could be.

Thanks!

nicolo-ribaudo:
> It'll be easier to maintain the various proposal if there are clear "branches" for each proposal,

Yeah, for example I put the two different decorators transformers in two different files (https://github.com/nicolo-ribaudo/babel/blob/33ac9a246c4b4e45f8f1b63122f4ba502f3cf32e/packages/babel-plugin-proposal-decorators/src/index.js)

thiagoarrais:
Nice. I'll definitely steal that idea.

js-choi:
This version of the transformer uses IIFEs, so it doesn’t support `await` and `yield`, right? See also a similar problem with the `do`-expressions transformer in #3780.

thiagoarrais:
Yeah . You're right, @js-choi

packages/babel-plugin-proposal-pipeline-operator/src/index.js
```diff
@@ -1,64 +1,23 @@
...
+import minimalVisitor from "./minimalVisitor";
+import smartVisitor from "./smartVisitor";
 
-export default declare(api => {
+export default declare((api, options) => {
   api.assertVersion(7);
 
+  // FIXME: Hacky... Need a way to get rid of this
+  delete smartVisitor._exploded;
+  delete smartVisitor._verified;
```
thiagoarrais: I did this because the plugin validation kept complaining that `.visitor["_exploded"]` and `.visitor["_verified"]` must be functions. Is there other way to honor that?

Wassna002:
"name": "Absent",
  "version": "0.1.0",
  "private": true,
  "devDependencies": {
    "babel-preset-react-native-stage-0": "^1.0.1",
    "jest": "^23.5.0",
    "jest-react-native": "^18.0.0",
    "react-test-renderer": "16.3.1"
  },

packages/babel-plugin-proposal-pipeline-operator/src/index.js
thiagoarrais: I haven't dug the root cause yet, but it seems like they are inserted by the plugin execution infrastructure. They somehow only showed up after the visitors were extracted to their own modules. When the visitors were declared inside the `declare` block itself, this doesn't arise. Maybe the visitor is getting reused between plugin executions?

packages/babel-plugin-proposal-pipeline-operator/src/minimalVisitor.js
```diff
@@ -0,0 +1,55 @@
+import { types as t } from "@babel/core";
+
+const minimalVisitor = {
+  BinaryExpression: {
+    exit(path) {
```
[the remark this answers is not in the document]
thiagoarrais: Some of the calls in this function need `Expression` nodes and the smart proposal child nodes aren't expressions, but can be converted by the smart proposal visitor. Running on exit guarantees that the child nodes are already converted.

packages/babel-plugin-proposal-pipeline-operator/src/smartVisitor.js
```diff
@@ -0,0 +1,36 @@
+import { types as t } from "@babel/core";
+import minimalVisitor from "./minimalVisitor";
+
+const updateTopicReferenceVisitor = {
+  PipelinePrimaryTopicReference(path) {
+    path.replaceWith(this.topicId);
+  },
+  AwaitExpression(path) {
+    throw path.buildCodeFrameError(
+      "await is not supported inside pipeline expressions yet",
```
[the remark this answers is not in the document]
thiagoarrais: Yup. But I'm holding off this one until I can write a test case for that. This should happen in #9178.

packages/babel-plugin-proposal-pipeline-operator/src/smartVisitor.js
```diff
@@ -0,0 +1,36 @@
...
+    );
+  },
+};
+
+const smartVisitor = {
+  ...minimalVisitor,
+  PipelineTopicExpression(path) {
+    path.get("expression").traverse(smartVisitor);
```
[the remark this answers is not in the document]
thiagoarrais: The main goal here is dealing with nested pipelines. This was introduced before switching to running on exit. Maybe that isn't necessary any longer. I'll take a look at that.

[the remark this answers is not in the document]
thiagoarrais: I'm coming back after fiddling with the code.

I believe I will be able to get rid of the subtraversal (maybe the one [on line 22](https://github.com/babel/babel/pull/9179/files/6921043b8f45abf124ae32c4e22557b3dbf2d659#diff-a4d6f2fbd24651e9f03fbc0acb332dd0R22) also) by storing a stack with some state in the visitor. But I'm not sure that is actually recommended (the handbook seems to discourage it). Please let me know if you think that is a worthline line of work.

Additionally can you please elaborate on your approach?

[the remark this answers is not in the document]
nicolo-ribaudo: If I understood it correctly, this traversal is needed to prevent parent topic expression from "stealing" topic references in nested expressions:

```js
foo |> # |> #;

// becomes

foo |> (_topic => _topic |> (_topic2 => _topic2));

// and not

foo |> (_topic => _topic |> (_topic2 => _topic));
```

This visitor does it by transforming the innter pipeline expressions first, but I am pretty sure that we can obtain the same behavior by skipping `PipelineTopicExpression` in `updateTopicReferenceVisitor`:

```js
const updateTopicReferenceVisitor = {
  PipelinePrimaryTopicReference(path) {
    path.replaceWith(this.topicId);
  },
  AwaitExpression(path) {
      "await is not supported inside pipeline expressions yet",
    );
  },
  PipelineTopicExpression(path) {
    path.skip();
  },
};
```

This probably would also fix https://github.com/babel/babel/pull/9179#discussion_r242001599, since we aren't using `smartVisitor` twice anymore.

packages/babel-plugin-proposal-pipeline-operator/src/index.js
```diff
@@ -1,64 +1,23 @@
...
-          : t.callExpression(right, [t.cloneNode(placeholder)]);
-        path.replaceWith(
-          t.sequenceExpression([
-            t.assignmentExpression("=", t.cloneNode(placeholder), left),
-            call,
-          ]),
-        );
-      },
-    },
+    visitor: visitorsPerProposal[options.proposal],
```
nicolo-ribaudo: We should add validation for the `proposal` optoin.

mAAdhaTTah: I think the syntax plugin already does this.

packages/babel-plugin-proposal-pipeline-operator/src/smartVisitor.js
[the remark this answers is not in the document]
thiagoarrais: That is much more clear now. It is an excellent idea! I'm on my phone now. But I'll definitely update the code as soon as possible.

> Btw, I really like how smartVisitor inherits from minimalVisitor: being able to comment out [...]

I hadn't noticed that. It is really nice indeed!

packages/babel-plugin-proposal-pipeline-operator/src/index.js
thiagoarrais: Do we need to add it here too?

js-choi: Is it possible for this change to be observable by a test?

nicolo-ribaudo: Yes, you can test the error message by adding `"error": "Expected error message"` in the options.json file of the fixture.

packages/babel-plugin-proposal-pipeline-operator/src/index.js
thiagoarrais: This was fixed by 36124fe because it eliminates the subtraversal. Thanks for the tip, @nicolo-ribaudo!

packages/babel-plugin-proposal-pipeline-operator/src/smartVisitor.js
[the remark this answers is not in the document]
thiagoarrais: Your suggestion worked wonders. This is resolved AFAIK.

packages/babel-plugin-proposal-pipeline-operator/src/index.js
js-choi: Well, even then: That test still wouldn’t distinguish between the parser throwing the error versus the transform plugin throwing the error, right?

nicolo-ribaudo: Yeah that's right. But usually parser error messages are different from plugin error messages. That is the important part; not where the error actually comes from.

thiagoarrais: I find the syntax plugin message (below) is clear enough in this case.

```
The pipeline operator plugin requires a 'proposal' option.
'proposal' must be one of: minimal, smart. More details:
https://babeljs.io/docs/en/next/babel-plugin-proposal-pipeline-operator
```

Maybe it should be moved to the transform plugin. What about holding it off for a later PR?

nicolo-ribaudo: Oh I didn't noticed that the link alreay points to the transform plugin. It is ok as-is :+1:

thiagoarrais:
Nice! I'm done here for now. Please let me know if you need anything else.

mAAdhaTTah:
Damn, I need to hurry up!

js-choi:
@thiagoarrais: Great work, again. Let me know on whether you plan to tackle #9178 (`yield` parsing) or the IIFEs-to-flattened-expressions rewrite. For what it’s worth, I expect to submit a pull request for adding `?`-placeholder support to babel-parser within two days, time permitting, and that will rename a lot of babel-parser’s pipeline test fixtures.

@mAAdhaTTah: No need to rush! We still don’t have `await`/`yield` support yet due to the use of IIFEs, after all.

thiagoarrais:
Since, #9216 landed, I've rebased to the latest master and added a check for yield.

thiagoarrais:
Bumping this. Let me know if you need anything else, @nicolo-ribaudo.

nicolo-ribaudo:
It looks good to me. I didn't merge it because we have a policy of two :heavy_check_mark: from core team members, but reviews by @mAAdhaTTah and @js-choi would probably be equally valuable on this topic.

packages/babel-plugin-proposal-pipeline-operator/test/fixtures/smart/topic-inside-arrow-function/input.js
```diff
@@ -0,0 +1,5 @@
+const result = -2.2   // -2.2
+  |> Math.floor       // -3
+  |> (() => Math.pow(#, 5))   // () => -243
```
mAAdhaTTah: @js-choi Question on the spec behavior: if we didn't use a topic token here, this arrow would get executed?

```js
-2.2
    |> (x => Math.pow(x, 5))
```

This is also valid in Smart. Are we checking whether the token appears in the arrow body?

packages/babel-plugin-proposal-pipeline-operator/src/minimalVisitor.js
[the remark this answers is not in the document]
mAAdhaTTah: This feels a bit like the logic is getting mixed between them. This might be a bit pendantic, but I'd extract the shared logic to a separate module and run them at different times.

packages/babel-plugin-proposal-pipeline-operator/src/index.js
```diff
@@ -1,64 +1,19 @@
...
+import smartVisitor from "./smartVisitor";
 
-export default declare(api => {
+export default declare((api, options) => {
   api.assertVersion(7);
 
+  const visitorsPerProposal = {
+    minimal: minimalVisitor,
+    smart: smartVisitor,
+  };
```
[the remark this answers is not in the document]
thiagoarrais: Good idea. Wil do it.

packages/babel-plugin-proposal-pipeline-operator/src/minimalVisitor.js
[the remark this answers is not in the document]
thiagoarrais: We could move this function to its own module and bind it to `entry` (or the default `BinaryExpression') on minimal and to `exit` on smart. But I'm not really sure about that. It feels like splitting hairs.

Plus: running on exit works great with the flattening PR that I've got in the works.

thiagoarrais:
Thank you everyone! I'll start work on flattening expressions (async/await) next.

PR #9184, opened 2018-12-14
Allow keywords to be used in type annotations

danez:
| Fixed Issues?            | 
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This allows keywords to be used as type annotations like

```js
function foo(a:function, b: switch) {}
```

Flow does also not error in this cases.

### DIFF
--- a/packages/babel-parser/src/plugins/flow.js
+++ b/packages/babel-parser/src/plugins/flow.js
@@ -1384,6 +1384,10 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         default:
           if (this.state.type.keyword === "typeof") {
             return this.flowParseTypeofType();
+          } else if (this.state.type.keyword) {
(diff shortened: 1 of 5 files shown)


nicolo-ribaudo:
```js
import type { s as switch, f as function } from "module";

function foo(a:function, b: switch) {}
```

Perfectly valid code :stuck_out_tongue:

jedwards1211:
Just imagine if JS didn't have any reserved words :smile:

```
import { as as default } from "module"

class class {
  async async() {
    function const() { }
    const function = let('foo') = () => { function(function) {...} }
    function().function(function)
    let switch = 'blah', case = 'blah', break = 'argh!'
    let await = () => ...
    await = await await()
    switch (switch) {
      case case: return break
      case default: return await
      default: return case
    }
  }
}

Issue #9251, opened 2018-12-28
@babel/traverse: can't traverse Property using path.traverse with AST parsers different then @babel/parser


coderaiser:
## Bug Report

**Current Behavior**
`@babel/traverse` dot not traverse `AST`-nodes `Property` of `ObjectPattern` generated by different parsers then `@babel/parser` using `path.traverse`.

**Input Code**
- REPL or Repo link if applicable:

As an example I used, `acorn`, `espree` and `cherow`. All of them use `Property` instead of `ObjectProperty` in `ObjectPattern`. There is a `fix` visitor that renames `Property` to `ObjectProperty` and with this fix everything works as expected. 

```js
const espree = require('espree');
const babel = require('@babel/parser');
const cherow = require('cherow');
const acorn = require('acorn');

const traverse = require('@babel/traverse').default;

const source = `
    const {a} = obj;
`;

const babelAST = babel.parse(source);
const cherowAST = cherow.parse(source);
const acornAST = acorn.parse(source)
const espreeAST = espree.parse(source, {
    ecmaVersion: 6,
});

const visitor = {
    enter(path) {
        if (path.type !== 'ObjectPattern')
            return;

        path.traverse({
            noScope: true,
            Property(path) {
                console.log('Property: only get here using babel');
            },
            ObjectProperty(path) {
                console.log('ObjectProperty: only get here using babel');
            },
        });
    }
};

const fix = {
    enter(path) {
        if (path.type !== 'Property')
            return;

        path.node.type = 'ObjectProperty';
    },
};

// uncoment to make things work as expected
//traverse(cherowAST, fix);
//traverse(espreeAST, fix);
//traverse(acornAST, fix);

traverse(babelAST, visitor);
traverse(cherowAST, visitor);
traverse(acornAST, visitor);
traverse(espreeAST, visitor);
```

**Expected behavior/code**
Would be great if `path.traverse` works the same way when `Property` used, because other parsers uses this name instead of `ObjectProperty` according to [estree standard](https://github.com/estree/estree/blob/master/es2015.md#objectpattern).

**Babel Configuration (.babelrc, package.json, cli command)**

```
npm i @babel/parser @babel/traverse acorn espree cherow
```

**Environment**
- Babel version(s): [e.g. v6.0.0, v7.0.0-beta.34]` latest`
- Node/npm version: [e.g. Node 8/npm 5] node `v11`
- OS: [e.g. OSX 10.13.4, Windows 10]: linux
[...]

PR #9253, opened 2018-12-28
@babel/types: added standard estree Property type

### LINKED ISSUE #8879: Add ESTree tokens definitions to @babel/types/src/definitions
## Feature Request

I want to propose it because of issue in https://github.com/babel/babel-eslint/issues/681#issuecomment-430101041. The cause behind the issue is some [monkey patches in `babel-eslint`](https://github.com/babel/babel-eslint/blob/master/lib/babylon-to-espree/toAST.js#L9) to make `@babel/types` supports the [ESTree](https://github.com/estree/estree) types though these types are generated by babel itself (specifically `@babel/parser` with option plugin: ['estree']).

I found it's tricky to fix it in `babel-eslint`, and it seems not necessary to pass in types to `@babel/traverse` as they are actually standard types defined by [ESTree](https://github.com/estree/estree). It's just [Babel outputs something different than ESTree](https://babeljs.io/docs/en/next/babel-parser.html#output) by default, but it can still outputs ESTree types.

As `@babel/parser` supports the ESTree types, it's rational that `@babel/types` supports them too. And the types could be placed with the other type system (like JSX, TypeScript) in `packages/babel-types/src/definitions/`.
Specifically, these types need to be added: `Literal`, `Property`, `MethodDefinition`.

Problem here it how `Program`, `BlockStatement`, `ClassMethod`, `ObjectProperty`, `ObjectMethod` should be treated here as though their definitions are slightly different in Babel & ESTree, but they share same name.

None

This is mainly for babel itself, no much user impact, except for the users who have already done something tricky to support ESTree types, like `babel-eslint`.

coderaiser:
Added standard `estree` [Property](https://github.com/estree/estree/blob/master/es2015.md#objectpattern) of `ObjectPattern`

Adds ability to use different parsers (like `acorn`, `espree`, `cherow` and all estree standard compatible) with @babel/traverse, not only @babel/parser

| Fixed Issues?            | Fixes #8879, #9251  
| Documentation PR Link    | 

### DIFF
--- a/packages/babel-types/src/definitions/core.js
+++ b/packages/babel-types/src/definitions/core.js
@@ -687,6 +687,49 @@ defineType("ObjectProperty", {
   aliases: ["UserWhitespacable", "Property", "ObjectMember"],
 });
 
+defineType("Property", {
+  builder: ["key", "value", "computed", "shorthand", "decorators"],
+  fields: {
+    computed: {
+      validate: assertValueType("boolean"),
+      default: false,
+    },
+    key: {
+      validate: (function() {
+        const normal = assertNodeType(
+          "Identifier",
+          "StringLiteral",
+          "NumericLiteral",
+        );
+        const computed = assertNodeType("Expression");
+
+        return function(node, key, val) {
+          const validator = node.computed ? computed : normal;
+          validator(node, key, val);
+        };
+      })(),
+    },
+    value: {
+      // Value may be PatternLike if this is an AssignmentProperty
+      // https://github.com/babel/babylon/issues/434
+      validate: assertNodeType("Expression", "PatternLike"),
+    },
+    shorthand: {
+      validate: assertValueType("boolean"),
+      default: false,
+    },
+    decorators: {
+      validate: chain(
+        assertValueType("array"),
+        assertEach(assertNodeType("Decorator")),
+      ),
+      optional: true,
+    },
+  },
+  visitor: ["key", "value", "decorators"],
+  aliases: ["UserWhitespacable", "ObjectProperty", "ObjectMember"],
+});
+
 defineType("RestElement", {
   visitor: ["argument", "typeAnnotation"],
   builder: ["argument"],


nicolo-ribaudo:
Babel doesn't aim to be compatible with ESTree (Babel's ast is a fork of it). In order to use other parsers, you should somethow convert their output to repect Babel's ast format.

coderaiser:
But babel has an [estree plugin](https://babeljs.io/docs/en/babel-parser#output) which makes it compatible with the standard and other parsers. Everybody will win if `@babel/traverse` can handle a couple standard cases.

Here is code of [@babel/traverse](https://github.com/babel/babel/blob/v7.2.5/packages/babel-traverse/src/index.js#L57) which observes types and can not find visitors of `Property` because there is no such property. We are talking not about `babel` or `@babel/parser` now, but about tools which works with a `JavaScript AST`, first of all.

>  In order to use other parsers, you should somethow convert their output to repect Babel's ast format.

There is a fix:

```js
const fix = {
    Property(path) {
        path.node.type = 'ObjectProperty';
   }
};
```

But it is doesn't works for [scope binding](https://github.com/cherow/cherow/issues/273).

fkleuver:
Hi @nicolo-ribaudo, I'm working with @coderaiser on this from the cherow side. I'm just trying to better understand something:

> Babel doesn't aim to be compatible with ESTree

I'm not judging or criticizing or questioning. I'm legitimately curious about the reasoning behind this. Did you find that ESTree had some problems or limitations with it, doesn't make sense in some way? Or did forking it simply allow you to do some things in a cleaner manner in general?

nicolo-ribaudo:
I was not there when Babel started diverging from ESTree, so I don't know why :confused:
FWIW, we will talk about it at the next meeting (babel/notes#77)

fkleuver:
Thanks @nicolo-ribaudo looking forward to the outcome :)

loganfsmyth:
I'm curious to hear what other think at the meeting, but I'll lay out my thoughts here. While this PR on its own isn't unreasonable, I think the motivation behind it requires much more consideration.

The intent of the `estree` parser plugin is purely for tooling that wants to use Babel's parser and then manipulate the AST using ESTree-specific tooling. `@babel/types` is not meant to be one of those tools. If you want to use `@babel/types` then the expectation is that you'd leave the AST in the form Babel expects.

In my opinion, it probably would have been ideal for Babel to have stayed with ESTree. That said, this PR concerns me because it's only a workaround for one specific case. If we only support ESTree in _some_ places, what places should they be and why? How are any users expected to know what does and doesn't? In my view, we can either be ESTree or not, but having our tooling try to support both is a recipe for missed edge cases and bugs across the whole codebase.

If we want to be ESTree, that would be a breaking major-version change that would then also be a breaking change for any community plugins written expecting our existing AST structure, which would have to be updated to handle both cases, or else release their own new major versions. Not necessarily the end of the world, but certainly not trivial and requiring a lot of thought.

I honestly don't know where I stand on ESTree support long-term. In a perfect world, I'd love Babel to be more parser/generator/language agnostic in general, because I think Babel's framework for code transformation could probably benefit a lot if non-JS things as well. Then whether you use an ESTree parser or a Shift parser or whatever would just mean your plugins needed to handle them. Unfortunately, I don't think our current traversal-based transformation architecture makes that very easy, so that will be a work-in-progress for a while and who even knows if it would be a net win for Babel.

nicolo-ribaudo:
In the jan-14 meeting we discussed about this issue. Supporting both our AST format and ESTree would be incompatible with our current plugins ecosystem, since plugins don't know what to do with a `Property` or a `Literal`. We might consider switching back to the ESTree ast format, but there is a problem: ESTree mainly specifies standard features, so we would have to invent our own nodes for experimental features. This would lead again to differences in the long term.

Our parser's `estree` plugin is only needed for `babel-eslint`.

If you have an ESTree AST and you need to traverse it using Babel, we reccomend converting the AST using a tool like @coderaiser's https://github.com/coderaiser/estree-to-babel

PR #9262, opened 2018-12-30
WIP: try to make flow enforce the signature of isType() 

alsuren:
The signature of isType() is `isType(nodeType: string, targetType: string): boolean` but if I assert that nodeType and targetType are both strings (as in this patch) I get a bunch of test failures:

    Test Suites: 86 failed, 47 passed, 133 total
    Tests:       155 failed, 54 skipped, 6600 passed, 6809 total
    Snapshots:   7 passed, 7 total
    Time:        47.333s, estimated 76s

In my other branch (re-writing bits of babel in rust) it's easier for me if I rely on nodeType and targetType both being strings, but in a bunch of tests they end up being undefined (the type property is not set on a bunch of nodes).

The question is: should I write a patch which fixes the above tests, or should I change the documentation/signature/tests for isType() to make it explicit that it needs to survive its arguments being undefined?

| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Not sure. Needs discussion.
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-traverse/src/path/index.js
+++ b/packages/babel-traverse/src/path/index.js
@@ -1,5 +1,7 @@
+// @flow
 import type { HubInterface } from "../hub";
 import type TraversalContext from "../context";
+// $FlowFixMe Not sure why flow can't find this.
 import * as virtualTypes from "./lib/virtual-types";
 import buildDebug from "debug";
 import traverse from "../index";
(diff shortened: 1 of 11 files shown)


nicolo-ribaudo:
I like the assertion, but we must defer it to Babel 8 since it is breaking

alsuren:
Okay. I will use the assertion to find the tests that need fixing and then
revert the assertion.

This is quite a performance sensitive function so I probably wouldn't
suggest adding the assertion in production.

I will continue developing on this branch.

D.

hulkish:
should these be symbol comparisons instead of strings?

packages/babel-traverse/src/path/index.js
```diff
@@ -138,44 +156,143 @@ export default class NodePath {
...
-  NodePath_inference,
-  NodePath_replacement,
-  NodePath_evaluation,
-  NodePath_conversion,
-  NodePath_introspection,
-  NodePath_context,
-  NodePath_removal,
-  NodePath_modification,
-  NodePath_family,
-  NodePath_comments,
```
alsuren: I feel like there's a sensible way to refactor this mixin pattern. It might be worth asking someone more adept with flow for suggestions. For now, I used a bit of grep and sed to generate the below and pasted it in.

The `is${type}` and `assert${type}` functions are going to want some kind of generated file.

I wonder if we could do make a module that does `export * from $file` for each of the above modules (plus some generated files with the is* definitions in, and then somehow `extend` from that?

PR #9264, opened 2019-01-01
Add follow on twitter badge

jamesgeorge007:
I think it's good to have a `follow on twitter` badge since there is already one for slack :thinking: 
> It looks like this:

[Twitter](https://twitter.com/intent/follow?screen_name=babeljs)


hzoo:
great thanks again!

PR #9275, opened 2019-01-03
flow type update: babel-types.isType(?string, string): boolean

alsuren:
NodePath.type is a ?string so we need to accept undefined here too.

I didn't see that @nicolo-ribaudo edited his comment on https://github.com/babel/babel/pull/9262 so I went down a massive rabbit-hole trying to get flow to check things properly. This is a micro-pr that fixes the original isType() discrepancy that I spotted.

If you want me to cherry-pick anything from the other PR, I'd be happy to.

| Fixed Issues?            |
| Patch: Bug Fix?          | bug in our type annotations
| Major: Breaking Change?  | no: isType()'s signature is now more permissive than before, and its return type has not changed.
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


existentialism:
@alsuren thanks!

PR #9288, opened 2019-01-06
Test262 update

existentialism:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Will follow-up this up with a few PRs to fix some spec parsing issues.


nicolo-ribaudo:
We should sort alphabetically the whitelist so that it is actually possible to see the real changes.

PR #9292, opened 2019-01-07
Throw error if TypeScript class has empty implements

existentialism:
| Fixed Issues?            | Y
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


armano2:
i just checked and looks like same there is also no check in interface heritage:

```ts
interface foo extends {
}
```

i can report this as separated ticket :)

existentialism:
Updated to handle interface/extends as well

PR #9294, opened 2019-01-07
@babel/generator: Add defined check to append function str

owap:
| Fixed Issues?            |
| Tests Added + Pass?      | Tests Pass
| Documentation PR Link    | 

If the `append` function is passed a null or undefined string, it will throw an error, causing some builds to fail. Instead, we simply don't append when nothing is passed :)

### DIFF
--- a/packages/babel-generator/src/buffer.js
+++ b/packages/babel-generator/src/buffer.js
@@ -117,6 +117,8 @@ export default class Buffer {
     filename: ?string,
     force?: boolean,
   ): void {
+    if (!str) return;
+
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
When do we pass null or undefined to that function? The type annotation says that it only accepts strings.

owap:
I suppose normally it shouldn't, unless the validator is invalid, which I must have done by hand when trying to troubleshoot some tree-shaking that wasn't working.

Feel free to close this issue if you feel like having an extra null/undefined check isn't important 👍

PR #9302, opened 2019-01-08
@babel/parser(ts): Add parsing of type import

armano2:
| Fixed Issues?            | #7749
| Documentation PR Link    | 👎 

This PR is continuation of work done in #8798 by @Kovensky few months ago.
This PR adds new node `TSImportType`.

Comparing to how TS parses this there is one difference
```ts
typeof import("./a")
```
in typescript is represented by `ImportType[typeof=true]`
this implementation will parse it into: `TSTypeQuery[exprName='TSImportType']`

i can easily change that if you think it will be better.

### DIFF
--- a/packages/babel-parser/src/plugins/typescript.js
+++ b/packages/babel-parser/src/plugins/typescript.js
@@ -220,26 +220,28 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return result;
     }
 
-    tsParseEntityName(allowReservedWords: boolean): N.TsEntityName {
-      let entity: N.TsEntityName;
-      // TODO: package this into its own tsParse function
-      if (this.state.type === tt._import) {
-        entity = this.startNode();
-        this.next();
-        if (!this.match(tt.parenL)) {
-          this.unexpected(null, tt.parenL);
-        }
-        this.eat(tt.parenL);
-        const args = this.parseCallExpressionArguments(tt.parenR, false, {
-          start: -1,
-        });
-        // TODO: assert args is length 1
-        entity.argument = args[0];
-        // TODO: make a TS-specific node type
-        entity = this.finishNode(entity, "Import");
-      } else {
-        entity = this.parseIdentifier();
+    tsParseImportType(allowReservedWords: boolean): N.TsImportType {
+      const node: N.TsImportType = this.startNode();
+      this.expect(tt._import);
+      this.expect(tt.parenL);
+      if (!this.match(tt.string)) {
+        throw this.unexpected();
       }
+      node.argument = this.parseLiteral(this.state.value, "StringLiteral");
+      this.expect(tt.parenR);
+
+      if (this.match(tt.dot)) {
+        this.eat(tt.dot);
+        node.qualifier = this.tsParseEntityName(allowReservedWords);
+      }
+      if (!this.hasPrecedingLineBreak() && this.isRelational("<")) {
+        node.typeParameters = this.tsParseTypeArguments();
+      }
+      return this.finishNode(node, "TSImportType");
+    }
+
+    tsParseEntityName(allowReservedWords: boolean): N.TsEntityName {
+      let entity: N.TsEntityName = this.parseIdentifier();
       while (this.eat(tt.dot)) {
         const node: N.TsQualifiedName = this.startNodeAtNode(entity);
         node.left = entity;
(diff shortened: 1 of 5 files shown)


armano2:
@existentialism you guys have to fix developing on windows, this is killing me xd

finally its finished xd

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -220,6 +220,26 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+        throw this.unexpected();
+      }
+      node.argument = this.parseLiteral(this.state.value, "StringLiteral");
+      this.expect(tt.parenR);
+
+      if (this.match(tt.dot)) {
+        this.eat(tt.dot);
+        node.qualifier = this.tsParseEntityName(allowReservedWords);
+      }
+      if (!this.hasPrecedingLineBreak() && this.isRelational("<")) {
```
DanielRosenwasser: TypeScript doesn't actually check for a preceding line break; do you do similar things throughout Babylon?

armano2: that's good point i will change that

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -220,6 +220,26 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+      this.expect(tt._import);
+      this.expect(tt.parenL);
+      if (!this.match(tt.string)) {
+        throw this.unexpected();
+      }
+      node.argument = this.parseLiteral(this.state.value, "StringLiteral");
+      this.expect(tt.parenR);
+
+      if (this.match(tt.dot)) {
+        this.eat(tt.dot);
```
nicolo-ribaudo: You can directly use `if (this.eat(tt.dot))`

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -220,6 +220,26 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+        node.qualifier = this.tsParseEntityName(allowReservedWords);
```
nicolo-ribaudo: Instead of getting `allowReservedWords` as a parameter, can't we just pass `true` here?

armano2: ok will do so

packages/babel-parser/src/plugins/typescript.js
armano2: i'm kinda new to this parser, thanks for suggestion

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -220,6 +220,25 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return result;
     }
 
+    tsParseImportType(): N.TsImportType {
+      const node: N.TsImportType = this.startNode();
+      this.expect(tt._import);
+      this.expect(tt.parenL);
+      if (!this.match(tt.string)) {
+        throw this.unexpected();
```
danez: Can we add a nice error message here please. Something like "A TypeScript type import requires a string literal as its argument."

armano2: typescript reports here: `String literal expected.`

armano2: but i'm ok with adding there whatever :)

armano2: but we should be consistent about this, and in `tsParseExternalModuleReference` there is no message.

```
      if (!this.match(tt.string)) {
        throw this.unexpected();
      }
```

armano2: @danez what do you think about `Argument in a type import must be an String Literal`

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -225,7 +225,10 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       this.expect(tt._import);
       this.expect(tt.parenL);
       if (!this.match(tt.string)) {
-        throw this.unexpected();
+        throw this.unexpected(
+          null,
+          "Argument in a type import must be an string literal",
```
danez: Nice, only thing it should be "... a string literal"

packages/babel-parser/src/plugins/typescript.js
danez: 👍

Kovensky:
Thanks for following up on my PR 🙏

hzoo:
@danez Going to make a PR tomorrow!

PR #9311, opened 2019-01-11
Disallow trailing comma after rest

nicolo-ribaudo:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

I started by fixing a bug where Babel did not throw for `[...a,]=[]`, than found that the logic to disallow it was duplicated in a lot of places: so I merged them together and properly used `this.state` instead of passing around the `refTrailingCommaPos` "reference" parameter.

NOTE: In the diff it looks like I added code (`312 insertions(+), 262 deletions(-)`), but excluding the test folder it is `79 insertions(+), 100 deletions(-)`


packages/babel-parser/src/tokenizer/state.js
```diff
@@ -87,6 +89,11 @@ export default class State {
   // TODO
   input: string;
 
+  // A comma after "...a" is only allowed in spread, but not in rest.
+  // Since we parse destructuring expressions ar array/object literals
```
existentialism: Nit: ar -> in (I think?)

nicolo-ribaudo: I think that I meant "as"

nicolo-ribaudo:
I'm merging this since another PR I'm working on depends on this PR.

PR #9313, opened 2019-01-11
Disallow `new import(x)` and `import(x,)`

nicolo-ribaudo:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes, 100% :heavy_check_mark: Test262 for dynamic import
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-parser/src/parser/expression.js
```diff
@@ -1306,21 +1316,24 @@ export default class ExpressionParser extends LValParser {
     }
 
     node.callee = this.parseNoCallExpr();
-    if (
+
+    if (node.callee.type === "Import") {
+      this.raise(node.callee.start, "import(...) can't be constructed");
```
existentialism: Feel like Chrome's error here is clearer?

> Cannot use new with import

nicolo-ribaudo:
`imports` should only enforce commas in statements,would you mind filing an issue?

Sorry to have broken your build, but every bug fix is breaking is someone relies on the buggy behavior 😅

nicolo-ribaudo:
@babel/babel We might consider reverting this PR until https://github.com/tc39/proposal-dynamic-import/issues/67 is closed.

PR #9315, opened 2019-01-11
Parse class heritage as strict mode code

### LINKED ISSUE #8186: Class heritage should be evaluated as strict code
## Bug Report

This is extremely low priority, but I'm just filing it since I noticed it.

**Current Behavior**
Class expressions/declarations are automatically strict. We've now added logic to ensure that `use strict` is added to classes properly, but the class parent heritage is still executed outside of the strict IIFE, so it is executed non-strictly.

**Input Code**

```js
var o = Object.freeze({});

var f = class extends (o.prop = 4, function(){}) { method(){}}
```

**Expected behavior/code**
Should throw an exception since `o.prop = 4` is trying to add a property to a frozen object.

**Possible Solution**
It seems like we'd need to consider moving the heritage evaluation inside the IIFE.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #8186
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/src/parser/statement.js
+++ b/packages/babel-parser/src/parser/statement.js
@@ -994,8 +994,16 @@ export default class StatementParser extends ExpressionParser {
     this.next();
     this.takeDecorators(node);
     this.parseClassId(node, isStatement, optionalId);
+
+    // class bodies and heritages are implicitly strict
+    const oldStrict = this.state.strict;
+    this.state.strict = true;
+
     this.parseClassSuper(node);
     this.parseClassBody(node);
+
+    this.state.strict = oldStrict;
+
     return this.finishNode(
       node,
       isStatement ? "ClassDeclaration" : "ClassExpression",
@@ -1020,9 +1028,6 @@ export default class StatementParser extends ExpressionParser {
   }
 
   parseClassBody(node: N.Class): void {
-    // class bodies are implicitly strict
-    const oldStrict = this.state.strict;
-    this.state.strict = true;
     this.state.classLevel++;
 
     const state = { hadConstructor: false };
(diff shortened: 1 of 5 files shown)


packages/babel-parser/src/parser/statement.js
```diff
@@ -994,8 +994,16 @@ export default class StatementParser extends ExpressionParser {
     this.next();
     this.takeDecorators(node);
```
loganfsmyth: Do you know if class decorators are also technically auto-strict? The spec currently says

> All parts of a ClassDeclaration or a ClassExpression are strict mode code. 

and I have no idea if that is something that is meant to include decorators, or if the decorators proposal will be changing that line to specifically call out the Body and Heritage.

I could see auto-strict decorators being annoying if support for decorating things other than classes were to be considered in the future.

nicolo-ribaudo: The decorators please doesn't say anything about it. @littledan wdyt about it?

nicolo-ribaudo:
I will merge this for now, if you are interested in the strictness of decorators you can watch tc39/proposal-decorators#204.

packages/babel-parser/src/parser/statement.js
nicolo-ribaudo: What of you directly write the function inside the decorator?

ljharb: I misunderstood the question :-) i believe without additional specification, it’s just like any other function - inherits strict from the lexical context or uses the pragma to turn it on.

PR #9316, opened 2019-01-11
Disallow super inside base classes

nicolo-ribaudo:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-parser/src/parser/statement.js
```diff
@@ -1336,23 +1370,35 @@ export default class StatementParser extends ExpressionParser {
...
+    this.state.inDerivedClass = oldInDerivedClass;
   }
 
   pushClassPrivateProperty(
     classBody: N.ClassBody,
     prop: N.ClassPrivateProperty,
+    inDerived: boolean,
   ) {
+    const oldInDerivedClass = this.state.inDerivedClass;
+    this.state.inDerivedClass = inDerived;
```
nicolo-ribaudo: All this things would be a nice usecase for a decorator, if we want to experiment with it. Probably it could even be removed at compile time to generate the current code, so that we avoid the runtime overhead.

```js
class Parser {
  // ...

  @saveState("inDerivedClass")
  pushClassPrivateProperty(
    classBody: N.ClassBody,
    prop: N.ClassPrivateProperty,
    inDerived: boolean,
  ) {
    this.state.inDerivedClass = inDerived;
    this.expectPlugin("classPrivateProperties", prop.key.start);
    classBody.body.push(this.parseClassPrivateProperty(prop));
  }

  // ...
}
```

nicolo-ribaudo:
Replaced by https://github.com/babel/babel/pull/9493

PR #9010, opened 2018-11-10
[WIP]Feature/pattern matching statement

vincentdchan:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  | No
| Minor: New Feature?      | Yes
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

This is an implementation of [tc39/pattern-matching](https://github.com/tc39/proposal-pattern-matching). Next to #7633 , which is an implementation that pattern matching was still an expression. Now, pattern matching becomes a statement.

Example:
```javascript
case (input) {
  when {x} -> console.log('object: ', x);
  when { name: { firstName },  age: x} -> console.log('nested object ', firstName, x);
  when [1, a, b] -> console.log('array: ', a, b);
  when [1, { type: 3 }, b] -> console.log('array', b);
  when undefined -> console.log('everything');
  when y -> console.log('equal to input: ', input, y);
}
```
It's awesome when you want to write something like a reducer in redux:
```javascript
function todoApp (state = initialState, action) {
  case (action) {
    when {type: 'set-visibility-filter', filter: visFilter} ->
      return {...state, visFilter}
    when {type: 'add-todo', text} ->
      return {...state, todos: [...state.todos, {text}]}
    when {type: 'toggle-todo', index} -> {
      return {
        ...state,
        todos: state.todos.map((todo, idx) => idx === index
          ? {...todo, done: !todo.done}
          : todo
        )
      }
    }
    when {} -> {} // ignore unknown actions
  }
}
```

Although it can work now, I am sorry to tell this implementation is currently **incomplete**. Features below are still working in progress:
- [ ] Initializers
- [ ] RestElement of Object Pattern

I will be happy if somebody want to help. Because there is still a large amount of work to do for this pattern matching feature. It's really cool! 💐💐💐

PR #9318, opened 2019-01-11
Pattern matching

gnprice:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is a continuation of @vincentdchan 's #9010 , implementing [tc39/pattern-matching](https://github.com/tc39/proposal-pattern-matching).

The code is still somewhat rough, and the two features @vincentdchan identified as unimplemented as still unimplemented:
* initializers
* object rest spread

But there are a lot more tests, and fixes for some bugs that the new tests exposed.

This is my first time really looking around inside the Babel codebase or any Babel plugins -- so feedback on the implementation, from people who've made other changes inside Babel or Babel plugins before, would be especially welcome.


gnprice:
(Thanks @vincentdchan !)

I spent some more time on this today, and just pushed a new version of the branch.

One change is that I added a bunch more tests, including to some subtle cases like where `case`/`when` statements are nested inside each other; and I converted all the existing tests to `exec.js` form, which I find more expressive and is more robust to changes in implementation details.

The other main change is that I replaced the transform code (none of the parsing, but the part that rewrites `case`/`when` to existing JavaScript) with a fresh implementation, because I had some ideas for another way to structure it. I think this new structure worked pretty well for making the implementation clearer, and closer to doing each job in a single place; and I think it will be a good foundation for adding the remaining unimplemented features.

Also I rebased the branch onto master; and fixed the lint error, which was from a lint rule added since the previous version.

(I see there's a CI failure on Node 6; I'll have to look into that. It's in one of the newly-added tests.)

With this version I think the code that's here is in good shape -- the main things to do are
* Implement initializers, and object rest spread
* There are still more tests that would be good to write -- particularly for testing that things are SyntaxErrors that should be
* Try it out, and iterate on the details of the spec!

As before, feedback welcome -- and particularly on the implementation, from people who've made other changes inside Babel or Babel plugins before.

gnprice:
Also, it'd be neat to play with it in the REPL, but when I follow the link in the CI box:
  https://babeljs.io/repl/build/9777/
I can't find how to enable the newly-added plugin -- it doesn't appear when I search in the "Add Plugin" modal. I'd love to know what I'm missing, either in the REPL's UI or in the code :slightly_smiling_face:

hzoo:
@gnprice thanks for the PR and @vincentdchan for the previous PR! For the repl, the "add plugin" thing is only from npm. We use @babel/standalone to rebuild babel to be bundled in the browser. So to test it, we add the plugin to the "stage x" preset in babel/standalone the package and it will show up in the repl. https://github.com/babel/babel/blob/4944e9e180a4cf329d7191e3ea0dffbab527b36c/packages/babel-standalone/src/index.js#L237 (We removed the stage presets as a package but have it in the REPL still, we will need to move away from it later though)

gnprice:
Thanks @hzoo for the tip about the REPL! I added the new plugin to babel-standalone/preset-stage-1, and it works now :smile: . I also fixed the one test case that was failing in CI.

For anyone interested in playing with the proposed feature, I suggest following the REPL link above (that's to https://babeljs.io/repl/build/9787/ right now) and trying some of the syntax described in the spec proposal:
https://github.com/tc39/proposal-pattern-matching/blob/latest/CORE.md

As I dug today into some further ideas I had for simplifying the implementation of this, and investigated some edge cases of how labels work in JS, I found a bug in those edge cases in `@babel/traverse`. I've just submitted a separate PR #9330 with a fix for that, along with some tests that reproduce it.

gnprice:
> May I ask what's the benefits of using `do ... while(0)` beyond `if ... else ...` (my original version)? It seems that it will cause more overhead.

I think for a v1.0 implementation the priorities are correctness and implementation simplicity -- with the simplicity a key ingredient in helping achieve and maintain correctness.

For that reason I wanted to try putting the assignments and tests as a sequence of statements:

    const _eventType = _val.eventType;
    if (_eventType !== 'new-message') continue;

rather than expressions:

    // if ( ... &&
    (_eventType = _val.eventType,
     eventType === 'new-message') &&
    // ... )

The statements version feels like a better conceptual match for what the patterns are really saying, even though the same behavior can be expressed either way. And the use of `const` makes it simpler to reason about the behavior; it's basically a version of the classic compiler technique of [SSA](https://en.wikipedia.org/wiki/Static_single_assignment_form).

Later, performance will be important. At runtime, I don't expect `do { ...; continue; ... } while (0)` to have any inherent overhead different from `if ( ... && ... )` -- they're both just ways of expressing the same branching control flow, and I'd expect them to turn into more or less the same thing in the JS engine.

For code size, there are some optimizations I've skipped in the current version and which will be good to have -- notably
* checking `typeof _x !== "undefined"` and then `_x === "foo"` can be reduced to just the latter
* binding `const _x = _val.x` and then immediately using `_x` for the first and last time can be reduced to using `_val.x` directly

The previous PR had a version of at least the first of these (maybe the second too, I'd have to go back and look), which means this version emits somewhat larger output. I think both these optimizations won't be fundamentally hard to add -- but they will make the implementation somewhat more complex (as they did in the previous PR), so I think they're best done after a v1.

gnprice:
> For code size, there are some optimizations I've skipped in the current version and which will be good to have

Also, for squeezing out the last few bytes of output size, it might well be helpful to drop the `const`s in favor of a few `var` declarations at the top -- and to switch (back) to `if ( ... && ... )` because `&&` is shorter to spell than `continue`!

I'd rather make that optimization after having a solid complete implementation, though, rather than at this stage.

And I'd also want to do at least some quick measurements of the effect -- in particular, dropping `const` throws away useful information that the JS engine will have to do some work to reconstruct. So it's possible that at least the earlier stages of a JS engine (where it emphasizes compiling quickly, and avoids spending a lot of time analyzing the code) -- which is as far as most code ever gets -- may perform worse on the `var _x; /* ... */ (_x = _val.x, /* ... */` version than the `const _x = _val.x` version.

vincentdchan:
@gnprice Great! The advantage of using `do...while..` is the readability of the generated code. It's a good perspective.

> they're both just ways of expressing the same branching control flow, and I'd expect them to turn into more or less the same thing in the JS engine.

I don't know if JS engine will optimize `while(0)` because it seems that determining not to jump in runtime is the overhead. But even if it's true, the cost is really is small. It's not essential. I am just curious.

nicolo-ribaudo:
For now we can prioritize the working implementation. Speed/code size can come later :+1:

vincentdchan:
Sure

gnprice:
With the commits I pushed today, I believe this implementation is now complete for a first version! All major features of the proposal are implemented, and there are a whole bunch of tests.

I just edited the description to say so, and added details which I guess I'll copy here for future readability of this thread:

---

All major features of the proposal are complete, including the two features @vincentdchan identified in the previous PR as remaining to implement:
* [x] object rest spread
* [x] initializers / default values

There are also a lot more tests, and fixes for some bugs that the new tests exposed.

There are still several places this implementation doesn't match the proposal spec; the ones I know of are marked with `TODO` comments. I believe they're all best handled in follow-up PRs after this one is merged, because at this point they're either
* a corner case which is off the main path;
* a case where I don't understand the spec, best addressed together with discussion (and possible clarification) there; or
* a case where I'm not convinced the implemented behavior isn't better than the spec'd behavior, again best addressed with discussion on the spec.

So I think the overall diff is ready for review.

If you like to read the individual commits when reviewing (as I certainly do!), some are quite clean and readable but others remain messy. I may go and clean the branch up in the next few days.

There are also some rebase conflicts. I'll rebase and resolve those, but in the meantime I think they shouldn't get in the way of code review.

goodmind:
@gnprice why you didn't implement if?

bradennapier:
Any update on when this may be merged in?

igl:
First thing i looked for in the 7.4.0 release notes :( but... now i am sad

antsmartian:
Eagerly, waiting for this update!

alexBrajkovic:
The is no type matching in proposal?

```js
function match_x(input) {
  case (input) {
  	 when Number -> return "number";
         when String -> return "string";
         when {x: 4} -> return x;
	 when _ -> return 'Default';  
}
  return false;
}

console.log(match_x({x: 1})) // number?
```

FireyFly:
@alexBrajkovic see the discussion in the repo for the proposal itself: https://github.com/tc39/proposal-pattern-matching/issues/127

In short, the identifier syntax would conflict with binding a part of an object to a name locally within the `when` branch. `when { x: Number } -> ...` would ambiguously either bind the field `x` under the name `Number` within the block, or perform an `instanceof` (I guess) check against the field's value.

There's also related discussion about matching constructors: https://github.com/tc39/proposal-pattern-matching/issues/117.

alexBrajkovic:
Another issue
```js
function match_x(input) {
  case (input) {
    when undefined -> return 'undefined';
  }
  return false;
}

console.log(match_x({x: 1})) // undefined
```

goodmind:
What's going on with this?

mAAdhaTTah:
@goodmind the proposal itself is still in flux, so the babel transform is probably not going to land until the proposal has settled down.

goodmind:
@mAAdhaTTah I didn't see any changes in proposal repository

Sceat:
pattern matching should have been officially implemented in node years ago, it's really needed :D hope it "settle down"

mAAdhaTTah:
@goodmind Some ideas are being sketched & debated here: https://github.com/tc39/proposal-pattern-matching/pull/140

luxueyan:
How long it can be merged! Can't wait to use it!

wyqydsyq:
The proposal hasn't been updated for 2 months, is there something blocking it from progressing or has it kind of just fallen on the backburner of discussion in TC39? This is one of the final "ergonomic" proposals I've been waiting for and can't wait to use it

jackHedaya:
Any updates?

coodoo:
Is there plan to have this merged any time soon?

mAAdhaTTah:
@coodoo My guess is no, as the proposal itself is still pretty unstable.

DenisSemionov:
This would be so great feature.. Even C# has pattern matching (since v8 if I'm not mistaken) and it makes code easily readable

Sceat:
being notified about this issue always increase my resting heart rate :see_no_evil:

aleksandar-b:
I think this should be merged as is. @hzoo

JLHwung:
Recently the pattern matching proposal has a new champion @ljharb, who is also seeking advancing stage-2. As the proposal does not seems abandoned now, I think we can revive this PR.

Before this PR can be merged, I will submit a new feature request to `estree` on the current AST spec.

ljharb:
@JLHwung i would suggest waiting until the next TC39 presentation; the syntax is likely to change significantly from what's currently on the proposal repo.

ljharb:
@andykais we haven't settled on that yet; there will be a PR to the repo updating it prior to the meeting where it's presented.

mbrowne:
Is there any way to try this out currently? The REPL links are broken, and `yarn install` doesn't work when checking out this branch. I was just curious to try it.

nikolalukovic:
Sooooo, what's the status with this PR? Is it going to get merged any time soon?

FireyFly:
Probably not anytime soon, no, given the [tc39/pattern-matching proposal][1] is still in stage 1. I wouldn't really expect the PR to land before it's hit stage 3, I think.. but for now it makes more sense to follow the proposal progress & development rather than specifically this PR.

[1]: https://github.com/tc39/proposal-pattern-matching

nicolo-ribaudo:
We usually implement new proposals before Stage 3, but as Jordan mentioned there will be major changes so it doesn't make sense to merge this PR yet (and probably it will need to be rewritten, but there are enough people excited about this proposal that it hopefully won't be a problem :grin:)

phyllisstein:
That's terribly exciting. Is there anything those of us in the user/tinkerer community can do to support that work?

coetry:
I think this is one of those additions to JavaScript that will make me happy for maaaaaany years to come. 

After experiencing pattern matching with ML and other statically types functional programming languages, it's really hard for me to go back to regular switch statements!

owattenmaker:
just adding a monthly i can't wait to use case matching! since seeing it in scala a few years ago i've been dying waiting to use something similar in javascript. i can't handle any more switch(true) statements :D

bbgvalayev:
Can anyone recommend library that they are currently using for pattern matching? Something light for client side. What do you think of z for client side?

ermouth:
@bbgvalayev there exists https://github.com/natefaubion/matches.js, which is very old and has moderate performance, however is proved to be reliable over years.

Here is the gist https://gist.github.com/ermouth/f954e71411590b66eb58, which extends above lib to be recursive (ie understand rules with subtrees). Gist has a comment with syntax example.

ljharb:
The pattern matching proposal has been significantly updated, in https://github.com/tc39/proposal-pattern-matching/pull/174. It's highly likely that this PR is thus significantly out of date.

wyqydsyq:
@bbgvalayev I've been using https://github.com/gvergnaud/ts-pattern which works well

fedeci:
Hey @gnprice, do you still intend to work on this or can I use it as a base for a new PR?

Issue #9112, opened 2018-12-02
Support @babel/template Placeholders at the Syntax Level in @babel/parser


tolmasky:
## Feature Request

**Problem**

Currently `@babel/template` allows you to supply a `placeholderPattern` that is applied *after* the successful parse of the template string. What this means is that it is impossible to create a pattern to unambiguously separate placeholders from common identifiers and string literals. This is fine if you are creating a template literal yourself, but makes things difficult if the templating is happening on user input where you have to communicate these restrictions. If you try a pattern like `/^@[a-z]+@$/` the parse will simply fail before it even gets to the replacement phase since `@x@` is not valid JavaScript syntax.

To concisely state the problem and its implications:

1. It is not currently possible to construct placeholder patterns that unambiguously separate placeholders from normal JavaScript identifiers.
2. The default pattern is too loose and too often catches cases not intended to be Placeholders.
3. The current state leads to a "forever maintenance" burden for either maintainers or users as new valid global identifiers are introduced to JavaScript runtimes that can be inadvertently caught by the matcher. The new `URL` constructor is a perfect example of this. While one could safely assume `URL` is a replacement in the past, it is now built-in on the web and on node. This means that something like `const NAME = new URL()` would surprisingly throw an error. The API then becomes immediately harder to use as you have to look through documentation to fix this. You have no option but to employ a very verbose solution for this very simple snippet because you *need* to use the full capital constructor. The options are: 1) provide a whitelist containing URL, 2) come up with a new pattern, or 3) pass in URL for URL. All three significantly increase the noise of the operation for a built-in type.
4. Strings are particularly confusing to deal with in this implementation [as seen here](https://github.com/babel/babel/issues/8067).
5. This often leads to fracturing of templating patterns in the community as people try to solve the above problems by coming up with their own more restrictive `placeholderPattern`s.
6. **It makes it hard (impossible?) to use babel templates for arbitrary user input as opposed to a developer's static templates.**

**Placeholder AST Nodes Solution**

It would be nice to add a new node type,`Placeholder`, to `@babel/parser` with the following shape:

```javascript
interface Placeholder {
    type: "Placeholder";
    name: string;
}
```

[...]

PR #9326, opened 2019-01-13
Refactor import and export parsing

nicolo-ribaudo:

While experimenting with https://github.com/babel/babel/issues/9112, I found out that I needed to hook into some parts of `import`/`export` parsing to disambiguate, for example `export %%FOO%%` and `export %%FOO%% from "path"`.
Since I'm working on #9112 using a plugin in a separate file (like we do for JSX or Flow), I can only replace logic overriding methods. With those monolithic `parseExport`/`parseImport` functions it is impossible without duplicating a lot of logic: for this reason I has to split them into multiple parts. While I was there, I fixed flow typings as noted in a comment before the `parseExport` function.

I'm not putting this changes in the PR I'm preparing for placeholders because they are not strictly related.

### DIFF
--- a/packages/babel-parser/src/parser/statement.js
+++ b/packages/babel-parser/src/parser/statement.js
@@ -1598,9 +1598,7 @@ export default class StatementParser extends ExpressionParser {
 
   parseExportFrom(node: N.ExportNamedDeclaration, expect?: boolean): void {
     if (this.eatContextual("from")) {
-      node.source = this.match(tt.string)
-        ? this.parseExprAtom()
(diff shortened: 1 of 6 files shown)


nicolo-ribaudo:
The actual placeholders (which depend on this PR) are implemented at https://github.com/nicolo-ribaudo/babel/compare/placeholders/refactor-import-export...nicolo-ribaudo:placeholders/1-parser

nicolo-ribaudo:
These commits are also included in #9364, which has already been approved by Daniel.

PR #9328, opened 2019-01-13
Fix handling newline with TypeScript declare and abstract classes

### LINKED ISSUE #9325: @babel/parser(ts): class with preceding keywords on different line
## Bug Report

**Current Behavior**
Babel ignores new lines between class and keyword, typescript does not

https://github.com/Microsoft/TypeScript/blob/master/tests/cases/conformance/classes/classDeclarations/classAbstractKeyword/classAbstractSingleLineDecl.ts

**Input Code**
```ts
abstract
class B {}
```
```ts
declare
class B {}
```

**Expected behavior/code**
in body there should be 2 statements
- ExpressionStatement `abstract` and `declare` and 
- ClassDeclaration

```js
    'typescript',
    'objectRestSpread',
    'decorators-legacy',
    'classProperties',
    'asyncGenerators',
    'dynamicImport',
    'estree',
    'bigInt',
    'importMeta'
  ];
  if (jsx) {
    plugins.push('jsx');
  }

  return babel.parse(text, {
    sourceType: 'unambiguous',
    ranges: true,
    plugins
  });
```

**Environment**
- Babel version(s):  7.2.3
- Node/npm version: 6-11
- OS: ----
- Monorepo: ---

existentialism:
| Fixed Issues?            | Fixes #9325 
| Major: Breaking Change?  | 
| Minor: New Feature?      | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/test/fixtures/typescript/module-namespace/module-new-line/input.js
+++ b/packages/babel-parser/test/fixtures/typescript/module-namespace/module-new-line/input.js
@@ -1,2 +1,3 @@
 module
-Foo = {}
+Foo
+{}
--- a/packages/babel-parser/test/fixtures/typescript/module-namespace/module-new-line/input.js
+++ b/packages/babel-parser/test/fixtures/typescript/module-namespace/module-new-line/input.js
@@ -1,0 +1,2 @@
+module
+Foo = {}
--- a/packages/babel-parser/src/plugins/typescript.js
+++ b/packages/babel-parser/src/plugins/typescript.js
@@ -1186,6 +1186,10 @@ export default (superClass: Class<Parser>): Class<Parser> =>
     }
 
     tsTryParseDeclare(nany: any): ?N.Declaration {
+      if (this.isLineTerminator()) {
+        return;
+      }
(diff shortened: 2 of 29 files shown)


nicolo-ribaudo:
Can you also add the check for type aliases? These should be two statements:

```js
type
Foo = string;
```

existentialism:
@nicolo-ribaudo updated (was going to originally save it for a follow up pr)

packages/babel-parser/test/fixtures/typescript/module-namespace/module-new-line/input.js
```diff
@@ -0,0 +1,2 @@
+module
+Foo = {}
```
nicolo-ribaudo: Since `module Foo = {}` woldn't be valid anyway, maybe the test should be like this?

```js
module
Foo
{}
```

Also for interfaces and namespaces

existentialism: K

PR #9330, opened 2019-01-14
Fix label handling in `scope`, to avoid causing collisions

gnprice:
| Fixed Issues?            | 
| Patch: Bug Fix?          | Yes!
| Major: Breaking Change?  | nope
| Minor: New Feature?      | nope
| Tests Added + Pass?      | :heavy_check_mark: 
| Documentation PR Link    | 
| Any Dependency Changes?  | nope

Labels work quite differently from identifier bindings in how they collide.  Because we track them on scopes in a very similar way to identifier bindings, this means that a transform which innocently tries to insert a label using `scope.generateUid`, like this:

    const label = path.scope.generateUidIdentifier("label");
    path.replaceWith(t.labeledStatement(label, path.node));

is subject to a bug where it can generate invalid code.

Specifically, it's actually perfectly fine to have several labels as siblings:

    function f() {
      loop: do { continue loop; } while (0);
      loop: while (1) { break loop; };
      loop: if (1) { break loop; /* :-P */ }
    }
  
but on the other hand, it's *not* permitted to have one as an ancestor of another:
  
    function f() {
      loop: do {
        loop: while (1) { // Error!
          break loop;     // After all -- what should this mean?
        }
        console.log("would this run?");
      } while (0);
    }

This commit adds several test cases to demonstrate the issue -- an end-to-end test featuring a transform, and a couple of unit tests -- and then changes the algorithm a Scope uses for tracking labels to one that more closely reflects the language semantics, fixing the issue.

### DIFF
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -238,7 +238,7 @@ export default class Scope {
       .replace(/[0-9]+$/g, "");
 
     let uid;
-    let i = 0;
+    let i = 1;
     do {
       uid = this._generateUid(name, i);
       i++;
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -107,11 +107,6 @@ const collectorVisitor = {
     },
   },
 
-  LabeledStatement(path) {
-    path.scope.getProgramParent().addGlobal(path.node);
-    path.scope.getBlockParent().registerDeclaration(path);
-  },
-
   AssignmentExpression(path, state) {
     state.assignments.push(path);
   },
@@ -148,6 +143,43 @@ const collectorVisitor = {
       }
     }
   },
+
+  LabeledStatement: {
+    enter(path, state) {
+      state.allLabels.push(path);
+      state.activeLabels.push(path);
+    },
+
+    exit(path, state) {
+      state.activeLabels.pop();
+    },
+  },
+
+  Scopable: {
+    /*
+     * state.activeLabels: All labels that are ancestors of the
+     *   current node.
+     *
+     * state.allLabels: All labels seen in traversal so far.
+     *
+     * state.labelIndexStack: A stack of indexes into allLabels,
+     *   so that `l === labelIndexStack[i]` means that the i'th
+     *   active scope started after the first l labels were seen.
+     *
+     * This logic is a little over-inclusive: it doesn't take advantage
+     * of the fact that scopes inside a function don't need to worry about
+     * labels outside, and vice versa.
+     */
+    enter(path, state) {
+      state.labelIndexStack.push(state.allLabels.length);
+    },
+
+    exit(path, state) {
+      const start = state.labelIndexStack.pop();
+      path.scope.registerInnerLabels(state.allLabels.slice(start));
+      path.scope.registerOuterLabels(state.activeLabels);
+    },
+  },
 };
 
 let uid = 0;
@@ -173,7 +205,8 @@ export default class Scope {
     this.block = node;
     this.path = path;
 
-    this.labels = new Map();
+    this.innerLabels = new Map();
+    this.outerLabels = new Map();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


packages/babel-traverse/src/scope/index.js
```diff
@@ -238,7 +271,7 @@ export default class Scope {
       .replace(/[0-9]+$/g, "");
 
     let uid;
-    let i = 0;
+    let i = 1;
```
danez: I don't quite get this change. Don't we now start with `_temp1`, although we should start with `_temp`

danez: Oh it is `> 1`, wonder if we should instead make `_generateUid()` do `>=1`. Although that might change all tests.

Guess we should do that separately.

danez:
The change looks okay to me, but I haven't fully understand what the problem was. Can you elaborate which invalid code we were generating?

gnprice:
Thanks for the review!

> The change looks okay to me, but I haven't fully understand what the problem was. Can you elaborate which invalid code we were generating?

Sure. Here's the invalid-syntax example in the commit message on 4e7f31fd8, the commit that adds the tests:
```
  function f() {
    loop: do {
      loop: while (1) { // Error!
        break loop;     // After all -- what should this mean?
      }
      console.log("would this run?");
    } while (0);
  }
```

The rule on label collisions in JavaScript is: you can't reuse a name for one label that's an ancestor (or descendant) of another, unless separated by a `function` or `class` declaration. In [the spec](https://www.ecma-international.org/ecma-262/9.0/index.html), this is expressed through the ContainsDuplicateLabels semantic rule -- I believe the total effect of all the various references to ContainsDuplicateLabels ends up being precisely that. (Though unfortunately if there's somewhere that this rule or others like it are stated more concisely, I haven't found it.)

One story you might tell as a rationale for this is that if the code inside were to then contain a `break loop` or `continue loop` statement, it'd be ambiguous what that statement should do. (And if it doesn't, the labels have no effect at all --  those are the only things that a label in JS can be used for.)

So, the new end-to-end test contains a plugin that innocently tries to introduce a label on a statement:
```
          const label = path.scope.generateUidIdentifier("label");
          path.replaceWith(t.labeledStatement(label, path.node));
          path.skip();
```
together with input code that just happens to already contain a colliding label `_label:` on an outer loop. Because the logic behind `generateUidIdentifier` wasn't seeing that, it would miss the collision and give us `_label` as the name to use for the new label, and the generated code would give a SyntaxError.

Likewise, the new unit test cases cover generating a name to use for a label, where the input code already contains a potentially-colliding label on either a descendant node or an ancestor node:
```
      const gen = (id, path) => path.scope.generateUid(id);
      expect(gen("foo", getPath("do { _foo: { } } while (0)"))).toBe("_foo2");
      expect(
	        gen("foo", getPathHere("_foo: do { HERE } while (0)")),
      ).toBe("_foo2");
```
Before this change:
* `generateUid('foo')` at the root of `do { _foo: { } } while (0)` would produce `_foo` -- so a transform could innocently end up with `_foo: do { _foo: { } } while (0)`;
* `generateUid('foo')` at the point marked `HERE` in `_foo: do { HERE } while (0)` would produce `_foo` -- so a transform could innocently end up with `_foo: do { _foo: { /* ... whatever ... */ } } while (0)`.

Does that help? Happy to put this longer explanation (or parts of it) in comments or commit messages, wherever you think it'd be helpful; or to explain more in another way.

packages/babel-traverse/src/scope/index.js
gnprice: > Oh it is `> 1`, wonder if we should instead make `_generateUid()` do `>=1`. Although that might change all tests.
> 
> Guess we should do that separately.

Yep -- agreed on all points :-)

gnprice: (Here's how I put it [in the commit message](https://github.com/babel/babel/pull/9330/commits/4f2f5d26b863e11c0d58404a3ec9e5ef7e0b20da), a bit more verbosely:

```
We were trying the names

  _temp  _temp  _temp2 _temp3 ...

Skip the first one -- so it's just

  _temp  _temp2 _temp3 ...

(Alternatively, the hyper-optimized way would be like this:

  _temp  _temp0 _temp1 _temp2 ...

That's what C++ compilers do in name-mangling, to make it a little
less common to have to spend the extra character for `_temp10`.
But of course that breaks almost every fragile, output-dependent test
case in the codebase... 182/8641 tests in 32/134 suites, by my count.)
```
)

PR #9333, opened 2019-01-14
Copy "optional" property when cloning Identifier node

### LINKED ISSUE #9331: Optional function params are not marked as optional after going through @babel/template
# v7 Regression

**Describe the regression**
If code passed to `@babel/template` had Flow type annotations and some of the function parameters were marked as optional (`?`) they won't be marked as optional in the produced code. This information gets lost

**Input Code**

```js
const pick = require('lodash.pick');
const template6 = require("babel-template");
const template7 = require("@babel/template").default;

const code = 'function f(foo?: number) {}';

const b6 = template6(code, { plugins: [ 'flow' ] })();
const b7 = template7(code, { plugins: [ 'flow' ] })();

console.log('Babel 6:', pick(b6.params[0], ['name', 'type', 'optional']));
console.log('Babel 7:', pick(b7.params[0], ['name', 'type', 'optional']));

// › Babel 6: {name: "bar", optional: true, type: "Identifier"}
// › Babel 7: {name: "bar", type: "Identifier"}
```

Sorry, couldn't even find the place in `@babel/template` at which point it goes missing. `optional: true` is still in the AST produced by the `@babel/parser`. This rabbit hole is too deep

**Expected behavior/code**
I expect result of executed template to match the code that was passed to it, especially if no replacements has been made

**Environment**
- Babel version(s): **v6.26.0** (behaves correctly), **v7.2.2** (behaves incorrectly)
- Node version: **8.12.0**, **10.15.0**
- How you are using Babel: generating code with it

unconfident:
Fixes #9331

| Fixed Issues?            | Fixes #9331
| Tests Added + Pass?      | None added, should pass
| Documentation PR Link    | 

Copying value of `optional` property of "Identifier" nodes when cloning AST nodes means that clone will be more similar to original node and optional parameters will continue to be optional. Very important when used through babel-template. 
See #9331 to find out how not copying it caused problems

### DIFF
--- a/packages/babel-types/src/clone/cloneNode.js
+++ b/packages/babel-types/src/clone/cloneNode.js
@@ -38,12 +38,14 @@ export default function cloneNode<T: Object>(node: T, deep: boolean = true): T {
   if (type === "Identifier") {
     newNode.name = node.name;
 
+    if (has(node, "optional") && typeof node.optional === "boolean") {
+      newNode.optional = node.optional;
+    }
+
     if (has(node, "typeAnnotation")) {
       newNode.typeAnnotation = deep
         ? cloneIfNodeOrArray(node.typeAnnotation, true)
         : node.typeAnnotation;
-
-      newNode.optional = node.optional;
     }
   } else if (!has(NODE_FIELDS, type)) {
     throw new Error(`Unknown node type: "${type}"`);
--- a/packages/babel-types/src/clone/cloneNode.js
+++ b/packages/babel-types/src/clone/cloneNode.js
@@ -42,6 +42,8 @@ export default function cloneNode<T: Object>(node: T, deep: boolean = true): T {
       newNode.typeAnnotation = deep
         ? cloneIfNodeOrArray(node.typeAnnotation, true)
         : node.typeAnnotation;
+
+      newNode.optional = node.optional;
     }
   } else if (!has(NODE_FIELDS, type)) {
     throw new Error(`Unknown node type: "${type}"`);


packages/babel-types/src/clone/cloneNode.js
```diff
@@ -42,6 +42,8 @@ export default function cloneNode<T: Object>(node: T, deep: boolean = true): T {
       newNode.typeAnnotation = deep
         ? cloneIfNodeOrArray(node.typeAnnotation, true)
         : node.typeAnnotation;
+
+      newNode.optional = node.optional;
```
nicolo-ribaudo: Shouldn't this be in a separate if check?

unconfident: I was under impression that it can only be marked as optional when there's a type annotation present. But you're right, looks like it's not necessary and `function f(foo?) {}` is still perfectly valid.

Will be fixed in a second

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -38,6 +38,10 @@ export default function cloneNode<T: Object>(node: T, deep: boolean = true): T {
   if (type === "Identifier") {
     newNode.name = node.name;
 
+    if (has(node, "optional") && typeof node.optional === "boolean") {
```
unconfident: Probably checking for boolean is not necessary

nicolo-ribaudo: Yeah I was going to leave this exact comment, but then I tought that it would just clean up cases like this:
```json5
{
  type: "Identifier",
  optional: null
}
```
Also, I don't think that it would jave a big runtime overhead given that it won't be looled up on the prototype (because of the "has" check).

PR #9336, opened 2019-01-14
Disallow usage of invalid keyword after export abstract statement in Typescript

### LINKED ISSUE #9304: @babel/parser(ts): exported abstract interface is parsed as a class
## Bug Report

**Current Behavior**
Exported `abstract interface` is parsed by babel/parser as a class, this is only a case when its exported

**Input Code**

```js
export abstract interface I {}
```

```json
{
	"type": "ExportNamedDeclaration",
	"specifiers": [],
	"source": null,
	"declaration": {
		"type": "ClassDeclaration",
		"abstract": true,
		"id": {
			"type": "Identifier",
			"identifierName": "I"
		},
		"name": "I"
	},
	"superClass": null,
	"body": {
		"type": "ClassBody",
		"body": []
	}
}
```

**Expected behavior/code**
I expect error that `abstract` keyword is not allowed. 

```js
    'typescript',
    'objectRestSpread',
    'decorators-legacy',
    'classProperties',
    'asyncGenerators',
    'dynamicImport',
    'estree',
    'bigInt',
    'importMeta'
  ];
  if (jsx) {
    plugins.push('jsx');
  }

  return babel.parse(text, {
    sourceType: 'unambiguous',
    ranges: true,
    plugins
  });
```

- [astexplorer](https://astexplorer.net/#/gist/de2632feeea6ac57b18f0bdbcf43a01e/b3cb4673e4cab889b2a75d77d4fe1ffc6ba5ecc3)
- [babeljs playground](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=KYDwDg9gTgLgBAQwEYGcZQQY3gSwHYzBQBmWwcAknAN4C-QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=true&sourceType=module&lineWrap=true&presets=stage-0&prettier=false&targets=&version=7.2.2&envVersion=1.6.2)

)**Environment**
- Babel version(s):  7.2.3
- Node/npm version: 6-11
- OS: ----
- Monorepo: ---

armano2:
| Fixed Issues?            | Fixes #9304
| Documentation PR Link    |

This PR makes parser error on unexpected keywords used after `export abstract` and disallow usage of unknown keywords as "classes":

example of fixed cases:
```ts
export abstract interface X {}
```
```ts
export abstract putHereWhathever X {}
```

-----------

`this.parseClass` does not check if its actually class, it's just doing `this.next()` on first token
it can be also solved by moving this check there

### DIFF
--- a/packages/babel-parser/src/plugins/typescript.js
+++ b/packages/babel-parser/src/plugins/typescript.js
@@ -1265,7 +1265,12 @@ export default (superClass: Class<Parser>): Class<Parser> =>
           if (next || this.match(tt._class)) {
             const cls: N.ClassDeclaration = node;
             cls.abstract = true;
-            if (next) this.next();
+            if (next) {
+              this.next();
+              if (!this.match(tt._class)) {
+                this.unexpected(null, tt._class);
+              }
+            }
             return this.parseClass(
               cls,
               /* isStatement */ true,
--- a/packages/babel-parser/test/fixtures/typescript/interface/export-abstract-interface/input.js
+++ b/packages/babel-parser/test/fixtures/typescript/interface/export-abstract-interface/input.js
@@ -1,0 +1,3 @@
+export abstract interface I {
+
+}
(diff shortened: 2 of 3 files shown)


nicolo-ribaudo:
Don't the other things inside that switch (enum, interface) have the same problem?

armano2:
no, in other cases we are checking identifier and only `abstract` in multi keyword `abstract class`

PR #9343, opened 2019-01-15
Partial Application Syntax: Stage 1

### LINKED ISSUE #9342: Partial Application syntax: Stage 1
> This proposal introduces a new syntax using the ? token in an argument list which allows you to partially apply an argument list to a call expression by acting as a placeholder for an argument.

## Info
Stage: 1
Champion: Ron Buckton (@rbuckton)
Proposal repo: https://github.com/tc39/proposal-partial-application

## Example
```javascript
const addOne = add(1, ?); // apply from the left
addOne(2); // 3

const addTen = add(?, 10); // apply from the right
addTen(2); // 12

// with pipeline
let newScore = player.score
  |> add(7, ?)
  |> clamp(0, 100, ?); // shallow stack, the pipe to `clamp` is the same frame as the pipe to `add`.

// partial template strings
const Diagnostics = {
  unexpected_token: `Unexpected token: ${?}`,
  name_not_found: `'${?}' not found.`
};
Diagnostics.name_not_found("foo"); // "'foo' not found."
```

## Basic Rules

- can be used one or more at the top level of the Arguments of a `CallExpression`, `CallMemberExpression`, or `SuperCall` (e.g. `f(?)` or `o.f(?)`)
- can be used in place of a `TemplateMiddleList`

## ESTree/Parsing

- I guess we can use the token `question`
- we can create a `Partial` node in `parseExprListItem` function

## Transform

We will add `Partial` node to the AST and it will be added to the spec.

byara:
| Fixed Issues?            | Fixes #9342 
| Documentation PR Link    | 
| Any Dependency Changes?  |

- I modified `parseExprListItem` and added a new condition that checks for `tt.comma` and another `boolean` called `allowPlaceholder`. Then I check for the plugin and add the node `Partial`
- I added a boolean argument `allowPlaceholder` in `parseCallExpressionArguments` and `parseExprListItem` to able to check if using `?` is allowed or not, so I can throw and error.
- I added 9 tests.


packages/babel-generator/src/generators/expressions.js
```diff
@@ -125,6 +125,10 @@ export function OptionalMemberExpression(node: Object) {
   }
 }
 
+export function Partial() {
```
nicolo-ribaudo: This name doesn't match with the name defined in the parser. Also, could you add a generator test?

packages/babel-parser/ast/spec.md
```diff
@@ -603,6 +605,16 @@ Any expression node. Since the left-hand side of an assignment may be any expres
 
 ## Super
 
+```js
+interface PartialExpression <: Node {
```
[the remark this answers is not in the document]
byara: Sure, I was just trying to follow the Grammar section of the proposal: [Grammar](https://github.com/tc39/proposal-partial-application#grammar). But I also think, `ArgumentPlaceholder` shows the purpose better.

packages/babel-parser/src/parser/expression.js
```diff
@@ -807,6 +809,9 @@ export default class ExpressionParser extends LValParser {
         }
         return this.finishNode(node, "Super");
 
+      case tt.question:
+        node = this.startNode();
+        this.raise(node.start, "Partial Application syntax is not allowed");
```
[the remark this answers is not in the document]
byara: I wanted to be able to throw a message whenever the user is using an invalid syntax.
I added a boolean `allowPlaceholder` to `parseExprListItem` and `parseCallExpressionArguments` and thought, I could set it to false whenever we do not want to allow the usage of `?`.
Maybe I could throw the error message with another conditional here:

```javascript
} else if (this.match(tt.question)) {
      if (!allowPlaceholder){
        this.raise(node.start, "Partial Application syntax is not allowed");
      }
      this.expectPlugin("partialApplication");
      this.next();
      elt = this.finishNode(node, "ArgumentPlaceholder");
}
```

packages/babel-generator/src/generators/expressions.js
byara: Sure, I'll add tests for the generator.
Do you mean in: `packages/babel-parser/src/types.js`
I see that I made a mistake there. I'll fix it 👍

byara: Additionally, I suppose, I can move this function to `src/generators/types.js`? Because we are not considering `Partial` (or if we call it `ArgumentPlaceholder`) to be an expression.

nicolo-ribaudo: Yeah good idea (I still prefer ArgumentPlaceholder)

packages/babel-parser/src/parser/expression.js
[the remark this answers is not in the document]
nicolo-ribaudo: The default error would be `unexpected ?`. it's ok either way.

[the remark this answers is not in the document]
danielcaldas: I like the `allowPlaceholder` idea from @byara , it can evolve into something more specific and make the users aware of when a partial application is allowed and when it's not, taking examples from https://github.com/tc39/proposal-partial-application#syntax we can tell the user we can point the user on the right direction instead of having him figure out why he's seeing an `unexpected ?` error.

byara:
- I renamed the node to `ArgumentPlaceholder` and added tests for the `babel-generator`.
- I'm not sure about the type of the node though. In `babel-types/src/validators/generated/index.js` and `babel-types/src/definitions/experimental.js` I have put the node under `UnaryLike` but I'm not sure about it. 
- After I made the change in `babel-types/src/validators/generated/index.js`, adding `ArgumentPlaceholder` under `UnaryLike`,  `make build`, creates [a few functions automatically](https://github.com/babel/babel/pull/9343/commits/65cdb981110d235f78dc021cf82a011326661351), not sure if that is normal.
- for now, I removed the Error message, but if we decide it is a good idea, I'll put it back.
(I'm also working on my own implementation of the plugin for this proposal)

nicolo-ribaudo:
> I renamed the node to ArgumentPlaceholder and added tests for the babel-generator.

:+1:

> I'm not sure about the type of the node though. In babel-types/src/validators/generated/index.js and babel-types/src/definitions/experimental.js I have put the node under UnaryLike but I'm not sure about it.

`?` is not an unary operator, I don't think that there should be any alias.

> After I made the change in babel-types/src/validators/generated/index.js, adding ArgumentPlaceholder under UnaryLike, make build, creates a few functions automatically, not sure if that is normal.

All the files inside `*/generated/*` are automatically created by the build process. You should only edit the files inside `definitions`.

byara:
> ? is not an unary operator, I don't think that there should be any alias.

removed the alias

> All the files inside */generated/* are automatically created by the build process. You should only edit the files inside definitions.

removed my changes from the `generated` folder.

packages/babel-types/src/definitions/experimental.js
```diff
@@ -10,6 +10,15 @@ import {
   classMethodOrDeclareMethodCommon,
 } from "./es2015";
 
+defineType("ArgumentPlaceholder", {
+  visitor: ["argument"],
+  fields: {
+    argument: {
+      validate: assertNodeType("ArgumentPlaceholder"),
```
nicolo-ribaudo: This would mean that the AST is something like this:
```jsonc
{
  "type": "ArgumentPlaceholder",
  "argument": {
    "type": "ArgumentPlaceholder",
    // ...
  }
}
```

It should only be this:
```js
defineType("ArgumentPlaceholder", {});
```

packages/babel-parser/src/parser/expression.js
```diff
@@ -1940,6 +1943,11 @@ export default class ExpressionParser extends LValParser {
       if (refTrailingCommaPos && this.match(tt.comma)) {
         refTrailingCommaPos.start = this.state.start;
       }
+    } else if (allowPlaceholder && this.match(tt.question)) {
```
nicolo-ribaudo: Wdyt about this logic, to have better errors?

```js
} else if (this.match(tt.question)) {
  this.expectPlugin("partialApplication");
  if (allowPlaceholder) {
    this.raise(this.state.start, "Unexpected argument placeholder");
  }
  // ...
```

packages/babel-parser/ast/spec.md
```diff
@@ -36,7 +37,7 @@ These are the core @babel/parser (babylon) AST node types.
     - [DoWhileStatement](#dowhilestatement)
     - [ForStatement](#forstatement)
     - [ForInStatement](#forinstatement)
-    - [ForOfStatement](#forofstatement)
+  - [ForOfStatement](#forofstatement)
```
nicolo-ribaudo: Are these indent changes auto-generated? :thinking:

packages/babel-parser/src/parser/expression.js
byara: I like this! It has a clear message and does not confuse the user. The user will understand that they should not use `ArgumentPlaceholder` there. I'll apply it in my next push.

packages/babel-parser/ast/spec.md
byara: Yeah, I didn't change anything about the list, I just added my change in spec and the list was generated for me 🤔

packages/babel-types/src/definitions/experimental.js
```diff
@@ -10,6 +10,8 @@ import {
   classMethodOrDeclareMethodCommon,
 } from "./es2015";
 
+defineType("ArgumentPlaceholder", {});
```
nicolo-ribaudo: We should also update the `CallExpression` definition to accept placeholder arguments

byara: @nicolo-ribaudo You mean in `/packages/babel-types/src/definitions/core.js` I have to add another assert node type like this:
```
defineType("CallExpression", {
  visitor: ["callee", "arguments", "typeParameters", "typeArguments"],
  builder: ["callee", "arguments"],
  aliases: ["Expression"],
  fields: {
    callee: {
      validate: assertNodeType("Expression"),
    },
    arguments: {
      validate: chain(
        assertValueType("array"),
        assertEach(
          assertNodeType("Expression", "SpreadElement", "JSXNamespacedName", "ArgumentPlaceholder"),
        ),
      ),
    },
...
```

packages/babel-plugin-syntax-partial-application/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@babel/plugin-syntax-partial-application",
+  "version": "7.2.0",
+  "description": "Allow parsing of optional properties",
```
nicolo-ribaudo: This description needs to be updated

byara:
@nicolo-ribaudo I was looking through the changes and I noticed that I have missed a few things:
- forbidding the usage with `SpreadElement `
- allowing the usage in `TemplateLiteral`
What do you think? Should I address them in this PR? or another one?

rbuckton:
Partial application in tagged templates was pretty much rejected in a previous TC39 meeting, so I plan to remove that from the proposal for the time being.

byara:
> > forbidding the usage with SpreadElement
> 
> Do you mean that `foo(?, ...arr)` isn't valid? Or are you talking about `foo(...)`?

Right now, we can parse `foo(?, ...arr)`. But as I was reading [this part](https://github.com/tc39/proposal-partial-application#support-for-restspread-) of the proposal, I had the impression that maybe we shouldn't parse it?

nicolo-ribaudo:
I think that that part is only disallowing the old proposed `foo(...)` syntax (rest placeholder), not `foo(?, ...a)`.

byara:
> I think that that part is only disallowing the old proposed `foo(...)` syntax (rest placeholder), not `foo(?, ...a)`.

I knew it got it wrong! Then I guess everything is alright!
I'm also ready to make a PR for the plugin (It's like 90% done).

nicolo-ribaudo:
Lets just wait for anather review so that this PR can be merged. Can you rebase it in the meantime?

byara:
> Lets just wait for anather review so that this PR can be merged. Can you rebase it in the meantime?

Sure! I'll do that 👍

packages/babel-parser/ast/spec.md
```diff
@@ -862,6 +865,14 @@ interface SpreadElement <: Node {
 }
 ```
 
+## ArgumentPlaceholder
```
danez: on more level please `###`

packages/babel-parser/ast/spec.md
```diff
@@ -72,6 +73,7 @@ These are the core @babel/parser (babylon) AST node types.
     - [LogicalExpression](#logicalexpression)
       - [LogicalOperator](#logicaloperator)
     - [SpreadElement](#spreadelement)
+  - [ArgumentPlaceholder](#argumentplaceholder)
```
danez: should also be intended one more level.

packages/babel-parser/ast/spec.md
danez: I don't think this file is autogenerated, maybe it was one of your ide extensions?
Can you revert this line and the on above in `Changes`?

packages/babel-parser/ast/spec.md
```diff
@@ -601,6 +603,7 @@ interface Expression <: Node { }
 
 Any expression node. Since the left-hand side of an assignment may be any expression in general, an expression can also be a pattern.
 
+
```
danez: not necessary

packages/babel-parser/ast/spec.md
byara: You are right. Apparently, `Markdown All in One` was doing auto formatting.

nicolo-ribaudo:
@byara Thank you! We will merge this PR when we decide to release v7.4.0. In the meantime, could you squash it and rebase the other PR on top of this one?

nicolo-ribaudo:
I'm merging this locally

PR #9348, opened 2019-01-16
Parser Performance Collection

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

So I tried to work on performance in the parser. I did not have any ground breaking success but found some small and minor stuff here and there.

Some of the changes might be debatable. If something feels weird to any of you feel free to mention and I can remove it.

The most impactful change here (depending on the tested source of course)  is the removal of flow code from the regular parser. (Only if not using the flow plugin)

The other big improvement is compiling the parser for node 6.9 or newer only. I was surprised to see it was using all transforms. This might be a problem if someone is using the parser in an older browser without first compiling.

The changes are separated into commits to make it easier to see what belongs together.

I also updated the benchmark tool to work better and more consistently. https://github.com/babel/babylon_performance/commit/1df708ccee211f4a539b36a608b6c6f6df672680


xtuc:
Really nice! Please let us know when it's ready for review.

danez:
I'm done now. Added some more changes, especially the keyword change brings some performance.

babel.config.js
```diff
@@ -14,12 +14,16 @@ module.exports = function(api) {
...
   let ignoreLib = true;
   let includeRuntime = false;
+  const nodeVersion = "6.9";
 
   switch (env) {
     // Configs used during bundling builds.
     case "babel-parser":
       convertESM = false;
       ignoreLib = false;
+      envOpts.targets = {
```
hzoo: It's possible we didn't add this before since Babylon was in another repo and a different config? Or I think it was actually slower before? Or it was because of browser support..

packages/babel-parser/src/parser/base.js
```diff
@@ -16,21 +16,21 @@ export default class BaseParser {
...
 
   hasPlugin(name: string): boolean {
-    return Object.hasOwnProperty.call(this.plugins, name);
+    return this.plugins.has(name);
   }
 
   getPluginOption(plugin: string, name: string) {
-    if (this.hasPlugin(plugin)) return this.plugins[plugin][name];
+    // $FlowIssue
+    if (this.hasPlugin(plugin)) return this.plugins.get(plugin)[name];
```
nicolo-ribaudo: Just out of curiosity, what does flow complain about here?

packages/babel-parser/src/plugins/flow.js
```diff
@@ -1630,7 +1631,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         this.match(tt.name) &&
         (this.state.value === "type" ||
           this.state.value === "interface" ||
-          this.state.value == "opaque")
+          this.state.value === "opaque")
```
nicolo-ribaudo: We should enable the eslint rule to only allow `==` for null/undefined.

packages/babel-parser/src/parser/base.js
danez: It is because of this: https://github.com/facebook/flow/issues/2751 
Flow does not realize that calling `has` ensures that `get` will return something `!=null` 
Other way would be just calling get and testing if `!= null`.

packages/babel-parser/src/plugins/flow.js
danez: 👍

ef4:
The changes in this PR introduced a regression, see https://github.com/babel/babel/issues/9374.

PR #9360, opened 2019-01-18
Update decorators transform

nicolo-ribaudo:
The decorators proposal had a few breaking changes recently: for this reason I had to make them opt-in.

While preparing this PR I had three different goals (in order of importance):
1. No breaking changes. The code must still work if a user updates `@babel/core` but not `@babel/helper-create-class-features`, and vice versa.
2. Users using the new version of the proposal shouldn't have a much bigger helper included because of the previous version: for this reason I used separate mixins (ref: https://github.com/babel/babel/pull/9166)
3. Users using the previous version of the proposal shouldn't have a much bigger helper because of the new version. This is less important than point 2 because they should update to the new proposal anyway.

I chatted with @littledan we decided to introduce a new `version` option, which represents the TC39 meeting where such features have been presented. These are de currently valid configurations:
```json5
{ version: "legacy" } === { legacy: true }
{ version: "nov-2018", decoratorsBeforeExport: true } === { decoratorsBeforeExport: true }
{ version: "nov-2018", decoratorsBeforeExport: false } === { decoratorsBeforeExport: false }
{ version: "jan-2019" }
```

Every other combination of options throw.

Notable changes:
- [x] Property descriptors have been flattened in the element descriptor.
- [x] Method descriptors' value has been remaned to "method"
- [x] Fields initializer->initialize
- [x] Finishers -> replace hook
- [x] Start hook
- [x] Finish hook


nicolo-ribaudo:
You can test this new transform at [https://deploy-preview-1970--babel.netlify.com/repl#?build=10089](https://deploy-preview-1970--babel.netlify.com/repl#?build=10089&code_lz=AIEwpgxg9gUBA2BDAzsgBAMSlNBvAvjDAHZgDum2AFAJQDcRAZgK7EQAuAllMWuNFXDIINPDDR8wwgHRh4YALZhi7ZNIAOzZAAsqucRLQBrTsRAAuNACJt2I1YA0BieqQRFy9patQyxR85oyOyIAE7stHhoiPLhVFYAkuxoZFChRsgAhFai-E4S-PQwhEA&debug=false&evaluate=true&presets=stage-2&version=7.3.2%2Bpr.9360)

nicolo-ribaudo:
This doesn't reflect the decorators proposal anymore.

PR #9364, opened 2019-01-18
Add parser support for placeholders

nicolo-ribaudo:
| Fixed Issues?            | https://github.com/babel/babel/issues/9112
| Patch: Bug Fix?          | 
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

I added support for `%%foo%%` style placeholders to `@babel/parser` throught the `placeholders` plugin.
Part of this PR, which is a small refactor of import/export parsing, is split up at https://github.com/babel/babel/pull/9326 since it can be reviewd by its own.

For now I only allowed the `Placeholder :: %% Identifier %%` syntax, it could be expanded later. `@babel/types`, `@babel/generator` and `@babel/template` support will come in separate PRs.

cc @tolmasky


packages/babel-parser/src/types.js
```diff
@@ -1409,6 +1423,16 @@ export type TsNonNullExpression = NodeBase & {
...
 };
 
+// ================
+// Babel placeholders %%foo%%
+// ================
+
+export type Placeholder<N: PlaceholderTypes> = NodeBase & {
+  type: "Placeholder",
+  id: Identifier,
+  expectedNode: N,
```
nicolo-ribaudo: What do you prefer?
- `expectedNode`
- `expectedType`
- `expected`
- `original`

tolmasky:
I just wanted to ask once again why we aren't using {% %} (I'm totally OK with %% if {% can't be done!), I just know that {%%} gets used for a lot of templating systems and it would be cool to allow it to fit naturally into those. In particular I am currently considering gitbook.io, where it would be nice to be able to use the same templating syntax in both the markdown itself as well as the codeblocks. Again though, I'm open to a different thing just wanted to raise this.

nicolo-ribaudo:
It is mainly because of https://github.com/keithamus/proposal-object-freeze-seal-syntax. Currently it uses `{| ... |}` and `{# ... #}`, but there are two different issues asking to replace `{#` with something else.
If it ends up using `{%`, we will have to change our placeholder syntax.

packages/babel-parser/src/parser/util.js
```diff
@@ -111,6 +111,13 @@ export default class UtilParser extends Tokenizer {
     this.eat(type) || this.unexpected(pos, type);
   }
 
+  // Throws if the current token and the prev one are separated by a space.
+  assertNoSpace(message: string = "Unexpected space."): void {
+    if (this.state.start > this.state.lastTokEnd) {
+      this.raise(this.state.start, message);
```
danez: Should we have the error at the position where the space starts (this.state.lastTokEnd)?

nicolo-ribaudo:
@keithamus Do you think that the `{% %}` syntax could conglict with your proposal?

keithamus:
@nicolo-ribaudo `{% %}` will not conflict with my proposal as it stands, but I think your point around the _potential_ for conflicting changes is true.

packages/babel-parser/src/plugins/placeholders.js
```diff
@@ -0,0 +1,313 @@
...
+          return this.finishNode(node, type);
+        } else {
+          this.unexpected(null, "A class name is required");
+        }
+      } else {
+        this.parseClassId(node, isStatement, optionalId);
+      }
+
+      this.parseClassSuper(node);
+      node.body = this.parsePlaceholder("ClassBody") || this.parseClassBody();
```
danez: I think we need to set `constructorAllowsSuper` here too. 
```suggestion
      node.body = this.parsePlaceholder("ClassBody") || this.parseClassBody(!!node.superClass);
```

packages/babel-parser/src/parser/statement.js
```diff
@@ -1144,7 +1151,7 @@ export default class StatementParser extends ExpressionParser {
     );
   }
 
-  parseClassBody(node: N.Class): void {
+  parseClassBody(constructorAllowsSuper): N.ClassBody {
```
danez:
```suggestion
  parseClassBody(constructorAllowsSuper: boolean): N.ClassBody {
```

packages/babel-parser/src/plugins/placeholders.js
```diff
@@ -0,0 +1,313 @@
...
+        super.parseBlock(...arguments)
+      );
+    }
+
+    parseFunctionId(): ?MaybePlaceholder<"Identifier"> {
+      return (
+        this.parsePlaceholder("Identifier") ||
+        super.parseFunctionId(...arguments)
+      );
+    }
```
danez: Is this function even necessary if we overwrite `parseIdentifier()` anyway ?

nicolo-ribaudo:
Rebased on top of the scope tracking PR

packages/babel-parser/src/plugins/placeholders.js
nicolo-ribaudo: Yes, because `this.match(tt.name)` is false for placeholders and thus it isn't parsed.

danez: Ah true.

danez:
> File name	Previous Size	New Size	Change
> babel-preset-env.js	[2.02 MB](https://10270-24560307-gh.circle-artifacts.com/0/home/circleci/babel/packages/babel-preset-env-standalone/babel-preset-env.js)	[2.02 MB](https://10272-24560307-gh.circle-artifacts.com/0/home/circleci/babel/packages/babel-preset-env-standalone/babel-preset-env.js)	7.39 KB (0%)
> babel-preset-env.min.js	[1.08 MB](https://10270-24560307-gh.circle-artifacts.com/0/home/circleci/babel/packages/babel-preset-env-standalone/babel-preset-env.min.js)	[1.08 MB](https://10272-24560307-gh.circle-artifacts.com/0/home/circleci/babel/packages/babel-preset-env-standalone/babel-preset-env.min.js)	4.48 KB (0%)
> babel.js	[2.74 MB](https://10270-24560307-gh.circle-artifacts.com/0/home/circleci/babel/packages/babel-standalone/babel.js)	[2.75 MB](https://10272-24560307-gh.circle-artifacts.com/0/home/circleci/babel/packages/babel-standalone/babel.js)	7.39 KB (0%)
> babel.min.js	[1.53 MB](https://10270-24560307-gh.circle-artifacts.com/0/home/circleci/babel/packages/babel-standalone/babel.min.js)	[1.54 MB](https://10272-24560307-gh.circle-artifacts.com/0/home/circleci/babel/packages/babel-standalone/babel.min.js)	4.48 KB (0%)

I changed the threshold to 30KB as the threshold is the total of changes for all assets together, so I thought 10KB per bundle + 5KB per minified bundle.

PR #9365, opened 2019-01-18
Update test262 sha

nicolo-ribaudo:
https://github.com/tc39/test262/pull/2026, https://github.com/tc39/test262/pull/2032 and https://github.com/tc39/test262/pull/2038

This commit removes almost 500 tests from the whitelist :tada:


Makefile
```diff
@@ -103,7 +103,7 @@ test-flow-update-whitelist:
 bootstrap-test262:
 	rm -rf ./build/test262
 	mkdir -p ./build
-	git clone --branch=master --single-branch --shallow-since=2018-11-01 https://github.com/tc39/test262.git ./build/test262
```
danez: Whats the reason to change the date? Now it is pretty much a non shallow clone i think

nicolo-ribaudo: I changed it because I thought that it would clone less things :thinking:

Before it would have downloaded every commit since November, now only since January. Unless I misunderstood what shallow-since does.

PR #9371, opened 2019-01-21
Allow toplevel await with option and correctly mark await keyword as unexpected

danez:
| Fixed Issues?            | 
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This has two changes

- Allow top level await in a statement like `for await ()...` when the option `allowAwaitOutsideFunction` is true

- If a `for await` is invalid because there is no expression that can be awaited, mark the await keyword as invalid instead of whatever node triggered the actual error.
For Example: `for await(x=0;;){}` triggers an error after the assignment, now it is the await keyword.
This change aligns babel parser with other parsers who all do it this way.

### DIFF
--- a/packages/babel-parser/test/fixtures/core/opts/allowAwaitOutsideFunction-false/options.json
+++ b/packages/babel-parser/test/fixtures/core/opts/allowAwaitOutsideFunction-false/options.json
@@ -1,0 +1,3 @@
+{
+  "throws": "Unexpected token, expected \"(\" (1:4)"
+}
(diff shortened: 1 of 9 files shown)


nicolo-ribaudo:
Instead of an option maybe it should be a plugin for the https://github.com/tc39/proposal-top-level-await proposal?

packages/babel-parser/test/fixtures/core/opts/allowAwaitOutsideFunction-false/options.json
```diff
@@ -0,0 +1,3 @@
+{
+  "throws": "Unexpected token, expected \"(\" (1:4)"
```
nicolo-ribaudo: This message could be better, something like `Unexpected top-level await`

danez:
I didn't know about this proposal. I will follow up with a PR to make it a plugin and making the option simply enabling the plugin. This would then also change the error messages similar to other plugins.

PR #9375, opened 2019-01-22
Make let a contextual keyword

danez:
| Fixed Issues?            | #6719
| Documentation PR Link    | 
| Any Dependency Changes?  |

Allows `let` being used as identifier in certain cases in non-strict mode. see new test cases

TODO:

- [x] Fix test262 failures


packages/babel-parser/src/parser/statement.js
```diff
@@ -71,6 +76,33 @@ export default class StatementParser extends ExpressionParser {
...
 
+  isLet() {
+    if (!this.isContextual("let")) {
+      return false;
+    }
+    skipWhiteSpace.lastIndex = this.state.pos;
+    const skip = skipWhiteSpace.exec(this.state.input);
+    // $FlowIgnore
+    const next = this.state.pos + skip[0].length;
+    const nextCh = this.state.input.charCodeAt(next);
```
nicolo-ribaudo: We should probably move those lines to a `lookaheadCh` function, since it can be used in many places where we currently use `lookahead()` (and it would be much faster)

packages/babel-parser/src/parser/statement.js
```diff
@@ -86,8 +118,14 @@ export default class StatementParser extends ExpressionParser {
   }
 
   parseStatementContent(declaration: boolean, topLevel: ?boolean): N.Statement {
-    const starttype = this.state.type;
+    let starttype = this.state.type;
     const node = this.startNode();
+    let kind;
+
+    if (this.isLet()) {
```
nicolo-ribaudo: If `declaration` is false and the identifier after `let` is on a new line, we shouldn't enter this `if`:

e.g.
```js
// VariableDeclaration (correct)
while (0) let
[]
```
```js
// ExpressionStatement + BlockStatement (correct)
while (0) let
{}
```
```js
// ExpressionStatement + ExpressionStatement (currently parsed as VariableDeclaration)
while (0) let
foo
```

packages/babel-parser/src/parser/statement.js
existentialism: 👍

packages/babel-parser/src/parser/statement.js
```diff
@@ -1661,7 +1694,7 @@ export default class StatementParser extends ExpressionParser {
     this.parseExportFrom(node, true);
   }
 
-  shouldParseExportDeclaration(): boolean {
```
nicolo-ribaudo: This renaming (which causes the merge conflict) can probably be reverted?

PR #9377, opened 2019-01-22
fix new keyword broken by recent refactoring

### LINKED ISSUE #9374: Syntax Error with yield new
## Bug Report

**Current Behavior**
The code `yield new Foo()` results in a syntax error:

```
SyntaxError: Unexpected token, expected ";" 
```

**Input Code**
```js
export function* test() {
  yield new Foo();
}
```

**Expected behavior/code**
I expect this valid code to parse successfully.

This will fail with no configuration (ie. the default). I've also tried a few different combinations of presets/plugins to get this work but it always fails.

**Environment**
- Babel version(s): v7.2.0
- Node/npm version: Node.js v11.6.0, NPM 6.6.0
- OS: MacOS 10.14.2
- Monorepo: no

danez:
| Fixed Issues?            | Fixes #9374 
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Copy and paste error

### DIFF
--- a/packages/babel-parser/src/tokenizer/types.js
+++ b/packages/babel-parser/src/tokenizer/types.js
@@ -184,7 +184,7 @@ export const keywords = Object.create(null, {
   const: makeKeywordProps("const"),
   while: makeKeywordProps("while", { isLoop }),
   with: makeKeywordProps("with"),
-  new: makeKeywordProps("new"),
+  new: makeKeywordProps("new", { beforeExpr, startsExpr }),
   this: makeKeywordProps("this", { startsExpr }),
   super: makeKeywordProps("super", { startsExpr }),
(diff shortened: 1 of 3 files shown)


packages/babel-parser/src/tokenizer/types.js
```diff
@@ -184,7 +184,7 @@ export const keywords = Object.create(null, {
   const: makeKeywordProps("const"),
   while: makeKeywordProps("while", { isLoop }),
   with: makeKeywordProps("with"),
-  new: makeKeywordProps("new"),
+  new: makeKeywordProps("new", { beforeExpr, startsExpr }),
```
loganfsmyth: I really hate this `beforeExpr` system, it makes it so easy to regress like this :(

PR #9379, opened 2019-01-22
Revert "Differentiate object spread and non-spread properties (#9341)"

### LINKED ISSUE #9376: Cannot convert undefined or null to object
## Bug Report

image

bug version: `@babel/runtime 7.3.0`,  #9334

danez:
This reverts commit 3ae5e79ec8ab1e6d1c6927f35d4a1eeac04faf9a.

| Fixed Issues?            | Fixes #9376
| Major: Breaking Change?  | reverse breaking change
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Not sure what the actual issue is yet, but it seems to throw in some cases.


saschanaz:
It seems it's because some environments doesn't support `Object.getOwnPropertyDescriptors`.

nicolo-ribaudo:
@saschanaz Do you want to open a PR moving these new changes to a new `objectSpread2` helper? That way old code won't be broken by the new helper.

saschanaz:
Yep, I'll do in a minute.

loganfsmyth:
@saschanaz I wouldn't rush, we can review a second pass on this in a future release.

danez:
The problem is for example this (which seems unlikely someone would do that, but there are probably other cases):

```
var a = { ...null, ...undefined };
```

in babel 7.2 this gets generated:
```
var a = _objectSpread({}, null, undefined);
```

in babel 7.3 this gets generated:
```
var a = _objectSpread({}, null, {}, undefined);
```

saschanaz:
I don't think adding `{}` there would cause any problem, though?

danez:
the problem is, that the code generated in 7.2 without `{}` still needs to work with the new helper.
and as we do `i%2` the first `t` will be in the `if(i%2)` clause of the helper the other `t` will be in the `else` branch. and if t is null `Object. getOwnPropertyDescriptors(null)` will throw.

Issue #8749, opened 2018-09-21
Spread operator mismatches native behavior when getter is used


Draccoz:
## Bug Report

**Current Behavior**
When using object spread on an object with a getter, the function is triggered only once, and the result is copied to the output object.

**Input Code**

```js
let counter = 0;

const x = { a: 1 };
const y = { ...x, get b() { counter++; return 2; } };
y.b;
y.b;

console.assert(counter === 2);
```

**Expected behavior/code**
Getter should be created in the output object so it can be triggered on each property access, as it happens with native browser implementation.

**Babel Configuration (.babelrc, package.json, cli command)**
Anything that compiles to ES5.

**Environment**
Any

**Possible Solution**
Use `Object.getOwnPropertyDescriptors(input)` to get a configuration of each input object and in case of getters/setters define them in the output object.

**Additional context/Screenshots**
This bug happens in both Babel and TypeScript, looks to be pretty common.

mathieutu:
The same here:
```ts
{
  ...attributes,
  get fullName() {
    return `${this.firstName} ${this.lastName}`.trim()
  },
}
```

Will transpile to (pseudo code): 
```js
Object.assign({}, attributes, {
   get fullName() {
    return `${this.firstName} ${this.lastName}`.trim()
  }
})
```

so all this.* parameters will be undefined.

Works well in browser with direct spread operators.

mathieutu:
Actually I found: 
https://github.com/babel/babel/issues/7524
https://github.com/babel/babel/pull/7034
https://github.com/babel/babel/issues/9322
and https://babeljs.io/docs/en/next/babel-plugin-proposal-object-rest-spread.html

About this topic.
This should be handled by now, I don't understand why it isn't working on our own.

I'm using a brand new create-react-app... 🤔

nicolo-ribaudo:
[...]

PR #9384, opened 2019-01-22
Retry to fix object spread helper compatibility

### LINKED ISSUE #9322: Object spread still calls getter function after transform
## Bug Report

**Current Behavior**
I still have the same issue as #7524.

**Input Code**

```js
var x = { a: 3 };
var y = { ...x, get y() { return this.a.toString(); } };
console.log(y.y === "3")
```

**Expected behavior/code**
It should log `true`.

```js
{
  "presets": [
    [
      "@babel/preset-env",
      {
        "targets": "last 1 Firefox version, last 1 Chrome version, last 1 Safari version, last 1 Edge version",
        "modules": "amd"
      }
    ]
  ]
}
```

**Environment**
- Babel version(s): 7.2.0
- Node/npm version: 11.6.0
- OS: Windows 10
- Monorepo: no

**Possible Solution**
`Object.getOwnPropertyDescriptors` should be used instead of just copying the values. This way the symbol case should be automatically covered.

The output does use helper function but still is broken.

```js
define([], function () {
  "use strict";

  function _objectSpread(target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i] != null ? arguments[i] : {}; var ownKeys = Object.keys(source); if (typeof Object.getOwnPropertySymbols === 'function') { ownKeys = ownKeys.concat(Object.getOwnPropertySymbols(source).filter(function (sym) { return Object.getOwnPropertyDescriptor(source, sym).enumerable; })); } ownKeys.forEach(function (key) { _defineProperty(target, key, source[key]); }); } return target; }

  var x = {
    a: 3
  };

  var y = _objectSpread({}, x, {
    get y() {
      return this.a.toString();
    }

  });

  console.log(y.y === "3");
});
//# sourceMappingURL=test.js.map
```

saschanaz:
| Fixed Issues?            | Fixes #9322
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Retries #9341.

This adds `helpers._objectSpread2()` where only the even arguments would be considered as spreads.


nicolo-ribaudo:
Can you add a comment that the old helper can be removed in the next major release?

saschanaz:
No idea what's causing failure to rename.

saschanaz:
>Can you add a comment that the old helper can be removed in the next major release?

Inside the helper or above `helpers.objectSpread` line?

saschanaz:
It seems somehow the helper name can't include a trailing number?

nicolo-ribaudo:
Yeah it can. It will be renamed to `_objectSpread` when inlined but it isn't a problem: `@babel/runtime/helpers/objectSpread` will still point to the old helper and  `@babel/runtime/helpers/objectSpread2` to the new one.

saschanaz:
Can this be merged or do this have any more blockers?

saschanaz:
Any news for this?

saschanaz:
@nicolo-ribaudo How about we just treat `null`/`undefined` always as `{}`? That way we can keep the compatibility with the previous behavior without adding `objectSpread2`.

nicolo-ribaudo:
It wouldn't work, because the old helper handles all the arguments the same way.

saschanaz:
The previous issue was that the old code didn’t work with the new helper because the helper assumed that the odd arguments are all objects. Modifying that behavior to also accept null/undefined should fix the specific issue.

________________________________

packages/babel-helpers/src/helpers.js
```diff
@@ -404,6 +405,30 @@ helpers.objectSpread = helper("7.0.0-beta.0")`
   }
 `;
 
+helpers.objectSpread2 = helper("7.0.0-beta.0")`
+  import defineProperty from "defineProperty";
+
+  export default function _objectSpread(target) {
```
danez: Should it be named objectSpread2? I wonder in the case someone uses babel-runtime, this might be a problem, but I don't know exactly how runtime works.

nicolo-ribaudo: Since this is a default export, the name doesn't matter.
Anyway, we should rename it for consistency.

danez: But where is the name in generated output coming from? Because right now it is called `_objectSpread` even with he new helper, which made me think it might collide in the runtime again.

nicolo-ribaudo: It does `generateUidIdentifier("objectSpread2")`, which removes `2` before generating the id.

saschanaz: Would `objectSpreadTwo` is okay or should I fix the function to not remove `2`?

nicolo-ribaudo: It's ok as it is now for me.

saschanaz: You mean this can be ignored?

>Anyway, we should rename it for consistency.

nicolo-ribaudo: We should rename the helper:
```js
export default function _objectSpread2
```
But is is not a problem if that `2` then gets automatically removed by `generateUid`.

nicolo-ribaudo:
Thanks!

mathieutu:
Hi folks. Thanks for you work. 

Just to be sure, do you confirm that this PR will fix the issue https://github.com/babel/babel/issues/8749?

Do you have any idea of its release planning?

Thanks!

cloudkite:
This change breaks IE11 compatibility when using object spread syntax without using polyfills as getOwnPropertyDescriptors is not available in IE11.

nicolo-ribaudo:
Could you please open a new issue? 🙏

QingLeiLi:
I got the error 
> ModuleNotFoundError: Module not found: Error: Can't resolve '@babel/runtime/helpers/objectSpread2'

There must be something wrong in `babel-plugin-proposal-object-rest-spread`.

pwnreza:
WhatsApp Image 2019-07-05 at 13 24 32

Same error as referenced by @edi9999
We are depending on building our RN app via pre-installed set of node_modules.
So we are pretty desperate for a fix 🙏

kimroen:
> I got the error
> 
> > ModuleNotFoundError: Module not found: Error: Can't resolve '@babel/runtime/helpers/objectSpread2'
> 
> There must be something wrong in `babel-plugin-proposal-object-rest-spread`.
> 
> ADD：`objectSpread2 ` was add in the `@babel/runtime-corejs2` package, but the `helperGenerator` method will search the helper in the `@babel/runtime` package only, except set the corejs option to 2.

We're getting a similar error after upgrading from `7.4.3` to `7.5.0`. The error we're getting is this:

"Unknown helper objectSpread2".

QingLeiLi:
> > I got the error
> > > ModuleNotFoundError: Module not found: Error: Can't resolve '@babel/runtime/helpers/objectSpread2'
> > 
> > 
> > There must be something wrong in `babel-plugin-proposal-object-rest-spread`.
> > ADD：`objectSpread2 ` was add in the `@babel/runtime-corejs2` package, but the `helperGenerator` method will search the helper in the `@babel/runtime` package only, except set the corejs option to 2.
> 
> We're getting a similar error after upgrading from `7.4.3` to `7.5.0`. The error we're getting is this:
> 
> "Unknown helper objectSpread2".

Yeah, add `@babel/plugin-transform-runtime` option 
`
{
      corejs: 2,
}
`
make it works again.

Udbhav12:
@QingLeiLi Can you please share your entire package.json file ?

QingLeiLi:
> package.json

package.json ? Did you mean babel config?

```
// babel.config.js
module.exports = {
    return {
        presets: [
            [
                '@babel/preset-env',
                {
                    modules: 'false',
                    targets: {
                        ios: '9',
                        android: '4.4'
                    },
                    useBuiltIns: 'usage'
                }
            ],
            ['@babel/preset-react'],
            ['@babel/preset-typescript']
        ],
        plugins: [
            ['@babel/plugin-transform-runtime', {
            	// point
                "corejs": 2,
            }],
            [
                {
                    "decoratorsBeforeExport": false
                }
            ]
        ],
    };
};
```

pwnreza:
> WhatsApp Image 2019-07-05 at 13 24 32
> 
> Same error as referenced by @edi9999
> We are depending on building our RN app via pre-installed set of node_modules.
> So we are pretty desperate for a fix 🙏

[See here](https://github.com/babel/babel/issues/10166)

kimroen:
Just wanted to mention that the issue I was describing above was fixed by this PR: https://github.com/babel/babel/pull/10170

PR #9400, opened 2019-01-23
Make yield a contextual keyword

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

This makes yield a contextual keyword. This allows yield to be used in more cases as identifier where it is valid.

I also changed the error messages for invalid `yield` and `await` usages to be more clear.

This also refactors where `await` is parsed, so that we do not have to do `this.parseIdentifier(true /* liberal */)` when an `await` could be allowed. This was blocking proper reserved word and keyword checks.


nicolo-ribaudo:
Thank you for all those parser PRs!

PR #9387, opened 2019-01-22
Pipelines as flattened expressions (support for async/await)

thiagoarrais:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Related to babel/proposals#29
| Patch: Bug Fix?          | Yes
| Major: Breaking Change?  | No
| Minor: New Feature?      | No
| Tests Added + Pass?      | Yes
| Documentation PR Link    |
| Any Dependency Changes?  | No
| License                  | MIT

This adds support for async/await in pipeline expressions by flattening them. Await and yield expressions will be transpiled from

```javascript
const r = a |> b |> await # |> c
```
to something like

```javascript
const $0 = a;
const $1 = b($0);
const $2 = await $1;
const r = c($2)
```

Most of the work here was inspired by js-choi's [description of statement flattening](https://gist.github.com/js-choi/446f79ea2b4a1cfac8a8cf594ca269a9) (thanks for that, @js-choi!).

Relates to https://github.com/babel/babel/projects/9#card-15461738

PR #9401, opened 2019-01-23
Support for await and yield in pipelines

thiagoarrais:
| Documentation PR Link    |

This is a follow up to #9387. It adds support for await and yield in the smart pipeline proposal by simply removing the checks put in place by the previous code. The included test cases for await and yield do run correctly. But I believe I am overlooking something. Please help me find what it is.

I'm leaving this as WIP to gather feedback because I honestly don't think something simple as this actually works.

Relates to https://github.com/babel/babel/projects/9#card-15461738

---

EDIT by @nicolo-ribaudo This PR is easier to review without whitespaces diff https://github.com/babel/babel/pull/9401/files?w=1


nicolo-ribaudo:
Wow this PR is too much simple :thinking: :laughing:

thiagoarrais:
And it apparently works! 🤯

nicolo-ribaudo:
The code internally does this (assume an async context):
```js
// input
foo |> await #

// intermediate 1
foo |> (_ => await _)

// optimized output
_tmp = foo, await _tmp;
```

The intermediate representation is invalid, but since it is a normal arrow function with one parameter it is guaranteed to be optimized away.
Since it is a bit hacky but works perfectly, I would prefer to assert that we are always removing that invalid function (it woun't have a big runtime cost): we should save arrow functions generated by the smart plugin in a `WeakSet` and, if `optimizeArrow` is false in the minimal transform but the node is present in the `WeakSet`, throw an error.

Also, please add a comment in the code about this "hack".

packages/babel-plugin-proposal-pipeline-operator/test/fixtures/minimal/async-arrow/exec.js
```diff
@@ -8,6 +8,6 @@ var result = 1
   |> (async (x) => await x + 1)
   |> then((x) => x + 1);
 
-result.then(val => {
```
nicolo-ribaudo: Good catch

nicolo-ribaudo:
Can you add a test for await without the async functions plugin enabled?

thiagoarrais:
Nice. I will add that instead of including the plugin. That was just my solution for fixing the build on Node 6 anyway...

thiagoarrais:
I've tried my hand at the WeakSet solution:

https://github.com/babel/babel/pull/9401/commits/8d6c4a594f5d70e3ab4e18fa8977c6daa114b274

But I don't like the design I came up with. Can anyone suggest any other way to keep state between the minimal and smart visitors?

cc @nicolo-ribaudo

nicolo-ribaudo:
Wouldn't a global `WeakSet` work?

thiagoarrais:
I guess it would but:

1. Where should it go?
2. Should we introduce a global global for that?

nicolo-ribaudo:
You can export it from the `minimalVisitor.js` file, without needing the factory function.
If you don't like it I'm ok with not having this check, since it should never fire.

nicolo-ribaudo:
Ok, this was too simple.
When the arrow functions plugin is enabled, the optimization bails out because the arrow function is transpiled to a normal function. [[repl](https://babeljs.io/repl/build/9918/#?babili=false&browsers=ie%2011&build=&builtIns=false&spec=false&loose=false&code_lz=IwAgPgfCDEDcQ&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=true&timeTravel=false&sourceType=script&lineWrap=true&presets=es2015%2Ces2016%2Ces2017%2Cstage-0&prettier=false&targets=&version=7.2.2%2Bpr.9401&envVersion=7.3.1%2Bpr.9387)]

@thiagoarrais I think that we should keep the two transforms separate and not rely on the intermediate function.

thiagoarrais:
I see... Maybe adding a clause for await/yield in the `updateTopicReferenceVisitor`should suffice. I will try to take a look at it today.

packages/babel-plugin-proposal-pipeline-operator/src/smartVisitor.js
```diff
@@ -1,38 +1,48 @@
...
+          .get("right")
+          .traverse(updateTopicReferenceVisitor, { topicId: placeholder });
 
-    const arrowFunctionExpression = t.arrowFunctionExpression(
-      [topicId],
-      path.node.expression,
-    );
+        call = right.body ? right.body : right.expression;
+      } else {
+        call = t.callExpression(right, [t.cloneNode(placeholder)]);
```
nicolo-ribaudo: Here you can use `right.callee`, so the separate `PipelineBareFunction` (and thus `exit(`) can be removed.

thiagoarrais:
You're right!

The one change I couldn't make was removing `exit` from smart. It is necessary because of `PipelineBareFunction`.

packages/babel-plugin-proposal-pipeline-operator/src/smartVisitor.js
thiagoarrais: Good idea. Will try that.

nicolo-ribaudo:
Sorry for "hating" exit, but it can re-queue it's children ast causing an extra traversal of the sub-tree after that it has already been fully traversed. Also, it increases the indentation of the code :stuck_out_tongue:

thiagoarrais:
No problem. Here you go: e308a1d

packages/babel-plugin-proposal-pipeline-operator/test/fixtures/smart/for-update/output.js
```diff
@@ -0,0 +1,9 @@
+let sum = 0;
+
+for (var i = 0; i <= 10; i = (_i = i, _i + 1)) {
+  var _i;
```
nicolo-ribaudo: I just noticed that this `_i` is in the wrong scope, but since it is a bug with `@babel/traverse` I won't block this PR.
I'll work on it in the meantime.

nicolo-ribaudo: Oh, `var`s are hoisted and `path.scope.push` uses `var`, so even if it is inside the loop body it is ok.

packages/babel-plugin-proposal-pipeline-operator/src/smartVisitor.js
```diff
@@ -1,41 +1,47 @@
...
+    if (t.isPipelineTopicExpression(right)) {
+      path
+        .get("right")
+        .traverse(updateTopicReferenceVisitor, { topicId: placeholder });
 
-    path.replaceWith(arrowFunctionExpression);
-  },
-  PipelineBareFunction(path) {
-    path.replaceWith(path.node.callee);
+      call = right.body ? right.body : right.expression;
```
nicolo-ribaudo: What is this check needed for?

thiagoarrais: Good catch!

It seems like an artifact from the refactor from minimal. Since we're dealing with a topic expression here, `body` will always be undefined. I'll remove the ternary.

PR #9403, opened 2019-01-24
Fix line continuation with Unicode line terminators

danez:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The spec says:

> The String Value of SingleStringCharacter :: LineContinuation is the empty code unit sequence.
> ```
> LineContinuation ::
>     \ LineTerminatorSequence
> LineTerminatorSequence ::
>     <LF>
>     <CR> [lookahead ≠ <LF>]
>     <LS>
>     <PS>
>     <CR><LF>
> ```


danez:
@nicolo-ribaudo We already do that correctly.

PR #9405, opened 2019-01-24
Simplify await and yield tracking in params

danez:
| Fixed Issues?            | 
| Patch: Bug Fix?          |n
| Major: Breaking Change?  |n
| Minor: New Feature?      |n
| Documentation PR Link    | 
| Any Dependency Changes?  |

This simply changes the tracking for yield and await to only have the position in state instead of the whole nodes. It also changes where the params are checked.

### DIFF
--- a/packages/babel-parser/src/parser/util.js
+++ b/packages/babel-parser/src/parser/util.js
@@ -147,4 +147,22 @@ export default class UtilParser extends Tokenizer {
       );
     }
   }
+
+  checkYieldAwaitInDefaultParams() {
+    if (
+      this.state.yieldPos &&
+      (!this.state.awaitPos || this.state.yieldPos < this.state.awaitPos)
+    ) {
+      this.raise(
+        this.state.yieldPos,
+        "Yield cannot be used as name inside a generator function",
+      );
+    }
+    if (this.state.awaitPos) {
+      this.raise(
+        this.state.awaitPos,
+        "Await cannot be used as name inside an async function",
+      );
+    }
+  }
 }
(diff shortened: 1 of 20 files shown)


packages/babel-parser/src/parser/util.js
```diff
@@ -147,4 +147,22 @@ export default class UtilParser extends Tokenizer {
       );
     }
   }
+
+  checkYieldAwaitInDefaultParams() {
+    if (
+      this.state.yieldPos &&
+      (!this.state.awaitPos || this.state.yieldPos < this.state.awaitPos)
```
nicolo-ribaudo: Can't we avoid this check by keeping the two state properties together?

packages/babel-parser/src/parser/util.js
```diff
@@ -147,4 +147,22 @@ export default class UtilParser extends Tokenizer {
...
+    ) {
+      this.raise(
+        this.state.yieldPos,
+        "Yield expression cannot be a default value",
```
nicolo-ribaudo: This error message isn't always accurate:

```js
function* f() {
  (yield) => {}
}
```

```js
function* f() {
  ({ [yield]: x }) => {}
}
```

danez: true will change.

packages/babel-parser/src/parser/util.js
danez: How would you put them together without loosing the information about if it is either await or yield?

nicolo-ribaudo: You are right, I didn't thought about it :+1:

PR #9406, opened 2019-01-24
Fix location/range on TypeScript ExportNamedDeclarations

### LINKED ISSUE #9399: @babel/parse(ts): Declare keyword is not included to range exported node in TypeScript
## Bug Report

**Current Behavior**
Babel does not include range of `declare` keyword to range exported node

exported node contains property `declare=true`

**Input Code**
```ts
export declare const foo = ''
```
```ts
export declare enum foo {}
```
```ts
export declare class foo {}
```

**Expected behavior/code**
Range of exported node should contain `declare` keyword

```js
    'typescript',
    'objectRestSpread',
    'decorators-legacy',
    'classProperties',
    'asyncGenerators',
    'dynamicImport',
    'estree',
    'bigInt',
    'importMeta'
  ];
  if (jsx) {
    plugins.push('jsx');
  }

  return babel.parse(text, {
    sourceType: 'unambiguous',
    ranges: true,
    plugins
  });
```

**Environment**
- Babel version(s):  7.3.0
- Node/npm version: 6-11
- OS: ----
- Monorepo: ---

https://astexplorer.net/#/gist/a748e5502863178b7d0d6dabd74c1900/6c73fa77be54cfbc8566ce5887944a93ea20ed99
issue is present in both 7.3 and 7.2

existentialism:
| Fixed Issues?            | Fixes #9399
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/src/plugins/typescript.js
+++ b/packages/babel-parser/src/plugins/typescript.js
@@ -1860,12 +1860,6 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       // "export declare" is equivalent to just "export".
       const isDeclare = this.eatContextual("declare");
 
-      // Reset location to include `declare` in range
-      if (isDeclare) {
-        this.state.start = startPos;
-        this.state.startLoc = startLoc;
-      }
-
       let declaration: ?N.Declaration;
 
       if (this.match(tt.name)) {
@@ -1876,6 +1870,9 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       }
 
       if (declaration && isDeclare) {
+        // Reset location to include `declare` in range
+        this.resetStartLocation(declaration, startPos, startLoc);
+
         declaration.declare = true;
       }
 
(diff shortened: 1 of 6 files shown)


packages/babel-parser/src/plugins/typescript.js
```diff
@@ -1853,10 +1853,21 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+      const startLoc = this.state.startLoc;
+
       // "export declare" is equivalent to just "export".
       const isDeclare = this.eatContextual("declare");
 
+      // Reset location to include `declare` in range
+      if (isDeclare) {
+        this.state.start = startPos;
+        this.state.startLoc = startLoc;
+      }
```
nicolo-ribaudo: I think that it would be better not to mock `this.state` and set the correct start to the `declaration` node after that it is parsed.

existentialism: Updated helper to support `range` option

PR #9409, opened 2019-01-24
install polyfill & runtime to dependency instead of devDependency [skip ci]

spondbob:
as per doc:
- https://babeljs.io/docs/en/next/babel-polyfill.html#installation
- https://babeljs.io/docs/en/next/babel-runtime.html#usage

| Documentation PR Link    | 

### DIFF
--- a/packages/babel-helper-create-class-features-plugin/README.md
+++ b/packages/babel-helper-create-class-features-plugin/README.md
@@ -1,19 +1,19 @@
-# @babel/plugin-class-features
+# @babel/helper-create-class-features-plugin
 
 > Compile class public and private fields, private methods and decorators to ES6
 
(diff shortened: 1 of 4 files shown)


nicolo-ribaudo:
Oh right, it needs to be special-cased in that script

spondbob:
Thanks for mentioning the script, sorry did not realize one exists!

I'll try to play around with the script see what I can come up with. It's just got a long weekend down here.

spondbob:
Managed to update the script so that it will print `npm install --save package` and `yarn add package` for `@babel/polyfill` and `@babel/runtime`.

Also found `@babel/helper-create-class-features-plugin` was written as `@babel/plugin-class-features` think that might be a typo?

nicolo-ribaudo:
> Also found @babel/helper-create-class-features-plugin was written as @babel/plugin-class-features think that might be a typo?

Yeah, it was the old name.

existentialism:
@spondbob thanks!

PR #9412, opened 2019-01-26
Destructuring: Fix array unpacking assignments with holes on RHS

motiz88:
| Fixed Issues?            | described in PR
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This fixes an issue where destructuring assignments eligible for the "array unpacking" optimization would fail to compile when the array literal on the right-hand side of the expression contained holes.

Example input:
```js
[a, b] = [, 2];
; // Avoid completion record special case
```

The error message was `Property right of AssignmentExpression expected node to be of a type ["Expression"] but instead got null`.

Now the above code compiles to:
```js
a = void 0;
b = 2;
;
```

This PR also adds a couple of related test cases that were missing, to ensure the change doesn't regress them:
* Normal assignment expression with unpacking
* Declaration with unpacking and a hole on RHS

### DIFF
--- a/packages/babel-plugin-transform-destructuring/test/fixtures/destructuring/array-unpack-optimisation/input.js
+++ b/packages/babel-plugin-transform-destructuring/test/fixtures/destructuring/array-unpack-optimisation/input.js
@@ -11,3 +11,7 @@ var [a, b] = [...foo, bar];
 var [a, b] = [foo(), bar];
 var [a, b] = [clazz.foo(), bar];
 var [a, b] = [clazz.foo, bar];
+var [a, b] = [, 2];
+[a, b] = [1, 2];
+[a, b] = [, 2];
+; // Avoid completion record special case
--- a/packages/babel-plugin-transform-destructuring/src/index.js
+++ b/packages/babel-plugin-transform-destructuring/src/index.js
@@ -83,7 +83,11 @@ export default declare((api, options) => {
 
       if (op) {
         node = t.expressionStatement(
-          t.assignmentExpression(op, id, t.cloneNode(init)),
+          t.assignmentExpression(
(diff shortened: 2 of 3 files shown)


packages/babel-plugin-transform-destructuring/test/fixtures/destructuring/array-unpack-optimisation/input.js
```diff
@@ -11,3 +11,7 @@ var [a, b] = [...foo, bar];
 var [a, b] = [foo(), bar];
 var [a, b] = [clazz.foo(), bar];
 var [a, b] = [clazz.foo, bar];
+var [a, b] = [, 2];
+[a, b] = [1, 2];
+[a, b] = [, 2];
+; // Avoid completion record special case
```
nicolo-ribaudo: Not in this PR, but this optimization is wrong: `;` keeps the previous completion value:
```js
eval("0;;;;")
```

Can you use something like `0;` instead?

motiz88: Yeah, I've just realised this on another PR. Ultimately it's a false negative in the `isCompletionRecord` check, which could be quite expensive to correct without memoising somehow.

I'll change this, but FWIW there are several other tests that currently use `;` in this way and would need to change similarly.

nicolo-ribaudo: This can/should be handled in another PR

nicolo-ribaudo:
Thank you!

PR #9414, opened 2019-01-26
Destructuring: Create assignments from ForX non-declaration patterns

motiz88:
| Fixed Issues?            | described in PR
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Fixes an issue where a `for...of` or `for...in` pattern binding that _wasn't_ a declaration would be compiled in much the same way as a declaration, causing new variables to be declared in the loop body instead of assigning to existing ones. This could lead to incorrect output from the `destructuring` plugin:

```js
// Input
let x;
for ([x] of array) {
}

// Output
let x;

for (var _ref of array) {
  var _ref2 = _slicedToArray(_ref, 1),
    x = _ref2[0]; // Attempts to redeclare x, which is an error
}
```

In common usage as part of a preset, the `block-scoping` plugin would then change `let x` to `var x` in-place, and things would happen to work; but that's still not ideal.

This PR fixes the issue by emitting assignments rather than declarations in this case.

```js
let x;

for (var _ref of array) {
  var _ref2 = _slicedToArray(_ref, 1);

  x = _ref2[0];
  void 0;
}
```

The extra `void 0` serves to hide the injected assignment statement's completion value, which could incorrectly become the loop's completion value if the body is empty. Edge cases! 👯  

The `object-rest-spread` plugin, which has an older version of the same code copied from `destructuring`, is similarly updated.


packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/for-x-completion-record/output.js
```diff
@@ -0,0 +1,9 @@
+for (var _ref of []) {
+  var _ref2 = _ref;
```
nicolo-ribaudo: Is it possible to avoid this double temp var? (Not a big deal)

motiz88: I suppose we could add the `isCompletionRecord` check from `destructuring` to inhibit repeating the expression and thus lose the second temp var in the general case. The problem is that's actually an unsound check for what it tries to achieve (which came up on another PR just now).

nicolo-ribaudo: It's ok like this then

PR #9415, opened 2019-01-26
@babel/traverse: Fix NodePath.getData

71:
| Fixed Issues?            | None.
| Patch: Bug Fix?          | Yup.
| Major: Breaking Change?  | Nope.
| Minor: New Feature?      | Nope.
| Tests Added + Pass?      | Nope.
| Documentation PR Link    | N/A.
| Any Dependency Changes?  | Nope.

Currently, if the obtained value is `false`, it will be replaced by the given default value, which is invalid. This makes sure that we only set the default value when the value is `undefined`, instead of falsy.

### DIFF
--- a/packages/babel-traverse/src/path/index.js
+++ b/packages/babel-traverse/src/path/index.js
@@ -116,7 +116,7 @@ export default class NodePath {
 
   getData(key: string, def?: any): any {
     let val = this.data[key];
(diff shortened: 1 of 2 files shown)


danez:
I added some tests and also ensured that the object protoype is not interfering.

danez:
The error is unrelated to this PR.

PR #9431, opened 2019-01-30
Typescript function destructured params

### LINKED ISSUE #9430: [TypeScript] fails when destructure an array param on function
## Bug Report

**Current Behavior**
Code fails with Babel, works with TSC.

**Input Code**
- [REPL](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=C4TwDgpgBAsiAq5oF4oAoDaBDANFARgLoBcUGAzsAE4CWAdgOZ50CuAtvhFYQJRTIA-KJVqMAUGID0kqKEiwESfugDeWAL6k1pEfQbq-g4dT0SAxgHs6lKAA9ScRPNQr1QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Ctypescript&prettier=false&targets=&version=7.2.0&envVersion=7.2.0) (ps: needs to install the @babel/plugin-transform-typescript plugin)
- [TypeScript Playground](http://www.typescriptlang.org/play/#src=type%20MyType%20%3D%20(%7B%20theme%20%7D%3A%20%7B%20theme%3A%20string%20%7D)%20%3D%3E%20string%0D%0A%0D%0Aconst%20x%3A%20MyType%20%3D%20%7B%7D)

This code fails on Babel but not on TSC:
```typescript
type MyType = ([a, b]: [string, number]) => string

const x: MyType = {}
```

**Expected behavior/code**
Support this syntax, because tsc supports it.

```js
{
}
```

**Environment**
- Babel version(s): 
    - "@babel/cli": "^7.2.3",
    - "@babel/core": "^7.2.2",
    - "@babel/preset-typescript": "^7.1.0"
- OS: OSX 10.13.4
- Monorepo: no

**Possible Solution**
see PR

**Related Issue**
[#8745](https://github.com/babel/babel/issues/8745)

mhcgrq:
| Fixed Issues?            | Fixes #9430
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-parser/src/plugins/typescript.js
```diff
@@ -794,6 +795,21 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         return true;
       }
 
+      if (this.match(tt.bracketL)) {
```
nicolo-ribaudo: The two checks should be merged, since there could be code like this:
```js
type Foo = ([{a}]: any) => {};
```

mhcgrq: Currently it just works for `[a]` `{a}` and `[{a}]`. Is that neccesary to check brace and bracket within a same if block? Since there is only brace and bracket, no more conditions.

PR #9434, opened 2019-01-30
Use a GitHub Action to generate the changelog

nicolo-ribaudo:

This is the first step towards publishing from GitHub actions, so that it will take 10 mins instead of one hour of the publisher's time.

The workflow I implemented in this PR is far from complete: it will automatically create the release at https://github.com/babel/babel/releases, but it won't publish anything to npm yet (it still needs to be done locally). Before adding publishing support, I want to check that this works: a wrong package on npm is far worst than a wrong GitHub release :wink:

I have done some testing at https://github.com/babel/test-github-actions (private repo, sorry) and it looks like this approach works. There is one big caveat: this action will be marked as "skipped" for every PR. It is correct, but it adds a bit of visual noise. When public repos will support actions other than "push", those three blocks will probably disappear since I will trigger them only on release.
I have contacted GitHub to understand if it is possible to hide it.
schermata da 2019-01-30 14-16-47

You can see the visual workflow at https://github.com/babel/babel/blob/d3032e53522505c490f0d0e64acdc673c4191a78/.github/main.workflow

### DIFF
--- a/.github/actions/trigger-github-release/release.js
+++ b/.github/actions/trigger-github-release/release.js
@@ -1,0 +1,25 @@
+"use strict";
+
+const [ /* node */, /* file */, tag ] = process.argv;
+
+const getStdin = require("get-stdin");
+const octokit = require("@octokit/rest")();
+
+octokit.authenticate({
+  type: "token",
+  token: process.env.GITHUB_TOKEN
+});
+
+const [ repoOwner, repoName ] = process.env.GITHUB_REPOSITORY.split("/");
+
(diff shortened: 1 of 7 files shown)


.github/actions/trigger-github-release/release.js
```diff
@@ -0,0 +1,25 @@
+"use strict";
+
+const [ /* node */, /* file */, tag ] = process.argv;
```
xtuc: oh nice trick

PR #9440, opened 2019-02-01
Remove charcodes from package as it is already in root

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/package.json
+++ b/packages/babel-parser/package.json
@@ -29,7 +29,6 @@
   },
   "devDependencies": {
     "@babel/helper-fixtures": "^7.2.0",
-    "charcodes": "0.1.0",
     "unicode-11.0.0": "^0.7.8"
   },
   "bin": {


danez:
I guess flow just simply does the same that node does and traverses up the hierarchy until it finds the package.

loganfsmyth:
Perhaps it would be better to move that plugin into `babel-parser` itself and load it via a `.babelrc`?

nicolo-ribaudo:
> Perhaps it would be better to move that plugin into babel-parser itself and load it via a .babelrc?

I find it easier to have all the config in one file. Is it such a big deal if this package is added both at the top-level and in `@babel/parser`? It is a devDep anyway, and probably it will be hoisted by lerna.

danez:
The main problem what I did this was that the version in the root package.json was updated, but not the one in the package. So we were installing two different versions.

nicolo-ribaudo:
Oh ok. I'm ok with both with removing the top-level dep or the one in babel/parser then.

loganfsmyth:
I think, since the parser references it, it should be in the parser. I don't really care where the plugin lives. If it's in the root, I'd code to load `charcodes` from the parser package maybe by doing `require("./packages/babel-parser/node_modules/charcodes")` or using the npm `resolve` package with `basedir: "packages/babel-parser"`.

Issue #1286, opened 2015-04-17
IE8 raises `Expected identifier` exception on `browser-polyfill.js`


pkese:
`babel-core/browser-polyfill.js` on a fresh npm install contains `iterator.throw` and `iterator.return` (rather than `["throw"]` and `["return"]`) which consequently makes it incompatible with older browsers.

The actual error was fixed in regenerator a few days ago, but the npm version of babel still installs the old syntax
https://github.com/facebook/regenerator/commit/97466d3d540243702496b24426765f0022f4bc07

To reproduce, I just freshly npm installed babel-core and searched for `.throw` in resulting browser-polyfill.js

sebmck:
It's completely unrelated to the regenerator issue. The problem is the uglify settings that are used to minify `browser-polyfill`.

zloirock:
@sebmck related. IIRC, uglify not transform `iterator["throw"]` -> `iterator.throw`, looks like simply need rebuild `browser-polyfill` with latest `regenerator`.

pkese:
Beats me...

If I run `make build` on a current version of babel, the `dist/polyfill.min.js` looks perfectly fine without any `.return` or `.throw`.

@sebmck I think @zloirock was right -- the make process just doesn't rebuild the `browser-polyfill`.

sebmck:
Should be fixed as of 5.1.12.

iam4x:
Thank's @sebmck !

alexlafroscia:
I'm still seeing this issue with the `browser-polyfill.js` file from `babel-core` v. `5.4.7`

sebmck:
Can you confirm this happens in latest?

On Friday, 5 June 2015, Alex LaFroscia <notifications@github.com> wrote:

> I'm still seeing this issue with the browser-polyfill.js file from
> babel-core v. 5.4.7
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/babel/babel/issues/1286#issuecomment-109381343>.
>


-- 
Sebastian McKenzie

alexlafroscia:
@sebmck I just pulled the latest `babel-core` release and imported the `browser-polyfill.js` file.  I'm seeing an `Expected Identifier` error coming from

> line 2880 character 39

which translates to this location in the source

```javascript
// Avoid propagating enqueueResult failures to Promises returned by
// later invocations of the iterator, and call generator.return() to
// allow the generator a chance to clean up.
previousPromise = enqueueResult.catch(invokeReturn);
```

[...]

Issue #9232, opened 2018-12-23
when @babel/polyfill is used with webpack 5 it throws a ReferenceError: global is not defined


capaj:
## Bug Report

**Current Behavior**
on startup, I get the ReferenceError

```
index.js:25 Uncaught ReferenceError: global is not defined
    at Object.../node_modules/@babel/polyfill/lib/index.js (index.js:25)
    at __webpack_require__ (bootstrap:22)
    at fn (hot module replacement:64)
    at Module../src/index/globals-import.ts (error-handler.js:37)
    at __webpack_require__ (bootstrap:22)
    at fn (hot module replacement:64)
    at Module../src/index/index.js (globals-import.ts:15)
    at __webpack_require__ (bootstrap:22)
    at bootstrap:47
    at bootstrap:47
```

**Expected behavior/code**
no ReferenceError is thrown

**Environment**
- Babel version(s): "@babel/polyfill": "^7.2.5"
- Node/npm version: Node 8/npm 6.4.1
- OS: ubuntu
- Monorepo: `no`
- How you are using Babel: [`loader`]

I understand that I probably should just provide a `global` and that could mitigate the startup issue, but I also wanted to ask why?
Isn't babel polyfill primarily targeted at old browsers? Why is it that it relies on `global` to bootstrap itself when it's obviously not going to be there? Unless using webpack 1-4 obviously....

nicolo-ribaudo:
I think that this has something to do with the fact that webpack 5 doesn't polyfill any node feature (i.e. `global`) by default anymore.

andersk:
The [documentation recommends](https://babeljs.io/docs/en/babel-polyfill#usage-in-node-browserify-webpack) using `@babel/preset-env` with the [`useBuiltIns`](https://babeljs.io/docs/en/babel-preset-env#usebuiltins) option set to `'entry'` or `'usage'`. In that case the top-level `@babel/polyfill` module isn’t used and this bug doesn’t appear.

We should still fix `@babel/polyfill`, but at least the `useBuiltIns` workaround is something you might want to be doing anyway.

Issue #9205, opened 2018-12-18
preset-env's `useBuiltIns: usage` doesn't work properly with current `regenerator-runtime` version


Hypnosphi:
## Bug Report

**Current Behavior**
`useBuiltIns: usage` output throws an error when using latest `regenerator-runtime` release.

```
$ yarn start
yarn run v1.12.3
$ babel index.js -d dist && node dist/index.js
Successfully compiled 1 file with Babel.
~/usebuiltins-usage-repro/dist/index.js:7
regeneratorRuntime.mark(foo);
^

ReferenceError: regeneratorRuntime is not defined
    at Object.<anonymous> (~/usebuiltins-usage-repro/dist/index.js:7:1)
    at Module._compile (module.js:653:30)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Function.Module.runMain (module.js:694:10)
    at startup (bootstrap_node.js:204:16)
    at bootstrap_node.js:625:3
error Command failed with exit code 1.

```

**Input Code**
https://github.com/Hypnosphi/usebuiltins-usage-repro

**Expected behavior/code**
No errors

**Babel Configuration (.babelrc, package.json, cli command)**

https://github.com/Hypnosphi/usebuiltins-usage-repro/blob/master/package.json

**Possible Solution**

https://github.com/babel/babel/issues/9044#issuecomment-444143528

Hypnosphi:
Does not reproduce with 7.4

PR #9442, opened 2019-02-01
Update regenerator-runtime to latest version

MattiasBuelens:
| Fixed Issues?            | facebook/regenerator#336 facebook/regenerator#346 facebook/regenerator#353 facebook/regenerator#355 #1286 
| Major: Breaking Change?  | Possibly, but probably not (facebook/regenerator#353)
| Tests Added + Pass?      | Existing tests pass
| Any Dependency Changes?  | `regenerator-runtime` (minor)

Update `regenerator-runtime` from version `0.12.0` to the latest version `0.13.1`.

facebook/regenerator#353 changes how the runtime defines the `regeneratorRuntime` global variable (when loaded as a script) or its module exports (when loaded as a CommonJS module). This obsoletes a previous attempt (facebook/regenerator#346) at fixing a CSP violation (facebook/regenerator#336). This change is a minor bump for `regenerator-runtime` (because it no longer has a separate `runtime-module.js`), but I don't think it will affect Babel users.

facebook/regenerator#355 fixes a regression where IE8 would throw a parse error when loading the runtime (#1286).

### DIFF
--- a/packages/babel-polyfill/package.json
+++ b/packages/babel-polyfill/package.json
@@ -12,6 +12,6 @@
   "main": "lib/index.js",
   "dependencies": {
     "core-js": "^2.5.7",
-    "regenerator-runtime": "^0.12.0"
+    "regenerator-runtime": "^0.13.2"
   }
 }
--- a/packages/babel-runtime-corejs2/package.json
+++ b/packages/babel-runtime-corejs2/package.json
@@ -10,7 +10,7 @@
   "author": "Sebastian McKenzie <sebmck@gmail.com>",
   "dependencies": {
     "core-js": "^2.5.7",
-    "regenerator-runtime": "^0.12.0"
+    "regenerator-runtime": "^0.13.2"
   },
   "devDependencies": {
     "@babel/helpers": "^7.3.1"
--- a/packages/babel-runtime/package.json
+++ b/packages/babel-runtime/package.json
@@ -9,7 +9,7 @@
   "repository": "https://github.com/babel/babel/tree/master/packages/babel-runtime",
   "author": "Sebastian McKenzie <sebmck@gmail.com>",
   "dependencies": {
-    "regenerator-runtime": "^0.12.0"
+    "regenerator-runtime": "^0.13.2"
   },
   "devDependencies": {
     "@babel/helpers": "^7.3.1"


wmertens:
Unless something around this code changed since November, I can say that it will break the build, because Babel relies on the polyfill being a global, IIRC.

How I found out: I use `pnpm` which enforces strict package dependencies, and this caused 0.13 to be exposed to my babel builds which work with 0.12. This broke the build.

I may be wrong, but that's basically what happened: I used 0.13 with Babel 7.something, and things broke.

MattiasBuelens:
You're right, the tests are failing. It looks like my local `yarn install` didn't pick up the dependency update before I ran the tests.

I guess `@babel/polyfill` should define the global then? 😕

MattiasBuelens:
Hmm, it looks like `@babel/preset-env` also needs to be updated to generate:
```js
var regeneratorRuntime = require('regenerator-runtime/runtime');
```
instead of just:
```js
require('regenerator-runtime/runtime')
```
as reported in #9205.

@benjamn @loganfsmyth Is it okay if I continue with this? Or do you think I should use a different approach?

danez:
Sorry for the long silence.

I love packages without Changelog ❤️. The changes look good to me, but yes we need to update `preset-env` and if I'm not mistaken we also need to update the generation in babel-runtime (https://github.com/babel/babel/blob/fdc869ce16d47ab5cc5f14d6174ee8f370c2ef85/packages/babel-plugin-transform-runtime/src/index.js#L203). Would you be willing to do this?

zloirock:
`global` is not available in browsers, `globalThis` still is a proposal and not available even with polyfills of stable features.

This way of `regenerator-runtime` import will cause a serious problem for `preset-env`. #7646 mean separate inclusion of `core-js` and `regenerator-runtime` and deprecation of `@babel/polyfill`. `preset-env` works just with `import 'regenerator-runtime/runtime'`.

MattiasBuelens:
> The changes look good to me, but yes we need to update `preset-env` ~and if I'm not mistaken we also need to update the generation in babel-runtime~. Would you be willing to do this?

Sure, I could help with that. I'm just not entirely certain on what approach we'd want to take, see below.

> Also I wonder what we should do in polyfill if the regenerator runtime does already exists on the global object? The old version of the regenerator-runtime was not overwriting and silently not doing anything if imported a second time. Should we do the same or trigger a warning?

Good point. I'd say we should not overwrite the global if it already exists, so you can load a modified version of `regenerator-runtime` if necessary. I don't know if we need to warn about that.

> `global` is not available in browsers, `globalThis` still is a proposal and not available even with polyfills of stable features.

Is this about [this line](https://github.com/babel/babel/blob/ebe3ff73e87ae9ab93aa0326a53448018a0a296d/packages/babel-polyfill/src/index.js#L28)?
```js
global.regeneratorRuntime = regeneratorRuntime;
```
I'm pretty sure that CommonJS `global` gets transpiled to something browser-friendly like:
```js
typeof global !== "undefined" ? global : typeof self !== "undefined" ? self : typeof window !== "undefined" ? window : {}
```
Otherwise, `global._babelPolyfill` wouldn't work either.

> This way of `regenerator-runtime` import will cause a serious problem for `preset-env`. #7646 mean separate inclusion of `core-js` and `regenerator-runtime` and deprecation of `@babel/polyfill`. `preset-env` works just with `import 'regenerator-runtime/runtime'`.

I hadn't seen your pull request yet, looks very interesting! But indeed, with this new version of `regenerator-runtime`, `preset-env` can no longer rely on it to define `regeneratorRuntime` globally if it's used as an `import`.

Do we need a wrapper around `regenerator-runtime` that ensures `regeneratorRuntime` is always globally defined, and import that from `polyfill` and `preset-env` instead? 🤔 

~For `useBuiltIns: 'usage'`, `preset-env` could insert `import regeneratorRuntime from 'regenerator-runtime/runtime'` instead so we don't need to define a global.~ Never mind, we have `transform-runtime` for that.

Alternatively, we update `regenerator-runtime` to always define the global, as it was before... 😕

zloirock:
> I'm pretty sure that CommonJS global gets transpiled to something browser-friendly like:

See [this issue](https://github.com/babel/babel/issues/9232).

> Do we need a wrapper around `regenerator-runtime` that ensures `regeneratorRuntime` is always globally defined, and import that from `polyfill` and `preset-env` instead?

Since now we haven't standard global object reference, without entry point which defines `regenerator` runtime globally, `preset-env` will have serious problems.

`usage` mode of `preset-env` should add global `regeneratorRuntime`. We can imports `globalThis` from `core-js` and use it. It's a complication of `preset-env`, but not something critically hard.

But what should do `entries` mode? It should remove `regenerator-runtime/runtime` import if no one plugin which uses `regenerator` transforms used in the target environment. It's not a problem to remove
```js
import `regenerator-runtime/runtime`;
```
but how we can detect that it should remove
```js
import regeneratorRuntime from `regenerator-runtime`;
Function('return this')().regeneratorRuntime = regeneratorRuntime;
```
?

Moreover, since we deprecate `@babel/polyfill`, we can without any problems say users instead of `@babel/polyfill` use
```js
import `core-js/stable`;
import `regenerator-runtime/runtime`;
```
but not
```js
import `core-js/stable`;
import `core-js/features/global-this`;
import regeneratorRuntime from `regenerator-runtime`;
globalThis.regeneratorRuntime = regeneratorRuntime;
```

benjamn:
Can we hold this PR until https://github.com/facebook/regenerator/pull/369 is released?

MattiasBuelens:
@benjamn Sure! I think whatever solution we come up with for Babel would involve defining `regeneratorRuntime` globally anyway. So it'd be better if `regenerator-runtime` did that for us. 🙂

packages/babel-polyfill/package.json
```diff
@@ -12,6 +12,6 @@
   "main": "lib/index.js",
   "dependencies": {
     "core-js": "^2.5.7",
-    "regenerator-runtime": "^0.12.0"
+    "regenerator-runtime": "^0.13.1"
```
benjamn:
```suggestion
    "regenerator-runtime": "^0.13.2"
```
Version 0.13.2 includes https://github.com/facebook/regenerator/pull/369, which defines `regeneratorRuntime` globally without needing to detect the global object, using [these two techniques](https://github.com/facebook/regenerator/blob/9c90408393bc8e4efc3b715d9e08a80c5f8367b9/packages/regenerator-runtime/runtime.js#L713-L726).

packages/babel-runtime-corejs2/package.json
```diff
@@ -10,7 +10,7 @@
   "author": "Sebastian McKenzie <sebmck@gmail.com>",
   "dependencies": {
     "core-js": "^2.5.7",
-    "regenerator-runtime": "^0.12.0"
+    "regenerator-runtime": "^0.13.1"
```
benjamn:
```suggestion
    "regenerator-runtime": "^0.13.2"
```

packages/babel-runtime/package.json
```diff
@@ -9,7 +9,7 @@
   "repository": "https://github.com/babel/babel/tree/master/packages/babel-runtime",
   "author": "Sebastian McKenzie <sebmck@gmail.com>",
   "dependencies": {
-    "regenerator-runtime": "^0.12.0"
+    "regenerator-runtime": "^0.13.1"
```
benjamn:
```suggestion
    "regenerator-runtime": "^0.13.2"
```

packages/babel-polyfill/src/index.js
```diff
@@ -25,4 +25,5 @@ if (global._babelPolyfill && typeof console !== "undefined" && console.warn) {
   );
 }
 
+global.regeneratorRuntime = regeneratorRuntime;
```
benjamn: This should no longer be necessary, though it also won't do any harm.

packages/babel-polyfill/src/noConflict.js
```diff
@@ -14,4 +14,5 @@ import "core-js/fn/promise/finally";
 // Ensure that we polyfill ES6 compat for anything web-related, if it exists.
 import "core-js/web";
 
-import "regenerator-runtime/runtime";
+import regeneratorRuntime from "regenerator-runtime/runtime";
+global.regeneratorRuntime = regeneratorRuntime;
```
benjamn: The old `import "regenerator-runtime/runtime"` should work fine here, too.

zloirock:
Cause of https://github.com/facebook/regenerator/pull/369, I bumped `regenerator-runtime` version in #7646 to `^0.13.2` in all required places.

MattiasBuelens:
Updated to version 0.13.2.

However, I think there's still a problem with the new approach. The new `regenerator-runtime` will always define `regeneratorRuntime` globally, **even when used through `@babel/plugin-transform-runtime`**. This kind of defeats the purpose of the runtime transform. 😕 

I've [suggested an alternative approach](https://github.com/facebook/regenerator/pull/369#issuecomment-472524249) in a comment on facebook/regenerator#369. I would love to hear your thoughts on this. 🙂

nicolo-ribaudo:
I agree with you. Ideally the global variable assignment should be opt-in (or at least it should be possible to opt out).
It shouldn't be a blocker for this to be merged, since we just generate `import "regenerator-runtime/runtime"` anyway, right?

MattiasBuelens:
> It shouldn't be a blocker for this to be merged, since we just generate `import "regenerator-runtime/runtime"` anyway, right?

Correct. Sure, if it's okay that `@babel/plugin-transform-runtime` defines a global `regeneratorRuntime` _for now_, then we can already land this.

PR #9446, opened 2019-02-02
Private Static Class Methods (Stage 3)

tim-mc:
| Fixed Issues?            | https://github.com/babel/proposals/issues/55
| Patch: Bug Fix?          | --
| Major: Breaking Change?  | --
| Minor: New Feature?      | Add private static class method support
| Tests Added + Pass?      | _WIP_
| Documentation PR Link    | 

cc @robpalme @nicolo-ribaudo @littledan

### DIFF
--- a/packages/babel-helper-create-class-features-plugin/src/fields.js
+++ b/packages/babel-helper-create-class-features-plugin/src/fields.js
@@ -24,7 +24,7 @@ export function buildPrivateNamesMap(props) {
         update.getId = prop.scope.generateUidIdentifier(`get_${name}`);
       } else if (prop.node.kind === "set") {
         update.setId = prop.scope.generateUidIdentifier(`set_${name}`);
-      } else if (prop.node.kind === "method" && isMethod) {
+      } else if (prop.node.kind === "method") {
         update.methodId = prop.scope.generateUidIdentifier(name);
       }
       privateNamesMap.set(name, update);
--- a/packages/babel-helper-create-class-features-plugin/src/fields.js
+++ b/packages/babel-helper-create-class-features-plugin/src/fields.js
@@ -24,7 +24,7 @@ export function buildPrivateNamesMap(props) {
         update.getId = prop.scope.generateUidIdentifier(`get_${name}`);
       } else if (prop.node.kind === "set") {
         update.setId = prop.scope.generateUidIdentifier(`set_${name}`);
-      } else if (prop.node.kind === "method" && isMethod && isInstance) {
+      } else if (prop.node.kind === "method" && isMethod) {
         update.methodId = prop.scope.generateUidIdentifier(name);
       }
       privateNamesMap.set(name, update);
@@ -147,6 +147,13 @@ const privateNameHandlerSpec = {
         [this.receiver(member), t.cloneNode(classRef), t.cloneNode(id)],
       );
     }
+    if (isStatic && isMethod) {
+      return t.callExpression(file.addHelper("classStaticPrivateMethodGet"), [
+        this.receiver(member),
+        t.cloneNode(classRef),
+        t.cloneNode(id),
+      ]);
+    }
     if (isMethod) {
       if (getId || setId) {
         return t.callExpression(file.addHelper("classPrivateFieldGet"), [
@@ -176,11 +183,25 @@ const privateNameHandlerSpec = {
       setId,
     } = privateNamesMap.get(name);
 
-    if (isStatic && !isMethod) {
-      return t.callExpression(
-        file.addHelper("classStaticPrivateFieldSpecSet"),
-        [this.receiver(member), t.cloneNode(classRef), t.cloneNode(id), value],
-      );
+    if (isStatic) {
+      if (!isMethod) {
+        return t.callExpression(
+          file.addHelper("classStaticPrivateFieldSpecSet"),
+          [
+            this.receiver(member),
+            t.cloneNode(classRef),
+            t.cloneNode(id),
+            value,
+          ],
+        );
+      }
+
+      return t.callExpression(file.addHelper("classStaticPrivateMethodSet"), [
+        this.receiver(member),
+        t.cloneNode(classRef),
+        t.cloneNode(id),
+        value,
+      ]);
     }
     if (isMethod) {
       if (setId) {
@@ -200,6 +221,12 @@ const privateNameHandlerSpec = {
   },
 
   call(member, args) {
+    const { privateNamesMap } = this;
+    const { name } = member.node.property.id;
+    const { static: isStatic, method: isMethod } = privateNamesMap.get(name);
+    if (isStatic && isMethod) {
+      return this.get(member);
+    }
     // The first access (the get) should do the memo assignment.
     this.memoise(member, 1);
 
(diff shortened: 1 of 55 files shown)


tim-mc:
@nicolo-ribaudo should this be considered a part of the `babel-plugin-proposal-private-methods` package?

nicolo-ribaudo:
I'd say yes, but let's also wait for @littledan 's opinion.

littledan:
Yes, I would include it in the private methods package. (If anyone has concerns, please let me know and I can provide more detailed rationale.)

tim-mc:
I've incorporated changes from https://github.com/babel/babel/pull/9508 in my PR since it solves a problem I ran into with handling `super`s.

tim-mc:
Support for both spec and loose modes have now been added. I'll see if I can figure out any more necessary test cases, but I think the PR is ready to start reviewing at this point. cc: @nicolo-ribaudo 

Keep in mind that un-merged changes from #9508 are present in this PR as well.

nicolo-ribaudo:
I have merged #9508

packages/babel-plugin-proposal-private-methods/test/fixtures/private-static-method/basic/input.js
```diff
@@ -0,0 +1,25 @@
...
+  static #privateStaticMethod() {
+    return 1017;
+  }
+
+  static staticMethod() {
+    return Cl.#privateStaticMethod();
+  }
+
+  static privateStaticMethod() {
+    return Cl.#privateStaticMethod();
```
nicolo-ribaudo: Can you also add a test for `this.#privateStaticMethod`?

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -176,11 +183,25 @@ const privateNameHandlerSpec = {
       setId,
     } = privateNamesMap.get(name);
 
-    if (isStatic && !isMethod) {
-      return t.callExpression(
-        file.addHelper("classStaticPrivateFieldSpecSet"),
-        [this.receiver(member), t.cloneNode(classRef), t.cloneNode(id), value],
-      );
+    if (isStatic) {
+      if (!isMethod) {
```
nicolo-ribaudo: Nit: we could do
```js
const helperName = isMethod ? "classStaticPrivateMethodSet" : "classStaticPrivateFieldSpecSet";
```

and then unify the two cases. Also in `get`.

packages/babel-helpers/src/helpers.js
```diff
@@ -1102,7 +1102,21 @@ helpers.classStaticPrivateFieldSpecSet = helper("7.0.2")`
...
   }
-  
+`;
+
+helpers.classStaticPrivateMethodGet = helper("7.3.2")`
+  export default function _classStaticPrivateMethodGet(receiver, classConstructor, descriptor) {
+    if (receiver !== classConstructor) {
+      throw new TypeError("Private static access of wrong provenance");
+    }
+    return descriptor.call(classConstructor);
```
nicolo-ribaudo: This should just return the method (without calling it), and then we should do `.call(receiver, arg1, arg2)`. Otherwise these cases don't work:

```js
var method = foo.#method;
var result = foo.#method(1, 2, 3);
```

packages/babel-plugin-proposal-private-methods/test/fixtures/private-static-method/basic/input.js
tim-mc: I have one here: https://github.com/babel/babel/pull/9446/files#diff-f49de9a84ff7fd3d1b84c3e6a076410f

tim-mc:
@nicolo-ribaudo thanks, updated.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -24,7 +24,7 @@ export function buildPrivateNamesMap(props) {
         update.getId = prop.scope.generateUidIdentifier(`get_${name}`);
       } else if (prop.node.kind === "set") {
         update.setId = prop.scope.generateUidIdentifier(`set_${name}`);
-      } else if (prop.node.kind === "method" && isMethod && isInstance) {
+      } else if (prop.node.kind === "method" && isMethod) {
```
nicolo-ribaudo: Nit: I don't think that this `isMethod` check is needed, since we are already checking `kind`.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -497,7 +531,7 @@ export function buildFieldsInitNodes(
     const isField = prop.isProperty();
     const isMethod = !isField;
 
-    if (isStatic && isField) {
+    if (isStatic) {
       const replaced = replaceThisContext(prop, ref, superRef, state, loose);
```
nicolo-ribaudo: In static methods, we should only replace `super` and not `this`:

```js
class A {
  static get a() { return 1 }
}

class B {
  static get b() { return 2 }

  static #getA() { return super.a }
  static #getB() { return this.b }

  static extract() {
    return [this.#getA, this.#getB];
  }
}

const [getA, getB] = B.extract();
expect(getA.call({ a: 3 })).toBe(1);
expect(getB.call({ b: 4 })).toBe(4);
```

You can add a flag to `replaceThisContext` to toggle this behavior.

packages/babel-helpers/src/helpers.js
```diff
@@ -1102,7 +1102,21 @@ helpers.classStaticPrivateFieldSpecSet = helper("7.0.2")`
...
   }
-  
+`;
+
+helpers.classStaticPrivateMethodGet = helper("7.3.2")`
+  export default function _classStaticPrivateMethodGet(receiver, classConstructor, descriptor) {
+    if (receiver !== classConstructor) {
+      throw new TypeError("Private static access of wrong provenance");
+    }
+    return descriptor;
```
nicolo-ribaudo: Nit: maybe call this `method` instead of `descriptor`?

packages/babel-helper-create-class-features-plugin/src/fields.js
tim-mc: Thanks, updated tests around this as well.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -547,9 +582,7 @@ export function buildFieldsInitNodes(
             privateNamesMap,
           ),
         );
-        staticNodes.push(
-          buildPrivateInstanceMethodDeclaration(prop, privateNamesMap),
-        );
+        staticNodes.push(buildPrivateMethodDeclaration(prop, privateNamesMap));
```
nicolo-ribaudo: Nit: here we should pass `loose` even if it isn't tecnically used for instance methods; but at least if someone changes `buildPrivateMethodDeclaration` to always rely on `loose` they won't unexpectedly get spec-mode output.

tim-mc:
Erm... not sure why [this](https://github.com/babel/babel/pull/9446#issuecomment-470343283) happened

nicolo-ribaudo:
We enabled it to avoid size regressions but it doesn't work well after rebases.

tim-mc:
I removed the "WIP" comment from the description. The PR is ready on my end.

nicolo-ribaudo:
Thank you!

PR #9450, opened 2019-02-03
Implement f# pipeline in parser

mAAdhaTTah:
| Fixed Issues?            | #6889

This is still very much "Work-in-Progress", but since I mostly don't know what I'm doing, I wanted to get this up early so I can get some feedback on alternate approaches, suggestions on the AST, etc. The two main things left are:

* [x] `await` parsing
* [x]  arrow body parsing

cc @js-choi @littledan


mAAdhaTTah:
Just pushed solo await parsing in the pipeline, courtesy of @thiagoarrais. Want to confirm the tests pass, but will also be pushing some tests so Thiago & I can collaborate on them.

mAAdhaTTah:
Actually may have gotten parenless arrow functions parsing. cc @littledan & @nicolo-ribaudo Definitely interested in suggested tests to add here / what cases I haven't covered.

packages/babel-parser/src/parser/expression.js
```diff
@@ -2269,4 +2295,36 @@ export default class ExpressionParser extends LValParser {
...
+    noIn: ?boolean,
+  ): N.Expression {
+    const startPos = this.state.start;
+    const startLoc = this.state.startLoc;
+
+    const node = this.startNode(startPos, startLoc);
+
+    if (this.lookahead().type === tt.arrow) {
+      this.state.potentialArrowAt = this.state.start;
+    }
```
mAAdhaTTah: This in particular I'm not sure is the best way to go about it.

mAAdhaTTah: Actually, the conditional wasn't needed at all. Always setting a potential arrow makes it parse correctly.

mAAdhaTTah:
Note that this AST is a modified implementation of what I inquired about [here](https://github.com/tc39/proposal-pipeline-operator/issues/104#issuecomment-383254158). The idea is to treat the `PipelineHead` & `PipelineBody` differently in regards to how they interact with arrow bodies. The operator will will terminate an arrow body when it's within a `PipelineBody` but **not** within a `PipelineHead`, allowing us to write:

```js
x => x |> y => y + 1 |> z => z * 2
```

which parses as (fully paren'd):

```js
x => (x |> (y => y + 1) |> (z => z * 2))
```

I don't know yet if the fact that it's feasible in babel means it's feasible from a spec perspective. Will need those with more expertise to weigh in.

Also note that accepting F# w/ parenless arrows may conflict with the movements toward making Smart Pipelines a superset of F# Pipelines.

packages/babel-parser/test/fixtures/experimental/pipeline-operator/proposal-fsharp-await/input.js
```diff
@@ -0,0 +1,3 @@
+async function test () {
+  return x |> await;
```
nicolo-ribaudo: Can you also test `x |> await |> y`?

packages/babel-parser/test/fixtures/experimental/pipeline-operator/proposal-fsharp-await/output.json
```diff
@@ -0,0 +1,181 @@
...
+                      "line": 2,
+                      "column": 14
+                    },
+                    "end": {
+                      "line": 2,
+                      "column": 19
+                    }
+                  },
+                  "body": {
+                    "type": "AwaitExpression",
```
nicolo-ribaudo: Maybe we could merge `PipelineBody` + `AwaitExpressoin` = `PipelineAwait`? So that plugins expecting `await` to have an `argument` don't break, and also because it's the only case where you can find `await` alone.

mAAdhaTTah:
> Question: why do we need the `PipelineHead` and `PipelineBody` nodes?

My understanding of the spec was we needed the different nodes to provide different precedence values related to the arrow function, depending on whether it's in the Head or Body. But this is not my strong suit, so I would need someone else to weigh in here.

packages/babel-parser/src/parser/expression.js
```diff
@@ -2269,4 +2295,33 @@ export default class ExpressionParser extends LValParser {
       this.state.topicContext.maxTopicIndex >= 0
     );
   }
+
+  makePipelineHead(left: N.Expression): N.Expression {
+    const node = this.startNode();
```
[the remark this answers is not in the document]
mAAdhaTTah: Fixed.

packages/babel-parser/src/parser/expression.js
```diff
@@ -333,8 +333,16 @@ export default class ExpressionParser extends LValParser {
 
         if (op === tt.pipeline) {
           this.expectPlugin("pipelineOperator");
+          const wasInPipeline = this.state.inPipeline;
           this.state.inPipeline = true;
           this.checkPipelineAtInfixOperator(left, leftStartPos);
+
+          if (
+            this.getPluginOption("pipelineOperator", "proposal") === "fsharp" &&
+            !wasInPipeline
```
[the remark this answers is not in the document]
mAAdhaTTah: FIxed.

packages/babel-parser/src/parser/expression.js
```diff
@@ -2269,4 +2295,33 @@ export default class ExpressionParser extends LValParser {
...
+
+  makePipelineHead(left: N.Expression): N.Expression {
+    const node = this.startNode();
+    node.head = left;
+
+    return this.finishNode(node, "PipelineHead");
+  }
+
+  parseFSharpPipelineBody(
+    op: TokenType,
```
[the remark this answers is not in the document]
mAAdhaTTah: Correct, done.

packages/babel-parser/test/fixtures/experimental/pipeline-operator/proposal-fsharp-await/input.js
mAAdhaTTah: Done.

packages/babel-parser/test/fixtures/experimental/pipeline-operator/proposal-fsharp-await/output.json
mAAdhaTTah: This is a new AST node, correct? I don't have a problem with it, but I think it would alter the spec.

mAAdhaTTah: According to the type for AwaitExpression, `argument` is already optional:

```js
export type AwaitExpression = NodeBase & {
  type: "AwaitExpression",
  argument: ?Expression,
};
```

nicolo-ribaudo:
I think that we should remove those nodes if (and only if) it is possible to understand if a node is the equivalent of a `PipelineHead` only by checking it's parent: this is how we disambiguate those cases in babel or babel/generator.

To my understanding, if `!t.isBinaryExpression(node.left, { operator: "|>" })`, then `node.left` is a pipeline head. Is it correct?

mAAdhaTTah:
@nicolo-ribaudo Yeah, that's correct. I can remove those nodes, but we do want to keep babel's produced AST in line with the spec, correct? If so, and we decide the spec requires them, would changing the AST be a breaking change? I'm not entirely clear on the spec needs here, and my understanding is Babel's AST should match the spec's AST, so I want to make sure we keep them in alignment before we remove anything. That said, I could easily be wrong about the spec's needs here, so I want to keep an eye on potential paths forward as we settle on the F# parenless AST.

nicolo-ribaudo:
Babel's AST is unrelated from the spec. For example, we didn't have a node for `ParenthesizedExpression`s until recently: the ast tree didn't need it, and we only added it for compatibility with Google Closure Compiler.

Since we have opposite opinions about it, maybe we could wait for someone else's opinoin? @babel/babel

mAAdhaTTah:
Oh, if the ASTs aren't related, then it doesn't really matter. I just wanted to make sure that if we made the change, we weren't backing ourselves into a corner for the future.

mAAdhaTTah:
Removed `PipelineHead`.

mAAdhaTTah:
The error below is a bug in Node 11:

>     TypeError: Cannot assign to read only property 'Symbol(Symbol.toStringTag)' of object '#<process>'

You can either upgrade Jest or downgrade node: https://github.com/facebook/jest/issues/8069#issuecomment-470520613

mAAdhaTTah:
It occurred to me this might be fixed in master. Merged that up.

packages/babel-parser/test/fixtures/experimental/pipeline-operator/proposal-fsharp-arrow-in-body-no-parens-indented-with-arg-parens/input.js
```diff
@@ -0,0 +1,3 @@
+(x) => x
+  |> (y) => y + 1
+  |> (z) => z * 2
```
nicolo-ribaudo: Maybe you could add a comment that this file is parsed as
```
(x) => (
    x |> (y) => (
             y + 1 |> (z) => z * 2
          )
)
```

mAAdhaTTah: Shoot, I don't think that's what we actually want, because the variables from the previous pipeline function shouldn't be available in the next function's scope.

mAAdhaTTah: So the underlying issue here is the arrow body's expression needs to stop parsing if it hits the pipeline operator token if it's nested within an arrow body expression already. I'm still poking around, but any thoughts / insights on where I might be able to implement this would be helpful.

thiagoarrais: Just checking here ('cause I think I've hit this). `|>` should be left-associative and parsed like this, right?

```
(x) => ((x
  |> (y) => y + 1 )
  |> (z) => z * 2)
```

nicolo-ribaudo: Sorry for the long delay. How should this code be parsed?

Is it correct that `a = b |> c` is `a = (b |> c)` and `a |> () => b |> c` is `a |> (() => b) |> c`?

Is so, how whould this be parsed?
```js
a |> () => b = c |> d
```

a) `|>` is an "hard break", which escapes the precedence even if it is nested inside other expressions:
  ```js
  a |> (() => b = c) |> d
  ```
b) `|>` and `() =>`'s precedences are modified, but nothing else:
  ```js
  a |> () => b = (c |> d)
  ```

mAAdhaTTah: It would be `a`–the mental shorthand is intended to be "pipeline operators terminate bodies of nested arrow functions".

nicolo-ribaudo: Ok, then you could add an `inFSharpPipelineDirectBody` flag to the state, and
1) Set it to `true` after `|>`
2) Set it to `false` inside `[({`/`})]`
3) If it is `true`, in `parseExprOp` return when you see `|>` instead of parsing it.

Another option is to pass a parameter (like `refShorthandDefaultPos`) down to the various parsing methods: it's not really Babel's style, but it is like you could probably write it in the spec.

I prefer the state flag.

mAAdhaTTah: Thanks! Will take a run at that. Agreed on the state flag.

mAAdhaTTah: @nicolo-ribaudo Updated mostly as suggested–I added tests for inside the `[` & `{` and I didn't have to change anything specific there for it to work as expected. Let me know if the AST looks off to you.

packages/babel-parser/src/parser/expression.js
```diff
@@ -2322,4 +2346,26 @@ export default class ExpressionParser extends LValParser {
...
+      this.parseMaybeUnary(),
+      startPos,
+      startLoc,
+      prec,
+      noIn,
+    );
+
+    this.state.inFSharpPipelineDirectBody = oldInFSharpPipelineDirectBody;
+
+    return this.finishNode(node, "PipelineBody");
```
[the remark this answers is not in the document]
mAAdhaTTah: I think so. The Smart proposal has a PipelineBody node as well.

packages/babel-parser/test/fixtures/experimental/pipeline-operator/proposal-fsharp-arrow-and-array-in-body/input.js
```diff
@@ -0,0 +1 @@
+x => x |> [y => y + 1] |> z => z * 2
```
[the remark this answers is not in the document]
mAAdhaTTah: Done. I may add a `withNonFSharpBodyContext` method to wrap these usages, cuz this pattern comes up a few times here.

Conceptually, this actually happens a decent amount in the parser: save values of current state, set them to new values, parse, reset to old values. Dunno if this is worthwhile to do in this PR, but a generic function for this would be useful.

packages/babel-parser/src/parser/expression.js
```diff
@@ -331,8 +331,11 @@ export default class ExpressionParser extends LValParser {
     const prec = this.state.type.binop;
     if (prec != null && (!noIn || !this.match(tt._in))) {
       if (prec > minPrec) {
-        const node = this.startNodeAt(leftStartPos, leftStartLoc);
         const operator = this.state.value;
+        if (operator === "|>" && this.state.inFSharpPipelineDirectBody) {
```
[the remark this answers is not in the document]
mAAdhaTTah: Done.

mAAdhaTTah:
@nicolo-ribaudo Addressed your comments / questions. I _think_ we still need `PipelineBody`, but can update that if you think we don't.

packages/babel-parser/test/fixtures/experimental/pipeline-operator/proposal-fsharp-arrow-and-array-in-body/input.js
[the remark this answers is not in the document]
nicolo-ribaudo: I have been working on adding something to abstract this pattern which will then be transpiled to normal assignments to avoid the performance overhead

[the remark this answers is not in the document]
mAAdhaTTah: Ok, cool, so I'll leave it as-is for now.

mAAdhaTTah:
Removed the `PipelineBody` node from the AST.

PR #9466, opened 2019-02-07
Update publishing command

nicolo-ribaudo:

This reflects how I have been releasing the last versions (https://github.com/babel/notes/issues/78#issuecomment-449519007).

In the future, we'll only need to run `make new-version && git push --tags` and it will run `make publish` from CI/Actions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,8 +130,12 @@ prepublish:
 	make prepublish-build
 	make test
 
+new-version:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -130,8 +130,12 @@ prepublish:
 	make prepublish-build
 	make test
 
+new-version:
+	./node_modules/.bin/lerna version --force-publish="@babel/runtime,@babel/runtime-corejs2,@babel/standalone,@babel/preset-env-standalone"
+
+# NOTE: Run make new-version fisrt
```
hzoo:
```suggestion
# NOTE: Run make new-version first
```

PR #9467, opened 2019-02-07
Show real version instead of core/preset-env versions in standalone pkgs

### LINKED ISSUE #6174: babel-standalone version number runs behind the actual version number

nicolo-ribaudo:
| Fixed Issues?            | Fixes the fact that https://babeljs.io/repl/ shows 7.2.2 instead of 7.3.2. Maybe also fixes babel/website#1793 and closes babel/website#1488, I'm not sure. Also fixes #6174 and closes #7039.

In `@babel/standalone` we showed the `@babel/core` version, but it doesn't really make sense because often it contains packages which are at a newer version. The same applies to `@babel/preset-env-standalone`: the preset is never updated, so the version was always old.

Note that the gulp script is run _after_ updating the version, so it should get the new one. THose packages are always force-updated, so they will always show the last version.

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -17,8 +17,7 @@ const source = require("vinyl-source-stream");
 const buffer = require("vinyl-buffer");
 const rollupBabel = require("rollup-plugin-babel");
 const rollupNodeResolve = require("rollup-plugin-node-resolve");
-const registerStandalonePackageTask = require("./scripts/gulp-tasks")
-  .registerStandalonePackageTask;
+const { registerStandalonePackageTask } = require("./scripts/gulp-tasks");
 
 const sources = ["codemods", "packages"];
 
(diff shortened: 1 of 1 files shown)


existentialism:
Also related: 

https://github.com/babel/babel/pull/7039
https://github.com/babel/babel/issues/6174

nicolo-ribaudo:
Probably https://github.com/babel/babel/pull/7039 has been fixed by https://github.com/babel/babel/pull/9466

PR #9468, opened 2019-02-07
Update flow to 0.92.1 and fix related issues

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  | flow

### DIFF
--- a/packages/babel-core/src/transformation/index.js
+++ b/packages/babel-core/src/transformation/index.js
@@ -127,10 +127,11 @@ function transformFile(file: File, pluginPasses: PluginPasses): void {
   }
 }
 
-function isThenable(val: Promise<*>): boolean {
+function isThenable(val: mixed): boolean {
   return (
     !!val &&
     (typeof val === "object" || typeof val === "function") &&
+    !!val.then &&
     typeof val.then === "function"
   );
 }
--- a/packages/babel-core/src/transformation/index.js
+++ b/packages/babel-core/src/transformation/index.js
@@ -127,7 +127,7 @@ function transformFile(file: File, pluginPasses: PluginPasses): void {
   }
 }
 
-function isThenable(val: mixed): boolean {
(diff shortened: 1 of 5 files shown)


packages/babel-core/src/transformation/index.js
```diff
@@ -127,7 +127,7 @@ function transformFile(file: File, pluginPasses: PluginPasses): void {
   }
 }
 
-function isThenable(val: mixed): boolean {
```
nicolo-ribaudo: I don't think that this is correct: `val` can be anything. Does something like this work?

```js
functin isThenable(val: mixed): boolean %checks (val instanceof Promise) {
  //...
}
```

https://flow.org/try/#0CYUwxgNghgTiAEAzArgOzAFwJYHtXywGcAVACxFSgCMIQAKANyggC54BbLADxGAEo2VHDlpR8AUjDkwAa0LxGzAqkIYxYEDkTwACjBydCIPgG4AUGdCRYCJjHhc2nHsHNms2ukTIVqtOlx8fPAA3mbwDgB0GOSodKZmAL7wIBBGoeFRMRTx5olAA

danez: The problem i tried to mitigate is this one: https://flow.org/try/#0GYVwdgxgLglg9mABDAzgFQBYFMwEMBGANlgBQBuuhAXIgLYwAeWAJgJQ35xzG5IDeAKESIATliggRSEgEhEQ4YgCESioUQAyDXPmLEJKAE8ADljjBEaxAF5biAERx8AKyzR7iAD6fER0+ctKGzt7UEhYBHtWTW09BWE-Mws1ADoobCRbawcw6Hgwex1dRFYAbjkBAF8gA

But I agree that the input can be anything.

danez: I changed it

PR #9469, opened 2019-02-07
Exclude generate @babel/types files from coverage report

nicolo-ribaudo:

I don't think that we should include these files in the coverage report, since testing them is not realistic anyway?
https://github.com/babel/babel/blob/master/packages/babel-types/src/validators/generated/index.js

Also, there are a lots of untested unused helpers in babel/types (e.g. https://codecov.io/gh/babel/babel/src/master/packages/babel-types/src/validators/isSpecifierDefault.js): we should check them one by one and decide if we want to deprecate them and remove in v8 or make `good first issue` issues to test them.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -8,3 +8,5 @@ coverage:
         target: "80%"
     patch:
       enabled: false
+ignore:
+  - packages/babel-types/src/*/generated/index.js


danez:
About the helpers I am not sure. Most of them I guess are used somewhere in this repo, at least I always assumed.

nicolo-ribaudo:
Well, if the coverage report says 0% it is because we aren't using them anywhere.

danez:
Or the location where it is used is also not tested :P

PR #9474, opened 2019-02-08
Partial application plugin

### LINKED ISSUE #9342: Partial Application syntax: Stage 1

byara:
| Fixed Issues?            | Fixes #9342
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR addresses the proposal regarding [Partial Application plugin](https://github.com/babel/proposals/issues/32). 

The syntax rules that was suggested in the proposal to be covered:

1. - [x] Given f(?), the expression f is evaluated immediately.
2. - [x] Given f(?, x), the non-placeholder argument x is evaluated immediately and fixed in its position.
3. - [x] Given f(?), excess arguments supplied to the partially applied function result are ignored.
4. - [x] Given f(?, ?) the partially applied function result will have a parameter for each placeholder token that is supplied in that token's position in the argument list.
5. - [x] Given f(this, ?), the this in the argument list is the lexical this.
6. - [x] Given f(?), the this receiver of the function f is fixed as undefined in the partially applied function result.
7. - [x] Given f(?), the length of the partially applied function result is equal to the number of ? placeholder tokens in the argument list.
8. - [x] Given f(?), the name of the partially applied function result is f.name.
9. - [x] Given o.f(?), the references to o and o.f are evaluated immediately.
10. - [x] Given o.f(?), the this receiver of the function o.f is fixed as o in the partially applied function result.
11. - [x] Given f(g(?)), the result is equivalent to f(_0 => g(_0)) not _0 => f(g(_0)). This is because the ? is directly part of the argument list of g and not the argument list of f.

- ~~I'm not 100% sure about 5, 6, 7 and 8 and how to address them.~~ 
- I welcome any suggestions and reviews to improve this, I'm doing this as part of my thesis and anything that could teach me something is great.


nicolo-ribaudo:
(I'm marking this as Low priority until the parser PR is merged - don't worry! :wink: )

packages/babel-plugin-proposal-partial-application/src/index.js
```diff
@@ -0,0 +1,286 @@
...
+        const allArgs = unwrapAllArguments(node, scope);
+        const finalArgs = mapNonPlaceholderToLVal(nonPlaceholderDecl, allArgs);
+        const placeholderVals = placeholderLVal(allArgs);
+
+        if (node.callee.type === "MemberExpression") {
+          const data = [
+            t.variableDeclaration("const", [
+              t.variableDeclarator(
+                receiverLVal,
+                t.identifier(receiverRVal(node)),
```
nicolo-ribaudo: You can use `path.scope.push` to generate those vars and avoid the iife:

```js
const g = o.f(?, x, 1);

// becomes

var _receiver, _func, _param, _param2; // generated by scope.push
const g = (_receiver = o, _func = o.f, _param = x, _param2 = 1, _arg => _func.call(_receiver, _arg, _param, _param2);
```

Btw, there is probably some scope/types helper function that says that `1` is a pure constant value and can thus can be transpiled as
```js
const g = (_receiver = o, _func = o.f, _param = x, _arg => _func.call(_receiver, _arg, _param, 1);
```

packages/babel-plugin-proposal-partial-application/src/index.js
```diff
@@ -0,0 +1,286 @@
...
+    return declarator;
+  }
+
+  /**
+   * It replaces the values of non-placeholder args in allArgs
+   * @param {Array<Declarator>} nonPlaceholderDecl that has no placeholder in them
+   * @param {Array<Arguments>} args
+   */
+  function mapNonPlaceholderToLVal(nonPlaceholderDecl, allArgsList) {
+    const clone = Array.from(allArgsList);
```
danielcaldas:
```javascript
Array.from(allArgsList)
  .map(cl => ...);
```

packages/babel-plugin-proposal-partial-application/src/index.js
```diff
@@ -0,0 +1,286 @@
...
+   * @param node a callExpression node
+   * @returns boolean
+   */
+  function hasArgumentPlaceholder(node) {
+    for (let i = 0; i < node.arguments.length; i++) {
+      if (t.isArgumentPlaceholder(node.arguments[i])) {
+        return true;
+      }
+    }
+    return false;
```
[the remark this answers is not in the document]
nicolo-ribaudo: `is*` functions can accept two arguments, so you'll need to wrap it in an arrow to only use one parameter (or use partial application lol)

byara:
@nicolo-ribaudo @danielcaldas thank you for all the reviews, I'll try to address them as soon as possible.

Meanwhile, I have a few questions:
1. Regarding this syntactic rule of the proposal:
> 8.  Given f(?), the name of the partially applied function result is f.name.

What do you think about this solution:
```
// if we have this:
const foo = bar(1, ?, 3, ?);

// currently converted to this:
const foo = (() => {
  const _func = bar;
  const _param = 1,
        _param2 = 3;
  return (_argPlaceholder, _argPlaceholder2) => _func(_param, _argPlaceholder, _param2, _argPlaceholder2);
})();

bar.name; // => bar
foo.name; // => empty
```
Instead we do this:
```
const foo = (() => {
  const _func = bar;
  const _param = 1,
        _param2 = 3;
{
  const bar = (_argPlaceholder, _argPlaceholder2) => _func(_param, _argPlaceholder, _param2, _argPlaceholder2);
  return bar;
}
})();

bar.name; // => bar
foo.name; // => bar
```
or maybe we can get the `foo.name` to be `foo`.

packages/babel-plugin-proposal-partial-application/src/index.js
byara: I would like to ask for a bit of clarification:
- what is the advantage of not using the iife?
- could you explain how pushing to `scope` creates the declarations for me?

nicolo-ribaudo:
Can we use normal functions?

```js
const add1 = add(1, ?);

// ->

var _func, _param;
const add1 = (_func = add, _param = 1, function add(_placeholder) { return _func(_param, _placeholder) });
```

byara:
> Can we use normal functions?
> 
> ```js
> const add1 = add(1, ?);
> 
> // ->
> 
> var _func, _param;
> const add1 = (_func = add, _param = 1, function add(_placeholder) { return _func(_param, _placeholder) });
> ```

That's a good idea and it looks better. I'll change the plugin accordingly.

nicolo-ribaudo:
Instead of using the custom `unfold` function, would `scope.generateUidIdentifierBasedOnNode` work?

packages/babel-plugin-proposal-partial-application/src/index.js
```diff
@@ -0,0 +1,279 @@
...
+        const allArgs = unwrapAllArguments(node, scope);
+        const finalArgs = mapNonPlaceholderToLVal(nonPlaceholderDecl, allArgs);
+        const placeholderVals = placeholderLVal(allArgs);
+
+        if (node.callee.type === "MemberExpression") {
+          const data = [
+            t.variableDeclaration("const", [
+              t.variableDeclarator(
+                receiverLVal,
+                t.identifier(receiverRVal(node)),
```
nicolo-ribaudo: I don't understand what `receiverRVal` (which, btw, doesn't generate a valid identifier since the result can contain `.`) is needed for: isn't using `node.callee.object` (and `node.callee.property` for `functionLVal`) enough?

nicolo-ribaudo: Also, it doesn't work:

```js
a.b.fn().c["d"].e(?);

// ->

(() => {
  const _receiver = a.b.fn.c.undefined;
  const _func = a.b.fn.c.undefined.e;
  return _argPlaceholder => _func.call(_receiver, _argPlaceholder);
})();
```

nicolo-ribaudo: Oh, and instead of using `newFuncLVal`/`newReceiverLVal` you could use `path.scope.generateUidIdentifierBasedOnNode(node.callee)` and `path.scope.generateUidIdentifierBasedOnNode(node.callee.object)` to generate  names that make it easier to understand where they come from.
```js
object.get().fn(?);

// ->

(() => {
  const _object$get = object.get;
  const _object$get$fn = object.get.fn;
  return _argPlaceholder => _object$get$fn.call(_object$get, _argPlaceholder);
})();
```

nicolo-ribaudo: Oh, and the previous example should be something like this, otuherwise we might execute some accessors twice:
```js
(() => {
  const _object$get = object.get;
  const _object$get$fn = _object$get.fn;
  return _argPlaceholder => _object$get$fn.call(_object$get, _argPlaceholder);
})();
```

packages/babel-plugin-proposal-partial-application/test/fixtures/general/member-expression/output.js
```diff
@@ -0,0 +1,13 @@
...
+
+var _param, _o, _o$f, _param2, _param3, _p, _p$b, _a$b$c$d$e, _a$b$c$d$e$foo;
+
+const g = (_o = o, _o$f = o.f, _param = x, function f(_argPlaceholder) {
+  return _o$f.call(_o, _argPlaceholder, _param, 1);
+});
+const h = (_p = p, _p$b = p.b, _param2 = y, _param3 = x, function b(_argPlaceholder2) {
+  return _p$b.call(_p, 1, _param2, _param3, 2, _argPlaceholder2);
+});
+const j = (_a$b$c$d$e = a.b.c.d.e, _a$b$c$d$e$foo = a.b.c.d.e.foo, function foo(_argPlaceholder3) {
```
nicolo-ribaudo: This should be

```js
const j = (_a$b$c$d$e = a.b.c.d.e, _a$b$c$d$e$foo = _a$b$c$d$e.foo, ...
```

because `a.b.c.d.e` could be getters which must be accessed exactly once.

packages/babel-plugin-proposal-partial-application/src/index.js
```diff
@@ -0,0 +1,178 @@
...
+    });
+  }
+
+  /**
+   * It replaces the values of non-placeholder args in allArgs
+   * @param {Array<Node>} nonPlaceholderArgs that has no placeholder in them
+   * @param {Array<Node>} args
+   */
+  function mapNonPlaceholderToLVal(nonPlaceholderArgs, allArgsList) {
+    const clonedArgs = Array.from(allArgsList);
```
nicolo-ribaudo: Why do you need to clone this array? You are only cloning the array, not its elements.

packages/babel-plugin-proposal-partial-application/src/index.js
```diff
@@ -0,0 +1,178 @@
...
+    clonedArgs.map(arg => {
```
nicolo-ribaudo: Nit: you can use `forEach`

packages/babel-plugin-proposal-partial-application/src/index.js
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * Takes the full list of arguments and extracts placeholders from it
+   * @param {Array<Argument>} allArgsList full list of arguments
+   * @returns {Array<ArgumentPlaceholder>} cloneList
+   */
+  function placeholderLVal(allArgsList) {
+    let cloneList = [];
+    allArgsList.forEach(item => {
+      if (item.name && item.name.includes("_argPlaceholder")) {
+        cloneList = cloneList.concat(item);
```
nicolo-ribaudo: `cloneList.push(item)`

packages/babel-plugin-proposal-partial-application/src/index.js
```diff
@@ -0,0 +1,178 @@
...
+      CallExpression(path) {
+        if (!hasArgumentPlaceholder(path.node)) {
+          return;
+        }
+        const { node, scope } = path;
+        const functionLVal = path.scope.generateUidIdentifierBasedOnNode(
+          node.callee,
+        );
+        const receiverLVal = path.scope.generateUidIdentifierBasedOnNode(
+          node.callee.object,
```
nicolo-ribaudo: This should only be generated when `node.callee` is a member expression.

nicolo-ribaudo:
Does `...spread` work?

```js
foo(a, ...b, ?);
```

Since the spec doesn't define how it should behave (or I can't find it), we can just throw an error when we find it. Or maybe it uses one of the following semantics? (cc @rbuckton)

```js
var _foo, _a, _b;
_foo = foo, _a = a, _b = b, function foo(_arg) { return _foo(_a, ..._b, _arg) }

// or maybe it calls [Symbol.iterator] earlier

_foo = foo, _a = a, _b = [...b], function foo(_arg) { return _foo(_a, ..._b, _arg) }
```

rbuckton:
@nicolo-ribaudo: Anything usable in a normal function call should be usable here. All evaluation should be eager, so the second example is correct.

nicolo-ribaudo:
(From the original post)

> * I'm not 100% sure about 5, 6, 7 and 8 and how to address them.

Don't they already work?

packages/babel-plugin-proposal-partial-application/src/index.js
byara: you are correct, this is useless. I removed it.

byara:
> (From the original post)
> 
> > * I'm not 100% sure about 5, 6, 7 and 8 and how to address them.
> 
> Don't they already work?

I added a few tests for 7 and 8. Regarding 5, I think, I sort of understand what "lexical this" means.
For 6, I don't think I understand the "this receiver" for `f(?)` 😅
For other reviews, I'll address them ASAP.

rbuckton:
Ignoring partial application for a second, when you call `f()` in *strict mode*, the receiver (the value used for `this` *inside* of `f`) is `undefined`. When you call `o.f()`, the receiver is `o`. When you do `f.call(o)`, the receiver is `o`. 

6 basically means that when you do `g = o.f(?); g()`, the value `o` is preserved as the receiver.

byara:
> Ignoring partial application for a second, when you call `f()` in _strict mode_, the receiver (the value used for `this` _inside_ of `f`) is `undefined`. When you call `o.f()`, the receiver is `o`. When you do `f.call(o)`, the receiver is `o`.
> 
> 6 basically means that when you do `g = o.f(?); g()`, the value `o` is preserved as the receiver.

@rbuckton Thank you for the explanation. I think the 6 is also addressed.

byara:
@nicolo-ribaudo one of the builds for travis times out. Would you retrigger it?

nicolo-ribaudo:
It seems that after the rebase the plugin disappeared? If you need it, you can restore from https://github.com/nicolo-ribaudo/babel/tree/pr/byara/9474-backup (it is 6 days old so commits after https://github.com/babel/babel/pull/9474#pullrequestreview-203689640 are not included). If you prefer, I suggest using `git reflog` to restore the lost commits.

byara:
> It seems that after the rebase the plugin disappeared? If you need it, you can restore from https://github.com/nicolo-ribaudo/babel/tree/pr/byara/9474-backup (it is 6 days old so commits after [#9474 (review)](https://github.com/babel/babel/pull/9474#pullrequestreview-203689640) are not included). If you prefer, I suggest using `git reflog` to restore the lost commits.

Thanks for the `git reflog` tip, It saved my life 😅

packages/babel-plugin-proposal-partial-application/src/index.js
```diff
@@ -0,0 +1,192 @@
...
+   * Unwrap the arguments of a CallExpression, and creates
+   * assignmentExpression from them and returns them as an array.
+   * @param {Object} node CallExpression node
+   * @returns {Array<AssignmentExpression>}
+   */
+  function unwrapArguments(node, scope) {
+    let newNode;
+    let argId;
+    return node.arguments.reduce((acc, arg) => {
+      if (!t.isArgumentPlaceholder(arg) && !t.isNumericLiteral(arg)) {
```
nicolo-ribaudo: Instead of `t.isNumericLiteral` you can use `t.isImmutable`, which also handles strings/booleans

packages/babel-plugin-proposal-partial-application/src/index.js
```diff
@@ -0,0 +1,192 @@
...
+        }
+        const { node, scope } = path;
+        const functionLVal = path.scope.generateUidIdentifierBasedOnNode(
+          node.callee,
+        );
+
+        const nonPlaceholderArgs = unwrapArguments(node, scope);
+        const allArgs = unwrapAllArguments(node, scope);
+        const finalArgs = mapNonPlaceholderToLVal(nonPlaceholderArgs, allArgs);
+        const placeholderVals = placeholderLVal(allArgs);
```
nicolo-ribaudo: While this works, I think that it can be made better: relying on an automatically generated name in `placeholderLVal` doesn't feel 100% ok, and the two nested loops in `mapNonPlaceholderToLVal` can be avoided.

You could merge those functions like this:
```
const argsInitializers = unwrapArguments(node, scope);
const [placeholdersParams, args] = replacePlaceholders(node, scope); 
```

Where `unwrapsArguments` returns the `_param1 = foo, _param2 = bar` part from `fn(foo, ?, bar, ?)` _and_ replaces the parameters: the call becomes `fn(_param1, ?, _param2, ?)`. It could be implemented roughly like this:

```js
function unwrapArguments(node, scope) {
  var init = [];
  for each (arg in node.arguments with index i) {
    if (arg is neither a placeholder nor an immutable value) {
      var id = generate param id;
      init.push("${id} = ${arg}")
      node.arguments[i] = t.cloneNode(id);
    }
  }
  return init;
}
```

`replacePlaceholders` would instead transform `fn(_param1, ?, _param2, ?)` to `fn(_param1, _placeholder1, _param2, _placeholder2)`, and additionally return an array `[_placeholder1, _placeholder2]` to be used to create the arguments of the new function:

```js
function replacePlaceholders(node, scope) {
  var placeholders = [];
  var args = [];
  for each (arg in node.arguments) {
    if (arg is a placeholder) {
      var id = generate placeholder id;
      placeholders.push(id);
      args.push(t.cloneNode(id));
    } else {
      args.push(arg);
    }
  }
  return [placeholders, args];
}
```

At the end, you would need to create the result like this:
```js
functionLVal = FN, ...argsInitializers, function NAME(...placeholdersParams) { return functionLVal(...args) }
```

byara: This is a really good idea, but I cannot figure out how to handle the `spreadElement`. My implementation of the two functions:

```js
function unwrapArguments(node, scope) {
    const init = [];
    for (let i = 0; i < node.arguments.length; i++) {
      if (
        !t.isArgumentPlaceholder(node.arguments[i]) &&
        !t.isImmutable(node.arguments[i])
      ) {
        const id = scope.generateUidIdentifierBasedOnNode(
          node.arguments[i],
          "param",
        );
        scope.push({ id });
        if (t.isSpreadElement(node.arguments[i])) {
          init.push(
            t.assignmentExpression(
              "=",
              t.cloneNode(id),
              t.arrayExpression([node.arguments[i]]),
            ),
          );
        } else {
          init.push(
            t.assignmentExpression("=", t.cloneNode(id), node.arguments[i]),
          );
        }
        node.arguments[i] = t.cloneNode(id);
      }
    }
    return init;
  }

  function replacePlaceholders(node, scope) {
    const placeholders = [];
    const args = [];

    node.arguments.forEach(arg => {
      if (t.isArgumentPlaceholder(arg)) {
        const id = scope.generateUid("_argPlaceholder");
        placeholders.push(t.identifier(id));
        args.push(t.cloneNode(t.identifier(id)));
      } else {
        args.push(arg);
      }
    });
    return [placeholders, args];
  }
```
In `unwrapArguments`, I can easily created the assignment for the `SpreadElement`, but when we are pushing the arguments into the `args` in `replacePlaceholders`, there is no information about the type of the argument to create the following form:
```js
const q = (_foo = foo, _a = a, _param = [...b], function foo(_argPlaceholder) {
   return _foo(_a, ..._param, _argPlaceholder);
});
```
and I would always endup with:

```js
const q = (_foo = foo, _a = a, _param = [...b], function foo(_argPlaceholder) {
   return _foo(_a, _param, _argPlaceholder);
});
```

byara: Maybe we could somehow in `unwrapArguments` send the needed information along with `init`?

nicolo-ribaudo: Instead of `node.arguments[i] = t.cloneNode(id);`, for spread element you need to do `node.arguments[i].argument = t.cloneNode(id);`.

packages/babel-plugin-proposal-partial-application/src/index.js
```diff
@@ -0,0 +1,151 @@
...
+
+  function replacePlaceholders(node, scope) {
+    const placeholders = [];
+    const args = [];
+
+    node.arguments.forEach(arg => {
+      if (t.isArgumentPlaceholder(arg)) {
+        const id = scope.generateUid("_argPlaceholder");
+        placeholders.push(t.identifier(id));
+        args.push(t.cloneNode(t.identifier(id)));
```
nicolo-ribaudo: No need to `cloneNode` here, `t.identifier` returns a new node.

nicolo-ribaudo:
The failing tests are fixed by https://github.com/babel/babel/pull/9558

packages/babel-plugin-proposal-partial-application/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@babel/plugin-proposal-partial-application",
+  "version": "7.2.0",
+  "description": "Transform pipeline operator into call expressions",
```
nicolo-ribaudo: This needs to be updated to match the readme.

nicolo-ribaudo:
Could you rebase this PR, since the parser PR has been merged?

packages/babel-plugin-proposal-partial-application/test/fixtures/general/multiple-calls/options.json
```diff
@@ -0,0 +1,3 @@
+{
+  "plugins": ["proposal-partial-application"]
+}
```
danez: nit: The `options.json` files are the same for all tests as I can see, so the config file could be moved on level up into the `general` folder. Then all the fixtures in `general` would share the config.

byara: @danez Thanks for the feedback. 
I moved `options.json` one level up and added a new test.

packages/babel-plugin-proposal-partial-application/test/fixtures/general/bound-method/input.js
```diff
@@ -0,0 +1,10 @@
+class Collator {
+  constructor() {
+    this.compare = this.compare(?, ?);
+  }
+  compare(a, b) {
+    if (a > b) {
+      return a;
+    };
```
existentialism: Nit, but is this semi necessary?

packages/babel-plugin-proposal-partial-application/src/index.js
```diff
@@ -0,0 +1,151 @@
...
+                    ),
+                  ),
+                ],
+                [],
+              ),
+              false,
+              false,
+            ),
+          ]);
+          path.replaceWith(finalExpression);
```
existentialism: Not big deal by any means, but since both of these end with replacing the path with a `sequenceExpression`, we could do:

```js
const sequenceParts = [];

if (node.callee.type === "MemberExpression") {
  sequenceParts.push(
    /* stuff */
  );
} else {
  sequenceParts.push(
    /* stuff */
  );
}

path.replaceWith(t.sequenceExpression(sequenceParts));
```

packages/babel-plugin-proposal-partial-application/test/fixtures/general/bound-method/input.js
byara: yep, unnecessary, I'll remove it.

packages/babel-plugin-proposal-partial-application/src/index.js
byara: good idea! made the necessary changes.

nicolo-ribaudo:
Awesome work!

PR #9477, opened 2019-02-08
Fix regression with let

### LINKED ISSUE #9472: Regression: Babel 7.3.2 fails to parse `let` statements with object destructuring after a line break
## Bug Report

**Current Behavior**
Babel 7.3.2 fails to parse `let` statements with object destructuring after a line break.

**Input Code**
Failing REPL from https://github.com/babel/babel/pull/9402
https://babeljs.io/repl/build/9899/#?code_lz=DYUwLgBAUAkA3hAhhAvhAvBBiBcECMqANFBGRAEYYQBMQA

> Unexpected reserved word 'let' (1:0)

Working REPL (for now, until it gets updated to 7.2.3)
https://babeljs.io/repl/#?code_lz=DYUwLgBAUAkA3hAhhAvhAvBBiBcECMqANFBGRAEYYQBMQA

```js
let
    { a } = { a: 1 },
    b = 2
```

**Environment**
- Babel version(s): 7.2.3
- Node/npm version: Node 10
- OS: Arch Linux
- How you are using Babel: cli / loader

Encountered this in a project using Quasar Framework: https://github.com/quasarframework/quasar/blob/e8c1cc69429b3e84f8d7512d33aba70b8073fe9d/src/utils/popup.js#L5-L12

danez:
| Fixed Issues?            | Fixes #9472
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This fixes the regression in 3.7.2 with let in certain cases and also allows a lot more new valid cases with functions (see test262 whitelist). This was backported from my branch which includes proper scope tracking. This is also the reason I did not care about good error messages here, as the code involving them will change anyway.


packages/babel-parser/src/parser/statement.js
```diff
@@ -148,18 +144,28 @@ export default class StatementParser extends ExpressionParser {
...
+        // TODO: Remove this once we have proper scope tracking in place.
+        if (context && result.generator) {
+          this.unexpected(node.start);
+        }
 
+        return result;
+      }
       case tt._class:
-        if (!declaration) this.unexpected();
+        if (context) this.unexpected();
```
nicolo-ribaudo: As a side note, `Function declaration is not allowed in this context` is much nicer :smile:

PR #9478, opened 2019-02-08
Add flow types to @babel/parser

goodmind:
| Documentation PR Link    | 


nicolo-ribaudo:
Is it possible to auto-generate them from the source code?

goodmind:
@nicolo-ribaudo is typescript typings autogenerated?

goodmind:
I checked now, they should be in lib folder to typecheck

nicolo-ribaudo:
No, but since we are using flow in our codebase I tought that it would be possible

goodmind:
@nicolo-ribaudo I think Flow team works on something like this, there was `gen-flow-files` but they removed it https://github.com/facebook/flow/issues/5871

nicolo-ribaudo:
Oh ok :+1:

goodmind:
@nicolo-ribaudo do you know where I can add build step to move this file to `lib` when publishing?

nicolo-ribaudo:
Could you add a comment at the top of the new file like this? Also in the ts definitions.

```js
/*================================!!! NOTE !!!===============================*\
|  When updating this file, don't forget to update typings/babel-parser.d.ts  |
\*===========================================================================*/
```

> do you know where I can add build step to move this file to `lib` when publishing?

You can add it at the end of `prepublish-build` in `Makefile`

nicolo-ribaudo:
I don't think that it should go in `typings`, since that folder is conventionally ued by typescript. Maybe in src?

goodmind:
@nicolo-ribaudo wouldn't it break internal types in`.flowconfig`
```
module.name_mapper='^@babel\/\([a-zA-Z0-9_\-]+\)$' -> '<PROJECT_ROOT>/packages/babel-\1/src/index'
```

nicolo-ribaudo:
I don't think so, since that type is trying to describe exactly `index.js`?

liuxingbaoyu:
Thank you for your PR! Since it is now obsolete, I will close it.

PR #9481, opened 2019-02-09
[preset-env] Don't use async-to-generator when already using regenerator

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          | 
| Major: Breaking Change?  |
| Minor: New Feature?      | Maybe?
| Documentation PR Link    | 
| Any Dependency Changes?  |

Since regenerator handles async functions, there is no need to load an additional transform. You can see the benefits at [f087c3f](https://github.com/babel/babel/commit/f087c3f5bbcafd6d9689f261798ae809b9fb4fcc?w=1). `async-to-generator` will still be loaded when `regenerator` is not needed.

NOTE: It would be better to also transform async generators only using regenerator when possible, but we can't do that because regenerator doesn't support `for await`.


nicolo-ribaudo:
Unless I have a very strong passion for resolving merge conflicts :stuck_out_tongue:

loganfsmyth:
I think we should be careful with this. It seems like a pretty large change in expectations, and regenerator hasn't always been very good at handling cases where ES6 syntax is used inside generators. I'm just worried that this could introduce changes in behavior because various things could run in a different order.

nicolo-ribaudo:
> regenerator hasn't always been very good at handling cases where ES6 syntax is used inside generators

This doesn't change with this PR. Consider this code:
```js
async function fn() {
  const { foo } = await syntaxRegeneratorDoesNotUnderstand();
}
```
with this PR, it would be directly passed to regenerator (which would probably break). Without this PR, that code would be first transformed to
```js
_asyncToGenerator(function* fn() {
  const { foo } = yield syntaxRegeneratorDoesNotUnderstand();
});
```
and then passed to regenerator, which would have broken anyway.

> I'm just worried that this could introduce changes in behavior because various things could run in a different order.

Regenerator transforms functions on exit, so I think that it will always run after that the function body (wether it is still an async function or it has become a generator function) has been transformed:
https://github.com/facebook/regenerator/blob/f44b30de0bad413e1256322004b63ca39ad8d75e/packages/regenerator-transform/src/visit.js#L19

PR #9483, opened 2019-02-10
Fix peer deps

jrjohnson:
| Any Dependency Changes?  | `peerDependencies` updated

Remove the -0 from all `peerDependencies` as it causes warnings when correct
babel/core@7 is installed.

I think this is just a continuation of @hzoo's work in #8445, `yarn` doesn't resolve `@babel/core 7.0.0` for `^7.0.0-0`. Which results in warnings like: 
```
warning "ember-cli > @babel/plugin-transform-modules-amd@7.2.0" has unmet peer dependency "@babel/core@^7.0.0-0".
```

With a depedency tree like
```
$yarn list
  yarn list v1.13.0
  ├─ @babel/code-frame@7.0.0
  │  └─ @babel/highlight@^7.0.0
  ├─ @babel/core@7.2.2
---snip
```

### DIFF
--- a/codemods/babel-plugin-codemod-object-assign-to-object-spread/package.json
+++ b/codemods/babel-plugin-codemod-object-assign-to-object-spread/package.json
@@ -16,7 +16,7 @@
     "@babel/plugin-syntax-object-rest-spread": "^7.0.0"
   },
   "peerDependencies": {
-    "@babel/core": "^7.0.0-0"
+    "@babel/core": "^7.0.0"
   },
   "devDependencies": {
     "@babel/core": "^7.0.0",
--- a/codemods/babel-plugin-codemod-optional-catch-binding/package.json
+++ b/codemods/babel-plugin-codemod-optional-catch-binding/package.json
@@ -16,7 +16,7 @@
     "@babel/plugin-syntax-optional-catch-binding": "^7.0.0"
   },
   "peerDependencies": {
-    "@babel/core": "^7.0.0-0"
+    "@babel/core": "^7.0.0"
   },
   "devDependencies": {
     "@babel/core": "^7.0.0",
--- a/packages/babel-cli/package.json
+++ b/packages/babel-cli/package.json
@@ -33,7 +33,7 @@
     "chokidar": "^2.0.3"
   },
   "peerDependencies": {
-    "@babel/core": "^7.0.0-0"
+    "@babel/core": "^7.0.0"
   },
   "devDependencies": {
     "@babel/core": "^7.2.0",
--- a/packages/babel-node/package.json
+++ b/packages/babel-node/package.json
@@ -26,7 +26,7 @@
     "v8flags": "^3.1.1"
   },
   "peerDependencies": {
-    "@babel/core": "^7.0.0-0"
+    "@babel/core": "^7.0.0"
   },
   "devDependencies": {
     "@babel/core": "^7.2.2",
--- a/packages/babel-plugin-external-helpers/package.json
+++ b/packages/babel-plugin-external-helpers/package.json
@@ -15,7 +15,7 @@
     "@babel/helper-plugin-utils": "^7.0.0"
   },
   "peerDependencies": {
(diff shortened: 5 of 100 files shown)


jrjohnson:
I couldn't tell you why @danez, I'm using yarn 1.13.0, may be something in the way babel core is installed through Ember.js. I figured this change would probably be welcome without full reproduceability, but I can dig in if its an issue (sounds like it isn't).

nicolo-ribaudo:
Does `^7.0.0` match things like `7.5.0-beta.1`? We don't currently plan to release beta versions, but I don't want to prevent us from doing it.

jrjohnson:
@nicolo-ribaudo I don't think it does, but I think given the two options that getting a warning for beta mis-match is better than the same warning for releases. In the case of a beta the user is aware that they have installed something different and won't be bothered. In the case of releases there is a warning when there shouldn't be.

danez:
But there shouldn't be a warning as `^7.0.0-0` matches stable releases. This is either a problem of yarn (old version?) or babel-core is actually not listed in `package.json`.

nicolo-ribaudo:
@jrjohnson Could you share a small repository which shows the warning? I have the same yarn version installed as you have, so I should be able to reproduce it.

jrjohnson:
I don't have a small one, but I created a fresh ember app and I see this. I created one at:
https://github.com/jrjohnson/test-babel-ember

```
git clone https://github.com/jrjohnson/test-babel-ember.git
Cloning into 'test-babel-ember'...
remote: Enumerating objects: 34, done.
remote: Counting objects: 100% (34/34), done.
remote: Compressing objects: 100% (27/27), done.
remote: Total 34 (delta 1), reused 34 (delta 1), pack-reused 0
Unpacking objects: 100% (34/34), done.
```

```
$ cd test-babel-ember/
```

```
$ yarn
yarn install v1.13.0
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
[4/5] 🔗  Linking dependencies...
warning "ember-cli > @babel/plugin-transform-modules-amd@7.2.0" has unmet peer dependency "@babel/core@^7.0.0-0".
[5/5] 🔨  Building fresh packages...
✨  Done in 7.92s.
```

This may very well be a bug somewhere else, as I said I didn't take a lot of time to debug since I figured this was more of a typo fix than anything else.

danez:
Looking at `test-babel-ember` it is not having `@babel/core` in its `devDependencies` that is why you see the message.
This is a general problem with peerDependencies as they are required to be installed in the root package.json. In the case of ember probably one of the ember packages installs `@babel/core` but that does not satisfy the peerDependency, as it needs to be listed in the root package.json file.

Unfortunately the problem can not be solved by changing the peerDependencies.

nicolo-ribaudo:
This PR doesn't remove the warning:
```
yarn install v1.12.3
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
info fsevents@1.2.7: The platform "linux" is incompatible with this module.
info "fsevents@1.2.7" is an optional dependency and failed compatibility check. Excluding it from installation.
[4/5] Linking dependencies...
warning " > @babel/plugin-transform-modules-amd@7.2.0" has unmet peer dependency "@babel/core@^7.0.0".
[5/5] Building fresh packages...
Done in 2.77s.
```

I have tested it by adding this code to you repo's `package.json`, and it points to the package in this PR:
```json
   "resolutions": {
     "ember-cli/@babel/plugin-transform-modules-amd": "file:../babel/babel/packages/babel-plugin-transform-modules-amd"
   }
```

nicolo-ribaudo:
@danez Wouldn't it work if `ember-cli` adds `@babel/core` to its dependencies?

danez:
I do not think so.

jrjohnson:
Oh ok, sorry for the noise then. This isn't really a bug, just an annoying message so I'm more than comfortable closing it if that makes the most sense to you.

Issue #4694, opened 2016-10-07
babel-traverse, var declarations not included in binding references


kentcdodds:
### Input Code

``` js
var a = 32
var {c: b} = obj

console.log(a, b)
```
### ASTExplorer.net example

https://astexplorer.net/#/jUTCWsdHdx/3
### Description of Behavior

The `VariableDeclarator` `a` should be included in it's own `path.scope.getBinding(path.node.name).referencePaths`, just like the `VariableDeclarator` `b` which is an `ObjectProperty` in an `ObjectPattern`.
### Possible Solution

I'm not sure what is causing this, but for some reason, [this code](https://github.com/babel/babel/blob/33eb56ab09e5dcabbd135f3f34fc707c5ea3fbb9/packages/babel-traverse/src/scope/binding.js#L84-L91) isn't called for the `VariableDeclarator`'s `Identifier`.
### Context

I'm writing a tool that removes unused variables. For example:

``` js
function sortRankedItems(a, b) {
  var {rank: aRank} = a
  var bRank = b.rank

  var same = aRank === bRank
  return 'blah'
}
```

With my tool, would result in:

``` js
function sortRankedItems(a, b) {
  return 'blah'
}
```

Based on the order the plugins are running, when the visitor comes across `aRank` and `bRank`, it looks like those variables are in use, but when the visitor gets to `same`, it removes the node. Then it searches for the references that were used by `same` and determines whether it is safe to now remove those as well. Because of this bug, `bRank` is not able to be removed because it's not in the referencePaths, so I can't find it from the `same` node. Hopefully that makes sense.

danharper:
It's only `ObjectPattern` including itself in it's own `referencePaths`. [`ArrayPattern` also doesn't](https://astexplorer.net/#/jUTCWsdHdx/4).

It looks like it's intentional: https://github.com/babel/babel/blob/03d772c2ecedc95f87493e293977e69c3db94d56/packages/babel-types/src/validators.js#L156-L159. But `ObjectPattern` doesn't seem to be used in that link as [the parent](https://github.com/babel/babel/blob/03d772c2ecedc95f87493e293977e69c3db94d56/packages/babel-traverse/src/path/lib/virtual-types.js#L18) of the `Identifier` is `ObjectProperty`, not `ObjectPattern`).

So it looks it's actually a bug that `b` has 2 reference paths? Although I don't know much about this part of the code.

liuxingbaoyu:
Fixed in https://github.com/babel/babel/pull/9492

PR #9492, opened 2019-02-11
Mark FOO in "var { x: FOO }˝ as a binding, not as a reference

nicolo-ribaudo:
| Major: Breaking Change?  | Maybe? I'd consider this as a bug fix
| Minor: New Feature?      | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

Before this PR, babel incorrectly marks `FOO` as a `ReferencedIdentifier` in `var { a: FOO }` and as a binding identifier in `({ a: FOO })`: this is because we needed to know not only the parent but also its grandparent to know it. Array patterns don't have this problem, since there isn't something like an `ArrayElement` node between the identifier and the pattern.

This code shows the bug; you can test the old version on ASTExplorer (https://astexplorer.net/#/gist/3dc1c7fd69ee5f7478c2229e4bdce428/latest)

<details>
<summary>Plugin</summary>

```js
function ({ types: t, template }) {
  let result = "";
  return {
    visitor: {
      Identifier(path) {
        if (path.node.name === "key") return;

        result += path.node.name +
          " - referenced: " + path.isReferencedIdentifier() +
          ", binding: " + path.isBindingIdentifier() + "\n";
      },
      Program: {
        exit() {
          alert(result);
        }
      }
    },
  };
}
```

</details>

<details>
<summary>Input code</summary>

```js
var a = 1;
var [ b ] = 1;
var { key: c } = 1;

x;
[ y ];
({ key: z });
```

</details>

<details>
<summary>Old output</summary>

```
a - referenced: false, binding: true
b - referenced: false, binding: true
c - referenced: true, binding: true
x - referenced: true, binding: false
y - referenced: true, binding: false
z - referenced: true, binding: true
```
</details>
<details>
<summary>New output</summary>

```
a - referenced: false, binding: true
b - referenced: false, binding: true
c - referenced: false, binding: true
x - referenced: true, binding: false
y - referenced: true, binding: false
z - referenced: true, binding: false
```
</details>

### DIFF
--- a/packages/babel-plugin-proposal-object-rest-spread/src/index.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/src/index.js
@@ -2,6 +2,17 @@ import { declare } from "@babel/helper-plugin-utils";
 import syntaxObjectRestSpread from "@babel/plugin-syntax-object-rest-spread";
 import { types as t } from "@babel/core";
 
+// TODO: Remove in Babel 8
+// @babel/types <=7.3.3 counts FOO as referenced in var { x: FOO }.
+// We need to detect this bug to know if "unused" means 0 or 1 references.
+const ZERO_REFS = (() => {
+  const node = t.identifier("a");
+  const property = t.objectProperty(t.identifier("key"), node);
+  const pattern = t.objectPattern([property]);
+
+  return t.isReferenced(node, property, pattern) ? 1 : 0;
+})();
+
 export default declare((api, opts) => {
   api.assertVersion(7);
 
@@ -98,7 +109,7 @@ export default declare((api, opts) => {
     Object.keys(bindings).forEach(bindingName => {
       const bindingParentPath = bindings[bindingName].parentPath;
       if (
-        path.scope.getBinding(bindingName).references > 1 ||
+        path.scope.getBinding(bindingName).references > ZERO_REFS ||
         !bindingParentPath.isObjectProperty()
       ) {
         return;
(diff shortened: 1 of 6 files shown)


packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -2,6 +2,17 @@ import { declare } from "@babel/helper-plugin-utils";
...
+// TODO: Remove in Babel 8
+// @babel/types <=7.3.3 counts FOO as referenced in var { x: FOO }.
+// We need to detect this bug to know if "unused" means 0 or 1 references.
+const ZERO_REFS = (() => {
+  const node = t.identifier("a");
+  const property = t.objectProperty(t.identifier("key"), node);
+  const pattern = t.objectPattern([property]);
+
+  return t.isReferenced(node, property, pattern) ? 1 : 0;
+})();
```
nicolo-ribaudo: This is why this might be considered a breaking change: we were relying on the wrong behavior.

packages/babel-plugin-transform-modules-commonjs/test/fixtures/misc/import-const-throw/output.js
```diff
@@ -18,24 +18,32 @@ _baz.Baz = (44, function () {
   throw new Error('"' + "Baz" + '" is read-only.');
 }());
 ({
-  Foo: _foo.default
-} = {});
+  Foo
+} = ({}, function () {
+  throw new Error('"' + "Foo" + '" is read-only.');
+}()));
```
nicolo-ribaudo: This change is correct. You can see the input file at https://github.com/babel/babel/blob/93ec6ee62a10ea022c902962dc989f1045fb1637/packages/babel-plugin-transform-modules-commonjs/test/fixtures/misc/import-const-throw/input.mjs

```js
import Foo from "foo";

// ...

({Foo} = {});
```

nicolo-ribaudo:
With the compact mode it shouldn't break, but as we did rely on the wrong behavior someone else could.

danez:
@nicolo-ribaudo this fixes #4694 ?

Issue #6283, opened 2017-09-20
`var`s which conflicts with a `let` declared in an outer block aren't reported.


nicolo-ribaudo:
This issue only appears if the let declaration is inside a block. If it is a direct child of a function body/program, babel throws correctly.

  [REPL](https://babeljs.io/repl/build/4985#?babili=false&browsers=&build=&builtIns=false&code_lz=N4KABGA2CmAuYEMDc4zDANwQJ0UsAviAUA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=stage-2&prettier=false&targets=&version=7.0.0-beta.1)

### Input Code
  ```js
  {
    let a;
    { var a; }
  }
  ```

### Expected Behavior
  ```
  Duplicate declaration "a"
  ```

### Current Behavior
No error

### Possible Solution

`Scope` should check the block scopes that variable declarations visit while they "walk up" the scope.
It should also keep track of them, in the case that the variable declaration comes before the lexical declaration (e.g. `{ { var a; }; let a; }`)

### Your Environment

| software         | version(s)
| ---------------- | -------
| Babel            |  7.0.0-beta.1
| node             | 
| npm              | 
| Operating System |

arthurchipdean:
I would like to claim this bug. I have narrowed the issue down to the Declaration method in the CollectorVisitor on line 65 in scope/index.js

If I comment out the if statement at that line the bug an error is correctly thrown:
```
    // delegate block scope handling to the `BlockScoped` method
    if (path.isBlockScoped()) return;
```

I just have to figure out why that code is there and what effects removing it may have. ( I think the solution may be adding another condition to that if statement. ) I'm confident I'll be able to solve it with a bit more time though.

arthurchipdean:
Can I claim this bug? I have narrowed the issue down to the Declaration method in the CollectorVisitor on line 65 in scope/index.js

If I comment out the if statement at that line an error is correctly thrown:

    // delegate block scope handling to the `BlockScoped` method
    if (path.isBlockScoped()) return;

I think the solution will be some modification to the BlockScoped method?

jridgewell:
@arthurchipdean: it's yours.

liuxingbaoyu:
[...]

PR #9493, opened 2019-02-11
Introduce scope tracking in the parser

### LINKED ISSUE #4331: Move SyntaxErrors to Babylon from Babel (T7410)
> Issue originally made by Francisco (tolmasky)
### Description

There are a number of SyntaxErrors that are thrown at the transformation phase instead of the parsing stage. A non-exhaustive (I think?) list:
1. SyntaxError: unknown: super() is only allowed in a derived constructor
2. SyntaxError: unknown: 'this' is not allowed before super()
3. SyntaxError: unknown: missing super() call in constructor

etc.

I think moving as many syntax errors as possible to Babylon would be immensely helpful. It would allow editors to use babylon and provide inline warnings, etc (whereas currently they need to additionally run the full transformation solely to get the errors). Additionally, in AST analysis/transformations, you are kind of dealing with illegal syntax trees (in the same way that you wouldn't want super() in a normal method, only to be warned about later in transformation).  Currently we run into this issue a lot, as it would be very much more time consuming to be running transformation on every keypress, and so, as we find these, we manually add the syntax checks ourselves.

### LINKED ISSUE #6722: Duplicate definitions of variables should throw syntax error
> Issue originally reported by @danez in babel/babylon#78

According to the spec when defining multiple variables with the same name, the parser should throw and SyntaxError when using `let` or `const`:

`let foo, foo;`
or
`let foo; let foo;`

From [13.2.1.1](https://tc39.github.io/ecma262/#sec-let-and-const-declarations-static-semantics-early-errors):

> It is a Syntax Error if the BoundNames of BindingList contains any duplicate entries.

V8 trows an error for `let` and `const`, but not vor `var`.

danez:
| Fixed Issues?            |  Fixes #6722 (Fixes maybe #6283) Fixes #9316 Fixes #4331
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT (acorn MIT)

This introduces scope tracking in the parser. The scope tracking replaces some of the Flags we currently have (inGenerator, inFunction, ...).

Spec changes:

- We now report an syntax error if `super` is used even though there is no superClass
- Duplicate bindings will throw according to spec

The idea here is to enter and exit scopes, during parsing so we always can check the current scope. A scope consists of flags (which adds information we need in several places like SCOPE_FUNCTION, SCOPE_ASYNC, ...) and 3 collections of declared bindings in this scope (var, lexical(let, const, classes) and functions).

TODO:

- [x] Add tests for duplicate variable bindings.
- [x] Migrate some traverse tests (duplicate declarations) to not use the parser

Big thanks to acorn, as most of this here is ported from acorn. :) Same goes for copyright on most of it.


danez:
We have some tests that test the duplicate declaration checks of babel-traverse, but as the parser now throws first we are not getting to the point of testing duplicate declarations in traverse.

I see 3 options:

1. Leave the tests as is and don't care that we are not testing traverse.
2. Add an option to the parser to disable the duplicate check.
3. Move the tests to the parser (as thats what they are testing now) but also recreate new tests for traverse which do not use the parser but node builders instead to create the AST.


1 and 2 are not really good options for me. So I guess I will go for 3 which is the most work but the best I fell? Any other ideas on this?

nicolo-ribaudo:
:+1: for 3. We could ask help to the community to rewrite the `@babel/traverse` tests.

danez:
> 👍 for 3. We could ask help to the community to rewrite the `@babel/traverse` tests.

Turns out this are only 4 tests. Probably took longer to write the message above, than to fix them :D

packages/babel-plugin-transform-classes/src/transformClass.js
```diff
@@ -19,22 +19,6 @@ function buildConstructor(classRef, constructorBody, node) {
...
-      const { node, parentPath } = path;
-      if (parentPath.isCallExpression({ callee: node })) {
-        throw path.buildCodeFrameError(
-          "super() is only allowed in a derived constructor",
-        );
-      }
-    },
-  },
-]);
-
```
danez: Not sure if i should leave that or not. If this check would be in traverse I would leave it, but in the transform, the only valid usecase now would be someone creating in a plugin invalid code and then running the transform.

nicolo-ribaudo: I don't think that we need it: currently all the other early errors (except for the duplicated bindings errors) are only in `@babel/parser`.

packages/babel-parser/src/parser/statement.js
```diff
@@ -564,6 +576,7 @@ export default class StatementParser extends ExpressionParser {
     const cases = (node.cases = []);
     this.expect(tt.braceL);
     this.state.labels.push(switchLabel);
+    this.enterScope(0);
```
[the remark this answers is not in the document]
danez: makes sense

packages/babel-parser/src/parser/scope.js
```diff
@@ -0,0 +1,152 @@
...
+  functions: string[] = [];
+
+  constructor(flags: ScopeFlags) {
+    this.flags = flags;
+  }
+}
+
+// The functions in this module keep track of declared variables in the
+// current scope in order to detect duplicate variable names.
+export default class ScopeParser extends UtilParser {
```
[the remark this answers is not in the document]
danez: I changed it to be a separate class, which works and might be nice to have like this if we would want to reuse this parts in traverse for example.

But when looking into flow I noticed another problem: extending the scope tracking is not as easy possible with the external class. So for typescript we probably would want to add another binding group for types. Flow has some custom rules too when it comes to `declare class|var|function`. I'm not sure if we actually want to support extending the scope tracking but it might be nice.

danez:
Okay so I added fixes for flow:
- declare module {} has it's own scope now
- `type` and `opaque type` are now registered as lexical declaration, so the are handled the same as `let` or `const`

I also added test for TS to ensure types do not count as duplicate declarations of other lexical declarations.

Any further additions for flow and typescript need changes to the logic in how duplicate declarations are checked. But the problem is that there is no easy way to extend the scope handler, as any addition (for example a new scope group for typescript type declarations) will need adjustments to all other parts of the scope tracking logic.
The only way to achieve correct scope tracking for flow and ts is to check on initialization which language we are parsing and depending on that choose a corresponding ScopeHandler. This ScopeHandlers for JS, TS and flow ultimately will need to be copy and pasted and kept in sync for bug fixes as I do not see a way to easy overwrite a subset of the logic of the duplicate detection algorithm.

For now this PR does not introduce any duplicate checks for flow an ts, besides the ones mentions above, and we can follow up in a separate PR.

packages/babel-parser/src/parser/statement.js
```diff
@@ -954,59 +989,66 @@ export default class StatementParser extends ExpressionParser {
...
-    this.state.inClassProperty = false;
-    this.state.yieldPos = 0;
-    this.state.awaitPos = 0;
+    const isStatement = statement & FUNC_STATEMENT;
+    const isHangingStatement = statement & FUNC_HANGING_STATEMENT;
 
     this.initFunction(node, isAsync);
 
+    if (this.match(tt.star) && isHangingStatement) {
+      this.unexpected();
```
nicolo-ribaudo: Maybe we can have a better error message here?

Chrome throws these errors:
```
Generators can only be declared at the top level or inside a block.

Async functions can only be declared at the top level or inside a block.
```

packages/babel-parser/src/parser/util.js
```diff
@@ -3,7 +3,9 @@
 import { types as tt, type TokenType } from "../tokenizer/types";
 import Tokenizer from "../tokenizer";
 import type { Node } from "../types";
-import { lineBreak } from "../util/whitespace";
+import { lineBreak, skipWhiteSpace } from "../util/whitespace";
+
+const literal = /^(?:'((?:\\.|[^'])*?)'|"((?:\\.|[^"])*?)"|;)/;
```
nicolo-ribaudo: I took 5 mins to understand what is this lol

I think that it could be simplified a bit:
```js
/^(?:;|('|")((?:\\?.)*?)\1)/
```

1. You don't need the `[^"]` part, because you are already using `*?` so it stops as soon as possible (and the closing quote this is parsed as the end of the literal).
2. You can merge the `"` and `'` cases, using `\1` to refer to the matched quotes kind.
3. Since this adds a new capturing group, you would have to chech `match[2]`.

Demo: https://regex101.com/r/aQTDzs/1

danez: Thanks, that is way easier to read.

packages/babel-parser/src/parser/statement.js
danez: I updated them and some more error messages

PR #9516, opened 2019-02-14
Dedupe Babel in @babel/standalone

nicolo-ribaudo:

`@babel/standalone` is 30% smaller :sunglasses: 

:warning: In the first screenshot `node_modules` is on the left, while in the second one it is on the right.

Before (4.2 MB unminified, 2.3 MB minified):
schermata da 2019-02-14 23-24-04

After (2.8 MB unminified, 1.6 MB minified):

schermata da 2019-02-15 00-27-01

The problem was that those two packages "leaked" and imported the version of Babel used to build our code (so an older version, not the one from packages/)


danez:
Maybe we should add https://github.com/darrenscerri/duplicate-package-checker-webpack-plugin

Issue #9495, opened 2019-02-11
TypeError: Cannot read property 'filter' of undefined (While processing preset ...)


PavelKastornyy:
This is my project structure:

```
script4j
|
|-modules
|   |-script4j.base  
|   |-script4jfx.base	
|   |-script4jfx.graphics
```
Module `script4jfx.base` depends on one module - `script4j.base` and here everything is OK. Module `script4jfx.graphics` depends on two modules - `script4j.base` and `script4jfx.base` and here I have a problem, at the same time if I leave only one module dependency everything is OK.

This is a piece of `script4jfx.graphics` [package.json](https://github.com/PavelKastornyy/script4j/blob/master/modules/script4jfx.graphics/package.json)

```
    "devDependencies": {
        "gulp": "4.0.0",
        "gulp-mocha": "6.0.0",
        "babel-core": "6.26.3",
        "babel-register": "6.26.0",
        "babel-preset-env": "1.7.0",
        "babel-preset-stage-0": "6.24.1",
        "chai": "4.2.0",
        "mocha": "5.2.0",
        "script4j.base": "file:./../script4j.base",
        "script4jfx.base": "file:./../script4jfx.base"
    },
    "dependencies": {
        "script4j.base": "file:./../script4j.base",
        "script4jfx.base": "file:./../script4jfx.base"
   }
```
And here is what I get when try to run tests of `script4jfx.graphics`:

```
[18:31:10] Requiring external module babel-register
[18:31:10] Failed to load external module @babel/register
[18:31:11] Using gulpfile ~/script4j/modules/script4jfx.graphics/gulpfile.babel.js
[18:31:11] Starting 'test-module'...
/home/Pavel/script4j/modules/script4jfx.graphics/node_modules/babel-core/lib/transformation/file/options/option-manager.js:328
        throw e;
        ^
TypeError: Cannot read property 'filter' of undefined (While processing preset: "/home/Pavel/script4j/modules/script4jfx.base/node_modules/babel-preset-env/lib/index.js")
    at normalize (/home/Pavel/script4j/modules/script4jfx.base/node_modules/browserslist/index.js:12:19)
    at /home/Pavel/script4j/modules/script4jfx.base/node_modules/browserslist/index.js:737:17
    at Object.<anonymous> (/home/Pavel/script4j/modules/script4jfx.base/node_modules/browserslist/index.js:756:2)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
[...]

PR #9518, opened 2019-02-15
Use `for..of Object.keys` instead of `for..in`

### LINKED ISSUE #9511: Configuring babel.config.js for for testing modules with mocha
After reading all the information I still can't configure babel 7 to compile files from modules when using mocha. This is my project structure:
```
script4j
|
|-modules
|   |-script4j.base
|   |      |-node_modules
|   |      |-package.json
|   |-script4jfx.base
|   |      |-node_modules
|   |      |-package.json
|   |-script4jfx.graphics
|   |      |-node_modules
|   |      |-package.json
|   |      |-register.js
|-babel.config.js
```
As you see I have three modules. No one module has `.babelrc` file. Module `script4jfx.graphics` depends on `script4j.base` and `script4jfx.base`. This is a piece of package.json of `script4jfx.graphics`:

```
    "devDependencies": {
        "gulp": "4.0.0",
        "chai": "4.2.0",
        "mocha": "5.2.0",
        "@babel/core": "7.2.2",
        "@babel/register": "7.0.0",
        "@babel/preset-env": "7.3.1",
        "@babel/cli": "7.2.3",
        "babel-plugin-module-resolver": "3.2.0",
        "typescript": "3.3.3",
        "script4j.base": "file:./../script4j.base",
        "script4jfx.base": "file:./../script4jfx.base"
    },
    "dependencies": {
        "script4j.base": "file:./../script4j.base",
        "script4jfx.base": "file:./../script4jfx.base"
    }
```
This is `register.js`:

```
require("@babel/register")({
    rootMode: "upward",
});
```
This is my babel.config.js:

```
module.exports = function (api) {
    api.cache(true);
    console.log("Point XXX");//       <-- NOTE THIS LINE
    return {
        babelrcRoots: [
            '.',
            './modules/script4j.base',
            './modules/script4jfx.base',
            './modules/script4jfx.graphics'
        ],
	ignore: [/node_modules/],
        presets: ["@babel/preset-env"]
    };
}
```

This is the way I run mocha:
```
$ pwd
home/Pavel/script4j/modules/script4jfx.graphics

$./node_modules/mocha/bin/mocha ./tmp/script4jfx.graphics.spec.js --require register.js --compilers js:@babel/register
```
And this is what I get:
```
Point XXX
/home/Pavel/script4j/modules/script4j.base/dist/script4j.base.js:1191
export { Class, Integer, ... };
^^^^^^
SyntaxError: Unexpected token export
```
So, as you see `script4j.base` wasn't compiled by babel. How to fix it?

nicolo-ribaudo:
| Fixed Issues?            | Fixes #9511

In https://github.com/babel/babel/issues/9511 (and #9495 is another symptom), @PavelKastornyy reported a node crash becaue the JavaScript heap run out of memory. The problem was that their code was adding enumerable properties to `Object.prototype`: it is something that shouldn't be done, but Babel shouldn't make node crash if someone adds them.
I reduced down the problem to `for...in` loops in `@babel/traverse` that grew the memory consumption exponentially because of that unexpected properties.


packages/babel-helper-define-map/src/index.js
```diff
@@ -98,7 +99,7 @@ export function push(
 }
 
 export function hasComputed(mutatorMap: Object): boolean {
-  for (const key in mutatorMap) {
+  for (const key in own(mutatorMap)) {
```
loganfsmyth: Do we need the macro? It's pretty easy to do `for (const key of Object.keys(mutatorMap)) {`.

nicolo-ribaudo: I already had the macro in another project so I published it and added there, but I can remove it if you prefer.
My fear is that `for ... of + Object.keys` is slower than `for ... in + Object.hasOwnProperty`, and it is used in some perf-sensitive places in `@babel/traverse`.

nicolo-ribaudo: Nvm, I was wrong https://jsperf.com/for-in-vs-for-ok/1

packages/babel-traverse/src/visitors.js
```diff
@@ -23,7 +23,7 @@ export function explode(visitor) {
   visitor._exploded = true;
 
   // normalise pipes
-  for (const nodeType in visitor) {
+  for (const nodeType of Object.keys(visitor)) {
```
nicolo-ribaudo: @loganfsmyth Since this is a pulic api, should I handle the case where `visitor` is null/undefined (that `for..in` handled by default)?

loganfsmyth: Since we're doing `visitor._exploded` above we shouldn't need to worry about that case here.

PR #9520, opened 2019-02-15
Use NODE_PATHS environment value as extra module paths

### LINKED ISSUE #5618: Babel not working with NODE_PATH
This is a bug report - I am getting this error:
```

Module build failed: Error: Couldn't find preset "latest" relative to directory "/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman-browser-polyfills/modules"
    at /Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/options/option-manager.js:293:19
    at Array.map (native)
    at OptionManager.resolvePresets (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/options/option-manager.js:275:20)
    at OptionManager.mergePresets (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/options/option-manager.js:264:10)
    at OptionManager.mergeOptions (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/options/option-manager.js:249:14)
    at OptionManager.init (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/options/option-manager.js:368:12)
    at File.initOptions (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/index.js:212:65)
    at new File (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/index.js:135:24)
    at Pipeline.transform (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/pipeline.js:46:16)
    at transpile (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-loader/lib/index.js:46:20)
    at Object.module.exports (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-loader/lib/index.js:163:20)
 @ ./lib/index.js 136:9-22
 @ multi babel-polyfill ./lib/index.js

ERROR in .-browser-polyfills/modules/assert.js
Module build failed: Error: Couldn't find preset "latest" relative to directory "/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman-browser-polyfills/modules"
    at /Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/options/option-manager.js:293:19
    at Array.map (native)
    at OptionManager.resolvePresets (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/options/option-manager.js:275:20)
    at OptionManager.mergePresets (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/options/option-manager.js:264:10)
    at OptionManager.mergeOptions (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/options/option-manager.js:249:14)
    at OptionManager.init (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/options/option-manager.js:368:12)
    at File.initOptions (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/index.js:212:65)
    at new File (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/file/index.js:135:24)
    at Pipeline.transform (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-core/lib/transformation/pipeline.js:46:16)
    at transpile (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-loader/lib/index.js:46:20)
    at Object.module.exports (/Users/alexamil/WebstormProjects/oresoftware/sumanjs/suman/node_modules/babel-loader/lib/index.js:163:20)
 @ ./lib/index.js 131:13-30
 @ multi babel-polyfill ./lib/index.js
```

Webpack should be able to find the babel modules, because I am using NODE_PATH, which allows me to store Babel deps in a particular location, and I run webpack like so:

```bash
#!/usr/bin/env bash
NODE_PATH=${NODE_PATH}:~/.suman/global/node_modules webpack
```

I have this webpack config:

```js
const path = require('path');

module.exports = {

  entry: ['babel-polyfill', './lib/index.js'],

  output: {
    path: path.resolve(__dirname + '/dist'),
    filename: 'suman.js'
  },

  module: {
    loaders: [
      {
        test: /\.js$/,
        loader: 'babel-loader',
        exclude: /node_modules/,
        options: {
          presets: ['latest'],
          plugins: ['transform-runtime']
        }
      },
      {
        test: /(\/lib\/init\/|cli.js$)/,
        loader: 'ignore-loader'
      }
    ]
  },

  resolve: {
    alias: {
      fs: require.resolve('suman-browser-polyfills/modules/fs'),
      assert: require.resolve('suman-browser-polyfills/modules/assert')
    },
    extensions: ['.js']
  },

  node: {
    // assert: true,
    buffer: true,
    child_process: 'empty',
    cluster: 'empty',
    console: true,
    constants: true,
    crypto: 'empty',
    dgram: 'empty',
    dns: 'mock',
    domain: true,
    events: true,
    // fs: 'empty',
    http: true,
    https: true,
    module: 'empty',
    net: 'mock',
    os: true,
    path: true,
    process: true,
    punycode: true,
    querystring: true,
    readline: 'empty',
    repl: 'empty',
    stream: true,
    string_decoder: true,
    timers: true,
    tls: 'mock',
    tty: true,
    url: true,
    util: true,
    v8: 'mock',
    vm: true,
    zlib: 'empty',
  }
};
```

as you can see, in `resolve.alias`, we are referencing another project. No Babel dependencies are installed in

But that should not matter, the Babel deps needed are located in the directory referenced in NODE_PATH.

What's probably happening is that Webpack/Babel is doing a relative path lookup check for the needed Babel deps, but it should also check on NODE_PATH, if it needs to do a "pre-check".

This is a bug.

strongholdmedia:
| Fixed Issues?            | Fixes #5618
| Patch: Bug Fix?          | *Depends*
| Minor: New Feature?      | Yes *(sort of)*
| Documentation PR Link    | 

Added `resolve` `sync` argument `paths` (that is currently omitted) set to `NODE_PATH` environment variable
It is necessary to rebuild packages installed with `npm -g` without tampering with their `babel.config.js` (that may make for upstream incompatibility)
I suggest that the discussion proceed at #5618

### DIFF
--- a/packages/babel-core/src/config/files/plugins.js
+++ b/packages/babel-core/src/config/files/plugins.js
@@ -19,6 +19,8 @@ const OTHER_PLUGIN_ORG_RE = /^(@(?!babel\/)[^/]+\/)(?![^/]*babel-plugin(?:-|\/|$
 const OTHER_PRESET_ORG_RE = /^(@(?!babel\/)[^/]+\/)(?![^/]*babel-preset(?:-|\/|$)|[^/]+\/)/;
 const OTHER_ORG_DEFAULT_RE = /^(@(?!babel$)[^/]+)$/;
 
+const nodePaths = process.env("NODE_PATH").split(":");
+
 export function resolvePlugin(name: string, dirname: string): string | null {
   return resolveStandardizedName("plugin", name, dirname);
 }
@@ -96,14 +98,17 @@ function resolveStandardizedName(
   const standardizedName = standardizeName(type, name);
 
   try {
-    return resolve.sync(standardizedName, { basedir: dirname });
+    return resolve.sync(standardizedName, {
+      basedir: dirname,
+      paths: nodePaths
+    });
   } catch (e) {
     if (e.code !== "MODULE_NOT_FOUND") throw e;
 
     if (standardizedName !== name) {
       let resolvedOriginal = false;
       try {
-        resolve.sync(name, { basedir: dirname });
+        resolve.sync(name, { basedir: dirname, paths: nodePaths });
         resolvedOriginal = true;
       } catch (e2) {}
 
@@ -116,6 +121,7 @@ function resolveStandardizedName(
     try {
       resolve.sync(standardizeName(type, "@babel/" + name), {
         basedir: dirname,
+        paths: nodePaths
       });
       resolvedBabel = true;
     } catch (e2) {}
@@ -127,7 +133,10 @@ function resolveStandardizedName(
     let resolvedOppositeType = false;
     const oppositeType = type === "preset" ? "plugin" : "preset";
     try {
-      resolve.sync(standardizeName(oppositeType, name), { basedir: dirname });
+      resolve.sync(standardizeName(oppositeType, name), {
+        basedir: dirname,
+        paths: nodePaths
+      });
       resolvedOppositeType = true;
     } catch (e2) {}
 


packages/babel-core/src/config/files/plugins.js
```diff
@@ -19,6 +19,8 @@ const OTHER_PLUGIN_ORG_RE = /^(@(?!babel\/)[^/]+\/)(?![^/]*babel-plugin(?:-|\/|$
 const OTHER_PRESET_ORG_RE = /^(@(?!babel\/)[^/]+\/)(?![^/]*babel-preset(?:-|\/|$)|[^/]+\/)/;
 const OTHER_ORG_DEFAULT_RE = /^(@(?!babel$)[^/]+)$/;
 
+const nodePaths = process.env("NODE_PATH").split(":");
```
nicolo-ribaudo: On windows it uses `;`

strongholdmedia: Ah, thanks. The latter I figured in the meantime.
As for the former, I am in the woods. I do not use Windows for like 4 years now.
Anyway, is there a *preferred* way to detect platform?
Would `os.type().toLowerCase().indexOf("win") === 0` do? Or is there any global constant I should be aware of?

nicolo-ribaudo: You can use `.split(path.delimiter)` (https://nodejs.org/api/path.html#path_path_delimiter)

nicolo-ribaudo:
Two other things:
1. It's `NODE_PATH`, not `NODE_PATHS`
2. Can you add a test?

strongholdmedia:
**Well..**
I guess adding a test is a bit less straightforward, as it'd involve setting the environment variable to a known but empty folder, then copying a module to it, and finally create a process that uses that module that is, however, not added to the normal requirements (as surely enough it would get loaded from there).

The module should be a babel plugin, but it should not be installed into the "node_modules" folder (because the builtin node.resolve surely enough *does* use `NODE_PATH`).

I am uncertain if this needs change to the current build environment or not.

I was running into the issue when building a plugin with something called `rollup`, that in turn used `rollup-plugin-babel`, that in turn uses @ babel, that in turn tries to resolve `babel-plugin-emotion`, that failed for some obscure reason.
Then just before I went mad, I `strace`d it, and found that it absolutely didn't use any extra paths `npm` in turn did.
This is how all this got created, but I am fairly uncertain how to recreate that whole quagmire inside this `lerna` thing.

nicolo-ribaudo:
If needed I can add the test later, but I want to know other maintainers' opinion first.

packages/babel-core/src/config/files/plugins.js
```diff
@@ -19,6 +19,18 @@ const OTHER_PLUGIN_ORG_RE = /^(@(?!babel\/)[^/]+\/)(?![^/]*babel-plugin(?:-|\/|$
...
 
+const nodePaths =
+  typeof process.env.NODE_PATH !== "undefined"
+    ? process.env.NODE_PATH.split(
+      os
+        .type()
+        .toLowerCase()
+        .indexOf("win") === 0
+          ? ";"
+          : ":"
```
nicolo-ribaudo: You can use `.split(path.delimiter)` (https://nodejs.org/api/path.html#path_path_delimiter)

strongholdmedia: Ah.

strongholdmedia: Think I've also missed an import.

strongholdmedia:
I am unsure how to fix 
> property paths is missing in object type [1] but exists in object literal

but will look into it tomorrow.
I guess there is some type definition as in typescript, but I can't seem to be able to locate it within the project.
(On the other hand, it certainly exists in the resolve module, it is even "documented" (as in "mentioned").)

existentialism:
Re: 
> property paths is missing in object type [1] but exists in object literal

You'll want to update our flow definition for it here:
https://github.com/babel/babel/blob/master/lib/third-party-libs.js.flow#L7

strongholdmedia:
I did, hope exact objects will work as expected with optional arguments, as it seems there are quite a few issues in relation to such in flow.

anru:
100% helpful feature. Sometimes NODE_PATH is very convenient way to solve issue quickly and it allows save time and money

strongholdmedia:
I have since migrated to _buble_ in my own projects to avoid a lot of hassle and wasted performance (currently using _webpack_ for resolution - this may change though).

But given that _a lot of libraries_ are using _babel_, the very point I've made this PR is still valid - namely, avoiding the necessity to tamper with the source codes of third-party/"vendor" libraries.

phyllisstein:
I got bit by this setting up a Dockerized dev environment for a Babel-driven project. (Moving the container’s `node_modules` directory out from under the bind-mounted source directory was the only way I could think of to synchronize dependencies without sacrificing performance or clobbering the host’s bindings.) First I figured it was Gatsby, then I reluctantly admitted it might be Webpack, but ultimately Babel broke my heart.

Anyway, I’d be happy to take this PR over the finish line, if only to avenge my lost Friday evening. Have the maintainers signaled whether or not it’s likely to merge? /cc @nicolo-ribaudo

nicolo-ribaudo:
@babel/babel What's your opinion about this PR?

ljharb:
`NODE_PATH` is deprecated, won't be included in node's ESM, and shouldn't be used in CJS anyways. I'm a strong -1 on supporting it anywhere.

strongholdmedia:
Please note that the node TC voted explicitly against the deprecation of NODE_PATH.
So - unless I am missing something latter - it is not deprecated.

See here: https://github.com/nodejs/node/issues/1627

ljharb:
It may not be officially deprecated, in that it won’t ever be removed, but it will never exist for ESM, and the community widely considers it deprecated and a bad practice.

Typically, deps aren’t  anyways; and using NODE_PATHS doesn’t save you from bundling every time. Separately, there’s other projects like pnpm, yarn pnp, tink, etc that are addressing this - using a half-baked bad practice solution isn’t saving anyone time or energy in the long run.

strongholdmedia:
Surely not from bundling, but you may create a (or, rather, multiple) "global" (rather, per-user) repository and use that for "concurrent versioning" of the same library without _either_ downloading it to every single location every single time _or_ including it in the actual code base (and the related crap with using containers, because of colliding versions, separate builds, blah, blah).

Heck, every other freaking language and most runtimes (from Free Pascal to PHP, and surely enough, C, C#, and Java) has a sensible means to both share and run  libraries/includes/components/whatever side-by-side without requiring to keep multiple freaking copies of every single one.

_Curtailing this for a whole language on the basis of a scope-specific community being hypothesized to be having some sort of "practice consideration" sounds, to me, unwise at the very least._

ljharb:
It's already (going to be) curtailed for ESM in node using this mechanism; similarly, you don't have to use NODE_PATHS to wrap `require` and pull from a location of your choosing.

nicolo-ribaudo:
@strongholdmedia Note that you can _already_ use `NODE_PATHS`:
```js
// babel.config.js
module.exports = {
  plugins: [
    require ("a-globally-installed-plugin")
  ]
};
```

strongholdmedia:
Okay, please consider this.
I have an "app", with 22843 (recursive total, even duplicate) dependencies in the local "node_modules" folder, of which 5248 has a dependency of, say, "acorn".

_For the mere sake of simplicity_, let's suppose that there are 32 versions of "acorn" mentioned, _but_ this merely arises out of the fact of the latest version having been npm i-d _at the time of the package creation_, and in fact, _all reverse dependencies_ work with the latest, current version of acorn.

Now I figure that I _absolutely don't want to keep, herd and maintain 5248 "nested local" copies_ of acorn, let alone 32 versions, and all I want is to have a local copy in, say, "/usr/share" (where this kind of stuff actually belongs), or even "/home/%s/.npm_global" (as I currently do).

_How to tell all the 5248 packages to use that very same fixed copy of acorn_ in the given location _without_ altering worst case: 5248 "package.json" files?_

**I can do this with Free Pascal, gcc, clang, java/openjdk, and PHP** currently, either using compiler/linker flags, runtime/VM configuration parameters, or environment variables.

_I am currently using a patched copy of Babel only for the packages requiring Babel, and use Buble with custom overlaying for others, that actually works great, just not very portable._
But I absolutely think this is a more than realistic requirement.

Please mind that objectively _at least 50%_ of Babel users are bundling for browsers, not Node; also, many of them are using Babel because their dependencies do; and for them, all of this npm mayhem may already be an unjustified layer of insanity, so the whole "ESM" rationale may fail to resonate at all with them.

ljharb:
In that case you'd run `npm dedupe` and it would hoist all those local copies up to a single one at the top level, solving your problem.

If you're bundling, your bundler could be doing aliasing and deduplication, not node or babel.

strongholdmedia:
But wouldn't "dedupe" update the affected packages' `package.json`?
It is paramount that the source trees remain unchanged so that this wouldn't litter projects' working trees when used by e.g. collaborators.

"Sadly", it is even said that one should commit their lockfile, let alone the  package.json, so that when they are changed, it does constitute an update.

Moreover, this hinders testing a lot: you cannot just drop in a random checked-out-and-altered copy of a package in some random folder (as above) to find out that indeed the problem is caused by the approach you suspected (that before you even bother to open an issue that may turn up unfounded, let alone working on fixes and work-arounds).

---

Surely enough, it is not the job of babel to do "deduplication".
But I think dependency resolution should be kept consistent throughout a toolchain, and even if I am wrong, means should be provided for this.
**But then again, even if my bundler does the aliasing and "deduplication", it can only do as long as a third party dependency doesn't use babel resolver.** Otherwise the library builds as its author imagined it, and that cannot really be changed without again tampering with that library's metadata, that, in turn, is part of that library's source tree.

strongholdmedia:
**Or perhaps I am dumb and misunderstand what you all are saying**, and it is indeed possible for me to alter resolver options in my babel config in the app level _that will then be recursively applied to all the deps within nested node_modules folders, no matter what their configuration specifies._

If so (which I still doubt), then it is indeed a means (if problematic, since I still have to manually symlink/copy the deps into _my_ node_modules folders) to achieve what I actually refer to.
Please let me know if so.

nicolo-ribaudo:
You don't need to do anything manually. If you use `require` to load plugins (like in https://github.com/babel/babel/pull/9520#issuecomment-521399337), it will use Node's resolution algorithm which uses `NODE_PATHS`

strongholdmedia:
But I can't change plugins that use babel's resolve and not node resolve.
The original problem that caused me to work on this is that this works for some modules, and not for others (that use rollup, then rollup-plugin-babel, then babel-plugin-whatever-resolver, ...).

So what you say works for some modules, and not for others, depending solely on whether they use babel resolve plugins or not.

bdkjones:
It seems kinda nuts that this fix has been available for years, would help folks out, has no apparent downsides, and never shipped. Is this something 6 reviewers can rectify for those of us who would really benefit from this small change?

strongholdmedia:
@bdkjones I actually had to resort to spinning up a separate npm organization and shamelessly maintaining useless and conflicting forks of required packages.
Sad but this is where we stand now.

ljharb:
@bdkjones anything that does not aggressively discourage use of NODE_PATHS is a huge downside; it's deprecated and should not be used.

bdkjones:
1. As discussed above, it's not actually deprecated. 

2. It solves a very real problem in many edge cases, including mine.

3. There's no downside, other than pedantic computer-science arguments about how things SHOULD work because that is the One True Way(tm).

4. All of Babel is a slowly-rolling deprecation train. Each year, more and more of Babel becomes unnecessary as browsers evolve and old ones fall away. In that context, it seems silly to fixate on one thing and say, "NOPE. We'll support these 37 experimental proposals that will likely set your server on fire and will probably change 62 times before actual adoption, but BY GOD we will not have NODE_PATH!"

strongholdmedia:
@ljharb I am open to any other suggestion in which I can instruct a third-party module to use my provided version of **its own** dependency without actually modifying (thus having to maintain) it.

ljharb:
@strongholdmedia that's something the package manager needs to solve, not babel - yarn has resolutions, npm has aliases and will have overrides. Until then, *there shouldn't be a solution* and you should, indeed, need to fork/patch/maintain all the things you want to modify.

@bdkjones it's deprecated in the colloquial meaning of the term: https://github.com/babel/babel/pull/9520#issuecomment-521369591

strongholdmedia:
> @strongholdmedia that's something the package manager needs to solve, not babel

At the _package manager level_, it is solved. I can use NODE_PATHS to override packages - but I can also use `npm shrinkwrap` to manage that, as well as several other approaches.

In fact my primary complaint is based on this - babel resolver seems to ignore _all this_.
This is not quite much of a problem when it is me running anything across babel - but for packages doing their own build / packaging using it, I am still uncertain how would even `npm` `overrides` solve this.

ljharb:
@strongholdmedia NODE_PATHS isn't part of the package manager, it's part of node; i was referring to aliases/resolutions/overrides.

PR #9522, opened 2019-02-16
Make tests spec compliant by avoiding duplicate declarations in input files

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

This does NOT change any real code, only test fixtures are change.
This changes make sure there are no duplicate declarations in any of the `input.m?js` files. I extracted this from #9493 as it can be already added independently and makes the review of #9493 easier.

For reviewing here the important part is that I did not change the meaning of any test by renaming stuff. This is the reason `import type, { imp } from "";` was extracted into a separate file as it collides with `import type from "";`, but renaming `type` to something else would probably change the meaning.

There is some additional fixes see inline comments.


packages/babel-traverse/test/arrow-transform.js
```diff
@@ -65,7 +65,7 @@ describe("arrow function conversion", () => {
       _supercall();
       () => _supercall();
     `,
-      { methodName: "constructor" },
+      { methodName: "constructor", extend: true },
```
danez: I think `extend: true` was missing here and not left out on purpose.
super call requires the class to extend another class.

packages/babel-parser/test/fixtures/esprima/invalid-syntax/migrated_0108/input.js
```diff
@@ -1 +0,0 @@
-function null() { }
```
danez: Removed because duplicate
https://github.com/babel/babel/blob/master/packages/babel-parser/test/fixtures/core/uncategorised/401/input.js

packages/babel-parser/test/fixtures/esprima/invalid-syntax/migrated_0109/input.js
```diff
@@ -1 +0,0 @@
-function true() { }
```
danez: Removed because duplicate
https://github.com/babel/babel/blob/master/packages/babel-parser/test/fixtures/core/uncategorised/402/input.js

packages/babel-parser/test/fixtures/esprima/invalid-syntax/migrated_0110/input.js
```diff
@@ -1 +0,0 @@
-function false() { }
```
danez: Removed because duplicate
https://github.com/babel/babel/blob/master/packages/babel-parser/test/fixtures/core/uncategorised/403/input.js

packages/babel-parser/test/fixtures/esprima/invalid-syntax/migrated_0111/input.js
```diff
@@ -1 +0,0 @@
-function if() { }
```
danez: Removed because duplicate
https://github.com/babel/babel/blob/master/packages/babel-parser/test/fixtures/core/uncategorised/404/input.js

nicolo-ribaudo:
You renamed a lot of function declaration: it doesn't really matter, but they can be duplicated, right?

danez:
in some cases yes, i just renamed all duplication in tests that were failing

PR #9524, opened 2019-02-16
Fix typescript generator params

### LINKED ISSUE #7109: [babel-generator] doesn't output parens for TS types when needed
I'm not really sure if it's a bug or it's expected from Typescript's AST, but at least it's a bit unintuitive for someone manipulating the AST. Basically the issue is that babel generator is not putting parens for nested unions and/or intersections, obeying the nesting of the AST. 

The reason why I think this is confusing and not sure if it's a bug is that TS has a node called TSParenthesizedType that is the one that is generated from parsing nested union or intersection expressions, and also for array types when the type is composite (a union or intersection for example). If I manually add that node the code is generated as expected, but I'd imagine it's responsibility of the generator to know when to put parentheses based on the nesting in the AST or not. If this is indeed a bug I'm willing to fix it.

### Input Code

```js
import * as t from '@babel/types';
import generate from '@babel/generator';

console.log(generate(
  t.TSArrayType(
    t.TSUnionType([
      t.TSIntersectionType([
        t.TSNumberKeyword(),
        t.TSBooleanKeyword()
      ]),
      t.TSNullKeyword()
    ])
  )
).code);
```

None in particular, just generating from AST manually.

### Expected Behavior
This should be the output:

```ts
((number & boolean) | null)[]
```

### Current Behavior
The following code is outputted instead:

```ts
number & boolean | null[]
```

### Possible Solution
Maybe the solution is to add this case in the `parentheses.js` file under the `node` folder. Not sure what needs to be done logic-wise though, but I suspect it is there where the logic should exist.

### Context
Adding code generation to Apollo with `apollo-codegen` I was trying to generate the types for graphql array fields when the internal type is nullable, which requires to generate `(Type | null)[]` but without manually adding the ParenthesizedType node, this was not done by the generator. https://github.com/apollographql/apollo-codegen/blob/master/src/javascript/typescript/helpers.ts#L51

### Your Environment

| ---------------- | -------
| Babel-types            | 7.0.0-beta.35 
| Babel-generator            | 7.0.0-beta.35

tanhauhau:
| Fixed Issues?            | Fixes #7109
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-generator/src/node/parentheses.js
+++ b/packages/babel-generator/src/node/parentheses.js
@@ -127,6 +127,17 @@ export function TSTypeAssertion() {
   return true;
 }
 
+export function TSUnionType(node: Object, parent: Object): boolean {
+  return (
+    t.isTSArrayType(parent) ||
+    t.isTSNullKeyword(parent) ||
+    t.isTSIntersectionType(parent) ||
+    t.isTSUnionType(parent)
+  );
+}
+
+export { TSUnionType as TSIntersectionType };
+
 export function BinaryExpression(node: Object, parent: Object): boolean {
   // let i = (1 in []);
   // for ((1 in []);;);
--- a/packages/babel-generator/src/node/parentheses.js
+++ b/packages/babel-generator/src/node/parentheses.js
@@ -130,9 +130,10 @@ export function TSTypeAssertion() {
 export function TSUnionType(node: Object, parent: Object): boolean {
   return (
     t.isTSArrayType(parent) ||
-    t.isTSNullKeyword(parent) ||
+    t.isTSOptionalType(parent) ||
     t.isTSIntersectionType(parent) ||
-    t.isTSUnionType(parent)
+    t.isTSUnionType(parent) ||
+    t.isTSRestType(parent)
   );
 }
 
--- a/packages/babel-generator/src/generators/typescript.js
+++ b/packages/babel-generator/src/generators/typescript.js
@@ -231,7 +231,7 @@ export function tsPrintBraced(members, node) {
 }
 
 export function TSArrayType(node) {
-  this.print(node.elementType);
+  this.print(node.elementType, node);
   this.token("[]");
 }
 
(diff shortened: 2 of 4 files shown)


packages/babel-generator/src/node/parentheses.js
```diff
@@ -127,6 +127,17 @@ export function TSTypeAssertion() {
   return true;
 }
 
+export function TSUnionType(node: Object, parent: Object): boolean {
+  return (
+    t.isTSArrayType(parent) ||
+    t.isTSNullKeyword(parent) ||
```
nicolo-ribaudo: Why? `null` doesn't have any child.

tanhauhau: I was meaning `isTSOptionalType`, basically I copied the logic from `UnionTypeAnnotation() {}` above, is there a way to tell `isArrayTypeAnnotation` and `isTSArrayType`, `isIntersectionTypeAnnotation` and `isTSIntersectionType`  is similar, so don't have to rewrite the same logic again

nicolo-ribaudo: > is there a way to tell `isArrayTypeAnnotation` and `isTSArrayType`, `isIntersectionTypeAnnotation` and `isTSIntersectionType` is similar, so don't have to rewrite the same logic again

I don't think so: typescript and flow can diverge at any time and we don't have a way to compare their ASTs.

PS. You should also add `TSRestType`.

tanhauhau: OK, I've added `TSRestType` too, as well as test

tanhauhau:
> Maybe the tests are better suited as input/output fixtures?

But seems like when babel parses `(number & string)[]`, it adds `TSParenthesizedType` around the `TSIntersectionType`. thus unable to test this edge case of creating the AST without using `TSParenthesizedType`

PR #9532, opened 2019-02-18
Migrate some duplicate binding tests to traverse

danez:
| Major: Breaking Change?  | ?
| Documentation PR Link    | 
| Any Dependency Changes?  |

The main goal here was to transfer all the tests which test duplicate bindings. They were in different transforms, although the code was in `@babel/traverse` and the transforms have no impact on that behavior, and imho also shouldn't. 

While doing this I discovered, that on case wasn't throwing an error:

```js
let foo;
function foo(){};
```

see inline comment.

### DIFF
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -349,9 +349,6 @@ export default class Scope {
     // class expression
     if (local.kind === "local") return;
 
-    // ignore hoisted functions if there's also a local let
-    if (kind === "hoisted" && local.kind === "let") return;
-
     const duplicate =
       // don't allow duplicate bindings to exist alongside
       kind === "let" ||
--- a/packages/babel-traverse/test/scope.js
+++ b/packages/babel-traverse/test/scope.js
@@ -328,46 +328,79 @@ describe("scope", () => {
     });
 
     describe("global", () => {
-      ["let", "const"].forEach(name => {
-        describe(name, () => {
-          it("class", () => {
-            const ast = [
-              t.variableDeclaration(name, [
-                t.variableDeclarator(t.identifier("foo")),
-              ]),
-              t.classDeclaration(t.identifier("foo"), null, t.classBody([])),
-            ];
+      // node1, node2, success
+      // every line will run 2 tests `node1;node2;` and `node2;node1;`
+      // unless node1 === node2
+      const cases = [
+        ["const", "let", false],
 
-            expect(() => getPath(ast)).toThrowErrorMatchingSnapshot();
-          });
-          it("function", () => {
-            const ast = [
-              t.variableDeclaration(name, [
-                t.variableDeclarator(t.identifier("foo")),
-              ]),
-              t.functionDeclaration(
-                t.identifier("foo"),
-                [],
-                t.blockStatement([]),
-              ),
-            ];
+        ["const", "const", false],
+        ["const", "function", false],
+        ["const", "class", false],
+        ["const", "var", false],
 
-            expect(() => getPath(ast)).toThrowErrorMatchingSnapshot();
-          });
-          it("var", () => {
-            const ast = [
-              t.variableDeclaration(name, [
-                t.variableDeclarator(t.identifier("foo")),
-              ]),
-              t.variableDeclaration("var", [
-                t.variableDeclarator(t.identifier("foo")),
-              ]),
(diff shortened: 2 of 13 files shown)


packages/babel-traverse/src/scope/index.js
```diff
@@ -349,9 +349,6 @@ export default class Scope {
     // class expression
     if (local.kind === "local") return;
 
-    // ignore hoisted functions if there's also a local let
-    if (kind === "hoisted" && local.kind === "let") return;
-
```
danez: For some reason we have this exception in the code for the case mentioned in the PR. I couldn't come up with a reason why we do this and the commit that introduced it doesn't help either 8a4296a948b496e9fde1d6271a4e0dbdeaacb128
Simply removing it also doesn't break any of the tests. 
Do you think it is safe to remove? Anyone has an idea why this code was there and which case it covers?

nicolo-ribaudo: Maybe it was this? It is an Annex B thing:
```js
let a;
{ function a() {} }
```

packages/babel-traverse/test/scope.js
```diff
@@ -250,4 +271,84 @@ describe("scope", function() {
...
+          const ast = createTryCatch(name);
+
+          expect(() => getPath(ast)).toThrowErrorMatchingSnapshot();
+        });
+      });
+
+      it("var", () => {
+        const ast = createTryCatch("var");
+
+        expect(getPath(ast).node).toMatchSnapshot();
```
nicolo-ribaudo: Nit: `generator(getPath(ast).node)` would make the snapshot output easier to check.

packages/babel-traverse/src/scope/index.js
danez: right but even with removing these lines, this case still works. will add a test.

packages/babel-traverse/test/scope.js
danez: I was thinking exactly the same, I guess what I'm really testing here is that it does not throw. So `expect().not.toThrow()` makes sense to me.

nicolo-ribaudo:
Maybe it's wort using `.toThrowErrorMatchingInlineSnapshot` instead of `toThrowErrorMatchingSnapshot`, since the error messages are small?

danez:
Oh I have never seen this. makes sense.

danez:
Nevermind this does not work obviously with loops and generated testcases.

packages/babel-traverse/test/scope.js
```diff
@@ -250,4 +271,103 @@ describe("scope", function() {
...
+            ),
+          ]),
+        ];
+
+        expect(() => getPath(ast)).not.toThrow();
+      });
+    });
+
+    describe("global", () => {
+      ["let", "const"].forEach(name => {
```
nicolo-ribaudo: We should also test `class/function` and `class/var` (and maybe `function/var` to be sure that it doesn't throw)

danez: Okay I added a more generic way to define tests.
Turns out we have some problems, unless I made a mistake in the tests.
```
var foo = 1;
class foo {};
```
and
```
class foo {};
var foo = 1;
```
Class and var never throw.

also does not throw:

```js
function foo() {}
const foo = 1;
```

```js
var foo = 1;
const foo = 1;
```

Thats why I disabled the reverse matching in the test.

danez:
It seems the scope tracking in traverse has some flaws. I guess we should have a look separately from this. I can maybe work on this once done with the scope tracking in the parser.
I was also already thinking that maybe we can make a shared module that handles scopes correctly and share it between the parser and traverse. That would be nice, but not sure what requirements the scope tracking in traverse has.

nicolo-ribaudo:
Yeah let's just update the tests for now.

> I was also already thinking that maybe we can make a shared module that handles scopes correctly and share it between the parser and traverse

traverse's scope tracker needs to be kept updated while parser's scope describes a static AST; I don't know how feasible it would be.

danez:
> traverse's scope tracker needs to be kept updated while parser's scope describes a static AST; I don't know how feasible it would be.

But in the end both implementations (parser and traverse) work on nodes, have a stack of scopes, add/remove bindings and check for duplicates. But yes there might be more stuff going on in traverse which I'm not yet aware of.

PR #9534, opened 2019-02-18
TypeScript Constant contexts

### LINKED ISSUE #9502: TypeScript 3.4 meta-issue
TypeScript 3.4 will be introducing a number of syntactic additions in the type space as well as the expression space.

* [ ] The `readonly` type modifier for array types and tuple types. (#9503)
    * https://github.com/Microsoft/TypeScript/pull/29435
* [ ] Constant contexts (a.k.a. `as const`)
    * https://github.com/Microsoft/TypeScript/pull/29510
    * This seems to already work?

Our target date for the release of TypeScript 3.4 is around the end of March 2019.

tanhauhau:
| Fixed Issues?            | Fixes #9502
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-parser/test/fixtures/typescript/cast/as-const-2/input.js
```diff
@@ -0,0 +1,3 @@
+let e1 = v1 as const;  // Error
+let e2 = (true ? 1 : 0) as const;  // Error
+let e3 = id(1) as const;  // Error
```
danez: For tests that contain multiple cases that all should error, they should be split up in multiple files. Otherwise the parser throws on the first line and never even parsers any consequent line.

tanhauhau: noted. i've split them into separate files

danez:
Looks very promising, let us know when you are done.

tanhauhau:
@danez it's ready for code review 😄

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -319,6 +319,30 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+        case "TemplateLiteral":
+        case "NumericLiteral":
+        case "BooleanLiteral":
+        case "ObjectExpression":
+        case "ArrayExpression":
+          return;
+        case "UnaryExpression":
+          return this.tsCheckLiteralForConstantContext(node.argument);
+        default:
+          throw this.unexpected(node.start);
```
nicolo-ribaudo:
```suggestion
          this.raise(node.start, "Only literal values are allowed in constant contexts");
```

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -319,6 +319,30 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+    }
+
+    tsCheckLiteralForConstantContext(node: N.Node) {
+      switch (node.type) {
+        case "StringLiteral":
+        case "TemplateLiteral":
+        case "NumericLiteral":
+        case "BooleanLiteral":
+        case "ObjectExpression":
+        case "ArrayExpression":
```
nicolo-ribaudo: For arrays and objects, don't we need to check their children?

e.g.
```js
[ foo() ] as const
```

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -935,12 +959,13 @@ export default (superClass: Class<Parser>): Class<Parser> =>
 
     tsParseTypeAssertion(): N.TsTypeAssertion {
       const node: N.TsTypeAssertion = this.startNode();
-      this.next(); // <
-      // Not actually necessary to set state.inType because we never reach here if JSX plugin is enabled,
-      // but need `tsInType` to satisfy the assertion in `tsParseType`.
-      node.typeAnnotation = this.tsInType(() => this.tsParseType());
+      const _const = this.tsTryNextParseConstantContext();
+      node.typeAnnotation = _const ? _const : this.tsNextThenParseType();
```
nicolo-ribaudo:
```suggestion
      node.typeAnnotation = _const || this.tsNextThenParseType();
```

packages/babel-parser/src/plugins/typescript.js
tanhauhau: updated and added new test case

packages/babel-parser/src/plugins/typescript.js
```diff
@@ -319,6 +319,44 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+        case "TemplateLiteral":
+        case "NumericLiteral":
+        case "BooleanLiteral":
+        case "SpreadElement":
+        case "ObjectMethod":
+        case "ObjectExpression":
+          return;
+        case "ArrayExpression":
+          return (node: N.ArrayExpression).elements.forEach(element => {
+            if (element) {
```
Andarist: when element might be missing? 🤔

nicolo-ribaudo: Maybe the second element of `[foo, ,bar]`?

Andarist: Seems like the reason 👌

packages/babel-parser/test/fixtures/typescript/cast/as-const/input.js
```diff
@@ -0,0 +1,42 @@
...
+let a9 = [...a8];
+
+let o1 = { x: 10, y: 20 } as const;
+let o2 = { a: 1, 'b': 2, ['c']: 3, d() {}, ['e' + '']: 4 } as const;
+let o3 = { ...o1, ...o2 } as const;
+let o4 = { a: 1, b: 2 };
+let o5 = { ...o4 } as const;
+let o6 = { ...o5 };
+let o7 = { ...d } as const;
+let o8 = { ...o7 };
```
nicolo-ribaudo: What are these tests without `as const` needed for? I think that you copied them from https://github.com/Microsoft/TypeScript/blob/08fe06f5275bae91a2b2d7eb6387218b95ce3566/tests/baselines/reference/constAssertions.js, but since Babel doesn't type-check they aren't needed.

nicolo-ribaudo:
Some tests are failing on travis/circle because of regenerator, but they will work if you rebase this PR on master.

packages/babel-parser/test/fixtures/typescript/cast/as-const/input.js
tanhauhau: Updated the test cases

tanhauhau:
@nicolo-ribaudo cool.

nicolo-ribaudo:
You also need to update the output 😛

PR #9542, opened 2019-02-19
Add placeholders support to @babel/types and @babel/generator

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  |

Related to #9364, but those two PRs can be merged in any order.

The "core" of this PR is at https://github.com/babel/babel/pull/9542/files#diff-70f395c5d000299ece93fa4d13b656f9R18: does it make sense? That same logic is replicated for the `is[Type]` helpers.


packages/babel-generator/test/fixtures/misc/.placeholders/input.js
```diff
@@ -0,0 +1,8 @@
+var %%a%% = %%b%%
+
+%%c%%
+
+class %%d%% {}
+class A %%e%%
+
+function %%f%%(...%%g%%) %%h%%
```
danez: None of the tests added a semicolon. But there should be cases that do, if I look at the code above?

nicolo-ribaudo: The semicolon has been added after the variable (but it has been inserted by the `VariableDeclaration` printer) and after the statement placeholder (by the placeholder printer).
It hasn't been added after function/class declaration because they are never followed by a semicolon.

danez: Oppss I looked at the wrong file.

nicolo-ribaudo:
Yeah, 99% of the times they will only be used by `@babel/template`. But I think that we should add them to `@babel/types` for two reasons:
1. Every other node is defined in `@babel/types`, even nodes that you will almost never use (e.g. `t.isFile`).
2. Even if they will mostly be used inside `@babel/template`, there are users which want to use placeholders in custom transforms. For example, point 3 of https://github.com/babel/babel/issues/9112#issuecomment-453716260

existentialism:
%%:+1:%%

PR #4986, opened 2016-12-10
Plugin to transform dynamic import using the spec mode require wrapper

Jessidhia:
| Q                 | A 
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     |
| License           | MIT
| Doc PR            |
| Dependency Changes| 

This transforms dynamic `import()` calls into `Promise`-wrapped `require` calls that go through the commonjs `spec: true` import helper.

Per the [latest TC39 meeting](https://github.com/rwaldron/tc39-notes/blob/master/es7/2016-11/dec-1.md#14ii-import-open-issues-and-stage-3-discussion), the resulting `Promise` always runs on the next tick (always asynchronously).

This continues off another PR so the diff is huge for now 😅 

*Input*

```js
async function x () {
  await import(`some${'module'}`)
}
```

*Output*

```js
async function x () {
  await Promise.resolve().then(() => _specRequireInterop(require(`some${'module'}`)));
}

function _specRequireInterop(obj) {
  // return obj if obj.__esModule, wrap in module with single default export otherwise
}
```

PR #9552, opened 2019-02-20
Create @babel/plugin-proposal-dynamic-import

### LINKED ISSUE #9551: Dynamic import transform
We currently only support _parsing_ of `import(foo)`, and ask people to use wither webpack or `babel-plugin-dynamic-import-node` to transpile it.

I think that we can easily add it to `@babel/plugin-transform-modules-*` (or create new `@babel/plugin-proposal-dynamic-import-*` packages):

```js
// input
import(source)

// commonjs
new Promise(resolve => resolve(_interopRequireWildcard(require(source))))

// amd
new Promise(resolve => require([source], imports => resolve(_interopRequireWildcard(imports))))

// systemjs
System.import(source)
```

My preference would be to add this feature (behind an option) to the existing plugins, since we can't transpile `import()` without also transpiling es6 `import/export`.
If you prefer the separte plugins, we could add an "internal option" to `@babel/plugin-transform-modules-*` which is enabled by `@babel/plugin-proposal-dynamic-import`. If the modules plugin is not enabled, `@babel/plugin-proposal-dynamic-import` would throw an error.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #9551
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is an implementation of what I proposed in #9551. I had two goals:
1) It shouldn't be possible, for example, to transform dynamic imports using commonjs and static imports using amd
2) Since dynamic import is still a proposal, it needs not to be enabled by default by the module transform plugins

Note: sysntemjs was already supported and enabled by default, but this PR merged it's entry point with the one for commonjs/amd.

#4986 is similar, but it only implements CommonJS support and introduces a `spec: true` mode for commonjs, using different helpers.


packages/babel-plugin-transform-modules-commonjs/src/index.js
```diff
@@ -10,9 +10,16 @@ import {
...
 import { template, types as t } from "@babel/core";
 
+import babelPluginDynamicImportNode from "babel-plugin-dynamic-import-node";
+
 export default declare((api, options) => {
   api.assertVersion(7);
 
+  // TODO: expose a better interface
+  const transformImportCall = Function.call.bind(
+    babelPluginDynamicImportNode(api).visitor.Import,
```
nicolo-ribaudo: Please ignore the ugliness of this line. If this PR gets approved, I will open a PR in the `babel-plugin-dynamic-import-node` repo to directly export the `transformImportCall` function.

xtuc:
As we did for ESM, we should force to preserve the import function when using Babel loader in webpack.

packages/babel-plugin-transform-modules-systemjs/src/index.js
```diff
@@ -21,6 +21,12 @@ const buildExportAll = template(`
   }
 `);
 
+const MISSING_PLUGIN_WARNING = `\
+WARNING: Dynamic import() transformation must be enabled using the
+         @babel/plugin-proposal-dynamic-import plugin. As of Babel 8,
+         without that plugin it won't be transformed.
```
Jessidhia: "Babel 8 will no longer transform import() without using that plugin."

packages/babel-plugin-proposal-dynamic-import/test/fixtures/amd/module/output.js
```diff
@@ -0,0 +1,5 @@
+define(["require"], function (_require) {
+  "use strict";
+
+  var modP = new Promise(_resolve => _require(["mod"], imported => _resolve(babelHelpers.interopRequireWildcard(imported))));
```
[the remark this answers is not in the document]
nicolo-ribaudo: I checked the require source, and apparentely the callback is always called after a `context.nextTick`:
https://github.com/requirejs/requirejs/blob/fdf4186d3e68df06a04bd71cb6ea0f24eb1600d1/require.js#L1460

It is always guaranteed to be async: https://github.com/requirejs/requirejs/issues/738#issuecomment-17708061

Reading RequireJS's source code made me realize that it can also generate an error.

nicolo-ribaudo:
Updated to use https://github.com/airbnb/babel-plugin-dynamic-import-node/pull/75

PR #9554, opened 2019-02-20
 Do not always print newlines around block comments

### LINKED ISSUE #9547: Spurious newlines are added around block comments
## Bug Report

**Current Behavior**
The code printer adds a newlines before and after block comments

**Input Code**

Input:

```js
f(/* hi */ b);
```

Output:

```js
f(
  /* hi */
  b
);

```

https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYCg9AVABAFgllCYoCMCUBuIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=true&sourceType=module&lineWrap=false&presets=es2016%2Ces2017%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3&prettier=true&targets=&version=7.3.3

```js
function f() {
  return /* bye */ 42;
}
```
Output:

```js
function f() {
  return (
    /* bye */
    42
  );
}
```

https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABMAFASkQbwFCMQJwFMoR8kB6AKkQCMBPQxS8xAFgCYBubAXyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=true&sourceType=module&lineWrap=false&presets=es2016%2Ces2017%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3&prettier=true&targets=&version=7.3.3

**Expected behavior/code**
I expect there to be no extra newlines added when the block comment is attached to an expression. Only add newlines when the comments are attached to statements.

**Environment**
- Babel version(s): 7.3.3

arv:
| Fixed Issues?            | Fixes #9547
| Patch: Bug Fix?          | Bug fix

When printing block comments (`/* comment */`) we used to always insert a newline before and after. This is not desired when the comment is *attached* to an expression node.

This changes the code to format the whitespace around these comments slightly to try to keep the comment inline with the expressions surrounding it.


arv:
I split this into two commits to make it easier to review. Second commit is only updates to output files

packages/babel-generator/test/fixtures/comments/block-comment-expressions/output.js
```diff
@@ -0,0 +1,9 @@
+fun(/* a */ a);
+
+function fun(/* a */ a, /* b */ b) {
+  return /* c */ c;
+}
+/** d */
+
+
```
nicolo-ribaudo: Too many newlines? :thinking: IIRC, we had some logic to avoid pushing more than two `\n`s

arv: yeah that looks ugly. Let me see if this is a regression and fix it if it is.

arv:
The failure looks unrelated. I will look into it though.

danez:
What is your problem/usecase? We tried to minimize the logic inside babel-generator. With babel 7 we remove a lot of special case handling like this for performance reason and easier code readability.

If this produces an error somewhere then we have to fix it, but if this is just for stylistic reasons, then I would be on the fence if we actually should do this as our goal is not necessarily to produce human readable code. If the output of babel should be readable it should be piped through prettier.

arv:
The main motivation was the following input code:

```js
return /** @type {T} */ (x);
```

which always gets printed as:

```js
return (
  /** @type {T} */
  (x)
);
```

danez:
> The main motivation was the following input code:

I understand what you are describing, but not what the problem is with printing i the way `@babel/generator` is printing it. Do you get an error message somewhere? Does some other tool not understand this if it is multiline?

arv:
Motivation: We are using a pipeline based on Babel to convert Google Closure Compiler JavaScript to Typescript and we are trying to minimize the diffs this pipeline introduces.

danez:
Have you had a look at https://github.com/codemod-js/codemod? It uses babel plugins to change code, but uses `recast` by default to reprint the source. `recast`s goal is to try to reprint source from an AST as close as possible to the input source. So it is probably exactly what you want for automatic code changes.

As I also said before it is not our goal here to print code as close as possible to the input or as beautiful as possible. The use case we are aiming for is printing it correctly and as fast as possible to be then executed by a machine or be fed into another tool (like a minifier).

So even if we fix this case, there might be a lot more cases coming up where you would run into the same problem that something is printed in a weird/not nice (but working) way. So you might have more success with the tool above in my opinion.

arv:
Thanks

PR #9559, opened 2019-02-21
Don't watch files that are passed to `--ignore`.

dahjelle:
I was hitting an error message of "System limit for number of file watchers reached", and it turns out that this was because I was doing something like:

```
npx babel . --out-dir=../es5 --watch --ignore="node_modules"
```

But, since `--ignore` was not getting passed to `chokidar`, it was trying to watch the entire `node_modules` directory!

---

I expect this could use a test or two, but it isn't clear to me how. The existing tests appear to check stderr, but what needs to be tested is the chokidar options. Suggestions?

| Tests Added + Pass?      | no (suggestions for tests welcome!)

### DIFF
--- a/packages/babel-cli/src/babel/dir.js
+++ b/packages/babel-cli/src/babel/dir.js
@@ -135,6 +135,7 @@ export default async function({ cliOptions, babelOptions }) {
 
     filenames.forEach(function(filenameOrDir) {
       const watcher = chokidar.watch(filenameOrDir, {
+        ignored: babelOptions.ignore,
         persistent: true,
         ignoreInitial: true,
         awaitWriteFinish: {


danez:
Thanks for creating this PR.

This seems plausible to me. We just have to look into detail if all cases of the ignore babel option are compatible to chokidars ignored option. (relative paths vs. absolute paths, not sure what chokidar expects and what our option internally is)

dahjelle:
Good call! I found where (I believe) [`babel` is using the `--ignore` parameters](https://github.com/dahjelle/babel/blob/879a59085b4c6bf7fae744abe758774047427a58/packages/babel-core/src/config/config-chain.js#L200), and it looks like [it is just using RegExp and not globs](https://github.com/dahjelle/babel/blob/879a59085b4c6bf7fae744abe758774047427a58/packages/babel-core/src/config/config-chain.js#L670-L694) for matching.

`chokidar`, on the other hand, uses an [anymatch](
https://github.com/es128/anymatch) definition, which is glob-based.

image

So, you are right. This won't work. Closing in favor of an issue.

dahjelle:
Replace the PR with an issue: https://github.com/babel/babel/issues/9565

epicfaace:
> Good call! I found where (I believe) [`babel` is using the `--ignore` parameters](https://github.com/dahjelle/babel/blob/879a59085b4c6bf7fae744abe758774047427a58/packages/babel-core/src/config/config-chain.js#L200), and it looks like [it is just using RegExp and not globs](https://github.com/dahjelle/babel/blob/879a59085b4c6bf7fae744abe758774047427a58/packages/babel-core/src/config/config-chain.js#L670-L694) for matching.

Is that actually the case? It seems like babel's --ignore does allow for globs -- see https://github.com/babel/website/pull/1818/files

gonzarodriguezt:
I just tried this solution and it actually worked! @dahjelle As chokidar's`ignored` seems to allow regExps as well as babel `--ignore`, or there is another caveat?

dahjelle:
@gonzarodriguezt I can't say I looked into it much further…as @danez pointed out, there are quite a few cases that need to be checked to make are that the options in `chokidar` and `babel` will do the same thing, otherwise things will be even more confusing than they are now.

This solution worked for my particular case, too, but (for reasons I don't remember) I ended up writing my own watch script anyways.

gonzarodriguezt:
You're right, and that's good to know as well, thanks for answering! @dahjelle 

I will try to test it a little and see if i can get to something stable out of it

PR #9561, opened 2019-02-22
Update CHANGELOG.md using the "Trigger GitHub release" action

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

We haven't been very good at keeping `CHANGELOG.md` updated (2 days ago I added the changelog for versions 7.2.2-7.2.3 21eb0837e8d367b2aa0d2ecffed5ec648f9e0bbc, and the previus commit was for versions 7.1.3-7.2.1 b60adce4cb4e0c705cb308b6eacf36605ed61db8).

This PR updates the action we use to generate the GitHub release so that it also updates `CHANGELOG.md` (commit author: @babel-bot :smile:). Btw, I'm trying to understand how I can test this action which chages the git tree and makes requests to the GH api, but I'm not doing it in this PR since it will take a lot of time to properly set up.

### DIFF
--- a/.github/actions/trigger-github-release/entrypoint.sh
+++ b/.github/actions/trigger-github-release/entrypoint.sh
@@ -5,21 +5,18 @@ set -e
 echo "INFO: Installing action dependencies..."
 (cd /action; npm ci)
 
-echo "INFO: Checking out current commit..."
-git -c advice.detachedHead=false checkout $GITHUB_SHA
-
 # GitHub doesn't support running actions on new tags yet: we need to run it on the commit.
 # For this reason, we can't be sure that the tag already exists. We can use the commit
 # message to create the tag. If the tag already exists locally, they won't conflict because
 # they have the same name and are on the same commit.
 echo "INFO: Getting release version..."
-# current_tag=$(git describe --abbrev=0 --tags HEAD)
-current_tag=$(git log --oneline --format=%B -1 HEAD)
+# current_tag=$(git describe --abbrev=0 --tags $GITHUB_SHA)
+current_tag=$(git log --oneline --format=%B -1 $GITHUB_SHA)
 
 echo "INFO: Creating new tag..."
 (git tag $current_tag $GITHUB_SHA) || echo "INFO: Tag already exists"
 
-last_tag=$(git describe --abbrev=0 --tags HEAD^)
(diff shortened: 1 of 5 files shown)


.github/main.workflow
```diff
@@ -7,6 +7,11 @@ action "Trigger GitHub release" {
   uses = "./.github/actions/trigger-github-release/"
   secrets = ["GITHUB_TOKEN"]
 
+  env = {
+    COMMIT_AUTHOR_NAME  = "Babel Bot"
+    COMMIT_AUTHOR_EMAIL = "babel@hopeinsource.com"
```
xtuc: shoudln't this be bot@babeljs.io?

.github/actions/trigger-github-release/entrypoint.sh
```diff
@@ -5,21 +5,18 @@ set -e
...
-git -c advice.detachedHead=false checkout $GITHUB_SHA
-
 # GitHub doesn't support running actions on new tags yet: we need to run it on the commit.
 # For this reason, we can't be sure that the tag already exists. We can use the commit
 # message to create the tag. If the tag already exists locally, they won't conflict because
 # they have the same name and are on the same commit.
 echo "INFO: Getting release version..."
-# current_tag=$(git describe --abbrev=0 --tags HEAD)
-current_tag=$(git log --oneline --format=%B -1 HEAD)
+# current_tag=$(git describe --abbrev=0 --tags $GITHUB_SHA)
```
xtuc: could you remove this line since its commented?

.github/main.workflow
nicolo-ribaudo: I don't know if it exists? `babel@hopeinsource.com` is the mail in the @babel-bot account config, so it works to link the commit to that account.

.github/actions/trigger-github-release/entrypoint.sh
nicolo-ribaudo: The commented line is the "correct" way to get the last release, but it will only work when this action is fired when the new tag is pushed. Currently GitHub only supports this for private repos.

PR #9573, opened 2019-02-24
Update esm and electron mappings in preset-env

existentialism:
Guess I was too slow for @zloirock :)

Jokes aside, my original PR had a few other changes that are worth landing, so here they are post-rebase!

### DIFF
--- a/packages/babel-preset-env/data/built-ins.json
+++ b/packages/babel-preset-env/data/built-ins.json
@@ -17,7 +17,7 @@
     "safari": "3.1",
     "node": "0.10",
     "ie": "9",
-    "android": "4.0",
+    "android": "4",
     "ios": "6",
     "phantom": "2",
     "electron": "1.1"
@@ -40,7 +40,7 @@
     "safari": "3.1",
     "node": "0.10",
     "ie": "9",
-    "android": "4.0",
+    "android": "4",
     "ios": "6",
     "phantom": "2",
     "electron": "1.1"
@@ -73,7 +73,7 @@
     "safari": "3.1",
     "node": "0.10",
     "ie": "9",
-    "android": "4.0",
+    "android": "4",
     "ios": "6",
     "phantom": "2",
     "electron": "1.1"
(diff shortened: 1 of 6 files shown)


packages/babel-preset-env/data/built-ins.json
```diff
@@ -215,10 +215,10 @@
     "edge": "12",
     "firefox": "5",
     "safari": "10.1",
-    "node": "10.0",
+    "node": "10",
```
nicolo-ribaudo: I think that `10.0` was better? Other features (e.g. `proposal-async-generator-functions`) have `node: 10.13`, so the minor version could be meaningful.

existentialism: For me, it’s about being consistent, so we don’t have multiple ways Of representing the same thing (`4.0` for one, and `4` for another).

I don’t find `10` vs `10.13` hard to decipher, in fact, I think that it makes minor versions _more_ noticeable.

But subjective stuff aside, thoughts on at least being consistent? Ultimately I’m fine with reverting this... not a huge deal :)

nicolo-ribaudo: If we handle `10` exactly as `10.0` it's ok like you prefer, I don't really care about it

packages/babel-preset-env/scripts/build-modules-support.js
```diff
@@ -3,23 +3,22 @@ const fs = require("fs");
...
 const allowedBrowsers = {};
 
 Object.keys(stats).forEach(browser => {
-  if (browser !== "and_chr") {
+  if (!skipList.has(browser)) {
     const browserVersions = stats[browser];
     const allowedVersions = Object.keys(browserVersions)
       .filter(value => {
-        return acceptedWithCaveats[browser]
+        // Edge 16/17 are marked as "y #6"
```
danez: isn't that usually a sign it is not fully supported? or behind a flag?

existentialism: It's a bit more nuanced.

If it's not fully supported it's marked as `a #<footnote number`.

In this case, it is fully supported, except the caveat is that it loads the file, but doesn't execute it.

Issue #9549, opened 2019-02-20
[bug report] - plugin-transform-typescript doesn't function correctly with 2 pragmas


vidarc:
## Bug Report

**Current Behavior**
Using version 7.3.2 of plugin-transform-typescript results in unexpected behavior. Using `/** @jsx jsx */` at the top of the file works for that one, but files that are just plain react will remove the `import React from 'react'`. 

Versions of plugin-transform-typescript before 7.3.2 work as expected. I assume this PR: https://github.com/babel/babel/pull/9095 is the one that broke this.

**Input Code**
- REPL or Repo link if applicable:

File A
```js
/** @jsx jsx */
import { css, jsx } from '@emotion/core'

....code
```

File B
```js
import React from 'react'

.....code
```

File A gets output as expected. File B gets output as expected, except for the React import being removed. Adding `/** @jsx React.createElement */` at the top does not fix this. Produces same result.

**Expected behavior/code**
I expect `/** @jsx jsx */` to only effect the file it is in

**Babel Configuration (.babelrc, package.json, cli command)**

```js
{
  "plugins": [
    ["@babel/plugin-transform-typescript", {
      "isTSX": true,
    }],
    "@babel/plugin-syntax-jsx",
    "@babel/plugin-transform-react-jsx",
  ]
}
```

**Environment**
    "@babel/cli": "7.2.3",
    "@babel/core": "7.3.3",
    "@babel/plugin-syntax-jsx": "7.2.0",
    "@babel/plugin-transform-react-jsx": "7.3.0",
    "@babel/plugin-transform-typescript": "7.2.0",
    "babel-loader": "8.0.5",

nicolo-ribaudo:
Fixed by https://github.com/babel/babel/pull/9798

PR #9577, opened 2019-02-25
update plugin-typescript-transform to support different pragma between files

vidarc:
| Fixed Issues?            | https://github.com/babel/babel/issues/9549
| Documentation PR Link    | 

JSX pragma appears to be "sticky" between files (unsure if this is by design) when using the typescript-transform plugin. This PR: https://github.com/babel/babel/pull/9095 removed the need to assign your jsx (like what emotion.sh uses) to a variable, but in doing so broke files that would be just React (or whatever the initial jsxPragma was). Could get around this by having different builds/folders for different pragmas, but seems like an unnecessary step.

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -25,6 +25,7 @@ const PARSED_PARAMS = new WeakSet();
 export default declare((api, { jsxPragma = "React" }) => {
   api.assertVersion(7);
 
+  const defaultPragma = jsxPragma;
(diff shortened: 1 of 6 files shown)


vidarc:
if anyone sees this, i gave up because i couldn't find a way to test this 🤣 
needs to run against multiple files using the --out-dir flag

billyjanitsch:
@nicolo-ribaudo can you provide any guidance on how to add a test for multiple files? I'm happy to pick this up but I'm not seeing other examples of it in the existing tests.

PR #9579, opened 2019-02-25
change var name for coherence

tanohzana:
| Fixed Issues?            |
| Documentation PR Link    | 

I was reding the code to understand how it worked, but I enciuntered this variable name called `printMethod` and I thought it was going to be used afterwards, but it turned out it as only to test that the function existed, so I think it makes more sense to name it as a boolean.

### DIFF
--- a/packages/babel-generator/src/printer.js
+++ b/packages/babel-generator/src/printer.js
@@ -368,7 +368,7 @@ export default class Printer {
 
     const loc = t.isProgram(node) || t.isFile(node) ? null : node.loc;
     this.withSource("start", loc, () => {
-      printMethod(node, parent);
+      printMethod.call(this, node, parent);
     });
 
     this._printTrailingComments(node);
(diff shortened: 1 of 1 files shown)


tanohzana:
It's either possible to rename this variable or use `printMethod` on line 370 so the variable is reused.

Just let me know what suits you !

tanohzana:
Yes it is, but it is used as a boolean. Let me know then, if what I do next makes more sense then.

packages/babel-generator/src/printer.js
```diff
@@ -367,7 +368,7 @@ export default class Printer {
 
     const loc = t.isProgram(node) || t.isFile(node) ? null : node.loc;
     this.withSource("start", loc, () => {
-      this[node.type](node, parent);
+      printMethod(node, parent);
```
danez: We could do
```suggestion
      printMethod.call(this, node, parent);
```

to get it working.

PR #9582, opened 2019-02-25
Minify bundles on circle for repl

danez:
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Seems I did not think of that the bundles build on circle are used in the repl. So currently the `min` bundle is not minified.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ jobs:
       - run: yarn --version
       - run: make test-ci-coverage
       # Builds babel-standalone with the regular Babel config
-      - run: make build
+      - run: IS_PUBLISH=true make build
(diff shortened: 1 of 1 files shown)


danez:
The push failed, well because I did not push to this repo but my fork. 🤔

nicolo-ribaudo:
That's strange :thinking: I have always only pushed to my fork

danez:
What should the threshold be for the buildsize comment?

Currently buildsize will comment on PRs when the bundles changes for more than 750 Bytes, which is a little bit low I think.

nicolo-ribaudo:
10 kB? It should catch regressions like duplicate packages while not firing for most PRs

danez:
Okay will change.

PR #9585, opened 2019-02-25
Fix strict mode prescanning with EmptyStatement

danez:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Correctly prescan for use strict directives.

For example this should work, because the "use strict" is not a directive, because an EmptyStatement is in front of it. (Currently throws, about non-simple parameters in strict mode)

```js
function a ([a] = []) {
 ; 'use strict'; with ({}) {}
}
```


nicolo-ribaudo:
Oh I have always thought that empty statements did not prevent strings from being directives.

PR #9589, opened 2019-02-26
Check exported bindings are defined

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

This ensures that exports are always defined before exporting them. This is required according to the spec here: https://www.ecma-international.org/ecma-262/9.0/index.html#sec-module-semantics-static-semantics-early-errors

> It is a Syntax Error if any element of the ExportedBindings of ModuleItemList does not also occur in either the VarDeclaredNames of ModuleItemList, or the LexicallyDeclaredNames of ModuleItemList.


packages/babel-generator/test/index.js
```diff
@@ -489,7 +489,7 @@ suites.forEach(function(testSuite) {
             const actualAst = parse(actualCode, {
               filename: actual.loc,
               plugins: task.options.plugins || [],
-              strictMode: false,
```
danez: This was not really correct as `sourceType:"module"` should always be in strict mode.

nicolo-ribaudo:
Can I `export { Object }`?

danez:
I was wondering the same while working on it, but node does not allow to export globals either. I will add a test.

shrinktofit:
Please take a look at #2829

benjamn:
https://github.com/babel/babel/issues/9720#issuecomment-481030523

It's fine to use the parser to help enforce spec compliance, as long as you make the behavior configurable, but as far as I can tell there's no way to disable this new and surprising behavior.

Babel has a powerful plugin system that allows sophisticated developers to produce valid code even if it looks "wrong" when it's first parsed. If the parser just refuses to parse the code, that power is lost.

Please consider making this behavior configurable! If you're open to that idea, I would be happy to submit a PR.

PR #9593, opened 2019-02-26
[TS] Correctly forget `await`s after parsing async arrows with type args

### LINKED ISSUE #9560: [typescript] async arrow function property with generics fails parser
## Bug Report

**Current Behavior**

In the current `@babel/parser` version 7.3.3, a class property that defines an async arrow function with generics breaks the parser.

```
    SyntaxError: boost/packages/core/src/Executor.ts: Unexpected token, expected ";" (85:52)

      83 |    * Execute a task with the provided value.
      84 |    */
    > 85 |   executeTask = async <T>(task: Task<Ctx>, value?: T): Promise<any> => {
         |                                                      ^
      86 |     const { console: cli } = this.tool;
      87 |     let result = null;
      88 | 

      at Object.raise (node_modules/@babel/parser/lib/index.js:3831:17)
```

If you remove the `<T>` generics or the `async/await`, the parser works.

**Input Code**

```ts
  executeTask = async <T>(task: Task<Ctx>, value?: T): Promise<any> => {
    const { console: cli } = this.tool;
    let result = null;

    cli.emit('task', [task, value, this.parallel]);

    try {
      result = await task.run(this.context, value);

      if (task.isSkipped()) {
        cli.emit('task.skip', [task, value, this.parallel]);
      } else {
        cli.emit('task.pass', [task, result, this.parallel]);
      }
    } catch (error) {
      cli.emit('task.fail', [task, error, this.parallel]);

      throw error;
    }

    return result;
  };
```

**Expected behavior/code**

It parses. This worked in the previous version.

```js
module.exports = {
  "babelrc": false,
  "comments": false,
  "plugins": [
    "@babel/plugin-proposal-export-default-from",
    "@babel/plugin-proposal-nullish-coalescing-operator",
    "@babel/plugin-proposal-optional-catch-binding",
    "@babel/plugin-proposal-optional-chaining",
    "@babel/plugin-syntax-dynamic-import",
    [
      "babel-plugin-transform-dev",
      {
        "evaluate": false
      }
    ]
  ],
  "presets": [
    [
      "@babel/preset-env",
      {
        "modules": "commonjs",
        "shippedProposals": true,
        "targets": {
          "node": "10.9.0"
        },
        "useBuiltIns": false
      }
    ],
    "@babel/preset-typescript"
  ]
};
```

**Environment**
- Babel version(s): 7.3.3
- Node/npm version: 10.14.1/6.5.0
- OS: OSX 10.14.3
- Monorepo: yes

nicolo-ribaudo:
| Fixed Issues?            | Fixes #9560
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The problem was that we didn't reset `state.awaitPos = 0` after parsing `async <T>() =>` arrow functions, and thus the following async function thought that `await` was used in its parameters. I reused an existing function to parse arrow bodies.

### DIFF
--- a/packages/babel-parser/test/fixtures/typescript/arrow-function/async-await-null/output.json
+++ b/packages/babel-parser/test/fixtures/typescript/arrow-function/async-await-null/output.json
@@ -93,7 +93,6 @@
           "params": [],
           "id": null,
           "generator": false,
-          "expression": true,
           "async": true,
           "body": {
             "type": "AwaitExpression",
--- a/packages/babel-parser/src/plugins/typescript.js
+++ b/packages/babel-parser/src/plugins/typescript.js
@@ -6,13 +6,7 @@ import { types as ct } from "../tokenizer/context";
 import * as N from "../types";
 import type { Pos, Position } from "../util/location";
 import Parser from "../parser";
-import {
-  type BindingTypes,
-  functionFlags,
-  BIND_NONE,
-  SCOPE_ARROW,
-  SCOPE_OTHER,
-} from "../util/scopeflags";
+import { type BindingTypes, BIND_NONE, SCOPE_OTHER } from "../util/scopeflags";
 
 type TsModifier =
   | "readonly"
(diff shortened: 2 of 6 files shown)


packages/babel-parser/test/fixtures/typescript/arrow-function/async-await-null/output.json
```diff
@@ -93,7 +93,6 @@
           "params": [],
           "id": null,
           "generator": false,
-          "expression": true,
```
nicolo-ribaudo: This doesn't exist anywhere else.

PR #9598, opened 2019-02-26
Update to webpack 4 and take advantage of es modules while bundling

nicolo-ribaudo:

While checking https://unpkg.com/@babel/standalone@7.3.4/babel.js I noticed that there are a lot of functions generated by out commonjs plugin's lazy mode (search for `function _`), e.g.
```js
function _types() {
  var data = _interopRequireWildcard(__webpack_require__(3));

  _types = function _types() {
    return data;
  };

  return data;
}
```
This is only needed in node: in webpack we shouldn't transpile import/export. I solved this problem by bundling our untranspiled sources, instead of the already transpiled to commonjs files.

Upgrading to webpack 4 wasn't my goal initially, but I needed a custom plugin and I couldn't figure out how to write one for webpack 3.
With webpack 4, we are getting some warnings because `webpack-dependency-suite` doesn't support it yet, but until webpack 5 it won't break:
```
(node:1616) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead
(node:1616) DeprecationWarning: Resolver: The callback argument was splitted into resolveContext and callback.
(node:1616) DeprecationWarning: Resolver#doResolve: The type arguments (string) is now a hook argument (Hook). Pass a reference to the hook instead.
(node:1616) DeprecationWarning: Pass resolveContext instead and use createInnerContext
```

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -110,6 +110,13 @@ module.exports = function(api) {
         sourceType: "unambiguous",
       },
       {
+        test: [
+          "packages/babel-runtime/helpers/esm",
+          "node_modules/@babel/runtime/helpers/esm",
+        ],
+        sourceType: "module",
+      },
+      {
         // The runtime transform shouldn't process its own runtime or core-js.
         exclude: [
           "packages/babel-runtime",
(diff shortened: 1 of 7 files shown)


nicolo-ribaudo:
Nvm, the bundle size increased

danez:
I thought about trying that too. But even with the increased size, we should update at some point anyway. Also babel-standalone is not an esmodule in master. You could try to use my PR: #9025

PR #9601, opened 2019-02-27
Don't loose "this" in helper-call-delegate

### LINKED ISSUE #9385: Typescript loses this in class method when parameter has type signature and default value
## Bug Report

**Current Behavior**
Having code as in example will produce invalid es5 with broken this

**Input Code**

```ts
export class Test {
  invite(options: { privacy: string } = {}) {
    const privacy = options.privacy || "Private"
    console.log(this)
  }
}
```

Babel output will be:
```js
export var Test =
/*#__PURE__*/
function () {
  function Test() {
    _classCallCheck(this, Test);
  }

  _createClass(Test, [{
    key: "invite",
    value: function invite() {
      var options = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : {};
      return function () {
        var privacy = options.privacy || "Private";
        console.log(this);
      }();
    }
  }]);

  return Test;
}();
```

**Expected behavior/code**
No additional scope should be created.

```js
{
  "plugins": [
    "@babel/plugin-transform-typescript"
  ],
  "env": {
    "production": {
      "presets": [
        [
          "@babel/preset-env",
          {
            "modules": false
          }
        ]
      ]
    }
  }
}
```

**Environment**
- Babel version(s): 7.2.0
- OS: OSX
- Monorepo: no
- How you are using Babel: `cli`, `loader`]

**Possible Solution**
A possible workaround is to move type from method declaration:

```ts
type Options = { privacy: string }
export class Test {
  invite(options: Options = {}) {
    const privacy = options.privacy || "Private"
    console.log(this)
  }
}
```

nicolo-ribaudo:
| Fixed Issues?            | Fixes #9385
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-transform-parameters/test/fixtures/parameters/iife-this-9385/output.js
+++ b/packages/babel-plugin-transform-parameters/test/fixtures/parameters/iife-this-9385/output.js
@@ -1,0 +1,35 @@
+"use strict";
+
+Object.defineProperty(exports, "__esModule", {
+  value: true
+});
+exports.Test = void 0;
+
+function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }
+
+function _defineProperties(target, props) { for (var i = 0; i < props.length; i++) { var descriptor = props[i]; descriptor.enumerable = descriptor.enumerable || false; descriptor.configurable = true; if ("value" in descriptor) descriptor.writable = true; Object.defineProperty(target, descriptor.key, descriptor); } }
+
+function _createClass(Constructor, protoProps, staticProps) { if (protoProps) _defineProperties(Constructor.prototype, protoProps); if (staticProps) _defineProperties(Constructor, staticProps); return Constructor; }
+
+var Test =
+/*#__PURE__*/
+function () {
+  function Test() {
+    _classCallCheck(this, Test);
+  }
+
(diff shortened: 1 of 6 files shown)


packages/babel-plugin-transform-parameters/test/fixtures/parameters/iife-this-9385/output.js
```diff
@@ -0,0 +1,35 @@
...
+function () {
+  function Test() {
+    _classCallCheck(this, Test);
+  }
+
+  _createClass(Test, [{
+    key: "invite",
+    value: function invite() {
+      var options = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : {};
+      return function () {
```
jridgewell: Why is the wrapper generated?

nicolo-ribaudo: Babel deopts because it thinks that the `privacy` key in the object type is a reference. I opened a PR to fix the iife case, but I'll see how we can avoid the deoptimization here.

nicolo-ribaudo: Will be fixed by #9605

Issue #9603, opened 2019-02-27
Proposal: Switch Expression


koblas:
## Feature Request

**Is your feature request related to a problem? Please describe.**

This is an improvement to syntax for a standard pattern.

**Describe the solution you'd like**

It's very common to need to compute a value based on a switch statement.  Currently, there are two ways to do this, but both are either hard to understand or prone to error.

This is unmaintainable -- 
```
    const x = v === 'foo' ? 1 : v === 'bar' ? 3 : v === 'baz' ? 6 : 99;  
```
This is maintainable, but is less than ideal:

```
   let x;

   switch (v) {
   case "foo":
     x = 1;
     break;
   case "bar":
     x = 3;
     break;
   case "baz":
     x = 6;
     break;
   default:
     x = 99;
     break;
   }
```

This would all be simpler with allowing switch statement to become switch expressions, like so:

```
   const x = switch (v) {
     case "foo" => 1;
     case "bar" => 3;
     case "baz" => 6;
     default => 99;
   }
```

What I really like is that it preserves all of the normal JavaScript syntax with the small change that a switch is allowed in an expression provided that all of the cases evaluate to expressions hence the use of the '=>' as an indicator.  Fundamentally this is a very basic concept where you have a state machine and need it switch based on the current state and evaluate to the new state.

```
   const nextState = switch (currentState) {
      case ... => 
   }
```

**Describe alternatives you've considered**

There are two Stage 1 proposals that touch on this, but in the case of pattern matching they are still a statement or do expressions which appears to be overly complex for the task.  The problem that I see and want to address is the need to have something that removes the need to chain trinary expressions together to have an Expression.

Pattern matching does shorten the code, but you have a weird default case and also still end up with a loose variable and since pattern matching is a statement you still have an initially undefined variable.

```
   let x;

   case (v) {
     when "foo" -> x = 1;
     when "bar" -> x = 3;     
     when "baz" -> x = 6;
     when v -> x = 99;
   }
```

Let's try do expressions, I'll leave people's thoughts to themselves.  
```
   const x = do {
     if (v === "foo") { 1; }
     else if (v === "bar") { 3; }
     else if (v === "baz") { 6; }
     else { 99; }
   }
```

Or as another do expression variant:

```
   const x = do {
     switch (v) {
       case "foo": 1; break;
       case "bar": 3; break;
       case "baz": 6; break;
       default: 99; break;
     }
   }
```

**Teachability, Documentation, Adoption, Migration Strategy**

[...]

PR #9604, opened 2019-02-27
Proposal: switch-expression support

koblas:
This is a pull request to add support for switch expressions #9603 please see the issue for more details.

### DIFF
--- a/packages/babel-parser/ast/spec.md
+++ b/packages/babel-parser/ast/spec.md
@@ -2,6 +2,7 @@ These are the core @babel/parser (babylon) AST node types.
 
 - [Node objects](#node-objects)
 - [Changes](#changes)
+    - [@babel/parser (Babylon) v7](#babelparser-babylon-v7)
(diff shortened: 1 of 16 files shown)


citycide:
This is great and all but this will never be merged without an official proposal to TC39 (as much as I want a pattern matching expression :laughing:), as @nicolo-ribaudo pointed out in the linked issue.

https://github.com/babel/babel/issues/9603#issuecomment-468062400

koblas:
I've been working on a TC39 proposal in parallel to this PR.  I wanted to provide a clear place to have an implementation where people could add some comments as needed.

When I look back at the pattern matching PR list, I see they did at least 3 PRs over time.

Is their a rule for babel TC39 proposals that they will only be consider if they're at least stage 1?

nicolo-ribaudo:
Generally we give precedence to proposals which are in higher stages, but the only real requirement is that it is at least stage 1.

PR #9605, opened 2019-02-27
[plugin-transform-typescript] Strip type imports used in Enums and object types

### LINKED ISSUE #8881: TypeScript: enum member with the same name as an imported type prevents import stripping
## Bug Report

**Current Behavior**
If the name of an `enum` member matches that of an imported type, the corresponding `import` statement remains through transpilation, even if it's never referenced.

**Input Code**

```ts
import Foo from 'some-module';

enum Enum {
  Foo
}

function myFunction(): Foo {}
```

([REPL link](https://babeljs.io/en/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=JYWwDg9gTgLgBAMQhOAzKERwOQGdMCmAtCBACYCuANgdgNwBQDBAdhVgKJtYDeDciZAwC-TVBRYBjGMAgs4IAJ4IJ02SwAUASgBcglD1FA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=typescript&prettier=false&targets=Node-6&version=7.0.0-beta.36&envVersion=7.0.0-beta.38))

**Expected behavior/code**
I expect _not_ to see `import Foo from 'some-module';` in the compiled output, since `Foo` is only ever used as a type.

Note that changing the name of the `enum` member, e.g. to `Foo2`, causes the `import` statement to be stripped.

**Environment**
- `@babel/core`: 7.1.2
- `@babel/plugin-transform-typescript`: 7.1.0

echenley:
| Fixed Issues?            | Fixes #8881 
| Documentation PR Link    | 

`TSEnumMember` and `TSPropertySignature` should not always be considered type references. Updates `isReferenced` to reflect this, and adds a few new tests for these use cases. Fixes the two cases mentioned in #8881.

Note: I don't have experience contributing to Babel, and I am not positive that this is the correct solution. Please let me know if you believe that there is another solution to this problem, or if I am missing some context with regards to this whitelist. Thank you!


nicolo-ribaudo:
This is correct for TSPropertySignature, but enums shouldn't be counted as references in the first place :thinking:

nicolo-ribaudo:
Those shouldn't count as references in the first place. You need to update https://github.com/babel/babel/blob/master/packages/babel-types/src/validators/isReferenced.js (you can add the needed TS-specific types at the end of the switch)

echenley:
So, basically this?

```js
    // no: enum X { NODE }
    case "TSEnumMember":
      return false;

```

nicolo-ribaudo:
Exactly :+1: For object type properties, we need to check that we are in the key and that it isn't computed (like we do for `ObjectProperty`, but without checking the grandparent).

packages/babel-types/src/validators/isReferenced.js
```diff
@@ -134,6 +134,18 @@ export default function isReferenced(
     // no: type X = { NODE: OtherType }
     case "ObjectTypeProperty":
       return parent.key !== node;
+
+    // no: enum X { NODE }
+    case "TSEnumMember":
+      return false;
+
+    // no: { NODE: NODE }
```
nicolo-ribaudo: This should be `{ NODE: value }`, since only the key isn't referenced.

Here, the type `string` _is_ a reference:

```js
type X = { foo: string }
```

echenley: Updated, appreciate the hand-holding. 😄

Going to add a couple more tests to babel-types instead of relying on the tests in plugin-transform-typescript.

packages/babel-types/src/validators/isReferenced.js
```diff
@@ -134,6 +134,18 @@ export default function isReferenced(
     // no: type X = { NODE: OtherType }
     case "ObjectTypeProperty":
       return parent.key !== node;
+
+    // no: enum X { NODE }
+    case "TSEnumMember":
+      return false;
```
nicolo-ribaudo: I just noticed that we need to check that node is the `id`, because `enum X { foo = NODE }` is referenced.

echenley: Good call, updated and added some extra tests.

packages/babel-plugin-transform-parameters/test/fixtures/parameters/iife-this-9385/output.js
```diff
@@ -22,10 +22,8 @@ function () {
     key: "invite",
     value: function invite() {
       var options = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : {};
-      return function () {
-        var privacy = options.privacy || "Private";
-        console.log(this);
-      }.apply(this);
+      var privacy = options.privacy || "Private";
+      console.log(this);
```
danez: Why did this change? It seems unrelated to the change.

echenley: There's an explanation here: https://github.com/babel/babel/pull/9601/files#r260984879

danez: ah, I see 👍

PR #9607, opened 2019-02-28
Refactor parsing object members

danez:
| Fixed Issues?            | 
| Minor: New Feature?      | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

This makes some changes around destructuring, restspread and objects.

- Split the actual parsing of object-members from `parseObj()`-loop into `parseObjectMember`.
- Parse `RestElement` in `parseObjectMember`  not with `parseSpread() + toAssignable()` as that seems an overkill if we can parse it directly.
- Parse `RestElement` in `parseObjectMember` not with `parseRest()` as only Identifier is allowed in this context
- Do allow object shorthand property duplicates named `__proto__`. The spec specifically says only duplicates which come from the form `*PropertyDefinition: PropertyName : AssignmentExpression*` : https://www.ecma-international.org/ecma-262/9.0/index.html#sec-__proto__-property-names-in-object-initializers
- Disallow a newline between async and method name.
- Refactored the trailing comma after rest element messages. We had lots of different variations for this error and I opted to go for one message for all cases (same as chrome): `"Rest element must be last element"` This makes the code a little bit simpler
- Rename `parseRest()` to `parseRestBinding()`


packages/babel-parser/src/parser/expression.js
```diff
@@ -1425,108 +1430,119 @@ export default class ExpressionParser extends LValParser {
...
+      prop.key.type === "Identifier" &&
+      prop.key.name === "async" &&
+      (this.match(tt.name) ||
+        this.match(tt.num) ||
+        this.match(tt.string) ||
+        this.match(tt.bracketL) ||
+        this.state.type.keyword ||
+        this.match(tt.star)) &&
+      !lineBreak.test(
+        this.state.input.slice(this.state.lastTokEnd, this.state.start),
```
nicolo-ribaudo: You can use `this.hasPrecedingLineBreak`

nicolo-ribaudo:
> * Parse `RestElement` in `parseObjectMember` not with `parseRest()` as only Identifier is allowed in this context
> * Disallow a newline between async and method name.

Could you add some tests for these two points? (Unless I just missed them)

packages/babel-parser/src/parser/expression.js
danez: Done

danez:
I added a test for the async-newline stuff. The RestElement has already tests. It is all the tests that switched from throwing `Invalid rest operator's argument` to `Unexpected token`.

PR #9610, opened 2019-02-28
Use `injectInitialization` to generate ts parameter properties

### LINKED ISSUE #9609: @babel/plugin-transform-typescript can insert public property initialization before the call to super.
## Bug Report

**Current Behavior**
`@babel/plugin-transform-typescript` inserts public property initialization at the start of the constructor when `super()` is not the very first statement.

**Input Code**

```js
class A {
}

class B extends A {
  constructor(public p: string) {
    console.log('anything before super');
    super();
  }
}
```

**Expected behavior/code**
`this.p = p;` should be added after the call to `super();`.

**Babel Configuration (.babelrc)**

```js
{
  "plugins": ["@babel/plugin-transform-typescript"]
}
```

**Environment**
- Babel version(s): @babel/core 7.3.4, @babel/plugin-transform-typescript 7.3.2
- Node/npm version: node 10.15.0, npm 6.4.1
- OS: Fedora 28
- Monorepo: no

**Possible Solution**
I'm not sure how to solve this but the issue is here:
https://github.com/babel/babel/blob/208195f42576a70798cedf1abe4d9f1d86979fcc/packages/babel-plugin-transform-typescript/src/index.js#L201-L212

This was found by users of istanbul, istanbuljs/istanbuljs#237 and istanbuljs/babel-plugin-istanbul#185.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #9609
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR fixes something where Babel is _tecnically_ correct and `babel-plugin-istanbul` is buggy, but there are an open feature request and PR to allow this behavior in TypeScript (see https://github.com/babel/babel/issues/9609#issuecomment-468349570).
I think that we can fix this in Babel.

cc @babel/typescript

### DIFF
--- a/packages/babel-helper-create-class-features-plugin/src/index.js
+++ b/packages/babel-helper-create-class-features-plugin/src/index.js
@@ -21,7 +21,7 @@ import {
 
 import pkg from "../package.json";
 
-export { FEATURES };
+export { FEATURES, injectInitialization };
 
 // Note: Versions are represented as an integer. e.g. 7.1.5 is represented
 //       as 70000100005. This method is easier than using a semver-parsing
--- a/packages/babel-plugin-transform-typescript/package.json
+++ b/packages/babel-plugin-transform-typescript/package.json
@@ -13,6 +13,7 @@
     "typescript"
   ],
   "dependencies": {
+    "@babel/helper-create-class-features-plugin": "^7.3.4",
     "@babel/helper-plugin-utils": "^7.0.0",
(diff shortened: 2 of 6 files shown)


nicolo-ribaudo:
Since the package is a direct dependency it will always use the correct version

SimenB:
Will this land? Seems like it'll fix downstream https://github.com/facebook/create-react-app/issues/5868

PR #9613, opened 2019-02-28
Don't add params of anonymous exported function decls to the outer scope

### LINKED ISSUE #9611: Duplicate declaration
## 🐛 Bug Report

When running babel with preset-env on this code

```javascript
import x from './x';

export default function(x) {}
```

It errors with 'Duplicate declaration "x"'

I think this is valid code however, as node has no problem executing the file (with esm)

nicolo-ribaudo:
| Fixed Issues?            | Fixes #9611
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-transform-modules-commonjs/test/fixtures/regression/issue-9611/input.mjs
+++ b/packages/babel-plugin-transform-modules-commonjs/test/fixtures/regression/issue-9611/input.mjs
@@ -1,0 +1,3 @@
+import x from './x';
+
+export default function(x) {}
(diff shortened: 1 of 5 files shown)


packages/babel-plugin-transform-modules-commonjs/test/fixtures/regression/issue-9611/input.mjs
```diff
@@ -0,0 +1,3 @@
+import x from './x';
+
+export default function(x) {}
```
danez: Looking at the code I would guess it throws, because on line 465 it tries to get the id of the declaration but this one does not have one. but maybe path.get() does never return null?

nicolo-ribaudo: It is set at https://github.com/babel/babel/blob/47bee79acf39e4fba0e21c4e2c52ea6cd11fce67/packages/babel-helper-split-export-declaration/src/index.js#L34

danez: 👍

PR #9618, opened 2019-02-28
Disallow escape sequences in contextual keywords

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

Resolves two problems:
- Disallow (async)(a) => {}
- Disallow escape sequences in contextual keywords like `async, as, of, get, set, static, ...`

The cases which were parsing successfully with escape sequences (although shouldn't) were: async in almost all cases. get & set in classes.

Also added tests for a lot of cases that were already correctly working.


packages/babel-parser/src/parser/expression.js
```diff
@@ -704,11 +712,11 @@ export default class ExpressionParser extends LValParser {
 
   atPossibleAsync(base: N.Expression): boolean {
     return (
-      !this.state.containsEsc &&
```
danez: `containsEsc` only says if the current token has escape sequences. This references the token after `async` which is the wrong one.

packages/babel-parser/src/parser/expression.js
```diff
@@ -576,13 +586,8 @@ export default class ExpressionParser extends LValParser {
         this.expect(tt.bracketR);
         return this.finishNode(node, "OptionalMemberExpression");
       } else if (this.eat(tt.parenL)) {
-        const possibleAsync = this.atPossibleAsync(base);
-
         node.callee = base;
-        node.arguments = this.parseCallExpressionArguments(
-          tt.parenR,
-          possibleAsync,
-        );
```
danez: I removed this, as I don't see a way what an `optionalCallExpression` has anything todo with the `async` keyword. No tests failed or changed. I guess this was copied wrong.

packages/babel-parser/src/parser/expression.js
```diff
@@ -525,12 +525,21 @@ export default class ExpressionParser extends LValParser {
     startLoc: Position,
     noCalls?: ?boolean,
   ): N.Expression {
+    const maybeAsyncArrow = this.atPossibleAsync(base);
```
danez: moving it from line 624 to here, basically prevents `(async)(a) => {}`

packages/babel-parser/src/parser/expression.js
```diff
@@ -641,7 +645,11 @@ export default class ExpressionParser extends LValParser {
         this.finishOptionalCallExpression(node);
       }
 
-      if (possibleAsync && this.shouldParseAsyncArrow()) {
+      if (
+        maybeAsyncArrow &&
+        !this.canInsertSemicolon() &&
```
danez: `!this.canInsertSemicolon()` prevents 
```js
async (x) 
=> {}
```

nicolo-ribaudo: Maybe we could move the check inside `shouldParseAsyncArrow`?
Also note that both flow and TS allow this code:
```js
async (f)
: t => { }
```
but disallow

```js
async (f) : t
=> { }
```

danez: Thanks, I added testcases and moved it into `shouldParseAsyncArrow `

PR #9622, opened 2019-03-02
babel-traverse: fix order of exit callbacks

Anber:

I've found a problem with the order of the tree traversing. Suppose we have a visitors configuration like this:
```
{
  enter() {
    // callback #1
  },
  exit() {
    // callback #4
  },
  Directive: {
    enter() {
      // callback #2
    },
    exit() {
      // callback #3
    },
  },
}
```

It is expected that 4th callback will be called after 3rd so we can do some common work before and after callbacks in Directive. However, exit callbacks are called in the same order as enter callbacks (eg. visitors from the example will be called in the order 1=>2=>4=>3).

### DIFF
--- a/packages/babel-traverse/src/path/context.js
+++ b/packages/babel-traverse/src/path/context.js
@@ -4,15 +4,22 @@ import traverse from "../index";
 
 export function call(key): boolean {
   const opts = this.opts;
+  const order = opts.exitOrder || "direct";
 
   this.debug(key);
 
   if (this.node) {
     const common = opts[key];
     const typed = opts[this.node.type] && opts[this.node.type][key];
-    return key === "exit"
-      ? this._call(typed) || this._call(common)
-      : this._call(common) || this._call(typed);
+    if (key === "exit" && order === "reverse") {
+      return this._call(typed) || this._call(common);
+    }
+
+    if (key === "exit" && common && typed && !opts.exitOrder) {
+      // There we can show a warning about unexpected behavior
+    }
+
+    return this._call(common) || this._call(typed);
   }
 
   return false;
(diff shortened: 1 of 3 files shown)


loganfsmyth:
This seems like a dangerous change since there could easily be existing code that relies on the existing behavior. If we wanted to consider this it would need to be part of a major version bump.

Anber:
It’s definitely a braking change but maybe we can add new behavior as an option?

loganfsmyth:
It seems like the easiest short-term solution would be
```
var visitor = {
  Directive: {
    enter() {
      // callback #2
    },
    exit() {
      // callback #3
    },
  },
}

traverse(ast, {
  enter(path, state) {
    // callback #1

    var handler = visitor[path.type];
    if (handler) handler(path, state);
  },
  exit() {
    var handler = visitor[path.type];
    if (handler) handler(path, state);

    // callback #4
  },
})
```

Anber:
Thank you. I have a workaround for my case but it is still workaround for unexpected and undocumented behavior :)
However, we can add the correct order as an optional behavior in one of the next minor releases in order to make it default in one of the future major releases.

loganfsmyth:
I think the problem really is that it'd just hard to understand no matter what. I agree that your expectation of ordering is one option, but I don't think it's as obvious as you think. Say you had instead written
```
{
  enter() {
    // callback #1
  },
  Directive: {
    enter() {
      // callback #2
    },
    exit() {
      // callback #3
    },
  },
  exit() {
    // callback #4
  },
}
```
what is the expected order?

I agree that it is underdefined right now, but I think the core issue is really that using an bunch of objects for all of this is just a bad way to do it.

Anber:
I mean if you want to implement something like callstack you can add push and pop logic to “global” enter and exit callback and some node related logic to “typed” callbacks. And if you do this you will realize that you don’t have expected stack frame in “typed” exit callback because it has been already popped in “global” exit.

Anber:
@loganfsmyth I've added an option which controls the order. What do you think about this variant?

PR #9626, opened 2019-03-03
[proposal-object-rest-spread] fix property using string template

### LINKED ISSUE #9563: Issue with rest spread and computed properties using string template
## Bug Report

**Current Behavior**
When using string templates in computed properties, the property name isn't correctly excluded from the rest properties object.
Looking at the compiled code, it puts 'undefined' instead of the actual property name. If you use string concatenation, this does work perfectly.

**Input Code**
[REPL sample](https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAlmADgV1gXhgbwL4G4BQA9ITAKIAeUATgIbCw0A2jMA1mCAO5gw1TVwARqgCmEGADMqIALYwoACxExkEEVXyhIsTPhgwA5nABuIsAH0wNGSIBchk2YBy1kQBo9MAOSMa0S65e9r7QLjYe-gDaAAagyGDUAJ7RALr2cQlUiREwkYhUIhJw5DAA1N7QfCJeaTCVUO6eMQAkmPmFxdhaagkQqengPVAQOcQwAILM8kowBdCeAHRLIIrq4_xUQqIQ-NgwGAgoULhAA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.3.3)

```js
const input = {};
// Extract all known attributes from the user
const {
  given_name: givenName,
  'last_name': lastName,
  [`country`]: country,
  [prefix + 'state']: state,
  [`${prefix}consents`]: consents,
  // All the rest
  ...otherAttributes
} = input;
```

**Expected behavior/code**
All given computed properties should be excluded from the `otherAttributes` object, but the ones using template strings remain.

**Environment**
- Babel version(s): v7.3.3 (probably older versions too)
- Node/npm version: any
- OS: any

**Possible Solution**
Make sure the transpiler knows what a template string is and correctly copy it over to the exclusion list. Sound like something that is easy to fix, but I have no idea where to start :).

AlbertLucianto:
| Fixed Issues?            | Fixes #9563
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

As suggested by @nicolo-ribaudo in the issue, template literal should be handled differently from the other literals because it does not have attribute value.

My first time contributing to babel, appreciate your review!


nicolo-ribaudo:
Thanks @AlbertLucianto! I'm sorry but I'm giving precedence to @pnowak (https://github.com/babel/babel/pull/9628), since they claimed the issue a few days ago (https://github.com/babel/babel/issues/9563#issuecomment-466812953).

AlbertLucianto:
@nicolo-ribaudo no worries.

Anyway, is there problem with the buildsize bot? The sizes are the same as the minimised one.

nicolo-ribaudo:
Probably your branch was based on a too old version of `master` (we added the bot last week).

PR #9628, opened 2019-03-03
[proposal-object-rest-spread] fix templateLiteral in extractNormalizedKeys 

pnowak:
| Fixed Issues?            | #9563
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Added additional check in extractNormalizedKeys function to handle templates differently than string.


packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -76,6 +76,8 @@ export default declare((api, opts) => {
       if (t.isIdentifier(prop.key) && !prop.computed) {
         // since a key {a: 3} is equivalent to {"a": 3}, use the latter
         keys.push(t.stringLiteral(prop.key.name));
+      } else if (t.isTemplateLiteral(prop.key)) {
+        keys.push(t.templateLiteral(prop.key.quasis, prop.key.expressions));
```
nicolo-ribaudo: Since this is the same as doing `t.cloneNode(prop.key)`, you can use that function. Note that this shouldn't set `allLiteral` to `false` (because templates are always strings), so it shouldn't be merged with the last `else` block.

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/template-literal-property/input.js
```diff
@@ -0,0 +1,10 @@
+const input = {};
+
+const {
+  given_name: givenName,
+  'last_name': lastName,
+  [`country`]: country,
+  [prefix + 'state']: state,
```
nicolo-ribaudo: Could you also add a test without this line, so that `allLiterals` will be true?

pnowak: Yes, I remove this `BinaryExpression`.

packages/babel-plugin-proposal-object-rest-spread/src/index.js
pnowak: Do you think something like check `isLiteral` and not `isTemplateLiteral`?

packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -76,7 +76,7 @@ export default declare((api, opts) => {
       if (t.isIdentifier(prop.key) && !prop.computed) {
         // since a key {a: 3} is equivalent to {"a": 3}, use the latter
         keys.push(t.stringLiteral(prop.key.name));
-      } else if (t.isLiteral(prop.key)) {
+      } else if (t.isLiteral(prop.key) && !t.isTemplateLiteral(prop.key)) {
```
nicolo-ribaudo: This will set `allLiterals` to false for template literals, but it isn't needed.

You can do something like
```suggestion
      } else if (t.isTemplateLiteral(prop.key)) {
        keys.push(t.cloneNode(prop.key));
      } else if (t.isLiteral(prop.key)) {
```

pnowak: Thank You very much.

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/template-literal-property/input.js
nicolo-ribaudo: Could you add this test, to test the last changes?

nicolo-ribaudo:
I meant to create a new test, to test both the `allLiterals: true` and `allLiterals: false` cases (possibly in two different files):

```js
const {
  given_name: givenName,
  'last_name': lastName,
  [`country`]: country,
  [prefix + 'state']: state,
  [`${prefix}consents`]: consents,
  ...rest
} = input;
```

```js
const {
  given_name: givenName,
  'last_name': lastName,
  [`country`]: country,
  ...rest
} = input;
```

nicolo-ribaudo:
Thank you!

pnowak:
Hi @nicolo-ribaudo.
Thank you for your help and the solution ;)

PR #9637, opened 2019-03-05
Update identifier parsing per Unicode v12

mathiasbynens:
https://github.com/tc39/ecma262/issues/1467
https://github.com/tc39/ecma262/pull/1468


packages/babel-parser/scripts/generate-identifier-regex.js
```diff
@@ -1,7 +1,7 @@
 "use strict";
 
 // Which Unicode version should be used?
```
nicolo-ribaudo:
```suggestion
// We should always use the latest version of Unicode
// https://tc39.github.io/ecma262/#sec-conformance
```

PR #9639, opened 2019-03-05
System modules - Hoist classes like other variables

guybedford:
| Fixed Issues?            | No tracking issue

The system module format hoists all variables, but does not currently hoist classes.

These leads to code like the following executing incorrectly:

```js
export function test () {
  return X;
}
class X {}
```

as the function is hoisted, but its reference to the class isn't.

Bug fix and tests included here.

In the process I've also combined _export calls where possible.


nicolo-ribaudo:
Thank you!

PR #9641, opened 2019-03-05
Allow context type annotation on getters/setters

### LINKED ISSUE #8069: Typescript: getters with this
## Bug Report

**Current Behavior**

**Input Code**
- https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAEAqCmEAu0DeAoa0Dm8VMSQAokALASwgC45CBKNTLaAJzwFcWA7aJF9-AG4mAX3QigA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=typescript%2Cenv&prettier=false&targets=Node-8.9&version=7.0.0-beta.42&envVersion=7.0.0-beta.42

```js
class Test {
  get test(this: Test) {
    return true;
  }
}
```

**Expected behavior/code**
No errors.

**Environment**

**Possible Solution**

https://www.typescriptlang.org/docs/handbook/functions.html#this

matt-tingen:
| Fixed Issues?            | Fixes #8069

Typescript allows adding [type annotations to the context (`this`) for functions](https://www.typescriptlang.org/docs/handbook/functions.html#this), including `get`ters and `set`ters. Presently, the [parser errors out](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEAqCmEAu0DeAoa0Dm8VMSQAokALASwgC45CBKNTLaAJzwFcWA7aJF9-AG4mAX3QigA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=typescript%2Cenv&prettier=false&targets=Node-8.9&version=7.0.0-beta.41) when such an annotation is present on a getter or setter because it expects exactly 0 or 1 parameter, respectively.

This change causes the parser to allow an extra first parameter to getters and setters when using the `typescript` plugin and the parameter's `name` is `this`. 

`babel-plugin-transform-typescript` already strips the `this: ...` annotation from functions, including getters and setters, but I added a test to verify this in addition to the parser test.

### DIFF
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -1547,19 +1547,18 @@ export default class ExpressionParser extends LValParser {
     );
   }
 
+  getGetterSetterExpectedParamCount(
+    method: N.ObjectMethod | N.ClassMethod,
+  ): number {
+    return method.kind === "get" ? 0 : 1;
+  }
+
   // get methods aren't allowed to have any parameters
   // set methods must have exactly 1 parameter which is not a rest parameter
   checkGetterSetterParams(method: N.ObjectMethod | N.ClassMethod): void {
-    const paramCount = method.kind === "get" ? 0 : 1;
+    const paramCount = this.getGetterSetterExpectedParamCount(method);
     const start = method.start;
-    // Allow `this` type annotation in TypeScript
-    if (
-      method.params.length === paramCount + 1 &&
-      method.params[0].type === "Identifier" &&
-      method.params[0].name === "this"
-    ) {
-      this.expectPlugin("typescript");
-    } else if (method.params.length !== paramCount) {
+    if (method.params.length !== paramCount) {
       if (method.kind === "get") {
         this.raise(start, "getter must not have any formal parameters");
       } else {
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -1552,7 +1552,14 @@ export default class ExpressionParser extends LValParser {
   checkGetterSetterParams(method: N.ObjectMethod | N.ClassMethod): void {
     const paramCount = method.kind === "get" ? 0 : 1;
     const start = method.start;
-    if (method.params.length !== paramCount) {
+    // Allow `this` type annotation in TypeScript
+    if (
+      method.params.length === paramCount + 1 &&
+      method.params[0].type === "Identifier" &&
+      method.params[0].name === "this"
+    ) {
+      this.expectPlugin("typescript");
+    } else if (method.params.length !== paramCount) {
       if (method.kind === "get") {
         this.raise(start, "getter must not have any formal parameters");
       } else {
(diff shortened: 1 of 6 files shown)


packages/babel-parser/src/parser/expression.js
```diff
@@ -1552,15 +1552,25 @@ export default class ExpressionParser extends LValParser {
...
     const paramCount = method.kind === "get" ? 0 : 1;
     const start = method.start;
-    if (method.params.length !== paramCount) {
+    // Allow `this` type annotation in TypeScript
+    if (
+      method.params.length === paramCount + 1 &&
+      method.params[0].type === "Identifier" &&
+      method.params[0].name === "this"
+    ) {
+      this.expectPlugin("typescript");
```
nicolo-ribaudo: When the first parameter is `this` the new implementation doesn't check the parameters count because the `} else if (method.params.length !== paramCount) {` branch is skipped:

```js
get prop(this: {}, foo, bar) {}
```

If we parsed `this` as a function parameter, then this plugin is already enabled (no need to assert). I would prefer to have as few as possible TypeScript-specific logic in the `parser/*` files. WDYT about extractions `method.kind === "get" ? 0 : 1` to a new `getAccessorsExpectedParamCount(method)` function and overwrite it in the typescript plugin?

danez: `method.params.length === paramCount + 1` wouldn't be true in that case I guess?

nicolo-ribaudo: Oh right :man_facepalming:

matt-tingen: @nicolo-ribaudo I have pushed a change for `getGetterSetterExpectedParamCount`. Thanks for suggesting the separation. I hadn't noticed I had the first usage of `this.expectPlugin("typescript")`.

At first, it felt a little odd that the error messages directly reference the required number of params despite that number being abstracted away. I considered extracting a `checkGetterSetterExpectedParamCount` method instead, but didn't find a good way to do so without duplicating logic in the TypeScript plugin. However, I found that there's less of a dissonance if I look at it as the error messages being based on target language and `getGetterSetterExpectedParamCount` being based on the source language, so I'm on board now 😄

nicolo-ribaudo:
Thanks!

matt-tingen:
My pleasure! Thanks for the quick reviews!

christianalfoni:
Hm, is this part of the latest 7.3.3 release? Still getting:

`getter must not have any formal parameters`

errors on this? Might be our setup, but yeah, just wanted to ask :)

matt-tingen:
@christianalfoni This was released in [7.4.0](https://github.com/babel/babel/releases/tag/v7.4.0). It's still working for me in the REPL.

christianalfoni:
Oh, hm, it seems that @babel/preset-typescript is not keeping up with latest versions?

matt-tingen:
@christianalfoni I'm not sure how the auxiliary packages are versioned. Most of them seem to keep in sync with core, but some may not need any changes for a given core bump.
In my project, we are using `@babel/core` `7.4.0` and `@babel/preset-typescript` `7.3.3`. If you need help getting it set up, I would suggest posting on StackOverflow.

nicolo-ribaudo:
We only publish updated packages when they are actually changed. `@babel/preset-typescript` hasn't changed, but only one of its dependencies: `@babel/plugin-transform-typescript`.

One of these commands should work:
```js
npm update @babel/plugin-transform-typescript

yarn upgrade @babel/plugin-transform-typescript
```

PR #9581, opened 2019-02-25
Parenthesized expressions (#8025)

nicolo-ribaudo:
This has been merged at https://github.com/babel/babel/pull/8025 and reverted by https://github.com/babel/babel/pull/9569 to release a patch version.
It already has 4 :heavy_check_mark:, so there is no need to review it again; it is just waiting for v7.4.0 :wink:

Sorry @arv for the confusion.

PR #9644, opened 2019-03-06
Do not remove comments

arv:
The fix for https://github.com/eslint/espree/issues/158 has an error; If
`lastChild === node` it ended up removing comments.

| Fixed Issues?            | For some reason I cannot test this in isolation
| Patch:          | Bug Fix
| Tests Added + Pass?      | No. It is not clear why I'm hitting this code path

The fix for https://github.com/eslint/espree/issues/158 incorrectly removes comments under some circumstances.

I can only repro this with #9581  applied but the code is clearly wrong. It should not remove the comment like it did

### DIFF
--- a/packages/babel-parser/src/parser/comments.js
+++ b/packages/babel-parser/src/parser/comments.js
@@ -141,9 +141,8 @@ export default class CommentsParser extends BaseParser {
     }
 
     if (lastChild) {
-      if (lastChild.leadingComments) {
+      if (lastChild.leadingComments && lastChild !== node) {
         if (
-          lastChild !== node &&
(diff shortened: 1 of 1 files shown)


arv:
I clearly need to update some expectations...

nicolo-ribaudo:
After merging #9581 it will be possible to write a test?

arv:
Thanks. I should be able to test this now.

arv:
I still cannot repro the problem here. It might be related to me only using HEAD for `@babel/parser` but I don't see how?

Closing this to reduce code churn until I find a test case that fails without these changes.

PR #9636, opened 2019-03-05
Update babel-plugin-proposal-unicode-property-regex for Unicode v12

mathiasbynens:
https://github.com/tc39/ecma262/issues/1467
https://github.com/tc39/ecma262/pull/1468

PR #9642, opened 2019-03-06
Update regexpu-core dependency

mathiasbynens:
```sh
yarn upgrade regexpu-core@4.5.3
```

This should resolve the issue @danez mentions here:
https://github.com/babel/babel/pull/9636#issuecomment-469870217

PR #9647, opened 2019-03-07
Add WarningsToErrorsPlugin to webpack to avoid missing build problems on CI

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

This should break the build if there are warnings in the webpack build, which we currently have because of #9636 
Let's see.


nicolo-ribaudo:
It is fixed by https://github.com/babel/babel/pull/9642, right?

danez:
Yes

Issue #8723, opened 2018-09-18
@babel/template crash when trying to parse template literal with an upper case variables


AlexMost:
## Bug Report

**Current Behavior**
Execution of template with a template literal with uppercase variable `DDD` - crashes. If I am changing variable name to `ddd` - it works.

**Input Code**
Thist code crashes in the node repl:

```js
const tpl = require('@babel/template').default;
tpl('`test test ${DDD}`')()
```

this one works:
```js
const tpl = require('@babel/template').default;
tpl('`test test ${ddd}`')()
```

**Expected behavior/code**
Should create an object with `ast` property.

**Babel Configuration (.babelrc, package.json, cli command)**
Was running this code in a repl without any additional babel configurations.

**Environment**
- Babel version(s): 7.1.0
- @babel/template 7.1.0
- Node/npm version: Node 8.9.1/npm 5.5.1
- OS: Ubuntu 14.04
- Monorepo [no]

**Additional context/Screenshots**
That actually worked in version `6.x` and started failing in `7.x`. 

In my case, I was able to fix that with calling the `ast` func:

```
const tpl = require('@babel/template').default;
tpl.ast('`test test ${DDD}`')
```

Stack trace:
```
TypeError: @babel/template placeholder "DDD": Property expressions[0] of TemplateLiteral expected node to be of a type ["Expression"] but instead got null
    at validate (/home/a.mostovenko/git/babel-plugin-ttag/node_modules/@babel/types/lib/definitions/utils.js:128:13)
    at validator (/home/a.mostovenko/git/babel-plugin-ttag/node_modules/@babel/types/lib/definitions/utils.js:97:7)
    at Object.validate (/home/a.mostovenko/git/babel-plugin-ttag/node_modules/@babel/types/lib/definitions/utils.js:172:7)
    at Object.validate (/home/a.mostovenko/git/babel-plugin-ttag/node_modules/@babel/types/lib/validators/validate.js:17:9)
    at applyReplacement (/home/a.mostovenko/git/babel-plugin-ttag/node_modules/@babel/template/lib/populate.js:130:9)
    at metadata.placeholders.slice.reverse.forEach.placeholder (/home/a.mostovenko/git/babel-plugin-ttag/node_modules/@babel/template/lib/populate.js:42:7)
    at Array.forEach (<anonymous>)
    at populatePlaceholders (/home/a.mostovenko/git/babel-plugin-ttag/node_modules/@babel/template/lib/populate.js:40:43)
[...]

Issue #8067, opened 2018-05-28
@babel/template tries to replace uppercase strings


djfarly:
# v7 Regression

> First check out: https://new.babeljs.io/docs/en/next/v7-migration.html
> Also a partial upgrade tool: https://github.com/babel/babel-upgrade

**Potential Commit/PR that introduced the regression**
Somewhere between v6 and v7 - not quite sure though… also the cafe is closing so i don't have time to investigate further, sorry! 

**Describe the regression**
AST generation fails for ALLCAPS strings. I think it should not.  🤔 

**Input Code**

[Reduced test case here.](https://repl.it/repls/GaseousMindlessTests)
```js
var template = require("@babel/template");

const string = "var a = { a: 'NormalString', b: 'UPPERCASESTRING' };"

console.log(template.default(string)())
```
This throws:
```
Error: @babel/template placeholder "UPPERCASESTRING": Expected string substitution
```

[Working v6 here.](https://repl.it/repls/ObeseHotpinkAddition)
```js
var template = require("babel-template");

const string = "var a = { a: 'NormalString', b: 'UPPERCASESTRING' };"

console.log(template(string)())
```
This works flawlessly.

**Babel Configuration (.babelrc, package.json, cli command)**
none

**Expected behavior/code**
It should create an AST normally, taking into account that 'UPPERCASESTRING' is just a string.

**Environment**
- Babel version(s): In REPL 7.0.0-beta.49
- Node/npm version: Tested in node 9 + 10
- OS: OSX (local) + linux/amd64 (REPL)
- Monorepo (does not matter?)
- How you are using Babel: babel/template directly in [babel-plugin-codegen](https://github.com/kentcdodds/babel-plugin-codegen)

Thanks a ton! 🎉 ❤️

loganfsmyth:
This was a breaking change in the API, but it was purposeful. If you're interested you can read over the docs in the README. https://github.com/babel/babel/blob/7846eaebaa69662134782b35d9bdff9b71616aa1/packages/babel-template/README.md

In this case, you have two easy options:

* Use `placeholderPattern: false` so it stops searching for replacement patterns entirely.
  * ```console.log(template(string, { placeholderPattern: false })())```
* Use an ES6 template literal with `.ast`
  * ```console.log(template.ast`var a = { a: 'NormalString', b: 'UPPERCASESTRING' };`)```

djfarly:
Awesome! Thank you @loganfsmyth.
That was rather quick. 🤐👍

PR #9648, opened 2019-03-07
Add %%placeholders%% support to @babel/template

nicolo-ribaudo:
| Fixed Issues?            | Not really, but the old behavior was confusing to some people: #8723, #8067, #7035
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | https://github.com/babel/website/pull/1988
| Any Dependency Changes?  |

This is the last step to make https://github.com/babel/babel/pull/9364 usable in Babel. I'm sorry for opening this PR so late, but I hope to get it in v7.4.

In this PR I added a new option to `@babel/template`, `syntacticPlaceholders: ?boolean`, which toggles between `%%foo%%` placeholders (when `true`) and `FOO` placeholders. If it isn't specified, Babel tries to be "smart" to avoid breaking backward compat: if `%%foo%%` is used `syntacticPlaceholders` defaults to `true`, otherwise to `false`.

0e58e252913efe84eba926cc9c9c19fb18d5c620 commit shows how some templates we used could be simplified by using this new placeholders syntax (we can't actually do it yet because we are importing `template` from `@babel/core` which could be an older version).

NOTE: Since I wanted to keep this PR as small as possible to make it easier to review, I didn't migrate `template.ast` to internally use the new syntax. It is an implementation detail, so it will be possible to change it in a patch release.

### DIFF
--- a/packages/babel-template/src/literal.js
+++ b/packages/babel-template/src/literal.js
@@ -74,6 +74,7 @@ function buildLiteralData<T>(
       ),
       placeholderPattern: opts.placeholderPattern,
       preserveComments: opts.preserveComments,
+      syntacticPlaceholders: opts.syntacticPlaceholders,
     });
   } while (
     metadata.placeholders.some(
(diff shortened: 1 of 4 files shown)


nicolo-ribaudo:
TODO: Docs

PR #9650, opened 2019-03-07
fix: Don't transpile ES2018 symbol properties

taion:
| Fixed Issues?            | #4783 (again)
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | No, per #5195
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is a repeat of #5195 to work around the same upstream issue.

### DIFF
--- a/packages/babel-plugin-transform-runtime/scripts/build-dist.js
+++ b/packages/babel-plugin-transform-runtime/scripts/build-dist.js
@@ -19,7 +19,15 @@ writeHelpers("@babel/runtime-corejs2", { corejs: 2 });
 function writeCoreJS2(runtimeName) {
   const pkgDirname = getRuntimeRoot(runtimeName);
 
-  const paths = ["is-iterable", "get-iterator"];
+  const paths = [
+    "is-iterable",
+    "get-iterator",
+
+    // This was previously in definitions, but was removed to work around
+    // zloirock/core-js#262. We need to keep it in @babel/runtime-corejs2 to
+    // avoid a breaking change there.
+    "symbol/async-iterator",
+  ];
 
   Object.keys(corejs2Definitions.builtins).forEach(key => {
     const path = corejs2Definitions.builtins[key];
--- a/packages/babel-plugin-transform-runtime/src/definitions.js
+++ b/packages/babel-plugin-transform-runtime/src/definitions.js
@@ -109,7 +109,7 @@ export default runtimeVersion => {
         : {}),
 
       Symbol: {
-        // FIXME: Disabled to work zloirock/core-js#262.
+        // FIXME: Disabled to work around zloirock/core-js#262.
         // asyncIterator: "symbol/async-iterator",
         for: "symbol/for",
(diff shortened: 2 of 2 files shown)


zloirock:
It could be a breaking change. If someone uses old transform and new runtime, entry point `@babel/runtime/core-js/symbol/async-iterator` will be missed.

taion:
@zloirock It won't be – the runtime transform will see that `Symbol` gets pulled in, and it will import the full `symbol` polyfill, which will work as before. this was the behavior in babel v6 with #5195.

taion:
And, yeah, ideally we'll kill this hack for good once we upgrade to core-js v3.

nicolo-ribaudo:
We should hardcode it in `build-dist`, so that it isn't removed from the package.

taion:
@nicolo-ribaudo Hardcode what?

taion:
The reason this isn't a breaking change is because we get the "full" `Symbol` polyfill rather than just the one for `Symbol.asyncIterator`.

In the below, note that it's now `_Symbol.asyncIterator` instead of `_Symbol$asyncIterator`. But this is fine – that's properly defined and present.

#### Before
```js
var _Symbol$iterator = require("../core-js/symbol/iterator");

var _Symbol$asyncIterator = require("../core-js/symbol/async-iterator");

var _Symbol = require("../core-js/symbol");

function _asyncIterator(iterable) {
  var method;

  if (typeof _Symbol === "function") {
    if (_Symbol$asyncIterator) {
      method = iterable[_Symbol$asyncIterator];
      if (method != null) return method.call(iterable);
    }

    if (_Symbol$iterator) {
      method = iterable[_Symbol$iterator];
      if (method != null) return method.call(iterable);
    }
  }

  throw new TypeError("Object is not async iterable");
}

module.exports = _asyncIterator;
```

#### After
```js
var _Symbol$iterator = require("../core-js/symbol/iterator");

var _Symbol = require("../core-js/symbol");

function _asyncIterator(iterable) {
  var method;

  if (typeof _Symbol === "function") {
    if (_Symbol.asyncIterator) {
      method = iterable[_Symbol.asyncIterator];
      if (method != null) return method.call(iterable);
    }

    if (_Symbol$iterator) {
      method = iterable[_Symbol$iterator];
      if (method != null) return method.call(iterable);
    }
  }

  throw new TypeError("Object is not async iterable");
}

module.exports = _asyncIterator;
```

zloirock:
I mean that someone could have `@babel/transform-runtime` bound to, for example, 7.3 (code sample **Before**) and `@babel/runtime@7.4` without `@babel/runtime/core-js/symbol/async-iterator` entry point.

nicolo-ribaudo:
We should force it to be included like we do for `is-iterator`: https://github.com/babel/babel/blob/b2b96e0b150b56d47fbe5ec1ef7727a13e86a899/packages/babel-plugin-transform-runtime/scripts/build-dist.js#L22

taion:
Makes sense. Done.

PR #4843, opened 2016-11-15
Add windows support for build and test.

garyjN7:
Added scripts to package.json to allow build and test to be used on Windows platforms

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | no
| Fixed tickets     | none
| License           | MIT
| Doc PR            | none

@danez I accidently deleted the original pull request for this (#4752) hence this new one.  Also, since I created the original, new functionality was added to the makefile which I have since added to the package.json scripts.

PR #5309, opened 2017-02-14
Bootstrap Babel on npm install

chicoxyzzy:
| Q                        | A 
| ------------------------ | ---
| Patch: Bug Fix?          | no
| Major: Breaking Change?  |no 
| Minor: New Feature?      | no
| Deprecations?            | no
| Spec Compliancy?         | no
| Tests Added/Pass?        | no
| Fixed Tickets            | Fixes #5299
| License                  | MIT
| Doc PR                   | 
| Dependency Changes       |

PR #9674, opened 2019-03-13
Discussion: Make developing babel on Windows easier

danez:
We are still `make` for a lot of tasks we run for publish, build, test, etc right now. As we all know `make` is not necessarily a native thing on Windows and it has been brought up a couple of times internally as well as in pull requests/issues that this is not an ideal situation for Windows users. 

We had a PR (#4843) before which migrated the make file to npm scripts. This worked, but did not solve the problem completely, as bash - another non windows native -  was still required. And imho it was also not very easy to read and understand.

In our build and bundle process we  use `gulp` hidden behind `make`. I'm not sure where using two tools originated, but looks like another reason to me to tackle this topic and maybe settle on one tool to orchestrate all the tasks.

As `gulp` was already setup it was simple for me to see how gulp would work running without `make`.

### POC `make` => `gulp`

> This is a proof of concept and does not cover all our cases.

This PR is a POC how `gulp` could potentially be used to run all our tasks in a structured way. All I did in this PR is to come up with an infrastructure for tasks to be defined in `gulp` without having to have everything in one big file.
I tried different things like `gulp-hub` for example, which did not work with the recommended way of defining gulp tasks (which is simply exporting them). 
I ended up creating a folder where each file is a gulp task. This tasks are then loaded in the main gulpfile and simply reexported.

I converted the main build as well as the rollup task. Additionally I added a `test` task which does right now exactly what the shell script did. This task is to demonstrate how we can run shell scripts from within `gulp`

>Now this might not make much sense now that we have jest. We could simply add some npm scripts to `package.json` (`"test": "jest"`, `"test:ci": "jest --ci --maxWorkers=4"`, `"test-inspect": "node --inspect-brk ./node_modules/jest/bin/jest.js --runInBand"`). This would cover all cases and make use of the `jest` cli instead of our `TEST_*` variables like `yarn test ./packages/babel-parser/` instead of `TEST_ONLY=babel-parser yarn test`

Other cli calls like `rm` and `mkdir` can be replaced with npm packages or node builtins. The only requirement left that I can currently see would be `git`, which everyone who has babel cloned obviously has. And `git` is only used for very specific tasks anyway (test262, flow tests and probably publish).

### Other task runners

Now I haven't looked at any other tools, but most that came to my mind require developers to install certain tool (e.g. `ant` or `gradle`) and I guess this is not any better, and require yet another language/syntax to handle. So having a tool where the configuration is in javascript seems like a big plus. Alternatives I know for `gulp` are `grunt` (which seems outdated and barely maintained) or `broccoli` (not sure how popular that is). To my knowledge `gulp` is the most prominent one.

### Questions

- Are there any other task runners we should consider or look at? 
- The approach I took for gulp is that the right direction?
- Does it make sense to have certain simple tools directly in `package.json`? Like `jest`? Or should we create a gulp task for everything no matter what?
- Is there something I missed? Anything that could be hard to do in `gulp`?
- Are there any *better* patterns to structure gulp tasks?
- Are there any big projects we could look at that solved this in a nice way?

> I'm happy for any input no matter if you are babel maintainer or first time contributor.

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -1,125 +1,16 @@
 "use strict";
-
-const plumber = require("gulp-plumber");
-const through = require("through2");
-const chalk = require("chalk");
-const newer = require("gulp-newer");
-const babel = require("gulp-babel");
-const gulpWatch = require("gulp-watch");
-const fancyLog = require("fancy-log");
-const filter = require("gulp-filter");
 const gulp = require("gulp");
-const path = require("path");
-const webpack = require("webpack");
-const merge = require("merge-stream");
-const rollup = require("rollup");
-const rollupBabel = require("rollup-plugin-babel");
-const rollupNodeResolve = require("rollup-plugin-node-resolve");
-const { registerStandalonePackageTask } = require("./scripts/gulp-tasks");
+const requireDir = require("require-dir");
 
-const sources = ["codemods", "packages"];
+const tasks = requireDir("./scripts/gulp/tasks/");
+module.exports = exports = tasks;
 
-function swapSrcWithLib(srcPath) {
-  const parts = srcPath.split(path.sep);
-  parts[1] = "lib";
-  return parts.join(path.sep);
-}
+exports.build = gulp.parallel(exports.bundle, exports.buildCommonJs);
+exports.default = exports.build;
 
-function getGlobFromSource(source) {
-  return `./${source}/*/src/**/*.js`;
-}
+//gulp.task("build-no-bundle", () => buildBabel());
 
-function getIndexFromPackage(name) {
-  return `${name}/src/index.js`;
-}
-
-function compilationLogger() {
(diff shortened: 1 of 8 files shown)


danez:
> FWIW I know eslint uses `shelljs/make` https://github.com/eslint/eslint/blob/06fa1655c3da8394ed9144d727115fc434b0416f/Makefile.js
> Don't know much about their experience though.

Oh interesting, so we could directly translate the Makefile to `shelljs/make`, that might be even less work.

hzoo:
I'm cool with gulp if it makes it possible for Windows and easier for contributors.

> Does it make sense to have certain simple tools directly in package.json? Like jest? Or should we create a gulp task for everything no matter what?

In theory it would be best if anyone can just use the same commands for other projects like `yarn/npm test` etc. We never ended up figuring out https://github.com/babel/babel/pull/5309 either

PR #9685, opened 2019-03-14
Only load es.symbol.description when using it.

nicolo-ribaudo:
:warning: This PR isn't targeting `master`, but the `update-to-core-js-3` branch.

@zloirock This was the change I was proposing earlier today on Slack. I don't think that this can cause missing polyfills or polyfills imported in the wrong order.

1)  `Symbol()`, chrome 40
    Before: es.symbol, es.symbol.description, es.object.to-string
    After: es.symbol, es.object.to-string

2) `Symbol()`, chrome 60
   Before: es.symbol.description
   After: (none)

3) `foo.description`, chrome 40
   Before: es.symbol, es.symbol.description
   After: es.symbol, es.symbol.description

3) `foo.description`, chrome 60
   Before: es.symbol.description
   After: es.symbol.description

I tried to do the same for Map/Set proposal methods, but I couldn't because whenever someone used the `[].filter` method it would have included the `es.map` polyfill (even if proposals were disabled).

### DIFF
--- a/packages/babel-preset-env/src/polyfills/corejs3/built-in-definitions.js
+++ b/packages/babel-preset-env/src/polyfills/corejs3/built-in-definitions.js
@@ -53,11 +53,7 @@ const PromiseDependenciesWithIterators = [
   ...CommonIterators,
 ];
 
-const SymbolDependencies = [
-  "es.symbol",
-  "es.symbol.description",
-  "es.object.to-string",
-];
+const SymbolDependencies = ["es.symbol", "es.object.to-string"];
 
 const MapDependencies = [
   "es.map",


zloirock:
As I said, it will not work for `es.symbol.description`. `Symbol#description` requires `Symbol` [constructor wrapper](https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/es.symbol.description.js#L17-L25), so we should load a polyfill on symbol creation.
```js
// file-1
require('core-js/modules/es.symbol');
module.exports = Symbol('');

// file-2
require('core-js/modules/es.symbol');
require('core-js/modules/es.symbol.description');
const symbol = require('./file-1');
console.log(symbol.description); // => undefined, but should be ''
```

zloirock:
If you wan't to load by this way collections methods - it will work - we can add some hacks for don't add not required modules without `proposals` flag. But, for example,
```js
something.filter
```
with `proposals` flag will be transpiled at least to 
```js
import "core-js/modules/es.array.filter";
import "core-js/modules/es.array.iterator";
import "core-js/modules/es.map";
import "core-js/modules/es.object.to-string";
import "core-js/modules/es.set";
import "core-js/modules/es.string.iterator";
import "core-js/modules/es.typed-array.filter";
import "core-js/modules/es.typed-array.to-string";
import "core-js/modules/esnext.map.filter";
import "core-js/modules/esnext.set.filter";
import "core-js/modules/web.dom-collections.iterator";

something.filter
```

It would be better to leave it on constructor detection.

zloirock:
BTW
```js
[].filter
```
anyway will be transpiled to 
```js
import "core-js/modules/es.array.filter";

[].filter
```
since we know what the source is an array.

nicolo-ribaudo:
> As I said, it will not work for es.symbol.description. Symbol#description requires Symbol constructor wrapper

Oh I tought that it was only needed for native symbols :man_facepalming:

Issue #6872, opened 2017-11-22
InitializeInstanceFields should be called before returning from super call


gsathya:
> Choose one: is this a bug report or feature request?

bug

### Input Code

```js
var log = [];

class Foo extends class {} {
  x = log.push(1);

  constructor() {
    super();
    super();
  }
}

try {
	new Foo;
} catch(e) {}

console.log(log)
```

### Babel/Babylon Configuration (.babelrc, package.json, cli command)

stage3,stage2

### Expected Behavior

`[1, 1]`

### Current Behavior

`[1]`

### Your Environment

[REPL](https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBANg9g5hAvBA2gXQNwCgsGMogDOhEAYjDBAKYAeALlQHYAmJ-RJA3gL4SdYQINJNHgA6AA4BXQgAsAFAEYAlNgERcMRoTpgpuOjDDzlfdYMJSJVY6vMRL129kHcsbrLoCeZgJCMqAHcyCmxeXBA6XAUqUx4cTW0YKCoxWDh5dOUgA&debug=false&circleciRepo=&evaluate=true&lineWrap=false&presets=stage-2%2Cstage-3&targets=&version=7.0.0-beta.32)

xtuc:
ah good catch!

Does it make sense to call `super` multiple times? Otherwise we should keep a state in the transformation to know if we already initialized stuff.

gsathya:
> Does it make sense to call super multiple times?

Nope. It's an error. But it should error out *after* the field initializers are run.

xtuc:
Ok so we should probably transform the second super call to a throw statement.

gsathya:
Except you still want to call the super constructor.

```
var i = 0;

class X {
  constructor() { i++; } 
}

class P extends X {
  constructor() {
    super();
    super();
  }
}

try { new P; } catch(e) {}
console.log(i);
```

This should print `2`

nicolo-ribaudo:
@xtuc It isn't always that easy:

e.g.
```js
class A {
  constructor() {
    if (Math.random() > .5) super();
    if (Math.random() > .5) super();
  }
}
```

The problem is that `_possibleConstructorReurn` throws before the initializers are executed, while the correct order should be (to my understanding)
1. Call the super constructor
2. Run properties initializers
3. If `this` was already defined, throw an error
(currently 2 and 3 are swapped)

[...]

PR #9686, opened 2019-03-14
Add test for checking duplicate super call after initialization

thiagoarrais:
| Fixed Issues?            | Relates to #6872
| Documentation PR Link    | no

This just adds @gsathya's example code from #6872 as a test case to (a) show that the bug is already fixed and (b) prevent regressions.

### DIFF
--- a/packages/babel-plugin-proposal-class-properties/test/fixtures/public/check-dup-after-initialize/exec.js
+++ b/packages/babel-plugin-proposal-class-properties/test/fixtures/public/check-dup-after-initialize/exec.js
@@ -1,0 +1,18 @@
+// from issue 6872
+
+var log = [];
+
+class Foo extends class {} {
+  x = log.push(1);
+
+  constructor() {
+    super();
+    super();
+  }
+}
+
+try {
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
I'm really surprised that this works.

thiagoarrais:
Added the i/o fixture. Will take a look into not transpiling later.

That's the second time we find something that shouldn't work but does. I'm already getting used to it. :smile:

thiagoarrais:
The issue shows up when not transpiling classes. I've added a failing test case and will work on fixing it next.

nicolo-ribaudo:
Please only fix it when a `spec: true` option is enabled, since 
<details>
<summary>Spoiler</summary>
it will require wrapping every `super()` with `try/finally`.
</details>

nicolo-ribaudo:
Note they initializers should run always after `super()` (even if it is the second one), but before that it throws.

thiagoarrais:
Will augment the test with that, then... Can you point to that specific behavior in the spec?

PR #9689, opened 2019-03-15
Update index.js

ikatun:
I'm working on a babel plugin to support decorators which emit metadata and I need to keep imports of classes which are stripped by this function.

Also, this might be useful to people who import a class for both side effects and to a type reference.

| Fixed Issues?            | `No issue reported` 
| Patch: Bug Fix?          |
| Minor: New Feature?      | New Feature
| Documentation PR Link    | 

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -22,7 +22,7 @@ interface State {
 
 const PARSED_PARAMS = new WeakSet();
 
-export default declare((api, { jsxPragma = "React" }) => {
+export default declare((api, { jsxPragma = "React", keepAllImports: false }) => {
   api.assertVersion(7);
 
   const JSX_ANNOTATION_REGEX = /\*?\s*@jsx\s+([^\s]+)/;
@@ -53,7 +53,7 @@ export default declare((api, { jsxPragma = "React" }) => {
 
         // remove type imports
         for (const stmt of path.get("body")) {
-          if (t.isImportDeclaration(stmt)) {
+          if (!keepAllImports && t.isImportDeclaration(stmt)) {
             // Note: this will allow both `import { } from "m"` and `import "m";`.
             // In TypeScript, the former would be elided.
             if (stmt.node.specifiers.length === 0) {


packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -22,7 +22,7 @@ interface State {
 
 const PARSED_PARAMS = new WeakSet();
 
-export default declare((api, { jsxPragma = "React" }) => {
+export default declare((api, { jsxPragma = "React", keepAllImports: false }) => {
```
danez:
```suggestion
export default declare((api, { jsxPragma = "React", keepAllImports = false }) => {
```

nicolo-ribaudo:
> I'm working on a babel plugin to support decorators which emit metadata and I need to keep imports of classes which are stripped by this function.

Can you expand a bit? If your plugin is injecting some usage of the class, you should update the scope to mark it as used: the TS plugin won't remove it.

nicolo-ribaudo:
Ok, my idea doesn't work because type imports are removed before visiting any other node.
I'd prefer to modify the typescript plugin to remove imports on `Program:exit` (instead of `Program:enter`), and then you can do something like this in your plugin:

https://runkit.com/embed/63nqqxt8dfje

ikatun:
Thanks for the correction and an example about updating the scope. I didn't know you could do that :)

Could removing the imports on the Program:exit instead of enter cause breaking changes somewhere?
If not, when could the change be implemented?

As for the expanding part of your question, I want to switch from tsc to babel on a typescript project we're working in my company because babel is faster by a factor of 10. However, we're using https://github.com/typeorm/typeorm and https://github.com/19majkel94/type-graphql which rely heavily on runtime reflection so I've implemented a babel plugin to reproduce the exact same calls for `Refect.metadata` injected by tsc.
These calls to Reflect.metadata fail when imports for types used in them are removed.

nicolo-ribaudo:
> Could removing the imports on the Program:exit instead of enter cause breaking changes somewhere?

Yes and no.
The [`imports/elide-injected`](https://github.com/babel/babel/blob/a35e5a314ab533828d0988a2da824edb04fc8d3c/packages/babel-plugin-transform-typescript/test/fixtures/imports/elide-injected/) would probably fail: as you can see, it is adding to the scope the import statement but not its usage: https://github.com/babel/babel/blob/a35e5a314ab533828d0988a2da824edb04fc8d3c/packages/babel-plugin-transform-typescript/test/fixtures/imports/elide-injected/plugin.js#L19-L22

I think that it is a bug in that test plugin: when injecting a new node, the scope should be updated accordingly. It might work without updating it, but then you can't complain if another plugin relying on the scope breaks your output.

cc @jeysal

> If not, when could the change be implemented?

You should split the `Program` visitor in the TS plugin. Instead of
```js
Program(path, state) {
  // pragma stuff
  // imports stuff
}
```
it should be
```js
Program: {
  enter(path, state) {
    // pragma stuff
  },
  exit(path, state) {
    // imports stuff
  }
}
```

---

Until the TypeScript plugin is fixed (I'm not sure if it should be done in a patch release or in a major release), I suggest injecting the new references before letting the TS plugin visit `Program`: this will be a little slower than how it would be done after the fix, but it will have the same behavior.

https://runkit.com/nicolo-ribaudo/use-imports-elided-by-ts

I don't like the `keepAllImports` because it's not something that TypeScript does, and it seems like it is only here to workaround a bug in Babel rather than because of a real need.

jeysal:
@nicolo-ribaudo Thanks for notifying me. IIUC with the changes to `transform-typescript`, `elide-injected` would fail due to `transform-typescript` eliding the import at `Program.exit` because the missing scope update makes it look like it is unused. However, if we add an additional `crawl` (or `registerBinding`) to the test between L21 and L22, it will work properly again.

If this matches your understanding then I think it's fine to update the test accordingly. The plugin should indeed make sure to update the scope after L21, I just didn't think about that when I wrote the test since its purpose was to test the behavior between adding the import and referencing it.

Of course there's always a small chance of breakage, in this case e.g. for plugins trying to add references in `Program.exit` after `transform-typescript`, but IIUC at least it would not be considered breaking *because of* the `elide-injected` test.

ikatun:
Great, thanks @nicolo-ribaudo for explanations and code snippets, I'll use the suggested approach with seperate plugin to add those references. I suppose that plugins will run before presets so I'll be able to use `@babel/preset-typescript`.

I guess I can close this pull request then?

nicolo-ribaudo:
I have opened #9991 to fix https://github.com/babel/babel/pull/9689#issuecomment-473510963 in Babel 8.

PR #9692, opened 2019-03-15
Fix TSFunctionType visitors definition

### LINKED ISSUE #9691: TSFunctionType doesn't specify parameters as a visitor
## Bug Report

**Current Behavior**
When traversing a tree parsing for TypeScript syntax and hitting a `TSFunctionType` node, the `typeParameters` and `typeAnnotation` fields are correctly visited but the `parameters` field isn't. As a result visitors by default don't visit `foo` below:

**Input Code**
```js
var x: (foo) => void; // foo is never visited
```
```
module.exports = function() {
  return {
    visitor: {
      Identifier(path) {
        if (path.node.name == "foo") {
          // Never hit because it's nested within TSFunctionType.parameters
          path.node.name = "bar";
        }
      }
    }
  };
}
```

**Expected behavior/code**
It appears to be a bug in `babel-types/src/definitions/typescript.js` which omits `parameters` in the visitors list for `fnOrCtr`.

```js
{
  "plugins": ["syntax-typescript", "./plugin"]
}
```

**Environment**
- Babel version(s): v7.3.4
- Node/npm version: Node 8.10/npm 3.5.2
- OS: Windows 10

**Possible Solution**
Add `parameters` to `visitor` array in `fnOrCtr` (in [https://github.com/babel/babel/blob/master/packages/babel-types/src/definitions/typescript.js](typescript.js)).

penielse:
| Fixed Issues?            | Fixes #9691
| Patch: Bug Fix?          | Unsure what the convention here is - it is a bug fix but it can break existing transforms if they assume the incorrect behavior that `TSFunctionType.parameters` shouldn't be visited.
| Major: Breaking Change?  |
| Minor: New Feature?      |

When traversing a tree parsing for TypeScript syntax and hitting a `TSFunctionType` node, the `typeParameters` and `typeAnnotation` fields are correctly visited but the `parameters` field isn't. As a result visitors by default don't visit `foo` below:

```
var x: (foo) => void; // foo is never visited
```
```
module.exports = function() {
  return {
    visitor: {
      Identifier(path) {
        if (path.node.name == "foo") {
          // Never hit because it's nested within TSFunctionType.parameters
          path.node.name = "bar";
        }
      }
    }
  };
}
```

It appears to be a bug in babel-types/src/definitions/typescript.js which omits `parameters` in the visitors list for `fnOrCtr`. Fixed by adding it.

### DIFF
--- a/packages/babel-traverse/test/fixtures/regression/visit-tsfunctiontype-parameters/input.ts
+++ b/packages/babel-traverse/test/fixtures/regression/visit-tsfunctiontype-parameters/input.ts
@@ -1,0 +1,1 @@
+let x: (number) => void;
--- a/packages/babel-traverse/test/fixtures/regression/visit-tsfunctiontype-parameters/options.json
+++ b/packages/babel-traverse/test/fixtures/regression/visit-tsfunctiontype-parameters/options.json
@@ -1,0 +1,3 @@
+{
+  "plugins": ["syntax-typescript", "./plugin"]
+}
--- a/packages/babel-traverse/test/fixtures/regression/visit-tsfunctiontype-parameters/output.js
+++ b/packages/babel-traverse/test/fixtures/regression/visit-tsfunctiontype-parameters/output.js
@@ -1,0 +1,1 @@
+let x: (string) => void;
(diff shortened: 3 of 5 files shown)


nicolo-ribaudo:
Yead I'd consider this a bug fix.

Issue #8172, opened 2018-06-15
TypeScript abstract classes break decorators, abstract keyword remains in compiled code


AlexanderOMara:
Bug report.

I wasn't sure if I should open a new bug or not, but I previously created [a bug report about how Babel was unable to compile decorators on TypeScript abstract classes](https://github.com/babel/babel/issues/7837). Since that update has been released, today I tried it out, and while it now compiles the code, it doesn't quite work due to it leaving the `abstract` keyword in the compiled code.

### Input Code
[GitHub Gist with all code](https://gist.github.com/AlexanderOMara/9301d66ac9f147a1940432f835318fbb).

```js
function decorate() {
	return (Class: Function) => {
		console.log(Class);
	};
}

@decorate()
abstract class Test extends Object {
	constructor() {
		super();
	}
}
```

### Babel/Babylon Configuration (.babelrc, package.json, cli command)

```js
{
	"presets": [
		["@babel/preset-env", {
			"modules": "commonjs",
			"targets": {
				"node": "10"
			}
		}],
		["@babel/preset-typescript"]
	],
	"plugins": [
		["@babel/plugin-proposal-decorators", {
			"legacy": true
		}]
	]
}
```

### Expected Behavior
Expect this code to compile to valid code successfully, like it does when using the official TypeScript compiler.

### Current Behavior

Attempting to run the code produces a syntax error:

```
$ node index.js
.../index.js:11
let Test = (_dec = decorate(), _dec(_class = abstract class Test extends Object {
                                             ^^^^^^^^

SyntaxError: missing ) after argument list
    at new Script (vm.js:74:7)
    at createScript (vm.js:246:10)
    at Object.runInThisContext (vm.js:298:10)
    at Module._compile (internal/modules/cjs/loader.js:670:28)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
    at Function.Module._load (internal/modules/cjs/loader.js:543:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:744:10)
    at startup (internal/bootstrap/node.js:238:19)
```

### Possible Solution
N/A

### Context
Just trying to use decorators on abstract classes, which should be possible in TypeScript.

### Your Environment

| software         | version(s)
| ---------------- | -------
| Babel            |  7.0.0-beta.51
| node             | v10.4.0
[...]

PR #9693, opened 2019-03-15
[plugin-transform-typescript] Fix transpiling of TS abstract classes with decorators 

agoldis:
| Fixed Issues?            | https://github.com/babel/babel/issues/8172
| Tests Added + Pass?      | Added, not passing 😞
| Documentation PR Link    | 

When `["@babel/plugin-proposal-decorators", { "legacy": true }]` and `@babel/preset-typescript` plugins are in use, using decorators with or within Typescript abstract classes causes generation of wrong syntax and preserves non-JS compliant `abstract` specification for class.

E.g. of code:

```typescript
export default abstract class SyncableEntity {
  @observable id: string;
}
```

becomes:
```javascript
let SyncableEntity = (_class = (_temp = abstract class SyncableEntity {
    _initializerDefineProperty(this, "id", _descriptor, this);
    // ...
}
```

This PR resolves it this issue by removing `abstract` property from `Class` visitor of `@babel/plugin-transform-typescript`.

I am having a trouble with tests, though - used in [test](https://github.com/babel/babel/pull/9693/files#diff-fac3cd7bf646c91b08f78aa272cd4a72R1) syntax is not accepted by parser:

```typescript
var foo = abstract class Foo {}
```

I get `BABEL_PARSE_ERROR`. Please help!

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -137,7 +137,6 @@ export default declare((api, { jsxPragma = "React" }) => {
           path.remove();
           return;
         }
-        if (node.abstract) node.abstract = null;
       },
 
       Class(path) {
@@ -146,6 +145,7 @@ export default declare((api, { jsxPragma = "React" }) => {
         if (node.typeParameters) node.typeParameters = null;
         if (node.superTypeParameters) node.superTypeParameters = null;
         if (node.implements) node.implements = null;
+        if (node.abstract) node.abstract = null;
 
         // Similar to the logic in `transform-flow-strip-types`, we need to
         // handle `TSParameterProperty` and `ClassProperty` here because the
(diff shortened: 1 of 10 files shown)


nicolo-ribaudo:
> I am having a trouble with tests, though - used in [test](https://github.com/babel/babel/pull/9693/files#diff-fac3cd7bf646c91b08f78aa272cd4a72R1) syntax is not accepted by parser:
>
> ```typescript
> var foo = abstract class Foo {}
> ```
>
> I get `BABEL_PARSE_ERROR`. Please help! 

That throws because it isn't valid typescript :stuck_out_tongue: You'll need to generate that code using a custom plugin in your test, similarly to what has been done in https://github.com/babel/babel/pull/9692

agoldis:
😀 so silly! Thanks  @nicolo-ribaudo!

agoldis:
@nicolo-ribaudo I've decided to not introduce a test with invalid TS syntax - instead I have added few tests with actual `["@babel/plugin-proposal-decorators", { "legacy": true }]` usage, as already appears at https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-typescript/test/fixtures/imports/parameter-decorators

PR #9697, opened 2019-03-17
Correctly preserve reference to array in for-of loop

danez:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #9696 Fixes #8913 Fixes #8920
| Patch: Bug Fix?          | y
| Major: Breaking Change?  | n
| Minor: New Feature?      | n
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

This change ensures that the reference to an array inside the header of an for-of loop cannot be unset.

PR #9698, opened 2019-03-18
Move array reference into `for` head initializer

danez:
| Fixed Issues?            | 
| Patch: Bug Fix?          |n
| Major: Breaking Change?  |n
| Minor: New Feature?      |n
| Documentation PR Link    | 
| Any Dependency Changes?  |

This moves the reference to the array of an for..of loop into the initializer of the transpiled for loop. Ass recommended in #9697

No functional change only moving the variable around.

### DIFF
--- a/packages/babel-plugin-transform-for-of/src/index.js
+++ b/packages/babel-plugin-transform-for-of/src/index.js
@@ -119,7 +119,7 @@ export default declare((api, options) => {
   function _ForOfStatementArray(path) {
     const { node, scope } = path;
 
-    const right = t.identifier(scope.generateUid("arr"));
+    const right = scope.generateUidIdentifierBasedOnNode(node.right, "arr");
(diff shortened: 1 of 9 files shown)


packages/babel-plugin-transform-for-of/src/index.js
```diff
@@ -119,20 +119,14 @@ export default declare((api, options) => {
...
 
-    const uid = scope.generateUid("arr");
-    const nodes = [
-      t.variableDeclaration("var", [
-        t.variableDeclarator(t.identifier(uid), node.right),
-      ]),
-    ];
-    const right = t.identifier(uid);
-
+    const right = t.identifier(scope.generateUid("arr"));
```
jridgewell: Nit:
```suggestion
    const right = scope.generateUidIdentifierBasedOnNode(node.right, "arr");
```

PR #9699, opened 2019-03-18
Correctly parse TS TypeAssertions around arrow functions

### LINKED ISSUE #9324: @babel/parser(ts): type assertion with arrow function
## Bug Report

**Current Behavior**
Babel parse `TypeAssertion` as `typeParameters` for arrow function.

this is checked with tsc parser

**Input Code**
```ts
var asserted1 = <any>((n) => { return n; });
```

ast produced by babel:

```json
{
  "id": {
    "name": "asserted1",
    "type": "Identifier"
  },
  "init": {
    "async": false,
    "body": {
      "elements": [
        {
          "name": "n",
          "type": "Identifier"
        }
      ],
      "type": "ArrayExpression"
    },
    "expression": true,
    "generator": false,
    "id": [null],
    "params": [
      {
        "name": "n",
        "type": "Identifier"
      }
    ],
    "type": "ArrowFunctionExpression",
    "typeParameters": {
      "params": [
        {
          "name": {
            "name": "any",
            "type": "Identifier"
          },
          "type": "TSTypeParameter"
        }
      ],
      "type": "TSTypeParameterDeclaration"
    }
  },
  "type": "VariableDeclarator"
}
```

**Expected behavior/code**

```json
{
  "id": {
    "name": "asserted1",
    "type": "Identifier"
  },
  "init": {
    "expression": {
      "async": false,
      "body": {
        "elements": [
          {
            "name": "n",
            "type": "Identifier"
          }
        ],
        "type": "ArrayExpression"
      },
      "expression": true,
      "generator": false,
      "id": [null],
      "params": [
        {
          "name": "n",
          "type": "Identifier"
        }
      ],
      "type": "ArrowFunctionExpression"
    },
    "type": "TSTypeAssertion",
    "typeAnnotation": {
      "type": "TSAnyKeyword"
    }
  },
  "type": "VariableDeclarator"
}
```

```js
    'typescript',
    'objectRestSpread',
    'decorators-legacy',
    'classProperties',
    'asyncGenerators',
    'dynamicImport',
    'estree',
    'bigInt',
    'importMeta'
  ];
  if (jsx) {
    plugins.push('jsx');
  }

  return babel.parse(text, {
    sourceType: 'unambiguous',
    ranges: true,
    plugins
  });
```

**Environment**
- Babel version(s):  7.2.3
- Node/npm version: 6-11
- OS: ----
- Monorepo: ---

danez:
| Fixed Issues?            | Fixes #9324
| Patch: Bug Fix?          | 👻
| Documentation PR Link    | 
| Any Dependency Changes?  |

Prior to this change it would only work if `createParenthesizedExpressions` was true, because we already check if the node is not an `ArrowFunctionExpression`. Now it also works with this setting set to false.

### DIFF
--- a/packages/babel-parser/test/fixtures/typescript/cast/arrow-in-parens-with-parens-node/options.json
+++ b/packages/babel-parser/test/fixtures/typescript/cast/arrow-in-parens-with-parens-node/options.json
@@ -1,0 +1,4 @@
+{
+  "createParenthesizedExpressions": true,
+  "plugins": ["typescript"]
+}
--- a/packages/babel-parser/src/plugins/typescript.js
+++ b/packages/babel-parser/src/plugins/typescript.js
@@ -2121,7 +2121,10 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         // This is similar to TypeScript's `tryParseParenthesizedArrowFunctionExpression`.
         typeParameters = this.tsParseTypeParameters();
         arrowExpression = super.parseMaybeAssign(...args);
-        if (arrowExpression.type !== "ArrowFunctionExpression") {
+        if (
+          arrowExpression.type !== "ArrowFunctionExpression" ||
+          (arrowExpression.extra && arrowExpression.extra.parenthesized)
+        ) {
           this.unexpected(); // Go to the catch block (needs a SyntaxError).
         }
       } catch (err) {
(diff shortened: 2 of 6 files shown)


packages/babel-parser/test/fixtures/typescript/cast/arrow-in-parens-with-parens-node/options.json
```diff
@@ -0,0 +1,4 @@
+{
+  "createParenthesizedExpressions": true,
```
armano2: should we add this option to `babel-plugin-syntax-typescript`?

danez: No this is only for tools using babel-parser directly.
Babel would not understand the AST that gets returned when this option is active.

PR #9704, opened 2019-03-19
Fix `super` Method Calls in Class Private Methods

### LINKED ISSUE #9580: super method calls do not work in private instance methods
## Bug Report

**Current Behavior**
Calling a `super` method or accessing a `super` property from within a private instance method or accessor does not work.

**Input Code**
[REPL link](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEDKCuAHApgJwMLitA3gKGkOgiTQFlkAXACwHsATACgEpcCiPhaA7CWkZADoQtAOaMA5CRSoJzANztCAXzyq8oSDAQAjTFujIAHpWTd620hiwx8HAMSJUASwBuYUxRoMWbDh2k0QUDULzomBSVodQ5EeB0QZ2Awn1Y7f0IaZwhBRxd3TypwlkUOVXVuZAB3OHj9KBZBOISklIjFIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=false&targets=&version=7.3.3)

I couldn't get the the repl link to include additional plugins, so you have to manually add these:
* `@babel/plugin-proposal-class-properties`
* `@babel/plugin-proposal-private-methods`

```js
class SuperClass {
    superMethod() {
        console.log('super');
    }
}

class SubClass extends SuperClass {
    #privateMethod() {
        super.superMethod();
    }

    publicMethod() {
        this.#privateMethod();
    }
}

new SubClass().publicMethod();
```

**Expected behavior/code**
The console logs a message:
```
super
```

**Actual behavior/code**
```js
function _classPrivateMethodGet(receiver, privateSet, fn) { if (!privateSet.has(receiver)) { throw new TypeError("attempted to get private field on non-instance"); } return fn; }

class SuperClass {
  superMethod() {
    console.log('super');
  }

}

class SubClass extends SuperClass {
  constructor(...args) {
    super(...args);

    _privateMethod.add(this);
  }

  publicMethod() {
    _classPrivateMethodGet(this, _privateMethod, _privateMethod2).call(this);
  }

}

var _privateMethod = new WeakSet();

var _privateMethod2 = function _privateMethod2() {
  super.superMethod();
};

new SubClass().publicMethod();
```

The console logs an error in `_privateMethod2`:
```
'super' keyword unexpected here
```

```js
{
  "plugins": [
    "@babel/plugin-proposal-private-methods"
  ]
}
```

**Environment**
- Babel version(s): 7.3.3
- Node/npm version: 10.13.0
- OS: Windows 10
- Monorepo: no

**Possible Solution**
Babel turns private methods into regular functions, however these functions cannot access `super`. If Babel wants to support this use case, I think it would need to "capture" `super` somehow, similar to how `@babel/plugin-transform-async-to-generator` creates helpers to call `super` methods ([example](https://babeljs.io/repl/#?babili=false&browsers=chrome%2050&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEDKCuAHApgJwMLitA3gKGkOgiTQFlkAXACwHsATACgEpcCiPhaA7CWkZADoQtAOaMA5CRSoJzANztCAXzyq8oSDAQAjTFujIAHpWTd620hiwx8HSAE9uwaADNatFmw6EwAdzAAS0piK0Fpcio6JgUlaFV1bmQ_OHg9GxZBRDSQQOAKGgYWRSA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.3.3)).

I couldn't find any `test262` tests that check the behavior of `super.method()` or `super.property` within a private method or accessor. The only tests I found were about [calling the super constructor within a private method](https://github.com/tc39/test262/blob/efa414ae008e1b2b57cd2e3e7a1f4234f7d9f17c/test/language/statements/class/elements/syntax/early-errors/grammar-special-meth-contains-super-private-method.js). Perhaps this is just ill-defined, and we need to file a spec bug first? 😕

tim-mc:
| Fixed Issues?            | Fixes #9580 
| Documentation PR Link    | 

This fixes an issue with the use of super method calls in class private methods. See https://github.com/babel/babel/issues/9580 for more info re: behavior of the bug.

### DIFF
--- a/packages/babel-helper-replace-supers/src/index.js
+++ b/packages/babel-helper-replace-supers/src/index.js
@@ -112,7 +112,12 @@ const specHandlers = {
 
   set(superMember, value) {
     return t.callExpression(this.file.addHelper("set"), [
-      getPrototypeOfExpression(this.getObjectRef(), this.isStatic, this.file),
+      getPrototypeOfExpression(
+        this.getObjectRef(),
+        this.isStatic,
+        this.file,
+        this.isPrivateMethod,
+      ),
       this.prop(superMember),
       value,
       t.thisExpression(),
--- a/packages/babel-helper-replace-supers/src/index.js
+++ b/packages/babel-helper-replace-supers/src/index.js
@@ -15,11 +15,12 @@ import * as t from "@babel/types";
  *
  *   helpers.getPrototypeOf(CLASS.prototype)
  */
-function getPrototypeOfExpression(objectRef, isStatic, file) {
+function getPrototypeOfExpression(objectRef, isStatic, file, isPrivateMethod) {
   objectRef = t.cloneNode(objectRef);
-  const targetRef = isStatic
-    ? objectRef
-    : t.memberExpression(objectRef, t.identifier("prototype"));
+  const targetRef =
+    isStatic || isPrivateMethod
+      ? objectRef
+      : t.memberExpression(objectRef, t.identifier("prototype"));
 
   return t.callExpression(file.addHelper("getPrototypeOf"), [targetRef]);
 }
@@ -98,7 +99,12 @@ const specHandlers = {
 
   get(superMember) {
     return t.callExpression(this.file.addHelper("get"), [
-      getPrototypeOfExpression(this.getObjectRef(), this.isStatic, this.file),
+      getPrototypeOfExpression(
+        this.getObjectRef(),
+        this.isStatic,
+        this.file,
+        this.isPrivateMethod,
+      ),
       this.prop(superMember),
       t.thisExpression(),
     ]);
@@ -170,8 +176,8 @@ export default class ReplaceSupers {
     const path = opts.methodPath;
 
     this.methodPath = path;
-    this.isStatic =
-      path.isClassMethod({ static: true }) || path.isObjectMethod();
+    this.isStatic = path.isObjectMethod() || path.node.static;
+    this.isPrivateMethod = path.isPrivate() && path.isMethod();
 
     this.file = opts.file;
     this.superRef = opts.superRef;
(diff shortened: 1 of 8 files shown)


tim-mc:
@nicolo-ribaudo

packages/babel-helper-replace-supers/src/index.js
```diff
@@ -170,8 +181,8 @@ export default class ReplaceSupers {
     const path = opts.methodPath;
 
     this.methodPath = path;
-    this.isStatic =
-      path.isClassMethod({ static: true }) || path.isObjectMethod();
+    this.isStatic = path.isObjectMethod() || path.node.static;
+    this.isPrivateMethod = path.isPrivate() && path.isMethod();
```
jridgewell: Nit: I would just add this as a new case to `isStatic`.

tim-mc: I kept them separate because rolling `isPrivateMethod` into `isStatic` will cause an issue with the loose handler. If the path is an _instance_ private method, [this](https://github.com/babel/babel/blob/673385038571b0d13b6bd9efafe6206b61f0c3eb/packages/babel-helper-replace-supers/src/index.js#L151) condition becomes truthy, and [this output](https://github.com/babel/babel/blob/673385038571b0d13b6bd9efafe6206b61f0c3eb/packages/babel-plugin-proposal-private-methods/test/fixtures/private-method-loose/super/output.js#L24-L26) is instead transformed to:

```js
var _privateMethod2 = function _privateMethod2() {
  return Base.superMethod.call(this);
};
```

PR #9714, opened 2019-03-20
Fix rest parameters indexing with TypeScript 'this parameter'

### LINKED ISSUE #8840: [plugin-transform-typescript@7.1.0] Combination of 'this' parameter and rest parameters goes wrong
## Bug Report

**Current Behavior**

I have a function that both has a `this` parameter, and a spread / rest parameter pattern, such as:

```
function logEmit(this: Socket, event: string, ...args: any[]) {
  console.log('given arguments:', args)
}
```

When I call this function, the value of `args` is missing the first given argument.

**Input code**

```
logEmit.call(socket, 'message', 1, 2, 3, 4, 5)
```

Yields: `given arguments: [2, 3, 4, 5]`

**Expected behavior/code**

I expect the above call to yield `given arguments: [1, 2, 3, 4, 5]`

I use React Native's `module:metro-react-native-babel-preset`, which uses `@babel/plugin-transform-typescript@7.1.0` under the hood.

**Environment**
- Babel version(s): v6.26.3
- Node/npm version: v10.8.0
- OS: OSX 10.13.6
- How you are using Babel: not sure, through React Native's metro bundler

**Possible Solution**

Inspecting the generated code of my `logEmit` function learns why this happens: the (virtual) `this` parameter is taken into account in the argument extraction algorithm.

```
function logEmit(event) {
    for (var _len = arguments.length, args = new Array(_len > 2 ? _len - 2 : 0), _key = 2; _key < _len; _key++) {
      args[_key - 2] = arguments[_key];
    }

    //...
}
```

The solution would be to ignore the `this` parameter, so that the number 2 would become 1 instead.

BenoitZugmeyer:
| Fixed Issues?            | Fixes #8840

The TypeScript ["this parameter"][0] is a fake parameter and should not be taken into account when counting the function parameters.  This patch skips it by using the condition taken from the `transform-typescript` plugin.

Note: since the `transform-typescript` plugin is removing this kind of parameter, including it before the `transform-parameters` plugin solves the issue.  This patch fixes the issue in other cases.

[0]: https://www.typescriptlang.org/docs/handbook/functions.html#this-parameters


packages/babel-plugin-transform-parameters/src/rest.js
```diff
@@ -295,12 +306,13 @@ export default function convertFunctionRest(path) {
     state.candidates.map(({ path }) => path),
   );
 
-  const start = t.numericLiteral(node.params.length);
+  const paramsCount = getParamsCount(node);
```
kesne: It might be better to hoist this const to above the `const state`, so that you can just re-use the `paramsCount` as the offset in state.

BenoitZugmeyer: Okay, PR edited!

BenoitZugmeyer:
Is there anything else I can do to make this PR right?

alamothe:
Any updates?

canpoyrazoglu:
I'm affected by this bug too, and there's this PR that fixes the issue. Any updates on when this would be merged?

packages/babel-plugin-transform-parameters/src/rest.js
```diff
@@ -153,6 +153,17 @@ const memberExpressionOptimisationVisitor = {
     }
   },
 };
+
+function getParamsCount(node) {
+  let count = node.params.length;
+  // skip the first parameter if it is a TypeScript 'this parameter'
```
JLHwung: Can you guard the following with `if (count)` ?, accessing `0` of empty array will result in expensive prototype chain lookup. https://v8.dev/blog/elements-kinds#avoid-reading-beyond-the-length-of-the-array

packages/babel-plugin-transform-parameters/src/rest.js
```diff
@@ -153,6 +153,17 @@ const memberExpressionOptimisationVisitor = {
...
+  const p0 = node.params[0];
+  if (p0 && t.isIdentifier(p0) && p0.name === "this") {
```
JLHwung: If `count` is guarded, we don't have to check `p0` because each node of `params` must not be nullish.

Also you can simplify the `isIdentifier` check to
```js
t.isIdentifier(p0, { name: "this" })
```

BenoitZugmeyer:
Thanks for the review! I processed your comments, and merged master to make sure the updated tests are still passing.

packages/babel-plugin-transform-parameters/test/fixtures/parameters/options.json
```diff
@@ -2,6 +2,7 @@
   "plugins": [
     "proposal-class-properties",
     "external-helpers",
+    "syntax-typescript",
     "syntax-flow",
```
nicolo-ribaudo: I'm surprised that this works, `flow` and `typescript` shouldn't be compatible.

packages/babel-plugin-transform-parameters/src/rest.js
```diff
@@ -155,6 +155,16 @@ const memberExpressionOptimisationVisitor = {
     }
   },
 };
+
+function getParamsCount(node) {
+  let count = node.params.length;
+  // skip the first parameter if it is a TypeScript 'this parameter'
+  if (count > 0 && t.isIdentifier(node.params[0], { name: "this" })) {
```
jridgewell: Should we be checking that there's a `typeAnnotation` attached to this?

nicolo-ribaudo: I don't think that it is needed: `this` is not a valid binding identifier, so it can't be anything else other than TypeScript.

PR #9716, opened 2019-03-20
Tweak preset-env corejs/useBuiltIns warning and error messages

existentialism:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-preset-env/src/normalize-options.js
+++ b/packages/babel-preset-env/src/normalize-options.js
@@ -204,7 +204,11 @@ export function normalizeCoreJSOption(
   if (useBuiltIns && corejs === undefined) {
     rawVersion = 2;
     console.log(
-      "\nWith `useBuiltIns` option, required direct setting of `corejs` option\n",
+      "\nWe noticed you're using the `useBuiltIns` option without declaring a " +
+        "core-js version. Currently, we assume version 2.x when no version " +
+        "is passed. Since this default version will likely change in future " +
+        "versions of Babel, we recommend explicitly setting the core-js version " +
+        "you are using via the `corejs` option.\n",
     );
   } else if (typeof corejs === "object" && corejs !== null) {
     rawVersion = corejs.version;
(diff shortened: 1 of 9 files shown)


packages/babel-preset-env/src/normalize-options.js
```diff
@@ -204,7 +204,11 @@ export function normalizeCoreJSOption(
   if (useBuiltIns && corejs === undefined) {
     rawVersion = 2;
     console.log(
-      "\nWith `useBuiltIns` option, required direct setting of `corejs` option\n",
+      "\nWe noticed you're using the `useBuiltIns` option without declaring a " +
+        "core-js version. Currently, we assume version 2.x when no version " +
+        "is passed. Since this default version will likely change in future " +
+        "versions of Babel, we recommend explicitly setting the core-js version " +
+        "you are using via the `corejs` option.\n",
```
nicolo-ribaudo: We should also say they users should be sure that they have the correct version of core-js in their dependencies.
For example, `angular-cli` was broken by the core-js@3 release because they were relying on `core-js@2` being in `node_modules` without specifying it in `dependencies`.

nicolo-ribaudo:
```suggestion
        "you are using via the `corejs` option.\n" +
        "You should also be sure that the version you pass to the `corejs` " +
        "option matches the version specified in your `package.json`'s " +
        "`dependencies` section. If it doesn't, you need to run one of the " +
        "following commands:\n\n" + 
        "  npm install --save core-js@2    npm install --save core-js@3\n" +
        "  yarn add core-js@2              yarn add core-js@3\n",
```

existentialism: 👍

htulipe:
The corsjs option is undocumented: https://babeljs.io/docs/en/babel-preset-env

nicolo-ribaudo:
Doc PR at https://github.com/babel/website/pull/1987

PR #9735, opened 2019-03-21
Set NPM_TOKEN in .npmrc when publishing

nicolo-ribaudo:

As you can see in https://github.com/babel/babel/runs/82101132, the publish failed because of an auth error:
```
lerna ERR! EWHOAMI Authentication error. Use `npm whoami` to troubleshoot
```

I tought that a `NPM_TOKEN` env var was enought, but apparentely it isn't: https://github.com/npm/npm/issues/8356.

I don't think that this PR could cause a security problem, because the .npmrc file is generated right before running lerna and deleted right after.

### DIFF
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -17,7 +17,7 @@ action "Publish to npm" {
   uses = "docker://node:10"
   secrets = ["NPM_TOKEN"]
   runs = "make"
-  args = "publish-c"
+  args = "publish-ci"
   env = {
     CI = "true"
   }
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -17,7 +17,7 @@ action "Publish to npm" {
   uses = "docker://node:10"
   secrets = ["NPM_TOKEN"]
   runs = "make"
-  args = "publish"
+  args = "publish-c"
   env = {
     CI = "true"
   }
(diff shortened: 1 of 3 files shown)


nicolo-ribaudo:
This (from the log above) might be a problem:
```
lerna http fetch GET 401 https://registry.npmjs.org/-/npm/v1/user 277ms
```

I'll need to check.

Makefile
```diff
@@ -137,9 +138,23 @@ new-version:
...
+ifeq ("$(CI)", "true")
+ifneq ("$(NPM_TOKEN)", "")
+	echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > .npmrc
+else
+	echo "Missing NPM_TOKEN env var"
+	exit 1
+endif
+else
+	echo "Skipping .npmrc generation..."
+endif
```
xtuc: I'm pretty sure this works on Linux, but we should try on Mac and Windows first I think.

nicolo-ribaudo: We are already using it in Makefile

.github/main.workflow
```diff
@@ -17,7 +17,7 @@ action "Publish to npm" {
   uses = "docker://node:10"
   secrets = ["NPM_TOKEN"]
   runs = "make"
-  args = "publish"
+  args = "publish-c"
```
itscaro: should it be `"publish-ci"`?

nicolo-ribaudo: Yes 🤦‍♂️

nicolo-ribaudo:
```suggestion
  args = "publish-ci"
```

PR #9742, opened 2019-03-21
docs(preset-env): remove changelog file

eddiemonge:
The changelog file is wildly out of date and not relevant. Changes are 
published to github releases and not to the changelog file


nicolo-ribaudo:
I think that we should keep it, but moved with the other "legacy" changelogs: https://github.com/babel/babel/tree/master/.github

PR #9747, opened 2019-03-22
Upgrade browserslist and caniuse in preset-env

TheLudd:
| Any Dependency Changes?         | y

There is a corrupt version of caniuse-lite that will cause installs with
yarn to fail. Upgrading these dependencies makes sure that the version
of caniuse-lite will not be the corrupt version.


existentialism:
@TheLudd thanks!

PR #9752, opened 2019-03-23
Normalize `core-js` entry points

zloirock:
Related [this comment](https://github.com/zloirock/core-js/issues/496#issuecomment-475681471). Sorry, one more time from the main repo.

### DIFF
--- a/packages/babel-preset-env/src/polyfills/corejs3/entry-plugin.js
+++ b/packages/babel-preset-env/src/polyfills/corejs3/entry-plugin.js
@@ -22,7 +22,10 @@ function isBabelPolyfillSource(source) {
 
 function isCoreJSSource(source) {
   if (typeof source === "string") {
-    source = source.replace(/([\\/](index)?)?(\.js)?$/i, "").toLowerCase();
+    source = source
+      .replace(/\\/g, "/")
+      .replace(/(\/(index)?)?(\.js)?$/i, "")
+      .toLowerCase();
   }
   return has(corejsEntries, source) && corejsEntries[source];
 }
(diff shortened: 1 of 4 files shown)


packages/babel-preset-env/src/polyfills/corejs3/entry-plugin.js
```diff
@@ -21,6 +21,9 @@ function isBabelPolyfillSource(source) {
 }
 
 function isCoreJSSource(source) {
+  if (typeof source === "string") {
+    source = source.replace(/(\/(index)?)?(\.js)?$/i, "").toLowerCase();
```
nicolo-ribaudo: I don't think we should use `i` and `toLoweCase`, otherwise imports that work when transpiled stop working if they are left as-is.

zloirock: What do you mean?
image

nicolo-ribaudo: Not on every OS:

Schermata da 2019-03-23 11-28-01

nicolo-ribaudo: Mh maybe it is ok to normalize paths; it won't cause any bug.

zloirock: I don't care, it's not something principal for me.

nicolo-ribaudo:
```suggestion
    source = source.replace(/([\\/](index)?)?(\.js)?$/i, "").toLowerCase();
```

We need to also handle windows paths.

zloirock: Makes sense 👍

PR #9756, opened 2019-03-23
Allow coreJS Symbol to be type object

### LINKED ISSUE #9755: TypeError: Object is not async iterable
I'm having a problem with

@babel/runtime-corejs2/helpers/asyncIterator.js (@7.4.2)

I does not recognize that certain items are iterable because it contains more or less the following:
```js
const Symbol = require("../core-js/symbol");

if (typeof Symbol === "function") {
  // ...
}
```

This is unfortunately incorrect. Symbol as exported by core-js 2 is an object not a function.

It seems like the code should simply allow object or function as the typeof symbol. In this case I am running in a browser with native Symbol support, and `require("../core-js/symbol").iterator === global.Symbol.iterator`, which is the necessary condition for this code to work properly.

conartist6:
| Fixed Issues?            | Fixes #9755
| Documentation PR Link    | None

See the #9755 for a detailed description of the problem.

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -80,7 +80,7 @@ helpers.jsx = helper("7.0.0-beta.0")`
 helpers.asyncIterator = helper("7.0.0-beta.0")`
   export default function _asyncIterator(iterable) {
     var method
-    if (typeof Symbol === "function") {
+    if (typeof Symbol !== "undefined") {
       if (Symbol.asyncIterator) {
         method = iterable[Symbol.asyncIterator]
         if (method != null) return method.call(iterable);


packages/babel-helpers/src/helpers.js
```diff
@@ -78,9 +78,10 @@ helpers.jsx = helper("7.0.0-beta.0")`
 `;
 
 helpers.asyncIterator = helper("7.0.0-beta.0")`
+  const symbolTypes = ["function", "object"];
   export default function _asyncIterator(iterable) {
     var method
-    if (typeof Symbol === "function") {
+    if (symbolTypes.includes(typeof Symbol)) {
```
nicolo-ribaudo: Maybe this could just be `typeof Symbol !== "undefined"`?

conartist6: Sounds good to me

nicolo-ribaudo:
@zloirock Is it expected that `core-js` creates an object-type `Symbol` when polyfilling only wks?

zloirock:
@nicolo-ribaudo yes. However, I'm not sure how it's related to `runtime` since here used `Symbol` directly.

conartist6:
This code appears to use Symbol directly, but it is transpiled prior to shipping such that it ends up importing Symbol from core-js.

nicolo-ribaudo:
Someone could also manually import the polyfill for `Symbol.asyncIterable` without importing `Symbol()` polyfill.

Issue #8526, opened 2018-08-24
Running script with @babel/node triggers SIGINT event twice


GProst:
## Bug Report

**Current Behavior**
As the title says.

**Input Code**
Run this code with babel-node
```
process.on('SIGINT', () => {
  console.log('SIGINT')
  setTimeout(() => {
    process.exit()
  }, 5000)
})
setTimeout(() => {}, 10000000)
```
Outputs two 'SIGINT' in stdout.

**Expected behavior/code**
Should be called once.

**Babel Configuration (.babelrc, package.json, cli command)**
You can test it without `.babelrc`

**Environment**
- Babel version: v7.0.0-rc.2
- Node/npm version: 8.11.2
- OS: OSX 10.13.6

dfranusic:
In my opinion `SIGINT` (**CTRL + C**) should not call `proc.kill("SIGINT")` (kill child process) since child process will be the first one to receive **CTRL + C** (being in the foreground) followed by the parent `babel-node` process. The way `babel-node` works now, `SIGINT` is received by the child process twice. The only potential benefit (a bit useless in my humble opinion) right now is when `SIGINT` is generated by invoking `kill -2 pid_of_babel_process`, the child process will then receive only one `SIGNINT`, the one caused by `proc.kill("SIGINT")`. 

This is the current code excerpt from `babel-node.js`:
````
106:  process.on("SIGINT", () => {
107:      proc.kill("SIGINT");
108:      process.exit(1);
109: });
````

And It should be changed to this:
````
106:  process.on("SIGINT", () => {
107:      //proc.kill("SIGINT");
108:      process.exit(1);
109: });
````

This is just my humble opinion and if everybody agrees, I would be quite happy to create a new pull request.

PR #9758, opened 2019-03-24
Remove process.exit(1) from babel-node

dword-design:
Because it breaked graceful shutdown of the sub process. For example, if you have an express server, catch the SIGINT signal and process.exit(0) after closing the server, the command would always return exit code 1. This PR allows the subprocess to decide how to exit itself.

### DIFF
--- a/packages/babel-node/src/babel-node.js
+++ b/packages/babel-node/src/babel-node.js
@@ -103,9 +103,6 @@ getV8Flags(function(err, v8Flags) {
         }
       });
     });
-    process.on("SIGINT", () => {
-      proc.kill("SIGINT");
-      process.exit(1);
-    });
+    process.on("SIGINT", () => proc.kill("SIGINT"));
   }
 });


mgreenw:
I don't think this fixes the other issue with this method from #8526 that the `SIGINT` signal gets sent to the child process twice. Should line 106 be removed entirely such that `babel-node` (the parent process) does not even handle `SIGINT` at all?

nicolo-ribaudo:
> Should line 106 be removed entirely such that babel-node (the parent process) does not even handle SIGINT at all?

Probably yes, but I'm not an expert on this topic.
I _think_ that currently we are doing this:

1. Spawn babel-node
  1.i Spawn user's program
  1.ii Ctrl+C
  1.iii SIGINT
2. SIGINT (it bubbles to the parent process?)
  2.i Forward SIGINT to the child process
3. `process.exit(1)`

mgreenw:
I'm not exactly sure what would happen if the `SIGINT` for some reason never got sent to the child process. Because the parent is dead, wouldn't the child get reaped at some point?

I think the initial SIGINT from the Ctrl-C should be plenty to alert the child that it should die, and the parent re-sending the signal (2.i in @nicolo-ribaudo's chart) does not make sense and should be removed.

mgreenw:
@nicolo-ribaudo Thanks for doing that testing! It looks like this kills two birds with one stone.

PR #9766, opened 2019-03-25
Add TS support to @babel/parser's Scope

### LINKED ISSUE #9480: error: bundling failed: SyntaxError: { path }\PinCode\index.ts: Exporting local ”IPinCodeParams“, which is not declared
## Bug Report

**Current Behavior**
I created a new project. I have special folders for interfaces, types and etc. I will have a lot of them, so I import them into one file and export them from there. When I want to export multiple interfaces I get an error:
```js
error: bundling failed: SyntaxError: C:\Users\QWERTY\Desktop\TestProjects\2019\TestProject\node_modules\react-native\scripts\src\types\PinCode\index.ts: Exporting local "IPinCodeParams", which is not declared.
  13 | }
  14 |
> 15 | export { PinStatus, IPinCodeParams }
     |                     ^
    at File.buildCodeFrameError (C:\Users\QWERTY\Desktop\TestProjects\2019\TestProject\node_modules\@babel\core\lib\transformation\file\file.js:261:12)
    at NodePath.buildCodeFrameError (C:\Users\QWERTY\Desktop\TestProjects\2019\TestProject\node_modules\@babel\traverse\lib\path\index.js:157:21)
    at getLocalMetadata (C:\Users\QWERTY\Desktop\TestProjects\2019\TestProject\node_modules\@babel\helper-module-transforms\lib\normalize-and-load-metadata.js:281:22)
    at child.get.forEach.spec (C:\Users\QWERTY\Desktop\TestProjects\2019\TestProject\node_modules\@babel\helper-module-transforms\lib\normalize-and-load-metadata.js:315:11)
    at programPath.get.forEach.child (C:\Users\QWERTY\Desktop\TestProjects\2019\TestProject\node_modules\@babel\helper-module-transforms\lib\normalize-and-load-metadata.js:307:33)
    at getLocalExportMetadata (C:\Users\QWERTY\Desktop\TestProjects\2019\TestProject\node_modules\@babel\helper-module-transforms\lib\normalize-and-load-metadata.js:294:27)
    at getModuleMetadata (C:\Users\QWERTY\Desktop\TestProjects\2019\TestProject\node_modules\@babel\helper-module-transforms\lib\normalize-and-load-metadata.js:86:21)
```

**Input Code / Babel Configuration (.babelrc, package.json, cli command)**
**package.json**
```js
{
  "name": "TestProject",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "android": "node node_modules/react-native/local-cli/cli.js run-android",
    "ios": "node node_modules/react-native/local-cli/cli.js run-ios",
    "lint": "yarn tslint && yarn typecheck",
    "tslint": "tslint --project tsconfig.json --config tslint.json src/**/*.ts{,x} --format stylish",
    "typecheck": "tsc --project tsconfig.json --noEmit",
    "postinstall": "typesync",
    "type-sync": "typesync"
  },
  "husky": {
    "hooks": {
      "pre-commit": "yarn lint",
      "pre-push": "yarn lint"
    }
  },
  "dependencies": {
    "mobx": "^4.9.2",
    "mobx-persist": "^0.4.1",
    "mobx-react": "^5.4.3",
    "react": "16.8.0",
    "react-native": "0.58.4",
    "react-native-vector-icons": "^6.2.0",
    "react-navigation": "^2.18.1"
  },
  "devDependencies": {
    "@babel/plugin-proposal-class-properties": "^7.3.0",
    "@babel/plugin-proposal-decorators": "^7.3.0",
    "@types/babel-core": "6.25.5",
    "@types/jest": "24.0.0",
    "@types/prettier": "^1.16.0",
    "@types/react": "16.8.2",
    "@types/react-native": "0.57.34",
    "@types/react-native-vector-icons": "^4.6.4",
    "@types/react-navigation": "^3.0.1",
    "@types/react-test-renderer": "16.8.0",
    "babel-core": "7.0.0-bridge.0",
    "babel-jest": "24.1.0",
    "babel-plugin-module-resolver": "^3.1.3",
    "husky": "^1.3.1",
    "jest": "24.1.0",
    "jsc-android": "236355.1.1",
    "metro-react-native-babel-preset": "0.51.1",
    "mobx-logger": "^0.7.1",
    "prettier": "^1.16.4",
    "react-native-typescript-transformer": "^1.2.11",
    "react-test-renderer": "16.6.3",
    "tslint": "^5.12.1",
    "tslint-config-prettier": "^1.18.0",
    "tslint-eslint-rules": "^5.4.0",
    "tslint-react": "^3.6.0",
    "typescript": "^3.2.4",
    "typesync": "^0.4.1"
  },
  "jest": {
    "preset": "react-native"
  }
}
```

**.babelrc**
```js
{
  "presets": [
    "module:metro-react-native-babel-preset",
  ],
  "plugins": [
    [
      {
        "legacy": true
      }
    ],
    "@babel/transform-react-jsx-source",
    [
      "module-resolver",
      {
        "root": [
          "/src"
        ],
        "alias": {
          "@api": "./src/api",
          "@actions": "./src/actions",
          "@components": "./src/components",
          "@constants": "./src/constants",
          "@hocs": "./src/hocs",
          "@managers": "./src/managers",
          "@locales": "./src/locales",
          "@reducers": "./src/reducers",
          "@screens": "./src/screens",
          "@static": "./src/static",
          "@stores": "./src/stores",
          "@styles": "./src/styles",
          "@types": "./src/types",
          "@utils": "./src/utils"
        }
      }
    ],
  ]
}
```

**tsconfing.json**
```js
{
  "compilerOptions": {
    "target": "es2015",
    "module": "commonjs",
    "moduleResolution": "node",
    "jsx": "react-native",
    "baseUrl": "./src",
    "outDir": "./dist",
    "allowSyntheticDefaultImports": true,
    "declaration": true,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "noEmit": true,
    "noImplicitAny": false,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "pretty": true,
    "skipLibCheck": true,
    "sourceMap": true,
    "strict": true,
    "paths": {
      "@api": [
        "api"
      ],
      "@actions": [
        "actions"
      ],
      "@components": [
        "components"
      ],
      "@constants": [
        "constants"
      ],
      "@hocs": [
        "hocs"
      ],
      "@managers": [
        "managers"
      ],
      "@locales": [
        "locales"
      ],
      "@reducers": [
        "reducers"
      ],
      "@screens": [
        "screens"
      ],
      "@static": [
        "static"
      ],
      "@stores": [
        "stores"
      ],
      "@styles": [
        "styles"
      ],
      "@types": [
        "types"
      ],
      "@utils": [
        "utils"
      ]
    }
  },
}
```

**tslist.json**
```js
{
  "defaultSeverity": "error",
  "extends": [
    "tslint-config-prettier",
    "tslint-eslint-rules",
    "tslint-react",
    "tslint:latest"
  ],
  "jsRules": {},
  "rules": {
    "await-promise": false,
    "interface-over-type-literal": false,
    "jsx-alignment": true,
    "jsx-boolean-value": false,
    "jsx-no-bind": true,
    "jsx-no-lambda": false,
    "max-classes-per-file": false,
    "member-access": false,
    "member-ordering": false,
    "no-constant-condition": false,
    "no-duplicate-imports": false,
    "no-empty": false,
    "no-implicit-dependencies": false,
    "no-submodule-imports": false,
    "object-literal-sort-keys": false,
    "ordered-imports": false,
    "quotemark": [
      true,
      "single",
      "jsx-single",
      "avoid-escape"
    ],
    "semicolon": [
      true,
      "always"
    ],
    "trailing-comma": [
      true,
      {
        "singleline": "never",
        "multiline": "always"
      }
    ],
    "triple-equals": [
      true,
      "allow-undefined-check"
    ]
  },
  "linterOptions": {
    "exclude": [
      "./src/typings/**/*.d.ts"
    ]
  },
  "rulesDirectory": []
}
```

[Reference to the source project (Github)][1]

  [1]: https://github.com/SergeyMyssak/TestProject

**Expected behavior/code**
I have this folder structure:
```
android\
ios\
node_modules\
src\
--->constants\
       |--->colors.ts
       |--->fonts.ts
       |--->index.ts
--->router\
       |--->index.ts
--->screen\
       |--->Splash\
               |--->containers\
                      |--->index.ts
                      |--->SplashContainer.tsx
               |--->views\
                      |--->index.ts
                      |--->SplashView.tsx
               |--->index.tsx
       |--->index.ts
--->types\
       |--->PinCode\
               |--->index.tsx
       |--->index.tsx
--->App.tsx
.babelrc
*
*
*
package.json
rn-cli.config.js
tsconfig.json
tslint.json
```

In the file **"PinCode"** I have an enum and an interface. 

When I write so (all right):
```js
export enum PinStatus {
  New = 'new',
  Confirm = 'confirm',
  Unlock = 'unlock',
  Edit = 'edit',
}

export interface IPinCodeParams {
  status: PinStatus;
  onComplete: () => void;
  onClose?: () => void;
  originalCode?: string;
}
```

But when I write so (I get the error):
```js
enum PinStatus {
  New = 'new',
  Confirm = 'confirm',
  Unlock = 'unlock',
  Edit = 'edit',
}

interface IPinCodeParams {
  status: PinStatus;
  onComplete: () => void;
  onClose?: () => void;
  originalCode?: string;
}

export { PinStatus, IPinCodeParams };
```

I have created some completely new projects, tried to rewrite this project several times, but nothing came out. I finally got this error anyway. Why do I get this error? How to fix it?

**Environment**
- Babel version(s): I don't know, but I use "babel-core": "7.0.0-bridge.0"
- Node/npm/yarn version: v10.4.1 / v6.5.0 / v1.9.4
- OS: Windows 10

### LINKED ISSUE #9763: New @babel/parser early errors break by-name typescript exports
## Bug Report

**Current Behavior**
`@babel/parser` doesn't consider identifiers declared with typescript syntax as identifiers in scope.

**Input Code**

```ts
enum Test {}

export { Test as default }
```

**Expected behavior/code**
`export`s of typescript types should not be a build error.

This case is particularly strange as `Test` is _actually_ a concrete value; it's just declared with a typescript declaration.

This is the expected output, for reference:

```js
var Test;

(function (Test) {})(Test || (Test = {}));

export { Test as default }
```

Direct `export enum Test {}` is unaffected.

Way too big, but just adding `@babel/plugin-syntax-typescript` to the REPL is sufficient to reproduce it. `@babel/plugin-transform-typescript` doesn't help as the error happens before any transforms can run.

**Environment**
- Babel version(s): 7.4.2
- Node/npm version: 11.10.0
- OS: macOS 18.2.0
- Monorepo: yes
- How you are using Babel: `babel-jest` / `babel-loader`

**Possible Solution**
Identifiers declared with `type`, `enum`, `interface` (and `namespace`) need to be considered as valid for `export`ing when the `typescript` plugin is enabled.

This same problem affects `flow` but the other way around. The following code is "valid syntax" in both flow and typescript; but babel can only parse the flow version, which is incorrect at runtime as flow requires explicitly annotating the export as a type export:

```ts
type Foo = {}
export { Foo }
```

nicolo-ribaudo:
| Fixed Issues?            | Fixes #9763, fixes #9480
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-parser/src/plugins/typescript/scope.js
```diff
@@ -0,0 +1,12 @@
+import type ScopeHandler, { Scope } from "../../util/scope";
+
+interface TypeScriptScope extends Scope {}
+
+const mixinScope = (superClass: Class<Scope>): Class<TypeScriptScope> =>
+  class TypeScriptScope extends superClass {};
```
danez: Do you think we will need that? I don't think we will ever have the need of multiple inheritances. as flow and ts will never be used together

I would just do
```
class TypeScriptScope extends Scope {};
```

packages/babel-parser/src/util/scope.js
```diff
@@ -33,19 +33,29 @@ class Scope {
...
 
+  // We need to copy this from the static property because new.target
+  // isn't available in the "enter" method.
+  Scope: Class<Scope>;
+
   constructor(raise: raiseFunction, inModule: boolean) {
     this.raise = raise;
     this.inModule = inModule;
+    // $FlowIgnore flow doesn't support new.target
+    this.Scope = new.target.Scope;
```
danez: What if we instead of all this magic, create a method?

```
createScope(flags) {
 return new Scope(flags);
}
```

this can easily be overwritten.

packages/babel-parser/src/util/scope.js
```diff
@@ -89,55 +93,79 @@ export default class ScopeHandler {
...
-        scope.var.indexOf(name) > -1;
-      scope.lexical.push(name);
-      if (this.inModule && scope.flags & SCOPE_PROGRAM) {
-        this.undefinedExports.delete(name);
-      }
-    } else if (bindingType === BIND_SIMPLE_CATCH) {
-      const scope = this.currentScope();
-      scope.lexical.push(name);
-    } else if (bindingType === BIND_FUNCTION) {
+    if (
```
nicolo-ribaudo: The changes in this file are mostly refactoring things in different functions, to make it possible to overwrite them.

packages/babel-parser/src/util/scopeflags.js
```diff
@@ -41,12 +41,14 @@ export const BIND_NONE = 0, // Not a binding
   BIND_LEXICAL = 2, // Let- or const-style binding
   BIND_FUNCTION = 3, // Function declaration
   BIND_SIMPLE_CATCH = 4, // Simple (identifier pattern) catch binding
-  BIND_OUTSIDE = 5; // Special case for function names as bound inside the function
+  BIND_OUTSIDE = 5, // Special case for function names as bound inside the function
+  BIND_TS_ENUM = 6; // TypeScript enums (block-scoped, but can be re-declared using var/function)
```
Jessidhia: A complete fix for https://github.com/babel/babel/issues/9763 would also need to keep track of other typescript declarations.

Also, this is not quite how TS Enums work. `enum` can be declared multiple times, but only with `enum` or `namespace`. It might be useful to keep track of which types are in "type space" or "value space".

For example, `type A = {}; function A() {}` are completely different and unrelated -- they are two separate bindings (and both are hoisted!); they just share the same name.

Here's an attempt to explain how the declarations (are supposed to) work. The "type" part of declarations is always hoisted.

\ | value | type
--- | --- | ---
`namespace` | maybe | yes, can hold other declarations
`enum` | yes, `let`-like | yes, `namespace`-like
`class` | yes, `let`-like | yes (refers to the instance)
`interface` | no | yes
`type` | no | yes
`var` | yes | no
`let` | yes | no
`const` | yes | no
`function` | yes | no

Generic arguments aren't here but they are `type`-like.

`namespace` and `enum` (besides an ES module itself) are the only kinds of type that can hold other types or declarations inside them. `class` "looks like" it would behave like that but they are `interface`-like instead of `namespace`-like -- but `class` _can_ merge with a `namespace`, which affects the value portion `class` (i.e. they're part of the constructor, not the instance). `class` can also merge with `interface`, which affects the type portion of `class` (the instance type).

✅ = merges
🔴 = invalid
⚪️ = no merging _and_ no conflict; they are separate entities
⚠️ = namespaces are weird (_as long as_ they do not contain values they don't conflict; if they do they can't merge)
❗️ = no conflict if and only if the type annotations are identical

Merging | `namespace` | `enum` | `class` | `interface` | `type` | `var` | `let` | `const` | `function`
--- | :-: | :-: | :-: | :-: | :-: | :-: | :-: | :-: | :-:
`namespace` | ✅ | ✅ | ✅ | ✅ | 🔴 | ⚠ | ⚠ | ⚠️ | ✅  
`enum` | ✅ | ✅ | 🔴 | 🔴 | 🔴 | 🔴  | 🔴 | 🔴 | 🔴  
`class` | ✅ | 🔴 | 🔴 | ✅ | 🔴 | 🔴 | 🔴 | 🔴 | 🔴 | 🔴 
`interface` | ✅ | 🔴 | ✅ | ✅ | 🔴 | ⚪️ | ⚪️ | ⚪️ | ⚪️ 
`type` | 🔴 | 🔴 | 🔴 | 🔴 | 🔴 | ⚪️ | ⚪️ | ⚪️ | ⚪️ 
`var` | ⚠ | 🔴 | 🔴 | ⚪️ | ⚪️ | ❗️ | 🔴 | 🔴 | 🔴 | 🔴  
`let` | ⚠ | 🔴 | 🔴 | ⚪️ | ⚪️ | 🔴 | 🔴 | 🔴 | 🔴 | 🔴 
`const` | ⚠ | 🔴 | 🔴 | ⚪️ | ⚪️ | 🔴 | 🔴 | 🔴 | 🔴 | 🔴 
`function` | ✅ | 🔴 | 🔴 | ⚪️ | ⚪️ | 🔴 | 🔴 | 🔴 | `*`

`*`: multiple declarations of a function / method on the same scope are only valid if the declarations are overloads; only one of the declarations is allowed to have an implementation, and it must be the last one.

`type` can maybe be understood as the type-only `const` equivalent, as it is block-scoped and it can't merge with anything or be reassigned.

nicolo-ribaudo: Thanks you for the detailed table!
I think that we for the (`var`...`functon`)x(`var`...`functon`) cells, we should follow JavaScript's behavior, since we can't rely on type info in Babel.

As for :warning:, I think that I won't implement redeclaration checks for them in this PR since they are only needed when transpiling namespaces to IIFEs, but Babel doesn't support doing so. It could be added later but it will be more difficult.

nicolo-ribaudo: > The "type" part of declarations is always hoisted.

Always to the block, not to the function, right?

airato:
today it seems we have 3 issues with named exports in typescript:
1) parser fails with undefined exports when we try to export TS types within named exports. this happens because TS type declarations are not counted as a declaration when checking exports (#9763) 
2) the scope for bodyless TS function declarations is getting opened, but not getting closed. This causes issues with function overloads
3) the typescript transform does not erase named type exports. 

this PR should fix 1) and maybe 2). Issue 3 is hidden today since parser fails before transform.

I have a branch attempting to fix these, but I'm doing so with existing scope module, and it's pretty dirty. 
Since this PR is doing a refactor of the scope module for TS, I'll wait for this to land and then will create a PR to fix issue 3.

Here is one of the tests from my branch
https://github.com/airato/babel/blob/2b6fe025ace1038b5d2e77420f50692a0b1bd22d/packages/babel-plugin-transform-typescript/test/fixtures/declarations/erased/input.mjs

The parser should be able to parse this with no errors

danez:
This testcases now fail although the shouldn't, and may be because the SIMPLE_CATCH was removed?:

```js
try {} catch (a) { 
  if(1) function a(){} 
}
```

https://tc39.github.io/ecma262/#sec-variablestatements-in-catch-blocks

I will add testcases to this PR.

nicolo-ribaudo:
@danez It is already failing on master: https://babeljs.io/repl/build/master#?babili=false&browsers=IE%207&build=&builtIns=false&spec=false&loose=false&code_lz=C4JwngBA3gvhDGBDY8AWEAUiCU0ICgIIBLAMwwEZdSBXAO3mGIHs6JENtYCYg&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=script&lineWrap=false&presets=es2015%2Creact%2Cstage-2&prettier=true&targets=&version=7.4.3&externalPlugins=

I'd prefer to fix it in a separate PR since it seems unrelated.

danez:
Sure, but I just wonder why the test works. https://github.com/babel/babel/blob/master/packages/babel-parser/test/fixtures/core/scope/dupl-bind-catch-hang-func/input.js

But in the repl it doesn't.

nicolo-ribaudo:
Because that tests doesn-t run `@babel/traverse`, only `@babel/parser`.

danez:
🤦‍♂️I thought the error comes from the parser.

nicolo-ribaudo:
@danez I will add a few more commits to address the link in https://github.com/babel/babel/pull/9766#issuecomment-479239380

nicolo-ribaudo:
@airato Your test is now correctly parsed.

henderea:
Hey, is this PR going anywhere?  It doesn't seem to have much activity in the past couple of weeks.  What all is it waiting on?

packages/babel-parser/src/plugins/typescript/scope.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  isRedeclaredInScope(
+    scope: TypeScriptScope,
+    name: string,
+    bindingType: BindingTypes,
+  ): boolean {
+    if (scope.enums.indexOf(name) > -1) {
+      // Enums can be merged with other enums
+      return !(bindingType & BIND_FLAGS_TS_ENUM);
```
Jessidhia: Enums can merge with namespaces too; I guess the error is avoided by not calling `super.declareName` for namespaces?

nicolo-ribaudo: Yeah exactly.

PR #9768, opened 2019-03-25
Correctly check for-in and for-of loop for invalid left-hand side

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

We were allowing assignments in for-in loops

```js
var x; for(x = 3 in list);
```
which is not allowed. The only exception is `for(var x = 1 in list);` in non-strict mode.

I refactored this checks and moved them into `parseForIn()`. This also improves the error messages.

Tests were already present

also did the same in acorn: https://github.com/acornjs/acorn/pull/821


packages/babel-parser/test/helpers/runFixtureTests.js
```diff
@@ -131,6 +131,12 @@ function runTest(test, parseFunction) {
       if (err.message === opts.throws) {
         return;
       } else {
+        /*
+        const fn = path.dirname(test.expect.loc) + "/options.json";
```
nicolo-ribaudo: We might want to enable this code when the `OVERWRITE=true` env var is set.

PR #9769, opened 2019-03-26
Don't accept '\08' or '\09' in strict mode

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-parser/src/tokenizer/index.js
```diff
@@ -1213,7 +1213,13 @@ export default class Tokenizer extends LocationParser {
             octalStr = octalStr.slice(0, -1);
             octal = parseInt(octalStr, 8);
           }
-          if (octal > 0) {
+          this.state.pos += octalStr.length - 1;
+          const next = this.input.charCodeAt(this.state.pos);
+          if (
+            octalStr !== "0" ||
```
nicolo-ribaudo: This is the same as the previous `octal > 0`, right? Or is it different in case of `\00`?

danez: Oh right, this is doing exactly the same thing.

PR #9794, opened 2019-03-30
fix(rest-spread): Do not require `Symbol.iterator` for strings

### LINKED ISSUE #9277: Spread transformation requires Symbol.iterator now
# v7 Regression

https://github.com/babel/babel/commit/4da3f3bc64841e0b4b7906c6c26af73920264095

**Describe the regression**

Transformation of `Spread` now requires `Symbol`

**Input Code**

```js
const divElements = [...document.getElementsByTagName('div')];
```

```js
// from webpack:

module: {
  rules: [{
    test: /\.js$/,
    exclude: /(node_modules|bower_components)/,
    use: {
      loader: 'babel-loader',
      options: {
        presets: [
          ['@babel/preset-env'],
        ],
      },
    },
  }
}
```

**Expected behavior/code**

Before this commit, all you needed was a polyfill for `Array.from()`. Now `Symbol.iterator` is needed as well. This breaks IE11.

**Environment**
- Babel version(s): @babel/preset-env: 7.2.3,
- Node/npm version: 8.11.3 / 6.5.0
- OS: OSX 10.14.2
- Monorepo: ??
- How you are using Babel: webpack

If, for some reason we still need `Symbol.iterator`, perhaps we perform the `isIterable` check *AFTER* the prototype check for `[object Arguments]`. That way, requiring `Symbol.iterator` would be a last-ditch effort.

clshortfuse:
| Fixed Issues?            | Fixes #9277
| Minor: New Feature?      |  No
| Tests Added + Pass?      | Pass (with no tests added) 
| Documentation PR Link    | 

This applies the same logic as other runtime helpers by checking if `typeof Symbol === 'function'` before using Symbol.iterator as seen here:

https://github.com/babel/babel/blob/3aaafae053fa75febb3aa45d45b6f00646e30ba4/packages/babel-helpers/src/helpers.js#L178

The CoreJS2 version of this example would be compiled as:

```
if (typeof _Symbol === "function" && _Symbol$asyncIterator) {
  AsyncGenerator.prototype[_Symbol$asyncIterator] = function () {
    return this;
  };
}
```

You can also see this at https://github.com/babel/babel/blob/3aaafae053fa75febb3aa45d45b6f00646e30ba4/packages/babel-helpers/src/helpers.js#L215

So I can confirm the syntax is correct. I've tested on IE11 and spread now works properly (after adding a polyfill for Array.from).

Tested with: 

```
function sum(x, y, z) {
  return x + y + z;
}

const numbers = [1, 2, 3];

console.log(sum(...numbers));
// expected output: 6
```

The only actual code change in `packages/babel-helpers/src/helpers.js` and everything else is auto generated by `make bootstrap`.

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -933,24 +933,27 @@ helpers.arrayWithHoles = helper("7.0.0-beta.0")`
 
 helpers.iterableToArray = helper("7.0.0-beta.0")`
   export default function _iterableToArray(iter) {
+    // ES6 prototypes that use @@iterator
     const prototypeWhiteList = [
-      '[object Arguments]',
+      String.prototype,
+      Array.prototype,
+      // Map.prototype,
+      // Set.prototype,
     ];
-    let hasSymbol = false;
-    try {
-      if (Symbol.iterator in Object(iter)) {
-        hasSymbol = true;
-      }
-    } catch (e) {}
-    if (hasSymbol) {
-      // Return outside catch block to allow any errors to be thrown by Array.from 
+    const iterObject = Object(iter);
+    if (prototypeWhiteList.some(prototype => prototype.isPrototypeOf(iterObject))) {
       return Array.from(iter);
     }
-    // Fallback to whitelist
-    if (prototypeWhiteList.indexOf(Object.prototype.toString.call(iter)) !== -1) {
+
+    // IsArguments
+    if (iterObject.toString() === "[object Arguments]") {
       return Array.from(iter);
     }
-    return null;
+
+    // Check for custom iterator outside ES6 spec, iterable WeakMap, or iterable TypedArray
+    if (Symbol.iterator in iterObject) {
+      return Array.from(iter);
+    }
   }
 `;
 
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -933,25 +933,19 @@ helpers.arrayWithHoles = helper("7.0.0-beta.0")`
 
 helpers.iterableToArray = helper("7.0.0-beta.0")`
   export default function _iterableToArray(iter) {
-    // ES6 prototypes that use @@iterator
-    const prototypeWhiteList = [
-      String.prototype,
-      Array.prototype,
-      // Map.prototype,
-      // Set.prototype,
-    ];
-    const iterObject = Object(iter);
-    if (prototypeWhiteList.some(prototype => prototype.isPrototypeOf(iterObject))) {
+    // String.prototype
+    if (typeof iter === 'string') {
       return Array.from(iter);
     }
 
-    // IsArguments
-    if (iterObject.toString() === "[object Arguments]") {
+    // Function Arguments
+    if (Object.prototype.toString.call(iter) === "[object Arguments]") {
       return Array.from(iter);
     }
 
-    // Check for custom iterator outside ES6 spec, iterable WeakMap, or iterable TypedArray
-    if (Symbol.iterator in iterObject) {
+    // Check for iterable WeakMap, Map, Set, or TypedArray
+    // Check for custom iterator outside ES6 spec
+    if (Symbol.iterator in Object(iter)) {
       return Array.from(iter);
     }
   }
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -933,10 +933,24 @@ helpers.arrayWithHoles = helper("7.0.0-beta.0")`
 
 helpers.iterableToArray = helper("7.0.0-beta.0")`
   export default function _iterableToArray(iter) {
-    if (
-      Symbol.iterator in Object(iter) ||
-      Object.prototype.toString.call(iter) === "[object Arguments]"
-    ) return Array.from(iter);
+    const prototypeWhiteList = [
+      '[object Arguments]',
+    ];
+    let hasSymbol = false;
+    try {
+      if (Symbol.iterator in Object(iter)) {
+        hasSymbol = true;
+      }
+    } catch (e) {}
+    if (hasSymbol) {
+      // Return outside catch block to allow any errors to be thrown by Array.from 
+      return Array.from(iter);
+    }
+    // Fallback to whitelist
+    if (prototypeWhiteList.indexOf(Object.prototype.toString.call(iter)) !== -1) {
+      return Array.from(iter);
+    }
+    return null;
   }
 `;
 
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -933,21 +933,11 @@ helpers.arrayWithHoles = helper("7.0.0-beta.0")`
 
 helpers.iterableToArray = helper("7.0.0-beta.0")`
   export default function _iterableToArray(iter) {
-    // String.prototype
-    if (typeof iter === 'string') {
-      return Array.from(iter);
-    }
-
-    // Function Arguments
-    if (Object.prototype.toString.call(iter) === "[object Arguments]") {
-      return Array.from(iter);
-    }
-
-    // Check for iterable WeakMap, Map, Set, or TypedArray
-    // Check for custom iterator outside ES6 spec
-    if (Symbol.iterator in Object(iter)) {
-      return Array.from(iter);
-    }
+    if (
+      typeof iter === 'string'
+      || Object.prototype.toString.call(iter) === "[object Arguments]"
+      || Symbol.iterator in Object(iter)
+    ) return Array.from(iter);
   }
 `;
 
--- a/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/parameters-exec/exec.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/parameters-exec/exec.js
@@ -1,0 +1,9 @@
+function sum(x, y, z) {
+  return x + y + z;
+}
+function test() {
+  var args = arguments;
+  return sum(...args);
+}
+
+expect(test(1, 2, 3)).toBe(6);
--- a/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/parameters-exec/options.json
+++ b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/parameters-exec/options.json
@@ -1,0 +1,3 @@
+{
+  "plugins": ["proposal-object-rest-spread"]
+}
--- a/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-spread/string-exec/exec.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-spread/string-exec/exec.js
@@ -1,0 +1,3 @@
+expect([...'']).toHaveLength(0);
+expect([...'abc']).toHaveLength(3);
+expect([...'def']).toMatchObject(['d','e','f']);
--- a/packages/babel-runtime-corejs2/helpers/esm/iterableToArray.js
+++ b/packages/babel-runtime-corejs2/helpers/esm/iterableToArray.js
@@ -1,20 +1,15 @@
 import _isIterable from "../../core-js/is-iterable";
 import _Array$from from "../../core-js/array/from";
 export default function _iterableToArray(iter) {
-  var prototypeWhiteList = [String.prototype, Array.prototype];
-  var iterObject = Object(iter);
-
-  if (prototypeWhiteList.some(function (prototype) {
-    return prototype.isPrototypeOf(iterObject);
-  })) {
+  if (typeof iter === 'string') {
     return _Array$from(iter);
   }
 
-  if (iterObject.toString() === "[object Arguments]") {
+  if (Object.prototype.toString.call(iter) === "[object Arguments]") {
     return _Array$from(iter);
   }
 
-  if (_isIterable(iterObject)) {
+  if (_isIterable(Object(iter))) {
     return _Array$from(iter);
   }
 }
--- a/packages/babel-runtime-corejs2/helpers/esm/iterableToArray.js
+++ b/packages/babel-runtime-corejs2/helpers/esm/iterableToArray.js
@@ -1,5 +1,22 @@
 import _Array$from from "../../core-js/array/from";
 import _isIterable from "../../core-js/is-iterable";
 export default function _iterableToArray(iter) {
-  if (_isIterable(Object(iter)) || Object.prototype.toString.call(iter) === "[object Arguments]") return _Array$from(iter);
+  var prototypeWhiteList = ['[object Arguments]'];
+  var hasSymbol = false;
+
+  try {
+    if (_isIterable(Object(iter))) {
+      hasSymbol = true;
+    }
+  } catch (e) {}
+
+  if (hasSymbol) {
+    return _Array$from(iter);
+  }
+
+  if (prototypeWhiteList.indexOf(Object.prototype.toString.call(iter)) !== -1) {
+    return _Array$from(iter);
+  }
+
+  return null;
 }
--- a/packages/babel-runtime-corejs2/helpers/esm/iterableToArray.js
+++ b/packages/babel-runtime-corejs2/helpers/esm/iterableToArray.js
@@ -1,15 +1,5 @@
-import _isIterable from "../../core-js/is-iterable";
 import _Array$from from "../../core-js/array/from";
+import _isIterable from "../../core-js/is-iterable";
 export default function _iterableToArray(iter) {
-  if (typeof iter === 'string') {
-    return _Array$from(iter);
-  }
-
-  if (Object.prototype.toString.call(iter) === "[object Arguments]") {
-    return _Array$from(iter);
-  }
-
-  if (_isIterable(Object(iter))) {
-    return _Array$from(iter);
-  }
+  if (typeof iter === 'string' || Object.prototype.toString.call(iter) === "[object Arguments]" || _isIterable(Object(iter))) return _Array$from(iter);
 }
--- a/packages/babel-runtime-corejs2/helpers/esm/iterableToArray.js
+++ b/packages/babel-runtime-corejs2/helpers/esm/iterableToArray.js
@@ -1,22 +1,20 @@
-import _Array$from from "../../core-js/array/from";
 import _isIterable from "../../core-js/is-iterable";
+import _Array$from from "../../core-js/array/from";
 export default function _iterableToArray(iter) {
-  var prototypeWhiteList = ['[object Arguments]'];
-  var hasSymbol = false;
+  var prototypeWhiteList = [String.prototype, Array.prototype];
+  var iterObject = Object(iter);
 
-  try {
-    if (_isIterable(Object(iter))) {
-      hasSymbol = true;
-    }
-  } catch (e) {}
-
-  if (hasSymbol) {
+  if (prototypeWhiteList.some(function (prototype) {
+    return prototype.isPrototypeOf(iterObject);
+  })) {
     return _Array$from(iter);
   }
 
-  if (prototypeWhiteList.indexOf(Object.prototype.toString.call(iter)) !== -1) {
+  if (iterObject.toString() === "[object Arguments]") {
     return _Array$from(iter);
   }
 
-  return null;
+  if (_isIterable(iterObject)) {
+    return _Array$from(iter);
+  }
 }
--- a/packages/babel-runtime-corejs2/helpers/iterableToArray.js
+++ b/packages/babel-runtime-corejs2/helpers/iterableToArray.js
@@ -1,19 +1,9 @@
-var _isIterable = require("../core-js/is-iterable");
-
 var _Array$from = require("../core-js/array/from");
 
+var _isIterable = require("../core-js/is-iterable");
+
 function _iterableToArray(iter) {
-  if (typeof iter === 'string') {
-    return _Array$from(iter);
-  }
-
-  if (Object.prototype.toString.call(iter) === "[object Arguments]") {
-    return _Array$from(iter);
-  }
-
-  if (_isIterable(Object(iter))) {
-    return _Array$from(iter);
-  }
+  if (typeof iter === 'string' || Object.prototype.toString.call(iter) === "[object Arguments]" || _isIterable(Object(iter))) return _Array$from(iter);
 }
 
 module.exports = _iterableToArray;
--- a/packages/babel-runtime-corejs2/helpers/iterableToArray.js
+++ b/packages/babel-runtime-corejs2/helpers/iterableToArray.js
@@ -1,26 +1,24 @@
-var _Array$from = require("../core-js/array/from");
-
 var _isIterable = require("../core-js/is-iterable");
 
+var _Array$from = require("../core-js/array/from");
+
 function _iterableToArray(iter) {
-  var prototypeWhiteList = ['[object Arguments]'];
-  var hasSymbol = false;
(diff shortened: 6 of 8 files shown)


zloirock:
It will cause the same problems with `Symbol` injection in `runtime` and `preset-env` like the previous PR, it should be fixed. However, this approach looks better for me since it does not add special cases and injection of `Map`, `Set`, etc.

clshortfuse:
@zloirock Do you have reference as to how it would produce an issue? Do you mean the conversion with corejs2? I did find other pieces of helper use `typeof Symbol !== 'undefined` which wouldn't care if `Symbol()` is available or not and then check `Symbol.iterator`. That would let `Symbol` be an object.

I did have another idea to solving this and that's ditching `Symbol` altogether. The reality is the code is checking to see if the Object has `Symbol.iterator`... and then never actually uses it. Unlike the other pieces of code that actually use `Symbol.asyncIterator` and `Symbol.iterator`, this one doesn't. If the code was going to call ` iter[Symbol.iterator]()` then that's different. 

In the end, it's just going to call `Array.from` and not use Symbol anyway. I think it makes for sense to just target what `Array.from` would actually do. Yes, all Objects that can be iterated with Symbol are arrays, but not that's exactly the best way to check for `Array.from` compatibility. 

I was only targeting the regression specifically, but the alternative of `Object.prototype.toString.call(iter) === "[object Arguments]"` isn't that great either. For example, some ArrayLikes such as HTMLCollections can be used with `Array.from` but its prototype would be `[object HTMLCollection]`. Trying to target every single variation in some sort of whitelist is just asking for issues down the line (like `Map` or `Set`).

According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Polyfill the only thing that `Array.from` really needs is to check the following is `Object(iter) != null`. To be a bit more robust, we can check if `iter.length > 0 || iter.length === 0`. (It'll make sure it's a number between 0 and  and not null or NaN). I'm pretty sure `Array.length` can't be negative, but that can easily checked as well.

zloirock:
@clshortfuse [this line](https://github.com/babel/babel/pull/9794/files#diff-ce878a969c5550bd6906bcf9cc15bcefR5) will add tens of kilobytes to the bundle and it's a serious regression.

It could be fixed on `runtime` and `preset-env` sides. But `typeof` checks should not be just ignored - they should return or be replaced to related boolean.

> In the end, it's just going to call `Array.from` and not use `Symbol` anyway.

I propose use this approach in the `loose` mode - it keeps all problems to a minimum.

clshortfuse:
I'm willing to dive in a bit more into the issue. I know sometimes solutions to problems tend to get shelved for lack of priority. I think finding the mechanics through which an Object can use Symbol.iterator can be a good goal. 

Plainly stated, all the current code checks is whether the object is a candidate for Symbol iterator. We don't have to replicate the entirely Symbol functionality, but, instead, only it's check it's candidacy. (ie: if it can get an iterator, then it can receive an Array.from call).

Also, I think it's important to note the that the only purpose for the entire check is to simply cast (of sorts) an Object to an Array. As far as I've seen, there's no reference to this function outside of ToArray, so maybe taking a step back and looking at the original purpose may prove useful.

As for the mention of bloat, I'm assuming the issue lies with the importing of _Symbol from an external library. This is my first time playing around with a PR for Babel, so I wasn't aware of what consequences some functions would have. Still, I'll try to see what alternatives I can come up with. I'll probably need to dive deep into the ECMA spec and review a handful of popular Symbol polyfills just to ensure we don't cause any conflicts, and ultimately, hopefully it can be used outside of loose mode.

clshortfuse:
I took another stab at this. I took another approach based on how `iterableToArrayLimit` attempts to use `Symbol.iterator`

https://github.com/babel/babel/blob/3aaafae053fa75febb3aa45d45b6f00646e30ba4/packages/babel-helpers/src/helpers.js#L891 

Basically, it's wrapped in a try block. If it fails, then it discards the error and tries the prototype. If that doesn't success, `toArray` is going to call `_nonIterableRest()` which throws a TypeError.

https://github.com/babel/babel/blob/3aaafae053fa75febb3aa45d45b6f00646e30ba4/packages/babel-helpers/src/helpers.js#L924

Doing it this way instead of how v6 was doing it (checking prototype first) avoids the performance penalty of the string comparison in contexts that support Symbol.iterator. I broke out the prototype whitelist into an array, so we can change it in the future to:

````
  const prototypeWhiteList = [
      '[object Arguments]',
      '[object HTMLCollection]',
      '[object NamedNodeMap]',
      '[object Map]',
      '[object Set]',
    ];
````

Out of curiosity, and not exactly related to the issue, but if we call `(Array.prototype.from || [].slice).call(iter)` we can then remove any requirements for IE11. It would only serve as a fallback. I'm sure there's some specific reason why we can't do this now (probably core-js), but it would be nice to have.

**Comment by zloirock** [comment]
@ljharb could you add support of `@@toStringTag` behavior for `Map` and `Set` to `es6-shim` (even without exposing `Symbol.toStringTag`)? It could seriously simplify of fixing this issue.

File: packages/babel-helpers/src/helpers.js [hunk 1]
```diff
@@ -860,10 +860,24 @@ helpers.arrayWithHoles = helper("7.0.0-beta.0")`
...
   export default function _iterableToArray(iter) {
-    if (
-      Symbol.iterator in Object(iter) ||
-      Object.prototype.toString.call(iter) === "[object Arguments]"
-    ) return Array.from(iter);
+    const prototypeWhiteList = [
+      '[object Arguments]',
+    ];
+    let hasSymbol = false;
+    try {
```
ljharb: Why is a try/catch needed here instead of just checking if Symbol.iterator exists or not?

clshortfuse: You'd have to check if `Symbol` exists before checking if `Symbol.iterator` exists. I squashed the previous commit, but checking if Symbol exists would bloat the code up, as explained the one of the previous comments.

Doing it this way removes the need for `typeof` checks and lets the browser pass the Symbol check failure to `catch`. Since we're not actually using `Symbol.iterator` the result would be same. If it can't check the object to see if it has it, then it wouldn't have it anyway.

ljharb: try/catch has a performance penalty that's far worse than a few extra gzippable bytes.

clshortfuse: As I mentioned in the other comment, this is the third try/catch in the same helper file. I don't see why this specific one should be treated any differently.

ljharb: because in that case, it's being used to catch errors thrown by user code (the iterator method). In this case it's used to catch errors in your code itself (attempting unchecked member access on a maybe-not-object).

clshortfuse: Ah, in the `next()` function. Okay, back to the drawing board.

clshortfuse:
According to [ES6 spec](https://www.ecma-international.org/ecma-262/6.0/), the following prototypes use `@@iterator`

* String
* Array
* Map
* Set

`WeakMap` isn't guaranteed to have an iterator. `TypedArray` is listed as well, but since it looks like a native abstract type, I couldn't find a way to target all of them.  I removed `Map` and `Set` since it would add more imports in `corejs2`. It doesn't seems possible to really check if an object uses `Symbol.iterator` without actually checking for `Symbol` and `iterator`. I did have another method of checking for Symbol, namely: 

````
const env = ((typeof window !== 'undefined') && window)
  || ((typeof global !== 'undefined') && global)
  || ((typeof self !== 'undefined') && self);
if (!env || 'Symbol' in env) {
  if (Symbol.iterator in iterObject) {
    return Array.from(iter);
  }
}
````

Since I'm only targeting fixing the V6 regression with this PR, I resorted to leaving it out.

File: packages/babel-helpers/src/helpers.js [hunk 2]
```diff
@@ -860,10 +860,27 @@ helpers.arrayWithHoles = helper("7.0.0-beta.0")`
 
 helpers.iterableToArray = helper("7.0.0-beta.0")`
   export default function _iterableToArray(iter) {
-    if (
-      Symbol.iterator in Object(iter) ||
-      Object.prototype.toString.call(iter) === "[object Arguments]"
-    ) return Array.from(iter);
+    // ES6 prototypes that use @@iterator
+    const prototypeWhiteList = [
+      String.prototype,
```
jridgewell: `typeof iter === "string"`

packages/babel-helpers/src/helpers.js
```diff
@@ -860,10 +860,27 @@ helpers.arrayWithHoles = helper("7.0.0-beta.0")`
...
+      Array.prototype,
```
jridgewell: `Array.isArray`

packages/babel-helpers/src/helpers.js
```diff
@@ -860,10 +860,27 @@ helpers.arrayWithHoles = helper("7.0.0-beta.0")`
...
+      // Map.prototype,
+      // Set.prototype,
+    ];
+    const iterObject = Object(iter);
+    if (prototypeWhiteList.some(prototype => prototype.isPrototypeOf(iterObject))) {
+      return Array.from(iter);
+    }
+
+    // IsArguments
+    if (iterObject.toString() === "[object Arguments]") {
```
jridgewell: `Object.prototype.toString.call(iterObject)`

packages/babel-helpers/src/helpers.js
```diff
@@ -860,10 +860,27 @@ helpers.arrayWithHoles = helper("7.0.0-beta.0")`
...
+      return Array.from(iter);
+    }
+
+    // Check for custom iterator outside ES6 spec, iterable WeakMap, or iterable TypedArray
+    if (Symbol.iterator in iterObject) {
```
jridgewell: Wasn't the point of this PR to do a `typeof Symbol !== 'undefined'` check?

clshortfuse: No. It's to fix the spread regression specifically.

jridgewell: But none of your changes will help spreading the `HTMLCollection` from the [issue](https://github.com/babel/babel/issues/9277): `[...document.getElementsByTagName('div')]`. I don't understand what you're looking for then.

packages/babel-helpers/src/helpers.js
clshortfuse: `Array.isArray` isn't the same a prototype check. For example, `Array.isArray` would return `true` when checked against `Array.prototype` where this wouldn't. Also, `Array.isArray` is already checked prior to this function.

packages/babel-helpers/src/helpers.js
clshortfuse: The input code shows the error, but doesn't accurately detail the regession. It's in the description, which is more accurate. Basically, all `spread` functions fail on IE11 because it's checking for Symbol before checking the prototype. It's a regression from v6. I believe it's caused by https://github.com/babel/babel/commit/4da3f3bc64841e0b4b7906c6c26af73920264095

jridgewell: The code in the OP goes through `arrayWithHoles`'s `Array.isArray` check, so this change does nothing still. I need an explicit "this fails in IE 11, doesn't fail with this change" case so that I can review the PR.

clshortfuse: I think you're missing the point of the `[object Arguments]` check.  That block never executes on environments without `Symbol`. Therefore any function with spread arguments would fail. 

````
function func1(a, b, c) {
  console.log(arguments[0]);
  // expected output: 1

  console.log(arguments[1]);
  // expected output: 2

  console.log(arguments[2]);
  // expected output: 3
}

func1(1, 2, 3);
````

Comment (clshortfuse) [reply packages/babel-helpers/src/helpers.js]: Looking back at the issue from 3 months ago, the request was a partial fix for the regression while the more complicated issue of completely eliminating Symbol is explored. That PR has been stuck for a couple of months now.

I think that's why I had the mindset of swapping the order of execution, at least to minimize the issue.

Comment (jridgewell) [reply packages/babel-helpers/src/helpers.js]: So, this is the kind of thing you're trying to fix? https://babeljs.io/repl/build/master#?babili=false&browsers=ie%20%3E%2010&build=&builtIns=false&spec=false&loose=false&code_lz=CYUwNiAuIAQO4EsB2wD2cB0BlAngWwCNUwBuAKADMBXJAY0gVSRgGcq8AKADwBoYc-ALwCUMAN5kYMAE5Qq05lxgBqfipiDyAX0o16jZtBaQOoiVIBuAQ2kwbAcxYwAvHen32IJJBbkpsyHlmNk4MMIcWYW0yMlomFmIQDDBUew4jEwBGPgAmPgBmYSigA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.4.3&externalPlugins=

A executable test case we can add is:

```js
function sum(x, y, z) {
  return x + y + z;
}
function test() {
  var args = arguments;
  return sum(...args);
}

assert.equal(test(1, 2, 3), 6);
```

All that's necessary for this was to change the order of the `||` expression in the old code.

clshortfuse: Yes, and wrap them in parenthesis. Also, any object that is a prototype of `Array` and `String` would align with the ES6 spec. 

I'm not seeing much more we can do that wouldn't trigger extra corejs2 imports. I have no way of checking for the internal `TypedArray` type. Also, `NodeList`, `HTMLCollection` and others aren't part of the ES6 spec. I realized that this function, at least as it's called, it's checking for array-likes. It's checking for iterables. Therefore, something like this:

````
[...({ 0: 'a', 1: 'b', 2: 'c', length: 3})]
````

would throw an error on an ES6 supporting browser, and therefore should throw an error on Babel. Support for array-likes would have to be `loose`. I think this the best we can do without violating spec.

Comment (jridgewell) [reply packages/babel-helpers/src/helpers.js]: We can add a string check, that's fine. The array check is already covered by `arrayWithHoles` (the difference between `Array.prototype` passing `Array.isArray` and not `Array.prototype.isPrototypeOf(Array.prototype)` doesn't matter). Let's do the string check, then arguments check, then isIterable check.

**Comment by clshortfuse** [comment]
Okay, I've made the changes to this code. I also changed the commit and topic to better reflect the limited scope of this fix. I'm only targeting the rest parameter regression here.

From my research, I've hit a bit of a wall getting a catch-all solution. I've come to agree with @zloirock that the `loose` option is probably the only possible way to make it all work, but it definitely wouldn't be to ES6 spec. For example, on a real ES6 browser, this:

````
[...({ 0: 'a', 1: 'b', 2: 'c', length: 3})]
````
would throw an non-iterable error. That means babel should as well. It would be bad if babel would run something fine when transpiled to ES5, and then if somebody makes a ES6 version without babel, realizes that it doesn't work. That's because `spread` only works for iterables, *NOT* array-likes.  Even though `Array.from` would accept it, it would be wrong to do so. Therefore any checks based on `.length` wouldn't be to spec.

There's more that can be done to target `Map` and `Set` which are part of the ES6 spec, but in my opinion, that would require changes to how `typeof` is transpiled. They should be checked without corejs2 importing the classes. Perhaps a special `$isUndefined()` could help in this case. Regardless, I think the best way to attack this problem is piece by piece and this is just a partial fix. We can explore that in another PR (I also have a couple of other theories as to checking if an object is using `Array.prototype.values()` as @@iterator).

Lastly, I want to apologize to the devs for the possibly annoying back and forth with the changes due to not fundamentally understanding the true underlying nature of `@@iterator` in ES6. 

Note to @jridgewell : The reason why I didn't put the String check outside `iterableToArray` is because I wasn't sure if I should put it with all functions that call it. I felt it was safer to keep it inside.

**Comment by clshortfuse** [comment]
If I add
````
// Do not add import if only checking typeof
if (t.isUnaryExpression(parent, { operator: "typeof" })) return;
````
to https://github.com/babel/babel/blob/3aaafae053fa75febb3aa45d45b6f00646e30ba4/packages/babel-plugin-transform-runtime/src/index.js#L264

Then I can add the following without forcing imports:
````
const iterObject = Object(iter);
// Map.prototype
if (typeof Map !== 'undefined' && (Map.prototype.isPrototypeOf(iterObject)) {
  return Array.from(iter);
}
// Set.prototype
if (typeof Set !== 'undefined' && (Set.prototype.isPrototypeOf(iterObject)) {
  return Array.from(iter);
}
// Check for iterable WeakMap or custom iterator outside ES6 spec
if (typeof Symbol !== 'undefined' && (Symbol.iterator in iterObject)) {
  return Array.from(iter);
}
````

Please advise to do so, or just keep this PR as is.

**Comment by jridgewell** [comment]
Please add an `exec.js` test from https://github.com/babel/babel/pull/9794#discussion_r273774691

**Comment by clshortfuse** [comment]
@jridgewell Let me know when you want me to squash the commits.

**Comment by jridgewell** [comment]
No need, we can squash merge them on GitHub.

File: packages/babel-helpers/src/helpers.js [hunk 6]
```diff
@@ -860,10 +860,21 @@ helpers.arrayWithHoles = helper("7.0.0-beta.0")`
...
+
+    // Function Arguments
+    if (Object.prototype.toString.call(iter) === "[object Arguments]") {
+      return Array.from(iter);
+    }
+
+    // Check for iterable WeakMap, Map, Set, or TypedArray
+    // Check for custom iterator outside ES6 spec
+    if (Symbol.iterator in Object(iter)) {
+      return Array.from(iter);
```
nicolo-ribaudo: Nit: could you merge these `if`s to have a smaller helper?
i.e.
```js
if (
  typeof iter === "string" ||
  Object.prototype.toString.call(iter) === "[object Arguments]" ||
  Symbol.iterator in Object(iter)
)
```

nicolo-ribaudo:
We will fix this to also work with Map/Set without loading the polyfills from `core-js`, but thank you for at least fixing this case!

zloirock:
I don't see any sense in this PR in the current form since without `Symbol.iterator` will work only `arguments` and strings, but even they require `Array.from`. Spread will not work even with arrays without `Symbol.iterator`.

clshortfuse:
@zloirock I don't understand. You rather it be in a more broken state? 

It's a regression. There is no catch-all fix that I've seen without polyfilling Symbol. If we can at least solve partially resolve the regression and people don't have to change perfectly working V6 code around.

zloirock:
@clshortfuse I'm not against of adding support of work spread without `Symbol. iterator` to `babel@7`, however I don't think that the current behavior is a regression - it's a breaking change in a major release, but by my vision, it's available since it makes `babel` behavior more correct by the spec. However, OK, let `babel@7`'s spread works even without `Symbol.iterator`. I don't understand what's the heck is going on in this PR, see my comment above - those changes are almost completely useless since it does not support even arrays.

zloirock:
Ok, my vision of this helper for `babel@7`:
```js
function iterableToArray(it) {
  switch (Object.prototype.toString.call(it).slice(8, -1)) {
    // As an option, they will work even without `Array.from` polyfill,
    // however, it could cause problems with custom iterators:
    case "Array":
    case "Arguments":      // It will not work in old IE without `core-js`
    case "HTMLCollection": // 2 most popular iterable DOM collections,
    case "NodeList":       // for work with other you should load `core-js`
      return Array.prototype.slice.call(it);
    case "String":         // Since of code units, we should use `Array.from`
    case "Set":            // It will work with `es6-shim` if they will add
    case "Map":            // internal @@toStringTag support
      return Array.from(it);
  }
  // It's more correct by the spec than `in` check
  const iterator = it[Symbol.iterator];
  if (typeof iterator !== 'function') throw TypeError();
  return Array.from({ [Symbol.iterator]: iterator.bind(it) });
}
```

clshortfuse:
@zloroick, I originally had an array prototype check but we agreed to remove it since the previous calling function is going to call `Array.isArray`.

I have my old commits here. I did have a list of every ES6 type in spec. I decided against the typeof check, before calling Set, Map or Symbol because it would lead to the wrong error. The user should get notified they don't exist and not a TypeError. That would be a spec violation. It can be confusing to give the user the incorrect error and we don't want people to think the object they are using spread against is an invalid one.

I opted out of non ES6 prototypes like HTMLCollection because they're not part of the actual spec. It's stricter, but whitelisting would be too loose, IMO, and go against the goal of being to spec.

I'm being pretty strict to limiting to just what ES6 has with String and rest arguments. Anything else should throw the correct error if it's not available. If users don't want it to be so strict, we can improve the loose mode.

zloirock:
Ok, seems I missed a moment about `Array.isArray` in `arrayWithoutHoles` -> `toConsumableArray`, however, this PR still makes spread work only on `arguments` and strings and only with polyfilled `Array.from`, which is not really useful. As you can see in the example of the helper above, you can unobtrusively check the argument on internal classes without type/reference errors and loading unnecessary polyfills. Iterators for DOM collections were added to `core-js` only because of a `babel`'s feature request, so adding them to `core-js` and not adding to special cases of this helper (if they are will be added) are double standards.

clshortfuse:
@zloirock I can easily add the code. I don't have an issue with that. That's how I originally envisioned it, but I wasn't sure we were doing whitelists.

I've researched that Map, Set, Array, String, TypedArray (which is abstract), and Arguments are in Spec. I honestly did try the prototype name because it's not exactly 100% safe. But then again, we're checking Arguments the same way, so it would be a bad double standard.

The roadblock I reached is realizing spread doesn't support Array-likes (objects with `length`). It only supports iterables. `Array.from` is more flexible than ES6 spread. There's going to have to be some flexibility and I wasn't sure how loose we were willing to go.

My idea is to whitelist as much as possible and then go for Symbol.Iterator. I think looser than that, then the user should consider loose. We don't want code that only works on Babel, but doesn't work on real browsers (using Spread on Array-likes without Symbol.Iterator).

I'll work on adding Map and Set as you suggested with the prototype name in String format.

ljharb:
Note that es6-shim’s `Array.from` would work in that example without any changes.

clshortfuse:
Okay, I'm checking @zloirock code at a time that's not 2AM and on my phone 😆, and it does seem to meet all the targets I was looking for and adds the `slice.call()` I mentioned in one of the previous comments.

On the whitelist I had prepared, I had couple of other types that are spread compatible that are part of spec:

* [object String Iterator]
* [object Array Iterator]
* [object Map Iterator]
* [object Set Iterator]
* [object Int8Array]
* [object Uint8Array]
* [object Uint8ClampedArray]
* [object Int16Array]
* [object Uint16Array]
* [object Int32Array]
* [object Uint32Array]
* [object Float32Array]
* [object Float64Array]
* [object BigInt64Array]
* [object BigUint64Array]

Basically, it's Iterators which are iterable and the TypedArray types. Thankfully, TypedArray iterators are `[object Array Iterator]` which reduces the big list.

nicolo-ribaudo:
If there are built in iterators, won't there also be `Symbol.iterator`?

clshortfuse:
@nicolo-ribaudo Yeah. It would just be a question of whether it's faster to bypass whatever we're doing with Symbol check and immediately return with `Array.from`.

clshortfuse:
How does everyone feel about just committing this as-is with the limited fix? I wasn't really targeting fixing everything. I just wanted rest arguments working. I feel like like the other PR that targeted all of the Symbol issues has been abandoned. The regression of breaking rest arguments is over a year old at this point (March 8th 2018). 

I could open up another PR for fixing everything and we can start fresh.

ljharb:
Iterative fixes are good, to be sure.

jridgewell:
I'm happy with merging this as is.

zloirock:
@clshortfuse `BigUint64Array` / `BigInt64Array` available only in engines with `Symbol` support.

I'd prefer a version with all required special cases.

eps1lon:
Is there anything we can do to help getting this merged? This seems like a good first step resolving Symbol issues in older environments.

nicolo-ribaudo:
Sorry for the long delay!

PR #9798, opened 2019-03-30
Prevent JSX pragma detection in transform-typescript from leaking between files

SamyPesse:
| Fixed Issues?            | 
| Tests Added + Pass?      | No test added
| Documentation PR Link    | 

This PR fixes a bug in support of JSX pragma in Typescript (introduced in https://github.com/babel/babel/pull/9095).

We've noticed this bug while building a large codebase with webpack+babel+typescript. Some files from our projects are using `/** @jsx h*/` to use a custom function for JSX syntax; and some others are using the React syntax.

We've noticed that our react imports were being removed by babel because the plugin instance was being reused by `babel-loader` and webpack; and it was causing an error :

```
ReferenceError: React is not defined
```

This PR resets the `jsxPragma` being used to the initial value for each new file being processed.

I can't add a unit test for this bug because it requires processing 2 files (one with `/** @jsx something */` then one with the react syntax).

I've tested a build of this PR and it correctly fixes the issue for us.

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -74,7 +74,10 @@ export default declare((api, { jsxPragma = "React" }) => {
               // just bail if there is no binding, since chances are good that if
               // the import statement was injected then it wasn't a typescript type
               // import anyway.
-              if (binding && isImportTypeOnly(file, binding, state.programPath)) {
+              if (
+                binding &&
+                isImportTypeOnly(file, binding, state.programPath)
+              ) {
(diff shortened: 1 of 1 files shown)


SamyPesse:
@nicolo-ribaudo Thank you for the review, I'm quite new to the babel internals, I'll make the changes right now :)

SamyPesse:
I've made the changes, and tested it on our application (it works fine) 🚀 

I'm just wondering what is the signature of `file.get`, I did `file.get(key) || defaultValue` since the JSX pragma should never be an empty string.

Retsam:
Looks like this would fix #9549.

PR #9799, opened 2019-03-30
Added es3 backward compatibility for babel-plugin-transform-react-constant-elements

### LINKED ISSUE #9226: babel-helpers - es3 support
## Bug Report

**Current Behavior**

`babel-helpers` package is using keywords that are reserved in es3, so even if target is set to es3 it is still not producing the valid es3 code.

**Input Code**

Any React code.

**Expected behavior/code**

It should produce es3 compatible code.

```js
{
  compact: false,
  presets: [
    // transform-function-name is breaking React 0.14.x: '_renderedComponent' is null or not an object
    ['@babel/preset-env', { targets: ['IE 7'], loose: true /* es3 */, exclude: ['transform-function-name'] }],
    ['@babel/preset-react', { development: true }],
    ['@babel/preset-typescript', { isTSX: true, allExtensions: true }]
  ],
  plugins: [
    ['@babel/proposal-class-properties', { loose: true /* es3 */ }],
    ['@babel/proposal-object-rest-spread', { loose: true /* es3 */ }],
    'transform-es3-member-expression-literals', // es3
    'transform-es3-property-literals',          // es3
    'add-react-displayname',
  ]
}
```

**Environment**

- Babel version(s): 7.1.6
- Node/npm version: Node 8
- OS: dev: macOS, runtime: Windows CE 7 + Internet Explorer 7 Mobile (zebra MC32 hand scanner)
- Monorepo: ?
- How you are using Babel: as a transform plugin with rollup

**Possible Solution**

Simple fix:

```diff
diff -ur helpers.old/lib/helpers.js helpers/lib/helpers.js
--- helpers.old/lib/helpers.js	2018-12-13 20:51:50.000000000 -0500
+++ helpers/lib/helpers.js	2018-12-13 20:52:17.000000000 -0500
@@ -47,7 +47,7 @@
   export default function _createRawReactElement(type, props, key, children) {
     if (!REACT_ELEMENT_TYPE) {
       REACT_ELEMENT_TYPE = (
-        typeof Symbol === "function" && Symbol.for && Symbol.for("react.element")
+        typeof Symbol === "function" && Symbol["for"] && Symbol["for"]("react.element")
       ) || 0xeac7;
     }
 ```

N/A

### PR DESCRIPTION
**Pull request by sormy**
| Fixed Issues?            | Fixes #9226

Fixes babel helpers code to fix backward compatibility with old ancient es3 browsers like IE6, IE7, IE8 when `babel-plugin-transform-react-constant-elements` is used.

### CODE REVIEW

**Comment by nicolo-ribaudo** [comment]
Is this still needed after #9781?

**Comment by sormy** [comment]
> Is this still needed after #9781?

Unfortunately, yes, babel helpers code is emitted as it is without passing to transform plugins.

**Comment by nicolo-ribaudo** [comment]
> babel helpers code is emitted as it is without processing through transform plugins.

I believe that this is a bug; I need to check

**Comment by nicolo-ribaudo** [comment]
@sormy I can't reproduce it.

I'm running this code: (`load` is similar to `require`, I use it to load plugins from `master`)
```js
const input = `
<foo />;
`;

const output = babel.transformSync(input, {
  configFile: false,
  plugins: [
    load("plugin-transform-react-inline-elements"),
    load("plugin-syntax-jsx"),
  ],
  presets: [[load("preset-env"), { targets: { ie: 7 } }]],
});

console.log(output.code);
```

The output is:
```js
"use strict";

var REACT_ELEMENT_TYPE;

function _jsx(type, props, key, children) { if (!REACT_ELEMENT_TYPE) { REACT_ELEMENT_TYPE = typeof Symbol === "function" && Symbol["for"] && Symbol["for"]("react.element") || 0xeac7; } var defaultProps = type && type.defaultProps; var childrenLength = arguments.length - 3; if (!props && childrenLength !== 0) { props = { children: void 0 }; } if (props && defaultProps) { for (var propName in defaultProps) { if (props[propName] === void 0) { props[propName] = defaultProps[propName]; } } } else if (!props) { props = defaultProps || {}; } if (childrenLength === 1) { props.children = children; } else if (childrenLength > 1) { var childArray = new Array(childrenLength); for (var i = 0; i < childrenLength; i++) { childArray[i] = arguments[i + 3]; } props.children = childArray; } return { $$typeof: REACT_ELEMENT_TYPE, type: type, key: key === undefined ? null : '' + key, ref: null, props: props, _owner: null }; }

_jsx("foo", {});
```

As you can see, `Symbol.for` is transpiled to `Symbol["for"]`.

loganfsmyth:
This seems reasonable to me. We've generally tried to keep the helpers as compatible as possible since the helpers in `@babel/runtime` for instance may not be transformed by user plugins.

nicolo-ribaudo:
> since the helpers in @babel/runtime for instance may not be transformed by user plugins.

We transform them with preset-env: https://unpkg.com/@babel/runtime@7.4.2/helpers/jsx.js

This PR is only needed for users supporting old IE without transpiling down every es feature to old IE.

nicolo-ribaudo:
Yeah, they have not been released yet.

sormy:
If you are 100% sure that new version will transform helpers with preset-env targeting ES3 by default then this PR could be safely rejected.

nicolo-ribaudo:
I'm closing this PR. @sormy If you need the es3-compatible helpers, Babel 7.4.3 has been released :slightly_smiling_face:

Issue #9788, opened 2019-03-29
super call in private instance method incorrectly calls overridden method


MattiasBuelens:
## Bug Report

**Current Behavior**
Calling a `super` method from within a private instance method or accessor calls the overridden method instead of the superclass's method. Similarly, accessing a `super` property accesses the overridden property instead of the superclass's property.

For reference: in #9580, I reported that `super` method calls and property accesses didn't work at all, which was fixed in #9704. This issue shows that although such a `super` call/access no longer throws, it can still call the wrong method/accessor.

**Input Code**
[REPL link](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEDKCuAHApgJwMLitA3gKGkOggHsBbZAWWQBcALEgEwAoBKXAor4EgO1JDIAdCBIBzZgHIISNJNYBuToQC-eNXlCQYCAEaZt0ZAA8ayXox2yMWGPi4BiRKgCWANzBnq9Jmw5cuGRRUIVIKbwYWRWVoDS5EeF0QF2AI33Z7AMJ6FwghJ1cPL1pItiUuOKIwqhL0_yzoHn4SQRFxKRldeXKiNQ1eZAB3OESDKDYhBKSUtKiFIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=false&targets=&version=7.4.2)

I still can't get the REPL link to include additional plugins, so you have to manually add these:
* `@babel/plugin-proposal-class-properties`
* `@babel/plugin-proposal-private-methods`

```js
class SuperClass {
    someMethod() {
        console.log('super');
    }
}

class SubClass extends SuperClass {
    #privateMethod() {
        super.someMethod();
    }

    publicMethod() {
        this.#privateMethod();
    }

    someMethod() {
        console.log('sub');
    }
}

new SubClass().publicMethod();
```

**Expected behavior/code**
The console logs a message:
```
super
```

**Actual behavior/code**
The console logs a message:
```
sub
```

<details>
<summary>Generated code</summary>

```js
[...]

PR #9801, opened 2019-03-31
Fix super method call in private instance method calling overridden method

MattiasBuelens:
| Fixed Issues?            | #9788 

Previously, we transpiled the following code:
```js
class Sub extends Base {
  #privateMethod() {
    return super.superMethod();
  }
}
```
to:
```js
var _privateMethod2 = function _privateMethod2() {
  return babelHelpers.get(babelHelpers.getPrototypeOf(this), "superMethod", this).call(this);
};
```
This is incorrect: the lookup happens on the wrong prototype. The prototype chain of `this` looks something like:
```
this ---> Sub.prototype ---> Base.prototype ---> Object.prototype ---> null
```
`getPrototypeOf(this) === Sub.prototype`, so we look up `Sub.prototype.superMethod`. This may be overridden by `Sub`, so it's incorrect for the `super` call.

With this PR, the transpiled code becomes:
```js
var _privateMethod2 = function _privateMethod2() {
  return babelHelpers.get(babelHelpers.getPrototypeOf(Sub.prototype), "superMethod", this).call(this);
};
```
`getPrototypeOf(Sub.prototype) === Base.prototype`, so we look up `Base.prototype.superMethod`. This is the expected method for the `super` call.

### DIFF
--- a/packages/babel-helper-create-class-features-plugin/src/fields.js
+++ b/packages/babel-helper-create-class-features-plugin/src/fields.js
@@ -501,7 +501,9 @@ function replaceThisContext(path, ref, superRef, file, loose) {
     file,
     getObjectRef() {
       state.needsClassRef = true;
-      return path.node.static ? ref : t.thisExpression();
+      return path.node.static
+        ? ref
+        : t.memberExpression(ref, t.identifier("prototype"));
     },
   });
   replacer.replace();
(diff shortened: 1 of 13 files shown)


MattiasBuelens:
You might want to remove the "needs triage" label on #9788, seeing that it's fixed now. 😉

PR #9812, opened 2019-04-02
Transpile babel-polyfill to es5, instead of targeting node 6

nicolo-ribaudo:
| Fixed Issues?            | Without this PR, `BABEL_ENV=production make build-dist` throws
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -5,11 +5,12 @@ module.exports = function(api) {
 
   const includeCoverage = process.env.BABEL_COVERAGE === "true";
 
-  const envOpts = {
+  const envOptsNoTargets = {
     loose: true,
     modules: false,
     exclude: ["transform-typeof-symbol"],
   };
+  const envOpts = Object.assign({}, envOptsNoTargets);
 
   let convertESM = true;
   let ignoreLib = true;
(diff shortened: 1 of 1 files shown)


babel.config.js
```diff
@@ -97,6 +98,11 @@ module.exports = function(api) {
           convertESM ? "@babel/transform-modules-commonjs" : null,
         ].filter(Boolean),
       },
+      {
+        test: "./packages/babel-polyfill",
+        presets: [["@babel/env", envOptsNoTargets]],
+        plugins: [["@babel/transform-modules-commonjs", null]],
```
danez: This could just be this ?
```suggestion
        plugins: ["@babel/transform-modules-commonjs"],
```

nicolo-ribaudo: I meant to use `false` here to disable the plugin, since it should be handled by browserify.

```suggestion
        plugins: [["@babel/transform-modules-commonjs", false]],
```

nicolo-ribaudo:
We recently introduced some code in `@babel/parser` which was stranspiled to es6. Previously there was only code which is always transpiled to es5 so we didn't notice it.

chrisbutler:
correct me if i'm wrong, but this is a pretty major regression, no? is there a way to prevent something like this happening again?

PR #9808, opened 2019-04-02
refactor: unify polyfill with noConflict mode

xtuc:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | `Fixes #1, Fixes #2` 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

PR #9780, opened 2019-03-28
Fix #9232, add some missed modules to `@babel/polyfill/noConflict`

zloirock:
Fixes #9232

PR #9815, opened 2019-04-03
babel-polyfill: move global import

xtuc:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Follow up from #9780 and #9808

### DIFF
--- a/packages/babel-polyfill/src/index.js
+++ b/packages/babel-polyfill/src/index.js
@@ -1,7 +1,5 @@
 import "./noConflict";
 
-import global from "core-js/library/fn/global";
-
 if (global._babelPolyfill && typeof console !== "undefined" && console.warn) {
   console.warn(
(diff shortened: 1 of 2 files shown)


packages/babel-polyfill/src/index.js
```diff
@@ -1,7 +1,5 @@
 import "./noConflict";
 
-import global from "core-js/library/fn/global";
-
 if (global._babelPolyfill && typeof console !== "undefined" && console.warn) {
```
nicolo-ribaudo: Isn't `global` only needed here?

xtuc: ah! I had the global object proposal in mind, but it's not even name that way anymore. Let's close this PR.

PR #9833, opened 2019-04-08
Use proxy wrap instead of function wrap for function.sent generator proposal

myuseringithub:
For `babel-plugin-proposal-function-sent` plugin the generator functions are wrap by a non generator function which prevents access to the original generator properties. Thus, detecting if a function is a generator after code transformation will be impossible. To solve this issue a proxy can be used to capture function calls through the `apply` handler, and reflect all other handlers to the original function.

### DIFF
--- a/packages/babel-helper-proxy-wrap-function/.npmignore
+++ b/packages/babel-helper-proxy-wrap-function/.npmignore
@@ -1,0 +1,3 @@
+src
+test
+*.log
--- a/packages/babel-helper-proxy-wrap-function/README.md
+++ b/packages/babel-helper-proxy-wrap-function/README.md
@@ -1,0 +1,19 @@
+# @babel/helper-proxy-wrap-function
+
+> Helper to wrap functions inside a proxy apply handler.
+
+See our website [@babel/helper-proxy-wrap-function](https://babeljs.io/docs/en/next/babel-helper-proxy-wrap-function.html) for more information.
+
+## Install
+
+Using npm:
+
+```sh
+npm install --save-dev @babel/helper-proxy-wrap-function
+```
+
+or using yarn:
+
+```sh
+yarn add @babel/helper-proxy-wrap-function --dev
+```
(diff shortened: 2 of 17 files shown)


myuseringithub:
How can I get the built REPL to work with a modified plugin (included in the PR) ? https://babeljs.io/repl/build/10677/

jridgewell:
I don't understand what this is trying to do?

myuseringithub:
@jridgewell I've updated in this PR the `plugin-proposal-function-sent` which an implementation that uses proxies instead of function wraps. How can I try it out in the provided [REPL link](https://babeljs.io/repl/build/10677/) 
It doesn't seem to be working.

loganfsmyth:
> detecting if a function is a generator after code transformation will be impossible

To me this seems like a poor API design. There's nothing stopping a user of an API from doing this themselves either, like say they have
```
doThing(function*(inputData){

});
```
and refactor their code to be
```
function* fn1(inputData){}
function* fn2(inputData){}
doThing(function(inputData) {
  if (inputData.type === "impl1") return fn1(inputData);
  return fn2(inputData);
});
```
I'd be totally surprised to find that they code suddenly broke. The type of the function itself is meaningless, the only thing that should matter to `doThing` is whether the object returned by the callback function is an iterable object. Treating generator functions as a special case is asking for problems, with or withing Babel in the loop.

nicolo-ribaudo:
Also, keep in mind that proxies only work in modern browsers. Babel will never be able to use unpolyfillable built-ins.

jridgewell:
Is there a bug that is solved with this? Proxies aren't available everywhere, so I'm not sure this is a good approach.

myuseringithub:
I would like to use generators for handing back control to the client, allowing them to manipulate internal function execution. 
Was experimenting with a controller/switch function that propagates execution to an implementing function that could be either a generator or a regular function, it executes it differently depending on it's type.

nicolo-ribaudo:
Did you try using `transform-regenerator` with only `@babel/plugin-syntax-function-sent`? I think that it would set the correct prototype.

PR #9847, opened 2019-04-10
Remove core-js dependency from @babel/register

coreyfarrell:
| Fixed Issues?            |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Existing tests pass
| Documentation PR Link    |
| Any Dependency Changes?  | :+1:

The dependency `core-js` is not actually used by `@babel/register` so this PR removes it.

### DIFF
--- a/packages/babel-register/package.json
+++ b/packages/babel-register/package.json
@@ -13,7 +13,6 @@
     "./lib/node.js": "./lib/browser.js"
   },
   "dependencies": {
-    "core-js": "^3.0.0",
     "find-cache-dir": "^2.0.0",
     "lodash": "^4.17.11",
     "mkdirp": "^0.5.1",


nicolo-ribaudo:
While this could be removed, I am a bit worried that it will break users using `@babel/register` `@babel/preset-env` with `corejs: 3` without explicitly listing it in their dependencies.

cc @babel/babel

suchipi:
Perhaps we should move the core-js dep to preset-env, then?

nicolo-ribaudo:
> Perhaps we should move the core-js dep to preset-env, then?

`preset-env` doesn't rely on `core-js`. Only it's output, when explicitly enabled by the user, does.
Also, we would need somehow to add both core-js 2 and 3 without aliasing them so it's not possible.

ljharb:
ok, fair point on that. still tho, theoretically any test of any kind on a babelified file would expose it?

PR #9848, opened 2019-04-10
[refactor] Use functions instead of classes in @babel/parser

nicolo-ribaudo:
This is my first attemp to make the parser faster by removing the class-based logic. I will add some data to this PR to show the numbers.

Most of the changes in this PR are `this.foo` -> `foo` and `ClassMethod` -> `FunctionDeclaration`.

The parser is organized in different "chunks": one for the tokenizer, one for the parser, and one for each plugin. Those chunks are bundled by rollup and put into the `createParser` function of `packages/babel-parser/src/index.js`. Every chunk exposes an "entry point" (specified in the `load` function called inside `createParser`). If an entry point has a default export, it becomes the return value of `load("entrypoint.js")`.
Those chunks can export their functions, which can either be imported by other chunks (e.g. `import { parseIdentifier } from "::build-tool::bindings/parser`) or overwritten, by exporting a function with the same name. The "previous version" of an overwritten method can be accessed using the `original(functionName)` function.

<details>
<summary>To understand how it works, take a look at this example:</summary>

```js
// index.js

import { load } from "::build-tool::";

function createParser() {
  load("./a.js");

  if (plugin) {
    load("./b.js");
  }
}
```

```js
// a.js

export function getOne() {
  return 1;
}

export function getThree() {
  return getOne() + getTwo();
}

function getTwo() {
  return getOne() + getOne();
}
```

```js
// b.js

import { original } from "::build-tool::";
import { getOne } from "::build-tool::bindings/a";

// Overwrites
export function getThree() {
  return original(getThree)() + getTen();
}

function getTen() {
  return getTwo() / 2;
}

function getTwo() {
  return 20;
}
```

Becomes

```js
function createParser() {
  function getOne() {
    return 1;
  }

  function getThree() {
    return getOne() + getTwo();
  }

  function getTwo() {
    return getOne() + getOne();
  }

  if (plugin) {
    let original_getThree = getThree;

    getThree = function () {
      return original_getThree() + getTen();
    }

    function getTen() {
      return _getTwo() / 2;
    }

    function _getTwo() {
      return 20;
    }
  }
}
```
</details>

If you want to check the bundled version of the parser, you can clone this branch and run `make build`: it will be generated at `packages/babel-parser/bundle/index.js` (and then transpiled in `babel-parser/lib/index.js`). Please note that `make watch` doesn't work yet with the new build step.

Next steps: I will check how a big change would be to remove the overwriting logic and just add some `if (hasPlugin(...))` where needed, like it was suggested on Slack: this will allow us to not introduce the custom build step.


nicolo-ribaudo:
I'm not able to reproduce anymore the 10% performance improvement I posted on Slack.
These numbers are similar both when using the big IIFE create by the custom build step and when inlining the plugins (I have only inlined the JSX one before running the benchmark) so that the IIFE isn't needed anymore.

```
┌──────────────────────┬─────────────────────────────┬─────────────────────────────┐
│ fixture              │ babel (old)                 │ babel (new)                 │
├──────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ ember.debug.js       │ 10.42 ops/sec ±4.53% (96ms) │ 9.99 ops/sec ±1.67% (100ms) │
├──────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ jquery.js            │ 55.11 ops/sec ±1.05% (18ms) │ 52.21 ops/sec ±0.49% (19ms) │
├──────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ angular.js           │ 25.14 ops/sec ±3.46% (40ms) │ 25.96 ops/sec ±0.59% (39ms) │
├──────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ babylon-dist.js      │ 46.58 ops/sec ±1.58% (21ms) │ 41.42 ops/sec ±0.77% (24ms) │
├──────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ backbone.js          │ 270 ops/sec ±0.85% (4ms)    │ 240 ops/sec ±0.34% (4ms)    │
├──────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ react-with-addons.js │ 25.79 ops/sec ±2.8% (39ms)  │ 23.3 ops/sec ±1.76% (43ms)  │
└──────────────────────┴─────────────────────────────┴─────────────────────────────┘
```

If I understand why it is slower (or why it isn't faster, since other people's experience suggested that it should have been), I might reopen this PR.

PR #9861, opened 2019-04-15
Fix: PrivateName Identifier should not be isReferenced.

### LINKED ISSUE #9859: transform-module-commonjs throws when a PrivateName matches an import name.
## Bug Report

**Current Behavior**
TypeError: Property id of PrivateName expected node to be of a type ["Identifier"] but instead got "MemberExpression"

**Input Code**
[REPL](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=JYWwDg9gTgLgBAbxgUwM4wIwBo4vQJhzxgGYBfOAMyghDgHIA6AegEMA7ATwHcALZKMnoBuAFCjkAD0iw4AYwA2rVKjgBRSa3ALkiUXDgBiYhjgBeOOwCuChWIPG0MfAAoAlIjL64Ac2TxHdBJ3RDhBGCsodksbOzgvLyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=&prettier=false&targets=&version=7.4.3&externalPlugins=%40babel%2Fplugin-syntax-class-properties%407.2.0%2C%40babel%2Fplugin-transform-modules-commonjs%407.4.3)

```js
import {test1, test2, test3} from './anywhere';

export class Example {
  #test1 = test1;
  #test2() { return test2}
  get #test3() { return test3; }
}
```

In this example each of these private declarations can independently cause the error.

**Expected behavior/code**
PrivateName's should not be seen by the `ReferencedIdentifier` visitor method.

```js
{
	"plugins": [
		"@babel/plugin-transform-modules-commonjs",
		"@babel/plugin-syntax-class-properties"
	]
}
```

**Environment**
- Babel version(s): 7.4.3
- Node/npm version: node 10.15.3, npm 6.4.1 / node v12.0.0-nightly20190411655c90b287, npm 6.9.0
- OS: Fedora 28
- Monorepo: no
- How you are using Babel: tested with CLI, register and REPL.

**Possible Solution**
Return false for `parent.type === 'PrivateName'` from `@babel/types/src/validators/isReferenced.js`

I have a commit prepared locally including a test, this commit is a followup to #9857 (due to sharing the new test fixture directory).  I'm very confident about what I've already posted to #9857, I don't want to add to that commit without feedback on my proposed fix.

Interesting note https://github.com/babel/babel/blob/master/packages/babel-helper-module-transforms/src/rewrite-live-references.js#L83-L88 has code to skip ClassProperty and Function paths, though I think this is ineffective and unneeded (these paths do not hit the ReferencedIdentifier visitor method even if `path.skip` is removed).

coreyfarrell:
| Fixed Issues?            | Fixes #9859
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    |
| Any Dependency Changes?  |

This resolves an exception from `@babel/transform-module-commonjs` when any kind of class private field is declared with the same name as a named import.

### DIFF
--- a/packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/canonical/output.js
+++ b/packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/canonical/output.js
@@ -3,7 +3,7 @@ var Point =
 function () {
   "use strict";
 
-  function Point(_x2 = 0, _y2 = 0) {
+  function Point(x = 0, y = 0) {
     babelHelpers.classCallCheck(this, Point);
     Object.defineProperty(this, _x, {
       writable: true,
@@ -13,8 +13,8 @@ function () {
       writable: true,
       value: void 0
     });
-    babelHelpers.classPrivateFieldLooseBase(this, _x)[_x] = +_x2;
-    babelHelpers.classPrivateFieldLooseBase(this, _y)[_y] = +_y2;
+    babelHelpers.classPrivateFieldLooseBase(this, _x)[_x] = +x;
+    babelHelpers.classPrivateFieldLooseBase(this, _y)[_y] = +y;
   }
 
   babelHelpers.createClass(Point, [{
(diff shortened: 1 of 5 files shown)


nicolo-ribaudo:
Thank you!

PR #9870, opened 2019-04-18
Fix flow types in traverse/path/family and enable flow

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-traverse/src/path/family.js
+++ b/packages/babel-traverse/src/path/family.js
@@ -1,3 +1,4 @@
+// @flow
 // This file contains methods responsible for dealing with/retrieving children or siblings.
 
 import type TraversalContext from "../index";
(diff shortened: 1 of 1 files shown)


packages/babel-traverse/src/path/family.js
```diff
@@ -203,9 +210,12 @@ export function getBindingIdentifierPaths(
     }
   }
 
+  // $FlowIssue Object.create() is object type
```
nicolo-ribaudo: Doesn't using `const id:  { [string]: NodePath } = Object.create(null);` above work?

danez: No flow says that the result of `Object.create()` is not compatible to object type. facebook/flow#4967

Issue #9305, opened 2019-01-09
parser: TypeScript: Parse error on abstract class with abstract method and optional methods in abstract/declare classes.


armano2:
## Bug Report

**Current Behavior**
Babel expect that abstract methods must have body, this is not a case for typescript. 
This is also issue for optional methods in abstract/declare classes.

```
SyntaxError: Unexpected token, expected "{"
```

**Input Code**
```ts
abstract class AbstractSocket {
    abstract createSocket(): Promise<string>;
}
```
```ts
abstract class AbstractSocket {
    createSocket?(): Promise<string>;
}
```
```ts
declare class Foo {
    bar?(): any;
}
```

**Expected behavior/code**
There should not be any parse errors

**Babel Configuration (.babelrc, package.json, cli command)**

```js
  const babel = require('@babel/parser');
  const plugins: ParserPlugin[] = [
    'typescript',
    'objectRestSpread',
    'decorators-legacy',
    'classProperties',
    'asyncGenerators',
    'dynamicImport',
    'estree',
    'bigInt',
    'importMeta'
  ];
  if (jsx) {
    plugins.push('jsx');
  }

  return babel.parse(text, {
    sourceType: 'unambiguous',
    allowImportExportEverywhere: true,
    allowReturnOutsideFunction: true,
    ranges: true,
    plugins
  });
```

**Environment**
- Babel version(s):  7.2.3
- Node/npm version: 6-11
- OS: ----
- Monorepo: ---
- How you are using Babel: @babel/parser -> `parse`

armano2:
@existentialism is it something what we want to support?

RWOverdijk:
Please, yes. If anything, do it for eslint.

nicolo-ribaudo:
This only happens when the `estree` plugin is enabled.

RWOverdijk:
I don't think I'm using that plugin and I'm seeing the same happen. Does that mean eslint adds that?

devongovett:
Should be fixed by https://github.com/babel/babel/pull/9890

armano2:
can confirm its fixed

PR #9890, opened 2019-04-22
Fix parsing typescript bodiless methods with the estree plugin also enabled

devongovett:
This fixes parsing of typescript bodiless methods when the estree plugin is also enabled. For context, I am adding TS support to react-docgen, which uses babel-parser with recast and therefore requires estree. https://github.com/reactjs/react-docgen/pull/348

Currently, when both the typescript plugin and estree are enabled, code like the following causes a parsing error.

```typescript
abstract class TSAbstractClass {
  abstract foo(name: string): boolean;
}
```

The problem was that the type was overridden prior to parsing, and the typescript plugin uses the type to determine whether a bodiless method is allowed. The PR fixes this issue by passing the original type through and only overriding it afterward.

| Fixed Issues?            | #9305
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/src/plugins/estree.js
+++ b/packages/babel-parser/src/plugins/estree.js
@@ -220,7 +220,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         isAsync,
         isConstructor,
         allowsDirectSuper,
-        "MethodDefinition",
+        "ClassMethod",
         true,
       );
       if (method.typeParameters) {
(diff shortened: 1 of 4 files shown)


devongovett:
I'm not sure why CI is failing. Seems to be something related to preset-env. Looks like a few other PRs are also failing.

nicolo-ribaudo:
Yeah, CI failure is unrelated.

Issue #9240, opened 2018-12-26
[flow] Parsing error when calling generic functions with type arguments when flow pragma is not first comment 


FermiDirak:
## Bug Report

**Current Behavior**
When the @flow pragma is not specified as the first comment of a JS or JSX file, babel parser will not be able to parse any code that passes type arguments.

**Input Code**
```js
// arbitrary comment
// @flow

const n: number = 5; // this works fine

const set = new Set<number>([1, 2, 3]);
        // unexpected token^
```

**Expected behavior/code**
Flow should be able to parse functions that are called with type arguments so long as the @flow pragma exists before any code. 

**Babel Configuration (.babelrc)**

```js
{
  presets: [
    "@babel/preset-env",
    "@babel/preset-react",
    "@babel/preset-flow",
  ]
}
```

**Environment**
- Babel version: [^7.0.0]
- Node/npm version: [Node 8.11.1]
- OS: [OSX 10.13.4]
- Monorepo: [yes]
- How you are using Babel: [`loader`]

**Possible Solution**

It appears this line ignores the flow pragma if it isn't found in the first comment of the source file.
https://github.com/babel/babel/blob/c586d4e8cac816941bf2ba559472fd9cc3150921/packages/babel-parser/src/plugins/flow.js#L77

This could be changed a-la Flow-Strip-Types, which first checks all comments for a flow pragma before it begins transforming. (https://github.com/babel/babel/blob/c586d4e8cac816941bf2ba559472fd9cc3150921/packages/babel-plugin-transform-flow-strip-types/src/index.js). Perhaps a solution like this could be used here as well.

nateshmbhat:
I had this problem because of older babel version which I fixed using a stackoverflow answer  : 
Check this stackoverflow answer : https://stackoverflow.com/a/53927430/6665568

JakubRimal:
@FermiDirak Thank you very much, you saved me at least a few hours as without seeing this issue I would never guess that the problem is because of a position of the `@flow`... :open_mouth: :smiley:

pwlmaciejewski:
I'm experiencing the same issue with flow enabled by default on all files:

```
// .flowconfig
[options]
all=true
```

Adding `@flow` pragma in the first line of the throwing files fixes fixes it.

nicolo-ribaudo:
Babel doesn't read your `.flowconfig` file.

pwlmaciejewski:
[...]

PR #9891, opened 2019-04-24
allow directives and other comments before flow pragma

tanhauhau:
handles 3 cases correctly:
- allow any arbitrary comments before `// @flow` flow pragma comment
- allow any arbitrary strings (assuming it's directive) before `// @flow` flow pragma comment
- disallow flow syntax even if `// @flow` flow pragma content is the first comment, but appear in the middle of the code

it is slightly different than the flow logic I described in https://github.com/babel/babel/issues/9240#issuecomment-485370957, but I think it's fair enough given the differences in the lexer logic between flow and babel, and it's definitely better than the current logic

| Fixed Issues?            | #9240 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/src/plugins/flow.js
+++ b/packages/babel-parser/src/plugins/flow.js
@@ -95,12 +95,21 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return this.getPluginOption("flow", "all") || this.flowPragma === "flow";
     }
 
+    finishToken(type: TokenType, val: any): void {
+      if (!(type === tt.string || type === tt.semi)) {
+        if (this.flowPragma === undefined) {
+          this.flowPragma = null;
+        }
+      }
+      return super.finishToken(type, val);
+    }
+
     addComment(comment: N.Comment): void {
       if (this.flowPragma === undefined) {
         // Try to parse a flow pragma.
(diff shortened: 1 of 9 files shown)


packages/babel-parser/src/plugins/flow.js
```diff
@@ -95,12 +95,26 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return this.getPluginOption("flow", "all") || this.flowPragma === "flow";
     }
 
+    finishNodeAt<T: N.Node>(
```
loganfsmyth: Would we be able to override `finishToken` to get the behavior you're looking for of bailing after a semi/string token?

tanhauhau: hmm.. interesting, let me try that. But which means it would have to handle whitespace, slash tokens for comments right?

tanhauhau:
Interesting the pipeline failed for unknown reasons

nicolo-ribaudo:
It's fixed by https://github.com/babel/babel/pull/9906

tanhauhau:
Thanks!

PR #9901, opened 2019-04-26
fix transform-flow-comments for import types

### LINKED ISSUE #6142: Flow comments plugin creates incorrect code
(I opened an issue on the affected repository [react-leaflet#363](https://github.com/PaulLeCam/react-leaflet/issues/363))

Flow-comments output incorrect code when transforming consecutive flow "statements". As shown in the issue linked above,
```
import { GeoJSON as LeafletGeoJSON, type LatLng, type Layer } from 'leaflet';
```
"renders" as 
```
import { GeoJSON as LeafletGeoJSON, /*:: type LatLng*/, /*:: type Layer*/ } from 'leaflet';
```
You'll notice the 2 hanging commas.

Reproducible on the repository @1.6.2

tanhauhau:
| Fixed Issues?            | Fixes #6142
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-transform-flow-comments/package.json
+++ b/packages/babel-plugin-transform-flow-comments/package.json
@@ -12,6 +12,7 @@
     "babel-plugin"
   ],
   "dependencies": {
+    "@babel/generator": "^7.0.0",
     "@babel/helper-plugin-utils": "^7.0.0",
     "@babel/plugin-syntax-flow": "^7.2.0"
   },
--- a/packages/babel-plugin-transform-flow-comments/src/index.js
+++ b/packages/babel-plugin-transform-flow-comments/src/index.js
@@ -1,19 +1,24 @@
 import { declare } from "@babel/helper-plugin-utils";
 import syntaxFlow from "@babel/plugin-syntax-flow";
 import { types as t } from "@babel/core";
+import generateCode from "@babel/generator";
 
 export default declare(api => {
   api.assertVersion(7);
 
-  function wrapInFlowComment(path, parent) {
+  function attachComment(path, comment) {
     let attach = path.getPrevSibling();
     let where = "trailing";
     if (!attach.node) {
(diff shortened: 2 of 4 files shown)


packages/babel-plugin-transform-flow-comments/package.json
```diff
@@ -16,10 +16,12 @@
     "@babel/plugin-syntax-flow": "^7.2.0"
   },
   "peerDependencies": {
-    "@babel/core": "^7.0.0-0"
+    "@babel/core": "^7.0.0-0",
+    "@babel/generator": "^7.0.0-0"
```
nicolo-ribaudo: Requiring users to install a new package is a breaking change.
I believe you can import it from `@babel/core`.

tanhauhau: interestingly `@babel/core` does not export `generate` function from `@babel/generator`, should I export it out from `@babel/core`?

nicolo-ribaudo: Oh I didn't know that.
If we added a `generate` export, we couldn't use it in this plugin anyway because it needs to be compatible with `@babel/core@^7.0.0`. We can add `@babel/generator` to the dependencies of this plugin.

tanhauhau: erm. then is it okay to use `@babel/generator` as `peerDependency`? since it's `@babel/core`'s dependency

nicolo-ribaudo: No, because peer dependencies must be explicitly installed by the user, otherwise npm will log a warning

tanhauhau:
@existentialism is there something any improvements I can make for this PR to be merged?

PR #9902, opened 2019-04-26
Fix destructuring assignments being transpiled for edge 15

eps1lon:
| Fixed Issues?            | No existing issue found
| Documentation PR Link    | 
| Any Dependency Changes?  |

[preset-env transpiles object destructuring in edge 17](https://babeljs.io/repl#?babili=false&browsers=edge%20%3E%3D%2017&build=&builtIns=false&spec=true&loose=false&code_lz=GYVwdgxgLglg9mABAYTgWwA4IKZigCgwCc4MBnASkQG8AoASAgTKhuDjgBoAjAQyIC-iALyJipMgG5aAoA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=script&lineWrap=true&presets=env&prettier=true&targets=Node-8.0&version=7.4.3&externalPlugins=%40babel%2Fplugin-transform-parameters%407.4.3).

Current behavior: https://github.com/babel/babel/pull/9902/commits/2254f7ccea0a144031b7477da21a2d7f9428acb3#diff-0d3abae52af132aba0fb19ce6f915ccf
Expected behavior: https://github.com/babel/babel/pull/9902/commits/24a0e71eeeb8e86c39842f90451d7ee4979ee640#diff-0d3abae52af132aba0fb19ce6f915ccf

The destructuring assignment is implemented as early as edge 15 (loosely in 14) though. Was this required for some transform dependencies when function parameters are transpiled? Specifically destructuring in the function parameters is only properly implemented in later browsers. 

Will need to check if those are still transpiled properly in the deploy preview in this PR.

On another note: preset-env is really nice but I wish the documentation would be more clear that there are legitimate use cases for specifiyng each transform instead of using the preset. `transform-destructuring` is currently used up to firefox 53 just because the return in custom iterator implementations is not called. I don't think this has any impact on basic object destructuring which would allow me to drop the transform for firefox 49. Maybe this is more of an issue with the [mdn page which lists object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Browser_compatibility) as being implemented in firefox 49 but then babel transpiles it anyway. Pretty confusing but I guess this is what browser support still requires.


nicolo-ribaudo:
Can you add a test for this code?
```
((a, {b = 0, c = 3}) => {
  return a === 1 && b === 2 && c === 3;
})(1, {b: 2});
```

It's the code for which compat-table reports edge 15 as unsupported.

eps1lon:
> MDN isn't really reliable, I suggest using [kangax.github.io/compat-table/es6](https://kangax.github.io/compat-table/es6/) instead.

For sure. But the outcome was the same. preset-env behavior didn't match what I would expect from the compat table.

> Can you add a test for this code?
> 
> ```
> ((a, {b = 0, c = 3}) => {
>   return a === 1 && b === 2 && c === 3;
> })(1, {b: 2});
> ```

> It's the code for which compat-table reports edge 15 as unsupported.

I would expect only `transform-parameters` to be applied in that case which results in 
```js
((a, _ref) => {
  let { b = 0, c = 3 } = _ref;
  return a === 1 && b === 2 && c === 3;
})(1, {
  b: 2
});
```
which is edge 15 compatible.

Didn't quite click for me where to put the test but I got it now. Reworking the PR.

nicolo-ribaudo:
Yeah, the parameters transform is enough

eps1lon:
e9b27ae10b96cf5db4ac46963c1675879740d456 should make my intentions clearer. Hope this is the right place for the test.

Got one test failing locally but this happened before any changes. Maybe a rebase will fix it.

nicolo-ribaudo:
Yeah, it's fixed on master. If the only failing test is that one, it's ok.

eps1lon:
[Looks good](https://babeljs.io/repl/build/10741/#?babili=false&browsers=edge%20%3E%3D%2015&build=&builtIns=false&spec=true&loose=false&code_lz=GYVwdgxgLglg9mABAYTgWwA4IKZigCgwCc4MBnASkQG8AoASAgTKhuDjgBoAjAQyIC-iALyJipMgG5aAoA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=script&lineWrap=true&presets=env&prettier=true&targets=Node-8.0&version=7.4.4%2Bpr.9902&externalPlugins=).

PR #9912, opened 2019-04-26
[legacy decorators] Allow decorating generator methods

### LINKED ISSUE #9852: Legacy Decorators + Generator Methods
## Bug Report

**Current Behavior**
Legacy decorators cannot decorate generator methods. An `Unexpected token` error is thrown:

```
/repl: Unexpected token (5:21)

  3 | class Foo {
  4 |   @deco
> 5 |   *generatorMethod() {}
    |                      ^
  6 | }
```

**Input Code**
[REPL](https://babeljs.io/en/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=true&code_lz=GYVwdgxgLglg9mABAEwKYTgCgJSIN4C-AUERADYCGAzlYgGJxz5GKIACaGLiAVAOaowqAE4UocYQFlUUABZxkOfMQJA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Cstage-2&prettier=false&targets=&version=7.4.3&externalPlugins=)

```js
function deco() {}

class Foo {
  @deco
  *generatorMethod() {}
}
```

**Expected behavior/code**
No error should be thrown and the method should be transpiled like normal methods.

**Babel Configuration**
Presets: `es2015` + `stage-2` 
Decorators: legacy

**Environment**
Babel version(s): 7.4.3

**Possible Solution**
#8742

#5704, https://github.com/babel/babylon/issues/382

nicolo-ribaudo:
The old proposal used LeftHandSideExpression (instead of
AssignmentExpression) to satisfy this usecase:
https://github.com/wycats/javascript-decorators/commit/e240cbc91a45d87b07fb2ba6fd7a499e4bf27e46

| Fixed Issues?            | Fixes #9852
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR will break the code of people doing things like
```js
class A {
  @foo && bar
  method() {}
}
```

I don't think that it is a problem because:
1) I have never seen that pattern anywhere
2) Prettier inserts parentheses, which still work (`@(foo && bar)`)
3) The "legacy" proposal disallows that code.

### DIFF
--- a/packages/babel-parser/test/fixtures/experimental/decorators/method-generator-decorator/input.js
+++ b/packages/babel-parser/test/fixtures/experimental/decorators/method-generator-decorator/input.js
@@ -1,0 +1,4 @@
+class Foo {
+  @deco
+  *generatorMethod() {}
+}
--- a/packages/babel-parser/src/parser/statement.js
+++ b/packages/babel-parser/src/parser/statement.js
@@ -390,7 +390,7 @@ export default class StatementParser extends ExpressionParser {
       node.expression = this.parseMaybeDecoratorArguments(expr);
       this.state.decoratorStack.pop();
     } else {
-      node.expression = this.parseMaybeAssign();
+      node.expression = this.parseExprSubscripts();
     }
     return this.finishNode(node, "Decorator");
   }
--- a/packages/babel-parser/test/fixtures/experimental/decorators/method-generator-decorator/output.json
+++ b/packages/babel-parser/test/fixtures/experimental/decorators/method-generator-decorator/output.json
@@ -1,0 +1,175 @@
+{
+  "type": "File",
+  "start": 0,
+  "end": 45,
+  "loc": {
+    "start": {
+      "line": 1,
+      "column": 0
+    },
+    "end": {
+      "line": 4,
+      "column": 1
+    }
+  },
+  "program": {
+    "type": "Program",
+    "start": 0,
+    "end": 45,
+    "loc": {
+      "start": {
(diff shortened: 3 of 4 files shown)


packages/babel-parser/test/fixtures/experimental/decorators/method-generator-decorator/input.js
```diff
@@ -0,0 +1,4 @@
+class Foo {
+  @deco
+  *generatorMethod() {}
```
NullVoxPopuli: YAAAAAS. This'll make ember-concurrency beautiful!

danez:
> This PR will break the code of people doing things like
>
> ```js
> class A {
>   @foo && bar
>   method() {}
> }
> ```

Just to understand correctly it wasn't allowed in legacy decorators, but it was in the old "new" implementation, but now in the "newest" it is disallowed?

nicolo-ribaudo:
It was allowed in a very early version of legacy decorators. It is disallowed in the "last" version of legacy decorators, in the "old new (a few months ago)" implementation and in the new not-yet-implemented version. Other than that, it is also disallowed by Flow and TypeScript.

robclancy:
Is this going to be in the next release? It's a pretty big issue requiring ugly workarounds if needing decorators for generators. A lot of people are waiting on this.

PR #9924, opened 2019-04-30
fix: Add missing dependency @babel/types

elliottsj:

This fixes an issue where '@babel/types' cannot be resolved when using
strict package managers like pnpm, e.g.

    /path/to/project/node_modules/.registry.npmjs.org/@babel/parser/7.4.4/node_modules/@babel/parser/typings/babel-parser.d.ts
    ERROR in /path/to/project/node_modules/.registry.npmjs.org/@babel/parser/7.4.4/node_modules/@babel/parser/typings/babel-parser.d.ts(11,71):
    TS2307: Cannot find module '@babel/types'.

    > Build error occurred
    Error: > Build failed because of webpack errors
        at Object.build [as default] (/path/to/project/node_modules/.registry.npmjs.org/next/8.1.0_react-dom@16.8.6+react@16.8.6/node_modules/next/dist/build/index.js:192:15)

### DIFF
--- a/packages/babel-parser/package.json
+++ b/packages/babel-parser/package.json
@@ -27,6 +27,9 @@
   "engines": {
     "node": ">=6.0.0"
   },
+  "dependencies": {
+    "@babel/types": "^7.4.4"
+  },
   "devDependencies": {
     "@babel/code-frame": "^7.0.0",
     "@babel/helper-fixtures": "^7.4.4",


loganfsmyth:
Hmm, I don't think this is the right fix. The only place in the parser that uses `@babel/types` is the type definitions, so adding a whole extra dependency that isn't used at runtime isn't ideal.

elliottsj:
Yes, TypeScript depends on npm to install dependencies for type declarations, as described on https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies:

>All dependencies are managed by npm. Make sure all the declaration packages you depend on are marked appropriately in the `"dependencies"` section in your `package.json`.

arzyu (approved):

PR #9974, opened 2019-05-14
avoid exceptions for control flow

### LINKED ISSUE #9972: Parser performance is very bad when typescript is enabled
## Bug Report

**Current Behavior**

I reported this before in #8779, and we made things a bit better, but it seems to have regressed again recently. The problem is the typescript parser plugin uses exceptions to back out of lookahead branches. For example, see the `tsTryParseAndCatch` method:

https://github.com/babel/babel/blob/1969e6b6aa7d90be3fbb3aca98ea96849656a55a/packages/babel-parser/src/plugins/typescript/index.js#L1249-L1260

This causes the typescript plugin to degrade the performance of the parser in a manner that is non-linear WRT the input size. 

**Repro**
Consider the following script to demo the problem:

```js
const {parse} = require('@babel/parser');
const fs = require('fs');

const input = fs.readFileSync(process.argv[2], 'utf8');

let start = Date.now();
parse(input, {
  plugins: [],
});
console.log('parse time with no TS ' + (Date.now() - start));


start = Date.now();
parse(input, {
  plugins: ['typescript'],
});
console.log('parse time with TS ' + (Date.now() - start));
```

The perf delta gets progressively bigger as the input size increases:

```
C:\Perf> ls .\samples\

Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        5/13/2019  11:50 AM        4364428 long.js
-a----        5/13/2019  11:45 AM        1454808 medium.js
-a----        5/13/2019  11:45 AM         290960 short.js

C:\Perf> npm run parser_perf .\samples\short.js
parse time with no TS 162
parse time with TS 221
C:\Perf> npm run parser_perf .\samples\medium.js
parse time with no TS 413
parse time with TS 1070
C:\Perf> npm run parser_perf .\samples\long.js
parse time with no TS 771
parse time with TS 6617
```

**Environment**
- Babel version(s): @babel/parser v7.4.4
- Node/npm version: Node v10.15.3

**Possible Solution**
We need to avoid using exceptions for control flow. Ideally, we would do some refactoring to completely remove the `tsTryParseAndCatch` from the typescript plugin. In the short term, adding a few conditionals to limit its use will probably be sufficient.

matthewrobertson:
| Fixed Issues?            | Fixes #9972 

This commit optimizes the performance of the typescript parser plugin by avoiding exceptions for control flow. Currently, the typescript parser is filled with code branches that look something like this:

```js
const state = this.state.clone();
try {
  this.parseSomething();
} catch (e) {
  if (!(e instanceof SyntaxError)) {
    // something is actually wrong
    throw e;
  }
  // we are just bailing out of a lookahead branch
  this.state = state;
}
```

Throwing exceptions in node is expensive so this commit adds a few conditionals to avoid throwing where possible. The most common issue was related to type argument parsing, which had many false positives due to less-than binary expressions (eg `const y = x <` could be `const y = x <z>();` but it much more likely to be `const y = x < 10;`), so I added  a new `tsTryParseTypeArguments` that resets state and returns `undefined` instead of throwing.

### DIFF
--- a/packages/babel-parser/src/plugins/typescript/index.js
+++ b/packages/babel-parser/src/plugins/typescript/index.js
@@ -1442,6 +1442,9 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       startPos: number,
       startLoc: Position,
     ): ?N.ArrowFunctionExpression {
+      if (!this.isRelational("<")) {
+        return undefined;
+      }
       const res: ?N.ArrowFunctionExpression = this.tsTryParseAndCatch(() => {
         const node: N.ArrowFunctionExpression = this.startNodeAt(
           startPos,
@@ -1464,6 +1467,36 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         /* params are already set */ null,
         /* async */ true,
       );
+    }
+
+    tsTryParseTypeArguments(): ?N.TsTypeParameterInstantiation {
+      if (!this.isRelational("<")) {
+        return undefined;
+      }
+      const state = this.state.clone();
+      const node = this.startNode();
+      const params = this.tsInType(() =>
+        // Temporarily remove a JSX parsing context, which makes us scan different tokens.
+        this.tsInNoContext(() => {
+          if (this.eatRelational("<")) {
+            return this.tsTryParseDelimitedList(
+              "TypeParametersOrArguments",
+              this.tsParseType.bind(this),
+            );
+          }
+        }),
+      );
+      if (params) {
+        node.params = params;
+        // This reads the next token after the `>` too, so do this in the enclosing context.
+        // But be sure not to parse a regex in the jsx expression `<C<number> />`, so set exprAllowed = false
+        this.state.exprAllowed = false;
+        if (this.eatRelational(">")) {
+          return this.finishNode(node, "TSTypeParameterInstantiation");
+        }
+      }
+      this.state = state;
+      return undefined;
     }
 
     tsParseTypeArguments(): N.TsTypeParameterInstantiation {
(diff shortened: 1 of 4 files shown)


packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -1466,6 +1469,36 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+    tsTryParseTypeArguments(): ?N.TsTypeParameterInstantiation {
+      if (!this.isRelational("<")) {
+        return undefined;
+      }
+      const state = this.state.clone();
+      const node = this.startNode();
+      const params = this.tsInType(() =>
+        // Temporarily remove a JSX parsing context, which makes us scan different tokens.
+        this.tsInNoContext(() => {
+          if (this.eatRelational("<")) {
```
nicolo-ribaudo: Nit: we have already checked (at line 1473) that this is `<`

```suggestion
          this.next(); // <
```

matthewrobertson: done

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -1466,6 +1469,35 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+        // But be sure not to parse a regex in the jsx expression `<C<number> />`, so set exprAllowed = false
+        this.state.exprAllowed = false;
+        if (this.eatRelational(">")) {
+          return this.finishNode(node, "TSTypeParameterInstantiation");
+        }
+      }
+      this.state = state;
+      return undefined;
+    }
+
```
danez: This is not nice, as we have this code now twice. But I see why you have done this.

What you should consider here is that we now clone the state twice. Once here and once probably in `tsTryParseDelimitedList`. Cloning the state is also something that is more on the slow side, but not sure how it compares to throwing exceptions.

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -1624,27 +1656,28 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
-            } else if (this.match(tt.backQuote)) {
-              return this.parseTaggedTemplateExpression(
-                startPos,
-                startLoc,
-                base,
-                state,
-                typeArguments,
-              );
-            }
+          const typeArguments = this.tsTryParseTypeArguments();
```
danez: Wouldn't this add an additional try catch?

danez: Ok my mistake, I should have looked first what `tsTryParseDelimitedList` is doing. I assumed everything with a try in the name uses a tray catch.

nicolo-ribaudo: Maybe we should rename things which don't use `try/catch` to `tsMaybeParse*`

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -1466,6 +1469,35 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+      );
+      if (params) {
+        node.params = params;
+        // This reads the next token after the `>` too, so do this in the enclosing context.
+        // But be sure not to parse a regex in the jsx expression `<C<number> />`, so set exprAllowed = false
+        this.state.exprAllowed = false;
+        if (this.eatRelational(">")) {
+          return this.finishNode(node, "TSTypeParameterInstantiation");
+        }
+      }
```
nicolo-ribaudo: Can lines 1478-1496 be shared with the `tsParseTypeArguments` method?

matthewrobertson:
Now that #9989 has landed the optimizations here are not as urgent, so I have updated it to remove the `tsTryParseTypeArguments` and left only the simpler optimizations where we could avoid exceptions in cases where the current state would always throw. 

If there is interest, I can change `tsParseTypeArguments` to an implementation that never throws, but I think we should do that in a separate PR and completely replace the existing method instead of adding a duplicate implementation (as was the case originally in the first version of this PR).

PR #9982, opened 2019-05-15
Avoid unnecessary work during lookahead

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

We were create a comment object but not doing anything with it. Then I noticed that even calling `pushComment()` is unnecessary and this way we avoid some slicing on the input.

Also we were calling `updateContext()` during lookahead which potentially could add a new context to the state, but as we do not clone arrays in state during lookahead this would mess up the state.
And we always only check the `type` or the `value` on the lookahead state.

And because we do no call `updateContext()` anymore I don't think we need to clone `this.state.context` every time, because it won't be change during lookahead.

I did not see any noticeable performance change, but we are doing `lookahead()` mostly in flow/ts plugins which I haven't tested.

### DIFF
--- a/packages/babel-parser/src/tokenizer/index.js
+++ b/packages/babel-parser/src/tokenizer/index.js
@@ -356,7 +356,7 @@ export default class Tokenizer extends LocationParser {
     this.state.type = type;
     this.state.value = val;
 
-    this.updateContext(prevType);
+    if (!this.isLookahead) this.updateContext(prevType);
   }
 
   // ### Token reading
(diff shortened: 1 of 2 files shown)


danez:
This is slightly faster with the fixture doing 30 dynamic imports (because I know there was one lookahead there). #9983 is not included.

```
❯ node --expose-gc --max-old-space-size=2000 --max-semi-space-size=1500 --noconcurrent_sweeping index.js
Node: v12.2.0
Running benchmark for es6/import.js ...
┌───────────────┬───────────────────────────────┬───────────────────────────────┐
│ fixture       │ babel                         │ babelNew                      │
├───────────────┼───────────────────────────────┼───────────────────────────────┤
│ es6/import.js │ 6100 ops/sec ±0.69% (0.164ms) │ 6358 ops/sec ±0.34% (0.157ms) │
└───────────────┴───────────────────────────────┴───────────────────────────────┘
```

PR #9985, opened 2019-05-15
Do not use lookahead when parsing declare module or declare module.exports in flow

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/src/plugins/flow.js
+++ b/packages/babel-parser/src/plugins/flow.js
@@ -240,12 +240,13 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       } else if (this.match(tt._var)) {
         return this.flowParseDeclareVariable(node);
       } else if (this.isContextual("module")) {
-        if (this.lookahead().type === tt.dot) {
+        this.next();
+        if (this.match(tt.dot)) {
           return this.flowParseDeclareModuleExports(node);
         } else {
           if (insideModule) {
             this.unexpected(
-              null,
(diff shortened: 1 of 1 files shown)


packages/babel-parser/src/plugins/flow.js
```diff
@@ -422,8 +421,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
     flowParseDeclareModuleExports(
       node: N.FlowDeclareModuleExports,
     ): N.FlowDeclareModuleExports {
-      this.expectContextual("module");
-      this.expect(tt.dot);
```
xtuc: I might be wrong, but this was eating two tokens? Now it only skip one.

danez: right the other is skipped outside.

packages/babel-parser/src/plugins/flow.js
```diff
@@ -240,12 +240,13 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       } else if (this.match(tt._var)) {
         return this.flowParseDeclareVariable(node);
       } else if (this.isContextual("module")) {
-        if (this.lookahead().type === tt.dot) {
+        this.next();
```
nicolo-ribaudo: Nit: you can use `this.eatContextual("module")` in the if condition

danez: done

PR #9994, opened 2019-05-17
Adding "[object NodeList]" === Object.prototype.toString.call(iter) t…

### LINKED ISSUE #9993: In IE11, NodeLists cannot be spread.
## Bug Report
Cause of Issue is known, will do a pull request.

**Current Behavior**
In IE11, NodeLists cannot be spread.
"Invalid attempt to spread non-iterable instance"

**Input Code**
```js
[...document.querySelectorAll('.filter-item')]
```

**Expected behavior/code**
The NodeList is turned into an Array.

```js
{
  "presets": ["@babel/preset-react", "@babel/preset-env"]
}
```
```js
{
"devDependencies": {
    "@babel/core": "^7.2.2",
    "@babel/polyfill": "^7.2.5",
    "@babel/preset-env": "^7.2.3",
    "@babel/preset-react": "^7.0.0",
    "babel-core": "^7.0.0-bridge.0",
    "babel-jest": "^23.6.0",
    "babel-loader": "^8.0.5",
    "babel-preset-vue": "^2.0.2"
}
"dependencies": {
    "babel-polyfill": "^6.26.0"
}
}
```

**Environment**
- Babel version(s): see above
- Node/npm version: node v10.14.1 / npm 6.4.1
- OS: Ubuntu 18.04
- Monorepo: no

**Possible Solution**
In _toConsumableArray, add `"[object NodeList]" === Object.prototype.toString.call(iter)` additionally to `"[object Arguments]" === Object.prototype.toString.call(iter)`. In Chrome, the method will return `[object Arguments]`, but in IE11 (Windows 8.1) it will return `[object NodeList]`.
_Edit: removed redundant quotes_

ZainlessBrombie:
```
Adding "[object NodeList]" === Object.prototype.toString.call(iter) to check if object is eligible for Array.from() call for _toConsumableArray(arr). In IE11, the object returns the former string rather than "[object Arguments]" as already covered.
```

| Fixed Issues?            | Fixes #9993 

When using the spread operator on an object to use as an array, the generated code passes the object to `_toConsumableArray()`. The function _toConsumableArray then decides how to handle the object, passing it to helper functions like `_iterableToArray(iter)`. One of the checks relies on `Object.prototype.toString.call(iter) === "[object Arguments]"`, which works in chrome. However in IE11 in Windows 8.1, `[object NodeList]` is returned instead and an exception is thrown, despite the object being processable by Array.from (or the replacement thereof). Since NodeLists are processable by Array.from, they should also be added to _iterableToArray.

edit: commit message put in code block

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -862,7 +862,8 @@ helpers.iterableToArray = helper("7.0.0-beta.0")`
   export default function _iterableToArray(iter) {
     if (
       Symbol.iterator in Object(iter) ||
-      Object.prototype.toString.call(iter) === "[object Arguments]"
+      Object.prototype.toString.call(iter) === "[object Arguments]" ||
+      Object.prototype.toString.call(iter) === "[object NodeList]"
     ) return Array.from(iter);
   }
 `;
--- a/packages/babel-runtime-corejs2/helpers/esm/iterableToArray.js
+++ b/packages/babel-runtime-corejs2/helpers/esm/iterableToArray.js
@@ -1,5 +1,5 @@
 import _Array$from from "../../core-js/array/from";
 import _isIterable from "../../core-js/is-iterable";
 export default function _iterableToArray(iter) {
-  if (_isIterable(Object(iter)) || Object.prototype.toString.call(iter) === "[object Arguments]") return _Array$from(iter);
+  if (_isIterable(Object(iter)) || Object.prototype.toString.call(iter) === "[object Arguments]" || Object.prototype.toString.call(iter) === "[object NodeList]") return _Array$from(iter);
 }
--- a/packages/babel-runtime-corejs2/helpers/iterableToArray.js
+++ b/packages/babel-runtime-corejs2/helpers/iterableToArray.js
@@ -3,7 +3,7 @@ var _Array$from = require("../core-js/array/from");
 var _isIterable = require("../core-js/is-iterable");
 
 function _iterableToArray(iter) {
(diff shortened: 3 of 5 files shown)


ZainlessBrombie:
Optimally, the object should have contained Symbol.iterable. I am still investigating as to why the Symbols mismatch.

ZainlessBrombie:
I have found the issue: The project is a vue project. Someone included the default cdn-version of google recaptcha so they could use it in vue instead of using the vue variant. The default version has its own polyfill and overwrites the main polyfill eventually. This is causing the iterator to be using a different key. I don't know why recaptcha overwrites without warning or even has global namespace pollution, but that is another matter entirely. The proposed "fix" here _may_ be included for robustnesses sake, but if that is necessary is questionable as other methods would surely have been affected as well, leaving more potential errors lingering.
TLDR; A script tag referencing google recaptcha overwrote the iterator. _Might_ want to merge the pull request for robustness, but it would only come to use in scenarios in which other components would likely be affected too.

loganfsmyth:
Unfortunately I'm not sure this is something we'd want to land. There are many types that could be iterable but aren't without the polyfill, and making them work is something we leave up to polyfills because it would be unrealistic to support them otherwise, and special-casing `NodeList` wouldn't set a precedent that I don't think we'd want to encourage.

ZainlessBrombie:
@loganfsmyth I absolutely agree! :)
Just wanted to offer the possibility of doing that, but it seems like a step sideways when we want to go forward, if that makes sense. No damage, but we shouldn't keep doing it so lets not start. Anyways, thank you for helping me out by pointing me to the code file, have a wonderful day!

PR #9998, opened 2019-05-19
Fix location for optional params in arrow functions

### LINKED ISSUE #9957: leadingComments drops last line comment before expression
## Bug Report

**Current Behavior**
`leadingComments` is mising last line comment before expression when expression contains parens.

**Input Code**
https://astexplorer.net/#/gist/ff2da32a9fe6bf725c83b66adadc148d/6aa7b8c2421146443f11eaaa5937c5ff990a61fa

```js
let fun = () => {
    // one
    // two
    // three
    return (1);
}
```

Current AST (@babel/parser 7.3.3+)
```
{
  "type": "ReturnStatement",
  ...
  "leadingComments": [
    {
      "type": "CommentLine",
      "value": " one",
      ...
    },
    {
      "type": "CommentLine",
      "value": " two",
      ...
    }
  ]
}
```

**Expected behavior/code**
Expect that the `leadingComments` should have three comments, as there are three in the source code. @babel/parser 7.3.2 is correct and outputs the following 
```
{
  "type": "ReturnStatement",
  ...
  "leadingComments": [
    {
      "type": "CommentLine",
      "value": " one",
      ...
    },
    {
      "type": "CommentLine",
      "value": " two",
      ...
    },
    {
      "type": "CommentLine",
      "value": " three",
      ...
    }
  ]
}
```

Repros on AST Explorer which may be a more standard environment than I'm running.

```js
{
  "dependencies": {
    "@babel/generator": "7.4.4",
    "@babel/parser": "7.4.4",
    "@babel/template": "7.4.4",
    "@babel/traverse": "7.4.4",
    "@babel/types": "7.4.4",
  }
}
```

```ts
import { parse } from '@babel/parser';

parse(sourceText, {
    sourceType: 'module',
    strictMode: false,
    plugins: ['classProperties', 'typescript', 'decorators-legacy'],
});
```

**Environment**
- Babel version(s): @babel/parser 7.3.3+ (**Regression** since @babel/parser 7.3.2)
- Node/npm version: 10.15.1
- OS: Windows 10
- Monorepo: no
- How you are using Babel: plugin, or AST explorer repros the issue too (link above)

This only seems to repro when the statement with leading comments has parens.

#### Example 1
The below works just fine
 ```js
let fun = () => {
    // one
    // two
    // three
    return 1;
}
```
The below drops comment `// three`
 ```js
let fun = () => {
    // one
    // two
    // three
    return (1);
}
```

#### Example 2
The below works just fine
 ```js
let fun = () => {
    // one
    // two
    // three
    if (!condition) {
       ...
    }
}
```
The below drops comment `// three`
 ```js
let fun = () => {
    // one
    // two
    // three
    if (!(condition1)) {
       ...
    }
}
```

danez:
| Fixed Issues?            | Fixes #9957
| Documentation PR Link    | 
| Any Dependency Changes?  |

This ensures we only add +1 to the location when there is an optional parameter. In TS we currently were calling `this.finishNode()` even if no question mark was present, which caused comments being processed twice and as a result comments were disappearing. Not sure why though, I didn't want to dive into the comments attachment algorithm.

It also made certain expressions locations wrong.

```js
((-10)) as const;
```

The numeric literal would include one of the closing parenthesis in its location.

I also made sure that the flow plugin does the same (include the question mark) as the typescript plugin.

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -15,6 +15,7 @@ const merge = require("merge-stream");
 const rollup = require("rollup");
 const rollupBabel = require("rollup-plugin-babel");
 const rollupNodeResolve = require("rollup-plugin-node-resolve");
+const rollupReplace = require("rollup-plugin-replace");
 const { registerStandalonePackageTask } = require("./scripts/gulp-tasks");
 
 const sources = ["codemods", "packages"];
@@ -92,6 +93,9 @@ function buildRollup(packages) {
         .rollup({
           input,
           plugins: [
+            rollupReplace({
+              "process.env.NODE_ENV": JSON.stringify(process.env.NODE_ENV),
+            }),
             rollupBabel({
               envName: "babel-parser",
             }),
@@ -103,6 +107,7 @@ function buildRollup(packages) {
             file: path.join(pkg, "lib/index.js"),
             format: "cjs",
             name: "babel-parser",
+            sourcemap: process.env.NODE_ENV !== "production",
           });
         });
     })
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,7 @@ prepublish-build:
 	rm -rf packages/babel-runtime/helpers
 	rm -rf packages/babel-runtime-corejs2/helpers
 	rm -rf packages/babel-runtime-corejs2/core-js
-	BABEL_ENV=production make build-dist
+	NODE_ENV=production BABEL_ENV=production make build-dist
 	make clone-license
 
 prepublish:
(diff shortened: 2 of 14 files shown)


danez:
I have no idea why the preset-env tests suddenly fail. It works find here. Do we have any dependencies that are not locked?

zloirock:
@danez `core-js@3.1` contains some new modules, so need to rebuild `preset-env` fixures.

zloirock:
Also, seems the replacing of `esnext.string.match-all` (and living it as a stub only) to `es.string.match-all` in some entry points broke `core-js-compat/entries` autogeneration. Need to fix it in `core-js@3.1.1`.

PR #10012, opened 2019-05-22
Optional chaining optimization

### LINKED ISSUE #10004: Optional chaining optimization
## Feature Request

**Is your feature request related to a problem?**

Current implementation of `@babel/plugin-proposal-optional-chaining` works like this:
- Input:
```js
const street = user?.address?.street;
```
- Output:
```js
var _user, _user$address;
const street = (_user = user) === null || _user === void 0
    ? void 0
    : (_user$address = _user.address) === null || _user$address === void 0
        ? void 0
        : _user$address.street;
```

Chaining ternary operators works fine, but theres a certain use case where it can be optimised:
- Input:
```js
if (user?.address?.street) {
    console.log('Street:', user.address.street);
}
```
- Output:
```js
var _user, _user$address;
if ((_user = user) === null || _user === void 0 ? void 0 : (_user$address = _user.address) === null || _user$address === void 0 ? void 0 : _user$address.street) {
  console.log('Street:', user.address.street);
}
```

Inside an `if` clause it only needs to check if the result is falsy or not. Transforming could be more space-efficient like this:
```js
if (user && user.address && user.address.street) {
    console.log('Street:', user.address.street);
}
```

I would only apply this optimisation for specific scenarios (`if` and `while`), and in case there's no additional operator/call:
```js
// Bail out. ternary chain works better here
if (user?.address?.street === 'some street') { ... }
```

Make this optimisation optional at start, then make it a default, keeping a way to opt-out:
```json
{
    "plugins": [
        ["@babel/plugin-proposal-nullish-coalescing-operator", {
            "fasly-expression-optimisation: true/false
        }],
    ]
}
```

tanhauhau:
| Fixed Issues?            | Fixes #10004
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-proposal-optional-chaining/src/index.js
+++ b/packages/babel-plugin-proposal-optional-chaining/src/index.js
@@ -5,7 +5,7 @@ import { types as t } from "@babel/core";
 export default declare((api, options) => {
   api.assertVersion(7);
 
-  const { loose = false } = options;
+  const { loose = false, falsyExpressionOptimisation = false } = options;
 
   return {
     name: "proposal-optional-chaining",
@@ -39,6 +39,23 @@ export default declare((api, options) => {
         if (parentPath.isUnaryExpression({ operator: "delete" })) {
           replacementPath = parentPath;
         }
+
+        if (
+          falsyExpressionOptimisation &&
+          loose &&
+          parentPath.isIfStatement()
+        ) {
+          if (hasNoCallExpressionInChain(path)) {
+            let right = path.node;
+            for (let i = 0; i < optionals.length; i++) {
+              const node = optionals[i];
+              right = t.logicalExpression("&&", node.object, right);
+            }
+            replacementPath.replaceWith(right);
+            return;
+          }
+        }
+
         for (let i = optionals.length - 1; i >= 0; i--) {
           const node = optionals[i];
 
@@ -122,3 +139,10 @@ export default declare((api, options) => {
     },
   };
 });
+
+function hasNoCallExpressionInChain(path) {
+  while (path.isMemberExpression()) {
+    path = path.get("object");
+  }
+  return !path.isCallExpression();
+}
--- a/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/if-optimisation-loose/input.js
+++ b/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/if-optimisation-loose/input.js
@@ -1,0 +1,16 @@
+if (a?.b?.c) {
+  console.log(a?.b?.c);
+} else if (a?.b.c?.d?.e.f) {
+  console.log(a?.b.c?.d?.e.f);
+}
+
+if (a?.b?.c === 'foobar') {
+}
+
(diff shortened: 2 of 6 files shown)


packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/if-optimisation/input.js
```diff
@@ -0,0 +1,16 @@
...
+if (a?.b?.c === 'foobar') {
+}
+
+if (a?.b()?.c) {
+  
+}
+
+if (a?.b?.()?.c) {
+
+}
```
hg-pyun: Need to EOF

packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/if-optimisation-loose/input.js
```diff
@@ -0,0 +1,16 @@
...
+if (a?.b?.c === 'foobar') {
+}
+
+if (a?.b()?.c) {
+  
+}
+
+if (a?.b?.()?.c) {
+
+}
```
hg-pyun: Need to EOF

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -39,6 +39,23 @@ export default declare((api, options) => {
         if (parentPath.isUnaryExpression({ operator: "delete" })) {
           replacementPath = parentPath;
         }
+
+        if (
+          falsyExpressionOptimisation &&
+          loose &&
+          parentPath.isIfStatement()
```
nicolo-ribaudo: Why only for if statements?

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -39,6 +39,23 @@ export default declare((api, options) => {
...
+        ) {
+          if (hasNoCallExpressionInChain(path)) {
```
nicolo-ribaudo: Why should we bail out in case of a function call? This seems ok:

```js
a?.b.c() 

// -> 

a && a.b.c()
```

tanhauhau: 🤔 so the function call can only happened at the end right? because middle of the chain could be problematic:

```
a?.b()?.c

// ->

a && a.b() && a.b().c
```

packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/if-optimisation/output.js
```diff
@@ -0,0 +1,17 @@
+var _a3, _a3$b, _a4, _a4$b, _a5, _a5$b;
+
+if (a && a.b && a.b.c) {
```
jridgewell: I don't think this is a worthwhile optimization, this is exactly the bug we're trying to avoid with the nullish check!

PR #10017, opened 2019-05-23
destructuring private fields with array pattern / object pattern

### LINKED ISSUE #9310: bug: exception with destructuring assignment with private fields
## Bug Report

**Current Behavior**
`Invalid destructuring assignment target` exception with destructuring assignment with private fields

**Input Code**
- [REPL](https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEBiD29oG8BQ1oGJQEsCmAdgC6rrTDwEREBOArsEfDQBQAON8bEAlCmRhbJyIfMQBc0IgAscEAHTZRhItAC-0ALzQOXXmTWpDQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=script&lineWrap=true&presets=stage-3&prettier=false&targets=&version=7.2.2&envVersion=)

```js
class Foo {
  #client

  constructor(props) {
    ;({ client: this.#client } = props) // throws: Invalid destructuring assignment target
  }
}
```

transpiles to _(excerpt)_:
```js
({
  client: _classPrivateFieldGet(this, _client)  // notice the getter
} = props);
```

**Environment**
- Babel version: **v7.2.2**

_update_:

I suppose it would need to use **_classPrivateFieldSet** as well as transpile the ES2015 destructuring assignment? unless destructuring assignments work with a function call ... which I don't think it does?

```js
var client = props.client
_classPrivateFieldSet(this, _client, client);
```

for what it's worth: works in `node --harmony` **v11.6.0**

tanhauhau:
| Fixed Issues?            | Fixes #9310 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


tanhauhau:
I guess I need to make sure it works on loose mode as well.

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -100,6 +100,26 @@ const handle = {
       return;
     }
 
+    // { KEY: MEMBER } = OBJ -> { KEY: _ref } = OBJ; _set(MEMBER, _ref);
+    // [MEMBER] = ARR -> [_ref] = ARR; _set(MEMBER, _ref);
+    if (
```
jridgewell: If I remember correct, this will affect `super.foo` expressions, too. Can you add a test for assigning `[super.foo] = array`, too?

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -100,6 +100,26 @@ const handle = {
...
+      (parentPath.isObjectProperty({ value: node }) &&
+        parentPath.parentPath.isObjectPattern()) ||
+      parentPath.isArrayPattern()
+    ) {
+      const { scope } = member;
+      const ref = scope.generateUidIdentifierBasedOnNode(node);
+      scope.push({ id: ref });
+
+      const assignmentParent = parentPath.getStatementParent();
+      assignmentParent.insertAfter(
```
jridgewell: I'm not sure we can insert after the statement, because later patterns can default to the property we're assigning:

```js
[
  this.#foo,
  x = this.#foo
] = [0];
```

Should have `x === 0`. The only way I can see this working in 100% of cases is if we use a special setter property:

```js
// input
[this.#foo, x = this.#foo] = [0];

// output
const _this = this;
const internal = { set foo(v) { _this.#foo = v } };

[internal.foo, x = this.#foo] = [0]
```

tanhauhau: so the order matters for array pattern right?
eg:
```
[this.#foo, x = this.#foo] = [0]
```
vs
```
[x = this.#foo, this.#foo] = [0];
```

how about object destructuring?
```
{
   x = this.#foo,
   y: this.#foo,
   z = this.#foo,
} = { y: 1 }
```

jridgewell: Yup, you're correct in both cases.

tanhauhau: would it be better if we defined `const internal = { set foo(v) { _this.#foo = v } };` at the constructor if we know there will be a destructuring, and reuse it throughout the scope of the class?

jridgewell: Yah, I would only create the object when we hit a destructure.

nicolo-ribaudo:
Depending on the order things are evaluated (need to check), we might be able to out things in the next property's key:
```js
{ x: this.#foo, y = this.#foo } = {}

// ->

{ x = tmp, [this.#foo = tmp, "y"] = this.#foo } = {}
```

Also, when we are destructing to member expressions instead of plain identifiers, this might work:
```js
[ this.#foo, foo.bar = this.#foo ] = [];

// ->

[ tmp, (this.#foo = tmp, foo).bar ] = []
```

I'm not sure about those optimizations, but I'd like to avoid using the setter when possible.

jridgewell:
I'm not sure I understand your input/output. Where does the array destructure come from in `{ x = tmp, [this.#foo = tmp, "y"] = this.#foo } = {}`?

There can also be indirection between a default value and the private through a getter:

```js
class Ex {
  get() {
    return this.#foo;
  },

  destructure(arr) {
    [ this.#foo, bar = this.get() ] = arr;
  }
}
```

Or maybe the private is a setter that mutates some other state.

nicolo-ribaudo:
It's a computed key.
I have a more efficient idea in mind anyway, I will post it soon

tanhauhau:
> ```js
> { x: tmp, [this.#foo = tmp, "y"] = this.#foo } = {}
> ```
wow, @nicolo-ribaudo this is cool, never thought of able to use comma operator this way.

tanhauhau:
> There can also be indirection between a default value and the private through a getter:

so you are saying, if there's a default value, you can't eliminate the possibility of reading from the updated private variable right

nicolo-ribaudo:
Currently this code:
<details>

```js
class A {
  #a;
  #b() {}
  get #c() {}
  set #d(a) {}
  
  test() {
    this.#a = 1;
    this.#b = 2;
    this.#c = 3;
    this.#d = 4;
  }
}
```
</details>

Is transpiled to this:
<details>

```js
class A {
  constructor() {
    _d.set(this, { set: _set_d });
    _c.set(this, { get: _get_c });
    _b.add(this);
    _a.set( writable: true, value: void });
  }

  test() {
    _classPrivateFieldSet(this, _a, 1);
    _classPrivateMethodSet();
    _classPrivateMethodSet();
    _classPrivateFieldSet(this, _d, 4);
  }

}
```
</details>

When using destruturing:
```js
class A {
  #a;
  #b() {}
  get #c() {}
  set #d(a) {}
  
  test() {
    [this.#a, this.#b, this.#c, this.#d] = [1, 2, 3, 4];
  }
}
```

It can be transpiled like this:
```js
class A {
  constructor() {
    _d.set(this, { set: _set_d });
    _c.set(this, { get: _get_c });
    _b.add(this);
    _a.set( writable: true, value: void });
  }

  test() {
    [
      _classPrivateFieldDestrSet(this, _a).value,
      _classPrivateMethodSet(),
      _classPrivateMethodSet(),
      _classPrivateFieldDestrSet(this, _d).value,
     ] = [1, 2, 3, 4];
  }

}
```

Where the `_classPrivateFieldDestrSet` helper is defined like this:
```js
export default function _classPrivateFieldDestrSet(receiver, privateMap) {
  if (!privateMap.has(receiver)) {
    throw new TypeError("attempted to set private field on non-instance");
  }
  var descriptor = privateMap.get(receiver);

  if ("value" in descriptor) return descriptor;
  if (!("__destrObj" in descriptor)) {
    descriptor.__destrObj = Object.defineProperty({}, "value", descriptor);
  }
  return descriptor.__destrObj;
}
```

jridgewell:
Ha, I couldn't parse the computed key when it was inline with the other keys... 😅 That's an obscure desugaring, but I guess it'll work for object patterns.

But array destructures can't use computed keys. I don't see a way to make this work in every case unless we use the setter objects. And if we have to maintain two ways to do it, why not just choose works-with-everything approach? Hopefully this will be rare enough to not matter (I didn't even realize you could destructure to a `MemberExpression` before this PR).

nicolo-ribaudo:
Yeah, in the last comment I posted another example which doesn't rely on computed keys (since I'm not even sure about the order in which they are evaluated).
The advantage of the helper I proposed is that instead of creating an intermediate object with a setter when destructuring to private fields, it reuses the object descriptor we already have. In the other case (setters), it creates an object reusing the descriptor we already have, and it creates it only once per key/instance pair.

jridgewell:
Sweet, just read the other comment. Let's use the private map's descriptor.

Though, I don't understand the `__destrObj` property in the helper file.

nicolo-ribaudo:
> Though, I don't understand the __destrObj property in the helper file.

In case of a setter, the decriptor looks like this:
```js
const desc = {
  get: _fn_get,
  set: _fn_set,
}
```
We can't just return the descriptor, because assigning to `.value` (or `.set`) wouldn't invoke the setter. For this reason, we need to actually create an object with a setter:
```js
const objWithSetter = {
  value: // This has the "descriptor" descriptor.
}
```
I stored it in `desc.__destrObj` only to avoid recreating it every time the helper is called on a given key (which probably would be called once for every invokaton of the class method which does the destructuring). Anyway, if you think that it just adds confusion this optimization can probably be removed, since destructuring to a private getter is a really rare edge case.

jridgewell:
Ooo, clever. But the `receiver` won't be correct during the set, we would have to bind it. Or we can just create the reusable internal object without reusing the descriptor.

The helper approach is definitely simpler, I like it the most.

```js
export default function _classPrivateFieldDestructureSet(receiver, privateMap) {
  if (!privateMap.has(receiver)) {
    throw new TypeError("attempted to set private field on non-instance");
  }
  var descriptor = privateMap.get(receiver);

  if ("value" in descriptor) return descriptor;
  
  if (!("__destrObj" in descriptor)) {
    descriptor.__destrObj = {
      set value(v) {
        descriptor.set.call(receiver, v)
      },
    };
  }
  return descriptor.__destrObj;
}

**Comment by nicolo-ribaudo** [comment]
Good catch, you're right!

**Comment by tanhauhau** [comment]
cool let me work on it!

File: packages/babel-helper-member-expression-to-functions/src/index.js (hunk 1 above)
Comment (tanhauhau) [reply packages/babel-helper-member-expression-to-functions/src/index.js]: where do you think should i add the test for `[super.foo] = array` ?

in `babel-plugin-transform-classes/test/get-set` ?

**Comment by tanhauhau** [comment]
@nicolo-ribaudo i noticed that the define descriptor could be non `writeable`, but is there a test case I can write that can test this?

**Comment by tanhauhau** [comment]
@jridgewell  Ok, I've read through the `set` helper for assigning super fields, I am not sure I should implement another helper that is similar to `_classPrivateFieldDestructureSet` above, because:
1. I would have somewhat duplicate logic in both `helpers.set` and `helpers.destructureSet`, and would need to retest all the edge cases
2. I am not entirely sure where to keep the `__destrObj` in this case, I certainly can't modify the descriptor object returned from `Object.getOwnPropertyDescriptor`.

So, I am thinking maybe I can 
1. call setter right after destructuring if there's no getter within the `ArrayPattern` statement:

```js
class A {
   foo() {
       ([super.bar] = [1]);
   }
}

// transpiled to

class A {
  foo() {
      ([_ref] = [1]);
      super.bar = _ref;
  }
}
```

2. use a temp `destructObject` if there's a getter:
```js
class A {
   foo() {
       ([super.bar, b = super.bar] = [1, 2]);
   }
}

// transpiled to

class A {
   foo() {
       const _this = this;
       const temp = {
           set value(_val) {
                babelHelpers.set(
                     babelHelpers.getPrototypeOf(_obj), 
                     "bar",
                     _val,
                     _this,
                     true
                )
           }
       };
       ([temp.value, b = super.bar] = [1, 2]);
   }
}
```

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/destructuring-array-pattern-1/output.js
```diff
@@ -0,0 +1,14 @@
+var Foo = function Foo(props) {
+  "use strict";
+
+  babelHelpers.classCallCheck(this, Foo);
+  Object.defineProperty(this, _client, {
+    writable: true,
+    value: void 0
+  });
+  babelHelpers.classPrivateFieldLooseBase(this, _client)[_client] = 1;
+  ;
```
nicolo-ribaudo: Where does this extra semicolon come from?

jridgewell:
We’ll have to discuss how this interacts with super properties, for now it’d be fine to just throw an error in the transformer.

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/destructuring-array-pattern-1/output.js
tanhauhau: oh, it's due to an extra comma in the `input.js`

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -100,6 +100,20 @@ const handle = {
...
 
+    // { KEY: MEMBER } = OBJ -> { KEY: _destructureSet(MEMBER) } = OBJ
+    // [MEMBER] = ARR -> [_destructureSet(MEMBER)] = ARR
+    // [...MEMBER] -> [..._destructureSet(MEMBER)]
+    // {...MEMBER} -> {..._destructureSet(MEMBER)}
+    if (
+      (parentPath.isObjectProperty({ value: node }) &&
+        parentPath.parentPath.isObjectPattern()) ||
+      parentPath.isArrayPattern() ||
+      parentPath.isRestElement()
```
nicolo-ribaudo: Do we also need to check for `AssignmentPattern`s?

```js
[this.#foo = 2] = []
```

tanhauhau: checked and added more test case

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/destructuring-array-pattern-1/output.js
tanhauhau: removed extra comma

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -100,6 +100,34 @@ const handle = {
...
+    // [MEMBER = _VALUE] = ARR -> [_destructureSet(MEMBER) = _VALUE] = ARR
+    // [...MEMBER] -> [..._destructureSet(MEMBER)]
+    if (
+      // { KEY: MEMBER } = OBJ
+      (parentPath.isObjectProperty({ value: node }) &&
+        parentPath.parentPath.isObjectPattern()) ||
+      // { KEY: MEMBER = _VALUE } = OBJ
+      (parentPath.isAssignmentPattern({ left: node }) &&
+        parentPath.parentPath.isObjectProperty({ value: parent }) &&
+        parentPath.parentPath.parentPath.isObjectPattern()) ||
```
jridgewell: Nit: I think we can actually join a few of the conditions here. `parentPath.isAssignmentPattern()` is sufficient to check for both array pattern and object pattern, we don't need to consult the grand parent.

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -100,6 +100,34 @@ const handle = {
...
+      // [MEMBER] = ARR
+      parentPath.isArrayPattern() ||
+      // [MEMBER = _VALUE] = ARR
+      (parentPath.isAssignmentPattern({ left: node }) &&
+        parentPath.parentPath.isArrayPattern()) ||
+      // {...MEMBER}
+      // [...MEMBER]
```
jridgewell: Man, some of these syntaxes are crazy.

packages/babel-helper-member-expression-to-functions/src/index.js
nicolo-ribaudo: We can move these checks to be assertions: (`t.assertObjectProperty(parentPath.parentPath, { value: parent })`

packages/babel-types/src/definitions/es2015.js
```diff
@@ -19,7 +19,13 @@ defineType("AssignmentPattern", {
   fields: {
     ...patternLikeCommon,
     left: {
-      validate: assertNodeType("Identifier", "ObjectPattern", "ArrayPattern"),
+      validate: assertNodeType(
+        "Identifier",
+        "ObjectPattern",
+        "ArrayPattern",
+        "MemberExpression",
+        "CallExpression",
```
jridgewell: How did `CallExpression` get into this?

tanhauhau: @jridgewell 
```js
[_destructureSet(MEMBER) = _VALUE] = ARR
```

packages/babel-helper-member-expression-to-functions/src/index.js
tanhauhau: @nicolo-ribaudo what is the rationale behind using `assertions` instead of `is`?

packages/babel-types/src/definitions/es2015.js
jridgewell: That's invalid, isn't it? Chrome throws a syntax error.

tanhauhau: Yes you are right. my bad. fixed it

packages/babel-helper-member-expression-to-functions/src/index.js
nicolo-ribaudo: Just that with an assertion is clear that if this is an assignment expression, the parent will always be an object/array destructing

nicolo-ribaudo: Oh ok nevermind, we can't assert two different parent types.

PR #10018, opened 2019-05-23
Create SECURITY.md

danez:
Not sure about 6.26.0, but I guess we would fix security issues if reported?

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,19 @@
+# Security Policy
+
+## Supported Versions
+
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
@xtuc We will only release 6.26.x in case of critical security vulnerabilities: we aren't maintaining it with bugfixes/backports.

danez:
The email is setup with mailgun, and forwards to the same ppl as the other team email.

PR #10021, opened 2019-05-24
Change root package.json#name

jamiebuilds:
I'm pretty sure GitHub's new Dependants feature is based on this package.json. I'm updating it here so it will reflect the new name for the core package.

D680031B-947A-4DCB-9160-51C628EFD805

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@babel/core",
+  "name": "babel-core",
   "private": true,
   "license": "MIT",
   "scripts": {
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
It seems that this breaks Jest :thinking:

danez:
I guess we should wait till the support it correctly.

hzoo:
Dang that sucks 😛! How bout we try `babel-core` in the meantime, I guess I'll be a little more accurate (v5 -> v6) although still not v7..

nicolo-ribaudo:
This doesn't make sense 😂

Screenshot_20190605-094152

PR #10029, opened 2019-05-26
Fixed computed keys for class expression

### LINKED ISSUE #8882: Invalid code emitted for computed properties
I am working with other developers to bring private fields into TypeScript. While working on that, @joeywatts discovered [this issue](https://github.com/Microsoft/TypeScript/issues/27864) that currently exists in TypeScript. While comparing TS with Babel, we discovered a bug in Babel's handling of this. Given the input code from the link above, the [Babel transformation](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBMA2BDCECmEYF4YG0C6A3AFABmIATjABTyqwCWWMADATIwDzYCMrMA1P3oBKGAG8iMOEhToAdAAcArhAAWVSVOnIMAQXGatueniYAiALYBPGAvIgFqclCtniRgL6bhxL2UpUoJCwCIgAXmEwICTashCiElJBECC0cvAgAOZUAPIARgBWqMBQcpl0OQDuYAAK9o7OVgByiBboVGColdoRVML9vkA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3&prettier=false&targets=&version=7.0.0-beta.50&envVersion=) produces code with a syntax error.

Input code:

```Javascript
const classes = [];
for (let i = 0; i <= 10; ++i) {
  classes.push(
    class A {
      [i] = "my property";
    }
  );
}
```

Snippet from Babel transformation:

```Javascript
var classes = [];
for (var i = 0; i <= 10; ++i) {
  var _class, _temp;
  classes.push(var _i = i;, (_temp = _class = function A() { // syntax error at `var` and `;`
    _classCallCheck(this, A);
    _defineProperty(this, _i, "my property");
  }, _temp));
}
```

tanhauhau:
| Fixed Issues?            | Fixes #8882
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

- used `let` instead of `var` for declaring computedKeys variable, so that the value is correctly scoped.
- declare the `let` at the top of current block, so that it wont get bailed out when converting the declaration to expression sequence.

### DIFF
--- a/packages/babel-helper-create-class-features-plugin/src/misc.js
+++ b/packages/babel-helper-create-class-features-plugin/src/misc.js
@@ -85,6 +85,9 @@ export function injectInitialization(path, constructor, nodes, renamer) {
 export function extractComputedKeys(ref, path, computedPaths, file) {
   const declarations = [];
 
+  const nearestBlock = path.find(
+    parentPath => parentPath.isBlockStatement() || parentPath.isProgram(),
+  );
   for (const computedPath of computedPaths) {
     computedPath.traverse(classFieldDefinitionEvaluationTDZVisitor, {
       classBinding: path.node.id && path.scope.getBinding(path.node.id.name),
@@ -98,10 +101,14 @@ export function extractComputedKeys(ref, path, computedPaths, file) {
       const ident = path.scope.generateUidIdentifierBasedOnNode(
         computedNode.key,
       );
+      nearestBlock.unshiftContainer(
+        "body",
+        t.variableDeclaration("let", [t.variableDeclarator(ident, null)]),
+      );
       declarations.push(
-        t.variableDeclaration("var", [
-          t.variableDeclarator(ident, computedNode.key),
-        ]),
+        t.expressionStatement(
+          t.assignmentExpression("=", t.cloneNode(ident), computedNode.key),
+        ),
       );
       computedNode.key = t.cloneNode(ident);
     }
--- a/packages/babel-helper-create-class-features-plugin/src/misc.js
+++ b/packages/babel-helper-create-class-features-plugin/src/misc.js
@@ -98,8 +98,8 @@ export function extractComputedKeys(ref, path, computedPaths, file) {
       const ident = path.scope.generateUidIdentifierBasedOnNode(
         computedNode.key,
       );
-      // Declaraing in the same block scope
-      // Fixing https://github.com/babel/babel/pull/10029/files#diff-fbbdd83e7a9c998721c1484529c2ce92
+      // Declaring in the same block scope
+      // Ref: https://github.com/babel/babel/pull/10029/files#diff-fbbdd83e7a9c998721c1484529c2ce92
       path.scope.push({
         id: ident,
         kind: "let",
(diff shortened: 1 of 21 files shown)


packages/babel-helper-create-class-features-plugin/src/misc.js
```diff
@@ -98,10 +101,14 @@ export function extractComputedKeys(ref, path, computedPaths, file) {
       const ident = path.scope.generateUidIdentifierBasedOnNode(
         computedNode.key,
       );
+      nearestBlock.unshiftContainer(
+        "body",
+        t.variableDeclaration("let", [t.variableDeclarator(ident, null)]),
+      );
```
nicolo-ribaudo: Can we use `path.scope.push` instead? So we don't have to re-implement the logic to decide where to put the var.

packages/babel-helper-create-class-features-plugin/src/misc.js
```diff
@@ -85,6 +85,9 @@ export function injectInitialization(path, constructor, nodes, renamer) {
 export function extractComputedKeys(ref, path, computedPaths, file) {
   const declarations = [];
 
+  const nearestBlock = path.find(
+    parentPath => parentPath.isBlockStatement() || parentPath.isProgram(),
```
nicolo-ribaudo: Sometimes the variable needs to be inserted where a block doesn't exist. Do we have a test for this code?

```js
() => class { [k] = 2 }
```

tanhauhau: good catch, added a new test case here (https://github.com/babel/babel/pull/10029/files#diff-fe07c1ba4a5c933d73ca27c26a08737a)

packages/babel-helper-create-class-features-plugin/src/misc.js
tanhauhau: changed

packages/babel-helper-create-class-features-plugin/src/misc.js
```diff
@@ -98,10 +98,14 @@ export function extractComputedKeys(ref, path, computedPaths, file) {
       const ident = path.scope.generateUidIdentifierBasedOnNode(
         computedNode.key,
       );
+      path.scope.push({
+        id: ident,
+        kind: "let",
```
nicolo-ribaudo: I don't see any advantage in using `let` here? Usually we try to generate an output as low-level as possible, if it isn't harder to do.

tanhauhau: It is for this scenario: https://github.com/Microsoft/TypeScript/issues/27864

if it is set as `var`, it will behave the same as pointed out in the issue link above

maybe it should be determined by whether the computedKeys original declaration kind, to preserve the original scoping intention?

nicolo-ribaudo: Oh ok; it's ok like this then.

existentialism: Seems like adding a comment explaining this might be useful?

nicolo-ribaudo:
Thank you!

PR #10034, opened 2019-05-28
Use scope for typescript export removals

### LINKED ISSUE #10033: Typescript export default erased in presence of namespace
## Bug Report

**Current Behavior**

The presence of a `declare namespace` where the name is shared erases the `export default`.

```js
export class N {
  f1(value) {
    value.f2();
  }

}
```

**Input Code**

```typescript
export class N {
    f1(value: N.I) {
        value.f2();
    }
}
export declare namespace N {
    export interface I {
        f2();
    }
}
export default N;
```

**Expected behavior/code**

```js
export class N {
  f1(value) {
    value.f2();
  }

}

export default N;
```

```js
{
  "your": { "config": "here" }
}
```

**Environment**
- Babel version(s): a596da28220b69df6265192d1ca788e6b9ca67aa
- Node/npm version: v8.12.0
- OS: Windows 10
- Monorepo: ?
- How you are using Babel: `cli` fixtures

**Possible Solution**

Same solution.

Independently: https://github.com/babel/babel/pull/10034

Required for: https://github.com/babel/babel/pull/9785

Introduced in https://github.com/babel/babel/pull/10019

Wolvereness:
| Fixed Issues?            | Fixes #10033
| Documentation PR Link    | 

Rewrites the removal of type exports to simply examine what's in the scope. This requires a fix to the `enum` transform so it adds itself to scope.

Exporting an `enum` before it appears is invalid typescript, so whether or not it's in the scope for that visitor is irrelevant.

This also changes a few related test cases to use a comment `; // Otherwise-empty file` so that they will fail-fast if the output changes, rather than quietly (though not silently) create an output file.


nicolo-ribaudo:
@airato  Does this look ok to you?

airato:
👍 thank you!

PR #10047, opened 2019-05-30
Refactor isImportTypeOnly helper function

Andarist:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

I've bumped onto the issue fixed by [this](https://github.com/babel/babel/commit/66be4aa5e3c2cd4c56c924d60eb0a74548b3af86) because I had outdated dependencies, was reading the code and noticed those few small things that could be improved so I've prepared this small PR.

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -17,12 +17,7 @@ function isInType(path) {
   }
 }
 
-interface State {
-  programPath: any;
-}
-
 const PARSED_PARAMS = new WeakSet();
-const PRAGMA_KEY = "@babel/plugin-transform-typescript/jsxPragma";
 
 export default declare((api, { jsxPragma = "React" }) => {
   api.assertVersion(7);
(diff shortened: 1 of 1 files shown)


packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -39,16 +34,15 @@ export default declare((api, { jsxPragma = "React" }) => {
       Identifier: visitPattern,
       RestElement: visitPattern,
 
-      Program(path, state: State) {
-        state.programPath = path;
```
Andarist: this was some leftover, not used anymore

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -39,16 +34,15 @@ export default declare((api, { jsxPragma = "React" }) => {
...
+      Program(path, state) {
         const { file } = state;
+        let fileJsxPragma = null;
 
         if (file.ast.comments) {
           for (const comment of (file.ast.comments: Array<Object>)) {
             const jsxMatches = JSX_ANNOTATION_REGEX.exec(comment.value);
             if (jsxMatches) {
-              file.set(PRAGMA_KEY, jsxMatches[1]);
+              fileJsxPragma = jsxMatches[1];
```
Andarist: Instead of storing this in a shared mutable state under some artificial key I think it's better to just pass it as explicit argument to `isImportTypeOnly`, as it makes the data flow clearer - especially that this info doesn't have to be passed through layers of functions, it's needed just one 1 level lower.

Issue #2471, opened 2015-10-01
Non-simple parameter messes up binding resolution


kangax:
The following snippet:

```js
(function(x, f = () => x) { 
  var x; 
  var y = x; 
  x = 2; 
  return [x, y, f()];
})(1)
```

gives `[2, null, 1]` in Babel. Aside from `x` shadowing behavior (and what `f` should resolve to — 1 or 2 (which Babel and FF disagree on)), `y` definitely shouldn't be `null` here (as far as I can tell), so this looks like a bug?

sebmck:
>  `y` definitely shouldn't be `null` here

It's not. It's actually `undefined`. You're just using the Babel REPL which has historically bad serialisation (/babel/babel.github.io#172).

sebmck:
> Aside from `x` shadowing behavior (and what `f` should resolve to — 1 or 2 (which Babel and FF disagree on))

Parameters create another scope so the Babel behaviour of `[2, undefined, 1]` is almost certainly correct. Firefox isn't following the spec here.

Skalman:
Is the scope really only created when you reference an earlier parameter?
```js
var a = (function(x, y = 1) { 
  var x;
  return x;
})(1); // 1

var b = (function(x, y = x) { 
  var x;
  return x;
})(1); // undefined
```

sebmck:
@Skalman The scope that default parameters are initialised in is separate. We don't always emit the IIFE because it's not necessary unless the default initialisers refer to body bindings.

Closing this issue as I've addressed all the problems found. The `null` issue is a REPL serialisation one and the default parameter behaviour is expected.

kangax:
> It's not. It's actually undefined. You're just using the Babel REPL which has historically bad serialisation (babel/babel.github.io#172).

Ah, got it, thanks. I was suspicious that such an obvious error would slip through :)

Speaking of separate scope for parameters — you're right and I forgot that we even have a test for it in compat table which FF currently fails (https://kangax.github.io/compat-table/es6/#test-default_function_parameters_separate_scope)

Skalman:
> We don't always emit the IIFE because it's not necessary unless the default initialisers refer to body bindings.

@sebmck I should have rephrased - *should* the scope really be created only when an earlier parameter is referenced? In my examples, I only change the definition of `y`, but the value of `x` is changed because of the IIFE.

loganfsmyth:
I'm going to reopen this since from my reading of the spec this should result in `[2, 1, 1]`.

[...]

PR #10053, opened 2019-06-04
Check shadow variable to identifier in default parameters

### LINKED ISSUE #9947: The scope of the arguments default value is different from what was expected in ES2015 (参数默认值的作用域与 ES2015 的预期结果不同)
## Bug Report

**Current Behavior**
The scope of the arguments default value is different from what was expected in ES2015

**Input Code**

[online view](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=DYUwLgBAZg9jEF4IHIYFcwgE7INwCh8o0A7AYzAEsYSIAjAQywApjzEJmBKRAPmjg8A3vggRQkWPCTJKJEtjyiIZGgGcYoAHTAYAc1aky3LgQC-hRi1MQA9LYjpMWfEA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.4.4&externalPlugins=)

This is source code of ES2015: 
```js
let foo = 'outer';

function bar(func = () => foo) {
  let foo = 'inner';
  console.log(func());
}

bar(); // outer

```
The result of run it: 

```js
outer
```

This is transform code by babel: 

```js
"use strict";

var foo = 'outer';

function bar() {
  var func = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : function () {
    return foo;
  };
  var foo = 'inner';
  console.log(func());
}

bar(); // outer

```

The result of run it: 

```js
inner
```

**Expected behavior/code**

Expect the results of the Babel conversion to be consistent with ES2015 performance

**Environment**
- Babel version(s): [v7.4.4]
- Node/npm version: [Node 8.9/npm 5]
- OS: [Windows 10]
- Monorepo: [no]
- How you are using Babel: [babeljs.io]

**Possible Solution**

```js
"use strict";

var foo = 'outer';

function bar() {
    var func = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : function() {
        return foo;
    }
    ;
    (function() {

        var foo = 'inner';
        console.log(func());

    }())
}

bar();

```

JLHwung:
| Fixed Issues?            | Fixes #9947 
| Documentation PR Link    | 

When there is a variable declaration inside the function body, which shares its name to a referenced identifier in default parameter expression, the function body should be wrapped into iife, otherwise the binding in default parameter scope will be shadowed by function body.

On the example, 
```js
// scope P
let x = "outside";
function outer(a = () => x // scope B) { // scope A
  let x = "inside";
  return a();
}
outer();
```

As is illustrated, there are 3 scopes in this program. Variable `x` in scope B will borrow from scope P. However, since there is `x` declared in scope A, we should wrap body in scope A to IIFE, otherwise the transformed code will have `x` in scope B borrowed from scope A, which is incorrect.

Here we check if it is a safe binding with respect to the function scope. Note that unlike the path scope changing within the traverse, this scope reference is invariant and always refers to the function declaration scope (in our example, it is scope A).

### DIFF
--- a/packages/babel-plugin-transform-parameters/src/params.js
+++ b/packages/babel-plugin-transform-parameters/src/params.js
@@ -32,7 +32,11 @@ function isSafeBinding(scope, node) {
 const iifeVisitor = {
   ReferencedIdentifier(path, state) {
     const { scope, node } = path;
-    if (node.name === "eval" || !isSafeBinding(scope, node)) {
+    if (
+      node.name === "eval" ||
+      !isSafeBinding(scope, node) ||
+      !isSafeBinding(state.scope, node)
+    ) {
       state.iife = true;
       path.stop();
     }
--- a/packages/babel-plugin-transform-parameters/test/fixtures/parameters/default-iife-9947/input.js
+++ b/packages/babel-plugin-transform-parameters/test/fixtures/parameters/default-iife-9947/input.js
@@ -1,0 +1,6 @@
+let x = "outside";
+function outer(a = () => x) {
+  let x = "inside";
+  return a();
+}
+outer();
--- a/packages/babel-plugin-transform-parameters/test/fixtures/parameters/default-iife-9947/output.js
+++ b/packages/babel-plugin-transform-parameters/test/fixtures/parameters/default-iife-9947/output.js
@@ -1,0 +1,13 @@
+var x = "outside";
+
+function outer() {
+  var a = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : function () {
+    return x;
+  };
+  return function () {
+    var x = "inside";
+    return a();
+  }();
+}
+
+outer();


nicolo-ribaudo:
Related: https://github.com/babel/babel/issues/2471

JLHwung:
Hi Nicolò,

Thanks for the quick response. The issue 2471 is still reproducible on master. I would try to look into this in couple of days. At first glance, renaming the inner binding looks better as it prevents from IIFE performance penalties. 

Where could I know all the possible values and the meaning of `binding.kind`? It would greatly help list every binding cases and evaluate whether it is safe to rename.

JLHwung:
Hi Nicolò,

I have tried to use `path.scope.rename` but it seems that the `renamer` will also rename the referenced identifier borrowed from parent scope. I have added extra test to showcase this potential issue. (PR is [here](https://github.com/babel/babel/pull/10055)) The `renamer` approach is blocked by this issue.

Though issue #2471 is related, I would suggest move things forward and try to fix them on different PR. Any thoughts?

packages/babel-plugin-transform-parameters/src/params.js
```diff
@@ -32,7 +32,11 @@ function isSafeBinding(scope, node) {
 const iifeVisitor = {
   ReferencedIdentifier(path, state) {
     const { scope, node } = path;
-    if (node.name === "eval" || !isSafeBinding(scope, node)) {
+    if (
+      node.name === "eval" ||
+      !isSafeBinding(scope, node) ||
```
JLHwung: This should be considered a workaround hotfix because the ReferencedIdentifier in params initializers should point to the parent of the function scope. Because `path.scope` is not accurate here, we have to do a hotfix and check `state.scope` again.

The ideal fix should construct a separate scope for parameter initializer.

PR #10055, opened 2019-06-05
test: add parameter default test to renamer

JLHwung:
| Fixed Issues?            | Reveal potential issue
| Patch: Bug Fix?          | Determined after further discussion
| Documentation PR Link    | 

This is a follow-up to https://github.com/babel/babel/pull/10053#pullrequestreview-245304223

Expected: 
- `Renamer` should not rename the bindings in the function default parameter if it is borrowed from parent scope
- `Renamer` should rename the bindings in the function default parameter if it is borrowed from the `param` binding in the function scope.

Actual:
- `Renamer` renames the bindings in the function default parameter even if it is borrowed from parent scope
- `Renamer` renames the bindings in the function default parameter when it is borrowed from the `param` binding in the function scope.

On the test example, we have renamer to rename `a` to `z` on `FunctionDeclaration` visitor.
In the first case,
```js
let a = "outside";
function f(g = () => a) {
  let a = "inside";
  return g();
}
```
The Renamer also rename `() => a` to `() => z`. This behavior is wrong since `a` is defined on global scope.

In the second case,
```js
function h(a, g = () => a) {
  return g();
}
```

The renamer should also `() => a` to `() => z` since `a` is a `param` binding now.

If we agree on the correctness of my proposed behavior, then this PR reveals a bug in `renamer` and should be fixed in later release.

Note that CI will break since the master does not conform to this behavior.

### DIFF
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -45,6 +45,20 @@ function gatherNodeParts(node: Object, parts: Array) {
   }
 }
 
+function isScopeInFunctionParameter(scope) {
+  //todo: check arrow function expression or any util method?
+  if (scope.parent && scope.parent.block.type === "FunctionDeclaration") {
+    const fdPath = scope.path.findParent(
+      path => path.type === "FunctionDeclaration",
+    );
+    if (fdPath) {
+      return !!fdPath.findParent(path => {
+        return fdPath.node.params.includes(path.node);
+      });
+    }
+  }
+  return false;
+}
 //
 
 const collectorVisitor = {
@@ -906,7 +920,21 @@ export default class Scope {
     do {
       const binding = scope.getOwnBinding(name);
       if (binding) return binding;
-    } while ((scope = scope.parent));
+      if (isScopeInFunctionParameter(scope)) {
+        scope = scope.parent;
+        if (!scope) {
+          break;
+        }
+        const binding2 = scope.getOwnBinding(name);
+        if (binding2 && binding2.kind === "param") {
+          return binding2;
+        }
+      }
+      scope = scope.parent;
+      if (!scope) {
+        break;
+      }
+    } while (true);
   }
 
   getOwnBinding(name: string) {
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -1,6 +1,6 @@
 import includes from "lodash/includes";
 import repeat from "lodash/repeat";
-import Renamer from "./lib/renamer";
+import Renamer, { isPathInFunctionParameter } from "./lib/renamer";
 import type NodePath from "../path";
 import traverse from "../index";
 import defaults from "lodash/defaults";
@@ -46,16 +46,8 @@ function gatherNodeParts(node: Object, parts: Array) {
 }
 
 function isScopeInFunctionParameter(scope) {
-  //todo: check arrow function expression or any util method?
-  if (scope.parent && scope.parent.block.type === "FunctionDeclaration") {
-    const fdPath = scope.path.findParent(
-      path => path.type === "FunctionDeclaration",
-    );
-    if (fdPath) {
-      return !!fdPath.findParent(path => {
-        return fdPath.node.params.includes(path.node);
-      });
-    }
+  if (scope.parent && t.isFunction(scope.parent.block)) {
+    return isPathInFunctionParameter(scope.path);
   }
   return false;
 }
@@ -916,25 +908,18 @@ export default class Scope {
 
   getBinding(name: string) {
     let scope = this;
+    let childScopeInFunctionParameter = false;
 
     do {
       const binding = scope.getOwnBinding(name);
-      if (binding) return binding;
-      if (isScopeInFunctionParameter(scope)) {
-        scope = scope.parent;
-        if (!scope) {
-          break;
+      if (binding) {
+        if (childScopeInFunctionParameter && binding.kind !== "param") {
+          continue;
         }
-        const binding2 = scope.getOwnBinding(name);
-        if (binding2 && binding2.kind === "param") {
-          return binding2;
-        }
+        return binding;
       }
-      scope = scope.parent;
-      if (!scope) {
-        break;
-      }
-    } while (true);
+      childScopeInFunctionParameter = isScopeInFunctionParameter(scope);
+    } while ((scope = scope.parent));
   }
 
   getOwnBinding(name: string) {
--- a/packages/babel-traverse/src/scope/lib/renamer.js
+++ b/packages/babel-traverse/src/scope/lib/renamer.js
@@ -126,8 +126,8 @@ export default class Renamer {
         if (!state.paramDefaultOuterBinding) {
           if (
             right.isIdentifier() &&
-            right.node.body.name === oldName &&
-            !isSafeBinding(scope, right.node.body)
+            right.node.name === oldName &&
+            !isSafeBinding(scope, right.node)
           ) {
             state.paramDefaultOuterBinding = true;
             break;
--- a/packages/babel-traverse/src/scope/lib/renamer.js
+++ b/packages/babel-traverse/src/scope/lib/renamer.js
@@ -2,18 +2,17 @@ import Binding from "../binding";
 import splitExportDeclaration from "@babel/helper-split-export-declaration";
 import * as t from "@babel/types";
 
-function isInFunctionParameterScope(path) {
-  const computableNode = path.findParent(path => {
-    return (
-      path.type === "AssignmentPattern" ||
-      (path.type === "ObjectProperty" && path.node.computed)
-    );
-  });
-  const fdPath = path.findParent(path => path.type === "FunctionDeclaration");
-  if (computableNode && fdPath) {
-    return !!path.findParent(path => {
-      return fdPath.node.params.includes(path.node);
-    });
+export function isPathInFunctionParameter(path) {
+  let p = path;
+  while (p != null) {
+    if (
+      p.parentPath &&
+      p.parentPath.isFunction() &&
+      p.parentPath.node.params.includes(p.node)
+    ) {
+      return true;
+    }
+    p = p.parentPath;
   }
   return false;
 }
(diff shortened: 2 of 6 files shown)


nicolo-ribaudo:
I think that, in order to fix this, we need to change this line:
https://github.com/babel/babel/blob/master/packages/babel-traverse/src/scope/lib/renamer.js#L119

instead of running `renameVisitor`, we should first check where the binding is defined.
If the current scope is a function scope and the binding is _not_ defined inside it's parameters, then we should only traverse its body instead of the whole declaration.

JLHwung:
Hi Nicolo,

Thanks for your reply and I have drafted a WIP commit. However, after I add another test case
```js
let a = "outside";

function j(g = a) {
  return g;
}
```
it successfully fails my WIP draft commit though the commit works good on those two testcase. I would work on this test case tomorrow. And I believe this case would also improve code coverage on the `transform-parameters`. I would sync it to #10053 later.

JLHwung:
Hi Nicolo, I have polished my work and it should be ready to review.

On this example,
```js
let a = "outside";

function j(g = a) {
  return g;
}
```

I think it is expected since the binding scope is not always same with the `scope` that invokes renamer. On this case, there is no `a` binding inside the Function scope, so it will rename `a` to `z` in the whole program.

packages/babel-traverse/src/scope/lib/renamer.js
```diff
@@ -97,6 +116,30 @@ export default class Renamer {
     );
   }
 
+  parameterDefaultHasOuterBinding(path, state) {
+    const { oldName, scope } = state;
+    const params = path.get("params");
+    for (let i = 0; i < params.length; i++) {
+      const param = params[i];
+      if (param.isAssignmentPattern()) {
```
nicolo-ribaudo: This is not enough; someone could also do this: (which is currently broken anyway)

```js
let name;

function fn({ [name]: value }) {
  let name;
}
```

or

```js
let name;

function fn({ key = name }) {
  let name;
}
```

packages/babel-traverse/src/scope/lib/renamer.js
```diff
@@ -97,6 +116,30 @@ export default class Renamer {
...
+        const right = param.get("right");
+        if (!state.paramDefaultOuterBinding) {
+          if (
+            right.isIdentifier() &&
+            right.node.body.name === oldName &&
```
nicolo-ribaudo:
```suggestion
            right.node.name === oldName &&
```

Identifiers don't have `body`, right?
Also, this should have been caught by a test.

packages/babel-traverse/src/scope/lib/renamer.js
```diff
@@ -97,6 +116,30 @@ export default class Renamer {
...
+            !isSafeBinding(scope, right.node.body)
```
nicolo-ribaudo:
```suggestion
            !isSafeBinding(scope, right.node)
```

nicolo-ribaudo:
Please write a few more tests, to ensure that everything works :slightly_smiling_face:

packages/babel-traverse/src/scope/lib/renamer.js
JLHwung: Ahh, good catch. Fixed and testcase is added.

packages/babel-traverse/src/scope/lib/renamer.js
JLHwung: Fixed in 840a26e.

JLHwung:
Hi Nicolo,

> If the current scope is a function scope and the binding is not defined inside it's parameters, then we should only traverse its body instead of the whole declaration.

I have crafted some examples and it turns out this "body/whole" approach would not cover some edge cases. In principal, what's missing here is that babel-traverse has not modeled the implicit intermediate scope for function parameters.

## Examples of different categories
### isolated scope
On the following example,
```js
let a = "outside"
f(g = () => a) {
  let a = "inside";
}
```

`() => a` is a new scope, but the `path.scope.bindingIdentifierEquals("a",  state.binding.identifier)` falsely returns `true` since `path.scope.getBinding()` returns `let a = "inside"` binding. If the implicit scope is modeled, `() => a` would not have binding in `g = () => a` scope and will pass on to the outer scope. Note that this is a separate bug in `scope.getBinding` and the "body/whole" approach does not fix it.

### same scope
Besides new scope, another category is expressions in various binding patterns. For example

```js
// computed property in `ObjectBindingPattern`
function l([{
  [a]: g
}]) {}

// expression in `KeyedBindingInitializer`
function k([{
  g = a
}]) {
  let a = "inside";
  return g;
}
```

These expressions do not introduce a new scope. So the `renameVisistor` will treat it as same as any other `ReferencedIdentifier`. If the parameter scope is modeled, the `Scope` visitor will kick in and check if could be skipped.

### nested binding patterns
Perhaps the most complicated examples comes from nested binding patterns. On this example
```js
function p(a, g = (a = a) => a) {
  g(a);
}
```
There are two `a` bindings in function `p`. One is `param` binding of `p`, the other is `param` binding of default parameter of `g`. When `scope.getBinding()` called on `(a = a) => a`, it returns the latter binding so the `Scope` visitor skip the whole path even though there is an `= a` binding, whose identifier exactly the param binding `a` of `p`.

The "body/whole" assumes that if `a` is defined as parameter in scope `p`, it should be safe to replace all bindings given that those in a new scope, who has different binding of `a`, should be skipped. However, the binding patterns can easily break across the scopes.

## New approach

So today I have tried a different approach other than "body/whole" to solve these. For those in separated scope, I patched the `scope.getBinding` so that it can returns correct binding when it is in function parameter scope.

For those expressions in the same scope, I added logic in `ReferencedIdentifier` visitor to check ancestry if it is in function parameter scope.

For the third cases, I think we should should do a recursive revised `renameVisitor` in the `Scope` visitor to pick up those possible nested binding patterns. It will be triggered only when the new scope is inside the parameter scope. It is a rough idea and I suggest to move this to a separate PR.

I have pushed a working edition of my new approach, it is still proof-of-concept and, of course need refactoring before ready to merge. I would like to discuss this idea before I go too far.

Thank you for your patience.

JLHwung:
Friendly ping @nicolo-ribaudo :)

JLHwung:
Ping @nicolo-ribaudo. :)

packages/babel-traverse/src/scope/index.js
```diff
@@ -45,6 +45,20 @@ function gatherNodeParts(node: Object, parts: Array) {
   }
 }
 
+function isScopeInFunctionParameter(scope) {
+  //todo: check arrow function expression or any util method?
+  if (scope.parent && scope.parent.block.type === "FunctionDeclaration") {
```
nicolo-ribaudo: This could be any function kind. It should be `t.isFunction(scope.parent.block)`

packages/babel-traverse/src/scope/index.js
```diff
@@ -45,6 +45,20 @@ function gatherNodeParts(node: Object, parts: Array) {
...
+    const fdPath = scope.path.findParent(
+      path => path.type === "FunctionDeclaration",
```
nicolo-ribaudo: Also here

packages/babel-traverse/src/scope/index.js
```diff
@@ -45,6 +45,20 @@ function gatherNodeParts(node: Object, parts: Array) {
...
+    );
+    if (fdPath) {
+      return !!fdPath.findParent(path => {
+        return fdPath.node.params.includes(path.node);
```
nicolo-ribaudo: I'm not sure that I understand this. If `path.node` must be _inside_ the parameters of `fdPath`, why are we searching it in the _parents_ of `fdPath`?

packages/babel-traverse/src/scope/index.js
```diff
@@ -906,7 +920,21 @@ export default class Scope {
...
+        const binding2 = scope.getOwnBinding(name);
+        if (binding2 && binding2.kind === "param") {
+          return binding2;
+        }
+      }
+      scope = scope.parent;
+      if (!scope) {
+        break;
+      }
+    } while (true);
```
nicolo-ribaudo: Can we keep this as `while ((scope = scope.parent))`?

packages/babel-traverse/src/scope/lib/renamer.js
```diff
@@ -2,9 +2,33 @@ import Binding from "../binding";
...
 import * as t from "@babel/types";
 
+function isInFunctionParameterScope(path) {
+  const computableNode = path.findParent(path => {
+    return (
+      path.type === "AssignmentPattern" ||
+      (path.type === "ObjectProperty" && path.node.computed)
+    );
+  });
+  const fdPath = path.findParent(path => path.type === "FunctionDeclaration");
```
nicolo-ribaudo: 1. We should only compute this is `computableNode` has been found
2. Can we share lines 12-17 with the same code in the `isScopeInFunctionParameter` function?

packages/babel-traverse/src/scope/lib/renamer.js
```diff
@@ -2,9 +2,33 @@ import Binding from "../binding";
 import splitExportDeclaration from "@babel/helper-split-export-declaration";
 import * as t from "@babel/types";
 
+function isInFunctionParameterScope(path) {
+  const computableNode = path.findParent(path => {
+    return (
+      path.type === "AssignmentPattern" ||
```
nicolo-ribaudo: WDYT about using `path.parentPath.isAssignmentPattern({ right: path.node })` instead?
I don't think that it would change anything, but the `left` of an assignment isn't computable anyway.

JLHwung: > WDYT about using path.parentPath.isAssignmentPattern({ right: path.node }) instead?

Good point. I have refactored this part and feel horrible on what I was writing a month ago.

packages/babel-traverse/src/scope/index.js
JLHwung: True, I have refactored this part.

packages/babel-traverse/src/scope/lib/renamer.js
```diff
@@ -2,9 +2,40 @@ import Binding from "../binding";
...
 const renameVisitor = {
-  ReferencedIdentifier({ node }, state) {
+  ReferencedIdentifier(path, state) {
+    const { node } = path;
     if (node.name === state.oldName) {
+      let p = path;
+      while (
+        p.parentPath &&
+        !p.parentPath.isAssignmentPattern({ right: p.node }) &&
+        !p.isObjectProperty({ computed: true })
```
nicolo-ribaudo: We should only stop if we were inside the key, not if we were inside the "value"/pattern of a computed property.

liuxingbaoyu:
Hi, why hasn't it been merged?

JLHwung:
Closing this PR in favor of #10912.

packages/babel-traverse/src/scope/lib/renamer.js
JLHwung: I reworked on #10912 to get rid of the spaghetti. 😝

PR #9754, opened 2019-03-23
[runtime-corejs3] Only polyfill instance methods when it might be needed

nicolo-ribaudo:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #9753
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | Yes?
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

This PR allows instance methods polyfills to specify on which base type they need to be applied. If Babel is 100% that it is called on another instance type, it will avoid polyfilling it.

For example, `"".concat` doesn't need to be polyfilled but `foo.concat` does because `foo` might be an array.
#9753 shows that polyfilling `.concat` breaks webpack logic to handle `require` calls with template literals.

Other instance methods could be optimized, but if this PR lands I'll leave it as a `[good first issue]` PR.

PR #10068, opened 2019-06-09
feat(transform-runtime): add more type hints for instance methods

devinus:
| Patch: Bug Fix?          | Potentially
| Documentation PR Link    |
| Any Dependency Changes?  | Yes, lodash added to transform-runtime

Continues the work from #9754 and adds more hints to the corejs3 definition type hints.


packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
```diff
@@ -185,47 +185,51 @@ export default () => {
...
+      fill: { stable: true, path: "fill", types: ["array"] },
+      filter: { stable: true, path: "filter", types: ["array"] },
+      find: { stable: true, path: "find", types: ["array"] },
+      findIndex: { stable: true, path: "find-index", types: ["array"] },
+      flags: { stable: true, path: "flags", types: ["regexp"] },
+      flatMap: { stable: true, path: "flat-map", types: ["array"] },
+      flat: { stable: true, path: "flat", types: ["array"] },
+      forEach: { stable: true, path: "for-each", types: ["array"] },
+      includes: { stable: true, path: "includes", types: ["array", "string"] },
+      indexOf: { stable: true, path: "index-of", types: ["array", "string"] },
```
zloirock: `core-js` does not add polyfills for `String#{ indexOf, lestIndexOf }` - they are not required.

packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
```diff
@@ -185,47 +185,51 @@ export default () => {
...
-      startsWith: { stable: true, path: "starts-with" },
-      trim: { stable: true, path: "trim" },
-      trimEnd: { stable: true, path: "trim-end" },
-      trimLeft: { stable: true, path: "trim-left" },
-      trimRight: { stable: true, path: "trim-right" },
-      trimStart: { stable: true, path: "trim-start" },
-      values: { stable: true, path: "values" },
+      copyWithin: { stable: true, path: "copy-within", types: ["array"] },
+      endsWith: { stable: true, path: "ends-with", types: ["string"] },
+      entries: { stable: true, path: "entries", types: ["array"] },
```
zloirock: `keys`, `values`, `entries`, `forEach` required also for DOM collection.

packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
```diff
@@ -185,47 +185,51 @@ export default () => {
...
+      map: { stable: true, path: "map", types: ["array"] },
+      matchAll: { stable: false, path: "match-all", types: ["string"] },
+      padEnd: { stable: true, path: "pad-end", types: ["string"] },
+      padStart: { stable: true, path: "pad-start", types: ["string"] },
+      reduce: { stable: true, path: "reduce", types: ["array"] },
+      reduceRight: { stable: true, path: "reduce-right", types: ["array"] },
+      repeat: { stable: true, path: "repeat", types: ["string"] },
+      replaceAll: { stable: false, path: "replace-all", types: ["string"] },
+      reverse: { stable: true, path: "reverse", types: ["array"] },
+      slice: { stable: true, path: "slice", types: ["array", "string"] },
```
zloirock: It's required only for `Array`.

packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
```diff
@@ -185,47 +185,51 @@ export default () => {
...
+      fill: { stable: true, path: "fill", types: ["array"] },
+      filter: { stable: true, path: "filter", types: ["array"] },
+      find: { stable: true, path: "find", types: ["array"] },
+      findIndex: { stable: true, path: "find-index", types: ["array"] },
+      flags: { stable: true, path: "flags", types: ["regexp"] },
+      flatMap: { stable: true, path: "flat-map", types: ["array"] },
+      flat: { stable: true, path: "flat", types: ["array"] },
+      forEach: { stable: true, path: "for-each", types: ["array"] },
+      includes: { stable: true, path: "includes", types: ["array", "string"] },
+      indexOf: { stable: true, path: "index-of", types: ["array"] },
```
nicolo-ribaudo: :exclamation: I think that this is the only actually needed line in this PR, since we don't need to load the polyfill for `"".indexOf`.
All the other additions only prevent us from adding polyfills for code that will throw at runtime.

packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
```diff
@@ -185,47 +185,51 @@ export default () => {
...
+      entries: { stable: true, path: "entries", types: ["array"] },
+      every: { stable: true, path: "every", types: ["array"] },
+      fill: { stable: true, path: "fill", types: ["array"] },
+      filter: { stable: true, path: "filter", types: ["array"] },
+      find: { stable: true, path: "find", types: ["array"] },
+      findIndex: { stable: true, path: "find-index", types: ["array"] },
+      flags: { stable: true, path: "flags", types: ["regexp"] },
+      flatMap: { stable: true, path: "flat-map", types: ["array"] },
+      flat: { stable: true, path: "flat", types: ["array"] },
+      forEach: { stable: true, path: "for-each", types: ["array"] },
```
nicolo-ribaudo: This is also for map/set, but I _think_ that Babel handles them as "unknown" and thus adds the polyfill anyway?

packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
nicolo-ribaudo: Does this work? (also for keys/values)

```js
Object.entries;
```

packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
devinus: > I think that Babel handles them as "unknown" and thus adds the polyfill anyway?

Correct.

packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
devinus: > Does this work?

Yes, it's treated as unknown.

packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
devinus: Doesn't it also prevent polyfilling for literals that may have had their prototypes monkey-patched? Which is really dumb but also possible.

packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
devinus: Yes, that's a static method rather than an instance method.

packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
nicolo-ribaudo: Could you add a test? I don't want to introduce bugs in case we decide to optimize the logic also for map/set other than array/etc.

zloirock: The problem with `Map` and `Set` that even if we will add naive detection of them to `typeAnnotationToString` to prevent adding unnecessary `.forEach` polyfill, `Map` anyway could be transpiled to something like `_Map` by `runtime` transform.

packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
zloirock: Seems, for better effect, we should extend `typeAnnotationToString` for work with more classes / types.

PR #10069, opened 2019-06-09
feat(transform-runtime): allow customizing corejs definitions

devinus:
| Documentation PR Link    |
| Any Dependency Changes?  | Yes, lodash added to transform-runtime

This allows users to customize the corejs definitions used during transform if needed. This is useful if for instance the user wants to add features already available in corejs but not yet added to the definitions.


zloirock:
Interesting solution 👍

packages/babel-plugin-transform-runtime/package.json
```diff
@@ -14,6 +14,7 @@
   "dependencies": {
     "@babel/helper-module-imports": "^7.0.0",
     "@babel/helper-plugin-utils": "^7.0.0",
+    "lodash": "^4.17.11",
```
nicolo-ribaudo: Can you add a dependency on `lodash.merge` instead?

devinus: Other packages depend on lodash, so rather than introducing a new dependency I figured I'd just depend on the same and allow it to be deduped.

Also I believe the lodash maintainers no longer recommend using the modularized packages.

PR #10070, opened 2019-06-09
Do expressions transform for switch statements

### LINKED ISSUE #3872: Do expressions transform incorrectly for switch statements (T6822)
> Issue originally made by @AppShipIt
### Bug information
- **Babel version:** 6.3.17
- **Node version:** 5.0.0
- **npm version:** 3.5.1
### Options

```
{
  "presets": [
    "es2015",
    "stage-0"
  ]
}
```
### Input code

```
const testSwitch = () => do {
  switch (new Date().getDay()) {
    case 0: case 6: "weekend 🚵"; break
    default: "weekday 🚴"
  }
}

alert(`Today is a ${testSwitch() || "🐛"}`)
```
### Description

Per [[ https://twitter.com/BrendanEich/status/676497551495856129 | tweet ]] from Brendan Eich:
### Expected

```
"use strict";

var testSwitch = function testSwitch() {
  return (function () {
    switch (new Date().getDay()) {
      case 0:case 6:
        return "weekend 🚵";
      default:
        return "weekday 🚴";
    }
  })();
};

alert("Today is a " + (testSwitch() || "🐛"));
```
### Got

```
"use strict";

var testSwitch = function testSwitch() {
  return (function () {
    switch (new Date().getDay()) {
      case 0:case 6:
        "weekend 🚵";break;
      default:
        "weekday 🚴";
    }
  })();
};

alert("Today is a " + (testSwitch() || "🐛"));
```

[[ http://codepen.io/dnalot/pen/VewJez | Live demo ]]

tanhauhau:
Rebased #6975, fixed PR review issues

| Fixed Issues?            | Fixes #3872, closes #6975
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-traverse/src/path/family.js
```diff
@@ -18,6 +18,34 @@ function addCompletionRecords(path, paths) {
...
+      // get the previous statement
+      const previousKey = breakStatement.key - 1;
+      if (previousKey >= 0 && consequent[previousKey].isExpressionStatement()) {
+        paths = addCompletionRecords(consequent[previousKey], paths);
+        breakStatement.remove();
+      } else {
+        breakStatement.replaceWith(switchCase.scope.buildUndefinedNode());
+        paths = addCompletionRecords(breakStatement, paths);
+      }
+    } else if (consequent.length > 0 && isDefaultCase) {
```
nicolo-ribaudo: `default` cases have nothing special, they should be handled exactly like `case`:

This should evaluate to `1`:
```js
const a = do {
  switch (2) {
    default: 0;
    case 1: 1; break;
  }
};
```

This should evaluate to `1`:
```js
const a = do {
  switch (1) {
    case 1: 1;
  }
};
```

The check should probably be more like `if (isLastStatement) {`

nicolo-ribaudo: Also this should evaluate to 1:
```js
const a = do {
  switch (1) {
    case 1: 1;
    case 2:
  }
};
```

tanhauhau: So can I say that the lastStatement is an expression statement , then no fall through, else will fall through say:

```
let a = 1
const b = do {
switch (1) {
   case 1: a = 2;
   case 2: a * 10;
}
```

So b = ?

nicolo-ribaudo: 10, because the last statement is `a*10`

tanhauhau: ok. interesting. is this documented somewhere within https://github.com/tc39/proposal-do-expressions? or where do I find the spec for this?

nicolo-ribaudo: 1. Switch expressions evaluate to the result of their block: https://tc39.es/ecma262/#sec-switch-statement-runtime-semantics-evaluation steps 7 and 9
2. It delegates to https://tc39.es/ecma262/#sec-runtime-semantics-caseblockevaluation, in the `CaseBlock: { CaseClauses }` section.
  i. At the first iteration of step 4 of that algorithm, `V` is set to `2` (the result of `a = 2`).
  ii. Since `case 1` isn't followed by a break/return/throw statement, `R` is a NormalCompletion and not an AbruptCompletion: for this reason the loop continues running to the following case.
  iii. At the second iteration, `V` is set to `20` (the result of `a * 10`).
  iv. The loop ends, because there are no more `case`s, and it returns `20`

nicolo-ribaudo: :man_facepalming: :man_facepalming: :man_facepalming: :man_facepalming: 
I just noticed I wrote 10 instead of 20

tanhauhau: I realised reading the ECMA262 Spec is giving me more confusion 🤦‍♂ 🤦‍♂ 🤦‍♂, but I guess I roughly know what you trying to point out. And I think we should document more in the do-expression proposal docs

nicolo-ribaudo: Btw, you can already "test" do expression in any browser.
```js
do { ANYTHING }
```
is
```js
eval(`{
  ANYTHING
}`);

File: packages/babel-plugin-proposal-do-expressions/test/fixtures/do-expression-switch-case/fallthrough-multiblock/exec.js [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+    case 1: 
+      { "c"; }
+      break;
+      { "d"; };
+    case 2: 
+      "a";
+      "b";
+    case 3:
+      {}
+      { break; }
```
tanhauhau: @nicolo-ribaudo i need some help over here, i've figured this as a plausible test case, but I not sure how i should transform this.

Previously, what I did is to find the `BreakStatement`, and turn the statement before `BreakStatement` as a completionPath, which would then be transformed into a return statement.

Problem I've created right now is that the previous statement could be anything from `BlockStatement` to nested `BlockStatement`s, and I need to find the `"last statement"` which I am kind stuck on how I should go for this...
not entirely sure I am in the right direction, that's why I am seeking some help over here

nicolo-ribaudo: I'm not sure about how to solve it, but it isn't a problem just for switchs. It's a problem for everything which uses `break`.

e.g.
```js
let a = do {
  while (1) {
    1;
    break;
    2;
  }
};
```

I think that we should try to fix it separately since it is a really hard problem and this PR is already a big improvement.

tanhauhau: OK I guess I shall disable that test case then

tanhauhau:
@nicolo-ribaudo any update on this?

packages/babel-traverse/src/path/family.js
```diff
@@ -18,6 +18,61 @@ function addCompletionRecords(path, paths) {
...
+        breakStatement.key > 0 &&
+        (prevSibling.isExpressionStatement() || prevSibling.isBlockStatement())
+      ) {
+        paths = addCompletionRecords(prevSibling, paths);
+        breakStatement.remove();
+      } else {
+        breakStatement.replaceWith(switchCase.scope.buildUndefinedNode());
+        paths = addCompletionRecords(breakStatement, paths);
+      }
+    } else {
```
JLHwung: I suggest check `isLastCaseWithConsequent` early so that it may skip the whole `hasConsequent` setup when it is false.

packages/babel-traverse/src/path/family.js
```diff
@@ -18,6 +18,61 @@ function addCompletionRecords(path, paths) {
...
+
+      const prevSibling = breakStatement.getPrevSibling();
+      if (
+        breakStatement.key > 0 &&
+        (prevSibling.isExpressionStatement() || prevSibling.isBlockStatement())
+      ) {
+        paths = addCompletionRecords(prevSibling, paths);
+        breakStatement.remove();
+      } else {
+        breakStatement.replaceWith(switchCase.scope.buildUndefinedNode());
```
JLHwung: nit: I think it should be `breakStatement.scope.buildUndefinedNode()`.

packages/babel-traverse/src/path/family.js
tanhauhau: @JLHwung updated!

packages/babel-traverse/src/path/family.js
tanhauhau: @JLHwung sure!

JLHwung: After #10243 they will be logically equivalent: `buildUndefinedNode` does not depend on scope any more. It is by reviewing this PR that I find the room for improvements in `buildUndefinedNode`.

tanhauhau:
@nicolo-ribaudo bump

PR #10071, opened 2019-06-09
feat(transform-runtime): update core-js@3 definitions

devinus:
| Documentation PR Link    |

This updates core-js@3 definitions used by transform-runtime to include every supportable polyfill and also now allows you to specify a `core-js` minor version like e.g. `preset-env`.


zloirock:
Typed arrays missed in the pure `core-js` version. Static methods of `Promise` or collections injected with constructors entry points, so separate definitions for them not required. Also, it could be a breaking change for a case of modern transform and old runtime, see a workaround for this situation in runtime for `core-js@2`.

packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
```diff
@@ -207,7 +301,7 @@ export default () => {
       keys: { stable: true, path: "keys" },
       lastIndexOf: { stable: true, path: "last-index-of" },
       map: { stable: true, path: "map" },
-      matchAll: { stable: false, path: "match-all" },
+      matchAll: { stable: true, path: "match-all" },
```
zloirock: It's a breaking change for a case of `core-js@3.0` runtime dependency.

packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js
```diff
@@ -168,16 +230,48 @@ export default () => {
         iterator: { stable: true, path: "symbol/iterator" },
         keyFor: { stable: true, path: "symbol/key-for" },
         match: { stable: true, path: "symbol/match" },
+        matchAll: { stable: false, path: "symbol/match-all" },
```
zloirock: `symbol/match-all` and `symbol/replace-all` available only from `core-js@3.1`, so it's also a breaking change for a case of old dependencies.

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -53,17 +54,17 @@ export default declare((api, options, dirname) => {
     rawVersion = corejs;
   }
 
-  const corejsVersion = rawVersion ? Number(rawVersion) : false;
+  const corejsVersion = rawVersion ? coerce(String(rawVersion)) : false;
 
-  if (![false, 2, 3].includes(corejsVersion)) {
+  if (corejsVersion !== false && !(2 <= corejsVersion.major <= 3)) {
```
zloirock: Since `core-js-pure` is a dependency of `@babel/runtime-corejs3` and missed in direct dependencies of users, unlike `preset-env`, it will not work without serious changes in the architecture.

devinus: Indeed, this is a dead end for now.

PR #10072, opened 2019-06-09
fix await and yield for do expression

### LINKED ISSUE #3780: Async functions with do expressions: inconsistent behavior (T6716)
> Issue originally made by @bathos
### Bug information
- **Babel version:** 5.8.34
- **Node version:** (browser)
### Options

```
async-functions, do-expressions
```
### Input code

``` js
const x = new Promise(resolve => resolve(5));

async function p() {
  const y = do {
    let z;
    await x;
  };

  return y;
}

p().then(::console.log);
```
### Description

If a do-expression contains an 'await' (in the context of an async function), Babel may or may not consider this an error depending on whether the do-expression was transpiled as a bare expression or as an IIFE. The following will successfully log '5' to the console:

```
const x = new Promise(resolve => resolve(5));

async function p() {
  const y = do {
    await x;
  };

  return y;
}

p().then(::console.log);
```

The following will throw an unexpected identifier error, because the do-expression required an IIFE when transpiled:

```
const x = new Promise(resolve => resolve(5));

async function p() {
  const y = do {
    let z;
    await x;
  };

  return y;
}

p().then(::console.log);
```

It seems that this should either work or not work regardless of the specific content of the do-expression. I'm not entirely sure what the correct behavior should be, though, since I can't find a formal spec for do expressions; however the summary is "treat a block statement as an expression" so it sounds like async should be available. If I understand right, the issue is that the IIFE translation of a do-expression containing an await would need to both inherit its parent's "asyncness", and also be automatically prepended with an await itself ... or something like that.

tanhauhau:
| Fixed Issues?            | Fixes #3780
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-traverse/src/path/introspection.js
+++ b/packages/babel-traverse/src/path/introspection.js
@@ -110,24 +110,6 @@ export function canSwapBetweenExpressionAndStatement(replacement) {
 }
 
 /**
- * Check whether the current path contains an `await` expression
- */
-
-export function containsAwaitExpression(): boolean {
-  let foundAwait = false;
-  this.traverse({
-    Function(path) {
-      path.skip();
-    },
-    AwaitExpression(path) {
-      foundAwait = true;
-      path.stop();
-    },
-  });
-  return foundAwait;
-}
-
-/**
  * Check whether the current path references a completion record
  */
 
--- a/packages/babel-traverse/src/path/introspection.js
+++ b/packages/babel-traverse/src/path/introspection.js
@@ -110,6 +110,24 @@ export function canSwapBetweenExpressionAndStatement(replacement) {
 }
 
 /**
+ * Check whether the current path contains an `await` expression
+ */
+
+export function containsAwaitExpression(): boolean {
+  let foundAwait = false;
+  this.traverse({
+    Function(path) {
+      path.skip();
+    },
+    AwaitExpression(path) {
+      foundAwait = true;
+      path.stop();
+    },
+  });
+  return foundAwait;
+}
+
+/**
  * Check whether the current path references a completion record
  */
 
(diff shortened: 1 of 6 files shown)


packages/babel-traverse/src/path/introspection.js
```diff
@@ -109,6 +109,24 @@ export function canSwapBetweenExpressionAndStatement(replacement) {
   return false;
 }
 
+/**
+ * Check whether the current path contains an `await` expression
+ */
+
+export function containsAwaitExpression(): boolean {
```
nicolo-ribaudo: You can/should use `hasType` instead:

https://github.com/babel/babel/blob/f92c2ae830dbb32013a36fa74facd2ef95b9947d/packages/babel-plugin-transform-block-scoping/src/index.js#L572-L576

nicolo-ribaudo: Disclaimer: I didn't know about that method until 5 mins ago. Usually, when someone adds a new method to `@babel/traverse` there is a 50% probability that it already exists.

tanhauhau: I tried to look for it too... But I guess you can find things better than me

nicolo-ribaudo:
Can you disable the failing test on node 6, since it doesn't support async functions?

Example:
https://github.com/babel/babel/blob/eac4c5bc17133c2857f2c94c1a6a8643e3b547a7/packages/babel-plugin-transform-new-target/test/fixtures/exec/reflect-class/options.json#L1-L3

nicolo-ribaudo:
Sorry for the bad example, the min node version should be 8 😅

tanhauhau:
Oh wait what was I thinking..  haha

tanhauhau:
@nicolo-ribaudo I'm thinking might as well fix for yield generators as well

nicolo-ribaudo:
I'd prefer it do be done in a separate PR, since there are many more edge cases (e.g. manually calling `.throw` or `.return`, and how `try/finally` with nested `do` expression behave).

tanhauhau:
@nicolo-ribaudo fixed the MR

PR #10093, opened 2019-06-13
removed duplicate tests for @babel/highlight getChalk method

Letladi:
Removed duplicate tests for the @babel/highlight getChalk method.

### DIFF
--- a/packages/babel-highlight/test/index.js
+++ b/packages/babel-highlight/test/index.js
@@ -94,8 +94,8 @@ describe("@babel/highlight", function() {
       });
     });
 
-    describe("when colors are supported", function() {
-      stubColorSupport(true);
+    describe("when colors are not supported", function() {
+      stubColorSupport(false);
 
       describe("when forceColor is not passed", function() {
         it("returns a Chalk instance", function() {


packages/babel-highlight/test/index.js
```diff
@@ -93,23 +93,5 @@ describe("@babel/highlight", function() {
         });
       });
     });
-
-    describe("when colors are supported", function() {
```
nicolo-ribaudo: Maybe we should test `when colors are not supported` instead?

Letladi: Maybe so @nicolo-ribaudo, currently I'm just learning the code and going through it so I can't say what should be done at this point 😅

nicolo-ribaudo: Yes, I think that it was duplicated because one of them was meant to test when colors were not supported

Letladi: Alright. I will update the pull request with that change.

Letladi: I've updated the pull request 👍🏼

Issue #10075, opened 2019-06-09
`super` method not present in type declarations


jhpratt:
## Bug Report

**Current Behavior**
Calling `t.super()` results in a compile-time type error, as `super` isn't present in type definitions. It does exist in [the documentation](https://babeljs.io/docs/en/babel-types#super) and works as expected after compilation.

**Expected behavior**
The `super` method should be present in the type definitions.

**Babel Configuration (.babelrc, package.json, cli command)**
Not applicable.

**Environment**
- Babel version(s): v7.4.0
- Node/npm version: Node 12.0.0, npm 6.9.0
- OS: Ubuntu 19.04
- Monorepo: no
- How you are using Babel: CLI

As someone using TypeScript to minimize errors, the amount of casting is _extremely_ verbose. By itself, that's not an issue, but coupled with errors in type definitions (this along with #10065) are forcing the use of unsafe casts and `// @ts-ignore`.

nicolo-ribaudo:
I'm not a TS expert, but how could we write this type definition so that it is valid code?
```ts
export function super() {}
```

jhpratt:
@nicolo-ribaudo That should be sufficient. It's my understanding type definitions are auto-generated, though, so I'm not sure where the error is truly at.

nicolo-ribaudo:
No, `super` is a keyword.

tanhauhau:
I figured we can do
```
declare function _super(): Super;
export { _super as super }
```

JLHwung:
Fixed by #10098

Issue #8375, opened 2018-07-24
@babel/types TypeScript definitions is missing some exported functions


mgroenhoff:
## Bug Report

**Current Behavior**
I am trying to use the TypeScript definitions from `@babel/types` but I am missing some functions:
https://github.com/babel/babel/blob/a811cf78e19442bbc10b05be2076c330fb702087/packages/babel-types/src/index.js#L96-L111

**Environment**
- Babel version(s): 7.0.0-beta.54

xtuc:
Would you mind sending a PR?

mgroenhoff:
I am not really sure where to start TBH. I know those types are automatically generated but that's about it. Can you give me some directions on how to approach this?

brandonkal:
cloneNode is also missing from the type definitions

nicolo-ribaudo:
@brandonkal Added!

JLHwung:
Fixed by #10098

brandonkal:
Looking through my babel plugins, I found some more types that are missing:
```ts
declare module '@babel/core' {
  namespace types {
    type VisitorKeys = {
      [T in keyof AllNodes]: Extract<
        keyof AllNodes[T],
        {
          [Key in keyof AllNodes[T]]: AllNodes[T][Key] extends (
            | t.Node
            | t.Node[]
            | null)
            ? Key
            : never;
        }[keyof AllNodes[T]]
      >;
    };

    const VISITOR_KEYS: { [T in keyof VisitorKeys]: VisitorKeys[T][] };
    const FLIPPED_ALIAS_KEYS: {
      [T in keyof t.Aliases]: t.Aliases[T]['type'][];
    };

    function shallowEqual(actual: object, expected: object): boolean;
  }
}
```
I'll post this here as it is related, but I can open new issue if desired.

mgroenhoff:
Thanks all for taking this over!

PR #10098, opened 2019-06-14
fix typescript for babel-types

### LINKED ISSUE #10065: `decorators` field is not present on `ClassPrivateProperty` in type declarations or documentation, but is in AST
## Bug Report

**Current Behavior**
Neither type declarations nor documentation indicate the presence of a `decorators` field on a `ClassPrivateProperty` node, while it is present in the AST.

**Input Code**
[AST explorer](https://astexplorer.net/#/gist/866ff157d68b2f148c75365361774b80/abf6c150530abdc194b2af45fc19ffd90fbcddb7)

```js
class Foo {
  @Decorator
  #prop = '';
}
```

**Expected behavior/code**
The `decorators` field should be present in both documentation and type declarations.

Just needs private class fields and decorators enabled, as indicated by link above.

**Environment**
- Babel version(s): v7.4.0
- Node/npm version: Node 12.0.0, npm 6.9.0
- OS: Ubuntu 19.04
- Monorepo: no
- How you are using Babel: CLI

tanhauhau:
| Fixed Issues?            | Fixes #10065, #10075, #8375
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-types/scripts/generators/flow.js
+++ b/packages/babel-types/scripts/generators/flow.js
@@ -126,6 +126,7 @@ lines.push(
   `declare function validate(n: BabelNode, key: string, value: mixed): void;`,
   `declare function clone<T>(n: T): T;`,
   `declare function cloneDeep<T>(n: T): T;`,
+  `declare function cloneNode<T>(n: T, deep?: boolean): T;`,
   `declare function removeProperties<T>(n: T, opts: ?{}): void;`,
   `declare function removePropertiesDeep<T>(n: T, opts: ?{}): T;`,
   `declare type TraversalAncestors = Array<{
--- a/packages/babel-types/scripts/generators/flow.js
+++ b/packages/babel-types/scripts/generators/flow.js
@@ -99,6 +99,14 @@ for (const type in t.NODE_FIELDS) {
         ", "
       )}): ${NODE_PREFIX}${type};`
     );
+  } else {
+    const functionName = toFunctionName(type);
+    lines.push(
+      `declare function _${functionName}(${args.join(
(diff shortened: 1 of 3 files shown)


jhpratt:
LGTM. I'd obviously prefer Babel used TS over Flow, as it's far more popular, but this PR should fix the issues I've found, at least.

tanhauhau:
> Looking at CI, there's one instance where you put object instead of Object for the type declaration.

Oops, because I copied over from `generators/typescript` 🙈

packages/babel-types/scripts/generators/flow.js
```diff
@@ -99,6 +99,14 @@ for (const type in t.NODE_FIELDS) {
         ", "
       )}): ${NODE_PREFIX}${type};`
     );
+  } else {
+    const functionName = toFunctionName(type);
+    lines.push(
+      `declare function _${functionName}(${args.join(
```
nicolo-ribaudo: Maybe this should be `export declare function ${type}`? They always start with an uppercase letter (thus are valid identifiers), and it would match https://github.com/babel/babel/blob/6467cd1e27ffe2a3a1cfe97bc8c67b407cb1e7c4/packages/babel-types/src/builders/generated/index.js

tanhauhau: @nicolo-ribaudo what do you think?

nicolo-ribaudo:
CircleCI failure is unrelated.

PR #10099, opened 2019-06-15
Disallow "let" as name at lexical bindings

g-plane:
| Fixed Issues?            |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 

Spec: https://www.ecma-international.org/ecma-262/9.0/#sec-let-and-const-declarations-static-semantics-early-errors


packages/babel-parser/src/parser/statement.js
```diff
@@ -1022,6 +1031,22 @@ export default class StatementParser extends ExpressionParser {
     );
   }
 
+  checkVarIdHasLet(id: ?N.Pattern): ?N.Identifier {
```
kaicataldo: I wonder if this would be clearer if it returned a boolean. Maybe something like `isValidVarName`?

g-plane: We need the identifier to report the error location.

kaicataldo: I guess it could return a tuple, but then I don’t know if that’s clearer or not.

packages/babel-parser/src/parser/statement.js
```diff
@@ -1014,6 +1014,15 @@ export default class StatementParser extends ExpressionParser {
       this.unexpected(null, "let is disallowed as a lexically bound name");
```
nicolo-ribaudo: We can now remove this, right?

packages/babel-parser/src/parser/statement.js
```diff
@@ -1014,6 +1014,15 @@ export default class StatementParser extends ExpressionParser {
...
+    if (kind === "const" || kind === "let") {
+      const invalid = this.checkVarIdHasLet(decl.id);
+      if (invalid) {
+        this.raise(
+          invalid.start,
+          "'let' is not allowed to be used as a name in 'let' or 'const' declarations.",
+        );
+      }
+    }
     this.checkLVal(
```
nicolo-ribaudo: If it works, I would prefer to move the check to `checkLVal`, to avoid looping over the destructuring tree twice.
We can check if `bindingType === BIND_LEXICAL` when checking identifiers, or pass an additional `disallowLetBinding` parameter to that function.

packages/babel-parser/src/parser/statement.js
g-plane: Sure!

g-plane:
@nicolo-ribaudo Are there anything else that should be changed?

nicolo-ribaudo:
I'm surprised that this didn't change any parser Test262 result

PR #10102, opened 2019-06-17
add bigint-transform-bigint

xtuc:
Previsouly only the BigInt literal syntax was support by Babel
(babel-plugin-syntax-bigint). However, emiting this syntax might break
other tool's parsers.

This change adds a package that does the trivial transformation between
BigInt literal syntax (1n) to BigInt function call (BigInt("1)).

Polyfilling the BigInt Object is up to the user.

| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-transform-bigint/.npmignore
+++ b/packages/babel-plugin-transform-bigint/.npmignore
@@ -1,0 +1,3 @@
+src
+test
+*.log
--- a/packages/babel-plugin-transform-bigint/README.md
+++ b/packages/babel-plugin-transform-bigint/README.md
@@ -1,0 +1,19 @@
+# @babel/plugin-transform-bigint
+
+> Transform BigInt literal syntax into calls
+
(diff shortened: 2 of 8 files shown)


nicolo-ribaudo:
I think that this plugin will very likely lead to confusion.
Users expect that `@babel/plugin-transform-*` plugin will make their code supported by older browsers. This plugin doesn't change anything about browser support from the input code to the output code.
If an engine doesn't support `1n + 2n`, it won't support `BigInt("1") + BigInt("2")`, no matter of which polyfill you load.

BigInts are stage 4 now, and other tools should start supporting them. ESLint can support them via `babel-eslint`, and `acorn` already supports them (https://github.com/acornjs/acorn/commit/09fe6e471ddd2be8f2bc1297e4ec8ed9507e2c52)

nicolo-ribaudo:
@xtuc I'm closing this PR per the above comment. If you still think that it should be discussed, feel free to re-open it.

PR #9057, opened 2018-11-21
preset-env: adds dynamic import flag for caller

thiagoarrais:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes babel/babel-loader#515
| Patch: Bug Fix?          | ❌
| Major: Breaking Change?  | ❌
| Minor: New Feature?      | 👍
| Tests Added + Pass?      | 👍
| Documentation PR Link    | 
| Any Dependency Changes?  | 👍
| License                  | MIT

This interprets the `supportsDynamicImport` caller flag. It was introduced in babel-loader by babel/babel-loader#660 and signals that the target env supports dynamic imports natively.

When this gets merged, it should no longer be necessary to manually include the dynamic import syntax plugin in the webpack config. Thus fixing babel/babel-loader#515.

Needs for conclusion:

- [ ] Docs for new flag

References: babel/babel-loader#714 (defunct), babel/babel-loader#515

PR #10109, opened 2019-06-18
Add @babel/plugin-proposal-dynamic-import to @babel/preset-env

nicolo-ribaudo:
| Fixed Issues?            | Closes babel/babel-loader#515, closes #9057 (by @thiagoarrais), ref https://github.com/babel/babel-loader/pull/714
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Follow up to https://github.com/babel/babel/pull/9552: that PR cretes the new plugin; this PR adds it to `@babel/preset-env`.

I'm really thankful that @loganfsmyth added the `supportsDynamicImport` flag to [`babel-loader`](https://github.com/babel/babel-loader/blob/7d8500ced72046d4c5c1dcdc7a9c0016d7e6c15b/src/injectCaller.js#L13) and [`rollup-plugin-babel`](https://github.com/rollup/rollup-plugin-babel/blob/8357816ddacf2881b63a8f822c0c93bc644c7a79/src/index.js#L25) a long time ago :slightly_smiling_face:

cc @Andarist Is this ok for `rollup-plugin-babel`?


Andarist:
Well - I guess it’s hard to tell if this can break somebody’s expectations, considering in how many ways babel configs may be written. But I wouldn’t stop this because of that because i think possibility of breakage is low.

nicolo-ribaudo:
I think that this could only break only if someone disables the modules transform plugin using `excludes` instead of `modules: false` (I don't know if it is actually possible?)

packages/babel-preset-env/src/index.js
```diff
@@ -124,16 +128,37 @@ export default declare((api, opts) => {
...
+      shouldTransformDynamicImport &&
+      shouldTransformESM &&
+      modules !== "umd"
+    ) {
+      plugins.push([getPlugin("proposal-dynamic-import"), { loose }]);
+    } else {
+      if (shouldTransformDynamicImport) {
+        console.warn(
+          "Dynamic import can only be supported when transforming ES modules" +
+            " to amd, commonjs or systemjs. Only the parser plugin will be enabled.",
```
existentialism: Nit: should we use proper capitalization here (AMD, CommonJS, SystemJS)?

nicolo-ribaudo:
```suggestion
            " to AMD, CommonJS or SystemJS. Only the parser plugin will be enabled.",
```

nicolo-ribaudo:
Merged as ef3f555be9ce1ef780e05cd1594a98e9567a1b80

klimashkin:
Why do we show `Dynamic import can only be supported when transforming ES modules` warning in case of `umd` target even if there is no `import()` in the code?

nicolo-ribaudo:
Could you open a new issue showing your config and code?

PR #10115, opened 2019-06-21
WIP: Eventual Send (wavy dot)

michaelfig:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    |

Add `@babel/parser` support for infix bang syntax, activated by:
```javascript
    parser.parse(s, {
      plugins: [['infixBang', { resolverObject: 'Promise' }]]
    })
```

where `resolverObject` can be any identifier whose `resolve()` method creates a `Promise` which understands the remote promise protocol.

```javascript
Foo!bar(arg)!baz![bingo]
```

becomes:

```javascript
Promise.resolve(Promise.resolve(Promise.resolve(Foo).post('bar', [arg]).get('baz')).get(bingo)
```

This PR will be fleshed out with more documentation when there is a stage0 proposal for it.


michaelfig:
Need to implement no-newline before bang (to prevent ASI from changing the meaning of existing programs).

loganfsmyth:
Did you perhaps submit this to the wrong repo? I see that you changed the package name, and as-is I don't think it makes sense for us to have a WIP patch open when there is no proposal.

michaelfig:
The stage 0 proposal is coming.  I thought this PR would be a good way to have other collaborators discuss the changes until it would no longer be WIP.  If that's not acceptable, I will close this PR and find a forum elsewhere.

jridgewell:
@michaelfig: I see you changed this to an agoric parser. Does that mean Mark will be presenting this at the next meeting?

loganfsmyth:
Assuming this is something that is being presented at the next meeting, I don't necessarily mind having the PR open. We definitely require something to have a stage 0 proposal before landing it in the parser. That said, we also generally want to be very cautious about landing things too early, because once things land, we can't change it and need to support it, even if there are further later changes to the proposal syntax that then might require the introduction of even more parser plugins.

michaelfig:
@jridgewell, possibly.  It depends on his schedule.  I'm currently experimenting with the parser, but the details could change significantly based on what is actually proposed.

michaelfig:
@loganfsmyth, I don't want to merge this PR until it actually satisfies all parties and doesn't look like it will change.  Until then, WIP!

loganfsmyth:
Okie doke, sounds good.

michaelfig:
Flaky tests?
```
 FAIL  packages/babel-plugin-transform-typescript/test/index.js
  ● Test suite failed to run

```

michaelfig:
This proposed syntax has been changed from infix bang (**!**) to wavy dot (**~.**), as documented in the [eventual send syntax proposal](https://github.com/Agoric/proposal-wavy-dot).

PR #10120, opened 2019-06-24
Avoid removing class fields using flow in `plugin-transform-flow-strip-types` plugin

### LINKED ISSUE #10039: Should not remove class fields when using flow
If the flow plugin runs before the class properties one (or if the class properties plugin isn't used), we remove uninitialize class fields:

```js
class Foo { x: string }
```

currently becomes
```js
class Foo {}
```

while it should be
```js
class Foo { x }
```

If someone still wants the old behavior, they can either use flow comments (as [offically recommended](https://github.com/facebook/flow/issues/6811#issuecomment-496808192)), or use the `ignoreUninitialized` option which will soon be added to the class properties plugin (https://github.com/babel/babel/pull/9141).

---

Is there anyone who wants to contribute for the first time to Babel? :slightly_smiling_face: 

> **NOTE** :warning::warning::warning: Since this is a breaking change, we won't merge the PR for a long time (before Babel 8), and it might seem that we are ignoring it.

1. Write a comment there to know other possible contributors that you are working on this bug.
1. Fork the repo
2. Run `git clone https://github.com/<YOUR_USERNAME>/babel.git && cd babel`
3. Run `yarn && make bootstrap`
4. Wait :hourglass_flowing_sand: 
5. Run `make watch` (or `make build` whenever you change a file)
6. Change the wrong `output.js` files in the `tests/fixture` folder of `@babel/plugin-transform-flow-strip-types` (if there are any)
7. Update the code!
8. `yarn jest [name-of-the-package-to-test]` to run the tests
9. If it is working, run `make test` tu run all the tests
10. Run `git push` and open a PR!

pajaydev:
| Fixed Issues?            | `Fixes #10039 ` 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Currently when using flow plugin `transform-flow-strip-types` removes the uninitialize class fields, after this change we won't remove the uninitialize class fields. However they can remove using `ignoreUninitialized` option mentioned here https://github.com/babel/babel/pull/9141


nicolo-ribaudo:
Can you add the `transform-class-properties` plugin to the failing test? Also, probably you'll need to update the output file: you can delete it and it will be re-generated.

pajaydev:
@nicolo-ribaudo okay cool. Let me try

pajaydev:
@nicolo-ribaudo  Thanks. I fixed it.

pajaydev:
@nicolo-ribaudo  Kindly review these changes.

nicolo-ribaudo:
This PR looks good, but we will have to wait until v8

packages/babel-plugin-transform-flow-strip-types/test/fixtures/regression/class-prop-types/options.json
```diff
@@ -1,7 +1,7 @@
 {
   "plugins": [
     "transform-flow-strip-types",
-    "transform-classes",
+    "@babel/plugin-proposal-class-properties",
```
nicolo-ribaudo: This should just be `proposal-class-properties`, to use the package from the monorepo and not from `node_modules`

nicolo-ribaudo:
According to our e2e tests, this breaking change affects Babel itself :upside_down_face: 

I'll fix it in `master` before merging this PR.

PR #10126, opened 2019-06-25
fix exportKind declaration in babel-types

zxbodya:
`babel-types`:
- added `exportKind` for `ExportNamedDeclaration`
- fixed `exportKind` type for `DeclareExportAllDeclaration`

~`babel-parser`:~
~- updated types in to include `exportKind=let`~

| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    |
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-types/src/definitions/es2015.js
+++ b/packages/babel-types/src/definitions/es2015.js
@@ -261,7 +261,7 @@ defineType("ExportNamedDeclaration", {
       validate: assertNodeType("StringLiteral"),
       optional: true,
     },
-    exportKind: validateOptional(assertOneOf("type", "value", "let")),
+    exportKind: validateOptional(assertOneOf("type", "value")),
   },
 });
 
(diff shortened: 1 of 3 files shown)


packages/babel-types/src/definitions/es2015.js
```diff
@@ -260,6 +261,7 @@ defineType("ExportNamedDeclaration", {
       validate: assertNodeType("StringLiteral"),
       optional: true,
     },
+    exportKind: validateOptional(assertOneOf("type", "value", "let")),
```
nicolo-ribaudo: When is it `let`?

zxbodya: found it in astexplorer when trying:
`export let something=1`

but can not reproduce it now… - will remove it

nicolo-ribaudo: Yeah, that should be a `value` export.

PR #10127, opened 2019-06-25
Bump compat-table and updating preset-env mappings

### LINKED ISSUE #6605: Block scope transform not included when targeting Safari 10
> Issue originally reported by @ssilve1989 in babel/babel-preset-env#411

Specifying a target of `"safari >= 10"` doesn't use the `transform-es2015-block-scoping` plugin. Yet Safari 10 does need this because of [this](https://bugs.webkit.org/show_bug.cgi?id=159270) bug.

Ex:
```javascript
{
  let a = 3
}

let a = 3
``` 
is an error in Safari 10.1.2 resulting in `SyntaxError: Can't create duplicate variable that shadows a global property: 'a'`.

I've had to manually include this plugin in the `include` option to get Safari working.

existentialism:
| Fixed Issues?            | Fixes #6605 
| Documentation PR Link    | 
| Any Dependency Changes?  |

All credit goes to @wessberg for making this happen!


packages/babel-preset-env/data/built-in-modules.json
```diff
@@ -6,7 +6,7 @@
     "safari": "10.1",
     "opera": "48",
     "ios_saf": "10.3",
-    "and_chr": "71",
-    "and_ff": "64"
+    "and_chr": "74",
```
nicolo-ribaudo: Is this technically a breaking change? A script which might have worked on and_chr 72 now it won't work anymore.

existentialism: Reverted this change and will push a new PR that updates the script to only update if it results in an earlier (or new) target

PR #10146, opened 2019-06-29
Inject core-js@3 imports in Program:exit instead of on post()

### LINKED ISSUE #10142: How to transform output of presets?
I want to transform the `core-js` imports added by `@babel/preset-env`, but plugins always run before presets, as such I don't have access to those imports. Is there any other way to achieve this?

nicolo-ribaudo:
| Fixed Issues?            | Fixes #10142
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Nodes injected during `post()` can't be visited again, so it should be avoided.

When reviewing this PR, disable whitespaces in the diff.


Jessidhia:
I wonder how this works with the combination of builtins-usage and the commonjs transform.

The commonjs transform uses `Object.defineProperty` and it's possible preset-env detects it as something to be polyfilled.

nicolo-ribaudo:
In that case, after modules have been transformed, useBuiltIns will inject a plain require call to the polyfill.
I believe we have a test for this, let me check.

packages/babel-preset-env/test/fixtures/corejs3/entry-entries-modules-chrome-71/output.mjs
```diff
@@ -1,2 +1,2 @@
-import "core-js/modules/es.object.from-entries";
-import "core-js/modules/esnext.string.replace-all";
+import 'core-js/modules/es.object.from-entries';
```
existentialism: These quote changes seem unintended/unnecessary?

nicolo-ribaudo: It's because Babel, when possible, reuses the original quotes.
The old implementation removed this import statements and replaced with a new copy of them, while this implementation avoids replacing an import which is already ok.

existentialism: :+1:

packages/babel-preset-env/src/polyfills/corejs3/entry-plugin.js
```diff
@@ -48,6 +49,20 @@ export default function(
 
   const available = new Set(getModulesListForTargetVersion(corejs.version));
 
+  function shouldReplace(source, modules) {
+    if (!modules) return false;
+    if (
+      // Avoid replaing an import with itself, to avoid infinite loops.
```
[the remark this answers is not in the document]
JLHwung: 🤔Is Brian's suggestion resolved by mistake?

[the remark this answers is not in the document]
nicolo-ribaudo: Oh I accidentally reverted it when rebasing

PR #10148, opened 2019-06-29
V8intrinsic syntax plugin

### LINKED ISSUE #10104: Add optional parsing support for V8 intrinsics
## Feature Request

V8 exposes [a number of intrinsics](https://cs.chromium.org/chromium/src/v8/src/runtime/runtime.h?rcl=c843aa75886abbd9aa4ea65818ece3d661ab2ed3) when the `--allow-natives-syntax` flag is set. Examples include:

```js
%DebugPrint(object);
%PrepareFunctionForOptimization(fn);
```

Because we generally don't want to encourage the use of these intrinsics in production code, they rely on non-standard syntax; standard JS parsers (including Babel's) consider this to be a modulo operator without a left-hand side. [Here's how V8 implements parsing support based on the flag.](https://cs.chromium.org/chromium/src/v8/src/parsing/parser-base.h?l=1785-1789&rcl=5e14bd7ec82b9021667402cc5c92310230bea48c)

However, some research projects intentionally make use of these V8-specific intrinsics, and of course the V8 code base itself makes use of them extensively in JS tests.

It would be great if Babel could optionally support this syntax extension, similar to how it supports JSX and Flow syntax.

If this seems more feasible as a Babel parser plugin, that'd be a good solution too! I just wanted to discuss this proposal with the core team first.

I'd be happy to write some docs if the team decides this makes sense in babel/parser core.

JLHwung:
| Fixed Issues?            | Fixes #10104  
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | ✔️
| Tests Added + Pass?      | ✔️
| Documentation PR Link    | 

Authored a new syntax plugin `v8intrinsic` to support V8 CallRuntime syntax
```
CallRuntime ::
  '%' Identifier Arguments
```

The original implementation in V8 parser `peek()` twice to check whether a `modulo` token is immediately followed by an identifier and a `parenL` token.

As in Babel `lookahead()` does not support multiple steps, here we implicitly track the execution branch inside `parseSubscript` to see if `parseCallExpressionArguments` is immediately called in the first round of `parseExprSubscripts`. If so we conclude that the modulo token is a valid V8Intrinsic marker.

It is also interesting to note that a V8Intrinsic call inside a _new_ expression does not raise a SyntaxError in V8. i.e.
```
node --allow-natives-syntax -e "new %DebugPrint(null)"
// throws TypeError: (intermediate value) is not a constructor

node -e "new %DebugPrint(null)"
// throws SyntaxError: Unexpected token %
```
@mathiasbynens I don't think there are any V8 intrinsics supposed to work with _new_ expression, aren't there?

If so, it should be reasonable to prevent v8intrisic from new expression, as is the behavior of this plugin. Otherwise we still need to take care on the _new_ expression.

### Action Items
- [x] Add `v8intrinsic` to babel-parser typings after #10259 gets merged.
- [x] Document updates: https://github.com/babel/website/pull/2074


packages/babel-parser/src/plugins/v8intrinsic.js
```diff
@@ -0,0 +1,48 @@
...
+      }
+    }
+
+    /* ============================================================ *
+     * parser/expression.js                                         *
+     * ============================================================ */
+
+    createIdentifier(): N.Identifier {
+      const identifier = super.createIdentifier(...arguments);
+      if (this.state.V8IntrinsicStart !== undefined) {
```
JLHwung: Are there any way to document the type of `V8IntrinsicStart` be a `Number` or `Undefined` ?

nicolo-ribaudo: You should specify it at https://github.com/babel/babel/blob/master/packages/babel-parser/src/tokenizer/state.js

Nit: I know that it's name is V8, but could you start this variable name with `v8`? :sweat_smile: We only use upper-first names for classes/interfaces.

packages/babel-parser/test/fixtures/v8intrinsic/expression/await-expression/input.js
```diff
@@ -0,0 +1 @@
+async () => { await %StringParseInt("42", 10) }
```
nicolo-ribaudo: Can you also add a test for `yield`?

```js
function* fn() {
  yield %StringParseInt("42", 10);
}
```

We should ensure that it isn't parsed as
```js
function* fn() {
  (yield) % (StringParseInt("42", 10));
}
```

JLHwung:
> I was thinking that maybe we can remove this.state.V8IntrinsicStart and put all the parsing logic in a single method

I agree, actually my first attack is on rewriting `parseIdentifier` but I find out it would mess up with the token position internal state. Apparently overriding `parseExprAtom` is way better than my previous approach. I have pushed the rewritten version.

> Can you also add a test for yield?

Great observation! The `parseYield` will check if a token can `startsExpr`, if not it ends up with an empty `yield` statement, which fails under this case where `%` can actually start an expression. I would try to figure out a way to do this in the plugin land.

packages/babel-parser/src/plugins/v8intrinsic.js
JLHwung: Good to know the naming convention here. Resolved as we have removed `V8IntrinsicStart` internal state.

mathiasbynens:
Thanks for your work on this!

> @mathiasbynens I don't think there are any V8 intrinsics supposed to work with new expression, aren't there?

That's correct. [Here's how V8 implements parsing support based on the flag.](https://cs.chromium.org/chromium/src/v8/src/parsing/parser-base.h?l=1785-1789&rcl=5e14bd7ec82b9021667402cc5c92310230bea48c)

packages/babel-parser/src/plugin-utils.js
```diff
@@ -89,13 +96,15 @@ import flow from "./plugins/flow";
 import jsx from "./plugins/jsx";
 import typescript from "./plugins/typescript";
 import placeholders from "./plugins/placeholders";
+import v8instrinsic from "./plugins/v8intrinsic";
```
mathiasbynens: typo: `s/v8instrinsic/vv8intrinsic/`

packages/babel-parser/test/fixtures/v8intrinsic/_errors/not-in-call-expression/input.js
```diff
@@ -0,0 +1,2 @@
+const i = %DebugpPrint;
```
mathiasbynens: typo: `%DebugPrint`

(not that it really matters for the purposes of the test case)

packages/babel-parser/test/fixtures/v8intrinsic/options.json
```diff
@@ -0,0 +1,3 @@
+{
+  "plugins": ["v8instrinsic"]
```
mathiasbynens: typo: `v8intrinsic`

packages/babel-parser/test/fixtures/v8intrinsic/_errors/not-in-call-expression/input.js
JLHwung: Ahh, good catch. Fixed and pushed.

packages/babel-parser/test/fixtures/v8intrinsic/options.json
JLHwung: Done in a1a4558.

packages/babel-parser/src/plugin-utils.js
JLHwung: Done in a1a4558.

JLHwung:
> That's correct. Here's how V8 implements parsing support based on the flag.

The implementation of V8 parsing `V8Intrinsic` will lead to the fact that `new %StringParseInt("42", 10)` will _not_ throw a syntax error, instead it throws a runtime type error. 

However as I have rewritten the plugin and now it closely follow the way how V8 parse. I have added this `newExpression` test case to the success test suites.

JLHwung:
I've talked to Nicolo offline a bit. Here we add `{ startsExpr: true }` configuration to `parser/types` since

- It is only used in `parseYield`. The impact is under control.
- There have been some plugin-defined tokens inside this file, i.e. `braceBarL`, `braceBarR`. It is acceptable to place plugin-defined token configurations here.

packages/babel-parser/test/fixtures/v8intrinsic/expression/await-expression/input.js
JLHwung: Test case is added and passed.

jridgewell:
@nicolo-ribaudo: Is this waiting for a release, or can it be merged?

nicolo-ribaudo:
We just released v7.5 so this is waiting for the next minor, but if it is urgent we could do it earlier than the usual time between two minor releases.

packages/babel-types/src/definitions/misc.js
```diff
@@ -19,3 +23,12 @@ defineType("Placeholder", {
     },
   },
 });
+
+defineType("V8IntrinsicIdentifier", {
```
nicolo-ribaudo: I just realized that we should add this to the allowed `callee`s of `CallExpression`

PR #10157, opened 2019-07-03
Fix incorrect usage of $o instead of %o in debug

ChlorideCull:
It would end up printing the literal "$o" instead of the directory name.

| Fixed Issues?            |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

When running with debug I saw that the root config loader wouldn't print the directory properly, instead printing `$o`. In other words, it would print the following:

> babel:config:loading:files:configuration Found root config 'babel.config.js' in $o. C:\Src\Proj +0ms

Decided to do a quick fix.


nicolo-ribaudo:
Thanks!

PR #10158, opened 2019-07-03
Package core-js@3 support data in @babel/preset-env

### LINKED ISSUE #10152: babel-cli postinstall spam
## Bug Report

**Current Behavior**
When installing babel-cli, a transitive dependency is spamming the installation log.  Consider removing `core-js` as a dependency.

```
$ npm install
...
> core-js@2.6.9 postinstall /usr/local/google/home/tildechris/src/tildechris-expeditors/node_modules/core-js
> node scripts/postinstall || echo "ignore"

Thank you for using core-js ( https://github.com/zloirock/core-js ) for polyfilling JavaScript standard library!

The project needs your help! Please consider supporting of core-js on Open Collective or Patreon: 
> https://opencollective.com/core-js 
> https://www.patreon.com/zloirock 

Also, the author of core-js ( https://github.com/zloirock ) is looking for a good job -)


> core-js-pure@3.1.4 postinstall /usr/local/google/home/tildechris/src/tildechris-expeditors/node_modules/core-js-pure
> node scripts/postinstall || echo "ignore"

Thank you for using core-js ( https://github.com/zloirock/core-js ) for polyfilling JavaScript standard library!

The project needs your help! Please consider supporting of core-js on Open Collective or Patreon: 
> https://opencollective.com/core-js 
> https://www.patreon.com/zloirock 

Also, the author of core-js ( https://github.com/zloirock ) is looking for a good job -)
```

**Expected behavior/code**

Installation without error should not produce any output by default.

**Environment**
- Babel version(s): [e.g. 6.26.0]
- Node/npm version: [e.g. node v11.10.0/npm 6.7.0]
- OS: [Debian GNU/Linux rodete]
- Monorepo: [e.g. no]
- How you are using Babel: [`cli`]

**Possible Solution**
Consider removing `core-js` as a dependency.

Please see below for more context:

https://github.com/zloirock/core-js/issues/548

tl;dr - The core-js maintainer isn't willing to revert this regression.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #10152
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This has the effect of removing the dependency on `core-js-compat`,
which in  turn depends on `core-js-pure`.
For this reason, `@babel/preset-env` users **who don't use `core-js`** won't
see the postinstall message anymore. `@babel/preset-env` users who
use `core-js` (using the `useBuiltIns` option) **will still see** the postinstall
message, because they need to install `core-js` anyway.

This behavior matches `@babel/plugin-transform-runtime`: people who
only use it for the helpers don't see the postinstall message, while
people who also use it for polyfills do, because they have to install
`@babel/runtime-corejs3` which depends on `core-js-pure`.


zloirock:
So, on each new data in `core-js-compat` or, for example, a new release of packages like `electron-to-chromium`, you should regenerate this data and publish a new release of `preset-env`. Or users will live with obsolete data - like now with obsolete data for Electron - the actual version is `1.3.184`, the version in dependencies bound to `1.3.113`. In this case, since compat data will be obsolete almost all time anyway, that will kill the main reasons for supporting `core-js-compat` for `preset-env` for me. I think in this case for me it would be better just create and maintain plugins for polyfilling in `babel` / `typescript` / `webpack` on `core-js` side.

nicolo-ribaudo:
@zloirock I hope you understand that we aren't hiding the message from core-js users (and we don't want to). We already regenerate our data often (from compat-table), thus the core-js@3 data would never be too much out of date (we would update them on every release).

I checked in core-js-compat and the message is logged because it depends on `core-js-pure`. `core-js-pure` is only used in `index.js`, a file which Babel doesn't use.

zloirock:
However, anyway it will be outdated and bound to `preset-env` version. The current architecture was done to prevent this problem. I don't think that some lines in the installation log are a big price for always actual data. If it will be done, `core-js-compat` will not be mantained in the current form anymore.

nicolo-ribaudo:
Whether or not it's a big price, I don't agree that it is a price that should be paid by people not using `core-js` and not using any code which depends on `core-js`.
I accept that core-js users (either directly or via `useBuiltIns` or `@babel/transform-runtime`) should pay this cost: you are the `core-js` maintainer, this decision is up to you. But this cost shouldn't be paid by every Babel users, since Babel internally doesn't use `core-js` and they don't use `core-js`.

What I'm asking is to replace this line in `core-js-compat`:
https://github.com/zloirock/core-js/blob/1508eb09a8105afd4ebecde3d41715305ae71305/packages/core-js-compat/index.js#L89
with
```js
const availableModules = new Set(getModulesListForTargetVersion(version));
modules = modules.filter(name => availableModules.has(name));
```
(which, by the way, would be more performant since it avoids creating two unecessary `Set`s and using array spread)

That's the only usage of `core-js-pure` in `core-js-compat`, and if you could remove it this PR would stop having any reason to exist.

zloirock:
You could add a PR to `core-js` which replaces this line, it's not something principled for me. However, I can't give any guarantees that it will not be added in the future again if it will be required. It's a much simpler and definitely better approach than the current PR, so I can't understand why you added the current PR at first. Anyway, it'll affect the existence of `core-js` postinstall output just ~1% of users.

> I have talked with other people from the team, and we have decided...

Already means a lot since it's a principled architectural question of polyfilling in Babel and it's news for me, so I will seriously think about the approach from my first comment in this thread.

nicolo-ribaudo:
> You could add a PR to core-js which replaces this line, it's not something principled for me. However, I can't give any guarantees that it will not be added in the future again if it will be required. It's a much simpler and definitely better approach than the current PR, so I can't understand why you added the current PR at first.

After reading https://github.com/zloirock/core-js/issues/582#issuecomment-507800137, I got the impression that you wouldn't be willing to limit the postinstall message in any case, so I tried to a workaround on our side. I apologize for not having asked for your opinion first.

> I think in this case for me it would be better just create and maintain plugins for polyfilling in babel / typescript / webpack on core-js side.

I think that we might have compatible opinion on this. As you probably know, I'm working on re-implementing how polyfills work in Babel, and allow any polyfill to provide an interop layer for Babel.
While implementing it, I'm also created polyfill providers for `core-js@2` and `core-js@3` for two reasons:
1) it's impossible for me to create support for polyfill providers without a polyfill provider to test it;
2) you have done a lot of work to add core-js support to Babel, and I didn't want to make you do that same work again.
When it's ready to be published (the polyfill injector isn't fully working yet), I plan to give you full access to the core-js providers repo/npm.

I'm going to close this PR now, and follow up with a PR to the `core-js-compat` package.

zloirock:
> After reading zloirock/core-js#582 (comment), I got the impression that you wouldn't be willing to limit the postinstall message in any case

It's about the presence of the `postinstall` message on `core-js` / `core-js-pure`, but not about depending `core-js-compat` from `core-js-pure`.

nicolo-ribaudo:
Yeah, sorry for misunderstanding.

zloirock:
> As you probably know, I'm working on re-implementing how polyfills work in Babel, and allow any polyfill to provide an interop layer for Babel.

Seems to need to discuss it since I do not completely agree with your proposed approaches - in some cases, it's overcomplicated, in some cases too simplified.

> you have done a lot of work to add core-js support to Babel, and I didn't want to make you do that same work again.

In that work, polyfilling plugins were prepared for extraction from `preset-env` and unification. Since, if I understood correctly, providers proposal supports usual visitors, I don't think that is a big difference will they be plugins or providers.

PR #10161, opened 2019-07-04
Improves the logic to import objects in helpers

ifsnow:

---------

In the case of `react-native`,

```js
import React, { PureComponent } from 'react';
```

Compiled as below.

```js
var _react = _interopRequireWildcard(_$$_REQUIRE(_dependencyMap[7], "react"));
```

The below logic(looping) is performed whenever `_interopRequireWildcard` with `react` is called.

```js
var newObj = {};

if (obj != null) {
  for (var key in obj) {
    if (Object.prototype.hasOwnProperty.call(obj, key)) {
      var desc = Object.defineProperty && Object.getOwnPropertyDescriptor ? Object.getOwnPropertyDescriptor(obj, key) : {};

      if (desc.get || desc.set) {
        Object.defineProperty(newObj, key, desc);
      } else {
        newObj[key] = obj[key];
      }
    }
  }
}
```

Generally, We have a lot of components that imports `react` or `react-native`. I don't think it's good to check the same object repeatedly on every call. so, I thought it would be better to put cache logic in here.

I counted the number of how many loops are executed to start my production-level App. It may feel like, but it feels like my app is a little faster. :)

|Version|Count
|Current | 11,032|
|New (Cache) | 128|

I don't fully understand `Babel` architecture. Please let me know if I have any mistakes.

I hope this helps.


packages/babel-helpers/src/helpers.js
```diff
@@ -588,28 +588,35 @@ helpers.interopRequireDefault = helper("7.0.0-beta.0")`
 `;
 
 helpers.interopRequireWildcard = helper("7.0.0-beta.0")`
+  var REQUIRE_WILD_CACHE = typeof WeakMap === 'function' ? new WeakMap() : undefined;
```
nicolo-ribaudo: This might have problems, in case `_interopRequireWildcard` is used before this variable declaraiton is run. To be safe, I would prefer that it is initialized inside the function:

```js
var REQUIRE_WILD_CACHE;

export default function _interopRequireWildcard(obj) {
  if (obj && obj.__esModule) {
    return obj;
  }

  if (REQUIRE_WILD_CACHE) {
    if (REQUIRE_WILD_CACHE.has(obj)) return REQUIRE_WILD_CACHE.get(obj);
  } else if (typeof WeakMap === "function") {
    REQUIRE_WILD_CACHE = new WeakMap();
  }

  ...
```

nicolo-ribaudo:
Note that this optimization would only work when using `@babel/plugin-transform-runtime`, otherwise the map is re-created in every file.

ifsnow:
@nicolo-ribaudo Thanks for your feedback. I will try something better based on what you said.

ifsnow:
@nicolo-ribaudo I've modified it to work well in the case you mentioned. Please let me know if there's anything else I need to consider.

packages/babel-helpers/src/helpers.js
```diff
@@ -588,28 +588,54 @@ helpers.interopRequireDefault = helper("7.0.0-beta.0")`
 `;
 
 helpers.interopRequireWildcard = helper("7.0.0-beta.0")`
+  function _createRequireWildcardCache(obj) {
+    if (typeof global._REQUIRE_WILD_CACHE === 'undefined') {
+      global._REQUIRE_WILD_CACHE = typeof WeakMap === 'function' ? new WeakMap() : null;
+    }
+
+    const cache = global._REQUIRE_WILD_CACHE;
```
nicolo-ribaudo: While `global` should always work in node, I would prefer to keep helpers self-contained.

I'd like to see this as:
```js
function _createRequireWildcardCache() {
  var map = typeof WeakMap !== "function" ? new WeakMap() : {
    has: function() { return false },
    set: function() {},
    get: function() { return null }
  };

  _createRequireWildcardCache = function() { return map };
  return map;
}

export default function _interopRequireWildcard(obj) {
  if (obj && obj.__esModule) {
    return obj;
  }

  var cache = _createRequireWildcardCache();
  if (cache.has(obj)) return cache.get(obj);

  // ...

  cache.set(obj, newObj);
  return newObj;
}
```

Since this disable the optimization when helpers are used inline, let's wait for @babel/babel's opinion.

ifsnow: @nicolo-ribaudo Your approach also looks pretty cool! But, the reason for using `global` is to keep the cache even if `@babel/plugin-transform-runtime` is not available.

Please let me know if babel team have an internal rule that does not use `global` or any other better approach.

Andarist: Using `global` referenced is often causing i.e. webpack to include polyfills for it when bundling for web (at least on 1-3 versions, most likely on v4 too, not sure about v5 - as I've heard about some changes regarding this logic in there). Certainly something to consider before merging this in.

zloirock: #9232

ifsnow:
@nicolo-ribaudo @Andarist @zloirock Thank you all for the feedback.

I realized that there might be no `global` object. https://github.com/webpack/webpack/blob/master/buildin/global.js#L16

In this case, I modified the cache to not work. I'm sure `cache` works well in modern development environments.

packages/babel-helpers/src/helpers.js
```diff
@@ -588,28 +588,46 @@ helpers.interopRequireDefault = helper("7.0.0-beta.0")`
 `;
 
 helpers.interopRequireWildcard = helper("7.0.0-beta.0")`
+  function _getRequireWildcardCache(obj) {
+    if (!global || typeof WeakMap !== 'function') {
```
ljharb:
```suggestion
    if (typeof global === 'undefined' || typeof WeakMap !== 'function') {
```
otherwise you’ll still trigger bundler’s polyfills

ifsnow:
@ljharb Thank you for pointing out what I did not think of.

packages/babel-helpers/src/helpers.js
```diff
@@ -588,28 +588,46 @@ helpers.interopRequireDefault = helper("7.0.0-beta.0")`
...
+    if (typeof global === 'undefined' || typeof WeakMap !== 'function') {
+      return null;
+    }
+
+    if (!global._REQUIRE_WILD_CACHE) {
+      global._REQUIRE_WILD_CACHE = new WeakMap();
+    }
+
+    return global._REQUIRE_WILD_CACHE;
+  }
```
nicolo-ribaudo: We can make it work when `global` is not defined but `@babel/runtime` is used (e.g. in rollup or webpack).
Also, could you rename the global var so that it is clear that comes from Babel?

```js
function _getRequireWildcardCache() {
  if (typeof WeakMap !== "function") return null;

  var cacheKey = "##__BABEL__REQUIRE__WILDCARD_CACHE";
  var map;
  if (typeof global !== "undefined") {
    if (global[cacheKey]) map = global[cacheKey];
    else map = global[cacheKey] = new WeakMap();
  } else {
    map = new WeakMap();
  }

  _getRequireWildcardCache = function() { return map };
  return map;
}
```

By doing so, multiple calls to `_getRequireWildcardCache` will return the same WeakMap even if `global` is not available.

ifsnow: @nicolo-ribaudo Your approach looks pretty cool. 

Come to think of it again, If we use function reassigning, there seems to be no reason to use `global`. I changed the implementation. what do you think about it?

nicolo-ribaudo: The global map would still be needed if someone has two different instances of this helper.
I prefer this as you currently implemented it (since it doesn't rely on `global`), even if it doesn't optimize 100% of the cases.

packages/babel-helpers/src/helpers.js
```diff
@@ -588,28 +588,43 @@ helpers.interopRequireDefault = helper("7.0.0-beta.0")`
...
+          } else {
+            newObj[key] = obj[key];
           }
         }
       }
-      newObj.default = obj;
-      return newObj;
     }
+    newObj.default = obj;
+    cache && cache.set(obj, newObj);
```
jridgewell: Nit: Use an if statement.

ifsnow: @jridgewell Thanks for your advice. I modified it.

ifsnow:
@nicolo-ribaudo @jridgewell @Andarist @zloirock @ljharb @cpojer Do you have plans to merge this PR?

I'm sure this cache feature makes React Native apps run faster. The following table shows the time spent running js bundle at startup.

| Device + Engine              | Current (ms) | Cache (ms) | Improving |
| -----------------------------|-------------:| ----------:|----------:|
| Android 6.0 (32bit) + JSC    | 3,182        | **2,990**      |6%🔻 |
| Android 7.0 (64bit) + JSC    | 1,118        | **1,089**      | 2.6%🔻 |
| Android 6.0 (32bit) + Hermes | 1,230        | **964**      |21.7%🔻 |
| Android 7.0 (64bit) + Hermes | 403          | **367**       |9%🔻 |

The following cases are more effective.
- Use many components
- Use low performance devices
- Use Hermes engine

I think we can improve the performance of many React Native apps without any effort. Please review it positively.

JLHwung:
@ifsnow On the CircleCI failure, please rebase on upstream master.

ifsnow:
@JLHwung Thanks for your feedback. Rebased my changes on upstream master to clear CircleCI failure. I think Travis CI failure is due to a build system issue (Network timeout).

nicolo-ribaudo:
Restarted

ifsnow:
@nicolo-ribaudo Do you have any plans to release the new version? I'm wondering when this fix will be available :)

PR #10170, opened 2019-07-05
Fix objectSpread2 backward compatibility

### LINKED ISSUE #10169: Unknown helper objectSpread2
Hello!

Thank you for your hard work.

However, I'm getting the following error when trying to compile my package using Babel with Rollup:

```
ReferenceError: Unknown helper objectSpread2
  at loadHelper (~/project/node_modules/@babel/helpers/lib/index.js:238:27)
  at Object.getDependencies (~/project/node_modules/@babel/helpers/lib/index.js:279:21)
  at File.addHelper (~/project/node_modules/@babel/core/lib/transformation/file/file.js:204:33)
  at PluginPass.addHelper (~/project/node_modules/@babel/core/lib/transformation/plugin-pass.js:31:22)
  at PluginPass.ObjectExpression (~/project/tumbleweed/node_modules/@babel/plugin-proposal-object-rest-spread/lib/index.js:379:25)
  at newFn (~/project/node_modules/@babel/traverse/lib/visitors.js:193:21)
  at NodePath._call (~/project/node_modules/@babel/traverse/lib/path/context.js:53:20)
  at NodePath.call (~/project/node_modules/@babel/traverse/lib/path/context.js:40:17)
  at NodePath.visit (~/project/node_modules/@babel/traverse/lib/path/context.js:88:12)
  at TraversalContext.visitQueue (~/project/node_modules/@babel/traverse/lib/context.js:118:16)

  code: 'PLUGIN_ERROR',
  helper: 'objectSpread2',
  pluginCode: 'BABEL_HELPER_UNKNOWN',
  plugin: 'babel',
  hook: 'transform',
  id: '~/project/file.ts'
```

I've tried to additionally install the `@babel/runtime`, but it didn't help.

Here's my config for `rollup-plugin-babel`:

```json
{
  "babelConfig": {
    "extensions": [
      ".js",
      ".jsx",
      ".ts",
      ".tsx"
    ],
    "root": "~/project",
    "babelrc": false,
    "runtimeHelpers": true,
    "presets": [
      [
        "@babel/preset-env",
        {
          "targets": "extends @acme/browserslist-config"
        }
      ]
    ],
    "plugins": [
      [
        "@babel/plugin-transform-runtime",
        {
          "corejs": 3
        }
      ]
    ],
    "exclude": [
      "**/node_modules/**"
    ]
  }
}
```

The file in question is just a regular TypeScript file, which uses spread operator to merge objects. I'm processing TypeScript code using `rollup-plugin-typescript2` before feeding it to Babel. The TypeScript compiler is configured to output ESNext code in ESM format.

Do you have any clues, what could be causing this? Any suggestions will be highly appreciated.

---

The dependencies are:

```
    "@babel/core": "^7.5.0",
    "@babel/plugin-transform-runtime": "^7.5.0",
    "@babel/preset-env": "^7.5.0",
    "@babel/runtime": "^7.5.0",
    "rollup": "^1.16.4",
    "rollup-plugin-babel": "^4.3.3",
    "rollup-plugin-commonjs": "^10.0.1",
    "rollup-plugin-typescript2": "^0.21.2",
    "typescript": "^3.5.2"
```

Thank you!

nicolo-ribaudo:
| Fixed Issues?            | Fixes #10169
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -406,7 +406,7 @@ helpers.objectSpread = helper("7.0.0-beta.0")`
   }
 `;
 
-helpers.objectSpread2 = helper("7.0.0-beta.0")`
+helpers.objectSpread2 = helper("7.5.0")`
   import defineProperty from "defineProperty";
 
   export default function _objectSpread2(target) {
--- a/packages/babel-plugin-proposal-object-rest-spread/src/index.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/src/index.js
@@ -452,7 +452,13 @@ export default declare((api, opts) => {
         if (loose) {
           helper = getExtendsHelper(file);
         } else {
-          helper = file.addHelper("objectSpread2");
+          try {
+            helper = file.addHelper("objectSpread2");
+          } catch {
+            // objectSpread2 has been introduced in v7.5.0
+            // We have to maintain backward compatibility.
+            helper = file.addHelper("objectSpread");
+          }
         }
 
         path.replaceWith(t.callExpression(helper, args));


XhmikosR:
@nicolo-ribaudo: Using v7.5.1 still fails for us https://travis-ci.org/twbs/bootstrap/jobs/554971554

nicolo-ribaudo:
It seems like there is a problem with the way `rollup-plugin-babel` injects the helpers. Can you try using `@babel/plugin-transform-runtime` and `@babel/runtime` instead? It should work until the `rollup-plugin-babel` problem is fixed.

benjamn:
I’m seeing the `objectSpread` helper injected as a function, even though I’m using `@babel/plugin-transform-runtime`:
```js
function _objectSpread(target) { for (var i = 1; i < arguments.length; i++) { if (i % 2) { var source = arguments[i] != null ? arguments[i] : {}; var ownKeys = Object.keys(source); if (typeof Object.getOwnPropertySymbols === 'function') { ownKeys = ownKeys.concat(Object.getOwnPropertySymbols(source).filter(function (sym) { return Object.getOwnPropertyDescriptor(source, sym).enumerable; })); } ownKeys.forEach(function (key) { _defineProperty(target, key, source[key]); }); } else { Object.defineProperties(target, Object.getOwnPropertyDescriptors(arguments[i])); } } return target; }
```
Is that something y’all’re aware of?

nicolo-ribaudo:
Did you tell `@babel/plugin-transform-runtime` which version of `@babel/runtime` you are using? Otherwise we must assume that it is `^7.0.0`, and can't rely on helpers introduced in newer versions.

billyjanitsch:
@nicolo-ribaudo FYI the `version` option is [not documented](https://babeljs.io/docs/en/babel-plugin-transform-runtime) in the runtime plugin docs. Would you like me to open an issue for that in the website repo?

Also, if the `version` option is incompatible with `objectSpread2`, shouldn't the effect of this PR be to fall back to importing the old `objectSpread` helper, rather than injecting the new `objectSpread2` helper? In other words, it doesn't seem like the try/catch block added in this PR is working as intended.

benjamn:
@nicolo-ribaudo That was the trick I needed, thanks!

@billyjanitsch I was just about to leave a similar comment about documentation. Full support for opening an issue.

nicolo-ribaudo:
> FYI the version option is not documented in the runtime plugin docs. Would you like me to open an issue for that in the website repo?

Thanks! I have opened https://github.com/babel/website/issues/2055 in the meantime, but a PR would be appreciated :blush:

> Also, if the version option is incompatible with objectSpread2, shouldn't the effect of this PR be to fall back to importing the old objectSpread helper, rather than injecting the new objectSpread2 helper? In other words, it doesn't seem like the try/catch block added in this PR is working as intended.

The try-catch block is used when loading the v7.5 plugin with an older `@babel/core` version, which doesn't even know about the existence of `objectSpread2`.
If you are using `@babel/core` 7.5 the helper exists. `@babel/transform-runtime` inlines it because it's better to use an inline working helper, than a broken helper from `@babel/runtime`.

PR #10172, opened 2019-07-05
fix: register injected importDeclaration

JLHwung:
| Fixed Issues?            | [REPL](https://babeljs.io/en/repl#?babili=false&browsers=ChromeAndroid%20versions&build=&builtIns=false&spec=false&loose=false&code_lz=KYDwDg9gTgLgBAIwIZTgMyhAtnARAOgHpkpcBuIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=false&targets=&version=7.5.1&externalPlugins=%40babel%2Fplugin-proposal-export-default-from%407.2.0%2C%40babel%2Fplugin-transform-typescript%407.5.1)
| Patch: Bug Fix?          | ✅
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | ✅
| Documentation PR Link    | 
| Any Dependency Changes?  |

When `transform-typescript` combines with `proposal-export-default-from`, the `export { _foo as foo };` will be erased as it seems that `_foo` is not defined in its own scope. This PR register the injected importDeclaration so that it can work well with elideImports from `transform-typescript`.

Note that #7592 uses `scope.crawl()` to manually sync nodes to AST. As `crawl` will incur great performance overkill, our approach check the injected import specifier only and register when necessary.

### DIFF
--- a/packages/babel-plugin-proposal-export-default-from/src/index.js
+++ b/packages/babel-plugin-proposal-export-default-from/src/index.js
@@ -35,6 +35,11 @@ export default declare(api => {
 
         path.replaceWithMultiple(nodes);
       },
+      ImportDefaultSpecifier(path) {
+        if (!path.scope.hasOwnBinding(path.node.local.name)) {
+          path.scope.registerDeclaration(path.parentPath);
+        }
+      },
     },
   };
 });
--- a/packages/babel-plugin-proposal-export-default-from/src/index.js
+++ b/packages/babel-plugin-proposal-export-default-from/src/index.js
@@ -33,12 +33,8 @@ export default declare(api => {
           nodes.push(node);
         }
 
-        path.replaceWithMultiple(nodes);
-      },
-      ImportDefaultSpecifier(path) {
-        if (!path.scope.hasOwnBinding(path.node.local.name)) {
-          path.scope.registerDeclaration(path.parentPath);
-        }
+        const [importDeclaration] = path.replaceWithMultiple(nodes);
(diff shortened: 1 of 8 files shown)


packages/babel-plugin-proposal-export-default-from/src/index.js
```diff
@@ -35,6 +35,11 @@ export default declare(api => {
 
         path.replaceWithMultiple(nodes);
       },
+      ImportDefaultSpecifier(path) {
+        if (!path.scope.hasOwnBinding(path.node.local.name)) {
+          path.scope.registerDeclaration(path.parentPath);
```
nicolo-ribaudo: I would prefer to do this as soon as the import declaration is created, in the `ExportNamedDeclaration` visitor. This makes it clearer _why_ we are doing it, and it prevents other plugin from running between the creating the declaration and registering it.

`path.replaceWithMultiple(nodes);` returns an array containing the new inserted paths: the first one should be the import declaration, and you can get it's default specifier.

JLHwung: > `path.replaceWithMultiple(nodes)` returns an array containing the new inserted paths.

Yes it is. It seems that `@types/babel__traverse` declares out-of-dated signature [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/babel__traverse/index.d.ts#L274).

Issue #10162, opened 2019-07-04
Re-export statements are omitted from output using @babel/plugin-transform-typescript@7.5.0


elliottsj:
## Bug Report

**Current Behavior**
Using `@babel/plugin-transform-typescript@7.5.0` via `@babel/preset-typescript@7.3.3`, re-export statements are omitted from the transpiled output.

Repro:

```sh-session
$ npm install @babel/cli @babel/core @babel/preset-typescript
$ echo "export { bar } from './bar';" > script.ts
$ ./node_modules/.bin/babel --presets @babel/preset-typescript script.ts


$ # observe: the above output is empty
```

Note that separate import + export statements transpile properly:

```sh-session
$ echo "import { bar } from './bar'; export { bar };" > script.ts
$ ./node_modules/.bin/babel --presets @babel/preset-typescript script.ts
import { bar } from './bar';
export { bar };

$ # observe: the output is as expected
```

**Input Code**

```js
export { bar } from './bar';
```

**Expected behavior/code**
The re-export statement should be included in the output.

**Babel Configuration (.babelrc, package.json, cli command)**

```js
{
  "presets": ["@babel/preset-typescript"]
}
```

**Environment**
- Babel version(s):

    ```
    $ npm ls --depth=1                                                                                                                                                                                                                                          (522ms)
    /Users/spencerelliott/Dev/temp
    ├─┬ @babel/cli@7.5.0
    │ ├── chokidar@2.1.6
    │ ├── commander@2.20.0
    │ ├── convert-source-map@1.6.0
    │ ├── fs-readdir-recursive@1.1.0
    │ ├── glob@7.1.4
    │ ├── lodash@4.17.11
    │ ├── mkdirp@0.5.1
    │ ├── output-file-sync@2.0.1
    │ ├── slash@2.0.0
    │ └── source-map@0.5.7
    ├─┬ @babel/core@7.5.0
    │ ├── @babel/code-frame@7.0.0
    │ ├── @babel/generator@7.5.0
    │ ├── @babel/helpers@7.5.0
    │ ├── @babel/parser@7.5.0
    │ ├── @babel/template@7.4.4
    │ ├── @babel/traverse@7.5.0
    │ ├── @babel/types@7.5.0
    │ ├── convert-source-map@1.6.0 deduped
    │ ├── debug@4.1.1 extraneous
    │ ├── json5@2.1.0
    │ ├── lodash@4.17.11 deduped
    │ ├── resolve@1.11.1
[...]

PR #10174, opened 2019-07-07
Do not trust Scope when removing TypeScript types

nicolo-ribaudo:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/10162, fixes https://github.com/gatsbyjs/gatsby/issues/1542v
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

In v7.5.0, we migrated the TypeScript plugin to use Scope to remove exported types. We assumed that if a node wasn't in the scope, it was a type. While this is _in theory_ true, plugins never did a good job at maintaining this "invariant".
Not only popular community plugins (e.g. https://github.com/gaearon/react-hot-loader/pull/1293), but also our own plugins (e.g. https://github.com/babel/babel/pull/10172).

While updating those plugins makes the scope more reliable, we can't drop support with older versions since it would be a breaking change.

This PR changes the TypeScript plugin so that it removes an export only if
1. It isn't registered in the scope
2. We saw it in a type declaration.

2 without 1 isn't enough, because we shouldn't remove the export in cases like this:
```js
var a = 1;
type a = 2;

export { a };
```

If others agree that this is the correct solution, I will release v7.5.2 as soon as this is merged.

cc @Wolvereness (author of the initial TS update), @JLHwung (fixed https://github.com/babel/babel/pull/10172), @theKashey (created https://github.com/gaearon/react-hot-loader/pull/1293).

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -23,7 +23,20 @@ const GLOBAL_TYPES = new WeakMap();
 
 function isGlobalType(path, name) {
   const program = path.find(path => path.isProgram()).node;
-  return GLOBAL_TYPES.get(program).has(name) && !path.scope.hasOwnBinding(name);
+  if (path.scope.hasOwnBinding(name)) return false;
+  if (GLOBAL_TYPES.get(program).has(name)) return true;
+
+  console.warn(
+    `The exported identifier "${name}" is not declared in Babel's scope tracker\n` +
+      `as a JavaScript value binding, but "@babel/plugin-transform-typescript"\n` +
+      `never encountered it as a TypeScript type declaration.\n` +
+      `It will be treated as a JavaScript value.\n\n` +
+      `This problem is likely caused by another plugin plugin which is injecting\n` +
+      `"${name}" without registering it in the scope tracker. If you are the author\n` +
+      ` of that plugin, please use "scope.registerDeclaration(declarationPath)".`,
+  );
+
+  return false;
 }
 
 function registerGlobalType(programScope, name) {
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -19,6 +19,16 @@ function isInType(path) {
 }
 
 const PARSED_PARAMS = new WeakSet();
+const GLOBAL_TYPES = new WeakMap();
+
+function isGlobalType(path, name) {
+  const program = path.find(path => path.isProgram()).node;
+  return GLOBAL_TYPES.get(program).has(name) && !path.scope.hasOwnBinding(name);
+}
+
+function registerGlobalType(programScope, name) {
+  GLOBAL_TYPES.get(programScope.path.node).add(name);
+}
 
 export default declare(
   (api, { jsxPragma = "React", allowNamespaces = false }) => {
@@ -40,6 +50,10 @@ export default declare(
           const { file } = state;
           let fileJsxPragma = null;
 
+          if (!GLOBAL_TYPES.has(path.node)) {
+            GLOBAL_TYPES.set(path.node, new Set());
+          }
+
           if (file.ast.comments) {
             for (const comment of (file.ast.comments: Array<Object>)) {
               const jsxMatches = JSX_ANNOTATION_REGEX.exec(comment.value);
@@ -50,7 +64,7 @@ export default declare(
           }
 
           // remove type imports
-          for (const stmt of path.get("body")) {
+          for (let stmt of path.get("body")) {
             if (t.isImportDeclaration(stmt)) {
               // Note: this will allow both `import { } from "m"` and `import "m";`.
               // In TypeScript, the former would be elided.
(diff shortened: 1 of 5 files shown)


Wolvereness:
I would suggest a warning being added as a canary.

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -91,6 +105,28 @@ export default declare(
...
+              stmt = stmt.get("declaration");
+            }
+
+            if (stmt.isVariableDeclaration({ declare: true })) {
+              for (const name of Object.keys(stmt.getBindingIdentifiers())) {
+                registerGlobalType(path.scope, name);
+              }
+            } else if (
+              stmt.isTSTypeAliasDeclaration() ||
+              stmt.isTSDeclareFunction() ||
```
JLHwung: Would `stmd.node.id.name` bail on `export default function() {}`?

nicolo-ribaudo: `TSDeclareFunction` (`declare function f(): type;`) can't be used in an `export default` declaration, so it can never be anonymous.

nicolo-ribaudo:
@babel/babel What do you think about the warning?

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -19,6 +19,29 @@ function isInType(path) {
...
+  const program = path.find(path => path.isProgram()).node;
+  if (path.scope.hasOwnBinding(name)) return false;
+  if (GLOBAL_TYPES.get(program).has(name)) return true;
+
+  console.warn(
+    `The exported identifier "${name}" is not declared in Babel's scope tracker\n` +
+      `as a JavaScript value binding, but "@babel/plugin-transform-typescript"\n` +
+      `never encountered it as a TypeScript type declaration.\n` +
+      `It will be treated as a JavaScript value.\n\n` +
+      `This problem is likely caused by another plugin plugin which is injecting\n` +
```
existentialism:
```suggestion
      `This problem is likely caused by another plugin injecting\n` +
```

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -19,6 +19,29 @@ function isInType(path) {
...
+const GLOBAL_TYPES = new WeakMap();
+
+function isGlobalType(path, name) {
+  const program = path.find(path => path.isProgram()).node;
+  if (path.scope.hasOwnBinding(name)) return false;
+  if (GLOBAL_TYPES.get(program).has(name)) return true;
+
+  console.warn(
+    `The exported identifier "${name}" is not declared in Babel's scope tracker\n` +
+      `as a JavaScript value binding, but "@babel/plugin-transform-typescript"\n` +
```
existentialism:
```suggestion
      `as a JavaScript value binding, and "@babel/plugin-transform-typescript"\n` +
```

Andarist:
IMHO error would be better than warning. From taking a brief look at it I understand that it's rather unsafe transformation in some cases due to how plugins are implemented. And if it's unsafe it would be better (IMHO) to fail fast and force people to implement correct logic in plugins rather than just warning which may pass unnoticed and break somebody.

nicolo-ribaudo:
While I would really like to throw an error, it would break compatibility with `@babel/plugin-proposal-export-default-from` <=7.5, `@babel/plugin-proposal-export-default-from` <=7.5 and `react-hot-loader` <4.12.5.
We found these plugin during a weekend because they are either _really_ popular, but it would probably break many other less known plugins.

JLHwung:
> While I would really like to throw an error, it would break compatibility with @babel/plugin-proposal-export-default-from <=7.5, @babel/plugin-proposal-export-default-from <=7.5 and react-hot-loader <4.12.5.

I think we can state in the warning that we _may_ throw error since the next major release. It is recommended for plugin user to fix potential problems. So we can finally get rid of these last resort extra check.

nicolo-ribaudo:
@JLHwung I don't think that we will ever throw an error, since when we would be able to throw an error for every binding which is not registered we would also be able to automatically register them.
I belive that such a long warning is already a big incentive for plugin authors to correctly register them anyway :laughing:

Peping:
FWIW if somebody stumbles upon this, suffering from an vast amount of warnings, updating `@babel/plugin-proposal-export-default-from` and `react-hot-loader` (as per [this comment](https://github.com/babel/babel/pull/10174#issuecomment-509033052)) might solve them for you.

Backstory: I switched from `typescript-loader` to `@babel/preset-typescript` recently and I got heaps of warnings about this particular issue whenever I launched the dev server or built my app. Kinda like this:

```
The exported identifier "_default" is not declared in Babel's scope tracker
as a JavaScript value binding, and "@babel/plugin-transform-typescript"
never encountered it as a TypeScript type declaration.
It will be treated as a JavaScript value.

This problem is likely caused by another plugin injecting
"_default" without registering it in the scope tracker. If you are the author
 of that plugin, please use "scope.registerDeclaration(declarationPath)".
The exported identifier "_default" is not declared in Babel's scope tracker
as a JavaScript value binding, and "@babel/plugin-transform-typescript"
never encountered it as a TypeScript type declaration.
It will be treated as a JavaScript value.

This problem is likely caused by another plugin injecting
"_default" without registering it in the scope tracker. If you are the author
 of that plugin, please use "scope.registerDeclaration(declarationPath)".
The exported identifier "_default" is not declared in Babel's scope tracker
as a JavaScript value binding, and "@babel/plugin-transform-typescript"
never encountered it as a TypeScript type declaration.
It will be treated as a JavaScript value.

This problem is likely caused by another plugin injecting
"_default" without registering it in the scope tracker. If you are the author
 of that plugin, please use "scope.registerDeclaration(declarationPath)".
[... this repeats ~80 more times]
```

PR #10180, opened 2019-07-08
[_objectSpread2] Do not use hoisted var from prev iteration

### LINKED ISSUE #10166: undefined is not a function(evaluation 'object.getOwnPropertyDescriptors(arguments[i])')
https://github.com/facebook/react-native/issues/23092

Unfortunately this exact same issue strikes again with the release of Babel v7.5.0.

These are the Babel related dependencies I am having:

```
    "@babel/core": "7.4.5",
    "@babel/plugin-proposal-class-properties": "7.4.4",
    "@babel/plugin-proposal-decorators": "7.4.4",
    "@babel/plugin-transform-flow-strip-types": "7.4.4",
    "@babel/preset-flow": "7.0.0",
    "babel-eslint": "10.0.2",
    "babel-jest": "24.8.0",
    "metro-react-native-babel-preset": "0.54.1",
```

With the `package-lock.json` I generated a few days ago everything's fine with `npm install`, but if I try to do it without the lockfile the error appears. Judging from my lockfile's diff I can say that a few secondary Babel dependencies were updated to v7.5.0 without the lockfile (eg. `@babel/generator`, `@babel/helper-create-class-features-plugin`) so am pretty sure this is a Babel issue again.

nicolo-ribaudo:
| Fixed Issues?            | Fix #10166 for real
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This was an error introduced in https://github.com/babel/babel/pull/10171/files.
The helper got the keys from `source`, which was defined at the previous iteration. 

Also, I added a test faking an old env by deleting `Object.getOwnPropertyDescriptors` :sunglasses: 	

If I had used `let` this would not have happened. Also, this is what happens when I merge a PR without waiting for the usual two :heavy_check_mark:; this bug could have been caught by more eyes looking at it.

@JordanProtin Could you check that if you update `_objectSpread2` like this in `@babel/helpers`, `@babel/runtime`, `@babel/runtime-corejs2` and/or `@babel/runtime-corejs3` in your `node_modules` (there could be multiple copied of each package) it works? I can reproduce the error without this fix but it seems to be fixed.


JordanProtin:
@nicolo-ribaudo Thank you for your hard work ! I'm sorry, it's a difficult for me to test this now ! I'll do this ASAP ! :D

PR #10181, opened 2019-07-09
feat(errors): validate preset when filename is absent

### LINKED ISSUE #10154: @babel/standalone transform failed when using preset-typescript
## Bug Report

**Current Behavior**
@babel/standalone transform failed when using preset-typescript

**Input Code**

```js
const _code = Babel.transform(code, { presets: ['typescript'] }).code;
```

**Environment**
- Babel version(s): @babel/standalone@7.4.5
- Node/npm version: node v8.9.3 / npm v6.9.0
- OS: OSX 10.13.2
- Monorepo: [no]
- How you are using Babel: standalone

JLHwung:
Closes #10154

| Fixed Issues?            | Fixes #10154 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 

Validate preset before `buildPresetChain`, when it is the last chance we can get the `UnloadedDescriptor` and include preset name in error message. We may add more business logic inside `validatePreset` to print preset-specific error message.

I have also considered the following alternative designs:

1) Pass down descriptor name to `matchPattern`.
Pro: The missing filename errors are still thrown in `matchPattern`
Cons: Introduce unnecessary signature complexity as this is not our primary execution path

2) Add descriptor name to `ConfigContext`
Pro: The missing filename errors are still thrown in `matchPattern`
Cons: ConfigContext has to be mutated or cloned when running `buildPresetChain`. The `ConfigContext` is widely used across config chain building but descriptor name is only used to print this error.

3) tryCatch `buildPresetChain`
Pros: Intuitive and fewer execution path
Cons: Any error from `buildPresetChain ` has to be matched once before they are re-thrown.

### DIFF
--- a/packages/babel-core/src/config/full.js
+++ b/packages/babel-core/src/config/full.js
@@ -284,9 +284,18 @@ const validateIfOptionNeedsFilename = (
   descriptor: UnloadedDescriptor,
 ): void => {
   if (options.test || options.include || options.exclude) {
+    const formattedPresetName = descriptor.name
+      ? `"${descriptor.name}"`
+      : "/* your preset */";
     throw new Error(
-      `Preset ${descriptor.name ||
-        ""} requires filename, but it was not passed.`.replace(/\s{2}/, " "),
+      [
+        `Preset ${formattedPresetName} requires a filename be set.`,
+        `For example, if you are calling Babel directly`,
+        `\`\`\``,
+        `babel.transform(code, { filename: 'file.js', presets: [${formattedPresetName}] });`,
+        `\`\`\``,
+        `See https://babeljs.io/docs/en/options#filename for more information.`,
+      ].join("\n"),
     );
   }
 };
--- a/packages/babel-core/src/config/full.js
+++ b/packages/babel-core/src/config/full.js
@@ -278,6 +278,22 @@ const instantiatePlugin = makeWeakCache(
   },
 );
 
+const validatePreset = (
+  preset: PresetInstance,
+  context: ConfigContext,
+  descriptor: UnloadedDescriptor,
+): void => {
+  if (!context.filename) {
+    const { options } = preset;
+    if (options.test || options.include || options.exclude) {
+      throw new Error(
+        `Preset ${descriptor.name ||
+          ""} requires filename, but it was not passed.`.replace(/\s{2}/, " "),
+      );
+    }
+  }
+};
+
 /**
  * Generate a config object that will act as the root of a new nested config.
  */
@@ -285,10 +301,9 @@ const loadPresetDescriptor = (
   descriptor: UnloadedDescriptor,
   context: ConfigContext,
 ): ConfigChain | null => {
-  return buildPresetChain(
-    instantiatePreset(loadDescriptor(descriptor, context)),
-    context,
-  );
+  const preset = instantiatePreset(loadDescriptor(descriptor, context));
+  validatePreset(preset, context, descriptor);
+  return buildPresetChain(preset, context);
 };
 
 const instantiatePreset = makeWeakCache(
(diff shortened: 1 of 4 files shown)


packages/babel-core/src/config/full.js
```diff
@@ -278,17 +278,32 @@ const instantiatePlugin = makeWeakCache(
...
+  preset: PresetInstance,
+  context: ConfigContext,
+  descriptor: UnloadedDescriptor,
+): void => {
+  if (!context.filename) {
+    const { options } = preset;
+    if (options.test || options.include || options.exclude) {
+      throw new Error(
+        `Preset ${descriptor.name ||
+          ""} requires filename, but it was not passed.`.replace(/\s{2}/, " "),
```
existentialism: Might be nice to link directly to the option:

> Preset ${name} requires a filename be set. See https://babeljs.io/docs/en/options#filename.

Or even:

> Preset ${name} requires a filename be set. For example, if you are calling Babel directly:
>
> ```
> babel.transform(code, { filename: 'file.js', presets: [${name || '/* your preset */'}] });
> ```
>
> See https://babeljs.io/docs/en/options#filename for more information.

(And, as you mentioned, improve this later with user- and preset-specific messages)

existentialism:
Should we also update the language on the options page (https://babeljs.io/docs/en/options#filename)?

Maybe simply changing:

> The filename is exposed to plugins. Some plugins may require the presence of the filename.

to:

> Some presets and plugins may require the presence of the filename.

Or, since we get the question a bunch, call out preset-typescript's behavior directly here?

I'm fine with either

packages/babel-core/src/config/full.js
```diff
@@ -278,17 +278,32 @@ const instantiatePlugin = makeWeakCache(
...
 );
 
+const validatePreset = (
+  preset: PresetInstance,
+  context: ConfigContext,
+  descriptor: UnloadedDescriptor,
+): void => {
+  if (!context.filename) {
+    const { options } = preset;
+    if (options.test || options.include || options.exclude) {
```
nicolo-ribaudo: We should also check if the `overrides` option contains a block which has one of these three options (like in the TS preset).

packages/babel-core/src/config/full.js
JLHwung: That delivers way better dev UX. Addressed in ffc6a72.

packages/babel-core/src/config/full.js
JLHwung: Good catch. Fixed in 056e197.

nicolo-ribaudo:
(Note: I labeled this as "New feature" but I I think that this PR can be merged in a patch version)

packages/babel-core/src/config/full.js
```diff
@@ -278,17 +279,53 @@ const instantiatePlugin = makeWeakCache(
...
+  options: ValidatedOptions,
+  descriptor: UnloadedDescriptor,
+): void => {
+  if (options.test || options.include || options.exclude) {
+    const formattedPresetName = descriptor.name
+      ? `"${descriptor.name}"`
+      : "/* your preset */";
+    throw new Error(
+      [
+        `Preset ${formattedPresetName} requires a filename be set.`,
```
JLHwung: I read again and feel like
```
Preset ${formattedPresetName} requires a filename to be set when babel is called directly,
`\`\`\``,
`babel.transform(code, { filename: 'file.ts', presets: [${formattedPresetName}] });`,
`\`\`\``
```

And we may can change `file.js` to `file.ts` since this is the popular path.

nicolo-ribaudo: > And we may can change file.js to file.ts since this is the popular path

Good idea

PR #10189, opened 2019-07-09
perf: match ownKeys perf to the one of objectSpread

### LINKED ISSUE #10192: objectSpread2 failure
## Bug Report

**Current Behavior**

While trying to update babel and its plugins to the latest v7.5.4 version, we still get some failures in our test suite https://travis-ci.org/twbs/bootstrap/jobs/556685620#L788

The related code should be https://github.com/twbs/bootstrap/blob/master/js/src/dom/manipulator.js#L47

Going back to v7.4.x and `objectSpread` works as expected.

**Input Code**

- https://github.com/twbs/bootstrap

```js
module.exports = {
  presets: [
    [
      '@babel/env',
      {
        loose: true,
        modules: false,
        exclude: ['transform-typeof-symbol']
      }
    ]
  ],
  plugins: [
    '@babel/plugin-proposal-object-rest-spread'
  ],
  env: {
    test: {
      plugins: [ 'istanbul' ]
    }
  }
};
```

**Environment**

- Babel version(s): v7.5.4
- Node/npm version: Node 10.16.0/npm 6.9.0
- OS: Windows 10
- Monorepo: no

JLHwung:
| Fixed Issues?            | Fixes a performance regression introduced in #10171, Fixes #10192 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR addresses a performance regression introduced at #10171.

`Object.keys` returns a list of enumerable property names. Hence we filter `enumerable` only on the property symbols. Note that this approach is also [practiced](https://github.com/babel/babel/blob/v7.5.4/packages/babel-helpers/src/helpers.js#L397) in `helpers.objectSpread`. 

In practice the most frequent use cases of objectSpread is merging two plain objects, which means we can skip a sweep of already enumerable property names. I also drafted a [jsPerf snippet](https://jsperf.com/object-spread-babel-helper) on the performance difference. It is 10% faster on V8 75 when merging two object with 512 keys.

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -412,11 +412,12 @@ helpers.objectSpread2 = helper("7.5.0")`
   function ownKeys(object, enumerableOnly) {
     var keys = Object.keys(object);
     if (Object.getOwnPropertySymbols) {
-      keys.push.apply(keys, Object.getOwnPropertySymbols(object));
+      var symbols = Object.getOwnPropertySymbols(object);
+      if (enumerableOnly) symbols = symbols.filter(function (sym) {
+        return Object.getOwnPropertyDescriptor(object, sym).enumerable;
+      });
+      keys.push.apply(keys, symbols);
     }
-    if (enumerableOnly) keys = keys.filter(function (sym) {
-      return Object.getOwnPropertyDescriptor(object, sym).enumerable;
-    });
     return keys;
   }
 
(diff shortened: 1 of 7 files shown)


nicolo-ribaudo:
Thanks!

(As a side note, I really like having more :heavy_check_mark: than usual!)

JLHwung:
> (especially the comment!)

The idea comes from discussion with @nicolo-ribaudo

PR #10198, opened 2019-07-11
add assertShape to validate templateElement

### LINKED ISSUE #10195: Value property of templateElement type missing documentation
Link: https://babeljs.io/docs/en/next/babel-types.html#templateelement

Docs reads: value (required)

According to astExplorer it should be: {raw: string, cooked: string}

Babel also doesn't throw if you construct a templateElement without this object, and produces wrong output.

tanhauhau:
Should we update the `templateElement`'s validator and [update the docs as well](https://github.com/babel/website/pull/2061)?

@sebmck need your advice, because i saw you left a `TODO` there.

| Fixed Issues?            | Fixes #10195 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-types/src/definitions/es2015.js
+++ b/packages/babel-types/src/definitions/es2015.js
@@ -539,8 +539,13 @@ defineType("TemplateElement", {
   fields: {
     value: {
       validate: assertShape({
-        raw: assertValueType("string"),
-        cooked: assertValueType("string"),
+        raw: {
+          validate: assertValueType("string"),
+        },
+        cooked: {
+          validate: assertValueType("string"),
+          optional: true,
+        },
       }),
     },
     tail: {
--- a/packages/babel-types/src/definitions/es2015.js
+++ b/packages/babel-types/src/definitions/es2015.js
@@ -1,5 +1,6 @@
 // @flow
 import defineType, {
+  assertShape,
   assertNodeType,
   assertValueType,
   chain,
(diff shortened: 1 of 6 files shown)


nicolo-ribaudo:
If you run `make build` does it updates the generated flow/typescript types?

tanhauhau:
```js
// index.js.flow
declare function templateElement(value: {raw:string, cooked:string}, tail?: boolean): BabelNodeTemplateElement;

// index.d.ts
export function templateElement(value: {raw:string, cooked:string}, tail?: boolean): TemplateElement;
```

packages/babel-types/src/definitions/utils.js
```diff
@@ -161,6 +161,35 @@ export function assertValueType(type: string): Validator {
   return validate;
 }
 
+export function assertShape(shape: {| [string]: Validator |}): Validator {
```
nicolo-ribaudo:
```suggestion
export function assertShape(shape: { [string]: Validator }): Validator {
```

indexers don't work with exact objects.

packages/babel-types/src/definitions/es2015.js
```diff
@@ -537,7 +538,10 @@ defineType("TemplateElement", {
   builder: ["value", "tail"],
   fields: {
     value: {
-      // todo: flatten `raw` into main node
+      validate: assertShape({
+        raw: assertValueType("string"),
+        cooked: assertValueType("string"),
```
jridgewell: `cooked` may also be `null` in a tagged template:

```js
tag`foo\x`
```

tanhauhau: @jridgewell made `cooked` optional

Issue #10163, opened 2019-07-04
@babel/parser 7.5.0 isn't published


alubbe:
## Bug Report

**Current Behavior**
@babel/traverse 7.5.0 wants @babel/parser ^7.5.0 but nothing matches that

Could you please publish @babel/parser 7.5.0?

nicolo-ribaudo:
I accidentally published it as "next" instead of "latest". It should be fixed now.

vincentholmen:
Hello @nicolo-ribaudo,

Unfortunately, I am still encountering this behavior and unable to install babel/parser 7.5.0, which is causing issues within CI Pipeline.

Can you confirm this has been corrected and the latest version 7.5.0 is available?

LuisEGR:
I am also having problems when installing this:

```
npm install  @babel/generator@^7.5.0
npm ERR! code ETARGET
npm ERR! notarget No matching version found for @babel/generator@^7.5.0
npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! notarget a package version that doesn't exist.
```
I can see the v7.5.0 on https://www.npmjs.com/package/@babel/generator but when I install with the default command I get the 7.4.4:

```
npm i @babel/generator
+ @babel/generator@7.4.4
updated 1 package in 0.8s
```
Can anyone correct the package on npm please?

Jrfidellis:
Maybe related with #10164?

nicolo-ribaudo:
`@babel/generator` was correctly published: https://www.npmjs.com/package/@babel/generator

Please try running the workarounds described at https://github.com/babel/babel/issues/10165

PR #10201, opened 2019-07-11
react-display-name: add names to lazy() components

motiz88:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | https://github.com/babel/website/pull/2060
| Any Dependency Changes?  |

https://github.com/facebook/react/pull/16104 (NOTE: not yet merged) adds support for setting `displayName` on the return value of `React.lazy()`:

```js
const Foo = React.lazy(() => import('./FooImplementation'));
Foo.displayName = 'Foo';
```

This PR amends `@babel/plugin-transform-react-display-name` to do this automatically in common cases.
* We reuse the name inference logic used for `createReactClass`, but additionally fall back to deriving a name from the import path, if one is found.
* In addition to blindly matching on calls to `React.lazy`, we also support `import { lazy } from 'react'`.

### DIFF
--- a/packages/babel-plugin-transform-react-display-name/src/index.js
+++ b/packages/babel-plugin-transform-react-display-name/src/index.js
@@ -52,59 +52,162 @@ export default declare(api => {
     return true;
   }
 
+  function isReactLazy(node) {
+    if (!node || !t.isCallExpression(node)) return false;
+
+    if (!t.matchesPattern(node.callee, "React.lazy")) {
+      return false;
+    }
+
+    return true;
+  }
+
+  function displayNameFromFilename(filename) {
+    if (!filename) {
+      return null;
+    }
+
+    let displayName = path.basename(filename, path.extname(filename));
+
+    // ./{module name}/index.js
+    if (displayName === "index") {
+      displayName = path.basename(path.dirname(filename));
+    }
+
+    return displayName;
+  }
+
+  function displayNameFromAncestors(path) {
+    let id;
+
+    // crawl up the ancestry looking for possible candidates for displayName inference
+    path.find(function(path) {
+      if (path.isAssignmentExpression()) {
+        id = path.node.left;
+      } else if (path.isObjectProperty()) {
+        id = path.node.key;
+      } else if (path.isVariableDeclarator()) {
+        id = path.node.id;
+      } else if (path.isStatement()) {
+        // we've hit a statement, we should stop crawling up
+        return true;
+      }
+
+      // we've got an id! no need to continue
+      if (id) return true;
+    });
+
+    // ensure that we have an identifier we can inherit from
+    if (!id) return;
+
+    // foo.bar -> bar
+    if (t.isMemberExpression(id)) {
+      id = id.property;
+    }
+
+    // identifiers are the only thing we can reliably get a name from
+    if (t.isIdentifier(id)) {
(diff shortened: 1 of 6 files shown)


packages/babel-plugin-transform-react-display-name/src/index.js
```diff
@@ -52,59 +52,162 @@ export default declare(api => {
...
+  }
+
+  function displayNameFromAncestors(path) {
+    let id;
+
+    // crawl up the ancestry looking for possible candidates for displayName inference
+    path.find(function(path) {
+      if (path.isAssignmentExpression()) {
+        id = path.node.left;
+      } else if (path.isObjectProperty()) {
```
JLHwung: Consider filter out the computed property on path, i.e.
```js
({
    [locale]: React.lazy(() => import("./someComponent"))
});
```
We would not expect `locale` to be `someComponent`'s display name.
```suggestion
      } else if (path.isObjectProperty({ computed: false })) {
```

packages/babel-plugin-transform-react-display-name/src/index.js
```diff
@@ -52,59 +52,162 @@ export default declare(api => {
...
+
+      // we've got an id! no need to continue
+      if (id) return true;
+    });
+
+    // ensure that we have an identifier we can inherit from
+    if (!id) return;
+
+    // foo.bar -> bar
+    if (t.isMemberExpression(id)) {
```
JLHwung: Similar to `isObjectProperty`, we should filter by `{ computed: false }`.

packages/babel-plugin-transform-react-display-name/src/index.js
motiz88: I can see the case for treating this specially for `lazy`, because we commonly have the additional fallback to the `import` path which is probably better than using a computed property key. But we also might not have an `import` - in which case I'd still want to grab a computed key as a last resort.

In any case I will not change the behaviour for `createClass` here since that is out of this PR's scope.

packages/babel-plugin-transform-react-display-name/src/index.js
```diff
@@ -52,59 +52,162 @@ export default declare(api => {
...
+    }
+    if (!displayName) {
+      return;
+    }
+    const { scope } = callPath;
+    const id = scope.generateDeclaredUidIdentifier(displayName);
+    scope.push({ id });
+    callPath.replaceWith(
+      t.sequenceExpression([
+        t.assignmentExpression("=", id, callPath.node),
```
JLHwung: The `id` needs to be cloned to prevent duplicate nodes in AST. For the background, see #7149 and #10163.

packages/babel-plugin-transform-react-display-name/src/index.js
JLHwung: > there might not be ... an import path we can use instead ... in which case I'd still want to grab a computed key as a last resort.

True. Actually computed property name will more likely to be combined with template element in import call.

```js
({
    [locale]: React.lazy(() => import(`./${locale}/adsComponent`))
});
```

In this case we can not get a valid import filename. Could you add a test case on the computed property? So we make it clear that it is part of design.

motiz88:
This feature has been rejected by the React team, closing.

PR #10203, opened 2019-07-11
[preset-env] Add targets for dynamic import

nicolo-ribaudo:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

In engines which support `import()` natively, it doesn't need to be transpiled. For example, someone might write this code:
```js
import color from "./static-module";

main();
async function main() {
  if (!window.$) await import("jquery-cdn");

  $("body").css("background", color);
}
```

and it could safely become
```js
define(["./static-module", function(foo) {
  main();
  async function main() {
    if (!window.$) await import("jquery-cdn");

    $("body").css("background", color);
  }
});
```

NOTE: I'm not 100% sure that this PR is correct. If someone doesn't want `import()` to be transpiled, you might argue that they shouldn't transpile static imports neither and thus use `modules: false`

### DIFF
--- a/packages/babel-preset-env/data/built-in-modules.json
+++ b/packages/babel-preset-env/data/built-in-modules.json
@@ -8,5 +8,15 @@
     "ios_saf": "10.3",
     "and_chr": "71",
     "and_ff": "64"
+  },
+  "dynamic-import": {
+    "edge": "75",
+    "firefox": "67",
+    "chrome": "63",
+    "safari": "11.1",
+    "opera": "50",
+    "ios_saf": "11.0",
+    "op_mob": "46",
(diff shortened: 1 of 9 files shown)


JLHwung:
Note that we have implemented similar checks for `export-ns-from`: #11849 I will see if I can revive this PR and targets for next minor release.

nicolo-ribaudo:
I don't think anymore that this is something that we should do. We should not implicitly mix two module systems (esm and something else). It's usually a mistake, and if someone really wants to do it they can already explicitly disable `proposal-dynamic-import`.

PR #10200, opened 2019-07-11
Workaround #10179 in proposal-object-rest-spread

nicolo-ribaudo:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #10179
| Patch: Bug Fix?          | Yes
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

The real cause of this bug is in `@babel/core` (https://github.com/babel/babel/issues/10187), but even if we fixed it new versions of `plugin-proposal-object-rest-spread` wouldn't work with older versions of `@babel/core` and `@babel/helpers`.
This bug also needs to be fixed in `@babel/core` anyway, otherwise it will happen again.

I couldn't figure out how to add a test for this bug; I tested it manually with a `package.json` like this (it requires `yarn`):
```json
{
  "devDependencies": {
    "@babel/cli": "^7.5.0",
    "@babel/core": "7.4.0",
    "@babel/plugin-proposal-object-rest-spread": "file:../../babel/babel/packages/babel-plugin-proposal-object-rest-spread"
  },
  "resolutions": {
    "@babel/helpers": "7.4.0"
  }
}
```

<details>
<summary>Input</summary>

```js
void { ...foo };
void { ...bar };
```
</details>

<details>
<summary>Output before the fix</summary>

```js
function _objectSpread2(target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i] != null ? arguments[i] : {}; var ownKeys = Object.keys(source); if (typeof Object.getOwnPropertySymbols === 'function') { ownKeys = ownKeys.concat(Object.getOwnPropertySymbols(source).filter(function (sym) { return Object.getOwnPropertyDescriptor(source, sym).enumerable; })); } ownKeys.forEach(function (key) { _defineProperty(target, key, source[key]); }); } return target; }

function _defineProperty(obj, key, value) { if (key in obj) { Object.defineProperty(obj, key, { value: value, enumerable: true, configurable: true, writable: true }); } else { obj[key] = value; } return obj; }

void _objectSpread2({}, foo);
void _objectSpread({}, bar);
```
</details>

<details>
<summary>Output after the fix</summary>

```js
[...]

PR #10208, opened 2019-07-12
always throw when add missing helpers

### LINKED ISSUE #10187: Missing helpers only throw once
## Bug Report

**Input Code**

```js
const input = ``;

const out = babel.transformSync(input, {
  configFile: false,
  plugins: [
    function() {
      return {
        visitor: {
          Program(path) {
            try {
              path.pushContainer("body", this.addHelper("fooBar"));
            } catch {}
            path.pushContainer("body", this.addHelper("fooBar"));
          },
        },
      };
    },
  ],
});

console.log(out.code);
```

**Expected behavior/code**
It should fail, because the second `addHelper` call should throw.

***Actual behavior**
```js
_fooBar
```

**Environment**
- Babel version(s): 7.5.0

tanhauhau:
| Fixed Issues?            | Fixes #10187
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The problem with current addHelper is that [this line will get execute first](https://github.com/babel/babel/blob/master/packages/babel-core/src/transformation/file/file.js#L206), before knowing whether the helper existed:

```js
// 2) will short circuit if `this.declarations[name]` exists
const declar = this.declarations[name];
if (declar) return t.cloneNode(declar);

// 1) will set this.declarations[name] no matter helper exists
const uid = (this.declarations[name] = this.scope.generateUidIdentifier(
   name,
));
```

so i added a `helpers.ensure(name)` before `(1)`.

Since `loadHelper` is cached, it's okay to call it multiple times.

### DIFF
--- a/packages/babel-core/src/transformation/file/file.js
+++ b/packages/babel-core/src/transformation/file/file.js
@@ -203,6 +203,9 @@ export default class File {
       if (res) return res;
     }
 
+    // make sure that the helper exists
+    helpers.ensure(name);
+
     const uid = (this.declarations[name] = this.scope.generateUidIdentifier(
       name,
     ));
--- a/packages/babel-core/test/api.js
+++ b/packages/babel-core/test/api.js
@@ -788,4 +788,28 @@ describe("api", function() {
       );
     });
   });
+
+  describe("missing helpers", function() {
+    it("should always throw", function() {
+      expect(() =>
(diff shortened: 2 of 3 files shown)


nicolo-ribaudo:
Related: https://github.com/babel/babel/pull/10200

PR #10217, opened 2019-07-14
Class Private Static Accessors

### LINKED ISSUE #8052: Static Class Features: Stage 3

tim-mc:
| Fixed Issues?            | https://github.com/babel/proposals/issues/56, closes #8052
| Patch: Bug Fix?          | --
| Major: Breaking Change?  | --
| Minor: New Feature?      | Add class private static accessor support
| Documentation PR Link    | 

cc @robpalme @nicolo-ribaudo @littledan

### DIFF
--- a/packages/babel-helper-create-class-features-plugin/src/fields.js
+++ b/packages/babel-helper-create-class-features-plugin/src/fields.js
@@ -44,6 +44,7 @@ export function buildPrivateNamesNodes(privateNamesMap, loose, state) {
     // because static fields are directly assigned to a variable in the
     // buildPrivateStaticFieldInitSpec function.
     const { id, static: isStatic, method: isMethod, getId, setId } = value;
+    const isAccessor = getId || setId;
     if (loose) {
       initNodes.push(
         template.statement.ast`
@@ -51,7 +52,7 @@ export function buildPrivateNamesNodes(privateNamesMap, loose, state) {
         `,
       );
     } else if (isMethod && !isStatic) {
-      if (getId || setId) {
+      if (isAccessor) {
         initNodes.push(template.statement.ast`var ${id} = new WeakMap();`);
       } else {
         initNodes.push(template.statement.ast`var ${id} = new WeakSet();`);
@@ -140,10 +141,11 @@ const privateNameHandlerSpec = {
       getId,
       setId,
     } = privateNamesMap.get(name);
+    const isAccessor = getId || setId;
 
     if (isStatic) {
       const helperName =
-        isMethod && !(getId || setId)
+        isMethod && !isAccessor
           ? "classStaticPrivateMethodGet"
           : "classStaticPrivateFieldSpecGet";
 
@@ -155,7 +157,7 @@ const privateNameHandlerSpec = {
     }
 
     if (isMethod) {
-      if (getId || setId) {
+      if (isAccessor) {
         return t.callExpression(file.addHelper("classPrivateFieldGet"), [
           this.receiver(member),
           t.cloneNode(id),
@@ -183,10 +185,11 @@ const privateNameHandlerSpec = {
       setId,
       getId,
     } = privateNamesMap.get(name);
+    const isAccessor = getId || setId;
 
     if (isStatic) {
       const helperName =
-        isMethod && !(getId || setId)
+        isMethod && !isAccessor
           ? "classStaticPrivateMethodSet"
           : "classStaticPrivateFieldSpecSet";
 
@@ -305,10 +308,11 @@ function buildPrivateInstanceFieldInitSpec(ref, prop, privateNamesMap) {
 function buildPrivateStaticFieldInitSpec(prop, privateNamesMap) {
   const privateName = privateNamesMap.get(prop.node.key.id.name);
   const { id, getId, setId, initAdded } = privateName;
+  const isAccessor = getId || setId;
 
-  if (!prop.isProperty() && (initAdded || !(getId || setId))) return;
+  if (!prop.isProperty() && (initAdded || !isAccessor)) return;
 
-  if (getId || setId) {
+  if (isAccessor) {
     privateNamesMap.set(prop.node.key.id.name, {
       ...privateName,
       initAdded: true,
@@ -351,8 +355,8 @@ function buildPrivateMethodInitLoose(ref, prop, privateNamesMap) {
         });
       `;
   }
-
-  if (getId || setId) {
+  const isAccessor = getId || setId;
+  if (isAccessor) {
     privateNamesMap.set(prop.node.key.id.name, {
       ...privateName,
       initAdded: true,
@@ -373,9 +377,11 @@ function buildPrivateMethodInitLoose(ref, prop, privateNamesMap) {
 function buildPrivateInstanceMethodInitSpec(ref, prop, privateNamesMap) {
   const privateName = privateNamesMap.get(prop.node.key.id.name);
   const { id, getId, setId, initAdded } = privateName;
+
   if (initAdded) return;
 
-  if (getId || setId) {
+  const isAccessor = getId || setId;
+  if (isAccessor) {
     privateNamesMap.set(prop.node.key.id.name, {
       ...privateName,
       initAdded: true,
@@ -420,9 +426,11 @@ function buildPublicFieldInitSpec(ref, prop, state) {
 function buildPrivateStaticMethodInitLoose(ref, prop, state, privateNamesMap) {
   const privateName = privateNamesMap.get(prop.node.key.id.name);
   const { id, methodId, getId, setId, initAdded } = privateName;
+
   if (initAdded) return;
 
-  if (getId || setId) {
+  const isAccessor = getId || setId;
+  if (isAccessor) {
     privateNamesMap.set(prop.node.key.id.name, {
       ...privateName,
       initAdded: true,
--- a/packages/babel-helper-create-class-features-plugin/src/fields.js
+++ b/packages/babel-helper-create-class-features-plugin/src/fields.js
@@ -319,8 +319,8 @@ function buildPrivateStaticFieldInitSpec(prop, privateNamesMap) {
         // configurable is false by default
         // enumerable is false by default
         // writable is false by default
-        get: ${getId ? getId.name : "null"},
-        set: ${setId ? setId.name : "null"}
+        get: ${getId ? getId.name : prop.scope.buildUndefinedNode()},
+        set: ${setId ? setId.name : prop.scope.buildUndefinedNode()}
       }
     `;
   }
@@ -363,8 +363,8 @@ function buildPrivateMethodInitLoose(ref, prop, privateNamesMap) {
         // configurable is false by default
         // enumerable is false by default
         // writable is false by default
-        get: ${getId ? getId.name : "undefined"},
-        set: ${setId ? setId.name : "undefined"}
+        get: ${getId ? getId.name : prop.scope.buildUndefinedNode()},
+        set: ${setId ? setId.name : prop.scope.buildUndefinedNode()}
       });
     `;
   }
@@ -383,8 +383,8 @@ function buildPrivateInstanceMethodInitSpec(ref, prop, privateNamesMap) {
 
     return template.statement.ast`
       ${id}.set(${ref}, {
-        get: ${getId ? getId.name : "undefined"},
-        set: ${setId ? setId.name : "undefined"}
+        get: ${getId ? getId.name : prop.scope.buildUndefinedNode()},
+        set: ${setId ? setId.name : prop.scope.buildUndefinedNode()}
       });
     `;
   }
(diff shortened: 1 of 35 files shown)


tim-mc:
Still a WIP. Spec support has been added. Working on loose now.

tim-mc:
Loose support has been added

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -429,7 +471,47 @@ function buildPublicFieldInitSpec(ref, prop, state) {
 }
 
 function buildPrivateStaticMethodInitLoose(ref, prop, state, privateNamesMap) {
-  const { id, methodId } = privateNamesMap.get(prop.node.key.id.name);
+  const privateName = privateNamesMap.get(prop.node.key.id.name);
+  const { id, methodId, getId, setId, initAdded } = privateName;
+  if (initAdded) return;
```
JLHwung: Why do we need guard against double entrance? IMO this function should run once for every `prop`.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -300,8 +303,47 @@ function buildPrivateInstanceFieldInitSpec(ref, prop, privateNamesMap) {
 }
 
 function buildPrivateStaticFieldInitSpec(prop, privateNamesMap) {
-  const { id } = privateNamesMap.get(prop.node.key.id.name);
+  const privateName = privateNamesMap.get(prop.node.key.id.name);
+  const { id, getId, setId, initAdded } = privateName;
   const value = prop.node.value || prop.scope.buildUndefinedNode();
```
JLHwung: I suggest move this line immediately before where `value` is used so we can early return for non-accessor class private methods.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -300,8 +303,47 @@ function buildPrivateInstanceFieldInitSpec(ref, prop, privateNamesMap) {
...
+    });
+
+    if (getId && setId) {
+      return template.statement.ast`
+        var ${id.name} = {
+          // configurable is false by default
+          // enumerable is false by default
+          // writable is false by default
+          get: ${getId.name},
+          set: ${setId.name}
```
jridgewell: Why don't we just reuse this template and set the undefined ones to `null`? Having the key shouldn't be an issue.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -300,8 +303,47 @@ function buildPrivateInstanceFieldInitSpec(ref, prop, privateNamesMap) {
 }
 
 function buildPrivateStaticFieldInitSpec(prop, privateNamesMap) {
-  const { id } = privateNamesMap.get(prop.node.key.id.name);
+  const privateName = privateNamesMap.get(prop.node.key.id.name);
+  const { id, getId, setId, initAdded } = privateName;
   const value = prop.node.value || prop.scope.buildUndefinedNode();
+  if (!prop.isProperty() && (initAdded || !(getId || setId))) return;
```
jridgewell: When would it not be a property?

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -429,7 +471,47 @@ function buildPublicFieldInitSpec(ref, prop, state) {
...
+    });
+
+    if (getId && setId) {
+      return template.statement.ast`
+        Object.defineProperty(${ref}, ${id}, {
+          // configurable is false by default
+          // enumerable is false by default
+          // writable is false by default
+          get: ${getId.name},
+          set: ${setId.name}
```
jridgewell: Ditto reuse.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -615,14 +697,14 @@ export function buildFieldsInitNodes(
       case isStatic && isPrivate && isMethod && !loose:
         needsClassRef = true;
         staticNodes.push(
+          buildPrivateStaticFieldInitSpec(prop, privateNamesMap),
+        );
+        staticNodes.unshift(
```
jridgewell: Why unshift?

packages/babel-helper-create-class-features-plugin/src/fields.js
tim-mc: That works in [this case](https://github.com/babel/babel/pull/10217/files/14f480de578c5f6220a8f8d62a31dd3fe1833ce2#r305591019), but setting `null` as a getter/setter is a `TypeError`.

packages/babel-helper-create-class-features-plugin/src/fields.js
tim-mc: It wouldn't be a property if it's a static private accessor.

packages/babel-helper-create-class-features-plugin/src/fields.js
tim-mc: I use `unshift` so that the declaration of the getter and/or setter occurs before the usage of the accessor(s) (e.g., when they're defined as the values on an object). Here's what a test failure looks like if that's changed to `push`:

```
    babel/packages/babel-plugin-proposal-private-methods/test/fixtures/static-accessors/basic/exec.js: expect(received).toEqual(expected) // deep equality

    Expected: "top secret string"
    Received: undefined
       1 | class Cl {
       2 |   static getValue() {
       3 |     return babelHelpers.classStaticPrivateFieldSpecGet(Cl, Cl, _privateStaticFieldValue);
       4 |   }
       5 |
       6 |   static setValue() {
       7 |     babelHelpers.classStaticPrivateFieldSpecSet(Cl, Cl, _privateStaticFieldValue, "dank");
       8 |   }
       9 |
      10 | }
      11 |
      12 | var _PRIVATE_STATIC_FIELD = {
      13 |   writable: true,
      14 |   value: "top secret string"
      15 | };
      16 | var _privateStaticFieldValue = {
      17 |   get: _get_privateStaticFieldValue,
      18 |   set: _set_privateStaticFieldValue
      19 | };
      20 |
      21 | var _get_privateStaticFieldValue = function () {
      22 |   return babelHelpers.classStaticPrivateFieldSpecGet(Cl, Cl, _PRIVATE_STATIC_FIELD);
      23 | };
      24 |
      25 | var _set_privateStaticFieldValue = function (newValue) {
      26 |   babelHelpers.classStaticPrivateFieldSpecSet(Cl, Cl, _PRIVATE_STATIC_FIELD, `Updated: ${newValue}`);
      27 | };
      28 |
      29 | expect(Cl.getValue()).toEqual("top secret string");
      30 | Cl.setValue();
      31 | expect(Cl.getValue()).toEqual("Updated: dank");
```

packages/babel-helper-create-class-features-plugin/src/fields.js
tim-mc: Updated.

packages/babel-helper-create-class-features-plugin/src/fields.js
tim-mc: Updated.

packages/babel-helper-create-class-features-plugin/src/fields.js
jridgewell: Ah, yah. We should be able to use `undefined` in all cases instead.

packages/babel-helper-create-class-features-plugin/src/fields.js
jridgewell: Can we change the get/set variables to function declarations instead? If not, this is still fine.

packages/babel-helper-create-class-features-plugin/src/fields.js
tim-mc: Good point, updated.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -300,9 +303,29 @@ function buildPrivateInstanceFieldInitSpec(ref, prop, privateNamesMap) {
...
+      ...privateName,
+      initAdded: true,
+    });
+
+    return template.statement.ast`
+      var ${id.name} = {
+        // configurable is false by default
+        // enumerable is false by default
+        // writable is false by default
+        get: ${getId ? getId.name : "null"},
```
jridgewell: Nit: Let's change this to undefined as well.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -335,35 +358,15 @@ function buildPrivateMethodInitLoose(ref, prop, privateNamesMap) {
...
-          set: ${setId.name}
-        });
-      `;
-    }
+    return template.statement.ast`
+      Object.defineProperty(${ref}, ${id}, {
+        // configurable is false by default
+        // enumerable is false by default
+        // writable is false by default
+        get: ${getId ? getId.name : "undefined"},
```
jridgewell: Nit: `undefined` is a binding, not a primitive. Use `scope.buildUndefinedNode()` instead.

packages/babel-helper-create-class-features-plugin/src/fields.js
nicolo-ribaudo: Nit: could you extract `getId || setId` to an `isAccessor` boolean variable?
I find it hard to read all these nested boolean expressions.

PR #10219, opened 2019-07-14
Added flow to babel cli

Letladi:
added flow to the babel-cli package code.


packages/babel-cli/src/babel-external-helpers.js
```diff
@@ -1,9 +1,15 @@
+// @flow
+
 import commander from "commander";
+import isString from "lodash/isString";
 import { buildExternalHelpers } from "@babel/core";
 
-function collect(value, previousValue): Array<string> {
+function collect(
+  value: String | any,
```
nicolo-ribaudo: This isn't a string object, it's a `string`

packages/babel-cli/src/babel-external-helpers.js
```diff
@@ -1,9 +1,15 @@
...
+  previousValue: Array<string>,
+): Array<string> {
   // If the user passed the option with no value, like "babel-external-helpers --whitelist", do nothing.
-  if (typeof value !== "string") return previousValue;
+  if (!isString(value)) return previousValue;
```
nicolo-ribaudo: Why is this needed?

packages/babel-cli/src/babel-external-helpers.js
Letladi: Oh yes. Sorry, that is my mistake. Will fix

packages/babel-cli/src/babel-external-helpers.js
Letladi: It isn't 🙈
Will fix.

packages/babel-cli/src/babel/file.js
```diff
@@ -91,25 +104,27 @@ export default async function({ cliOptions, babelOptions }) {
...
-      process.stdin.setEncoding("utf8");
+        process.stdin.setEncoding("utf8");
 
-      process.stdin.on("readable", function() {
-        const chunk = process.stdin.read();
-        if (chunk !== null) code += chunk;
-      });
+        process.stdin.on("readable", function() {
+          const chunk = process.stdin.read();
+          if (isString(chunk)) code += toString(chunk);
```
JLHwung: We don't need `isString` and `toString` here.

Note that `chunk` could also be a `Buffer` ([docs](https://nodejs.org/dist/latest-v12.x/docs/api/stream.html#stream_event_data)), so `isString(chunk)` is not equivalent to `chunk !== null`.

`code += chunk` will implicitly convert `chunk` to string since `code` is a string.

packages/babel-cli/src/babel/file.js
```diff
@@ -184,15 +199,14 @@ export default async function({ cliOptions, babelOptions }) {
           }
 
           console.error(err);
-          return null;
```
JLHwung: I would suggest leave `return null` here so that we can make this PR focus on adding flow type annotation _without_ changing any internal logic.

Note that you can use union type to specify this function returns either a `Promise<Object>` or `null`.
```js
_filenames.map(async function(filename: string): Promise<Object> | null {
```

packages/babel-cli/src/babel/file.js
Letladi: I can remove `isString`, but without `toString` flow will throw an error because it sees `Buffer` being incompatible with `String`

nicolo-ribaudo: We can `// $FlowIgnore` it

Letladi: Alright.

packages/babel-cli/src/babel/file.js
Letladi: Good point. I'll make the fix.

packages/babel-cli/src/babel/file.js
```diff
@@ -6,10 +8,19 @@ import path from "path";
...
+  code: string,
+  map: Object,
+};
+
+export default async function({
+  cliOptions,
+  babelOptions,
+}: CmdOptions): Promise<void> {
+  function buildResult(fileResults: Array<Object>): CompilationOutput {
+    const map: Object = new sourceMap.SourceMapGenerator({
```
nicolo-ribaudo: Doesn't flow already infer this type to the `SourceMapGenerator` class? (defined at https://github.com/babel/babel/blob/7dc5fdb39091b1e9249bcc20487e5cc2aefa44b6/lib/third-party-libs.js.flow#L147)

Letladi: It does, but the `fromObject` [method](https://github.com/babel/babel/blob/7dc5fdb39091b1e9249bcc20487e5cc2aefa44b6/lib/third-party-libs.js.flow#L180) expects a source but we pass it SourceMapGenerator, so maybe I should add `SourceMap | SourceMapGenerator` to that module declaration?

Letladi: I'm thinking something like this `fromObject(obj: SourceMap | SourceMapGenerator): Converter,'. This works without errors.

nicolo-ribaudo: I think that `fromObject(obj: SourceMap)` was correct, and it isn't intended to be called with a `SourceMapGenerator` object.
It should be `.fromObject(map.toJSON())`.

Letladi: Alright. I will make that change.

JLHwung:
Could you rebase your branch? Test fails because we have moved our fork of lerna into babel organization.

Letladi:
@JLHwung, will do.

Letladi:
Sorry, I'm still working on the rebase (haven't done this before)

Letladi:
I'm closing this. I will send in a new pull request shortly.
The rebasing is showing me flames. 🤦🏽‍♂️

nicolo-ribaudo:
I suggest squashing this commits before rebasing; it makes it easier to do.

Letladi:
I actually deleted everything. forked the repo again and added my changes; #10244

PR #10242, opened 2019-07-19
fix: normalize moduleName of absolute runtime

### LINKED ISSUE #10209: Failing tests on a windows machine due to path format
When running `yarn jest transform-runtime`
The `packages/babel-plugin-transform-runtime/test/index.js` tests fails with (not the backslashes `\` -> `//`):
```
- var _classCallCheck = require("<CWD>/packages/babel-plugin-transform-runtime/node_modules/@babel/runtime/helpers/classCallCheck");
+ var _classCallCheck = require("<CWD>\\packages\\babel-plugin-transform-runtime\\node_modules\\@babel\\runtime/helpers/classCallCheck");
```
When running `make test`, all of these fail
- packages/babel-core/test/config-chain.js
- packages/babel-preset-env-standalone/test/babel-preset-env.js
- packages/babel-generator/test/index.js
- packages/babel-plugin-transform-react-jsx-source/test/index.js
- packages/babel-preset-react/test/preset-options.js
- packages/babel-plugin-transform-runtime/test/index.js
- packages/babel-preset-env/test/debug-fixtures.js
- packages/babel-generator/test/index.js

Windows 7 (x64)

JLHwung:
| Fixed Issues?            | Fixes #10209
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Nope 😢
| Documentation PR Link    | 
| Any Dependency Changes?  |

The moduleName of resolved absolute runtime should be normalized, otherwise it will return
`PATH\\TO\\MODULE` on Windows 7, which is apparently incorrect.

I don't know how to add a unit test given that CI runs on Linux only.

### DIFF
--- a/packages/babel-plugin-transform-runtime/src/index.js
+++ b/packages/babel-plugin-transform-runtime/src/index.js
@@ -10,9 +10,9 @@ import { typeAnnotationToString } from "./helpers";
 
 function resolveAbsoluteRuntime(moduleName: string, dirname: string) {
   try {
-    return path.dirname(
-      resolve.sync(`${moduleName}/package.json`, { basedir: dirname }),
-    );
+    return path
+      .dirname(resolve.sync(`${moduleName}/package.json`, { basedir: dirname }))
+      .replace(/\\/g, "/");
   } catch (err) {
     if (err.code !== "MODULE_NOT_FOUND") throw err;
 
--- a/packages/babel-plugin-transform-runtime/src/index.js
+++ b/packages/babel-plugin-transform-runtime/src/index.js
@@ -10,9 +10,9 @@ import { typeAnnotationToString } from "./helpers";
 
 function resolveAbsoluteRuntime(moduleName: string, dirname: string) {
   try {
-    return path
-      .dirname(resolve.sync(`${moduleName}/package.json`, { basedir: dirname }))
-      .replace(/\\/g, "/");
+    return path.posix.dirname(
+      resolve.sync(`${moduleName}/package.json`, { basedir: dirname }),
+    );
   } catch (err) {
     if (err.code !== "MODULE_NOT_FOUND") throw err;
 


JLHwung:
@sag1v Could you test on your machine to see if the test is now fixed on this branch?

```
git fetch origin refs/pull/10242/head
git checkout FETCH_HEAD
// now run the test
yarn jest transform-runtime
```

sag1v:
@JLHwung When i run `git fetch origin refs/pull/10242/head`
I get 
> fatal: Couldn't find remote ref refs/pull/10242/head

I think its because i'm on a fork and not the actual repo?

JLHwung:
> I think its because i'm on a fork and not the actual repo?

Oh I just assume everyone is cloning babel/babel, in this case you can add it as a new remote

```
git remote add upstream https://github.com/babel/babel
git fetch upstream refs/pull/10242/head
```

> maybe i could just clone @JLHwung's fork and test it...

Surely you can. Note that `git remote add` would save your time on `make bootstrap` and `yarn`.

sag1v:
Thanks @JLHwung !
As for the tests, still failing.
When i run `yarn jest transform-runtime` i get 2 failing tests:
```
- var _classCallCheck = require("<CWD>/packages/babel-plugin-transform-runtime/node_modules/@babel/runtime/helpers/classCallCheck");
+ var _classCallCheck = require("./helpers/classCallCheck");
```
```
- var _classCallCheck = require("<CWD>/packages/babel-plugin-transform-runtime/node_modules/@babel/runtime/helpers/classCallCheck");
+ var _classCallCheck = require("./helpers/classCallCheck");
```

And when i run `make test` i get 13 faild tests:
Example:
```
- "fixtures/sourcemaps/call-identifiers/input.js",
+ "fixtures\\sourcemaps\\call-identifiers\\input.js",
```

JLHwung:
> When i run yarn jest transform-runtime i get 2 failing tests:

Interesting. It seems like I have replaced a bug with another one. I will investigate later.

<blockquote>
<code>
- "fixtures/sourcemaps/call-identifiers/input.js",
+ "fixtures\\sourcemaps\\call-identifiers\\input.js",
</code>
</blockquote>

This one is related to how sourcemap is generated.

@sag1v Thank you for posting test results. I guess I had better get a Windows environment running and go through these issues.

JLHwung:
@nicolo-ribaudo My perspective on this PR has changed: Even if we _may_ normalize the path separator in the import call, the transform result injected with absolute path can only run on Windows -- we cannot get rid of the leading disk label anyway.

As Windows build of Node.js support both `\` and `/` in module resolution. The transformed result is effectively working on Windows machine. Therefore I think we don't need fix this.

The test failure is _expected_, as the behavior of injecting absolute path runtime, is os-dependent.
If we agree, we should close this PR and mark it as won't fix.

@sag1v There are other issues, i.e. source maps that may need fix. The discussions above focus only on the `absoluteRuntime` of `transform-runtime`. We may track these issues on different thread.

nicolo-ribaudo:
I'm ok with closing it as "wontfix", but I'd really like to have an option to skip running a test on some platform. Something like this:
```json
{
  "plugins": [ /* ... */ ],
  "os": ["Linux"]
}
```
Similar to `minNodeVersion`:
https://github.com/babel/babel/blob/07eaa3c63ff06edbeb14b4cc30aca44d294a4731/packages/babel-helper-fixtures/src/index.js#L167

JLHwung:
@nicolo-ribaudo Good idea. Let's discuss the `os` field design on #10252 . Closing this PR now as wontfix.

PR #10243, opened 2019-07-19
perf: always return `void 0` as undefined node

JLHwung:
| Fixed Issues?            |
| Patch: Bug Fix?          | Performance tweak
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Effectively, `this.hasBinding("undefined")` will always returns true since `undefined` is a global property. As `void 0` is safer than `undefined`, here we remove the `if` conditional check. `buildUndefinedNode` will be a bit faster as it doesn't have to look through the nested scopes.

I would also suggest that we move `buildUndefinedNode` to babel-types and clearly state that it is not relevant to the Scope class.

### DIFF
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -490,12 +490,7 @@ export default class Scope {
   }
 
   buildUndefinedNode() {
-    if (this.hasBinding("undefined")) {
-      return t.unaryExpression("void", t.numericLiteral(0), true);
-    } else {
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
> I would also suggest that we move buildUndefinedNode to babel-types and clearly state that it is not relevant to the Scope class.

Maybe in the next major?

JLHwung:
> Maybe in the next major?

Sure. We would not get rid of it on 7.x

PR #10244, opened 2019-07-19
added flow to babel cli

Letladi:
Added flow to the  babel-cli package.

### DIFF
--- a/packages/babel-cli/src/babel/file.js
+++ b/packages/babel-cli/src/babel/file.js
@@ -76,7 +76,7 @@ export default async function({
       babelOptions.sourceMaps === "inline" ||
       (!cliOptions.outFile && babelOptions.sourceMaps)
     ) {
-      code += "\n" + convertSourceMap.fromObject(map.toJSON()).toComment();
+      code += "\n" + convertSourceMap.fromObject(map).toComment();
     }
 
     return {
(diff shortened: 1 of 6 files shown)


Letladi:
I re-created my changes from the latest master branch. I don't know what changes I need to make to make the CI checks pass.

JLHwung:
> I don't know what changes I need to make to make the CI checks pass.

Test is good as long as Travis CI passes. Sometimes CircleCI returns flaky result.

packages/babel-cli/src/babel/file.js
```diff
@@ -65,7 +76,7 @@ export default async function({ cliOptions, babelOptions }) {
       babelOptions.sourceMaps === "inline" ||
       (!cliOptions.outFile && babelOptions.sourceMaps)
     ) {
-      code += "\n" + convertSourceMap.fromObject(map).toComment();
+      code += "\n" + convertSourceMap.fromObject(map.toJSON()).toComment();
```
JLHwung: `toJSON()` will incur extra Object memory allocation here given that `convertSourceMap` can work with `SourceMapGenerator` to generate a base64 comment.

I would suggest we extend the signature of `fromObject` to state that it can work with `SourceMapGenerator`.
https://github.com/babel/babel/blob/fced5cea430cc00e916876b663a8d2a84a5dad1f/lib/third-party-libs.js.flow#L180

Letladi: Alright. I can make that change.

PR #10258, opened 2019-07-23
Add ";" token to the end of TS construct & call signature declaration

skyiea:
| Fixed Issues?            | 
| Documentation PR Link    | --
| Any Dependency Changes?  |


JLHwung:
Could you also check if the printer will add `;` for the `CallSignature`?
```ts
interface a {
    (a: string) : number
}
```

Per [Object Type Literals Grammar], only the last `TypeMember` of `TypeMemberList` can omit trailing semicolon. I suggest we add another `TypeMember` to the test case so that `;` can not be omitted.
```ts
interface a {
    new (a: string): void
    new (a: number): void
}
```

[Object Type Literals Grammar]: https://github.com/microsoft/TypeScript/blob/master/doc/spec.md#383-object-type-literals

Issue #10260, opened 2019-07-24
NullishCoalescing forbids mixing with LogicalExpressions without parenthesis


jridgewell:
## Bug Report

**Current Behavior**
The nullish coalescing operator (`??`) currently allows mixing with the other logical operators (`||` and `&&`) without parenthesis. This is invalid per the latest specs, parenthesis are required to mix them:

```js
// Currently valid but should be a syntax error
a ?? b || c

// Currently valid but should be a syntax error
a || b ?? c

// Currently valid but should be a syntax error
a ?? b && c

// Currently valid but should be a syntax error
a && b ?? c
```

https://babeljs.io/repl/#?babili=false&browsers=Safari%2011&build=&builtIns=false&spec=false&loose=false&code_lz=PTAEGEFcCdoUwHYBcA2BPUA3AhiglgCagBGkSoAzgBYD2kKRxco2lay2AHqHLDdAChWAfmElQAHwmgAxgIEgIMeMnRZchEmUq16jZqwrskXHn0Gsp40bPmKosRKgw58jbdToMSBth2680PxCoDbEoABkEbYKYA4qzupuWuSeej4sfiYB5iFR1mJyQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=true&targets=&version=7.5.5&externalPlugins=%40babel%2Fplugin-proposal-nullish-coalescing-operator%407.4.4

**Expected behavior/code**
These should each throw a syntax error when parsing. It requires parens to disambiguate the intended grouping.

```js
// Valid according to spec
a ?? (b || c)

// Valid according to spec
(a ?? b) || c

// Valid according to spec
(a || b) ?? c

// Valid according to spec
a || (b ?? c)

// Valid according to spec
a ?? (b && c)

// Valid according to spec
(a ?? b) && c

// Valid according to spec
(a && b) ?? c

// Valid according to spec
a && (b ?? c)
```

**Babel Configuration (.babelrc, package.json, cli command)**

```js
{
  "plugins": ["proposal-nullish-coalescing-operator"]
}
```

**Environment**
- Babel version(s): 7

**Possible Solution**
These should throw during parsing. Somewhere around https://github.com/babel/babel/blob/fced5cea430cc00e916876b663a8d2a84a5dad1f/packages/babel-parser/src/parser/expression.js#L383-L390.

[...]

PR #10269, opened 2019-07-25
Fix parenthesis for nullish coalescing

vivek12345:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/10260 
| Documentation PR Link    | 
| Any Dependency Changes?  | NO

With this change all the following operations which are invalid will throw error and ask to add parenthesis to them
```
// will throw syntax error
a ?? b || c

// will throw syntax error
a || b ?? c

// will throw syntax error
a ?? b && c

// will throw syntax error
a && b ?? c
```


packages/babel-parser/test/fixtures/experimental/nullish-coalescing-operator/and-nullish/input.js
```diff
@@ -1 +1 @@
-a && b ?? c;
+(a && b) ?? c;
```
vivek12345: test updated to pass and new failing test cases added

packages/babel-parser/test/fixtures/experimental/nullish-coalescing-operator/no-paren-and-nullish/input.js
```diff
@@ -0,0 +1 @@
+c && d ?? e;
```
vivek12345: new failing test cases fixtures

packages/babel-parser/src/parser/expression.js
```diff
@@ -380,6 +380,58 @@ export default class ExpressionParser extends LValParser {
...
 
+        /* this check is needed because for operations like
+         * a ?? b || c
+         * a ?? b => This is considered as a logical expression in the ast tree
+         * c => Identifier
+         * so we need to check, only for || cases that if left is logical expression without paren
+         * then raise an error
+         */
+
+        if (op === tt.logicalOR) {
```
jridgewell: We can actually avoid this by lowering the precedence of the `??` operator compared to `||`. You can do that by raising the `binop` for the operators in https://github.com/babel/babel/blob/4d30379d3657142cb079a41f37642f519d699593/packages/babel-parser/src/tokenizer/types.js#L142-L156. (This is actually how the spec is implemented, `??` has lower precedence than `||` and the rest)

packages/babel-parser/src/parser/expression.js
```diff
@@ -380,6 +380,58 @@ export default class ExpressionParser extends LValParser {
...
+         */
+        if (op === tt.nullishCoalescing) {
+          if (
+            left.type === "LogicalExpression" &&
+            left.operator !== "??" &&
+            !(left.extra && left.extra.parenthesized)
+          ) {
+            throw this.raise(
+              left.start,
+              `Wrap left hand side of the expression in parenthesis`,
```
jridgewell: Let's change this message, since it's not clear if the `??` was meant to be a group or the other operator was meant to be a group. I'd suggest "Nullish coalescing operator requires parens when mixing with logical operators", but I'm not sure end users will understand what nullish coalescing operator is or what the other logical operators are.

For reference, the exponentiation operator uses "Wrap left hand side or entire exponentiation in parentheses."

packages/babel-parser/test/fixtures/experimental/nullish-coalescing-operator/no-paren-nullish-or/input.js
```diff
@@ -0,0 +1 @@
+e ?? f ?? g || h;
```
jridgewell: Let's add an explicit test case for `a ?? b ?? c` not requiring parens.

vivek12345: a test for this already exists here
https://github.com/babel/babel/blob/master/packages/babel-parser/test/fixtures/experimental/nullish-coalescing-operator/associativity/input.js#L1

packages/babel-parser/src/parser/expression.js
vivek12345: Thank you @jridgewell. As per your suggestion adding this as a message
```
Nullish coalescing operator(??) requires parens when mixing with logical operators
```

packages/babel-parser/src/tokenizer/types.js
```diff
@@ -139,21 +139,21 @@ export const types: { [name: string]: TokenType } = {
...
-  bitwiseXOR: createBinop("^", 4),
-  bitwiseAND: createBinop("&", 5),
-  equality: createBinop("==/!=/===/!==", 6),
-  relational: createBinop("</>/<=/>=", 7),
-  bitShift: createBinop("<</>>/>>>", 8),
-  plusMin: new TokenType("+/-", { beforeExpr, binop: 9, prefix, startsExpr }),
-  modulo: createBinop("%", 10),
-  star: createBinop("*", 10),
-  slash: createBinop("/", 10),
+  logicalOR: createBinop("||", 2),
```
vivek12345: priority of other operators incremented by +1 so that they have more priority over the `nullishCoalescing` operator which is as per spec.

packages/babel-parser/src/parser/expression.js
vivek12345: Thank you @jridgewell. This change has been made.

packages/babel-parser/src/tokenizer/types.js
vivek12345: It seems like this incrementing of priority broke this test case
```
FAIL  packages/babel-generator/test/index.js
  ● generation/typescript › cast as
```

nicolo-ribaudo: Why is priority important? It can be the same. `a ?? b && c` is an error anyway, regardless of which has higher priority.

vivek12345: @nicolo-ribaudo earlier `||` and `??` had same priority which was not right. `??` should have lower priority then `||`.

packages/babel-parser/test/fixtures/experimental/nullish-coalescing-operator/no-paren-nullish-or/input.js
jridgewell: Perfect.

packages/babel-parser/src/tokenizer/types.js
jridgewell: See the discussion at https://github.com/babel/babel/pull/10269#discussion_r307572385.

packages/babel-parser/src/tokenizer/types.js
```diff
@@ -139,21 +139,21 @@ export const types: { [name: string]: TokenType } = {
...
+  logicalAND: createBinop("&&", 3),
+  bitwiseOR: createBinop("|", 4),
+  bitwiseXOR: createBinop("^", 5),
+  bitwiseAND: createBinop("&", 6),
+  equality: createBinop("==/!=/===/!==", 7),
+  relational: createBinop("</>/<=/>=", 8),
```
jridgewell: Oh, we'll also need to increase the `in` and `instanceof` operator precedence to match the relational precedence ([spec](https://tc39.es/ecma262/#sec-relational-operators)). That should fix this `typescript/cast/as` test case, which is currently incorrect.

packages/babel-parser/test/fixtures/typescript/cast/as/output.json
```diff
@@ -236,7 +236,7 @@
           }
         },
         "expression": {
-          "type": "BinaryExpression",
+          "type": "TSAsExpression",
```
jridgewell: This change is incorrect.

```js
// input
x === 1 as number;

// Before
x === (1 as number);

// After
(x === 1) as number;
```

It's caused by the  `in` operator having a difference precedence than the relational operators. We should add a test case that `types. _in.binop === types.relational.binop`.

packages/babel-parser/src/tokenizer/types.js
vivek12345: For this https://github.com/babel/babel/pull/10269#discussion_r307880391
In the comment you had mentioned to raise the binOp for all other operators in order to lower the binOp for ??.
That is done and after that I faced the cast as error.
That incrementing the binOp for all operators is that correct?

packages/babel-parser/test/fixtures/typescript/cast/as/output.json
vivek12345: Understood. Fixing the error and adding a test case now.

packages/babel-generator/test/index.js
```diff
@@ -298,6 +298,29 @@ describe("generation", function() {
...
+      t.identifier("a"),
+      t.logicalExpression(
+        "&&",
+        t.identifier("b"),
+        t.logicalExpression("&&", t.identifier("c"), t.identifier("d")),
+      ),
+    );
+    const output = generate(nullishCoalesc).code;
+    expect(output).toBe(`a ?? (b && c && d)`);
+  });
```
jridgewell: Nit: Newline

packages/babel-generator/test/index.js
```diff
@@ -298,6 +298,29 @@ describe("generation", function() {
...
+  it("should add parenthesis when NullishCoalescing is used along with ||", function() {
+    // https://github.com/babel/babel/issues/10260
+    const nullishCoalesc = t.logicalExpression(
+      "??",
+      t.logicalExpression("||", t.identifier("a"), t.identifier("b")),
+      t.identifier("c"),
+    );
+    const output = generate(nullishCoalesc).code;
+    expect(output).toBe(`(a || b) ?? c`);
+  });
```
jridgewell: Nit: newline

vivek12345:
@jridgewell The build failed with this error
```
```
Can we restart the build?

packages/babel-generator/src/node/index.js
```diff
@@ -96,6 +96,18 @@ export function needsParens(node, parent, printStack) {
   if (t.isNewExpression(parent) && parent.callee === node) {
     if (isOrHasCallExpression(node)) return true;
   }
```
jridgewell: Nit: newline

packages/babel-generator/test/index.js
```diff
@@ -298,6 +298,29 @@ describe("generation", function() {
 });
 
 describe("programmatic generation", function() {
+  it("should add parenthesis when NullishCoalescing is used along with ||", function() {
```
vivek12345: test cases for checking programmatic creation of NullishCoalescing with && ||

packages/babel-generator/test/index.js
vivek12345: my editor settings :( I will remove the new lines added.

vivek12345: Seems I mis understood the statement. What you meant was add a new line between the tests. Adding them now.

packages/babel-parser/test/unit/tokenizer/types.js
```diff
@@ -0,0 +1,13 @@
+import { types } from "../../../src/tokenizer/types";
+
+describe("token types", () => {
+  it("should check if the binOp for relational === in", () => {
+    expect(types.relational.binop).toEqual(types._in.binop);
+  });
+  it("should check if the binOp for relational === instanceOf", () => {
```
jridgewell: Nit: newlines here as well.

vivek12345: Done 👍

vivek12345:
Thank you @jridgewell for your patience and the kind words. It means a lot. First time contribution to babel has been delight with your help and support 🎉

jridgewell:
Nullish coalescing is only available if you enable `proposal-nullish-coalescing-operator`, so I think people are aware that this is still a proposal. The change has an explicit message on how to fix, too. Might be ok to release as a minor with a very bold warning.

packages/babel-parser/src/parser/expression.js
```diff
@@ -380,6 +380,39 @@ export default class ExpressionParser extends LValParser {
...
+          } else if (
+            node.right.type === "LogicalExpression" &&
+            node.right.operator !== "??" &&
+            !(node.right.extra && node.right.extra.parenthesized)
+          ) {
+            throw this.raise(
+              node.right.start,
+              `Nullish coalescing operator(??) requires parens when mixing with logical operators`,
+            );
+          }
```
KFlash: @jridgewell @vivek12345 Can I ask why you are checking the AST node property? It should be enough to use current token as a func arg and validate against it. That way you 1) improve performance 2) skip duplicate code 3) simplify

This will also improve the performance.

I did it that way when I implemented similar in [my own parser Meriyah](https://github.com/meriyah/meriyah/blob/master/src/parser.ts#L3268-L3270) a

nicolo-ribaudo:
Regarding https://github.com/babel/babel/pull/10269#pullrequestreview-267464010, I think that we should merge it since

1) Most people would probably have already used parentheses, to improve readability
2) Prettier did already add the now required parenthesis

packages/babel-parser/src/parser/expression.js
nicolo-ribaudo: It would only work for `a && b ?? c`, but not for `a ?? b && c` since in the second case the precedence is `a ?? (b && c)`

nicolo-ribaudo: It would only work for `a && b ?? c`, but not for `a ?? b && c` since in the second case the precedence is `a ?? (b && c)`

KFlash: @nicolo-ribaudo I'm confused because it will work both lhs and rhs.  I implemented it that way I described. See [here](https://github.com/meriyah/meriyah/blob/master/src/parser.ts#L3406) . And it works very well to if you try it in the [REPL](https://meriyah.github.io/meriyah/)

montogeek:
What is left to be able to merge it?

nicolo-ribaudo:
We are going to release a minor version with this patch soon!

PR #10275, opened 2019-07-27
fix object rest in array pattern

### LINKED ISSUE #7215: proposal-object-rest-spread doesn't transform the rest/spread syntax in some cases
This looks superficially similar to https://github.com/babel/babel/issues/7099 but that one is supposedly [resolved in Babel 7](https://github.com/babel/babel/issues/7099#issuecomment-357065187) but my bug isn’t. Therefore I’m filing a new one.

Here is a complete repro:

```sh
npm i --save @babel/core@7.0.0-beta.37 @babel/plugin-proposal-object-rest-spread@7.0.0-beta.37
```

Then run in Node shell:

```js
console.log(
  require('@babel/core').transform(
`
const { ...rest1 } = {};
const [{ ...rest2 }] = [{}];
`,
    {
      plugins: ['@babel/plugin-proposal-object-rest-spread']
    }
  ).code
)
```

Notice how I have two different rest destructurings, but the second one is *also* inside an array destructuring.

Output:

```js
function _objectWithoutProperties(source, excluded) { if (source == null) return {}; var target = {}; var sourceKeys = Object.keys(source); var key, i; for (i = 0; i < sourceKeys.length; i++) { key = sourceKeys[i]; if (excluded.indexOf(key) >= 0) continue; target[key] = source[key]; } if (Object.getOwnPropertySymbols) { var sourceSymbolKeys = Object.getOwnPropertySymbols(source); for (i = 0; i < sourceSymbolKeys.length; i++) { key = sourceSymbolKeys[i]; if (excluded.indexOf(key) >= 0) continue; if (!Object.prototype.propertyIsEnumerable.call(source, key)) continue; target[key] = source[key]; } } return target; }

const rest1 = _objectWithoutProperties({}, []);

const [{ ...rest2
}] = [{}];
```

Notice how `...rest1` got transformed but `...rest2` is left intact.

I expect that `...rest2` is also transformed.

---

Original issue where this got reported: https://github.com/facebookincubator/create-react-app/issues/3694

### LINKED ISSUE #10257: Object rest destructuring doesn't work properly when nested inside array destructuring
## Bug Report

**Current Behavior**

**Input Code**
[REPL](https://babeljs.io/en/repl#?babili=false&browsers=last%202%20Chrome%20versions%2C%0Alast%202%20Firefox%20versions%2C%0Alast%202%20Edge%20versions%2C%0Alast%202%20Safari%20versions&build=&builtIns=false&spec=false&loose=false&code_lz=PTAEBcAsEsGdQOYFNz3AJwIYDtYAdoAbJAE1D3QHs8l1CBPAKGnCQFtYA6NzPACj4BvAEaZ0AGlCdp6JLHABfAJSgAvAD5Q2AK6FCSgNyNi4UCLGTpnWfIVrQADyOMQEGPBKU52AOThmrBzcvAIA2gBmlJSS5hJSMnKKALoqGlq6-kYmoBFRAEwxougF8daJeQpJ9k6MQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env&prettier=false&targets=&version=7.5.5&externalPlugins=)

```js
// this gets transpiled properly
items.map(({bar, ...rest}) => null);
let {bar, ...rest} = x;

// this doesn't
items.map(([foo, {bar, ...rest}]) => null);
let [foo2, {bar2, ...rest2}] = x;
```

**Expected behavior/code**
`[foo, {bar, ...rest}]` being transpiled since Edge doesn't support it.

tanhauhau:
| Fixed Issues?            | Fixes #7215, Fixes #10257
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Edit by @JLHwung : Replace `Fixes #7215, #10257` by `Fixes #7215, Fixes #10257`.


JLHwung:
@tanhauhau Could you rebase on the master branch? The linting errors have been fixed.

JLHwung:
Could you change _Fixed Issues_ to
```
Fixes #7215, Fixes #10257
```
I don't think GitHub support `Fixes` comma separated issues.

hzoo:
Yeah @JLHwung is correct, it's required to put a word after for each number https://help.github.com/en/articles/closing-issues-using-keywords for ref!

packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -401,27 +416,54 @@ export default declare((api, opts) => {
...
 
-          return;
+          const key = scope.generateUidIdentifier("ref");
+          node.left = t.variableDeclaration(left.kind, [
+            t.variableDeclarator(key, null),
+          ]);
+
+          path.ensureBlock();
+
+          node.body.body.unshift(
```
JLHwung: `node.body` could be an `EmptyStatement`
```js
for (const foo of []);
```

tanhauhau: `path.ensureBlock` has ensure that it will be a block statement.

packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -37,6 +37,17 @@ export default declare((api, opts) => {
     return foundRestElement;
   }
 
+  function hasObjectPatternRestElement(path) {
+    let foundRestElement = false;
+    visitRestElements(path, restElement => {
+      if (restElement.parentPath.isObjectPattern()) {
+        foundRestElement = true;
+        path.stop();
```
nicolo-ribaudo: Shouldn't this be `restElement.stop()`?

tanhauhau: if i understand correctly, `restElement.stop()` will stop the traversing into `restElement` but continue on siblings, but `path.stop` will stop the entire `visitRestElements` traversal?

nicolo-ribaudo: That should be the difference between `.skip()` and `.stop()`. `path` is not generated by `visitElemenrs`: you should use it when you want to stop the traversal which provided it.

tanhauhau: oh. understood.

I found out that this line was copied over from https://github.com/babel/babel/blob/088d2200e1c473622d0ac7790af7be29dd8b4dc1/packages/babel-plugin-proposal-object-rest-spread/src/index.js#L31-L38

does this means that this shouldnt use `path.stop()` either?

nicolo-ribaudo: Yeah, that looks like an error

PR #10283, opened 2019-07-30
`babel --watch` should have equivalent file selection logic with `babel`

### LINKED ISSUE #10270: babel --watch does initial compile but ignores changes when --extensions is missing
## Bug Report

1. Create a typscript file `foo.ts`
2. Run `babel foo.ts --out-file foo.js --watch --verbose`
3. Make a change to `foo.ts`

**Current Behavior**
Babel runs an initial compile, but does not react to any changes in the file.

**Expected behavior/code**
Babel should recompile the file when it changes.

If you add `--extensions ".ts"` to the command line, babel will properly recompile when the file changes. However, since an exact filename has been given on the command line, the `--extensions` option shouldn't be needed.

**.babelrc**

```json
{
    "presets": [
        "@babel/env",
        "@babel/typescript"
    ]
}
```

**Environment**
- Babel version(s): v7.5.5
- Node/npm version: Node 10.16.0 / npm 6.9.0
- OS: macOS Mojave 10.14.6
- Monorepo: no

JLHwung:
| Fixed Issues?            | Fixes #10270 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Nope
| Documentation PR Link    | 
| Any Dependency Changes?  |

In this PR, we align file selection logic of `--watch` to one time building. In
https://github.com/babel/babel/blob/f08062b1deaf34d28726d0ef825821d060cb8554/packages/babel-cli/src/babel/file.js#L144-L163
The cli options `--extensions` is used to filter compilable files if the `filename` argument is a directory. However the watch logic excludes all non-compilable files even if it is specified in `filename` argument.

We also introduce minor performance tweak by passing `disableGlobbing: true` into chokidar as `@babel/cli` does not support globbing.

Fixture-based unit test does not work here since we need separate process to trigger a file change.

### DIFF
--- a/packages/babel-cli/src/babel/file.js
+++ b/packages/babel-cli/src/babel/file.js
@@ -216,6 +216,7 @@ export default async function({
       const chokidar = util.requireChokidar();
       chokidar
         .watch(filenames, {
+          disableGlobbing: true,
           persistent: true,
           ignoreInitial: true,
           awaitWriteFinish: {
--- a/packages/babel-cli/src/babel/file.js
+++ b/packages/babel-cli/src/babel/file.js
@@ -224,7 +224,10 @@ export default async function({
           },
         })
         .on("all", function(type: string, filename: string): void {
-          if (!util.isCompilableExtension(filename, cliOptions.extensions)) {
+          if (
+            !util.isCompilableExtension(filename, cliOptions.extensions) &&
+            !filenames.includes(filename)
+          ) {
             return;
           }
 


packages/babel-cli/src/babel/file.js
```diff
@@ -213,6 +213,7 @@ export default async function({
       const chokidar = util.requireChokidar();
       chokidar
         .watch(filenames, {
+          disableGlobbing: true,
```
nicolo-ribaudo: This is just for perf, since the filenames are already resolved. Right?

nicolo-ribaudo:
@JLHwung Could you rebase and merge?

PR #10291, opened 2019-08-01
Always use `sourceFileName` instead of `sourceFilename`.

eventualbuddha:
| Fixed Issues?            | n/a
| Patch: Bug Fix?          |
| Major: Breaking Change?  | maybe
| Minor: New Feature?      |
| Tests Added + Pass?      | No, I'm not sure what changes are needed
| Documentation PR Link    | 
| Any Dependency Changes?  |

There seems to be some confusion around which of these two is correct within the codebase. This consolidates all instances to be `sourceFileName` since that one had more instances.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -1,5 +1,5 @@
 type BabelParserOptions = {
-  sourceFilename?: string;
+  sourceFileName?: string;
   sourceType?: "module" | "script";
   plugins?: Array<Object>;
 };
--- a/packages/babel-parser/src/parser/index.js
+++ b/packages/babel-parser/src/parser/index.js
@@ -26,7 +26,7 @@ export default class Parser extends StatementParser {
     this.inModule = this.options.sourceType === "module";
     this.scope = new ScopeHandler(this.raise.bind(this), this.inModule);
     this.plugins = pluginsMap(this.options.plugins);
-    this.filename = options.sourceFilename;
+    this.filename = options.sourceFileName;
   }
 
   // This can be overwritten, for example, by the TypeScript plugin.
--- a/packages/babel-parser/src/parser/index.js
+++ b/packages/babel-parser/src/parser/index.js
@@ -26,7 +26,7 @@ export default class Parser extends StatementParser {
     this.inModule = this.options.sourceType === "module";
     this.scope = new ScopeHandler(this.raise.bind(this), this.inModule);
     this.plugins = pluginsMap(this.options.plugins);
-    this.filename = options.sourceFileName;
+    this.filename = options.sourceFilename || options.sourceFileName;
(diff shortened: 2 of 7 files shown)


packages/babel-parser/src/parser/index.js
```diff
@@ -26,7 +26,7 @@ export default class Parser extends StatementParser {
     this.inModule = this.options.sourceType === "module";
     this.scope = new ScopeHandler(this.raise.bind(this), this.inModule);
     this.plugins = pluginsMap(this.options.plugins);
-    this.filename = options.sourceFilename;
+    this.filename = options.sourceFileName;
```
hzoo: Yeah I guess we wouldn't want to drop the other case until later (although I also don't expect it to be used often) as Logan said on Slack.

But I believe this is the only part that is breaking since the others are test cases, types, or local variables. We could do something like

```suggestion
    this.filename = options.sourceFilename || options.sourceFileName;
```

eventualbuddha: I think this is a good idea, but it still doesn't seem to make the tests pass.

packages/babel-parser/src/parser/index.js
```diff
@@ -26,7 +26,7 @@ export default class Parser extends StatementParser {
     this.inModule = this.options.sourceType === "module";
     this.scope = new ScopeHandler(this.raise.bind(this), this.inModule);
     this.plugins = pluginsMap(this.options.plugins);
-    this.filename = options.sourceFilename;
+    this.filename = options.sourceFilename || options.sourceFileName;
```
nicolo-ribaudo: Since the nullish coaleshing proposal is now stage 3, we could enable `@babel/plugin-proposal-nullish-coalescing-operator` and use `options.sourceFilename ?? options.sourceFileName` :sunglasses:

eventualbuddha:
I'm not sure what should be done about the test failures. Anyone else want to advise?

nicolo-ribaudo:
I'm not sure, maybe try rebasing it :thinking:

lib/parser.js
```diff
@@ -1,5 +1,5 @@
 type BabelParserOptions = {
-  sourceFilename?: string;
+  sourceFileName?: string;
```
nicolo-ribaudo: Isn't this a breaking change?

packages/babel-parser/src/parser/index.js
nicolo-ribaudo: It has been enabled on master, you just need to rebase.

eventualbuddha:
Not sure what should be done to make this work, so I'm just going to close it. If someone else wants to pick it up, feel free.

Issue #6226, opened 2017-09-11
copy-files flag does not respect ignore block in .babelrc


mattgoucher:
**Bug:** If in your build script, you have `--copy-files` passed, files that are ignored in .babelrc files will be copied over, instead of being ignored.

### Input Code
```js
babel src --out-dir lib --copy-files
```

### Babel Configuration (.babelrc, package.json, cli command)

```js
{
  "presets": [
    "es2015",
    "stage-1",
    "react"
  ],
  "ignore": [
    "**__mocks__/*.js"
  ]
}
```

### Expected Behavior
Babel should not copy files that are explicitly ignored in `.babelrc` configuration. In this case, babel should not transpile or copy Jest mocks.

### Current Behavior
Babel copies files that are ignored are explicitly `.babelrc`.
```sh
> npm run build 
> babel src --out-dir lib --copy-files
  src/api/__mocks__/index.js -> lib/api/__mocks__/index.js
  src/api/endpoints.js -> lib/api/endpoints.js
```
If you pass an `--ignore` flag (with the same blob) _after_ the `--copy-files` flag, it will behave as expected.
```sh
> npm run build;
> babel src --out-dir lib --copy-files --ignore **__mocks__/*.js
  src/api/endpoints.js -> lib/api/endpoints.js
  (output does not include ignored directories)
```

### Context
This issue comes up when you work with Jest manual mocks inside of your source directory. Jest warns that you have duplicate __mocks__ directories.

| software         | version(s)
| ---------------- | -------
| Babel            |  6.26.0
| node             | 6.9.2
| npm              | 5.4.0
| Operating System | OSX El Capitan

loganfsmyth:
The objective of `--copy-files` is to say "please copy the files that Babel doesn't process", so this is working as expected. If you only want it to copy the files Babel processes, just leave off `--copy-files`. If you need a bit of both, seems like you'd want to piece that together with Gulp.

mattgoucher:
@loganfsmyth That makes sense, I was curious if that was the expected. However, it's strange that it works differently when the `--ignore` flag is passed to the command, versus inside the .babelrc.

jridgewell:
Yup, that seems like a bug.

donaldpipowitch:
[...]

PR #10301, opened 2019-08-05
Filter "dot directories" in addition to "dot files" when `includeDotFiles` is falsy.

### LINKED ISSUE #10037: babel-cli chokes on VisualStudio lock file
## Bug Report

**Current Behavior**
When I run babel with `--copy-files` on a directory with a VisualStudio project that I have opened, babel tries to copy a file ...\sqlite3\db.lock from the .vs directory to the output folder. This file cannot be copied, and babel stops with an error.

I tried ignoring / excluding this folder in the babel config, but could not get that working (it tries to copy it anyway).

**Expected behavior/code**
I would expect that ignoring / excluding the `.vs` folder would skip the folder.
Alternatively, since `.vs` is a "dotfile", it should automatically skip it already.
(Looks like cli/util.js always passes directories, even dot-directories.)

**Environment**
- Babel version(s): v7.4.4
- Node/npm version: Node 10.15/npm 6.4
- OS: Windows 10

**Possible Solution**
Update the `ignoreDotFiles` to also skip dot-directories

rhpijnacker:
Fixes #10037

### DIFF
--- a/packages/babel-cli/src/babel/util.js
+++ b/packages/babel-cli/src/babel/util.js
@@ -20,10 +20,8 @@ export function readdir(
   return readdirRecursive(dirname, (filename, _index, currentDirectory) => {
     const stat = fs.statSync(path.join(currentDirectory, filename));
 
-    if (stat.isDirectory()) return true;
-
     return (
-      (includeDotfiles || filename[0] !== ".") && (!filter || filter(filename))
+      (includeDotfiles || filename[0] !== ".") && (stat.isDirectory() || !filter || filter(filename))
     );
   });
 }


JLHwung:
IMO we should not modify the behavior of `includeDotFiles: false`. The objective of `--copy-files` is to copy every file that babel does _not_ process. If you would like to ignore some specific folder, i.e. `.vs` in your cases during copy, then I would recommend you use gulp tasks to handler extra logics here. 

Above all, babel/cli should focus on code transforms and there are better tools to handle file copy.

Related: https://github.com/babel/babel/issues/6226#issuecomment-328660064

rhpijnacker:
As #10037 explains, I tried to exclude the `.vs` folder using Babel's exclusion mechanism but that does not work. 

| The objective of --copy-files is to copy every file that babel does not process.
This behaviour can be tweaked by `includeDotFiles`, so it is not necessarilly _every_ file. I do not think that excluding dot folders is so different.
@JLHwung to keep the semantics of `includeDotFiles` intact, how would you feel about an `includeDotFolders` option (set to `true` by default)?
That would be slightly more complicated to implement, but would work.

rhpijnacker:
Implemented a "--exclude-dotdirs" option.
@JLHwung is this a more acceptable approach?

JLHwung:
@rhpijnacker Thanks. My major concern here is that this feature may be open-ended. We try to avoid introducing extra file/folder copy logic into babel cli and I believe there are many better tools out there to handle file copy: Grunt/Gulp/Rollup, etc.

Besides, it may confuse users especially when there is `include/exclude` configurations out there in babel.config.js to specify which file should be transformed. I will be happy if I am wrong, and I will leave this PR to seek consensus in communities.

Back at the time, the `--include-dotfiles` is meant to be a fix when users expect `--copy-files` literally should copy every file.

rhpijnacker:
@JLHwung My initial implementation was `--include-dotdirs`, but as far as I can see there is no way to pass `true` / `false` to that and to have it set to `true` by default. That forced me to implement `--exclude-dotdirs` instead.

I'm still looking for a better (Babel) way to exclude this `.vs` folder, but so far without much luck.
I would really like to avoid having to introduce other tooling (Grunt/Gulp/etc.) just to copy the other files, when Babel can already to that for me.

What is the next step? If this PR does not get accepted, I will have look for alternative options, and it would be nice to know whether it will or will not get accepted.

PR #10315, opened 2019-08-09
fix: ensure `@babel/parser` can reference `@babel/types`

eventualbuddha:
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The type declarations for `@babel/parser` reference `@babel/types` but previously there was no explicit dependency on `@babel/parser`. This meant that TypeScript is unable to process `@babel/parser` unless `@babel/types` also happens to be installed. Furthermore, if installing using a stricter `node_modules` layout (like the one favored by `pnpm`), even if `@babel/types` is installed by another package `@babel/parser` will be unable to see it. With an explicit dependency this problem goes away.

See [discussion on Slack](https://babeljs.slack.com/archives/C062RC35M/p1565380370022200).

### DIFF
--- a/packages/babel-parser/package.json
+++ b/packages/babel-parser/package.json
@@ -35,5 +35,8 @@
   },
   "bin": {
     "parser": "./bin/babel-parser.js"
+  },
+  "dependencies": {
+    "@babel/types": "^7.5.5"
   }
 }


JLHwung:
@eventualbuddha Could you please rebase on master?

nicolo-ribaudo:
There is also #9924, where @loganfsmyth wrote a big valid concern.

eventualbuddha:
I agree with @loganfsmyth that it's a little strange to include a package as a dependency simply because you need the types. However, I don't see a good way around it. As-is, `@babel/parser` is broken when used by itself with respect to TypeScript.

eventualbuddha:
I’m unsure why this is failing. Looks like a git issue?

```
error Couldn't find match for "89eab830be040fbaabc31ad7206c5efab878c6c3" in "refs/heads/babel-collect-updates,refs/heads/backup,refs/heads/lerna-collect-updates,refs/heads/master,refs/heads/readme,refs/tags/v1.0.0…
```

nicolo-ribaudo:
Yeah sorry, we have fixed in on master

eventualbuddha:
More unrelated failures?

```
Summary of all failing tests
 FAIL  packages/babel-preset-env/test/fixtures.js
  ● Test suite failed to run

      at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:443:43)
      at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:513:25)
      at Object.<anonymous> (packages/babel-helper-fixtures/node_modules/lodash/_baseRest.js:1:43)
```

loganfsmyth:
It seems like if we actually wanted to fix this, we'd have to make a new package that was _just_ the  type definitions or something.

Can we just ask people using TS to install `@babel/types` alongside the parser if they want to use the type definitions?

eventualbuddha:
You could go back to using type definitions from DefinitelyTyped. Or you could also include a comment in the type declarations file at the point where TS shows an error message that instructs users to add `@babel/types` to their package as a `devDependency`.

Both of these feel kinda janky. I'm curious what the actual harm with this is, besides possibly adding a copy of `@babel/types` to someone's `node_modules` folder. If that's all it is, is that so bad?

PR #10319, opened 2019-08-11
Add a builder definition including name for tsTypeParameter

### LINKED ISSUE #10317: t.tsTypeParameter should take 3 parameters but throws
## Bug Report

**Current Behavior**
calling [t.tsTypeParameter](https://babeljs.io/docs/en/babel-types.html#tstypeparameter) with the name parameter yields `Error: TSTypeParameter: Too many arguments passed. Received 3 but can receive no more than 2`

**Input Code**
```js
const generate = require("@babel/generator").default;

const ast = t.tsTypeParameter(
  t.tsTypeReference(t.identifier("bar")),
  t.tsTypeReference(t.identifier("baz")),
  "foo"
);

const { code } = generate(ast);
console.log(code);
```

**Expected behavior/code**
should output `foo extends bar = baz`

N/A

**Environment**
- Babel version(s): 7.4.4
- Node/npm version: Node v10.12.0 / yarn 1.17.3
- OS: Ubuntu 18.04

**Possible Solution**
Add `name` to [visitor](https://github.com/babel/babel/blob/master/packages/babel-types/src/definitions/typescript.js#L497)

deificx:
| Fixed Issues?            | Fixes #10317 
| Documentation PR Link    | https://github.com/babel/website/pull/2080

Adds `name` to the visitor field for `TSTypeParameter`, this is a required field but since it is not listed here the assert fails rendering the function useless, since it cannot be called with either 2 or 3 arguments. 

Documentation: I found this when trying to run all the t.* [builder functions](https://babeljs.io/docs/en/babel-types.html), but my generation "fails" here because it doesn't look like [these arguments are required](https://deploy-preview-2080--babel.netlify.com/docs/en/babel-types#tstypeparameter), they are (which has been corrected in the current master). So I will need to re-run that generation once that fix is available.

### DIFF
--- a/packages/babel-types/src/definitions/typescript.js
+++ b/packages/babel-types/src/definitions/typescript.js
@@ -494,6 +494,7 @@ defineType("TSTypeParameterDeclaration", {
 });
 
 defineType("TSTypeParameter", {
+  builder: ["constraint", "default", "name"],
   visitor: ["constraint", "default"],
   fields: {
     name: {
--- a/packages/babel-types/test/builders/typescript/__snapshots__/tsTypeParameter.js.snap
+++ b/packages/babel-types/test/builders/typescript/__snapshots__/tsTypeParameter.js.snap
@@ -1,0 +1,24 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`builders typescript tsTypeParameter accept name as argument for tsTypeParameter 1`] = `
+Object {
+  "constraint": Object {
+    "type": "TSTypeReference",
+    "typeName": Object {
+      "name": "bar",
+      "type": "Identifier",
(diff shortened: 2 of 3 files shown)


packages/babel-types/src/definitions/typescript.js
```diff
@@ -494,7 +494,7 @@ defineType("TSTypeParameterDeclaration", {
 });
 
 defineType("TSTypeParameter", {
-  visitor: ["constraint", "default"],
+  visitor: ["constraint", "default", "name"],
```
nicolo-ribaudo: Only nodes can be visited, so we can't add `name` to `visitor`.
You can define `builder: ["constraint", "default", "name"],` instead.

PR #10325, opened 2019-08-12
Warn when the version specified in transform-runtime doesn't match

### LINKED ISSUE #10323: Flow comments plugin moves the generic type arguments of the left hand side of the `extends` keyword to its right side
## Bug Report

**Current Behavior**
`@babel/plugin-transform-flow-comments` moves the generic type arguments of the left hand side of the `extends` keyword to its right side

```js
class A extends B /*:: <X>*/ {}
```

**Input Code**
[REPL link](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=true&code_lz=MYGwhgzhAECCA8ANAfNApgDwC5oHYBMYAhaAbwF8AoIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=&prettier=false&targets=&version=7.5.5&externalPlugins=%40babel%2Fplugin-transform-flow-comments%407.5.5)

```js
class A<X> extends B {}
```

**Expected behavior/code**

```js
class A /*:: <X>*/ extends B {}
```

**Environment**
- Babel version(s): v7.5.5
- How you are using Babel: Online REPL

**Possible Solution**
```js

  return {
    //...
    visitor: {
      //...
      ClassDeclaration(path) {
        if (path.parentPath.isFlow()) {
          return;
        }

        const { node } = path;
        if (node.typeParameters) {
          const typeParameters = path.get("typeParameters");
          const id = path.get("id");
          id.addComment("trailing", generateComment(typeParameters, node));
          typeParameters.remove();
        }
      },
      //...
    },
  };
```

nicolo-ribaudo:
| Fixed Issues?            | Fixes #10323 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

We can't automatically infer the version of `@babel/runtime`, because there is a too high risk of introducing a breaking change by inferring the wrong version because of a dependency depending on `@babel/runtime`, but we can at least warn explaining the problem.
I suspect that currently almost everyone isn't using the `verison` option, leading to many helpers to be inlined.

I will extract 055b16c to another PR, since it will be useful for testing other warning messages, to clean up our tests console output and to remove the distinction between fixtures and debug-fixtures in preset-env.


packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -27,6 +28,30 @@ function resolveAbsoluteRuntime(moduleName: string, dirname: string) {
...
+    const runtimeDir = path.dirname(
+      resolve.sync(`${modulePath}/package.json`, {
+        basedir: dirname,
+      }),
+    );
+    const pkgFilename = path.resolve(runtimeDir, "../../../package.json");
+    const pkg = JSON.parse(fs.readFileSync(pkgFilename, "utf8"));
+
+    return (
+      pkg.dependencies?.[moduleName] ||
```
nicolo-ribaudo: It is stage 3 now :pray:

nicolo-ribaudo:
Rebased to include #10326

packages/babel-plugin-transform-runtime/test/fixtures/absoluteRuntime/relative/output.js
```diff
@@ -1,4 +1,4 @@
-var _classCallCheck = require("<CWD>/packages/babel-plugin-transform-runtime/node_modules/@babel/runtime/helpers/classCallCheck");
+var _classCallCheck = require("<CWD>/packages/babel-plugin-transform-runtime/test/node_modules/@babel/runtime/helpers/classCallCheck");
```
JLHwung: 🤔The extra `test` here looks strange.

nicolo-ribaudo: It's because I created a "fake" node_modules folder inside of `/test/`, to avoid relying on a "real" `@babel/runtime` which isn't part of the test suite and whose version will change.
For this reason, `transform-runtime` is resolving this new package.

JLHwung: Ahh, I see. Thank you for the explainer.

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -27,6 +28,30 @@ function resolveAbsoluteRuntime(moduleName: string, dirname: string) {
...
+  modulePath: string,
+  dirname: string,
+) {
+  try {
+    const runtimeDir = path.dirname(
+      resolve.sync(`${modulePath}/package.json`, {
+        basedir: dirname,
+      }),
+    );
+    const pkgFilename = path.resolve(runtimeDir, "../../../package.json");
```
loganfsmyth: Hmm, I'm not sure this is what I'd recommend for verifying this. Could we consider looking for the `package.json` in a parent directory starting from `dirname`? It seems like treating the root as relative to the plugin usage location might be safer than using the location of the plugin, since that could be in an unknown place.

packages/babel-plugin-transform-runtime/src/helpers.js
```diff
@@ -32,6 +32,57 @@ export function hasMinVersion(minVersion, runtimeVersion) {
...
+  moduleName,
+  runtimeVersion,
+  actualRuntimeVersion,
+) {
+  if (!actualRuntimeVersion) return;
+
+  const minActualVersion = semver.minVersion(actualRuntimeVersion).version;
+  const minVersion = semver.minVersion(runtimeVersion).version;
+
+  if (minActualVersion === minVersion) return;
```
loganfsmyth: Treating semver ranges as anything other than ranges is asking for trouble I think. Avoiding logic like this is why I lean toward `intersects` for things in https://github.com/babel/babel/blob/f0c2364d0177908da4fab336fb2b1ad1b1a082d7/packages/babel-core/src/transformation/file/file.js#L191.

What we really want to know in this context is whether every possible version in `actualRuntimeVersion` is contained within `runtimeVersion`, I think? As in, every possible version of the runtime that could be installed by this package.json is valid given the semver range passed to the Babel plugin.

This current logic would pass if for instance we had
```
runtimeVersion: "^7.2.0",
actualRuntimeVersion: "~7.2.0"
```
since the minimum for both is `7.2.0`.

I'm not sure if the semver module actually exposes and API for doing the comparison we need to do here, but we can certainly check.

packages/babel-plugin-transform-runtime/src/helpers.js
```diff
@@ -32,6 +32,57 @@ export function hasMinVersion(minVersion, runtimeVersion) {
...
+          { ...options, version: minActualVersion },
+        ],
+      ],
+    },
+    null,
+    2,
+  ).replace(/^/gm, "  ".repeat(2));
+
+  if (semver.gt(minActualVersion, minVersion)) {
+    console.warn(
```
loganfsmyth: I'd be very hesitant to add new logging to a plugin in a non-major if it isn't opt-in. Some tools will treat any stderr output as an indication that something in the build has failed.

JLHwung:
> I'd be very hesitant to add new logging to a plugin in a non-major if it isn't opt-in

I think the concern here is valid, @nicolo-ribaudo Can you rebase this one onto `next-8-dev`? We can target this change to Babel 8.

PR #10326, opened 2019-08-12
Allow testing logs with `@babel/helper-transform-fixture-test-runner`

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is used by https://github.com/babel/babel/pull/10325, and it would also allow to clean our tests console output and to remove the distinction between fixtures and debug-fixtures in preset-env.

### DIFF
--- a/packages/babel-helper-fixtures/src/index.js
+++ b/packages/babel-helper-fixtures/src/index.js
@@ -141,11 +141,14 @@ export default function get(entryLoc): Array<Suite> {
       const taskOptsLoc = resolve(taskDir + "/options");
       if (taskOptsLoc) extend(taskOpts, require(taskOptsLoc));
 
+      const taskLogLoc = resolve(taskDir + "/expected-log.txt");
+
       const test = {
         optionsDir: taskOptsLoc ? path.dirname(taskOptsLoc) : null,
         title: humanize(taskName, true),
         disabled: taskName[0] === ".",
         options: taskOpts,
+        expectedLog: taskLogLoc ? readFile(taskLogLoc) : null,
         exec: {
           loc: execLoc,
           code: readFile(execLoc),
@@ -221,6 +224,12 @@ export default function get(entryLoc): Array<Suite> {
             "Test cannot throw and also return sourcemaps: " + sourceMapLoc,
           );
         }
+      }
+
+      if (test.exec.code && test.expectedLog) {
+        throw new Error(
+          "Test cannot have logs and also be executed: " + taskLogLoc,
+        );
       }
     }
   }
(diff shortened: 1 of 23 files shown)


packages/babel-helper-fixtures/src/index.js
```diff
@@ -141,11 +141,14 @@ export default function get(entryLoc): Array<Suite> {
       const taskOptsLoc = resolve(taskDir + "/options");
       if (taskOptsLoc) extend(taskOpts, require(taskOptsLoc));
 
+      const taskLogLoc = resolve(taskDir + "/expected-log.txt");
```
JLHwung: The `debug-fixtures` in preset-env is testing against `stdout.txt` and `stderr.txt`. I think we could keep consistency here? And it would be great if preset-env does not log to test output after this PR gets merged.

nicolo-ribaudo: > The `debug-fixtures` in preset-env is testing against `stdout.txt` and `stderr.txt`. I think we could keep consistency here?

I like that, by using the same file, you are also keeping the messages ordered. I would slightly prefer to replace `stdout.txt` and `stderr.txt` with `expected-log.txt`, but I don't have a string opinion.

> And it would be great if preset-env does not log to test output after this PR gets merged.

:+1:

packages/babel-helper-transform-fixture-test-runner/src/index.js
```diff
@@ -263,6 +270,26 @@ function run(task) {
...
+function validateFile(actualCode, expectedLoc, expectedCode) {
+  try {
+    expect(actualCode).toEqualFile({
+      filename: expectedLoc,
+      code: expectedCode,
+    });
+  } catch (e) {
+    if (!process.env.OVERWRITE) throw e;
+
+    console.log(`Updated test file: ${expectedLoc}`);
```
JLHwung: If we don't restore the `console.log` mock, would this line go to the `actualLogs`?

nicolo-ribaudo: That's right

JLHwung: I don't think this is expected. `Updated test file: foo.js` is used to remind the developer and therefore should always print to stdout, regardless of `validateLogs`.

nicolo-ribaudo: Yeah, I meant: "You are right, I will fix it" :sweat_smile:

PR #10331, opened 2019-08-13
Commit generated code

JLHwung:
| Fixed Issues?            | Commit generated code artifacts
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The generated code artifacts should be committed. I am wondering if we could add an extra step
```
git diff-index --quiet HEAD --
```
after `make bootstrap` in CI so that we can remind people there are artifacts changes need to be committed.


nicolo-ribaudo:
We could do it with GH actions, to test them.
I have added you to https://github.com/babel/test-github-actions in case you want to test them.

PR #10332, opened 2019-08-13
Do not allow member expressions to start async arrows

### LINKED ISSUE #10314: Conditional operator confusion with variable named async
## Bug Report

**Current Behavior**
With `@babel/preset-flow`, when a variable is named `async`, the compiler gets confused within a conditional operator and emits a syntax error.

**Input Code**

```js
// @flow

import * as async from 'async';

true ? async.waterfall() : null;
```

```
/repl: Unexpected token, expected "=>" (5:31)

  3 | import * as async from 'async';
  4 | 
> 5 | true ? async.waterfall() : null;
    |                                ^
  6 | 
```

**Expected behavior/code**
It's expected that  `async` would be treated as a normal variable; it's seemingly only within this particular type of expression that things act oddly. Also, the flow compiler when not going through babel does not seem to have this issue.

Note that working around this is easy enough; I'm just reporting the bug to be a good citizen.

```json
{
  "presets": [
    ["@babel/preset-env", {
      "targets": {
        "node": "10.15"
      }
    }],
    "@babel/preset-flow"
  ],
  "plugins": ["add-module-exports"]
}
```

**Environment**
- Babel version(s): 7.5.5 (@babel/core 7.5.5)
- Node version: v10.15.3
- npm version: 6.10.3
- OS: macOS 10.14.6
- Monorepo: no
- How you are using Babel: `cli` and `register`

nicolo-ribaudo:
| Fixed Issues?            | Fixes #10314
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/src/types.js
+++ b/packages/babel-parser/src/types.js
@@ -1447,5 +1447,6 @@ export type Placeholder<N: PlaceholderTypes> = NodeBase & {
 
 export type ParseSubscriptState = {
   optionalChainMember: boolean,
+  maybeAsyncArrow: Boolean,
   stop: boolean,
 };
--- a/packages/babel-parser/src/types.js
+++ b/packages/babel-parser/src/types.js
@@ -1447,6 +1447,6 @@ export type Placeholder<N: PlaceholderTypes> = NodeBase & {
 
 export type ParseSubscriptState = {
   optionalChainMember: boolean,
-  maybeAsyncArrow: Boolean,
+  maybeAsyncArrow: boolean,
   stop: boolean,
 };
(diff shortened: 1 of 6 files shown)


packages/babel-parser/src/types.js
```diff
@@ -1447,5 +1447,6 @@ export type Placeholder<N: PlaceholderTypes> = NodeBase & {
 
 export type ParseSubscriptState = {
   optionalChainMember: boolean,
+  maybeAsyncArrow: Boolean,
```
JLHwung: I think it should be `boolean` anyway.

nicolo-ribaudo:
```suggestion
  maybeAsyncArrow: boolean,
```

PR #10342, opened 2019-08-15
Ensure the corejs-plugin injected imports is handled by module transforms

### LINKED ISSUE #10333: AMD modules with useBuiltIns not correctly implemented
## Bug Report

**Current Behavior**
I am using Babel with `babel-preset-react` and `babel-preset-env` in an environment where modules are managed by [requirejs](https://requirejs.org/).

Everything is latest version.  The generated `.js` files include code like this (in the top level of the `.js` file):
```
require("core-js/modules/es.symbol");

require("core-js/modules/es.symbol.description");

require("core-js/modules/es.symbol.iterator");

require("core-js/modules/es.array.find");

require("core-js/modules/es.array.iterator");

require("core-js/modules/es.array.map");

require("core-js/modules/es.function.name");

require("core-js/modules/es.object.assign");

require("core-js/modules/es.object.get-own-property-descriptor");

require("core-js/modules/es.object.get-prototype-of");

require("core-js/modules/es.object.to-string");

require("core-js/modules/es.string.iterator");

require("core-js/modules/es.string.match");

require("core-js/modules/web.dom-collections.iterator");
```

This applies to configurations with `core-js@2` and `babel-transform-runtime` as well.

**Expected behavior/code**
I expect AMD modules to be compatible with `requirejs`, which in this case means that modules are not loaded with the code above. The following code will fail with the error [Module name ... has not been loaded yet for context](https://requirejs.org/docs/errors.html#notloaded)

```js
require("core-js/modules/es.string.match");
```

```js
{
  "targets": {"ie": 11, "chrome": "58", "firefox": "50"},
    "modules": "amd",
    "useBuiltIns": "usage",
    "corejs": 3,
    "ignoreBrowserslistConfig": true}],
    "@babel/preset-react"],
  "plugins": ["@babel/plugin-proposal-class-properties", "@babel/plugin-proposal-object-rest-spread"],
  "sourceMaps": "inline"
}
```

**Environment**
- Babel version(s): 7.5.5
- Node/npm version: Node v8.9.4/npm v6.10.3
- OS: Windows 10
- How you are using Babel: through a [sbt plubin](https://github.com/zenato/sbt-babel), but I'm pretty sure that it has no relation to the problem at hand

**Possible Solution**
Include these dependencies in the module definition as normal dependencies (i.e. when calling `define`) like this:
```js
define(["exports", "core-js/modules/es.symbol", "core-js/modules/es.symbol.description"], function ...);
```

or use the [simplified define wrapper](https://github.com/requirejs/requirejs/wiki/Differences-between-the-simplified-CommonJS-wrapper-and-standard-AMD-define#cjs) as defined by the `requirejs` documentation, though I'm not 100% sure this is the right solution.

The main issue here is that with AMD modules `require` is asynchronous so code like below does not work as expected.
```js
var mymodule = require("mymodule");
```
```js
require("mymodule");
```

JLHwung:
| Fixed Issues?            | Fixes #10333 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The current sequences of corejs-plugin and module-transforms plugin is
```
// traverse begins
[Program.exit]: module-transforms
// traverse ends
[post]: corejs-plugin addImport
```
However, if user specifies `module: "amd"`, the injected corejs imports will not be transpiled.

This PR changes the sequences above into
```
// traverse begins
[Program.exit]: corejs-plugin addImport
[Program.exit]: module-transforms
// traverse ends
```
Hence I don't adjust the timing of module-transforms since they are public available and doing so may introduce breaking changes. However, the `corejsPlugin` is private so it should be good.

Therefore, this PR can be split into two parts:
1. adjust timing of `addImport` inside `corejs-plugin`
2. move `corejs-plugin` before `module-transforms` plugin.

/cc @zloirock

### DIFF
--- a/packages/babel-preset-env/src/index.js
+++ b/packages/babel-preset-env/src/index.js
@@ -203,22 +203,22 @@ export default declare((api, opts) => {
 
     if (corejs) {
       if (useBuiltIns === "usage") {
-        if (corejs.major === 2) {
-          plugins.push([addCoreJS2UsagePlugin, pluginOptions]);
-        } else {
-          plugins.push([addCoreJS3UsagePlugin, pluginOptions]);
-        }
         if (regenerator) {
-          plugins.push([addRegeneratorUsagePlugin, pluginOptions]);
+          plugins.unshift([addRegeneratorUsagePlugin, pluginOptions]);
+        }
+        if (corejs.major === 2) {
+          plugins.unshift([addCoreJS2UsagePlugin, pluginOptions]);
+        } else {
+          plugins.unshift([addCoreJS3UsagePlugin, pluginOptions]);
         }
       } else {
+        if (!regenerator) {
+          plugins.unshift([removeRegeneratorEntryPlugin, pluginOptions]);
+        }
         if (corejs.major === 2) {
-          plugins.push([replaceCoreJS2EntryPlugin, pluginOptions]);
+          plugins.unshift([replaceCoreJS2EntryPlugin, pluginOptions]);
         } else {
-          plugins.push([replaceCoreJS3EntryPlugin, pluginOptions]);
-          if (!regenerator) {
-            plugins.push([removeRegeneratorEntryPlugin, pluginOptions]);
-          }
+          plugins.unshift([replaceCoreJS3EntryPlugin, pluginOptions]);
         }
       }
     }
--- a/packages/babel-preset-env/src/polyfills/corejs2/entry-plugin.js
+++ b/packages/babel-preset-env/src/polyfills/corejs2/entry-plugin.js
@@ -33,14 +33,12 @@ export default function(
   );
 
   const isPolyfillImport = {
-    ImportDeclaration(path: NodePath) {
-      if (isPolyfillSource(getImportSource(path))) {
-        this.replaceBySeparateModulesImport(path);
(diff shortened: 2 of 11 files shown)


zloirock:
`core-js` internally uses commonjs, so I'm not sure that it's enough.

JLHwung:
@zloirock Good point. Both `regenerator` and `core-js` is using commonjs, so we may need upstream support to effectively run the app bundle.

On Babel's side we should always generate semantically correct code, so `import foo` is translated to `define(["foo"])` even though the library `foo` may _not_ support AMD formats.

RequireJS offers [r.js](https://requirejs.org/docs/commonjs.html#manualconversion) to transform commonjs modules to AMD modules. So it is possible to transform core-js to AMD bundles.

/cc @roti

roti:
> 
> 
> @zloirock Good point. Both `regenerator` and `core-js` is using commonjs, so we may need upstream support to effectively run the app bundle.
> 
I can confirm that. It is not enough to generate correct code on babel's side, core-js needs to support amd as well.

JLHwung:
Close it since upstream polyfills do not support AMD format.

PR #10343, opened 2019-08-15
Do not remove let bindings even they are wrapped in closure

### LINKED ISSUE #10339: [@babel/plugin-proposal-object-rest-spread] TypeError: Cannot read property 'references' of undefined
## Bug Report

**Current Behavior**

While transpiling some code involving object destructuring, the following error is thrown:
```
TypeError: Cannot read property 'references' of undefined
    at Object.keys.forEach.bindingName (.../node_modules/@babel/plugin-proposal-object-rest-spread/lib/index.js:146:45)
```

**Input Code**

- REPL link: https://babeljs.io/en/repl#?babili=false&browsers=safari%2010&build=&builtIns=false&spec=false&loose=true&code_lz=GYVwdgxgLglg9mABFApgZygCgCZwiAWxTCjQEpEBvAKEUWDgCdFMIEMqZsAaRAOgHAYKADbY0AX0RxgiXPiIlyVWnURswHUJEQBeFhV0A-FWrUa0cESj4i4Ac0xcyAblV0Jb9_XARMr7wsrGztHIVFxAI9qCWogA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.5.5&externalPlugins=

```js
function test(documents) {
  for (const {id, ...fields} of documents) {
    const func = () => {
      console.log(id);
    };

    func();
    console.log(fields);
  }
}
```

**Expected behavior/code**

@babel/plugin-proposal-object-rest-spread shouldn't throw an error.

**Babel Configuration**

Env preset browsers: 'safari 10' (see REPL link above)

**Environment**

- Babel v7.5.5 running in the official REPL (see REPL link above)

JLHwung:
| Fixed Issues?            | Fixes #10339
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

I would like to talk more details of #10339 than I did usually because it is an interesting bug: It is an issue that, on the surface, the `object-rest-spread` should be blamed but after digging you would find that it comes from `block-scoping` which seems completely unrelated.

I have simplified the OP's snippet a bit
```js
for (const {foo, ...bar} of {}) {
  () => foo;
  bar;
}
```

The minimum configuration to reproduce this bug is
```
{
  "plugins": ["transform-block-scoping", ["proposal-object-rest-spread", { "loose": true }]]
}
```
note that the plugin order matters here. Theoretically a loose approximation of this configuration would be a two-pass transformation: First transform with `transform-block-scoping` and then transform the intermediate result with `proposal-object-rest-spread`. The Babel REPL works good on two-pass transformation, which means there must be something wrong after the `block-scoping` transforms. Something that does not touch the syntax but the abstract metadata.

The error in #10339 comes from https://github.com/babel/babel/blob/eb3767d58b5e376fa4f4e95e1ba9dc86c667908d/packages/babel-plugin-proposal-object-rest-spread/src/index.js#L113-L116
in the `removeUnusedExcludeKeys` function that is invoked only in `loose` mode. This routine will lookup the binding informations of the ObjectPattern `{ foo, ...bar }` in the for-of statements. For unknown reasons the binding informations are corrupted so exception is thrown.

The block-scoping transform will try to wrap the loop body into a closure once a let reference is used inside a function expressions in the loop body. Which means
```js
for (const {foo, ...bar} of {}) {
  () => foo;
  bar;
}
```
will be transformed to
```js
var _loop = function (foo, bar) {
  () => foo;
  bar;
};

for (var { foo, ...bar } of qux) {
  _loop(foo, bar);
}
```
since `foo` is inside an arrow function expression. However, the `block-scoping` transform incorrectly removes the binding after the loop body is wrapped -- maybe because we tend to feel it is useless since the whole loop body is wrapped in a new scope.

This operation creates dichotomy between the actual code and the abstract syntax tree: The variable declarator is still in our code while its binding information is removed. This error has been hidden for a while until it is luckily detected by `object-rest-spread` loose mode, which tries to manipulate the code according to the binding information.

In the end, the fix is pretty straightforward: do not remove the binding information and always synchronize code with AST.

### DIFF
--- a/packages/babel-plugin-transform-block-scoping/src/index.js
+++ b/packages/babel-plugin-transform-block-scoping/src/index.js
@@ -453,9 +453,7 @@ class BlockScoping {
       if (binding.kind === "let" || binding.kind === "const") {
         binding.kind = "var";
 
-        if (wrappedInClosure) {
-          scope.removeBinding(ref.name);
-        } else {
+        if (!wrappedInClosure) {
           scope.moveBindingTo(ref.name, parentScope);
         }
       }
--- a/packages/babel-plugin-transform-block-scoping/src/index.js
+++ b/packages/babel-plugin-transform-block-scoping/src/index.js
@@ -441,20 +441,25 @@ class BlockScoping {
   }
 
   updateScopeInfo(wrappedInClosure) {
-    const scope = this.scope;
+    const blockScope = this.blockPath.scope;
 
-    const parentScope = scope.getFunctionParent() || scope.getProgramParent();
+    const parentScope =
+      blockScope.getFunctionParent() || blockScope.getProgramParent();
     const letRefs = this.letReferences;
 
     for (const key of Object.keys(letRefs)) {
       const ref = letRefs[key];
-      const binding = scope.getBinding(ref.name);
+      const binding = blockScope.getBinding(ref.name);
       if (!binding) continue;
       if (binding.kind === "let" || binding.kind === "const") {
         binding.kind = "var";
 
-        if (!wrappedInClosure) {
-          scope.moveBindingTo(ref.name, parentScope);
+        if (wrappedInClosure) {
+          if (blockScope.hasOwnBinding(ref.name)) {
+            blockScope.removeBinding(ref.name);
+          }
+        } else {
+          blockScope.moveBindingTo(ref.name, parentScope);
         }
       }
     }
--- a/packages/babel-plugin-transform-block-scoping/test/fixtures/general/issue-10339/exec.js
+++ b/packages/babel-plugin-transform-block-scoping/test/fixtures/general/issue-10339/exec.js
@@ -8,27 +8,31 @@ const code = multiline([
   "}"
 ]);
 
+let programPath;
+let forOfPath;
+let functionPath;
+
 transform(code, {
   configFile: false,
   plugins: [
     "../../../../lib",
     {
       post({ path }) {
+        programPath = path;
         path.traverse({
-          Program(path) {
-            expect(Object.keys(path.scope.bindings)).toEqual(["foo", "bar"]);
-          },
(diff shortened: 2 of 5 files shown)


JLHwung:
@nicolo-ribaudo Addressed in 0db9416.

Actually the issue you mentioned above also exists in current master, so it should be another bug. 🤦‍♂️... Looking into
https://github.com/babel/babel/blob/eb3767d58b5e376fa4f4e95e1ba9dc86c667908d/packages/babel-plugin-transform-block-scoping/src/index.js#L450-L452

When scope is attached to a ForOfStatement, any variable declaration in the loop body is _not_ visible to ForOfStatement and would be skipped here. Given that `needsClosure` only works in loop expression. The goal of removing those bindings that would be defined in loop body closure has never been met.

I have changed `this.scope` to `this.blockPath.scope` so that it works as expected. While it may look like a subtle but drastic change, the only behavior difference will be only in the ForOfStatement case if we look through the BlockScoping usage [here](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-block-scoping/src/index.js#L22).

| Pattern | Equivalence |
| --- | --- |
| Loop | No: blockPath.scope !== this.scope |
| CatchClause | Yes: CatchClause and the its body shares the same scope |
| BlockStatement/SwitchStatement/Program | Yes: blockPath === path |

nicolo-ribaudo:
> Actually the issue you mentioned above also exists in current master

Oh I didn't notice it :sweat_smile:
Anyway, thanks for fixing it!

nicolo-ribaudo:
I changed the `exec` test because the assertion in `Program` did never run.

PR #10344, opened 2019-08-15
Flow enums parsing

gkz:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | Yes (added behind optional flag)
| Documentation PR Link    | 

Flow is working on adding an optional (and by default off) enums feature, gated by the `experimental.enums` Flowconfig flag. Parsing was implemented in the Flow Parser 1.5 months ago (https://github.com/facebook/flow/commit/e3ee40cc1939e8475eae2e064362e6137f6af4dd). 

Here, we add support for parsing Flow Enums in the Flow babel-parser plugin, also gated by an option, `enums`. This brings Babel's Flow parsing in line with Flow Parser's. 

Here are some examples:

```
enum E {
  A,
  B,
}
```

```
enum E {
  A = 1,
  B = 2,
}
```
For more information, see this document: https://github.com/gkz/enums


packages/babel-parser/src/plugins/flow.js
```diff
@@ -2823,4 +2861,416 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+    }
+
+    flowEnumParseExplicitType({
+      enumName,
+    }: {
+      enumName: string,
+    }): EnumExplicitType {
+      if (this.isContextual("of")) {
+        this.next();
+        if (this.state.type === tt.name) {
```
nicolo-ribaudo:
```suggestion
        if (this.matches(tt.name)) {
```

packages/babel-types/src/definitions/flow.js
```diff
@@ -464,3 +465,80 @@ defineType("Variance", {
...
+  },
+});
+
+defineType("EnumBooleanBody", {
+  visitor: ["members"],
+  fields: {
+    explicit: validate(assertValueType("boolean")),
+    members: validateArrayOfType("EnumBooleanMember"),
+  },
+});
```
nicolo-ribaudo: I would prefer this:

```js
defineType("EnumBody", {
  visitor: ["members"],
  fields: {
    explicit: validate(assertValueType("boolean")),
    type: validate(oneOf("boolean", "string", "symbol", "number"),
    members: validateArrayOfType("EnumBooleanMember"),
  },
});
```

Because I think that most plugins would visit any type of enum.

packages/babel-parser/src/plugins/flow.js
```diff
@@ -2823,4 +2861,416 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+    ): void {
+      this.raise(
+        pos,
+        `String enum members need to consistently either all use initializers, or use no initializers, ` +
+          `in enum \`${enumName}\`.`,
+      );
+    }
+
+    flowEnumMemberInit(): EnumMemberInit {
+      const lookahead = this.lookahead();
```
nicolo-ribaudo: Lookahead is really expensive, and should be avoided if possible.
Can we first parse the literal, and then check if the new token is `comma || braceR`?

gkz:
Will add support for parsing export of enums

packages/babel-parser/src/plugins/flow.js
```diff
@@ -2823,4 +2861,416 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+    }
+
+    flowEnumErrorInvalidMemberName(
+      pos: number,
+      { enumName, memberName }: { enumName: string, memberName: string },
+    ): void {
+      const suggestion = memberName[0].toUpperCase() + memberName.slice(1);
+      this.raise(
+        pos,
+        `Enum member names cannot start with lowercase 'a' through 'z'. Instead of using ` +
```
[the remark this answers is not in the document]
gkz: It's specifically `'a' through 'z'`, because JavaScript can be written in any language (even those which do not have the concept of uppercase/lowercase letters), and our restriction is purely for 'a' through 'z'.

packages/babel-parser/src/plugins/flow.js
gkz: Will address

packages/babel-types/src/definitions/flow.js
gkz: We have designed the AST representation with the goal of making invalid states impossible to represent (as much as possible). For example, a symbol based enum must always be explicit, so an `EnumSymbolBody` does not have an `explicit` field.

packages/babel-parser/src/plugins/flow.js
gkz: Will address

gkz:
I have updated the pull request as per comments

packages/babel-types/src/definitions/flow.js
```diff
@@ -464,3 +465,80 @@ defineType("Variance", {
...
+  fields: {
+    members: validateArrayOfType("EnumDefaultedMember"),
+  },
+});
+
+defineType("EnumBooleanMember", {
+  visitor: ["id"],
+  fields: {
+    id: validateType("Identifier"),
+    init: validate(assertValueType("boolean")),
```
nicolo-ribaudo: This is inconsistent with the other types, it should be `BooleanLiteral`

packages/babel-types/src/definitions/flow.js
nicolo-ribaudo: I would still prefer to have a single `EnumBody` with a `type` attribute and maybe a single `EnumMeber`, but if the rest of the team doesn't think that it is needed maybe an alias would be enough:
```js
defineType("EnumNumberBody", {
  aliases: ["EnumBody"],
  visitor: ["members"],
  fields: {
    explicit: validate(assertValueType("boolean")),
    members: validateArrayOfType("EnumNumberMember"),
  },
});
```

packages/babel-types/src/definitions/flow.js
gkz: Will update.

packages/babel-types/src/definitions/flow.js
gkz: I like the idea of using aliases, it works well. I will add an alias for both `EnumBody` and `EnumMember`.

packages/babel-parser/src/plugins/flow.js
```diff
@@ -1567,6 +1585,18 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
     // ==================================
 
+    checkReservedWord(
+      word: string,
+      startLoc: number,
+      checkKeywords: boolean,
+      isBinding: boolean,
+    ): void {
+      if (this.shouldParseEnums() && word === "enum") {
+        return;
```
nicolo-ribaudo: `enum`s should still be reserved words, like any other keyword:

```js
var x = enum;
```

packages/babel-parser/src/plugins/flow.js
```diff
@@ -1615,6 +1645,14 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         const node = this.startNode();
         this.next();
         return this.flowParseInterface(node);
+      } else if (
+        this.shouldParseEnums() &&
+        this.match(tt.name) &&
+        this.state.value === "enum"
```
nicolo-ribaudo: Nit: you can use `this.isContextual` instead of these two lines

packages/babel-parser/src/plugins/flow.js
```diff
@@ -2830,4 +2884,419 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+    flowEnumMemberInit(): EnumMemberInit {
+      const startPos = this.state.start;
+      const endOfInit = () => this.match(tt.comma) || this.match(tt.braceR);
+      switch (this.state.type) {
+        case tt.num: {
+          const literal = this.parseLiteral(this.state.value, "NumericLiteral");
+          if (endOfInit()) {
+            return { type: "number", pos: literal.start, value: literal };
+          }
+          return { type: "invalid", pos: startPos };
```
nicolo-ribaudo: Can't we just report the error here?

packages/babel-parser/src/plugins/flow.js
```diff
@@ -2830,4 +2884,419 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+      defaultedMembers: Array<N.Node>,
+      { enumName }: { enumName: string },
+    ): N.Node {
+      if (initializedMembers.length === 0) {
+        bodyNode.members = defaultedMembers;
+      } else if (defaultedMembers.length === 0) {
+        bodyNode.members = initializedMembers;
+      } else if (defaultedMembers.length > initializedMembers.length) {
+        bodyNode.members = defaultedMembers;
+        for (const member of initializedMembers) {
```
nicolo-ribaudo: Why do we need both the cases? (`defaultMembers` and `initializedMembers`)

packages/babel-parser/src/plugins/flow.js
```diff
@@ -2830,4 +2884,419 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+      return this.finishNode(bodyNode, "EnumStringBody");
+    }
+
+    flowEnumParseExplicitType({
+      enumName,
+    }: {
+      enumName: string,
+    }): EnumExplicitType {
+      if (this.isContextual("of")) {
+        this.next();
```
nicolo-ribaudo: Nit: You can use `if (this.eatContextual("foo"))`

packages/babel-parser/src/plugins/flow.js
```diff
@@ -2830,4 +2884,419 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+            bodyNode.members = [];
+            return this.finishNode(bodyNode, "EnumStringBody");
+          };
+          bodyNode.explicitType = false;
+          const boolsLen = members.booleanMembers.length;
+          const numsLen = members.numberMembers.length;
+          const strsLen = members.stringMembers.length;
+          const defaultedLen = members.defaultedMembers.length;
+
+          if (!(boolsLen | numsLen | strsLen | defaultedLen)) {
```
nicolo-ribaudo: Why `|` and not `||`? Those are not used as boolean bit flags.

packages/babel-parser/src/plugins/flow.js
gkz: It avoid us having to thread the context required for the error message (enum name and explicit type) through two function calls. Additionally, we need to return something from here anyway, and later on we go through all cases of initializer exhaustively. If we threaded through the context, we'd still return some type of `{type: "invalid"}` object, and then later on when we'd have to add a case for it and ignore it. By keeping it this way, I think the code is cleaner, and we deal with reporting errors for initializers all in one place, in the switch exhaustively going through the different cases of initializer (in `flowEnumMembers`)

packages/babel-parser/src/plugins/flow.js
gkz: Basically, we either allow all enum members to be initialized, or not initialized. It has to be consistent. This is unlike TypeScript (for example), so users will probably try to mix up initializing and not-initializing (defaulting). That's why we have extra code and logic to come up with good errors, as it is likely for users to experiment and try it out. We track both initialized and defaulted members, and report errors on whatever is the minority. For example:

```
enum E {
  A,
  B = 2,
  C = 3,
}
```
A naive implementation would take the first member as correct (using a defaulted member), and error on the other two (`B` and `C`), as they are initialized. We are smarter: we see that the majority of members are initialized, and instead error on the minority - member `A`.

gkz:
Not sure exactly the purpose of the "request a review" button on GitHub, but I clicked it. Actually looking at this it doesn't look like PRs need to be "approved" before they are merged, so feel free to ignore the request.

nicolo-ribaudo:
We need to :heavy_check_mark: to merge a PR (and I won't merge it until the 7.7.0 release anyway)

packages/babel-generator/src/generators/flow.js
```diff
@@ -134,6 +134,96 @@ export function DeclareExportAllDeclaration(/*node: Object*/) {
...
+function enumInitializedMember(context: Object, node: Object) {
+  const { id, init } = node;
+  context.print(id, node);
+  context.space();
+  context.token("=");
+  context.space();
+  if (typeof init === "boolean") {
+    context.word(init ? "true" : "false");
+  } else {
+    context.print(init, node);
```
JLHwung: What if `init` is `""`
```js
enum E of string {
  A = ""
}
```

packages/babel-generator/src/generators/flow.js
```diff
@@ -134,6 +134,96 @@ export function DeclareExportAllDeclaration(/*node: Object*/) {
...
+  this.token(",");
+}
+
+function enumInitializedMember(context: Object, node: Object) {
+  const { id, init } = node;
+  context.print(id, node);
+  context.space();
+  context.token("=");
+  context.space();
+  if (typeof init === "boolean") {
```
JLHwung: When will `init` be a boolean? I guess they are always an AST node?

packages/babel-parser/src/plugins/flow.js
```diff
@@ -2839,4 +2877,418 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+    // Flow enum parsing
+
+    flowEnumErrorBooleanMemberNotInitialized(
+      pos: number,
+      { enumName, memberName }: { enumName: string, memberName: string },
+    ): void {
+      this.raise(
+        pos,
+        `Boolean enum members need to be initialized. Use either \`${memberName} = true,\` ` +
+          `or \`${memberName}\` = false,\` in enum \`${enumName}\`.`,
```
JLHwung:
```suggestion
          `or \`${memberName} = false,\` in enum \`${enumName}\`.`,
```

packages/babel-parser/src/plugins/flow.js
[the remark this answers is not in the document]
JLHwung: Troll: It is a valid flow enum.
```js
enum A of string { е = "" }
```

packages/babel-parser/src/plugins/flow.js
```diff
@@ -2839,4 +2877,418 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+          `in enum \`${enumName}\`.`,
+      );
+    }
+
+    flowEnumMemberInit(): EnumMemberInit {
+      const startPos = this.state.start;
+      const endOfInit = () => this.match(tt.comma) || this.match(tt.braceR);
+      switch (this.state.type) {
+        case tt.num: {
+          const literal = this.parseLiteral(this.state.value, "NumericLiteral");
```
JLHwung: Not related to this PR but I am curious why
```js
enum E of number { A = 0 }
```
is valid but
```js
enum E of number { A = -0 }
```
is not valid.

packages/babel-generator/src/generators/flow.js
gkz: Ah yes, thanks for pointing this out. It used to be able to be a `boolean` but I updated this to use a `BooleanLiteral` node after a comment from Nicolò

packages/babel-parser/src/plugins/flow.js
gkz: Not a super strong reason, other than: 
- As a general principle, we're keeping the initial version as restrictive as possible while still being useful, with the idea that you can always allow more behaviour in the future if there is some compelling use-case, but it is difficult to make things more restrictive
- When talking to potential users, no one mentioned wanting negative numbers as initializers
- It simplifies the implementation for v0 of Flow enums

I can imagine a future iteration of Flow enums where we allow arbitrary expressions as initializers, as long as they can be statically evaluated to a number/string/boolean, which would allow for this (and other more complicated, but still statically evaluable, expressions)

Issue #8951, opened 2018-10-31
SyntaxError: Expected atom at position 2     \p{Script_Ex 


satishmane:
## Bug Report

**Current Behavior**

```
> babel . --out-dir dist

SyntaxError: Expected atom at position 2
    \p{Script_Ex
      ^
    at bail (/home/shm/dpp-query-ws/node_modules/regjsparser/parser.js:1093:13)
    at parseTerm (/home/shm/dpp-query-ws/node_modules/regjsparser/parser.js:486:9)
    at parseAlternative (/home/shm/dpp-query-ws/node_modules/regjsparser/parser.js:457:21)
    at parseDisjunction (/home/shm/dpp-query-ws/node_modules/regjsparser/parser.js:437:16)
    at parse (/home/shm/dpp-query-ws/node_modules/regjsparser/parser.js:1108:18)
    at rewritePattern (/home/shm/dpp-query-ws/node_modules/regexpu-core/rewrite-pattern.js:324:15)
    at PluginPass.RegExpLiteral (/home/shm/dpp-query-ws/node_modules/@babel/plugin-transform-unicode-regex/lib/index.js:50:51)
    at newFn (/home/shm/dpp-query-ws/node_modules/@babel/traverse/lib/visitors.js:193:21)
    at NodePath._call (/home/shm/dpp-query-ws/node_modules/@babel/traverse/lib/path/context.js:53:20)
    at NodePath.call (/home/shm/dpp-query-ws/node_modules/@babel/traverse/lib/path/context.js:40:17)
```

**Input Code**
- REPL or Repo link if applicable:

```js
var your => (code) => here;
```

**Expected behavior/code**
Successful transpiled code

**Babel Configuration (.babelrc, package.json, cli command)**
package.json
```json5
"devDependencies": {
    "@babel/cli": "^7.1.2",
    "@babel/core": "^7.1.2",
    "@babel/node": "^7.0.0",
    "@babel/plugin-proposal-object-rest-spread": "^7.0.0",
    "@babel/preset-env": "^7.1.0",
    "@babel/preset-flow": "^7.0.0",
    "nodemon": "^1.18.4"
  }
```
babel.config.js
```js
module.exports = {
  "presets": ["@babel/preset-env","@babel/preset-flow"],
  "plugins": [
    "@babel/plugin-proposal-object-rest-spread"
  ]
}
```

**Environment**
- Babel version(s): [e.g. v6.0.0, v7.0.0-beta.34]
[...]

PR #10347, opened 2019-08-16
Ensure proposal-unicode-property-regex runs before transform-unicode-regex

meowtec:
| Fixed Issues?            |  #8951
| Documentation PR Link    |  N/A

base case:

```
"devDependencies": {
  "@babel/cli": "^7.5.5",
  "@babel/core": "^7.5.5",
  "@babel/preset-env": "^7.5.5"
},
```

babelrc:
```
{
  "presets": [
    ["@babel/preset-env", {
    }]
  ]
}
```
input:

```
/^\p{Unified_Ideograph}$/u;
```

throw with:
```
SyntaxError: Expected atom at position 3
    ^\p{Unified_I
```

input: 
```
/^[\p{Unified_Ideograph}]+$/u;

```

output:
```
/^[IU_ad-in-pr\{\}]+$/;
```


packages/babel-preset-env/test/fixtures/sanity/regex-unicode-property/exec.js
```diff
@@ -0,0 +1 @@
+expect(/^\p{Unified_Ideograph}+$/u.test('中')).toBe(true);
```
JLHwung: Could you check if `/s` flags work with unicode property escape?
```js
expect(/\p{Unified_Ideograph}./su.test('中\n')).toBe(true);
```

meowtec: It got SyntaxError `Expected atom ^\p{Unified_I` because `transform-dotall-regex` runs before `proposal-unicode-property-regex`.

However, if I moved `proposal-unicode-property-regex` before `transform-dotall-regex`, it would get a wrong result. see https://github.com/mathiasbynens/regexpu/issues/68

JLHwung: When `proposal-unicode-property-regex` is applied _before_ `transform-dotall-regex`, the `.` token is transformed to equivalent code points given the default `dotAll: false`, rendered the subsequent `transform-dotall-regex` inactive.

https://github.com/babel/babel/blob/eb3767d58b5e376fa4f4e95e1ba9dc86c667908d/packages/babel-plugin-proposal-unicode-property-regex/src/index.js#L22-L25

Preferably `regexpu-core` could offer a `preserveDot: boolean` option so that `.` token can be consumed by `transform-dotall-regex` _after_ unicode-property transforms.

Note that we may workaround this by plug in `dotAllFlags` in `proposal-unicode-property-regex`,

```js
rewritePattern(node.pattern, node.flags, {
  dotAllFlags: node.flags.includes("s")
}
```
/cc @mathiasbynens

nicolo-ribaudo:
I wouldn't be against creating a single plugin which delegates to regexpu, with options to toggle parsing/transform for the various feature. I will propose it to the team.

mathiasbynens:
FWIW, I'm happy to defer to y'all for making the call on what's best for Babel. I'm open to changes on the regexpu-core side as well, if that would be helpful.

JLHwung:
> I would be in favor of creating a single plugin which delegates to regexpu, with options to toggle parsing/transform for the various feature. I will propose it to the team.

#10367 also justifies the approach to create a single regexpu plugin since `namedGroups` is also supported in regexpu.

PR #10352, opened 2019-08-17
Do not register ambient classes to the TS scope

### LINKED ISSUE #10350: Ambient classes and functions (concrete or ambient) can merge in TypeScript 3.6
The following code snippets will be considered valid in TypeScript 3.6 to more-easily express the "callable class" pattern.

For all intents and purposes, `C` should be considered a hoisted declaration like any other function.

```ts
declare class C { }
declare function C() { }
```

```ts
declare class C { }
function C() { }
```

CC @wesleywigham @RyanCavanaugh

nicolo-ribaudo:
| Fixed Issues?            | Fixes #10350
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | Y?
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is what we already did for ambient functions, since they
follow complex rules regarding declaration merging.
This changes effectively disables duplicate declaration errors
about ambitent classes. While it's not exactly the TypeScript
compiler's behavior, is a close enough approximation which
makes it easy for us to allow ambient classes to merge with
function declarations.

### DIFF
--- a/packages/babel-parser/src/parser/statement.js
+++ b/packages/babel-parser/src/parser/statement.js
@@ -20,6 +20,7 @@ import {
   SCOPE_OTHER,
   SCOPE_SIMPLE_CATCH,
   SCOPE_SUPER,
+  type BindingTypes,
 } from "../util/scopeflags";
 
 const loopLabel = { kind: "loop" },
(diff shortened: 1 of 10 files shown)


nicolo-ribaudo:
> This will unintentionally accept `declare class C {}; declare class C {}` as well but that shouldn't be a big deal.

Yeah, I think that even if we don't throw for all the invalid TS programs it's ok, since people will also use `tsc` to type check anyway.

PR #10357, opened 2019-08-20
Slice Notation (Stage 1)

JLHwung:
This PR has been split into the following PRs:
- [ ] [parser] Add support for slice notation (#10596)
- [ ] Transform slice notation ( #10597 )

They should now be reviewed individually instead of here.


packages/babel-types/src/definitions/experimental.js
```diff
@@ -258,3 +258,25 @@ defineType("BigIntLiteral", {
...
 });
+
+defineType("SliceExpression", {
+  visitor: ["object", "begin", "end", "step"],
+  aliases: ["Expression", "LVal"],
+  fields: {
+    object: {
+      validate: assertNodeType("Expression"),
+    },
+    lower: {
```
JLHwung: May consider rename `lower/upper` to `startIndex/endIndex` given that step can be negative when `lower` is not necessarily lower than `upper`.
```js
const arr = [0, 1, 2]
arr[2:1:-1] // [2]
```

jridgewell: Is this actually valid?

packages/babel-types/src/definitions/experimental.js
```diff
@@ -258,3 +258,25 @@ defineType("BigIntLiteral", {
...
+
+defineType("SliceExpression", {
+  visitor: ["object", "lower", "upper", "step"],
+  aliases: ["Expression", "LVal"],
+  fields: {
+    object: {
+      validate: assertNodeType("Expression"),
+    },
+    lower: {
+      validate: assertNodeType("Identifier", "NumericLiteral"),
```
jridgewell: This can't be a generic expression? It can at least be a `UnaryExpression`.

packages/babel-parser/src/parser/expression.js
```diff
@@ -704,6 +695,73 @@ export default class ExpressionParser extends LValParser {
...
+        node.optional = false;
+        return this.finishNode(node, "OptionalMemberExpression");
+      }
+      return this.finishNode(node, "MemberExpression");
+    }
+    if (expr) {
+      this.checkSliceArgument(pos, expr);
+      node.lower = expr;
+    }
+    expr = this.parseSliceArgument();
```
jridgewell: Nit: don't reuse the `expr` binding.

packages/babel-parser/src/parser/expression.js
```diff
@@ -704,6 +695,73 @@ export default class ExpressionParser extends LValParser {
...
+    startPos: number,
+    startLoc: Position,
+    base: N.Expression,
+    state: N.ParseSubscriptState,
+  ): N.Expression {
+    const node = this.startNodeAt(startPos, startLoc);
+    node.object = base;
+    let expr;
+    const pos = this.state.pos;
+    if (!this.hasPlugin("sliceNotation") || !this.match(tt.colon)) {
```
jridgewell: This is kinda complicated to follow.

packages/babel-types/src/definitions/experimental.js
JLHwung: > If the step argument is negative, then the object is traversed in reverse.

Yes if you mean the negative step.

packages/babel-types/src/definitions/experimental.js
JLHwung: > This can't be a generic expression?

Currently the proposal [restricts](https://github.com/tc39/proposal-slice-notation#can-the-upper-bound-lower-bound-or-the-step-argument-be-an-arbitrary-expression) the argument to be IdentifierReference and DecimalDigits.

> It can at least be a UnaryExpression

Yes given that it can be a negative number or negative infinity. I should have added tests on that. 🤦‍♂️

packages/babel-parser/src/parser/expression.js
JLHwung: Yeah I admit it looks complicated. The code was basically rewritten from
```js
if (this.hasPlugin("sliceNotation")) {
  if (!this.match(tt.colon)) {
    expr = this.parseExpression();
  } else {
    // SliceExpression stuffs.
  }
  if (!this.match(tt.colon)) {
    // MemberExpression stuffs
  } else {
    // SliceExpression stuffs.
  }
} else {
  expr = this.parseExpression();
  // MemberExpression stuffs
}
```

Maybe I could extract `// MemberExpression stuffs` into a dedicate function after I come up with a proper name.

babel.config.js
```diff
@@ -72,6 +72,7 @@ module.exports = function(api) {
       "@babel/plugin-transform-flow-strip-types",
       ["@babel/proposal-class-properties", { loose: true }],
       "@babel/proposal-export-namespace-from",
+      "@babel/proposal-nullish-coalescing-operator",
```
JLHwung: Nullish Coalescing is Staged 3 now. 🙏

packages/babel-plugin-proposal-slice-notation/src/index.js
```diff
@@ -0,0 +1,25 @@
+import { declare } from "@babel/helper-plugin-utils";
+import syntaxSliceNotation from "@babel/plugin-syntax-slice-notation";
+import * as t from "@babel/types";
+
+export default declare(api => {
+  api.assertVersion(7);
+
+  return {
+    name: "proposal-pipeline-operator",
```
jridgewell: 😛

packages/babel-helpers/src/helpers.js
```diff
@@ -1973,3 +1973,122 @@ helpers.wrapRegExp = helper("7.2.6")`
...
+  export default function slice(t, begin, end, step) {
+    // The default value for the step argument is 1
+    if (step === void 0) {
+      step = 1;
+    }
+    step = toInteger(step);
+    if (step === 0 || step === Infinity || step === -Infinity) {
+      // todo: the proposal does not specify details about these cases
+      // https://github.com/tc39/proposal-slice-notation/issues/28
+      throw new Error("Slice step can not be zero, NaN of Infinity");
```
jridgewell: s/of/or

babel.config.js
jridgewell: Can we make this `loose`?

packages/babel-helpers/src/helpers.js
```diff
@@ -1973,3 +1973,122 @@ helpers.wrapRegExp = helper("7.2.6")`
...
+    }
+    return Math.min(len, 9007199254740991);
+  }
+
+  function toInteger(value) {
+    const number = +value;
+    if (isNaN(number)) {
+      return 0;
+    }
+    if (number === 0 || number === Infinity || number === -Infinity) {
```
jridgewell: Nit: none of these are necessary, the flooring won't affect them.

packages/babel-helpers/src/helpers.js
```diff
@@ -1973,3 +1973,122 @@ helpers.wrapRegExp = helper("7.2.6")`
...
+    }
+    step = toInteger(step);
+    if (step === 0 || step === Infinity || step === -Infinity) {
+      // todo: the proposal does not specify details about these cases
+      // https://github.com/tc39/proposal-slice-notation/issues/28
+      throw new Error("Slice step can not be zero, NaN of Infinity");
+    }
+    var tIsString = isString(t);
+    var tIsArray = Array.isArray(t);
+    if (
```
jridgewell: I think this is too complicated for a proposal that doesn't have defined semantics yet.

packages/babel-helpers/src/helpers.js
```diff
@@ -1973,3 +1973,122 @@ helpers.wrapRegExp = helper("7.2.6")`
...
+    if (end === void 0) {
+      final = step < 0 ? low : high;
+    } else {
+      end = toInteger(end);
+      final = end < 0 ? Math.max(low, len + end) : Math.min(end, high);
+    }
+
+    var o;
+    if (tIsString) {
+      o = "";
```
jridgewell: Nit: We can use the array path for this, and join at the end.

packages/babel-parser/src/parser/expression.js
```diff
@@ -704,6 +699,71 @@ export default class ExpressionParser extends LValParser {
     }
   }
 
+  checkSliceArgument(pos: number, expr: N.Expression) {
+    if (
+      !expr ||
+      expr.type === "NumericLiteral" ||
+      expr.type === "Identifier" ||
```
jridgewell: It seems super weird that Identifier is allowed, but normal expressions are not. 😞

packages/babel-parser/src/parser/expression.js
```diff
@@ -704,6 +699,71 @@ export default class ExpressionParser extends LValParser {
...
+    state: N.ParseSubscriptState,
+  ): N.Expression {
+    if (state.optionalChainMember) {
+      node.optional = false;
+      return this.finishNode(node, "OptionalMemberExpression");
+    }
+    return this.finishNode(node, "MemberExpression");
+  }
+
+  parseMaybeSliceExpression(
```
jridgewell: I like this way much better.

packages/babel-plugin-proposal-slice-notation/src/index.js
JLHwung: 🤦‍♂️

packages/babel-helpers/src/helpers.js
JLHwung: Yeah I copy pasted from ES specification without really thinking.

babel.config.js
JLHwung: Sure we can apply `loose`, until we code in a too fancy way.

packages/babel-helpers/src/helpers.js
JLHwung: > This notation can be used for slice operations on primitives like String and any object that provides indexed access using [[Get]] like Array and TypedArray.

It is used to limit the slicee to string, array and typed array. Surely we can debate about whether we should loose the control here and allow any fancy objects as long as they implement indexed access.

packages/babel-helpers/src/helpers.js
JLHwung: My major concern on implementation is memory footprint and performance. We could shorten the helper if we convert string to array, but that means given a string `str` with length `N`, `str[::M]` will require `N` byte memory while the optimal should be `N/M`.

packages/babel-parser/src/parser/expression.js
JLHwung: Actually I like this design choice. If nested general expression is supported, we can expect pretty nested code and one would easily get confused with ternary operator. Think of

```js
a[
 foo : (
   get_bar_from_qux()
 ) : (
  get_qux_from_quux()
 )
]
```
Surely we can open an issue on [the proposal](https://github.com/tc39/proposal-slice-notation#can-the-upper-bound-lower-bound-or-the-step-argument-be-an-arbitrary-expression) to discuss this choice.

packages/babel-helpers/src/helpers.js
jridgewell: That matches array-likes, too, eg `{ length: 1, 0: 'test' }`

packages/babel-generator/test/fixtures/misc/slice-notation/input.js
```diff
@@ -0,0 +1 @@
+a[b:c]
```
nicolo-ribaudo: Can you add more tests?
```js
a[b:c];
a[b:c:d]
a[b:];
a[b::];
a[:c];
a[::d];
a[b::d];
a[:];
a[::];
```

packages/babel-parser/src/parser/expression.js
```diff
@@ -704,6 +699,69 @@ export default class ExpressionParser extends LValParser {
...
+  checkSliceArgument(pos: number, expr: N.Expression) {
+    if (
+      !expr ||
+      expr.type === "NumericLiteral" ||
+      expr.type === "Identifier" ||
+      (expr.type === "UnaryExpression" &&
+        expr.operator === "-" &&
+        (expr.argument.type === "NumericLiteral" ||
+          expr.argument.type === "Identifier"))
+    ) {
```
nicolo-ribaudo: We also need to check that `expr.extra && expr.extra.parenthesized` is false.

JLHwung: Correct, dogfooding another stage 3 proposal. 😝

packages/babel-generator/test/fixtures/misc/slice-notation/input.js
JLHwung: Thanks. Addressed in 452c6cd.

packages/babel-helpers/src/helpers.js
JLHwung: > That matches array-likes, too, eg { length: 1, 0: 'test' }.

Yes, array-like objects is supported.

```js
const obj = { 0: 'a', 1: 'b', 2: 'c', 3: 'd', length: 4 };
obj[1:3];
// → ['b', 'c']
```

But we should reject the following Proxy exotic object
```js
new Proxy({} , { 
  get(target, prop) {
    if (prop === "length") { return 1 }
    if (prop == 0) { return prop }
    return undefined
  }
});
```

nicolo-ribaudo: > any object that provides indexed access using [[Get]]

I think that this means that proxies should be allowed. For example, when using a properly defined proxy, `[].slice` works:

```js
var obj = new Proxy({} , { 
  get(target, prop) {
    if (prop === "length") { return 1 }
    if (prop == 0) { return 3 }
  },
  has(target, prop) {
    return prop == 0;
  }
});

[].slice.call(obj); // [3]
```

cc @gsathya What should it do here?

packages/babel-parser/src/parser/expression.js
```diff
@@ -704,6 +699,75 @@ export default class ExpressionParser extends LValParser {
...
+    }
+    this.raise(
+      pos,
+      "Slice argument requires to be either numeric literal or identifier",
+    );
+  }
+
+  parseSliceArgument(): ?N.Expression {
+    let expr = null;
+    this.eat(tt.colon);
```
nicolo-ribaudo: Shouldn't this be `this.expect`, or be inside an `if (this.eat(tt.colon))`?
Also please add a test where this expectation is not respected, e.g. `foo[1:2;3]`

JLHwung: It would throw "unexpected token, expected `]`" now since `parseExpression` is not guarded by `tt.colon` check.

JLHwung:
Abandoning this PR due to https://github.com/babel/babel/pull/10596#issuecomment-653683160

PR #10361, opened 2019-08-22
feat: if code frame error is on a single line, highlight the whole path

SimenB:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 

Before:
image

After: 
image

`buildCodeFrameError` doesn't seem to have any tests, at least not any I could find. Thoughts on where I can add one (if you want this)

This _might_ make sense as an option passed to `path.buildCodeFrameError` rather than `buildCodeFrameError` forcing it on consumers?

### DIFF
--- a/packages/babel-core/src/transformation/file/file.js
+++ b/packages/babel-core/src/transformation/file/file.js
@@ -19,6 +19,17 @@ const errorVisitor = {
   },
 };
 
+export type NodeLocation = {
+  loc?: {
+    end?: { line: number, column: number },
+    start: { line: number, column: number },
+  },
+  _loc?: {
+    end?: { line: number, column: number },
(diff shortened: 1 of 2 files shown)


JLHwung:
The highlighted path may contain many subpaths even though they are within the same line, i.e.
```js
foo(a, b(f, { g: h(i) }))
```
If the whole callExpression `foo(a, b(f, { g: h(i) }))` is highlighted, user may even get more confused than we simply highlight the start `foo`.

nicolo-ribaudo:
I prefer this PR to the current behavior. In @JLHwung example, if the problem is the whole call expression I don't see why it shouldn't be highlighted.

PR #10371, opened 2019-08-28
perf: replace lookahead by lookaheadCharCode

JLHwung:
| Fixed Issues?            | 
| Patch: Bug Fix?          | Perf
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Introduced `tokenizer.lookaheadCharCode()` method, which simply skips the whitespaces and comments, until the next non-whitespace charcode. The idea is [borrowed from acorn](https://github.com/acornjs/acorn/blob/7e9817d17639d95cc6dbacfde734a0626b2a7dea/acorn/src/statement.js#L38-L40).

`tokenizer.lookahead()` is expensive in the way that it would cache the whole state and restore after `nextToken`. However, in most of times `tokenizer.lookahead` is used to compare the `type` of next token only, therefore we can replace `tokenizer.lookahead()` by `tokenizer.lookaheadCharCode()` which does _not_ interfere with the tokenizer state.

## Logic difference

The replacement is logically equivalent when the character maps to only one possible token type.

It is okay to replace `tt.parenL` with `charcode === 40`.

It will introduce logic difference when replacing `tt.dot` with `charcode === 46` since it may be part of `tt.ellipsis`. However `tt.ellipsis` should not immediately follow the `function` keyword and would bail in `parseExpression`. The same applies when following `import` and `null`.

## Benchmark report
Environment
CPU: Intel Core-i7 6820HQ
RAM: 16 GB 2133 MHz LPDDR3
Kernel: Darwin 18.7.0
Node.js: v12.9.0

Control group versions
acorn: v7.0.0
babel-parser: v7.5.5
```
┌──────────────────────────┬─────────────────────────────┬──────────────────────────────┬──────────────────────────────┬──────────────────────────────┬──────────────────────────────┐
│ fixture                  │ acorn                       │ babel                        │ babelDev                     │ esprima                      │ meriyah                      │
├──────────────────────────┼─────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┤
│ es5/angular.js           │ 30.26 ops/sec ±4.75% (33ms) │ 21.27 ops/sec ±4.57% (47ms)  │ 23.56 ops/sec ±3.79% (42ms)  │ 24.56 ops/sec ±7.19% (41ms)  │ 41.04 ops/sec ±2.64% (24ms)  │
├──────────────────────────┼─────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┤
│ es5/ember.debug.js       │ 11.48 ops/sec ±3.63% (87ms) │ 8.04 ops/sec ±3.97% (124ms)  │ 9.01 ops/sec ±3.5% (111ms)   │ 8.87 ops/sec ±14.86% (113ms) │ 17.27 ops/sec ±8.29% (58ms)  │
├──────────────────────────┼─────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┤
│ es5/babylon-dist.js      │ 54.39 ops/sec ±3.17% (18ms) │ 33.18 ops/sec ±5.38% (30ms)  │ 34.1 ops/sec ±5.29% (29ms)   │ 44.85 ops/sec ±1.74% (22ms)  │ 93.23 ops/sec ±1.13% (11ms)  │
├──────────────────────────┼─────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┤
│ es5/jquery.js            │ 65.05 ops/sec ±2.82% (15ms) │ 42.83 ops/sec ±5.56% (23ms)  │ 46.42 ops/sec ±4.71% (22ms)  │ 57.87 ops/sec ±1.11% (17ms)  │ 115 ops/sec ±2% (8.683ms)    │
├──────────────────────────┼─────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┤
│ es5/backbone.js          │ 270 ops/sec ±2.2% (3.707ms) │ 192 ops/sec ±2.09% (5.197ms) │ 203 ops/sec ±2.61% (4.915ms) │ 227 ops/sec ±2.78% (4.402ms) │ 465 ops/sec ±2.11% (2.148ms) │
├──────────────────────────┼─────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┤
│ es5/react-with-addons.js │ 32.19 ops/sec ±4.13% (31ms) │ 20.24 ops/sec ±4.75% (49ms)  │ 22.79 ops/sec ±3.07% (44ms)  │ 25.02 ops/sec ±4.86% (40ms)  │ 52.12 ops/sec ±1.53% (19ms)  │
├──────────────────────────┼─────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┤
│ es6/angular-compiler.js  │ 14.17 ops/sec ±5.99% (71ms) │ 10.35 ops/sec ±4.55% (97ms)  │ 11.39 ops/sec ±5.81% (88ms)  │ 14.55 ops/sec ±3.34% (69ms)  │ 26.89 ops/sec ±2.4% (37ms)   │
├──────────────────────────┼─────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┼──────────────────────────────┤
│ es6/material-ui-core.js  │ 21.06 ops/sec ±3.95% (47ms) │ 13.44 ops/sec ±3% (74ms)     │ 14.73 ops/sec ±2.93% (68ms)  │ 20.12 ops/sec ±3.46% (50ms)  │ 34.92 ops/sec ±2.13% (29ms)  │
└──────────────────────────┴─────────────────────────────┴──────────────────────────────┴──────────────────────────────┴──────────────────────────────┴──────────────────────────────┘
```

### DIFF
--- a/packages/babel-parser/src/parser/statement.js
+++ b/packages/babel-parser/src/parser/statement.js
@@ -170,7 +170,7 @@ export default class StatementParser extends ExpressionParser {
       case tt._for:
         return this.parseForStatement(node);
       case tt._function:
-        if (this.lookahead().type === tt.dot) break;
+        if (this.lookaheadCharCode() === charCodes.dot) break;
         if (context) {
           if (this.state.strict) {
             this.raise(
@@ -223,8 +223,11 @@ export default class StatementParser extends ExpressionParser {
         return this.parseEmptyStatement(node);
       case tt._export:
       case tt._import: {
-        const nextToken = this.lookahead();
-        if (nextToken.type === tt.parenL || nextToken.type === tt.dot) {
+        const nextTokenCharCode = this.lookaheadCharCode();
+        if (
+          nextTokenCharCode === charCodes.leftParenthesis ||
+          nextTokenCharCode === charCodes.dot
+        ) {
           break;
         }
 
--- a/packages/babel-parser/src/parser/statement.js
+++ b/packages/babel-parser/src/parser/statement.js
@@ -8,7 +8,7 @@ import {
   isIdentifierStart,
   keywordRelationalOperator,
 } from "../util/identifier";
-import { lineBreak, skipWhiteSpace } from "../util/whitespace";
+import { lineBreak } from "../util/whitespace";
 import * as charCodes from "charcodes";
 import {
   BIND_CLASS,
@@ -105,10 +105,7 @@ export default class StatementParser extends ExpressionParser {
     if (!this.isContextual("let")) {
       return false;
     }
-    skipWhiteSpace.lastIndex = this.state.pos;
-    const skip = skipWhiteSpace.exec(this.input);
-    // $FlowIgnore
-    const next = this.state.pos + skip[0].length;
+    const next = this.nextTokenStart();
     const nextCh = this.input.charCodeAt(next);
     // For ambiguous cases, determine if a LexicalDeclaration (or only a
     // Statement) is allowed here. If context is not empty then only a Statement
@@ -1760,18 +1757,9 @@ export default class StatementParser extends ExpressionParser {
 
   isAsyncFunction(): boolean {
     if (!this.isContextual("async")) return false;
-
-    const { pos } = this.state;
-
-    skipWhiteSpace.lastIndex = pos;
-    const skip = skipWhiteSpace.exec(this.input);
-
-    if (!skip || !skip.length) return false;
-
-    const next = pos + skip[0].length;
-
+    const next = this.nextTokenStart();
     return (
-      !lineBreak.test(this.input.slice(pos, next)) &&
+      !lineBreak.test(this.input.slice(this.state.pos, next)) &&
       this.input.slice(next, next + 8) === "function" &&
       (next + 8 === this.length ||
         !isIdentifierChar(this.input.charCodeAt(next + 8)))
--- a/packages/babel-parser/src/parser/statement.js
+++ b/packages/babel-parser/src/parser/statement.js
@@ -1738,11 +1738,11 @@ export default class StatementParser extends ExpressionParser {
   maybeParseExportDeclaration(node: N.Node): boolean {
     if (this.shouldParseExportDeclaration()) {
       if (this.isContextual("async")) {
-        const next = this.lookahead();
+        const next = this.nextTokenStart();
 
         // export async;
-        if (next.type !== tt._function) {
-          this.unexpected(next.start, `Unexpected token, expected "function"`);
+        if (!this.isUnparsedContextual(next, "function")) {
+          this.unexpected(next, `Unexpected token, expected "function"`);
         }
       }
 
@@ -1760,9 +1760,7 @@ export default class StatementParser extends ExpressionParser {
     const next = this.nextTokenStart();
     return (
       !lineBreak.test(this.input.slice(this.state.pos, next)) &&
-      this.input.slice(next, next + 8) === "function" &&
-      (next + 8 === this.length ||
-        !isIdentifierChar(this.input.charCodeAt(next + 8)))
+      this.isUnparsedContextual(next, "function")
     );
   }
 
@@ -1824,10 +1822,10 @@ export default class StatementParser extends ExpressionParser {
       return false;
     }
 
-    const lookahead = this.lookahead();
+    const next = this.nextTokenStart();
     return (
-      lookahead.type === tt.comma ||
-      (lookahead.type === tt.name && lookahead.value === "from")
+      this.input.charCodeAt(next) === charCodes.comma ||
+      this.isUnparsedContextual(next, "from")
     );
   }
 
--- a/packages/babel-parser/src/tokenizer/index.js
+++ b/packages/babel-parser/src/tokenizer/index.js
@@ -13,11 +13,11 @@ import {
   lineBreakG,
   isNewLine,
   isWhitespace,
+  skipWhiteSpace,
 } from "../util/whitespace";
 import State from "./state";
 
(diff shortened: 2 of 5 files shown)


packages/babel-parser/src/tokenizer/index.js
```diff
@@ -168,6 +169,18 @@ export default class Tokenizer extends LocationParser {
...
   }
 
+  nextTokenStart(): number {
+    const thisTokEnd = this.state.pos;
+    skipWhiteSpace.lastIndex = thisTokEnd;
+    const skip = skipWhiteSpace.exec(this.input);
+    // $FlowIgnore: The skipWhiteSpace ensures to match any string
+    return thisTokEnd + skip[0].length;
+  }
+
```
KFlash: @JLHwung In most cases this will not improve the performance. Using regex for this purpose may have opposite effect. I looked at the lexer code and ... well... I understand why you do this.., but you should only need to use current index like this `this.input.charCodeAt(this.index)`.  Refactor the lexer into a `while loop` and you get everything like this for free without any need for regular expressions. You only need to do `index + 1` if you hit `0x20`. If you need the length, store the length before and after the iteration in a locale variable. Doing it like this will also get rid of some overhead like `[0]` and `.length`.  For this case you simply use a locale variable at the start of the function 
`let length = 0` and inside the loop you do this `length++`. And your return will be `TokEnd + length`.

packages/babel-parser/src/parser/statement.js
```diff
@@ -1757,18 +1757,9 @@ export default class StatementParser extends ExpressionParser {
...
-    const skip = skipWhiteSpace.exec(this.input);
-
-    if (!skip || !skip.length) return false;
-
-    const next = pos + skip[0].length;
-
+    const next = this.nextTokenStart();
     return (
-      !lineBreak.test(this.input.slice(pos, next)) &&
+      !lineBreak.test(this.input.slice(this.state.pos, next)) &&
```
KFlash: @JLHwung You don't gain performance here. Try replace `!lineBreak.test(this.input.slice(this.state.pos, next))` with this

```ts
const { input, pos } = this;
const nextChar = input.slice(pos, next);
return ( (nextChar=== 0x10 || nextChar === 0x13 || (nextChar^ 0x2028) <= 1) &&
input.slice(next, next + 8) === "function" && (next + 8 === this.length ||
        !isIdentifierChar(input.charCodeAt(next + 8)))
```

Not ideal either, but an improvement :)

packages/babel-parser/src/tokenizer/index.js
KFlash: You shouldn't blindly trust jsperf, but a regex can be super fast if done in a single operation / task, but in the case of Babel parser this have multiple purposes and slow down. In the long run a `while loop` is the best solution when it comes to a parser. Even if you super optimize the regular expression.
I just give you a friendly tip. Everyone code differently :) But to illustrate this run a benchmark against Acorn. You see Babel parser is 1x slower (*estimated*). Reason I guess is that even the small things haven't been optimized when adding new stuff, and changing the small things as you do in this case will have larger impact then you think. In a positive way.
You can also run a benchmark against my parser too - Meriyah. I guess Babel parser is 4 -5x slower (*estimated*).

packages/babel-parser/src/parser/statement.js
KFlash: `isIdentifierChar` can be super optimized, but only use a direct lookup for it. Similiar to what I did [here](https://github.com/meriyah/meriyah/blob/master/src/lexer/identifier.ts#L15). But I recently found a faster way to scan for identifier than what V8 does which include direct table lookup without bitmasks etc.
 
Btw. I couldn't get babel parser to run in my benchmark, but where do I find a benchmark with it online? And try run [this benchmark](https://meriyah.github.io/meriyah/performance/) and see how Meriyah does it vs Acorn.

KFlash: The source for the benchmark is [located here](https://github.com/meriyah/meriyah/tree/gh-pages/performance) and the website is in the root folder.

Estimate 14 days hard work and you could replicate the Babel parser if you write it from scratch.  Another 6 - 8 days to get all plug-ins working. And the Babel parser should perform same as Meriyah :)

KFlash: @JLHwung Meriyah's REPL is [located here](https://meriyah.github.io/meriyah/) in case of interest. Inspired by Babel's REPL because I found out that the REPL was loading the page very slow too :)

KFlash: @JLHwung You mentioned table lookup table for identifier scanning. I would say that the V8 solution isn't as fast as it could be either, but what you can do is have a table lookup for the token kinds. Then you know that keywords are only lower letters, and e.g. no keyword starts with letter 'u'. This knowledge you can use to optimize for this, and you shouldn't need to check what's between "k" and "d" in "keyword" if you use a while loop iteration and checks for IdentifierPart chars only. That way you would only need to check what's follow after "keyword". That's faster than what V8 does. And you skip unnecessary branching. 
I just implemented this in my own lexer refactoring - seen [here](https://github.com/meriyah/meriyah/blob/recovery/src/scanner/identifier.ts#L9)

I just mentioned it because you mentioned it first, and it could be a good optimization trick for Babel :)

PR #10372, opened 2019-08-29
Don't allow instance properties transformation on namespace

rhyzx:
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    |
| Any Dependency Changes?  |

Properties transformation on namespace is unnecessary and will break webpack's [Tree-Shaking](https://webpack.js.org/guides/tree-shaking/).

Input:
```js
import * as op from 'rxjs/operators';
op.map(fn);
```

Before output:
```js
import _mapInstanceProperty from "@babel/runtime-corejs3/core-js-stable/instance/map";
import * as op from 'rxjs/operators';

_mapInstanceProperty(op).call(op, fn);
```

After output:
```js
import * as op from 'rxjs/operators';
op.map(fn);
```

### DIFF
--- a/packages/babel-plugin-transform-runtime/src/index.js
+++ b/packages/babel-plugin-transform-runtime/src/index.js
@@ -113,7 +113,14 @@ export default declare((api, options, dirname) => {
     );
   }
 
+  function isNamespaced(path) {
+    const binding = path.scope.getBinding(path.node.name);
+    if (!binding) return false;
+    return binding.path.isImportNamespaceSpecifier();
+  }
+
   function maybeNeedsPolyfill(path, methods, name) {
+    if (isNamespaced(path.get("object"))) return false;
     if (!methods[name].types) return true;
 
     const typeAnnotation = path.get("object").getTypeAnnotation();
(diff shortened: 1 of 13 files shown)


packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -113,7 +113,14 @@ export default declare((api, options, dirname) => {
...
   }
 
+  function isNamespaced(path) {
+    const binding = path.scope.getBinding(path.node.name);
+    if (!binding) return false;
+    return binding.path.isImportNamespaceSpecifier();
+  }
+
   function maybeNeedsPolyfill(path, methods, name) {
+    if (isNamespaced(path.get("object"))) return false;
```
JLHwung: The `maybeNeedsPolyfill` is called in `CallExpression` and `MemberExpression` visitors. If we would like to know whether `path.get("object")` is an import namespace specifier, it suffices to test this only in `MemberExpression`

https://github.com/babel/babel/blob/8a775a32ba3dc74c3bb784550ff9a78bdc98e47d/packages/babel-plugin-transform-runtime/src/index.js#L399-L404

zloirock:
Seems something like that also required for the `usage` plugin of `preset-env`.

packages/babel-plugin-transform-runtime/src/index.js
rhyzx: Just made some experiments, `CallExpression` is called before `MemberExpression`, seems there is no way to check `isNamespaced` in `MemberExpression` and prevent transformation in `CallExpression`.

rhyzx:
@zloirock updated

Issue #10355, opened 2019-08-19
Babel plugin absoluteRuntime should support relative paths


arcanis:
## Feature Request

**Is your feature request related to a problem? Please describe.**
The path stored because of `absoluteRuntime` is absolute. As detailed in the doc:

> Using absolute paths is not desirable if files are compiled for use at a later time

**Describe the solution you'd like**
Instead of either full absolute paths or absolutely not resolved bare identifiers, a third option should be relative paths. This would solve the problems of both approaches:

- The relative path would be guaranteed to work provided that the install location and the transpiled files keep the same hierarchy from a computer to another, which is much easier to do than absolute paths (since you generally don't control where a repository is cloned).

- The relative path would prevent cross-packages boundary issues when using PnP (PnP would prevent packages that don't depend on `runtime-corejs2` to require it, which is problematic in environments such as Next.js which transpile everything).

From an implementation perspective, depending on how safe you would want to play it, there are two possible options:

- `absoluteRuntime` could use a relative path by default
- We could introduce a `relativeRuntime` field with the same semantic, but relative

**Describe alternatives you've considered**
Using relative paths is the approach we took for the PnP file itself, and so far it worked really well. I don't see an alternative that would work as well.

loganfsmyth:
Would you be able to elaborate more on the usecase, maybe with a clearer example of how the existing logic is not workable for you? I'm trying to understand the case where you'd want to compile node_modules, but then not be bundling the code in such a way that the path is irrelevant. Are you packaging up the code in a different way?

arcanis:
My use case is abstract in that it doesn't affect me right now - it's more that I think it would be a problem for other users.

For example, I could see a case for transpiling all the files used by a server application offline for speed before sending it to a deployment server. If the deployment path isn't the same as the one used when transpiling the application, then it would crash.

loganfsmyth:
My concern with `relativeRuntime` is that:
* the plugin runs on many files with many paths, so there is no one relative path.
* `absoluteRuntime` lets us resolve the location of the runtime once when the plugin loads, instead of per-file.

What we _could_ do, which I'd be fine with, is add a `useRelativeRuntime` boolean option or something, which would do what we do now, but then instead of injecting the absolute path, it would first compute `path.relative(jsAbsoluteFilename, runtimeAbsolutePath)` and then use that?

arcanis:
[...]

PR #10378, opened 2019-08-31
Implements relativeRuntime

arcanis:
| Fixed Issues?            | https://github.com/babel/babel/issues/10355
| Documentation PR Link    | 

This PR adds a counterpart to `absoluteRuntime` named `relativeRuntime` which, when used, will hardcode the relative path between the transpiled file and the runtime within its `import` tag. This will help with setups where the runtime comes from a package deep within the dependency tree (think Next.js or CRA) and you want the build artifacts to work the same way across various machines (where absolute paths would break).

### DIFF
--- a/packages/babel-plugin-transform-runtime/src/index.js
+++ b/packages/babel-plugin-transform-runtime/src/index.js
@@ -27,6 +27,11 @@ function resolveAbsoluteRuntime(moduleName: string, dirname: string) {
   }
 }
 
+function resolveRelativeRuntime(moduleName: string, dirname: string) {
+  const absoluteRuntime = resolveAbsoluteRuntime(moduleName, dirname);
+  return path.relative(dirname, absoluteRuntime);
+}
+
 function supportsStaticESM(caller) {
   return !!(caller && caller.supportsStaticESM);
 }
@@ -41,6 +46,7 @@ export default declare((api, options, dirname) => {
     useESModules = false,
     version: runtimeVersion = "7.0.0-beta.0",
     absoluteRuntime = false,
+    relativeRuntime = false,
   } = options;
 
   let proposals = false;
@@ -91,6 +97,22 @@ export default declare((api, options, dirname) => {
   ) {
     throw new Error(
       "The 'absoluteRuntime' option must be undefined, a boolean, or a string.",
+    );
+  }
+  if (
+    typeof relativeRuntime !== "boolean" &&
+    typeof relativeRuntime !== "string"
+  ) {
+    throw new Error(
+      "The 'relativeRuntime' option must be undefined, a boolean, or a string.",
+    );
+  }
+  if (
+    absoluteRuntime !== false &&
+    relativeRuntime !== false
+  ) {
+    throw new Error(
+      "The 'absoluteRuntime' and 'relativeRuntime' settings cannot be enabled at the same time",
     );
   }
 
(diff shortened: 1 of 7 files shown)


hg-pyun:
It seems to resolve conflict.

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -27,6 +27,11 @@ function resolveAbsoluteRuntime(moduleName: string, dirname: string) {
   }
 }
```
hg-pyun: It seems to resolve conflict.

arcanis:
🥳 Ready to review, @nicolo-ribaudo!

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -93,6 +99,19 @@ export default declare((api, options, dirname) => {
...
+    typeof relativeRuntime !== "boolean" &&
+    typeof relativeRuntime !== "string"
+  ) {
+    throw new Error(
+      "The 'relativeRuntime' option must be undefined, a boolean, or a string.",
+    );
+  }
+  if (absoluteRuntime !== false && relativeRuntime !== false) {
+    throw new Error(
+      "The 'absoluteRuntime' and 'relativeRuntime' settings cannot be enabled at the same time",
```
nicolo-ribaudo:
```suggestion
      "The 'absoluteRuntime' and 'relativeRuntime' options cannot be enabled at the same time",
```

(I think? I'm not a native speaker)

arcanis: Yep, plus it's more in line with the other messages 👍

packages/babel-plugin-transform-runtime/test/fixtures/relativeRuntime/relative/output.js
```diff
@@ -0,0 +1,7 @@
+var _classCallCheck = require("../../../../node_modules/@babel/runtime/helpers/classCallCheck");
```
nicolo-ribaudo: I think that this is wrong. Even if it is resolved starting from `./subfolder`, the location of `@babel/runtime` relative to this file has only three `../../../`.

I would expect a test for the string version of this option to have another `node_modules` folder inside `./subfolder`: then this path should look like
```js
var _classCallCheck = require("./subfolder/node_modules/@babel/runtime/helpers/classCallCheck");
```

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -202,6 +221,11 @@ export default declare((api, options, dirname) => {
       moduleName,
       path.resolve(dirname, absoluteRuntime === true ? "." : absoluteRuntime),
     );
+  } else if (relativeRuntime !== false) {
+    modulePath = resolveRelativeRuntime(
+      moduleName,
+      path.resolve(dirname, relativeRuntime === true ? "." : relativeRuntime),
```
loganfsmyth: This doesn't match my expectations. Wouldn't we need to calculate this path relative to the JS file being compiled, rather than using the directory?

I also wanted this to strictly be a boolean option because I think of the flags this way:

- `absoluteRuntime` 
    - `false` to just inject `@babel/runtime/<whatever>` 
    - `true` to resolve that to a specific path first
- `useRelativeRuntime`
    - only allowed when `absoluteRuntime` is `true` or a `string`
    - `false` - current behavior
    - `true` - take the resolved path from before and resolve it relative to the JS file it is being inserted into, instead of inserting an absolute path.

If I were doing this from scratch, these might have been nicer as
```
type Opts = {
  resolveRuntime?: boolean | string,
  resolveRelativePath?: boolean,
};
```
or
```
type Opts = {
  resolve?: boolean | string | {
    dirname?: string,
    relative?: boolean,
  },
};
```
or something

arcanis:
Yep you're both right - I'll update this PR. Would you know what's the best way to obtain the path of the file being processed during transformation? Local variable set during `pre(file)`, maybe?

loganfsmyth:
The same way we assign `this.addDefaultImport` in `pre`, we should probably assign `this.runtimeModuleName` generated from `moduleName` and `this.filename`. 

Note that `this.filename` can be `undefined`, so we should throw a useful error in that case.

PR #10369, opened 2019-08-26
Retain trailing comments in array expressions

banga:
This is a proposed fix for https://github.com/babel/babel/issues/10368
with a simple test.

| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #10368 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

This is based on the fix from https://github.com/babel/babel/pull/8488 but for array expressions. I see some opportunity for code reuse here, but wanted to keep the changes isolated.

PR #10380, opened 2019-08-31
Refactor trailing comment adjustment

banga:
| Fixed Issues?            | Fixes issues similar to 10368
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Following up from https://github.com/babel/babel/pull/10369

- Unify the logic for adjusting trailing comments into a separate function
- Use it for all three cases, which fixes a bug for ObjectExpressions and CallExpressions
- Update tests to check for the fixed bug

### DIFF
--- a/packages/babel-parser/src/parser/comments.js
+++ b/packages/babel-parser/src/parser/comments.js
@@ -132,6 +132,7 @@ export default class CommentsParser extends BaseParser {
     if (firstChild) {
       switch (node.type) {
         case "ObjectExpression":
+        case "ObjectPattern":
           this.adjustCommentsAfterTrailingComma(node, node.properties);
           break;
         case "CallExpression":
--- a/packages/babel-parser/src/parser/comments.js
+++ b/packages/babel-parser/src/parser/comments.js
@@ -38,7 +38,7 @@ export default class CommentsParser extends BaseParser {
     this.state.leadingComments.push(comment);
   }
 
-  adjustCommentsAfterTrailingComma_(node: Node, elements: Node[]) {
+  adjustCommentsAfterTrailingComma(node: Node, elements: Node[]) {
     if (elements.length === 0) {
       return;
     }
@@ -132,13 +132,13 @@ export default class CommentsParser extends BaseParser {
     if (firstChild) {
       switch (node.type) {
         case "ObjectExpression":
-          this.adjustCommentsAfterTrailingComma_(node, node.properties);
+          this.adjustCommentsAfterTrailingComma(node, node.properties);
           break;
         case "CallExpression":
-          this.adjustCommentsAfterTrailingComma_(node, node.arguments);
+          this.adjustCommentsAfterTrailingComma(node, node.arguments);
           break;
         case "ArrayExpression":
-          this.adjustCommentsAfterTrailingComma_(node, node.elements);
+          this.adjustCommentsAfterTrailingComma(node, node.elements);
           break;
       }
     }
--- a/packages/babel-parser/src/parser/comments.js
+++ b/packages/babel-parser/src/parser/comments.js
@@ -45,9 +45,6 @@ export default class CommentsParser extends BaseParser {
     if (this.state.leadingComments.length === 0) {
       return;
     }
-    if (!this.state.commentPreviousNode) {
-      return;
-    }
 
     const lastElement = last(elements);
 
--- a/packages/babel-parser/src/parser/comments.js
+++ b/packages/babel-parser/src/parser/comments.js
@@ -60,17 +60,23 @@ export default class CommentsParser extends BaseParser {
       }
     }
 
-    lastElement.trailingComments = [];
+    const newTrailingComments = [];
     while (this.state.leadingComments.length) {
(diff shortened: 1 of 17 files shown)


packages/babel-parser/src/parser/comments.js
```diff
@@ -38,6 +38,48 @@ export default class CommentsParser extends BaseParser {
     this.state.leadingComments.push(comment);
   }
 
+  adjustCommentsAfterTrailingComma_(node: Node, elements: Node[]) {
```
nicolo-ribaudo: Why the trailing underscore?

packages/babel-parser/src/parser/comments.js
```diff
@@ -84,85 +126,20 @@ export default class CommentsParser extends BaseParser {
...
-              }
-              node.trailingComments.push(leadingComment);
-            }
-          }
-        }
+    // Adjust comments that follow a trailing comma on the last element in a
+    // comma separated list of nodes to be the trailing comments on the last
+    // element
+    if (firstChild) {
+      switch (node.type) {
```
nicolo-ribaudo: Maybe also `SequenceExpression`? (I didn't check)

JLHwung: I don't think `SequenceExpression` will allow trailing comma.

packages/babel-parser/src/parser/comments.js
banga: Oh I wanted to mark it as a private function and out of habit followed the convention from another repo. I don't see a convention in babel so I'll just drop the underscore.

packages/babel-parser/src/parser/comments.js
banga: I skimmed through https://github.com/babel/babylon/blob/master/ast/spec.md and looks like `ImportDeclaration` and `ArrayPattern` also allow trailing commas. Should I add those?

nicolo-ribaudo: Yeas please! And also `ExportNamedDeclaration` and `ObjectPattern`

packages/babel-parser/test/fixtures/es2015/modules/export-declaration-trailing-comma/input.js
```diff
@@ -0,0 +1,5 @@
+export {
+  /* One */ Foo
+  /* Two */,
+  /* Three */
+} /* Four */ from "foo";
```
nicolo-ribaudo: Shouldn't "Four" be an `innerComment` of `ExportNamedDeclaration`?

banga: That would be better, but from what I can tell, there's no way to figure out which comments are before the closing bracket and which ones are after. Before the PR, the last three comments get attached as `leadingComments` of `foo`, so we were at least not dropping them. So I could remove handling of `ImportDeclaration` and `ExportDeclaration` from this PR if you prefer the prior logic.

nicolo-ribaudo: I tried to call `adjustCommentsAfterTrailingComma` right before eating the `}` in `parseNamedImportSpecifiers`, instead of doing so in `processComment`. By doing so, `Four` gets attached as a `leadingComments` of `"foo"`.

I'm ok with either behaviors (`leadingComments` of `"foo"` or `trailingComments` of `Foo`); attaching it as an `innerComment` is too hard with the current comments attachment logic.

JLHwung: Agree. Comment attachment currently could only be invoked at `finishNode`, and it will definitely misclassify some inner comments as long as there are multiple whitespaces between the `end` location of any two nodes.

packages/babel-parser/src/parser/comments.js
```diff
@@ -38,6 +38,48 @@ export default class CommentsParser extends BaseParser {
...
   }
 
+  adjustCommentsAfterTrailingComma(node: Node, elements: Node[]) {
+    if (elements.length === 0) {
+      return;
+    }
+    if (this.state.leadingComments.length === 0) {
+      return;
+    }
+    if (!this.state.commentPreviousNode) {
```
JLHwung: I think `this.state.commentPreviousNode` will always be non-empty as long as `elements` is not empty.

banga: Oh yeah good point.

packages/babel-parser/test/fixtures/es2015/modules/export-declaration-trailing-comma/input.js
banga: Yeah, looks like something like #9084 is needed to overhaul comment parsing properly, but there might be challenges around backwards compatibility.

I'm slightly more inclined to keep them as `trailingComments` on `Foo`, because I expect there to be more examples of trailing comments before the `}` than after, so I'll leave this in.

PR #10388, opened 2019-09-03
Add parser support for static decorators

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

After a long time since the proposal update, the PR is here :stuck_out_tongue: 
I'll close all the PRs for the older decorators versions. Note that they will still be maintianed with bug fixes and performance improvements if possible/needed.

---

**Open question**:
I created a new plugin (`staticDecorators`) because we had two different plugins for the other decorators flavors (`decorators` and `decorators-legacy`). If we can come up with a name for the proposal we had one year ago, I would prefer to use a single plugin live we did for the pipeline operator:
```
["decorators", { version: "static" | "legacy" | "?" }]
```
(and leave `decorators-legacy` for backward compatibility)

---

I'm cc-ing a few people which are/might be interested: @littledan @tomdale @kesne @tjcrowder @pabloalmunia

### DIFF
--- a/packages/babel-parser/src/tokenizer/index.js
+++ b/packages/babel-parser/src/tokenizer/index.js
@@ -13,6 +13,7 @@ import {
   lineBreakG,
   isNewLine,
   isWhitespace,
+  skipWhiteSpace,
 } from "../util/whitespace";
 import State from "./state";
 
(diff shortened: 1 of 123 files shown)


packages/babel-parser/src/tokenizer/index.js
```diff
@@ -168,6 +169,21 @@ export default class Tokenizer extends LocationParser {
...
+  lookaheadChPos(noNewLine?: boolean): number {
+    skipWhiteSpace.lastIndex = this.state.pos;
+    // $FlowIgnore
+    const [skip] = skipWhiteSpace.exec(this.input);
+
+    if (noNewLine && lineBreak.test(skip)) return -1;
+    return this.state.pos + skip.length;
+  }
+
+  lookaheadCh(noNewLine?: boolean): number {
```
nicolo-ribaudo: This does the same thing introduced by https://github.com/babel/babel/pull/10371 (when I initially wrote this code that PR didn't exist :stuck_out_tongue:)

I'll rebase this branch when that PR is merged, to use the method it introduces.

mbrowne:
Just happened upon this PR. For the naming, how about just `v2` for the proposal prior to the introduction of static decorators, and `v3` for the current proposal?
```
["decorators", { version: "legacy" | "v2" | "v3" }]
```

(I'm assuming that renaming `legacy` would be a breaking change and needs to remain as-is, but if not, it could be renamed `v1` for consistency.)

nicolo-ribaudo:
The proposal has been rewritten

PR #10395, opened 2019-09-05
fix: transform name capturing regex once

### LINKED ISSUE #10393: Complex regex causes build to fail
## Bug Report

**Current Behavior**
When I create a regex that has both a named capture group and positive lookbehind, babel will throw a stack overflow.  I initially discovered this bug while writing in a big webpack project - I narrowed down the source of the issue and was able to reproduce it in a new repo. I also tried to compile the offending regex after removing the lazy quantifier (In `[a-zA-Z]+?`, the `?` part), but it still throws the same error

**Input Code**
- REPL or Repo link if applicable: https://github.com/csakai/broken-babel-example  
  
For thoroughness, I am including code snippets in this issue post, but you can get all the code you need to replicate this bug using the repo I shared above.
```js
/* src/index.js */
const resourceIdPathPattern = /(?<=\/):(?<key>[a-zA-Z]+?Id|id)/


/* command line */
npm run build
```

**Expected behavior/code**
I expected this valid regex to transpile with no issues.

```js
/* package.json */
{
  "name": "broken-babel-example",
  "scripts": {
    "transpile": "babel src/index.js --config-file ./.babelrc -o dist/bundle.js",
    "clean": "rimraf dist; mkdirp dist",
    "build": "npm run clean; npm run transpile"
  },
  "devDependencies": {
    "@babel/cli": "^7.5.5",
    "@babel/core": "^7.5.5",
    "@babel/preset-env": "^7.5.5",
    "mkdirp": "^0.5.1",
    "rimraf": "^3.0.0"
  }
}

/* .babelrc */
{
  "presets": [ "@babel/preset-env" ]
}

/* command line */
npm run build
```

**Environment**
- Babel version(s): v7.5.5
- Node/npm version: [e.g. Node 10/npm 5]
- OS: WSL (Bash on Ubuntu on Windows 10)
- Monorepo: no
- How you are using Babel: This error occurs when transpiling babel-cli and babel-loader. I have not tested it with register. If you remove `@babel/preset-env` from presets, since there are no changes made to the regex, the build succeeds.

**Possible Solution**
🤷‍♂  I got around this by rewriting my code to eliminate my reliance on lookahead/lookbehind and named capture groups. I realize my code was overly complicated, so I simplified it, but this regex is valid, so it should transpile.

I was only able to test this in WSL, 
The pattern I wrote is for finding a path parameter in a uri template. IE:
```js
const a = 'api/users/:id'
const b = 'api/users/:userId/posts'
a.match(resourceIdPathPattern)
// matches array: [ ':id', 'id' ]
// matches.groups: { key: 'id' }

b.match(resourceIdPathPattern)
// matches array: [ ':userId', 'userId' ]
// matches.groups: { key: 'userId' }
```

This is my first time submitting an issue to a project of this scale, so please let me know if you need any further info.

JLHwung:
| Fixed Issues?            | Fixes #10393 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

When `runtime && !isRegExpTest(path)` is true, a new RegExp node will be inserted to AST and visited later by the `transform-named-capturing-groups-regex` again. Most of cases when regex consists of only named capturing groups, the transformed regex node will be prevented by `indexOf("(?<")` early return, however when there is a lookbehind `<?=`, babylon will recurse infinitely.

Here we revise `indexOf("(?<")` early return by looking ahead`(?<` to see if it is a lookbehind pattern.


nicolo-ribaudo:
What about using `.search(/\?<(?!=/)/)` instead of `indexOf`, to avoid unnecessarily running  the regexp transformer?

JLHwung:
@nicolo-ribaudo Good idea, I revise your suggested regex to include both positive and negative lookbehind.

PR #10396, opened 2019-09-05
fix: early return when instance is not iterable

### LINKED ISSUE #9210: TypeError: undefined is not iterable
## Bug Report
I'm not familiar with Webpack's specifies, or Babel's internals, as I'm being lazy and using `laravel-mix`. However, the error I'm receiving seems like a bug in the compiler to me. But you, the experts, can be the judge of that.

**Current Behaviour**
A clear and concise description of the behaviour.

Using `for ... of` in a certain context doesn't work, but it works elsewhere as expected.

The error I receive is:

```
TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
```

**Input Code**

This is a method from the containing class. It's a very simple CSS parser (no block support) for a specific use case I have. The whole module can be seen under "Additional Context", below.

```js
toObject() {
    let statements = this.css.split(';'),
        object = { };

    console.log(statements);

    for (let statement of statements) {
        let [property, value] = this.parseStatement(statement);

        object[property] = value;
    }

    return object;
}
```

**Output Code**
You didn't ask for it, but I think it may be useful.

```js
{
    key: "toObject",
    value: function toObject() {
      var statements = this.css.split(';'),
          object = {};
      console.log(statements);
      var _iteratorNormalCompletion = true;
      var _didIteratorError = false;
      var _iteratorError = undefined;

      try {
        for (var _iterator = statements[Symbol.iterator](), _step; !(_iteratorNormalCompletion = (_step = _iterator.next()).done); _iteratorNormalCompletion = true) {
          var statement = _step.value;

          var _this$parseStatement = this.parseStatement(statement),
              _this$parseStatement2 = _slicedToArray(_this$parseStatement, 2),
              property = _this$parseStatement2[0],
              value = _this$parseStatement2[1];

          object[property] = value;
        }
      } catch (err) {
        _didIteratorError = true;
        _iteratorError = err;
      } finally {
        try {
          if (!_iteratorNormalCompletion && _iterator.return != null) {
            _iterator.return();
          }
        } finally {
          if (_didIteratorError) {
            throw _iteratorError;
          }
        }
      }

      return object;
    }
  }
```

**Expected behavior/code**
I expected it to iterate over the plain JS array, like it does elsewhere I use `for ... of`.

Unknown. I've not yet forced myself to learn more about what's beneath `laravel-mix`.

But whatever `laravel-mix` uses by default, basically.

But, as follows, these are the package versions.
```
├─ laravel-mix@4.0.12
│  ├─ @babel/core@^7.2.0
```

**Environment**
- Babel version(s): 7.2.0
- Yarn version: `yarn@1.12.3`
- OS: macOS 10.14.2 (18C54)
- How you are using Babel: via `laravel-mix`

**Additional context**

Here is the whole module:

```js
class CSSParser {
    constructor(css) {
        this.css = css;
    }

    getProperty(property, defaultValue) {
        return this.toObject()[property] || defaultValue;
    }

    toObject() {
        let statements = this.css.split(';'),
            object = { };

        console.log(statements);

        for (let statement of statements) {
            let [property, value] = this.parseStatement(statement);

            object[property] = value;
        }

        return object;
    }

    parseStatement(statement) {
        let trim = statement.trim(),
            split = trim.split(':');

        // If empty statement, move on. Not exactly invalid. Also prevents error when no code at all is specified.
        if ( ! trim) {
            return;
        }

        if (split.length < 2) {
            throw new SyntaxError(`Invalid CSS syntax at statement: ${statement}`);
        }

        let [property, ...value] = split;

        value = value.join(':');

        if ( ! CSSParser.isValidPropertyName(property)) {
            throw new SyntaxError(`Invalid property name at statement: ${statement}`);
        }
        
        return [property, value];
    }

    static isValidPropertyName(property) {
        return /^[a-z\-]+$/.test(property);
    }

    static isValid(css) {
        try {
            this.parse(css);
        } catch (error) {
            return false;
        }
    }

    static parse(css) {
        return (new this(css)).toObject();
    }
}

export default CSSParser;
```

JLHwung:
| Fixed Issues?            | Fixes #9210
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The `slicedToArray` is supposed to throw `nonIterableRest` error when applied to an non-iterable instance. However, the `iterableToArrayLimit` will always throw error in `arr[Symbol.iterator]` when `arr` is not iterable.

Here we early return in `iterableToArrayLimit` so that we can offer better error message in runtime. Note that by doing so we also align `iterableToArrayLimit` with `iterableToArray`.

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -943,7 +943,7 @@ helpers.iterableToArrayLimit = helper("7.0.0-beta.0")`
     // _e = _iteratorError
     // _i = _iterator
     // _s = _step
-
+    if (!(Symbol.iterator in Object(arr))) return;
     var _arr = [];
     var _n = true;
     var _d = false;
@@ -969,6 +969,7 @@ helpers.iterableToArrayLimit = helper("7.0.0-beta.0")`
 
 helpers.iterableToArrayLimitLoose = helper("7.0.0-beta.0")`
   export default function _iterableToArrayLimitLoose(arr, i) {
+    if (!(Symbol.iterator in Object(arr))) return;
     var _arr = [];
     for (var _iterator = arr[Symbol.iterator](), _step; !(_step = _iterator.next()).done;) {
       _arr.push(_step.value);
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -943,7 +943,9 @@ helpers.iterableToArrayLimit = helper("7.0.0-beta.0")`
     // _e = _iteratorError
     // _i = _iterator
     // _s = _step
-    if (!(Symbol.iterator in Object(arr))) return;
+    if (!(
+      Symbol.iterator in Object(arr) || Object.prototype.toString.call(arr) === "[object Arguments]"
+    )) { return }
     var _arr = [];
     var _n = true;
     var _d = false;
@@ -969,7 +971,9 @@ helpers.iterableToArrayLimit = helper("7.0.0-beta.0")`
 
 helpers.iterableToArrayLimitLoose = helper("7.0.0-beta.0")`
   export default function _iterableToArrayLimitLoose(arr, i) {
-    if (!(Symbol.iterator in Object(arr))) return;
+    if (!(
+      Symbol.iterator in Object(arr) || Object.prototype.toString.call(arr) === "[object Arguments]"
+    )) { return }
     var _arr = [];
     for (var _iterator = arr[Symbol.iterator](), _step; !(_step = _iterator.next()).done;) {
       _arr.push(_step.value);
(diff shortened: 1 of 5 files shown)


packages/babel-helpers/src/helpers.js
```diff
@@ -943,7 +943,7 @@ helpers.iterableToArrayLimit = helper("7.0.0-beta.0")`
     // _e = _iteratorError
     // _i = _iterator
     // _s = _step
-
+    if (!Symbol.iterator in Object(arr)) return;
```
JLHwung: The `iterableToArray` has guarded this check with an `arguments` check
```js
export default function _iterableToArray(iter) {
    if (
      Symbol.iterator in Object(iter) ||
      Object.prototype.toString.call(iter) === "[object Arguments]"
    ) return Array.from(iter);
  }
```

But 
```js
(function () { return Symbol.iterator in Object(arguments) })()
```
returns true in Chrome 76. I am not sure if this behavior is inconsistent across different browsers.

JLHwung: Oh, there is a browser support matrix on `arguments[@@iterator]`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments/@@iterator

PR #10402, opened 2019-09-06
fix: pass optionLoc when validating plugin object

### LINKED ISSUE #9877: Invalid error message when validating plugin options
## Bug Report

**Current Behavior**
When a plugin has an invalid option, it throws a misleading error, making harder to discover the issue.

This occurs in [validatePluginObject](https://github.com/babel/babel/blob/master/packages/babel-core/src/config/validation/plugins.js#L90) which calls the validator with the option key (string) and value. But the validator (e.g. [assertFunction](https://github.com/babel/babel/blob/master/packages/babel-core/src/config/validation/option-assertions.js#L179)) expects a NestingPath | GeneralPath instead of a string. When the validator fails it calls [msg](https://github.com/babel/babel/blob/master/packages/babel-core/src/config/validation/option-assertions.js#L27) with the string param instead of a loc giving a "Assertion failure: Unknown type undefined" error instead of the proper error; "inherits must be a function, or undefined"

Probably introduced by https://github.com/babel/babel/pull/8493

**Input Code**
- https://codepen.io/blikblum/pen/QPrwRp
See the error message in console

**Expected behavior/code**
Should throw a meaningful message when plugin option is invalid like "inherits must be a function, or undefined"

Using babel standalone:
https://unpkg.com/@babel/standalone/babel.min.js

**Environment**
- Babel version(s): [7.4]
- Node/npm version: [browser]
- OS: [Windows 10]
- Monorepo: [no]
- How you are using Babel: [standalone]

**Possible Solution**

JLHwung:
| Fixed Issues?            | Fixes #9877 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Construct option loc and pass it to the validator so that it can print helpful messages on validation error.


packages/babel-core/test/config-loading.js
```diff
@@ -322,6 +322,21 @@ describe("@babel/core config loading", () => {
...
+        inherits: "inhertis-should-not-be-string",
+      };
+      const opts = {
+        cwd: path.dirname(FILEPATH),
+        filename: FILEPATH,
+        plugins: [fooPlugin],
+      };
+
+      expect(() => loadConfig(opts)).toThrow(
+        /\.inhertis must be a function, or undefined/,
```
nicolo-ribaudo: What is the current message?

JLHwung: Quotes from #9877 
> "Assertion failure: Unknown type undefined" error instead of the proper error; "inherits must be a function, or undefined"

PR #10409, opened 2019-09-07
Fix typo

FND:
> s/compliancy/compliancy/g - seems both correct and consistent with more recent occurrences


nicolo-ribaudo:
Could you also update it in the first few lines of https://github.com/babel/babel/blob/master/CHANGELOG.md? The rest of the file is automatically generated

FND:
I'm not sure what you mean, @nicolo-ribaudo: My patch should include that already (see
 https://github.com/FND/babel/blob/compliance-typo/CHANGELOG.md)?

FWIW, I'd used `git grep -il compliancy | xargs sed -i -e "s/ompliancy/ompliance/g"` to make sure I didn't miss anything.

PR #10417, opened 2019-09-09
Do not guess relative execution status for exported fns

### LINKED ISSUE #10416: Weird behaviour after recent release
## Bug Report

**Current Behavior**
`Cannot read property 'name' of null` error is coming when babel is parsing.
Since `target.node.id is null`

**Input Code**
- Error code - http://bit.ly/2k7mzXL
- Working code - http://bit.ly/2lZu8QG
- Working code - http://bit.ly/2kzBt9e

This code breaks on repl with only es2015 preset selected
```js

const E_ARR = [];

export default function () {
  const someVar = E_ARR;
  return [...someVar];
}
```

**Expected behavior/code**
This code is working though

```js
export default function () {
  const E_ARR = [];
  const someVar = E_ARR;
  return [...someVar];
}
```
```js
const E_ARR = [];

export default function namedFunction() {
  const someVar = E_ARR;
  return [...someVar];
}

```
**Environment**
- Babel version(s): https://babeljs.io/repl
- Node/npm version: https://babeljs.io/repl
- OS: [e.g. OSX 10.13.4, Windows 10]: OSX
- Monorepo: [e.g. yes/no/Lerna]: N/A
- How you are using Babel: [e.g. `cli`, `register`, `loader`]: https://babeljs.io/repl

**Possible Solution**
We believe it's being caused by the latest release this weekend.

This same code was working before this weekend's release

nicolo-ribaudo:
| Fixed Issues?            | Fixes #10416
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-traverse/src/path/introspection.js
+++ b/packages/babel-traverse/src/path/introspection.js
@@ -370,7 +370,12 @@ const executionOrderCheckedNodes = new WeakSet();
 export function _guessExecutionStatusRelativeToDifferentFunctions(
   target: NodePath,
 ): RelativeExecutionStatus {
-  if (!target.isFunctionDeclaration()) return "unknown";
+  if (
+    !target.isFunctionDeclaration() ||
+    target.parentPath.isExportDeclaration()
+  ) {
+    return "unknown";
+  }
 
   // so we're in a completely different function, if this is a function declaration
   // then we can be a bit smarter and handle cases where the function is either
--- a/packages/babel-plugin-transform-block-scoping/test/fixtures/tdz/exported-fn/input.mjs
+++ b/packages/babel-plugin-transform-block-scoping/test/fixtures/tdz/exported-fn/input.mjs
@@ -1,0 +1,14 @@
+import { foo } from "somewhere";
+
+// foo might call "bar"
+foo();
+
+a;
+
+let a;
+
+a;
+
+export function bar() {
+  return a;
+}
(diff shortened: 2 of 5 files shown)


packages/babel-traverse/src/path/introspection.js
```diff
@@ -370,7 +370,12 @@ const executionOrderCheckedNodes = new WeakSet();
 export function _guessExecutionStatusRelativeToDifferentFunctions(
   target: NodePath,
 ): RelativeExecutionStatus {
-  if (!target.isFunctionDeclaration()) return "unknown";
+  if (
+    !target.isFunctionDeclaration() ||
+    target.parentPath.isExportDeclaration()
```
JLHwung: I am not familiar with this topic, but looking in the code, does it mean that `foo` will pass this check
```js
function foo() {}
export default foo;
```
while `bar` wouldn't?
```js
export function bar() {}
```

nicolo-ribaudo: Yes, but they are handled the same because of these lines:

https://github.com/babel/babel/blob/373afbf3a604d7aa8a0f5ea64603b49e281acc4d/packages/babel-traverse/src/path/introspection.js#L394-L404

In the `foo` example, `export default foo` references it without calling the function, so it returns `"unknown"`.
In the `bar` example, it returns `"unknown"` because it's directly exported.

JLHwung: Ah, I see. So we fix the function declaration case without id and also early return when target is an export declaration. Kill two birds with one stone!

PR #10418, opened 2019-09-09
docs: add homepage link

DanArthurGallagher:
| Fixed Issues?            | No issue raised that I can see.
| Patch: Bug Fix?          | I haven't incremented the version because there is no functional change.
| Major: Breaking Change?  | No.
| Minor: New Feature?      | No.
| Tests Added + Pass?      | There are no tests in the package. I ran the global tests and none of the reported errors were in this package.
| Documentation PR Link    | This is the documentation change.
| Any Dependency Changes?  | No.

Adds a homepage link to populate npmjs.com from the `package.json` so that a link is available there in the sidebar. This link is exactly the one taken from `README.md`.

Update the repository property syntax to the object syntax including `type: "git"`.

### DIFF
--- a/packages/babel-runtime/package.json
+++ b/packages/babel-runtime/package.json
@@ -7,8 +7,9 @@
     "access": "public"
   },
   "repository": {
-    "type": "git",
-    "url": "https://github.com/babel/babel/tree/master/packages/babel-runtime"
+    "type" : "git",
+    "url" : "https://github.com/babel/babel.git",
+    "directory": "packages/babel-runtime"
   },
   "homepage": "https://babeljs.io/docs/en/next/babel-runtime.html",
   "author": "Sebastian McKenzie <sebmck@gmail.com>",
--- a/packages/babel-runtime/package.json
+++ b/packages/babel-runtime/package.json
@@ -11,7 +11,7 @@
     "url" : "https://github.com/babel/babel.git",
     "directory": "packages/babel-runtime"
   },
-  "homepage": "https://babeljs.io/docs/en/next/babel-runtime.html",
+  "homepage": "https://babeljs.io/docs/en/next/babel-runtime",
   "author": "Sebastian McKenzie <sebmck@gmail.com>",
   "dependencies": {
     "regenerator-runtime": "^0.13.2"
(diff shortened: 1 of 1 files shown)


packages/babel-runtime/package.json
```diff
@@ -6,7 +6,11 @@
   "publishConfig": {
     "access": "public"
   },
-  "repository": "https://github.com/babel/babel/tree/master/packages/babel-runtime",
+  "repository": {
```
JLHwung: On the `repository`, NPM offers an extra [`directory` field](https://docs.npmjs.com/files/package.json#repository) for monorepo:

```json
"repository": {
  "type" : "git",
  "url" : "https://github.com/babel/babel.git",
  "directory": "packages/babel-runtime"
}
```
To be consistent across different packages in this repo, I think we had better to replace all the repository fields.

JLHwung:
`npm docs` will respect the `homepage` field, I think it is a good idea to align `homepage` to the README.

packages/babel-runtime/package.json
DanArthurGallagher: Thanks for pointing that out! I will make the update here.

DanArthurGallagher:
Hi there @JLHwung, I'm not sure of what you meant by [your review comment](https://github.com/babel/babel/pull/10418#issuecomment-529508652) 

> &hellip; I think it is a good idea to align `homepage` to the README in all the sub packages.

Do you mean that I should link directly to the README in the `homepage` link, or that the `homepage` of the other submodules should also point directly to their documentation pages?

I can make either change, but I did think that as a consumer it would be useful to have a link directly to the intended location rather than replicate the link from the README.
However the npmjs.com sidebar UI would be populated as long as something is added, and that was my intention.

packages/babel-runtime/package.json
```diff
@@ -6,7 +6,12 @@
...
+    "type" : "git",
+    "url" : "https://github.com/babel/babel.git",
+    "directory": "packages/babel-runtime"
+  },
+  "homepage": "https://babeljs.io/docs/en/next/babel-runtime.html",
```
JLHwung: nit: `.html` is not necessary.

nicolo-ribaudo:
@DanArthurGallagher It would be really useful is the "homepage" link pointing to the correct page on the website is added to all the packages which have a docs page!

PR #10421, opened 2019-09-09
Miscellaneous perf tweak

JLHwung:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Some miscellaneous performance tweak when reading comment attachment code.

### DIFF
--- a/packages/babel-parser/src/parser/comments.js
+++ b/packages/babel-parser/src/parser/comments.js
@@ -39,10 +39,11 @@ export default class CommentsParser extends BaseParser {
   }
 
   adjustCommentsAfterTrailingComma(node: Node, elements: Node[]) {
-    if (elements.length === 0) {
+    if (this.state.leadingComments.length === 0) {
       return;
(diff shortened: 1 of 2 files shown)


packages/babel-parser/src/parser/comments.js
```diff
@@ -39,10 +39,11 @@ export default class CommentsParser extends BaseParser {
   }
 
   adjustCommentsAfterTrailingComma(node: Node, elements: Node[]) {
-    if (elements.length === 0) {
+    if (this.state.leadingComments.length === 0) {
       return;
     }
-    if (this.state.leadingComments.length === 0) {
+
+    if (elements.length === 0) {
```
jridgewell: I'm surprised by this change. Checking an array's length should be much faster than accessing 3 props.

PR #10422, opened 2019-09-09
Correctly delegate .return() in async generator

### LINKED ISSUE #9021: Async Iteration Bug - yield* for await break
## Bug Report

**Current Behavior**
When `for await` is used on a generator that uses `yield*` to delegate to another generator and we attempt to `break` out of the loop, the iteration stops but the code after `yield*` still runs. This is wrong and the native implementation in V8 doesn't have this problem.

This essentially means that with the code below, we get `done`:
```js
0
'done'
```

**Input Code**
[Link to example repo](https://github.com/amirmohsen/babel-7-async-iteration-bug)

```js
async function* inner() {
  const ids = [0, 1, 2, 3];
  for await (const id of ids) {
    yield id;
  }
}

async function* wrapper() {
  yield* inner();
  console.log('done');
}

const run = async() => {
  for await (const id of wrapper()) {
    console.log(id);
    break;
  }
};

run();
```

**Expected behavior/code**
If you run the above code in the Chrome console, you get:
```js
0
```
This is the desired behavior (as implemented natively in V8). Since we are breaking out of the loop, the execution inside the generator must not continue.

Please check the repo link above for more details.

babel.config.js
```js
module.exports = (api) => {
  api.cache(true);
  return {
    presets: [
      [
        '@babel/preset-env',
        {
          targets: {
            node: '8.11.3'
          }
        }
      ]
    ]
  }
}
```

package.json (partial)
```json
{
  "scripts": {
    "build": "babel ./src -d ./dist -s",
    "start:dist": "node ./dist",
    "start": "babel-node ./src"
  },
  "devDependencies": {
    "@babel/cli": "^7.1.5",
    "@babel/core": "^7.1.6",
    "@babel/node": "^7.0.0",
    "@babel/polyfill": "^7.0.0",
    "@babel/preset-env": "^7.1.6"
  }
}
```

**Environment**
- Babel version(s): v7.1.6
- Node/npm version: Node 8.11.3/npm 6.4.1
- OS: OSX 10.13.6
- Monorepo: no

### LINKED ISSUE #9905: Async generator with yield*  in inner loop cannot manually return as expected
## Bug Report

Manually return an async generator with yield* in a inner loop it will remain to yield.

**Input Code**

```js
async function* func1()
{
    yield await "yield 1"
    yield await "yield 2"
    yield await "yield 3"
}

async function* func2()
{
    let i = 5
    while (i--) // Inner loop here
    {
        yield* await func1()
    }
}

(async () =>
{
    const iterator = func2();
    for await (let item of iterator)
    {
        console.log(item)
        iterator.return() // manual return at first iteration
    }
})()
```

**Expected behavior/code**
logs
```
yield 1
````

**Current Behavior**
logs
```
yield 1
yield 2
yield 2
yield 2
yield 2
```

```js
{
  "plugins": [
    [
    ]
  ],
  "presets": [
    [
      "@babel/preset-env",
      {
        "targets": {
          "ie": "11"
        }
      }
    ]
  ]
}
```

**Environment**
- Babel version(s): 7.3.4
- Node/npm version: v10.14.1
- OS: Windows 10
- How you are using Babel: With ParcelJS

nicolo-ribaudo:
| Fixed Issues?            | Fixes #9905, fixes #9021
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

(rant) Debugging this thing was really painful. When working with async code and generators, node was trying to be helpful by stepping into code in the next tick and it missed a lot of stack frames.

Disclaimer: I'm only 80% sure that this PR is correct.

cc @zenparsing "Git blame" attributes this code to only you. If you have time to review this code it would be really appreciated, since propbably you are one of the only persons which fully understand how it works :sweat_smile:


packages/babel-plugin-proposal-async-generator-functions/test/fixtures/yield-star/throw-method-with-catch/exec.js
```diff
@@ -0,0 +1,37 @@
+const log = [];
+
+async function* inner() {
+  try {
+    log.push(1);
+    yield "a";
+    log.push(2);
+    yield "b";
+    log.push(3);
+  } catch {
```
JLHwung: Adding a catch parameter should make Node.js 6/8 happy.

nicolo-ribaudo:
I think that the tests are failing on node 6 because of a generator bug in that version.

nicolo-ribaudo:
:heavy_check_mark:

Issue #10403, opened 2019-09-06
Invalid Typescript Types in lib/index.d.ts since 7.6.0


adamdbradley:
## Bug Report

**Current Behavior**
`node_modules/@babel/types/lib/index.d.ts` has invalid typescript since the 7.6.0 release: https://unpkg.com/browse/@babel/types@7.6.0/lib/index.d.ts line 2098

```
export function isNode(node: ?object): boolean
```

**Environment**
- Babel version(s): 7.6.0

**Possible Solution**
Move the `?` before the `:`

forstermatth:
We can't move the `?` unless you move the argument order, since the affected arguments are before others that are not optional.

milesj:
@forstermatth Not sure what you mean. The `?<type>` is invalid TS and is crashing any TS apps using Babel 7.6. This PR caused the issue: https://github.com/babel/babel/pull/10098

@tanhauhau

forstermatth:
Yes - the possible solution offered in the report body would be incorrect. See my PR #10404

trs:
@milesj It looks like the `?<type>` syntax is a [flow maybe type](https://flow.org/en/docs/types/maybe/).
Changing these to a typescript optional type would fail in cases such as [`export function isPlaceholderType(placeholderType: ?string, targetType: string): boolean`](https://github.com/babel/babel/blob/master/packages/babel-types/scripts/generators/typescript.js#L158) since non-optional types cannot follow an optional type.
If `null` or `undefined` are also explicitly allowed, then https://github.com/babel/babel/pull/10404 would be correct.

starksds:
fyi, I reverted back to 7.5.5 and the error went away

lancegliser:
Just for clarity. If you're going move back to 7.5.5, you should declare the `@babel/types` dependency directly. I had devDependencies for `@babel/cli`, `@babel/core`, `@babel/preset-env,` all pointing at 7.5.5. but they all allowed the upgraded version causing this issue.

dantehemerson:
I have the same error using Nestjs

cperryk:
For reference/searchability, the precise error on compiling is this: `node_modules/@babel/types/lib/index.d.ts:2098:30 - error TS8020: JSDoc types can only be used inside documentation comments.`

nicolo-ribaudo:
Fixes in `@babel/types` 7.6.1. Thanks @forstermatth!

PR #10427, opened 2019-09-10
chore: add lint-ts rule

JLHwung:

Add `lint-ts` to `lint` rule. Basically it runs typescript compiler on generated babel types `index.d.ts` and reports any errors. This should prevent #10403 happens again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,8 +54,13 @@ watch: clean clean-lib
 flow:
 	./node_modules/.bin/flow check --strip-root
 
-lint:
+lint: lint-js lint-ts
+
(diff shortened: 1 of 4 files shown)


nicolo-ribaudo:
Generated flow types are already checked by `make flow`, right?

JLHwung:
> Generated flow types are already checked by make flow, right?

Yes. `make flow` will check the flow types in the project scope. But typescript files are only limited to babel-types and it is auto-generated, so I don't think it suffices to add `make typescript` unless we have typescript in our codebase.

PR #10430, opened 2019-09-11
refactor: replace regexp-tree by regexpu

JLHwung:
| Fixed Issues?            | Standardize on regexpu
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is our first step to standardize regexp transforms on regexpu. It helps reduce the package size and pave the way for an integral approach to treat unicode regex.

Currently the test failure is due to an upstream [issue](https://github.com/mathiasbynens/regexpu-core/issues/27).

---

Edit by @nicolo-ribaudo 

Depends on:
- [x] https://github.com/mathiasbynens/regexpu-core/pull/28
- [x] https://github.com/mathiasbynens/regexpu-core/pull/29
- [x] #10439


packages/babel-plugin-transform-named-capturing-groups-regex/test/fixtures/wrapper/basic/output.js
```diff
@@ -1,3 +1,3 @@
-"foo".match(babelHelpers.wrapRegExp(/(.)\1/, {
+"foo".match(babelHelpers.wrapRegExp(/([\0-\t\x0B\f\x0E-\u2027\u202A-\uFFFF])\1/, {
```
JLHwung: This is an known issue for regexpu, it would produce suboptimal output for dot: https://github.com/mathiasbynens/regexpu-core/issues/24

nicolo-ribaudo: Fixed at https://github.com/mathiasbynens/regexpu-core/pull/29

JLHwung:
The size change of `@babel/preset-env-standalone` comes from the fact that `@babel/plugin-transform-named-capturing-groups` is _not_ included in https://github.com/babel/babel/blob/master/packages/babel-standalone/scripts/pluginConfig.json, therefore, it will instead be bundled into `@babel/preset-env-standalone` instead of `@babel/standalone`.

Looking into the [pluginConfig.json](https://github.com/babel/babel/blob/master/packages/babel-standalone/scripts/pluginConfig.json), it seems that even while Stage 0 proposal (function-bind) is included in `pluginConfig`, a Stage 4 proposal (named-capturing-groups) is _not_ included. I believe it is a bug and would propose to move all `notIncludedPlugins`

https://github.com/babel/babel/blob/256fa5aa0455c6313e39f74d4675e676420336cc/packages/babel-preset-env-standalone/src/available-plugins.js#L3-L8

 into `pluginsConfig.json`. We expect a slight size increase of `babel.js` and decrease of `babel-preset-env.js`. In all the total size will be reduced since we are not including `regexp-tree`.

Side note: we may need specify the `peerDependencies` of `@babel/preset-env-standalone` to be `@babel/standalone`.

nicolo-ribaudo:
IIRC, `notIncludedPlugins` includes all the plugin from preset-env which are not part of the es* or stage-* presets.

Ideally, `@babel/standalone` should be as small as possible and all the plugins should only be in `@babel/preset-env-standalone`, but for legacy reasons we can't just move them.
By doing so, users can update `preset-env-standalone` without having to udpate `@babel/standalone` when a new plugin is added.
Unless there is a webpack hack to load `regexpu-core` from the `@babel/standalone` bundle in `@babel/preset-env-standalone`, I don't think that there is anything that we can do to remove that duplicated dependency.

JLHwung:
Thank you for the explainers!

> Unless there is a webpack hack to load regexpu-core from the @babel/standalone bundle in @babel/preset-env-standalone, I don't think that there is anything that we can do to remove that duplicated dependency.

The webpack dll is supposed to offer a runtime `require`-able libraries for webpack application. But from my previous experience I would rather stay away from it especially when we required cross-version compatibility between babel/standalone and preset-env-standalone.

I guess we have to accept the size change. But maybe we can deprecate plugin usage in `babel-standalone`? So we can move plugins to preset-env on v8 (not the JS engine).

PR #6104, opened 2017-08-14
Add real world tests for TypeScript

andy-ms:
| Q                        | A
| ------------------------ | ---
| Fixed Issues             |
| Patch: Bug Fix?          | No 
| Major: Breaking Change?  | No 
| Minor: New Feature?      | No
| Tests Added/Pass?        | Tests added
| Spec Compliancy?         | No
| License                  | MIT
| Doc PR                   |
| Any Dependency Changes?  | Only devDependencies

This adds an additional test suite for `babel-plugin-transform-typescript`, that downloads from two real-world code base and runs some tests:
* Every file parses correctly and produces a valid AST.
* Generator outputs code which parses back to the same AST.
* Transform produces output with an equivalent AST to the output produced by `tsc`. (Whitespace may differ.)

These tests will not run as part of the normal test suite -- they take a long time to run. But since they were useful during the development of the TS transform I thought I would make a pull request for them.

PR #10444, opened 2019-09-15
Run parser tests from the official TypeScript parser

nicolo-ribaudo:
| Fixed Issues?            | Closes #6104
| Any Dependency Changes?  | Yes, dev dependency

The two commits can/should be reviewed one by one.

In the first commit, I abstracted the logic to run external Babel tests: I now that abstractions can introduce bugs, but I didn't want to have three different files to do basically the same thing :stuck_out_tongue: 	

In the second commit I added the tests from https://github.com/Microsoft/TypeScript. The whitelist is so big because many invalid JavaScript programs are accepted by the TypeScript parser without any error (because they report them in the typechecker, which is too strict compared to `@babel/parser`).

cc @babel/typescript


KFlash:
@nicolo-ribaudo Regarding the whitelist file. There are plenty of cases where the Babel TS parser doesn't fail / pass as it should which isn't related to the type checker.. 

Just to mention a few ... The current dynamic import isn't equal to what TS does, the subtract type node should also accept `type reference` and not only `literal`.  See [TS source](https://github.com/microsoft/TypeScript/blob/master/src/compiler/parser.ts#L3020). 
There are other cases as well related to the class and function implementation etc. 

I think I discovered around 60 different cases in total.

nicolo-ribaudo:
Yeah I know, but a significant part of them are for things like duplicate declarations, or invalid modifiers (e.g. `static const x = 2` as a class field)

Makefile
```diff
@@ -100,12 +101,26 @@ bootstrap-flow:
...
+bootstrap-typescript:
+	rm -rf ./build/typescript
+	mkdir -p ./build
+	git clone --branch=master --single-branch --shallow-since=2019-09-01 https://github.com/microsoft/TypeScript.git ./build/typescript
+	cd build/typescript && git checkout $(TYPESCRIPT_COMMIT)
+
+test-typescript:
+	node scripts/tests/typescript
+
+test-typescript-ci: bootstrap test-typescript
```
JLHwung: It will fail on `make -j test-typescript-ci` since `test-typescript` depends on `bootstrap`. Could you rebase on master?

scripts/tests/typescript/index.js
```diff
@@ -0,0 +1,59 @@
+const path = require("path");
+const fs = require("fs").promises;
+const ts = require("typescript");
+const TestRunner = require("../utils/parser-test-runner");
+
+async function* loadTests(dir) {
+  const names = await fs.readdir(dir);
+
+  for (const name of names) {
+    //if (name !== "abstractProperty.ts") continue;
```
JLHwung: nit: code-like comments should be removed.

nicolo-ribaudo:
https://travis-ci.com/babel/babel/jobs/262389187

PR #10449, opened 2019-09-17
Create parser plugin "topLevelAwait"

nicolo-ribaudo:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | https://github.com/babel/website/pull/2115
| Any Dependency Changes?  |

It is a Stage 3 proposal: https://github.com/tc39/proposal-top-level-await

This is different from the `allowAwaitOutsideFunction` because the top-level-await proposal only allows it in modules, while `allowAwaitOutsideFunction` also allows it in scripts. Also, I think that we need the plugin rather than the option for consistency with all the other plugins.

Ref: https://github.com/babel/proposals/issues/44, https://github.com/babel/babel/pull/7637


hzoo:
Awesome! Should we add something like a flag using `caller` for webpack support?

https://github.com/babel/babel/blob/fcb77de9013378c1089abffee3c648b02828e572/packages/babel-preset-env/src/index.js#L160-L166

nicolo-ribaudo:
We can do it when we have `@babel/plugin-syntax-top-level-await`

nicolo-ribaudo:
The failing test262 tests are fixed by https://github.com/tc39/test262/pull/2388

packages/babel-parser/src/parser/expression.js
```diff
@@ -2183,6 +2183,7 @@ export default class ExpressionParser extends LValParser {
   isAwaitAllowed(): boolean {
     if (this.scope.inFunction) return this.scope.inAsync;
     if (this.options.allowAwaitOutsideFunction) return true;
+    if (this.hasPlugin("topLevelAwait")) return this.inModule;
```
eventualbuddha: If top-level await is only valid in modules, should seeing one set `sawUnambiguousESM`? In other words, how should this behave with a `sourceType` of `unambiguous`? Currently, parsing `await 0` with `unambiguous` yields an AST with `sourceType` of `script`, even though if you'd specified `script` it would not have parsed:

```
> require('./').parse('await 0', { plugins: ['topLevelAwait'], sourceType: 'unambiguous' }).program.sourceType
'script'
> require('./').parse('await 0', { plugins: ['topLevelAwait'], sourceType: 'script' }).program.sourceType
Thrown:
{ [SyntaxError: Unexpected token, expected ";" (1:6)
] pos: 6, loc: Position { line: 1, column: 6 } }
```

nicolo-ribaudo: This is tricky because, for exmaple, this is both a valid script and a valid module:
```js
await
0
```

I can set it to a module when `await` the next invalid token is on the same line.

packages/babel-parser/test/fixtures/experimental/top-level-await/unambiguous-ambiguous-allowAwaitOutsideFunction/output.json
```diff
@@ -0,0 +1,85 @@
...
+      "start": {
+        "line": 1,
+        "column": 0
+      },
+      "end": {
+        "line": 1,
+        "column": 7
+      }
+    },
+    "sourceType": "script",
```
eventualbuddha: This is a script that contains an await expression? I thought that wasn’t allowed.

nicolo-ribaudo: We allow it using the `allowAwaitOutsideFunction` option, but it's not in the spec.

packages/babel-parser/src/parser/expression.js
```diff
@@ -2234,9 +2235,18 @@ export default class ExpressionParser extends LValParser {
       );
     }
 
+    if (!this.scope.inFunction && !this.options.allowAwaitOutsideFunction) {
+      if (this.hasPrecedingLineBreak()) {
```
JLHwung: I think the following examples are also unambiguous
```js
await + 0
await - 0
await ( 0 )
await [ 0 ]
await / 0 /u
```

nicolo-ribaudo: I will try to fix it without using lookahead if possible, to avoid special casing different char codes (which can cause errors if we forget some of them / add new ones)

packages/babel-parser/src/parser/expression.js
```diff
@@ -2234,9 +2235,34 @@ export default class ExpressionParser extends LValParser {
...
+        // All the following expressions are ambiguous:
+        //   await + 0
+        //   await - 0
+        //   await ( 0 )
+        //   await [ 0 ]
+        //   await / 0 /u
+        this.match(tt.plusMin) ||
+        this.match(tt.parenL) ||
+        this.match(tt.bracketL) ||
+        this.match(tt.backQuote) ||
```
JLHwung: Awesome! I missed ``await ` 0 ` ``.

I come up with a new case: If one turns on `v8intrinsic`, the following could be ambiguous, too.

```js
await %a()
```

nicolo-ribaudo:
I was initially going to add an `afterExpr` property to the tokens and then check `this.state.type.afterExpr && this.state.type.startsExpr`, but since it would only be used here I kept the `this.match` logic instead.

packages/babel-parser/src/index.js
```diff
@@ -25,15 +25,32 @@ export function parse(input: string, options?: Options): File {
...
+      if (parser.sawUnambiguousESM) {
+        return ast;
+      }
+
+      if (parser.ambiguousScriptDifferentAst) {
+        // Top level await introduces code which can be both a valid script and
+        // a valid module, but which produces different ASTs:
+        //    await
+        //    0
+        // can be parser either as an AwaitExpression, or as two ExpressionStatements.
```
nicolo-ribaudo:
```suggestion
        // can be parsed either as an AwaitExpression, or as two ExpressionStatements.
```

packages/babel-parser/src/parser/expression.js
```diff
@@ -2234,9 +2235,33 @@ export default class ExpressionParser extends LValParser {
...
+        //   await + 0, await - 0, await ( 0 ), await [ 0 ], await / 0 /u, await ``
+        this.match(tt.plusMin) ||
+        this.match(tt.parenL) ||
+        this.match(tt.bracketL) ||
+        this.match(tt.backQuote) ||
+        // Sometimes the tokenizer generates tt.slash for regexps, and this is
+        // handler by parseExprAtom
+        this.match(tt.regexp) ||
+        this.match(tt.slash) ||
+        // This code couls be parsed both as a modulo operator or as an intrinsic:
```
existentialism: Nit:

```suggestion
        // This code could be parsed both as a modulo operator or as an intrinsic:
```

PR #10451, opened 2019-09-17
"use strict" directive should be properly identified as an expression statement

### LINKED ISSUE #10412: [parser] Fake directive with "tail" can still trigger strict mode
## Bug Report

Super edge case: A property access, call, or tagged template on a `"use strict"` string that would otherwise be a directive should not be considered as such. Babel seems to do this anyways.

**Current Behavior**

A "directive" is a string that appears at the first series of statements of a script or function that are exclusively strings (ignoring whitespace, comments, and semi colons / ASI).

So when the parser sees a property access, call, or tagged template, it should no longer be tempted to consider the string a directive. Such strings can not trigger strict mode. It seems Babel is not resilient against this.

**Input Code**

https://astexplorer.net/#/gist/15eb31fbb1024faa6f42a4940d8beb4a/aa32936fd03f63e1e8084610fe5ded643823694b

```js
function f(){
  "use strict".foo;
  eval = 1;
}

function f(){
  "use strict"[foo];
  eval = 1;
}

function f(){
  "use strict"`foo`; // Only works syntactically, but same bug
  eval = 1;
}

function f(){
  "use strict"(foo); // Only works syntactically, but same bug
  eval = 1;
}
```

Note: changing the string contents will allow the example to pass, proving it's caused by the string.
Note: this seems to work correctly in script top level. Just not in a function.

**Expected behavior/code**

This should pass.

**Environment**

Babel 7.8.4

**Possible Solution**

Add more checks when parsing directives :)

Very much an edge case. Low priority.

CrCliff:
| Fixed Issues?            | Fixes #10412 
| Documentation PR Link    | 

This PR fixes the issue of a 'fake' directive with a property access being recognized as a real directive by only matching strings that are not followed by a period.

Also added test case for this situation.

We accomplish this with a lookahead in the regex used by `strictDirective()`:
```javascript 
const literal = /^('|")((?:\\?.)*?)\1(?!\.)/;
```

This way the following two function still uses strict mode:
```javascript
function f() {
  "use strict"
  eval = 1; // Error: Assigning to 'eval' in strict mode
}
```
whereas the following does not:
```javascript
function f() {
  "use strict".foo;
  eval = 1
}
```


JLHwung:
Thank you for contributing to babel!

While this approach does prevent the false positive `"use strict"` directives, could you check the following cases?

```js
// property access with whitespace
"use strict"
.foo

"use strict"["foo"]

// binary operator
"use strict" + "bar";

// sequence expression
"use strict", "bar"
```

The original approach of
https://github.com/babel/babel/blob/99f4f6c3b03c7f3f67cf1b9f1a21b80cfd5b0224/packages/babel-parser/src/parser/util.js#L178
 is to try cheap lookahead to find `"use strict"` string literal. However, per [ECMA Standard](https://tc39.es/ecma262/#use-strict-directive), the Use Strict Directive should be an expression statement, therefore `"use strict".foo` will be a false positive since the lookahead have no information on statement level.

Note that we construct the User Strict Directive here
https://github.com/babel/babel/blob/99f4f6c3b03c7f3f67cf1b9f1a21b80cfd5b0224/packages/babel-parser/src/parser/statement.js#L870

if we could pass `nonSimple` from
https://github.com/babel/babel/blob/99f4f6c3b03c7f3f67cf1b9f1a21b80cfd5b0224/packages/babel-parser/src/parser/expression.js#L1911

and throw when `nonSimple` is true, we could prevent this issue since we are now processing on statement level instead of token level.

CrCliff:
Thank you! 

And when we say `nonSimple`, we're referring to [IsSimpleParameterList](https://tc39.es/ecma262/#sec-destructuring-binding-patterns-static-semantics-issimpleparameterlist), correct?

JLHwung:
@CrCliff 

> And when we say nonSimple, we're referring to IsSimpleParameterList, correct?

Correct, `isSimpleParamList` should be a faithful implementation of `IsSimpleParameterList`.
https://github.com/babel/babel/blob/45a484f0a2802a932a951fca2f9cf1ba63b0aed5/packages/babel-parser/src/parser/expression.js#L1956-L1958

CrCliff:
After looking at the code a bit more, isn't that already what we're doing in expression.js?

https://github.com/babel/babel/blob/99f4f6c3b03c7f3f67cf1b9f1a21b80cfd5b0224/packages/babel-parser/src/parser/expression.js#L1917-L1929

nicolo-ribaudo:
Yes, but the problem is that, before running that if statement, we get `useStrict` with the buggy logic.

What @JLHwung proposes is to move that error inside the `parseBlockOrModuleBlockBody` function, where we actually parse the `"use strict"` directive. In order to do so, we could pass the `nonSimple` and `errorPos` parameters down to `parseBlock` (called inside `parseFunctionBody`), forward it to `parseBlockBody` and then down to `parseBlockOrModuleBlockBody`.

However, I think that this would bring up another problem: we don't need to know if there is `"use strict"` only for disallowing non simple parameter lists, but we also need it to disallow duplicate parameter names. That why `useStrict` is used in the arguments of `checkParams`:
https://github.com/babel/babel/blob/9c1ad0a9f7f8c57e02a20e9f151fce8076d80725/packages/babel-parser/src/parser/expression.js#L1937-L1943

We could also pass to `parseBlockOrModuleBlockBody` the parameters which needs to be passed to `checkParams`, and call it there after parsing the directives.

At this point I wonder: instead of moving all this function-related code into a method whose goal is only to parse blocks, could we pass down a single callback which is called after parsing the directives?

We already have a precedent for this: the `parseMaybeAssign` function, used to parse things like `a += b`, accepts an `afterLeftParse` callback which is called right before parsing the assignment operator:
https://github.com/babel/babel/blob/9c1ad0a9f7f8c57e02a20e9f151fce8076d80725/packages/babel-parser/src/parser/expression.js#L191-L193

I think that we could introduce an `afterDirectivesParse` callback, and then in `parseFunctionBody` do something like this:
```js
node.body = this.parseBlock(true, false, () => {
  // Disallow nonSimple if now this.state.strict is true
  // Validate parameters
});
```

CrCliff:
Hopefully I'm beginning to understand this a bit better. Basically if we finish parsing the directives and there is a use strict directive the `IsSimpleParamsList` must be true. But when I implement this using the following, many tests fail:
``` javascript
      node.body = this.parseBlock(true, false, () => {
        if (this.state.strict) {
          // This logic is here to align the error location with the estree plugin
          const errorPos =
            // $FlowIgnore
            (node.kind === "method" || node.kind === "constructor") &&
            // $FlowIgnore
            !!node.key
              ? node.key.end
              : node.start;
          this.raise(
            errorPos,
            "Illegal 'use strict' directive in function with non-simple parameter list",
          );
        }
        // Add the params to varDeclaredNames to ensure that an error is thrown
        // if a let/const declaration in the function clashes with one of the params.
        this.checkParams(
          node,
          !oldStrict &&
            !useStrict &&
            !allowExpression &&
            !isMethod &&
            !nonSimple,
          allowExpression,
        );
      });
```

nicolo-ribaudo:
In `checkParams` you probably need to use `this.state.strict` instead of `useStrict`.
Can you push your changes, even if they aren't working? Otherwise it's really difficult to say what is broken.

packages/babel-parser/src/parser/statement.js
```diff
@@ -818,6 +819,9 @@ export default class StatementParser extends ExpressionParser {
     if (createNewLexicalScope) {
       this.scope.exit();
     }
+    if (afterDirectivesParse) {
+      afterDirectivesParse.call();
```
nicolo-ribaudo: By calling it here, you are calling it after that the whole block has been parsed, not just the directives. You should call it inside `parseBlockOrModuleBlockBody`, when setting `parsedNonDirective` for the first time.

https://github.com/babel/babel/blob/ad29d91be6fb6969e3a15eac4037684ec7935800/packages/babel-parser/src/parser/statement.js#L885-L888

We also need to be careful: if we never reach the linked code (in case there are only directives and nothing else), we also need to call the function.

Also, `.call()` without any argument isn't needed, you can just use `afterDirectivesParse`.

---

Note that this would make the function be called after we parse a non-directive statement, but I think that it's ok since I couldn't think of a case which could be broken by this change.

This would still work, which is the only affected case:
```js
function f(a) {
  "use strict";
  let a;
}
```
because before registering the `a` parameter we have already registered `let a` to the scope, and thus we can catch duplicates.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1934,14 +1934,33 @@ export default class ExpressionParser extends LValParser {
...
-      // Add the params to varDeclaredNames to ensure that an error is thrown
-      // if a let/const declaration in the function clashes with one of the params.
-      this.checkParams(
-        node,
-        !oldStrict && !useStrict && !allowExpression && !isMethod && !nonSimple,
-        allowExpression,
-      );
-      node.body = this.parseBlock(true, false);
+      node.body = this.parseBlock(true, false, () => {
+        if (this.state.strict && nonSimple) {
```
nicolo-ribaudo: Note that the previous implementation only threw if there is an `"use strict"` directive inside the current function, not if the function is in strict mode.

This makes me realize that my suggestion wasn't 100% correct: we need to pass a parameter `hasStrictDirective` to this function, which is `true` if the `parseBlockOrModuleBlockBody` found a `"use strict"` directive.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1934,14 +1934,33 @@ export default class ExpressionParser extends LValParser {
       const oldLabels = this.state.labels;
       this.state.labels = [];
       if (useStrict) this.state.strict = true;
```
nicolo-ribaudo: This (and the whole old `useStrict` logic) can be removed. You only need to keep the labels logic.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1934,14 +1934,33 @@ export default class ExpressionParser extends LValParser {
...
+            errorPos,
+            "Illegal 'use strict' directive in function with non-simple parameter list",
+          );
+        }
+        // Add the params to varDeclaredNames to ensure that an error is thrown
+        // if a let/const declaration in the function clashes with one of the params.
+        this.checkParams(
+          node,
+          !oldStrict &&
+            !this.state.strict &&
```
nicolo-ribaudo: Nit: since `this.state.strict` "includes" `oldString`, you can remove `!oldStrict &&`

packages/babel-parser/test/fixtures/estree/directives/function-strict-property-access/input.js
```diff
@@ -0,0 +1,5 @@
+
+function f() {
+  "use strict".foo;
```
nicolo-ribaudo: Can you move this rest inside `fixtures/core` instead of `fixtures/estree`?

CrCliff: That makes sense. What is `fixtures/estree` for?

packages/babel-parser/src/parser/expression.js
CrCliff: This means we can remove all of the logic around `oldStrict`, correct?

packages/babel-parser/test/fixtures/estree/directives/function-strict-property-access/input.js
nicolo-ribaudo: It's for the parser plugin `estree`

nicolo-ribaudo: It's for the parser plugin `estree`

packages/babel-parser/src/parser/expression.js
nicolo-ribaudo: Yes :+1:

packages/babel-parser/src/parser/statement.js
CrCliff: This means we need to pass `afterDirectivesParse` all the way down through `parseBlock`, `parseBlockBody`, and finally `parseBlockOrModuleBlockBody`, correct?

I might be having trouble with calling the callback when only directives are parsed. I'll go ahead and push the changes that I have.

nicolo-ribaudo: If after the loop it `parsedNonDirective` is still true, you'll need to call the callback.

CrCliff: Ah that may be where I was having trouble. I was doing the following:
https://github.com/babel/babel/blob/355ee3ee2b62ac020c947e35117f6faffd1915e0/packages/babel-parser/src/parser/statement.js#L900-L902
which means I was calling the callback if `parsedNonDirective` was false.

nicolo-ribaudo: Oh I was wrong, your code is correct :man_facepalming:

packages/babel-parser/src/parser/statement.js
```diff
@@ -880,8 +892,14 @@ export default class StatementParser extends ExpressionParser {
       }
 
       parsedNonDirective = true;
+      if (afterDirectivesParse) {
+        afterDirectivesParse(hasStrictDirective);
```
nicolo-ribaudo: This function should only be called once, the first time we find a non-directive:

```js
if (!parsedNonDirective && afterDirectivesParse) {
   afterDirectivesParse(hasStrictDirective);
}
parsedNonDirective = true
```

JLHwung:
Closing this PR in favour of #11188 

@CrCliff Thank you all the same!

PR #10456, opened 2019-09-17
[parser] Disallow duplicate and undeclared private names

nicolo-ribaudo:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Note: the tests inside `class-private-names-duplicated` are automatically generated. They are (`instance|static`*`field|method|get|set`)^2

This PR reduces the Test262 whitelist from 194 entries to 18 :tada: We are now 99.97% compliant.


JLHwung:
As long as we make it syntax error finally, do we need check duplicated private name in the transformer?

https://github.com/babel/babel/blob/45a484f0a2802a932a951fca2f9cf1ba63b0aed5/packages/babel-helper-create-class-features-plugin/src/index.js#L80-L85

nicolo-ribaudo:
I'm ok with removing it since we don't check other early errors in the transform plugins, but:
- we already report errors about duplicate variables also _after_ parsing, in case the AST is modified. I think that these two kinds of errors are similar;
- it will accept invalid code in case someone has `@babel/helper-create-class-features-plugin` >v7.6.1 and `@babel/parser` <v7.6.1 (not that it is a big deal)

JLHwung:
> we already report errors about duplicate variables also after parsing, in case the AST is modified

If we want to report errors on duplicated private names _after_ parsing, ideally it should be done in babel-traverse instead of the plugin since we could not make any assumptions of plugin ordering.

> @babel/helper-create-class-features-plugin >v7.6.1 and @babel/parser <v7.6.1

We could bump its peerDependency of `@babel/core` to 7.6.1 or let's target to v8 release.

nicolo-ribaudo:
> If we want to report errors on duplicated private names after parsing, ideally it should be done in babel-traverse instead of the plugin since we could not make any assumptions of plugin ordering

Well, we can be sure that this plugin will be the last one run on private names, since then they are removed.

> We could bump its peerDependency of @babel/core to 7.6.1 or let's target to v8 release.

This is a breaking change, since npm would suddently start to warn about valid dependencies.

---

I'm not going to die on this hill since I am just _slightly_ in favor of the duplicated error :stuck_out_tongue:
@babel/babel @tim-mc Do you think that we can remove it from the transform plugin?

JLHwung:
> Well, we can be sure that this plugin will be the last one run on private names, since then they are removed.

What would happen if a plugin inserts private names _after_ they are removed? Would it still be transpiled?

If it is a breaking change to bump peerDependencies, we could target to v8 anyway. Not a big deal.

tim-mc:
It makes sense to move it to `babel-traverse` given @JLHwung's reasons (plugin ordering, reporting the error after parsing). Also, if errors aren't typically reported in the transforms, it's probably best to stick to that convention rather than unnecessarily spreading that concern out to more areas of the codebase.

I also agree that v8 should be targeted for the change.

scripts/tests/test262/test262_whitelist.txt
```diff
@@ -10,102 +10,22 @@ language/expressions/async-arrow-function/early-errors-arrow-await-in-formals-de
...
-language/expressions/class/elements/syntax/early-errors/invalid-names/method-outter-call-expression-this.js(strict mode)
-language/expressions/class/elements/syntax/early-errors/invalid-names/method-outter-member-expression-bad-reference.js(default)
-language/expressions/class/elements/syntax/early-errors/invalid-names/method-outter-member-expression-bad-reference.js(strict mode)
-language/expressions/class/elements/syntax/early-errors/invalid-names/method-outter-member-expression-this.js(default)
-language/expressions/class/elements/syntax/early-errors/invalid-names/method-outter-member-expression-this.js(strict mode)
-language/expressions/class/elements/syntax/early-errors/super-private-access-invalid.js(default)
-language/expressions/class/elements/syntax/early-errors/super-private-access-invalid.js(strict mode)
 language/expressions/object/method-definition/private-name-early-error-async-gen-inside-class.js(default)
 language/expressions/object/method-definition/private-name-early-error-async-gen-inside-class.js(strict mode)
+language/expressions/object/method-definition/private-name-early-error-async-gen.js(default)
```
nicolo-ribaudo: This test (correctly) threw but for the wrong reason. I will open a [good first issue] for it after that this PR is merged.

packages/babel-parser/src/util/scope.js
```diff
@@ -124,6 +165,58 @@ export default class ScopeHandler<IScope: Scope = Scope> {
...
+      const accessor = redefined && classScope.loneAccessors.get(name);
+      if (accessor) {
+        const differences = elementType ^ accessor;
+
+        // The private name can be duplicated only if it is used by
+        // two accessors with different kind (get and set), and if
+        // they have the same placement (static or not).
+        redefined =
+          !(differences & CLASS_ELEMENT_KIND_ACCESSOR) ||
+          differences & CLASS_ELEMENT_FLAG_STATIC;
```
jridgewell: This makes my head hurt. Why not split into multiple if-statements?

jridgewell: Or, just compare them in two separate steps:

```js
// staticBit will be zero if both are static or both are instance.
const staticBit = (accessor & CLASS_ELEMENT_FLAG_STATIC) ^ (elementType & CLASS_ELEMENT_FLAG_STATIC);
// accessorBit will be zero if they are opposite types.
const accessorBit = (accessor & CLASS_ELEMENT_KIND_ACCESSOR) & (elementType & CLASS_ELEMENT_KIND_ACCESSOR);

redefined = !!(staticBit | accessorBit);
```

nicolo-ribaudo: Re-reading it made my head hurt too :stuck_out_tongue: 

I'm rewriting it to this code:
```js
        const oldStatic = accessor & CLASS_ELEMENT_FLAG_STATIC;
        const newStatic = elementType & CLASS_ELEMENT_FLAG_STATIC;

        const oldKind = accessor & CLASS_ELEMENT_FLAG_STATIC;
        const newKind = elementType & CLASS_ELEMENT_FLAG_STATIC;

        // The private name can be duplicated only if it is used by
        // two accessors with different kind (get and set), and if
        // they have the same placement (static or not).
        redefined = oldKind === newKind || oldStatic !== newStatic;
```

PR #10461, opened 2019-09-17
[parser] Disallow static fields named `constructor`

### LINKED ISSUE #10459: Static fields named "constructor" should be a parser error
## Bug Report

**Expected Behavior**
Static class fields can't be named `constructor`. We already disable it for instance fields.

**Input Code**

```js
class A {
  static constructor;
}
```

[repl](https://babeljs.io/repl#?babili=false&browsers=last%202%20versions%20and%20%3E%201%25%2C%20not%20ie%20%3C%3D%2011%2C%0Anot%20op_mini%20all&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAECC0G8BQ1oQC5nQS2NYA9gHYYBOArsOgaQNxIC-QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=react%2Cstage-2%2Ctypescript%2Cenv&prettier=false&targets=&version=7.6.0&externalPlugins=)

Specification: https://tc39.es/proposal-static-class-features/#sec-static-semantics-early-errors

---

If anyone wants to work on this issue, please leave a comment below! If you need any help, plase ask me (or to any other Babel team member) on [Slack](https://slack.babeljs.io/) :slightly_smiling_face: 

1. Fork the repo
2. Run `git clone https://github.com/<YOUR_USERNAME>/babel.git && cd babel`
3. Run `yarn && make bootstrap`
4. Wait :hourglass_flowing_sand: 
5. Run `make watch` (or `make build` whenever you change a file)
6. Add a test in `packages/babel-parser/test/fixtures/experimental/class-properties` (you can take the `no-ctor` test in the same folder as an example)
7. Update the code!
8. `yarn jest parser` to run the tests
9. If it is working, run `make test` to run all the tests
10. Note: this change will likely also impact the results of running the [Test262](https://github.com/tc39/test262) suite on Babel.
    - Before running those tests you need to run `make bootstrap-test262` to download them
    - Use `make test-test262` to run them
    - If some tests which were previously failing now pass, you can run `make test-test262-update-whitelist`
10. Run `git push` and open a PR!

Tip: if you don't know where to start from, try to Ctrl+F for the error generated by non-static fields named `constructor`! :smile:

guywaldman:
| Fixed Issues?            | Fixes #10459
| Patch: Bug Fix?          |  👍
| Major: Breaking Change?  |  👍
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

---

## Main Changes

- Disallowed static fields named `constructor` in a class (previously only non-static were disallowed)
- Updated the text for the error for one consolidated warning, for both static and non-static fields
- Added a test with a static constructor field
- Updated an existing test in the `flow` test suite of the parser to reflect the parse error (see note below)

---

**Note for reviewers**:

I updated an existing test in the `flow` test suite of the parser (`constructor-override-with-class-prop-plugin`)  to reflect the parse error. Strangely, the test _used_ to pass and started failing _when inlining_ the `isNonstaticConstructor` method, without any changes.
In that test, `constructor` was a field, so in theory it should never have passed.
The test relies on the `jsx` (which is probably unrelated), `flow` and `classProperties` plugins.
So would appreciate some feedback on this, as I'm not 100% sure if this is somehow related to Flow.


packages/babel-parser/test/fixtures/flow/classes/constructor-override-with-class-prop-plugin/options.json
```diff
@@ -1,3 +1,4 @@
 {
+  "throws": "Classes may not have a field named 'constructor' (2:2)",
```
nicolo-ribaudo: It's correct :+1:

https://flow.org/try/#0MYGwhgzhAEBiD29oG8BQ1rHgOwgFwCcBXYPeAgLmgAoBKaAXgD5o8ALASwgG5UBfVEA

packages/babel-parser/src/parser/statement.js
```diff
@@ -1498,13 +1498,16 @@ export default class StatementParser extends ExpressionParser {
   }
 
   pushClassProperty(classBody: N.ClassBody, prop: N.ClassProperty) {
-    // This only affects properties, not methods.
-    if (this.isNonstaticConstructor(prop)) {
+    if (
+      !prop.computed &
```
```suggestion
      !prop.computed &&
```

packages/babel-parser/src/parser/statement.js
```diff
@@ -1498,13 +1498,16 @@ export default class StatementParser extends ExpressionParser {
...
+      (prop.key.name === "constructor" || prop.key.value === "constructor")
```
nicolo-ribaudo: Could you add a comment that the second test is needed for strings?

packages/babel-parser/src/parser/statement.js
guywaldman: Woops, thanks.

packages/babel-parser/src/parser/statement.js
guywaldman: WDYT?

nicolo-ribaudo: Maybe like it's in the original function:
```js
  isNonstaticConstructor(method: N.ClassMethod | N.ClassProperty): boolean {
    return (
      !method.computed &&
      !method.static &&
      (method.key.name === "constructor" || // Identifier
        method.key.value === "constructor") // String literal
    );
  }
```

nicolo-ribaudo:
Thanks!

PR #10467, opened 2019-09-19
[parser] Invalid NonOctal Decimal

### LINKED ISSUE #10465: The BigInt suffix should be disallowed in NonOctal Decimal Integer Literal
## Bug Report

**Current Behavior**
The following code should throw a syntax error if the parser plugin bigInt is enabled.

**Input Code**
- REPL or Repo link if applicable: https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=AwDgnAdkA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=script&lineWrap=true&presets=env&prettier=false&targets=&version=7.6.0&externalPlugins=%40babel%2Fplugin-syntax-bigint%407.4.4

```js
089n
```

**Expected behavior/code**
It should throw `Invalid BigIntLiteral` error.

**Background**

We use tokenizer to process the source code into a stream of tokens and `08n` will be processed in `readNumber`
https://github.com/babel/babel/blob/9c1ad0a9f7f8c57e02a20e9f151fce8076d80725/packages/babel-parser/src/tokenizer/index.js#L981

The NonOctal Decimal Integer Literal is defined as

```
NonOctalDecimalInteger:
0 DecimalDigits

DecimalDigits:
DecimalDigit DecimalDigits_opt

DecimalDigit:
0 1 2 3 4 5 6 7 8 9
```
The above grammar requires a NonOctal Decimal Integer Literal to be a sequence of decimal digits starting with zero and the length is at least 2.

**Possible Solution**
1. Detect NonOctal Decimal Integer Literal in `readNumber`. Throw an error here when it is detected.

https://github.com/babel/babel/blob/9c1ad0a9f7f8c57e02a20e9f151fce8076d80725/packages/babel-parser/src/tokenizer/index.js#L1027-L1028

2. Add a test on https://github.com/babel/babel/tree/master/packages/babel-parser/test/fixtures/experimental/bigint

3. Run `make test-test262-update-whitelist` to update the whitelist, the following lines are expected to be removed.
https://github.com/babel/babel/blob/9c1ad0a9f7f8c57e02a20e9f151fce8076d80725/scripts/tests/test262/test262_whitelist.txt#L113-L116

1. Write a comment there to let other possible contributors know that you are working on this bug.
1. Fork the repo
1. Run `make watch` (or `make build` whenever you change a file)
1. Add a test (only `input.js`; `output.js` will be automatically generated)
1. Update the code!
1. `yarn jest parser` to run the tests

gonzarodriguezt:
| Fixed Issues?            | Fixes #10465 

I added a regex verification for a NonOctal Decimal Integer, defined as a sequence of decimal digits starting with 0 and with at least two characters, i did it within the scope of the `if (octal)` because `octal` ensure two of the requirements, and there was a similar octal verification.

**Todo:**
- [x] Update the whitelist


packages/babel-parser/src/tokenizer/index.js
```diff
@@ -1024,8 +1025,11 @@ export default class Tokenizer extends LocationParser {
 
     if (this.hasPlugin("bigInt")) {
       if (next === charCodes.lowercaseN) {
-        // disallow floats and legacy octal syntax, new style octal ("0o") is handled in this.readRadixNumber
-        if (isFloat || octal) this.raise(start, "Invalid BigIntLiteral");
+        // disallow floats, legacy octal syntax and non octal decimals
+        // new style octal ("0o") is handled in this.readRadixNumber
+        if (isFloat || octal || isNonOctalDecimal) {
```
JLHwung: Does `isNonOctalDecimal == true` _imply_ `(octal || isNonOctalDecimal) == true` ? If so we could consider removing `octal` check here.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -990,15 +991,15 @@ export default class Tokenizer extends LocationParser {
...
         this.raise(
           start,
           "Legacy octal literals are not allowed in strict mode",
         );
       }
-      if (/[89]/.test(this.input.slice(start, this.state.pos))) {
-        octal = false;
-      }
+      if (/[89]/.test(number)) octal = false;
+      if (/^[0-9]*$/.test(number)) isNonOctalDecimal = true;
```
JLHwung: `octal == true` implies that `startsWithDot == false` (because it starts with zero), which in turn means that `this.readInt(10) != null`, or otherwise we would throw `Invalid number` error here.
```js
if (!startsWithDot && this.readInt(10) === null) {
      this.raise(start, "Invalid number");
}
```
IIRC `this.readInt(10)` would ensure the input from `start` to `this.state.pos` to always be `[0-9]`, if so we don't have to check `[0-9]` again, which means `isNonOctalDecimal` equals to the previous definition of `octal`:
```js
this.state.pos - start >= 2 && this.input.charCodeAt(start) === charCodes.digit0;
```

nicolo-ribaudo: Yeah, so it would be enought to do
```js
if (/[89]/.test(number)) {
  octal = false;
  isNonOctalDecimal = true;
}
```

So that `isNonOctalDecimal` literally means "we tought that it was an octal number, but it turns out that it isn't"

gonzarodriguezt: You're right, i didn't see it in that way, it took me a while to realize why the test i added was passing, but it's because when `octal` is set to false here:

```javascript 
if (/[89]/.test(number)) octal = false;
```
`isNonOctalDecimal` remains true, so it can throw the error here:

```javascript
 if (isFloat || octal || isNonOctalDecimal) {...}
```
It's confusing, thanks for the review, i'll come back with some changes!

gonzarodriguezt: @nicolo-ribaudo Nice! Also, the `decimal` part isn't confusing? Sounds like fractional numbers, maybe it can only be `isNonOctal`?

nicolo-ribaudo: The spec calls it [`NonOctalDecimalIntegerLiteral`](https://tc39.es/ecma262/#prod-annexB-NonOctalDecimalIntegerLiteral), so maybe `nonOctalDecimalInt`?

But if you prefer `isNonOctal`, it's ok :+1:

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -1024,8 +1026,23 @@ export default class Tokenizer extends LocationParser {
 
     if (this.hasPlugin("bigInt")) {
       if (next === charCodes.lowercaseN) {
-        // disallow floats and legacy octal syntax, new style octal ("0o") is handled in this.readRadixNumber
-        if (isFloat || octal) this.raise(start, "Invalid BigIntLiteral");
+        // disallow numeric separators in non octal decimals
+        if (this.hasPlugin("numericSeparator") && isNonOctalDecimalInt) {
+          const underscorePos = this.input
```
nicolo-ribaudo: Could you move this whole `if` out of the `if (this.hasPlugin("bigInt"))`? It also applies to normal numbers:
```js
0_1;
```

gonzarodriguezt: Sure! I thought it was only for bigInts, can i ask if that's because bigInt is no more going to be experimental?

gonzarodriguezt: Done!

gonzarodriguezt:
Now all green again 🎉

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -1022,10 +1024,26 @@ export default class Tokenizer extends LocationParser {
...
     }
 
+    // disallow numeric separators in non octal decimals
+    if (this.hasPlugin("numericSeparator") && isNonOctalDecimalInt) {
+      const underscorePos = this.input
+        .slice(start, this.state.pos)
+        .indexOf("_");
+      if (underscorePos > 0) {
+        this.raise(
+          underscorePos,
```
JLHwung: Shouldn't it be `start + underscorePos`? 
```js
void 0_n;
```

scripts/tests/test262/test262_whitelist.txt
```diff
@@ -104,26 +104,12 @@ language/expressions/object/method-definition/private-name-early-error-gen-insid
...
-language/literals/bigint/numeric-separators/numeric-separator-literal-nonoctal-09-err.js(default)
-language/literals/bigint/numeric-separators/numeric-separator-literal-nonoctal-0_8-err.js(default)
-language/literals/bigint/numeric-separators/numeric-separator-literal-nonoctal-0_9-err.js(default)
-language/literals/bigint/numeric-separators/numeric-separator-literal-nzd-nsl-dds-leading-zero-err.js(default)
 language/literals/numeric/numeric-separators/numeric-separator-literal-lol-00-err.js(default)
 language/literals/numeric/numeric-separators/numeric-separator-literal-lol-01-err.js(default)
 language/literals/numeric/numeric-separators/numeric-separator-literal-lol-07-err.js(default)
 language/literals/numeric/numeric-separators/numeric-separator-literal-lol-0_0-err.js(default)
 language/literals/numeric/numeric-separators/numeric-separator-literal-lol-0_1-err.js(default)
 language/literals/numeric/numeric-separators/numeric-separator-literal-lol-0_7-err.js(default)
```
nicolo-ribaudo: If you want to open another PR to fix these tests, you are welcome to do so!

gonzarodriguezt:
Thanks guys! Appreciate the review, I learned a lot 🙏

scripts/tests/test262/test262_whitelist.txt
gonzarodriguezt: Great, i would like to do so! Should i make an issue or just as a follow up of this PR?

nicolo-ribaudo: Naa, just a PR

gonzarodriguezt: Goodd 👌

PR #10468, opened 2019-09-19
[parser] Disallow numeric separator in unicode scape sequences

### LINKED ISSUE #10460: Numeric separators should be disallowed in unicode escape sequences
## Bug Report

**Current and Expected Behavior**

Currently `@babel/parser` allows numeric separators (i.e. the `_` in `123_456`) inside unicode escape sequences, both in the `\u{...}` and `\uXXXX` forms. They should be disallowed.

**Input Code**

```js
"\u{1F_639}";
"\u12_34";
"\x1_2";
```

[repl](https://babeljs.io/repl#?babili=false&browsers=last%202%20versions%20and%20%3E%201%25%2C%20not%20ie%20%3C%3D%2011%2C%0Anot%20op_mini%20all&build=&builtIns=false&spec=false&loose=false&code_lz=EQHQrg3gjAYg-gNgMwE4C-wDcAoUYoBMcSALFkA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=script&lineWrap=true&presets=&prettier=false&targets=&version=7.6.0&externalPlugins=%40babel%2Fplugin-syntax-bigint%407.4.4%2C%40babel%2Fplugin-syntax-numeric-separator%407.2.0)

You may need to enable the `numericSeparator` parser plugin in order to debug this issue.

**Environment**
- Babel version(s): v7.6

**Possible Solution**

This is a bug in the `Tokenizer`, the part of `@babel/parser` which transforms the input source code into a lazy list of *tokens* (e.g. strings, numbers, identifiers, punctuators, ...).

The problem is that [`readInt`](https://github.com/babel/babel/blob/45a484f0a2802a932a951fca2f9cf1ba63b0aed5/packages/babel-parser/src/tokenizer/index.js#L883), when the `numericSeparator` plugin is enabled, always allows numeric separators.
When `readInt` is called by `readHexChar`, if it finds `_` it should throw something like `Numeric separators are not allowed inside unicode escape sequences`.

---

If anyone wants to work on this issue, please leave a comment below! If you need any help, plase ask me (or to any other Babel team member) on [Slack](https://slack.babeljs.io/) :slightly_smiling_face: 

1. Fork the repo
1. Run `make watch` (or `make build` whenever you change a file)
1. Add some tests (at least one for `\u{...}`, `\uXXXX` and one for `\xXX`) in `packages/babel-parser/test/fixtures/experimental/numeric-separator`. Please don't call it "invalid-*", but use a descriptive name :stuck_out_tongue:
    - If you want an example of a test for an error, you can check one of the `invalid-*` tests in that folder.
1. Update the code!
1. `yarn jest parser` to run the tests
1. Note: this change will likely also impact the results of running the [Test262](https://github.com/tc39/test262) suite on Babel.
    - Before running those tests you need to run `make bootstrap-test262` to download them
    - Use `make test-test262` to run them
    - If some tests which were previously failing now pass, you can run `make test-test262-update-whitelist`

ivandevp:
| Fixed Issues?            | Fixes #10460


packages/babel-parser/src/tokenizer/index.js
```diff
@@ -897,6 +897,19 @@ export default class Tokenizer extends LocationParser {
 
     let total = 0;
 
+    // Called from readHexChar
```
JLHwung: Other than `readCodePoint`, `readHexChar` could also be called from `readRadixNumber`. Did you checked whether `0xf_f` can be successfully parsed?

We could add an extra `allowNumSeparator: boolean` to `readHexChar`
```js
readHexChar(len: number, throwOnInvalid: boolean, allowNumSeparator: boolean): number | null
```
and set `allowNumSeparator = true` when calling from `readCodePoint`.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -897,6 +897,19 @@ export default class Tokenizer extends LocationParser {
...
+    if (radix === 16) {
+      const unicode = this.input.slice(this.state.pos, this.state.pos + len);
+
+      if (unicode.includes("_")) {
+        const numSeparatorPos = this.input.indexOf("_");
+        this.raise(
```
JLHwung: I don't think we should throw an error like this if user does _not_ enable `numericSeparator` parser plugin.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -897,6 +897,21 @@ export default class Tokenizer extends LocationParser {
 
     let total = 0;
 
+    if (this.hasPlugin("numericSeparator")) {
+      // Called from readHexChar
+      if (radix === 16) {
+        const unicode = this.input.slice(this.state.pos, this.state.pos + len);
+
+        if (unicode.includes("_")) {
+          const numSeparatorPos = this.input.indexOf("_");
```
nicolo-ribaudo: This doesn't work in case there is another `_` in the input somewhere before the number. e.g.

```js
_;

"\x1_2"
```

It would be better to move this check inside the `for` loop right after, where there already are the other errors about numeric separators.

packages/babel-parser/src/tokenizer/index.js
ivandevp: Good idea.

packages/babel-parser/src/tokenizer/index.js
ivandevp: You're right.

packages/babel-parser/src/tokenizer/index.js
ivandevp: Got it.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -917,6 +921,13 @@ export default class Tokenizer extends LocationParser {
             this.raise(this.state.pos, "Invalid or unexpected token");
           }
 
+          if (allowNumSeparator) {
```
nicolo-ribaudo: Shouldn't it only throw if it is _not_ allowed?

ivandevp: Oh... My bad. I'll fix it right now.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -917,6 +921,13 @@ export default class Tokenizer extends LocationParser {
             this.raise(this.state.pos, "Invalid or unexpected token");
           }
 
+          if (!allowNumSeparator) {
+            this.raise(
+              this.state.pos,
+              "Numeric separators are not allowed inside unicode escape sequences",
```
JLHwung: The Numeric separators are also not allowed in hex escape sequences:
```js
var c = \x1_0
```

ivandevp: Sure.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -1250,9 +1262,13 @@ export default class Tokenizer extends LocationParser {
 
   // Used to read character escape sequences ('\x', '\u').
 
-  readHexChar(len: number, throwOnInvalid: boolean): number | null {
+  readHexChar(
```
JLHwung: `readHexChar` is only used for `\x` and `\u`, so we don't have to expose `allowNumSeparator` to its caller.

ivandevp: :thinking: Good catch. So, now it is exposed to `readInt`'s callers and changed to `false` when called from `readHexChar`.

Issue #10470, opened 2019-09-20
[parser] Disallow private properties access on super


nicolo-ribaudo:
## Bug Report

**Input Code**

```js
class A extends B {
  #x;

  method() {
    super.#x;
  }
}
```

[repl](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAECC0FMAeAXBA7AJjAQtA3gFDTQDESA3IcdALYIoAWA9pgBQCUBNJEArgAcEAJwB05KiQC-hKUA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=script&lineWrap=false&presets=&prettier=false&targets=&version=7.6.0&externalPlugins=%40babel%2Fplugin-syntax-class-properties%407.2.0)

**Expected behavior/code**

It should throw a parser error.

**Specification**

In the class fields proposal, [`MemberExpression`](https://tc39.es/proposal-class-fields/#prod-MemberExpression) (which is the technical tern for "property access") is defined as follows:
```
MemberExpression:
  PrimaryExpression
  MemberExpression [ Expression ]           <-- foo[bar]
  MemberExpression . IdentifierName         <-- foo.bar
  MemberExpression TemplateLiteral          <-- foo`bar`
  SuperProperty
  MetaProperty
  new MemberExpression Arguments            <-- new foo.bar()
  MemberExpression . PrivateIdentifier      <-- foo.#x
```

As you can see, `.PrivateIdentifier` (where a `PrivateIdentifier` is like `#x`) is only allowed after another `MemberExpression`.

Let's see what [`SuperProperty`](https://tc39.es/ecma262/#prod-SuperProperty) is. It is not redefined by the proposal, so we must search it in the main specification.
```
SuperProperty:
  super [ Expression ]                      <-- super[foo]
  super . IdentifierName                    <-- super.#x
```

As you might see, there is nothing allowing `super.#x`

**Babel Configuration (.babelrc, package.json, cli command)**

In order to test this bug, you must enable the `classPrivateProperties` parser plugin

**Environment**
- Babel version(s): v7.6

**Possible Solution**

When we parse the `super` keyword, we check if it is followed by `(`, `[` or `.` and throw otherwise:

[...]

PR #10472, opened 2019-09-20
added check to disallow super.private variable access and test case added

vivek12345:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/10470 
| Documentation PR Link    | 


packages/babel-parser/src/parser/expression.js
```diff
@@ -1170,7 +1179,9 @@ export default class ExpressionParser extends LValParser {
     if (node.property.name !== propertyName || containsEsc) {
       this.raise(
         node.property.start,
-        `The only valid meta property for ${meta.name} is ${meta.name}.${propertyName}`,
+        `The only valid meta property for ${meta.name} is ${
```
vivek12345: this is my linter doing something crazy. I will try and remove this change.

packages/babel-parser/src/parser/expression.js
```diff
@@ -655,6 +655,15 @@ export default class ExpressionParser extends LValParser {
       node.object = base;
       node.property = this.parseMaybePrivateName();
       node.computed = false;
+      if (
+        node.property.type === "PrivateName" &&
+        node.object.type === "Super"
+      ) {
+        this.raise(
+          startPos,
+          "super is not allowed to be called on a private identifier of a class",
```
nicolo-ribaudo: Well, in this case `super` is not being called (it's not `super(#x)`).
What about `Private fields can't be accessed on super?`

vivek12345: Yes that message makes more sense. I will make the changes accordingly and also update the tests. Thank you

vivek12345:
@nicolo-ribaudo when I am doing
```
make test-test262-update-whitelist
```
I get the following error:
```
Error: Unsupported version of Test262: '4.0.0'
```

Any help here?

nicolo-ribaudo:
Could you try running `yarn upgrade test262-stream`?

vivek12345:
That did the trick. Thank you 👍

PR #10475, opened 2019-09-21
Correctly disambiguate / after async fuctions

### LINKED ISSUE #10410: [parser] Regex on new line after `typeof async function` is division
## Bug Report

Parser is incorrectly handling a parsing edge case. When a forward slash follows a `typeof` expression on a new line, without semi-colon, there is no ASI if the forward slash can form a division. In this case it can and so it should be a division.

**Current Behavior**

It'll try to parse a regular expression on the input code

**Input Code**
https://astexplorer.net/#/gist/15eb31fbb1024faa6f42a4940d8beb4a/20685a8eccf4257cb4e2d357f2ce58c94d1bc1f2

Babel 7.5.5

```js
typeof async function f(){}
/foo/g
```

(It seems to behave correctly for a regular function or just `x` as the `typeof` argument)

For reference, the next should pass but will crash in Babel (and Acorn):

```js
typeof async function f(){}
/ foo
```

And for the sake of completion, the next example should fail, because the first forward slash can form a division (and so it still fails on the second one because the parser shouldn't backtrack):
```js
typeof async function f(){}
/foo/
```

**Expected behavior/code**

It should parse a division (so `((typeof x) / foo) / g`). See output of Esprima, for example. Acorn has the same problem.

**Possible Solution**
At its core, the "are we parsing a regex or division" flag is incorrectly passed to the lexer. In the actual parser its implementation may be more complex than that.

It's a super edge case and pretty low prio.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #10410
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -19,6 +19,7 @@
 // [opp]: http://en.wikipedia.org/wiki/Operator-precedence_parser
 
 import { types as tt, type TokenType } from "../tokenizer/types";
+import { types as tc } from "../tokenizer/context";
 import * as N from "../types";
 import LValParser from "./lval";
 import {
@@ -949,6 +950,18 @@ export default class ExpressionParser extends LValParser {
           this.match(tt._function) &&
           !this.canInsertSemicolon()
         ) {
+          if (this.state.context.pop() !== tc.functionStatement) {
+            // Since "async" is an identifier and normally identifiers
+            // can't be followed by expression, the tokenizer assumes
+            // that "function" starts a statement.
+            // Fixing it in the tokenizer would mean traking not only the
+            // previous token ("async"), but also the one before to know
+            // if its beforeExpr value.
+            // It's easier and more efficient to adjust the context here.
+            throw new Error("Internal error");
+          }
+          this.state.context.push(tc.functionExpression);
+
           this.next();
           return this.parseFunction(node, undefined, true);
         } else if (
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -19,7 +19,7 @@
 // [opp]: http://en.wikipedia.org/wiki/Operator-precedence_parser
 
 import { types as tt, type TokenType } from "../tokenizer/types";
-import { types as tc } from "../tokenizer/context";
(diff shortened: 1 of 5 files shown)


packages/babel-parser/src/parser/expression.js
```diff
@@ -949,6 +950,18 @@ export default class ExpressionParser extends LValParser {
           this.match(tt._function) &&
           !this.canInsertSemicolon()
         ) {
+          if (this.state.context.pop() !== tc.functionStatement) {
+            // Since "async" is an identifier and normally identifiers
+            // can't be followed by expression, the tokenizer assumes
+            // that "function" starts a statement.
+            // Fixing it in the tokenizer would mean traking not only the
```
JLHwung:
```suggestion
            // Fixing it in the tokenizer would mean tracking not only the
```

packages/babel-parser/src/parser/expression.js
```diff
@@ -949,6 +950,18 @@ export default class ExpressionParser extends LValParser {
...
+            // previous token ("async"), but also the one before to know
+            // if its beforeExpr value.
```
JLHwung:
```suggestion
            // its beforeExpr value.
```

packages/babel-parser/src/parser/expression.js
```diff
@@ -949,6 +950,18 @@ export default class ExpressionParser extends LValParser {
...
+            // It's easier and more efficient to adjust the context here.
+            throw new Error("Internal error");
+          }
+          this.state.context.push(tc.functionExpression);
```
JLHwung: It should be more efficient to check the last element of `this.state.context` and simply replace it with `tc.functionExpression` without `push` and `pop`.

packages/babel-parser/src/parser/expression.js
```diff
@@ -19,6 +19,7 @@
 // [opp]: http://en.wikipedia.org/wiki/Operator-precedence_parser
 
 import { types as tt, type TokenType } from "../tokenizer/types";
+import { types as tc } from "../tokenizer/context";
```
JLHwung: nit: prefer `ct` to be consistent with naming in `tokenizer`.

PR #10477, opened 2019-09-21
Changes UMD callsite to be more likely to pass in the intended object.

### LINKED ISSUE #10476: Cannot load UMD module created with transform-es2015-modules-umd as an ES module.
## Bug Report

**Current Behavior**
If you apply the transform-es2015-modules-umd plugin to a module, the resulting code cannot be executed as an ES module.

The problem stems from the fact that the generated code passes `this` as the first parameter to the UMD function.  When executed as an ES module, `this` is undefined at the script scope so `undefined` gets passed into the UMD function.  `globalThis` and `window` are both defined in the browser context (I haven't tested NodeJS), so the UMD module _can_ still function if something besides `this` was passed in to the UMD function.

**Input Code**
```html
<!DOCTYPE html>
<html lang='en'>
	<head><meta charset='utf-8'></head>
	<body>
		<script type='module' src='my-module.js'></script>
	</body>
</html>
```
Original ES module (`my-module.mjs`):
```js
export default 42;
```
Generated UMD module (`my-module.js`):
```js
(function (global, factory) {
  if (typeof define === "function" && define.amd) {
    define(["exports"], factory);
  } else if (typeof exports !== "undefined") {
    factory(exports);
  } else {
    var mod = {
      exports: {}
    };
    factory(mod.exports);
    global.input = mod.exports;
  }
})(this, function (_exports) {
  "use strict";

  Object.defineProperty(_exports, "__esModule", {
    value: true
  });
  _exports.default = void 0;
  var _default = 42;
  _exports.default = _default;
});
```

**Expected behavior/code**
The module would execute when loaded as an ES module.

```js
{ 
  globals: { 
    "webextension-polyfill": "browser", 
  }, 
  exactGlobals: true, 
}
```

**Environment**
- Babel version(s): `"babel-plugin-transform-es2015-modules-umd": "^6.24.1"`
- Node/npm version: Unknown
- OS: Unknown
- Monorepo: No.
- How you are using Babel: Grunt

**Possible Solution**
Change the UMD function call to pass `globalThis || window || this` as the first parameter, rather than passing `this` as the first parameter.  This change will make the module fully functional when loaded as an ES module, while still retaining the current behavior when loaded as a script.

The problem showed up for me over at the webextension-polyfill page, spelunking lead me to realize the problem is with Babel, which that project uses for transpiling to UMD.
https://github.com/mozilla/webextension-polyfill/issues/202

MicahZoltu:
| Fixed Issues?            | Fixes #10476
| Documentation PR Link    |
| Any Dependency Changes?  |
| License                  | CC0/Unlicense/MIT/WTFPL/whatever you want.

See #10476 for details.

Note: This PR should be considered pseudocode and used as an illustration of the proposed fix.  I do not know nearly enough about this project to know if this is an appropriate solution to the problem, nor do I have the confidence to update the tests appropriately.

### DIFF
--- a/packages/babel-plugin-transform-modules-umd/src/index.js
+++ b/packages/babel-plugin-transform-modules-umd/src/index.js
@@ -27,7 +27,11 @@ const buildWrapper = template(`
 
       GLOBAL_TO_ASSIGN;
     }
-  })(this, function(IMPORT_NAMES) {
+  })(
+    typeof globalThis === "object" ? globalThis
+      : typeof self === "object" ? self
+      : this,
+    function(IMPORT_NAMES) {
   })
 `);
 
--- a/packages/babel-plugin-transform-function-name/test/fixtures/function-name/export-default-arrow-renaming-module-umd/output.js
+++ b/packages/babel-plugin-transform-function-name/test/fixtures/function-name/export-default-arrow-renaming-module-umd/output.js
@@ -10,7 +10,7 @@
     factory(mod.exports);
     global.input = mod.exports;
   }
-})(this, function (_exports) {
+})(typeof globalThis === "object" ? globalThis : typeof self === "object" ? self : this, function (_exports) {
   "use strict";
 
   Object.defineProperty(_exports, "__esModule", {
--- a/packages/babel-plugin-transform-modules-umd/test/fixtures/loose/export-default-10/output.js
+++ b/packages/babel-plugin-transform-modules-umd/test/fixtures/loose/export-default-10/output.js
@@ -10,7 +10,7 @@
     factory(mod.exports);
     global.input = mod.exports;
   }
-})(this, function (_exports) {
+})(typeof globalThis === "object" ? globalThis : typeof self === "object" ? self : this, function (_exports) {
   "use strict";
 
   _exports.__esModule = true;
--- a/packages/babel-plugin-transform-modules-umd/test/fixtures/loose/export-default-11/output.js
+++ b/packages/babel-plugin-transform-modules-umd/test/fixtures/loose/export-default-11/output.js
@@ -10,7 +10,7 @@
     factory(mod.exports);
     global.input = mod.exports;
   }
-})(this, function (_exports) {
+})(typeof globalThis === "object" ? globalThis : typeof self === "object" ? self : this, function (_exports) {
   "use strict";
 
   _exports.__esModule = true;
--- a/packages/babel-plugin-transform-modules-umd/test/fixtures/loose/export-default-2/output.js
+++ b/packages/babel-plugin-transform-modules-umd/test/fixtures/loose/export-default-2/output.js
@@ -10,7 +10,7 @@
     factory(mod.exports);
     global.input = mod.exports;
   }
-})(this, function (_exports) {
+})(typeof globalThis === "object" ? globalThis : typeof self === "object" ? self : this, function (_exports) {
   "use strict";
 
   _exports.__esModule = true;
--- a/packages/babel-plugin-transform-modules-umd/test/fixtures/loose/export-default-3/output.js
+++ b/packages/babel-plugin-transform-modules-umd/test/fixtures/loose/export-default-3/output.js
@@ -10,7 +10,7 @@
     factory(mod.exports);
     global.input = mod.exports;
   }
-})(this, function (_exports) {
+})(typeof globalThis === "object" ? globalThis : typeof self === "object" ? self : this, function (_exports) {
   "use strict";
 
   _exports.__esModule = true;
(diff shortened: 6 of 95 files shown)


packages/babel-plugin-transform-modules-umd/src/index.js
```diff
@@ -27,7 +27,7 @@ const buildWrapper = template(`
 
       GLOBAL_TO_ASSIGN;
     }
-  })(this, function(IMPORT_NAMES) {
+  })(globalThis || window || this, function(IMPORT_NAMES) {
```
nicolo-ribaudo: Wouldn't this throw a `globalThis is not defined` error in browsers with modules supprot (where the code is executed in strict mode) but without `globalThis` support?

Those browsers are Edge 16, Firefox 60-64, Chrome 61-70, Safari 10.1-12.0 and Opera 48-57

nicolo-ribaudo:
```suggestion
  })(
    typeof globalThis === "object" ? globalThis
      : typeof window === "object" ? window
      : this,
    function(IMPORT_NAMES) {
```

jridgewell: Definitely need Nicolo's suggestion, but we should use `self` instead of `window` to support workers as well.

nicolo-ribaudo:
After the code changes, you can run `make bootstrap` and `OVERWRITE=true make test` to update the failing tests

packages/babel-plugin-transform-modules-umd/src/index.js
MicahZoltu:
````javascript
typeof globalThis === "object" ? globalThis
: typeof self === "object" ? self
: typeof window === "object" ? window
: typeof global === "object" ? global
: this
```
Is that too overboard on trying?  I think there are environments where `window` is defined but not `self` or `globalThis` (older browsers), and older versions of NodeJS only define `global` I believe.  Does it make sense to fallback to `this` at the end?  Perhaps for test environments the `this` fallback is valuable?

jridgewell: I think just `self` is enough. IE has supported it forever. As for `global`, let's assume people can just load a `globalThis` polyfill in node. Client side we have to worry about bytes.

MicahZoltu:
> After the code changes, you can run make bootstrap and OVERWRITE=true make test to update the failing tests

Unfortunately I'm on Windows, so I don't have `make`.

MicahZoltu:
Also the pre-commit hook is failing on my machine in what appears to be unrelated code, so the current commit is pushed with `--no-verify`.

packages/babel-plugin-transform-modules-umd/src/index.js
MicahZoltu: The suggested changes have been applied and the code rebased.  Anyone can feel free to takeover and finish this up (e.g., update the tests).  I'm on Windows and this repo appears to have a number of issues with Windows so I doubt I'll be able to finish it.

jridgewell:
Pushed updated tests.

jridgewell:
Actually, do we even need a `globalThis` check? The `global` parameter is only used if we're not using requirejs and we're not in node. That means we're in browser mode, and `self` will be enough.

MicahZoltu:
`globalThis` is _intended_ to be the future of the universal global object.  It feels like a good idea to try that first as one day (far, far in the future) `window`, `self`, and/or `global` may be deprecated.  Also, glancing at the `globalThis` docs on MDN it also references `frames` in some situations, not sure if we need to or should handle that as well?

JLHwung:
> Unfortunately I'm on Windows, so I don't have make.

[gnuwin32](http://gnuwin32.sourceforge.net/packages/make.htm) works for me on Windows.

MicahZoltu:
Does the PR build process really take over 21 days to complete, or is it stalled?  If it is stalled, what can be done to restart it so we can move forward with this PR?

jridgewell:
I kicked the build servers. It definitely shouldn't take this long.

MicahZoltu:
The Travis build failure does not appear to be related to the changes I have made in this PR.  Can someone more familiar with this project glance at the build failure and help me understand how they are related to these changes if they are?

MicahZoltu:
The build failure doesn't appear to have anything with these changes.  Can I get someone familiar with the project to validate that claim?

MicahZoltu:
Is there anything I can or should do to help move this along, or is it just a matter of waiting for it to get pickup up in a release?

jridgewell:
Waiting on @nicolo-ribaudo to merge before the next appropriate release.

nicolo-ribaudo:
(Soon)

PR #1752, opened 2015-06-15
Initialize properties to avoid hidden class thrashing.

loganfsmyth:
I *think* this is all the properties for `NodePath`?

On my machine, building ember goes down 10-15%ish.

PR #10480, opened 2019-09-22
Traverse performance

JLHwung:
| Minor: Bug Fix? | 👍
| Major: Breaking Change?  | No if one is using public API.

In this PR we improve the performance of babel-traverse by revising the memory layout of `NodePath`. It results to 10% performance gain compared to the control branch on the following test cases.

```js
const babel = require("/path/to/local/babel-core");
const code = require("fs").readFileSync("./fixtures/es5/babylon-dist.js");
babel.transform(code, { preset: ["@babel/env"] });
```

I can observe this branch outperforms control branch consistently.
```
┌─────────────────────┬──────────────────────────────┐
│ fixture             │ devTransform                 │
├─────────────────────┼──────────────────────────────┤
│ es5/babylon-dist.js │ 1.02 ops/sec ±31.84% (978ms) │
└─────────────────────┴──────────────────────────────┘
┌─────────────────────┬───────────────────────────────┐
│ fixture             │ baseTransform                 │
├─────────────────────┼───────────────────────────────┤
│ es5/babylon-dist.js │ 0.86 ops/sec ±20.52% (1167ms) │
└─────────────────────┴───────────────────────────────┘

```

And the memory usage improvements due to compressing traverse flags and adding accessor methods for `inList` and `parentPath`.
```
┌─────────────────────┬──────────────────┐
│ fixture             │ devTransform x5  │
├─────────────────────┼──────────────────┤
│ es5/babylon-dist.js │ heap: 239.37 MiB │
└─────────────────────┴──────────────────┘
┌─────────────────────┬──────────────────┐
│ fixture             │ baseTransform x5 │
├─────────────────────┼──────────────────┤
│ es5/babylon-dist.js │ heap: 305.47 MiB │
└─────────────────────┴──────────────────┘
```

The following changes are implemented:
- Removed `NodePath.inList` data property, a get accessor is provided for backward compatibility (partially). 
- Removed `NodePath.parentKey` data property, a get accessor is provided for backward compatibility (partially). 
- Removed `NodePath.typeAnnotation` data property, it is introduced in #1752 but since then it is not used by any routines. I believe it is a mistype of `Node.typeAnnotation`.
- Compressed `NodePath.shouldSkip/shouldStop/removed` into `_traverseFlags` bit array, both get accessor and set accessor is provided for backward compatibility.
- Lazily assign `this.skipKeys` in `setContext`.
- Lazily initialize `this.data` in `NodePath.constructor`.

In practice `inList` and `parentKey` should be treated as read-only since these two properties are rewritten in `setContext`. The current behavior accepts one assigning these two properties.

The `shouldSkip/shouldStop/removed` traverse flags are handy and turbofan would inline these simple accessors we don't introduce performance overhead here. By compressing these states into bit array we optimize the performance of `setContext`.

### DIFF
--- a/packages/babel-traverse/src/index.js
+++ b/packages/babel-traverse/src/index.js
@@ -31,6 +31,10 @@ export default function traverse(
     }
   }
 
+  if (!t.VISITOR_KEYS[parent.type]) {
+    return;
+  }
+
   visitors.explode(opts);
 
   traverse.node(parent, opts, scope, state, parentPath);
--- a/packages/babel-traverse/src/path/context.js
+++ b/packages/babel-traverse/src/path/context.js
@@ -44,6 +44,7 @@ export function _call(fns?: Array<Function>): boolean {
     // node has been replaced, it will have been requeued
     if (this.node !== node) return true;
 
+    // this.shouldSkip || this.shouldStop || this.removed
     if (this._traverseFlags > 0) return true;
   }
 
@@ -100,6 +101,7 @@ export function skipKey(key) {
 }
 
 export function stop() {
+  // this.shouldSkip = true; this.shouldStop = true;
   this._traverseFlags |= SHOULD_SKIP | SHOULD_STOP;
 }
 
@@ -123,6 +125,7 @@ export function setContext(context) {
   if (this.skipKeys != null) {
     this.skipKeys = {};
   }
+  // this.shouldSkip = false; this.shouldStop = false; this.removed = false;
   this._traverseFlags = 0;
 
   if (context) {
--- a/packages/babel-traverse/src/path/context.js
+++ b/packages/babel-traverse/src/path/context.js
@@ -216,7 +216,6 @@ export function pushContext(context) {
 
 export function setup(parentPath, container, listKey, key) {
   this.listKey = listKey;
-  this.parentKey = listKey || key;
   this.container = container;
 
   this.parentPath = parentPath || this.parentPath;
--- a/packages/babel-traverse/src/path/context.js
+++ b/packages/babel-traverse/src/path/context.js
@@ -93,6 +93,9 @@ export function skip() {
 }
 
 export function skipKey(key) {
+  if (this.skipKeys == null) {
+    this.skipKeys = {};
+  }
   this.skipKeys[key] = true;
 }
 
@@ -117,7 +120,9 @@ export function setScope() {
 }
 
 export function setContext(context) {
-  this.skipKeys = {};
+  if (this.skipKeys != null) {
+    this.skipKeys = {};
+  }
   this._traverseFlags = 0;
 
   if (context) {
--- a/packages/babel-traverse/src/path/context.js
+++ b/packages/babel-traverse/src/path/context.js
@@ -1,6 +1,7 @@
 // This file contains methods responsible for maintaining a TraversalContext.
 
 import traverse from "../index";
+import { SHOULD_SKIP, SHOULD_STOP } from "./index";
 
 export function call(key): boolean {
   const opts = this.opts;
@@ -43,7 +44,7 @@ export function _call(fns?: Array<Function>): boolean {
     // node has been replaced, it will have been requeued
     if (this.node !== node) return true;
 
-    if (this.shouldStop || this.shouldSkip || this.removed) return true;
+    if (this._traverseFlags > 0) return true;
   }
 
   return false;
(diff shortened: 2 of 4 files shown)


packages/babel-traverse/src/index.js
```diff
@@ -33,7 +33,11 @@ export default function traverse(
 
   visitors.explode(opts);
 
-  traverse.node(parent, opts, scope, state, parentPath);
+  if (!t.VISITOR_KEYS[parent.type]) {
+    return;
```
nicolo-ribaudo: Can we move this check before calling `visitors.explode`?

packages/babel-traverse/src/path/interface-compatibility.js
```diff
@@ -0,0 +1,77 @@
+let deprecateShown = false;
+function showDeprecationWarning() {
+  if (!deprecateShown) {
+    console.trace(
+      "Accessing NodePath.skipKeys is deprecated, use NodePath.skipKey(key) instead",
```
nicolo-ribaudo: I'm ok with adding this deprecation message, but I'd prefer to wait for a minor release.

packages/babel-traverse/src/path/interface-compatibility.js
```diff
@@ -0,0 +1,77 @@
...
+    );
+  }
+  deprecateShown = true;
+}
+
+export function generateSkipKeysGetter(skipKeysSet) {
+  return new Proxy(
```
nicolo-ribaudo: Can we cache this proxy, to avoid recreating it every time that `skipKeys` is read?

packages/babel-traverse/src/index.js
```diff
@@ -55,11 +59,17 @@ traverse.node = function(
   const keys: Array = t.VISITOR_KEYS[node.type];
   if (!keys) return;
 
-  const context = new TraversalContext(scope, opts, state, parentPath);
-  for (const key of keys) {
-    if (skipKeys && skipKeys[key]) continue;
-    if (context.visit(node, key)) return;
+  if (!skipKeys) {
+    skipKeys = {};
```
nicolo-ribaudo: You can define it as a default parameter

packages/babel-traverse/src/path/interface-compatibility.js
JLHwung: Oh I find out another way to skip the whole `skipKeysSet` things and achieve similar performance.

packages/babel-traverse/src/path/context.js
```diff
@@ -92,12 +93,14 @@ export function skip() {
 }
 
 export function skipKey(key) {
+  if (this.skipKeys == null) {
+    this.skipKeys = {};
+  }
```
nicolo-ribaudo: I wish https://github.com/tc39/proposal-logical-assignment wasn't just stage 1 :stuck_out_tongue:

JLHwung:
@nicolo-ribaudo I am sorry I make a typo before. It should be the `parentKey` instead of `parentPath`, but your point is still valid on `parentKey`.

> I think that any plugin relying on the old behavior was actually broken: if someone wanted to change the parent node by doing path.parentKey = ... it wouldn't have worked.

I agree with you that setting `inList` and/or `parentKey` never works in practice. Instead of failing silently, babel will now report `can not assign to an accessor`. In this way it seems more like a bug fix, though. 😁

packages/babel-traverse/src/path/context.js
JLHwung: Offtopic: we still need a PR to add nullish coalescing and optional chaining.

hzoo: We can always use new stage features if you are willing to move off of them if they change (I think that one is small enough that it's fine)

packages/babel-traverse/src/path/index.js
```diff
@@ -23,29 +23,29 @@ import * as NodePath_comments from "./comments";
 
 const debug = buildDebug("babel");
 
+export const REMOVED = 1 << 0;
+export const SHOULD_STOP = 1 << 1;
+export const SHOULD_SKIP = 1 << 2;
```
hzoo: Some great work doing this research!

I would just like to note that using bitwise flags and operators is for sure going to make contributing and reading the coding that much more difficult (for me, and I will assume some others). I think I mentioned it earlier on Slack and it's a hard call but trading off readability/performance. I don't know what else we can do other than writing comments to explain things though

PR #10484, opened 2019-09-23
[parser] Add support for TS declare modifier on fields

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Ref: https://github.com/microsoft/TypeScript/pull/33401

I also updated the parser to generate more useful errors when a class member has a disallowed modifier (e.g. `readonly method() {}`) 

cc @babel/typescript

### DIFF
--- a/packages/babel-parser/src/plugins/typescript/index.js
+++ b/packages/babel-parser/src/plugins/typescript/index.js
@@ -127,12 +127,14 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return undefined;
     }
 
-    tsParseModifiers<T: TsModifier>(allowedModifiers: T[]): { [*]: true } {
-      const modifiers = {};
+    tsParseModifiers<T: TsModifier>(
+      allowedModifiers: T[],
+    ): { [key: TsModifier]: ?true, __proto__: null } {
+      const modifiers = Object.create(null);
 
       while (true) {
         const startPos = this.state.start;
-        const modifier = this.tsParseModifier(allowedModifiers);
+        const modifier: ?T = this.tsParseModifier(allowedModifiers);
 
         if (!modifier) break;
 
--- a/packages/babel-parser/src/plugins/typescript/index.js
+++ b/packages/babel-parser/src/plugins/typescript/index.js
@@ -127,6 +127,11 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return undefined;
     }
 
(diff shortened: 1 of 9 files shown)


sandersn:
Note that the combined PR that I intend to merge is microsoft/TypeScript#33509.

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -124,6 +127,24 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+        const startPos = this.state.start;
+        const modifier = this.tsParseModifier(allowedModifiers);
+
+        if (!modifier) break;
+
+        if (Object.hasOwnProperty.call(modifiers, modifier)) {
+          this.raise(startPos, `Duplicate modifier: '${modifier}'`);
+        }
+        modifiers[modifier] = true;
+      }
```
Jessidhia: This seems to allow any modifier in any order; I presume that's okay? or are there some modifiers which have a specific required order (e.g. `declare` first)

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -124,6 +127,24 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return undefined;
     }
 
+    tsParseModifiers<T: TsModifier>(allowedModifiers: T[]): { [*]: true } {
+      const modifiers = {};
```
Jessidhia: I'd feel better with `Object.create(null)` but it probably isn't really relevant here.

packages/babel-parser/src/plugins/typescript/index.js
nicolo-ribaudo: Some modifiers must come in a specific order, some others don't. For example, `abstract` and `readonly` can be swapped, but `public` must be before them.

We parse them using a logic similar to this one, which ensures the correct order:
```js
this.tsParseModifiers(["public"]);
this.tsParseModifiers(["abstract", "readonly"]);
```

existentialism: Not super important, but it might be nice to add a comment around modifier order/usage.

nicolo-ribaudo:
I'll merge this to a separate branch so that I can start working on the transform.

PR #10491, opened 2019-09-24
Trailing comma after rest - The final fix

nicolo-ribaudo:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Tracking a comma after a spread/rest element is ticky, because we don't know when we should reset the state. We can't do it in `parseExpression` because, for example, in `[[...a,]] = []` `parseExpression` is finished after parsing the inner array.


packages/babel-parser/src/parser/lval.js
```diff
@@ -60,6 +60,14 @@ export default class LValParser extends NodeUtils {
             const prop = node.properties[i];
             const isLast = i === last;
             this.toAssignableObjectExpressionProp(prop, isBinding, isLast);
+
+            if (
+              isLast &&
+              prop.type === "RestElement" &&
+              node.extra?.trailingComma
+            ) {
+              this.raiseRestNotLast(node.extra.trailingComma);
```
JLHwung: By raising error at comma position we are not really consistent with where this error is throw in other situation, i.e. in `toAssignableObjectExpressionProp`.

https://github.com/babel/babel/blob/11fa2461cec930168fd43cfeda733aab056d8318/packages/babel-parser/src/parser/lval.js#L132-L134

Hence we could throw at `prop.start`.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1488,7 +1489,11 @@ export default class ExpressionParser extends LValParser {
         first = false;
       } else {
         this.expect(tt.comma);
-        if (this.eat(tt.braceR)) break;
+        if (this.match(tt.braceR)) {
+          this.addExtra(node, "trailingComma", this.state.lastTokStart);
```
JLHwung: What if we instead record the `traillingCommaAttachedNodeStart`? So we can always throw at the node.

nicolo-ribaudo: I'd prefer to make `trailingComma` a boolean, and then throw at `props[props.length - 1].start`

nicolo-ribaudo:
@JLHwung I ended up fixing the error location in a different way.
It now throws two different errors:
1. `Rest element must be last element`, pointing to the rest element
1. `Unexpected trailing comma after rest element`, pointing to the trailing comma

packages/babel-parser/src/parser/lval.js
```diff
@@ -455,19 +469,22 @@ export default class LValParser extends NodeUtils {
...
     if (this.match(tt.comma)) {
-      this.raiseRestNotLast(this.state.start);
-    }
-  }
-
-  checkCommaAfterRestFromSpread(): void {
-    if (this.state.commaAfterSpreadAt > -1) {
-      this.raiseRestNotLast(this.state.commaAfterSpreadAt);
+      // TODO: Use lookaheadCharCode after that https://github.com/babel/babel/pull/10371 is merged
+      if (this.lookahead().type == close) {
```
nicolo-ribaudo: https://github.com/babel/babel/pull/10371

Issue #10486, opened 2019-09-24
`@babel/types` should validate the number of elements inside template literals


nicolo-ribaudo:
## Feature Request

`@babel/types`, when calling `t.templateLiteral` , should throw an error if the number of `quasis` (static template strings) is not exactly one more than the number of `expressions` (substitutions).

**Possible Solution**

`@babel/types` already already provides the ability to define custom validation functions, for example
https://github.com/babel/babel/blob/a00f3716106a0811ff9b915783bb90f2949ef48a/packages/babel-types/src/definitions/core.js#L408-L412

You should add a custom validator to `TemplateLiteral`, defined in the `packages/babel-types/src/definitions/es2015.js` file.

---

If you don't know how to clone Babel, follow these steps: (you need to have `make` and `yarn` available on your machine).
1. Write a comment there to let other possible contributors know that you are working on this bug.
1. Fork the repo
1. Run `git clone https://github.com/<YOUR_USERNAME>/babel.git && cd babel`
1. Run `yarn && make bootstrap`
1. Wait :hourglass_flowing_sand: 
1. Run `make watch` (or `make build` whenever you change a file)
1. Add a test at `packages/babel-types/test/builders/es2015/templateElement.js`
1. Update the code!
1. `yarn jest babel-types` to run the tests
    - If you need to update some snapshots generated by Jest, you can run the tests again with the `-u` cli flag
1. If it is working, run `make test` to run all the tests
1. Run `git push` and open a PR!

Basaingeal:
I'll give this one a go.

tsareg:
Hey folks, after upgrading to recent @babel/types (namely 7.6.3) I started to get:
"Number of TemplateLiteral quasis should be exactly one more than the number of expressions.
Expected 2 quasis but got 1"

The code I'm trying is dynamic import with `@loadable/component`:
```js
const AsyncIcon = loadable(props => import(`./icons/${props.name}.js`));
```

This was working fine with @babel/types 7.4.4.
Can you please explain what's wrong now with this code and what needs to be changed?

PR #10492, opened 2019-09-24
Template literal validation

Basaingeal:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/10486
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    |
| Any Dependency Changes?  |

I added a custom validation rule to the definition of `TemplateLiteral` in `packages\babel-types\test\builders\es2015\templateElement.js` to check that the number of quasis is exactly one more than the number of expressions.

I'm open to suggestions about the wording of the TypeError. I based it off of the number of quasis based off the wording of the issue, but I'm not sure if it needs to be in more layman's terms or more generalized or not.

### DIFF
--- a/packages/babel-types/src/definitions/es2015.js
+++ b/packages/babel-types/src/definitions/es2015.js
@@ -568,6 +568,16 @@ defineType("TemplateLiteral", {
       validate: chain(
         assertValueType("array"),
         assertEach(assertNodeType("TemplateElement")),
+        function(node, key, val) {
+          if (val.length !== node.expressions.length + 1) {
+            throw new TypeError(
+              `Length of ${
+                node.type
+              } ${key} should be exactly one more than the number of expressions.\nExpected ${node
+                .expressions.length + 1} ${key} but got ${val.length}`,
+            );
+          }
+        },
       ),
     },
     expressions: {
--- a/packages/babel-types/src/definitions/es2015.js
+++ b/packages/babel-types/src/definitions/es2015.js
@@ -568,25 +568,22 @@ defineType("TemplateLiteral", {
       validate: chain(
         assertValueType("array"),
         assertEach(assertNodeType("TemplateElement")),
-        function(node, key, val) {
-          if (
-            Array.isArray(node.expressions) &&
-            val.length !== node.expressions.length + 1
-          ) {
-            throw new TypeError(
-              `Length of ${
-                node.type
-              } ${key} should be exactly one more than the number of expressions.\nExpected ${node
-                .expressions.length + 1} ${key} but got ${val.length}`,
-            );
-          }
-        },
       ),
     },
     expressions: {
       validate: chain(
         assertValueType("array"),
         assertEach(assertNodeType("Expression")),
+        function(node, key, val) {
+          if (node.quasis.length !== val.length + 1) {
+            throw new TypeError(
+              `Number of ${
+                node.type
+              } quasis should be exactly one more than the number of expressions.\nExpected ${val.length +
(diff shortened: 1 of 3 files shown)


packages/babel-types/src/definitions/es2015.js
```diff
@@ -568,6 +568,19 @@ defineType("TemplateLiteral", {
       validate: chain(
         assertValueType("array"),
         assertEach(assertNodeType("TemplateElement")),
+        function(node, key, val) {
+          if (
+            Array.isArray(node.expressions) &&
```
jridgewell: Nit: If we move this to `expressions` validation instead, we won't have to check that `node.quasis` is an array because that will already have been validated.

packages/babel-types/test/builders/es2015/templateElement.js
```diff
@@ -21,5 +21,31 @@ describe("builders", function() {
         expect(() => t.templateElement("foo")).toThrowErrorMatchingSnapshot();
       });
     });
+    describe("templateLiteral", function() {
+      it("should validate", function() {
+        const foo = t.templateElement({ raw: "foo" });
+        const bar = t.templateElement({ raw: "bar" });
+
+        expect(() => t.templateLiteral([foo], [])).toMatchSnapshot();
```
jridgewell: Don't wrap this in an arrow function. That's why we're getting `[Function]` in the snapshots.

packages/babel-types/test/builders/es2015/templateElement.js
```diff
@@ -21,5 +21,31 @@ describe("builders", function() {
...
+        expect(() =>
+          t.templateLiteral([foo, bar], ["baz", "qux"]),
+        ).toThrowErrorMatchingSnapshot();
+
+        expect(() =>
+          t.templateLiteral([foo, bar], []),
+        ).toThrowErrorMatchingSnapshot();
+
+        expect(() =>
+          t.templateLiteral({}, ["baz"]),
```
jridgewell: The first needs to be an array, and the second an array of expressions (not a string).

Basaingeal: Those tests are put there intentionally to show that the previous validators still trigger and that the only time the new validation rule triggers is when that's the only thing wrong.

I can remove it though if it's considered irrelevant.

nicolo-ribaudo:
Thanks!

nicolo-ribaudo:
@Basaingeal I added `Co-authored-by` to the commit message to link the commit to your GitHub accout, but I suggest configuring it so that both your emails get linked!

tsareg:
Hey folks, after upgrading to recent @babel/types (namely 7.6.3) I started to get:
"Number of TemplateLiteral quasis should be exactly one more than the number of expressions.
Expected 2 quasis but got 1"

The code I'm trying is dynamic import with `@loadable/component`:
```js
const AsyncIcon = loadable(props => import(`./icons/${props.name}.js`));
```

This was working fine with @babel/types 7.4.4.
Can you please explain what's wrong now with this code and what needs to be changed?

nicolo-ribaudo:
Could you open a new issue with your configuration?

tsareg:
> Could you open a new issue with your configuration?

I've created a bug for @loadable/component, see above. I was just wondering what have actually changed with the fix, I think this is where they use `templateLiteral` https://github.com/smooth-code/loadable-components/blob/master/packages/babel-plugin/src/properties/chunkName.js#L79-L88

PR #10494, opened 2019-09-24
Enable optional chaining and nullish coalescing plugins

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  |

They are both stage 3, and they are used in a few PRs (e.g. https://github.com/babel/babel/pull/10491, https://github.com/babel/babel/pull/10357)

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -25,4 +25,5 @@ suppress_type=$FlowFixMe
 suppress_type=$FlowSubtype
 esproposal.export_star_as=enable
 esproposal.optional_chaining=enable
+esproposal.nullish_coalescing=enable
 module.name_mapper='^@babel\/\([a-zA-Z0-9_\-]+\)$' -> '<PROJECT_ROOT>/packages/babel-\1/src/index'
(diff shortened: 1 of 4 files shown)


nicolo-ribaudo:
Well, they aren't shipped anywhere yet :stuck_out_tongue: 

https://kangax.github.io/compat-table/esnext/#test-optional_chaining_operator_(?.)

mrtnzlml:
@nicolo-ribaudo It's also a good idea to enable this lint rule to make sure it's used correctly: [unnecessary-optional-chain](https://flow.org/en/docs/linting/rule-reference/#toc-unnecessary-optional-chain)

Just a small tip 🍻 :)

jsg2021:
> Nice! Should we add these to preset-env shippedProposals eventually? Not sure how many people are using that anyway though

I use `shippedProposals` 😉

nicolo-ribaudo:
Spoiler: given that we release a new minor about every 2 months, probably `shippedProposals` won't be the correct place for at least one of them 😉

PR #10500, opened 2019-09-26
Ci/test262

rricard:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,7 +89,7 @@ jobs:
             git clone --recurse-submodules https://github.com/babel/babel-test262-runner
             cd babel-test262-runner
             yarn
-            # yarn add tap-xunit@2.1.0 --dev
+            # yarn add tap-mocha-reporter --dev
             curl -L https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64 > jq
             chmod +x ./jq
             for package in ../packages/*/package.json; do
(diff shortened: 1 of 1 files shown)


rricard:
- [ ] Revert back full build and hold
- [ ] run test262 against the built version of babel in the repo

rricard:
Nice, the test output is not super useful yet but we might be able to do something out of the xunit file for further analysis.

jbhoosreddy:
@rricard yes but I don't want to run test262 twice for master and PR. Ideally we should be able to reuse the build artifact from master to compare and analyse.

It already takes 1.5 hours to run the tests once.

I am hoping @circleci can highlight new errors from existing errors like Jenkins can simply by giving it an xunit file

jbhoosreddy:
> test262.tap	[new file]	[110.87 KB](https://11741-24560307-gh.circle-artifacts.com/0/home/circleci/babel/babel-test262-runner/test262.tap)

This doesn't get added to babel build somehow, right? Or I need to add it in .npmignore or something?

nicolo-ribaudo:
> This doesn't get added to babel build somehow, right?

No, because it isn't committed to the repo or present in our local copies.

jbhoosreddy:
So I put a support ticket to @circleci and turns out they do not have the Jenkins-style rendering of new vs existing test failures.
Please vote for: https://ideas.circleci.com/ideas/CCI-I-1192
If you find it relevant.

In the meanwhile, we may have to do this:

“ And also, we can store the previous test results data as artifacts and it can download via API, so I think you can also create a script that compares the test results.”

For CI, I think if we can highlight if a PR has failed new tests. But in terms of a web UI or anything fancy we should work on getting it added to test262.report

PR #10501, opened 2019-09-26
Add support for babel.config.json

### LINKED ISSUE #10495: RFC: Support babel.config.json for better cacheability
At the moment, Babel supports several configuration options: `.babelrc`, `.babelrc.js`, and `babel.config.js`. Many projects have migrated to `babel.config.js` since it was introduced, especially monorepos. However, many configs are still quite simple and could be JSON instead of JavaScript. This has several benefits, including better cacheability. Babel should support `babel.config.json` in addition to its existing options for those cases.

Using JavaScript for configs can be nice when you need to do complex condition-based configuration, but it has several downsides including preventing caching. In Parcel 2, we have to bail on caching all files compiled with Babel when a JavaScript config file is used instead of JSON. Unfortunately, that leaves no recourse to users if they want the benefits of caching, but want to use a root config. If Babel supported `babel.config.json`, these issues would be alleviated somewhat. For simple configs that are JSON serializable, Parcel and other tools could correctly cache files as is already possible with `.babelrc`.

I am aware of the Babel config [caching](https://babeljs.io/docs/en/config-files#apicache) API, but it's not really designed for use in a cache that is serialized to disk. It is designed to avoid re-calling the config function, not to avoid re-compiling files. In addition, the API relies on callback functions to determine if the cache should be invalidated, but that only works when the callback can be stored in memory - we cannot serialize it to disk.

devongovett:
Implements support for `babel.config.json` in addition to the existing `babel.config.js`. This allows root configuration to be static JSON, which has many benefits including better cacheability, easier linting, better autocomplete, and better upgrade tools. See #10495 for more details.

| Fixed Issues?            | Closes #10495 
| Documentation PR Link    | 

### DIFF
--- a/packages/babel-core/src/config/files/configuration.js
+++ b/packages/babel-core/src/config/files/configuration.js
@@ -19,6 +19,11 @@ import type { CallerMetadata } from "../validation/options";
 const debug = buildDebug("babel:config:loading:files:configuration");
 
 const BABEL_CONFIG_JS_FILENAME = "babel.config.js";
+const BABEL_CONFIG_JSON_FILENAME = "babel.config.json";
+const ROOT_CONFIG_FILENAMES = [
+  BABEL_CONFIG_JS_FILENAME,
+  BABEL_CONFIG_JSON_FILENAME,
+];
 
 const BABELRC_FILENAME = ".babelrc";
 const BABELRC_JS_FILENAME = ".babelrc.js";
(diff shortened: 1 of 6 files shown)


nicolo-ribaudo:
We should throw if both are used in the same folder

devongovett:
Good idea @nicolo-ribaudo! Refactored to match how `findRelativeConfig` did this.

PACelah:
Nice-!"^")\,`DevonGovett

nicolo-ribaudo:
@devongovett Could you add this new config file to https://github.com/babel/website/blob/master/docs/config-files.md?

devongovett:
@nicolo-ribaudo https://github.com/babel/website/pull/2106

nicolo-ribaudo:
(Rebased)

PR #10504, opened 2019-09-27
Add declarations for more of @babel/types exports

Jessidhia:
| Patch: Bug Fix?          |
| Major: Breaking Change?  | not yet
| Tests Added + Pass?      | do we even have infrastructure for testing types?
| Documentation PR Link    | 
| Any Dependency Changes?  |

This adds lots of missing declarations to the types of @babel/types,
including many utility functions useful to plugin authors.

With the typescript types, I tried to make them as useful as possible
for control flow analysis / inference, but Flow doesn't seem to
support overloads and I'm not as familiar with it anyway so it has
simpler types.

### DIFF
--- a/packages/babel-types/scripts/generators/typescript.js
+++ b/packages/babel-types/scripts/generators/typescript.js
@@ -62,6 +62,13 @@ for (const type in t.NODE_FIELDS) {
 
   fieldNames.forEach(fieldName => {
     const field = fields[fieldName];
+    // Future / annoying TODO:
+    // MemberExpression.property, ObjectProperty.key and ObjectMethod.key need special cases; either:
+    // - convert the declaration to chain() like ClassProperty.key and ClassMethod.key,
+    // - declare an alias type for valid keys, detect the case and reuse it here,
+    // - declare a disjoint union with, for example, ObjectPropertyBase,
+    //   ObjectPropertyLiteralKey and ObjectPropertyComputedKey, and declare ObjectProperty
+    //   as "ObjectPropertyBase & (ObjectPropertyLiteralKey | ObjectPropertyComputedKey)"
     let typeAnnotation = stringifyValidator(field.validate, "");
 
     if (isNullable(field) && !hasDefault(field)) {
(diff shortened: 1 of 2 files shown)


packages/babel-types/scripts/generators/typescript.js
```diff
@@ -109,25 +116,113 @@ for (const type in t.NODE_FIELDS) {
...
+    `export function is${typeName}(node: object | null | undefined, opts?: object | null): ${result};`,
+    // TypeScript 3.7: https://github.com/microsoft/TypeScript/pull/32695 will allow assert declarations
+    // eslint-disable-next-line max-len
+    `// export function assert${typeName}(node: object | null | undefined, opts?: object | null): asserts ${
+      result === "boolean" ? "node" : result
+    };`
+  );
 }
 
 lines.push(
```
Jessidhia: I think this whole thing would be much nicer to modify if there was, say, a `typescript.template.d.ts` file that is complete enough to be typechecked and with placeholder comments / types for where the generated types should get injected; the generator script then just needs to slurp the template and edit it.

Jessidhia:
CircleCI's build failed.... because it ran out of memory? o_O

nicolo-ribaudo:
Yeah it's flaky

packages/babel-types/scripts/generators/typescript.js
```diff
@@ -109,25 +116,113 @@ for (const type in t.NODE_FIELDS) {
...
-  lines.push(decl);
+for (const typeName of t.TYPES) {
+  const result =
+    t.NODE_FIELDS[typeName] || t.FLIPPED_ALIAS_KEYS[typeName]
+      ? `node is ${typeName}`
+      : "boolean";
+
+  lines.push(
+    `export function is${typeName}(node: object | null | undefined, opts?: object | null): ${result};`,
+    // TypeScript 3.7: https://github.com/microsoft/TypeScript/pull/32695 will allow assert declarations
```
Jessidhia: We can use [typesVersions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions) to publish typings for typescript <3.7 and typescript >=3.7 at the same time, but we'll need to generate types twice.

I want to refactor the typescript generation to use .d.ts files as templates (to make it much easier to get editor feedback while modifying them) so I can try to do that at the same time.

PR #10506, opened 2019-09-27
Use `make -j` for parallel build

JLHwung:

This PR introduces general improvements to current build scripts. The following jobs have seen significant time reduce

| Travis Job | Control | Experiment |
| --- | --- | --- |
| lint | [4min 57s](https://travis-ci.com/babel/babel/jobs/239869697) | [3min 14s](https://travis-ci.com/JLHwung/babel/jobs/239949154) |
| parser-flow-test | [3min 38s](https://travis-ci.com/babel/babel/jobs/239869699) | [2min 28s](https://travis-ci.com/JLHwung/babel/jobs/239949161) |
| parser-test262-test | [3min 48s](https://travis-ci.com/babel/babel/jobs/239869700) | [2min 51s](https://travis-ci.com/JLHwung/babel/jobs/239949162) |

The following improvements are implemented:

- Provide `gulp build-babel-types` for `make lint`, so that `lint` does not have to build all babel packages.
- Remove `merge-stream` by refactoring `buildBabel`
- Introduce `make code-quality`: an umbrella rule (with parallelism) of `make lint && make flow`.
- Skip `make build-standalone` and `make build-preset-env-standalone` in `lint`, `parser-flow-test` and `parser-test262-test` jobs.
- Parallelize `make bootstrap-only` and `make bootstrap-flow` as they requests different services and writes to different destinations.
- Parallelize `make bootstrap-only` and `make bootstrap-test262` for the same reason above
- Parallelize `make build-standalone` and `make build-preset-env-standalone` as they writes to different destinations.
- Replace nested `make` call to `$(MAKE)` variable

Note that now the lint job will run first since it is less expensive than unit testing.

Other thoughts: I did a rough measure on travis: it will take 2 min for machine setup + yarn install + lerna bootstrap, even with yarn cache. Could we merge `lint`, `parser-flow-test` and `parser-test262-test` into one job so that we can remove machine setup overhead?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,7 @@ matrix:
       env: JOB=lint
   # We test the latest version on circleci
     - node_js: "11"
-    - node_js: "10"
-  # Move `windows` build to be the fourth since it is slow
+  # Move `windows` build to be the third since it is slow
     - os: windows
       node_js: "node"
       env:
@@ -38,8 +37,9 @@ matrix:
       # https://travis-ci.community/t/build-doesnt-finish-after-completing-tests/288/9
       - YARN_GPG=no
   # Continue node_js matrix
-    - node_js: "8"
     - node_js: "6"
+    - node_js: "10"
+    - node_js: "8"
     - node_js: "node"
       env: JOB=babel-parser-flow-tests
     - node_js: "node"
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,6 @@ cache:
 
 os: linux
 
-node_js:
-  # We test the latest version on circleci
-  - "11"
-  - "10"
-  - "8"
-  - "6"
-
 env:
   global:
     - PATH=$HOME/.yarn/bin:$PATH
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -6,48 +6,40 @@ cache:
...
 
 matrix:
   fast_finish: true
   include:
+    - node_js: "node"
+      env: JOB=lint
+  # We test the latest version on circleci
+    - node_js: "11"
+    - node_js: "10"
+  # Move `windows` build to be the fourth since it is slow
```
nicolo-ribaudo: Also node 6 is slow. What do you think about this order?

- Lint
- Node 11 (fast, for quick feedback)
- Windows
- Node 6
- Node 8
- Node 10
- Flow
- Test262

JLHwung: We can definitely give it a try. As long as `Windows` or `Node 6` is not the last I think they would not make significant difference.

How many concurrent jobs do we have on circle ci? If we have two concurrent jobs we can move windows testing to circle ci.

nicolo-ribaudo: 4 I think

Issue #9888, opened 2019-04-22
Support asynchronous configuration


guybedford:
I couldn't find an existing tracking issue for this feature, please mark as a duplicate if I've missed anything.

This is something that would be required for jspm support in Babel CLI, where we'd want to use dynamic `import()` to reference plugins directly, and so would need some kind of promise support in the configuration.

Currently an asynchronous configuration gives the following error:

```
Error: You appear to be using an async configuration, which your current version of Babel does not support. We may add support for this in the future, but if you're on the most recent version of @babel/core and still seeing this error, then you'll need to synchronously return your config.
```

And an asynchronous plugin gives the following error:

```
You appear to be using an async plugin, which your current version of Babel
does not support.If you're using a published plugin, you may need to upgrade your @babel/core version.
```

Either or both of these features would be really great to support.

The async plugin case effectively provides parity with top-level await workflows in future too. Happy to discuss further as well.

nicolo-ribaudo:
While developing Babel 7 beta, we added that errors to be able to introduce async config/plugins during the Babel 7 lifetime without breaking changes.

The biggest problem is that currently Babel can be called synchronously (`var out = babel.transform(code, options)`), and there is no way to make it work with async config/plugins.

guybedford:
Thanks for clarifying. Ok so how can we do this safely? Just throw when running under the sync transform only?

nicolo-ribaudo:
> Just throw when running under the sync transform only?

Either that, or https://github.com/abbr/deasync.

I have done some research to understand how hard would be for the ecosystem to support async Babel configurations:
1. `babel-loader` will support it without any change, it already uses Babel's async api.
2. `rollup-plugin-babel` can be easily updated: it call's Babel sync `transform` inside an async function
3. ESLint doesn't support async parsers: this is a problem for `babel-eslint`

loganfsmyth:
When I added the error my intention was to eventually
1. Throw an exception if a plugin/config returns a promise in a context where Babel is used synchronously
2. Expose an API on the first function parameter so that plugins could check which whether promises would be allowed, and change their behavior depending on that. That way a plugin could be async ideally and fall back to a sync implementation if it wanted or something.

That way if a plugin _really_ wanted to use `deasync` or something, it could, but it wouldn't be something we'd do in our own plugins, or necessarily recommend, since that module really is ugly.

[...]

PR #10507, opened 2019-09-27
Prepare @babel/core for asynchronicity

nicolo-ribaudo:
| Fixed Issues?            | First step for #9888
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

In this PR, I am refactoring `@babel/core` to use [`gensync`](https://github.com/loganfsmyth/gensync/). It's a tool by @loganfsmyth which allows defining functions as generators which can then be run either synchronously or asynchronously, pushinc control to the edges of the program.

For example, it allows doing this:

```js
const fs = require("fs-for-gensync");

const readJSON = gensync(function* (path) {
  const contents = yield* fs.readFile(path, "utf8"); // Sync or async? It doesn't matter
  return JSON.parse(contents);
});

readJson.sync("file.json"); // internally calls fs.readFileSync(path, encoding)
readJson.async("file.json"); // internally calls fs.readFile(path, encoding, cb)
```

By doing so, we will be able to easily allow some parts of Babel to be asynchronous. I identified 4 possible parts:
- `.pre` functions of plugins
- `.post` functions of plugins
- Configuration functions (i.e. in `babel.config.json`)
- Configuration file loading

The latter is necessary because, starting from node v12.11.0, is a user has `"type": "module"` in their `package.json` we won't be able to load `babel.config.js` using `require` but we'll need to use `import()`. (https://github.com/nodejs/modules/issues/389)

Since JavaScript doesn't have native algebraic effects<sup>[1]</sup>, which would have allowed me to use `yield` only in those 4 places, we have to use `yield*` in the whole call stack. The last two parts are deep inside `@babel/core`, and that's why `yield*` spread almost everywhere.

This PR already makes I/O actions asynchronous when using the async functions.

This PR aligns the API of all the functions which could interact with the disk: now `parse`, `transform`, `transformFile`, `transformFromAst`, `loadPartialConfig` and `loadOptions` all have three version:
- `fooSync`, which synchronously returns the result;
- `fooAsync`, which returns a promise;
- `foo`, which accepts a callback. From compatibility reasons, some of these behave synchronously when the callback is not provided.

TODO:
- [x] https://github.com/loganfsmyth/gensync/pull/1
- [x] Unify caching logic for generators and normal functions
- [x] Test performance differences

<sup>[1]</sup> I learned exactly what they are while preparing this PR

---

**Performance differences**: Not significant.
In order to test this PR, I compiled every file inside `packages/*/src` (405 files) both synchronously and asynchronously, using the following script. It's always run using `node demo.js --sync && node demo.js --sync && node demo.js --sync && node demo.js --sync && node demo.js --async && node demo.js --async && node demo.js --async && node demo.js --async`.

<details>

```js
"use strict";

const load = name => require(`./packages/babel-${name}`);

const babel = load("core");
const fs = require("fs");

function walkSync(dir, filelist) {
  const files = fs.readdirSync(dir);
  filelist = filelist || [];
  files.forEach(file => {
    if (fs.statSync(dir + "/" + file).isDirectory()) {
      filelist = walkSync(dir + "/" + file, filelist);
    } else {
      filelist.push(dir + "/" + file);
    }
  });
  return filelist;
}

const files = [];

const packages = fs.readdirSync("packages");
for (const name of packages) {
  if (name === "README.md" || name.startsWith("babel-runtime")) continue;
  files.push.apply(files, walkSync("packages/" + name + "/src"));
}


const start = process.hrtime();
global.results = {};

if (process.argv.includes("--sync")) {
  process.env.SYNC_OR_ASYNC = "sync";
  for (const file of files) {
    global.results[file] = babel.transformFileSync(file);
  }
  console.log("Sync: " + process.hrtime(start));
} else {
  process.env.SYNC_OR_ASYNC = "async";
  let i = 0;
  for (const file of files) {
    i++;
    babel.transformFile(file, (err, out) => {
      if (err) {
        console.log(err);
        process.exit(1);
      }
      global.results[file] = out;
  
      if (--i === 0) console.log("Async: " + process.hrtime(start));
    });
  }
}
```

</details>

<details>
<summary>Results</summary>

`NODE_ENV` is always `"test"`

- `master` branch, with `api.cache.never()`:
  ```
  Sync: 5,886560734
  Sync: 5,885043757
  Sync: 5,867874524
  Sync: 5,803315282
  Async: 5,949791895
  Async: 5,789683786
  Async: 5,806937770
  Async: 5,853401118
  ```
- `core-async` branch, with `api.cache.never()`:
  ```
  Sync: 5,960720491
  Sync: 5,894450713
  Sync: 5,891051295
  Sync: 5,946484727
  Async: 5,913818398
  Async: 5,929368030
  Async: 6,11599499
  Async: 6,58814871
  ```

- `master` branch, with `api.cache.using(() => process.env.SYNC_OR_ASYNC)`:
  ```
  Sync: 5,883534231
  Sync: 4,840278891
  Sync: 4,994791415
  Sync: 4,897095093
  Async: 4,874467892
  Async: 4,824938870
  Async: 4,826639093
  Async: 4,851304427
  ```
- `core-async` branch, with `api.cache.using(() => process.env.SYNC_OR_ASYNC)`:
  ```
  Sync: 4,866956272
  Sync: 4,820806372
  Sync: 4,857302419
  Sync: 4,783584388
  Async: 5,869067030
  Async: 5,931280575
  Async: 6,17525481
  Async: 5,993042102
  ```
</details>

**Conclusion**: When running synchronously, the performance is almost the same. On the other hand, when running asynchronously this branch is consistently a bit slower than `master`. I think that it's because `gensync` waits one tick for each yielded value, the same as `await` behaves in async functions.

### DIFF
--- a/packages/babel-core/package.json
+++ b/packages/babel-core/package.json
@@ -49,6 +49,7 @@
     "@babel/types": "^7.7.4",
     "convert-source-map": "^1.7.0",
     "debug": "^4.1.0",
+    "gensync": "^1.0.0-beta.1",
     "json5": "^2.1.0",
     "lodash": "^4.17.13",
     "resolve": "^1.3.2",
--- a/packages/babel-core/src/config/caching.js
+++ b/packages/babel-core/src/config/caching.js
@@ -1,5 +1,17 @@
 // @flow
 
+import gensync, { type Handler } from "gensync";
+import {
+  maybeAsync,
+  isAsync,
+  onFirstPause,
+  waitFor,
+  isThenable,
+} from "../gensync-utils/async";
+import { isIterableIterator } from "./util";
+
+export type { CacheConfigurator };
+
 export type SimpleCacheConfigurator = SimpleCacheConfiguratorFn &
   SimpleCacheConfiguratorObj;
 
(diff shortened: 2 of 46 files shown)


nicolo-ribaudo:
https://github.com/nicolo-ribaudo/babel/commit/3f640b2d8c1649d73904f2b7d556dcc238a0b4db (not included in this PR) shows how with the new architecture it is straightforward to allow some parts of Babel to be async.

packages/babel-core/src/config/full.js
```diff
@@ -45,10 +47,10 @@ type SimpleContext = {
   caller: CallerMetadata | void,
 };
 
-export default function loadFullConfig(
+export default gensync<[any], ResolvedConfig | null>(function* loadFullConfig(
```
loganfsmyth: Why the `any` instead of `mixed`?

packages/babel-core/src/config/caching-a.js
```diff
@@ -0,0 +1,108 @@
...
+
+    if (cachedValue) {
+      for (const { value, valid } of cachedValue) {
+        if (valid(data)) return value;
+      }
+    }
+
+    const cache = new CacheConfigurator(data);
+
+    const value = yield* handler(arg, cache);
```
loganfsmyth: One big complexity here is that, if this is async, you can end up with multiple calls to `handler` running in parallel. In my opinion, we should only ever allow one call to `handler` at a time for a given `arg`.

packages/babel-core/src/config/caching-a.js
```diff
@@ -0,0 +1,108 @@
...
+  handler: (ArgT, CacheConfigurator<SideChannel>) => Handler<ResultT>,
+): Gensync<[ArgT, SideChannel], ResultT> {
+  return gensync(function* cachedFunction(arg: ArgT, data: SideChannel) {
+    let cachedValue: CacheEntry<ResultT, SideChannel> | void = callCache.get(
+      arg,
+    );
+
+    if (cachedValue) {
+      for (const { value, valid } of cachedValue) {
+        if (valid(data)) return value;
```
loganfsmyth: Do you think we should consider allowing this to be async too? We're currently stuck with the one `statSync` because it is sync-only. We shouldn't expose gensync to consumers, but we could repeat the `.then` check here too and allow returning a promise, if we wanted.

Not needed now and easy to expand later, but curious for your thoughts.

loganfsmyth:
This is exactly the usage I had in mind when designing `gensync`, so I'm happy to hear that it seems to have worked pretty smoothly overall.

packages/babel-core/src/config/full.js
nicolo-ribaudo: It's because I get errors about objects not being compatible with `mixed`:
https://github.com/babel/babel/blob/0da4f7ef50b5299d67d7bd3a10fb66453f214dc3/packages/babel-core/src/transformation/block-hoist-plugin.js#L14-L18
https://github.com/babel/babel/blob/0da4f7ef50b5299d67d7bd3a10fb66453f214dc3/packages/babel-core/src/config/index.js#L18-L19

packages/babel-core/src/config/caching-a.js
nicolo-ribaudo: I'm not sure about how they could run in parallel. If it is async, `handler`'s effects will be yielded and no other code will be run until they are completed, including other `handler` calls.

packages/babel-core/src/config/caching-a.js
nicolo-ribaudo: I didn't change it because I still have to finish refactoring the caching logic, but I will do it :+1:

nicolo-ribaudo:
I'm also extending `gensync` to reach the point where we call plugin functions, since we already have a check to throw if they are async.

packages/babel-core/src/config/caching-a.js
loganfsmyth: For instance, if you do
```
var doThing = makeWeakCache(function*(arg) {
  yield* delay(100);
  return {};
});

var obj = {};
const [r1, r2] = yield* gensync.all[doThing(obj), doThing(obj)]);

console.log(r1 === r2);
```
would be `true` in sync code and `false` in async code with your current implementation because the first would run then populate the cache with the `r1` object, but then the second one would finish and replace the `callCache` data with `r2`. So really the second call to see that there is already a call in-progress and wait for it to finish before checking if the cache entry matches.

packages/babel-core/src/config/caching-a.js
loganfsmyth: Cool!

packages/babel-core/src/config/full.js
loganfsmyth: Huh, seems like a Flow bug, but fair enough.

packages/babel-core/package.json
```diff
@@ -42,6 +42,7 @@
     "@babel/types": "^7.6.0",
     "convert-source-map": "^1.1.0",
     "debug": "^4.1.0",
+    "gensync": "1.0.0-beta.0",
```
loganfsmyth: Don't forget to bump this!

packages/babel-core/src/config/full.js
```diff
@@ -12,7 +15,8 @@ import {
 } from "./config-chain";
 import type { UnloadedDescriptor } from "./config-descriptors";
 import traverse from "@babel/traverse";
-import { makeWeakCache, type CacheConfigurator } from "./caching";
+import { makeWeakCache, type CacheConfigurator } from "./caching-a";
+import { makeWeakCache as makeWeakCacheSyncSync } from "./caching";
```
loganfsmyth: `makeWeakCacheSyncSync` => `makeWeakCacheSync`?

packages/babel-core/src/gensync-utils/forward.js
```diff
@@ -0,0 +1,32 @@
...
+    adapted: (...args: ActionArgs) => MaybePromise<ActionReturn>,
+  ) => MaybePromise<Return>,
+): Handler<Return> {
+  return isAsync(kind => {
+    const adapted = cachedGensync<ActionArgs, ActionReturn>(action)[kind];
+    return cb(adapted);
+  });
+}
+
+export const isAsync = gensync<[any], any>({
```
loganfsmyth: Is there a way we can avoid `any`? I tried really hard to never use `any` in the core codebase, which is why I keep mentioning it.

packages/babel-core/src/gensync-utils/forward.js
```diff
@@ -0,0 +1,32 @@
+// @flow
+
+import gensync, { type Handler } from "gensync";
+
+type MaybePromise<T> = T | Promise<T>;
+
+export default function forward<ActionArgs: mixed[], ActionReturn, Return>(
```
loganfsmyth: It would be good to have some comments in these, since I don't think it's very clear what their expected usecases would be.

packages/babel-core/src/gensync-utils/maybe-async.js
```diff
@@ -0,0 +1,29 @@
...
+    },
+    async(...args) {
+      return Promise.resolve(fn.apply(this, args));
+    },
+  });
+}
+
+export function isThenable(val: mixed): boolean %checks {
+  return (
+    /*:: val instanceof Promise && */
```
loganfsmyth: Tricky!

packages/babel-core/src/config/caching-a.js
loganfsmyth: Was this explanation enough to make my concern clearer?

nicolo-ribaudo:
@loganfsmyth please don't review yet because I'm still iterating 😛

packages/babel-core/src/gensync-utils/forward.js
nicolo-ribaudo: This can definitely be removed, I don't know why I used it

packages/babel-core/src/config/full.js
nicolo-ribaudo: I still have to refactor it, but there is a difference: `Sync` allows an async cache invalidation function and a sync cached function, while with `SyncSync` both are sinchronous.

packages/babel-core/src/config/caching-a.js
nicolo-ribaudo: Yes sorry, I just didn't have time to update it! There is a reason if this PR is marked as "draft" 😛

packages/babel-core/src/config/caching-a.js
nicolo-ribaudo: # TODO

packages/babel-core/src/config/caching-a.js
nicolo-ribaudo: # TODO

nicolo-ribaudo: @loganfsmyth I was implementing this, and I'm not sure that it is something that we want, since it can cause a few problems:
1. If a function is called recursively (like a plugin loading another plugin via `inherits`), it will cause a deadlock: the outer function needs the inner one's return value, but the inner one won't run until the outer one is finished.
2. It will make async calls to Babel slower. Suppose that we have this code:
    ```js
    var maybeDelay = gensync({
      sync: () => {},
      async: t => delay(t),
    });

    var doThing = gensync(makeWeakCache(function*(arg) {
      yield* maybeDelay(100);
      return {};
    }));

	doThing.async();
    await delay(50);
    doThins.sync();
    ```

    here `doThins.sync()` runs while `doThing.async` is paused by the `delay` call. If we can't call `doThing.sync` until `doThing.async` is finished, the only option we have is to throw an error, because it can't wait.

    `maybeDelay` might seem like a function crafted only to create this example, without real-world applications, but it is exactly like any `fs.*` operation.

I think that randomly throwing errors caused by race conditions is a too hight cost for to be introduced only for performance reasons. I think that we have two options:
1. Leave is as-is, allowing two different async calls to happen in parallel
2. Always invalidate the cache when switching between sync/async. I already had to do it when interacting with an either sync or async `cache.invalidate` call, but it shouldn't be observable.

loganfsmyth: The big issue for me is that I could really see this being a noticable perf hit for something that starts compiling a bunch of files in parallel, like `@babel/cli`. If you do something like

```
var results = Promise.all(files.map(filename => babel.transformFileAsync(filename)));
```
then you'll likely end up calling every plugin and preset function for every single file, because config resolving will all be kicked off in parallel and each file's process will kick off an operation that then won't get the cached valid of the previous file compilation get, because the operation will have started before the previous one finished. I haven't actually tested that though, so if you want to verify that could help the discussion. Maybe I'm missing something.

> If a function is called recursively (like a plugin loading another plugin via inherits), it will cause a deadlock: the outer function needs the inner one's return value, but the inner one won't run until the outer one is finished.

We should probably disallow that anyway, if we don't already, since it would be an unresolvable problem in sync code too. I agree that recursion could be a concern, but I'm not sure it's one that wouldn't already be an issue in sync cases in general too.

> If we can't call doThing.sync until doThing.async is finished, the only option we have is to throw an error, because it can't wait.

I think it's probable find to have entirely separate caching for sync and async calls to avoid this conflict. In fact we likely explicitly want that, because plugins could return different results depending on whether the system is sync or async.

packages/babel-core/src/config/caching.js
```diff
@@ -14,91 +27,253 @@ type SimpleCacheConfiguratorObj = {
...
+ *       a. Wait for that lock to be released
+ *       b. RETURN the value associated with that lock
+ *   4. Acquire to locks: ConfigLock and FinishLock
+ *   5. Start executing the cached function
+ *   6. When the cache if configured, release ConfigLock
+ *   7. Wait for the function to finish executing
+ *   8. Release FinishLock
+ *   9. Send the function result to anyone waiting on FinishLock
+ *   10. Store the result in the cache
+ *   11. RETURN the result
```
nicolo-ribaudo: @loganfsmyth What do you think abuot this algorithm, to solve the concurrency issue?
It is sub optimal in cases like the following one (it will take 2s instead of 1s), but I don't think that they are common:

```js
const fn = makeStrongCache(function* (arg) {
  yield* wait(1000);
  cache.never();
}).async;

fn(); fn();
```

I initially tried to force `cache.*` to be called synchronously (before any `yield`), but then found out that it doesn't even work for our internal code: https://github.com/babel/babel/pull/10507/files#diff-83f25d98662c0f8e8ad7cc7d1ed06f8eR32

nicolo-ribaudo: I tested the performance of this change, and there is a problem.
I have compiled every src file of this repo four times (405 x 4 files) and there are the results:

|           | **`cache.forever()`** | **`cache.never()`** |
|----------:|:---------------------:|:-------------------:|
| **async** |     15.372567052 s    |   470.434076412 s   |
|  **sync** |     15.862861386 s    |    19.743430078 s   |

As you can see, there is a *big* problem with `cache.never()` + async.
It is related to re-compiling the same file. If I compile every file once instead of four times, it doesn't have this big difference:

|           | **`cache.never()`** |
|----------:|:-------------------:|
| **async** |    10.391877456 s   |
|  **sync** |    6.149023702 s    |

nicolo-ribaudo: @loganfsmyth We can solve this big performance problem if we disallow the cache to be configured asynchronously. Synchronously configuring the cache with an async validator would still be allowed:

```js
// waitFor transforms a "Promise<T> | T" into a "Gensync<T>"

makeWeakCache(function* (arg) {
  // Allowed, because cache.using is called synchronously

  const val = yield* waitFor(
    cache.using(async () => {
      await wait(100);
      return process.env.FOO;
    })
  );
});


makeWeakCache(function* (arg) {
  // Disallowed, because cache.using is called asynchronously

  yield* waitFor(wait(100));
  const val = cache.using(() => process.env.FOO);
});
```

In the future, this change will be observable to our users because they won't be able to use `await` before configuring the cache.

In order to implement this behaviour I had to modify `gensync`'s internals, because I couldn't find a way "from the outside" of knowing if a generator has been suspended. I will open a PR soon because I edited it in my `node_modules` and I fear to loose those changes, but if you have a better solution feel free to propose it!

nicolo-ribaudo: These are the new numbers:

| **#** |    **async**    | **sync**       | **sync `master`** |
|:-----:|:---------------:|:--------------:|:-----------------:|
|  100  |  2.219376366 s  |  2.256803449 s | 2.143184428 s     |
|  200  |  3.714403922 s  |  3.641993385 s | 3.591130842 s     |
|  400  |  6.338499966 s  |  6.638144645 s | 6.299714871 s     |
|  800  |  11.772867713 s | 12.198001927 s | 11.539516664 s    |
|  1600 |  21.811404623 s | 22.632490371 s | 21.436522896 s    |

packages/babel-core/src/config/caching-a.js
Jessidhia: Sounds like you need a mutex.

nicolo-ribaudo:
Ready for review!

nicolo-ribaudo:
Note that node 13.0.0 will be released soon (https://github.com/nodejs/node/pull/29504), and v13.1.0 will have unflagged modules (https://twitter.com/jasnell/status/1184265267951493120?s=20), around the beginning of November.

nicolo-ribaudo:
I need help to fix the windows build. https://github.com/babel/babel/pull/10507/commits/53c74e888b51f2fef01c24919cc9a536a49f9cd3 is the last "real" commit of this PR; the following commits are just failed attempts at fixing it.

nicolo-ribaudo:
Ok, after many failed attempts I managed to fix the windows build :tada:

PR #10511, opened 2019-09-30
Add filename to transform error

### LINKED ISSUE #10502: ICE when compiling a *.d.ts file
## Bug Report

**Current Behavior**
An exception is thrown when attempting to transform a *.d.ts file here: https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-typescript/src/index.js#L142
```
TypeError: Cannot read property 'name' of null
  at PluginPass.Program (node_modules/@babel/plugin-transform-typescript/lib/index.js:167:57)
  at newFn (node_modules/@babel/traverse/lib/visitors.js:195:21)
  at NodePath._call (node_modules/@babel/traverse/lib/path/context.js:53:20)
  at NodePath.call (node_modules/@babel/traverse/lib/path/context.js:40:17)
  at NodePath.visit (node_modules/@babel/traverse/lib/path/context.js:88:12)
  at TraversalContext.visitQueue (node_modules/@babel/traverse/lib/context.js:120:16)
  at TraversalContext.visitSingle (node_modules/@babel/traverse/lib/context.js:92:19)
  at TraversalContext.visit (node_modules/@babel/traverse/lib/context.js:148:19)
  at Function.traverse.node (node_modules/@babel/traverse/lib/index.js:96:17)
  at traverse (node_modules/@babel/traverse/lib/index.js:78:12)
```

**Input Code**

```typescript
export default function<T>(x: T): T;
```

**Expected behavior/code**

Babel should not process *.d.ts files, or should give a better error.

```js
{
  "presets": [["@babel/preset-env", { "targets": { "node": 10 } }], "@babel/preset-typescript"],
  "plugins": ["babel-plugin-macros"]
}
```

**Environment**
- Node 10 / Yarn 1.17.3
- OS: MacOS 10.14.6
- Monorepo: yes: yarn workspace
- How you are using Babel: cli: `./node_modules/.bin/babel -d lib -x '.ts' src`

**Possible Solution**

When I add this to my config, it works:
```
{
  ...
  "ignore": ["src/dbg.macro.d.ts"]
}

```

### LINKED ISSUE #10505: Babel should output the current file being transformed if there's an unexpected error
## Bug Report

**Current Behavior**
After reading https://github.com/babel/babel/issues/10502, I noticed that an important issue was that the error messages from `@babel/traverse` and thus `@babel/core` don't include the file being processed which would of made debugging much easier. I assumed this was the case already since we do output the file for an expected error like a syntax error, but there are a lot like this?

```
SyntaxError: /home/nicolo/Documenti/dev/babel/babel/demo.js: Legacy octal literals are not allowed in strict mode (11:0)

   9 | const parser = load("parser");
  10 | const generate = load("generator").default;
> 11 | 00 0
     | ^
```

But when there's an internal error it doesn't:

```
TypeError: Cannot read property 'name' of null
    at PluginPass.Program (/Users/henryzhu/dev/a/node_modules/@babel/plugin-transform-typescript/lib/index.js:167:57)
    at newFn (/Users/henryzhu/dev/a/node_modules/@babel/traverse/lib/visitors.js:193:21)
```

**Input ~~Code~~**

This kind of error doesn't happen from a parsing error so it's more around how things are configured. There's probably an easier way to throw this kind of error via other bug reports but for now, I would reproduce the error from #10502? Another option is to just throw an error anywhere in `@babel/core` or some plugin/preset.

```sh
yarn add @babel/cli @babel/core @babel/preset-typescript
```

//.babelrc
```json
{
}
```

make a `a.d.ts` file

```ts
export default function<T>(x: T): T;
```

run `npx babel a.d.ts`

**Expected behavior/code**

We'd need to add the filename (and maybe other relevant information could be in a debug mode like https://github.com/babel/babel/issues/2960) to the error message somehow like for our `SyntaxError`. 

**Possible Solution**

We would need to wrap our main transform() call in `@babel/core` with a try/catch to modify the error message.

https://github.com/babel/babel/blob/b459f6a0a4fa5d45814158eb271e5aed66c001d7/packages/babel-core/src/transformation/index.js#L58

```
try {
  transformFile(file, config.passes);
} catch (err) {
  // throw new Error(err); and add the filename
}
```

JLHwung:
| Fixed Issues?            | Fixes #10502, fixes #10505 

Previous art: In the parsing stage babel-core will add filename to the error message

https://github.com/babel/babel/blob/94fcabc4e3bed021e1a05d6c7a9b0a52261b2018/packages/babel-core/src/transformation/normalize-file.js#L145

In this PR we add `filename` to the error message to both transforming phrase and generating phrase. It should help developer debug which file actually causes babel transformer/generator to throw.

### DIFF
--- a/packages/babel-core/src/transformation/index.js
+++ b/packages/babel-core/src/transformation/index.js
@@ -55,19 +55,28 @@ export function runSync(
     ast,
   );
 
+  const opts = file.opts;
   try {
     transformFile(file, config.passes);
   } catch (e) {
-    e.message = `${file.opts.filename ?? "unknown"}: ${e.message}`;
+    e.message = `${opts.filename ?? "unknown"}: ${e.message}`;
     if (!e.code) {
       e.code = "BABEL_TRANSFORM_ERROR";
     }
     throw e;
   }
 
-  const opts = file.opts;
-  const { outputCode, outputMap } =
-    opts.code !== false ? generateCode(config.passes, file) : {};
+  let outputCode, outputMap;
+  try {
+    ({ outputCode, outputMap } =
+      opts.code !== false ? generateCode(config.passes, file) : {});
+  } catch (e) {
+    e.message = `${opts.filename ?? "unknown"}: ${e.message}`;
+    if (!e.code) {
+      e.code = "BABEL_GENERATE_ERROR";
+    }
+    throw e;
+  }
 
   return {
     metadata: file.metadata,
--- a/packages/babel-core/src/transformation/index.js
+++ b/packages/babel-core/src/transformation/index.js
@@ -55,7 +55,15 @@ export function runSync(
     ast,
   );
 
-  transformFile(file, config.passes);
+  try {
+    transformFile(file, config.passes);
+  } catch (e) {
+    e.message = `${file.opts.filename ?? "unknown"}: ${e.message}`;
+    if (!e.code) {
+      e.code = "BABEL_TRANSFORM_ERROR";
+    }
+    throw e;
+  }
 
   const opts = file.opts;
   const { outputCode, outputMap } =
--- a/packages/babel-core/src/transformation/index.js
+++ b/packages/babel-core/src/transformation/index.js
@@ -68,8 +68,9 @@ export function runSync(
 
   let outputCode, outputMap;
   try {
-    ({ outputCode, outputMap } =
-      opts.code !== false ? generateCode(config.passes, file) : {});
+    if (opts.code !== false) {
+      ({ outputCode, outputMap } = generateCode(config.passes, file));
+    }
   } catch (e) {
     e.message = `${opts.filename ?? "unknown"}: ${e.message}`;
     if (!e.code) {
--- a/packages/babel-core/src/transformation/file/file.js
+++ b/packages/babel-core/src/transformation/file/file.js
@@ -259,7 +259,7 @@ export default class File {
   ): Error {
     let loc = node && (node.loc || node._loc);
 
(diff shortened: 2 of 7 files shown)


packages/babel-core/src/transformation/index.js
```diff
@@ -55,11 +55,28 @@ export function runSync(
...
+    if (!e.code) {
+      e.code = "BABEL_TRANSFORM_ERROR";
+    }
+    throw e;
+  }
+
+  let outputCode, outputMap;
+  try {
+    ({ outputCode, outputMap } =
+      opts.code !== false ? generateCode(config.passes, file) : {});
```
nicolo-ribaudo: Nit: since we don't use an expression anymore, we could rewrite this as
```js
if (opts.code !== false) {
  ({ outputCode, outputMap } = generateCode(config.passes, file))
}
```

I find it a bit more readable, but it's up to you.

JLHwung: Your suggestion is better. I am too lazy to refactor code styles. :P

JLHwung:
> maybe later we can think about normalizing error messages or moving into util

I think it is still slightly early to extract the error messages processing into separate utils. I would suggest refactoring when we are working on the debug mode error message.

jrop:
@JLHwung @hzoo Thanks guys!  I definitely was not expecting a fix this quick.  Thank you much!

hzoo:
Same haha! I think not being to know the file where the error is happening to due an internal error (not syntax error) would be pretty bad on our part so best to make filing an issue easy! 🙌

PR #10515, opened 2019-10-02
Revert "Try different name for used-by feature again [skip ci]"

existentialism:
Reverts babel/babel#10513


hzoo:
Don't need this since the setting should cover it either way https://twitter.com/left_pad/status/1179121411270610945?s=20

PR #10516, opened 2019-10-02
Bump flow@0.108.0

existentialism:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-preset-env/src/utils.js
+++ b/packages/babel-preset-env/src/utils.js
@@ -62,7 +62,7 @@ export function findSuggestion(options: string[], option: string): string {
       return validOption;
     }
     return suggestion;
-  }, undefined);
+  }, "");
 }
 
 export function prettifyVersion(version: string) {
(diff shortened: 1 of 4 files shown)


packages/babel-preset-env/src/utils.js
```diff
@@ -62,7 +62,7 @@ export function findSuggestion(options: string[], option: string): string {
       return validOption;
     }
     return suggestion;
-  }, undefined);
+  }, "");
```
existentialism: `Cannot return options.reduce(...) because undefined [1] is incompatible with string [2].`

packages/babel-preset-env/test/utils.spec.js
```diff
@@ -56,6 +56,8 @@ describe("utils", () => {
       const options = ["one", "two", "three"];
       expect(findSuggestion(options, "onr")).toEqual("one");
       expect(findSuggestion(options, "tree")).toEqual("three");
+      expect(findSuggestion(options, "")).toEqual("one");
+      expect(findSuggestion(options, "xxx")).toEqual("one");
```
nicolo-ribaudo: Is this a behaviour change?

existentialism: Assuming a set of options is passed, it would always return a string.

PR #10519, opened 2019-10-02
Fix generator missing parens around an arrow returning function type

### LINKED ISSUE #10517: Parsing and generating flow type moves parentheses
## Bug Report

**Input Code**

`packages/babel-generator/test/fixtures/types/ArrowFunctionExpression2/input.js`:
```js
const pair = <T>(a: T): (T => T[]) => b => [a, b]
```
`packages/babel-generator/test/fixtures/types/ArrowFunctionExpression2/options.json`:
```json
{ "plugins": ["flow"] }
```

**Expected behavior/code**
The output should be the same as the input, with the parentheses around `(T => T[])`.

**Current Behavior**
Output code is:
```js
const pair = <T>(a: T): (T) => T[] => b => [a, b];
```
The parentheses have moved from `(T => T[])` to `(T) => T[]`.
Flow now gives a parse error on the output, where it worked for the original.

**Environment**
- Babel version(s):  master
- OS: Linux Mint 18.1
- Monorepo: yes
- How you are using Babel: ` env TEST_ONLY=babel-generator make test`

existentialism:
| Fixed Issues?            | Fixes #10517
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-generator/src/node/parentheses.js
+++ b/packages/babel-generator/src/node/parentheses.js
@@ -35,14 +35,22 @@ export function NullableTypeAnnotation(node: Object, parent: Object): boolean {
   return t.isArrayTypeAnnotation(parent);
 }
 
-export function FunctionTypeAnnotation(node: Object, parent: Object): boolean {
+export function FunctionTypeAnnotation(
+  node: Object,
+  parent: Object,
+  printStack: Array<Object>,
+): boolean {
   return (
     // (() => A) | (() => B)
     t.isUnionTypeAnnotation(parent) ||
     // (() => A) & (() => B)
     t.isIntersectionTypeAnnotation(parent) ||
     // (() => A)[]
-    t.isArrayTypeAnnotation(parent)
+    t.isArrayTypeAnnotation(parent) ||
+    // <T>(A: T): (T => T[]) => B => [A, B]
+    (t.isTypeAnnotation(parent) &&
+      // Check grandparent
+      t.isArrowFunctionExpression(printStack[printStack.length - 3]))
   );
 }
 
--- a/packages/babel-generator/test/fixtures/flow/arrow-functions/output.js
+++ b/packages/babel-generator/test/fixtures/flow/arrow-functions/output.js
@@ -11,3 +11,5 @@ const bar5 = (x): string => {};
 const bar6 = (x: number) => {};
 
 const bar7 = <T>(x) => {};
+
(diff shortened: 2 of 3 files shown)


packages/babel-generator/src/node/parentheses.js
```diff
@@ -35,14 +35,21 @@ export function NullableTypeAnnotation(node: Object, parent: Object): boolean {
...
     // (() => A) | (() => B)
     t.isUnionTypeAnnotation(parent) ||
     // (() => A) & (() => B)
     t.isIntersectionTypeAnnotation(parent) ||
     // (() => A)[]
-    t.isArrayTypeAnnotation(parent)
+    t.isArrayTypeAnnotation(parent) ||
+    // <T>(A: T): (T => T[]) => B => [A, B]
+    (t.isTypeAnnotation(parent) &&
+      t.isArrowFunctionExpression(printStack[printStack.length - 3]))
```
Jessidhia: What does `printStack[printStack.length - 3]`...... mean?

I assume this is related to Flow emit; the example syntax looks like invalid TypeScript.

Also, would the structure be the same if `<T>(A: T): // ...` (declaration signature) was written as `<T>(A: T) => /...` (expression signature)?

packages/babel-generator/test/fixtures/flow/arrow-functions/output.js
```diff
@@ -10,4 +10,6 @@ const bar5 = (x): string => {};
 
 const bar6 = (x: number) => {};
 
-const bar7 = <T>(x) => {};
\ No newline at end of file
+const bar7 = <T>(x) => {};
+
+const bar8 = <T>(x: T): ((T) => T[]) => y => [x, y];
```
Jessidhia: hmm, I see. Maybe the correct output should be the one in the code comment.

packages/babel-generator/src/node/parentheses.js
existentialism: Yeah, it's Flow emit related. I can add a comment, the printStack bit is just checking the grandparent (since these are only passed node, its parent, and the whole stack).

packages/babel-generator/test/fixtures/flow/arrow-functions/output.js
existentialism: I was look into dropping the parens around `(T)` later, initial experiment looked like it touched a bunch of other fixtures.

PR #10521, opened 2019-10-03
[parser] Enable "exportNamespaceFrom" by default

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | Yes (default change)
| Documentation PR Link    | 
| Any Dependency Changes?  |

It has been merged to the main spec: https://github.com/tc39/ecma262/pull/1174


packages/babel-parser/typings/babel-parser.d.ts
```diff
@@ -106,7 +106,7 @@ export type ParserPlugin =
     'dynamicImport' |
     'estree' |
     'exportDefaultFrom' |
-    'exportNamespaceFrom' |
+    'exportNamespaceFrom' | // deprecated
```
Jessidhia: I wonder if a `/** @deprecated */` annotation preceding the literal would work, or if it only works with enums

nicolo-ribaudo: I'll check 👍

nicolo-ribaudo:
No, it's otherwise people using new `@babel/core` versions with an older parser won't get the nice error.

PR #10529, opened 2019-10-07
Do not hoist jsx referencing a mutable binding

### LINKED ISSUE #10522: transform-react-constant-elements: using mutable variable doesn't lead to deoptimization
## Bug Report

**Current Behavior**
When a React component uses a `let` variable, hoisting of the returned React Element leads to incorrect behavior

**Input Code**
https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=DYUwLgBAZg9jEF4IHIAWJjBsg3AKDxAA8AHGAJ0gGMYA7AZ0jHIEsBzNkc-gQSrBZ16ANRYBDAEIBXMGDqIIACgCUiAHwQA3ngjQ4C5GzgATAEYBPELh0Ry4KeVoQAPPRJjaazbBgBfZwD0bh5q-L74QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=react&prettier=false&targets=&version=7.6.2&externalPlugins=%40babel%2Fplugin-transform-react-constant-elements%407.6.0

```js
let foo = 'hello';

export const Component = () => {
  foo = 'goodbye';
  return <span>{foo}</span>;
};
```

**Expected output**

```js
let foo = 'hello';

export const Component = () => {
  foo = 'goodbye';
  return React.createElement("span", null, foo);
};
```

**Actual output**

```js
let foo = 'hello';

var _ref =
/*#__PURE__*/
React.createElement("span", null, foo);

export const Component = () => {
  foo = 'goodbye';
  // renders "hello" instead of "goodbye"
  return _ref;
};
```

nicolo-ribaudo:
| Fixed Issues?            | Fixes #10522
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Related tests:
- https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-react-constant-elements/test/fixtures/constant-elements/dont-hoist-before-hoc/
- https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-react-constant-elements/test/fixtures/constant-elements/reassignment

### DIFF
--- a/packages/babel-traverse/src/path/lib/hoister.js
+++ b/packages/babel-traverse/src/path/lib/hoister.js
@@ -32,6 +32,15 @@ const referenceVisitor = {
     const binding = path.scope.getBinding(path.node.name);
     if (!binding) return;
 
+    // we can handle reassignments only if they happen in the same scope as the declaration
+    for (const violation of binding.constantViolations) {
+      if (violation.scope !== binding.path.scope) {
+        state.mutableBinding = true;
+        path.stop();
+        return;
+      }
+    }
+
     // this binding isn't accessible from the parent scope so we can safely ignore it
     // eg. it's in a closure etc
     if (binding !== state.scope.getBinding(path.node.name)) return;
@@ -46,6 +55,9 @@ export default class PathHoister {
     this.breakOnScopePaths = [];
     // Storage for bindings that may affect what path we can hoist to.
     this.bindings = {};
+    // "true" if the current path contains a reference to a binding whose
+    // value can change and thus can't be safely hoisted.
+    this.mutableBinding = false;
     // Storage for eligible scopes.
     this.scopes = [];
     // Our original scope and path.
@@ -194,6 +206,8 @@ export default class PathHoister {
 
   run() {
     this.path.traverse(referenceVisitor, this);
+
+    if (this.mutableBinding) return;
 
     this.getCompatibleScopes();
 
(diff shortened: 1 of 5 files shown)


packages/babel-traverse/src/path/lib/hoister.js
```diff
@@ -32,6 +32,15 @@ const referenceVisitor = {
     const binding = path.scope.getBinding(path.node.name);
     if (!binding) return;
 
+    // we can handle reassignments only if they happen in the same scope as the declaration
+    for (const violation of binding.constantViolations) {
+      if (violation.scope !== binding.path.scope) {
```
jridgewell: Can't we still hoist to the level of the violation?

```js
let foo = 'hello';
export const Component = () => {
  foo = 'goodbye';
  return <span>{array.map(() => <inner>{foo}</inner>}</span>;
};

Comment (nicolo-ribaudo) [reply packages/babel-traverse/src/path/lib/hoister.js]: No, we can't. Consider this example:

```js
let foo = 0;
export const Component = () => {
  foo++;
  return Promise.resolve().then(() => <span>{foo}</span>);
};

Component();
Component();
```

Here both the components should use `2`, but if we hoist it the first one will use `1`.

PR #10532, opened 2019-10-09
Allow duplicate __proto__ keys in patterns, simple case (#6705)

### LINKED ISSUE #6705: should allow duplicate __proto__ keys in patterns
> Issue originally reported by @babel-bot in babel/babylon#306

Original issue submitted by @gsathya in https://github.com/babel/babel/issues/5145

### Input Code

```js
result = { __proto__: x, __proto__: y } = value;
```

presets - latest

### Expected Behavior
No error

### Current Behavior
Error: `Redefinition of __proto__ property`

### Your Environment

| software         | version
| ---------------- | -------
| Babel            |  
| node             | 
| npm              | 
| Operating System |

alejo90:
| Fixed Issues?            | Fixes #6705 
| Documentation PR Link    | 

This is a fix for an error being thrown when there is a duplicate `__proto__` in a destructuring pattern. It only fixes the simple case when the pattern is immediately followed by a `=`. I renamed the `checkPropClash` function to `checkDuplicatedProto`, as advised in the issue. Instead of relying exclusively on a boolean flag to determine whether `__proto__` had appeared before or not, I store the amount of times it has appeared and the position of the first duplicated occurrence. In `parseObj`, I check for the existence of the first duplicated occurrence property I stored before and raise an error if it's there and the next token is not a `=`.
Please let me know if there are any concerns or improvements I can make. I'm thinking about the more complicated case as well but I'll need a little more time with the code before I can tackle it.
Thanks @nicolo-ribaudo for the hints on how to get started.


nicolo-ribaudo:
We also need to update the test262 whitelist. Could you run `make bootstrap-test262` and `make test-test262-update-whitelist`=

packages/babel-parser/src/parser/expression.js
```diff
@@ -67,14 +67,13 @@ export default class ExpressionParser extends LValParser {
...
-  // strict mode, init properties are also not allowed to be repeated.
+  // Check if property __proto__ has been used more than once.
+  // If the expression is a destructuring assignment, then __proto__ may appear
+  // multiple times. Otherwise, __proto__ is a duplicated key.
 
-  checkPropClash(
+  checkDuplicatedProto(
     prop: N.ObjectMember | N.SpreadElement,
-    propHash: { [key: string]: boolean },
+    propHash: { [key: string]: number },
```
nicolo-ribaudo: Could you rename this to something like `protoRef`?
Also, ideally it's type should be `{ used: boolean, start?: number }`.
And then, when we find `__proto__`,
1. If `used` is true, set `start`
2. Set `used` to true

alejo90: Thanks, @nicolo-ribaudo. Wouldn't there be an issue if we had more than two occurences of `__proto__`? In this way, `start` would be the last occurrence's start position instead of the first duplicated one. It would be an error all the same if it happens outside of a destructuring pattern but it wouldn't give us the absolute first error position. Does that matter in this case?

nicolo-ribaudo: > In this way, start would be the last occurrence's start position instead of the first duplicated one.

Ok this is the best behavior. We can do `if (protoRef.used && !protoRef.start)` to preserve it.

nicolo-ribaudo:
@JLHwung I have repoened the issue

PR #10543, opened 2019-10-12
add asserts predicate

### LINKED ISSUE #10527: TypeScript assertion signatures
TypeScript 3.7 will support an `asserts Identifier is Type` syntax in function annotation positions (much like with type predicates).

```ts
declare function assertIsFoo(x): asserts x is Foo;

function assertIsDefined<T>(x: T): asserts x is NonNullable<T> {
    if (x == null) throw "oh no";
}
```

See more at the original pull request: https://github.com/microsoft/TypeScript/pull/32695

tanhauhau:
| Fixed Issues?            | Fixes #10527
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-types/src/definitions/typescript.js
```diff
@@ -175,10 +176,11 @@ defineType("TSTypeReference", {
 
 defineType("TSTypePredicate", {
   aliases: ["TSType"],
-  visitor: ["parameterName", "typeAnnotation"],
+  visitor: ["parameterName", "typeAnnotation", "assertsModifier"],
   fields: {
+    assertsModifier: validateOptionalType("TSAssertsKeyword"),
```
nicolo-ribaudo: This should be a boolean, since the only needed information is if it is present or not.

tanhauhau: I was following the TypeScript AST.
But I understand Babel AST doesn't have to be the same as Typescript AST.

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -904,21 +904,34 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         const t: N.TsTypeAnnotation = this.startNode();
         this.expect(returnToken);
 
+        const assertsModifier = this.tsTryParse(
```
nicolo-ribaudo: Note that this could be parser without using `tryParse`: it would be possible to parse the `assert` identifier and then convert it to a type annotation if the next token is not another identifier.

On the other hand, we use `tryParse` everywhere when parsing TypeScript. I think that it's better to be consistent here, and then the whole TS parser can be optimized on other PRs.

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -904,21 +904,34 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
-          return this.tsParseTypeAnnotation(/* eatColon */ false, t);
+          if (!assertsModifier) {
+            // : type
+            return this.tsParseTypeAnnotation(/* eatColon */ false, t);
+          }
+
+          // : asserts foo
+          const node = this.startNodeAtNode(t);
+          node.parameterName = this.parseIdentifier();
+          node.assertsModifier = assertsModifier;
```
nicolo-ribaudo: Nit
```suggestion
          node.assertsModifier = true;
```

packages/babel-parser/test/fixtures/typescript/arrow-function/predicate-types/output.json
```diff
@@ -73,12 +73,12 @@
             },
             "typeAnnotation": {
               "type": "TSTypePredicate",
-              "start": 10,
+              "start": 8,
```
nicolo-ribaudo: :+1:

nicolo-ribaudo:
The CircleCI failure is unrelated and already fixed on master.

bradzacher:
Hi there!

We on `typescript-eslint` added support for this to our parser a few weeks ago (in https://github.com/typescript-eslint/typescript-eslint/pull/1045)

Could we look to align the AST representations?
Specifically we chose to go with `asserts: boolean` instead of `assertsModifier: boolean`

nicolo-ribaudo:
I'm 100% ok with it. We are going to release a new Babel version soon; would you be able to prepare a quick PR?

thorn0:
> quick PR

I'll do it

existentialism:
@thorn0 thanks!

PR #10546, opened 2019-10-13
Transform TypeScript "declare" fields

### LINKED ISSUE #10311: TypeScript definite assignment emit
## Bug Report

**Current Behavior**
In loose mode, `@babel/plugin-proposal-class-properties` emits a property assignment for TypeScript definite assignments.

**Input Code**
[REPL](https://babeljs.io/en/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=PTAEBkEsGsFNQPICMBWsDGAXAdAQwM76QDmAdqEgK6agDukANg6KQPY0FFmgAOATqx6w-mSLHx0AFpHSTQuPvFwNFuACYBPUGtgAzSKVhqAUDvQMF8dK1L4OhEqQAie3JQaZ8ALlCYNQ1l1EVAwcTkdjY3NOUAAxVlZQAG9jUAoFAEIfUkoAWyRhAG5ItOtbTD5KLFY-AApWADdhPkgdb2T0vmy8gr5QAF8ASmTUtPkHMhddNw98WsxpfAAaUEbm1vFB4rT-436gA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=typescript&prettier=false&targets=Node-8.9&version=7.5.5&externalPlugins=%40babel%2Fplugin-syntax-typescript%407.3.3%2C%40babel%2Fplugin-proposal-class-properties%407.5.5%2C%40babel%2Fplugin-transform-typescript%407.5.5) (I'm not sure if it's possible to turn turn on loose mode for the plugin)
[TS Playground](https://www.typescriptlang.org/play/index.html#code/PTAEBkEsGsFNQPICMBWsDGAXAdAQwM76QDmAdqEgK6agDukANg6KQPY0FFmgAOATqx6w+mSLHx0AFpHSTQuPvFwNFuACYBPUGtgAzSKVhqAsACgd6BgvjpWpfB0IlSAET25KDTPgBcoTBpCrLqIqBg4nM5mZpacoABirKygAN5moBQKAIR+pJQAtkjCANzRphm29ph8lFisfAAUrABuwnyQOr6pmXy5BUV8oAC+AJSp6RnyTmRuuh5e+A2Y0vgANKAtbR3iI6Xlw2ZDQA)

```ts
// Like Object.assign but will not assign properties which are already defined
declare const assignDefaults: typeof Object.assign

class Foo {
  bar!: number;

  constructor(overrides: { bar: number }) {
    assignDefaults(this, overrides);
  }
}
```

**Expected behavior/code**
A definite assignment should not emit any JS.

```js
// Like Object.assign but will not assign properties which are already defined
class Foo {
  constructor(overrides) {
    assignDefaults(this, overrides);
  }

}
```

```json
{
  "presets": ["@babel/preset-typescript"],
  "plugins": [["@babel/plugin-proposal-class-properties", { "loose": true }]]
}
```

**Environment**
- Babel version(s): 7.5.5
- Node/npm version: Node 12.6 / yarn 1.15.2
- OS: macOS 10.14.5
- Monorepo: no

This was previously raised in #7997, but considered expected behavior. It was mentioned that it could be reconsidered for `loose` mode, however. 

I do feel like it should be reconsidered one way or another; emitting for these declarations is problematic in the same way emitting for a `declare` would be. Additionally, this syntax is not valid JS so there the only applicable spec compliance is that of TS.

nicolo-ribaudo:
| Fixed Issues?            |  
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Since there is a breaking change (uninitialized fields are now initialized to undefined), I enabled it behind an option (`allowDeclareFields`). This option is similar to `useDefineForClassFields` (https://github.com/microsoft/TypeScript/issues/27644), except that [[Set]] vs [[Define]] is set by the `loose` option of the properties transform plugin.

Incidentally, this PR fixes #10311 when the plugins are used in the correct order.


packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -44,11 +44,90 @@ function registerGlobalType(programScope, name) {
...
+                "Parameter properties can not be destructuring patterns.",
+              );
+            }
+
+            return template.statement.ast`this.${id} = ${id}`;
+          });
+
+          injectInitialization(classPath, path, assigns);
+        }
+      },
```
nicolo-ribaudo: The code for `method` and `constructor` has just been moved here from the lines before.

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -44,11 +44,90 @@ function registerGlobalType(programScope, name) {
...
+  (
+    api,
+    { jsxPragma = "React", allowNamespaces = false, declareFields = false },
+  ) => {
     api.assertVersion(7);
 
     const JSX_ANNOTATION_REGEX = /\*?\s*@jsx\s+([^\s]+)/;
 
+    const classMemberVisitors = {
+      field(path) {
```
nicolo-ribaudo: I didn't use the `ClassProperty` visitor because the class fields transform transforms them using the `Class` visitor, so they would never be visited in practice.

nicolo-ribaudo:
cc @babel/typescript

packages/babel-helper-create-class-features-plugin/src/typescript.js
```diff
@@ -0,0 +1,19 @@
...
+
+export function assertFieldTransformed(path: NodePath) {
+  // TODO (Babel 8): Also check path.node.definite
+
+  if (path.node.declare) {
+    throw path.buildCodeFrameError(
+      `TypeScript 'declare' fields must first be transformed by ` +
+        `@babel/plugin-transform-typescript.\n` +
+        `If you have already enabled  that plugin (or '@babel/preset-typescript'), make sure ` +
+        `that it runs  before any plugin related to additional class features:\n` +
```
JLHwung: nit: `s/  / /g`
(Even though I could comment on multiple lines, I couldn't submit a multiple line suggestion, else I would have created one.)

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -44,11 +44,90 @@ function registerGlobalType(programScope, name) {
 }
 
 export default declare(
-  (api, { jsxPragma = "React", allowNamespaces = false }) => {
+  (
+    api,
+    { jsxPragma = "React", allowNamespaces = false, declareFields = false },
```
existentialism: Should we name this `allowDeclareFields` (or `allowDeclare`) to match `allowNamespaces`?

nicolo-ribaudo: :+1: `allowDeclareFields`

nicolo-ribaudo:
Merged to #10545 (not in master yet)

PR #10555, opened 2019-10-15
[TS] Correctly transform computed strings and templates in enums

### LINKED ISSUE #10298: Flagging: Improved backticks support coming to TypeScript
TypeScript 3.7 will likely improve its support for template strings that have no substitutions, and make them usable in much more places than it currently does (basically every time it would accept a regular literal string, in particular in enum declarations), so the Babel parser will need to be updated accordingly.

This will be very useful for TS codebases since they'll finally be able to enable the `backtick` version of the ESLint [`quote` rule](https://eslint.org/docs/rules/quotes).

Ref https://github.com/microsoft/TypeScript/pull/32064#issuecomment-517402758

nicolo-ribaudo:
| Fixed Issues?            | Fixes #10298
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

TS playground demo: https://www.typescriptlang.org/play/index.html?ts=3.7-Beta&ssl=6&ssc=3&pln=4&pc=2#code/PTAEHsGsFgCg4KYDsCuBbUBBUBvOoDQAPUAXlACJMLQBqSgIwoBp9CBPM0AAwC9vWsAL4BuIA


packages/babel-plugin-transform-typescript/src/enum.js
```diff
@@ -161,6 +160,11 @@ function evaluate(
         return evalConstant(expr.expression);
       case "Identifier":
         return seen[expr.name];
+      case "TemplateLiteral":
+        if (expr.quasis.length === 1) {
+          return expr.quasis[0].value.cooked;
+        }
+      /* falls throught */
```
nicolo-ribaudo:
```suggestion
      /* falls through */
```

Issue #4987, opened 2016-12-10
Run tc39/test262 against Babel 


xtuc:
### Context
The idea is to run [tc39/test262](https://github.com/tc39/test262) tests suite against Babel.

> Running the official spec tests against Babel can make sure we comply to the spec or at least know when we don’t. We’ll want to figure out how to do filtering on things we can’t compile (proxy, TCO, etc) and set up an easy way to check failing tests and file issues and PRs for them.

This issue will be used to track this.

###  Previous discussions:
* https://github.com/kangax/compat-table/issues/830
* https://github.com/bakkot/test262-web-runner/
* https://github.com/bterlson/test262-harness/issues/58
* https://github.com/motiz88/babel-smoke-tests

### Possible Solution
* https://github.com/bterlson/test262-harness (it seems pretty straightforward, replace node with babel-node ?)

I think we should use our own test runner. We might want to skip some tests.

There are some restriction in order to run the tests. They are well explained in https://github.com/tc39/test262/blob/master/INTERPRETING.md#test-execution.

The tests could be run automaticly on the CI (they are taking some time).

/cc @kangax 

---
**Edit**: 11/12

My test is available here: https://github.com/xtuc/babel-test262.

Some observations I made:
* Running all tests takes too long ([build log](https://travis-ci.org/xtuc/babel-test262/jobs/182910702))
* Mocha reporter and only the language `built-ins` tests ([build log](https://travis-ci.org/xtuc/babel-test262/builds/183020833)) and this time the ouput was too long.
* Mocha dot reporter and only the language `built-ins` tests seems to work pretty well ([build log](https://travis-ci.org/xtuc/babel-test262/builds/183025498))

kangax:
I did this couple months ago when testing Babili. Here's my script to run from within chakra-core repo — https://gist.github.com/kangax/d4e96384bdce3528d41808ae1843df14

Here's one of the issues I filed back then — https://github.com/babel/babel/issues/4756

xtuc:
The project now allows to run tests with given Babel transformations (for convienience I choose to use babel-preset-env).

I wanted it to be as simple and flexible as possible. See the README to see its usage.

I used an environement variable to configure the target browser.

```sh
TARGET_BROWSERS="last 200 versions"
```

Please tell me if this make sense to you and/or way to improve it.

Example build job: https://travis-ci.org/xtuc/babel-test262/builds/184788391.

### next

- [ ] @kangax, could this suffice to run your tests on Babili?
[...]

PR #10556, opened 2019-10-15
Add master branch workflow for test262 tests

jbhoosreddy:
Fixes #10500

| Fixed Issues?            | N/A (doesn't completely fix anything yet. only a step towards the goal
| Major: Breaking Change?  | N/A
| Minor: New Feature?      | N/A
| Tests Added + Pass?      | test262 tests added. Definitely don't pass 😝 
| Any Dependency Changes?  | No direct dependencies. Dependencies for CI

This PR adds a master branch job for running test262 tests. The results of
these tests will be stored in build artifacts which will be used to
compare with test262 runs on PR (future work!). Robin has also done a
lot of work to ensure all babel packages are properly linked to the
test runner. Additionally, Nicolo has helped in mentorship and
contributions to babel-test262-runner which is used here (https://github.com/babel/babel-test262-runner/compare/a59509...399abd).

Co-authored-by: Robin Ricard <rricard2@bloomberg.net>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,7 +118,7 @@ workflows:
   nightly:
       triggers:
         - schedule:
-            cron: "0 0 * * *"
+            cron: "0 5 * * *"
             filters:
               branches:
                 only:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -116,10 +116,9 @@ workflows:
     jobs:
       - test
   master:
-    filters:
-      branches:
-        only:
-          - master
     jobs:
-      - test
-      - test262
+      - test262:
+          filters:
+            branches:
+              only:
+                - master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,13 +115,11 @@ workflows:
   test:
     jobs:
       - test
-  nightly:
-      triggers:
-        - schedule:
-            cron: "0 5 * * *"
-            filters:
-              branches:
-                only:
-                  - master
-      jobs:
-        - test262
+  master:
+    filters:
+      branches:
+        only:
+          - master
+    jobs:
+      - test
+      - test262
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,12 +102,10 @@ jobs:
           <<: *test262_workdir
       - store_artifacts: *artifact_test262_tap
       - run:
-          name: Convert Tap output to Xunit
+          name: Output test262 results
           command: |
-            mkdir -p ~/test-results/test262
-            cat ~/test262.tap | $(npm bin)/tap-mocha-reporter xunit | tee ~/test-results/test262/results.xml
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -54,3 +63,65 @@ jobs:
...
+
+workflows:
+  version: 2
+  test:
+    jobs:
+      - test
+  nightly:
+      triggers:
+        - schedule:
+            cron: "0 0 * * *"
```
nicolo-ribaudo: According to https://api.github.com/repos/babel/babel/stats/punch_card, it's better to to this at

```suggestion
            cron: "0 7 * * *"
```

(is I'm understanding the "cron" format correctly: it should be 7 am UTC)

| Time | Commits |
|------|---------|
| 0 | 539 |
| 1 | 427 |
| 2 | 298 |
| 3 | 213 |
| 4 | 174 |
| 5 | 136 |
| 6 | 201 |
| 7 | 164 |
| 8 | 309 |
| 9 | 399 |
| 10 | 552 |
| 11 | 686 |
| 12 | 546 |
| 13 | 563 |
| 14 | 599 |
| 15 | 634 |
| 16 | 698 |
| 17 | 628 |
| 18 | 634 |
| 19 | 507 |
| 20 | 475 |
| 21 | 566 |
| 22 | 659 |
| 23 | 704 |

.circleci/config.yml
```diff
@@ -54,3 +63,65 @@ jobs:
...
+            yarn add tap-mocha-reporter --dev
+            curl -L https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64 > jq
+            chmod +x ./jq
+            for package in ../packages/*/package.json; do
+              yarn link $(./jq -j ".name" $package)
+            done
+            node lib/download-node
+      - run:
+          name: Run Test262
+          command: node lib/run-tests I_AM_SURE | tee ~/test262.tap
```
nicolo-ribaudo: Rather than linking every Babel package, what do you think of adding a `--babelMonorepo ..` cli clag to `lib/run-tests`?

jbhoosreddy: That is something I wanted to do in a followup PR in which I want to:
1. Use something like yargs: http://yargs.js.org/
2. Refactor existing cli arguments
3. Add path to babel argument
4. Add path to whitelist files so that they can remain colocated to babel/babel repo
5. Add babel plugins/config argument

.circleci/config.yml
jbhoosreddy: confirmed that it's the right syntax

jbhoosreddy:
Thinking a bit more about this I am realizing that only running these tests nightly doesn’t achieve our end result. Because when I query master branch artifacts, it is necessary that I find it in the latest master job, of other non-nightly jobs have run. Based on that thinking, I’ll add a commit to make these tests always run on master, so that I can rely on them existing when I’m in a PR job.

jbhoosreddy:
Relates to: #4987

PR #10557, opened 2019-10-15
fix: disable caching when babel could not read/write cache

### LINKED ISSUE #6906: @babel/register crashes when node_modules/.cache folder is not writable
Bug report.

`@babel/register` crashes when the `node_modules/.cache` folder is not writable. This can happen if the node process is running as a different user than the user who owns the `node_modules` folder.

### Babel Register Configuration

The specifics of my config are not relevant to this issue, but here it is anyway.

```js
babelRegister({
  only: [/views/, /lib/],
  extensions: ['.js', '.mjs']
})
```

### Expected Behavior

I noticed that `@std/esm` also uses the `node_modules/.cache` folder, but it recovers gracefull when this folder is not writable, probably falling back to another folder or simply disabling the cache. This seems like much better behavior than crashing the process.

### Current Behavior

The process crashes when `node_modules/.cache` is not writable.

### Context

I run my production Node.js process as the `www-data` user, which does not have permission to write to the app folder, including the `node_modules` folder. Therefore, `@babel/register` is unable to write into the `node_modules/.cache` folder.

### Your Environment

| ---------------- | -------
| Babel            |  @babel/core@7.0.0-beta.32 
| Babel-register | @babel/register@7.0.0-beta.32
| Babylon          | 
| node             | 8.9.0
| npm              | 5.5.1
| Operating System | macOS 10.13.1

### LINKED ISSUE #9649: Stack overflow exception when using Azure Run from Package feature
## Bug Report

**Current Behavior**
When using Azure DevOps to deploy to an Azure App Service or Function the default deployment method is `Run from Package`.
This will simply deploy a zip file with your whole web-app. This zip gets mounted as a read-only `wwwroot` directory.
The problem is that babel tries to write a cache file in the read-only directory. For some reason this generates a stack exception in the `mkdirp`.

**Expected behavior/code**
I expect a warning that tells me the cache file could not be created and caching will not be used.

**Current Solution**
I have succesfully used the `BABEL_CACHE_PATH` setting to point to a location outside of the default Azure `wwwroot` directory.

JLHwung:
| Fixed Issues?            | Fixes #6906, fixes #9649 
| Tests Added + Pass?      | Added test on permission issues
| Documentation PR Link    | 
| Any Dependency Changes?  |

In this PR we catch the IO error and disable caching when permission issue arises, or it the cache path resides in a readonly filesystem.

The following `existsSync` usage
```js
if (fs.existsSync(file)) {
  fs.readFileSync(file)
}
```
is also replaced as it introduces duplicate IO operations. In general [`fs.existsSync`](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback) should only be used
> if the file won’t be used directly, for example when its existence is a signal from another process.

In the future I will open a separate PR to replace all similar usage in our codebase.

### DIFF
--- a/packages/babel-register/src/cache.js
+++ b/packages/babel-register/src/cache.js
@@ -24,6 +24,7 @@ function isCacheDisabled() {
  */
 
 export function save() {
+  if (isCacheDisabled()) return;
   let serialised: string = "{}";
 
   try {
@@ -70,6 +71,9 @@ because it resides in a readonly filesystem. Cache is disabled.`,
 export function load() {
   if (isCacheDisabled()) return;
 
+  process.on("exit", save);
+  process.nextTick(save);
+
   let cacheContent;
 
   try {
@@ -90,9 +94,6 @@ due to a permission issue. Cache is disabled.`,
         throw e;
     }
   }
-
-  process.on("exit", save);
-  process.nextTick(save);
 
   try {
     data = JSON.parse(cacheContent);
--- a/packages/babel-register/src/cache.js
+++ b/packages/babel-register/src/cache.js
@@ -69,7 +69,10 @@ because it resides in a readonly filesystem. Cache is disabled.`,
  */
 
 export function load() {
-  if (isCacheDisabled()) return;
+  if (isCacheDisabled()) {
+    data = {};
+    return;
+  }
 
   process.on("exit", save);
   process.nextTick(save);
@@ -80,8 +83,6 @@ export function load() {
     cacheContent = fs.readFileSync(FILENAME);
   } catch (e) {
     switch (e.code) {
-      case "ENOENT":
-        return;
       case "EACCES":
       case "EPERM":
         console.warn(
@@ -89,15 +90,15 @@ export function load() {
 due to a permission issue. Cache is disabled.`,
         );
         cacheDisabled = true;
-        return;
+      /* fall through */
       default:
-        throw e;
+        return;
     }
   }
 
   try {
     data = JSON.parse(cacheContent);
-  } catch (err) {}
+  } catch {}
 }
 
 /**
--- a/packages/babel-register/src/cache.js
+++ b/packages/babel-register/src/cache.js
@@ -14,6 +14,11 @@ const DEFAULT_FILENAME = path.join(
 const FILENAME: string = process.env.BABEL_CACHE_PATH || DEFAULT_FILENAME;
 let data: Object = {};
 
+let cacheDisabled = false;
+
+function isCacheDisabled() {
+  return process.env.BABEL_DISABLE_CACHE ?? cacheDisabled;
+}
 /**
  * Write stringified cache to disk.
  */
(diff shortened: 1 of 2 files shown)


packages/babel-register/src/cache.js
```diff
@@ -32,27 +38,66 @@ export function save() {
...
+        cacheDisabled = true;
+        return;
+      default:
+        throw e;
+    }
   }
+
+  try {
+    data = JSON.parse(cacheContent);
+  } catch (err) {}
```
nicolo-ribaudo: Nit:

```suggestion
  } catch {}
```

I'm surprised that eslint doesn't warn about the unused variable.

nicolo-ribaudo:
Is it possible to add a test somehow?

packages/babel-register/src/cache.js
JLHwung: > eslint doesn't warn about the unused variable

Because optional catching binding is a stage-4 proposal?

packages/babel-register/src/cache.js
```diff
@@ -32,27 +38,66 @@ export function save() {
...
+    switch (e.code) {
+      case "EACCES":
+      case "EPERM":
+        console.warn(
+          `Babel could not write cache to file: ${FILENAME} 
+due to a permission issue. Cache is disabled.`,
+        );
+        cacheDisabled = true;
+        break;
+      case "EROFS":
```
JLHwung: Disclaimer: I have not tested on a readonly filesystem, this error code is copied from `man 2 open`
> [EROFS]            The named file resides on a read-only file system, and the file is to be modified.

JLHwung:
@nicolo-ribaudo Added test on permission issues. On the readonly file system error, I think it is hard to reproduce on unit testing.

packages/babel-register/test/cache.js
```diff
@@ -73,5 +73,42 @@ describe("@babel/register - caching", () => {
       expect(fs.existsSync(testCacheFilename)).toBe(true);
       expect(get()).toEqual({});
     });
+
+    it("should create the cache after load", cb => {
+      load();
+
+      process.nextTick(() => {
+        process.nextTick(() => {
```
nicolo-ribaudo: [Q]: Why isn't one `nextTick` enough?

nicolo-ribaudo:
@JLHwung The failing test on windows it a test of this PR.

packages/babel-register/test/cache.js
JLHwung: [A]: I thought the execution order of `next tick` tasks is not defined. Apparently I am wrong after reading https://nodejs.org/docs/latest/api/process.html#process_process_nexttick_callback_args, thank you for reminding me!

JLHwung:
@nicolo-ribaudo The failing test on Windows is due to the following [caveats](https://nodejs.org/api/fs.html#fs_file_modes)
> on Windows only the write permission can be changed [by `fs.chmod`]

I guess that the read error could only occur when the file is owned by another user and have ACL settings excluding current user to read it. There is no `isReadable` mode on Windows.

As it is difficult to construct such example on Windows CI, I disable this test on Windows.

PR #10559, opened 2019-10-15
fix: Exclude catch clause from let identifier error

### LINKED ISSUE #10413: [parser] `let` in catch clause is not an error in sloppy
## Bug Report

The `let` identifier should be valid in a catch clause in sloppy (non-strict) mode but this appears to be invalid.

**Current Behavior**

The `let` identifier is special in strict mode but not a keyword in non-strict mode. In this case Babel still throws over it where I think it should not.

**Input Code**

https://astexplorer.net/#/gist/15eb31fbb1024faa6f42a4940d8beb4a/50b40e18038bb65c40bc3257e886093a206ad411

```js
try {} catch (let) {}
```

**Expected behavior/code**

Should pass.

**Environment**

Babel 7.5.5

**Possible Solution**

Update heuristics to detect this :)

I hope I'm not missing some weird exception in the spec but as far as I can see there is none and this should pass.

gonzarodriguezt:
| Fixed Issues?            | Fixes #10413 

I added a new getter in `scope` to know if the current scope is a catch block, and then use that property to exclude catch blocks from the original condition

I also tried to go the other way around and implicitly exclude the catch block by raising the error not on lexical binding, but only on `let` and `const` declarations, but i couldn’t find a way to check the kind of the variable declaration when handling an identifier's name

I think other solution could be adding a new scope flag for catch blocks, but i'm not really sure

What do you guys think?

### DIFF
--- a/packages/babel-parser/src/parser/lval.js
+++ b/packages/babel-parser/src/parser/lval.js
@@ -383,7 +383,11 @@ export default class LValParser extends NodeUtils {
             checkClashes[key] = true;
           }
         }
-        if (bindingType === BIND_LEXICAL && expr.name === "let") {
+        if (
+          bindingType === BIND_LEXICAL &&
+          expr.name === "let" &&
+          !this.scope.inCatchBlock
+        ) {
           this.raise(
             expr.start,
             "'let' is not allowed to be used as a name in 'let' or 'const' declarations.",
--- a/packages/babel-parser/src/parser/lval.js
+++ b/packages/babel-parser/src/parser/lval.js
@@ -17,7 +17,7 @@ import type {
 import type { Pos, Position } from "../util/location";
 import { isStrictBindReservedWord } from "../util/identifier";
 import { NodeUtils } from "./node";
-import { type BindingTypes, BIND_NONE, BIND_LEXICAL } from "../util/scopeflags";
+import { type BindingTypes, BIND_NONE } from "../util/scopeflags";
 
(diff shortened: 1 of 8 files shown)


packages/babel-parser/src/parser/lval.js
```diff
@@ -383,7 +383,11 @@ export default class LValParser extends NodeUtils {
             checkClashes[key] = true;
           }
         }
-        if (bindingType === BIND_LEXICAL && expr.name === "let") {
+        if (
+          bindingType === BIND_LEXICAL &&
+          expr.name === "let" &&
+          !this.scope.inCatchBlock
+        ) {
```
nicolo-ribaudo: Does this approach work with this code, which should throw?

```js
try {} catch (err) {
  let let;
}
```

gonzarodriguezt: It doesn’t, thanks for pointing it out, i'll work on that

packages/babel-parser/src/parser/statement.js
```diff
@@ -1021,6 +1021,7 @@ export default class StatementParser extends ExpressionParser {
       kind === "var" ? BIND_VAR : BIND_LEXICAL,
       undefined,
       "variable declaration",
+      kind === "var" ? false : true,
```
```suggestion
      kind !== "var",
```

gonzarodriguezt: Good one lol

PR #10567, opened 2019-10-16
[parser] Exception to 8 and 9 in tagged template

### LINKED ISSUE #10437: [parser] Tagged template with \8 should set value to null
## Bug Report

**Current Behavior**

Minor AST issue; The `\8` and `\9` are illegal escapes. In tagged template these should be recognized as bad escapes and set the `.value` property in the AST to `null` to reflect this.

**Input Code**

```js
`\8`
```

```js
`\9`
```

**Expected behavior/code**

For templates, an escape `\8` and `\9` end up as `NotEscapeSequence`
https://tc39.es/ecma262/#prod-NotEscapeSequence

The string value of that escape is undefined:
https://tc39.es/ecma262/#sec-static-semantics-tv-and-trv
> The TV of TemplateCharacter :: `\` NotEscapeSequence is `undefined`.

As such, the `.value` property in the AST nodes should be `null` to reflect this, like with other bad escapes.

**Environment**
Babel 7.5.5

**Possible Solution**

I think there's an exception to 8 and 9. Do the same as happens with `\1` ~ `\7`. This is for _tagged_ templates only.

pnowak:
| Fixed Issues?            | Fixes #10437 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


nicolo-ribaudo:
Thanks!

pvdz:
Thanks! :)
(Could have used a tagged template test case, since tagged templates have a special case for illegal escapes)

PR #10572, opened 2019-10-18
[transform-react-jsx] Add useSpread option to transform JSX

### LINKED ISSUE #9652: Add option to preset-react / transform-react-jsx to output object spread syntax
## Feature Request

Currently, transform-react-jsx either outputs uses of the `extend` helper or calls to `Object.assign` for handling spread props (any attribute with `{...Expression}`).

An option to instead convert such instances to an inline object with spread elements.

This would still allow `@babel/preset-env` to further transform the inline object if the allowed targets don't support object spread.

If it wasn't a semver-major change I'd be as bold as to say that it's safe to make it enabled by default -- I suspect almost no project runs `@babel/preset-react` without also using `@babel/preset-env`, and native object rest spread has been supported by it even without requiring the `shippedProposals` option. It has also been shipping on browsers and node (LTS, even) for over a year.

ivandevp:
| Fixed Issues?            | Fixes #9652 
| Documentation PR Link    | https://github.com/babel/website/pull/2110 


packages/babel-helper-builder-react-jsx/src/index.js
```diff
@@ -172,6 +172,14 @@ You can turn on the 'throwIfNamespace' flag to bypass this warning.`,
     let _props = [];
     const objs = [];
 
+    const useSpread = file.opts.useSpread || false;
```
nicolo-ribaudo: I don't like that this accepts `0`, but let's keep it as-is for consistency with the other options of this file 🤷‍♀️

We should revisit it in Babel 8.

JLHwung: IMO, we could begin on `useSpread` now because it is a new option. `useBuiltins: 0` is strange enough and it does not justify that you can even use `useSpread: 0`.

JLHwung: I think we should also add `useSpread` to `preset-react`, too.

packages/babel-helper-builder-react-jsx/src/index.js
```diff
@@ -172,7 +180,15 @@ You can turn on the 'throwIfNamespace' flag to bypass this warning.`,
...
-    const useBuiltIns = file.opts.useBuiltIns || false;
+    const useSpread = setDefaultValue(file.opts.useSpread, false);
+    if (typeof useSpread !== "boolean") {
+      throw new Error(
+        "transform-react-jsx currently only accepts a boolean option for " +
+          "useSpread (defaults to false)",
+      );
+    }
+
+    const useBuiltIns = setDefaultValue(file.opts.useBuiltIns, false);
```
nicolo-ribaudo: While this makes 100% sense, it is a breaking change (for example, if you pass `0`) and we should defer it to Babel 8.

packages/babel-helper-builder-react-jsx/src/index.js
```diff
@@ -172,7 +180,15 @@ You can turn on the 'throwIfNamespace' flag to bypass this warning.`,
     let _props = [];
     const objs = [];
 
-    const useBuiltIns = file.opts.useBuiltIns || false;
+    const useSpread = setDefaultValue(file.opts.useSpread, false);
```
nicolo-ribaudo: You can use
```js
const { useSpread = false } = file.opts;
```

packages/babel-helper-builder-react-jsx/src/index.js
ivandevp: So. Should I go back to the previous version?

nicolo-ribaudo: Yes 👍

ivandevp: Oh ok, got it.

ivandevp:
> I think when `useSpread` is enabled, we can further simplify the logic to
> 
> ```js
> return attribs.map(convertAttribute)
> ```
> 
> because `objs` is used to support breaking at spreadElement only.
> 
> And we may teach `convertAttribute` to replace `JSXSpreadAttribute` by `spreadElement`.

Hey @JLHwung , do you mean to add a condition before `while` loop to check if `useSpread` is set and put an early return mapping all attributes to `convertAttribute`? If that's the case, I would need to return and object expression instead of an array, right? Something like:

```js
if (useSpread) {
  const props = attribs.map(convertAttribute);
  return t.objectExpression(props);
}
```

JLHwung:
@ivandevp correct, I missed the `objectExpression` but you got me.

ivandevp:
I've made the requested changes guys, let me know if you think we can do any other enhancement.

JLHwung:
@ivandevp Could you add `useSpread` to `preset-react`, too? The other changes look good to me.

ivandevp:
Sure @JLHwung! I forgot that one :sweat_smile:

ivandevp:
Hey @JLHwung ! I've tried to follow the `useBuiltIns` flow in `preset-react` in order to do the same for `useSpread`. Is my change ok?

PR #10573, opened 2019-10-18
Create @babel/plugin-syntax-top-level-await

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

As suggested by @hzoo in https://github.com/babel/babel/pull/10449#issuecomment-537207583, I added a `caller` flag for TLA used by `preset-env`, because both webpack and rollup already have experimental support for top level await.

The tests are currently failing because this PR depends on #10449


packages/babel-plugin-syntax-top-level-await/.npmignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+*.log
+src
```
JLHwung: nit: should also ignore `test`

nicolo-ribaudo: I just copy-pasted this file from somewhere else in the repo; we should fix many of them.

PR #10576, opened 2019-10-18
[parser] validation for parentheses in the left-hand side of assignment expressions

### LINKED ISSUE #8796: Parser: Should not treat assignment to group as pattern
## Bug Report

**Current Behavior**

When assigning to a group that is an assignment the parser currently creates a Pattern node for the group. It should consider this a syntax error.

**Input Code**

```js
(x = x) = x
```

**Expected behavior/code**

Syntax error.

```js
        babelAst = babelParse(code, {
          sourceType: goalMode === GOAL_MODULE ? 'module' : 'script',
          strictMode: startInStrictMode ? true : false,
          plugins: ['estree'],
        }).program;
```

**Environment**
- Babel version(s): v7.0.0 (regular)
- Node/npm version: node 10
- OS: linux
- Monorepo only on sundays
- How you are using Babel: code

**Possible Solution**

Whatever the mechanism for checking assignable is in Babel should do so properly across parens. 

Current:
```
{ type: 'Program',
  body:
   [ { type: 'ExpressionStatement',
       expression:
        { type: 'AssignmentExpression',
          left:
           { type: 'AssignmentPattern',
             left: { type: 'Identifier', name: 'x' },
             right: { type: 'Identifier', name: 'x' } },
          operator: '=',
          right: { type: 'Identifier', name: 'x' } } } ] }
```

Expected:

https://tc39.github.io/ecma262/#sec-assignment-operators-static-semantics-assignmenttargettype

This rule says that an assignment itself is not a valid assignment. 

Note that assignments are right associative, so `x = x = x` is really `x = (x = x)`, which is why this is still valid while still considering assignments invalid assignment targets. 

https://tc39.github.io/ecma262/#sec-semantics-static-semantics-assignmenttargettype

This rule says that the AssignmentTargetType state of a (non-arrow) group is the AssignmentTargetType of its contents.

In other words; it's an invalid AssignmentTargetType and it should throw a syntax error.

### LINKED ISSUE #9899: Destructuring assignment left-hand side allows parens
## Bug Report

**Current Behavior**

`babel-parser` accepts the following but should not.

**Input Code**

```js
[(a = 1)] = t;
[([b])] = t;
```

**Expected behavior/code**

It should error.

boweihan:
| Fixed Issues?            | Fixes #8796, fixes #9899

@nicolo-ribaudo has a good description of the problem in the linked issue. Basically, the parser is missing some validation for parentheses in the left-hand side of assignment expressions.

This means that expressions such as `[a = 1] = t` and `[ { x: y = 3 } ] = t` are valid (and correctly so!).

The issue with the current behaviour is that expressions such as 1:`(a = 1) = t` and 2:`[(a = 1)] = t` are treated as valid. The left-hand side of assignment expressions should be invalid if there are any parenthesized child nodes which are not of type `Identifier` or `MemberExpression`.

The fix in this pr involves:
- adding the parens check to `toAssignable`
- updating / adding test cases to reflect new behavior


packages/babel-parser/src/parser/lval.js
```diff
@@ -60,7 +65,12 @@ export default class LValParser extends NodeUtils {
           ) {
             const prop = node.properties[i];
             const isLast = i === last;
-            this.toAssignableObjectExpressionProp(prop, isBinding, isLast);
+            this.toAssignableObjectExpressionProp(
+              prop,
+              isBinding,
+              isLast,
+              allowAssign,
```
nicolo-ribaudo: Why is it passed down? Inside an object property, assignments are always allowed (for default props)

boweihan: The original approach needed to pass it down to "remember" that there had been a parenthesis somewhere up the stack - to invalidate expressions such as `[({ a: [b = 2]})] = t`. It's become more clear to me now that the presence of any parenthesis in the left-hand side of an assignment expression is the problem so I'll clean this up.

packages/babel-parser/test/fixtures/es2015/destructuring/parenthesized-lhs-array/options.json
```diff
@@ -1,3 +1,3 @@
 {
-  "throws": "You're trying to assign to a parenthesized expression, eg. instead of `([a]) = 0` use `([a] = 0)` (1:1)"
```
boweihan: We lose this specific error message by applying a more general fix but I think it's probably worth it in terms of code cleanliness - specific cases can be added later on.

boweihan:
@nicolo-ribaudo makes sense - I've updated the code. I think we can get away with not having an extra parameter in `toAssignable`.

The new logic covers most of the parens validation needed except for when the `createParenthesizedExpressions` parser option is set to `true` - github.com/babel/babel/pull/8025. There is separate validation for that code path which I have kept as-is (`unwrapParenthesizedExpression` in parser/expression.js).

packages/babel-parser/src/parser/lval.js
```diff
@@ -44,6 +44,14 @@ export default class LValParser extends NodeUtils {
     contextDescription: string,
   ): Node {
     if (node) {
+      if (
+        node.extra &&
+        node.extra.parenthesized &&
```
nicolo-ribaudo: Nit: you can use `node.extra?.parenthesized`

boweihan:
@nicolo-ribaudo @JLHwung originally I thought there might need to be a separate flow for handling the `createParenthesizedExpressions` parser option but I've figured out a more straightforward approach that works for both. Have added tests to cover each case.

packages/babel-parser/src/parser/lval.js
```diff
@@ -19,6 +19,12 @@ import { isStrictBindReservedWord } from "../util/identifier";
 import { NodeUtils } from "./node";
 import { type BindingTypes, BIND_NONE } from "../util/scopeflags";
 
+const unwrapParenthesizedExpression = (node: Node) => {
```
boweihan: this is needed to handle nested parenthesis when `createParenthesizedExpressions` is enabled. i.e. `[(((x)))]`

JLHwung: For nested parenthesis, `toAssignable` will be called with _parenthesizee_ recursively. Therefore we may check if `node` is one of `Identifier`, `MemberExpression` or `ParenthesizedExpression`.

JLHwung:
@boweihan Circle CI is flaky. I have rerun the build.

JLHwung:
@boweihan Well done! Thanks.

PR #10579, opened 2019-10-19
Implement PR workflow for running test262 on babel PRs

jbhoosreddy:
This PR introduces testing of babel transforms with test262 in every PR on each commit. It is a followup to https://github.com/babel/babel/pull/10556 which runs test262 tests on every commit to master. This PR utilizes changes to babel-test262-runner, here: (https://github.com/babel/babel-test262-runner/compare/399abd...b5eaa57)

A major aspect of the PR workflow is that it compares the test results on a PR against the test results on master. This is because there are frankly too many tests and many esoteric failing tests that to present them to a new contributor seems like an excessive barrier to entry for new contributions. Therefore, we have pivoted to compare the results from the previous master job and show only the results **that have changed**.

Due to @hzoo's suggestion to focus on testing only `/language` tests, and @JLHwung's babel config caching PR https://github.com/babel/babel-test262-runner/pull/12, we're able to run the tests in about 30m. This. while being a lot, is comparable to the existing job on TravisCI.

I also want to highlight authors' of some really cool and helpful packages that I was able to utilize which made this task much easier than implementing from scratch!

+ tap-mocha-reporter
+ tap-merge
+ make-tap-output
+ highland
+ @tap-format/parser

Also, also props to @nicolo-ribaudo for the quick turnaround on many PR reviews and help throughout this process!

| Fixed Issues?            | Relates to #4987 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,9 @@ aliases:
   - &artifact_test262_xunit
     path: ~/test-results
 
+  - &artifact_test262_diff_tap
+    path: ~/diff.tap
+
 executors:
   node-executor:
     docker:
@@ -73,6 +76,14 @@ jobs:
     executor: node-executor
     steps:
       - checkout
+      - run:
+          name: Sync with latest master branch (only on PRs)
+          command: |
+            if [ -n "$CIRCLE_PULL_REQUEST" ]
+            then
+              git fetch origin refs/pull/$CIRCLE_PR_NUMBER/merge
+              git checkout -qf FETCH_HEAD
+            fi
       - restore_cache: *restore-yarn-cache
       - restore_cache: *restore-node-modules-cache
       - run:
@@ -101,15 +112,33 @@ jobs:
             done
             node lib/download-node
       - run:
+          name: Download master branch Test262 artifact
+          command: node lib/download-master-artifact ~/master.tap
+          <<: *test262_workdir
+      - run:
           name: Run Test262
           command: node lib/run-tests I_AM_SURE | tee ~/test262.tap
           <<: *test262_workdir
       - store_artifacts: *artifact_test262_tap
       - run:
-          name: Output test262 results
+          name: Output Test262 results
           command: |
             cat ~/test262.tap | $(npm bin)/tap-mocha-reporter spec || true
           <<: *test262_workdir
+      - run:
+          name: Compare previous master branch & current job results
+          command: |
+            mkdir -p ~/test-results/test262
+            node lib/compare-results ~/master.tap ~/test262.tap | tee ~/diff.tap
+          <<: *test262_workdir
+      - store_artifacts: *artifact_test262_diff_tap
+      - run:
+          name: Output comparision results and report to CircleCI
+          command: |
+            mkdir -p ~/test-results/test262
+            cat ~/diff.tap | $(npm bin)/tap-merge | $(npm bin)/tap-mocha-reporter xunit | tee ~/test-results/test262/results.xml
+          <<: *test262_workdir
+      - store_test_results: *artifact_test262_xunit
 
   publish-verdaccio:
     executor: node-executor
@@ -136,7 +165,6 @@ workflows:
   test:
     jobs:
       - test
-
   master:
     jobs:
       - test262:
@@ -144,7 +172,13 @@ workflows:
             branches:
               only:
                 - master
-
+  test262:
+    jobs:
+      - approve-test262-run:
+          type: approval
+      - test262:
+          requires:
+            - approve-test262-run
   e2e:
     jobs:
       - publish-verdaccio
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -165,7 +165,7 @@ workflows:
   test:
     jobs:
       - test
-  master:
+  test262-master:
     jobs:
       - test262:
           filters:


jbhoosreddy:
There is an existing travis CI job which takes 21m to finish. Running all test262 language tests average 30 minutes. Given that, I'm inclining towards running them on every commit instead of requiring maintainers to trigger it manually. Thoughts @hzoo @nicolo-ribaudo ?

.circleci/config.yml
```diff
@@ -86,6 +96,7 @@ jobs:
       - run:
           name: Setup Test Runner
           command: |
+            # Switch to master branch later
```
JLHwung: Do we still need switching branch as long as we are comparing the artifacts now?

.circleci/config.yml
```diff
@@ -69,6 +72,13 @@ jobs:
       - image: circleci/node:12
     steps:
       - checkout
+      - run:
```
JLHwung: I would prefer leave it as-is as sometimes the PR branch may be conflict with master.

jbhoosreddy: So this one is interesting. Here's the problem:

Based on what we previously discussed, we're comparing with the previous master job. Now, if the PR is out of date, then it's not really clear what we're comparing against as the jobs could mean entirely different things if they are not at the same base.

To bring clarity to this situation, I chose to sync with the latest master. This opens us up to the possibility of the job failing if there is merge conflicts, but that's a failure case that'd be remediated anyway. As the author will fix this issue, and the failing job will simply point to the same issue.

But the benefit then, is that it is easy to compare the results and make sense of it..

I'm open to a different direction here

nicolo-ribaudo: Would it be possible to compare with the last common commit between the PR and master?

JLHwung: I have thought about this approach before but you will have to run test262 again since the artifact may be out dated if the master has fixed any broken tests.

Another approach may be running on the undocumented `/refs/pull/12345/merge` instead so we don't have to perform git _writing_ operation on CI.

nicolo-ribaudo: > I have thought about this approach before but you will have to run test262 again since the artifact may be out dated if the master has fixed any broken tests.

If it has been fixed in master, but not in the last common commit, than it will still be broken in the PR and it's status shouldn't be shown as changed.

> Another approach may be running on the undocumented /refs/pull/12345/merge instead so we don't have to perform git writing operation on CI.

Better than the current one :+1:

jbhoosreddy: So that I'm clear, what's `/refs/pull/12345/merge` and how can I use it?

jbhoosreddy: also,

> we don't have to perform git writing operation on CI

it's only a local git writing operation inside CI, afaik, it shouldn't affect the PR or any remote branch.

JLHwung: `/refs/pull/PULL_REQUEST_ID/merge` is a undocumented [refspec](https://git-scm.com/book/en/v2/Git-Internals-The-Refspec) on GitHub repositories. You may consider it as a pre-merged version of any pull request. Assume the pull request ID is `12345`, you may fetch and checkout such refs
```
git fetch origin refs/pull/12345/merge
git checkout -qf FETCH_HEAD
```

jbhoosreddy:
The test262 failure with 2 failure tests is a bit concerning as a) my PR shouldn't cause any existing tests to fail and b) it wasn't failing in my previous approach of syncing with master @JLHwung. I think we might need to explore this space a little bit more, or potentially I'm not using it correctly.

So I'm going to experiment by leaving the changeset as is, and manually rebase against master.

jbhoosreddy:
While this is running, I do want to note, that I don't think that I've completely mapped out all the error scenarios that can happen, and when you factor in the mechanics of running about 40,000 tests on two different branches, and situations like test case timeouts, and other weirdness, and then comparing them together, I can't guarantee that this MVP will not have false negatives. Certainly this can be improved over time.

Given the seemingly daunting difficulty in that, I've pivoted my approach to sort of *how can this be a useful source of information for the maintainers*. While it's a bit disappointing not to always get all green ticks (although I think it'll be green most of the time), I don't think failing the test262 job (in its current form) should block a PR, but rather be a signal for maintainers and contributors to inspect the results and see if they are related or not and make a judgement based on that.

We could circle back to running the test262 job only on maintainer approval. Or we can explore other options. I'll defer to the maintainers' judgement on what they'll find most helpful here. @nicolo-ribaudo @JLHwung @hzoo

nicolo-ribaudo:
@jbhoosreddy Would it be possible to run this workflow automatically on master, and with approval on PRs?

Running on master should always be safe, and it won't run on PRs by default since it's not guaranteed that it should pass.

.circleci/config.yml
```diff
@@ -69,6 +72,14 @@ jobs:
...
+          name: Sync with latest master branch (only on PRs)
+          command: |
+            # if [ -n "$CIRCLE_PULL_REQUEST" ]
+            # then
+            #   git fetch origin refs/$CIRCLE_BRANCH/merge
```
JLHwung: 🤔, from the CircleCI document [Builtin Environment Variables](https://circleci.com/docs/2.0/env-vars/?gclid=CjwKCAiAzanuBRAZEiwA5yf4utDULIqLsdnr437DHKlQo30KqzhOa5Nxk7DS6ipsCaUO_ovC8V9osRoCD4YQAvD_BwE#built-in-environment-variables), shouldn't it be
```
git fetch origin refs/pull/$CIRCLE_PR_NUMBER/merge
```

jbhoosreddy: circle branch was = pull/1234. I can swap to the other

jbhoosreddy:
> @jbhoosreddy Would it be possible to run this workflow automatically on master, and with approval on PRs?
>
> Running on master should always be safe, and it won't run on PRs by default since it's not guaranteed that it should pass.

Yes, absolutely! If that's the consensus, I'll update the PR to do that then.

.circleci/config.yml
JLHwung: That's okay if it works. Although it is a bit confused here since `pull/1234` is technically _not_ "the name of the Git branch currently being built", which should be something like `test262-pr-workflow`.

jbhoosreddy: I know what you mean.. Jenkins does the same thing here. weird

.circleci/config.yml
```diff
@@ -119,3 +148,10 @@ workflows:
             branches:
               only:
                 - master
+  test262:
+    jobs:
+      - approve-test262-run:
```
JLHwung: While `approval` grants contributor rights to run test262 on specific PR, I am wondering if it will end up with many Circle CI builds in the `ON HOLD` status but no one approves them at all.

The good news is that it does not count on job numbers so our new CircleCI job would not be blocked.

Ideally after we merge the PR into master, the regarding test262 job should be suspended if it is still on hold. Is it possible to do so?

.circleci/config.yml
```diff
@@ -69,6 +72,14 @@ jobs:
       - image: circleci/node:12
     steps:
       - checkout
+      - run:
+          name: Sync with latest master branch (only on PRs)
+          command: |
+            if [ -n "$CIRCLE_PULL_REQUEST" ]
```
nicolo-ribaudo: Is it possible to use [`branches`](https://circleci.com/docs/2.0/configuration-reference/#branches-2) or `filters` instead?

.circleci/config.yml
nicolo-ribaudo: I _think_ that Circle CI will clean up jobs of merged branches after some time anyway, but I'm not 100% sure.

.circleci/config.yml
jbhoosreddy: I don't know if we can configure them on dynamic refspecs though?

jbhoosreddy: I only see this sparse doc on this: https://circleci.com/docs/2.0/workflows/#using-regular-expressions-to-filter-tags-and-branches

nicolo-ribaudo: Oh ok then, it's probably not worth investigating for a single line.

jbhoosreddy: that was also motivator for this PR: https://github.com/babel/babel-test262-runner/pull/16

.circleci/config.yml
```diff
@@ -119,3 +148,10 @@ workflows:
             branches:
```
nicolo-ribaudo: Nit: could we rename this workflow to `update-test262-status`? `master` doesn't explain much.

.circleci/config.yml
JLHwung: We can merge it and watch.

nicolo-ribaudo:
Thanks 🎉

PR #10581, opened 2019-10-19
[action] Automatically add PR labels based on the modified files

nicolo-ribaudo:
| Fixed Issues?            | babel/babel-bot#1

This was an action I created a month ago and forgot to add to the repo :sweat_smile: 

Demo (private repo, sorry): https://github.com/babel/test-github-actions/pull/54

### DIFF
--- a/.github/workflows/pr-triage.yml
+++ b/.github/workflows/pr-triage.yml
@@ -1,0 +1,72 @@
+name: PR Triage
+
+on: pull_request
+
+jobs:
+  add_pkg_labels:
+    name: "Add \"pkg: *\" labels"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get corresponding labels
+        id: labels
+        uses: babel/actions/get-modified-files-labels@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          pull_number: ${{ github.event.pull_request.number }}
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-triage.yml
```diff
@@ -0,0 +1,73 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get corresponding labels
+        id: labels
+        uses: babel/actions/get-modified-files-labels@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          pull_number: ${{ github.event.pull_request.number }}
+          labels: >
+            {
```
JLHwung: Could it be extracted as a dedicated JSON/yaml configuration? It could be similar to what it is done in [Labeller](https://github.com/actions/labeler#create-githublabeleryml). 

By doing so It will benefit from better error check / prettier styling and I would expect it can be more visible to contributors.

nicolo-ribaudo: Oh well, I could probably use Labeller directly

nicolo-ribaudo:
This is not going to work, because GitHub doesn't pass a token with the correct permissions to Actions run from a fork. Labeler has the same problem (https://github.com/WordPress/gutenberg/issues/17324)

PR #10589, opened 2019-10-22
corrected the grammar of option descriptions

m-ahmadi:
corrected some grammar typos of commander option descriptions. (punctuation/capitalization/etc)

### DIFF
--- a/packages/babel-cli/src/babel/options.js
+++ b/packages/babel-cli/src/babel/options.js
@@ -101,6 +101,7 @@ commander.option(
 // Config params for certain module output formats.
 commander.option(
   "--module-root [filename]",
+  // eslint-disable-next-line max-len
   "Optional prefix for the AMD module formatter that will be prepended to the filename on module definitions.",
 );
 commander.option("-M, --module-ids", "Insert an explicit id for modules.");
--- a/packages/babel-cli/src/babel/options.js
+++ b/packages/babel-cli/src/babel/options.js
@@ -12,19 +12,19 @@ import pkg from "../../package.json";
 // Standard Babel input configs.
 commander.option(
   "-f, --filename [filename]",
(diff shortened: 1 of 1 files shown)


packages/babel-cli/src/babel/options.js
```diff
@@ -40,114 +40,114 @@ commander.option(
...
   "--source-root [filename]",
-  "the root from which all sources are relative",
+  "The root from which all sources are relative.",
 );
 
 // Config params for certain module output formats.
 commander.option(
   "--module-root [filename]",
-  "optional prefix for the AMD module formatter that will be prepend to the filename on module definitions",
+  "Optional prefix for the AMD module formatter that will be prepended to the filename on module definitions.",
```
JLHwung: Could you add `// eslint-disable-next-line max-len` above this line?

JLHwung:
CI failure has been fixed on master. Thus it is not relevant.

packages/babel-cli/src/babel/options.js
```diff
@@ -40,114 +40,115 @@ commander.option(
...
-  "retain line numbers - will result in really ugly code",
+  "Retain line numbers. This will result in really ugly code.",
 );
 commander.option(
   "--compact [true|false|auto]",
-  "do not include superfluous whitespace characters and line terminators",
+  "Do not include superfluous whitespace characters and line terminators.",
   booleanify,
 );
 commander.option("--minified", "save as much bytes when printing [true|false]");
```
m-ahmadi: Dang it, I missed this line.
@JLHwung Could you replace this description with `"Save as many bytes when printing [true|false]"` in a future commit?

JLHwung: Ah, could you submit another PR?

m-ahmadi: OK.

PR #10591, opened 2019-10-22
chore: test against Node.js 13

JLHwung:
| Any Dependency Changes?  | announce Node.js 13 support

Run CI against Node.js 13. Fixes may follow if necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,8 @@ matrix:
   include:
     - node_js: "node"
       env: JOB=lint
-  # We test the latest version on circleci
-    - node_js: "11"
+  # We test the Node.js 12 on circleci
+    - node_js: "13"
   # Move `windows` build to be the third since it is slow
     - os: windows
       node_js: "node"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -27,8 +27,8 @@ matrix:
   include:
     - node_js: "node"
       env: JOB=lint
-  # We test the latest version on circleci
-    - node_js: "11"
+  # We test the Node.js 12 on circleci
```
existentialism: Why not run 13 on Circle and make this 12?

JLHwung: Because `circleci/node:13` is [not available](https://circleci.com/docs/2.0/circleci-images/#nodejs) yet. After it is available we could swap 13 to circleci.

existentialism: Gotcha, fwiw, I'm okay with waiting :)

nicolo-ribaudo:
Lets merge the update to package.json update ASAP since CI is currently broken, and we will add it to circle ci as soon as it is available (which shouldn't take too much anyway)

nicolo-ribaudo:
Let's merge this now since CI is completely broken; we can move the tests to CircleCI as soon as it is available.

PR #10592, opened 2019-10-22
Allow TypeScript type assertions in array destructuring

### LINKED ISSUE #10066: [TypeScript] SyntaxError on array destructuring
## Bug Report

**Current Behavior**
SyntaxError: Unexpected type cast in parameter position

**Input Code**
```js
[x as Foo] = arr
```

**Expected behavior/code**
no error

@babel/parser with typescript plugin

**Environment**
- Babel version(s): 7.4.5
- Node/npm version: Node 11.13 / npm 6.7

SakibulMowla:
| Fixed Issues?            | Fixes #10066 
| Patch: Bug Fix?         | Yes
| Documentation PR Link    | 

### DIFF
--- a/packages/babel-parser/src/plugins/typescript/index.js
+++ b/packages/babel-parser/src/plugins/typescript/index.js
@@ -2427,10 +2427,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
     toAssignableList(
       exprList: N.Expression[],
       isBinding: ?boolean,
-      contextDescription: string,
     ): $ReadOnlyArray<N.Pattern> {
-      const isAssignmentExpression =
-        contextDescription === "assignment expression";
       for (let i = 0; i < exprList.length; i++) {
         const expr = exprList[i];
         if (!expr) continue;
@@ -2440,7 +2437,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
             break;
           case "TSAsExpression":
           case "TSTypeAssertion":
-            if (isAssignmentExpression) {
+            if (!isBinding) {
               exprList[i] = this.typeCastToParameter(expr);
             } else {
               this.raise(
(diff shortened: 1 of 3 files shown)


packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -2365,7 +2365,13 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
     }
 
-    toAssignableList(exprList: N.Expression[]): $ReadOnlyArray<N.Pattern> {
+    toAssignableList(
+      exprList: N.Expression[],
+      isBinding: ?boolean,
+      contextDescription: string,
+    ): $ReadOnlyArray<N.Pattern> {
+      const isAssignmentExpression =
+        contextDescription === "assignment expression";
```
nicolo-ribaudo: I would prefer not to rely on the exact value of this string, since it's only used for error messages. Could we use `!isBinding` instead of `isAssignmentExpression`?

packages/babel-parser/test/fixtures/typescript/cast/destructure-and-assign/input.ts
```diff
@@ -0,0 +1 @@
+(a as number) = 42;
```
existentialism: Shouldn't this be:

```
[a as number] = 42
```

?

packages/babel-parser/src/plugins/typescript/index.js
SakibulMowla: Highly agree on not using the value of the string. What does `isBinding` represent? Can you point me to any doc or link?

packages/babel-parser/test/fixtures/typescript/cast/destructure-and-assign/input.ts
SakibulMowla: It sure should be. Was playing around with input and forgot to change it back. Thanks for noticing!

PR #10582, opened 2019-10-21
[parser] Allow optional async methods

gonzarodriguezt:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #10525 
| Patch: Bug Fix?          | Yes
| Major: Breaking Change?  | No
| Minor: New Feature?      | No
| Tests Added + Pass?      | Yes
| Any Dependency Changes?  | No
| License                  | MIT

async methods were not parsing post member name modifiers, causing `?` to be an unexpected token

PR #10594, opened 2019-10-23
[parser] Parse only modifiers of actual methods

gonzarodriguezt:
| Fixed Issues?            | Follow-up from #10582 comment 

i've did what in my opinion was the most trivial solution and added a test as an start point, reviews are highly appreciated

### DIFF
--- a/packages/babel-parser/src/parser/statement.js
+++ b/packages/babel-parser/src/parser/statement.js
@@ -1348,9 +1348,9 @@ export default class StatementParser extends ExpressionParser {
     const isPrivate = key.type === "PrivateName";
     // Check the key is not a computed expression or string literal.
     const isSimple = key.type === "Identifier";
+    const maybeQuestionTokenStart = this.state.start;
 
-    // "async" should not be treated as normal porperty names
-    if (key.name !== "async") this.parsePostMemberNameModifiers(publicMember);
+    this.parsePostMemberNameModifiers(publicMember);
 
     if (this.isClassMethod()) {
       method.kind = "method";
@@ -1403,6 +1403,10 @@ export default class StatementParser extends ExpressionParser {
     ) {
       // an async method
       const isGenerator = this.eat(tt.star);
+
+      if (publicMember.optional) {
+        this.unexpected(maybeQuestionTokenStart);
+      }
 
       method.kind = "method";
       // The so-called parsed name would have been "async": get the real name.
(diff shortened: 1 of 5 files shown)


packages/babel-parser/src/parser/statement.js
```diff
@@ -1349,7 +1349,8 @@ export default class StatementParser extends ExpressionParser {
     // Check the key is not a computed expression or string literal.
     const isSimple = key.type === "Identifier";
 
-    this.parsePostMemberNameModifiers(publicMember);
+    // "async" should not be treated as normal porperty names
+    if (key.name !== "async") this.parsePostMemberNameModifiers(publicMember);
```
JLHwung: `async` could be a class property, i.e.
```ts
class A {
  async?: boolean
}
```

gonzarodriguezt: Wow nice, thank you

gonzarodriguezt: hey @JLHwung i was trying to follow this suggestion, but i'm not sure of the proper way to check if a token was eaten, this is what i've tried:

```javascript
let token;
if ((token = this.state.tokens.find(s => s.type === tt.question))) {
  this.unexpected(token.start);
}
```

with this modification to `parsePostMemberNameModifiers`

```javascript
parsePostMemberNameModifiers(
  methodOrProp: N.ClassMethod | N.ClassProperty,
  saveToken: boolean,
): void {
  this.options.tokens = saveToken;
  const optional = this.eat(tt.question);
  if (optional) methodOrProp.optional = true;
} 
```
the problem with that is that previous tests are failing because the absence of the property `options.tokens` in their `output.json` files

thanks you! 🙏

JLHwung:
```js
if ((token = this.state.tokens.find(s => s.type === tt.question))) {
  this.unexpected(token.start);
}
```
I am afraid we may not iterate `this.state.tokens` here because it may run out of the public member method scope, i.e. when parsing `foo()` it would throw on the previous consumed `?` token, which is incorrect.
```
class A {
  async?: boolean,
  async foo(): Promise<void>
}
```

You may consider to log the `maybeQuestionTokenStart = this.state.start` before the first `parsePostMemberNameModifiers` call. After that, if we are in an `async` method later, we may check if `publicMethod.optional` is true. If it is, then it must come from the `?` token after `async` keyword, and now we should throw at `maybeQuestionTokenStart`.

gonzarodriguezt: This is awesome, thanks a lot for taking the time and the explanation, i've updated the PR

JLHwung:
@gonzarodriguezt Could you rebase on master?

gonzarodriguezt:
@JLHwung Sorry i was afk!

It's a pleasure @nicolo-ribaudo 🙂

PR #10599, opened 2019-10-23
Add support for .cjs config files

### LINKED ISSUE #10595: Loading .cjs config files on Node 13
## Feature Request

If you have Node 13 and `babel.config.js` file in a package with `type=module` then Node will emit a warning:

```text
(node:22720) Warning: require() of ES modules is not supported.
require() of /home/spyke/Projects/undercut/packages/undercut/babel.config.js from /home/spyke/Projects/undercut/node_modules/@babel/core/lib/config/files/configuration.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename babel.config.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /home/spyke/Projects/undercut/packages/undercut/package.json.
```

The solution should be renaming `babel.config.js` into `babel.config.cjs`, but it isn't supported by Babel yet.

I've considered ignoring this warning, but it is annoying.

The solution has no braking changes. Documentation changes are minimal: adding another possible name for config file.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #10595
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | babel/website#2106 
| Any Dependency Changes?  |

The three commits should be reviewed one by one:
- 6c80fa1 refactors a bit the config loading logic, since the check to disallow duplicated files was triplicated in three different places
- 6aa196c adds support for `babel.config.cjs` and `.babelrc.js`
- db4f8a4 adds the necessary tests and refactors them to use `test.each` (since we are testing the same things for each config file)

I'm marking this PR as `High Priority` because we need it to support using Babel in packages using `"type": "module"` in their package.json. Note that while ES modules are still experimental in Node 13.0.0, they will be enabled by default in 13.1.0 (https://twitter.com/jasnell/status/1184265267951493120).


packages/babel-core/test/config-chain.js
```diff
@@ -14,6 +15,24 @@ function loadOptions(opts) {
...
+    }
+  }
+  return pairs;
+}
+
+async function getTemp(name) {
+  const cwd = await fs.promises.mkdtemp(os.tmpdir() + path.sep + name);
+  const tmp = name => path.join(cwd, name);
+  const config = name =>
+    fs.promises.copyFile(fixture("config-files-templates", name), tmp(name));
```
JLHwung: `fs.promises` is not supported until 10.x

nicolo-ribaudo: Oh right 🤦‍♂️

nicolo-ribaudo:
Docs added to https://github.com/babel/website/pull/2106

packages/babel-core/src/config/files/configuration.js
```diff
@@ -18,26 +18,22 @@ import type { CallerMetadata } from "../validation/options";
 
 const debug = buildDebug("babel:config:loading:files:configuration");
 
-const BABEL_CONFIG_JS_FILENAME = "babel.config.js";
-const BABEL_CONFIG_JSON_FILENAME = "babel.config.json";
 const ROOT_CONFIG_FILENAMES = [
-  BABEL_CONFIG_JS_FILENAME,
-  BABEL_CONFIG_JSON_FILENAME,
+  "babel.config.js",
```
yordis: What about `.mjs`? Would that be an issue in the future?

nicolo-ribaudo: We don't support configurations which must be loaded asynchronously (like es modules) yet.

nicolo-ribaudo:
@fisker In our tests there isn't that warning :thinking:

https://travis-ci.com/babel/babel/jobs/250695663

JLHwung:
@fisker `fs.promises` is marked as stable in v10.17.0: https://github.com/nodejs/node/commit/21604db52f5a8cf5f0e1dd564a53794be2b530a0

PR #10604, opened 2019-10-24
Nit: Use ?.  in the example used by the proposal readme

nicolo-ribaudo:

The only reason for this PR is that I noticed that these lines are used in the readme of the optional chaining proposal as a use case for optional chains after `delete`. :stuck_out_tongue:

https://github.com/tc39/proposal-optional-chaining#optional-deletion (click on `Why do we support optional deletion?`)

(cc @claudepache, in case you'll want to update the link in the readme)

### DIFF
--- a/packages/babel-helper-builder-react-jsx/src/index.js
+++ b/packages/babel-helper-builder-react-jsx/src/index.js
@@ -91,9 +91,7 @@ You can turn on the 'throwIfNamespace' flag to bypass this warning.`,
       value.value = value.value.replace(/\n\s+/g, " ");
 
       // "raw" JSXText should not be used from a StringLiteral because it needs to be escaped.
-      if (value.extra && value.extra.raw) {
-        delete value.extra.raw;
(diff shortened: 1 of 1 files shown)


packages/babel-helper-builder-react-jsx/src/index.js
```diff
@@ -91,9 +91,7 @@ You can turn on the 'throwIfNamespace' flag to bypass this warning.`,
       value.value = value.value.replace(/\n\s+/g, " ");
 
       // "raw" JSXText should not be used from a StringLiteral because it needs to be escaped.
-      if (value.extra && value.extra.raw) {
-        delete value.extra.raw;
-      }
+      delete value.extra?.raw;
```
JLHwung: Although `delete value.extra?.raw` is equivalent to
```js
value.extra == null ? true : delete value.extra.raw
```
it looks good here to remove `value.extra.raw` even it is nullish.

PR #10616, opened 2019-10-30
[BUGFIX] Allows legacy decorators to work with strict class fields

### LINKED ISSUE #9779: Inconsistent field assignment for Legacy/Stage 1 Decorators
## Bug Report

**Current Behavior**

Currently, the legacy decorators transform requires the class fields to be configured to `loose` mode. In loose mode, class fields are assigned using _set_ semantics, instead of `Object.defineProperty`. This makes sense - in TS, fields are assigned with _set_ semantics, even when they are decorated - and is referenced frequently in the current discussion about the downsides of stage 1 decorators, and the differences between them and stage 2-3 decorators.

However, it seems that currently, fields which are decorated [actually _are_ set using `Object.defineProperty`](https://github.com/babel/babel/blob/master/packages/babel-helpers/src/helpers.js#L982-L991).

**Input Code**
- The REPL currently doesn't seem to support compiling with class fields in loose mode to demonstrate, but [here is the code with them in strict, with the error/warning](https://babeljs.io/repl#?babili=false&browsers=Last%201%20Chrome%20Version&build=&builtIns=false&spec=false&loose=true&code_lz=MYGwhgzhAEBiD29oG8BQ1oAEAmBTY8ATmAC67QBGYh0AvNAIwBMAzANzTqcZUBedjVm1QBfIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=stage-1%2Cenv&prettier=true&targets=Node-8.9&version=7.4.2)

```js
class Foo {
  @decorate bar = 123; 
  
  baz = 123;
}
```

**Expected behavior/code**

Class field assignment should always be consistent. Both decorated and undecorated fields should be defined using `Object.defineProperty`, or _set_.

```js
{
  "plugins": [
    ["@babel/plugin-proposal-decorators", { legacy: true }],
    ["@babel/plugin-proposal-class-properties", { loose: true }]
  ]
}
```

**Environment**
- Babel version(s): v7.4.2
- OS: OSX 10.14.3
- Monorepo: N/A
- How you are using Babel: `ember-cli-babel`

**Possible Solution**

One possible solution would be to change the decorators transform to assign values using _set_, but that seems like it could be a breaking change. Alternatively, we could remove the restriction requiring `loose` mode for class properties, and have them be assigned with `Object.defineProperty` in a spec compliant way. Then it will be consistent. 

Also, apologies if I'm missing anything nuanced here, and this behavior is required for other reasons 😅

cc @nicolo-ribaudo, was told you'd be interested in this

pzuraq:
| Fixed Issues?            | Fixes #9779
| Minor: New Feature?      | Sort of
| Documentation PR Link    | 

This PR allows legacy decorators to work with strict class fields, which
are now stage 3 and have shipped in a number of browsers. Allowing this will
allow users of the legacy transform (which is currently recommended by the
[champions of the decorator proposal](https://github.com/tc39/proposal-decorators#how-should-i-use-decorators-in-transpilers-today)) to use the proper class field semantics
for non-decorated fields, which should help prevent breakage later on.

This change is not a breaking change, since users had to explicitly opt into
loose mode in class fields before. This just gives them the option to remove
that opt-in.

### DIFF
--- a/packages/babel-plugin-proposal-decorators/src/transformer-legacy.js
+++ b/packages/babel-plugin-proposal-decorators/src/transformer-legacy.js
@@ -287,4 +287,24 @@ export default {
       ]),
     );
   },
+
+  CallExpression(path, state) {
+    if (path.node.arguments.length !== 3) return;
+    if (!WARNING_CALLS.has(path.node.arguments[2])) return;
+
+    // If the class properties plugin isn't enabled, this line will add an unused helper
+    // to the code. It's not ideal, but it's ok since the configuration is not valid anyway.
+    if (path.node.callee.name !== state.addHelper("defineProperty").name) {
+      return;
+    }
+
+    path.replaceWith(
+      t.callExpression(state.addHelper("initializerDefineProperty"), [
+        t.cloneNode(path.get("arguments")[0].node),
+        t.cloneNode(path.get("arguments")[1].node),
+        t.cloneNode(path.get("arguments.2.arguments")[0].node),
+        t.cloneNode(path.get("arguments.2.arguments")[1].node),
+      ]),
+    );
+  },
 };
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -1033,9 +1033,7 @@ helpers.initializerWarningHelper = helper("7.0.0-beta.0")`
     export default function _initializerWarningHelper(descriptor, context){
         throw new Error(
           'Decorating class property failed. Please ensure that ' +
-          'proposal-class-properties is enabled and set to use loose mode. ' +
-          'To use proposal-class-properties in spec mode with decorators, wait for ' +
-          'the next major version of decorators in stage 2.'
+          'proposal-class-properties is enabled and runs after the decorators transform.'
         );
     }
 `;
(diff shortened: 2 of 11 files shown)


pzuraq:
cc @littledan @nicolo-ribaudo

pzuraq:
Failures appear to be unrelated to this change in particular, though I'm not entirely sure

packages/babel-plugin-proposal-decorators/src/transformer-legacy.js
```diff
@@ -287,4 +287,19 @@ export default {
       ]),
     );
   },
+
+  CallExpression(path, state) {
+    // insert assignment if class properties are in strict mode
+    if (path.node.callee.name !== "_defineProperty") return;
```
nicolo-ribaudo: This is not safe. If a user already has a `_defineProperty` binding in their code, Babel will use another name.

The check should look more like:

```js
if (path.node.arguments.length !== 2) return;
if (!WARNING_CALLS.has(path.node.arguments[2])) return;

// If the class properties plugin isn't enabled, this line will add an unused helper
// to the code. It's not ideal, but it's ok since the configuration is not valid anyway.
if (path.node.callee.name !== state.addHelper("defineProperty").name) return;
```

nicolo-ribaudo:
Yes CI failures are unrelated.

pzuraq:
@nicolo-ribaudo is there anything else that needs to be done to get this merged? Do we need to wait until CI is green again?

endel:
Hi there, thanks for this. Is this patch available on NPM already? 👀

JLHwung:
@endel We are preparing on 7.7 release.

PR #10619, opened 2019-10-31
chore: remove output-file-sync dependency

JLHwung:
| Any Dependency Changes?  | Yes, removed `output-file-sync` dependency

This PR replaces `output-file-sync` by `make-dir` + `fs.writeFileSync`. Under the hood `output-file-sync` uses [`mkdirp` + `fs.writeFileSync`](https://github.com/shinnn/output-file-sync/blob/2b7664b027fae76a76de5f58947ed97e151d2ba6/index.js#L111-L118) and adds options checking. As we didn't use options in our codebase, it's safe to replace `output-file-sync`.

### DIFF
--- a/packages/babel-cli/src/babel/dir.js
+++ b/packages/babel-cli/src/babel/dir.js
@@ -1,7 +1,6 @@
 // @flow
 
 import defaults from "lodash/defaults";
-import outputFileSync from "output-file-sync";
 import { sync as makeDirSync } from "make-dir";
 import slash from "slash";
 import path from "path";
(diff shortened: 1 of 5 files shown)


packages/babel-cli/src/babel/dir.js
```diff
@@ -10,6 +9,11 @@ import fs from "fs";
 import * as util from "./util";
 import { type CmdOptions } from "./options";
 
+function outputFileSync(filePath: string, data: string | Buffer): void {
```
hg-pyun: Seems good.

PR #10621, opened 2019-10-31
throw a TypeError if identifier validation fails

dentrado:
| Fixed Issues?            | 
| Documentation PR Link    | 

I noticed that `t.identifier(name)` accepts invalid identifiers, this PR fixes that bug. Thanks to @JLHwung for [finding the reason](https://babeljs.slack.com/archives/C0DFJT81H/p1572540253206000?thread_ts=1572537708.205200&cid=C0DFJT81H) so quickly!


JLHwung:
@jridgewell We do have plan to refactor it. In the near future babel-types will not depend on `esutils` anymore, instead it will reuse the big identifier name regex in babel-parser.

PR #10628, opened 2019-11-02
Don't throw when destructuring into a var named as an import

### LINKED ISSUE #10627: Invalid read-only when using destructuring shadowed by import
## Bug Report

**Current Behavior**
The commonjs and amd plugins create invalid `throw new Error('"' + "foo" + '" is read-only.');` statements in the output when there is an import with the same name as used in the destructuring even though it's shadowed by a function parameter.

**Input Code**
- REPL link: https://babeljs.io/en/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=JYWwDg9gTgLgBAbzgMwhOBfFUIjgIgDoB6AD3wG4AoK5AVwDsBjGYCBuGAUwGcYAKVBACUiKnDhN2PCABsuhWRADmgtMOoSA2kIC6cALxwtARl2a4_BEKxGkQgFxwTmDeMnS5CpaqFuMNNx8_AAMbkA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=false&targets=&version=7.6.4&externalPlugins=%40babel%2Fplugin-transform-modules-commonjs%407.6.0

```js
import { foo } from "./x";

function test(foo) {
  console.log(foo);
  [foo] = [1];
  ({foo} = { foo: 1 });
  console.log(foo);
}

test(0);
```

Output:

```js
"use strict";

var _x = require("./x");

function test(foo) {
  console.log(foo);
  [foo] = ([1], function () {
    throw new Error('"' + "foo" + '" is read-only.');
  }());
  ({
    foo
  } = ({
    foo: 1
  }, function () {
    throw new Error('"' + "foo" + '" is read-only.');
  }()));
  console.log(foo);
}

test(0);
```

**Expected behavior/code**
No Error's in output, like this when the import name is changed:

```js
"use strict";

var _x = require("./x");

function test(foo) {
  console.log(foo);
  [foo] = [1];
  ({
    foo
  } = {
    foo: 1
  });
  console.log(foo);
}

test(0);
```

```js
  "babel": {
    "plugins": [
      "@babel/plugin-transform-modules-commonjs"
    ]
  }
```

**Environment**
- Babel version(s): v7.6.4
- Node/npm version: Node 12
- OS: OSX
- Monorepo: no
- How you are using Babel: cli or REPL

**Possible Solution**

nicolo-ribaudo:
| Fixed Issues?            | Fixes #10627
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-helper-module-transforms/src/rewrite-live-references.js
+++ b/packages/babel-helper-module-transforms/src/rewrite-live-references.js
@@ -224,6 +224,10 @@ const rewriteReferencesVisitor = {
       seen.add(path.node);
 
       const left = path.get("left");
+
+      // No change needed
+      if (left.isMemberExpression()) return;
+
       if (left.isIdentifier()) {
         // Simple update-assign foo += 1; export { foo };
         // =>   exports.foo =  (foo += 1);
@@ -234,9 +238,9 @@ const rewriteReferencesVisitor = {
           return;
         }
 
-        const exportedNames = exported.get(localName) || [];
+        const exportedNames = exported.get(localName);
         const importData = imported.get(localName);
-        if (exportedNames.length > 0 || importData) {
+        if (exportedNames?.length > 0 || importData) {
           assert(path.node.operator === "=", "Path was not simplified");
 
           const assignment = path.node;
@@ -259,13 +263,14 @@ const rewriteReferencesVisitor = {
           );
           requeueInParent(path);
         }
-      } else if (left.isMemberExpression()) {
-        // No change needed
       } else {
         const ids = left.getOuterBindingIdentifiers();
-        const id = Object.keys(ids)
-          .filter(localName => imported.has(localName))
-          .pop();
+        const programScopeIds = Object.keys(ids).filter(
+          localName =>
+            scope.getBinding(localName) === path.scope.getBinding(localName),
+        );
+        const id = programScopeIds.find(localName => imported.has(localName));
+
         if (id) {
           path.node.right = t.sequenceExpression([
             path.node.right,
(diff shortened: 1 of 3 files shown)


packages/babel-plugin-transform-modules-commonjs/test/fixtures/misc/import-shadowed-assign/output.js
```diff
@@ -0,0 +1,23 @@
...
+
+function f(foo) {
+  foo = 2;
+  [foo] = [];
+  ({
+    foo
+  } = {});
+}
+
+_x.foo = (2, function () {
```
nicolo-ribaudo: In a follow up PR, this could be left as `foo` since we are throwing anyway.

packages/babel-helper-module-transforms/src/rewrite-live-references.js
```diff
@@ -259,13 +263,14 @@ const rewriteReferencesVisitor = {
...
-      } else if (left.isMemberExpression()) {
-        // No change needed
       } else {
         const ids = left.getOuterBindingIdentifiers();
-        const id = Object.keys(ids)
-          .filter(localName => imported.has(localName))
-          .pop();
+        const programScopeIds = Object.keys(ids).filter(
+          localName =>
+            scope.getBinding(localName) === path.scope.getBinding(localName),
```
JLHwung: nit: consider to guard with `scope.hasBinding` to avoid unnecessary `path.scope.getBinding`.

nicolo-ribaudo: I tried to do it and found a bug which, if fixed, would break `@babel/plugin-transform-runtime` <= 7.6 :cry: 

This is what currently happens:
1) The imports are analyzed
2) The imports are removed -> The bindings are removed
3) References to imported bindings are rewritten (this file)
    i) When we are comparing the bindings, the global one is `undefined`
    ii) If the local binding is `undefined`, it hasn't been redefined.

I tried to swap 2 and 3, but I still can optimize away the check for the local binding because `@babel/transform-runtime` doesn't register the import it injects in the scope tracker.

PR #10639, opened 2019-11-04
Add CodeSandbox CI Config

CompuIves:
After I've heard some interest from people at Babel about CodeSandbox CI I decided to start fiddling around with it. We've made some changes to CodeSandbox CI to make it work, and this is the result! You can see the service in action here: https://github.com/CompuIves/babel/pull/1.

To have a better idea of what CodeSandbox CI does you can read more about it here: https://u2edh.csb.app/. I recommend merging this PR first before installing the GitHub App, to ensure that all future PRs will work from the get-go.

The gist of it is:

The goal of this app is to make it easier for you to test your library without publishing to npm yet. Since CodeSandbox is already used for bug reports we wanted to make it possible to also test the PR version of your library with the existing bug reports.

In an ideal flow, it would work like this:

Someone opens an issue for a bug with a sandbox reproducible
Someone opens a fix PR mentioning the issue
We build the library from the PR, fork the sandbox repro and install the new library in that sandbox
This way you will only have to open the generated sandbox to confirm that the fix works. No need to clone, install or test locally.

I'm curious what you think of this! Here's a generated sandbox as reference, it's a mini Babel REPL that you can use to play with the version of the library at that moment: https://codesandbox.io/s/babel-playground-44xtv

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,0 +1,5 @@
+{
+  "installCommand": "bootstrap",
+  "buildCommand": false,
+  "sandboxes": ["kypop"]
+}
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "private": true,
   "license": "MIT",
   "scripts": {
+    "bootstrap": "make bootstrap",
     "build": "make build",
     "fix": "make fix",
     "lint": "make lint",


CompuIves:
Really interesting here is that we build and publish all 151 packages referenced by Lerna, this way you can also test all these packages individually! That was the big test for us 😄.

thiagoarrais:
Nice work, @CompuIves! Thanks for the PR and for Codesandbox itself. It'll be an excellent addition to the Babel toolbox and I'll try to get this merged as soon as possible.

Do you have any docs on the `ci.json` contents? As I understand, `installCommand` is something that is supposed to install deps and `buildCommand` is supposed to build the lib itself. Is this correct? 

What about `sandboxes`? Is this the name for the base codesandbox where the pull request sandboxes will be based from? Why an array?

thiagoarrais:
Found out about https://u2edh.csb.app/docs.html just after publishing. But it seems like `installCommand` isn't covered.

CompuIves:
Oh yes, that's true! I've added `installCommand` specifically for babel, I still need to update the docs. Will do that now!

CompuIves:
I've updated the docs with more info! In this case I've set the `installCommand` to `bootstrap`,  because `bootstrap` installs dependencies. However, it also builds the library already, so I've decided to set `buildCommand` to `false`, which makes it skip the build.

existentialism:
❤️ @CompuIves!

CompuIves:
We're about to improve our uploading of packages next week by moving the infra to GCP. Would it be possible to hold off with merging until Wednesday? Then it should be faster than now 😄

nicolo-ribaudo:
Definitely! We are really good at not merging PRs 😛

Btw, would it be possible to run Babel using node rather than using standalone? Then it would be possible to have a proper babel.config.js file.

I tried to do it myself (using Babel in the `"start"` command), but it's the first time I'm using CodeSandbox and I couldn't figure it out.

CompuIves:
> Btw, would it be possible to run Babel using node rather than using standalone? Then it would be possible to have a proper babel.config.js file.

Yep! I think that should work too. With CodeSandbox we have two kinds of projects, sandboxes and containers. Sandboxes are run in the browser, containers are run on a node server. The sandbox that I'm using as example here is run in the browser, which means that `yarn start` et al don't work. The advantage of running in the browser is that the load time is super fast, there is no node container that needs to spin up. The disadvantage is that not all "local" code will work.

In this case `@babel/preset-env` and all the plugins don't need node specifically, I was thinking that `@babel/core` might need some node functionality, so I went with `babel-standalone` as the initial implementation. We do stub `fs` and allow reading from other sandbox files though, so it might work.

We're releasing CodeSandbox CI today, and we're now ready to merge this PR to master, now that we've improved our uploading mechanics (to not upload the packages sequentially, but as one tar). Let me know if anything else is needed!

Here's the sandbox example using `@babel/core`. The caveat is that you need to `require` what you need in the config in one of the files, we use this to determine which files we need to download so we don't download all node_modules but only the files we need: https://codesandbox.io/s/babel-playground-0bkuw.

Let me know if you'd like to use the example of `@babel/core` instead of `babel-standalone`. I can also include both if that would be nicer 😄.

PR #10640, opened 2019-11-04
Remove import bindings after rewriting live references

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  |

Ref: https://github.com/babel/babel/pull/10628#discussion_r342282695

### DIFF
--- a/packages/babel-helper-module-transforms/src/rewrite-live-references.js
+++ b/packages/babel-helper-module-transforms/src/rewrite-live-references.js
@@ -5,6 +5,13 @@ import simplifyAccess from "@babel/helper-simple-access";
 
 import type { ModuleMetadata } from "./";
 
+function isSameBinding(scope1, scope2, name) {
+  const binding = scope1.getBinding(name);
+  // TODO: In Babel 8, check if the next code can be uncommented
+  // https://github.com/babel/babel/pull/10628#discussion_r342282695
+  return /* !!binding && */ binding === scope2.getBinding(name);
+}
+
 export default function rewriteLiveReferences(
   programPath: NodePath,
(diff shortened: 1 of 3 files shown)


packages/babel-helper-module-transforms/src/rewrite-live-references.js
```diff
@@ -171,11 +178,8 @@ const rewriteReferencesVisitor = {
 
     const localName = path.node.name;
 
-    const localBinding = path.scope.getBinding(localName);
-    const rootBinding = scope.getBinding(localName);
-
     // redeclared in this scope
-    if (rootBinding !== localBinding) return;
+    if (!isSameBinding(scope, path.scope, localName)) return;
```
JLHwung: Consider add it to a scope method, such as
```js
scope.hasSameBinding(localName, otherScope)
```
because the `// redeclared in this scope` check is also used in `helper-simple-access` and `transform-modules-commonjs`.

nicolo-ribaudo: Then it wouldn't work with `@babel/core` <= 7.6.0. Let's defer this PR to Babel 8 since it's not urgent.

JLHwung: 🤔Doesn't uncommenting `!!binding` cause it fail with `@babel/core` <= 7.6.0?

> Let's defer this PR to Babel 8 since it's not urgent.

Yup.

nicolo-ribaudo: Also relying on `scope.hasSameBinding`, since in older versions it's undefined.

PR #10642, opened 2019-11-04
Ignore convert-source-map module in browser builds, fixes @babel/core for browsers

tajo:
| Fixed Issues?            |  
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 

I am using `transformFromAstSync` function from `@babel/core` to compile a code in the browser. The issue is that bundler (webpack in my case) eventually reaches the import of `convert-source-map` module which further imports `fs` and that results in an error since the `fs` module is not present in the browser environment.

There a few workarounds / solutions:
- significantly refactor the `@babel/core` so `convert-source-map` is not imported anymore (it's needed only when dealing with sourceMaps)
- webpack config allows to mock out modules through `config.node = { fs: 'empty' }` but that adds a lot of friction on the consumer side
- all popular bundlers understand `browser` field in the package.json, so we can use it to ignore the  `convert-source-map` import

This PR uses the browser field since it doesn't require any code modifications.

(I was also trying to fork `@babel/core` in meantime but the problem is that every single preset (plugin) imports `@babel/core`, so I would have to fork the whole plugin ecosystem which is not feasible.)

### DIFF
--- a/packages/babel-core/package.json
+++ b/packages/babel-core/package.json
@@ -30,7 +30,8 @@
   },
   "browser": {
     "./lib/config/files/index.js": "./lib/config/files/index-browser.js",
-    "./lib/transform-file.js": "./lib/transform-file-browser.js"
+    "./lib/transform-file.js": "./lib/transform-file-browser.js",
+    "convert-source-map": false
   },
   "dependencies": {
     "@babel/code-frame": "^7.5.5",


nicolo-ribaudo:
Since `convert-source-map` often works without the `fs` module, I don't think that we should just drop its functionality.

I'd like to see this fixed directly in `convert-source-map`.

This function should be extracted to a separate file:
https://github.com/thlorenz/convert-source-map/blob/ad25a3e4373c641a2241cf1470b81ae123e4c23d/index.js#L28

This file should have a corresponding browser file which exports a function which throws when called.

tajo:
Hm, not sure if this can be fixed only by refactoring `convert-source-map`. You allow to pass any `SourceMap` configuration [here](https://github.com/babel/babel/blob/master/packages/babel-core/src/transformation/normalize-file.js#L49). So users of `@babel/core` can pass in something like:

```
{
  inputSourceMap: {
    isFileComment: true,
    commentFileDir: '...'
  }
}
```

which would still need to reach `fs` / `path` calls. 

I guess the ultimate solution would be to browser bundle different `normalize-file.js` and `generate.js`that would filter out `isFileComment` and called a modified version of `convert-source-map`. (Also it might be easier to just inline that library into babel-core (~120LoCs) than asking them to completely reorganize it.)

Anyway, could we still merge this in meantime as a workaround? Afaik, there is no point to bundle `convert-source-map` into the browser build today since the first import is `fs` and that always ends up in an exception. 

Agreed that this PR doesn't solve the underlying problem but at least if user doesn't care about source maps, `@babel/core` can be ran in browser.

nicolo-ribaudo:
> Afaik, there is no point to bundle convert-source-map into the browser build today since the first import is fs and that always ends up in an exception.

`@babel/standalone` currently has source maps support, and this PR would remove it. It relies on webpack's ability to provide `fs` mocks.

tajo:
I see. I opened a [PR for `convert-source-map`](https://github.com/thlorenz/convert-source-map/pull/68) that adds this into its `package.json`:

```js
"browser": {
  "fs": false
},
```

which is the library level version of [this mock](https://github.com/babel/babel/blob/master/scripts/gulp-tasks.js#L56).

Seems like the most pragmatic way forward.

(Note: it doesn't help when placed into @babel/core's package.json since it's a different module)

nicolo-ribaudo:
Thanks!

JLHwung:
Closing this in favour of https://github.com/thlorenz/convert-source-map/pull/68

@tajo Thank you.

nicolo-ribaudo:
@tajo Could you open a PR to update `convert-source-map` in Babel?

tajo:
@nicolo-ribaudo there you go: https://github.com/babel/babel/pull/10667

PR #10647, opened 2019-11-05
fix typo: identifer -> identifier

leonid-shevtsov:
Initially I only noticed the `babel-types` typo, but searching revealed two more - including one in a test that shouldn't be passing with the typo? Weird.

| Fixed Issues?            | - 
| Tests Added + Pass?      | (Unchanged)

### DIFF
--- a/packages/babel-generator/test/index.js
+++ b/packages/babel-generator/test/index.js
@@ -197,7 +197,7 @@ describe("generation", function() {
 
     const id2 = fn.body.body[0].expression;
     id2.name += "2";
-    id2.loc.identiferName = "bar";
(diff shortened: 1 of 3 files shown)


packages/babel-generator/test/index.js
```diff
@@ -197,7 +197,7 @@ describe("generation", function() {
 
     const id2 = fn.body.body[0].expression;
     id2.name += "2";
-    id2.loc.identiferName = "bar";
+    id2.loc.identifierName = "bar";
```
nicolo-ribaudo: It didn't fail because if `name` !== `identifierName`, `@babel/generator` uses `name`. Good catch!

PR #10648, opened 2019-11-05
bump @babel/* dev dependencies

JLHwung:
| Any Dependency Changes?  | Bump dev dependencies.

Bump `@babel/*` devDependencies to 7.7.0, after that the `yarn.lock` is processed by `yarn-deduplicate`.


existentialism:
Do we need to bump the `version` value in our [transform-runtime config](https://github.com/babel/babel/blob/master/babel.config.js#L126)?

JLHwung:
@existentialism Good point! I am shot by footgun.

Let's wait for the 7.7.2 release and I will update again afterwards.

babel.config.js
```diff
@@ -123,7 +123,10 @@ module.exports = function(api) {
         ],
         plugins: [
           includeRuntime
-            ? ["@babel/transform-runtime", { version: "7.4.4" }]
+            ? [
+                "@babel/transform-runtime",
+                { version: require("@babel/runtime/package").version },
```
nicolo-ribaudo: Good idea

JLHwung: Actually this option is _not_ effective until #10668 is merged, because we are always building against the `lib` version of `@babel/*`, which has inlined all the helper imports.

Issue #10646, opened 2019-11-05
throw error: api.caller is not a function in babel-preset-env/index.js


love2me:
## Bug Report

**Current Behavior**

` api.caller is not a function
    at _default
 (/Code/node_modules/@babel/preset-env/lib/index.js:234:35)
    at Object.default (/Code/node_modules/@babel/helper-plugin-utils/lib/index.js:19:12)
    at getEnvPlugins (/Code/node_modules/parcel-bundler/src/transforms/babel/env.js:62:34)
    at getEnvConfig (/Code/node_modules/parcel-bundler/src/transforms/babel/env.js:12:25)
`

**Environment**
- Babel version(s): [7.7.0]
- Node/npm version: [Node 8.11.3/npm 5.6.0]
- OS: [ OSX 10.13.4]
- Monorepo: [no]
- How you are using Babel: [cli]

**Possible Solution**
I think need to check api.caller is not a function in [packages/babel-preset-env/index.js line 240](https://github.com/babel/babel/blob/5f2240e7ab319c8e22cc6d2405364e439968e1e0/packages/babel-preset-env/src/index.js#L240).

nicolo-ribaudo:
Could you open a quick PR? 🙏

love2me:
> Could you open a quick PR? 🙏

I had to create a PR. Please review.
@nicolo-ribaudo

PR #10649, opened 2019-11-05
Fix(babel-preset-env): check api.caller is a function to avoid to thr…

love2me:
Fix [#10646](https://github.com/babel/babel/issues/10646)
check api.caller is a function to avoid to throw error: api.caller is not a function.


hjanuschka:
i can confirm, that it worked!

hjanuschka:
any chances you can publish a hotfix release?

nicolo-ribaudo:
Fixed in 7.7.1

PR #10656, opened 2019-11-05
fix: add inList setter for compatibility with babel-minify

### LINKED ISSUE #10654: Babel-minify throws cannot set property inList of [object Object] which has only a getter
## Bug Report

**Current Behavior**
Babel-minify will throw on @babel/traverse 7.7

**Input Code**
- [REPL](https://babeljs.io/repl#?babili=true&browsers=&build=&builtIns=false&spec=true&loose=false&code_lz=BQMwlABAvAfBLDEA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=babili%2Cenv&prettier=false&targets=&version=7.7.1&externalPlugins=)

**Expected behavior/code**
It should not throw

**Environment**
- Babel version(s): 7.7.0

**Possible Solution**
Add a setter of `inList` which will set `listKey` to `null` when `false` is passed.

** Additional Context **
This regression was introduced in #10480.

JLHwung:
| Fixed Issues?            | Fixes #10654 
| Tests Added + Pass?      | Nope
| Documentation PR Link    | 
| Any Dependency Changes?  |

Ultimately we should deprecate manipulating `inList`. A compatibility setter is offered for [`inList = false` usage](https://github.com/babel/minify/blob/e1d0c52f5b501f5849741be6db56f968094854eb/packages/babel-plugin-minify-simplify/src/index.js#L557) in babel-minify.

### DIFF
--- a/packages/babel-traverse/src/path/index.js
+++ b/packages/babel-traverse/src/path/index.js
@@ -161,6 +161,13 @@ export default class NodePath {
     return !!this.listKey;
   }
 
+  set inList(inList) {
+    if (!inList) {
+      this.listKey = null;
+    }
+    // ignore inList = true as it should depend on `listKey`
+  }
+
   get parentKey() {
     return this.listKey || this.key;
   }


CarterLi:
I'm encountering this issue. Hopefully a bugfix version will be released soon

ksegal:
Thanks for this patch @JLHwung! I was encountering the same errors for a babel project serving a storybook which surfaced a nested storybook ui (see screenshot [here](https://github.com/storybookjs/storybook/issues/5269)), so for now I downgraded babel versions until this fix is out in a release.

PR #10663, opened 2019-11-06
Publish from GitHub actions

nicolo-ribaudo:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,9 +5,25 @@ on:
     tags: ["v*"]
 
 jobs:
+  npm_publish:
+    name: Publish to npm
+    runs-on: ubuntu-latest
+
+    if: github.actor == "nicolo-ribaudo"
+
+    steps:
+      - name: Checkout the new tag
+        uses: actions/checkout@v1.0.0
+      
+      - name: Publish to npm
+        run: make publish-ci
+        env:
+          NPM_TOKEN: ${{ secrets.NICOLO_NPM_TOKEN }}
+
   github_release:
     name: Trigger GitHub release
     runs-on: ubuntu-latest
+    needs: npm_publish
     steps:
       - name: Checkout the new tag
         uses: actions/checkout@v1.0.0


.github/workflows/release.yml
```diff
@@ -5,9 +5,25 @@ on:
...
+    if: github.actor == "nicolo-ribaudo"
+
+    steps:
+      - name: Checkout the new tag
+        uses: actions/checkout@v1.0.0
+      
+      - name: Publish to npm
+        run: make publish-ci
+        env:
+          NPM_TOKEN: ${{ secrets.NICOLO_NPM_TOKEN }}
```
nicolo-ribaudo: If this ends up working, we can use @babel-bot's token and add other the people who currently have publish access

.github/workflows/release.yml
```diff
@@ -5,9 +5,25 @@ on:
...
+  npm_publish:
+    name: Publish to npm
+    runs-on: ubuntu-latest
+
+    if: github.actor == "nicolo-ribaudo"
+
+    steps:
+      - name: Checkout the new tag
+        uses: actions/checkout@v1.0.0
+
```
JLHwung: Consider move `Check if releasing from master` step to here now that `npm_publish` will run _before_ `github_release`.

nicolo-ribaudo: I think that if we publish a v* tag we would always want to publish it to npm (for example, for a special version), but we might not want to add it to the changelog.

JLHwung: I wound rather be conservative here, we could add branch to whitelist if publishing special version. But we should prevent if people push their local tags in their own branch and it ends up with a public npm release.

nicolo-ribaudo:
I don't know if I still like this PR anymore.
Currently only 5-6 people have publish access to npm, but a lot more people have push access to this repo.
A possible attacker (either someone with push access or someone who stole their account) could simply add theirselves to this new action and then publish.

JLHwung:
> A possible attacker (either someone with push access or someone who stole their account) could simply add theirselves to this new action and then publish.

We could separate the publish process by
1) tagging a release (running `lerna version` and sign the tags with GPG)
2) the bot will verify against the signature and publish to npm

By doing so we could make sure even if someone's GitHub account is stolen, the attacker could not forge a new release tag with trusted signature.

PR #10668, opened 2019-11-06
Reduce standalone build size

JLHwung:

In this PR we build babel-standalone against the `src` version of `@babel/*` dependencies instead of `lib`, which is specified in their `package.json`. By doing so the `transform-runtime` can replace babel helper calls by runtime helper imports. It is impossible to achieve this on `lib` version where the helpers are inlined.

Here is a brief summary on the number of inlined `function _interopRequireDefault(obj)` declarations in the artifacts. (produced by [this script](https://gist.github.com/JLHwung/a42e22b86c1b91a1382287279c39f639), lower is better) 

| | 7.7.1 | This branch
| --- | --- | --- |
| @babel/standalone | 151 | 4 |
| @babel/preset-env-standalone | 101 | 0 |

Note that there are still 4 inlined `_interopRequireDefault` declaration not removed in `@babel/standalone`, all of which comes from third party dependencies, i.e. `regenerator-runtime`.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -11,6 +11,7 @@ module.exports = function(api) {
     exclude: ["transform-typeof-symbol"],
   };
   const envOpts = Object.assign({}, envOptsNoTargets);
+  let transformRuntimeOpts = null;
 
   let convertESM = true;
   let ignoreLib = true;
@@ -48,6 +49,17 @@ module.exports = function(api) {
         node: "current",
       };
       break;
+  }
+
+  if (includeRuntime) {
+    const babelRuntimePackageJSONPath = require.resolve(
+      "@babel/runtime/package.json"
+    );
+    const path = require("path");
+    transformRuntimeOpts = {
+      version: require(babelRuntimePackageJSONPath).version,
+      absoluteRuntime: path.dirname(babelRuntimePackageJSONPath),
+    };
   }
 
   const config = {
@@ -123,10 +135,7 @@ module.exports = function(api) {
         ],
         plugins: [
           includeRuntime
-            ? [
-                "@babel/transform-runtime",
-                { version: require("@babel/runtime/package").version },
-              ]
+            ? ["@babel/transform-runtime", transformRuntimeOpts]
             : null,
         ].filter(Boolean),
       },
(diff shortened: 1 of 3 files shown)


packages/babel-core/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "7.7.0",
   "description": "Babel compiler core.",
   "main": "lib/index.js",
+  "_main:used-by-babel-standalone": "src/index.js",
```
JLHwung: By prefixing a pragmatic underscore and suffixing a paranoid signature I hope none of our user will depend on these fields in their bundling scripts. It should not kick in for most of users who are bundling _with_ babel instead of bundling babel itself.

JLHwung:
The size change of `babel-preset-env.js` comes from `forceAllTransforms: true`.

packages/babel-core/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "7.7.2",
   "description": "Babel compiler core.",
   "main": "lib/index.js",
+  "_main:used-by-babel-standalone": "src/index.js",
```
nicolo-ribaudo: I don't like this :laughing: 

Is it possible to use webpack's [`resolve`](https://webpack.js.org/configuration/resolve/) instead?

JLHwung: By using `resolve.alias` like the other packages, webpack will ignore the `browsers` field in package.json.

nicolo-ribaudo: I'm not sure I follow. If we did something like this, wouldn't it work?

```js
resolve: {
  alias: {
    "@babel/core$": "packages/babel-core/src/indec.js"
  }
}
```

babel.config.js
```diff
@@ -50,6 +51,17 @@ module.exports = function(api) {
       break;
   }
 
+  if (includeRuntime) {
+    const babelRuntimePackageJSONPath = require.resolve(
+      "@babel/runtime/package.json"
+    );
+    const path = require("path");
```
hg-pyun: It's a simple question. What is the advantage of declaring within this scope rather than at the top?

JLHwung: Not a strong reason: `require("path")` is used only when building babel-*-standalone. However, this config is shared with other building process (build/jest), by doing so we could avoid the overhead of requiring a builtin module.

Issue #10651, opened 2019-11-05
@babel/parser -> Flow Type Error when 


kubilaykiymaci:
## Bug Report

When i am getting bundle, @babel/parser is failed. (this has been since upgrading version 7.7.0)

```js
(data: any): void => onSwipeValueChange && onSwipeValueChange({ ...data, key })
```

**Environment**
- Babel version(s): [v7.6.4]
- Node/npm version: [v8.12.0]
- OS: [OSX]

**Additional context/Screenshots**

SyntaxError: /***/Documents/projects/gitrepos/***/index.js: Unexpected token (242:34)
 at Object.raise (/****/***/node_modules/@babel/parser/lib/index.js:6931:17)

JLHwung:
I could not reproduce this error on [REPL](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=true&loose=false&code_lz=BQEwhgLmBcAEYDsCeBKOA3A9gSxLAvAHyyYIDKA7tgA4CmAamADYCutAwgBaIDmtsAMgElyVOo1YduCPsADesAHTLwUADSwA1rSSwAviiA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=flow%2Cenv&prettier=true&targets=&version=7.7.1&externalPlugins=), could you attach your babel config and complete error stack trace?

kubilaykiymaci:
**my babel config**
```
{
  "presets": [
    "module:metro-react-native-babel-preset"
  ],
  "plugins": [
    [
      "module-resolver",
      {
        "root": [
          "./src",
          "./public"
        ]
      }
    ],
    ["@babel/plugin-proposal-decorators", { "legacy": true }]
  ]
}
```

**full of error stack**
```
error SyntaxError: 
***/swipelistview/index.js: Unexpected token (242:34)

  240 |           onSwipeValueChange={
  241 |             onSwipeValueChange
> 242 |               ? (data: any): void => onSwipeValueChange && onSwipeValueChange({ ...data, key })
      |                                   ^
  243 |               : null
  244 |           }
  245 |           ref={(row: any): void => (this._rows[key] = row)}. Run CLI with --verbose flag for more details.
SyntaxError: ***/swipelistview/index.js: Unexpected token (242:34)

  240 |           onSwipeValueChange={
  241 |             onSwipeValueChange
[...]

PR #10669, opened 2019-11-06
Parse arrows with params annotations in conditional expressions

nicolo-ribaudo:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/10651
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-parser/src/plugins/flow.js
```diff
@@ -2061,6 +2061,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
           return node.operator === "=";
 
         case "ParenthesizedExpression":
+        case "TypeCastExpression":
```
JLHwung: Should also check for `TSTypeAssertion`.

packages/babel-parser/test/fixtures/flow/regression/arrow-in-ternary-with-param-type-and-return-type-like/input.js
```diff
@@ -0,0 +1,2 @@
+test
+  ? (x: T): U => y
```
JLHwung: `U => y` must be parenthesized, otherwise flow [could not](https://flow.org/try/#0C4UwzsBQAE0PzQBQA8Bc0AqBKdBVaAvAHzQCeQA) parse it.

packages/babel-parser/src/plugins/flow.js
nicolo-ribaudo: This is just for flow.
Arrow functions in conditional were broken in our TS parser since forever. I wanted to keep this PR small to fix the regression (fixing it in TS requires much more work)

JLHwung: (hmm... GitHub should offer a facepalm reply)

packages/babel-parser/test/fixtures/flow/regression/arrow-in-ternary-with-param-type-and-return-type-like/input.js
nicolo-ribaudo: It is a known bug with flow, because it also rejects this valid JS code:
```js
test ? (x) : I => y
```

https://github.com/facebook/flow/issues/2006

nicolo-ribaudo: For context, here is one of my first PRs :stuck_out_tongue: 
https://github.com/babel/babylon/pull/573

PR #10677, opened 2019-11-07
Add `asserts this [is type]` parsing support

### LINKED ISSUE #10671: TypeScript 3.7 syntax: `asserts this`
## Bug Report

**Current Behavior**
In #10527, it wasn't taken into account that the `asserts` keyword can be followed by `this`, not only by an identifier. Now it's a parser error, `Unexpected token`. [Babel REPL](https://babeljs.io/en/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEAqCmEAu0DeAoaXqQvATkgJIQLIBMAFAJQBcOUBSMSAFgJYydyJLlqZsQ9gDNolNt3YA7ZGGnB4AezFk-1aPnhIArvmkBuQUKxt8SgO7Rp8KwFF85_DSNCAvujdA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2%2Ctypescript&prettier=false&targets=&version=7.7.1&externalPlugins=%40babel%2Fplugin-syntax-typescript%407.3.3).

**Input Code**
```js
class Test {
    assertIsTest2(): asserts this is Test2 {
        if (this instanceof Test2) return;
        throw new Error();
    }
}
```

**Expected behavior/code**
Should parse. [TypeScript playground](https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=8&pc=29#code/MYGwhgzhAEAqCmEAu0DeAoa1IXgJyQEkIFkAmACgEoAubKfJGJACwEsYO5Eky1Ms0NgDNoFVlzYA7ZGCnB4Ae1GleVaHnhIArnikBuAVlZ5FAd2hT4FgKJ5TeaoawBfdG-Sgc3ctHgAPJHgpABMYVTRoFyA).

**Environment**
- Babel version(s): 7.7.0
- How you are using Babel: @babel/parser

JLHwung:
| Fixed Issues?            | Fixes #10671 
| Patch: Bug Fix?          | Yes, (sort of)

As a follow-up to #10543, this PR adds `asserts this [is type]` support. In addition, it also fixes an edge case where `asserts` keyword contains escaped sequence.

Since we declared TypeScript 3.7 support on 7.7.0, I would expect this PR to be shipped in a patch release.


packages/babel-parser/test/fixtures/typescript/assert-predicate/asserts-this-with-predicate/output.json
```diff
@@ -0,0 +1,263 @@
...
+              "column": 8
+            },
+            "end": {
+              "line": 3,
+              "column": 1
+            }
+          },
+          "body": [
+            {
+              "type": "TSDeclareMethod",
```
nicolo-ribaudo: This seems to be wrong

JLHwung: Yeah it is wrong, it should be `ClassMethod`, in this [example](https://github.com/babel/babel/blob/0287c0f02f8844adbfd7d84a97a1a8a94c17fa1a/packages/babel-parser/test/fixtures/typescript/class/method-computed/input.ts#L2) the class property is also [finished](https://github.com/babel/babel/blob/0287c0f02f8844adbfd7d84a97a1a8a94c17fa1a/packages/babel-parser/test/fixtures/typescript/class/method-computed/output.json#L80) as TSDeclareMethod. I would suggest we address it in a separate PR.

nicolo-ribaudo: Well, in that example it's correct because there isn't the class body.

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -989,18 +1008,25 @@ export default (superClass: Class<Parser>): Class<Parser> =>
     }
 
     tsParseTypePredicateAsserts(): boolean {
-      if (!this.tsIsIdentifier()) {
-        return false;
-      }
-
-      const id = this.parseIdentifier();
       if (
```
JLHwung: This is refactor of `tsParseTypePredicateAsserts` which aims to avoid creating node by `parseIdentifier`.

packages/babel-parser/test/fixtures/typescript/assert-predicate/asserts-this-with-predicate/output.json
JLHwung: Addressed in e6bf63e.

PR #10680, opened 2019-11-08
Add cloneDeepWithoutLoc

### LINKED ISSUE #10670: Add cloneDeepWithoutLoc
I had a need for a `t.cloneDeepWithoutLoc` function today. We have a `t.cloneDeep` and a `t.cloneWithoutLoc` (both deprecated, but that's OK), but we do not have a `t.cloneDeepWithoutLoc`.

These functions all live in https://github.com/babel/babel/tree/bdb4318f47ec42ca4e325b5163e2e8ad57de4849/packages/babel-types/src/clone.

They are exported publicly at https://github.com/babel/babel/blob/bdb4318f47ec42ca4e325b5163e2e8ad57de4849/packages/babel-types/src/index.js#L19-L23.

We can add the functionality as a new optional-boolean parameter `t.cloneNode`. Let's call it `withoutLoc`. While we're add it, we can update `t.cloneWithoutLoc` to use this new parameter.

Let's also add [tests](https://github.com/babel/babel/blob/bdb4318f47ec42ca4e325b5163e2e8ad57de4849/packages/babel-types/test/cloning.js), and [flow](https://github.com/babel/babel/blob/bdb4318f47ec42ca4e325b5163e2e8ad57de4849/packages/babel-types/scripts/generators/flow.js#L133-L137) and [typescript](https://github.com/babel/babel/blob/bdb4318f47ec42ca4e325b5163e2e8ad57de4849/packages/babel-types/scripts/generators/typescript.js#L151-L155) types.

Taym95:
Fixes #10670

### DIFF
--- a/packages/babel-types/scripts/generators/flow.js
+++ b/packages/babel-types/scripts/generators/flow.js
@@ -135,9 +135,9 @@ lines.push(
   // clone/
   `declare function clone<T>(n: T): T;`,
   `declare function cloneDeep<T>(n: T): T;`,
+  `declare function cloneDeepWithoutLoc<T>(n: T): T;`,
   `declare function cloneNode<T>(n: T, deep?: boolean, withoutLoc?: boolean): T;`,
   `declare function cloneWithoutLoc<T>(n: T): T;`,
-  `declare function cloneDeepWithoutLoc<T>(n: T): T;`,
 
   // comments/
   `declare type CommentTypeShorthand = 'leading' | 'inner' | 'trailing'`,
--- a/packages/babel-types/scripts/generators/flow.js
+++ b/packages/babel-types/scripts/generators/flow.js
@@ -135,8 +135,9 @@ lines.push(
   // clone/
   `declare function clone<T>(n: T): T;`,
   `declare function cloneDeep<T>(n: T): T;`,
-  `declare function cloneNode<T>(n: T, deep?: boolean): T;`,
+  `declare function cloneNode<T>(n: T, deep?: boolean, withoutLoc?: boolean): T;`,
   `declare function cloneWithoutLoc<T>(n: T): T;`,
+  `declare function cloneDeepWithoutLoc<T>(n: T): T;`,
 
   // comments/
   `declare type CommentTypeShorthand = 'leading' | 'inner' | 'trailing'`,
--- a/packages/babel-types/scripts/generators/typescript.js
+++ b/packages/babel-types/scripts/generators/typescript.js
@@ -154,9 +154,9 @@ lines.push(
   // clone/
   `export function clone<T extends Node>(n: T): T;`,
   `export function cloneDeep<T extends Node>(n: T): T;`,
+  `export function cloneDeepWithoutLoc<T extends Node>(n: T): T;`,
   `export function cloneNode<T extends Node>(n: T, deep?: boolean, withoutLoc?: boolean): T;`,
   `export function cloneWithoutLoc<T extends Node>(n: T): T;`,
-  `export function cloneDeepWithoutLoc<T extends Node>(n: T): T;`,
 
   // comments/
   `export type CommentTypeShorthand = 'leading' | 'inner' | 'trailing'`,
--- a/packages/babel-types/scripts/generators/typescript.js
+++ b/packages/babel-types/scripts/generators/typescript.js
@@ -154,8 +154,9 @@ lines.push(
   // clone/
   `export function clone<T extends Node>(n: T): T;`,
   `export function cloneDeep<T extends Node>(n: T): T;`,
-  `export function cloneNode<T extends Node>(n: T, deep?: boolean): T;`,
+  `export function cloneNode<T extends Node>(n: T, deep?: boolean, withoutLoc?: boolean): T;`,
   `export function cloneWithoutLoc<T extends Node>(n: T): T;`,
+  `export function cloneDeepWithoutLoc<T extends Node>(n: T): T;`,
 
   // comments/
   `export type CommentTypeShorthand = 'leading' | 'inner' | 'trailing'`,
--- a/packages/babel-types/src/clone/cloneDeepWithoutLoc.js
+++ b/packages/babel-types/src/clone/cloneDeepWithoutLoc.js
@@ -1,0 +1,12 @@
+// @flow
+import cloneNode from "./cloneNode";
+/**
+ * Create a deep clone of a `node` and all of it's child nodes
+ * including only properties belonging to the node.
+ * excluding `_private` and location properties.
+ */
+export default function cloneDeepWithoutLoc<T: Object>(node: T): T {
+  const newNode = cloneNode(node, /* deep */ true, /* withoutLoc */ true);
+
+  return newNode;
+}
--- a/packages/babel-types/src/clone/cloneDeepWithoutLoc.js
+++ b/packages/babel-types/src/clone/cloneDeepWithoutLoc.js
@@ -6,7 +6,5 @@ import cloneNode from "./cloneNode";
  * excluding `_private` and location properties.
  */
 export default function cloneDeepWithoutLoc<T: Object>(node: T): T {
-  const newNode = cloneNode(node, /* deep */ true, /* withoutLoc */ true);
-
-  return newNode;
+  return cloneNode(node, /* deep */ true, /* withoutLoc */ true);
 }
--- a/packages/babel-types/src/clone/cloneNode.js
+++ b/packages/babel-types/src/clone/cloneNode.js
@@ -27,8 +27,7 @@ function cloneIfNodeOrArray(obj, deep) {
 /**
  * Create a clone of a `node` including only properties belonging to the node.
  * If the second parameter is `false`, cloneNode performs a shallow clone.
- * If the third parameter is `true`, cloneNode performs a shallow clone
- * of a `node` excluding `_private` location properties.
+ * If the third parameter is true, the cloned nodes exclude location properties.
  */
 export default function cloneNode<T: Object>(
   node: T,
--- a/packages/babel-types/src/clone/cloneNode.js
+++ b/packages/babel-types/src/clone/cloneNode.js
@@ -2,7 +2,7 @@ import { NODE_FIELDS } from "../definitions";
 
 const has = Function.call.bind(Object.prototype.hasOwnProperty);
 
-function cloneIfNode(obj, deep) {
+function cloneIfNode(obj, deep, withoutLoc) {
   if (
     obj &&
     typeof obj.type === "string" &&
@@ -11,17 +11,17 @@ function cloneIfNode(obj, deep) {
     obj.type !== "CommentLine" &&
     obj.type !== "CommentBlock"
   ) {
-    return cloneNode(obj, deep);
+    return cloneNode(obj, deep, withoutLoc);
   }
 
   return obj;
 }
 
-function cloneIfNodeOrArray(obj, deep) {
+function cloneIfNodeOrArray(obj, deep, withoutLoc) {
   if (Array.isArray(obj)) {
-    return obj.map(node => cloneIfNode(node, deep));
+    return obj.map(node => cloneIfNode(node, deep, withoutLoc));
   }
-  return cloneIfNode(obj, deep);
+  return cloneIfNode(obj, deep, withoutLoc);
 }
 
 /**
@@ -49,7 +49,7 @@ export default function cloneNode<T: Object>(
 
     if (has(node, "typeAnnotation")) {
       newNode.typeAnnotation = deep
-        ? cloneIfNodeOrArray(node.typeAnnotation, true)
+        ? cloneIfNodeOrArray(node.typeAnnotation, true, withoutLoc)
         : node.typeAnnotation;
     }
   } else if (!has(NODE_FIELDS, type)) {
@@ -58,7 +58,7 @@ export default function cloneNode<T: Object>(
     for (const field of Object.keys(NODE_FIELDS[type])) {
       if (has(node, field)) {
         newNode[field] = deep
-          ? cloneIfNodeOrArray(node[field], true)
+          ? cloneIfNodeOrArray(node[field], true, withoutLoc)
           : node[field];
       }
     }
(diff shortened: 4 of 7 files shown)


packages/babel-types/scripts/generators/flow.js
```diff
@@ -133,8 +133,9 @@ lines.push(
   // clone/
   `declare function clone<T>(n: T): T;`,
   `declare function cloneDeep<T>(n: T): T;`,
-  `declare function cloneNode<T>(n: T, deep?: boolean): T;`,
+  `declare function cloneNode<T>(n: T, deep?: boolean, withoutLoc?: boolean): T;`,
   `declare function cloneWithoutLoc<T>(n: T): T;`,
+  `declare function cloneDeepWithoutLoc<T>(n: T): T;`,
```
jridgewell: Nit: Can we alphabetize this?

packages/babel-types/scripts/generators/typescript.js
```diff
@@ -151,8 +151,9 @@ lines.push(
   // clone/
   `export function clone<T extends Node>(n: T): T;`,
   `export function cloneDeep<T extends Node>(n: T): T;`,
-  `export function cloneNode<T extends Node>(n: T, deep?: boolean): T;`,
+  `export function cloneNode<T extends Node>(n: T, deep?: boolean, withoutLoc?: boolean): T;`,
   `export function cloneWithoutLoc<T extends Node>(n: T): T;`,
+  `export function cloneDeepWithoutLoc<T extends Node>(n: T): T;`,
```
jridgewell: Ditto

packages/babel-types/src/clone/cloneDeepWithoutLoc.js
```diff
@@ -0,0 +1,12 @@
...
+import cloneNode from "./cloneNode";
+/**
+ * Create a deep clone of a `node` and all of it's child nodes
+ * including only properties belonging to the node.
+ * excluding `_private` and location properties.
+ */
+export default function cloneDeepWithoutLoc<T: Object>(node: T): T {
+  const newNode = cloneNode(node, /* deep */ true, /* withoutLoc */ true);
+
+  return newNode;
```
jridgewell: Nit: Just `return cloneNode(…)`, no need for a local variable.

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -28,7 +28,11 @@ function cloneIfNodeOrArray(obj, deep) {
  * Create a clone of a `node` including only properties belonging to the node.
  * If the second parameter is `false`, cloneNode performs a shallow clone.
```
jridgewell: Please add documentation for the `withoutLoc` parameter.

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -77,5 +81,9 @@ export default function cloneNode<T: Object>(node: T, deep: boolean = true): T {
     };
   }
 
+  if (withoutLoc) {
```
jridgewell: Let's move this up to the `has(node, "loc")` check above

packages/babel-types/src/clone/cloneWithoutLoc.js
```diff
@@ -1,12 +1,10 @@
 // @flow
-import clone from "./clone";
-
+import cloneNode from "./cloneNode";
 /**
  * Create a shallow clone of a `node` excluding `_private` and location properties.
  */
 export default function cloneWithoutLoc<T: Object>(node: T): T {
-  const newNode = clone(node);
-  newNode.loc = null;
```
jridgewell: 😄

packages/babel-types/test/cloning.js
```diff
@@ -72,4 +72,21 @@ describe("cloneNode", function() {
       node.declarations[0].id.typeAnnotation,
     );
   });
+
+  it("should support deep cloning without loc", function() {
```
jridgewell: Can you add a test for a shallow `cloneWithoutLoc`, too?

packages/babel-types/test/cloning.js
```diff
@@ -72,4 +72,21 @@ describe("cloneNode", function() {
...
+    const node = t.variableDeclaration("let", [
+      t.variableDeclarator({
+        ...t.identifier("value"),
+        typeAnnotation: t.anyTypeAnnotation(),
+      }),
+    ]);
```
jridgewell: I think we need to add `loc` objects to some of the nodes to test that loc isn't copied.

packages/babel-types/test/cloning.js
```diff
@@ -72,4 +72,21 @@ describe("cloneNode", function() {
...
+    const cloned = t.cloneNode(node, /* deep */ true, /* withoutLoc */ true);
+    expect(cloned.declarations[0].id.typeAnnotation).toEqual(
```
jridgewell: We can remove the `typeAnnotation` checks, we're only interested in `loc` for the test.

packages/babel-types/test/cloning.js
Taym95: Thanks for the clarification, just question how I can add `loc` objects to some of the nodes and witch nodes we should add loc objects to?

jridgewell: Just doing `node.loc = {}` is enough to test that `loc` isn't copied over. To test the `deep` part, we need to set a loc object on both the root node and some deeper node, like the `value` identifier.

Taym95: I added `loc` for `value` the deeper node.

packages/babel-types/test/cloning.js
```diff
@@ -91,6 +92,7 @@ describe("cloneNode", function() {
         typeAnnotation: t.anyTypeAnnotation(),
       }),
     ]);
+    node.loc = {};
```
jridgewell: Add a loc to `node.id`, too.

packages/babel-types/test/cloning.js
```diff
@@ -91,6 +92,7 @@ describe("cloneNode", function() {
...
     const cloned = t.cloneNode(node, /* deep */ true, /* withoutLoc */ true);
     expect(cloned.loc).toBeNull();
```
jridgewell: Check that `cloned.id.loc` is null, too.

Taym95: Thanks, somehow `cloned.id.loc` is always `{}` not null! any idea why?

packages/babel-types/test/cloning.js
Taym95: done!

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -27,8 +27,14 @@ function cloneIfNodeOrArray(obj, deep) {
 /**
  * Create a clone of a `node` including only properties belonging to the node.
  * If the second parameter is `false`, cloneNode performs a shallow clone.
+ * If the third parameter is `true`, cloneNode performs a shallow clone
+ * of a `node` excluding `_private` location properties.
```
jridgewell: Nit: "If the third parameter is `true`, the cloned nodes exclude location properties."

packages/babel-types/src/clone/cloneWithoutLoc.js
```diff
@@ -1,12 +1,10 @@
...
+  const newNode = cloneNode(node, /* deep */ false, /* withoutLoc */ true);
 
   return newNode;
```
jridgewell: Nit: We can also eliminate this local variable.

JLHwung:
@Taym95 The [CI error](https://travis-ci.com/babel/babel/jobs/255049174#L1078) is related to this PR, could you check why it does not work as intended?

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -27,8 +27,13 @@ function cloneIfNodeOrArray(obj, deep) {
 /**
  * Create a clone of a `node` including only properties belonging to the node.
  * If the second parameter is `false`, cloneNode performs a shallow clone.
+ * If the third parameter is true, the cloned nodes exclude location properties.
  */
-export default function cloneNode<T: Object>(node: T, deep: boolean = true): T {
+export default function cloneNode<T: Object>(
+  node: T,
+  deep: boolean = true,
+  withoutLoc: boolean = false,
```
jridgewell: Deep clones also need to receive the `withoutLoc` param, which is why the tests are failing.

Taym95: Yes, I will fix it!

Taym95: I added `withoutLoc` to deep clones wich means `cloned.declarations[0].id.loc` needs to be null but I am still receiving `{}` in the test, any idea why?

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -44,7 +49,7 @@ export default function cloneNode<T: Object>(node: T, deep: boolean = true): T {
 
     if (has(node, "typeAnnotation")) {
       newNode.typeAnnotation = deep
-        ? cloneIfNodeOrArray(node.typeAnnotation, true)
+        ? cloneIfNodeOrArray(node.typeAnnotation, true, withoutLoc)
         : node.typeAnnotation;
```
JLHwung: if `withoutLoc` is true, we should also clone it and pass down `withoutLoc`.

packages/babel-types/src/clone/cloneNode.js
jridgewell: This looks like is working.

Taym95: Ah yes, sorry :sweat_smile:!

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -53,14 +58,18 @@ export default function cloneNode<T: Object>(node: T, deep: boolean = true): T {
...
   if (has(node, "loc")) {
-    newNode.loc = node.loc;
+    if (withoutLoc) {
+      newNode.loc = null;
+    } else {
+      newNode.loc = node.loc;
+    }
   }
   if (has(node, "leadingComments")) {
     newNode.leadingComments = node.leadingComments;
```
JLHwung: when both `deep` and `withoutLoc` is `true`, we should apply `cloneIfNodeOrArray` on these comment nodes, same applied to `innerComments` and `trailingComments`.

Taym95: something like this? 

```
newNode.leadingComments = deep && withoutLoc ? cloneIfNodeOrArray(node.leadingComments, true, true) : node.leadingComments;

```

JLHwung: Yep! And please add a test on the new behaviour.

Taym95: Done!

packages/babel-types/test/cloning.js
```diff
@@ -72,4 +72,53 @@ describe("cloneNode", function() {
...
+        ...t.identifier("value"),
+        typeAnnotation: t.anyTypeAnnotation(),
+      }),
+    ]);
+    node.loc = {};
+    node.declarations[0].id.loc = {};
+    const cloned = t.cloneNode(node, /* deep */ true, /* withoutLoc */ true);
+    expect(cloned.loc).toBeNull();
+    expect(cloned.declarations[0].id.loc).toBeNull();
+    expect(cloned.declarations[0].id.leadingComments).toEqual(
```
JLHwung: The `cloned.declarations[0].id.leadingComments` will be `undefined` given that we did not add such comments when constructing `node`.

Similar to what we did in `node.declarations[0].id`, we can create the leading comments node
```js
node.leadingComments = [{ loc: {} }]
```
And then we assert that after it is cloned, the `loc` will be `null`:
```js
expect(cloned.leadingComments[0].loc).toBeNull()
```

Tip: you can view the node structures on [this website](https://lihautan.com/babel-ast-explorer/#?eyJiYWJlbFNldHRpbmdzIjp7InZlcnNpb24iOiI3LjYuMCIsIm9wdGlvbmFsQ2hhaW5pbmciOnsib3B0aW9ucyI6e30sImVuYWJsZWQiOnRydWV9LCJkeW5hbWljSW1wb3J0Ijp7Im9wdGlvbnMiOnt9LCJlbmFibGVkIjp0cnVlfX0sInRyZWVTZXR0aW5ncyI6eyJoaWRlRW1wdHkiOnRydWUsImhpZGVMb2NhdGlvbiI6ZmFsc2UsImhpZGVUeXBlIjp0cnVlLCJoaWRlQ29tbWVudHMiOmZhbHNlfSwiY29kZSI6Ii8vIGxlYWRpbmcgY29tbWVudHNcbmxldCB2YWx1ZSA9IDQyO1xuLy8gdHJhaWxpbmcgY29tbWVudHMifQ==), and `cloneWithoutLoc` aims to eliminate all the `loc` location information from the AST nodes.

Taym95: Thank you for the tip and the link, now I understand what I am doing!

Taym95: Thank you for the tip and the link, now I understand what I am doing!

packages/babel-types/src/clone/cloneNode.js
```diff
@@ -11,24 +11,29 @@ function cloneIfNode(obj, deep) {
     obj.type !== "CommentLine" &&
     obj.type !== "CommentBlock"
```
Taym95: @JLHwung because of this condition `cloneNode` will never be called on `comments` so `comment.loc` will never be null! I am adding comments to node using:
```
t.addComment(node, "leading", "leadingComments", true);
node.leadingComments[0].loc = {};
```
comment type is either `CommentLine` or `CommentBlock`!

JLHwung: Good catch! I think we should short-circuit this condition when `withoutLoc` and `deep` are both true. Otherwise it is inconsistent that after `cloneWithoutLoc`, the cloned `CommentLine` still contain `loc`.

Taym95: Yes, we can use:
```
((deep && withoutLoc) ||
      (obj.type !== "CommentLine" && obj.type !== "CommentBlock"))
```
But this will new an error here:

https://github.com/babel/babel/blob/bdb4318f47ec42ca4e325b5163e2e8ad57de4849/packages/babel-types/src/clone/cloneNode.js#L50

```
Unknown node type: "CommentBlock"
```
```
Unknown node type: "CommentLine"
```

nicolo-ribaudo: @Taym95 Maybe we could special-case comments and have a small helper function (not exposed as part of the API) `cloneCommentWithoutLoc`?

PR #10683, opened 2019-11-08
Fix: Don't call Object.keys on non-objects (babel#10482)

### LINKED ISSUE #10482: objectSpread not working with boolean arguments
Hi,
We have a scenario where objectSpread throws an exception when the arguments list contains a boolean.

This particular piece of code 

```
function _objectSpread(target) {
    var source = arguments[i] != null  ? arguments[i] : {};

    if (typeof Object.getOwnPropertySymbols === 'function') {
      ownKeys = ownKeys.concat(Object.getOwnPropertySymbols(source).filter(function (sym) {
        return Object.getOwnPropertyDescriptor(source, sym).enumerable;
      }));
    }

    ownKeys.forEach(function (key) {
      defineProperty(target, key, source[key]);
    });
  }

  return target;
}
```

gives an exception : Object.keys argument is not an object.

Shouldn't the check be like:
`   var source = arguments[i] != null && typeof arguments[i] === 'object' ? arguments[i] : {};`

chrishinrichs:
| Fixed Issues?            | Fixes #10482 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This changes the polyfill for the object spread operator to ensure that the argument to be spread is an Object before calling `Object.keys`. This fixes a JS error in IE11 which implements the old ES5 version of `Object.keys` which does not support non-object arguments.

A common pattern that will run into this issue:

```
const someBool = false;
const obj = {
  ...someBool && { foo: 'bar' }
};
```


packages/babel-helpers/src/helpers.js
```diff
@@ -396,7 +396,7 @@ helpers.objectSpread = helper("7.0.0-beta.0")`
 
   export default function _objectSpread(target) {
     for (var i = 1; i < arguments.length; i++) {
-      var source = (arguments[i] != null) ? arguments[i] : {};
+      var source = (typeof arguments[i] === 'object' && arguments[i] != null) ? arguments[i] : {};
       var ownKeys = Object.keys(source);
```
ljharb: it would be better to do this:
```suggestion
      var ownKeys = Object.keys(Object(source));
```

chrishinrichs: Thanks for the review! I have made the suggested change @ljharb, can you take another look?

PR #10685, opened 2019-11-09
Support removing nodes when deeply inside a traversal

jridgewell:
| Fixed Issues?            | 
| Major: Breaking Change?  | Possibly
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

We had an issue where removing a `VariableDeclarator` when inside the `id` key (in this case an `ObjectPattern`) would still iterate the `init` key. Well, when it tried to determine if it the init was a `ReferencedIdentifier`, it would pass a null `parent` to [`isReferenced`](https://github.com/babel/babel/blob/70c0ed512a5fb2e8f461c7d14e493a3ce131e686/packages/babel-types/src/validators/isReferenced.js#L7) (because removing the declarator turns `path.node = null`). That threw an error.

This could probably be extended to `shouldSkip` and `shouldStop` as well.

### DIFF
--- a/packages/babel-traverse/src/context.js
+++ b/packages/babel-traverse/src/context.js
@@ -9,6 +9,9 @@ export default class TraversalContext {
     this.scope = scope;
     this.state = state;
     this.opts = opts;
+
+    const parentContext = parentPath?.context;
+    if (parentContext) parentContext.child = this;
   }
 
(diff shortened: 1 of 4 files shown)


jridgewell:
I have a better approach.

PR #10689, opened 2019-11-09
fix: update asserts types

JLHwung:
| Fixed Issues?            | Update `asserts` related type definitions

This is a byproduct of #10677. We align the `asserts` definition to the current behaviour (optional boolean) and remove `asserts` from visitors since it's not an AST Node.


JLHwung:
Closing this PR as #10677 depends on it.

PR #10691, opened 2019-11-09
corrected the grammar of an option description

m-ahmadi:
Corrected the missed line from #10589

### DIFF
--- a/packages/babel-cli/src/babel/options.js
+++ b/packages/babel-cli/src/babel/options.js
@@ -73,7 +73,7 @@ commander.option(
   "Do not include superfluous whitespace characters and line terminators.",
   booleanify,
 );
-commander.option("--minified", "Save as many bytes when printing. [true|false]");
+commander.option("--minified [true|false]", "Save as many bytes when printing.");
 commander.option(
   "--auxiliary-comment-before [string]",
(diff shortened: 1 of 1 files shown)


packages/babel-cli/src/babel/options.js
```diff
@@ -73,7 +73,7 @@ commander.option(
   "Do not include superfluous whitespace characters and line terminators.",
   booleanify,
 );
-commander.option("--minified", "save as much bytes when printing [true|false]");
+commander.option("--minified", "Save as many bytes when printing. [true|false]");
```
JLHwung:
```suggestion
commander.option("--minified [true|false]", "Save as many bytes when printing.");
```
To be consistent with other descriptions.

packages/babel-cli/src/babel/options.js
```diff
@@ -73,7 +73,7 @@ commander.option(
   "Do not include superfluous whitespace characters and line terminators.",
   booleanify,
 );
-commander.option("--minified", "save as much bytes when printing [true|false]");
+commander.option("--minified [true|false]", "Save as many bytes when printing.");
```
JLHwung: Could you run prettier on this file? Otherwise we are good to go.

PR #10694, opened 2019-11-10
Fix optional method chaining in derived classes

### LINKED ISSUE #10673: Optional chaining on super.method() breaks
## Bug Report

**Current Behavior**
Optional call of a super method (`super.method?.()`) attempts to use `super` as an identifier, which is invalid. This might only happen in strict mode?

**Input Code**
[REPL](https://babeljs.io/repl/#?babili=false&browsers=last%201%20and_chr%20version%0A&build=&builtIns=usage&spec=false&loose=false&code_lz=MYGwhgzhAEBCkFNoG8BQ0PQLYIC4AsB7AEwAoBKFaAX1VtVEhgBEEAnASwDcFjoEAHrgQA7YjHgQkaTNjxEylGbMxs8AVzYjoEdQAd2AOhwESAfkMV0mWtSA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=stage-3&prettier=false&targets=&version=7.7.2&externalPlugins=)

```js
class Base {
    method() { }
}

class Sub extends Base {
    method() {
        return super.method?.()
    }
}
```

**Expected behavior/code**
Super method must always be invoked as `super.method()`, not `method.call(super)`

```js
{
}
```

**Environment**
- Babel version(s): 7.7.2
- Node/npm version: any
- OS: any
- Monorepo: any

**Possible Solution**

Two different possible solutions:

1. ~Access `super.method` twice~ (Don't do this, see https://github.com/babel/babel/issues/10673#issuecomment-550959334)
2. Call the cached method with `this` instead of super

<details>

<summary>Solution 1 (ignore this)</summary>

```js
// Solution 1
// Input
super.method?.();

// Output
((_super$method = super.method) === null || _super$method === undefined)
  ? undefined
  : super.method();
```

</details>

```js
// Solution 2
// Input
super.method?.();

// Output
((_super$method = super.method) === null || _super$method === undefined)
  ? undefined
  : _super$method.call(this)
```

This should be a beginner level fix.

Shriram-Balaji:
| Fixed Issues?            | Fixes #10673  
| Documentation PR Link    |  NA 

### DIFF
--- a/packages/babel-plugin-proposal-optional-chaining/src/index.js
+++ b/packages/babel-plugin-proposal-optional-chaining/src/index.js
@@ -86,6 +86,10 @@ export default declare((api, options) => {
                 context = object;
               }
 
+              if (t.isSuper(context)) {
+                context = t.thisExpression();
+              }
+
               node.arguments.unshift(t.cloneNode(context));
               node.callee = t.memberExpression(
                 node.callee,
--- a/packages/babel-plugin-proposal-optional-chaining/src/index.js
+++ b/packages/babel-plugin-proposal-optional-chaining/src/index.js
@@ -82,12 +82,10 @@ export default declare((api, options) => {
               let context = scope.maybeGenerateMemoised(object);
               if (context) {
                 chain.object = t.assignmentExpression("=", context, object);
+              } else if (t.isSuper(object)) {
+                context = t.thisExpression();
               } else {
                 context = object;
-              }
-
-              if (t.isSuper(context)) {
-                context = t.thisExpression();
               }
 
               node.arguments.unshift(t.cloneNode(context));
--- a/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/super-method-call/input.js
+++ b/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/super-method-call/input.js
@@ -1,0 +1,12 @@
+"use strict";
+class Base {
+  method() {
+    return 'Hello!';
+  }
+}
+
+class Derived extends Base {
+    method() {
+        return super.method?.()
+    }
+}
--- a/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/super-method-call-loose/input.js
+++ b/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/super-method-call-loose/input.js
@@ -1,0 +1,12 @@
+"use strict";
+class Base {
+  method() {
+    return 'Hello!';
+  }
+}
+
+class Derived extends Base {
+    method() {
+        return super.method?.()
+    }
+}
--- a/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/super-method-call-loose/options.json
+++ b/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/super-method-call-loose/options.json
@@ -1,0 +1,3 @@
+{
+  "plugins": [["proposal-optional-chaining", { "loose": true }]]
+}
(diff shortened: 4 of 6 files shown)


packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -86,6 +86,10 @@ export default declare((api, options) => {
                 context = object;
               }
 
+              if (t.isSuper(context)) {
```
jridgewell: Nit: We can put this with the `if else` above.

packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/super-method-call/input.js
```diff
@@ -0,0 +1,12 @@
+"use strict";
```
jridgewell: Could you also add a `loose` test?  We can reuse this same `input.js` file, just copy it to a new folder and also copy https://github.com/babel/babel/blob/76cb2010e350a1ef69966d92c0d706e6115efca3/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/function-call-loose/options.json into the same fodler.

packages/babel-plugin-proposal-optional-chaining/src/index.js
Shriram-Balaji: @jridgewell Would you like for `t.isSuper` to be called twice in the `if else`? because context would be null if it wasnt memoized before, right?

packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/super-method-call/input.js
Shriram-Balaji: Sure, added this.

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -86,6 +86,10 @@ export default declare((api, options) => {
...
+                context = t.thisExpression();
+              }
+
               node.arguments.unshift(t.cloneNode(context));
```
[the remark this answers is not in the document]
nicolo-ribaudo: But it's already used inside the `t.assignmentExpression` of the `if` branch.

packages/babel-plugin-proposal-optional-chaining/src/index.js
jridgewell: There should only be one check needed. The only way `context` would be a `Super` here is if it went into the `else` case (because if it when through the `if` case it would be an `AssignmentExpression`). So we need to do an `else if` (before the current `else`) to check if `t.isSuper(object)`.

Shriram-Balaji: @jridgewell Alright, that makes sense. The changes are done now 😄

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -82,6 +82,8 @@ export default declare((api, options) => {
               let context = scope.maybeGenerateMemoised(object);
               if (context) {
                 chain.object = t.assignmentExpression("=", context, object);
+              } else if (t.isSuper(context)) {
```
JLHwung:
```suggestion
              } else if (t.isSuper(object)) {
```
`context` is `null` in the `else` branch.

Shriram-Balaji: @JLHwung Oh right, that's what I meant to do. Fixed now.

existentialism:
@Shriram-Balaji thanks! great work!

PR #10695, opened 2019-11-11
Adding funding field for babel/core

hzoo:
via https://blog.npmjs.org/post/188841555980/updates-to-community-docs-more, https://github.com/npm/rfcs/blob/latest/accepted/0017-add-funding-support.md

Not sure if we need to add it to all of them? I think this is fine (or we add it to some other ones like the parser specifically); maybe no need for each package (although it's not hard to do either)?


nicolo-ribaudo:
https://github.com/babel/babel/pull/10687 😛

PR #10698, opened 2019-11-12
Babel should not silently remove unknown options after commander arguments

### LINKED ISSUE #4825: Validate babel-cli program options
It is frustrating that `babel-cli` program does not validate options that are passed to it, e.g.

```sh
babel ./src --out-dir ./dist --copy-files --spurce-maps
```

This will run the program without any error (note the spelling error in `--spurce-maps`; the intetended value is `--source-maps`).

Ideally, a CLI program should throw an error whenever an unknown property/ value is passed to it.

As far as I can tell, https://github.com/tj/commander.js does not support "strict" options parameter (like [yargs#strict](https://github.com/yargs/yargs#strict)). Therefore, this could be achieved by:

* switching the implementation to yargs; or
* implementing JSON schema validation of the options object

JLHwung:
| Fixed Issues?            | Fixes #4825
| Any Dependency Changes?  | `commander.js` is bumped to v4.0.1

In this PR we bump `commander.js` to v4.0.1, which can throw on unknown options _after_ arguments when an action handler is provided. To do so we register an empty action handler so that unknown options error can be thrown.

### Impact analysis of upgrading `commander.js` from 2.8 to 4.0

Breaking changes from 2.0 to 3.0
- Change TypeScript to use overloaded function for .command. 
Not related
- custom event listeners: --no-foo on cli now emits option:no-foo (previously option:foo)
Not related since we are not using custom event listeners
- default value: defining --no-foo after defining --foo leaves the default value unchanged (previously set it to false)
Not related since we don't have any option that has both `--no-` and `--` prefixes.

Breaking changes from 3.0 to 4.0

- keep command object out of program.args when action handler called
Not related since we are not using the command object in an action handler
- Commander is only officially supported on Node 8 and above, and requires Node 6
It should be fine since we will drop Node 6 and 8 soon, too. If CI is good we may say it should work on Node 6 from now on.
- Testing for no arguments is changed
Not related since I can't find any `.args.length` usage in our codebase.

### DIFF
--- a/packages/babel-cli/package.json
+++ b/packages/babel-cli/package.json
@@ -19,7 +19,7 @@
     "compiler"
   ],
   "dependencies": {
-    "commander": "^2.8.1",
+    "commander": "^4.0.1",
     "convert-source-map": "^1.1.0",
     "fs-readdir-recursive": "^1.1.0",
     "glob": "^7.0.0",
--- a/packages/babel-cli/src/babel/options.js
+++ b/packages/babel-cli/src/babel/options.js
@@ -163,6 +163,10 @@ commander.option(
 
 commander.version(pkg.version + " (@babel/core " + version + ")");
 commander.usage("[options] <files ...>");
+// register an empty action handler so that commander.js can throw on
+// unknown options _after_ args
+// see https://github.com/tj/commander.js/issues/561#issuecomment-522209408
+commander.action(() => {});
 
 export type CmdOptions = {
   babelOptions: Object,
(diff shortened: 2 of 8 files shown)


packages/babel-cli/package.json
```diff
@@ -19,7 +19,7 @@
     "compiler"
   ],
   "dependencies": {
-    "commander": "^2.8.1",
+    "commander": "^4.0.1",
```
JLHwung: @nicolo-ribaudo I don't think a major bump of dependencies is a breaking change as long as we have handled necessary breaking changes on our side, isn't it?

PR #10701, opened 2019-11-13
Circumvent typeof transform for umd build template

### LINKED ISSUE #10662: [Reopen]_typeof function in generated code calls itself before it is defined (T7149)
It's a reopen of ISSUE https://github.com/babel/babel/issues/4125.

I have found a reproducible example. My transform code:
```js
const babel = require('@babel/core');
const fs = require('fs');
(async () => {
	const result = await babel.transformFileAsync('predefine.ts', {
		presets: [
		    ["@babel/preset-env", { modules: 'umd' }],
		    [require("@babel/preset-typescript")],
		],
	});
	fs.writeFileSync("_typeof_out.js", result.code);
})();
```

Generated code:

[predefine.zip](https://github.com/babel/babel/files/3812495/predefine.zip)

babel version: 7.6.4
transform-modules-umd version:7.7.0, but 7.2.0 does not reference `_typeof` before declare it

@nicolo-ribaudo

JLHwung:
| Fixed Issues?            | Fixes #10662 

In this PR we revise the umd module build template to avoid typeof transforms. It is not an ideal fix as ultimately we should hoist helper declaration to the program block if referenced by any nodes in the build template.

On the other hand we may skip polyfilling the `typeof` calls in build template as it is rare to assign a symbol to these global properties.

This PR is a follow-up to #10477.

### DIFF
--- a/packages/babel-plugin-transform-function-name/test/fixtures/function-name/export-default-arrow-renaming-module-umd/output.js
+++ b/packages/babel-plugin-transform-function-name/test/fixtures/function-name/export-default-arrow-renaming-module-umd/output.js
@@ -10,7 +10,7 @@
     factory(mod.exports);
     global.input = mod.exports;
   }
-})(typeof globalThis === "object" ? globalThis : typeof self === "object" ? self : this, function (_exports) {
+})(typeof globalThis !== "undefined" ? globalThis : typeof self !== "undefined" ? self : this, function (_exports) {
   "use strict";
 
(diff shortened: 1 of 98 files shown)


langi93:
[foundation-sites-develop.zip](https://github.com/babel/babel/files/3843400/foundation-sites-develop.zip)

JLHwung:
@langi93 Is that related to this PR?

jridgewell:
It's a spam account.

PR #10702, opened 2019-11-13
Ignore PR job on master branch

jbhoosreddy:
| Fixed Issues?            | Relates to #4987 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Followup to #10579. If we look at a commit which landed after, like https://github.com/babel/babel/commit/1d1101eb7c8d89a2179b098f6dd131b4b479020a, you can see that even though it's passing all checks the PR check is failing since the PR job is not approved to run.

This will incorrectly show all jobs as pending, which is noisy.

This PR aims at fixing this issue. Thanks to @nicolo-ribaudo for pointing it out!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -176,9 +176,17 @@ workflows:
     jobs:
       - approve-test262-run:
           type: approval
+          filters:
+            branches:
+              ignore:
+                - master
       - test262:
           requires:
             - approve-test262-run
+          filters:
+            branches:
+              ignore:
+                - master
   e2e:
     jobs:
       - publish-verdaccio


jbhoosreddy:
These are the docs I used to implement this: https://circleci.com/docs/2.0/configuration-reference/#filters-1

The only issue here is that there is not way for me to properly test this unless it's on master.

nicolo-ribaudo:
> The only issue here is that there is not way for me to properly test this unless it's on master.

I'll merge this PR since it's minimal; let's see if it works.

PR #10705, opened 2019-11-14
Add eslint/ for @babel/eslint-* packages

kaicataldo:
| Fixed Issues?            | No issue!
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | I think this would be considered a minor?
| Documentation PR Link    | 
| Any Dependency Changes?  | Only the ones in the packages themselves

This is the result of running the following commands:
```sh
$ ./node_modules/.bin/lerna import ~/Code/babel-eslint-config-internal --dest=eslint --preserve-commit --flatten
$ ./node_modules/.bin/lerna import ~/Code/babel-eslint-parser --dest=eslint --preserve-commit --flatten
$ ./node_modules/.bin/lerna import ~/Code/babel-eslint-plugin --dest=eslint --preserve-commit --flatten
$ ./node_modules/.bin/lerna import ~/Code/babel-eslint-plugin-development/ --dest=eslint --preserve-commit --flatten
```

This PR uses `lerna import` to bring in all the ESLint-related packages and adds them to a top-level `eslint/` directory.

I didn't change the package names yet or update the versions. Is it okay to follow up with that? Figured it would be nice to bring them in as they are now and update them afterwards for posterity.


kaicataldo:
Ah, one tricky thing. ESLint dropped support for Node <8, and we're running tests in Node 6 here.

JLHwung:
We can extract the jest configuration into `jest.config.js` and ignore the test files begin with `@babel/eslint-*` when `process.version` is less than 8.

nicolo-ribaudo:
I'd prefer not to modify anything in this PR (just import the packages), and then modify things in a separate PR

JLHwung:
Yep we can hold it for it depends on other PRs.

kaicataldo:
Is it okay to have CI failing until I can follow up? Looks like we'll have to do this in a few steps:

1) Have ESLint ignore the new directories
2) Remove custom linting config, testing, etc. configuration from the individual packages
3) Fix Linting errors
4) Disable testing `@babel/eslint-parser` related packages in Node <8

One thing I'm a bit confused about is that it looks like the repo is already on ESLint v6, which doesn't support Node 6.

kaicataldo:
I think we can fix CI by updating the ESLint config in this PR, which seems like the best thing to do to me.

kaicataldo:
Ah, I see what it is - it's currently 6.0.1, which doesn't use any syntax that would throw in Node 6. This will stop working if the repo upgrades to the latest version, though, as we now have syntax that would be unsupported (see https://github.com/eslint/eslint/pull/12407).

kaicataldo:
Are there plans to drop support for Node 6 (and 11) now that they're not longer officially supported?

JLHwung:
IIRC, Nodejs 11 has been dropped from CI. We will drop node 6 and 8 in babel 8 after node 8 reaches EOL, which is the end of this year.

nicolo-ribaudo:
Ok I think that my magic worked:

```
git filter-branch -f --msg-filter "node $PWD/../rewrite-commit-msg.js" master..import-babel-eslint
```

```js
// ../rewrite-commit-msg.js

"use strict";
const fs = require("fs");
const cp = require("child_process");

let msg = fs.readFileSync("/dev/stdin", "utf8");

const folderToRepo = {
  "babel-eslint-config-internal": "eslint-config-babel",
  "babel-eslint-parser": "babel-eslint",
  "babel-eslint-plugin": "eslint-plugin-babel",
  "babel-eslint-plugin-development": "eslint-plugin-babel-plugin"
};

let file = cp
  .execSync(
    `git diff-tree --no-commit-id --name-only -r ${process.env.GIT_COMMIT}`,
    { encoding: "utf8" }
  )
  .trim()
  .split("\n")[0];

if (!file.startsWith("eslint/")) {
  return msg;
}
const repo = folderToRepo[file.split("/")[1]];
if (!repo) {
  throw new Error(`Folder does not map to a known repo: ${file} ${process.env.GIT_COMMIT}`);
}

process.stdout.write(msg.replace(/(?<!\w)#(\d+)/g, `babel/${repo}#$1`));
```

Issue #10709, opened 2019-11-14
ESLint-related packages migration - TODO


nicolo-ribaudo:
- [x] Merge https://github.com/babel/babel/pull/10705
- [x] Remove lock files from new packages (https://github.com/babel/babel/pull/10711)
- [ ] ~~Ignore new packages while linting~~
- [x] Make "Lint" pass on CI (https://github.com/babel/babel/pull/10712)
- [x] Move CHANGELOG files to `.github/` (https://github.com/babel/babel/pull/10714)
- [x] Actually rename the new packages (https://github.com/babel/babel/pull/10715)
- [x] Prevent lerna from publishing those packages until we are ready (@nicolo-ribaudo) (https://github.com/babel/babel/pull/10722)
- [x] Unify build process (https://github.com/babel/babel/pull/10723, https://github.com/babel/babel/pull/10729)
- [x] Migrate from `mocha` to `jest` (which is used by the other `@babel/` packages) (https://github.com/babel/babel/pull/10716)
- [x] Disable tests for ESLint related packages on unsupported node versions (https://github.com/babel/babel/pull/10730)

rajasekarm:
Can I pick this? Migrate from mocha to jest

nicolo-ribaudo:
Ok :+1:

nicolo-ribaudo:
If you wait 5 mins, I will push the linting PR first

JLHwung:
Yep, let's make the CI green again.

rajasekarm:
Sure, I'm making the change. I'm using https://github.com/skovhus/jest-codemods to make the change.

kaicataldo:
Assigning this to myself so I don't lose track of this, but that doesn't mean others should hold off on working on these tasks.

kaicataldo:
Looks like we’re all set except for renaming the packages. Thoughts on how to proceed? I’m of the mind that we should all of this with the next minor released that occurs after the new packages are ready.

thiagoarrais:
Maybe related? https://github.com/babel/babel-eslint/pull/807

@rajasekarm: what do you think?

rajasekarm:
Yeah, This changes should be added here.

nicolo-ribaudo:
Please wait for Kai's go-ahead before opening new features/bug fixes in this repo.

rajasekarm:
Sure

PR #10715, opened 2019-11-14
Rename ESLint packages and fix publishing

nicolo-ribaudo:
| Fixed Issues?            | Ref: https://github.com/babel/babel/issues/10709

The goal of this PR is to make the e2e CircleCI workflow pass.

### DIFF
--- a/eslint/babel-eslint-parser/package.json
+++ b/eslint/babel-eslint-parser/package.json
@@ -4,6 +4,9 @@
   "description": "Custom parser for ESLint",
   "author": "Sebastian McKenzie <sebmck@gmail.com>",
   "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
   "repository": {
     "type": "git",
     "url": "https://github.com/babel/babel-eslint.git"
--- a/eslint/babel-eslint-parser/package.json
+++ b/eslint/babel-eslint-parser/package.json
@@ -17,7 +17,6 @@
   "homepage": "https://github.com/babel/babel-eslint",
   "scripts": {
     "test": "cd test && mocha specs && cd -",
-    "preversion": "npm test",
     "changelog": "git log `git describe --tags --abbrev=0`..HEAD --pretty=format:' * %s (%an)' | grep -v 'Merge pull request'"
   },
   "engines": {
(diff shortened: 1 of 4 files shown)


eslint/babel-eslint-parser/package.json
```diff
@@ -1,9 +1,12 @@
 {
-  "name": "babel-eslint",
-  "version": "11.0.0-beta.0",
+  "name": "@babel/eslint-parser",
```
JLHwung: I would prefer to defer the rebranding until babel v8, and at that time we can align the `@babel/eslint-parser`'s version to `@babel/core`.

nicolo-ribaudo: We could already align it with 7.8.0

JLHwung: It will make people confused, things like when one should use `babel-eslint` and `@babel/eslint-parser`. If we rebrand them in v8, it's more clear: `@babel/eslint-parser` 8.0 works with `@babel/core` 8.0, all the `babel-eslint` supports up to `@babel/core` 7.x

nicolo-ribaudo: I will update the PR tomorrow morning

cc @kaicataldo @existentialism @hzoo

kaicataldo: I'm in agreement with @nicolo-ribaudo. I actually like that we'll be publishing a different package with the v11 features, because there's already been a lot of confusion around it. Historically, it's always been a challenge to denote the compatibility of a specific version of babel-eslint with Babel and ESLint, and I think this name change will go a long way in solving it on Babel's end.

I also think that releasing babel-eslint v11 (which has a major breaking change in the configuration-aware parsing), forcing users to upgrade, and then releasing @babel/eslint-parser v7.8 a little while later, forcing another migration, is actually more burden on the end user than it would be to just upgrade once.

babel-eslint has been incomplete/buggy for a long time now, and I'm viewing this move as a fresh start to make it more of a first class citizen. Can we just say that @babel/eslint-parser is compatible with whatever the first version we release it in (whether that's 7.8 or 8.0), mark babel-eslint as deprecated, and just guarantee @babel/eslint-parser works from here on out? Old versions won't be any worse than they are now, but hopefully going forward the versioning will make it clear what version of @babel/eslint-* packages one should be using with a particular version of Babel.

All that being said, I'm new around here, and will definitely defer to team consensus 😄

nicolo-ribaudo:
Closing until we decide (there is time until 7.8.0 to decide anyway)

PR #10719, opened 2019-11-15
Add main file for babel-node

dword-design:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    |
| Any Dependency Changes?  |

It sometimes makes sense to `require.resolve` a package in order to look for certain information/files. E.g. [https://www.npmjs.com/package/depcheck](depcheck) parses the binaries from packages in order to find out if it's an unused dependency. However, this requires the packages to have a main file. This PR simply adds `lib/babel-node.js` as the main fine.

### DIFF
--- a/packages/babel-node/package.json
+++ b/packages/babel-node/package.json
@@ -39,5 +39,5 @@
   "bin": {
     "babel-node": "./bin/babel-node.js"
   },
-  "main": "src/babel-node.js"
+  "main": "lib/babel-node.js"
 }
--- a/packages/babel-node/package.json
+++ b/packages/babel-node/package.json
@@ -38,5 +38,6 @@
   },
   "bin": {
     "babel-node": "./bin/babel-node.js"
-  }
+  },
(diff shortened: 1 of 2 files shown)


JLHwung:
I'm conservative on this change as adding a `main` entry will enable people to use `require("@babel/node")`, which we are not meant for.

Could [`Special` options](https://www.npmjs.com/package/depcheck#special) help `depcheck` to detech `@babel/node`?

dword-design:
Hey @JLHwung, bin special is in fact the special I'm using. This uses resolveFrom, which in turn uses require.resolve. Other options like resolve-bin also use require.resolve, so it always seems to come down to that :P.

JLHwung:
Closing this PR as I think it should be fixed on `resolve-bin`'s side: https://github.com/thlorenz/resolve-bin/issues/7

dword-design:
Yeah you're right

PR #10720, opened 2019-11-15
polish: skip creating extra reference for safely re-used node

### LINKED ISSUE #10717: @babel/plugin-proposal-null-coalescing-operator injects unnecessary variable
## Bug Report

**Current Behavior**
@babel/plugin-proposal-null-coalescing-operator creates unnecesary variable declaration when variable used in conditional expression is within function scope

**Input Code**
https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABACwIZgCYBsCmBhNMAcxwAoYwAHEKANVSxBwEpEBvAKEUQgQGcoiAG4MmiALyIK1OqJyIA_AsRRkMPgDoRjHAG4OAXyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.7.3&externalPlugins=

```js
function handleChange(inputValue) {
  const value = inputValue ?? this.value;
}
```

**Expected behavior/code**
I expect output to not have additional `var _inputValue;` declaration because it's unnecessary.

```js
function handleChange(inputValue) {
  // since we don't accessing nested property in inputValue the output should be:
  var value = inputValue !== null && inputValue !== void 0 ? inputValue : this.value;
}
```

```js
{
  presets: [
    ['@babel/preset-env', {
      useBuiltIns: true,
      loose: true,
      modules: false,
      corejs: 3,
    }],
  ],
  plugins: [
    '@babel/plugin-proposal-optional-chaining',
    '@babel/plugin-proposal-nullish-coalescing-operator'
  ],
}
```

**Environment**
- Babel version(s): 7.7.2
- Node/npm version: 13.1.0
- OS: Windows 10
- Monorepo: no
- How you are using Babel: babel-loader

**Possible Solution**
-

-

JLHwung:
| Fixed Issues?            | Fixes #10717 

In this PR we skip creating extra reference when the `left` of `??` is a [static reference](https://github.com/babel/babel/blob/5bb4ee73da59fd9be7c94ef9337ab3917f12dce2/packages/babel-traverse/src/scope/index.js#L309), which is safely to reused when later unfolding into conditional comparison.

Compared to the master branch, this PR generate less code for the common case `var bar = foo ?? "bar"`
Before:
```js
var bar = (_foo = foo) !== null && _foo !== void 0 ? _foo : "bar"
// minified (52 B)
// var bar=null!==(_foo=foo)&&void 0!==_foo?_foo:"bar";
```
After:
```js
var bar = foo !== null && foo !== void 0 ? foo : "bar"
// minified (43 B)
// var bar=null!==foo&&void 0!==foo?foo:"bar";
```


packages/babel-plugin-proposal-nullish-coalescing-operator/src/index.js
```diff
@@ -16,14 +16,17 @@ export default declare((api, { loose = false }) => {
           return;
         }
 
-        const ref = scope.generateUidIdentifierBasedOnNode(node.left);
-        scope.push({ id: ref });
+        let ref, assignment;
+        // skip creating extra reference when `left` is semantically safe to re-use
+        if (t.isIdentifier(node.left)) {
```
nicolo-ribaudo: What about using `isConstantExpression`, or maybe better `maybeGenerateMemoised`?

JLHwung: Good idea! I end up with `maybeGenerateMemoised` because it is rare that the `left` is a literal --even if it is, creating a reference other than inlining this literal should costs less bytes in most situations, given that we generate `[_]+` as identifier name for literals.

packages/babel-plugin-proposal-nullish-coalescing-operator/src/index.js
```diff
@@ -16,14 +16,15 @@ export default declare((api, { loose = false }) => {
...
-          node.left,
-        );
+        let ref = scope.maybeGenerateMemoised(node.left);
+        let assignment;
+        // skip creating extra reference when `left` is static
+        if (ref === null) {
+          ref = node.left;
+          assignment = t.cloneNode(node.left);
+        } else {
+          assignment = t.assignmentExpression("=", ref, node.left);
```
JLHwung: if `ref` is non-null, it is always cloned from `node.left`.

PR #10723, opened 2019-11-16
Unify build step for ESLint packages

kaicataldo:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | refs https://github.com/babel/babel/issues/10709
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

This PR makes it so that we're running the new `eslint/*` packages through the same build process as the rest of our packages. Files also started getting linted, so had to fix some ESLint errors for the precommit Git hook.

PR #10724, opened 2019-11-16
Update ESLint/Prettier config for eslint/* packages

kaicataldo:
| Fixed Issues?            | refs #10709
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

~~This is based on https://github.com/babel/babel/pull/10723 and I'll rebase it once we land that. Converts to ES Modules so that we can be make sure we're compiling everything correctly.~~

~~Side benefit: fixes ESLint for this package!~~

Originally intended to convert some of the new files to use ES Modules, but I realized we should fix linting/formatting first.


.prettierrc
```diff
@@ -13,10 +13,12 @@
...
       "**/codemods/*/test/**/*.js",
       "**/packages/*/src/**/*.js",
-      "**/packages/*/test/**/*.js"
+      "**/packages/*/test/**/*.js",
+      "**/eslint/*/src/**/*.js",
+      "**/eslint/*/test/**/*.js"
     ],
-    "parser": "babylon",
     "options": {
+      "parser": "babel",
```
kaicataldo: This was actually incorrect!

PR #10730, opened 2019-11-17
Do not run eslint/* tests in Node v6

kaicataldo:
| Fixed Issues?            | refs https://github.com/babel/babel/issues/10709
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The `eslint/* package tests are failing in Node v6 because ESLint v6 dropped support for Node <8. This is a stopgap measure until Babel follows suit.

One thing to note: this repo is already using ESLint v6, but ESLint didn't add any Node>=8 syntax until a later minor release, which is why it's passing in CI in Node v6 right now. If we upgrade, linting will begin failing as well.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -8,6 +8,7 @@ module.exports = {
     "eslint/*/src/**/*.js",
   ],
   // The eslint/* packages use ESLint v6, which has dropped support for Node v6.
+  // TODO: Remove this process.version check in Babel 8.
   testRegex: `./(packages|codemods${
     /^v6./u.test(process.version) ? "" : "|eslint"
   })/[^/]+/test/.+\\.m?js$`,
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,50 @@
+module.exports = {
+  collectCoverageFrom: [
+    "packages/*/src/**/*.mjs",
+    "packages/*/src/**/*.js",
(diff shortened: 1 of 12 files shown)


nicolo-ribaudo:
> One thing to note: this repo is already using ESLint v6, but ESLint didn't add any Node>=8 syntax until a later minor release, which is why it's passing in CI in Node v6 right now. If we upgrade, linting will begin failing as well.

This shouldn't be a problem, because we lint using node 12.

jest.config.js
```diff
@@ -0,0 +1,50 @@
+module.exports = {
+  collectCoverageFrom: [
+    "packages/*/src/**/*.mjs",
+    "packages/*/src/**/*.js",
+    "codemods/*/src/**/*.mjs",
+    "codemods/*/src/**/*.js",
+    "eslint/*/src/**/*.mjs",
+    "eslint/*/src/**/*.js",
+  ],
+  // The eslint/* packages use ESLint v6, which has dropped support for Node v6.
```
JLHwung: Nit: could you add a todo item here that we should remove the process.version check in Babel 8? By that time we will drop Node < 8.9 support.

kaicataldo:
Woohoo, Travis is passing! 🎉

Issue #10735, opened 2019-11-19
rest spread ignores default values


eps1lon:
## Bug Report

**Current Behavior**
Default values are ignored for rest variables.

**Input Code**
- https://babeljs.io/repl#?babili=false&browsers=chrome%2051&build=&builtIns=false&spec=false&loose=false&code_lz=AQYw9gdgzgLsDeAoYLgEkIAcCuMAKATmJlAFwLKoogA2AhlFAKZQA0lVAdNxjvkSQDyMABZMCHYAF9gAXgTAAZmDDkAjNPZduYUeMoz5mAVADcQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.7.3&externalPlugins=

```js
const props = {};
const {
  InputProps: {
    classes,
    ...InputPropsOther
  } = { foo: 1 },
  ...other
} = props;

console.log(InputPropsOther);
```

**Expected behavior/code**

logs `{ foo: 1 }`

Can't provide a runkit since no node version supports that syntax but Chrome 78 does and it matches the expected behavior.

**Babel Configuration (.babelrc, package.json, cli command, .eslintrc)**

preset-env with `chrome 51` as the target

**Environment**
- Babel version(s): 7.7.3
- Node/npm version: repl
- OS: [e.g. OSX 10.13.4, Windows 10]
- Monorepo: no
- How you are using Babel: repl

**Possible Solution**

**Additional context/Screenshots**
Causes crashes for nested `{}` default values because it transpiles to `props.InputProps.classe` where `props.InputProps` can be `undefined` but guarded against with a default `{}`. Omitting that default value results in "cannot read property classes of undefined"

PR #10736, opened 2019-11-19
test: Add testcase for rest with default values

eps1lon:
Testcase for #10735

| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes but they should produce a different output
| Documentation PR Link    |  Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/nested-default-value-2/output.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/nested-default-value-2/output.js
@@ -1,0 +1,9 @@
+const {
+  InputProps: {
+    classes
+  } = {
+    foo: 1
+  }
+} = props,
+      InputPropsOther = babelHelpers.objectWithoutProperties(props.InputProps, ["classes"]),
+      other = babelHelpers.objectWithoutProperties(props, ["InputProps"]);
--- a/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/nested-default-value-2/input.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/nested-default-value-2/input.js
@@ -1,0 +1,7 @@
+const {
+  InputProps: {
+    classes,
+    ...InputPropsOther
+  } = { foo: 1 },
+  ...other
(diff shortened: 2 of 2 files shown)


packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/nested-default-value-2/output.js
```diff
@@ -0,0 +1,9 @@
+const {
+  InputProps: {
+    classes
+  } = {
+    foo: 1
+  }
+} = props,
+      InputPropsOther = babelHelpers.objectWithoutProperties(props.InputProps, ["classes"]),
```
eps1lon: The default value should probably be declared as a new variable and then referenced here e.g. 
```suggestion
      InputPropsOther = babelHelpers.objectWithoutProperties(props.InputProps === undefined ? defaultInputProps : props.InputProps, ["classes"]),
```

JLHwung: Thanks for the test case and possible solution! I will work on this case.

Note that for the `AssignmentPattern`, we should memorize the transformed member expression so that we would not execute them twice:
```js
const ref$propsInputProps = props.InputProps;
InputPropsOther = babelHelpers.objectWithoutProperties( 
  ref$propsInputProps === undefined ? defaultInputProps : ref$propsInputProps,
  ["classes"]
),
```

It is similar to what babel has done on array rest. ([REPL](https://babeljs.io/repl#?babili=false&browsers=chrome%2036&build=&builtIns=false&spec=false&loose=false&code_lz=G4QwTgBA3gUBEgFwQNp3hARuANO-AdEQPYAuAFgKZjoC6EAvNBAGbHHIAsATBAL54-jaHwDcQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.7.3&externalPlugins=))

JLHwung:
Close it as #10738 includes all commits of this PR.
@eps1lon Thanks for the test case.

PR #10737, opened 2019-11-19
Flow enums: fix enum body location

gkz:
The location of the Flow enum bodys was incorrect. I fixed this in flow-parser (https://github.com/facebook/flow/commit/498a2092abc6a9b5e2776c45e31aa7354575478a), and this PR fixes this in Babel, making it consistent with flow-parser, and also consistent with what the enum body contains (everything after `enum E`).

The diff also does some minor refactoring and cleaning up:
- Splits out parsing of the enum body into a separate function (separate from parsing the enum declaration overall)
- Rather than have a helper function for enum string bodies, it returns the string members, so the creation of enum bodies is completely contained in the new `flowEnumBody` function.

This incorrect location was discovering when adding support in Prettier https://github.com/prettier/prettier/pull/6860

After this lands and they update their parser deps, Prettier can remove their hacks.

| Fixed Issues?            |  
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/src/plugins/flow.js
+++ b/packages/babel-parser/src/plugins/flow.js
@@ -3250,34 +3250,32 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return members;
     }
 
-    flowEnumStringBody(
-      bodyNode: N.Node,
+    flowEnumStringMembers(
       initializedMembers: Array<N.Node>,
       defaultedMembers: Array<N.Node>,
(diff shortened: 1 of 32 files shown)


nicolo-ribaudo:
(Please run `yarn install && make fix`)

existentialism:
@gkz thanks!

PR #10738, opened 2019-11-19
Assignment pattern support in transform-object-rest-spread 

### LINKED ISSUE #10735: rest spread ignores default values

JLHwung:
| Fixed Issues?            | Fixes #10735

In this PR we add support for assignment pattern in transform-object-rest-spread. 

When the object pattern is inside an assignment pattern, the memoized reference for both `objRef` and the `right`(default value) of `AssignmentPattern` will be created so that they will only be executed once.


packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -131,6 +131,48 @@ export default declare((api, opts) => {
     });
   }
 
+  function createObjRefExpression(path, objRef) {
+    let impureDefaultValueDeclarator;
```
JLHwung: nit: consider rename it to `nonStaticDefaultValueDeclarator` because it is defined when `t.isStatic(path)` is `false`.

packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -131,6 +131,48 @@ export default declare((api, opts) => {
...
+    let objRefExpression;
+    if (path.parentPath?.isAssignmentPattern()) {
```
nicolo-ribaudo: When can `parentPath` be null?

packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -131,6 +131,48 @@ export default declare((api, opts) => {
...
+      const { scope } = path;
+      const parentRight = path.parentPath.get("right");
+      const parentRightId = scope.maybeGenerateMemoised(parentRight.node);
+      if (parentRightId) {
+        impureDefaultValueDeclarator = t.variableDeclarator(
+          parentRightId,
+          t.cloneNode(parentRight.node),
+        );
+      }
+      const objRefName = scope.generateUidBasedOnNode(objRef, "_ref");
```
nicolo-ribaudo: Nit
```suggestion
      const objRefName = scope.generateUidBasedOnNode(objRef, "ref");
```

packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -131,6 +131,48 @@ export default declare((api, opts) => {
...
+          t.cloneNode(objRef),
+        ),
+        t.conditionalExpression(
+          t.binaryExpression(
+            "!==",
+            t.identifier(objRefName),
+            scope.buildUndefinedNode(),
+          ),
+          t.identifier(objRefName),
+          parentRightId ?? t.cloneNode(parentRight.node),
```
nicolo-ribaudo: It `parentRightId` is null, won't this generate an invalid AST? AFAIK the third parameter of `conditionalExpression` is mandatory.

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/assignment-pattern-static/output.js
```diff
@@ -0,0 +1,11 @@
+function foo() {
+  var _props$Global;
+
+  const {
+    Global: {
+      classes
+    } = this
+  } = props,
+        InputPropsOther = babelHelpers.objectWithoutProperties((_props$Global = props.Global, _props$Global !== void 0 ? _props$Global : this), ["classes"]),
```
nicolo-ribaudo: This is accessing `props.Global` twice: once for the destructuring, and once here.

packages/babel-plugin-proposal-object-rest-spread/src/index.js
JLHwung: If `parentRightId` is `null`, it means `parentRight.node` is static and can be safely referenced. This case is covered in [assignment-static](https://github.com/babel/babel/blob/ac487d4a55cb70f96ac3e1019680943560a56f71/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/assignment-pattern-static/output.js).

If a `parentPath` is an assignment pattern, the `right` must presents, so `t.cloneNode(parentRight.node)` should always be truthy.

Maybe I can rename it into `nonStaticParentRightId`?

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/assignment-pattern-static/output.js
JLHwung: Good point! It seems that this issue has been around [for a while](https://github.com/babel/babel/blob/da5e355f995e46601aaeeb67bdae48012c78ea27/packages/babel-plugin-transform-object-rest-spread/test/fixtures/object-rest/nested/actual.js). I would suggest we address it in a separate PR.

packages/babel-plugin-proposal-object-rest-spread/src/index.js
nicolo-ribaudo: I'm not sure I follow. If `parentRightId` is `null`, this will call `conditionalExpression` like this:
```js
t.conditionalExpression(/* objRefName !== void 0 */, /* objRefName */, null)
```
which isn't valid, because the third parameter can't be null.

JLHwung: `parentRightId` is generated from `scope.maybeGenerateMemoised(parentRight.node)`, which is `null` when `t.isStatic(parentRight.node)` is true according to the [implementation](https://github.com/babel/babel/blob/fee4d936a3443c4d8812cac208e567e0ba8edc9d/packages/babel-traverse/src/scope/index.js#L330).

packages/babel-plugin-proposal-object-rest-spread/src/index.js
JLHwung: `parentPath` can be `undefined` when [`traverse.hasType`](https://github.com/babel/babel/blob/fee4d936a3443c4d8812cac208e567e0ba8edc9d/packages/babel-traverse/src/index.js#L103) is called, but I admit I may be too paranoid here.

packages/babel-plugin-proposal-object-rest-spread/src/index.js
nicolo-ribaudo: Omg I was stupid :man_facepalming: I was reading `a ?? b` as `a == null ? null : b`

JLHwung: (damn programming homework should also be blamed)

nicolo-ribaudo:
@JLHwung It has been almost six months since you wrote this code: I think your review/:heavy_check_mark: would be "valid" because you have a way deeper understanding of how things work now.

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/assignment-pattern-non-static/output.js
```diff
@@ -0,0 +1,12 @@
+var _foo, _props$InputProps;
+
+const _foo = {
```
nicolo-ribaudo, line 3: `_foo` is declared twice.

PR #10741, opened 2019-11-20
Update chokidar to v3

pipobscure:
chokidar depends on fsevents. Version 3 updates fsevents to version 2 which is based on napi. V1.x is now deprecated. Therefore chokidar should now urgently be updated to v3.

| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-cli/package.json
+++ b/packages/babel-cli/package.json
@@ -29,7 +29,7 @@
     "source-map": "^0.5.0"
   },
   "optionalDependencies": {
-    "chokidar": "^2.1.8"
+    "chokidar": "^3.3.0"
   },
   "peerDependencies": {
     "@babel/core": "^7.0.0-0"


nicolo-ribaudo:
What are the breaking changes in v3?

JLHwung:
Chokidar v3 requires nodejs 8+. Since CI is green, we may state `"chokidar": "^3.0.0 || ^2.0.0" so that people using nodejs 6 still gets proper support.

pipobscure:
No breaking changes. It’s just that the native part was entirely rewritten to take advantage of N-API so we felt it merrited a major bump.

pipobscure:
Also be aware fsevents v1 (and by extension chokidar v2) is ending EOL. Meaning every user will compile the native part on their own (no more precompiled binaries).

nicolo-ribaudo:
As @JLHwung wrote, https://github.com/paulmillr/chokidar#changelog reports a big breaking change.
Without node 6 support we can't currently merge this PR until Babel 8.

milesj:
Supporting `^3.0.0 || ^2.0.0` at minimum would unblock all users who want to use Node v12 locally, because at the moment, it's super broken on OSX.

Are there any metrics on how many users are actually on Node v6?

nicolo-ribaudo:
If `^3.0.0 || ^2.0.0` would still download v2 on node 6, then it would be ok.

Btw, either the next minor version or the one after will probably be a major, so this PR wouldn't have to wait too long anyway.

JLHwung:
A new install from `^3.0.0 || ^2.0.0` will still resolve to `3.x` unless the user specified explicitly that it should be `2.x`. But yeah we do have plans to chokidar as many other dependencies to the latest version on Babel 8.

pipobscure:
Sounds like a plan. Thanks!

JLHwung:
Closing it in favour of #10747 

@pipobscure Thanks for the PR!

PR #10742, opened 2019-11-20
Update preset-env mappings

### LINKED ISSUE #10740: @babel/preset-env includes transform-template-literals for safari
https://github.com/babel/babel/blob/master/packages/babel-preset-env/data/plugins.json#L2-L10

https://caniuse.com/#feat=template-literals

Seems like there was [a critical bug](https://bugs.webkit.org/show_bug.cgi?id=190756) that affected template literals in safari 12 and ios 12.1. However, the issue was fixed and doesn't affect newer versions.

It's probably impossible to skip just those broken versions... but we're punishing future versions unnecessarily.

I propose we re-enable excluding this transform for versions newer than the broken versions.

```js
"transform-template-literals": {
    "chrome": "41",
    "edge": "13",
    "firefox": "34",
    "safari": "12.1", // <--
    "node": "4",
    "ios": "12.2", // <--
    "samsung": "3.4",
    "opera": "28",
    "electron": "0.24"
  },
```

I'm running into this issue while trying to configure multiple builds for my project. I noticed that targeting only latest chrome, firefox and safari was still including this transform "for no reason". My best guess is that safari and ios are excluded because of that bug.

existentialism:
| Fixed Issues?            | Fixes #10740
| Documentation PR Link    | 
| Any Dependency Changes?  |

After this, we can rebase https://github.com/babel/babel/pull/10225 and land that as well.

### DIFF
--- a/packages/babel-preset-env/data/corejs2-built-in-features.js
+++ b/packages/babel-preset-env/data/corejs2-built-in-features.js
@@ -95,7 +95,31 @@ const es = {
     ],
   },
 
-  "es6.map": "Map",
+  // This is explicit to prevent Map-related proposals (like
+  // Map.prototype.upsert) from being included
+  "es6.map": {
+    features: [
+      "Map / basic functionality",
+      "Map / constructor arguments",
+      "Map / constructor requires new",
+      "Map / constructor accepts null",
+      "Map / constructor invokes set",
+      "Map / iterator closing",
+      "Map / Map.prototype.add returns this",
+      "Map / -0 key converts to +0",
+      "Map / Map.prototype.size",
+      "Map / Map.prototype.delete",
+      "Map / Map.prototype.clear",
+      "Map / Map.prototype.forEach",
+      "Map / Map.prototype.keys",
+      "Map / Map.prototype.values",
+      "Map / Map.prototype.entries",
+      "Map / Map.prototype[Symbol.iterator]",
+      "Map / Map.prototype isn't an instance",
+      "Map / Map iterator prototype chain",
+      "Map / Map[Symbol.species]",
+    ],
+  },
 
   "es6.math.acosh": "Math methods / Math.acosh",
   "es6.math.asinh": "Math methods / Math.asinh",
--- a/packages/babel-preset-env/data/corejs2-built-ins.json
+++ b/packages/babel-preset-env/data/corejs2-built-ins.json
@@ -77,6 +77,7 @@
     "safari": "12",
     "node": "11",
     "ios": "12",
+    "samsung": "10.2",
     "opera": "56",
     "electron": "4"
   },
@@ -1496,6 +1497,7 @@
     "safari": "12",
     "node": "10",
     "ios": "12",
+    "samsung": "9.2",
     "opera": "53",
     "electron": "3.1"
   },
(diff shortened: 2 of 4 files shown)


packages/babel-preset-env/data/built-in-modules.json
```diff
@@ -6,7 +6,7 @@
     "safari": "10.1",
     "opera": "48",
     "ios_saf": "10.3",
-    "and_chr": "71",
-    "and_ff": "64"
+    "and_chr": "74",
```
JLHwung: I don't think we need to bump `and_chr` and `and_ff` because:
1) CanIUse doesn't support mobile browsers versions
2) `caniuse-db` is not bumped

existentialism: > CanIUse doesn't support mobile browsers versions

What do you mean? This data is generated directly via our [script](https://github.com/babel/babel/blob/master/packages/babel-preset-env/scripts/build-modules-support.js)?

image

> `caniuse-db` is not bumped

FWIW, if you run `build-data` on master, it updates these two values with the version of `caniuse-db` that it's pinned to. In fact, if we bump it, it'll update `and_chr` to 78.

JLHwung: They should not be updated because caniuse does not track old mobile browser versions, so `and_chr` will always return the latest version of all available Android Chrome, same as `and_ff`. (i.e. they are bumped in this [commit](https://github.com/Fyrd/caniuse/commit/3bc36b569c7506b38d34c4a2da07ed7794436ec5) of caniuse-db)

Could we skip `and_ff` and `and_chr` in https://github.com/babel/babel/blob/fee4d936a3443c4d8812cac208e567e0ba8edc9d/packages/babel-preset-env/scripts/build-modules-support.js#L6

existentialism: Oh, gotcha. Yep I'll make an update.

nicolo-ribaudo: Related: https://github.com/babel/babel/pull/9978/files

existentialism: Hard to keep everything in your head 🤦‍♂

JLHwung: (So true. `git-blame` is my very best reference for code review.)

nicolo-ribaudo: According to https://github.com/babel/babel/pull/7344#issue-167553342, it was you who explained it to me :stuck_out_tongue:

JLHwung: @nicolo-ribaudo Any idea why `and_chr` is removed from `skipList` in #9978?

existentialism: Adding them to `skipList` would remove them from the output altogether

nicolo-ribaudo: Yeah, and it doesn't really matter since browserlists returns the desktop version anyway.

JLHwung: Then we need a `doNotUpdateList` then. Anyway let's address it in a separate PR.

PR #10744, opened 2019-11-21
Rename babel-eslint-config-internal -> @babel/eslint-config-internal

kaicataldo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

As discussed in today's meeting, this PR renames this package. I think we should release this with the next release.

### DIFF
--- a/eslint/babel-eslint-config-internal/index.js
+++ b/eslint/babel-eslint-config-internal/index.js
@@ -1,9 +1,11 @@
+"use strict";
+
 module.exports = {
   parser: "babel-eslint",
   extends: "eslint:recommended",
   plugins: ["flowtype"],
   parserOptions: {
-    ecmaVersion: 2017,
+    ecmaVersion: 2019,
     sourceType: "module",
   },
   rules: {
--- a/eslint/babel-eslint-config-internal/index.js
+++ b/eslint/babel-eslint-config-internal/index.js
@@ -5,8 +5,18 @@ module.exports = {
   extends: "eslint:recommended",
   plugins: ["flowtype"],
   parserOptions: {
-    ecmaVersion: 2019,
+    ecmaVersion: 2020,
     sourceType: "module",
+  },
+  globals: {
+    // Flow
+    Iterator: true,
+    $Keys: true,
+  },
+  env: {
+    node: true,
+    es2020: true,
+    browser: true,
   },
   rules: {
     camelcase: "off",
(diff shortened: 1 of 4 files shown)


eslint/babel-eslint-config-internal/package.json
```diff
@@ -1,21 +1,15 @@
 {
-  "name": "eslint-config-babel",
-  "version": "9.0.0",
-  "description": "ESLint config for babel",
+  "name": "@babel/eslint-config-internal",
+  "version": "7.7.0",
```
kaicataldo: Does this seem right? Or should I leave this as is for now until we actually publish this?

nicolo-ribaudo: As long as it's <= than the current Babel version, it should be ok. Mabe `0.0.0` so that it's clear that it isn't published yet?

eslint/babel-eslint-config-internal/package.json
```diff
@@ -1,21 +1,15 @@
...
+  "description": "The Babel Team's ESLint configuration",
```
nicolo-ribaudo:
```suggestion
  "description": "The Babel Team's ESLint configuration. Since it's internal, it might not respect semver.",
```

:stuck_out_tongue: :pray:

eslint/babel-eslint-config-internal/index.js
```diff
@@ -1,9 +1,11 @@
+"use strict";
+
 module.exports = {
   parser: "babel-eslint",
   extends: "eslint:recommended",
   plugins: ["flowtype"],
   parserOptions: {
-    ecmaVersion: 2017,
```
JLHwung: Nit: we can set [`env.es2020: true`](https://eslint.org/docs/user-guide/configuring#specifying-environments) and get rid of `parserOptions.ecmaVersion`.

nicolo-ribaudo: Does `babel-eslint` actually need it? :thinking:

kaicataldo: @nicolo-ribaudo Are you asking about `ecmaVersion`? I think it's useful to set it because rules can access it when running (example [here](https://github.com/eslint/eslint/blob/master/lib/rules/no-lone-blocks.js#L83)).

I'd be fine with setting the `es2020` env as well.

existentialism: Yeah let's do `2020`

PR #10747, opened 2019-11-21
Bump Node support to >= 10.13

### LINKED ISSUE #9981: Drop support for node <8.9.0

JLHwung:
| Fixed Issues?            | Closes #9981
| Documentation PR Link    | 
| Any Dependency Changes?  | Dropped Node.js <10.13 support, use `rollup-plugin-node-polyfills`

In this PR we drop Node.js < 10.9 support.
- [x] Update `engines` field in package.json
- [x] Update babel.config.js
- [x] Update CI config
- [x] Bump dependencies that requires node.js >= 10.13
- [x] remove ad-hoc polyfills for Node.js 6 and Node.js 8 in code base
- [x] remove `minNodeVersion: "8.0.0"` test case config
- [x] replace `make-dir` by `fs.mkdir`
- [x] replace `node-environment-flags` by `process.allowedNodeEnvironmentFlags`
- [ ] ~~https://github.com/calvinmetcalf/rollup-plugin-node-globals/pull/34~~


kaicataldo:
Do we also need to update the `engines` field in the individual packages? Or is that something we should do in a separate PR?

packages/babel-plugin-proposal-unicode-property-regex/package.json
```diff
@@ -8,9 +8,6 @@
     "access": "public"
   },
   "main": "lib/index.js",
-  "engines": {
```
JLHwung: The `engines` field is redundant here as long as it peer depends on `@babel/core`.

packages/babel-cli/package.json
```diff
@@ -24,12 +24,12 @@
     "fs-readdir-recursive": "^1.1.0",
     "glob": "^7.0.0",
     "lodash": "^4.17.13",
-    "make-dir": "^2.1.0",
-    "slash": "^2.0.0",
+    "make-dir": "^3.0.0",
```
coreyfarrell: Bumping the requirement to `>10.12.0` would allow babel to drop `make-dir` in favor of using `fs.mkdirSync(dir, {recursive: true})` directly.  10.13.0 was the first LTS, I would suggest babel should just specify that is the minimum version since dependencies which require node.js `>=10` are likely to only support LTS versions.

JLHwung:
The CI error is due to an upstream issue: https://github.com/calvinmetcalf/rollup-plugin-node-globals/pull/34

milesj:
Jest 25 just landed that updated chokidar to v3. Seems like Babel is the last one. Looking forward to this landing :)

nicolo-ribaudo:
:tada:

Issue #8462, opened 2018-08-13
ESM runtime helper files should have the .mjs extension


jaydenseric:
#### Config

```js
{
  presets: [
    [
      '@babel/env',
      {
        targets: {
          node: '6.10',
          browsers: '>1%'
        },
        modules: false,
        shippedProposals: true,
        loose: true
      }
    ]
  ],
  plugins: [
    ['@babel/transform-runtime', { useESModules: true }]
  ]
}
```

#### Input code

```js
import React from 'react'

class Demo extends React.Component {}
```

Outputs:

```js
import _inheritsLoose from '@babel/runtime/helpers/esm/inheritsLoose'
import React from 'react'

var Demo = (function(_React$Component) {
  _inheritsLoose(Demo, _React$Component)

  function Demo() {
    return _React$Component.apply(this, arguments) || this
  }

  return Demo
})(React.Component)
```

When run as native ESM in Node.js with `--experimental-modules` enabled:

```
node --experimental-modules lib/demo.mjs
```

`@babel/runtime/helpers/esm/inheritsLoose` resolves to `node_modules/@babel/runtime/helpers/esm/inheritsLoose.js`. Because Node.js can only run native ESM in files with the `.mjs` extension, this results in a runtime error:

```
$ node --experimental-modules lib/demo.mjs
(node:97782) ExperimentalWarning: The ESM module loader is experimental.
/Users/jaydenseric/Sites/demo/node_modules/@babel/runtime/helpers/esm/inheritsLoose.js:1
(function (exports, require, module, __filename, __dirname) { export default function _inheritsLoose(subClass, superClass) {
                                                              ^^^^^^

SyntaxError: Unexpected token export
    at new Script (vm.js:74:7)
    at createScript (vm.js:246:10)
    at Proxy.runInThisContext (vm.js:298:10)
    at Module._compile (internal/modules/cjs/loader.js:657:28)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at createDynamicModule (internal/modules/esm/translators.js:56:15)
    at setExecutor (internal/modules/esm/create_dynamic_module.js:50:23)
```

### Expected behavior/code

The ESM helper import should resolve a helper file with a `.mjs` extension.

### Environment
[...]

PR #10748, opened 2019-11-21
Add support for native esm to @babel/runtime

nicolo-ribaudo:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/8462
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Currently `esm/` helpers are broken: when you try to natively `import` them, they are parsed as scripts and thus throw an error.

Assing `type: module` is a better fix than renaming the files to `*.mjs`, because this isn't a breaking change.

I only tested this PR manually because:
1) Babel can't be built on Node 13.2 which supports unflagged modules (https://github.com/nodejs/node/issues/30581)
2) Jest doesn't support native modules

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,6 +222,7 @@ ifneq ("$(I_AM_USING_VERDACCIO)", "I_AM_SURE")
 endif
 	$(MAKE) prepublish-build
 	yarn lerna version patch --force-publish=$(FORCE_PUBLISH)  --no-push --yes --tag-version-prefix="version-e2e-test-"
+	git --no-pager diff HEAD
 	yarn lerna publish from-git --registry http://localhost:4873 --yes --tag-version-prefix="version-e2e-test-"
 	$(MAKE) clean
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -249,8 +249,9 @@ clean-lib:
 			$(call clean-source-lib, $(source))))
 
 clean-runtime-helpers:
-	rm -rf packages/babel-runtime/helpers
-	rm -rf packages/babel-runtime-corejs2/helpers
+	rm -f packages/babel-runtime/helpers/**/*.js
+	rm -f packages/babel-runtime-corejs2/helpers/**/*.js
+	rm -f packages/babel-runtime-corejs3/helpers/**/*.js
 	rm -rf packages/babel-runtime-corejs2/core-js
```
JLHwung: nit: I think we should also clean `packages/babel-runtime-corejs3/core-js(-stable)?`.

nicolo-ribaudo:
@JLHwung Why isn't the package.json enough?

jaydenseric:
@nicolo-ribaudo native ESM in Node.js requires all the import specifiers to have the full path to files, including the extension:

https://nodejs.org/api/esm.html#esm_mandatory_file_extensions

nicolo-ribaudo:
Thanks, I reopened the issue

Makefile
Leooo: Could that explain something like that appearing in the past hours?

```
Global error: Uncaught Error: Could not find module `@babel/runtime/helpers/esm/typeof.js` imported from `ember-exam/test-support/-private/ember-exam-qunit-test-loader` at http://localhost:7357/assets/vendor.js, line 252
```

PR #10768, opened 2019-11-26
"regenerator-runtime" import now can be controlled through include/exclude option

### LINKED ISSUE #10759: transform-runtime is not removing regenerator-runtime import added by preset-env
Hello!

Thank you for this great tool!

# Problem

However, I'm looking at situation when the `transform-runtime` plugin is not removing global `regenerator-runtime/runtime` import, which is introduced by the `preset-env` preset.

# Use case

We were working on some libraries and were using `preset-env` to actually transpile the code to the required target execution environment as well as to add needed global `core-js` and `regenerator-runtime` imports (usage mode). We were also using `transform-runtime` to optimize babel helper imports.

And now we would like to migrate from using global `regeneratorRuntime` usages to the explicitly imported ones (pure), so we've enabled `regenerator` option of the `transform-runtime`.

# More details

## Source code

```javascript
// src/main.js

export async function foo() {
  return await new Promise(resolve => setTimeout(resolve, 1000));
}
```

## Babel config

```json
// .babelrc

{
  "presets": [
    [
      "@babel/preset-env",
      {
        "targets": "IE 11",
        "useBuiltIns": "usage",
        "corejs": 3,
        "modules": false,
        "debug": true
      }
    ]
  ],
  "plugins": [
    [
    ]
  ],
  "exclude": "node_modules/**"
}
```

## Transpiled result

```javascript
// dist/main.js

import "core-js/modules/es.object.to-string";
import "core-js/modules/es.promise";

import "regenerator-runtime/runtime";
import _regeneratorRuntime from "@babel/runtime/regenerator";

export function foo() {
  return _regeneratorRuntime.async(function foo$(_context) {
    while (1) {
        case 0:
          _context.next = 2;
          return _regeneratorRuntime.awrap(new Promise(function (resolve) {
            return setTimeout(resolve, 1000);
          }));

        case 2:
          return _context.abrupt("return", _context.sent);

        case 3:
        case "end":
          return _context.stop();
      }
    }
  });
}
```

As you can see, the generated file contains both pure and global imports for `regenerator-runtime`, however, only explicit (pure) import should be left.

---

I believe, that `transform-runtime` should remove such global imports (isn't it it's purpose in the first place?), or at least there should be an option for the `preset-env` to skip adding the global import for `regenerator-runtime` (I couldn't find any).

1. What do you think about this issue?

2. Do you know of any workarounds I can use right now to actually remove such superfluous imports?

Thank you.

slavafomin:
| Fixed Issues?            | Fixes #10759
| Patch: Bug Fix?          | Sort of (see #10759)
| Documentation PR Link    | Will add if needed

When using `core-js` you can control what polyfills gets added or omitted from the generated code using `include`/`exclude` preset options. This PR adds the same functionality for `regenerator-runtime`.

With this, you could remove global `regenerator-runtime` imports by adding `regenerator-runtime` to the `exclude` option. This is required if you want to use `transform-runtime` to replace `regenerator-runtime` imports from global ones to a pure local.

This also allows to forcibly add import for `regenerator-runtime`, which could also be useful.

I will be glad to update the documentation as well if this PR will get accepted.

### DIFF
--- a/packages/babel-preset-env/src/index.js
+++ b/packages/babel-preset-env/src/index.js
@@ -47,9 +47,10 @@ const getPlugin = (pluginName: string) => {
 export const transformIncludesAndExcludes = (opts: Array<string>): Object => {
   return opts.reduce(
     (result, opt) => {
-      const target = opt.match(/^(es|es6|es7|esnext|web)\./)
-        ? "builtIns"
-        : "plugins";
+      const target =
+        opt === "regenerator-runtime" || opt.match(/^(es|es6|es7|esnext|web)\./)
+          ? "builtIns"
+          : "plugins";
       result[target].add(opt);
       return result;
     },
@@ -249,7 +250,13 @@ export default declare((api, opts) => {
     getOptionSpecificExcludesFor({ loose }),
     pluginSyntaxMap,
   );
+
   removeUnnecessaryItems(pluginNames, overlappingPlugins);
+
+  const regenerator =
+    (pluginNames.has("transform-regenerator") ||
+      include.builtIns.has("regenerator-runtime")) &&
+    !exclude.builtIns.has("regenerator-runtime");
 
   const polyfillPlugins = getPolyfillPlugins({
     useBuiltIns,
@@ -259,7 +266,7 @@ export default declare((api, opts) => {
     exclude: exclude.builtIns,
     proposals,
     shippedProposals,
-    regenerator: pluginNames.has("transform-regenerator"),
+    regenerator,
     debug,
   });
 
--- a/packages/babel-preset-env/src/normalize-options.js
+++ b/packages/babel-preset-env/src/normalize-options.js
@@ -54,6 +54,7 @@ const getValidIncludesAndExcludes = (
         ? [...Object.keys(corejs2Polyfills), ...defaultWebIncludes]
         : Object.keys(corejs3Polyfills)
       : []),
+    "regenerator-runtime",
   ]);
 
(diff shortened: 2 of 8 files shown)


JLHwung:
I feel like a dedicated `regenerator: boolean` option in `preset-env` will provide better ergonomics. 

The `transform-runtime` provides both `corejs` and `regenerator` to control how the helper imports will be rewritten. It feels more straight-forward now that we have `corejs` to control the `core-js` imports, and `regenerator` the other.

Technically `exclude: ["regenerator-runtime"]` is correct but I think it is more error-prone: what about `exclude: ["regenerator"]`? and less likely to get proper support from JSON schema, which can autocomplete `regenerator: boolean` options.

nicolo-ribaudo:
The good thing about the current name is that it can't be confused with the already valid `exclude: [transform-regenerator]`.

JMarkoski:
I think adding `exclude: ["regenerator-runtime"]` or `exclude: ["regenerator"]` will just add to the current confusion with using `usingBuiltIns` and polyfilling. I like @JLHwung idea, with adding `regenerator: boolean` which is the same config as in `@babel/transform-runtime`. Having two ways of adding similar config is redundant IMHO.

nicolo-ribaudo:
Regenerator is now handled as any other `@babel/runtime` helper (but we still support the legacy behavior for backward compatibility).

Additionally, with https://github.com/babel/babel-polyfills/ users can now choose to just enable polyfill injection (`babel-plugin-polyfill-corejs3`) without enabling `babel-plugin-polyfill-regenerator`.

Issue #10752, opened 2019-11-22
@babel/eslint-* v8


kaicataldo:
Making an issue here to track the work that needs to be done for the new @babel/eslint-* packages for the v8 release.

#### `@babel/eslint-parser` (taken from https://github.com/babel/babel-eslint/milestone/1):
* ~~[add support for preset & plugins in babelOptions](https://github.com/babel/babel-eslint/pull/784)~~ We can add this in a later minor version if need be.
* ~~[11.0.0-beta.0 - Private class methods no-undef error](https://github.com/babel/babel-eslint/issues/746)~~ This is not an issue specific to `babel-eslint` since the error is being thrown by `@babel-core` and the example code doesn't compile either.
* ~~[11.0.0-beta.0 - Private Instance Methods - Parsing error: params is not iterable](https://github.com/babel/babel-eslint/issues/749) - Dupe of https://github.com/babel/babel/issues/9506~~ https://github.com/babel/babel/pull/10914
* ~~[11.0.0-beta.0 - Typescript interface/type no-undef error](https://github.com/babel/babel-eslint/issues/760)~~ I don't think we should consider this a blocker for v8, since TS support is currently fairly limited. I think it makes sense to focus on making TS support more robust in the future, though I'm still of the mind that folks should just use `typescript-eslint`.
* ~~[Doesn't work with custom parser through `parserOverride`](https://github.com/babel/babel-eslint/issues/786)~~ I don't think we should consider this a blocker for v8.

#### `@babel/eslint-plugin`
* ~~https://github.com/babel/babel/issues/10904~~
* ~~Ensure `ClassPrivateProperty` works correctly with `no-unused-vars` and `no-undef`.~~ https://github.com/babel/babel/pull/10913
* ~~Can we create a shared config that enables the parser and corresponding rules automatically?~~ I no longer think this is a good idea because ESLint doesn't provide any runtime hooks to see what the user has configured. Turning off incompatible rules works, but we can't enable the corresponding rule in our plugin.
* ~~Audit current rules to see if they need to exist.~~ https://github.com/babel/babel/pull/10975 https://github.com/babel/babel/pull/10977
* ~~Remove deprecated rules?~~ https://github.com/babel/babel/pull/10975
* Figure out how we can be more proactive rather than reactive. Run core rules over fixtures that contain new experimental syntax? Fuzzer?
* https://github.com/babel/babel-eslint/issues/814

#### `@babel/parser`

[...]

PR #10774, opened 2019-11-28
Eslint config

swansontec:
| Fixed Issues?            | #10752 (partially)
| Patch: Bug Fix?          | Yes |
| Major: Breaking Change?  | No |
| Minor: New Feature?      | Yes |
| Tests Added + Pass?      | N/A |
| Documentation PR Link    | 
| Any Dependency Changes?  | No |

These commits add `recommended` ESLint configs to the two plugins. With this, users can enable the plugins and activate their rules in a single line.

While doing this work, I noticed that the documentation for `@babel/eslint-plugin` had some typos, so I fixed those.

I also noticed that `@babel/eslint-plugin` exports a `rulesConfig`, which was [removed in ESLint 2](https://github.com/eslint/eslint/blob/master/docs/user-guide/migrating-to-2.0.0.md#plugins-no-longer-have-default-configurations). I left that in with a comment, since removing it would be a breaking change, but then again this whole package is being renamed, so maybe now is a good time to remove it.

### DIFF
--- a/eslint/babel-eslint-plugin-development/README.md
+++ b/eslint/babel-eslint-plugin-development/README.md
@@ -36,3 +36,13 @@ Then, load the plugin in your `.eslintrc` configuration file. You can omit the `
   value, since it might cause problem if `undefined` is redeclared.
 * `@babel/development/plugin-name`: Requires plugins to have a `name` property, which
   can be useful for debugging purposes.
+
+## Recommended Configuration
+
+You can enable this plugin and all its rules by extending the recommended configuration:
+
+```json
+{
+  "extends": ["plugin:@babel/development/recommended"]
+}
+```
--- a/eslint/babel-eslint-plugin-development/src/index.js
+++ b/eslint/babel-eslint-plugin-development/src/index.js
@@ -8,4 +8,14 @@ export default {
     "no-undefined-identifier": noUndefinedIdentifier,
     "plugin-name": pluginName,
   },
+  configs: {
+    recommended: {
+      plugins: ["@babel/development"],
+      rules: {
+        "@babel/development/no-deprecated-clone": "error",
+        "@babel/development/no-undefined-identifier": "error",
+        "@babel/development/plugin-name": "error",
+      },
+    },
+  },
 };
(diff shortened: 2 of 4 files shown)


swansontec:
Hrm, the test failure seems unrelated to these changes. I'll try a rebase once `master` is clean.

swansontec:
Thanks for the quick reply!

> Thanks for the PR. For some context, all of the ESLint-related packages are in the process of being ported over to the monorepo and being updated so that we can publish them under the new name in a hopefully less broken and much easier to maintain state.

I noticed that this work was ongoing, and tried to contribute along those lines.

> I’m hoping to remove all the deprecated rules, audit the remaining ones to see if they’re still necessary,

I can update this PR to do that if you like.

> and try to come up with a better strategy for discovery of where core rules break with experimental syntax.

Yeah, that's a tricky one. You need to exercise the ESLint core rules on some codebase that includes the experimental features, but you also need to know what the "correct" outcome for those rules should be. If TC39 allows dangling commas in some new location, for example, a rule like `no-comma-dangle` wouldn't actually realize it needs to check there. It's not clear how to automate this, besides the "million monkeys" approach of just waiting for bug reports to come in.

> The readme was intentionally left in an incomplete state because I was assuming we’d have to rewrite it all anyway 🙂

It's not that bad, actually, especially now with the fixes.

> My end goal is to have a way to configure Babel-related rules and the parser with a single config, but I don’t think we should add the recommended config until after we do the other work and can take stock of what we want this to actually look like.

The config I added does this. The only question at this point is what the final rule-set should be.

kaicataldo:
I’m just starting in on this process and need to be spend some time investigating before I come up with an actual proposal.

Just so you’re aware, my current priorities are:

1. Fix bugs in`@babel/eslint-parser` and finalize the API.
1. Investigate the current status of `@babel/eslint-plugin` and do the work I discussed above.
1. Audit and make sure `@babel/eslint-plug-in-development` is in a good place.

My inclination right now is to hold off on merging this and work on the parser first. Once we have the parser in a good place, we’ll be able to evaluate what the plugin should do.

Would love to have your help along the way if you’re interested! I’ll see if I can make a Slack channel for this work later today :)

PR #10776, opened 2019-11-29
fix: wrong node name when add LabeledStatement to Program scope.globals.

lovetingyuan:
| Tests Added + Pass?      | Not added
| Documentation PR Link    | 

### DIFF
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -107,7 +107,7 @@ const collectorVisitor = {
   },
 
   LabeledStatement(path) {
-    path.scope.getProgramParent().addGlobal(path.node);
+    path.scope.getProgramParent().addGlobal(path.node.label);
     path.scope.getBlockParent().registerDeclaration(path);
   },
 


nicolo-ribaudo:
What are the old and new behaviours?

lovetingyuan:
input
```
test()
foo: bar.name
```
### old
visitor
```javascript
{
  Program(path) {
   assert.equal(Object.keys(path.scope.globals), ['test', 'undefined', 'bar'])
 }
}
```
### new
```javascript
{
  Program(path) {
   assert.equal(Object.keys(path.scope.globals), ['test', 'foo', 'bar'])
 }
}
```

nicolo-ribaudo:
I think that the correct result should be `['test', 'bar']`: labels aren't global variables.

lovetingyuan:
Yes, I think almost nobody still use label statement.
However, babel call `addGlobal` in `LabeledStatement`, I am not clear the reason to do that.
Thanks.

JLHwung:
@lovetingyuan This code has been here since babel 6.0. Since it is always `undefined`, I don't think removing this would break anyone practically.

lovetingyuan:
Yes, I think it does not matter, feel free to close this PR.
Thanks.

nicolo-ribaudo:
@lovetingyuan I would still accept a PR which produces the output at https://github.com/babel/babel/pull/10776#issuecomment-559779436

PR #10778, opened 2019-11-29
refactor: Improve error message in @babel/core

jaroslav-kubicek:
Hi 👋 

I would like to propose updating the error message when running babel with `rootMode:"upward"` so user gets feedback on what's expected.

Motivation:

Given I have a repository and I migrated recently to babel 7.x and still using just `.babelrc` for configuration. I ran a script provided to me by a third-party library which tried to search for my config but failed with this error. I would like to understand more about what is expected of me from an error message.

### DIFF
--- a/packages/babel-core/src/config/partial.js
+++ b/packages/babel-core/src/config/partial.js
@@ -12,7 +12,12 @@ import {
   type RootMode,
 } from "./validation/options";
 
-import { findConfigUpwards, type ConfigFile, type IgnoreFile } from "./files";
+import {
+  findConfigUpwards,
+  ROOT_CONFIG_FILENAMES,
+  type ConfigFile,
+  type IgnoreFile,
+} from "./files";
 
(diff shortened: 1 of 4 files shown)


JLHwung:
@jaroslav-kubicek Thanks for fixing the outdated test fixtures. I have cherry-picked your fix to the master branch. Could you rebase?

packages/babel-core/src/config/partial.js
```diff
@@ -31,7 +36,9 @@ function resolveRootMode(rootDir: string, rootMode: RootMode): string {
       throw Object.assign(
         (new Error(
           `Babel was run with rootMode:"upward" but a root could not ` +
-            `be found when searching upward from "${rootDir}"`,
+            `be found when searching upward from "${rootDir}".` +
```
nicolo-ribaudo:
```suggestion
            `be found when searching upward from "${rootDir}".\n` +
```

nicolo-ribaudo:
Thanks!

PR #10781, opened 2019-11-29
Tune makefile scripts

JLHwung:

In this PR we wrap `node` call into a `yarn` call so it can be future-proof on upcoming yarn 2 upgrade.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,9 @@ EMPTY :=
 SPACE := $(EMPTY) $(EMPTY)
 COMMA_SEPARATED_SOURCES = $(subst $(SPACE),$(COMMA),$(SOURCES))
 
+YARN := yarn --silent
+NODE := $(YARN) node
+
 
 .PHONY: build build-dist watch lint fix clean test-clean test-only test test-ci publish bootstrap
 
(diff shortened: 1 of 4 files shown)


scripts/test-cov.sh
```diff
@@ -1,10 +1,11 @@
...
 
+node="yarn node"
 jestArgs="--coverage"
 
 if [ -n "$CI" ]; then
   jestArgs="${jestArgs} --maxWorkers=4 --ci"
 fi
 
-node_modules/.bin/jest $jestArgs
+$node node_modules/.bin/jest $jestArgs
```
JLHwung: We leave `jest` to be relative path as we plan to unplug jest on yarn 2 so that it can work with `--inspect-brk`.

Makefile
```diff
@@ -14,6 +14,9 @@ EMPTY :=
 SPACE := $(EMPTY) $(EMPTY)
 COMMA_SEPARATED_SOURCES = $(subst $(SPACE),$(COMMA),$(SOURCES))
 
+YARN := yarn --silent
```
JLHwung: This will suppress the `yarn foo v1.2.3` header printed by yarn. I think it is okay to use `$(YARN)` when we use it to invoke another node binaries, e.g. eslint/prettier.

PR #10788, opened 2019-12-01
Do not transpile typeof helper with itself in babel/runtime

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  |

I noticed that the `typeof.js` helper also contained a copy of itself as `_typeof2`: https://unpkg.com/browse/@babel/runtime@7.7.4/helpers/typeof.js
This PR removes it:
```js
function _typeof(obj) {
  if (typeof Symbol === "function" && typeof Symbol.iterator === "symbol") {
    module.exports = _typeof = function _typeof(obj) {
      return typeof obj;
    };
  } else {
    module.exports = _typeof = function _typeof(obj) {
      return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj;
    };
  }

  return _typeof(obj);
}

module.exports = _typeof;
```

In other helpers it is still transformed:
```js
var _typeof = require("../helpers/typeof");

var toPrimitive = require("./toPrimitive");

function _toPropertyKey(arg) {
  var key = toPrimitive(arg, "string");
  return _typeof(key) === "symbol" ? key : String(key);
}

module.exports = _toPropertyKey;
```

### DIFF
--- a/packages/babel-plugin-transform-runtime/scripts/build-dist.js
+++ b/packages/babel-plugin-transform-runtime/scripts/build-dist.js
@@ -158,7 +158,7 @@ function buildHelper(
     presets: [
       [
         "@babel/preset-env",
-        { modules: false, exclude: ["transform-typeof-symbol"] },
+        { modules: false, exclude: ["@babel/plugin-transform-typeof-symbol"] },
       ],
     ],
     plugins: [
@@ -177,7 +177,6 @@ function buildHelper(
         exclude: /typeof/,
         plugins: ["@babel/plugin-transform-typeof-symbol"],
       },
-      {},
     ],
   }).code;
 }
--- a/packages/babel-plugin-transform-runtime/scripts/build-dist.js
+++ b/packages/babel-plugin-transform-runtime/scripts/build-dist.js
@@ -151,8 +151,6 @@ function buildHelper(
   );
   tree.body.push(...helper.nodes);
 
-  console.log(helperFilename);
-
   return babel.transformFromAst(tree, null, {
     filename: helperFilename,
     presets: [
--- a/packages/babel-plugin-transform-runtime/scripts/build-dist.js
+++ b/packages/babel-plugin-transform-runtime/scripts/build-dist.js
@@ -9,6 +9,7 @@ const t = require("@babel/types");
 
 const transformRuntime = require("../");
 
+const runtimeVersion = require("@babel/runtime/package.json").version;
 const corejs2Definitions = require("../lib/runtime-corejs2-definitions").default();
 const corejs3Definitions = require("../lib/runtime-corejs3-definitions").default();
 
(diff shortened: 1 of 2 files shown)


packages/babel-plugin-transform-runtime/scripts/build-dist.js
```diff
@@ -150,16 +151,34 @@ function buildHelper(
...
         path.relative(path.dirname(helperFilename), pkgDirname),
         helperName
       ),
     ],
+    overrides: [
+      {
+        exclude: /typeof/,
+        plugins: ["@babel/plugin-transform-typeof-symbol"],
+      },
+      {},
```
JLHwung: Why empty object here?

packages/babel-plugin-transform-runtime/scripts/build-dist.js
```diff
@@ -150,16 +151,34 @@ function buildHelper(
...
 
+  console.log(helperFilename);
+
   return babel.transformFromAst(tree, null, {
-    presets: [[require("@babel/preset-env"), { modules: false }]],
+    filename: helperFilename,
+    presets: [
+      [
+        "@babel/preset-env",
+        { modules: false, exclude: ["transform-typeof-symbol"] },
```
JLHwung:
```suggestion
        { modules: false, exclude: ["@babel/plugin-transform-typeof-symbol"] },
```

packages/babel-plugin-transform-runtime/scripts/build-dist.js
```diff
@@ -150,16 +151,33 @@ function buildHelper(
   );
   tree.body.push(...helper.nodes);
 
+  console.log(helperFilename);
```
JLHwung: Oops.

nicolo-ribaudo:
CircleCI is failing because of core-js 3.4.6. It's fixed in 3.4.7, but I don't know how to clear the cache before triggering a build.

Andarist:
Could we exclude this plugin altogether and only allow using `typeof` helper as an explicit dependency of other helpers when needed? 

Even after this change, the output is slightly wrong which can be observed in the pasted in `toPropertyKey`, which actually should look like this (if it should reuse `typeof` helper):
```diff
-var _typeof = require("../helpers/typeof");
+var _typeof = require("./typeof");

var toPrimitive = require("./toPrimitive");

function _toPropertyKey(arg) {
  var key = toPrimitive(arg, "string");
  return _typeof(key) === "symbol" ? key : String(key);
}

module.exports = _toPropertyKey;
```

This is especially important for ESM-flavour of those helpers - right now ESM helpers reference a CJS file by accident.

nicolo-ribaudo:
I have a better alternative to this PR.

nicolo-ribaudo:
Nevermind, it doesn't work. (I was trying to rewrite the helper so that it wouldn't be transformed by the `typeof` plugin)

PR #10790, opened 2019-12-02
Use chrome data when android is absent

### LINKED ISSUE #10789: Preset-env loads unnecessary polyfills for Android >= 52
## Bug Report

**Current Behavior**
The following snippet is transpiled

**Input Code**
- REPL or Repo link if applicable: https://babeljs.io/repl#?browsers=Android%20%3E%3D%2052&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAECC0G8C-Q&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.7.4&externalPlugins=

```js
class A {};
```

**Expected behavior/code**
It is not transpiled on `Chrome >= 52`, so it should not be transpiled when `Android >= 52`, which shares the same language features support with Chrome.

- Filename: `babel.config.js`

```json
{
  "presets": ["@babel/env", { "targets": "Android >= 52" }]
}
```

**Environment**
REPL

JLHwung:
| Fixed Issues?            | Fixes #10789 

In this PR we introduce a `getLowestImplementedVersion` which adds chrome support data as fallback when `android` support data is absent, by doing preset-env would not introduce unnecessary transforms for `Android >= 52`, which should share the same features sets as `Chrome >= 52`.

The previous behaviour is revealed in this [CI failure](https://travis-ci.com/JLHwung/babel/jobs/261977352#L1085). This behaviours renders settings like `last 2 versions` unusable and preset-env will thus apply all the transforms.


nicolo-ribaudo:
Maybe this should be added to https://github.com/browserslist/browserslist/blob/4da6f9c683ce27816f8804663c4cd29ef38d01a7/index.js#L454?

zloirock:
The existence of any `android` versions greater than 4.4.4 in `browserslist` looks like a bug.

JLHwung:
According to the caniuse description, the `android` whose version are greater than `4.4.4` refers to Android 5-6.x WebView. We could propose to add a new `android-webview` agent to caniuse-db but I think a single `android` works fine here as long as the version does not overlap with Android Browser.

My first approach was somehow similar to core-js-compat: map `android >= 36` to `chrome >= 36`. But I realize that it would make queries like `android >= 4` converted to `{ android: "4.0.0", chrome: "76.0.0" }` so I switch to current implementation.

@zloirock Wouldn't current approach in `core-js-compat` map `android 5` to `chrome 5`? I think we should emit warnings for `5 <= android < 30` in which both WebView and Android Browser are not available.

zloirock:
@JLHwung `browserslist` will throw an error on `android 5`, so not.

For Android Chrome and WebView we have `and_chr` -> `chrome`.

See https://github.com/kangax/compat-table/issues/801

packages/babel-preset-env/test/fixtures/debug/browserslists-last-2-versions-not-ie/options.json
```diff
@@ -0,0 +1,20 @@
...
+    [
+      "env",
+      {
+        "debug": true,
+        "targets": {
+          "browsers": [
+            "last 2 versions",
+            "not ie <= 11",
+            "not ie_mob <= 11",
+            "not op_mob <= 12.1"
```
JLHwung: This shouldn't be needed. It is a bug of browserslist that returns `op_mob 12.1` for `last 2 versions`, which should be instead `op_mob 59, op_mob 60`.

packages/babel-preset-env/test/fixtures/debug/browserslists-last-2-versions-not-ie/stdout.txt
```diff
@@ -0,0 +1,31 @@
...
+  "firefox": "68",
+  "ios": "13",
+  "opera": "46",
+  "safari": "12.1",
+  "samsung": "9.2"
+}
+
+Using modules transform: auto
+
+Using plugins:
```
JLHwung: The plugins listed here is generated from a sort-of modern browserslist configuration. It can help us prioritize bug fixes in the current transforms. because these plugins are expected to be used for a longer time than other standard transforms.

packages/babel-preset-env/test/fixtures/debug/browserslists-last-2-versions-not-ie/stdout.txt
```diff
@@ -0,0 +1,31 @@
+@babel/preset-env: `DEBUG` option
+
+Using targets:
+{
+  "android": "77",
+  "chrome": "77",
```
JLHwung: Recently the chrome 49 usage just drops to 0.5% and therefore excluded from browserslists defaults `> 0.5%` now.

PR #10801, opened 2019-12-03
Use scope flags to check arguments

### LINKED ISSUE #10771: Use of `arguments` should be permitted even in a field initializer, if the field is initialized with a class expression.
## Bug Report

**Current Behavior**
Any use of `arguments` within a class field initializer leads to `'arguments' is not allowed in class field initializer`, even when such a limitation appears unwarranted - e.g. if the class field is initialized with a class expression and `arguments` appears inside one of the declared class' methods.

**Input Code**
```js
class Voodoo {
  static magic = {
    wizardry: class extends Witchcraft {
      cast() {
        super.cast(...arguments)
      }
    }
  }
}
```
See also [babeljs.io/repl](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEBqD28AmjoG8BQ1oQC5lwEthoBbMAc2OgF50ttoB3QgLzACckOBPALmihIMAKYAPXCIB2SGAHVCuYAAtgHMADNc9Ro2CRcACgCUO3YwgBXAA4iOAOn15D915wqXS03BGMNdAL7-0EHYQQFAA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Cstage-2&prettier=false&targets=&version=7.7.4&externalPlugins=)

**Expected behavior/code**
The example code should transpile without error; the class field initializer scope should not prevent use of `arguments` in a field initializer once it has been bound to some scope declared within the initializer and can no longer refer to the original class' constructor arguments.

JLHwung:
| Fixed Issues?            | Fixes #10771 

In this PR we introduce `inClass` accessor in `ScopeHandler` and use it to check `arguments` keyword. Please refer to this [thread](https://github.com/babel/babel/issues/10771#issuecomment-559277848) for the background.

### DIFF
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -2208,7 +2208,11 @@ export default class ExpressionParser extends LValParser {
       }
     }
 
-    if (this.state.inClassProperty && word === "arguments") {
+    if (
+      this.scope.inClass &&
+      !this.scope.inNonArrowFunction &&
+      word === "arguments"
+    ) {
       this.raise(
         startLoc,
         "'arguments' is not allowed in class field initializer",
--- a/packages/babel-parser/src/util/scope.js
+++ b/packages/babel-parser/src/util/scope.js
@@ -64,6 +64,9 @@ export default class ScopeHandler<IScope: Scope = Scope> {
   get allowDirectSuper() {
     return (this.currentThisScope().flags & SCOPE_DIRECT_SUPER) > 0;
   }
+  get inClass() {
+    return (this.currentThisScope().flags & SCOPE_CLASS) > 0;
+  }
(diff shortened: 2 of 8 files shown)


nicolo-ribaudo:
What if we have `arguments` inside an arrow function inside a class field?

JLHwung:
@nicolo-ribaudo It has been tested in https://github.com/babel/babel/blob/e94da0dce5f3101cb3a77f78e1b6802ddf2c6cc9/packages/babel-parser/test/fixtures/experimental/class-properties/arguments-in-arrow-function/input.js

PR #10806, opened 2019-12-04
fix(optional chaining): Optional delete returns true with nullish base

### LINKED ISSUE #10805: Optional delete must return true when base is nullish
## Bug Report

**Current Behavior**

When using optional delete to delete a property of a nullish base, the delete operation returns `undefined`. It should return `true`

**Input Code**
[repl](https://babeljs.io/repl/#?browsers=Chrome%2078&build=&builtIns=usage&spec=false&loose=false&code_lz=DYUwLgBAZg9jDcAoAxjAdgZ0gJ3BAvBACYihgjRwD8AdAEYCG2SK6GMoNDGGI2YAClyR8oiGGwBXEAEp4QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015&prettier=false&targets=&version=7.7.4&externalPlugins=%40babel%2Fplugin-proposal-optional-chaining%407.6.0)

```js
let foo;
const ret = delete foo?.bar;

console.assert(ret === true);
```

**Expected behavior/code**
The `ret` value should be `true`, not `undefined`.

- Filename: `babel.config.js`

```js
{
  "plugins": ["@babel/plugin-proposal-optional-chaining"]
}
```

**Possible Solution**
When replacing optional deletes, we must use `t.booleanLiteral(true)` in the consequent (truthy branch of a ternary expression) instead of `scope.buildUndefinedNode()`.

Specifically, we must fix https://github.com/babel/babel/blob/d18afbd7c3e977208194acbea4c0de0313a6e1eb/packages/babel-plugin-proposal-optional-chaining/src/index.js#L116

mpaarating:
| Fixed Issues?            | Fixes #10805
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Per issue #10805, the return value when using delete on a nullish base is
currently undefined. The correct return type should be true.

### DIFF
--- a/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/delete/output.js
+++ b/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/delete/output.js
@@ -7,7 +7,7 @@ const obj = {
     b: 0
   }
 };
-let test = obj === null || obj === void 0 ? void 0 : (_obj$a = obj.a) === null || _obj$a === void 0 ? void 0 : delete _obj$a.b;
-test = obj === null || obj === void 0 ? void 0 : delete obj.a.b;
-test = obj === null || obj === void 0 ? void 0 : (_obj$b = obj.b) === null || _obj$b === void 0 ? void 0 : delete _obj$b.b;
-obj === null || obj === void 0 ? void 0 : delete obj.a;
+let test = obj === null || obj === void 0 ? true : (_obj$a = obj.a) === null || _obj$a === void 0 ? true : delete _obj$a.b;
+test = obj === null || obj === void 0 ? true : delete obj.a.b;
+test = obj === null || obj === void 0 ? true : (_obj$b = obj.b) === null || _obj$b === void 0 ? true : delete _obj$b.b;
+obj === null || obj === void 0 ? true : delete obj.a;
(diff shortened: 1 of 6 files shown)


packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/delete/output.js
```diff
@@ -7,7 +7,7 @@ const obj = {
     b: 0
   }
 };
-let test = obj === null || obj === void 0 ? void 0 : (_obj$a = obj.a) === null || _obj$a === void 0 ? void 0 : delete _obj$a.b;
-test = obj === null || obj === void 0 ? void 0 : delete obj.a.b;
-test = obj === null || obj === void 0 ? void 0 : (_obj$b = obj.b) === null || _obj$b === void 0 ? void 0 : delete _obj$b.b;
-obj === null || obj === void 0 ? void 0 : delete obj.a;
+let test = obj === null || obj === void 0 ? true : (_obj$a = obj.a) === null || _obj$a === void 0 ? true : delete _obj$a.b;
```
jridgewell: Excellent!

PR #10811, opened 2019-12-04
Add optional-chaining and nullish-coalescing to preset-env

Druotic:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #10809 
| Patch: Bug Fix?          | N
| Major: Breaking Change?  | N
| Minor: New Feature?      | Y
| Tests Added + Pass?      | Y
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

PR #10814, opened 2019-12-04
chore: map mobile browser data to their desktop version

JLHwung:
| Fixed Issues?            | Fixes issue mentioned in [here](https://github.com/babel/babel/pull/10811#discussion_r354005444)
| Patch: Bug Fix?          | Yep

In this PR we map the ES Module support data for mobile browser to their desktop versions. caniuse now preserve only latest version for the mobile browser other than `iOS Safari`. It fixes the issue that when `caniuse-db` is upgraded, the `and_chr` and `and_ff` will be incorrectly bumped.

### DIFF
--- a/packages/babel-compat-data/data/built-in-modules.json
+++ b/packages/babel-compat-data/data/built-in-modules.json
@@ -6,7 +6,6 @@
     "safari": "10.1",
     "opera": "48",
     "ios_saf": "10.3",
-    "android": "61",
     "op_mob": "48",
     "and_chr": "61",
     "and_ff": "60",
--- a/packages/babel-compat-data/data/built-in-modules.json
+++ b/packages/babel-compat-data/data/built-in-modules.json
@@ -6,6 +6,7 @@
     "safari": "10.1",
     "opera": "48",
     "ios_saf": "10.3",
+    "android": "61",
     "op_mob": "48",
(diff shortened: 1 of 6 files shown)


JLHwung:
@nicolo-ribaudo It doesn't. Actually it works _like_ https://github.com/browserslist/browserslist/pull/428 by replacing the caniuse mobile desktop agents by their desktop versions.

nicolo-ribaudo:
The upstream fix has been released

JLHwung:
@nicolo-ribaudo unskipping `android` will introduce `android 61` to the browser sets from `esmodules: true`, which does not work correctly until #10790 since `android >= 37` data is absent (and of course unnecessary to be present) from compat-table.

Therefore, I suggest we merge it as-is.

packages/babel-preset-env/test/targets-parser.spec.js
```diff
@@ -120,7 +120,7 @@ describe("getTargets", () => {
     it("does throws on unsupported versions", () => {
       expect(() => {
         getTargets({
-          browsers: "node 15.0.0, chrome 1000",
+          browsers: "node 1000.0.0, chrome 1000",
```
JLHwung: Time flies: node 15 is scheduled at Oct. 2020.

PR #10815, opened 2019-12-04
Add support for nullish coalescing operator in preset-env

Druotic:
| Fixed Issues?            |  
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |


Druotic:
@nicolo-ribaudo You motivated me 😄https://github.com/babel/babel/pull/10811#issuecomment-561861646

nicolo-ribaudo:
Ok sorry, I meant if you could do it in that same PR :sweat_smile: 

Because since alphabetically the two plugins are one after the other, they are in the same place in the various lists and thus every single line of this PR would conflict with the other PR.

Druotic:
Ah, okay - I thought with trailing commas always being in place, no collisions would happen. It's easy enough to combine, though. Updating to the other PR incoming.

PR #10823, opened 2019-12-05
chore: remove unnecessary transform-for-of when building babel-parser

JLHwung:

<s>This PR applies `transform-for-of: assumeArray` on standalone build, of which the size decrease is expected. </s>

It removes this transform from babel-parser build config since it is supported in Node >= 6.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -14,6 +14,7 @@ module.exports = function(api) {
 
   let convertESM = true;
   let ignoreLib = true;
+  let isStandalone = false;
   const nodeVersion = "6.9";
   // The vast majority of our src files are modules, but we use
   // unambiguous to keep things simple until we get around to renaming
@@ -45,6 +46,7 @@ module.exports = function(api) {
     case "standalone":
       convertESM = false;
       ignoreLib = false;
+      isStandalone = true;
       unambiguousSources.push(
         "**/node_modules",
         "packages/babel-preset-env/data"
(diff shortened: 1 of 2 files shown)


babel.config.js
```diff
@@ -101,14 +103,12 @@ module.exports = function(api) {
       ["@babel/plugin-proposal-nullish-coalescing-operator", { loose: true }],
 
       convertESM ? "@babel/transform-modules-commonjs" : null,
+      isStandalone ? ["@babel/transform-for-of", { assumeArray: true }] : null,
```
nicolo-ribaudo: I'd prefer that we also apply this transform when running tests outside of `@babel/standalone`, otherwise we risk introducing a `for-of` with a `Map` which isn't caught by the tests.

nicolo-ribaudo: Actually, this probably already breaks part of `@babel/core` and all the modules plugins.

JLHwung: That's an interesting usage of this plugin! I will try it later.

nicolo-ribaudo: Yup, it doesn't work:

https://babeljs.io/repl/build/13296#?browsers=Chrome%2023&build=&builtIns=false&spec=false&loose=false&code_lz=JYWwDg9gTgLgBAbzgDzgXzgMyhEcBEARvgNwBQZApspLHAG4CGUcjcAvCiUA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015&prettier=false&targets=&version=7.7.4%2Bpr.10823&externalPlugins=

JLHwung:
@nicolo-ribaudo I revised this PR to be minimal: It only removes unnecessary `transform-for-of` for babel-parser.

nicolo-ribaudo:
How much bigger it the new bundle size? (Since it will now use the "default" transform for `for-of`)

JLHwung:
@nicolo-ribaudo The unminified babel-preset-env.js increases from 3.0MB to 3.1MB. I will work on a fix later.

PR #10825, opened 2019-12-06
babel-eslint-parser: remove allowImportExportEverywhere

kaicataldo:
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

I'm actually not entirely sure when this was removed, but it doesn't look like this is an option we have in Espree any more (though it still exists in [Acorn](https://github.com/acornjs/acorn/tree/master/acorn#interface)) and I don't think is necessary any more? Please let me know if I'm missing something!

### DIFF
--- a/eslint/babel-eslint-parser/src/parse.js
+++ b/eslint/babel-eslint-parser/src/parse.js
@@ -26,7 +26,7 @@ export default function(code, options) {
     ignore: options.babelOptions.ignore,
     only: options.babelOptions.only,
     parserOpts: {
-      allowImportExportEverywhere: options.allowImportExportEverywhere, // consistent with espree
+      allowImportExportEverywhere: true,
       allowReturnOutsideFunction: true,
       allowSuperOutsideMethod: true,
       ranges: true,
@@ -34,7 +34,7 @@ export default function(code, options) {
       plugins: ["estree"],
     },
     caller: {
-      name: "babel-eslint",
+      name: "@babel/eslint-parser",
     },
   };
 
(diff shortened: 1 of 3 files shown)


packages/babel-parser/test/fixtures/experimental/class-properties/arguments-in-nested-class-decorator-call-expression/options.json
```diff
@@ -1,3 +1,6 @@
 {
-  "plugins": ["classProperties", ["decorators", { "decoratorsBeforeExport": false }]]
+  "plugins": [
```
kaicataldo: Fixed by running `npm run fix`

eslint/babel-eslint-parser/src/parse.js
```diff
@@ -26,15 +26,15 @@ export default function(code, options) {
     ignore: options.babelOptions.ignore,
     only: options.babelOptions.only,
     parserOpts: {
-      allowImportExportEverywhere: options.allowImportExportEverywhere, // consistent with espree
```
JLHwung: This option was introduced at https://github.com/babel/babel-eslint/pull/329 and intended to pass through to `@babel/parser`, so the comment here might be a little bit confusing. You can change it to `// consistent with babel-parser`.

Therefore, I would say we leave it as-is, since there is no strong reason that we should drop this option.

kaicataldo: Ah okay, that makes sense. Thanks!

Issue #10826, opened 2019-12-06
@babel/parser: differences between Espree when using ESTree plugin


kaicataldo:
## Bug Report

**Current Behavior**
There are a few cases for newer where `@babel/parser` deviates from ESTree. These should be aligned so that `@babel/eslint-parser` works correctly with existing ESLint rules.

##### Dynamic Imports
`@babel/parser` currently parses dynamic imports (`import('a')`) as `CallExpression`s (code [here](https://github.com/babel/babel/blob/master/packages/babel-parser/src/parser/expression.js#L784-L802)). The ESTree spec differs from this and gives it its own type called [`ImportExpression`](https://github.com/estree/estree/blob/master/es2020.md#expressions). 

##### BigInt Literals
`@babel/parser` currently parses BigInt Literals (`1n`) as `BigIntLiteral`. The ESTree spec differs from this and gives it the type `Literal` and adds the `bigint` property (see spec [here](https://github.com/estree/estree/blob/master/es2020.md#bigintliteral)).

**Input Code**
N/A

**Expected behavior/code**
N/A

**Babel Configuration (babel.config.js, .babelrc, package.json#babel, cli command, .eslintrc)**
N/A

**Environment**
N/A

**Possible Solution**
This should be fixed in the ESTree plugin for `@babel/parser`.

kaicataldo:
These issues have been fixed.

PR #10827, opened 2019-12-06
@babel/eslint-parser: fix BigIntLiteral node to match ESTree spec

kaicataldo:
| Fixed Issues?            | Refs https://github.com/babel/babel/issues/10826
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  | Yes, this adds the `@babel/plugin-syntax-bigint` plugin to the `@babel/eslint-parser` package.

`@babel/parser` currently parses BigInt Literals (`1n`) as `BigIntLiteral`. The ESTree spec differs from this and gives it the type `Literal` and adds the `bigint` property (see spec [here](https://github.com/estree/estree/blob/master/es2020.md#bigintliteral)).

Additionally, Espree tokenizes BigInt Literals as type `"Numeric"`. 

I have tested this in Node 8 (no `BigInt` support) and Node 12, and these are the nodes that are created, respectively:

#### Node 8
```js
{
  type: 'Literal',
  start: 10,
  end: 12,
  loc: {
    start: Position { line: 1, column: 10 },
    end: Position { line: 1, column: 12 } 
  },
  range: [ 10, 12 ],
  value: null,
  raw: '1n',
  bigint: '1' 
}
```

#### Node 12
```js
{
  type: 'Literal',
  start: 10,
  end: 12,
  loc: {
    start: Position { line: 1, column: 10 },
    end: Position { line: 1, column: 12 }
  },
  range: [ 10, 12 ],
  value: 1n,
  raw: '1n',
  bigint: '1'
}
```


packages/babel-parser/src/plugins/estree.js
```diff
@@ -31,6 +33,24 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+      try {
+        bigInt = BigInt(value);
+      } catch (e) {
+        // https://github.com/estree/estree/blob/master/es2020.md#bigintliteral
+        // In environments that don't support BigInt, this value
+        // will be null.
+      }
+
+      // Espree creates a "Numeric" token for BigIntLiterals.
+      this.state.type = tt.num;
```
kaicataldo: Is this safe?

nicolo-ribaudo: Maybe it would be safer to overwrite the last element of the `this.state.tokens` array in the overwritten `parseLiteral` method?

kaicataldo: Oh, I like that! Will update.

kaicataldo: I hadn't realized we do ESTree token conversion in `babel-eslint` - will update doing that instead!

eslint/babel-eslint-parser/src/babylon-to-espree/convertToken.js
```diff
@@ -77,6 +77,9 @@ export default function(token, tt, source) {
       flags: value.flags,
     };
     token.value = `/${value.pattern}/${value.flags}`;
+  } else if (type === tt.bigint) {
```
kaicataldo: Muuuuuch better 😄

packages/babel-parser/test/helpers/runFixtureTests.js
```diff
@@ -41,6 +41,11 @@ class FixtureError extends Error {
   }
 }
 
+// https://github.com/estree/estree/blob/master/es2020.md#bigintliteral
+function bigIntReplacer(key, value) {
```
kaicataldo: `BigInt`s can't be serialized. In the ESTree spec, `BigInt` values are set to `null` when the type isn't supported by an environment, which seemed like a reasonable behavior. Another option we have is to use `BigInt.prototype.toString()` - thoughts on this? Note that with the latter, `1n` will become just `1`, so if we do that, we might want to add the `n` to it. Not sure if this would have some negative side effects though, since we're altering the type.

packages/babel-parser/test/helpers/runFixtureTests.js
```diff
@@ -107,19 +109,43 @@ export function runThrowTestsWithEstree(fixturesPath, parseFunction) {
...
+  }
+
+  for (const obj of notJSONparseableObj) {
+    const { toJSON } = obj.prototype;
+    originalToJSONMap.set(obj, toJSON);
+    obj.prototype.toJSON = function(...args) {
+      let isBigInt = false;
+
+      try {
+        isBigInt = typeof obj("1") === "bigint";
```
kaicataldo: Is there a better way to check this?

packages/babel-parser/test/helpers/runFixtureTests.js
```diff
@@ -107,19 +109,33 @@ export function runThrowTestsWithEstree(fixturesPath, parseFunction) {
...
+
+  if (typeof BigInt !== "undefined") {
+    notJSONparseableObj.push(BigInt);
+  }
+
+  for (const obj of notJSONparseableObj) {
+    const { toJSON } = obj.prototype;
+    originalToJSONMap.set(obj, toJSON);
+    obj.prototype.toJSON = function() {
+      return this.toString();
```
kaicataldo: We need to ensure we're not passing parameters through here because `BigInt.prototype.toString()` takes a radix argument that must be between `[2-36]` and throws if it gets called with a different argument. It has a default value, though, so it's fine to not call it with anything.

packages/babel-parser/test/helpers/runFixtureTests.js
```diff
@@ -1,3 +1,5 @@
+/* global BigInt */
```
kaicataldo: We'll be able to remove this when we enable the latest version of `@babel/eslint-config-internal`.

packages/babel-parser/test/fixtures/estree/bigInt/basic/output.json
```diff
@@ -0,0 +1,102 @@
...
+                "start": {
+                  "line": 1,
+                  "column": 10
+                },
+                "end": {
+                  "line": 1,
+                  "column": 12
+                }
+              },
+              "value": "1",
```
kaicataldo: This isn't the actual correct value in the AST (it's the `BigInt.prototype.toString()` representation since `BigInt`s can't be serialized to JSON), but it should be fine for testing purposes.

kaicataldo:
Hmm this is failing because `BigInt` doesn't appear to be defined in Flow. Should I just add a type for now?

nicolo-ribaudo:
It's better to add `// $FlowIgnore` before the line which is reporting an error so that it will tell us to remove it when we upgrade to a version that supports bigints.

kaicataldo:
Makes sense! Will do.

nicolo-ribaudo:
:heavy_check_mark:

packages/babel-parser/src/plugins/estree.js
```diff
@@ -31,6 +33,16 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return node;
     }
 
+    estreeParseBigIntLiteral(value: any): N.Node {
+      // https://github.com/estree/estree/blob/master/es2020.md#bigintliteral
+      // $FlowIgnore
+      const bigInt = typeof BigInt !== "undefined" ? BigInt(value) : null;
```
JLHwung: We should add a test case to show that the parsed result is different on node.js 6 and 8, which _does not_ have BigInt support.

https://github.com/estree/estree/blob/master/es2020.md#bigintliteral
> In environments that don't support BigInt values, value property will be null as the BigInt value can't be represented natively.

kaicataldo: Ah, are we not testing in older Node versions at the moment? I tested it locally with older Node versions and assumed this was already being done in CI?

kaicataldo: Oh, right - sorry, I was thinking of the `@babel/eslint-parser` tests. Yes, we should add a test for this in `@babel/parser`.

PR #10828, opened 2019-12-06
@babel/eslint-parser: fix ImportExpression node to match ESTree spec 

kaicataldo:
| Fixed Issues?            | Refs https://github.com/babel/babel/issues/10826
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  | 

`@babel/parser` currently parses dynamic imports (`import('a')`) as `CallExpression`s (code [here](https://github.com/babel/babel/blob/master/packages/babel-parser/src/parser/expression.js#L784-L802)). The ESTree spec differs from this and gives it its own type called [`ImportExpression`](https://github.com/estree/estree/blob/master/es2020.md#expressions).

### DIFF
--- a/eslint/babel-eslint-parser/test/babel-eslint-parser.js
+++ b/eslint/babel-eslint-parser/test/babel-eslint-parser.js
@@ -524,5 +524,11 @@ describe("babylon-to-espree", () => {
         const a = 1n;
       `);
     });
+
+    it("Dynamic Import", () => {
+      parseAndAssertSame(`
+        const a = import('a');
+      `);
+    });
   });
 });
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -759,7 +759,6 @@ export default class ExpressionParser extends LValParser {
     node: T,
     optional: boolean,
   ): T {
-    let nodeType = optional ? "OptionalCallExpression" : "CallExpression";
     if (node.callee.type === "Import") {
       if (node.arguments.length !== 1) {
         this.raise(node.start, "import() requires exactly one argument");
@@ -769,13 +768,11 @@ export default class ExpressionParser extends LValParser {
           this.raise(importArg.start, "... is not allowed in import()");
         }
       }
-
-      if (this.hasPlugin("estree")) {
-        nodeType = "ImportExpression";
-        node.source = node.arguments[0];
-      }
     }
-    return this.finishNode(node, nodeType);
+    return this.finishNode(
+      node,
+      optional ? "OptionalCallExpression" : "CallExpression",
+    );
   }
 
   parseCallExpressionArguments(
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -758,7 +758,7 @@ export default class ExpressionParser extends LValParser {
   finishCallExpression<T: N.CallExpression | N.OptionalCallExpression>(
     node: T,
     optional: boolean,
-  ): T {
+  ): N.Expression {
     if (node.callee.type === "Import") {
       if (node.arguments.length !== 1) {
         this.raise(node.start, "import() requires exactly one argument");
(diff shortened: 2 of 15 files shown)


packages/babel-parser/src/parser/expression.js
```diff
@@ -794,11 +795,12 @@ export default class ExpressionParser extends LValParser {
           this.raise(importArg.start, "... is not allowed in import()");
         }
       }
+
+      if (this.hasPlugin("estree")) {
```
kaicataldo: I'm not seeing any other instances of the `estree` plugin check happening in the parser itself. Is this okay?

packages/babel-parser/test/fixtures/experimental/class-properties/arguments-in-nested-class-decorator-call-expression/options.json
```diff
@@ -1,3 +1,6 @@
 {
-  "plugins": ["classProperties", ["decorators", { "decoratorsBeforeExport": false }]]
+  "plugins": [
```
kaicataldo: Result of running `npm run fix`.

packages/babel-parser/src/parser/expression.js
nicolo-ribaudo: I would prefer to keep the `estree` plugin self-contained. We can override `finishCallExpression` in the `estree` plugin class.

eslint/babel-eslint-parser/test/babel-eslint-parser.js
```diff
@@ -518,5 +518,11 @@ describe("babylon-to-espree", () => {
         }
       `);
     });
+
+    it("Dynamic Import", () => {
```
kaicataldo: Nice to be able to test this across the parser and the estree plugin :)

kaicataldo:
Wrestling with some Flow types right now (I haven't used Flow in years 😬), but will hopefully have that fixed up soon.

kaicataldo:
Still struggling with the flow types on this one (trying to avoid using `any`). If someone sees what I'm doing wrong, I'd love your input :)

packages/babel-parser/test/fixtures/estree/dynamic-import/basic/output.json
```diff
@@ -0,0 +1,150 @@
...
+                "start": {
+                  "line": 1,
+                  "column": 10
+                },
+                "end": {
+                  "line": 1,
+                  "column": 21
+                }
+              },
+              "callee": {
```
nicolo-ribaudo: Is this expected? The spec doesn't mention it.

nicolo-ribaudo:
The problem with flow is that a function returns `X`, you can't override it with a function which returns `X | Y` because it's not compatible: the caller wouldn't know how to handle `Y`. (I have fixed it)

packages/babel-parser/test/fixtures/estree/dynamic-import/basic/output.json
kaicataldo: Good find - I misread the spec and thought it was inheriting properties from `CallExpression` 😓. Will update.

packages/babel-parser/src/plugins/estree.js
```diff
@@ -411,4 +411,28 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+        delete ((node: N.Node): N.EstreeImportExpression).arguments;
+        delete ((node: N.Node): N.EstreeImportExpression).callee;
+      }
+
+      return node;
+    }
+
+    // ImportExpressions do not have an arguments array.
+    toReferencedListDeep(
+      exprList: $ReadOnlyArray<?N.Expression> = [],
```
kaicataldo: This was crashing because it's given the `CallExpression`'s `arguments` array, which is `undefined` for `ImportExpression`s (deleted above).

nicolo-ribaudo: What about doing this? We avoid using the empty array and iterating it:

```js
toReferencedListDeep(exprList: ?$ReadOnlyArray<?N.Expression>) {
  if (exprList) return super.toReferencedListDeep(exprList);
}
```

If flow complains that it can't return undefined, we can just change the original `toReferencedListDeep` signature to return `void`, since it's return value is never used.

kaicataldo: Sure, that sounds good to me. Updated!

PR #10831, opened 2019-12-06
Prevent ignored files in out dir

### LINKED ISSUE #6226: copy-files flag does not respect ignore block in .babelrc

### LINKED ISSUE #10829: '--ignore' conflicts with  '--copy-files'
## Bug Report

**Current Behavior**
cli option '--copy-files' copies all non-compiled files, including ignored files

**Input Code**
- Not behave as what i expect
```json
// package.json
{
  "scripts": {
     "build": "babel src --out-dir dist --copy-files --ignore src/**/*.test.js  --presets=@babel/env"
  }
}
```
run in terminal
```bash
$ npm run build
```

**Expected behavior/code**
'--copy-files' copies all non-compiled files, and skip ignored files

**Environment**

- Babel version(s): [v7.7.4]
- Node/npm version: [ Node v12.13]
- OS: [OSX 10.15.1,]
-  How you are using Babel: [`cli`]

rajasekarm:
| Fixed Issues?            | Fixes #6226
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

--ignore folder should not be present in out dir.


JLHwung:
C.f. https://github.com/babel/babel/issues/10829#issuecomment-562614238
> but what if i try to copy the ignored files after this issue fixed?

I am afraid people could rely on this behaviour to copy ignored files. Since `--copy-files` is meant to copy files that _will not_ be compiled and the ignored files meets the definition.

I propose to add an `includeIgnoredFiles` cli option so people can configure the behaviour.

rajasekarm:
That can be done, But I assume it should copy only the non supported format files and not the ignored one. Example css is non supported format, can be copied and required for the app to run. Test files are ignored and not copied and not required for the app to run.

JLHwung:
@rajasekarm Given the behaviour has been existed since v7, I would prefer we add this behind a flag. And later we can change the defaults in v8, so that practically it will skip the `ignore` files when `--copy-files` is enabled.

rajasekarm:
Sure, I'll make the change. Thanks for your input @JLHwung.

rajasekarm:
@JLHwung Added the flag, I named it as **includeIgnore**.

packages/babel-cli/src/babel/options.js
```diff
@@ -161,6 +161,11 @@ commander.option(
   "Delete the out directory before compilation.",
 );
 
+commander.option(
+  "--include-ignore",
+  "Include ignored files when compiling and copy to destination",
```
rajasekarm: Do we need to rephrase this help text?

packages/babel-cli/src/babel/options.js
```diff
@@ -161,6 +161,11 @@ commander.option(
   "Delete the out directory before compilation.",
 );
 
+commander.option(
+  "--include-ignore",
+  "Copy the ignored files to destination",
```
existentialism: To match to `--include-dotfiles`,

```suggestion
  "Include ignored files when compiling and copying non-compilable files.",
```

packages/babel-cli/src/babel/options.js
```diff
@@ -161,6 +161,11 @@ commander.option(
   "Delete the out directory before compilation.",
 );
 
+commander.option(
+  "--include-ignore",
+  "Include ignored files when compiling and copying non-compilable files.",
```
JLHwung: 🤔I don't think ignored file will be compiled when `--include-ignore` is true, it is different than `--include-dotfiles`.

Personally I prefer `--include-ignored` than `--include-ignore` because `ignore` is a verb.

existentialism: Yeah, we can drop the compiling bit.

rajasekarm: @JLHwung , I'll make the the change for the flag. Is the below text ok for you? Just want to confirm before pushing the change.

Include ignored files and copying non-compilable files.

JLHwung: I feel like it is

> Include ignored files when copying non-compilable files.

rajasekarm: @JLHwung Done ✅

PR #10549, opened 2019-10-14
Add ".js" extension to injected polyfill imports

shimataro:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #10548, fixes #8462 
| Patch: Bug Fix?          | Yes
| Major: Breaking Change?  | No
| Minor: New Feature?      | No
| Tests Added + Pass?      | No
| Documentation PR Link    | 
| Any Dependency Changes?  | No
| License                  | MIT

input:

```javascript
for (const x of [1, 2, 3]) {
  console.log(x);
}
```

output(before):

```javascript
import "core-js/modules/es.array.iterator";

for (const x of [1, 2, 3]) {
  console.log(x);
}
```

output(after):

```javascript
import "core-js/modules/es.array.iterator.js"; // <- extension added!

for (const x of [1, 2, 3]) {
  console.log(x);
}
```

As of Node.js v12, `import` syntax requires extension.
<https://medium.com/@nodejs/announcing-a-new-experimental-modules-1be8d2d6c2ff>

Please refer #10548 for more details.

PR #10833, opened 2019-12-06
Do not add `.js` extension to injected helper imports when module transform presents

### LINKED ISSUE #10832: Unresolvable runtime helper modules after upgrading to v7.7.5
## Bug Report

https://github.com/babel/babel/pull/10549 introduced a bit of breakage to consumers of `ember-cli-babel`.

When using `@babel/plugin-transform-modules-amd` to transpile the runtime helpers (from `@babel/runtime/helpers/esm`) into AMD modules, the resulting define statement's do **not** include the `.js` extension. However, when using `@babel/plugin-transform-runtime@7.7.5` (which includes #10549) along with `useESModules: true` the import statements will include `.js` which results in "module not found" errors.

JLHwung:
| Fixed Issues?            | Fixes #10832 

As a follow up to #10549, this PR partially reverts the recently introduced behaviour on the helper imports. It will remove `.js` added in #10549 when other module transforms presents.

The `.js` extension is only required by native ES Modules and we can safely remove it when it _will_ be transformed to other module formats.

**Compatibility Note**:

When using transform-runtime 7.7.6 with
- modules-umd < 7.7.6
- modules-amd < 7.5.0

it will still add the `.js` extension to imports due to lack of context variables. This will cause trouble to AMD users. If you are experience this issue, please upgrade these plugins to 7.7.6.

### DIFF
--- a/packages/babel-plugin-transform-runtime/src/index.js
+++ b/packages/babel-plugin-transform-runtime/src/index.js
@@ -254,7 +254,14 @@ export default declare((api, options, dirname) => {
         if (cached) {
           cached = t.cloneNode(cached);
         } else {
-          cached = addDefault(file.path, source, {
+          const moduleTransforms = file.get(
+            "@babel/plugin-transform-modules-*",
+          );
+          // Remove the trailing `.js` requires by ES Module when module transforms presents
+          const validSource = moduleTransforms
+            ? source.replace(/\.js$/, "")
+            : source;
+          cached = addDefault(file.path, validSource, {
             importedInterop: "uncompiled",
             nameHint,
             blockHoist,
--- a/packages/babel-plugin-transform-runtime/test/fixtures/runtime/modules-helpers/input.mjs
+++ b/packages/babel-plugin-transform-runtime/test/fixtures/runtime/modules-helpers/input.mjs
@@ -1,5 +1,6 @@
 import foo from "foo";
 
 class Example {
-  method() {}
+  property
 }
+
--- a/packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop-loose/imports-hoisting/output.js
+++ b/packages/babel-plugin-transform-modules-commonjs/test/fixtures/interop-loose/imports-hoisting/output.js
@@ -1,8 +1,8 @@
 "use strict";
 
-var _interopRequireDefault = require("@babel/runtime/helpers/interopRequireDefault.js");
+var _interopRequireDefault = require("@babel/runtime/helpers/interopRequireDefault");
 
-var _taggedTemplateLiteral2 = _interopRequireDefault(require("@babel/runtime/helpers/taggedTemplateLiteral.js"));
+var _taggedTemplateLiteral2 = _interopRequireDefault(require("@babel/runtime/helpers/taggedTemplateLiteral"));
 
 function _templateObject() {
   const data = (0, _taggedTemplateLiteral2.default)(["foo"]);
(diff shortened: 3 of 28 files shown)


packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -254,7 +254,14 @@ export default declare((api, options, dirname) => {
         if (cached) {
           cached = t.cloneNode(cached);
         } else {
-          cached = addDefault(file.path, source, {
+          const moduleTransforms = file.get(
+            "@babel/plugin-transform-modules-*",
+          );
+          // Remove the trailing `.js` requires by ES Module when module transforms presents
+          const validSource = moduleTransforms
```
JLHwung: Technically only `amd` and `umd` are affected by this issue since by default RequireJS does not support resolving `foo.js => foo.js` as-is out of the box. You may configure to support this but we did cause breaking changes to these users.

Both CommonJS and SystemJS support `foo.js => foo.js`, but at the same time they also support `foo => foo.js`. Hence I remove `.js` for the sake of briefness. C.f. https://github.com/babel/babel/pull/10549#pullrequestreview-306294014

packages/babel-plugin-transform-runtime/test/fixtures/runtime/modules-helpers/input.mjs
```diff
@@ -1,5 +1,6 @@
 import foo from "foo";
 
 class Example {
-  method() {}
+  property
```
JLHwung: I rewrite this test case since it looks more practical that an ESM supported environment does not support class-properties and so we need to transform.

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -254,7 +254,14 @@ export default declare((api, options, dirname) => {
         if (cached) {
           cached = t.cloneNode(cached);
         } else {
-          cached = addDefault(file.path, source, {
+          const moduleTransforms = file.get(
+            "@babel/plugin-transform-modules-*",
```
nicolo-ribaudo: IIRC, this was only supported since 7.4.0.
Let's try anyway if it is enough. Otherwise, we will have to revert the whole PR which introduced the `.js` extensions and only enable it behind an option.

LarsDenBakker:
This should not apply to systemjs transform, or be configurable. Systemjs is an es module compliant polyfills. We use this in https://www.npmjs.com/package/es-dev-server where we use es modules in the browser during development and fallback to systemjs if modules are not supported.

nicolo-ribaudo:
@babel/core-team What do you think about reverting the PR which introduced the regression, releasing it before Monday and then re-introducing the `.js` suffix behind an option?

> Systemjs is an es module compliant polyfill.

Note that the es modules specification doesn't say anything about file extension. The `.js` extension is never mentioned in the ECMAScript standard, and I think that it isn't mentioned by all the web platform stands neither (or at least, if it's meaningless).

chriskrycho:
> The `.js` extension is never mentioned in the ECMAScript standard, and I think that it isn't mentioned by all the web platform stands neither (or at least, if it's meaningless).

Indeed! More, the spec is *intentionally* silent on how a string is to be resolved. A given module loader (browser, Node, etc.) may choose to implement its own rules for the string parse. It is [required to be a string literal](https://www.ecma-international.org/ecma-262/10.0/index.html#prod-ModuleSpecifier), but that is the *only* thing specified about the module specifier.

JLHwung:
@LarsDenBakker This PR restores to behaviour for SystemJS to pre 7.7.5 so it should work.

I discussed with @nicolo-ribaudo a bit and we are now [reverting #10549](https://github.com/babel/babel/pull/10835) for 7.7.6. This PR is thus closed.

A new option `removeImportsExtension: boolean` will be introduced in 7.8.0. It defaults to `true` so the behaviour will be aligned to pre-7.7.5, which means we still generate `@babel/runtime/foo` imports. For those people publishing the es modules, they will have to set it `false`.

We are planing to change its default to `false` on 8.x, which is more meaningful to current ES Modules implementation while still providing backward supports to other module standards, i.e. amd.

LarsDenBakker:
Makes sense guys, thanks!

PR #10847, opened 2019-12-09
Add missing dev dependencies

JLHwung:

This PR is extracted from a messy working branch on the upcoming Yarn 2 support: https://github.com/JLHwung/babel/commits/standalone-rollup-bundle

The following devDependencies is required since Yarn 2 does _not_ resolve package in `../../node_modules`, which is the root `node_modules` in the project.

Note that `eslint-import-resolver-node` would not bail on these situations since the Node.js module resolution algorithm will resolve all the way up to `GLOBAL_FOLDERS`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "enhanced-resolve": "^3.0.0",
     "eslint": "^6.0.1",
     "eslint-config-babel": "^9.0.0",
+    "eslint-import-resolver-node": "^0.3.2",
     "eslint-plugin-flowtype": "^3.8.2",
     "eslint-plugin-import": "^2.17.2",
     "eslint-plugin-prettier": "^3.1.0",
--- a/package.json
+++ b/package.json
@@ -78,7 +78,8 @@
     "typescript": "^3.6.3"
   },
   "resolutions": {
-    "@lerna/**/@lerna/collect-updates": "https://github.com/babel/lerna.git#babel-collect-updates"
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -40,6 +40,7 @@
     "enhanced-resolve": "^3.0.0",
     "eslint": "^6.0.1",
     "eslint-config-babel": "^9.0.0",
+    "eslint-import-resolver-node": "^0.3.2",
```
nicolo-ribaudo: What is it needed for? `eslint-plugin-import` (which, from what I understood, is the package using it) already has it in `dependencies`.

nicolo-ribaudo:
@kaicataldo Do you have any clue about where the linting errors might come from?

JLHwung:
@nicolo-ribaudo We should map `eslint-scope` to 5.0.0. I will push later.

package.json
```diff
@@ -77,7 +78,9 @@
     "typescript": "^3.6.3"
   },
   "resolutions": {
-    "@lerna/**/@lerna/collect-updates": "https://github.com/babel/lerna.git#babel-collect-updates"
+    "@lerna/**/@lerna/collect-updates": "https://github.com/babel/lerna.git#babel-collect-updates",
+    "#todo: remove eslint-scope resolution when babel-eslint 11.0.0-beta.2 is published": "",
```
nicolo-ribaudo: AFAIK, comments aren't supported

JLHwung: It looks like comment but semantically it is a long sentence key.

PR #10848, opened 2019-12-09
Tune eslint packages test configuration

JLHwung:
| Fixed Issues?            | ESLint integration test should not load root babel configuration

This PR is extracted from a messy working branch on the upcoming Yarn 2 support: https://github.com/JLHwung/babel/commits/standalone-rollup-bundle

As a follow up to #10705 , the integration test should not read root babel configuration.

### DIFF
--- a/eslint/babel-eslint-parser/test/non-regression.js
+++ b/eslint/babel-eslint-parser/test/non-regression.js
@@ -18,25 +18,22 @@ function verifyAndAssertMessagesWithSpecificESLint(
       node: true,
       es6: true,
     },
+    ...overrideConfig,
     parserOptions: {
       sourceType,
       ecmaFeatures: {
         globalReturn: true,
       },
+      requireConfigFile: false,
       babelOptions: {
         configFile: path.resolve(
           __dirname,
           "./fixtures/config/babel.config.js",
         ),
       },
+      ...overrideConfig?.parserOptions
     },
   };
-
-  if (overrideConfig) {
-    for (const key in overrideConfig) {
-      config[key] = overrideConfig[key];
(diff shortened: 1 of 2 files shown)


JLHwung:
@nicolo-ribaudo The test error is expected, because now `ecmaFeatrues.globalReturn` is preserved. It aligns with ESLint behaviour: https://eslint.org/demo#eyJ0ZXh0IjoidmFyIGxlYWtlZEdsb2JhbCA9IDE7Iiwib3B0aW9ucyI6eyJwYXJzZXJPcHRpb25zIjp7ImVjbWFWZXJzaW9uIjo2LCJzb3VyY2VUeXBlIjoic2NyaXB0IiwiZWNtYUZlYXR1cmVzIjp7fX0sInJ1bGVzIjp7Im5vLWltcGxpY2l0LWdsb2JhbHMiOjJ9LCJlbnYiOnt9fX0=

I have added another test case when `ecmaFeatures.globalReturn` is `false`.

eslint/babel-eslint-parser/test/non-regression.js
```diff
@@ -18,26 +18,23 @@ function verifyAndAssertMessagesWithSpecificESLint(
       node: true,
       es6: true,
     },
+    ...overrideConfig,
     parserOptions: {
       sourceType,
       ecmaFeatures: {
         globalReturn: true,
```
nicolo-ribaudo: Can we remove this? It's better to set it explicitly in the only test where we need it rather than having a default value.

nicolo-ribaudo:
The Travis failures are caused by network errors

PR #10849, opened 2019-12-09
chore: specify package type [skip-ci]

JLHwung:

Per [Node.js ECMAScript Modules Docs](https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field), package should include "type" field and be explicit about it.

### DIFF
--- a/codemods/babel-plugin-codemod-object-assign-to-object-spread/package.json
+++ b/codemods/babel-plugin-codemod-object-assign-to-object-spread/package.json
@@ -4,6 +4,7 @@
   "description": "Transforms Object.assign into object spread syntax",
   "repository": "https://github.com/babel/babel/tree/master/codemods/babel-plugin-codemod-object-assign-to-object-spread",
   "license": "MIT",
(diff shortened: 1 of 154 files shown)


nicolo-ribaudo:
While this PR doesn't add any new feature, I would prefer to merge it in v7.8.0. Since it changes every package, it will cause every package of the monorepo to be published even if the code is the same.
I think that we should keep patch releases small.

JLHwung:
@nicolo-ribaudo I am also going to add `exports: false` to every package (other than `@babel/runtime`). I can prepend `exports: false` to `"licence": "MIT"` so that we can avoid merge conflicts, though.

nicolo-ribaudo:
Since this PR is quite simple, you push the `exports` changes here as a separate commit and then we'll use "Rebase and merge" instead of "Squash and merge" if you prefer.

nicolo-ribaudo:
I have added the `Ready to be Merged` label even if there is a single review, since this PR is quite straightforward.

PR #10850, opened 2019-12-10
Disable submodule exports

JLHwung:

This PR adds `exports: false` to every package except `@babel/runtime(-corejs[23])?`. By doing so we are stating explicitly that every package does not offer submodule exports, i.e. `import "@babel/core/src/config"` will bail because this file belongs to internal implementation details.

Note that even we don't specify `exports: false`, the following imports will still bail
```
import "@babel/core/src/config/index.js";
```
because the subpath `./src/config/index.js` is _not_ defined as export path in package.json. So it should not break anyone.

**Merge Tips**
This PR includes commits in #10849. Therefore it is intended to merge #10849 first and then use `Reabse to Merge` option to merge this one.

### DIFF
--- a/codemods/babel-plugin-codemod-object-assign-to-object-spread/package.json
+++ b/codemods/babel-plugin-codemod-object-assign-to-object-spread/package.json
@@ -4,6 +4,7 @@
   "description": "Transforms Object.assign into object spread syntax",
   "repository": "https://github.com/babel/babel/tree/master/codemods/babel-plugin-codemod-object-assign-to-object-spread",
   "license": "MIT",
+  "type": "commonjs",
   "publishConfig": {
     "access": "public"
   },
--- a/codemods/babel-plugin-codemod-object-assign-to-object-spread/package.json
+++ b/codemods/babel-plugin-codemod-object-assign-to-object-spread/package.json
@@ -5,6 +5,7 @@
   "repository": "https://github.com/babel/babel/tree/master/codemods/babel-plugin-codemod-object-assign-to-object-spread",
   "license": "MIT",
   "type": "commonjs",
+  "exports": false,
   "publishConfig": {
     "access": "public"
   },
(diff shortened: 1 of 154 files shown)


nicolo-ribaudo:
Does this also applies when using `require` instead of `import`?
If it does, then it will break some popular projects:
- [`vue-cli`](https://github.com/vuejs/vue-cli/blob/02f2436bdb44a435122e0de4b3924199b475983e/packages/%40vue/babel-preset-app/index.js#L19)
- [`ember-cli-babel`](https://github.com/babel/ember-cli-babel/blob/b750535ac24c5400df39744bab3f7546ae5cdbaa/index.js#L468)

We should probably consider exposing `@babel/preset-env/lib/targets-parser` as a separate helper package.

JLHwung:
> Does this also applies when using require instead of import?

Yes, because `exports` take over `main`, which means Node.js 13 will have different module resolution behaviour than previous versions if package exports presents. Therefore, it is a breaking change.

People may use `import "@babel/core/lib/whatever"` in the wild and 8.x is expected in weeks. As you said there _are_ some cases when the submodule exports is inevitable, so I propose to limit the scope of this PR:

1) add `exports: false` to every `@babel/plugin-*` and `@babel/helper-*`. They either features a single `lib/index.js` or not supposed to be directly used by end users. We can target this change to 8.x

2) for other packages, add necessary exports for compatibility with popular downstream tools. We can always create new package in 8.x if it is reasonable. Finally we can expect to remove unintended submodule exports on 9.x

JLHwung:
This PR has been revamped according to the latest best practise: https://nodejs.org/api/esm.html#esm_main_entry_point_export, to specify both `exports` and `main` as relative path starting with `.`

```json
{
  "main": "./lib/index.js",
  "exports": "./lib/index.js"
}
```

JLHwung:
The e2e test is failing because `vue-cli` invokes `require("@babel/core/package.json")` and `package.json` is not defined in `@babel/core`. The node communities are still discussing whether `package.json` should be allowed: https://github.com/nodejs/node/issues/33460

Before decisions are made by node TSC, I consider `package.json` as our package metadata and thus it falls into part of our public interfaces. I will add `./package.json` entries to every packages.

PR #10854, opened 2019-12-10
Restore `@babel/parser` tests

nicolo-ribaudo:

Since a test is considered as failed if `misMatch(expected, actual) !== undefined`, the missing returns effectively disabled parser tests.

### DIFF
--- a/packages/babel-parser/test/fixtures/core/uncategorised/369/output.json
+++ b/packages/babel-parser/test/fixtures/core/uncategorised/369/output.json
@@ -13,6 +13,7 @@
     }
   },
   "errors": [
+    "SyntaxError: Invalid parenthesized assignment pattern (1:1)",
     "SyntaxError: Invalid left-hand side in assignment expression (1:1)"
   ],
   "program": {
--- a/packages/babel-parser/test/fixtures/es2015/uncategorised/223/output.json
+++ b/packages/babel-parser/test/fixtures/es2015/uncategorised/223/output.json
@@ -1,7 +1,7 @@
 {
   "type": "File",
   "start": 0,
-  "end": 22,
+  "end": 20,
   "loc": {
     "start": {
       "line": 1,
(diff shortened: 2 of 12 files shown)


packages/babel-parser/test/fixtures/core/uncategorised/369/output.json
```diff
@@ -13,6 +13,7 @@
     }
   },
   "errors": [
+    "SyntaxError: Invalid parenthesized assignment pattern (1:1)",
     "SyntaxError: Invalid left-hand side in assignment expression (1:1)"
```
nicolo-ribaudo: I will deduplicate these errors in a separate PR.

packages/babel-parser/test/fixtures/es2015/uncategorised/223/output.json
```diff
@@ -1,32 +1,32 @@
 {
   "type": "File",
   "start": 0,
-  "end": 22,
+  "end": 20,
```
nicolo-ribaudo: This is because https://github.com/babel/babel/blame/master/packages/babel-parser/test/fixtures/es2015/uncategorised/223/input.js changed recently

packages/babel-parser/test/helpers/runFixtureTests.js
```diff
@@ -127,15 +129,21 @@ function overrideToJSON(cb) {
     const { toJSON } = obj.prototype;
     originalToJSONMap.set(obj, toJSON);
     obj.prototype.toJSON = function() {
+      if (typeof this === "bigint") {
+        return { [serialized]: "BigInt", value: serialize(this) };
```
nicolo-ribaudo: This serialization using a "type annotation" is needed because, in envs which don't support BigInts, we must know if `"1"` represented an integer (which should thus be compared as if it was `null`) or the string `"1"`.

I will expand this serialization to regular expressions, but in a separate PR to keep this one as small as possible so that it can be merged quickly (and thus we can rely on parser tests again).

PR #10858, opened 2019-12-11
Properly serialize non-json values in parser tests

nicolo-ribaudo:
| Fixed Issues?            | Follow up to #10854 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

1) I reused the same serialization logic used for BigInts also for RegExps
2) I removed the `.prototype.toJSON` override hack and used the mysterious second parameter of `JSON.stringify`

Please disable whitespace diff while reviewing this PR.

### DIFF
--- a/packages/babel-parser/test/helpers/runFixtureTests.js
+++ b/packages/babel-parser/test/helpers/runFixtureTests.js
@@ -111,39 +111,7 @@ export function runThrowTestsWithEstree(fixturesPath, parseFunction) {
 }
 
 function save(test, ast) {
-  overrideToJSON(() =>
-    fs.writeFileSync(test.expect.loc, JSON.stringify(ast, null, 2)),
-  );
-}
-
-// Ensure that RegExp, BigInt, and Errors are serialized as strings
-function overrideToJSON(cb) {
-  const originalToJSONMap = new Map();
-  const notJSONparseableObj = [RegExp, Error];
-
-  if (typeof BigInt !== "undefined") {
-    notJSONparseableObj.push(BigInt);
-  }
-
-  for (const obj of notJSONparseableObj) {
-    const { toJSON } = obj.prototype;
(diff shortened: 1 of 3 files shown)


packages/babel-parser/test/helpers/runFixtureTests.js
```diff
@@ -111,39 +111,7 @@ export function runThrowTestsWithEstree(fixturesPath, parseFunction) {
...
-  }
-
-  const result = cb();
-
-  for (const obj of notJSONparseableObj) {
-    obj.prototype.toJSON = originalToJSONMap.get(obj);
-  }
-
-  return result;
+  fs.writeFileSync(test.expect.loc, JSON.stringify(ast, serialize, 2));
```
JLHwung: nit:
consider explicitly specify the _ReplaceFunction_:
```js
JSON.stringify(ast, (key, value) => serialize(value), 2)
```

packages/babel-parser/test/helpers/runFixtureTests.js
```diff
@@ -221,22 +189,43 @@ function runTest(test, parseFunction) {
   }
 }
 
-function serialize(value) {
+function serialize(key, value) {
   if (typeof value === "bigint") {
```
JLHwung: Does it imply that the serialization result is dependent on node versions? Because `BigInt` is not supported in Node.js 6 and 8.

I am surprised that they do not bail.

nicolo-ribaudo: When the AST should contain a bigint, `null` is used instead. For this reason, I made the serialisation dependent on BigInt support.

This is also true before this PR: if you delete `packages/babel-parser/test/fixtures/estree/bigInt/basic/output.json` and run the tests on node 6, you'll get a different output.

Issue #6341, opened 2017-09-29
object-rest not behaving as expected with computed properties


lslv1243:
Choose one: is this a bug report or feature request?
bug

### Input Code

```js
const a = { 1: 0, 2: 1, 3: 2 }
const i = 1
const { [i]: val, ...rest } = a
```

### Babel Configuration (.babelrc, package.json, cli command)

```json
{
  "presets": ["babel-preset-expo"],
  "env": {
    "development": {
      "plugins": ["transform-react-jsx-source"]
    }
  }
}
```

### Expected Behavior

rest should be 
```javascript
{
  "2": 1,
  "3": 2,
}
```

### Current Behavior

rest is 
```javascript
{
  "1": 0,
  "2": 1,
  "3": 2,
}
```

### Possible Solution

indexOf compares with 3 equals and the object key is represented as string
https://github.com/babel/babel/blob/88b7983e4f1b4d472f37c057afce0e86da7feb0b/packages/babel-helpers/src/helpers.js#L469

### Context

running react-native application

### Your Environment

using react native with expo
```json
{
  "expo": {
    "sdkVersion": "21.0.0"
  }
}
```
dont know about configuration

Andarist:
Thanks for the report @lslv1243 . It is indeed a bug, let me look into fixing it.

Andarist:
This already got fixed by https://github.com/babel/babel/pull/5757 and the fix is available in v7. 

For some reason when having preset-es2015 + stage-3 active in the repl it creates buggy output. stage-3 alone is working as expected.

Andarist:
Ok, got down to it - in v7 there is still a conflict when using both:
- babel-plugin-transform-es2015-destructuring
- babel-plugin-transform-object-rest-spread

**Actual output**
```
var a = {
  1: 0,
  2: 1,
  3: 2
};
var i = 1;
var val = a[i],
    rest = babelHelpers.objectWithoutProperties(a, [i]);
```

**Expected output**
```
var a = {
  1: 0,
  2: 1,
  3: 2
};
var i = 1;
var val = a[i],
    rest = babelHelpers.objectWithoutProperties(a, [i].map(babelHelpers.toPropertyKey));
```

jridgewell:
There's all kinds of bugs with using destructuring and object-rest-spread together. It's on my list of todos.

cgood92:
Having the same problem.

```
const obj = {
  0: true,
  1: 'hi',
  2: true,
}

[...]

PR #10863, opened 2019-12-12
fix: add computed property support for object Ref

JLHwung:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/6341#issuecomment-564896831

In this PR we add support for computed properties when generating `objRef` for `object-rest-spread`. In the following example
```js
var prop = "a";
var obj = { a: { b: 42 } };
var {
  [prop]: { ...x }
} = obj;

// transformed on v7.7.5

var prop = "a";
var obj = {
  a: {
    b: 42
  }
};

var {} = obj,
    x = _extends({}, obj.prop);
```

Here the `objRef` is `obj.prop`, which is the parent path of the rest element `...x`. Apparently `obj.prop` is incorrect because `[prop]` is a _computed_ property. This is fixed by adding `computed` argument when building `t.memberExpression` for `objRef`.

However, the computed properties could be impure: which means we could not evaluate these properties twice. Therefore we refactored the interface of `replaceImpureComputedKeys` and reuse it to _cache_ the computed properties.

Note that although this PR only addresses https://github.com/babel/babel/issues/6341#issuecomment-564896831, #6341 can also be closed after merging this PR because I can not reproduce OP's issue on REPL. ([link](https://babeljs.io/repl#?browsers=&build=&builtIns=usage&spec=false&loose=false&code_lz=MYewdgzgLgBAhjAvDA3jAjALhgBgDQwBM26BAzNoTAL4BQoksAZkjABQCUSAfBveNFQwA2k04BdbADc4AGwIA6JQCcApoOqs4QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=false&targets=Node-12&version=7.7.5&externalPlugins=%40babel%2Fplugin-proposal-object-rest-spread%407.7.4))

### DIFF
--- a/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/object-ref-computed-exec/exec.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/object-ref-computed-exec/exec.js
@@ -1,0 +1,20 @@
+var key, x, y, z;
+// impure
+key = 1;
+var { [key++]: { y, ...x } } = { 1: { a: 1, y: 1 } };
+expect(x).toEqual({ a: 1 });
+expect(key).toBe(2);
+expect(y).toBe(1);
+
+// takes care of the order
+
+key = 1;
+var {
+  [++key]: { y, ...rest_y },
+  [++key]: { z, ...rest_z }
+} = {2: { y: 2, z: 3 }, 3: { y: 2, z: 3 } };
+expect(y).toBe(2);
+expect(rest_y).toEqual({z: 3});
+expect(z).toBe(3);
+expect(rest_z).toEqual({ y: 2 });
+
--- a/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/object-ref-computed-exec/exec.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/object-ref-computed-exec/exec.js
@@ -17,4 +17,4 @@ expect(y).toBe(2);
 expect(rest_y).toEqual({z: 3});
 expect(z).toBe(3);
 expect(rest_z).toEqual({ y: 2 });
-
+expect(key).toBe(3);
--- a/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/object-ref-computed/input.js
+++ b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/object-ref-computed/input.js
@@ -1,0 +1,20 @@
+var key, x, y, z;
+// impure
+key = 1;
+var { [key++]: { y, ...x } } = { 1: { a: 1, y: 1 } };
+expect(x).toEqual({ a: 1 });
+expect(key).toBe(2);
+expect(y).toBe(1);
+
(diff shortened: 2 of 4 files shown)


packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/object-ref-computed/input.js
```diff
@@ -0,0 +1,20 @@
+var key, x, y, z;
```
JLHwung: This test case is revised from https://github.com/babel/babel/blob/0b3f883ed19b4958449f3dbf0f6104b4eeda570b/packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/impure-computed/input.js#L1

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/object-ref-computed-exec/exec.js
```diff
@@ -0,0 +1,20 @@
...
+key = 1;
+var {
+  [++key]: { y, ...rest_y },
+  [++key]: { z, ...rest_z }
+} = {2: { y: 2, z: 3 }, 3: { y: 2, z: 3 } };
+expect(y).toBe(2);
+expect(rest_y).toEqual({z: 3});
+expect(z).toBe(3);
+expect(rest_z).toEqual({ y: 2 });
+
```
nicolo-ribaudo:
```suggestion
expect(key).toBe(3);

```

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/object-ref-computed/input.js
```diff
@@ -0,0 +1,20 @@
...
+key = 1;
+var {
+  [++key]: { y, ...rest_y },
+  [++key]: { z, ...rest_z }
+} = {2: { y: 2, z: 3 }, 3: { y: 2, z: 3 } };
+expect(y).toBe(2);
+expect(rest_y).toEqual({z: 3});
+expect(z).toBe(3);
+expect(rest_z).toEqual({ y: 2 });
+
```
nicolo-ribaudo:
```suggestion
expect(key).toBe(3);

```

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-rest/object-ref-computed/output.js
```diff
@@ -0,0 +1,61 @@
...
+    rest_z = babelHelpers.objectWithoutProperties(_$[_ref2], ["z"]);
+
+expect(y).toBe(2);
+expect(rest_y).toEqual({
+  z: 3
+});
+expect(z).toBe(3);
+expect(rest_z).toEqual({
+  y: 2
+});
```
nicolo-ribaudo:
```suggestion
});
expect(key).toBe(3);
```

nicolo-ribaudo:
The failing tests are fixed in master

PR #10865, opened 2019-12-13
Added support for record and tuple syntax.

rickbutton:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

TC39 Proposal Info:
Champion: @rricard @rickbutton
Spec Repo: https://github.com/tc39/proposal-record-tuple
Slides (Stage 1 at Oct 2019 TC39 Meeting): https://button.dev/talks/records-and-tuples-tc39-october-2019.pdf

This PR adds support for both the hash style `#{ a: 1 }` and the bar style `{| a: 1 |}` record and tuple syntax to `babel-parser` under the option `recordAndTuple.{hash/bar}`. I've also added a new plugin `babel-plugin-syntax-record-and-tuple` for enabling the parser option.

This does not include a plugin for transforming the syntax, we intend to iterate on a transform and polyfill separately for now, similar to the approach taken for BigInt.

This is my first `babel` PR, so please let me know if I'm missing something!


rickbutton:
@littledan the changes to the tokenzier check for the existence of the `bar` config options in order to tokenize the `{|` and `|}` tokens, the parser itself will consume those tokens regardless of the options provided (but more defense in depth would make sense here, I should probably check for the option at both levels).

I'm also not opposed to switching to `syntax: "bar"` instead of `bar: true` so that one or the other must be chosen.

packages/babel-parser/ast/spec.md
```diff
@@ -718,6 +720,26 @@ interface ObjectMethod <: ObjectMember, Function {
 }
 ```
 
+## RecordExpression
+
+```js
+interface RecordExpression <: Expression {
+  type: "RecordExpression";
+  syntaxType: string;
+  properties: [ ObjectProperty | ObjectMethod | SpreadElement ];
```
JLHwung: 🤔Is `ObjectMethod` allowed in `RecordExpression`?

rickbutton: Good question! My initial attempt disallowed `ObjectMethod` inside a `RecordExpression` but after the discussion on https://github.com/tc39/proposal-record-tuple/issues/88 I opted for the current behavior.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1046,11 +1060,27 @@ export default class ExpressionParser extends LValParser {
         this.state.inFSharpPipelineDirectBody = oldInFSharpPipelineDirectBody;
         return this.finishNode(node, "ArrayExpression");
       }
+      case tt.braceBarL:
+      case tt.braceHashL: {
```
JLHwung: should assert `recordAndTuple` presents otherwise it is a syntax error.
```js
this.expectPlugin("recordAndTuple");
```
Can you add a test case when `recordAndTuple` is not enabled?

packages/babel-types/src/definitions/experimental.js
```diff
@@ -269,3 +270,40 @@ defineType("BigIntLiteral", {
   },
   aliases: ["Expression", "Pureish", "Literal", "Immutable"],
 });
+
+defineType("RecordExpression", {
+  visitor: ["properties"],
+  aliases: ["Expression"],
+  fields: {
+    syntaxType: {
+      validate: assertValueType("string"),
```
JLHwung: Let's be more explicit
```js
assertOneOf("bar", "hash")
```

packages/babel-plugin-syntax-record-and-tuple/src/index.js
```diff
@@ -0,0 +1,19 @@
...
+  api.assertVersion(7);
+
+  return {
+    name: "syntax-record-and-tuple",
+
+    manipulateOptions(opts, parserOpts) {
+      parserOpts.plugins.push([
+        "recordAndTuple",
+        {
+          hash: options.hash,
```
JLHwung: I prefer to expose `syntaxType` to the user land, because
- When either syntax variants is accepted, we can make that as default. We have similar [design](https://babeljs.io/docs/en/next/babel-plugin-proposal-pipeline-operator.html#via-babelrc-recommended) on pipeline-operator. 
- We don't have to process option precedence issues like `hash: false, bar: false` and `hash: true, bar: true`.
- More open to competing proposals, if any

packages/babel-types/src/definitions/experimental.js
nicolo-ribaudo: I'm not sure that this information should be there. This seems like syntax-specific information, not related to the AST.
It's similar to how the presence or absence of a semicolon is not represented in the AST.

I would prefer it as an option in `@babel/generator`, which should throw when printing a `RecordExpression` node if that option is not specified.
`@babel/plugin-syntax-record-and-tuple` would set both the parser option and the generator option.

packages/babel-parser/src/tokenizer/types.js
```diff
@@ -93,9 +93,13 @@ export const types: { [name: string]: TokenType } = {
 
   // Punctuation token types.
   bracketL: new TokenType("[", { beforeExpr, startsExpr }),
+  bracketHashL: new TokenType("#[", { beforeExpr, startsExpr }),
+  bracketBarL: new TokenType("{|", { beforeExpr, startsExpr }),
```
nicolo-ribaudo:
```suggestion
  bracketBarL: new TokenType("[|", { beforeExpr, startsExpr }),
```

packages/babel-types/src/definitions/experimental.js
rickbutton: @nicolo-ribaudo do you mean an additional property on the node specifically for the generator to use to pick syntax to output? If so, why wouldn't you just use the existing `syntaxType` property?

JLHwung: The generator options is independent on AST. By doing so we can have the AST agnostic about the syntax types, which is semantically equivalent.

You can add a new format option here https://github.com/babel/babel/blob/c35ba3d3abd0b915849be13fc682934c54f41599/packages/babel-generator/src/index.js#L39

and reference this option in the generator via `this.format.recordTupleSyntaxType`.

In the `syntax-*` you can mutate the generator options via `options.generatorOpts.recordTupleSyntaxType`.

rickbutton:
I've pushed a set of commits that I think addresses all of your concerns so far:

- I've moved the parser options to use `syntaxType: "hash"` instead of `hash: true` cc @littledan @JLHwung
- added checks to the parser to check for the `recordAndTuple` plugin instead of relying only on the tokenizer to pick up the error (and added tests to check for missing plugin + invalid `syntaxType`
- Moved the `syntaxType` node property into an option to the generator (and added some tests for this)
- added a test for flow fixed type syntax with record/tuple bar syntax

Let me know if I missed anything! Thank you for the super detailed + quick reviews!

rickbutton:
Just pushed a rebase, looks like the standalone tests pass, but e2e and coverage are failing, but they look like unrelated failures. Any ideas? @nicolo-ribaudo  @JLHwung

JLHwung:
The CI failure is unrelated. Did you fetch the upstream? If you rebase on the latest upstream the CI error should be gone.

rickbutton:
@JLHwung yep, fetched upstream and rebased. parent commit of this PR is current HEAD of babel/babel#master `8fd532d`

JLHwung:
@rickbutton Can you rebase again? The CI error has been fixed in recent commits.

rickbutton:
@JLHwung rebased again, everything is green! (minus test262). Thank you!

packages/babel-generator/src/generators/types.js
```diff
@@ -106,6 +106,67 @@ export function ArrayExpression(node: Object) {
...
+  let endToken;
+  if (this.format.recordAndTupleSyntaxType === "bar") {
+    startToken = "{|";
+    endToken = "|}";
+  } else if (this.format.recordAndTupleSyntaxType === "hash") {
+    startToken = "#{";
+    endToken = "}";
+  } else {
+    throw new Error(
+      `${this.format.recordAndTupleSyntaxType} is not a valid recordAndTuple syntax type`,
```
nicolo-ribaudo: This error message should mention the option name, and also be easy to understand when the option is not provided. WDYT about this?

```js
      `The "recordAndTupleSyntaxType" generator option must be either "bar" or "hash" (${JSON.stringify(this.format.recordAndTupleSyntaxType)} received).`,
```

Also, please add a test for an invalid value of the option and for when the option is not set.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -439,12 +452,12 @@ export default class Tokenizer extends LocationParser {
...
     if (this.state.pos !== 0 || this.length < 2) return false;
 
+    let ch = this.input.charCodeAt(this.state.pos + 1);
+    if (ch !== charCodes.exclamationMark) return false;
+
     const start = this.state.pos;
     this.state.pos += 1;
 
-    let ch = this.input.charCodeAt(this.state.pos);
-    if (ch !== charCodes.exclamationMark) return false;
```
nicolo-ribaudo: What is this change for?

packages/babel-parser/src/types.js
```diff
@@ -389,11 +389,21 @@ export type ArrayExpression = NodeBase & {
...
+};
+
 export type ObjectExpression = NodeBase & {
   type: "ObjectExpression",
   properties: $ReadOnlyArray<ObjectProperty | ObjectMethod | SpreadElement>,
 };
 
+export type RecordExpression = NodeBase & {
+  type: "RecordExpression",
+  properties: $ReadOnlyArray<ObjectProperty | SpreadElement>,
```
nicolo-ribaudo:
```suggestion
  properties: $ReadOnlyArray<ObjectProperty | ObjectMethod | SpreadElement>,
```

packages/babel-parser/src/tokenizer/index.js
rickbutton: I moved this because it will incorrectly eat a `#` character if the next character isn't a `!`, I think that the `this.state.pos` should only be advanced if it is actually going to be an interpreter directive. (This happens if the first line of the file is `#{}`

packages/babel-generator/src/generators/types.js
rickbutton: @nicolo-ribaudo pushed a change to include this new message, and added the tests you mentioned.

The `babel-generator` test runner doesn't support the standard `throws` option in `options.json` (looks like it isn't using the same infra as everything else). I took the liberty of adding support for `throws` to the package's test suite, let me know if this should exist somewhere else.

packages/babel-parser/test/fixtures/experimental/record-and-tuple/invalid-type-bar-record/options.json
```diff
@@ -0,0 +1,4 @@
+{
+  "plugins": [["recordAndTuple", { "syntaxType": "invalid" }]],
+  "throws": "Unexpected token (1:1)"
```
nicolo-ribaudo: Here we should throw an error similar to `@babel/generator`'s for the invalid option.

rickbutton: Good call, how about this? (used the pipelineOperator option checks for a reference here).

```
'recordAndTuple` requires `syntaxType` option whose value should be one of: 'hash', 'bar'
```

nicolo-ribaudo: Yes :+1:

rickbutton: done :)

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -668,16 +698,34 @@ export default class Tokenizer extends LocationParser {
...
         return;
       case charCodes.rightSquareBracket:
         ++this.state.pos;
         this.finishToken(tt.bracketR);
         return;
       case charCodes.leftCurlyBrace:
-        ++this.state.pos;
-        this.finishToken(tt.braceL);
+        if (
+          this.getPluginOption("recordAndTuple", "syntaxType") === "bar" &&
```
nicolo-ribaudo: Instead of checking this in the `if` head, can you do this in the body?

```js
if (this.getPluginOption("recordAndTuple", "syntaxType") !== "bar") {
  this.unexpected(this.state.pos, "Record expressions starting with '{|' are only allowed when the 'syntaxType' option of the 'recordAndTuple' plugin is set to 'bar'");
}
```

It would give us nicer error messages. (Also in the other places in this file).

rickbutton: great idea! I pushed a commit that does this, and added test cases for each of the possible error messages.

nicolo-ribaudo:
The failing tests are not related to this PR.

nicolo-ribaudo:
@rickbutton I know that rebasing can be a pain, but... could you adapt this to https://github.com/babel/babel/pull/11192? :pray: :sweat_smile:

rickbutton:
absolutely!

rickbutton:
@nicolo-ribaudo rebased my changes against the latest master, and updated my changes to use the new message templates support.

There are two files that still have inline error messages:

`packages/babel-parser/src/plugin-utils.js` - looks like none of the errors in this file were moved over, should I worry about this?

`packages/babel-generator/src/generators/types.js` - record/tuple is the only thing that throws in this location, so there isn't any support for Error constants/`this.raise` here, what would you like me to do here?

rickbutton:
also, looks like the test failures are still unrelated

nicolo-ribaudo:
The other errors are ok :+1:

:tada:

rickbutton:
awesome! thank you @JLHwung and @nicolo-ribaudo for your reviews!

JLHwung:
@rickbutton Can you add plugin mappings to

https://github.com/babel/babel/blob/0616f06b00eb1d3611258553c06daef56efe79d0/packages/babel-core/src/parser/util/missing-plugin-helper.js#L3

So end users can have better error message.

rickbutton:
@JLHwung no problem at all! I've pushed a commit to add it.

nicolo-ribaudo:
:shipit: :rocket:

rricard:
Thanks nicolo!

benjamn:
@rickbutton @rricard I'd love to get involved with the runtime implementation of this syntax (whether or not that work is already underway)!

<details>
I have some ideas about how to make the internalization of records and tuples efficient (the trick that allows `===` wherever possible), and I think I know what that could look like in terms of `@babel/runtime` helpers. I'm looking forward to using this syntax to represent immutable structures like GraphQL queries and result trees (I work for Apollo GraphQL, formerly Meteor). I also sit on TC39 with you (see you next week?) and have been involved with Babel mostly because I wrote and maintain https://github.com/facebook/regenerator.
</details>

rricard:
Hey @benjamn, make sure to catch our update at the next meeting as we already have some kind of runtime implementation thanks to Rick! That being said, we would be happy to get your feedback on it especially on the performance side. If you also want to help with the proposal itself there is a lot to do as well on the spec side, if that’s the case please let me know! See you next week!

benjamn:
Ahh, now I see https://github.com/tc39/proposal-record-tuple/blob/rb/babel-experiment/babel-plugin-experiment. Is that the best place for issues/PRs? Do you ultimately want to move those plugins into the `@babel/*` namespace, or keep them independent?

rickbutton:
@benjamn this is the official repo for the polyfill https://github.com/bloomberg/record-tuple-polyfill

Feel free to open issues/PRs against that repo!

The polyfill and transform will not be moved into the `@babel` namespace for the forseeable future, as they are purely experimental and the semantics of the proposal are still in flight; we don't want anyone to get attached to a specific implementation of the polyfill.

rickbutton:
(also, on the implementation itself, I encourage you to look at the above link! It is the most up to date version, and does exactly that via interning frozen objects in a graph that is cleaned up via WeakRef and FinalizationRegistry)

PR #10881, opened 2019-12-17
WIP: Chore: add Makefile.js

kaicataldo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  | Added `shelljs`

This is a WIP. Still have to finish converting some scripts and then will have to figure out how to test them all.

I'm not sure how best to handle environment variables in a cross environment-friendly way. I have to look into if `shelljs.env()` environment vars are persisted in child processes spawned by `exec` (I'm guessing not?).


JLHwung:
> I have to look into if shelljs.env() environment vars are persisted in child processes spawned by exec (I'm guessing not?).

It should be persisted because `exec` [defaults](https://github.com/shelljs/shelljs/blob/aed6d994b57bccea313e85eeaa9650077b12f4de/src/exec.js#L42) `env` option to `process.env`, which can be [exported](https://github.com/shelljs/shelljs/blob/aa9d443e7b8126763cfb76b1230a7bdb578a6a5e/test/env.js#L15) from `shell.env`.

.circleci/config.yml
```diff
@@ -57,9 +57,9 @@ jobs:
       - restore_cache: *restore-yarn-cache
       - restore_cache: *restore-node-modules-cache
       - run: yarn --version
-      - run: make test-ci-coverage
+      - run: node Makefile.js testCiCoverage
       # Builds babel-standalone with the regular Babel config
-      - run: IS_PUBLISH=true make build
+      - run: IS_PUBLISH=true node Makefile.js build
```
nicolo-ribaudo: Please use `yarn node` instead of `node`, so that we won't have to update it when migrating to yarn 2.

kaicataldo:
Seems like this might not be the best way to go about this due to the limitations discussed [here](https://github.com/babel/babel/issues/10889#issue-539830191). Will explore some other options.

PR #10882, opened 2019-12-17
Ignore some files in npm package

JLHwung:
| Fixed Issues?            | Some files were unexpectedly published to npm

Ignore unused file in npm package. Here is an unpacked size comparison

| **package** | **before** | **after** |
| --- | --- | --- |
| `generator` | 117 kB | 116.5 kB |
| `plugin-proposal-pipeline-operator` | 29.4 kB | 8.3 kB |
| `plugin-syntax-pipeline-operator` | 3.74 kB | 3.0 kB |
| `plugin-transform-dotall-regex` | 4.2 kB | 2.9 kB |
| `preset-typescript` | 6.34 kB | 3.6 kB |
| `standalone` | 3.69 MB | 3.69 MB |
| `env-standalone` | 4.33 MB | 4.33 MB |


packages/babel-standalone/package.json
```diff
@@ -5,8 +5,7 @@
   "main": "babel.js",
   "files": [
     "babel.js",
-    "babel.min.js",
-    "src"
```
JLHwung: I remove the `src` here because
1) It is untranspiled since day 1 of `@babel/standalone` and nobody ever complains about that
2) `src` is not intended to be used by end users

packages/babel-plugin-syntax-pipeline-operator/.npmignore
```diff
@@ -0,0 +1,2 @@
+*.log
+src
```
nicolo-ribaudo: Let's also put `test` here, since those files are often copy-pasted when creating a new package.

Issue #10356, opened 2019-08-20
Newline having a trailing whitespace character is removed in JSX attribute value


beliayeu:
## Bug Report

**Current Behavior**
new line (\n character) is removed from jsx attribute value when there is a whitespace character (\s+) right after new line.
new line (\n character) is NOT removed from jsx attribute value when there are no whitespace characters (\s+) right after new line.

**Input Code**
- [REPL](https://babeljs.io/en/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=DwEwlgbgfAUABHUk5gHaoKYCcASAVAWQBkBeAIhggEMs4AzAewbhLgEYBuOAem4BcMAZz4pUAGzQY4AYwYBbORlR9KNOACM1rAExcE-g4bgwyPWAiQRRmXIVIVqtRs1YBmLrwHDREzDPmKyvBGIQaqtJq0rAAseqHGptywwNzg0EA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=false&targets=&version=7.5.5&externalPlugins=%40babel%2Fplugin-transform-react-jsx%407.3.0)

**Expected behavior/code**
Newline should be processed more consistently. Either it should be removed or not removed in both cases mentioned in the current behavior.

**Impact**
react helmet is used for injecting scripts to the document head. react helmet allows to add inline scripts using innerHTML attribute. once we add inline scripts containing inline comments then we end up with broken script produced by react render

example
```
<Helmet>
    <script innerHTML="
var firstVariable = 1; //test

var lostVariable = 2;
" />
</Helmet>
```
will produce
```
<script data-react-helmet="true" >
var firstVariable = 1; //test var lostVariable = 2;
</script>
```

**Possible Solution**
Option 1. do not remove new line, remove only trailing whitespace characters. 

change
```
value.value = value.value.replace(/\n\s+/g, " ");
```
into
```
value.value = value.value.replace(/\n\s+/g, "\n");
```
in [babel-helper-builder-react-jsx](https://github.com/babel/babel/blob/master/packages/babel-helper-builder-react-jsx/src/index.js#L91)

**Concerns**
were there any reasons why the new line was replaced with whitespace?

JLHwung:
[...]

PR #10884, opened 2019-12-17
@babel/helper-builder-react-jsx: Make EOL handling plateform independant

### LINKED ISSUE #10426: Builder-react-jsx: Inconsistent string literal value normalization across different platforms
## Bug Report

**Current Behavior**
Babel-helper-builder-react-jsx modifies string literal attribute values. Though it itself is already not nice, but even more, it does it in inconsistent way.

**Input Code**

`index.js`
```js
import React from 'react'

const style = {
  height: '100%',
}

const svgStyle = {
  fill: 'currentColor',
}

const toTitle = (error, componentStack) => `${error.toString()}\n\nThis is located at:${componentStack}`

const ErrorBoundaryFallbackComponent = ({ componentStack, error }) => (
  <div style={style} title={toTitle(error, componentStack)}>
    <svg style={svgStyle} viewBox="0 0 24 24" preserveAspectRatio="xMidYMid">
      <path
        d="M20,12A8,8 0 0,0 12,4A8,8 0 0,0 4,12A8,8 0 0,0 12,20A8,8 0 0,0 20,
        12M22,12A10,10 0 0,1 12,22A10,10 0 0,1 2,12A10,10 0 0,1 12,2A10,10 0 0,1 22,
        12M15.5,8C16.3,8 17,8.7 17,9.5C17,10.3 16.3,11 15.5,11C14.7,11 14,10.3 14,
        9.5C14,8.7 14.7,8 15.5,8M10,9.5C10,10.3 9.3,11 8.5,11C7.7,11 7,10.3 7,9.5C7,
        8.7 7.7,8 8.5,8C9.3,8 10,8.7 10,9.5M12,14C13.75,14 15.29,14.72 16.19,
        15.81L14.77,17.23C14.32,16.5 13.25,16 12,16C10.75,16 9.68,16.5 9.23,
        17.23L7.81,15.81C8.71,14.72 10.25,14 12,14Z"
      />
    </svg>
  </div>
)

export default ErrorBoundaryFallbackComponent
```
`index.test.js`
```js
import React from 'react'
import { shallow } from "enzyme"

import { configure } from "enzyme"
import Adapter from "enzyme-adapter-react-16"

configure({ adapter: new Adapter() })

import {default as Fallback} from './'

describe('<Fallback />', () => {
  test('renders error correctly', () => {
    const props = { componentStack: 'dummy stack that goes on forever', error: 'crash, burn, die' }
    const tree = shallow(React.createElement(Fallback, props))

    expect(tree.render()).toMatchSnapshot()
  })
})
```
Due to [this line](https://github.com/babel/babel/blob/daf25af47c4161025dfa552fd41b837f062efe47/packages/babel-helper-builder-react-jsx/src/index.js#L91) babel loader produces different generated code on Windows and *nix machines because of different line endings (`CRLF` vs `LF`).
And even though it is not a big deal itself and does not impact code functionality, it causes different snapshots generated by Jest on different OSs.

**Expected behavior/code**
It should either not modify attribute value or do it in platform-independent way.

```js
module.exports = {
	presets: [
		['@babel/preset-env', {
			targets: {
				node: 'current',
			}
		}],
		"@babel/preset-react"
	]
}
```
```
"dependencies": {
    "react": "16.5.2",
    "react-dom": "16.5.2",
    "react-hot-loader": "4.3.12",
    "react-radio-group": "3.0.3"
  },
  "devDependencies": {
    "@babel/core": "7.1.2",
    "@babel/node": "7.0.0",
    "@babel/preset-env": "7.1.0",
    "@babel/preset-react": "7.0.0",
    "babel-core": "7.0.0-bridge.0",
    "babel-engine-plugin": "0.2.1",
    "babel-eslint": "8.2.6",
    "babel-jest": "^23.6.0",
    "babel-loader": "8.0.4",
    "babel-plugin-transform-ensure-ignore": "0.1.0",
    "babel-plugin-transform-es2015-modules-commonjs": "^6.23.0",
    "babel-polyfill": "^6.26.0",
    "babel-preset-env": "1.7.0",
    "babel-register": "^6.26.0",
    "enzyme": "^3.4.1",
    "enzyme-adapter-react-16": "1.6.0",
    "enzyme-to-json": "3.3.4",
    "jest": "23.6.0",
    "jest-fetch-mock": "1.6.6",
    "react-test-renderer": "^16.5.0",
    "snapshot-diff": "0.4.0"
  },
  "jest": {
    "verbose": true,
    "transform": {
      "^.+\\.js$": "babel-jest"
    },
    "snapshotSerializers": [
      "enzyme-to-json/serializer"
    ]
  },
```
`npm run jest`

**Environment**
- Babel version: 7.1.2
- Node 8.12.0/npm 6.4.1
- Windows 7, Windows 10, Debian 9
- Monorepo: no

**Possible Solution**
Do not replace `\n` in attribute value or replace additionally `\r\n`

screen1182
screen1183
screen1184
These are screenshots of debug run on Windows using file with `LF` endings. In case of `CRLF` value would remain line breaks.

ghostd:
| Fixed Issues?            | Fixes #10426
| Documentation PR Link    | 

Join all the lines of JSXText whatever the plateform is (use the same regex than [jsx-transform](https://github.com/alexmingoia/jsx-transform/blob/master/lib/visitor.js#L276))


JLHwung:
Related: #10356

Note that this behaviour has been existed since day 1 of the jsx transform, and people may rely on the behaviour in the wild. We should discuss about it on v8 release.

nicolo-ribaudo:
cc @babel/react

mathiasbynens:
> PS. @mathiasbynens Will something like `\p{JavaScriptNewline}` ever work? 😛

I know you were half-joking, but I'll give a serious answer anyway :)

For `\p{JavaScriptLineTerminator}`, we could make it work today, either by convincing the Unicode consortium to standardize this property. (It's probably not the right thing to do though since the definition of a JS `LineTerminator` belongs in the JS spec.)

For `\p{JavaScriptLineTerminatorSequence}` which seems to be what you really want here, we need [sequence properties](https://github.com/tc39/proposal-regexp-unicode-sequence-properties) in the language first, since as you pointed out it could expand to a sequence of multiple code points (CRL+F).

nicolo-ribaudo:
@babel/react Do you have any opinion about this?

lunaruan:
This seems okay to me, though would this be a breaking change? 
Also, we should update [`@babel/babel-helper-builder-react-jsx-experimental`](https://github.com/babel/babel/blob/master/packages/babel-helper-builder-react-jsx-experimental/src/index.js#L511) with this change as well

nicolo-ribaudo:
Yeah, if we do this we would definitely wait for Babel 8 :+1:.

jridgewell (commented):
I think we're doing this at the wrong level here, this should be performed by the Parser not the transformer.

nicolo-ribaudo:
Currently all the other existing JSX parser (or at least, every JS parser on ASTExplorer supporting JSX) keep the newlines in the AST, similarly to multiline template literals.

jridgewell:
> Currently all the other existing JSX parser (or at least, every JS parser on ASTExplorer supporting JSX) keep the newlines in the AST, similarly to multiline template literals.

Sorry, I'm talking about normalization to `\n` here, so that transformers only ever need to think about `\n`. We do a similar parsing-normalization for literal source-text `\r\n` -> `\n` in strings.

PR #10623, opened 2019-10-31
Fix: inputSourceMap should work when it is an external file

JLHwung:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #10534 
| Patch: Bug Fix?          | Yes
| Tests Added + Pass?      | Yes
| License                  | MIT

This PR fixes invalid `fromMapFileComment` call signature via prepending parsed comment text by `//`, which is [required in `convert-source-map`](https://github.com/thlorenz/convert-source-map/blob/ad25a3e4373c641a2241cf1470b81ae123e4c23d/index.js#L15).

This PR is a follow up to #9960.

PR #10885, opened 2019-12-17
fix: avoid string copy when processing input source-map

### LINKED ISSUE #10875: Very high memory usage compiling mozilla/PDF.js
## Bug Report

**Current Behavior**
When running npm start or npm run build, while pdf.js is used in create-react-app project, memory usage seems to skyrocket to unreasonable amounts and fail with "JavaScript heap out of memory" error.
unreasonable memory usage
NPM Error
I've reported it to PDF.js guys, and they found that ejecting and downgrading babel from 7.7.0+ to 7.6.4 seems to fix the issue so they think it must be regression issue with Babel. The issue can be reproduced in any fresh create-react-app (version 3.3.0) project with pdfjs-dist installed. ([link to original issue](https://github.com/mozilla/pdf.js/issues/11410))

***Steps to reproduce the problem***
1. Create new react app with create-react-app
2. npm install pdfjs-dist
3. Copy [example](https://github.com/mozilla/pdf.js/tree/master/examples/create-react-app) 
4. npm start or npm run build

**Input Code**
[Example repo](https://github.com/astrada/pdf.js-cra-test)

**Expected behavior/code**
npm start or npm run build succesfully without unreasonable memory usage

No additional configuration except of what CRA gives us

**Environment**

```
System:
    OS: Windows 10 10.0.18362
  Binaries:
    Node: 12.13.1 - C:\Program Files\nodejs\node.EXE
```
- Babel version(s): v7.7.4
- Node/npm version: Node 12/npm 6
- OS: Windows 10 Enterprise 1903

JLHwung:
| Fixed Issues?            | Fixes #10875
| Tests Added + Pass?      | No test is added, it is covered by current test suite

This PR fixes a performance regression introduced in #10623. 

https://github.com/babel/babel/blob/c35ba3d3abd0b915849be13fc682934c54f41599/packages/babel-core/src/transformation/normalize-file.js#L68-L69

The `lastComment` was copied into a new string, prefixed by `//` and later unwrapped by convert-source-map. 
https://github.com/thlorenz/convert-source-map/blob/9e9a7a9b652c30878c9c9aa591d861a9fdf61a7e/index.js#L31

This approach introduced significant memory footprint when there are many files with input source-map, i.e. the [test repo](https://github.com/astrada/pdf.js-cra-test) mentioned in #10875. Here we use `convertSourceMap.fromJSON` to avoid unnecessary string copy.

@artisb45 Thank you for the reproduction repo, it helps me a lot! You can test flight this fix by downloading and overwrite `node_modules/@babel/core/lib/transformation/normalize-file.js` with [this file](https://gist.githubusercontent.com/JLHwung/a4e735de002261d84764b8c3ed136c24/raw/48239c4844129b04d7aed1ac1985d5ee5ea72e29/normalize-file.js). We are expected to land it in a patch release if review is good.

### DIFF
--- a/packages/babel-core/src/transformation/normalize-file.js
+++ b/packages/babel-core/src/transformation/normalize-file.js
@@ -1,5 +1,6 @@
 // @flow
 
+import fs from "fs";
 import path from "path";
 import buildDebug from "debug";
 import cloneDeep from "lodash/cloneDeep";
@@ -64,10 +65,12 @@ export default function normalizeFile(
       const lastComment = extractComments(EXTERNAL_SOURCEMAP_REGEX, ast);
       if (typeof options.filename === "string" && lastComment) {
         try {
-          inputMap = convertSourceMap.fromMapFileComment(
-            // fromMapFileComment requires the whole comment block
-            `//${lastComment}`,
-            path.dirname(options.filename),
+          const inputMapFilename = EXTERNAL_SOURCEMAP_REGEX.exec(lastComment)
+            .groups.url;
+          inputMap = convertSourceMap.fromJSON(
+            fs.readFileSync(
+              path.resolve(path.dirname(options.filename), inputMapFilename),
+            ),
           );
         } catch (err) {
           debug("discarding unknown file input sourcemap", err);
@@ -157,7 +160,7 @@ function parser(
 
 // eslint-disable-next-line max-len
 const INLINE_SOURCEMAP_REGEX = /^[@#]\s+sourceMappingURL=data:(?:application|text)\/json;(?:charset[:=]\S+?;)?base64,(?:.*)$/;
-const EXTERNAL_SOURCEMAP_REGEX = /^[@#][ \t]+sourceMappingURL=(?:[^\s'"`]+?)[ \t]*$/;
+const EXTERNAL_SOURCEMAP_REGEX = /^[@#][ \t]+sourceMappingURL=(?<url>[^\s'"`]+?)[ \t]*$/;
 
 function extractCommentsFromList(regex, comments, lastComment) {
   if (comments) {
--- a/packages/babel-core/src/transformation/normalize-file.js
+++ b/packages/babel-core/src/transformation/normalize-file.js
@@ -65,8 +65,9 @@ export default function normalizeFile(
       const lastComment = extractComments(EXTERNAL_SOURCEMAP_REGEX, ast);
       if (typeof options.filename === "string" && lastComment) {
         try {
-          const inputMapFilename = EXTERNAL_SOURCEMAP_REGEX.exec(lastComment)
-            .groups.url;
+          const inputMapFilename = EXTERNAL_SOURCEMAP_REGEX.exec(
+            lastComment,
+          )[1];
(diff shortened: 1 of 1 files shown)


packages/babel-core/src/transformation/normalize-file.js
```diff
@@ -157,7 +160,7 @@ function parser(
 
 // eslint-disable-next-line max-len
 const INLINE_SOURCEMAP_REGEX = /^[@#]\s+sourceMappingURL=data:(?:application|text)\/json;(?:charset[:=]\S+?;)?base64,(?:.*)$/;
-const EXTERNAL_SOURCEMAP_REGEX = /^[@#][ \t]+sourceMappingURL=(?:[^\s'"`]+?)[ \t]*$/;
+const EXTERNAL_SOURCEMAP_REGEX = /^[@#][ \t]+sourceMappingURL=(?<url>[^\s'"`]+?)[ \t]*$/;
```
JLHwung: Dogfooding `named-capturing-regex` here.

The file size is bloated by helpers (6678 B -> 11276 B) but I think it is okay since `named-capturing-regex` is supported on Node.js 10+, which means we can get rid of the helpers in v8.

packages/babel-core/src/transformation/normalize-file.js
```diff
@@ -157,7 +161,7 @@ function parser(
 
 // eslint-disable-next-line max-len
 const INLINE_SOURCEMAP_REGEX = /^[@#]\s+sourceMappingURL=data:(?:application|text)\/json;(?:charset[:=]\S+?;)?base64,(?:.*)$/;
-const EXTERNAL_SOURCEMAP_REGEX = /^[@#][ \t]+sourceMappingURL=(?:[^\s'"`]+?)[ \t]*$/;
+const EXTERNAL_SOURCEMAP_REGEX = /^[@#][ \t]+sourceMappingURL=([^\s'"`]+?)[ \t]*$/;
```
jridgewell:
```suggestion
const EXTERNAL_SOURCEMAP_REGEX = /^[@#][ \t]+sourceMappingURL=([^\s'"`]+)[ \t]*$/;
```

The capturing group is mutually exclusive with the trailing quantifier group, so there's no need to be lazy about it. This should speed up the capture for very large string.

artisb45:
Hello, thanks for such a quick response to this issue @JLHwung ! Howerver I tried to replace the file you provided and it did not seem to do anything. Not sure if I'm doing it wrong, but I'm just supposed to replace the file at the path right? I'm using the same repo I linked in the issue. Also tried rolling back before ejection and then replacing - same result.

nicolo-ribaudo:
> https://github.com/babel/babel/pull/10885#issuecomment-566879115

Ops, I didn't see your comment :sweat_smile:

JLHwung:
@artisb45 Oops, you may need clear the babel-loader cache first because it is an ad-hoc approach.
```sh
# Clear babel-loader cache
rm -rf node_modules/.cache/babel-loader
cp normalize-file.js node_modules/@babel/core/lib/transformation/normalize-file.js
npm run build
```
It went well on my local environment. You can also wait for the upcoming 7.7.7 release.

PR #10886, opened 2019-12-18
Revert "Prevent ignored files in out dir"

rajasekarm:
Reverts babel/babel#10831

This will be added in Next minor release.


nicolo-ribaudo:
Reason: this is a new feature, it needs to wait for 7.8.0 😛

PR #10887, opened 2019-12-18
--copy-ignored flag added to CLI 

### LINKED ISSUE #6226: copy-files flag does not respect ignore block in .babelrc

rajasekarm:
| Fixed Issues?            | Fixes #6226 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |


nicolo-ribaudo:
Reviews at https://github.com/babel/babel/pull/10831

packages/babel-cli/src/babel/dir.js
```diff
@@ -84,8 +84,16 @@ export default async function({
...
-    if (!written && cliOptions.copyFiles) {
+    const relative = path.relative(base, src);
+    const isCompilableExtension = util.isCompilableExtension(
+      relative,
+      cliOptions.extensions,
+    );
+    if (
+      !written &&
+      ((!isCompilableExtension && cliOptions.copyFiles) ||
+        cliOptions.includeIgnored)
```
JLHwung: I think when `copyFiles` is not enabled but `includeIgnored` is enabled, we should throw in `parseArgv`, otherwise we will have another flag that _could_ copy non compilable files.

packages/babel-cli/test/fixtures/babel/--copy-files with ignore with includeIgnored/options.json
```diff
@@ -0,0 +1,12 @@
+{
+  "args": [
```
JLHwung: Can you add a test for `--only`?
```
babel src --out-dir lib --copy-files --only src/main/* --include-ignored
```

packages/babel-cli/src/babel/dir.js
existentialism: 🚲🏠 warning:

Should we call it `--copy-ignored`? 😝

rajasekarm: Changed to --copy-ignored

rajasekarm:
I'm throwing error if `--copy-ignored` is used without `ignore` flag. Which will solve other corner cases.

packages/babel-cli/test/fixtures/babel/--copy-files with ignore with includeIgnored/options.json
rajasekarm: Is this still needed?

packages/babel-cli/test/fixtures/babel/--copy-files --include-dotfiles with ignore/out-files/lib/foo/.foo.js
```diff
@@ -1 +0,0 @@
-a;
```
JLHwung: 🤔Why is `.foo.js` not copied?

rajasekarm: we've --include-dotfile to Include dotfiles when compiling and copying non-compilable files.

https://github.com/rajasekarm/babel/blob/master/packages/babel-cli/src/babel/options.js#L148

JLHwung: Yeah, it _was_ copied before this PR. But now .foo.js has been removed from the out files, which means it is _not_ copied.

rajasekarm: I got it, I'll check.

rajasekarm: I've that file in my branch, https://github.com/rajasekarm/babel/blob/master/packages/babel-cli/test/fixtures/babel/--copy-files%20--include-dotfiles%20with%20ignore/out-files/lib/foo/.foo.js

I dono why github shows this as removed.

nicolo-ribaudo: The linked one is your master branch, not the PR branch :stuck_out_tongue:

packages/babel-cli/test/fixtures/babel/--copy-files with ignore with includeIgnored/options.json
JLHwung: Good to have.😀

packages/babel-cli/test/fixtures/babel/--copy-files --include-dotfiles with ignore/out-files/lib/foo/.foo.js
rajasekarm: Yeah, I'm check this.

rajasekarm: Found the root cause of this bug, I'm fixing it.

rajasekarm: .foo.js is inside src/foo and it is in ignored list. So it is not copied.

packages/babel-cli/test/fixtures/babel/--copy-files with ignore with includeIgnored/options.json
rajasekarm: Added a test case for this.

rajasekarm:
I've exposed pathToPattern function from @babel/core for the ignore file check.

nicolo-ribaudo:
> I've exposed pathToPattern function from @babel/core for the ignore file check.

I have a bed news for you :stuck_out_tongue: 
This is a breaking change because if someone used `@babel/core` v7.7.7 and updates `@babel/cli` to 7.8.0 won't work.
For now, we can duplicate the code (with comments to explain that it's duplicated and if it is updated it should be done in both places) and then revisit it for Babel 8.

rajasekarm:
Sure, I'll make the change.

rajasekarm:
@nicolo-ribaudo Done 👍

nicolo-ribaudo:
I was looking at the `write` function in `packages/babel-cli/src/babel/dir.js`. From what I understood, it can return `false` in three cases:
1. The extension is not compatible (first `return false`)
2. The file has been ignored, and thus `babel.transformSync` (aka `util.compile`) returns `null`
3. There was an error.

If we make the `write` function return a string enum rather than a boolean, maybe we could avoid checking if the file has been ignored? Also, I think that the current approach doesn't work well if a file has been ignored not using a `cli` option, but using a config file.

rajasekarm:
@nicolo-ribaudo I made the requested change. Can you please check.

packages/babel-cli/src/babel/dir.js
```diff
@@ -9,6 +9,13 @@ import fs from "fs";
 import * as util from "./util";
 import { type CmdOptions } from "./options";
 
+const FILE_TYPE = Object.freeze({
+  NON_COMPACTIBLE: "NON_COMPACTIBLE",
```
JLHwung: It should be `NON_COMPILABLE`.

packages/babel-cli/src/babel/dir.js
```diff
@@ -20,11 +27,11 @@ export default async function({
 }: CmdOptions): Promise<void> {
   const filenames = cliOptions.filenames;
 
-  async function write(src: string, base: string): Promise<boolean> {
+  async function write(src: string, base: string): Promise<string> {
```
JLHwung:
```suggestion
  async function write(src: string, base: string): Promise<$Keys<typeof FILE_TYPE>> {
```

packages/babel-cli/src/babel/dir.js
rajasekarm: Done 👍

packages/babel-cli/src/babel/dir.js
rajasekarm: Done 👍

JLHwung:
@rajasekarm Can you rebase on upstream master? 🤔It should not have legacy CircleCI test check here.

nicolo-ribaudo:
@rajasekarm Could you open a PR to babel/website for the docs?

rajasekarm:
@nicolo-ribaudo I'll do it tonight.

PR #10890, opened 2019-12-18
fix: skip merging large input sourcemaps

### LINKED ISSUE #10875: Very high memory usage compiling mozilla/PDF.js

JLHwung:
| Fixed Issues?            | Fixes #10875 

This PR addresses #10875 where a large input source map is consumed and causing OOM error. It is not an ideal fix but still preserve basic support for practical size input source map.

Note that #10885 misidentified the memory bottleneck. But I preserve the change here since we need to detect the source map file size.

Ideally the mergeSourceMap in babel generator should operate in a streaming mode. But this does not fit in a patch release.

I don't consider it as a regression since in 7.6.4 the `inputSourceMap` is not working properly on external sourcemaps.

@artisb45 @midgethoen Though the URL does not change, [here](https://gist.github.com/JLHwung/a4e735de002261d84764b8c3ed136c24) is a new fix. You can download it and overwrite `node_modules/@babel/core/lib/transformation/normalize-file.js`.

### DIFF
--- a/packages/babel-core/src/transformation/normalize-file.js
+++ b/packages/babel-core/src/transformation/normalize-file.js
@@ -13,6 +13,7 @@ import File from "./file/file";
 import generateMissingPluginMessage from "./util/missing-plugin-helper";
 
 const debug = buildDebug("babel:transform:file");
+const LARGE_INPUT_SOURCEMAP_THRESHOLD = 1_000_000;
 
 export type NormalizedFile = {
   code: string,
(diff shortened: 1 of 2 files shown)


packages/babel-core/src/transformation/normalize-file.js
```diff
@@ -65,14 +66,18 @@ export default function normalizeFile(
...
-          if (!match) throw new Error("Invalid source map comment format.");
-          inputMap = convertSourceMap.fromJSON(
-            fs.readFileSync(
-              path.resolve(path.dirname(options.filename), match[1]),
-              "utf8",
-            ),
+          // when `lastComment` is non-null, EXTERNAL_SOURCEMAP_REGEX must have matches
+          const match: [string, string] = (EXTERNAL_SOURCEMAP_REGEX.exec(
+            lastComment,
+          ): any);
```
nicolo-ribaudo: Is this guaranteed to never return `null`?

JLHwung: Yes because the capturing group `([^\s'"`]+)` has a `+` quantifier, if the whole regex passed, the capturing group must not be `null`.

JLHwung:
@nicolo-ribaudo Correct. I have tested on `pdfjs-dist/build/pdf.js.map` (1.21 MB) and it does not bail. To be conservative I shrink the threshold to 1M. We can easily change it to an option if people really need to configure.

PR #10897, opened 2019-12-19
Try to resolve browserslist config when targets are not specified

### LINKED ISSUE #9962: Only change browserslist defaults if no config file exists
https://github.com/babel/babel/pull/8897#issuecomment-490659840

> I guess the best solution would be to check ourself if there is a browserlist config in the fs if there are not targets specified. We could do this with `browserlist.findConfig(from)`. If we find one we do not change defaults. But that would also be a breaking change so I guess we would need to wait for v8. https://github.com/browserslist/browserslist/blob/master/node.js#L238

JLHwung:
| Fixed Issues?            | Resolves #9962 
| Documentation PR Link    | 

This PR changes the targets parsing logic when targets are empty. It will now try to resolve any `browserslists` external configuration and fallback to `forceAllTransforms: true` when neither `targets` is specified nor `browserslists` configuration is resolved.

This PR also generally improves integrations with `browserslists`: the `browserslists.default` is never changed during targets parsing. Surprisingly, by doing so we can remove a bunch of dead code.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/10897"></a>


packages/babel-preset-env/src/targets-parser.js
```diff
@@ -194,35 +177,36 @@ const getTargets = (targets: Object = {}, options: Object = {}): Targets => {
...
 
-  if (shouldParseBrowsers || shouldSearchForConfig) {
-    // If no targets are passed, we need to overwrite browserslist's defaults
-    // so that we enable all transforms (acting like the now deprecated
-    // preset-latest).
-    //
-    // Note, if browserslist resolves the config (ex. package.json), then usage
-    // of `defaults` in queries will be different since we don't want to break
-    // the behavior of "no targets is the same as preset-latest".
-    if (!hasTargets) {
```
JLHwung: When `hasTargets` is `false`, `objectToBrowserslist(targets)` always return `[]`.

packages/babel-preset-env/src/targets-parser.js
```diff
@@ -194,35 +177,36 @@ const getTargets = (targets: Object = {}, options: Object = {}): Targets => {
...
+    // we return an empty array so that "no targets and no browserslist is the same as
+    // preset-latest"
+    if (!browsersquery) {
+      browsersquery = [];
+    }
+  }
 
+  if (shouldParseBrowsers || shouldSearchForConfig) {
     const browsers = browserslist(browsersquery, {
-      path: options.configPath,
```
JLHwung: Because `browsersquery` is never `undefined` here, `browserslists` will never `loadConfig` and therefore we don't need to pass `path` option here.

packages/babel-preset-env/src/targets-parser.js
nicolo-ribaudo: Is there any difference between `[]` and `["defaults"]`?

packages/babel-preset-env/src/targets-parser.js
```diff
@@ -194,35 +177,36 @@ const getTargets = (targets: Object = {}, options: Object = {}): Targets => {
...
+      browsersquery = browserslist.loadConfig({
+        path: options.configPath,
+      });
     }
+    // If there is no external browserslist config and no targets are specified,
+    // we return an empty array so that "no targets and no browserslist is the same as
+    // preset-latest"
+    if (!browsersquery) {
+      browsersquery = [];
+    }
```
nicolo-ribaudo: I think that at this point it would make sense to default to `["defaults"]`?

JLHwung: Well we can defaults to `defaults` here, need to tell people that if no targets are specified, it will always be `browserslists configuration ?? browserslist defaults`. And one can only use `forceAllTransforms: true` to apply all of them. I prefer `["defaults"]` here, let's see if people have different opinions.

nicolo-ribaudo: I thought that we were discussing about this during the last meeting :sweat_smile:

JLHwung: Then I misunderstood the plans. It is now addressed in 8d5cef3.

JLHwung:
@nicolo-ribaudo Note that although technically `defaults` is not static, it is almost invisible for our users as long as `ie11` is still included in `defaults`, which is accounted for most es6 transformers.

nicolo-ribaudo:
Superseded by https://github.com/babel/babel/pull/12989

PR #10898, opened 2019-12-20
Move coverage to GitHub actions

JLHwung:
| Fixed Issues?            | `ENOMEM: not enough memory, stat /home/circleci/babel/packages/just/click/the/rerun/button`

This PR moves the coverage report job to GitHub Actions. The ENOMEM error should be alleviated because

| | CPU | RAM |
| --- | --- | --- |
| [GitHub](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources) | 2 core | 7GB |
| [CircleCI](https://circleci.com/docs/2.0/configuration-reference/#docker-executor) | 2 core | 4GB |

Merge tips:
Please add `CODECOV_TOKEN` to the `Settings > Secrets` after merging to master. See https://github.com/codecov/codecov-action#usage

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -18,22 +18,7 @@ jobs:
       - name: Environment log
         id: env
         run: |
-          echo "::set-output name=yarn-cache-dir::$(yarn cache dir)"
           yarn --version
-      - name: Get node_modules cache
-        uses: actions/cache@v1
-        with:
-          path: node_modules
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-yarn-
-      - name: Get yarn cache
-        uses: actions/cache@v1
-        with:
-          path: ${{ steps.env.outputs.yarn-cache-dir }}
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-yarn-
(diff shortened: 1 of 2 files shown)


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Generate coverage report
+        run: |
+          yarn --version
+          make test-ci-coverage
+          # Builds babel-standalone with the regular Babel config
+          # test-ci-coverage doesn't test babel-standalone, as trying to gather coverage
+          IS_PUBLISH=true make build-standalone
+          # data for a JS file that's several megabytes large is bound to fail. Here,
+          # we just run the babel-standalone test separately.
+          yarn jest "\-standalone/test"
```
nicolo-ribaudo: Since this is just for coverage, maybe we can avoid testing standalone here?

JLHwung: The standalone is only tested on the coverage job in the CI, we have skipped the test on other executors.

nicolo-ribaudo: Oh then maybe we should test it separately? I don't really see how it fits in the coverage tests.

JLHwung: Yeah but standalone will rely on building and installing all the dependencies, I think we can test it here to save redundant machine hours.

nicolo-ribaudo: Ok fair

PR #10899, opened 2019-12-20
Extract targets parser and compat data from preset-env

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is PR only moves code around, to expose as public packages internal API which are already used by some Babel integrations:
- https://github.com/vuejs/vue-cli/blob/00acca5f14be9f69b763ba4b755c6290368cae2c/packages/%40vue/babel-preset-app/index.js#L19
- https://github.com/babel/ember-cli-babel/blob/fb97b9fddc598786558f8ce8cfc70882b6fb9618/index.js#L468
- https://github.com/babel/ember-cli-babel/blob/fb97b9fddc598786558f8ce8cfc70882b6fb9618/index.js#L203
Otherwise, they will have problems in Babel 8 with https://github.com/babel/babel/pull/10850.

I created two new packages: `@babel/compat-data` and `@babel/helper-compilation-targets`.
I wasn't sure about where to put the info about overlapping plugins which will be useful for `preset-modules` (I put it in `@babel/compat-data`) and about the mappings from transform plugins to syntax plugins (I left it in `@babel/preset-env`).


loganfsmyth:
I swear at this point we should just append a random UUID to every filename except `index.js` on publish :P

nicolo-ribaudo:
Did you see https://github.com/babel/babel/pull/10850? 😛

loganfsmyth:
Hah, nope! Had no idea that was a thing Node allowed!

packages/babel-compat-data/scripts/build-data.js
```diff
@@ -7,7 +7,7 @@ const flattenDeep = require("lodash/flattenDeep");
 const isEqual = require("lodash/isEqual");
 const mapValues = require("lodash/mapValues");
 const pickBy = require("lodash/pickBy");
-const unreleasedLabels = require("../data/unreleased-labels");
+const { unreleasedLabels } = require("../../babel-helper-compilation-targets");
```
JLHwung: Can we just require `@babel/helper-compilation-targets`?

nicolo-ribaudo: Good catch; I'll update it. Note that this file isn't published to npm so it isn't a problem, but using `@babel/helper-compilation-targets` is more clear.

packages/babel-compat-data/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@babel/compat-data",
+  "version": "0.0.0",
+  "author": "The Babel Team (https://babeljs.io/team)",
+  "license": "MIT",
+  "description": "",
```
JLHwung: Maybe "Engine compat data used in @babel/preset-env"?

packages/babel-compat-data/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "./native-modules": "./data/native-modules.json",
+    "./corejs2-built-ins": "./data/corejs2-built-ins.json",
+    "./overlapping-plugins": "./data/overlapping-plugins.json"
+  },
+  "scripts": {
+    "build-data": "./scripts/download-compat-table.sh; node ./scripts/build-data.js; node ./scripts/build-modules-support.js; node ./scripts/build-overlapping-plugins.js"
+  },
+  "keywords": [
+    "babel",
+    "babel-plugin"
```
JLHwung:
```suggestion
    "compat"
```
`babel-plugin` may be misleading here.

packages/babel-helper-compilation-targets/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@babel/helper-compilation-targets",
+  "version": "0.0.0",
+  "author": "The Babel Team (https://babeljs.io/team)",
+  "license": "MIT",
+  "description": "",
+  "repository": "https://github.com/babel/babel/tree/master/packages/babel-helper-compilation-targets",
+  "main": "lib/index.js",
```
JLHwung: we can disable `exports` for new package.

existentialism:
Any reason to not also move `isPluginRequired` as another primitive to the helper?

In fact, erring on the side of _less_ packages to maintain, can we just have one package that exports `getTargets`, our plugin/target data and something like:

```js
isRequired(
  name: string,
  targetEnvironments: Targets,
  options: {
    compatData?: { [key in string]: Targets }
    // ...
  }
) { ... }
```

This seems like a key use-case (as both vue/ember use it), and would allow anyone to pass their own mappings in, if they wanted.

nicolo-ribaudo:
> Any reason to not also move `isPluginRequired` as another primitive to the helper?

Maybe I forgot it, I'll check.

> In fact, erring on the side of less packages to maintain

I don't think that this should be a goal? The maintenance cost of 2 packages in a monorepo is exactly the same as 1 package.

I created two separate packages because, if we are going to disable sub-package imports in Babel 8.0.0, `@babel/compat-data` will already be an exception when you can import internal files. If we merge it with the helper package, then _some_ internal files will be importable while others won't, and I think that it would be more confusing.

packages/babel-helper-compilation-targets/src/filter-items.js
```diff
@@ -0,0 +1,85 @@
+// @flow
+
+import semver from "semver";
+
+import type { Targets } from "./types";
+import { isUnreleasedVersion, semverify } from "./utils";
+
+export function isItemRequired(supportedEnvironments: Targets, item: Targets) {
```
nicolo-ribaudo: Oh, I already extracted it. I called it `isItemRequired` because it's not just used for plugins, but also for polyfills. Maybe I should choose another name?

existentialism: I think the name is fine.

nicolo-ribaudo:
@existentialism I just noticed that the function you are proposing is different. I agree that it's worth to add it, given that as you pointed out both ember and vue (which would probably be the main "famous" users of this package) are implementing that logic.

nicolo-ribaudo:
Whops I accidentally deleted my comment

existentialism:
> The maintenance cost of 2 packages in a monorepo is exactly the same as 1 package.

I'm not sure I agree with that... perhaps mechanically, but I think we _should_ always consider the balance between a package's public API vs. adding more packages.

> Also, do you think that isRequired should default to using the plugins and core-js 2 data, or just the plugins data? And then there is also core-js 3, which keeps its data in a separate package.

I think we can go with just plugins data for now.

nicolo-ribaudo:
@existentialism I ended up not exporting `isItemRequired` (which I also renamed) to avoid confusion with `isRequired`.

PR #10900, opened 2019-12-20
Print Babel config

### LINKED ISSUE #10617: RFC: Print Babel Configuration
## Introduction

This document aims to address the requirement of printing babel configuration. The concept of configuration and environment information is clarified. Two new config items are introduced.

There have been multiple issues requesting this feature (#2960, #6856, #10538), of which we shall agree on the popularity and value.

There are two PRs (#4806, #7201) implementing this feature. Both of them were stale due to inactivity or lack of consensus. The scope of #4806 and #7201 are slightly different: the former prints the _resolved configuration_ while the latter prints the config items `plugins` and `presets`. Note that both of them prints babel plugin versions, which should be part of environment information.

Before we propose a solution, we should first clarify the concept of babel configuration and environment information.

## Concept

__Babel Configuration__ is a specific file content and/or user-supplied JavaScript object. Here is an example list: 1) contents of `.babelrc`, 2) contens of `babel` field in `package.json` and 3) the `opt` argument of `babel.transform(code, opt)`. `@babel/core` will process this  configuration lists and merge them into a __resolved configuration__.

__Environment Information__ is an umbrella term includes but is not limited to: versions of the OS, `node`, `npm`; versions of `babel` and its plugins/presets; versions of monorepo tools `lerna`, `yarn workspace`. A list of environment information is required in Babel's bug template to convey the context of this issue.

## Scope

Printing Environment Information is orthogonal to Babel and thus should not be implemented in `@babel/cli`. We have proposed a [`babel` preset](https://github.com/tabrindle/envinfo/pull/126) for `envinfo` and we expect it will soon be installed to issue templates.

In this document we will only address printing babel configs.

## Solution

Add a new config item: `showConfig: boolean`. Its default value is `false`.

This boolean option is an opt-in switch to output configs. The alternative environment variable is considered to use only when `@babel/core` is a transitive dependency or the configuration is not controlled by end user. 

Add a new config item: `showConfigForPath: string`. If it is not specified and `BABEL_SHOW_CONFIG_FOR_PATH` environment variable exists, set `showConfigForPath` to be its value. If this value is not nullish, it will imply that `showConfig` is `true`. When `showConfig` is specified, its default value is assigned from the `filename` transform option.

This option is used to eliminate the ambiguity of configuration chain when multiple babel configurations exist in different paths. Instead of printing all configurations which may be irrelevant, only applicable config with respect to `showConfigForPath` will be printed.

A new `@babel/cli` options `--show-config` is added. Note that `--show-config-for-path` is redundant as it can be inferred from the `path` argument.

When multiple paths are passed in `@babel/cli`, only the first one will be respected.

### Procedure
When loading configuration:

<pre>
if `showConfig` is `true`, compare `options.filename` with `showConfigForPath`:
  if it does not match, continue;
  else 
   a) serialize the applicable configuration descriptor
   b) print to stdout
   c) terminate the process with code 1
</pre>

The last step ensures that 
- the configuration would not be overwhelmed by other logs
- debug feedback is faster than running through the whole process.

### Using `@babel/cli`
```shell
babel src/foo.js --show-config --plugins=@babel/proposal-class-properties,@babel/transform-class
```
Example output:
```
// cli config
{
  "plugins": [
    "@babel/transform-class"
  ],
  "showConfig": true
}

// babel config ./src/.babelrc
{
  "plugins": [
    "lodash"
  ]
}

// babel config ./babel.config.cjs (serialized)
{
  "presets": [
    ["@babel/preset-env", { useBuiltIns: true }]
  ],
  "overrides": [
    { "test": "./src/foo.js", "compact": true }
  ]
}

// babel config ./package.json#babel
{
  "babelrcRoots": [
    ".",
    "./src/*"
  ]
}
```

### Using `webpack` and `babel-loader`

```shell=bash
BABEL_SHOW_CONFIG_FOR_PATH="./src/app.jsx" webpack --config webpack.config.js
```
Example output:
```
// user supplied config (serialized)
{
  "sourceMaps": "inline"
}

// babel config ./babel.config.cjs
{
  "presets": ["@babel/react"]
}
```

### Using `create-react-app`

```shell=bash
BABEL_SHOW_CONFIG_FOR_PATH="./src/app.jsx" react-scripts build
```
Example output:
```
// user supplied config (serialized)
{
  ... babel configuration injected by CRA
}

// babel config ./src/.babelrc.js (serialized)
{
  "plugins": [
    "lodash"
  ]
}
```

### static (.babelrc, babel.config.json, package.json#babel)

The static configuration is JSON format, therefore the config we get is a plain object, it is straightforward to stringify it back and output to console.

### dynamic (babel.config.js, babel.config.cjs, .babelrc.js, .babelrc.cjs)

The configuration object is produced from `require`ing CommonJS module before processing config items.

Serializing this object with customized JSON replacer `replace(key, value)`:

If `key` is `plugins` or `presets`, for every array item, check by `item`, `item.name`, `item[0].name` for plugin/preset name

Otherwise if `value` is of function type, omit this property

Mark the output as `serialized` as it may be different from the original dynamic config.

## Questions

Q: Should we support `--show-merged-config` flag to output a _merged_ JSON configuration?

A: Would be good to include it in next phase. It should help reasoning about the ignore/include config items.

Q: Should we print the resolved presets config? (`presets: ["@babel/env"]` vs `plugins: [...]`)

A: Not prioritized. It doesn't help on plugin ordering issues and it makes printed result more verbose.

Q: Should we print the _extended_ config item? i.e. `extends: "babel-preset-foo"`

A: Not prioritized. I don't think it is popular.

## Prior Art
This document is inspired from the following prior arts.

TypeScript Compiler `tsc` has a `--showConfig` option. It prints the merged configuration from the contents of `tsconfig.json` and default configs.
```
$ tsc --showConfig foo.ts

{
    "compilerOptions": {},
    "files": [
        "./foo.ts"
    ]
}
```

`npm` has a subcommand `npm config list`. It lists a chain of applicable configs with respect to current working directory.
```
$ npm config list

; cli configs
metrics-registry = "https://registry.npmjs.org/"

; userconfig /Users/foo/.npmrc
tag-version-prefix = ""

; globalconfig /usr/local/etc/npmrc
sign-git-tag = true
```

rajasekarm:
| Fixed Issues?            | Fixes #10617 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### Through CLI
- [x] Add showConfig flag in babel-cli
- [x] Expose babelrc config from loadconfig
- [ ] Print config for babel/cli, babelrc, package.json, 
- [ ] If path is a file, print and exit, If path is a dir print for first file and exit.

###  Through webpack and babel-loader
- [ ] Print config if BABEL_SHOW_CONFIG_FOR_PATH env is available

### DIFF
--- a/packages/babel-cli/src/babel/dir.js
+++ b/packages/babel-cli/src/babel/dir.js
@@ -47,15 +47,16 @@ export default async function({
     const dest = getDest(relative, base);
 
     try {
-      const res = await util.compile(
-        src,
-        defaults(
-          {
-            sourceFileName: slash(path.relative(dest + "/..", src)),
-          },
-          babelOptions,
-        ),
+      const opts = defaults(
+        {
+          sourceFileName: slash(path.relative(dest + "/..", src)),
+        },
+        babelOptions,
       );
+      if (!print) {
+        opts.showConfig = false;
+      }
+      const res = await util.compile(src, opts);
 
       if (!res) return FILE_TYPE.IGNORED;
 
@@ -96,8 +97,12 @@ export default async function({
     return path.join(cliOptions.outDir, filename);
   }
 
-  async function handleFile(src: string, base: string): Promise<boolean> {
-    const written = await write(src, base);
+  async function handleFile(
+    src: string,
+    base: string,
+    print: boolean,
+  ): Promise<boolean> {
+    const written = await write(src, base, print);
 
     if (
       (cliOptions.copyFiles && written === FILE_TYPE.NON_COMPILABLE) ||
@@ -122,17 +127,19 @@ export default async function({
       let count = 0;
 
       const files = util.readdir(dirname, cliOptions.includeDotfiles);
+      let print = true;
       for (const filename of files) {
         const src = path.join(dirname, filename);
 
-        const written = await handleFile(src, dirname);
+        const written = await handleFile(src, dirname, print);
         if (written) count += 1;
+        print = false;
       }
 
       return count;
     } else {
       const filename = filenameOrDir;
-      const written = await handleFile(filename, path.dirname(filename));
+      const written = await handleFile(filename, path.dirname(filename), true);
 
       return written ? 1 : 0;
     }
--- a/packages/babel-cli/src/babel/file.js
+++ b/packages/babel-cli/src/babel/file.js
@@ -164,7 +164,7 @@ export default async function({
         _filenames.push(filename);
       }
     });
-
+    let print = true;
     const results = await Promise.all(
       _filenames.map(async function(filename: string): Promise<Object> {
         let sourceFilename = filename;
@@ -175,7 +175,10 @@ export default async function({
           );
         }
         sourceFilename = slash(sourceFilename);
-
+        if (!print) {
+          babelOptions.showConfig = false;
+        }
+        print = false;
         try {
           return await util.compile(
             filename,
--- a/packages/babel-cli/src/babel/options.js
+++ b/packages/babel-cli/src/babel/options.js
@@ -169,6 +169,7 @@ commander.option(
   "--copy-ignored",
   "Include ignored files when copying non-compilable files.",
 );
+commander.option("--show-config", "show config");
 
 commander.version(pkg.version + " (@babel/core " + version + ")");
 commander.usage("[options] <files ...>");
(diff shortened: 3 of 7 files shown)


kaicataldo:
Wow, I remember this being discussed years ago. Thanks for working on this!

rajasekarm:
Waiting for this PR to be merged. https://github.com/babel/babel/pull/10507

JLHwung:
@rajasekarm Can you rebase on upstream master? #10507 has been merged.

rajasekarm:
Yeah, I've rebased it. It's in my local. I'll push with the new changes shortly.

rajasekarm:
Closing this PR and opened https://github.com/babel/babel/pull/11018

PR #10902, opened 2019-12-20
fix: Object.getOwnPropertySymbols called on non-object

bassaer:
| Fixed Issues?            | Fix related #10482 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR is related to https://github.com/babel/babel/pull/10683
`TypeError: Object.getOwnPropertySymbols called on non-object` occurred in android default browser (android 4.x 5.x) when it uses the spread operator. 
Because this browser dose not support Object.getOwnPropertySymbols with non-object argument.

スクリーンショット 2019-12-20 23 13 18

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -399,7 +399,7 @@ helpers.objectSpread = helper("7.0.0-beta.0")`
       var source = (arguments[i] != null) ? arguments[i] : {};
       var ownKeys = Object.keys(Object(source));
       if (typeof Object.getOwnPropertySymbols === 'function') {
-        ownKeys = ownKeys.concat(Object.getOwnPropertySymbols(source).filter(function(sym) {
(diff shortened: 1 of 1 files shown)


packages/babel-helpers/src/helpers.js
```diff
@@ -399,7 +399,7 @@ helpers.objectSpread = helper("7.0.0-beta.0")`
       var source = (arguments[i] != null) ? arguments[i] : {};
```
JLHwung: I feel like we can add `Object` to `arguments[i]` and forget about `Object(source)` later.

bassaer: you mean like this?
```
var source = (arguments[i] != null) ?  Object(arguments[i]) : {};
```

bassaer: fixed https://github.com/babel/babel/pull/10902/commits/1366ed3c1ef80dbe4b11996b0da8de08eceb4ac8

Issue #10755, opened 2019-11-23
Add support for .mjs config files


xiaoxiangmoe:
## Feature Request

Add `babel.config.mjs` config file.

nicolo-ribaudo:
https://github.com/babel/babel/pull/10507 is the first step in this direction.

PR #10903, opened 2019-12-21
Add support for babel.config.mjs and .babelrc.mjs

nicolo-ribaudo:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/10755
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | https://github.com/babel/website/pull/2150
| Any Dependency Changes?  |

**NOTE**: This PR is a diff on top of https://github.com/babel/babel/pull/10507, but they can be reviewed in any order.

This PR adds support for `*.mjs` config files: this is one of the last steps (with https://github.com/babel/babel/pull/10783) to support as configuration files all the extensions natively supported by nodejs.

Since native ECMAScript modules can only be loaded asynchronously, `.mjs` files can only be used when calling Babel with `.parseAsync`, `.transformFileAsync`, `.transformAsync`, `.loadPartialConfigAsync` or `.loadOptionsAsync`. Synchronous calls will throw an error. This is a problem for `@babel/eslint-parser` which will still only support synchronous configuration files, but the ESLint team is working on allowing asynchronous ESLint calls.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -107,6 +107,10 @@ module.exports = function(api) {
       ["@babel/plugin-proposal-nullish-coalescing-operator", { loose: true }],
 
       convertESM ? "@babel/transform-modules-commonjs" : null,
+      // Until Jest supports native mjs, we must simulate it 🤷
+      env === "test" || env === "development"
+        ? "@babel/plugin-proposal-dynamic-import"
+        : null,
     ].filter(Boolean),
     overrides: [
       {
--- a/packages/babel-core/src/config/files/module-types.js
+++ b/packages/babel-core/src/config/files/module-types.js
@@ -1,5 +1,6 @@
 import { isAsync, waitFor } from "../../gensync-utils/async";
 import type { Handler } from "gensync";
+import path from "path";
 
 let import_;
 try {
@@ -28,8 +29,8 @@ export default function* loadCjsOrMjsDefault(
   }
 }
 
-function guessJSModuleType(path: string): "cjs" | "mjs" | "unknown" {
-  switch (path.slice(-4)) {
+function guessJSModuleType(filename: string): "cjs" | "mjs" | "unknown" {
+  switch (path.extname(filename)) {
     case ".cjs":
       return "cjs";
     case ".mjs":
@@ -41,6 +42,7 @@ function guessJSModuleType(path: string): "cjs" | "mjs" | "unknown" {
 
 function loadCjsDefault(filepath: string) {
   const module = (require(filepath): mixed);
+  // TODO (Babel 8): Remove "undefined" fallback
   return module?.__esModule ? module.default || undefined : module;
 }
 
--- a/packages/babel-core/src/config/files/module-types.js
+++ b/packages/babel-core/src/config/files/module-types.js
@@ -1,0 +1,57 @@
+import { isAsync, waitFor } from "../../gensync-utils/async";
+import type { Handler } from "gensync";
+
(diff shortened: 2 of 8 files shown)


babel.config.js
```diff
@@ -121,6 +121,10 @@ module.exports = function(api) {
       ["@babel/plugin-proposal-nullish-coalescing-operator", { loose: true }],
 
       convertESM ? "@babel/transform-modules-commonjs" : null,
+      // Until Jest supports native mjs, we must simulate it 🤷
+      env === "test" || env === "development"
+        ? "@babel/plugin-proposal-dynamic-import"
```
nicolo-ribaudo: I hate this, but this is the best way we have so far to test it :cry:

packages/babel-core/src/config/files/module-types.js
```diff
@@ -0,0 +1,57 @@
...
+    case ".mjs":
+      return "mjs";
+    default:
+      return "unknown";
+  }
+}
+
+function loadCjsDefault(filepath: string) {
+  const module = (require(filepath): mixed);
+  return module?.__esModule ? module.default || undefined : module;
```
JLHwung: Why `undefined` is offered as fallback here?

nicolo-ribaudo: I'm not sure; I only moved this code here from https://github.com/babel/babel/blob/5b907e9bb7e356fe869aaa9e3f8aed8c4991f1b5/packages/babel-core/src/config/files/configuration.js#L181-L185

nicolo-ribaudo: Maybe in case someone exports `false` as their default config? :thinking:

packages/babel-core/test/config-chain.js
```diff
@@ -1024,24 +1020,64 @@ describe("buildConfigChain", function() {
...
+      test("should not load babel.config.mjs synchronously", async () => {
+        const { cwd, tmp, config } = await getTemp(
+          "babel-test-load-config-sync-babel.config.mjs",
+        );
+        const filename = tmp("src.js");
+
+        await config("babel.config.mjs");
+
+        expect(() => loadOptions({ filename, cwd })).toThrow(
+          /is only supported when running Babel asynchronously/,
```
JLHwung: You can use `toThrowErrorMatchingSnapshot()` here and don't have to write the same string again.

packages/babel-core/src/config/files/module-types.js
JLHwung: Alright, we should investigate this in later PRs, or just remove it in Babel 8 because it looks suspicious.

packages/babel-core/src/config/files/module-types.js
```diff
@@ -0,0 +1,57 @@
...
+    case "mjs":
+      if (yield* isAsync()) {
+        return yield* waitFor(loadMjsDefault(filepath));
+      }
+      throw new Error(asyncError);
+  }
+}
+
+function guessJSModuleType(path: string): "cjs" | "mjs" | "unknown" {
+  switch (path.slice(-4)) {
```
kaicataldo: Is it worth using `path.extname()` here?

packages/babel-core/test/config-chain.js
nicolo-ribaudo: I can't use it because the error contains the path of the file in the tmp folder, which might change.

PR #10912, opened 2019-12-23
Fix parameter expression get binding

JLHwung:
| Fixed Issues?            | `getBinding` will return declaration inside the function body for patterns in function parameters.

This PR is a rework of #10055. In this PR I extend the definition of `Scope` to include all the `Pattern` nodes when it immediately descends from the `params` of a Function node. By doing so we can intercept the ancestry queries in `scope.getBinding` and skip the binding if it is _not_ a `param`-kind, because the spec ([9.2.10](https://tc39.es/ecma262/#sec-functiondeclarationinstantiation), Step 28) requires that any closure defined in the parameter expressions should not access the declarations inside the function body.

An example can be found in the `scope` unit test:
https://github.com/babel/babel/blob/fb4125a816f6cc55f608be111939dc7ceccee118/packages/babel-traverse/test/scope.js#L77-L92

Personal sidenote:
This PR is dedicated to @nicolo-ribaudo, who gave insightful comments to #10055, one of my first works when I spent days to merely learn what is scope, and who inspires me and encourages me to work on Babel. Thank you and Merry Christmas @nicolo-ribaudo!

### DIFF
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -902,14 +902,14 @@ export default class Scope {
     do {
       const binding = scope.getOwnBinding(name);
       if (binding) {
-        // When a pattern is a Scope, it is a part of parameter expressions.
+        // Check if a pattern is a part of parameter expressions.
         // 9.2.10.28: The closure created by this expression should not have visibility of
         // declarations in the function body. If the binding is not a `param`-kind,
         // then it must be defined inside the function body, thus it should be skipped
         if (
           previousPath &&
           previousPath.isPattern() &&
-          previousPath.isScope() &&
+          previousPath.parentPath.isFunction() &&
           binding.kind !== "param"
         ) {
           // do nothing
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -897,10 +897,27 @@ export default class Scope {
 
   getBinding(name: string) {
     let scope = this;
+    let previousPath;
 
     do {
       const binding = scope.getOwnBinding(name);
-      if (binding) return binding;
+      if (binding) {
+        // When a pattern is a Scope, it is a part of parameter expressions.
+        // 9.2.10.28: The closure created by this expression should not have visibility of
+        // declarations in the function body. If the binding is not a `param`-kind,
+        // then it must be defined inside the function body, thus it should be skipped
+        if (
+          previousPath &&
+          previousPath.isPattern() &&
+          previousPath.isScope() &&
+          binding.kind !== "param"
+        ) {
+          // do nothing
+        } else {
+          return binding;
+        }
+      }
+      previousPath = scope.path;
     } while ((scope = scope.parent));
   }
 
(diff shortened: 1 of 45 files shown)


packages/babel-types/src/validators/isScope.js
```diff
@@ -18,5 +19,9 @@ export default function isScope(node: Object, parent: Object): boolean {
     return false;
   }
 
+  if (isPattern(node) && isFunction(parent)) {
```
JLHwung: If a `Pattern` is a child of `Function`, it must be one of the `params` key because both `body` and `typeAnnotation` can not _be_ a `Pattern`.

packages/babel-traverse/src/scope/index.js
```diff
@@ -897,10 +897,27 @@ export default class Scope {
 
   getBinding(name: string) {
     let scope = this;
+    let previousPath;
 
     do {
       const binding = scope.getOwnBinding(name);
-      if (binding) return binding;
+      if (binding) {
+        // When a pattern is a Scope, it is a part of parameter expressions.
```
JLHwung: This case is one (likely the only one?) of the exception that although the `Pattern` is descended from `Function`, the scope of `Pattern` is _not_ descended from the one of `Function`. This breaks the assumption that a scope is always initialized from the scope of its nearest Scopable ancestry, which becomes `scope.parent` later.

https://github.com/babel/babel/blob/f651f22f96b2715010cd8a7b65bb12648eba8307/packages/babel-traverse/src/path/context.js#L113

I feel like it is more cumbersome to fix that on `scope.parent` accessor, so I implement the fix on `getBinding`.

packages/babel-traverse/test/scope.js
```diff
@@ -73,6 +73,25 @@ describe("scope", () => {
       ).toBe("Identifier");
     });
 
+    describe("function paramater expression", function() {
+      it("should not has visibility of declarations inside function body", () => {
```
nicolo-ribaudo: Nit: after `should` you shouldn't use the third person but the base form of the verb (`should have`).

packages/babel-types/src/validators/isScope.js
nicolo-ribaudo: If you need to add a comment to explain your PR, then why don't add it directly in the code? :wink:

packages/babel-traverse/src/scope/lib/renamer.js
```diff
@@ -9,8 +9,9 @@ const renameVisitor = {
     }
   },
 
-  Scope(path, state) {
+  "Pattern|Scope"(path, state) {
```
nicolo-ribaudo: Since now the correct `Pattern`s are already scope, aren't they already matched by the `Scope` visitor?

packages/babel-traverse/test/scope.js
JLHwung: Oops.

nicolo-ribaudo:
@JLHwung wrong button?

JLHwung:
@nicolo-ribaudo Oops.

PR #10913, opened 2019-12-23
@babel/eslint-parser: update scope analysis for ClassPrivateProperty

kaicataldo:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Refs https://github.com/babel/babel/issues/10752 
| Patch: Bug Fix?          | :+1:
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

This PR updates the scope analysis for `ClassPrivateMethod`s once https://github.com/babel/babel/pull/10914 is merged and we're parsing these nodes correctly.

Note that `ClassPrivateMethod`s still aren't 100% compatible with `no-undef`/`no-unused-vars`, but I didn't want to throw too much into this PR. I'll handle that as part of auditing the the plugin and making sure rules work with `@babel/eslint-parser`.

This is the last known issue on the parser side of things!

These changes rely on https://github.com/babel/babel/pull/10914 being merged first.

PR #10914, opened 2019-12-23
@babel-core: parse should parse only

kaicataldo:
| Fixed Issues?            | Refs https://github.com/babel/babel/issues/10752 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This fixes the issue of `ClassPrivateMethod` nodes causing `@babel/core.parse()` to throw. 

Currently, `@babel/core.parse()` is doing a lot more than just parsing - this change makes it so that it now just loads the config and parses with the normalized options.

This will unblock https://github.com/babel/babel/pull/10913.


packages/babel-core/src/parse.js
```diff
@@ -1,14 +1,11 @@
...
+import parser from "./parser";
+import type { ParseResult } from "./parser";
 import normalizeOptions from "./transformation/normalize-opts";
 
-type AstRoot = BabelNodeFile | BabelNodeProgram;
-
-export type ParseResult = AstRoot;
-
-export type FileParseCallback = {
+type FileParseCallback = {
```
nicolo-ribaudo: (Note for reviewers)
This wasn't already imported anywhere.

PR #10915, opened 2019-12-23
breaking: remove parserOpts.objectRestSpread from typescript

JLHwung:

Although it is technically a breaking change, I don't think it is visible to end users as long as they use babel/core 8.x with syntax-typescript 8.x.


nicolo-ribaudo:

PR #5971, opened 2017-07-19
Cleanup babel types

jridgewell:
| Q                        | A 
| ------------------------ | ---
| Patch: Bug Fix?          | No
| Major: Breaking Change?  | Yes
| Minor: New Feature?      | No
| Deprecations?            | Bad node types
| Spec Compliancy?         | Yes?
| Tests Added/Pass?        | No
| Fixed Tickets            | Fixes #4059
| License                  | MIT
| Doc PR                   | 
| Dependency Changes       | 

This provides a much stricter node system. It also does a deep validation of the AST before printing (should we only do this in testing?).

PR #10917, opened 2019-12-23
(opt-in) Improve @babel/types with env.BABEL_TYPES_8_BREAKING

nicolo-ribaudo:
| Fixed Issues?            | Closes #5971
| Patch: Bug Fix?          |
| Major: Breaking Change?  | Yes, but opt-in.
| Minor: New Feature?      | 
| Documentation PR Link    | 
| Any Dependency Changes?  |

:warning: This PR isn't actually breaking, but I used that label for changelog reasons.

As discussed during the [Dec 04](https://github.com/babel/notes/blob/master/2019/12/04.md) meeting, I'm introducing the stricter AST validation of #5971 behind a flag.

We decided to introduce those breaking changes when the `process.env.BABEL_TYPES_8_BREAKING` env variable is enabled, to give time to some popular projects (like `react-native`) to update their code before the Babel 8 release, to provide a smoother update experience to our users.

I didn't include the changes related to the default value of `decorators` (since they are still highly experimental, I don't think we should pollute the AST unless the parser plugin is enabled). Also, I didn't include the `validateDeep` logic, but we can introduce it later in our test runner.

Our tests are already passing when `BABEL_TYPES_8_BREAKING=true`.


packages/babel-types/src/asserts/generated/index.js
```diff
@@ -224,12 +224,12 @@ export function assertArrowFunctionExpression(
 export function assertClassBody(node: Object, opts?: Object = {}): void {
   assert("ClassBody", node, opts);
 }
-export function assertClassDeclaration(node: Object, opts?: Object = {}): void {
-  assert("ClassDeclaration", node, opts);
-}
 export function assertClassExpression(node: Object, opts?: Object = {}): void {
   assert("ClassExpression", node, opts);
 }
+export function assertClassDeclaration(node: Object, opts?: Object = {}): void {
```
JLHwung: I think they are sorted alphabetically?

nicolo-ribaudo: This file is automatically generated. I had to swap then in the `definitions/es2015` file, so they got swapped here.

packages/babel-types/src/definitions/experimental.js
```diff
@@ -26,9 +26,16 @@ defineType("AwaitExpression", {
 defineType("BindExpression", {
   visitor: ["object", "callee"],
   aliases: ["Expression"],
-  fields: {
-    // todo
-  },
+  fields: !process.env.BABEL_TYPES_8_BREAKING
```
nicolo-ribaudo: I used `!` everywhere so that, when deleting it for Babel 8, I only have to make one selection (from `!` to `:`) rather than two (from `!` to `?` and then from `:` to the end).

Dev-optimization? :smile:

PR #10924, opened 2019-12-25
Replace custom "fundSuggestion" function with "levenary"

nicolo-ribaudo:
| Documentation PR Link    | 

This PR is extracted from https://github.com/babel/babel/pull/10899. In https://github.com/babel/babel/pull/10899 I needed to use the `findSuggestion` function in two different libraries, so I would have had to duplicate that code.
Instead of duplicating it, I decided to dig into the universe of fastest levenshtein distance functions on npm to find one with the interface that we actually need.
I initially used `didyoumean2`, but then switched to `levenary` because:
1) it's faster and because :wink:
2) it's created by @tanhauhau 
3) it provides flow typings out of the box.

I extracted this commit to a separate PR because @JLHwung is going to work on some content which touches similar lines/depends on this package, so we can merge this without waiting for https://github.com/babel/babel/pull/10899.

### DIFF
--- a/lib/third-party-libs.js.flow
+++ b/lib/third-party-libs.js.flow
@@ -190,12 +190,6 @@ declare module "convert-source-map" {
   };
 }
 
-declare module "js-levenshtein" {
-  declare module.exports: {
-    (string, string): number,
-  };
-}
-
 declare module "core-js-compat/data" {
   declare type Target = "node" | "chrome" | "opera" | "edge" | "firefox" | "safari" | "ie" | "ios" | "android" | "electron" | "samsung";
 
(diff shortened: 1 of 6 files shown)


tanhauhau:
Published [levenary@1.1.0](https://www.npmjs.com/package/levenary/v/1.1.0) with @JLHwung's PR https://github.com/tanhauhau/levenary/pull/1

SimenB:
This is a breaking change - levenary depends on Node 8

SimenB:
`leven` itself [is node 6](https://github.com/sindresorhus/leven/blob/252edb6e8c274b78fbc3a3bc4a01b4896a513354/package.json#L13), so it should just be a matter of downgrading the requirement in `levenary`: https://github.com/tanhauhau/levenary/pull/3. Hopefully that's all that's needed

nicolo-ribaudo:
Thanks for catching this! cc @tanhauhau

PR #10925, opened 2019-12-25
support different token,make support different languages javascript

zy445566:
support different token,make support different languages.
for example:
```js
const babelCore = require('@babel/core')
babelCore.tokTypes._function.keyword = '函数'
babelCore.tokTypes._if.keyword = '如果'
babelCore.tokTypes._return.keyword = '返回'
const result = babelCore.transformSync(`
函数 斐波那契(所求数) {
    如果(所求数<=2){返回 1};
    返回 斐波那契(所求数-2)+斐波那契(所求数-1);
}
`);
console.log(result.code)
/**
 * output:
 * function 斐波那契(所求数) {
 *    if (所求数 <= 2) {
 *       return 1;
 *    };
 *    return 斐波那契(所求数 - 2) + 斐波那契(所求数 - 1);
 *  }
 */
```

### DIFF
--- a/packages/babel-parser/src/tokenizer/types.js
+++ b/packages/babel-parser/src/tokenizer/types.js
@@ -66,11 +66,11 @@ export class TokenType {
         return keyword;
       },
       set(val) {
+        if (keyword && keywords.has(keyword)) {
+          keywords.delete(keyword);
+        }
         if (!val) {
           return;
-        }
-        if (keywords.has(keyword)) {
-          keywords.delete(keyword);
         }
         keywords.set(val, this);
         this.label = val;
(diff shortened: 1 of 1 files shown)


JLHwung:
`babelCore.tokTypes` is considered internal implementation details and thus never meant to be exposed to public APIs. If you would like to support custom keywords, you can fork `@babel/parser` and use `parserOverride` to instruct babel uses your fork to parse the custom keywords.

```js
// babel-plugin-漢語關鍵字
const parse = require("漢語JavaScript解析器");

module.exports = {
  plugins: [{
    parserOverride(code, opts) {
      return parse(code, opts);
    },
  }]
}

// test.js
const babelCore = require('@babel/core')
babelCore.transformSync(`
函数 斐波那契(所求数) {
    如果(所求数<=2){返回 1};
    返回 斐波那契(所求数-2)+斐波那契(所求数-1);
}
`, { plugins: ["babel-plugin-漢語關鍵字"] })
```

nicolo-ribaudo:
I know that @hzoo had an idea about this.
_In case_ we end up including something like this in the official package, I think that we should only export the `keywords` map from `@babel/parser`.

zy445566:
Good idea, I will try.

PR #10927, opened 2019-12-26
Type checking preset options

JLHwung:
| Fixed Issues?            | Add strict type checking and _did you mean_ tips to preset-{flow,react,typescript}

The design is copied from preset-env. To avoid duplicate codes, we can extract the validators into a helper library. Ideally we can use schema to check these types, but this PR should include the only visible breaking changes: we are now throwing type errors like `useBuiltIns: 0` or invalid options `useBuiltins: true` instead of accepting/discarding them silently.

Todo: 
- [x] Add test on invalid top level options.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/10927"></a>

### DIFF
--- a/packages/babel-preset-flow/src/normalize-options.js
+++ b/packages/babel-preset-flow/src/normalize-options.js
@@ -1,0 +1,42 @@
+// @flow
+
+import findSuggestion from "levenary";
+
+const PACKAGE_NAME = "@babel/preset-flow";
+
+const TopLevelOptions = ["all", "allowDeclareFields", "ignoreExtensions"];
+
+function validateTopLevelOptions(options) {
+  for (const option of Object.keys(options)) {
+    if (!TopLevelOptions.includes(option)) {
+      const suggestion = findSuggestion(option, TopLevelOptions);
+      throw new Error(
+        `${PACKAGE_NAME}: '${option}' is not a valid top-level option.\n` +
+          `Maybe you meant to use '${suggestion}'?`,
+      );
+    }
+  }
+}
+
+function validateBoolOption(
+  opts: {},
+  key: string,
+  defaultValue: ?boolean,
+): ?boolean {
+  const value = opts[key] ?? defaultValue;
+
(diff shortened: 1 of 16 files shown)


milesj:
Shameless plug. I wrote a library to do just this with a clean fluent API. https://github.com/milesj/optimal

JLHwung:
@milesj Thanks for the info. 

We haven't decided yet how the preset option validation should be reworked. During talks to @nicolo-ribaudo, he proposes a JSON-schema based validator so we can ship schema for better integration to IDE. The goal is to help developers write correct configuration instead of to discourage them via complicated validate message.

This PR is only meant to extend the current validation logic from preset-env to other three babel presets. We can switch to a more formal validation libraries later but this PR should be complete for UI visibility.

JLHwung:
This PR has been revamped featuring less boilerplate of option validation.

packages/babel-preset-flow/src/normalize-options.js
```diff
@@ -0,0 +1,42 @@
...
+      const suggestion = findSuggestion(option, TopLevelOptions);
+      throw new Error(
+        `${PACKAGE_NAME}: '${option}' is not a valid top-level option.\n` +
+          `Maybe you meant to use '${suggestion}'?`,
+      );
+    }
+  }
+}
+
+function validateBoolOption(
```
kaicataldo, line 21: Can we make this DRYer by extracting these helper functions into a separate module?

JLHwung: These helpers are derived from https://github.com/babel/babel/blob/cfaa70dcf48a0ea9a6c7adb4ae1f7925934b3933/packages/babel-preset-env/src/normalize-options.js#L139

I think we should address this in a separate PR on v7, then we can rebase this PR on that.

kaicataldo: Sounds good to me :)

Issue #10933, opened 2019-12-27
plugin-transform-modules-commonjs: imported template tag has implied receiver


ajafff:
## Bug Report

- [x] I would like to work on a fix!

**Current Behavior**

Using an imported identifier (default or named import doesn't matter) as tag of a tagged template literal, the transpiled code calls the tag with an implied receiver.

**Input Code**
https://babeljs.io/en/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=JYWwDg9gTgLgBAMwhANHA3gIwIZQL6JQQhwDkSEpA3AFA0UAGDtOUTVQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2016&prettier=false&targets=&version=7.7.4&externalPlugins=%40babel%2Fplugin-transform-modules-commonjs%407.7.5

```js
import foo, {bar} from 'foo';

foo``;
bar``;
```

**Expected behavior/code**

The function should be dereferenced like it's already done for call expressions:

```js
(0, _foo.default)``;
(0, _foo.bar)``;
```

PR #10934, opened 2019-12-27
helper-module-transforms: dereference imported template tag

ajafff:
| Fixed Issues?            | Fixes: #10933
| Documentation PR Link    | 

Prevent imported template tags to be called with an implied receiver.

### DIFF
--- a/packages/babel-helper-module-transforms/src/rewrite-live-references.js
+++ b/packages/babel-helper-module-transforms/src/rewrite-live-references.js
@@ -183,6 +183,7 @@ const rewriteReferencesVisitor = {
 
       if (
         (path.parentPath.isCallExpression({ callee: path.node }) ||
+          path.parentPath.isOptionalCallExpression({ callee: path.node }) ||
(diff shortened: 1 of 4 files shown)


packages/babel-helper-module-transforms/src/rewrite-live-references.js
```diff
@@ -182,7 +182,8 @@ const rewriteReferencesVisitor = {
       ref.loc = path.node.loc;
 
       if (
-        path.parentPath.isCallExpression({ callee: path.node }) &&
+        (path.parentPath.isCallExpression({ callee: path.node }) ||
+          path.parentPath.isTaggedTemplateExpression({ tag: path.node })) &&
```
JLHwung: Not related to this PR, but the following case also has been broken for the same reason:
```js
foo3?.()
```
Can you also fix the `OptionalCallExpression` case?

ajafff: done

PR #10938, opened 2019-12-29
StringNumericLiteral does not include NumericLiteralSeparator

JLHwung:
| Fixed Issues?            | `proposal-numeric-separator` incorrectly removes `_` in `Number` argument.
| Patch: Bug Fix?          | Yes, but given that it does change the behaviour, minor is expected.
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | Nope, `Number("4_2")` is never mentioned in the docs.

The String Type Interpretation Grammar of `ToNumber` was updated in https://github.com/tc39/proposal-numeric-separator/commit/fc53cdbb9d2e755c1e14d44f9c4ef131d1bef74f, the `StringNumericLiteral` does not allow a NumericLiteralSeparator. Therefore the following example should return `NaN`

```js
new Number("4_2")
```

but Babel will transform `Number("4_2")` into `Number("42")`, which is incorrect.
This PR removes the transformation on `Number` arguments.


JLHwung:
@nicolo-ribaudo Nope. I find it by searching `CallExpression:` in the code base.

Background: I was reviewing an estree AST design PR and inspired from `https://github.com/estree/estree/pull/204#issuecomment-565686514` (prevent GitHub backtrace)

> It's a severe concern that non-optional member accesses and function calls are getting two different representations.

It is exactly how things are done in babel-parser, so I feel like there will be some potential bugs when a `CallExpression` visitor does not take `OptionalCallExpression` into account. But in this case I fix this bug by removing the code lol.

PR #10940, opened 2019-12-29
Remove unnecessary trailing comma record when `maybeAsyncArrow: false`

JLHwung:

This PR includes commits in #10937, so only commits descending from 7d5e8c204d74ab9009490f02af9d87e490d85345 are relevant.

This PR pass the `nodeForExtra` argument in `parseCallExpressionArguments` only when `state. maybeAsyncArrow: true`. The `nodeForExtra` is used to bookkeeping the trailing comma of the last parameters, it is used in `toAssignable` later to throw on unexpected trailing comma.

When `maybeAsyncArrow` is `false`, the `parseAsyncArrowFromCallExpression => parseArrowExpression => setArrowFunctionParameters => toAssignable` chain is never called, which means `node.extra.trailingComma` is never used. Therefore we don't need to add the `extra .traillingComma` property when `maybeAsyncArrow: false`.


nicolo-ribaudo:
@JLHwung Do you want to revive this?

JLHwung:
@nicolo-ribaudo Thanks for the heads-up. I will rebase this PR.

liuxingbaoyu (approved):

nicolo-ribaudo (approved):

nicolo-ribaudo:
Hi @sxzz, we didn't consider this patch as a breaking change because `extra.trailingComma` is already not always generated for async functions, so it's not possible to consistently rely on it.

However, if it's causing problem we can delay this change to Babel 8.

PR #10944, opened 2019-12-30
When reading a new string, U+2028/2029 should correctly set the new column

### LINKED ISSUE #10435: [parser] x2028 and x2029 should increment location lines
## Bug Report

**Current Behavior**

While line terminators are not allowed in strings (unless continuation), and `\u2028` and `\u2029` are considered line terminators explicitly, they are still allowed inside a string without needing to be escaped. Should this increase the line counter? If multi-line comments and templates do, then I think this should as well.

The same goes for 2028 and 2029 in templates.

**Input Code**

Unfortunately it's not "safe" to just copy paste these characters for a test so I'll try to make it very unambiguous;

```js
babel.parse("\u2028")
babel.parse("\u2029")
```

**Expected behavior/code**

The location should have its line incremented.

```
>    Babel AST is different:
>
>    -  loc:{start:{line:1,column:0},end:{line:2,column:2}},
>    +  loc:{start:{line:1,column:0},end:{line:2,column:4}},
>    -        loc:{start:{line:1,column:0},end:{line:2,column:1}},
>    +        loc:{start:{line:1,column:0},end:{line:2,column:3}},
>    -      loc:{start:{line:1,column:0},end:{line:2,column:2}}
>    +      loc:{start:{line:1,column:0},end:{line:2,column:4}}
```

**Environment**

Babel 7.5.5

**Possible Solution**

Treat 2028 and 2029 the same as CR or LF when it comes to location tracking.

JLHwung:
| Fixed Issues?            | Fixes #10435 

As a follow-up to #8866, this PR fixes when the tokenizer reading a string, it does not update the `lineStart` after U+2028/2029 is read. Note that the `lineStart` is correctly updated when reading a template token
https://github.com/babel/babel/blob/b91720c1ccbfc066cbc6af7e3db6114acefac17a/packages/babel-parser/src/tokenizer/index.js#L1229

So I consider it as a bug fix to unify the different behaviour here.

### DIFF
--- a/packages/babel-parser/test/fixtures/es2019/json-strings/template-line-separator/input.js
+++ b/packages/babel-parser/test/fixtures/es2019/json-strings/template-line-separator/input.js
@@ -1,0 +1,2 @@
+(`before after`);
+//      ^ That's a U+2028 LINE SEPARATOR UTF-16 char (between 'before' and 'after')
--- a/packages/babel-parser/test/fixtures/es2019/json-strings/template-line-separator/output.json
+++ b/packages/babel-parser/test/fixtures/es2019/json-strings/template-line-separator/output.json
@@ -1,0 +1,128 @@
+{
+  "type": "File",
+  "start": 0,
+  "end": 101,
+  "loc": {
+    "start": {
+      "line": 1,
+      "column": 0
+    },
+    "end": {
+      "line": 3,
+      "column": 83
+    }
+  },
+  "program": {
+    "type": "Program",
+    "start": 0,
+    "end": 101,
+    "loc": {
+      "start": {
+        "line": 1,
+        "column": 0
+      },
+      "end": {
+        "line": 3,
+        "column": 83
+      }
+    },
+    "sourceType": "script",
+    "interpreter": null,
+    "body": [
+      {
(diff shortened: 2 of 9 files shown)


packages/babel-parser/test/fixtures/es2019/json-strings/template-line-separator/input.js
```diff
@@ -0,0 +1,2 @@
+(`before after`);
```
JLHwung: The behaviour of U+2028 in string template is _not_ changed. Adding test here for coverage.

packages/babel-parser/test/fixtures/es2019/json-strings/template-line-separator/output.json
```diff
@@ -0,0 +1,128 @@
...
+        "start": 0,
+        "end": 17,
+        "loc": {
+          "start": {
+            "line": 1,
+            "column": 0
+          },
+          "end": {
+            "line": 2,
+            "column": 8
```
JLHwung: Keep in mind that the `loc` here should equals to that of `packages/babel-parser/test/fixtures/es2019/json-strings/string-paragraph-separator/output.json`.

PR #10946, opened 2019-12-30
fix: Class Field Initializer should not allow await expression as immediate child

### LINKED ISSUE #6687: Wrong await parsing
> Issue originally reported by @babel-bot in babel/babylon#636

Original issue submitted by @droooney in https://github.com/babel/babel/issues/5966

Babel version: **6.25.0**.
NodeJS and NPM version is irrelevant (the issues can be reproduced at http://babeljs.io/repl/).

1. `const await = foo;` throws a syntax error because it is a reserved word, though it is a reserved word only inside an async function.

2. `await` cannot appear inside function arguments (inside default values or computed keys).

```
async function foo(bar = await baz()) {}
```

or

```
async function foo({ [await baz()]: bar }) {}
```

Here Babel throws an error because `await` is an unexpected reserved word (see 1, apparently Babel thinks that `await` is not inside a function). Though the reason why it's a syntax error is because it's not allowed to use `await` inside arguments (it is a function body actually). That's why the following snippet does not throw any error (though it should).

```
async function foo() {
  async function bar(a = await baz()) {}
}
```

Chrome (as well as NodeJS) throws for this "Illegal await-expression in formal parameters of async function".
Firefox throws "await can't be used in default expression".

3. The same as for 2 goes for class properties:

```
async function a() {
  class Foo {
    bar = await baz();
  }
}
```

is also should be a syntax error because the `bar = await baz()` part is actually inside a constructor which can't be an async function.

Probably, there are more similar cases.

P.S.: actually all of these "work" for `yield` as well.

JLHwung:
| Fixed Issues?            | Fixes #6687
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR fixes the `isAwaitAllowed` logic by excluding `await` identifier reference immediately nested in a class property initializer.

Per https://tc39.es/proposal-class-fields/#sec-new-syntax
```
FieldDefinition[Yield, Await]:
   ClassElementName[?Yield, ?Await] Initializer[In, ~Yield, ~Await]opt
```
the spec dictates that the `Initializer` can _not_ contain `IdentifierReference_Yield` or `IdentifierReference_Await`, which means `await` should be considered as identifier reference here, unless its `[Await]` parameter is enabled by other language structures.

The original snippet in #6687 
```js
async function a() {
  class Foo {
    bar = await baz();
  }
}
```
will still throw because `await` is an identifier instead of await expression, but not because the constructor is _not_ an async function. Think of the following example, which should still throw even the static property does not need a constructor to initialize.
```js
async function a() {
  class Foo {
    static bar = await baz();
  }
}
```

Note that we have a similar issue regarding to the generators:
Parsing the following code should throw because `yield` is an identifier, which turns out to be a reserved word since it resides in a class declaration where all parts are in strict code mode.
```js
function* f() {
  return class { p = yield + 42 }
}
```
I will address this in a separate PR.

### DIFF
--- a/packages/babel-parser/src/util/scope.js
+++ b/packages/babel-parser/src/util/scope.js
@@ -56,7 +56,18 @@ export default class ScopeHandler<IScope: Scope = Scope> {
     return (this.currentVarScope().flags & SCOPE_GENERATOR) > 0;
   }
   get inAsync() {
-    return (this.currentVarScope().flags & SCOPE_ASYNC) > 0;
+    for (let i = this.scopeStack.length - 1; ; i--) {
+      const scope = this.scopeStack[i];
+      const isVarScope = scope.flags & SCOPE_VAR;
+      const isClassScope = scope.flags & SCOPE_CLASS;
+      if (isClassScope && !isVarScope) {
+        // If it meets a class scope before a var scope, it means it is a class property initializer
+        // which does not have an [Await] parameter in its grammar
+        return false;
+      } else if (isVarScope) {
+        return (scope.flags & SCOPE_ASYNC) > 0;
+      }
+    }
   }
   get allowSuper() {
     return (this.currentThisScope().flags & SCOPE_SUPER) > 0;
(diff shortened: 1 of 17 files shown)


JLHwung:
@nicolo-ribaudo Test are added. The computed properties are not affected here because it is never in the class scope. The class scope are applied to the property initializer and the method body.

But I feel like it we really want to be 100% spec compliant, we should introduce `noAwait` and `noYield` as acorn did in `noIn`, and align these context states to the grammar. Checking `noAwait` via scope and `awaitPos` still looks ad-hoc to me.

PR #10947, opened 2019-12-31
Fix: TopLevelAwait should respect await identifiers defined in sub scope.

JLHwung:
| Fixed Issues?            | Babel incorrectly parse AwaitExpression when `topLevelAwait` is enabled

This PR includes commits in #10946 because one of the test case relies on that PR. Only commits descending from b60de33 should be reviewed.

This PR assigns `SCOPE_ASYNC` to the program scope if topLevelAwait is enabled and sourceType is module. By doing so we can reused the `inAsync` check in `isAwaitAllowed`, this can avoid parsing await expression when await is an identifier which is not defined in the top level.

The added test cases are parsed successfully before this PR, which should throw.

### DIFF
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -39,6 +39,7 @@ import {
   SCOPE_DIRECT_SUPER,
   SCOPE_SUPER,
   SCOPE_PROGRAM,
+  SCOPE_ASYNC,
 } from "../util/scopeflags";
 
 export default class ExpressionParser extends LValParser {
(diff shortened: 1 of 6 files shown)


packages/babel-parser/src/parser/expression.js
```diff
@@ -2186,7 +2191,9 @@ export default class ExpressionParser extends LValParser {
   isAwaitAllowed(): boolean {
     if (this.scope.inFunction) return this.scope.inAsync;
     if (this.options.allowAwaitOutsideFunction) return true;
-    if (this.hasPlugin("topLevelAwait")) return this.inModule;
+    if (this.hasPlugin("topLevelAwait")) {
+      return this.inModule && this.scope.inAsync;
```
nicolo-ribaudo: Since at this point `this.scope.inFunction` is false, doesn't `this.inModule` already imply `this.scope.inAsync`?

JLHwung: Nope. For example
```js
class C {
  p = await 0;
}
```
Even top level await is enabled, await cant not be allowed here because it is _not_ allowed in field initializers.

I think the idea of top level await is that the whole module is wrapped in an async function, which is what I do here to assign `SCOPE_ASYNC` to the program scope. Then we will need to handle cases where `await` is _not_ allowed.

PR #10949, opened 2020-01-01
Override toString in case this function is printed

jayenashar:
| Documentation PR Link    | 

I hit an error, but the error message pointed me to this code instead of my code.  This change will put my function in the error message instead of this function.  The same is done in `wrapCheck()` in the same file.  Related to https://stackoverflow.com/questions/59543968/unexpected-return-value-from-visitor-method

### DIFF
--- a/packages/babel-traverse/src/visitors.js
+++ b/packages/babel-traverse/src/visitors.js
@@ -218,15 +218,18 @@ function wrapWithStateOrWrapper(oldVisitor, state, wrapper: ?Function) {
       let newFn = fn;
 
       if (state) {
-        newFn = function(path) {
+        newFn = function (path) {
           return fn.call(state, path, state);
         };
-        // Override toString in case this function is printed, we want to print the wrapped function, same as we do in `wrapCheck`
-        newFn.toString = () => fn.toString();
       }
 
       if (wrapper) {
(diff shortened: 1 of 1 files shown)


packages/babel-traverse/src/visitors.js
```diff
@@ -221,6 +221,8 @@ function wrapWithStateOrWrapper(oldVisitor, state, wrapper: ?Function) {
         newFn = function(path) {
           return fn.call(state, path, state);
         };
+        // Override toString in case this function is printed, we want to print the wrapped function, same as we do in `wrapCheck`
+        newFn.toString = () => fn.toString();
       }
 
       if (wrapper) {
```
JLHwung: Can you override `toString` immediately before returning a `newFn`? Otherwise it will be overridden if `wrapper` is true.

Note that we have the same `toString` hack in `wrapCheck` so I consider it a bug fix.
https://github.com/babel/babel/blob/26eb8918702552c98bbe2dd9f41dd1d6aaeb7a1f/packages/babel-traverse/src/visitors.js#L261

jayenashar: Yes, committed.

I did notice the hack, as mentioned in the comment I added.

Will change issue description to bug fix.

PR #10950, opened 2020-01-01
test: add invalid-lone-import test

JLHwung:

<s>This PR removes an unreachable `unexpected` check, so it is not covered in any unit test:
https://codecov.io/gh/babel/babel/src/b91720c1ccbfc066cbc6af7e3db6114acefac17a/packages/babel-parser/src/parser/expression.js#L929

If the next token is neither `tt.dot` nor `tt.parenL`, it will be parsed in `parseStatementContent`
https://github.com/babel/babel/blob/daaa2063bb8755249d61bd3f5617336983bf4193/packages/babel-parser/src/parser/statement.js#L222-L229</s>

Add a new `lone-import` test for coverage.


nicolo-ribaudo:
While looking at this error, I realized that we can use `this.raise` to provide a better error message: `"import" can only be used as part of dynamic import() or import.meta`.
Also, since the most common error would probably be thinking of `import()` as a function call and thus of `import` as a norma identifier that you can pass around, recovering from this error by using it as a norma expression makes sense.

PR #10956, opened 2020-01-03
Refactor await/yield production parameter tracking

JLHwung:
| Fixed Issues?            | Class Field Initializers should not allow YieldExpression

This PR includes commits from #10947, I will rebase once that PR is merged.

The current parser scoping is meant to track declarations and bindings, it serves well for the purpose of declaration scope. It is also mixed with `[Await]` and `[Yield]` production parameter tracks, historically it was okay because when `[Await]`/`[Yield]` are changed, it must be when a function declaration/expression is parsed, where a declaration scope is also created. 
However, class fields initializers is an exception where a declaration scope is _not_ created but production parameter is changed. Fixing this issue in parser scoping is cumbersome and error-prone: https://github.com/babel/babel/pull/10946/files#diff-a9acee4ca26b68744fe0a1cb5aa03e34R60, therefore I extract the `SCOPE_ASYNC` and `SCOPE_GENERATOR` tracking into a dedicate stack fashioned handler, the code is now more intuitive.

I should point out the idea comes out from an offline discussion with @nicolo-ribaudo and I am really appreciated.

Note that after this PR is merged, ultimately `isAwaitAllowed`
https://github.com/babel/babel/blob/9f832c271647b71a21baf5ba61fd75cb5db6f754/packages/babel-parser/src/parser/expression.js#L2186-L2191
should be simplified to
```js
return (this.param.hasAwait || this.options.allowAwaitOutsideFunction)
```
which is a faithful implementation of the spec text

> BindingIdentifier: await 
> It is a Syntax Error if this production has an [Await] parameter

But we still have trouble on tracking arrow / async arrow expressions. We are try to recover such error in `awaitPos` when parsing arrow expressions. Unless we are parsing the ambiguous `ArrayExpression/ParenthesizedExpression` (like what v8 does in https://docs.google.com/document/d/1FAvEp9EUK-G8kHfDIEo_385Hs2SUBCYbJ5H-NnLvq8M/edit ) and record these errors to be fulfilled or rejected later when it becomes unambiguous, we still relies on the `scope.inFunction` guard.

### DIFF
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -43,7 +43,7 @@ import {
 import { ExpressionErrors } from "./util";
 import {
   PARAM_AWAIT,
-  PARAM_,
+  PARAM,
   functionFlags,
 } from "../util/production-parameter";
 
@@ -110,12 +110,12 @@ export default class ExpressionParser extends LValParser {
 
   // Convenience method to parse an Expression only
   getExpression(): N.Expression {
-    let paramFlags = PARAM_;
+    let paramFlags = PARAM;
     if (this.hasPlugin("topLevelAwait") && this.inModule) {
       paramFlags |= PARAM_AWAIT;
     }
     this.scope.enter(SCOPE_PROGRAM);
-    this.param.enter(paramFlags);
+    this.prodParam.enter(paramFlags);
     this.nextToken();
     const expr = this.parseExpression();
     if (!this.match(tt.eof)) {
@@ -178,7 +178,7 @@ export default class ExpressionParser extends LValParser {
     const startPos = this.state.start;
     const startLoc = this.state.startLoc;
     if (this.isContextual("yield")) {
-      if (this.param.hasYield) {
+      if (this.prodParam.hasYield) {
         let left = this.parseYield(noIn);
         if (afterLeftParse) {
           left = afterLeftParse.call(this, left, startPos, startLoc);
@@ -370,7 +370,7 @@ export default class ExpressionParser extends LValParser {
           if (
             this.match(tt.name) &&
             this.state.value === "await" &&
-            this.param.hasAwait
+            this.prodParam.hasAwait
           ) {
             throw this.raise(
               this.state.start,
@@ -1191,7 +1191,7 @@ export default class ExpressionParser extends LValParser {
     this.next();
     meta = this.createIdentifier(meta, "function");
 
-    if (this.param.hasYield && this.eat(tt.dot)) {
+    if (this.prodParam.hasYield && this.eat(tt.dot)) {
       return this.parseMetaProperty(node, meta, "sent");
     }
     return this.parseFunction(node);
@@ -1868,10 +1868,10 @@ export default class ExpressionParser extends LValParser {
         (inClassScope ? SCOPE_CLASS : 0) |
         (allowDirectSuper ? SCOPE_DIRECT_SUPER : 0),
     );
-    this.param.enter(functionFlags(isAsync, node.generator));
+    this.prodParam.enter(functionFlags(isAsync, node.generator));
     this.parseFunctionParams((node: any), allowModifiers);
     this.parseFunctionBodyAndFinish(node, type, true);
-    this.param.exit();
+    this.prodParam.exit();
     this.scope.exit();
 
     this.state.yieldPos = oldYieldPos;
@@ -1890,7 +1890,7 @@ export default class ExpressionParser extends LValParser {
     trailingCommaPos: ?number,
   ): N.ArrowFunctionExpression {
     this.scope.enter(SCOPE_FUNCTION | SCOPE_ARROW);
-    this.param.enter(functionFlags(isAsync, false));
+    this.prodParam.enter(functionFlags(isAsync, false));
     this.initFunction(node, isAsync);
 
     const oldMaybeInArrowParameters = this.state.maybeInArrowParameters;
@@ -1903,7 +1903,7 @@ export default class ExpressionParser extends LValParser {
     if (params) this.setArrowFunctionParameters(node, params, trailingCommaPos);
     this.parseFunctionBody(node, true);
 
-    this.param.exit();
+    this.prodParam.exit();
     this.scope.exit();
     this.state.maybeInArrowParameters = oldMaybeInArrowParameters;
     this.state.yieldPos = oldYieldPos;
@@ -2177,7 +2177,7 @@ export default class ExpressionParser extends LValParser {
     checkKeywords: boolean,
     isBinding: boolean,
   ): void {
-    if (this.param.hasYield && word === "yield") {
+    if (this.prodParam.hasYield && word === "yield") {
       this.raise(
         startLoc,
         "Can not use 'yield' as identifier inside a generator",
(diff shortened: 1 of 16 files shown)


packages/babel-parser/src/parser/statement.js
```diff
@@ -1573,9 +1576,12 @@ export default class StatementParser extends ExpressionParser {
     node: N.ClassPrivateProperty,
   ): N.ClassPrivateProperty {
     this.scope.enter(SCOPE_CLASS | SCOPE_SUPER);
+    // [In] production parameter is tracked in parseMaybeAssign
+    this.param.enter(PARAM_);
```
JLHwung: I am not sure if I can still understand what I have written at https://github.com/babel/babel/pull/10946/files#diff-a9acee4ca26b68744fe0a1cb5aa03e34R60 three months later.

But this change should be more intuitive as it is copied from the spec:
```
FieldDefinition[Yield, Await]:
  ClassElementName[?Yield, ?Await]  Initializer[In, ~Yield, ~Await]opt
```

packages/babel-parser/src/parser/expression.js
```diff
@@ -97,11 +101,12 @@ export default class ExpressionParser extends LValParser {
...
   getExpression(): N.Expression {
-    let scopeFlags = SCOPE_PROGRAM;
+    let paramFlags = PARAM_;
     if (this.hasPlugin("topLevelAwait") && this.inModule) {
-      scopeFlags |= SCOPE_ASYNC;
+      paramFlags |= PARAM_AWAIT;
     }
-    this.scope.enter(scopeFlags);
+    this.scope.enter(SCOPE_PROGRAM);
+    this.param.enter(paramFlags);
```
nicolo-ribaudo: I don't like this name, since `param` could be anything. What about `this.prodp`, or better `this.prodParam`?

packages/babel-parser/src/parser/expression.js
```diff
@@ -33,14 +33,18 @@ import * as charCodes from "charcodes";
...
   SCOPE_CLASS,
   SCOPE_DIRECT_SUPER,
+  SCOPE_FUNCTION,
   SCOPE_SUPER,
   SCOPE_PROGRAM,
-  SCOPE_ASYNC,
 } from "../util/scopeflags";
+import {
+  PARAM_AWAIT,
+  PARAM_,
```
nicolo-ribaudo: Nit: what about `PARAM_DEFAULT`, or just `PARAM`? The trailing underscore looks... weird.

packages/babel-parser/src/util/scopeflags.js
```diff
@@ -5,8 +5,6 @@
 export const SCOPE_OTHER        = 0b0000000000,
              SCOPE_PROGRAM      = 0b0000000001,
              SCOPE_FUNCTION     = 0b0000000010,
-             SCOPE_ASYNC        = 0b0000000100,
-             SCOPE_GENERATOR    = 0b0000001000,
              SCOPE_ARROW        = 0b0000010000,
```
nicolo-ribaudo: Nit: You can now shift to the left all these other bits.

packages/babel-parser/src/parser/expression.js
JLHwung: Well the spec does have `[Default]` parameter so `PARAM_DEFAULT` will be confusing as we don't check `[Default]` yet.

I have come up with names like `PARAM_NONE`, `PARAM_OTHER` when authoring but I am afraid they maybe used at the spec one day.

I will go with `PARAM` then.

packages/babel-parser/src/parser/expression.js
JLHwung: Well if we don't require `prodParam` to be as easy to speak as `param`, I am good with `prodParam` which is still shorter than the daunting `this.productionParameter`.

JLHwung:
> What do you think about also moving [+Return] to the new class rather than using this.scope.inFunction? It provides way fewer benefits, but at least we are consistent.

Good idea, I have include it in this PR. Not my usual move but the change is minimal.

> I think that in Babel 8 allowAwaitOutsideFunction and allowReturnOutsideFunction should simply set the corresponding production parameter.

Yeah, we are on the same page. I would like to interpret these options as
```
ScriptBody:
  StatementList[~Yield, +Await, +Return]

ModuleItem:
  StatementListItem[~Yield, +Await, +Return]
```

Since the all the TypeScript declaration grammars, including `NamespaceDeclaration` is in the same level as `Statement`. (Link: https://github.com/microsoft/TypeScript/blob/master/doc/spec.md#112-scripts) `allow*OutsideFunction` should not change the production parameter of these nodes, though technically TypeScript Spec does not give any specifications of that. 

By doing so we can throw
```ts
namespace N {
  await 42
}
declare enum E = { P = await 42 };
```
when `allowAwaitOutsideFunction: true`, which should be done in Babel 8.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1158,7 +1170,7 @@ export default class ExpressionParser extends LValParser {
     this.next();
     meta = this.createIdentifier(meta, "function");
 
-    if (this.scope.inGenerator && this.eat(tt.dot)) {
+    if (this.prodParam.hasYield && this.eat(tt.dot)) {
```
KFlash: 2x member access is expensive.

JLHwung: Yeah, but `this.prodParam.hasYield` is still faster than `scope.inGenerator` because we don't have to walkup the scope stacks to find a var scope:

https://github.com/babel/babel/blob/09cb4273bc74e1916db17920f068cad563eefd00/packages/babel-parser/src/util/scope.js#L209

KFlash: What about a mutual parser flag instead?  `state.flag |= HasYield` and then 
`if (state.flag & HasYield) ...``

nicolo-ribaudo: Do you think that it would make any difference? I have always assumed that we have way bigger perf problems in our parser, and that those optimizations would only be noticeable after we resolve things like "Every method access has two walk different levels of the prototype chain" and "Almost every function is megamorphic" and "Our AST objects continuously change their shape".

JLHwung: `this.prodParam.hasYield` is implemented by parser flags. V8 will inline these Get accessors eventually.

The scope stack is used to track nested production params. It is equivalent to do things like
```js
const oldHasYield = this.state.hasYield
this.state.hasYield = true;
parseFunctionBody()
this.state.hasYield = oldHasYield;
```
The only difference is that the `hasYield` above is stored implicitly in stack while `this.prodParams.hasYield` stored explicitly in heap.

KFlash: @nicolo-ribaudo If you see the whole picture, maybe no :) There are too many bottle necks in this parser such as the lexer, tokens itself etc. 

Why not develop a new parser from bottom up? It shouldn't take more than 10 days (estimated) to replicate current Babel parser with working plugins.

nicolo-ribaudo: If you see the even bigger picture, then it would probably be more valuable to re-implement `@babel/traverse` which is where Babel spends most of its time :joy:

KFlash: Just looked at `@babel/traverse`. Should be easy :) Just get rid of the prototype and use functions instead, plus use bitmasks to handle the flow.
The `findParent` function is one of the root causes I guess for why it's slow :)

PR #10961, opened 2020-01-04
fix: optional-chaining should work correctly with ts non-null operator

### LINKED ISSUE #10959: a?.b.c() produces different result than a?.b!.c()
## Bug Report

**Input Code**

```js
a?.b.c()
a?.b!.c()
```

**Actual behavior**
```js
"use strict";

var _a, _a2;

(_a = a) === null || _a === void 0 ? void 0 : _a.b.c();
((_a2 = a) === null || _a2 === void 0 ? void 0 : _a2.b).c();
```

**Expected behavior/code**
The `!` should have no effect on the produced code.

Problem: Suppose `a` is null, then the second line can be reduced to:
```js
   ((_a2 = a) === null || _a2 === void 0 ? void 0 : _a2.b).c()
=> ((a2 = null) === null || _a2 === void 0 ? void 0 : _a2.b).c()
=> (null === null || _a2 === void 0 ? void 0 : _a2.b).c()
=> (true ? void 0 : _a2.b).c()
=> undefined.c()
```

- Filename: `babel.config.js`

```js
{
  "presets": ["@babel/preset-typescript", "@babel/preset-stage-3"]
}
```

**Environment**
- Babel version(s): 7.7.7
- How you are using Babel: https://babeljs.io/repl

**Possible Solution**
¯\\\_(ツ)_/

macabeus:
| Fixed Issues?            | Fixes #10959
| Patch: Bug Fix?          | 👍 Yeah
| Major: Breaking Change?  | 🧐 kind off
| Minor: New Feature?      | 🙅‍♂️ nop
| Any Dependency Changes?  | 🙅‍♂️ nop

This PR fix when use `!` operator within an optional chaining.

```ts
a?.b!.c
``` 

should be compiled to

```js
(_a = a) === null || _a === void 0 ? void 0 : _a.b.c;
```

But previous this fix, it was compiling to

```js
((_a2 = a) === null || _a2 === void 0 ? void 0 : _a2.b).c;
```

That could raise wrongly an exception if `a` is `null` or `undefined`.

The approach to fix this bug is unwrapping `TSNonNullExpression` on the `OptionalMemberExpression`.


packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -34,6 +34,16 @@ export default declare((api, options) => {
             optionalPath.node.type = "CallExpression";
             optionalPath = optionalPath.get("callee");
           }
+
+          // unwrap a TSNonNullExpression if need
+          if (optionalPath.isTSNonNullExpression()) {
+            optionalPath.replaceWith(optionalPath.node.expression);
+
+            if (optionalPath.isOptionalMemberExpression()) {
```
JLHwung: I think we should only skip TSNonNullExpression and leave the `node.expression` as-is. Because the non-null expression does not imply any runtime checks, which is done by the optional member.

macabeus: Oh, yeah. I tested now removing this `if` block and works fine. Okay, I'm updating this PR. Thank you.

packages/babel-plugin-proposal-optional-chaining/test/fixtures/regression/10959/options.json
```diff
@@ -0,0 +1,4 @@
+{
+  "plugins": ["proposal-optional-chaining"],
```
nicolo-ribaudo: Can you also add a test using the syntax plugin instead?

macabeus:
@nicolo-ribaudo could you explain for me these `transforms` and `syntax`, please? What it's and what do. I still didn't understand very well about that.

JLHwung:
@macabeus The `transforms` are the plugins that desugar `OptionalChaining` so that it can run on other platforms, i.e. browsers does not have optional-chaining support. The `syntax` are the plugins that _allow_ the parser to parse `optional-chaining`, so that it will not throw unexpected token.

> Optional chaining transform and typescript syntax only

So this is to say, one wants to desugar `optional-chaining` but keep typescript syntax intact, in this case the `NonNullExpression` should be preserved.

> Optional chaining syntax only and typescript transform

In this case one wants to remove TypeScript stuffs but leave optional-chaining as-is.

macabeus:
I'm agree that we should follow the output in `tsc`, so [I opened an issue in TypeScript repository](https://github.com/microsoft/TypeScript/issues/36031) saying about this problem.

proteriax:
@jridgewell It’s labeled as a bug in the TypeScript repo now. I think we’re good to go.

macabeus:
Wait, we need to add the tests that @nicolo-ribaudo said.
I'll do it tonight.

macabeus:
@nicolo-ribaudo Done. I added the two tests that you requested.
Maybe we could remove `10959-proposal-optional-chaining-and-preset-typescript`, because it's redundant since there is the `10959-proposal-optional-chaining-and-typescript-syntax`... What do you think?

macabeus:
Some random testes failed... but it's sounds like a false negative, because I just added more tests that works fine locally. I didn't changed anymore.

JLHwung:
@macabeus The CI error is not related. I will fix that later.

macabeus:
The fix was just merged on TS: https://github.com/microsoft/TypeScript/pull/36539

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -34,6 +34,11 @@ export default declare((api, options) => {
             optionalPath.node.type = "CallExpression";
             optionalPath = optionalPath.get("callee");
           }
+
+          // unwrap a TSNonNullExpression if need
+          if (optionalPath.isTSNonNullExpression()) {
+            optionalPath.replaceWith(optionalPath.node.expression);
```
jridgewell: I'm not sure that this will correctly fix the bug. We may be able to see it more in test cases like `a?.b!.c.d` and `a?.b.c!.d`

nicolo-ribaudo:
This is technically a breaking change, because it drops support for TypeScript 3.8. However, without merging this PR we can't support TypeScript 3.9.

I don't think we should wait for Babel 8: does anyone thinks that it would be better to merge this in a minor rather than in the next patch?

macabeus:
@nicolo-ribaudo Well... TS already merged that on a minor release: `3.8` -> `3.9`
Maybe we could follow the same approach?

existentialism:
> This is technically a breaking change, because it drops support for TypeScript 3.8. However, without merging this PR we can't support TypeScript 3.9.
> 
> I don't think we should wait for Babel 8: does anyone thinks that it would be better to merge this in a minor rather than in the next patch?

I'm good with a minor.

PR #10962, opened 2020-01-05
added basic support for module attributes and tests updated

vivek12345:
| New proposal?            | https://github.com/tc39/proposal-module-attributes
| Patch: Bug Fix?          | Feature
| Documentation PR Link    | 
| Any Dependency Changes?  | 

This change is to let the babel parser parse the custom module attributes as discussed in the proposal above.
Since it is a stage-1  proposal we will definitely have a lot more changes coming up but this is an early draft of what the implementation would look like.

Now we can specify module attributes in the following way
```
import foo from "foo.json" with type: "json";
```
* We will be parsing the `with` keyword and will then parse any custom key value pairs which are passed after the `with` attribute.
* In the AST node, we will set node.attributes = [array of these key value pairs]
* Also we will only be parsing this if the `plugin: ["moduleAttributes"]` is enabled. otherwise we will raise an error.

We still need to discuss
* If the attributes do not specify a `type` attribute, then should we throw an error?
* If multiple `type` attributes are present then should we raise an error, as only one type should be allowed?


nicolo-ribaudo:
FYI @littledan @xtuc

packages/babel-parser/src/parser/statement.js
```diff
@@ -2042,6 +2055,15 @@ export default class StatementParser extends ExpressionParser {
     node.specifiers.push(this.finishNode(specifier, type));
   }
 
+  maybeParseModuleAttributes() {
+    if (this.match(tt._with) && !this.hasPlugin("moduleAttributes")) {
+      this.expectPlugin("moduleAttributes");
+    } else if (this.eat(tt._with)) {
+      return this.parseKeyValuePairs();
+    }
+    return [];
```
nicolo-ribaudo: I think that we should do something like this:
1. If there is `with`,
  i. Expect the plugin
  ii. return `parseKeyValuePairs`
2. If there is the plugin
  i. Return `[]`

By doing so, we don't return an empty array if the plugin is disabled (since disabled plugins shouldn't have any effect on the AST).
Even better, we should only set `node.attributes` if `maybeParseModuleAttributes` returns the array.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1460,6 +1460,26 @@ export default class ExpressionParser extends LValParser {
     return this.finishNode(node, "TemplateLiteral");
   }
 
+  parseKeyValuePairs() {
+    const propHash: any = Object.create(null);
+    const node = this.startNode();
+    node.properties = [];
+
+    do {
+      const prop = this.parseObjectMember();
```
JLHwung: It maybe overkill to use `parseObjectMember` here because it is _likely_ to include only primitive types (see https://github.com/tc39/proposal-module-attributes/issues/22#issuecomment-558754321), i.e.
```js
import foo from "foo" with loaders: ["bar", "qux"]
```
are not allowed because array is not a primitive type. As a stage-1 proposal, I suggest we implement the minimal superset of proposal. Should the proposal accepts more types, we can always drops restrictions later.

packages/babel-parser/src/parser/statement.js
vivek12345: Understood. That makes sense. I will make the changes accordingly.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1460,6 +1460,26 @@ export default class ExpressionParser extends LValParser {
...
+      if (prop.shorthand) {
+        this.addExtra(prop, "shorthand", true);
+      }
+
+      node.properties.push(prop);
+    } while (this.eat(tt.comma));
+    if (!this.match(tt.eq) && propHash.start !== undefined) {
+      this.raise(propHash.start, "Redefinition of __proto__ property");
+    }
+    return this.finishNode(node, "ObjectExpression");
```
nicolo-ribaudo: I don't think that module attributes allow arbitrary values (I might be wrong).

I think that this should be simplified to (pseudo-JavaScript)
```js
const attrs = [];

do {
  const node = startNode();
  parseIdentifier();
  expect(":");
  parseLiteral();
} while (eat(","));

return attrs; // Return the array of attributes, not a wrapper node.
```

It could also be inlined in `maybeParseModuleAttributes`, since it's short.

packages/babel-parser/src/parser/expression.js
vivek12345: Understood. I will write a simple parseKeyValue function. Thank you 👍

packages/babel-parser/src/parser/expression.js
vivek12345: This looks way better and simpler. I will make the changes. Thank you for the inputs 👍

vivek12345: few questions here, should the above implementation look something like this
```js
const attrs = [];
do {
  node.key = this.parseIdentifier();
  node.computed = false;
  node.method = false;
  node.shorthand = false;
  this.expect(tt.colon);
  node.value = this.parseLiteral(this.state.value);
  this.finishNode(node, "ObjectProperty");
  attrs.push(node);} while (this.eat(tt.comma));
```

vivek12345: or should we just do
```js
node.value = this.parseLiteral(this.state.value, "Literal");
```

JLHwung: The second `type` argument is meant to set type of the AST nodes. Babel supports the following `types` (but not general `"Literal"`)
```
RegExpLiteral
NullLiteral
StringLiteral
BooleanLiteral
NumericLiteral
BigIntLiteral
```
A primitive value is a member of one of the following built-in types: Undefined, Null, Boolean, Number, String, and Symbol. So you can guard against `parseExprAtom` with (pseudoJS)
```
this.match(tt._null) || ... tt._true, tt._false, tt.name (and value is "undefined"), tt.string, tt.num, tt.bigint
```

Note that `Symbol` is also a primitive type and not covered above, but I think we can forbid symbol in module attributes? @xtuc

packages/babel-parser/src/parser/statement.js
vivek12345: updated the code with these changes.

packages/babel-parser/src/parser/expression.js
vivek12345: Refactored with the suggested changes

packages/babel-parser/src/parser/expression.js
vivek12345: These changed have been updated. Thank you

packages/babel-parser/src/parser/expression.js
```diff
@@ -852,6 +852,35 @@ export default class ExpressionParser extends LValParser {
...
+      case tt._null:
+        node = this.startNode();
+        this.next();
+        return this.finishNode(node, "NullLiteral");
+
+      case tt._true:
+      case tt._false:
+        return this.parseBooleanLiteral();
+      default:
+        throw this.unexpected();
```
nicolo-ribaudo:
```suggestion
        throw this.raise(this.state.start, "Only primitive literals are allowed in this position.");
```

packages/babel-parser/src/parser/statement.js
```diff
@@ -2005,13 +2005,29 @@ export default class StatementParser extends ExpressionParser {
...
+      // now we check that if we need to parse the next imports
+      // but only if he/she is not importing * which means all
       if (parseNext && !hasStar) this.parseNamedImportSpecifiers(node);
       this.expectContextual("from");
     }
     node.source = this.parseImportSource();
+    const attributes = this.maybeParseModuleAttributes();
+    if (Array.isArray(attributes)) {
+      node.attributes = attributes;
+    }
```
nicolo-ribaudo: As a micro-optimization (I know that this logic was suggested by me :stuck_out_tongue:) we could do
```js
if (attributes) node.attributes = attributes;
```
which looks more like "if attributes are present, then set them" and avoids needing to know that it is an array when skimming throughout this function.

packages/babel-parser/src/parser/statement.js
```diff
@@ -2042,6 +2058,27 @@ export default class StatementParser extends ExpressionParser {
...
+    const attrs = [];
+    do {
+      const node = this.startNode();
+      node.key = this.parseIdentifier();
+      node.computed = false;
+      node.method = false;
+      node.shorthand = false;
+      this.expect(tt.colon);
+      node.value = this.parsePrimitiveValue();
+      this.finishNode(node, "ObjectProperty");
```
nicolo-ribaudo: I don't think that we should re-use `ObjectProperty`: it's conceptually and syntactically different. We should introduce a new `ModuleAttribute` node.

packages/babel-parser/src/parser/expression.js
littledan: I'd recommend starting with just supporting string literals at this point.

JLHwung:
@littledan Sure. We are really good at not merging PRs.

packages/babel-parser/src/parser/statement.js
```diff
@@ -2042,6 +2058,27 @@ export default class StatementParser extends ExpressionParser {
...
+    this.expectPlugin("moduleAttributes");
+    const attrs = [];
+    do {
+      const node = this.startNode();
+      node.key = this.parseIdentifier();
+      node.computed = false;
+      node.method = false;
+      node.shorthand = false;
+      this.expect(tt.colon);
+      node.value = this.parsePrimitiveValue();
```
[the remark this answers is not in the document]
vivek12345: Noted. Thank you @littledan 👍

packages/babel-parser/src/parser/expression.js
xtuc: I like the suggestion, but the text isn't very human friendly IMO.

What about: "Only a primitive literal is allowed here", assuming it will show the code frame?

packages/babel-parser/src/parser/statement.js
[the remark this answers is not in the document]
xtuc: I agree with @littledan 

That said, we didn't define the grammar yet and we might want to consider things like:

```
import u from "a" with lazy: true;
```

instead of a string literal?

packages/babel-parser/src/parser/statement.js
xtuc: Shouldn't the module attribute be a map instead of an array? I would prefer rejecting duplicated attributes to avoid confusion.

xtuc:
Nice work @vivek12345 and thanks for the help @nicolo-ribaudo!

packages/babel-parser/src/parser/statement.js
```diff
@@ -2005,13 +2005,29 @@ export default class StatementParser extends ExpressionParser {
...
+       * but if we do have a default import
+       * we need to check if we have a comma after that and
+       * that is where this || this.eat comes into play
+       */
       const parseNext = !hasDefault || this.eat(tt.comma);
+      // if we do have to parse the next set of specifiers, we first check for start imports
+      // import React, * from "react";
       const hasStar = parseNext && this.maybeParseStarImportSpecifier(node);
+      // now we check that if we need to parse the next imports
+      // but only if he/she is not importing * which means all
```
xtuc:
```suggestion
      // but only if they are not importing * which means all
```

packages/babel-parser/src/parser/statement.js
```diff
@@ -2005,13 +2005,29 @@ export default class StatementParser extends ExpressionParser {
...
       if (parseNext && !hasStar) this.parseNamedImportSpecifiers(node);
       this.expectContextual("from");
     }
     node.source = this.parseImportSource();
+    const attributes = this.maybeParseModuleAttributes();
```
xtuc: nit: the documentation above is nice, would you mind adding more details where?

> parse module attributes if the next token is `with` or ignore and finish the ImportDeclaration node.

packages/babel-parser/src/parser/statement.js
xtuc: Would it help if we agreed on the Babel AST before the implemation?

packages/babel-parser/src/parser/statement.js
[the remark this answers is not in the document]
littledan: Yes, we might want to consider it. But I don't think this is worth complicating the Babel codebase at the moment.

packages/babel-parser/src/parser/expression.js
```diff
@@ -852,6 +852,35 @@ export default class ExpressionParser extends LValParser {
...
+
+      case tt._null:
+        node = this.startNode();
+        this.next();
+        return this.finishNode(node, "NullLiteral");
+
+      case tt._true:
+      case tt._false:
+        return this.parseBooleanLiteral();
+      default:
```
jridgewell: Nit: newline
```suggestion

      default:
```

packages/babel-parser/src/parser/expression.js
```diff
@@ -1459,7 +1488,6 @@ export default class ExpressionParser extends LValParser {
     this.next();
     return this.finishNode(node, "TemplateLiteral");
   }
-
```
jridgewell: Nit: undo

packages/babel-parser/src/parser/statement.js
jridgewell: > Shouldn't the module attribute be a map instead of an array?

Since the syntax isn't `{ foo: bar }`, I'm not sure that I'd expect a map.

nicolo-ribaudo: The key of a module attribute must be it's own AST node, which can have comments and location info.
We are also using an array for imported bindings (i.e. `{ x as y, x as w }`).

xtuc: Ok, that's fine to me too

packages/babel-parser/src/parser/expression.js
xtuc: We aren't sure if we want to allow any value there, let's keep it simple for now I think; only stringliterals

xtuc:
I submited https://github.com/babel/babel/pull/11015 that sketches the AST for Module Attributes

nicolo-ribaudo:
@vivek12345 The AST specification PR (https://github.com/babel/babel/pull/11015) has been merged.

vivek12345:
Understood. I will start making the changes that have been recommended in the reviews along with the AST specification for module attributes.

xtuc:
To follow up here. We are planning to restrict the module attributes to StringLiteral only, for now.

Is there anything I can do to help @vivek12345?

packages/babel-parser/src/parser/statement.js
vivek12345: added this information. Thank you 👍

packages/babel-parser/src/parser/expression.js
vivek12345: Made the changes and now we are only allowing string literals.

packages/babel-parser/src/parser/expression.js
vivek12345: Made the changes and now we are only supporting String Literals.

packages/babel-parser/src/parser/expression.js
vivek12345: The code has changed now and we no longer need this function. So marking this as resolved.

packages/babel-parser/src/parser/statement.js
vivek12345: Changed the code to now use the new AST node `ImportAttribute`

packages/babel-parser/src/parser/statement.js
vivek12345: made the changes. Thank you 👍

packages/babel-parser/src/parser/expression.js
vivek12345: Have removed this function as we are only allowing String Literals. So marking this as resolved. Thank you 👍

vivek12345:
> Can you also add the following tests, to check ASI?
> 
> ```js
> import "x" with type: "json"
> [0]
> ```
> 
> ```js
> import "x"
> with
> type: "json"
> ```
> 
> This one should throw either about `with` in strict mode, or about the `(` where we expect an identifier.
> 
> ```js
> import "x"
> 
> with ({});
> ```
> 
> PS: It looks like you commits are not linked to your account. You can run
> 
> ```
> git config user.email [your email used for github]
> ```
> 
> or
> 
> ```
> git config --global user.email [your email used for github]
> ```

Added all these tests to the PR.

packages/babel-parser/src/parser/statement.js
```diff
@@ -2068,6 +2087,33 @@ export default class StatementParser extends ExpressionParser {
     node.specifiers.push(this.finishNode(specifier, type));
   }
 
+  maybeParseModuleAttributes() {
+    if (!this.eat(tt._with)) {
+      if (this.hasPlugin("moduleAttributes")) return [];
```
xtuc: Do you think it's worth having two plugins? One for `as` and the other for the attributes

nicolo-ribaudo: We should have a single plugin with an option, similarly to the pipeline operator.

vivek12345: @nicolo-ribaudo can you help me understand which file to look at to understand this option to allow different syntax with the same plugin?

nicolo-ribaudo: You can search for `getPluginOption(` in `expression.js`.

vivek12345: Understood 👍

xtuc: Sure, that works too. We just need to make sure that both can be enabled at the same time

nicolo-ribaudo: Why? I don't think that alternative versions of the same proposal should be mixed.

xtuc: We don't see them as different versions of the same proposal. Imagine `as "json"` being a shorthand for `with type:"json"`. We are going to present that idea to the committee, but eventually they will live together. Does that make sense to you?

nicolo-ribaudo: Oh ok :+1:

Then we can have two boolean option, `as` and `with`.

packages/babel-parser/src/parser/statement.js
```diff
@@ -2068,6 +2087,33 @@ export default class StatementParser extends ExpressionParser {
...
+      return null;
+    }
+    this.expectPlugin("moduleAttributes");
+    const attrs = [];
+    do {
+      // we are trying to parse a node which has the following syntax
+      // with type: "json"
+      // [with -> keyword], [type -> Identifier], [":" -> token for colon], ["json" -> StringLiteral]
+      const node = this.startNode();
+      node.key = this.parseIdentifier();
```
JLHwung:
```suggestion
      node.key = this.parseIdentifier(true);
```
To align with Property Name, the identifier name should be literal, that is,
```js
import foo from "bar" with for: "test"
```
should be valid.

nicolo-ribaudo:
NOTE: Given some pushback from the committee regarding the `as` syntax, we are only going to implement the `with type: "foo"` syntax.

packages/babel-parser/src/parser/statement.js
```diff
@@ -2042,6 +2061,52 @@ export default class StatementParser extends ExpressionParser {
...
+      }
+      // set the attribute name entry in the attributeMap to true
+      attributeMap[node.key.name] = true;
+      // check for colon
+      this.expect(tt.colon);
+      // check if the value set to the module attribute is a string as we only allow string literals
+      if (!this.match(tt.string)) {
+        return this.unexpected(
+          this.state.start,
+          "Only string literal values are allowed",
```
nicolo-ribaudo: Please move this to the `Errors` object.

Also, let's give context to the error message:
```suggestion
          "Only string literals are allowed as module attributes values",
```

packages/babel-parser/src/parser/location.js
```diff
@@ -98,6 +98,9 @@ export const Errors = Object.freeze({
   MissingUnicodeEscape: "Expecting Unicode escape sequence \\uXXXX",
   MixingCoalesceWithLogical:
     "Nullish coalescing operator(??) requires parens when mixing with logical operators",
+  ModuleAttributesWithDuplicateKeys:
+    "Module attributes should not include duplicate keys",
+  ModuleAttributesWithoutType: "Module attributes should include a type",
```
nicolo-ribaudo: Instead of throwing if there isn't `type`, can we throw if there is anything other than `type`?

```suggestion
  ModuleAttributeDifferentFromType: "The only accepted module attribute is `type`",
```

packages/babel-parser/src/parser/statement.js
```diff
@@ -2042,6 +2061,52 @@ export default class StatementParser extends ExpressionParser {
     node.specifiers.push(this.finishNode(specifier, type));
   }
 
+  maybeParseModuleAttributes() {
+    if (!this.eat(tt._with)) {
+      if (this.hasPlugin("moduleAttributes")) return [];
+      return null;
+    }
+    this.expectPlugin("moduleAttributes");
```
nicolo-ribaudo: Nit: can we do something like this?

```suggestion
    if (this.matches(tt._with)) {
      this.expectPlugin("moduleAttributes");
      this.next(); // Consume the `with` keyword
    } else {
      if (this.hasPlugin("moduleAttributes")) return [];
      return null;
    }
```

By doing so, we throw the "missing plugin" error pointing at `with` rather than at what comes after.

packages/babel-parser/src/parser/statement.js
```diff
@@ -2042,6 +2061,52 @@ export default class StatementParser extends ExpressionParser {
...
+    const attrs = [];
+    const attributeMap = {};
```
nicolo-ribaudo: This is not safe: what if there is an attribute named, for example, `hasOwnProperty`? In cases like this, we should use either `Object.create(null)` or `new Map()`.

nicolo-ribaudo: Also, maybe add a test where an attribute is called `hasOwnProperty`

packages/babel-parser/src/parser/statement.js
```diff
@@ -2042,6 +2061,52 @@ export default class StatementParser extends ExpressionParser {
...
+      // we are trying to parse a node which has the following syntax
+      // with type: "json"
+      // [with -> keyword], [type -> Identifier], [":" -> token for colon], ["json" -> StringLiteral]
+      const node = this.startNode();
+      node.key = this.parseIdentifier(true);
+
+      // check if we already have an entry for an attribute
+      // if a duplicate entry found, throw an error
+      if (attributeMap[node.key.name]) {
+        throw this.raise(
```
nicolo-ribaudo: This error is easily recoverable: you can remove `throw`.

packages/babel-parser/src/parser/statement.js
```diff
@@ -2042,6 +2061,52 @@ export default class StatementParser extends ExpressionParser {
...
+          Errors.ModuleAttributesWithDuplicateKeys,
+        );
+      }
+      // set the attribute name entry in the attributeMap to true
+      attributeMap[node.key.name] = true;
+      // check for colon
+      this.expect(tt.colon);
+      // check if the value set to the module attribute is a string as we only allow string literals
+      if (!this.match(tt.string)) {
+        return this.unexpected(
```
```suggestion
        throw this.unexpected(
```

packages/babel-parser/src/parser/statement.js
```diff
@@ -2042,6 +2061,52 @@ export default class StatementParser extends ExpressionParser {
...
+      const node = this.startNode();
+      node.key = this.parseIdentifier(true);
+
+      // check if we already have an entry for an attribute
+      // if a duplicate entry found, throw an error
+      if (attributeMap[node.key.name]) {
+        throw this.raise(
+          this.state.start,
+          Errors.ModuleAttributesWithDuplicateKeys,
+        );
```
nicolo-ribaudo: If you modify the error message like this:

```
Module attributes should not include duplicate keys ("%1")
```

then you can pass `node.key.name` as an additional parameter to `this.raise` and it will be included in the error message.

packages/babel-parser/test/fixtures/experimental/import-module-attributes/invalid-syntax-with-no-attributes-identifier/options.json
```diff
@@ -0,0 +1,8 @@
+{
+  "plugins": [
+    "moduleAttributes",
+    "estree"
+  ],
+  "sourceType": "module",
+  "throws": "Unexpected token (2:5)"
```
nicolo-ribaudo: Note for reviewers: this is because ASI doesn't apply, since `with` statements are disallowed in modules.

packages/babel-parser/src/parser/location.js
vivek12345: But we do allow other attributes too right? 
Like you can specify 
```
import file from "file.json". with loading: true, hasData: false
```
In this case the error should be `Module attributes should include a type`

Or are we restricting any other attribute other than `type` for now
@nicolo-ribaudo

packages/babel-parser/src/parser/statement.js
vivek12345: This makes sense. I will update the code 👍

packages/babel-parser/src/parser/statement.js
vivek12345: Noted 👍

packages/babel-parser/src/parser/statement.js
vivek12345: Added this test as well 👍

packages/babel-parser/src/parser/location.js
nicolo-ribaudo: Ok, I thought that we discussed this but I can't even find the discussion where we decided to throw if `type` isn't provided :thinking:

cc @xtuc What should we allow for now?

vivek12345: @nicolo-ribaudo I think that was a very initial discussion that we had when I had just started working on this PR.
I can revert the changes if the `type` key is optional as well after @xtuc confirms

packages/babel-parser/test/fixtures/experimental/import-module-attributes/invalid-syntax-with-no-attributes-identifier/options.json
JLHwung: The unexpected token error comes from `(` is not an ID_Start character. If it was parsed as a `with` statement, it should throw `Errors.StrictWith`.

packages/babel-parser/test/fixtures/experimental/import-module-attributes/invalid-syntax-with-no-attributes-identifier/input.js
```diff
@@ -0,0 +1,2 @@
+import "x"
+with ({});
```
JLHwung: @xtuc Although `with` statement is forbidden in modules, do we allow LineTerminator before `WithClause`? I notice that`AsClause` does forbid leading line terminators: https://tc39.es/proposal-module-attributes/#prod-AsClause

The current implementation allow the following to be parsed.
```js
import "x"
with type: "jpeg"
```

JLHwung:
nit: @vivek12345 Can you rename the test folder by `module-attributes`? It's shorter and also the slash name of the tc39 `proposal-module-attributes`.

packages/babel-parser/src/parser/statement.js
```diff
@@ -2042,6 +2061,55 @@ export default class StatementParser extends ExpressionParser {
...
+  maybeParseModuleAttributes() {
+    if (this.match(tt._with)) {
+      this.expectPlugin("moduleAttributes");
+      this.next();
+    } else {
+      if (this.hasPlugin("moduleAttributes")) return [];
+      return null;
+    }
+    const attrs = [];
+    const attributeMap = Object.create(null);
```
JLHwung: I think it suffices to model `attributeMap` as a `Set<string>`. The value of `attributeMap` is redundant.

packages/babel-parser/src/parser/location.js
```diff
@@ -98,6 +98,11 @@ export const Errors = Object.freeze({
   MissingUnicodeEscape: "Expecting Unicode escape sequence \\uXXXX",
   MixingCoalesceWithLogical:
     "Nullish coalescing operator(??) requires parens when mixing with logical operators",
+  ModuleAttributesIncorrectValue:
```
JLHwung: suggestion: the `ModuleAttributeInvalidValue` seems more consistent to me with other naming conventions.

packages/babel-parser/src/parser/location.js
JLHwung:
```suggestion
  ModuleAttributesWithoutType: "Module attributes should include a \"type\" key",
```
`Type` is a bit general and it may confuse TypeScript users.

packages/babel-parser/src/parser/location.js
```diff
@@ -98,6 +98,11 @@ export const Errors = Object.freeze({
...
+    "Only string literals are allowed as module attributes values",
+  ModuleAttributesWithDuplicateKeys:
+    "Module attributes should not include duplicate keys (%0)",
```
JLHwung:
```suggestion
    "Duplicate key \"%0\" is not allowed in module attributes",
```

vivek12345: This definitely is better. Made the changes 👍

packages/babel-parser/src/parser/location.js
vivek12345: Updated the code with `ModuleAttributeInvalidValue`

packages/babel-parser/src/parser/statement.js
vivek12345: Set is great. Thank you 👍 Have replaces `Object.create` with `Set` now.

packages/babel-parser/src/plugin-utils.js
```diff
@@ -84,6 +84,20 @@ export function validatePlugins(plugins: PluginList) {
...
+    const moduleAttributesVerionPluginOption = getPluginOption(
+      plugins,
+      "moduleAttributes",
+      "version",
+    );
+    if (moduleAttributesVerionPluginOption !== "feb-2020") {
+      throw new Error(
+        "The 'moduleAttributes' plugin requires a 'version' option," +
+          " whose value must be feb-2020.",
+      );
```
nicolo-ribaudo:
```suggestion
      throw new Error(
        "The 'moduleAttributes' plugin requires a 'version' option," +
          " representing the last proposal update. Currently, the" +
          " only supported value is 'feb-2020'.",
      );
```

packages/babel-parser/src/parser/statement.js
```diff
@@ -2042,6 +2061,55 @@ export default class StatementParser extends ExpressionParser {
     node.specifiers.push(this.finishNode(specifier, type));
   }
 
+  maybeParseModuleAttributes() {
+    if (this.match(tt._with)) {
```
nicolo-ribaudo: To implement the `[no LineTerminator]` here, you can add this:
```suggestion
    if (this.match(tt._with) && !this.isLineTerminator()) {
```

packages/babel-parser/src/parser/statement.js
```diff
@@ -2042,6 +2061,55 @@ export default class StatementParser extends ExpressionParser {
...
+        this.raise(
+          this.state.start,
+          Errors.ModuleAttributesWithDuplicateKeys,
+          node.key.name,
+        );
+      }
+      // set the attribute name entry in the attributeMap
+      attributes.add(node.key.name);
+      // check for colon
+      this.expect(tt.colon);
```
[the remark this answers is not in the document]
vivek12345: Makes sense 👍 Will remove them

packages/babel-parser/src/parser/statement.js
vivek12345: Do we need to add this?

packages/babel-parser/src/parser/statement.js
```diff
@@ -2042,6 +2061,55 @@ export default class StatementParser extends ExpressionParser {
...
+      // with type: "json"
+      // [with -> keyword], [type -> Identifier], [":" -> token for colon], ["json" -> StringLiteral]
+      const node = this.startNode();
+      node.key = this.parseIdentifier(true);
+
+      // check if we already have an entry for an attribute
+      // if a duplicate entry found, throw an error
+      if (attributes.has(node.key.name)) {
+        this.raise(
+          this.state.start,
```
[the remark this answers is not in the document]
vivek12345: done. Added 👍

packages/babel-parser/src/parser/statement.js
nicolo-ribaudo: Yes, as mentioned at https://github.com/babel/babel/pull/10962#discussion_r391979487

vivek12345: @nicolo-ribaudo Added this check as well and the tests are also updated. Can you review them once and see if that is the error that we expect?

packages/babel-parser/src/parser/location.js
```diff
@@ -98,6 +98,11 @@ export const Errors = Object.freeze({
   MissingUnicodeEscape: "Expecting Unicode escape sequence \\uXXXX",
   MixingCoalesceWithLogical:
     "Nullish coalescing operator(??) requires parens when mixing with logical operators",
+  ModuleAttributeInvalidValue:
+    "Only string literals are allowed as module attributes values",
+  ModuleAttributesWithDuplicateKeys:
+    'Duplicate key "%0" is not allowed in module attributes',
+  ModuleAttributesWithoutType: 'Module attributes should include a "type" key',
```
nicolo-ribaudo: @littledan @xtuc We decided to add this error in the early stages of this PR. Is it still needed?

packages/babel-parser/src/parser/location.js
```diff
@@ -98,6 +98,11 @@ export const Errors = Object.freeze({
   MissingUnicodeEscape: "Expecting Unicode escape sequence \\uXXXX",
   MixingCoalesceWithLogical:
     "Nullish coalescing operator(??) requires parens when mixing with logical operators",
+  ModuleAttributeInvalidValue:
+    "Only string literals are allowed as module attributes values",
```
xtuc:
```suggestion
    "Only string literals are allowed as module attribute values",
```

packages/babel-parser/src/parser/location.js
xtuc: To echo want I said on Slack. I think we should restrict the module attribute to the `type` only for now. It's easy enough to lift after and will avoid plugin author to already create custom attributes

packages/babel-parser/src/parser/statement.js
vivek12345: @nicolo-ribaudo I am not sure why it's showing this conversation here. But to re-confirm should we remove throw here as well?

In this condition we are checking if attribute is anything other than type

packages/babel-parser/src/parser/location.js
vivek12345: Thank you @xtuc . That seems correct to me and was a mistake at my end.

packages/babel-parser/src/parser/location.js
vivek12345: Updated the code to only allow as `type` as module attribute. The tests have now been updated to throw an error if anything is entered other than `type`.

packages/babel-parser/src/parser/statement.js
nicolo-ribaudo: Yes, let's remove it. `throw` is only needed when it would be impossible to continue paring after the error.

vivek12345:
As per recent discussions with @xtuc and @nicolo-ribaudo , we are only going to allow `type` as the module attribute for now.

This is the reason why the code is now updated to support that condition.
The tests have been kept as is - with the only change that for now they will all throw an error if any other attribute other than `type` is being used.

Later on when this condition is relaxed, we can update those tests too.

packages/babel-parser/src/parser/statement.js
```diff
@@ -2115,6 +2115,7 @@ export default class StatementParser extends ExpressionParser {
     if (!attributes.has("type")) {
       throw this.raise(this.state.start, Errors.ModuleAttributesWithoutType);
```
```suggestion
      this.raise(this.state.start, Errors.ModuleAttributesWithoutType);
```

nicolo-ribaudo: Oh actually I think that we can remove this now, since `import "foo" with x: "y";` will already throw because `x` is not a valid attribute, and `import "foo" with;` isn't valid anyway.
Even without this error, the only thing that doesn't throw is `import "foo" with type: "xx"`.

vivek12345: Understood. I have now removed this condition.

packages/babel-plugin-syntax-import-module-attributes/README.md
```diff
@@ -1,6 +1,6 @@
 # @babel/plugin-syntax-import-module-attributes
```
nicolo-ribaudo:
```suggestion
# @babel/plugin-syntax-module-attributes
```

Since the proposal is just called "module attributes"

vivek12345: Made the changes 👍

packages/babel-parser/src/parser/statement.js
JLHwung: nit: It suffices to check `!this.hasPrecedingLineBreak()` because it is guarded by `this.match(tt._with)`.

packages/babel-parser/test/fixtures/experimental/module-attributes/invalid-syntax-with-no-attributes-identifier/input.js
```diff
@@ -0,0 +1,2 @@
+import "x"
+with ({});
```
JLHwung: This test case is not testing `import-with-no-attributes-identifier` but `withClause: [noLineTerminator] with type: StringLiteral`. Can you rename the test to `import-with-statement`?

We can add a new test for `invalid-syntax-with-no-attributes-identifier`
```js
import "x" with;
```

packages/babel-parser/test/fixtures/experimental/module-attributes/invalid-syntax-with-attributes-multiple-lines/input.js
```diff
@@ -0,0 +1,3 @@
+import "x"
+with
+type: "json"
```
JLHwung: This error is thrown because invalid with statement parsing but not invalid with attributes.
I don't think currently we need testing invalid multiple line attributes because only `"type"` is allowed, but it is worthy to test ASI (it is currently valid).
```js
import "x" with
type: "json"
```

littledan:
I'm happy to see the approvals, but note that this doesn't match the current state of the repo (which is based on `as "json"`, but this was not taken very well by the committee). Let's wait a bit before landing this until that can be settled down--tbh I'm suspecting we should stick with what this patch does, but maybe with errors also for unknown attributes.

nicolo-ribaudo:
Currently, this PR reports an error when an attribute other than `type` is used.
We plan to release 7.9.0 today or tomorrow. Should we defer this PR to the next minor version?

littledan:
That sounds good. The only remaining possibility is that we might want to maintain a whitelist of possible `type` values; Babel might want to enforce this. I'd prefer if this waited until the next minor release, unless you're OK with breaking compatibility for people who opt into the early version.

xtuc:
We changed the proposal syntax and form: https://github.com/tc39/proposal-module-attributes. I think we can move forward with this PR

JLHwung:
@vivek12345 Hey how are you doing? Recently TC39 has updated the spec so Nicolò has changed the version from `feb-2020` to `apr-2020`. I realize that the new syntax
```
import(a, b)
```
is not yet implemented in this PR. Can you add a test case and work on that? Under the hood the dynamic `import()` is parsed as a `CallExpression` and we may have additional checks on the number of arguments. We should change that from 1 to 2.

vivek12345:
I am doing well @JLHwung . Thank you for asking. I hope you are doing well too. I will start looking into this today in the evening.

vivek12345:
I have a question around this dynamic imports and the second argument
1) What node type should we consider them. In a normal import we consider them as ImportAttribute. But here since we are parsing these as call expression and it is an argument for this call expression, should we consider the second expression as an object expression?

2) The argument check here 
https://github.com/babel/babel/blob/master/packages/babel-parser/src/parser/expression.js#L768
Should we change that to 
```
if (node.arguments.length < 1) {
        this.raise(node.start, Errors.ImportCallArity);
```

nicolo-ribaudo:
1. No, the second argument can be anything. Consider this example:
    ```js
    const opts = { with: { type: "json" } };
    import("foo.js", opts);
    ```
2. Having more than two arguments is still invalid. We could change it to
   ```js
   const maxArity = this.hasPlugin("moduleAttributes") ? 2 : 1;
   if (node.arguments.length === 0 || node.arguments.length > maxArity) {
        this.raise(node.start, Errors.ImportCallArity);
   ```

littledan:
Note that the proposal also permits trailing commas in dynamic import.

vivek12345:
Thank you. In that case do we parse the second argument as we were parsing module attributes and throw an error if anything other than type is used. Also the value of the attribute should not be anything other than a string? Do the same rules apply here as well?

nicolo-ribaudo:
No, the second argument could be anything: it's enough to relax the restriction on `node.arguments.length`. This is because it's not possible to statically analyze dynamic imports.

packages/babel-parser/src/parser/expression.js
```diff
@@ -758,8 +758,13 @@ export default class ExpressionParser extends LValParser {
...
   ): N.Expression {
     if (node.callee.type === "Import") {
-      if (node.arguments.length !== 1) {
-        this.raise(node.start, Errors.ImportCallArity);
+      const maxArity = this.hasPlugin("moduleAttributes") ? 2 : 1;
+      if (node.arguments.length === 0 || node.arguments.length > maxArity) {
+        this.raise(
+          node.start,
+          Errors.ImportCallArity,
+          this.hasPlugin("moduleAttributes") ? "two arguments" : "one argument",
```
vivek12345, line 773: we are passing a message here to explain the exact number of arguments needed. for module attributes it is two arguments and for every other thing it is only one argument.

packages/babel-parser/src/plugin-utils.js
```diff
@@ -86,6 +86,20 @@ export function validatePlugins(plugins: PluginList) {
...
+    const moduleAttributesVerionPluginOption = getPluginOption(
+      plugins,
+      "moduleAttributes",
+      "version",
+    );
+    if (moduleAttributesVerionPluginOption !== "apr-2020") {
+      throw new Error(
+        "The 'moduleAttributes' plugin requires a 'version' option," +
+          " representing the last proposal update. Currently, the" +
+          " only supported value is 'feb-2020'.",
```
JLHwung, line 99: Nit: apr-2020

vivek12345: Fixed. Thank you for pointing it.

vivek12345:
@nicolo-ribaudo @JLHwung Is there something else too that needs to be done here or can we merge this now?

packages/babel-parser/src/parser/location.js
```diff
@@ -67,7 +67,7 @@ export const Errors = Object.freeze({
   IllegalReturn: "'return' outside of function",
   ImportCallArgumentTrailingComma:
     "Trailing comma is disallowed inside import(...) arguments",
-  ImportCallArity: "import() requires exactly one argument",
+  ImportCallArity: "import() requires exactly %0",
```
nicolo-ribaudo, line 70: We should also move `exactly` "inside" `%0`, since the error "requires exactly two arguments" when the proposal is enabled is wrong.

JLHwung:
@vivek12345 I have pushed some changes related to feedback about `import(a, b)` and trailing comma as we are going to release 7.10 soon. This PR is awesome! Thanks for your contribution.

packages/babel-parser/src/parser/statement.js
```diff
@@ -2041,13 +2041,31 @@ export default class StatementParser extends ExpressionParser {
...
     if (!this.match(tt.string)) {
+      // check if we have a default import like
+      // import React from "react";
       const hasDefault = this.maybeParseDefaultImportSpecifier(node);
+      /* we are checking if we do not have a default import, then it is obvious that we need named imports
+       * import { get } from "axios";
+       * but if we do have a default import
+       * we need to check if we have a comma after that and
+       * that is where this `|| this.eat` condition comes into play
+       */
```
nicolo-ribaudo, line 2052: @vivek12345 I don't know if I already thanked you, but I love these comments!

packages/babel-parser/src/parser/expression.js
```diff
@@ -765,12 +765,26 @@ export default class ExpressionParser extends LValParser {
...
-          this.raise(importArg.start, Errors.ImportCallSpreadArgument);
+        const importArgSpreadElement = node.arguments.find(
+          arg => arg.type === "SpreadElement",
+        );
+        if (importArgSpreadElement) {
+          this.raise(
+            importArgSpreadElement.start,
+            Errors.ImportCallSpreadArgument,
+          );
         }
```
nicolo-ribaudo, line 788: This is a minor point, but if we have two spreads (`import(...a, ...b)`) is should report both.

```suggestion
        for (const arg of node.arguments) {
          if (arg.type === "SpreadElement") {
            this.raise(arg.start, Errors.ImportCallSpreadArgument);
          }
        }
```

packages/babel-parser/src/parser/statement.js
vivek12345: Thank you @nicolo-ribaudo  . Means a lot 🙏

vivek12345:
Thank you @JLHwung . Really happy to see this PR getting merged soon 🎉

PR #10976, opened 2020-01-10
Update Test262, Flow and TS parser tests

nicolo-ribaudo:


scripts/parser-tests/typescript/whitelist.txt
```diff
@@ -58,6 +59,7 @@ augmentedTypesEnum2.ts
 augmentedTypesFunction.ts
 augmentedTypesInterface.ts
 augmentedTypesVar.ts
+awaitInClassInAsyncFunction.ts
```
JLHwung: Hasn't this case been fixed in master?

nicolo-ribaudo: This PR is based on c3388ea42f3ec2d9614503fd0eff81b679ba12a7, which was master 20 mins ago :thinking:

JLHwung: Oh I just realized that parseDiagnostics is incomplete. And this case is considered a valid program. I will see if I can shrink the false negatives in the whitelist tomorrow.

PR #10975, opened 2020-01-09
@babel/eslint-plugin: remove deprecated rules

kaicataldo:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | refs https://github.com/babel/babel/issues/10752
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

This PR is an initial pass at removing deprecated rules.

PR #10977, opened 2020-01-10
@babel/eslint-plugin: Update rules/tests to use @babel/eslint-parser

kaicataldo:
| Fixed Issues?            | refs #10752, https://github.com/babel/babel/pull/10975
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR builds on https://github.com/babel/babel/pull/10975 (we should land that first or, if folks prefer, this PR can supersede that one) and does the following:

1. Removes the remaining unnecessary rules from `@babel/eslint-plugin`
1. Uses `@babel/eslint-parser` for `@babel/eslint-plugin`'s tests
1. Start a new private packages called `eslint/babel-eslint-shared-fixtures` so that we can centralize shared fixtures for the ESLint packages
1. Remove rule overrides for stage 4 features. Some of these aren't handled in the ESLint core rules yet (currently blocked by https://github.com/estree/estree/pull/204), but they will be, and so I think it's best to not duplicate throwaway work here.
1. I've deleted all test cases that are not specifically testing the overridden behavior in each rule. I'd like to figure out a way to continue testing core rule tests using these packages, but manually managing copying over tests every release is a huge hassle and not a road I think we should go down. Note that some of these tests cases are very old and contained options that no longer exist.

This should get us back to a pretty sane state that we can use as a foundation to figure out what our maintenance strategy will be here. Excited to start figuring that stuff out!


eslint/babel-eslint-parser/test/babel-eslint-parser.js
```diff
@@ -7,7 +7,10 @@ import { parseForESLint } from "../src";
 import assertImplementsAST from "./helpers/assert-implements-ast";
 
 const babelOptions = {
-  configFile: path.resolve(__dirname, "./fixtures/config/babel.config.js"),
+  configFile: path.resolve(
+    __dirname,
+    "../../babel-eslint-shared-fixtures/config/babel.config.js",
```
JLHwung: You can add `@babel/eslint-shared-fixtures` as devDependency (even if that is private, but it should be okay to be dev deps) and use
```js
configFile: "@babel/eslint-shared-fixtures/config/babel.config.decorators-legacy.js"
```

eslint/babel-eslint-plugin/src/rules/no-unused-expressions.js
```diff
@@ -31,7 +32,8 @@ function isInDoStatement(node) {
   if (
     node.type === "IfStatement" &&
     node.parent &&
-    node.parent.type === "IfStatement"
+    node.parent.type === "IfStatement" &&
+    node.parent.alternate === node
```
JLHwung: Is that fixing a bug? I suggest we do the refactor only and isolate bug fix to another PR.

kaicataldo: Yes, this is guarding against `if (foo) if (baz) console.log('hi friend');`.

I'm not against removing this change, but a lot of test cases actually throw parser errors with the updated parser (this plugin is reeeaaaaaallly out of date 😬 ), so it seemed easier to update it in one fell swoop.

kaicataldo: But I can definitely remove the bug fixes/refactors from this PR if you prefer!

eslint/babel-eslint-plugin/src/rules/object-curly-spacing.js
```diff
@@ -1,26 +1,20 @@
...
-      metadata.sourceCode.getTokenBefore(node.specifiers[0]).value === "export"
-    ) {
-      return false;
-    }
+  // Allow exportDefaultFrom syntax:
+  // export x from '...';
+  if (
+    node.type === "ExportNamedDeclaration" &&
+    node.specifiers.length === 1 &&
+    node.specifiers[0].type === "ExportDefaultSpecifier"
```
kaicataldo: In addition to the bug fix above, this code change is to allow the core rule to handle `exportNamespaceFrom`, since it's stage 4.

eslint/babel-eslint-plugin/src/rules/semi.js
```diff
@@ -102,21 +103,3 @@ const semiRuleWithClassProperty = ruleComposer.joinReports([
     },
   }),
 ]);
-
-export default ruleComposer.filterReports(
```
kaicataldo: Async iterators should be handled by ESLint core

eslint/babel-eslint-plugin/src/rules/no-unused-expressions.js
kaicataldo: Also, want to note that this also includes the changes from https://github.com/babel/babel/pull/10975, so the diff isn't actually this big 😄

eslint/babel-eslint-shared-fixtures/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@babel/eslint-shared-fixtures",
+  "version": "0.0.0",
+  "description": "Shared fixtures for testing @babel/eslint-* packages",
+  "license": "MIT",
+  "private": true,
+  "dependencies": {
```
kaicataldo: Wanted to check to make sure this seemed like the best way to share these configs.

nicolo-ribaudo: We don't do anything like this currently, but I think that it's ok.

eslint/babel-eslint-plugin/src/rules/semi.js
```diff
@@ -1,13 +1,15 @@
...
 
-const isSemicolon = token => token.type === "Punctuator" && token.value === ";";
+const rule = new eslint.Linter().getRules().get("semi");
+
+function isSemicolon(token) {
+  return token.type === "Punctuator" && token.value === ";";
+}
 
-const isUnnecessarySemicolon = (context, lastToken) => {
+function isUnnecessarySemicolon(context, lastToken) {
```
nicolo-ribaudo: Fwiw, I love this change :joy:

eslint/babel-eslint-parser/package.json
```diff
@@ -29,22 +29,8 @@
     "semver": "^6.3.0"
   },
   "devDependencies": {
+    "@babel/eslint-shared-fixtures": "0.0.0",
```
nicolo-ribaudo: I think that lerna will update `0.0.0`. Maybe we could use `*` instead?

nicolo-ribaudo:
Should we also update the `peerDependency` on `eslint`, since we now are assuming that those rules don't need to be fixed?

kaicataldo:
Thanks for the reviews!

> Should we also update the peerDependency on eslint, since we now are assuming that those rules don't need to be fixed?

I thought I did this, but maybe I'm misunderstanding. Mind pointing me to which lines you're talking about?

nicolo-ribaudo:
Nevermind, I missed it somehow :sweat_smile:

kaicataldo:
Thanks again for the reviews! Any other concerns with this?

PR #10979, opened 2020-01-10
Reduce false negative cases of typescript parser tests

JLHwung:
| Fixed Issues?            | Reduced the false positive cases of parser-typescript-tests

For details please refer to the comment section of new files.

I have commented out the error codes which would introduce new false positive cases, so that this PR is only removing the false negative cases.

Before
```
 ✔ 4293 valid programs parsed without error
 ✔ 277 invalid programs produced a parsing error
 ✔ 4 invalid programs did not produce a parsing error (and allowed by the whitelist file)
 ✔ 505 valid programs produced a parsing error (and allowed by the whitelist file)
```

After
```
 ✔ 4293 valid programs parsed without error
 ✔ 388 invalid programs produced a parsing error
 ✔ 4 invalid programs did not produce a parsing error (and allowed by the whitelist file)
 ✔ 395 valid programs produced a parsing error (and allowed by the whitelist file)
```

When reading the TypeScript tests, I realize that some false negative cases comes from that we are not running test properly. i.e. [`declarationMapsOutFile.ts`](https://github.com/microsoft/TypeScript/blob/master/tests/cases/compiler/declarationMapsOutFile.ts) uses `@filename` annotation to specify multiple typescript files inside a single testcase. But we are parsing the whole testcase a single file, this would introduce FN especially if the case is testing import/exports. That's why we have many import/export related case in the whitelists. I would address this issue in a separate PR.


scripts/parser-tests/typescript/index.js
```diff
@@ -1,7 +1,8 @@
 const path = require("path");
 const fs = require("fs").promises;
-const ts = require("typescript");
+const ts = require("../../../build/typescript");
```
JLHwung: Use the bundled typescript bin since the dev dependency could be out-of-dated.

scripts/parser-tests/typescript/error-codes.js
```diff
@@ -0,0 +1,71 @@
...
+in the parser test runner. We check these error codes against the stderr log in the build/typescript/tests/baselines/reference
+
+Note that babel-parser should not throw for the TypeChecking Diagnostics
+
+The commented out diagnostic codes will introduce false positive cases that should be addressed in separate PRs.
+*/
+
+module.exports = [
+  // "TS1005", // '{0}' expected.
+  "TS1009", // Trailing comma not allowed.
```
JLHwung: TypeScript does not push `parseDiagnostics` for all and not limited to these errors. 🤷

PR #10980, opened 2020-01-10
Disallow private name in object elements and TS type elements

JLHwung:

In this PR we add a new boolean argument `isPrivateNameAllowed` for `parseMaybePrivateName` so that we can throw if private name is not allowed in the current grammar context.

This PR also clears the test262 whitelist, which marks the milestone where babel-parser is now 100% test262 compliant.


packages/babel-parser/src/plugins/flow.js
```diff
@@ -2265,10 +2265,11 @@ export default (superClass: Class<Parser>): Class<Parser> =>
     }
 
     parsePropertyName(
-      node: N.ObjectOrClassMember | N.ClassMember | N.TsNamedTypeElementBase,
+      node: N.ObjectOrClassMember | N.TsNamedTypeElementBase,
```
JLHwung: `N.ObjectOrClassMember` includes `N.ClassMember`.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1131,11 +1131,19 @@ export default class ExpressionParser extends LValParser {
...
+    isPrivateNameAllowed: boolean,
+  ): N.PrivateName | N.Identifier {
     const isPrivate = this.match(tt.hash);
 
     if (isPrivate) {
       this.expectOnePlugin(["classPrivateProperties", "classPrivateMethods"]);
+      if (!isPrivateNameAllowed) {
+        this.raise(
+          this.state.pos,
+          "Private name can only be used as name of class element (i.e. class C { #p = 42 #m() {} } )\n or property of member expression (i.e. this.#p).",
```
nicolo-ribaudo: Nit: the code in the error is not valid.

```suggestion
          "Private name can only be used as name of class element (i.e. class C { #p = 42; #m() {} } )\n or property of member expression (i.e. this.#p).",
```

packages/babel-parser/src/parser/expression.js
```diff
@@ -1131,11 +1131,19 @@ export default class ExpressionParser extends LValParser {
...
+    isPrivateNameAllowed: boolean,
+  ): N.PrivateName | N.Identifier {
     const isPrivate = this.match(tt.hash);
 
     if (isPrivate) {
       this.expectOnePlugin(["classPrivateProperties", "classPrivateMethods"]);
+      if (!isPrivateNameAllowed) {
+        this.raise(
+          this.state.pos,
+          "Private name can only be used as name of class element (i.e. class C { #p = 42; #m() {} } )\n or property of member expression (i.e. this.#p).",
```
existentialism:
```suggestion
          "Private names can only be used as the name of a class element (i.e. class C { #p = 42; #m() {} } )\n or a property of member expression (i.e. this.#p).",
```

PR #10987, opened 2020-01-11
Duplicate `__proto__` key should be allowed in object patterns

### LINKED ISSUE #6705: should allow duplicate __proto__ keys in patterns

JLHwung:
| Fixed Issues?            | Fixes #6705 

This PR can be reviewed commit by commit.

The first commit is to replace `refShorthandDefaultPos` argument by a new general struct: `refExpressionErrors`. By doing so we rebrand the `refShorthandDefaultPos` into a context variable within the expression scope.

This paves the way to fix the duplicate `__proto__` tracking issue: In the second commit we add a new property `doubleProto` to the `refExpressionErrors` and reset if we are sure the parsed ambiguous object patterns is indeed a LHS.

I have definitely borrowed a lot from acorn's design but carefully crafted `parseMaybeAssign` to avoid regressions we have solved before.


packages/babel-parser/src/parser/util.js
```diff
@@ -268,4 +268,26 @@ export default class UtilParser extends Tokenizer {
...
+
+  checkExpressionErrors(
+    refExpressionErrors: ?ExpressionErrors,
+    andThrow: boolean,
+  ) {
+    if (!refExpressionErrors) return false;
+    const { shorthandAssign } = refExpressionErrors;
+    if (!andThrow) return shorthandAssign >= 0;
+    if (shorthandAssign >= 0) {
+      this.unexpected(shorthandAssign);
```
nicolo-ribaudo: (In a separate PR)
Probably this error would be easily recoverable.

JLHwung: Yeah, I try to preserve the error output of shorthand assign error so it's indeed a refactor.

PR #10821, opened 2019-12-05
Merge env-standalone to babel-standalone

JLHwung:
| Q                        | A 
| ------------------------ | ---
| Minor: New Feature?      | Yes
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  | Add more devDependencies to babel-standalone
| License                  | MIT

This PR registers `env` in `babel-standalone` and copies the plugins in `babel-preset-env-standalone` to `babel-standalone`. By doing so babel-standalone will now includes all the functions in env-standalone, and we intend to remove env-standalone on 8.x.

The following size change is produced from [CI Artifacts before](https://app.circleci.com/jobs/github/babel/babel/13289) and [CI Artifcats after](https://app.circleci.com/jobs/github/babel/babel/13293)

| | **Previous Size** | **New Size** | **Change** |
| --- | --- | --- | --- |
| babel.js | 2.4 MB | 2.9 MB | +20% |
| babel.min.js | 1.2 MB | 1.5 MB | +25% |
| babel-preset-env.js | 3.0 MB | 3.0 MB | - |
| babel-preset-env.min.js | 1.3 MB | 1.3 MB | - |

**Edit by @nicolo-ribaudo**
The previous table should be read as this. These sizes refer to the whole bundle including all the plugins + preset-env.

|  | **Previous Total Size** | **New Total Size** | **Change** |
| --- | --- | --- | --- |
| Not minified | 5.4 MB | 2.9 MB | -46% |
| Minified | 2.5 MB | 1.5 MB | -40% |

PR #10993, opened 2020-01-12
Archive @babel/preset-env-standalone

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  |

This package shouldn't be used anymore after https://github.com/babel/babel/pull/10821.

:warning: This PR shouldn't be merged until the REPL supports _not_ loading `@babel/preset-env-standalone` from CircleCI artifacts.

---

Depends on https://github.com/babel/website/pull/2157


JLHwung:
> until the REPL supports not loading @babel/preset-env-standalone from CircleCI artifacts.

Note that REPL still needs to load preset-env-standalone for older versions than 7.8.0. I suggest we preserve its backward compatibility to 7.0.0 until, I don't know, maybe 2021.

nicolo-ribaudo:
Yes, but I think that we should check the version. If <7.8.0 then load the second file, otherwise don't.

PR #11002, opened 2020-01-13
fix: generated builder parameter should respect builder keys

### LINKED ISSUE #10998: @babel/types TypeScript declaration file has some errors above v7.8.0
## Bug Report

When I upgrade `@babel/types` to `7.8.0`, the following TypeScript source files had some compilation errors:

```typescript
import {
  arrowFunctionExpression,
  identifier,
  blockStatement,
  returnStatement,
  program,
  variableDeclaration,
  variableDeclarator,
  stringLiteral,
} from '@babel/types';

// error TS2554: Expected 4-7 arguments, but got 2.
arrowFunctionExpression(
  [identifier('arg')],
  blockStatement([returnStatement(identifier('arg'))]),
);

//  error TS2554: Expected 5 arguments, but got 1.
program([
  variableDeclaration('const', [
    variableDeclarator(identifier('foo'), stringLiteral('bar')),
  ]),
]);

```

I followed the error message and fixed it like this:

```typescript
arrowFunctionExpression(
  [identifier('arg')],
  blockStatement([returnStatement(identifier('arg'))]),
  false,
  false,
);

program(
  [
    variableDeclaration('const', [
      variableDeclarator(identifier('foo'), stringLiteral('bar')),
    ]),
  ],
  undefined,
  undefined,
  undefined,
  '',
);
```

Typescript compilation was successful，But when I use `node` to execute the compiled JS file，`@babel/types` threw errors:

```textplain
Error: ArrowFunctionExpression: Too many arguments passed. Received 4 but can receive no more than 3

Error: Program: Too many arguments passed. Received 5 but can receive no more than 4
```

**tsconfig.json**

```json
{
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    "strict": true,
    "esModuleInterop": true
  },
}
```

**Environment**

```
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 12.14.0 - /usr/local/bin/node
    Yarn: 1.21.1 - /usr/local/bin/yarn
    npm: 6.13.4 - /usr/local/bin/npm
  npmPackages:
    @babel/types: ^7.8.0 => 7.8.0 
```

JLHwung:
| Fixed Issues?            | Fixes #10998 
| Major: Breaking Change?  | No, see explainers below
| Tests Added + Pass?      | Manual tested
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR fixes a regression introduced in #10917. 

In https://github.com/babel/babel/pull/10917/files#diff-dd23409d66e27b24936f7ae4e316073eR265, the `optional` meta-field is added only if the `default` meta-field presents. This change makes sense but unexpectedly it breaks 
https://github.com/babel/babel/blob/d0a8982c124ac495b6008f9e5958ec8ad9b5a3f8/packages/babel-types/scripts/generators/typescript.js#L336

which relies on `optional` to detect if it is a nullable key, so the non-builder fields are marked as non-optional and included in the definition. For example, here is what `Program` definition 

https://github.com/babel/babel/blob/d0a8982c124ac495b6008f9e5958ec8ad9b5a3f8/packages/babel-types/src/definitions/core.js#L642

generates in different versions.

v7.7.4
```ts
export function program(
  body: Array<Statement>,
  directives?: Array<Directive>,
  sourceType?: "script" | "module",
  interpreter?: InterpreterDirective | null,
  sourceFile?: string | null
): Program;
```

v7.8.0
```ts
export function program(
  body: Array<Statement>,
  directives: Array<Directive> | undefined,
  sourceType: "script" | "module" | undefined,
  interpreter: InterpreterDirective | null | undefined,
  sourceFile: string
): Program;
```
The `sourceFile` is not nullable so all the parameters before `sourceFile` becomes non optional parameters. This is a breaking change for types consumers.

The problem here is that we should not include `sourceFile` at all because it is not defined in `builders` and we are really lucky that nobody ever complains that the fifth argument `sourceFile?` does not work _at_ _all_.

This PR filter the args against `BUILDER_KEYS` so the generated types become
```ts
export function program(
  body: Array<Statement>,
  directives?: Array<Directive>,
  sourceType?: "script" | "module",
  interpreter?: InterpreterDirective | null
): Program;
```
where the `sourceFile` is removed from builder arguments definition, technically it is a breaking change but practically it should not break anyone because the builder will check the arguments length at runtime.

In the long term we should add the generated `index.js.flow` and `index.d.ts` to the codebase so we can track such changes.

### DIFF
--- a/packages/babel-types/scripts/generators/typescript.js
+++ b/packages/babel-types/scripts/generators/typescript.js
@@ -56,6 +56,7 @@ const lines = [];
 for (const type in t.NODE_FIELDS) {
   const fields = t.NODE_FIELDS[type];
   const fieldNames = sortFieldNames(Object.keys(t.NODE_FIELDS[type]), type);
+  const builderNames = t.BUILDER_KEYS[type];
 
   const struct = ['type: "' + type + '";'];
   const args = [];
@@ -75,18 +76,20 @@ for (const type in t.NODE_FIELDS) {
       typeAnnotation += " | null";
     }
 
-    if (areAllRemainingFieldsNullable(fieldName, fieldNames, fields)) {
-      args.push(
-        `${t.toBindingIdentifierName(fieldName)}${
-          isNullable(field) ? "?:" : ":"
-        } ${typeAnnotation}`
-      );
-    } else {
-      args.push(
-        `${t.toBindingIdentifierName(fieldName)}: ${typeAnnotation}${
-          isNullable(field) ? " | undefined" : ""
-        }`
-      );
+    if (builderNames.includes(fieldName)) {
+      if (areAllRemainingFieldsNullable(fieldName, builderNames, fields)) {
+        args.push(
+          `${t.toBindingIdentifierName(fieldName)}${
+            isNullable(field) ? "?:" : ":"
+          } ${typeAnnotation}`
+        );
+      } else {
+        args.push(
+          `${t.toBindingIdentifierName(fieldName)}: ${typeAnnotation}${
+            isNullable(field) ? " | undefined" : ""
+          }`
+        );
+      }
     }
 
     const alphaNumeric = /^\w+$/;
--- a/packages/babel-types/scripts/generators/flow.js
+++ b/packages/babel-types/scripts/generators/flow.js
@@ -54,6 +54,7 @@ for (const type in t.NODE_FIELDS) {
 
   const struct = ['type: "' + type + '";'];
   const args = [];
+  const builderNames = t.BUILDER_KEYS[type];
 
   Object.keys(t.NODE_FIELDS[type])
     .sort((fieldA, fieldB) => {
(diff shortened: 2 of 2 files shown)


packages/babel-types/scripts/generators/typescript.js
```diff
@@ -75,18 +76,20 @@ for (const type in t.NODE_FIELDS) {
...
-          isNullable(field) ? "?:" : ":"
-        } ${typeAnnotation}`
-      );
-    } else {
-      args.push(
-        `${t.toBindingIdentifierName(fieldName)}: ${typeAnnotation}${
-          isNullable(field) ? " | undefined" : ""
-        }`
-      );
+    if (builderNames.includes(fieldName)) {
```
nicolo-ribaudo: Wouldn't it be better to directly iterate over `builderNames` instead of `fieldNames`? Since `fieldNames` is generated from an object, I think that we risk changing the order of its keys without thinking about this problem.

JLHwung: The `fieldNames` has been sorted according to the orders of `buildKeys` and anything not in the `builders` are at tails.
https://github.com/babel/babel/blob/d0a8982c124ac495b6008f9e5958ec8ad9b5a3f8/packages/babel-types/scripts/generators/typescript.js#L349

PR #11011, opened 2020-01-15
Bug/missing references after crawl

### LINKED ISSUE #10896: Missing referencePaths after crawl
## Bug Report

**Current Behavior**
`scope.crawl()` does not add references to bindings correctly in some cases. In particular I get the bug when referencing a class inside a class method.

**Input Code**
https://github.com/regiontog/babel/compare/master...bug/missing-references-after-crawl

```js
const path = getPath("class a { build() { return new a(); } }");
path.scope.crawl();
const referencePaths = path.scope.bindings.a.referencePaths;
expect(referencePaths).toHaveLength(1);
```

**Expected behavior/code**
I expect the binding `a` to have 1 reference inside the scope of the `build` class method after a `crawl`. The binding has 1 reference before the `crawl` call is made.

**Environment**

```
System:
    OS: Linux 5.4 Arch Linux
  Binaries:
    Node: 13.3.0 - /usr/bin/node
    Yarn: 1.21.0 - /usr/bin/yarn
    npm: 6.12.1 - /usr/bin/npm
  Monorepos:
    Lerna: 3.19.0
npmPackages:
    @babel/cli: ^7.7.0 => 7.7.5
    @babel/core: ^7.7.2 => 7.7.5
    @babel/eslint-plugin-development: ^1.0.1 => 1.0.1
    @babel/plugin-proposal-class-properties: ^7.7.0 => 7.7.4
    @babel/plugin-proposal-export-namespace-from: ^7.5.2 => 7.7.4
    @babel/plugin-proposal-nullish-coalescing-operator: ^7.4.4 => 7.7.4
    @babel/plugin-proposal-numeric-separator: ^7.2.0 => 7.7.4
    @babel/plugin-proposal-object-rest-spread: ^7.7.4 => 7.7.4
    @babel/plugin-proposal-optional-chaining: ^7.6.0 => 7.7.5
    @babel/plugin-transform-flow-strip-types: ^7.7.4 => 7.7.4
    @babel/plugin-transform-for-of: ^7.7.4 => 7.7.4
    @babel/plugin-transform-modules-commonjs: ^7.7.0 => 7.7.5
    @babel/plugin-transform-runtime: ^7.6.2 => 7.7.6
    @babel/preset-env: ^7.7.1 => 7.7.6
    @babel/preset-flow: ^7.0.0 => 7.7.4
    @babel/register: ^7.7.0 => 7.7.4
    @babel/runtime: ^7.7.2 => 7.7.6
    babel-eslint: ^11.0.0-beta.2 => 11.0.0-beta.2
    babel-jest: ^24.9.0 => 24.9.0
    babel-plugin-transform-charcodes: ^0.2.0 => 0.2.0
    eslint: ^6.0.1 => 6.7.2
    eslint-config-babel: ^9.0.0 => 9.0.0
    gulp-babel: ^8.0.0 => 8.0.0
    jest: ^24.9.0 => 24.9.0
    lerna: ^3.19.0 => 3.19.0
    rollup-plugin-babel: ^4.0.0 => 4.3.3
```
- How you are using Babel: API

**Possible Solution**

It seems suspect to me to add the reference to a binding in the references own scope [here](https://github.com/babel/babel/blob/d389b702e65bfb0a3cc8258907db76560491547a/packages/babel-traverse/src/scope/index.js#L756)

Output of test linked above:
```
scope › binding paths › reference paths after crawl

    expect(received).toHaveLength(expected)

    Expected length: 1
    Received length: 0
    Received array:  []

      276 |       path.scope.crawl();
      277 |       const referencePaths = path.scope.bindings.a.referencePaths;
    > 278 |       expect(referencePaths).toHaveLength(1);
          |                              ^
      279 |     });
      280 |   });
      281 |

      at Object.<anonymous> (packages/babel-traverse/test/scope.js:278:30)
```

regiontog:
| Fixed Issues?            | Fixes #10896 
| Major: Breaking Change?  |Uncertain| Minor: New Feature?      |No
| Documentation PR Link    | 

When traversing during a `crawl()` babel currently sets the class identifier inside the class to the _old_ binding. This has caused a number of issues in parcel(https://github.com/parcel-bundler/parcel/issues/3932, https://github.com/parcel-bundler/parcel/issues/2513, https://github.com/parcel-bundler/parcel/issues/2981, https://github.com/parcel-bundler/parcel/issues/3664). This PR sets the binding to the binding in the parent. See #10896 for in depth explanation. 

In addition the assumption that a `Declaration` visitor is visited before `ClassDeclaration` in `crawl()` was removed by merging the visitors.


packages/babel-traverse/src/scope/index.js
```diff
@@ -131,14 +139,6 @@ const collectorVisitor = {
     scope.getBlockParent().registerDeclaration(path);
   },
 
-  ClassDeclaration(path) {
-    const id = path.node.id;
-    if (!id) return;
-
-    const name = id.name;
-    path.scope.bindings[name] = path.scope.getBinding(name);
```
jridgewell: Isn't the following change all that's necessary?

```diff
-    path.scope.bindings[name] = path.scope.getBinding(name);
+    path.scope.bindings[name] = path.scope.parent.getBinding(name);
```

nicolo-ribaudo: I agree. `parent` is the nearest function scope, but classes are block-scoped:
```js
let A;
function fn() {
  {
    class A {}
  }
  A; // This references the let declaration, not the path.
}
```

JLHwung: This change
```diff
-    path.scope.bindings[name] = path.scope.getBinding(name);
+    path.scope.bindings[name] = path.scope.parent.getBinding(name);
```
implicitly depends on the fact that the `Declaration` visitor runs _before_ `ClassDeclaration` visitor, so that `path.scope.parent.getBinding` always points to the new binding which may be initialized in `Declaration` visitor or `BlockScoped` visitor.

See https://github.com/babel/babel/issues/10896#issuecomment-574415561

Besides that, I agree we should do `path.scope.parent` instead of `parent`.

JLHwung: @regiontog Ah you are right! The `ClassDeclaration` binding is registered in `BlockScoped` and `BlockScoped` runs _before_ `ClassDeclaration`. I am curious why the test can even get passed.

PR #11014, opened 2020-01-15
Fix parentheses removal in nullish-coalescing operation

### LINKED ISSUE #10966: Parentheses removed falsely when transforming nullish-coalescing-operator
## Bug Report

**Current Behavior**
When transforming an AST containing a nullish coalescing operator, parentheses are removed falsely, leading to different behavior.

**Input Code**
```js
const babel = require('@babel/core')
const { readFileSync, writeFileSync } = require('fs')

const code = readFileSync('src/index.js')
// src/index.js:
// const foo = 'test'
// console.log((foo ?? '') == '')

const babelConfig = {
  plugins: [
    '@babel/plugin-proposal-nullish-coalescing-operator',
  ],
}

// this works as expected
const ast = babel.parseSync(code, babelConfig)

// this removes the parentheses
const { code: transformedCode } = babel.transformFromAstSync(ast, code)

writeFileSync('dist/index.js', transformedCode)
// dist/index.js:
// const foo = 'test';
// console.log(foo ?? '' == '');

const babelRegister = require('@babel/register')
babelRegister(babelConfig)

require('./dist')
// outputs 'test' instead of false
```

See also:
https://github.com/dword-design/test-babel-missing-parentheses

**Expected behavior/code**
The resulting file should still contain the parentheses.

**Environment**

- Babel version(s): 7.7.7
- Node/npm version: Node 11.15.0/npm 6.7.0
- OS: macOS 10.14.6
- Monorepo: no
- How you are using Babel: core, register

sidntrivedi012:
| Fixed Issues?            | Fixes #10966  
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


sidntrivedi012:
@nicolo-ribaudo Made a draft PR for working on #10966 . I tried adding precedence to `??`` operator in the list but no success.

sidntrivedi012:
@nicolo-ribaudo Its working. The problem was that I was not running `make build` or `make watch` for compiling the changes made to `parentheses.js`.

Also, I have assigned precedence 0 to it since https://tc39.es/ecma262/#sec-binary-logical-operators mentions `Logical OR ` and `??` at same precedence. PTAL. Thanks. 🙂

packages/babel-generator/src/node/parentheses.js
```diff
@@ -2,6 +2,7 @@ import * as t from "@babel/types";
 
 const PRECEDENCE = {
   "||": 0,
```
jridgewell: While it's unobservable now, `??` should actually be the lowest precedence of all tokens. Meaning we'll need to increase every other token's precedence.

packages/babel-generator/test/fixtures/parentheses/nullish-coalescing/output.js
```diff
@@ -0,0 +1,2 @@
+const foo = 'test';
+console.log(foo ?? '' == '');
```
jridgewell: Don't forget to update the test

packages/babel-generator/src/node/parentheses.js
sidntrivedi012: Ok, so I am assigning a precedence of 0 to `??` and will increment that of others by 1. BTW, I am still not clear as to how to get to know the precedence that we follow here.

packages/babel-generator/test/fixtures/parentheses/nullish-coalescing/output.js
sidntrivedi012: Oops. My bad. Making the changes.

packages/babel-generator/src/node/parentheses.js
jridgewell: Oh, actually, you're right. `SortCircuitExpression` is both `CoalesceExpression` and `LogicalORExpression`, so they have the same precedence. I was remember when we had the spec using grammar flags, and there it was `CoalesceExpression -> LogicalORExpression`.

sidntrivedi012: Oh, I had already made the changes. 😅 Reverting them.

sidntrivedi012:
@jridgewell Made the changes. PTAL. Thanks.

sidntrivedi012:
@nicolo-ribaudo Thanks a lot for such a detailed explanation. I was confused earlier but not now. And yes, I fully trust your words, no need to read the spec. 😉

PR #11015, opened 2020-01-15
add AST for the module attribute proposal

xtuc:
Introduce a `ImportAttribute` structure and add an `attributes` key on
the `ImportDeclaration` node.

| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/ast/spec.md
+++ b/packages/babel-parser/ast/spec.md
@@ -107,6 +107,7 @@ These are the core @babel/parser (babylon) AST node types.
     - [ImportSpecifier](#importspecifier)
     - [ImportDefaultSpecifier](#importdefaultspecifier)
     - [ImportNamespaceSpecifier](#importnamespacespecifier)
+    - [ImportAttribute](#importattribute)
   - [Exports](#exports)
     - [ExportNamedDeclaration](#exportnameddeclaration)
     - [ExportSpecifier](#exportspecifier)
@@ -1221,6 +1222,7 @@ interface ImportDeclaration <: ModuleDeclaration {
   importKind: null | "type" | "typeof" | "value";
   specifiers: [ ImportSpecifier | ImportDefaultSpecifier | ImportNamespaceSpecifier ];
   source: Literal;
+  attributes?: [ ImportAttribute ];
 }
 ```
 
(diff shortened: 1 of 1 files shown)


packages/babel-parser/ast/spec.md
```diff
@@ -1259,6 +1261,18 @@ interface ImportNamespaceSpecifier <: ModuleSpecifier {
 
 A namespace import specifier, e.g., `* as foo` in `import * as foo from "mod.js"`.
 
+### ImportModuleAttribute
+
+```js
+interface ImportModuleAttribute <: Node {
+  type: "ImportModuleAttribute";
+  key: Identifier;
+  value: StringLiteral;
```
JLHwung: Does it mean that tc39 has decided it will only support string types primitive?

nicolo-ribaudo: Probably not, but it's better to be conservative for now. Allowing new things is "less breaking" than disallowing other kinds of literals.

xtuc: It's still in discussion. I would ideally support all the primtives, but Babel could get keep it simple for now

packages/babel-parser/ast/spec.md
```diff
@@ -1259,6 +1261,18 @@ interface ImportNamespaceSpecifier <: ModuleSpecifier {
 
 A namespace import specifier, e.g., `* as foo` in `import * as foo from "mod.js"`.
 
+### ImportModuleAttribute
+
+```js
+interface ImportModuleAttribute <: Node {
+  type: "ImportModuleAttribute";
```
jridgewell: Nit: I would call it "ImportAttribute".

nicolo-ribaudo: @xtuc WDYT?

xtuc: that works for me. I will update that

packages/babel-parser/ast/spec.md
littledan: We're still discussing this among the people working on module attributes. I'd rather hold off on landing this a bit longer.

PR #11018, opened 2020-01-16
Print Babel config

rajasekarm:
| Fixed Issues?            | https://github.com/babel/babel/issues/10617
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-core/src/utils/print-config.js
```diff
@@ -0,0 +1,41 @@
...
+      plugins: config.babelrc.options.plugins || [],
+    };
+    console.log(babelrc);
+  }
+
+  if (config.cli) {
+    console.log("--------------");
+
+    const babelcli = {
+      from: "@babel/cli",
```
nicolo-ribaudo: What about using
```suggestion
      from: `caller (${config.options.caller.name})`,
```

packages/babel-core/src/utils/print-config.js
```diff
@@ -0,0 +1,41 @@
...
+      presets: config.cli.presets || [],
+      plugins: config.cli.plugins || [],
+    };
+    console.log(babelcli);
+    console.log("--------------");
+  }
+
+  // if (config.options.caller.name === "babel-loader") {
```
JLHwung: The commented-out code should be removed.

packages/babel-cli/src/babel/dir.js
```diff
@@ -43,7 +44,9 @@ export default async function({
         ? path.extname(relative)
         : cliOptions.outFileExtension,
     );
-
+    if (!print) {
+      babelOptions.showConfig = false;
```
xtuc: maybe it's just a preference but I would use ```
babelOptions.showConfig = print;
```

and add a default value to `print`

PR #11021, opened 2020-01-16
chore: replace yarn-upgrade by bump-babel-dependencies in vuejs e2e tests

JLHwung:
| Fixed Issues?            | Fix failing vue-cli e2e tests

### DIFF
--- a/scripts/integration-tests/utils/bump-babel-dependencies.js
+++ b/scripts/integration-tests/utils/bump-babel-dependencies.js
@@ -1,0 +1,29 @@
+const fs = require("fs");
+const path = require("path");
+const cwd = process.cwd();
+const packageJSONPath = path.resolve(cwd, "./package.json");
+const content = JSON.parse(fs.readFileSync(packageJSONPath));
+
+let bumped = false;
+function bumpBabelDependency(dependencies) {
+  for (const dep of Object.keys(dependencies)) {
+    if (dep.startsWith("@babel/")) {
+      dependencies[dep] = "latest";
+      bumped = true;
+    }
+  }
+}
+
+if ("peerDependencies" in content) {
+  bumpBabelDependency(content.peerDependencies);
+}
+if ("devDependencies" in content) {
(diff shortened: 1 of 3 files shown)


scripts/integration-tests/utils/bump-babel-dependencies.js
```diff
@@ -0,0 +1,29 @@
+const fs = require("fs");
```
JLHwung: @nicolo-ribaudo WDYT replacing `yarn-upgrade` by this snippet. 🤷‍♂️

scripts/integration-tests/utils/bump-babel-dependencies.js
```diff
@@ -0,0 +1,29 @@
...
+const path = require("path");
+const cwd = process.cwd();
+const packageJSONPath = path.resolve(cwd, "./package.json");
+const content = JSON.parse(fs.readFileSync(packageJSONPath));
+
+let bumped = false;
+function bumpBabelDependency(dependencies) {
+  for (const dep of Object.keys(dependencies)) {
+    if (dep.startsWith("@babel/")) {
+      dependencies[dep] = "latest";
```
JLHwung: Technically if a downstream project specifies `latest` for `@babel/` in the `package.json`, this approach will not work at all because `yarn` will then lock `latest` to old versions. But practically this scenario should never happen.

JLHwung:
@nicolo-ribaudo The [first error](https://app.circleci.com/jobs/github/babel/babel/15442) is due to https://github.com/yarnpkg/yarn/issues/7797, instead of manually updating every babel plugins, I fall back to Yarn 1.18, but then there is a [second error](https://app.circleci.com/jobs/github/babel/babel/15478) which never happens on Yarn >= 1.19. I'd rather not to die on this hill and come up with `bump-babel-dependencies.js`, which is exactly what we meant to achieve.

nicolo-ribaudo:
Ok then, let's use that script everywhere

Issue #10988, opened 2020-01-11
await should be allowed when it is not in AsyncArrowHead


JLHwung:
## Bug Report

- [ ] I would like to work on a fix!

**Current Behavior**
A clear and concise description of the behavior.
The following snippet will throw SyntaxError: Await cannot be used as name inside an async function (1:1)

**Input Code**
- REPL or Repo link if applicable: [REPL](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=BQQw7iCWAuCUAEBeAfPA3gXyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=script&lineWrap=true&presets=env%2Cenv&prettier=false&targets=&version=7.8.4&externalPlugins=)

```js
(await) => {}
```

**Expected behavior/code**
It should not throw because it is not in async arrow parameters, which means `await` is a plain identifier and should be allowed in ArrowParameters because it does not have `[Await]` production parameter.
https://tc39.es/ecma262/#sec-arrow-function-definitions

**Babel Configuration**
REPL, no plugins are enabled

**Environment**

REPL

**Possible Solution**

Should revisit how `awaitPos` is tracked in the parser. You can pay attention to
https://github.com/babel/babel/blob/a0a9c64a470acec77bec1da0b4bceec4ccf7f446/packages/babel-parser/src/parser/expression.js#L2172-L2175

where `awaitPos` is updated when it may be in arrow parameters. It was used to track `awaitPos` on cases like
```js
async ( x = (await) => {}) => {}
```
however it becomes overkill when it is _not_ in _async_ arrow parameters. Consider add a new parser state `maybeInAsyncArrowHead` when parsing ambiguous async arrow

https://github.com/babel/babel/blob/master/packages/babel-parser/src/parser/expression.js#L649

It should be similar to `maybeAsyncArrow` but it should have longer lifecycle -- `maybeInAsyncArrowHead` should be true during the whole async arrow parameters. i.e.
```js
// top level
async ( // maybeInAsyncArrowHead : true
  x = ( // maybeInAsyncArrowHead : true
    await // register the `awaitPos`
  ) => async ( // oldMaybeInAsyncArrowHead should be saved, new maybeInAsyncArrowHead should be true
    y
  ) => {}
); // we does not see an arrow, so the parsed ambiguous async arrow is indeed a CallExpression, reset `awaitPos`.
```

[...]

PR #11022, opened 2020-01-16
Replace switch statement with an object literal

arku:
This is a super minor refactor I did on a piece of code which I came across while working on https://github.com/babel/babel/issues/10988

Feel free to close it if you don't think it adds any value.

### DIFF
--- a/packages/babel-parser/src/plugins/typescript/index.js
+++ b/packages/babel-parser/src/plugins/typescript/index.js
@@ -57,34 +57,28 @@ type ParsingContext =
   | "TypeMembers"
   | "TypeParametersOrArguments";
 
+const tsKeywordTypes = {
+  any: "TSAnyKeyword",
+  boolean: "TSBooleanKeyword",
+  bigint: "TSBigIntKeyword",
+  never: "TSNeverKeyword",
+  number: "TSNumberKeyword",
+  object: "TSObjectKeyword",
+  string: "TSStringKeyword",
+  symbol: "TSSymbolKeyword",
+  undefined: "TSUndefinedKeyword",
+  unknown: "TSUnknownKeyword",
+};
+
 // Doesn't handle "void" or "null" because those are keywords, not identifiers.
 function keywordTypeFromName(
   value: string,
 ): N.TsKeywordTypeType | typeof undefined {
-  switch (value) {
-    case "any":
-      return "TSAnyKeyword";
-    case "boolean":
-      return "TSBooleanKeyword";
-    case "bigint":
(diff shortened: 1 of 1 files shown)


packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -61,30 +61,20 @@ type ParsingContext =
...
-    case "symbol":
-      return "TSSymbolKeyword";
-    case "undefined":
-      return "TSUndefinedKeyword";
-    case "unknown":
-      return "TSUnknownKeyword";
-    default:
-      return undefined;
-  }
+  const keywordTypes = {
```
arku: Is this variable name fine? Or should we name it `tsKeywordTypes`?

arku:
@JLHwung Thanks for the quick review!

Interesting points. I definitely wouldn't have thought of those. Both points can be addressed easily, I think. Would you be interested in the PR if your points are addressed? 🙂

arku:
@JLHwung I have addressed your comments. Take a look when you get a chance.

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -57,34 +57,28 @@ type ParsingContext =
...
-      return "TSStringKeyword";
-    case "symbol":
-      return "TSSymbolKeyword";
-    case "undefined":
-      return "TSUndefinedKeyword";
-    case "unknown":
-      return "TSUnknownKeyword";
-    default:
-      return undefined;
+  if (!Object.prototype.hasOwnProperty.call(tsKeywordTypes, value)) {
```
arku: @JLHwung We could avoid this check by creating the `tsKeywordTypes` object with `Object.create(null)` which should have no prototype -- hence no additional prototype chain checks when we try to access a key.

Let me know which solution you prefer.

existentialism:
I think the switch statement is fine?

kaicataldo:
Mind explaining what problem you're trying to solve with this change? The current code seems fine to me.

arku:
I spent quite a bit of time thinking about this but couldn't come up with a reason other than improving the readability by a tad (the `switch` statement was a bit long with more than 10 `case`s). And I admit I didn't think about the memory implications of creating an object vs using a `switch` statement.

I am going to close the PR as it doesn't add much value to the current version. Thank you all, especially @JLHwung for teaching me something, even from this trivial PR.

And sorry for the noise!

existentialism:
@arku there is no need to apologize, thanks for opening the PR!

PR #11024, opened 2020-01-16
Check that generated compat-data is up to date on CI

nicolo-ribaudo:

This will prevent cases like 6363b594cc96caa55b293bad4da1d948a31d1702: the file used to generate the data was updated but the data itself wasn't, so we didn't notice the regression in the PR which updated the compat-table mappings.

### DIFF
--- a/packages/babel-compat-data/scripts/build-data.js
+++ b/packages/babel-compat-data/scripts/build-data.js
@@ -4,7 +4,6 @@ const fs = require("fs");
 const path = require("path");
 const semver = require("semver");
 const flattenDeep = require("lodash/flattenDeep");
-const isEqual = require("lodash/isEqual");
 const mapValues = require("lodash/mapValues");
 const pickBy = require("lodash/pickBy");
 const { unreleasedLabels } = require("@babel/helper-compilation-targets");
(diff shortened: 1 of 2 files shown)


packages/babel-compat-data/scripts/build-data.js
```diff
@@ -297,19 +296,19 @@ const generateData = (environments, features) => {
...
 
-    if (!isEqual(currentData, newData)) {
+    // Compare as JSON strings to also check keys ordering
+    if (currentData !== stringified) {
       console.error(
         "The newly generated plugin/built-in data does not match the current " +
-          "files. Re-run `npm run build-data`."
+          "files. Re-run `make build-compat-data`."
       );
       process.exit(1);
```
JLHwung:
```suggestion
      process.exitCode = 1;
```

nicolo-ribaudo: Why?

JLHwung: > In most situations, it is not actually necessary to call process.exit() explicitly.

https://nodejs.org/api/process.html#process_process_exit_code

packages/babel-compat-data/scripts/build-data.js
```diff
@@ -306,7 +306,7 @@ const generateData = (environments, features) => {
         "The newly generated plugin/built-in data does not match the current " +
           "files. Re-run `make build-compat-data`."
       );
-      process.exit(1);
```
JLHwung: Can you revert this commit? I realize that this function is in a `forEach` callback.

PR #11025, opened 2020-01-16
Enable `no-process-exit`

JLHwung:

This PR includes commits from #11026.

Enabled `no-process-exit` rule and fixed linting errors. Rationale: https://nodejs.org/api/process.html#process_process_exit_code

In most case the synchronous `process.exit()` is unnecessary, and it causes asynchronous IO operations terminated even when they are not fulfilled, e.g. writes to `stdout` or `stderr`.

In our codebase, `process.exit()` is often used as top level return: It is straightforward to fix: wrap the statements into an `else` branch so we are not scheduling new tasks for the event loop.

### DIFF
--- a/packages/babel-compat-data/scripts/build-data.js
+++ b/packages/babel-compat-data/scripts/build-data.js
@@ -290,7 +290,7 @@ const generateData = (environments, features) => {
   });
 };
 
-["plugin", "corejs2-built-in"].forEach(target => {
+for (const target of ["plugin", "corejs2-built-in"]) {
   const newData = generateData(
     environments,
     require(`./data/${target}-features`)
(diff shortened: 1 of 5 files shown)


packages/babel-node/src/babel-node.js
```diff
@@ -92,7 +92,7 @@ getV8Flags(function(err, v8Flags) {
         if (signal) {
           process.kill(process.pid, signal);
         } else {
-          process.exit(code);
+          process.exitCode = code;
```
JLHwung: node.js will terminate when all `exit` listener is executed, so we don't need to call `process.exit` again in the listener.

packages/babel-parser/bin/babel-parser.js
```diff
@@ -7,10 +7,9 @@ var fs = require("fs");
 var filename = process.argv[2];
 if (!filename) {
   console.error("no filename specified");
-  process.exit(0);
```
JLHwung: `process.exit(0)` is unnecessary after refactor.

packages/babel-compat-data/scripts/build-data.js
```diff
@@ -305,11 +305,10 @@ const generateData = (environments, features) => {
         "The newly generated plugin/built-in data does not match the current " +
           "files. Re-run `npm run build-data`."
       );
-      process.exit(1);
+      process.exitCode = 1;
+      break;
     }
-
-    process.exit(0);
```
JLHwung: This is rather a bug because when `--check` is provided, it will still exit after checking only `"plugin"` data. Removed after refactor.

PR #11028, opened 2020-01-17
Added jsescOptions to Numeric Literals

### LINKED ISSUE #10960: [babel-generator]: jsescOption.numbers does not affect NumericLiterals
## Bug Report

**Current Behavior**
Currently, when I set:

```
{
    jsescOption: {
        numbers: 'hexadecimal'
    }
}
```
This option has not affects any numbers that `@babel/generator` is generated.

For example i need the following in the output generated code:
```
var foo = 0x1;
```

**Expected behavior/code**
```
{
    jsescOption: {
        numbers: 'hexadecimal'
    }
}
```
This option should affect any numbers that `@babel/generator` is generated

- @babel/generator version: 7.7.7 
- Node/npm version: Node 12

**Possible Solution**
Looks like here: https://github.com/babel/babel/blob/master/packages/babel-generator/src/generators/types.js#L121
number value should be generated using `jsesc` like in `StringLiteral`

sidntrivedi012:
| Fixed Issues?            | Fixes #10960  
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | https://github.com/babel/website/pull/2172
| Any Dependency Changes?  |


sidntrivedi012:
Reffered the code from StringLiteral function for Numeric Literals. What can we use here instead of `this.format.jsonCompatibleStrings` for Numeric Literals?

packages/babel-generator/src/generators/types.js
```diff
@@ -120,7 +120,11 @@ export function NullLiteral() {
 
 export function NumericLiteral(node: Object) {
   const raw = this.getPossibleRaw(node);
-  const value = node.value + "";
+  const opts = this.format.jsescOption;
+  if (this.format.jsonCompatibleStrings) {
```
JLHwung: Why `jsonCompatibleString` should be applied fro `NumericLiteral`? Also `jsonCompatibleString` will be removed in 8.x: https://github.com/babel/babel/pull/9958

packages/babel-generator/src/generators/types.js
```diff
@@ -120,7 +120,11 @@ export function NullLiteral() {
...
+    opts.json = true;
+  }
+  const value = jsesc(node.value, opts);
```
JLHwung: Can you add tests for these two `jsesc` options? One should expect both `numbers` and `lowercasehex` should work for `NumericLiteral`.

packages/babel-generator/src/generators/types.js
sidntrivedi012: @JLHwung Tried looking for but did not find what to use instead of `jsonCompatibleString`. That's why used it. What can we use in place of it?

Oh, if it's going to be removed then we will have to replace it in the `StringLiteral` function. Should I make an issue for it?

packages/babel-generator/src/generators/types.js
sidntrivedi012: Ok:+1:, do I also need to set
```
{
    jsescOption: {
        numbers: 'hexadecimal'
    }
}
```
in the tests?

JLHwung: Yep that is exactly what #10960 is requiring.

packages/babel-generator/src/generators/types.js
JLHwung: `jsonCompatibleString` is a legacy option and it is equivalent to say `jsescOptions: { json: true }`.

Since it is a legacy option and we never support `jsescOptions` on `NumericLiteral` before, I would say we'd better remove this `jsonCompatibleString` thing and call `jsesc` directly.

```js
const value = jsesc(node.value, opts);
```

sidntrivedi012:
@JLHwung Made the changes. But it is affecting other tests. PTAL.

packages/babel-generator/test/fixtures/escapes/jsonEscape/output.js
```diff
@@ -1,2 +1,2 @@
 0;// Not a directive
-"\u00A9";
\ No newline at end of file
+"\u00a9";
```
JLHwung: You can add an `options.json` to this testcase
```json
{
  "minified": true,
  "jsonCompatibleStrings": true
}	}
```
So it will _not_ be affected by `numbers: true`.

packages/babel-generator/test/fixtures/escapes/options.json
```diff
@@ -1,4 +1,6 @@
 {
```
JLHwung: ditto, the `numbers: "hexadecimal"` should be moved to `options.json` _inside_ `escapes/numeric-literals`.

packages/babel-generator/test/fixtures/minified/literals/output.js
```diff
@@ -1 +1 @@
-5;5;50;5e4;"foobar";" ";"\n\r";"\uD83D\uDE02";`😂`;/foobar/g;null;true;false;5;2;56;31;
```
JLHwung: I don't think this file should be updated.

packages/babel-generator/test/fixtures/escapes/jsonEscape/options.json
```diff
@@ -1,4 +1,4 @@
 {
   "minified": true,
   "jsonCompatibleStrings": true
-}
+}
```
JLHwung: Can you exclude this change?

packages/babel-generator/src/index.js
```diff
@@ -47,6 +47,8 @@ function normalizeOptions(code, opts): Format {
     minified: opts.minified,
     concise: opts.concise,
     jsonCompatibleStrings: opts.jsonCompatibleStrings,
+    numbers: opts.numbers,
+    lowercaseHex: opts.lowercaseHex,
```
nicolo-ribaudo: Sicne we plan to remove `jsonCompatibleStrings` because it can already be set using `jsescOption`, I don't think that we should introduce `numbers` and `lowercaseHex`.
Users can directly set them on `jsescOption`.

JLHwung: Yeah, I talked to @sidntrivedi012 offline on this problem, he is working on a fix.

sidntrivedi012:
@JLHwung @nicolo-ribaudo PTAL now. Have added conditions in `types.js` to check where `jsesc` will need to be applied. Have also added a test. Alongwith this, have made separate `optional.json` files for `jsonEscape` and `NumericLiteral` folders in `fixtures`.

sidntrivedi012:
@nicolo-ribaudo @JLHwung PTAL now. Have modified the conditions as mentioned above.

packages/babel-generator/src/generators/types.js
```diff
@@ -120,8 +120,11 @@ export function NullLiteral() {
 
 export function NumericLiteral(node: Object) {
   const raw = this.getPossibleRaw(node);
+  const opts = this.format.jsescOption;
   const value = node.value + "";
-  if (raw == null) {
+  if (this.format.jsescOption.numbers) {
```
nicolo-ribaudo: Nit: use `opts.numbers`

packages/babel-generator/src/generators/types.js
```diff
@@ -120,8 +120,11 @@ export function NullLiteral() {
...
+    this.number(jsesc(parseInt(value), opts));
```
nicolo-ribaudo: Why use `parseInt` when we already have the original value at `node.value`?

sidntrivedi012:
@nicolo-ribaudo Made the changes. PTAL now.

packages/babel-generator/src/generators/types.js
sidntrivedi012: `jsesc` was not working with `raw` until I added `parseInt` to it. Thinking the same, I added it to `value` too. 😅

packages/babel-generator/src/generators/types.js
```diff
@@ -120,8 +120,11 @@ export function NullLiteral() {
 
 export function NumericLiteral(node: Object) {
   const raw = this.getPossibleRaw(node);
+  const opts = this.format.jsescOption;
   const value = node.value + "";
-  if (raw == null) {
+  if (opts.numbers) {
+    this.number(jsesc(value, opts));
```
nicolo-ribaudo: `value` is a string, but `jsesc` [expects a number](https://github.com/mathiasbynens/jsesc#numbers). You can use `node.value`.

sidntrivedi012:
@nicolo-ribaudo Done. Please review.

sidntrivedi012:
@JLHwung where should I add the docs for this PR?

nicolo-ribaudo:
We should mention in https://github.com/babel/website/blob/master/docs/generator.md that `jsescOption` is also applied to numbers.

sidntrivedi012:
Ok. Will add the details to the doc page 👍

PR #11029, opened 2020-01-17
fix(babel-parser): avoid state.clone() to clone the whole token store

3cp:
| Documentation PR Link    |

Fixed the performance issue on large input when turned on option {tokens: true} and typescript plugin which uses quite a few state.clone().

## To reproduce the performance issue.
Create a nodejs project with following 2 files:

package.json
```json
{
  "dependencies": {
    "@babel/parser": "latest",
    "typescript": "latest"
  }
}
```

index.js
```js
var parser = require('@babel/parser');
var fs = require('fs');

var code = fs.readFileSync(require.resolve('typescript'), 'utf8');

var start = new Date().getTime();
var result = parser.parse(code, {
  plugins: ['typescript'],
  tokens: true
});
var spend = new Date().getTime() - start;
console.log('spend ' + (spend/1000).toFixed(2) + 's');
console.log('tokens length: ' + result.tokens.length);
```
We parse the typescript package main file (7.8m) to demonstrate the performance issue.
After `npm i`, run `node index.js`

On my machine:
```
spend 18.28s
tokens length: 897894
```

To compare the other two results:
1. remove `tokens: true`
```
spend 3.51s
```
2. or remove `plugins: ['typescript'],`
```
spend 2.70s
tokens length: 897894
```

## Why the huge difference?
Because babel-parser's typescript plugin uses `state.clone()` api quite few times. (flow plugin also uses the clone api)
But `State` class has a token store `state.tokens` to track all parsed tokens. This means everytime `state.clone()` is called, it duplicates the huge token store `state.tokens`.

This is not only a performance issue, it also causes memory spike.
On my mac, roughly read from activity monitor, the nodejs process goes up to around 800m vs 160m when tokens option is off).

## The fix
This fix is very simple, move up the token store from State to Tokenizer itself, so that the tokens aren't cloned in copied states. A cleanup logic is added to deal with typescript/flow's try-catch parsing branches.

This fix restored the running time of the code sample above back to 2~3 seconds range.


3cp:
There are additional few comments stores on state too. I will looking into them in next few days. I don't think state should hold those stores.

3cp:
@nicolo-ribaudo thx, I now see the duplicated tokens with my fix.

I don't quite understand. From what I saw, the token store is only populated when `options.tokens` is set to true, I only see mutation happen in `Tokenizer`. It means for majority use cases where `options.tokens` is false, the token store is always empty. **Parsing does not rely on it**.

There must be somewhere in typescript plugin to mutate the token store (only when it is not empty), but I cannot find relevant code. Can you point me to where the token store is touched by typescript plugin?

Also I see only typescript and flow plugins are using `tryParser()` (it clones state too) and `state.clone`. I can avoid the two plugins to bypass this issue for now.

Nevertheless, we need to find way to avoid cloning (shallowly currently) the huge token store array. This doesn't scale with input code size.

3cp:
@nicolo-ribaudo I see now what typescript plugin did.
It is not mutating the tokens directly, but keep the old state, try parse something, if it fails, restore state to old state, then try parse something else.
We can do better than this, but require bit of redesign to avoid expensive clone.

3cp:
@nicolo-ribaudo pls review the added tests (output.json generated by old master), and the tiny trick to maintain clean token store.

3cp:
I am not sure what's going on on CI. The whole units tests passed on my local machine.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -123,13 +127,25 @@ export default class Tokenizer extends LocationParser {
...
 
+  pushToken(token: Token | N.Comment) {
+    // Pop out invalid tokens trapped by try-catch parsing.
+    // Those parsing branches are mainly created by typescript and flow plugins.
+    while (
+      this.tokens.length > 0 &&
+      this.tokens[this.tokens.length - 1].end > token.start
+    ) {
+      this.tokens.pop();
+    }
```
nicolo-ribaudo: To make this faster, we can store `this.state.tokensLength: number`.
Then, when we `pushToken`, we first set `this.tokens.length = this.state.tokensLength`.
By doing so we can delete the wrong tokens in `O(1)` time rather than looping over the list.

3cp: I thought about that.

If some future plugin calls `this.state = oldState` directly without knowing to restore tokensLength, that will cause trouble.

Now I think I can hide the details in `get state()` and `set state(state)`, that requires no code changes in plugins.

nicolo-ribaudo: `state` is accessed a lot, I don't know how much a getter will impact performance.

> If some future plugin calls this.state = oldState directly without knowing to restore tokensLength, that will cause trouble.

`oldState` will already have the old `tokensLength`, so it would be automatically restored.

3cp: I just missed your example. Is it better to restore the length when set state, or just before push token?

nicolo-ribaudo: I think that it's ok to do it in `pushToken` since it will always be called.
My suggestion was to avoid looping over the existing tokens list and checking their `start`/`end` position.

3cp: I have made the suggested change.

nicolo-ribaudo:
Out of curiosity, could you run the benchmark you posted in your PR description using this PR, to see how much it improved performance?

3cp:
Good idea!
```
#master
spend 17.06s
tokens length: 897894

#patched
spend 3.27s
tokens length: 897894
```

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -123,13 +127,21 @@ export default class Tokenizer extends LocationParser {
     this.isLookahead = false;
   }
 
+  pushToken(token: Token | N.Comment) {
+    // Pop out invalid tokens trapped by try-catch parsing.
+    // Those parsing branches are mainly created by typescript and flow plugins.
+    this.tokens.length = this.state.tokensLength;
+    this.tokens.push(token);
```
JLHwung: I know the `array#splice` is always hacky, but since it is a frequent operation, can you simplify this to
```js
this.tokens.splice(this.state.tokensLength, Infinity, token)
```
It should be marginally faster since the engine don't have to reallocate memory if `this.tokens.length > this.state.tokensLength`

3cp:
I ran a rough benchmark for avg of 10 times.

`.length =` avg 2.55s
`.splice` avg 2.65s

There might be something else (my system load) to make the splice slight worse. I would say we leave the optimization out for no obvious benefit even for huge input (7.8m file).

packages/babel-parser/src/tokenizer/index.js
3cp: BTW, there would not be a situation for `this.tokens.length > this.state.tokensLength`.

PR #11031, opened 2020-01-19
fix: properly parse member expression after property initializer

### LINKED ISSUE #10989: Unexpected token when parsing a member expression after a property initializer in an object pattern
## Bug Report

**Current Behavior**

The following snippet will throw SyntaxError: Unexpected token, expected ","
**Input Code**
- REPL or Repo link if applicable: [REPL](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=BQbwBAhmC8YCwCYA0YBGAuMIC-A6AJmNjFtgJRA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=true&sourceType=script&lineWrap=false&presets=&prettier=false&targets=&version=7.8.0&externalPlugins=)

```js
({ a = 42, b: {}.d } = {})
```

**Expected behavior/code**
It should parse successfully.

**Babel Configuration (babel.config.js, .babelrc, package.json#babel, cli command, .eslintrc)**: No plugin.

**Environment**
REPL

**Possible Solution**
We may reset `refExpressionErrors.shortAssign` when parsing Assignment Expression after eating `tt.colon` in `parseObjectProperty`.

https://github.com/babel/babel/blob/bfb82e778fea5b766e18066a68f79ae3fbbab9bc/packages/babel-parser/src/parser/expression.js#L1732

vedantroy:
Fixes issue 10989 where only the identifier in a member expression that is the value of an object property would be parsed. Removing `checkExpressionErrors` in `parseExprSubscripts` results in the subscript also being parsed.

| Fixed Issues?            | Fixes #10989 

When parsing `({a = 42, b: test.d} = {})`, Babel expects a "," after `test`. This is because Babel only parses `test` instead of `test.d`. So it expects a comma after `b: test`, since it believes a new object property is starting. This happens because inside `parseExprSubscripts`, `this.checkExpressionErrors(refExpressionErrors, false)` returns `true`, which causes the method to return early and not call `this.parseSubscripts(expr, startPos, startLoc)`. Removing that check allows `parseSubscripts` to be called, resulting in `test.d` to be parsed as the object property value.


packages/babel-parser/src/parser/expression.js
```diff
@@ -532,10 +532,6 @@ export default class ExpressionParser extends LValParser {
       return expr;
     }
 
-    if (this.checkExpressionErrors(refExpressionErrors, false)) {
```
JLHwung: We should investigate if the other `this.checkExpressionErrors(refExpressionErrors, false)` checks is necessary. They are not caught by any test cases in babel-parser now.

(They are not even caught by acorn testcases)

vedantroy: I was thinking of removing those checks, but I wasn't sure if I should do that in this PR, since removing those checks wouldn't be necessary to fix this specific bug.

I will look into removing these checks.

vedantroy: I will look into removing the other instances of `this.checkExpressionErrors(refExpressionErrors, false)`. I didn't remove them in this PR, because it wasn't necessary to fix the bug, but I can always add some more changes to this PR/open a new one.

nicolo-ribaudo:
Thanks!

PR #11039, opened 2020-01-21
Update compat table

nicolo-ribaudo:
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Our data never actually included `"safari": "tp"`, so we never found the bug in the build script before :sweat_smile:

### DIFF
--- a/packages/babel-compat-data/data/plugins.json
+++ b/packages/babel-compat-data/data/plugins.json
@@ -263,7 +263,7 @@
   },
   "transform-parameters": {
     "chrome": "49",
-    "edge": "14",
+    "edge": "18",
     "firefox": "53",
     "safari": "10",
     "node": "6",
--- a/packages/babel-compat-data/data/plugins.json
+++ b/packages/babel-compat-data/data/plugins.json
@@ -7,6 +7,7 @@
   },
   "proposal-optional-chaining": {
     "chrome": "80",
+    "firefox": "74",
     "safari": "tp",
     "opera": "67"
   },
@@ -152,7 +153,7 @@
   },
   "transform-block-scoped-functions": {
     "chrome": "41",
-    "edge": "79",
+    "edge": "12",
     "firefox": "46",
     "safari": "10",
     "node": "4",
@@ -285,7 +286,7 @@
   },
   "transform-block-scoping": {
     "chrome": "49",
-    "edge": "79",
+    "edge": "14",
     "firefox": "51",
     "safari": "11",
     "node": "6",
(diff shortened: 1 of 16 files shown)


packages/babel-compat-data/data/plugins.json
```diff
@@ -275,7 +285,7 @@
   },
   "transform-block-scoping": {
     "chrome": "49",
-    "edge": "14",
+    "edge": "79",
```
nicolo-ribaudo: I should check why this changed

existentialism:
> Our data never actually included "safari": "tp"

You mean recently? We've definitely had `tp` in the past.

nicolo-ribaudo:
Yeah, I mean "after that line in the build was changed". `unreleasedLabelForEnv` is a string, so I don't see how returning it could have worked were we expect an object with `version` and `semver` properties.

packages/babel-compat-data/data/plugins.json
existentialism: I'm guessing https://github.com/kangax/compat-table/pull/1569?

nicolo-ribaudo: That commit didn't change the data files :thinking:

existentialism: We use essentially the same code as compat-table's HTML generation to interpolate versions via `environments.json`. As I said, was just a guess... I'll try to dig a bit further when I get home.

chicoxyzzy: It's because of [`equals` field](https://github.com/kangax/compat-table/pull/1569/files#diff-ca0f8d6af0c57ea63bddc9086afc0bc2R2214). Should be correct.

nicolo-ribaudo: Well, `edge14` still supports `let`/`const`. Maybe it's  bug in our script?

packages/babel-compat-data/data/plugins.json
```diff
@@ -253,7 +263,7 @@
   },
   "transform-parameters": {
     "chrome": "49",
-    "edge": "18",
+    "edge": "15",
```
nicolo-ribaudo: This is wrong

packages/babel-compat-data/data/plugins.json
nicolo-ribaudo: https://kangax.github.io/compat-table/es6/ has the same bug. Any version of Edge before 79 is marked as not supporting `let`/`const`.

nicolo-ribaudo:
I have fixed the `compat-table` bug. This is ready for review!

PR #11046, opened 2020-01-23
feat(babel-parser): throw syntax error for `}` and `>` in JSX text

### LINKED ISSUE #11042: Proposal: throw syntax error for `}` and `>` in JSX text.
## Feature Request

  - If needs be, I can look into it

The JSX spec [1][2] lists `}` and `>` as invalid JSX text characters.

> JSXTextCharacter :
> - SourceCharacter but not one of {, <, > or }

However, babel does not throw a syntax error when it encounters a naked `}` or `>` in a JSXText string.

According to the spec, the only way you should be able to include either of these characters (similarly for `{` and `<`), is via a string expression: `{'>'}`.

I propose that babel should follow the JSX spec, and throw a syntax error if it encounters a "naked" `}` or `>` within jsx text.

From my experience, it's uncommon that these characters are intentionally put in JSX text, and are instead the result of typos (i.e. typing `{}`, but having an editor auto-complete an extra `}` => `{}}`), or copy pasting errors.

Flow has already updated their parser to flag this as an error [3][4]. TypeScript is happy to make this change as well [5] (if Babel makes the change, as babel is generally considered the "reference implementation" for JSX).
I have also raised an issue in acorn for the same functionality [6].

- Custom lint rules
  - We had these implemented at facebook before flow made it a parsing error. (The lint rules actually spurred the change).
  - Outside of a tightly controlled monorepo, I can see that it would be hard to gain adoption for the rule, as not everyone uses linting tools, not everyone uses the react plugins, etc etc.
- Prettier auto-formatting to convert `}` to `{'}'}`.
  - This is an AST change, and prettier doesn't like to change the AST when formatting.

If you can, explain how users will be able to use this and possibly write out a version the docs.
Maybe a screenshot or design?

I think the best way to document this would be via clear and explicit error messaging.
For example, the following code:
```js
function F() {
  return <div>></div>;
}

function G() {
  return <div>{1}}</div>;
}
```
produces the following flow errors:
```
4:   return <div>></div>;
                 ^ Unexpected token `>`. Did you mean `{'>'}`?
8:   return <div>{1}}</div>;
                    ^ Unexpected token `}`. Did you mean `{'}'}`?
```
[flow.org repl](https://flow.org/try/#0MYewdgzgLgBASgUwIbFgXhgJwQRwK4CW2AFAOTYpSkCUA3AFD0BmeYqB4MAYsdTAN70YWBFDyYwMADwATAgDcAfIqkB6OUoYBfRizZQOkgOK8BQkWInSNi-gEYtWtTe1A)

---

Related:
[1] https://facebook.github.io/jsx/
[2] https://github.com/facebook/jsx/pull/18
[3] https://github.com/facebook/flow/blob/master/Changelog.md#01160
[4] facebook/flow@e1d0038
[5] https://github.com/microsoft/TypeScript/issues/36341
[6] https://github.com/acornjs/acorn-jsx/issues/106

bradzacher:
| Fixed Issues?            | Fixes #11042
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

See the full context in #11042.
The JSX spec lists `}` and `>` as invalid JSX text characters.
Flow recently updated their parser so that it errors in this case.
This PR just does the same for Babel.

Looking for feedback, as I've never contributed to babel before.
Do we need explicit documentation for this change, or are the error messages "good enough"?


packages/babel-parser/src/plugins/jsx/index.js
```diff
@@ -507,13 +507,35 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return this.jsxParseElementAt(startPos, startLoc);
     }
 
+    // Validates and parses the current state as JSXText
+
+    parseJSXText(): N.Expression {
+      const startPos = this.state.start;
+      for (let idx = 0; idx < this.state.value.length; idx += 1) {
+        const character = this.state.value[idx];
+        if (character === "}") {
```
kaicataldo: Do we also need to need to throw an error for `{` and `<`?

packages/babel-parser/src/plugins/jsx/index.js
```diff
@@ -507,13 +507,35 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+    }
+
     // ==================================
     // Overrides
     // ==================================
 
     parseExprAtom(refExpressionErrors: ?ExpressionErrors): N.Expression {
       if (this.match(tt.jsxText)) {
-        return this.parseLiteral(this.state.value, "JSXText");
+        return this.parseJSXText();
```
jridgewell: It seems I pointed you to the wrong location, sorry about that!

At this point, we've already tokenized the source into a `tt.jsxText` token (that's the `this.match()` check above). But we can do this earlier, when we're still creating the `jsxText` token. It's located at https://github.com/babel/babel/blob/1815c349877acca1d3275ea72512ee7824db49e3/packages/babel-parser/src/plugins/jsx/index.js#L81-L120.

packages/babel-parser/src/plugins/jsx/index.js
kaicataldo: Ah, ignore me - this already throws.

packages/babel-parser/src/plugins/jsx/index.js
bradzacher: ezpz I'll fix that up tomorrow morning.
thanks for the pointers!

bradzacher: ezpz I'll fix that up tomorrow morning.
thanks for the pointers!

bradzacher: ezpz I'll fix that up tomorrow morning.
thanks for the pointers!

packages/babel-parser/src/plugins/jsx/index.js
```diff
@@ -107,6 +107,15 @@ export default (superClass: Class<Parser>): Class<Parser> =>
             chunkStart = this.state.pos;
             break;
 
+          case charCodes.greaterThan:
+          case charCodes.rightCurlyBrace:
+            throw this.raise(
+              this.state.pos,
+              `Unexpected token \`${
+                this.input[this.state.pos]
+              }\`. Did you mean \`{'${this.input[this.state.pos]}'}\`?`,
```
jridgewell: Nit: we should encourage the HTML entities `&rbrace;` and `&gt;`, which shouldn't require the separate `JSXExpressionContainer`.

bradzacher: Would you be okay with the message suggesting both? i.e.
``Unexpected token `>`. Did you mean `{'>'}`, or `&gt;`?``

Personally I don't like seeing html entity codes in JSX.

jridgewell: Sure.

nicolo-ribaudo:
Thanks! It will take a bit to merge this PR since we need to wait for Babel 8.

packages/babel-parser/src/plugins/jsx/index.js
```diff
@@ -107,6 +107,20 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
 
+          case charCodes.greaterThan:
+          case charCodes.rightCurlyBrace: {
+            const htmlEntity =
+              ch === charCodes.rightCurlyBrace ? "&rbrace;" : "&gt;";
+            throw this.raise(
+              this.state.pos,
+              `Unexpected token \`${
+                this.input[this.state.pos]
+              }\`. Did you mean \`${htmlEntity}\`, or \`{'${
```
kaicataldo:
```suggestion
              }\`. Did you mean \`${htmlEntity}\` or \`{'${
```

Small thing I just noticed: Can we get rid of this extra comma in the error message?

nicolo-ribaudo:

Issue #9903, opened 2019-04-26
Cannot set core-js version for helper-only transformation in `@babel/tranform-runtime`


sodatea:
## Feature Request

**Is your feature request related to a problem? Please describe.**

Basically we are trying to resolve this issue in [@vue/babel-preset-app](https://github.com/vuejs/vue-cli/blob/v3.0.5/packages/%40vue/babel-preset-app):
> https://github.com/babel/babel/issues/6628
> How can we write ES20xx and ship the smallest amount of code (for a given target environment)?

To be more specific, this item:
> * babel helpers: reference the same helpers instead of recreating per file (either via [/packages/babel-plugin-transform-runtime@`7.0`#options](https://github.com/babel/babel/tree/7.0/packages/babel-plugin-transform-runtime#options)

- To achieve that we need to combine `@babel/preset-env` with `@babel/transform-runtime`, and the latter one is only used to do helper transformation.
- The default runtime has a blocking issue though, see https://github.com/babel/babel/issues/7597
- So to circumvent it we have to use `@babel/runtime-corejs2`. I know it's not the best solution but it's an acceptable compromise/tradeoff in practice.
- Before beta.56 we were able to disable `polyfills` option while still referring the helper functions to `@babel/runtime-corejs2`.
- But after [this commit](https://github.com/babel/babel/commit/3210a7a873d6aced7aa3313a8be5121d8c4e5072#diff-76ec2b062a79e02e8adc6218241c667e), `polyfills` option is removed because it is now the default behavior.
- An unintended consequence is that we can no longer point `helpers` to any other runtime package without introducing core-js polyfills along with it.

**Describe the solution you'd like**

Bring back the `polyfills` option, allowing the user to opt-out it while still being able to specify a core-js version for helpers.

**Describe alternatives you've considered**
N/A

**Teachability, Documentation, Adoption, Migration Strategy**
N/A

nicolo-ribaudo:
As a workaround, you can use https://github.com/tleunen/babel-plugin-module-resolver to alias `@babel/runtime` -> `@babel/runtime-corejs2`

nicolo-ribaudo:
@sodatea I have been thinking about this issue today, and I think that the proper solution would be to transpile `@babel/runtime`'s helpers using `@babel/plugin-transform-runtime` itself.
This is tricky, because we should aboud transpiling everything else except for polyfills, to avoid injecting references to themselves inside the helpers.
A working config might look like this:
```js
module.exports = {
  overrides: [{
[...]

Issue #2954, opened 2015-11-10
 (0 , _typeof3.default) is not a function


arshtepe:
this code in generated file
````javascript
/*!*******************************************!*\
  !*** ./~/babel-runtime/helpers/typeof.js ***!
  \*******************************************/
/***/ function(module, exports, __webpack_require__) {

	"use strict";
	
	var _typeof2 = __webpack_require__(/*! babel-runtime/helpers/typeof */ 88);
	
	var _typeof3 = _interopRequireDefault(_typeof2);
	
	function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
	
	var _Symbol = __webpack_require__(/*! babel-runtime/core-js/symbol */ 89)["default"];
	
	exports["default"] = function (obj) {
	  return obj && obj.constructor === _Symbol ? "symbol" : typeof obj === "undefined" ? "undefined" : (0, _typeof3.default)(obj);
	}; // this i have error: Uncaught TypeError: (0 , _typeof3.default) is not a function
	
	exports.__esModule = true;

/***/ },
````
If change this on code from source-map, it is work
```` javascript
  return obj && obj.constructor === _Symbol ? "symbol" : typeof obj;
````

also this error have in
core-js/library/modules/es6.symbol.js
babel-runtime/~/core-js/library/modules/$.is-object.js

**this error only with  plugins: ['transform-runtime']**

**Comment by ysmood** [comment]
Same problem with me, please fix it soon

**Comment by egoist** [comment]
Waiting for updates, temporarily I use babel-runtime@6.0.14 to avoid this bug.

**Comment by egoist** [comment]
I solved this with:

```bash
npm install babel-preset-es2015-node5 --save-dev
```
```json
{
  "presets": ["es2015-node5", "stage-0"],
  "plugins": ["transform-runtime"]
}
```

viciu:
Is it possible to add an automatic test to avoid this regression? Looks like it's coming back and forth.

Nickman87:
Same problem here, reverted to 6.0.14 but had to provide a temp. polyfill for JSON.stringify

astephenb:
+1

EricMCornelius:
+1

jeff-blaisdell:
:+1:

jamestalmage:
[...]

PR #11049, opened 2020-01-23
Avoid compiling the _typeof helper with itself

### LINKED ISSUE #8571: Name collision when re-processing transpiled code
## Bug Report

**Current Behavior**

This is a very confusing bug, and I want to say first that I’m not sure where the bug is (in Babel or elsewhere), so I am hoping for feedback in order to refine this bug report and possibly move it. (Is it more likely a Babel bug, a Webpack bug, a minifier bug?)

The problem: Our project uses https://www.npmjs.com/package/clipboard, which is published to NPM in transpiled form. As such, it contains the `_typeof` helper:

```js
 var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) {
        return typeof obj;
    } : function (obj) {
        return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj;
    };
```

After transformation via Webpack (4.17.1), Babel (6.26.3) `uglifyjs-webpack-plugin` or (currently) `terser-webpack-plugin` (1.0.2, `terser` 3.8.2), it seems that the `typeof Symbol.iterator` part has been transformed a second time, and resulted in a name collision. The minified code, reformatted for legibility:

```js
r = "function" == typeof Symbol && "symbol" === r(Symbol.iterator) ?
	function (e) {
		return void 0 === e ? "undefined" : r(e)
	} :
	function (e) {
		return e && "function" == typeof Symbol && e.constructor === Symbol && e !== Symbol.prototype ?
			"symbol" :
			void 0 === e ? "undefined" : r(e)
	}
```

Since *both* `typeof` instances were minified, presumably via the `_typeof` helper, to the same name (here: `r`), the expression fails to evaluate because the `r` in `r(Symbol.iterator)` isn’t initialized yet.

**Input Code**

Unfortunately, I have been unable to reproduce this in a minimal repro case. Importing the same library (`clipboard`) in a minimal sample and minifying it with the same options did not reproduce the behaviour.

```js
{
  "presets": ["babel-preset-env", {
    "browsers": [
	">5%",
	"IE >= 11",
	"last 2 Chrome versions",
	"last 2 ChromeAndroid versions",
	"last 2 Edge versions",
	"last 2 Firefox versions",
	"last 2 iOS versions"
    ]
  }],
  "plugins": [
	"babel-plugin-lodash",
	"babel-plugin-syntax-dynamic-import",
	"babel-plugin-transform-decorators-legacy",
	"babel-plugin-transform-class-properties",
	"babel-plugin-transform-flow-comments",
	"babel-plugin-transform-object-rest-spread",
	"babel-plugin-transform-react-jsx"
  ]
}
```

**Environment**
- Babel version(s): 6.26.3
- Node/npm version: Node 8.11.1/npm 6.4.0
- OS: Ubuntu 18.04
- Monorepo: No
- How you are using Babel: `babel-loader`/Webpack

### LINKED ISSUE #9127: Use of transform-runtime + proposal-decorators plugins causes Maximum call stack size exceeded
## Bug Report

**Current Behavior**
The use of `@babel/plugin-proposal-decorators` together with `@babel/plugin-transform-runtime` generates a bundle that causes a infinite loop (around `_typeof()` and `_typeof2()` eventually causing the browser to throw `Uncought RangeError: Maximum call stack size exceeded`.

If I remove `transform-runtime` from the configuration, the bundle generated does not emit any errors.

**Input Code**

```js
class Foo {
    @propDecorator
    get method1() {
        return  "one";
    }
}

function propDecorator(d) {
    return d;
}
```

**Expected behavior/code**
The transformed code should not cause an infinite loop

```js
{
    "test": {},
    "loader": "babel-loader",
    "options": {
        "babelrc": false,
        "presets": [
            [
                "@babel/preset-env",
                {
                    "modules": false,
                    "loose": false,
                    "forceAllTransforms": true
                }
            ]
        ],
        "plugins": [
            [
                "@babel/plugin-transform-runtime",
                {
                    "corejs": false,
                    "helpers": true,
                    "regenerator": true,
                    "useESModules": true
                }
            ],
            [
                {
                    "legacy": false,
                    "decoratorsBeforeExport": true
                }
            ]
        ]
    }
}
```

**Environment**
- Babel version(s): 7.2.0
- Node/npm version: Node v8.9.0, NPM v5.5.1
- OS: Windows
- Monorepo No
- How you are using Babel: Loader (webpack)

All packages:
```
    "@babel/core": "^7.2.0",
    "@babel/plugin-proposal-class-properties": "^7.2.1",
    "@babel/plugin-proposal-decorators": "^7.2.0",
    "@babel/plugin-transform-async-to-generator": "^7.2.0",
    "@babel/plugin-transform-runtime": "^7.2.0",
    "@babel/preset-env": "^7.2.0",
    "@babel/runtime": "^7.2.0",
    "babel-loader": "^8.0.4",
    "webpack": "^4.27.0",
    "webpack-cli": "^3.1.2"
```

Thank you in advance for your help/assistance.

### LINKED ISSUE #10996: TypeError: _typeof2 is not a function
var _typeof2 = __webpack_require__(/*! @babel/runtime/helpers/typeof */ "./node_modules/_@babel_runtime@7.7.7@@babel/runtime/helpers/typeof.js");

function _typeof(obj) {
  if (typeof Symbol === "function" && _typeof2(Symbol.iterator) === "symbol") {
    module.exports = _typeof = function _typeof(obj) {
      return _typeof2(obj);
    };
  } else {
    module.exports = _typeof = function _typeof(obj) {
      return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : _typeof2(obj);
    };
  }

  return _typeof(obj);
}

module.exports = _typeof;//# sourceURL=[module]
//# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJmaWxlIjoiLi9ub2RlX21vZHVsZXMvX0BiYWJlbF9ydW50aW1lQDcuNy43QEBiYWJlbC9ydW50aW1lL2hlbHBlcnMvdHlwZW9mLmpzLmpzIiwic291cmNlcyI6WyJ3ZWJwYWNrOi8vLy4vbm9kZV9tb2R1bGVzL19AYmFiZWxfcnVudGltZUA3LjcuN0BAYmFiZWwvcnVudGltZS9oZWxwZXJzL3R5cGVvZi5qcz9kYjVlIl0sInNvdXJjZXNDb250ZW50IjpbImZ1bmN0aW9uIF90eXBlb2Yob2JqKSB7XG4gIGlmICh0eXBlb2YgU3ltYm9sID09PSBcImZ1bmN0aW9uXCIgJiYgdHlwZW9mIFN5bWJvbC5pdGVyYXRvciA9PT0gXCJzeW1ib2xcIikge1xuICAgIG1vZHVsZS5leHBvcnRzID0gX3R5cGVvZiA9IGZ1bmN0aW9uIF90eXBlb2Yob2JqKSB7XG4gICAgICByZXR1cm4gdHlwZW9mIG9iajtcbiAgICB9O1xuICB9IGVsc2Uge1xuICAgIG1vZHVsZS5leHBvcnRzID0gX3R5cGVvZiA9IGZ1bmN0aW9uIF90eXBlb2Yob2JqKSB7XG4gICAgICByZXR1cm4gb2JqICYmIHR5cGVvZiBTeW1ib2wgPT09IFwiZnVuY3Rpb25cIiAmJiBvYmouY29uc3RydWN0b3IgPT09IFN5bWJvbCAmJiBvYmogIT09IFN5bWJvbC5wcm90b3R5cGUgPyBcInN5bWJvbFwiIDogdHlwZW9mIG9iajtcbiAgICB9O1xuICB9XG5cbiAgcmV0dXJuIF90eXBlb2Yob2JqKTtcbn1cblxubW9kdWxlLmV4cG9ydHMgPSBfdHlwZW9mOyJdLCJtYXBwaW5ncyI6Ijs7Ozs7Ozs7Ozs7Ozs7OztBQUFBO0FBQ0E7QUFDQTtBQUNBO0FBQ0E7QUFDQTtBQUNBO0FBQ0E7QUFDQTtBQUNBO0FBQ0E7QUFDQTtBQUNBO0FBQ0E7QUFDQSIsInNvdXJjZVJvb3QiOiIifQ==

### LINKED ISSUE #11036: `TypeError: (0 , _typeof3.default) is not a function` when executing test inside sub-package of monorepo
## Bug Report

**Current Behavior**
Executing tests written in ES6+ inside sub-package of monorepo results in error: `TypeError: (0 , _typeof3.default) is not a function`

**Input Code**
Repro repo: https://github.com/soryy708/typeof3.default-repro

**Expected behavior/code**
Executed tests without errors

- Filename: `babel.config.js`

```js
module.exports = {
    presets: [
        ['@babel/preset-env', {
            modules: 'commonjs'
        }]
    ],
    plugins: [
        '@babel/transform-runtime'
    ],
    ignore: [
        'node_modules'
    ]
};
```

**Environment**

```
npx: installed 1 in 2.935s

  System:
    OS: Windows 10 10.0.18362
  Binaries:
    Node: 12.13.0 - C:\Program Files\nodejs\node.EXE
    npm: 6.12.0 - C:\Program Files\nodejs\npm.CMD
```
- Babel version(s): 7.8.3
- Node/npm version: Node11.13 npm6.12
- OS: Windows 10 x64 Home
- Monorepo: Yes
- How you are using Babel: register

**Possible Solution**
 ¯\\_(ツ)_/¯

Perhaps related?

* https://swizec.com/blog/dont-worry-its-a-compiler-bug/swizec/6764
* https://github.com/babel/babel/issues/2954
* https://github.com/babel/babel/issues/3737
* https://github.com/babel/babel/issues/10996

### LINKED ISSUE #11047: Disable Babel in the @babel/runtime package.
## Feature Request

- [ ] I would like to work on this feature!

The combination of Webpack with Babel may cause a problem such as #10996 if the `@babel/runtime` package is not explicitly excluded from Babel. This package contains helper functions which must not be transpiled by Babel. However, if care is not taken to exclude, Webpack will apply Babel to `@babel/runtime` resulting in a broken bundle. 

For example, the Babel runtime contains helper function `_typeof(obj)` which replaces the `typeof` operator to recognize `Symbol`s. This function contains few typeof operators internally, and when transpiled by Babel these operators are replaced with the same helper function. Whoops!

If Babel contains any exclusion mechanism, it must be applied to package `@babel/runtime` to prevent the situations as the above.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #11047, fixes #10996, fixes #8571, fixes #9127, fixes #11036, ref #2954, ref #3737
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

We already had some logic to prevent the `typeof-symbol` plugin from transpiling its own helper. It worked by generating the helper name and checking if the current path is inside a function or a variable declaration with that name.

However, it had a few problems:
1. It doesn't work with `@babel/runtime`, because the generated name is a unique identifier so it's guaranteed _not_ to be equal to the `_typeof` helper name.
1. Sometimes, multiple names are generated (`_typeof2` and `_typeof3`), causing circular dependencies. This is probably caused by running twice the plugin, but we can avoid failing in that case.

I initially hard-coded the helper name in the `isUnderHelper` logic to make it work with `@babel/runtime`, but:
1. It doesn't fix (2)
1. If `@babel/runtime` is bundled in an app/library by `rollup`, the function name might change
1. It doesn't work with minifiers.

I then decided to introduce a `"@babel/helpers - typeof"` directive, which has a few advantage:
1. We don't need to generate the helper code _before_ checking if we are inside the helper, which would always lead to duplicated code when transpiling `@babel/runtime/helpers/typeof`
1. It fixes (2) and works with Rollup.

The directive still doesn't work with Terser and UglifyJS because they strip them by default. We could either suggest to whoever finds this bug to set the `directives: false` option in their minifier settings, or we could ask UglifyJS and Terser's maintainers if they could keep this specific directive in the minified output.

This PR makes it safe to transpile `@babel/runtime`, which should provide a solution for #9903

### DIFF
--- a/packages/babel-plugin-transform-typeof-symbol/test/helper.spec.js
+++ b/packages/babel-plugin-transform-typeof-symbol/test/helper.spec.js
@@ -1,0 +1,52 @@
+import * as babel from "@babel/core";
+import resolvePath from "resolve";
+import fs from "fs";
+
+import transformTypeofSymbol from "..";
+
+const resolve = path =>
+  new Promise((resolve, reject) =>
+    resolvePath(path, (err, path) => (err ? reject(err) : resolve(path))),
+  );
+const readFile = path =>
+  new Promise((resolve, reject) =>
+    fs.readFile(path, "utf8", (err, contents) => {
+      if (err) reject(err);
+      else resolve(contents);
+    }),
+  );
+
+describe("@babel/plugin-transform-typeof-symbol", () => {
+  test.each`
+    runtime                     | type
+    ${"@babel/runtime"}         | ${"esm"}
+    ${"@babel/runtime"}         | ${"cjs"}
+    ${"@babel/runtime-corejs2"} | ${"esm"}
+    ${"@babel/runtime-corejs2"} | ${"cjs"}
+    ${"@babel/runtime-corejs3"} | ${"esm"}
+    ${"@babel/runtime-corejs3"} | ${"cjs"}
+  `(
+    "shouldn't transpile the $type $runtime helper",
+    async ({ type, runtime }) => {
+      const path = await resolve(
+        `${runtime}/helpers${type === "esm" ? "/esm/" : "/"}typeof`,
+      );
+      const src = await readFile(path);
+
+      const ast = babel.parseSync(src, {
+        configFile: false,
+        sourceType: type === "esm" ? "module" : "script",
+      });
+
+      const withPlugin = babel.transformFromAstSync(ast, src, {
+        configFile: false,
+        plugins: [transformTypeofSymbol],
+      });
+      const withoutPlugin = babel.transformFromAstSync(ast, src, {
+        configFile: false,
+      });
+
+      expect(withPlugin.code).toBe(withoutPlugin.code);
+    },
+  );
+});
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -12,6 +12,8 @@ const helper = (minVersion: string) => tpl => ({
 
 helpers.typeof = helper("7.0.0-beta.0")`
   export default function _typeof(obj) {
+    "@babel/helpers - typeof";
+
     if (typeof Symbol === "function" && typeof Symbol.iterator === "symbol") {
       _typeof = function (obj) { return typeof obj; };
     } else {
--- a/packages/babel-plugin-proposal-class-properties/test/fixtures/regression/6154/output.js
+++ b/packages/babel-plugin-proposal-class-properties/test/fixtures/regression/6154/output.js
@@ -1,4 +1,4 @@
-function _typeof(obj) { if (typeof Symbol === "function" && typeof Symbol.iterator === "symbol") { _typeof = function _typeof(obj) { return typeof obj; }; } else { _typeof = function _typeof(obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; }; } return _typeof(obj); }
+function _typeof(obj) { "@babel/helpers - typeof"; if (typeof Symbol === "function" && typeof Symbol.iterator === "symbol") { _typeof = function _typeof(obj) { return typeof obj; }; } else { _typeof = function _typeof(obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; }; } return _typeof(obj); }
 
 function _possibleConstructorReturn(self, call) { if (call && (_typeof(call) === "object" || typeof call === "function")) { return call; } return _assertThisInitialized(self); }
 
--- a/packages/babel-plugin-proposal-decorators/test/fixtures/legacy-regression/7030/output.js
+++ b/packages/babel-plugin-proposal-decorators/test/fixtures/legacy-regression/7030/output.js
@@ -1,4 +1,4 @@
-function _typeof(obj) { if (typeof Symbol === "function" && typeof Symbol.iterator === "symbol") { _typeof = function _typeof(obj) { return typeof obj; }; } else { _typeof = function _typeof(obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; }; } return _typeof(obj); }
+function _typeof(obj) { "@babel/helpers - typeof"; if (typeof Symbol === "function" && typeof Symbol.iterator === "symbol") { _typeof = function _typeof(obj) { return typeof obj; }; } else { _typeof = function _typeof(obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; }; } return _typeof(obj); }
 
 function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }
 
--- a/packages/babel-plugin-transform-classes/test/fixtures/regression/T7537/output.js
+++ b/packages/babel-plugin-transform-classes/test/fixtures/regression/T7537/output.js
@@ -1,4 +1,4 @@
-function _typeof(obj) { if (typeof Symbol === "function" && typeof Symbol.iterator === "symbol") { _typeof = function _typeof(obj) { return typeof obj; }; } else { _typeof = function _typeof(obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; }; } return _typeof(obj); }
+function _typeof(obj) { "@babel/helpers - typeof"; if (typeof Symbol === "function" && typeof Symbol.iterator === "symbol") { _typeof = function _typeof(obj) { return typeof obj; }; } else { _typeof = function _typeof(obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; }; } return _typeof(obj); }
 
 function _possibleConstructorReturn(self, call) { if (call && (_typeof(call) === "object" || typeof call === "function")) { return call; } return _assertThisInitialized(self); }
 
--- a/packages/babel-plugin-transform-parameters/test/fixtures/regression/6057-expanded/output.js
+++ b/packages/babel-plugin-transform-parameters/test/fixtures/regression/6057-expanded/output.js
@@ -9,7 +9,7 @@ var _args = _interopRequireDefault(require("utils/url/args"));
 
 function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { "default": obj }; }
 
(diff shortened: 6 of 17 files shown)


packages/babel-plugin-transform-typeof-symbol/test/helper.spec.js
```diff
@@ -0,0 +1,52 @@
...
+import fs from "fs";
+
+import transformTypeofSymbol from "..";
+
+const resolve = path =>
+  new Promise((resolve, reject) =>
+    resolvePath(path, (err, path) => (err ? reject(err) : resolve(path))),
+  );
+const readFile = path =>
+  new Promise((resolve, reject) =>
```
soryy708: Can be refactored using [`util.promisify`](https://nodejs.org/api/util.html#util_util_promisify_original)
Example:

    const readFile = path => util.promisify(fs.readFile)(path, "utf8")

nicolo-ribaudo: No, it's not supported by Node 6.
When we'll drop it, we'll directly use `fs.promises`.

nicolo-ribaudo:
If the file is already bundled, then Babel can't reference it and wouldn't introduce a circular dependency but only duplicate the helper. Anyway, I don't think that bundlers strip directives, especially if they are inside a nested function.

soryy708:
Don't assume, verify.

nicolo-ribaudo:
[Rollup doesn't](https://rollupjs.org/repl/?version=1.29.1&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMGZuJTIwZnJvbSUyMCU1QyUyMi4lMkZmaWxlLmpzJTVDJTIyJTNCJTVDbiU1Q25mbigpJTNCJTIyJTJDJTIyaXNFbnRyeSUyMiUzQXRydWUlN0QlMkMlN0IlMjJuYW1lJTIyJTNBJTIyZmlsZS5qcyUyMiUyQyUyMmNvZGUlMjIlM0ElMjJleHBvcnQlMjBkZWZhdWx0JTIwZnVuY3Rpb24lMjBmbigpJTIwJTdCJTVDbiU1Q3QlNUMlMjJkaXJlY3RpdmUlNUMlMjIlM0IlNUNuJTVDdGNvbnNvbGUubG9nKDEpJTNCJTVDbiU3RCUyMiUyQyUyMmlzRW50cnklMjIlM0FmYWxzZSU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJlc20lMjIlMkMlMjJuYW1lJTIyJTNBJTIybXlCdW5kbGUlMjIlMkMlMjJhbWQlMjIlM0ElN0IlMjJpZCUyMiUzQSUyMiUyMiU3RCUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlN0QlMkMlMjJleGFtcGxlJTIyJTNBbnVsbCU3RA==), [Parcel doesn't](https://codesandbox.io/s/parcel-demo-pliyc?fontsize=14&hidenavigation=1&theme=dark), [Webpack doesn't](https://codesandbox.io/s/webpack-starter-g5qj4?fontsize=14&hidenavigation=1&theme=dark).

davixyz:
Hi! when is this getting released?

nicolo-ribaudo:
Tomorrow probably :grin:

PR #11053, opened 2020-01-23
Import type for TS

### LINKED ISSUE #10981: `import type` in TypeScript

  ```ts
  ```

  ```ts

  let r: typeof RegExp;

  new RegExp()
  ```

See more at

hzoo:
| Fixed Issues?            | fixes #10981
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | Yea
| Documentation PR Link    | 
| Any Dependency Changes?  |

Implemented for import type, still need to implement export


packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -1945,7 +1967,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
             member.start,
             `Index signatures cannot have an accessibility modifier ('${
               (member: any).accessibility
-            }')`,
+            } ')`,
```
existentialism:
```suggestion
            }')`,
```

nicolo-ribaudo:
```suggestion
            }')`,
```

nicolo-ribaudo:
I have changed the base branch to `feat-7.9.0/ts-import-type` so that we can merge it there and work on `export type` and the transform plugin option.

nicolo-ribaudo:
:warning: This PR is not merged to master yet

Issue #10675, opened 2019-11-07
Parsing error for generic function named 'of'


alexwalterbos:
Thanks for the great tool guys! I found a parsing bug regarding generic function definitions and specific naming:

## Bug Report

**Current Behavior**
Babel-eslint parses a generic function just fine:
```JS
function someGenericFunction<R>(args: R): Array<R> {
  return [args];
}
```

It fails to parse that same function when it is named `of`, however:
```JS
function of<R>(args: R): Array<R> {
  return [args];
}
```
```
/path/to/index.js
  5:12  error  Parsing error: Unexpected token, expected "("

  3 | }
  4 | 
> 5 | function of<R>(args: R): Array<R> {
    |            ^
  6 | 	return [args];
  7 | }
  8 | 

✖ 1 problem (1 error, 0 warnings)

error Command failed with exit code 1.
```
Flow accepts the function just fine and shows 'No errors!'. It confused me for longer than I'd like to admit, hopefully fixing this will save others some time.
It would seem that `of` is treated as a keyword - maybe because of the [`for...of` syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of)?

**Input Code**
Reproduction repository: https://github.com/alexwalterbos/generic-function-of

**Expected behavior/code**
Parse the generic function named `of` just like any other generic function.

**Babel Configuration (.babelrc, package.json, cli command)**

```js
// package.json
{
  "devDependencies": {
    "babel-eslint": "^10.0.3",
    "eslint": "^6.6.0"
  }
}

// .eslintrc.js
module.exports = {
    "parser": "babel-eslint",
    "env": {
        "browser": true,
        "es6": true
    },
    "extends": [
        "eslint:recommended",
    ],
    "globals": {
        "Atomics": "readonly",
        "SharedArrayBuffer": "readonly"
    },
    "parserOptions": {
        "ecmaVersion": 2018,
        "sourceType": "module"
    },
    "rules": {
    }
};
```
Error is triggered simply with these commands in the reproduction repository linked above;
```
yarn install
yarn eslint .
```

**Environment**
- Babel version(s): v7.7.0 (`@babel/parser`)
- Node/npm version: Node v12.13.0, yarn v1.19.1
- OS: Fedora 29 x64
- Monorepo: no
- How you are using Babel: cli

**Workaround**
Don't name your generic functions `of`.

rajasekarm:
I can pick this.

rajasekarm:
[...]

PR #11055, opened 2020-01-24
Fix parse error for generics function named of

rajasekarm:
| Fixed Issues?            | https://github.com/babel/babel/issues/10675 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/src/plugins/jsx/index.js
+++ b/packages/babel-parser/src/plugins/jsx/index.js
@@ -559,7 +559,9 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       if (
         code === charCodes.lessThan &&
         this.state.exprAllowed &&
-        this.input.charCodeAt(this.state.pos + 1) !== charCodes.exclamationMark
+        this.input.charCodeAt(this.state.pos + 1) !==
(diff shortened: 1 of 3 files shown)


packages/babel-parser/src/plugins/jsx/index.js
```diff
@@ -559,7 +559,9 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       if (
         code === charCodes.lessThan &&
         this.state.exprAllowed &&
-        this.input.charCodeAt(this.state.pos + 1) !== charCodes.exclamationMark
+        this.input.charCodeAt(this.state.pos + 1) !==
+          charCodes.exclamationMark &&
+        context !== tc.functionStatement
```
nicolo-ribaudo: Also `functionExpression`.
However, I think that it would be better to set `this.state.exprAllowed` to `false` when we find `of` after `function`, rather than working it around here.

rajasekarm: I'll check this.

PR #11063, opened 2020-01-28
[@babel/cli] Copy ignored files by default

### LINKED ISSUE #11052: v7.8.0 introduces a non-backwards-compatible change
Babel design docs claim that [Babel follows semver](https://github.com/babel/babel/blob/master/doc/design/versioning.md). However, Semantic Versioning claims that non-backwards-compatible bug fixes should be released as major versions.

`v7.8.0` included this PR: https://github.com/babel/babel/pull/10887. This PR changes the behavior of the public API, namely `--ignore` combined with `--copy-files`.

`^v7.8.0` breaks all of our apps, and we have Dependabot configured to auto-merge minor/patch bumps, so this has been rather disruptive.

From reading through the history of #6226, it seems like it was pretty well established that this would be a breaking change. It seems like the idea of this even being a bug [was contentious](https://github.com/babel/babel/issues/6226#issuecomment-328660064).

Is my understanding of semver out of line? The way I see it, to maintain previous behavior when upgrading to 7.8.0, I need to add `--copy-ignored`. If I need to make a change to maintain previous behavior, that's a breaking change.

Please consider either:
1. reverting this change and including it in the v8 line instead
2. changing the default for the new `--copy-ignored` flag to `true`, to maintain previous behavior (and possibly changing the default to `false` in the v8 line)

(Note: I didn't submit this as a bug report and I didn't include repro steps because everything seems to be working as expected. The only issue is that this is a breaking change published as a minor. I can provide repro steps if necessary.)

nicolo-ribaudo:
| Fixed Issues?            | Fixes #11052
| Major: Breaking Change?  | 
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This commit restores the pre-7.8.0 behavior of the `--copy-files` option, by making `--copy-ignored` default to true when `--copy-files` is enabled.

@serhalp Could you check if the `--copy-files` tests in https://github.com/nicolo-ribaudo/babel/tree/regression-7.8-cli/packages/babel-cli/test/fixtures/babel match your expectations?

@STRML I didn't take a look at the other bug you mentioned in #11052


nicolo-ribaudo:
@kaicataldo We also have other `--no-` options. Also, I think that it's quite standard for boolean flags that default to `true`, since `commander` automatically handles the `--no-` prefix as such.

STRML:
Great. We can fix separately. 👍

PR #11065, opened 2020-01-29
Add TypeScript declarations to scope tracker

devongovett:
TypeScript declarations were missing from Babel's scope tracking due to missing config in `getBindingIdentifiers`. This adds the missing identifiers and adds tests to ensure TypeScript declarations are accessible via the scope.

### DIFF
--- a/packages/babel-traverse/test/scope.js
+++ b/packages/babel-traverse/test/scope.js
@@ -188,6 +188,64 @@ describe("scope", () => {
       ).toBe("ImportSpecifier");
     });
 
+    it("typescript declare function", function() {
+      expect(
+        getPath("declare function test(): void", {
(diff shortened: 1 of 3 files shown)


nicolo-ribaudo:
All the failing tests make me think that it would be better to defer this to Babel 8 :thinking:

devongovett:
Oops. Hmm, a lot of these look like strange TypeScript-isms to me. e.g. TS [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html). Not sure how Babel's scope tracker could handle those. The bindings are a 1-1 mapping of names to nodes right now, and TS breaks that assumption. I suppose we could only point to the first one for the short term to avoid scope collision errors?

PR #11068, opened 2020-01-29
Fix classNameTDZError in computed prototype methods with class fields 

### LINKED ISSUE #10766: classNameTDZError in computed prototype methods with class fields
## Bug Report

**Input Code**

```js
class Foo {
  static nickname = 'Tom';

  ['HELLO']() {
    console.log('>>>>', Foo);
  }
}
```

**Output code with babel**

```js
"use strict";

function _classNameTDZError(name) { throw new Error("Class \"" + name + "\" cannot be referenced in computed property keys."); }

class Foo {
  ['HELLO']() {
    console.log('>>>>', (_classNameTDZError("Foo"), Foo));
  }

}

Foo.nickname = 'Tom';
```

However, when I put the `static nickname = 'Tom';` out of the class,  like `Foo.nickname = 'Tom';` or just remove this line, the output code is correct

- Filename: `.babelrc`

```json
{
  "presets": [
    ["@babel/preset-env", {
      "targets": {
        "browsers": ["Chrome >= 65", "Firefox >= 55", "Safari >= 11"]
      },
    }],
  ],
  "plugins": [
    ["@babel/plugin-proposal-class-properties", { "loose": true }],
  ]
}
```

**Environment**

```
- Babel version(s): 7.6.4
- Node/npm version: 12.13.0 / 6.12.0
- OS: OSX 10.14.6

### PR DESCRIPTION
**Pull request by sidntrivedi012**
| Fixed Issues?            | Fixes #10766  
| Documentation PR Link    | 
| Any Dependency Changes?  |

### CODE REVIEW

File: packages/babel-helper-create-class-features-plugin/src/misc.js [hunk 1]
```diff
@@ -86,10 +83,17 @@ export function extractComputedKeys(ref, path, computedPaths, file) {
...
+      classFieldDefinitionEvaluationTDZVisitor.ReferencedIdentifier.call(
+        computedPath.get("key"),
+        {
+          classBinding:
+            path.node.id && path.scope.getBinding(path.node.id.name),
+          file,
+        },
+      );
+    }
+    computedPath.get("key").traverse(classFieldDefinitionEvaluationTDZVisitor);
```
nicolo-ribaudo: Here you should still pass the state object (the second parameter)

packages/babel-helper-create-class-features-plugin/test/fixtures/plugin-proposal-class-properties/tdz-error-computed-prototypes/options.json
```diff
@@ -0,0 +1,10 @@
+{
+  "plugins": [
+    [
+      "proposal-class-properties",
+      {
+        "loose": true
```
nicolo-ribaudo: Can you also add a test in non-loose mode?

sidntrivedi012: Added them. PTAL.

packages/babel-helper-create-class-features-plugin/src/misc.js
sidntrivedi012: Done. PTAL.

sidntrivedi012:
@nicolo-ribaudo Have extracted the method `handleClassTDZ` from `classFieldDefinitionEvaluationTDZVisitor`. 

But, `this.classBinding` throws error : `Cannot read classBinding of undefined.`

nicolo-ribaudo:
Instead of `this`, you can try to use the second parameter

sidntrivedi012:
Done. PTAL.

packages/babel-helper-create-class-features-plugin/src/misc.js
```diff
@@ -84,9 +82,14 @@ export function injectInitialization(path, constructor, nodes, renamer) {
...
+    if (computedPath.get("key").isReferencedIdentifier()) {
+      handleClassTDZ(computedPath.get("key"), {
+        classBinding: path.node.id && path.scope.getBinding(path.node.id.name),
+        file,
+      });
+    }
+    computedPath.get("key").traverse(classFieldDefinitionEvaluationTDZVisitor, {
       classBinding: path.node.id && path.scope.getBinding(path.node.id.name),
       file,
     });
```
nicolo-ribaudo: Instead of rewriting the object, you can extract it to a variable.

sidntrivedi012: Done. PTAL.

packages/babel-helper-create-class-features-plugin/src/misc.js
```diff
@@ -25,26 +25,24 @@ const referenceVisitor = {
...
-        const throwNode = t.callExpression(classNameTDZError, [
-          t.stringLiteral(path.node.name),
-        ]);
-
-        path.replaceWith(t.sequenceExpression([throwNode, path.node]));
-        path.skip();
-      }
-    },
-  },
-  environmentVisitor,
```
nicolo-ribaudo: Context for reviewers: `environmentVisitor` wasn't actually used. It defines a visitor for `Method` so that it can skip all its children except for the computed key, but since we were calling `methodPath.traverse` that visitor was never called.

sidntrivedi012: @nicolo-ribaudo Should I add this as a comment in the code as well? Can help future contributors as a reference.

nicolo-ribaudo: Maybe we could add a comment in the file where `environmentVisitor` is defined, explained that it should be used when traversing the whole class and not for specific class elements.

sidntrivedi012: @nicolo-ribaudo Done. PTAL.

packages/babel-helper-create-class-features-plugin/src/misc.js
```diff
@@ -84,12 +82,17 @@ export function injectInitialization(path, constructor, nodes, renamer) {
...
+  };
   for (const computedPath of computedPaths) {
-    computedPath.traverse(classFieldDefinitionEvaluationTDZVisitor, {
-      classBinding: path.node.id && path.scope.getBinding(path.node.id.name),
-      file,
-    });
+    if (computedPath.get("key").isReferencedIdentifier()) {
+      handleClassTDZ(computedPath.get("key"), state);
+    }
+    computedPath
```
nicolo-ribaudo: Nit: we can use `else` here, since an identifier doesn't have any child anyway.

sidntrivedi012: Done. Please review :)

packages/babel-helper-create-class-features-plugin/src/misc.js
```diff
@@ -84,12 +82,18 @@ export function injectInitialization(path, constructor, nodes, renamer) {
...
+  const state = {
+    classBinding: path.node.id && path.scope.getBinding(path.node.id.name),
+    file,
+  };
   for (const computedPath of computedPaths) {
-    computedPath.traverse(classFieldDefinitionEvaluationTDZVisitor, {
-      classBinding: path.node.id && path.scope.getBinding(path.node.id.name),
-      file,
-    });
+    if (computedPath.get("key").isReferencedIdentifier()) {
```
JLHwung: nit: can we cache `computedPath.get("key")`? It is also used in `!computedPath.get("key").isConstantExpression()` later.

sidntrivedi012: @JLHwung Done. PTAL 🙂

PR #11069, opened 2020-01-29
Update deps and babel-eslint-* package.jsons

kaicataldo:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Noticed that these need to be updated!


kaicataldo:
Whoops, looks like I'll need to update some tests

kaicataldo:
Hmmm actually I don't think we want to do this yet. I'm having trouble getting `@babel/eslint-parser` to be installed at the top level. It looks like https://github.com/typescript-eslint/typescript-eslint does this, so we'll have to look into how they do it.

kaicataldo:
Moved these changes into https://github.com/babel/babel/pull/11070 because they're interdependent.

PR #11070, opened 2020-01-30
Clean babel-eslint-*/**/lib

kaicataldo:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

I was having trouble running `make lint` locally because I had an old `babel-eslint-config-internal/lib` that was being used. Since the main file in question isn't in `src`, it wasn't getting recompiled and I was therefore using the old compiled output.

I don't _think_ we need to avoid cleaning the ESLint packages now that they're all using `src` if they're compiled` and are not in `src` if not.

### DIFF
--- a/eslint/babel-eslint-config-internal/package.json
+++ b/eslint/babel-eslint-config-internal/package.json
@@ -7,7 +7,7 @@
   "license": "MIT",
   "private": true,
   "repository": "https://github.com/babel/babel/tree/master/eslint/babel-eslint-config-internal",
-  "main": "lib/index.js",
+  "main": "index.js",
   "peerDependencies": {
     "babel-eslint": "^10.0.0 || ^11.0.0-0",
     "eslint-plugin-flowtype": "^3.0.0"
--- a/Makefile
+++ b/Makefile
@@ -268,10 +268,8 @@ bootstrap: bootstrap-only
 	$(MAKE) build
 
 clean-lib:
-	# TODO: Don't delete eslint/*/lib when they use src
 	$(foreach source, $(SOURCES), \
-		$(if $(filter-out $(source), eslint), \
-			$(call clean-source-lib, $(source))))
+		$(call clean-source-lib, $(source)))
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -12,7 +12,9 @@
   "devDependencies": {
     "@babel/cli": "^7.7.0",
     "@babel/core": "^7.7.2",
-    "@babel/eslint-plugin-development": "^1.0.1",
+    "@babel/eslint-config-internal": "./eslint/babel-eslint-config-internal",
```
kaicataldo: Is this the best way to use these internal packages for linting before we publish them? Or is there another way?

kaicataldo: Hmmm okay, so this doesn't actually work because it only copies in `src`. Will have to keep digging.

kaicataldo:
Once we figure this out I'll start working on ensuring ESLint's core rules work!

kaicataldo:
Okay, so I've tried a number of different things to try to get the local packages hoisted, including using Yarn workspaces and trying to install from the package directory itself. I think I'm going to give up on this and just wait until we publish these packages. 

We should still merge this though, since `babel-eslint-config-internal` does not generate a `lib/index.js` file any more!

eslint/babel-eslint-config-internal/package.json
```diff
@@ -7,7 +7,7 @@
   "license": "MIT",
   "private": true,
   "repository": "https://github.com/babel/babel/tree/master/eslint/babel-eslint-config-internal",
-  "main": "lib/index.js",
```
kaicataldo: This file doesn't exist because we don't compile this file.

nicolo-ribaudo:
> Okay, so I've tried a number of different things to try to get the local packages hoisted, including using Yarn workspaces and trying to install from the package directory itself. I think I'm going to give up on this and just wait until we publish these packages.

I'll try to make it work tomorrow.

PR #11074, opened 2020-01-30
Fix export bindings not updated by 'for ... in' and 'for ... of'

### LINKED ISSUE #10941: plugin-transform-modules-commonjs: export binding not updated by 'for...in' and 'for...of'
## Bug Report

**Current Behavior**
If an exported binding is used as LHS in a ForOfStatement or ForInStatement, only the local variable is updated. The exported binding stays the same.

**Input Code**
- REPL or Repo link if applicable: https://babeljs.io/en/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=KYDwDg9gTgLgBAG2PAZhCcC8cDkBLAOzxjwEMEcBuAKGrSjgAo0NC4BvMKCMALjgCMAXwCUHIbXpMWcCCjgBtHMCQBbYARg4AumPYS60JuxZDZ8hSfT8cAE2ABnGFACuAYxguowWziG7xWiA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2016&prettier=false&targets=&version=7.7.4&externalPlugins=%40babel%2Fplugin-transform-modules-commonjs%407.7.5

```js
export let foo = 'initial';

for (foo in {prop: 1}) {}

for (foo of ['element']) {}

for ({foo} of [{foo: 'destructured'}]) {}
```

**Expected behavior/code**
`exports.foo` should be updated every time `foo` is assigned a new value.
I tested the code without transpilation in Node.js and it works as expected.

vedantroy:
| Fixed Issues?            | Fixes #10941
| Major: Breaking Change?  | No?
| Documentation PR Link    | 

This PR makes Babel correctly transform exported variables in `for of` and `for in` loops.


packages/babel-plugin-transform-modules-commonjs/test/fixtures/misc/for-of-in-export/output.js
```diff
@@ -0,0 +1,62 @@
...
+
+Object.defineProperty(exports, "__esModule", {
+  value: true
+});
+exports.bar = exports.foo = void 0;
+let foo;
+exports.bar = exports.foo = foo;
+
+for (let _foo of []) {
+  exports.bar = exports.foo = foo = _foo
```
nicolo-ribaudo: Why is this reassigning `bar`?

vedantroy: Because the original input has `export {foo as bar}`.

I don't manually build the export expressions, I use `buildBindingExportAssignmentExpression`.

nicolo-ribaudo: Oh ok :+1:

nicolo-ribaudo:
Btw, you can use `t.getOuterBindingIdentifiers(left)` to get all the bindings updated by the destructuring pattern, and then rename the bindings declared in the nested scope. By doing so, you don't need the custom `transformArrayPattern`.

vedantroy:
@nicolo-ribaudo I think moving `PATTERN` into the loop body would break the following case:
```
for ({foo} of [{foo: 1}]) {
  let foo;
}
```
because this would be transformed into
```
for (let _foo of [{foo: 1}]) {
  ({foo} = _foo)
  let foo;
}
```
and Babel would transform that into
```
for (let _foo of [{foo: 1}]) {
  ({ foo } = _foo);
  exports.foo = foo;
  let foo;
}
```
which results in an error.

nicolo-ribaudo:
I was proposing to do something like this:

1. Let `bindings` be `t.getOuterBindingIdentifiers(left)`
2. For each `binding` in `bindings`,
  a. If the inner scope declares `binding`, call `rename(binding)` on the inner scope
3. Move `PATTERN` to `PATTERN = _foo` inside the loop body.

vedantroy:
@nicolo-ribaudo 

I changed the code to use your method.

packages/babel-helper-module-transforms/src/rewrite-live-references.js
```diff
@@ -306,4 +306,41 @@ const rewriteReferencesVisitor = {
...
+          didTransform = true;
+          if (loopBodyScope.hasOwnBinding(name)) {
+            loopBodyScope.rename(name);
+          }
+        }
+      }
+      if (!didTransform) {
+        return;
+      }
+      const newLoopId = scope.generateUidIdentifier();
```
nicolo-ribaudo: Nit: we can pass an "hint" to generate a better name

```suggestion
      const newLoopId = scope.generateUidBasedOnNode(left);
```

vedantroy: Done!

packages/babel-helper-module-transforms/src/rewrite-live-references.js
```diff
@@ -306,4 +306,41 @@ const rewriteReferencesVisitor = {
...
+          didTransform = true;
+          if (loopBodyScope.hasOwnBinding(name)) {
+            loopBodyScope.rename(name);
+          }
+        }
+      }
+      if (!didTransform) {
+        return;
+      }
+      const newLoopId = t.identifier(scope.generateUidBasedOnNode(left));
```
JLHwung:
```suggestion
      const newLoopId = scope.generateUidIdentifierBasedOnNode(left);
      scope.push({ id: newLoopId });
```

The first change is for shorter code size. The second change is to make sure the scope is synced with AST changes.

vedantroy: @JLHwung The second change seems to lead to broken output? For example:
```
for (foo of []) {}
```
is transformed into
```
for (let _foo of []) {
  exports.bar = exports.foo = foo = _foo;

  var _foo;
}
```
after the second change. I don't think the `var _foo;` should be there?

nicolo-ribaudo: Yeah, `push()` declares a new variable. We need `registerBinding()` or `registerDeclaration()`.

PR #11078, opened 2020-01-31
Do not publish build directory of compat-data (#11078) [skip ci]

danez:
| Fixed Issues?            | 
| Documentation PR Link    | 

The build directory should not be published.

Also we do not have a src or test directory in this package.

### DIFF
--- a/packages/babel-compat-data/.npmignore
+++ b/packages/babel-compat-data/.npmignore
@@ -1,3 +1,4 @@
+build
 scripts
 src
 test


packages/babel-compat-data/.npmignore
```diff
@@ -1,4 +1,3 @@
+build
 scripts
-src
-test
```
nicolo-ribaudo: Why remove src and test?

danez: There is no src or test directory in this module. But I can also leave it in. Should I?

nicolo-ribaudo: Oh I edited my comment exactly at the same time as your reply :sweat_smile:

PR #11081, opened 2020-01-31
Only download compat-table if it's not already downloaded

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  |

I'm currently working on a PR where I need to run `make build-compat-data` many times, and it's a bit annoying that it re-clones the repo every time :stuck_out_tongue:

### DIFF
--- a/packages/babel-compat-data/scripts/download-compat-table.sh
+++ b/packages/babel-compat-data/scripts/download-compat-table.sh
@@ -2,6 +2,12 @@
 set -e
 
 COMPAT_TABLE_COMMIT=4e9369a699b0e15ba5c21586ce3bdd34299db9c1
+GIT_HEAD=build/compat-table/.git/HEAD
+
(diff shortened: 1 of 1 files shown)


packages/babel-compat-data/scripts/download-compat-table.sh
```diff
@@ -2,6 +2,12 @@
 set -e
 
 COMPAT_TABLE_COMMIT=4e9369a699b0e15ba5c21586ce3bdd34299db9c1
+GIT_HEAD=build/compat-table/.git/HEAD
+
+if [ -f "$GIT_HEAD" -a "$(cat $GIT_HEAD)" == $COMPAT_TABLE_COMMIT ]; then
```
JLHwung: It feels ad-hoc to rely on git internals. How about
```suggestion
if [ -f "build/compat-table" -a "$(git rev-parse HEAD)" == $COMPAT_TABLE_COMMIT ]; then
```

PR #11083, opened 2020-01-31
Include preset modules

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | https://github.com/babel/website/pull/2193
| Any Dependency Changes?  |

This PR introduces a new option for `@babel/preset-env`: `bugfixes`. When it is enabled,
Babel uses the plugins from `@babel/preset-modules` to replace "normal" plugins when possible.


nicolo-ribaudo:
Ready for review!

hzoo:
This is something we would want to at least move to defaulting to true in Babel 8, correct? And we are just waiting since we don't want to break anyone in case (minor version)

packages/babel-compat-data/data/plugin-bugfixes.json
```diff
@@ -0,0 +1,125 @@
...
+    "ios": "10.3",
+    "samsung": "6",
+    "opera": "42",
+    "electron": "1.6"
+  },
+  "bugfix/transform-async-arrows-in-class": {
+    "chrome": "55",
+    "edge": "15",
+    "firefox": "52",
+    "safari": "11",
```
developit: ~Safari should be 10.1 / 10.3 here I think? Same as async-to-generator. I may have had references to Safari 11 in the plugin since that's what I had chosen to target, but the fix should work in all ES-Modules-supporting versions of Safari (10.1+ desktop / 10.3+ ios)~

Edit: I keep thinking this list is "versions that are fixed by X", but it's actually "oldest version where this plugin is not needed", right?

packages/babel-compat-data/data/plugin-bugfixes.json
```diff
@@ -0,0 +1,125 @@
+{
```
developit: @nicolo-ribaudo - meta-comment, more on the data format itself: do you think it might be easier to maintain this file if the data only specified which versions were actually affected by the bugfix plugins?  For example, the edge function name plugin could just be listed as `{ "edge": "79" }` rather than including the other browser versions which it doesn't have an effect on. A missing browser version could just be taken as "no change".

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -275,6 +275,11 @@ export default function normalizeOptions(opts: Options) {
     ) || corejs.proposals;
 
   return {
+    bugfixes: validateBoolOption(
+      TopLevelOptions.bugfixes,
+      opts.bugfixes,
+      false,
```
developit: I recall we'd said this would be enabled by default - I'm guessing this is `false` until Babel 8?

packages/babel-compat-data/data/plugin-bugfixes.json
existentialism: > oldest version where this plugin is not needed

:+1:

packages/babel-preset-env/src/normalize-options.js
nicolo-ribaudo: Yes :+1:
This shouldn't technically be a breaking change, but it's safer to wait (Babel 8 will be released soonish anyway).

packages/babel-compat-data/data/plugin-bugfixes.json
nicolo-ribaudo: This data is generated by the same script which generated the old data, so it's easier to generate all the targets :sweat_smile:

packages/babel-compat-data/data/overlapping-plugins.json
```diff
@@ -1,3 +1,18 @@
 {
-  "transform-regenerator": []
-}
\ No newline at end of file
+  "transform-async-to-generator": [
```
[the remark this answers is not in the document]
nicolo-ribaudo: Ideally yes. We did it in the past,but had to revert it because regenerator doesn't support async functions with polyfilled promises.

I'd like to revisit it in Babel 8 somehow.

packages/babel-compat-data/data/plugin-bugfixes.json
developit: Ah! gotcha.

packages/babel-compat-data/data/overlapping-plugins.json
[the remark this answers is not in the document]
developit: Okay. At least when bugfixes is enabled though, regenerator will never be included, right?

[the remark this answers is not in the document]
nicolo-ribaudo: It's only included when targeting browsers not supporting generators (we don't use it for async functions).

It's unrelated to `bugfixes`: it isn't included because of the `esmodules` option.

packages/babel-compat-data/data/plugin-bugfixes.json
hzoo: it could be interesting to create a rendered version based on the data on our website or something? Or just showcasing this in general in a more readable manner?

nicolo-ribaudo: Something like compat-table but with plugin names rather than ECMAScript features?

hzoo: Yeah, then you could generate what @developit was saying allow about seeing the diff's or what was actually affected? Maybe we should add a simple readme to explain each json file

packages/babel-preset-env/src/available-plugins.js
```diff
@@ -49,7 +49,20 @@ import transformTemplateLiterals from "@babel/plugin-transform-template-literals
 import transformTypeofSymbol from "@babel/plugin-transform-typeof-symbol";
 import transformUnicodeRegex from "@babel/plugin-transform-unicode-regex";
 
+import bugfixAsyncArrowsInClass from "@babel/preset-modules/lib/plugins/transform-async-arrows-in-class";
```
JLHwung: nit: We should export theses plugins from `@babel/preset-modules` instead of relying on internal file structures.

nicolo-ribaudo: I would like to just move the plugin files inside `@babel/preset-env` in the future, if @developit is ok with that.

PR #11088, opened 2020-02-03
Allow redeclaration of loop variable in body

### LINKED ISSUE #8913: for...of array optimization can introduce conflicting variables into the scope

openorclose:
| Fixed Issues?            | Fixes #8913 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The following code
```js
for (let i of 'foo') {
  let i = 'a';
  console.log(i);
}
```
now gives the error message 
```
/repl.js: Duplicate declaration "i"
  1 | for (let i of 'foo') {
> 2 |   let i = 'a';
    |       ^
  3 |   console.log(i);
  4 | }
```
on 7.8.4 [repl link](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYewTgBAFANgpgFwgSwiYEDkoSYJQQDeAUBBPEqgLxYCGmA3KRAMYgB2AziPAHQwgA5lGR4mAXyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Cstage-3&prettier=false&targets=&version=7.8.4&externalPlugins=)

when this is valid js code.

### Fix

We check to see if the immediate body of the for-of loop contains any bindings to variables declared in the `left` side of the for-of loop.

If so, we wrap the original for-of body in a block, and then prepend the necessary declarations/assignments during transformation:

So the above code would transform into:
```js
for (let _i = 0, _foo = 'foo'; _i < _foo.length; _i++) {
  let i = _foo[_i]; /// Line A
  {
    let i = 'a';
    console.log(i);
  }
}
```

Of note is that this extra block is necessary, and we cannot simply omit Line A in the above code to avoid creating the extra block, because we need to account for cases like:

```js
for ( const [first, ...rest] of array) {
  const first = 1;
  // no redeclaration of rest
}
```


packages/babel-plugin-transform-for-of/src/index.js
```diff
@@ -48,8 +48,19 @@ export default declare((api, options) => {
             );
           }
 
-          const block = t.toBlock(body);
-          block.body.unshift(assignment);
+          let blockBody;
+          const body = path.get("body");
+          if (
+            body.isBlockStatement() &&
+            Object.keys(path.getBindingIdentifiers()).some(id =>
```
JLHwung: Can you filter the binding identifiers `id` by
```js
id.type !== "var"
```
So we don't add extra block statement when the for binding is of `var` kind.
```js
for (var [head, ...tail] of array) {
  const head = 1;
  console.log(tail);
  console.log(head);
}
```

nicolo-ribaudo: Don't we still need the block? This code would be invalid:

```js
for (let _i = 0, _array = array; _i < _array.length; _i++) {
  var [head, ...tail] = _array[_i];
  const head = 1;
  console.log(tail);
  console.log(head);
}
```

openorclose:
> Thanks!

Thanks for the review!

PR #11089, opened 2020-02-03
Make `isReferenced` return false for method parameter name

### LINKED ISSUE #11087: types.isReferenced incorrectly returns true for object/class method param identifiers
## Bug Report

**Current Behavior**

For code like this:

```js
class Foo {
  bar(baz) {}
}
```

calling `isReferenced` for the `baz` node returns `true`.

**Input Code**

Here's a (failing) test case to demonstrate:

```js
test('demonstrate isReferenced issue', () => {
  const results = {};
  function plugin() {
    return {
      visitor: {
        Identifier(path) {
          results[path.node.name] = path.isReferenced();
        },
      },
    };
  }

  const code = 'class Foo { bar(baz) {} }';
  babel.transformSync(code, { plugins: [plugin], configFile: false });
  expect(results.baz).toBe(false);
});
```

**Expected behavior/code**

`isReferenced` for `baz` should be `false` because it is simply declaring a param for the method.

The test case above uses `configFile: false` when calling into babel; I'm running the test case itself with this config:

- Filename: `babel.config.js`

```js
module.exports = {
  presets: ['@babel/preset-env'],
};
```

**Environment**

- Babel version(s): 7.8.3
- Node/npm version: node 12.13.1, npm 6.13.6
- OS: OSX 10.14.6
- Monorepo: Lerna
- How you are using Babel: writing a babel plugin

**Possible Solution**

See this commit, let me know if I should submit a PR: https://github.com/brokensandals/babel/commit/63767c5841ff48c1a13f9af43c4bcf4a61ccf5b3

I'm writing a plugin that tries to transform all expressions in the source code. It needs to distinguish between identifiers that will be evaluated as expressions, and those which simply name functions/params/etc. My understanding is that `path.isReferenced` returning `true` should indicate that the Identifier node represents retrieval of a value by name, so I've been using that. It seems to work right in most cases, but not in the case described here.

brokensandals:
| Fixed Issues?            | Fixes #11087 

The `isReferenced` method (if I understand correctly) tries to determine if a given node is a reference to the value stored in some variable, as opposed to being e.g. a declaration of a variable or the left-hand-side of an assignment. As described in issue #11087, it currently misidentifies method parameter names as references; this fixes that.

The logic in `isReferenced` is based on the type of the parent node, and currently Object/ClassProperty and Object/Class(Private)Method parent nodes are handled by the same case block. That block checks for when the child node is the `key` of the property/method node, and for when it's part of an ObjectPattern, but otherwise returns true. This change adds a check for when the child node is part of the `params` list of the parent node.

### DIFF
--- a/packages/babel-types/src/validators/isReferenced.js
+++ b/packages/babel-types/src/validators/isReferenced.js
@@ -67,7 +67,7 @@ export default function isReferenced(
       if (parent.value === node) {
         return !grandparent || grandparent.type !== "ObjectPattern";
       }
-      if (parent.params && parent.params.indexOf(node) !== -1) {
+      if (parent.params.includes(node)) {
         return false;
       }
       return true;
--- a/packages/babel-types/src/validators/isReferenced.js
+++ b/packages/babel-types/src/validators/isReferenced.js
@@ -45,6 +45,15 @@ export default function isReferenced(
     case "PrivateName":
       return false;
 
+    // no: class { NODE() {} }
+    // yes: class { [NODE]() {} }
+    // no: class { foo(NODE) {} }
+    case "ClassMethod":
+    case "ClassPrivateMethod":
+    case "ObjectMethod":
+      if (parent.params.includes(node)) {
+        return false;
+      }
     // yes: { [NODE]: "" }
     // no: { NODE: "" }
     // depends: { NODE }
@@ -55,20 +64,11 @@ export default function isReferenced(
     // yes: class { key = NODE; }
     case "ClassProperty":
     case "ClassPrivateProperty":
-    // no: class { NODE() {} }
-    // yes: class { [NODE]() {} }
-    // no: class { foo(NODE) {} }
-    case "ClassMethod":
-    case "ClassPrivateMethod":
-    case "ObjectMethod":
       if (parent.key === node) {
         return !!parent.computed;
       }
       if (parent.value === node) {
         return !grandparent || grandparent.type !== "ObjectPattern";
-      }
-      if (parent.params.includes(node)) {
-        return false;
       }
       return true;
 
--- a/packages/babel-types/src/validators/isReferenced.js
+++ b/packages/babel-types/src/validators/isReferenced.js
@@ -67,12 +67,8 @@ export default function isReferenced(
       if (parent.value === node) {
         return !grandparent || grandparent.type !== "ObjectPattern";
       }
-      if (parent.params) {
-        for (const param of parent.params) {
-          if (param === node) {
-            return false;
-          }
-        }
+      if (parent.params && parent.params.indexOf(node) !== -1) {
+        return false;
       }
       return true;
 
--- a/packages/babel-types/src/validators/isReferenced.js
+++ b/packages/babel-types/src/validators/isReferenced.js
@@ -54,6 +54,8 @@ export default function isReferenced(
       if (parent.params.includes(node)) {
         return false;
       }
+    // Fall-through to next case clause to check whether the node is the method's name.
+
     // yes: { [NODE]: "" }
     // no: { NODE: "" }
     // depends: { NODE }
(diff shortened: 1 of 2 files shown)


packages/babel-types/src/validators/isReferenced.js
```diff
@@ -66,6 +67,13 @@ export default function isReferenced(
       if (parent.value === node) {
         return !grandparent || grandparent.type !== "ObjectPattern";
       }
+      if (parent.params) {
+        for (const param of parent.params) {
+          if (param === node) {
+            return false;
+          }
+        }
```
nicolo-ribaudo: What about using `parent.params.indexOf(node) !== -1`?

brokensandals: Ah, yeah, that's much better. Updated.

packages/babel-types/src/validators/isReferenced.js
```diff
@@ -66,6 +67,9 @@ export default function isReferenced(
       if (parent.value === node) {
         return !grandparent || grandparent.type !== "ObjectPattern";
       }
+      if (parent.params && parent.params.indexOf(node) !== -1) {
```
jridgewell:
```suggestion
      if (parent.params.includes(node)) {
```

brokensandals: @nicolo-ribaudo you asked me to remove the use of `.includes` for the sake of Node 6 support, but now that I check, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes indicates it's supported there. Is using `.includes` ok?

nicolo-ribaudo: Oh I thought that it wasn't 😅

If it works, it's ok to use it.

brokensandals: Cool, I checked a node 6 repl and `.includes` seems to work.

@jridgewell removing the check for whether `parent.params` is present before calling `.includes`, does mean an error would be thrown if we reach this line when parent is an `ObjectProperty` or `ClassProperty`. The only way that could happen is if `node` is neither the key nor the value of the parent Object/ClassProperty, which I think currently could only happen if it's a decorator or type annotation. I don't know whether it's important for `isReferenced` to be able to handle being called with non-identifier-like nodes such as Decorator.

nicolo-ribaudo: What about refactoring the switch like this then?

```js
case "ClassMethod":
case "ClassPrivateMethod":
case "ObjectMethod":
  check parent.params.includes(node);
  /* falls through */
case "ObjectProperty":
case "ClassProperty":
case "ClassPrivateProperty":
  // all the existing checks
```

brokensandals: That works, updated.

I'm now getting some failing tests in babel-preset-env related to "edge" being 18 instead of the fixture's expectation of 17, but I get them on master too - I'm guessing it was caused by a new version of caniuse-lite being released today.

nicolo-ribaudo: Yeah ignore those tests

packages/babel-types/src/validators/isReferenced.js
```diff
@@ -45,6 +45,15 @@ export default function isReferenced(
     case "PrivateName":
       return false;
 
+    // no: class { NODE() {} }
+    // yes: class { [NODE]() {} }
+    // no: class { foo(NODE) {} }
+    case "ClassMethod":
+    case "ClassPrivateMethod":
+    case "ObjectMethod":
+      if (parent.params.includes(node)) {
```
existentialism: May want to add that a comment that fall-through is intentional

brokensandals: Done.

PR #11090, opened 2020-02-03
[parser] Recover after using await expressions in sync contexts

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | Yes, but I don't think that it needs to wait for a minor version.
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -463,8 +463,13 @@ export default class ExpressionParser extends LValParser {
   // Parse unary operators, both prefix and postfix.
 
   parseMaybeUnary(refExpressionErrors: ?ExpressionErrors): N.Expression {
-    if (this.isContextual("await") && this.isAwaitAllowed()) {
-      return this.parseAwait();
+    const startPos = this.state.start;
+    const startLoc = this.state.startLoc;
+    const isAwait = this.isContextual("await");
+
+    if (isAwait && this.isAwaitAllowed()) {
+      this.next();
+      return this.parseAwait(startPos, startLoc);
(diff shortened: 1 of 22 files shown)


packages/babel-parser/test/fixtures/es2017/async-functions/await-inside-parameters-of-async-arrow-function/output.json
```diff
@@ -0,0 +1,158 @@
...
+      "column": 0
+    },
+    "end": {
+      "line": 1,
+      "column": 26
+    }
+  },
+  "errors": [
+    "SyntaxError: Await expressions are only allowed inside async functions. (1:11)",
+    "SyntaxError: Await cannot be used as name inside an async function (1:11)"
```
JLHwung: In this case there are two contradicting errors in the same position, IMO it is confusing and we should fix it.

packages/babel-parser/src/parser/expression.js
```diff
@@ -514,6 +517,29 @@ export default class ExpressionParser extends LValParser {
...
+      !this.hasPrecedingLineBreak()
+    ) {
+      this.raise(
+        startPos,
+        this.hasPlugin("topLevelAwait")
+          ? "Await expressions are only allowed inside async functions or modules."
+          : "Await expressions are only allowed inside async functions.",
+      );
+
+      return this.parseAwait(startPos, startLoc);
```
JLHwung: We can reset the `awaitPos` after await expression is parsed since we have thrown an recoverable error before.

PR #11092, opened 2020-02-04
Fix Async Generic After Await Parsing Error

### LINKED ISSUE #11007: Error when Mixing Nested Async Functions and TypeScript Generics
## Bug Report

I'm interested in helping, but I don't have much context! If someone else would like to take point on a fix, that is also fine with me.

**Current Behavior**
When mixing TypeScript generics and async/await in a certain way, I'm seeing an `Unexpected Token, expected ";"` error.

**Input Code**
- [REPL](https://babeljs.io/en/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=KYDwDg9gTgLgBAYwgOwM7wGYQnAvHAQ1QE9kE4AKASjwD44BvAKDkIHcCBLeABSggC2nVMAB0UYKggAbAG7AKDAL5UA3ExaIU6OACMCUPIRJk4AHgAqtanTgByLBDvqlqoA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=typescript&prettier=false&targets=&version=7.8.3&externalPlugins=)

This is the smallest reproducible example I could come up with:

```ts
export const foo = async () => {
  await Promise.resolve({});

  const bar = async <T>() => 'foo';
};
```

Looks like it fails if you have an `await` before an `async` function with a generic defined.

**Expected behavior/code**
As far as I can tell, this should transpile. Its individual components do.

(Note that I was able to repro this on the REPL)

- Filename: `.babelrc`

```json
{
  "presets": [
    [
      "@babel/preset-env",
      {
        "targets": {
          "node": "8.10"
        },
        "useBuiltIns": "entry",
      }
    ],
    "@babel/preset-typescript"
  ],
  "plugins": [
    "lodash",
    [
      "babel-plugin-transform-builtin-extend",
      {
        "globals": ["Error"]
      }
    ]
  ]
}
```

**Environment**

```
  System:
    OS: macOS 10.15.1
  Binaries:
    Node: 8.10.0 - ~/.nvm/versions/node/v8.10.0/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 5.6.0 - ~/.nvm/versions/node/v8.10.0/bin/npm
  npmPackages:
    @babel/cli: 7.8.3 => 7.8.3 
    @babel/core: 7.8.3 => 7.8.3 
    @babel/node: 7.8.3 => 7.8.3 
    @babel/polyfill: 7.8.3 => 7.8.3 
    @babel/preset-env: 7.8.3 => 7.8.3 
    @babel/preset-typescript: 7.8.3 => 7.8.3 
    @babel/register: 7.8.3 => 7.8.3 
    babel-jest: 24.9.0 => 24.9.0 
    babel-loader: 8.0.5 => 8.0.5 
    babel-plugin-lodash: 3.3.4 => 3.3.4 
    babel-plugin-transform-builtin-extend: 1.1.2 => 1.1.2 
    eslint: 5.13.0 => 5.13.0 
    jest: 24.9.0 => 24.9.0 
    webpack: 4.29.3 => 4.29.3 
```

**Possible Solution**

Raw output (from a `yarn babel-node` invocation):

```
/Users/liam/fabric/linen/node_modules/@babel/core/lib/parser/index.js:95
    throw err;
    ^

SyntaxError: /path/to/file.ts: Unexpected token, expected ";" (31:26)

  29 |   await Promise.resolve({});
  30 | 
> 31 |   const bar = async <T>() => 'foo';
     |                           ^
  32 | };
  33 | 
    at Object.raise (/Users/liam/fabric/linen/node_modules/@babel/core/node_modules/@babel/parser/lib/index.js:7013:17)
    at Object.unexpected (/Users/liam/fabric/linen/node_modules/@babel/core/node_modules/@babel/parser/lib/index.js:8384:16)
    at Object.semicolon (/Users/liam/fabric/linen/node_modules/@babel/core/node_modules/@babel/parser/lib/index.js:8366:40)
    at Object.parseVarStatement (/Users/liam/fabric/linen/node_modules/@babel/core/node_modules/@babel/parser/lib/index.js:11129:10)
    at Object.parseStatementContent (/Users/liam/fabric/linen/node_modules/@babel/core/node_modules/@babel/parser/lib/index.js:10725:21)
    at Object.parseStatementContent (/Users/liam/fabric/linen/node_modules/@babel/core/node_modules/@babel/parser/lib/index.js:5972:18)
    at Object.parseStatement (/Users/liam/fabric/linen/node_modules/@babel/core/node_modules/@babel/parser/lib/index.js:10658:17)
    at Object.parseBlockOrModuleBlockBody (/Users/liam/fabric/linen/node_modules/@babel/core/node_modules/@babel/parser/lib/index.js:11234:25)
    at Object.parseBlockBody (/Users/liam/fabric/linen/node_modules/@babel/core/node_modules/@babel/parser/lib/index.js:11221:10)
    at Object.parseBlock (/Users/liam/fabric/linen/node_modules/@babel/core/node_modules/@babel/parser/lib/index.js:11205:10)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I believe this is a regression - I'm in the middle of upgrading from `7.2.x` to `7.8.x`, so I haven't found the exact version that introduced it yet, but the code in question worked in `7.2.x`.

liamfd:
| Fixed Issues?            | Fixes #11007
| Tests Added + Pass?      | yes

As outlined in #11007, code like the following would fail to compile:

```ts
export const foo = async () => {
  await Promise.resolve({});

  const bar = async <T>() => 'foo';
};
```

This would only happen with an `await` followed by an async arrow function with a TypeScript generic defined. @JLHwung helped me understand the issue, and [suggested the fix](https://github.com/babel/babel/issues/11007#issuecomment-574232329) I've implemented in this PR. 

In short, when parsing the snippet above, Babel would throw an `Unexpected token, expected ";"` error, which was caused by an `Await cannot be used as name inside an async function (2:2)` error. I've attempted to address this by resetting `awaitPos` in `tsTryParseGenericAsyncArrowFunction`.

Thanks in advance for any feedback, and thank you @JLHwung for your guidance on this issue!

### DIFF
--- a/packages/babel-parser/src/plugins/typescript/index.js
+++ b/packages/babel-parser/src/plugins/typescript/index.js
@@ -1532,7 +1532,7 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       const oldMaybeInArrowParameters = this.state.maybeInArrowParameters;
       const oldYieldPos = this.state.yieldPos;
       const oldAwaitPos = this.state.awaitPos;
-      this.state.maybeInArrowParameters = false;
+      this.state.maybeInArrowParameters = true;
       this.state.yieldPos = -1;
       this.state.awaitPos = -1;
 
--- a/packages/babel-parser/src/plugins/typescript/index.js
+++ b/packages/babel-parser/src/plugins/typescript/index.js
@@ -1528,6 +1528,14 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       if (!this.isRelational("<")) {
         return undefined;
       }
+
+      const oldMaybeInArrowParameters = this.state.maybeInArrowParameters;
+      const oldYieldPos = this.state.yieldPos;
+      const oldAwaitPos = this.state.awaitPos;
+      this.state.maybeInArrowParameters = false;
+      this.state.yieldPos = -1;
+      this.state.awaitPos = -1;
+
       const res: ?N.ArrowFunctionExpression = this.tsTryParseAndCatch(() => {
         const node: N.ArrowFunctionExpression = this.startNodeAt(
           startPos,
@@ -1540,6 +1548,10 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         this.expect(tt.arrow);
         return node;
       });
+
+      this.state.maybeInArrowParameters = oldMaybeInArrowParameters;
+      this.state.yieldPos = oldYieldPos;
+      this.state.awaitPos = oldAwaitPos;
 
       if (!res) {
         return undefined;
--- a/packages/babel-parser/test/fixtures/typescript/arrow-function/async-generic-after-await/input.ts
+++ b/packages/babel-parser/test/fixtures/typescript/arrow-function/async-generic-after-await/input.ts
@@ -1,0 +1,4 @@
+async () => {
+  await null;
+  async <T>() => null;
+};
--- a/packages/babel-parser/test/fixtures/typescript/arrow-function/async-generic-after-await/output.json
+++ b/packages/babel-parser/test/fixtures/typescript/arrow-function/async-generic-after-await/output.json
@@ -12,9 +12,6 @@
       "column": 2
     }
   },
-  "errors": [
-    "SyntaxError: Await cannot be used as name inside an async function (2:2)"
-  ],
   "program": {
     "type": "Program",
     "start": 0,
--- a/packages/babel-parser/test/fixtures/typescript/arrow-function/async-generic-after-await/output.json
+++ b/packages/babel-parser/test/fixtures/typescript/arrow-function/async-generic-after-await/output.json
@@ -1,0 +1,217 @@
+{
+  "type": "File",
+  "start": 0,
+  "end": 53,
+  "loc": {
+    "start": {
+      "line": 1,
+      "column": 0
+    },
+    "end": {
+      "line": 4,
+      "column": 2
+    }
+  },
+  "errors": [
+    "SyntaxError: Await cannot be used as name inside an async function (2:2)"
+  ],
+  "program": {
+    "type": "Program",
+    "start": 0,
+    "end": 53,
+    "loc": {
+      "start": {
+        "line": 1,
+        "column": 0
+      },
+      "end": {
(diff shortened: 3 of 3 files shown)


packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -1528,6 +1528,14 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       if (!this.isRelational("<")) {
         return undefined;
       }
+
+      const oldMaybeInArrowParameters = this.state.maybeInArrowParameters;
```
liamfd: I wasn't 100% sure where to put this change, and the one below to reset things. 

For resetting these values, it seemed best to do it before any attempt to parse is started, but strictly speaking it only seemed to need to happen before `parseFunctionParams`.

For setting state back to oldState below, it looks like typically this is done before `parseArrowExpression` is called (e.g. https://github.com/liamfd/babel/blob/master/packages/babel-parser/src/parser/expression.js#L972), so I tried to match that.

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -1528,6 +1528,14 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+      const oldYieldPos = this.state.yieldPos;
```
liamfd: I added `oldMaybeInArrowParameters` and `oldYieldPos`  at @JLHwung 's suggestion. From looking around, it does seem like these are generally reset together. 

I also wasn't entirely sure whether `oldMaybeInArrowParameters` should be set here, or whether it should be `true` or `false`. I opted to match https://github.com/liamfd/babel/blob/master/packages/babel-parser/src/parser/expression.js#L972, `true` seemed more appropriate.

liamfd:
Looks like I've got some failures, I'll try to address them tonight.

PR #11093, opened 2020-02-04
Turn on no-fallthrough rule

JLHwung:

Turn on the `no-fallthrough` ESLint rules, which is included in the `eslint:recommended` preset.

Related: https://github.com/babel/babel/pull/11089#discussion_r374728817

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -36,6 +36,7 @@ module.exports = function(api) {
     case "standalone":
       includeRegeneratorRuntime = true;
       unambiguousSources.push("packages/babel-runtime/regenerator");
+    // fall through
     case "rollup":
       convertESM = false;
       ignoreLib = false;
--- a/packages/babel-types/src/validators/isReferenced.js
+++ b/packages/babel-types/src/validators/isReferenced.js
@@ -54,16 +54,18 @@ export default function isReferenced(
       if (parent.params.includes(node)) {
         return false;
       }
-    // Fall-through to next case clause to check whether the node is the method's name.
+    // fall through
 
     // yes: { [NODE]: "" }
     // no: { NODE: "" }
     // depends: { NODE }
     // depends: { key: NODE }
+    // fall through
     case "ObjectProperty":
     // no: class { NODE = value; }
     // yes: class { [NODE] = value; }
     // yes: class { key = NODE; }
+    // fall through
     case "ClassProperty":
     case "ClassPrivateProperty":
       if (parent.key === node) {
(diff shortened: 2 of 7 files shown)


babel.config.js
```diff
@@ -36,6 +36,7 @@ module.exports = function(api) {
     case "standalone":
       includeRegeneratorRuntime = true;
       unambiguousSources.push("packages/babel-runtime/regenerator");
+    // fall through
```
nicolo-ribaudo: I hate that prettier doesn't include this comment, since conceptually it's a control flow statement similar to `break`/`continue`.

kaicataldo: Classic ambiguous comment indentation 😆

packages/babel-types/src/validators/isReferenced.js
```diff
@@ -54,16 +54,18 @@ export default function isReferenced(
...
         return false;
       }
-    // Fall-through to next case clause to check whether the node is the method's name.
+    // fall through
 
     // yes: { [NODE]: "" }
     // no: { NODE: "" }
     // depends: { NODE }
     // depends: { key: NODE }
+    // fall through
```
nicolo-ribaudo: @kaicataldo The comment on line 57 is not enough. Could https://github.com/eslint/eslint/issues/10608 be reopened? :pray: :sweat_smile:

kaicataldo: I wonder if we should allow it to be either the first or last, but not the middle? I know it's unlikely, but I think it would be good to avoid something like this triggering it:

```js
/* eslint no-fallthrough: ["error", { "commentPattern": "fall-through[\\s\\w]*not allowed" }] */

switch(foo) {
    case 1:
        doSomething();
        // We don't want to fall through here
        // because fall-through is not allowed,
        // oh look, a run on sentence.

    default:
        doSomething();
}

```

nicolo-ribaudo: Doesn't this code already have that problem?

```js
/* eslint no-fallthrough: ["error", { "commentPattern": "fall-through[\\s\\w]*not allowed" }] */

switch(foo) {
    case 1:
        doSomething();
        // We don't want to fall through here because fall-through is not allowed.

    default:
        doSomething();
}
```

kaicataldo: Yeah, it would. Might just be me, but the grouped line comments seem more error prone. It probably is the same thing in practice, though.

PR #11095, opened 2020-02-04
Remove functions present only for backwards-compatibility (#6993)

nicolo-ribaudo:
Closes #6993 (rebased)


nicolo-ribaudo:
(Also approved in the other PR)

PR #11096, opened 2020-02-05
Wrap type-only class fields in flow comments

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  |

Our current code will break when https://github.com/babel/babel/issues/10039 is fixed. This has been caught by https://github.com/babel/babel/pull/10120.

I will push this commit to https://github.com/babel/babel/pull/10120 to check if this is enough, please don't merge this PR until we are sure that e2e tests pass.

### DIFF
--- a/packages/babel-parser/src/parser/index.js
+++ b/packages/babel-parser/src/parser/index.js
@@ -1,7 +1,7 @@
 // @flow
 
 import type { Options } from "../options";
-import type { File, JSXOpeningElement } from "../types";
+import type { File /*::, JSXOpeningElement */ } from "../types";
 import type { PluginList } from "../plugin-utils";
 import { getOptions } from "../options";
 import StatementParser from "./statement";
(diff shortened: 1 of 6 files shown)


packages/babel-parser/src/parser/index.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 
 import type { Options } from "../options";
-import type { File, JSXOpeningElement } from "../types";
+import type { File /*::, JSXOpeningElement */ } from "../types";
```
nicolo-ribaudo: This is needed otherwise ESLint marks it as unused (because it's "used" inside a comment).

nicolo-ribaudo:
Ok, this seems to work: https://circleci.com/workflow-run/3cdcd424-4d94-4a43-9982-cccf7dad1d29?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

PR #11101, opened 2020-02-06
Remove stage presets from the monorepo

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  |

`babel/babel-archive` already includes the last working version of these presets (v7.0.0-beta.52), and they have been throwing since 7.0.0


nicolo-ribaudo:
Probably yes, but now they are hardcoded in `@babel/standalone` (sigh).
Maybe we could remove them from `@babel/standalone` in Babel 8 and hard-code it in `babel/website`.

hzoo:
Oh right! We'll we aren't expecting anyone to be using any stage presets anymore for a long time now so we could remove it there too. It's only for convenience in the repl and that we haven't made the UI change there?

nicolo-ribaudo:
> It's only for convenience in the repl and that we haven't made the UI change there?

Yeah, exactly.

nicolo-ribaudo:
(The build failure is unrelated)

PR #11102, opened 2020-02-06
fix(transform-react-constant-elements): wrap hoisted values to enable tree-shaking

### LINKED ISSUE #11037: @babel/plugin-transform-react-constant-elements generates non-tree-shakeable output
## Bug Report

Consider the following input:

```jsx
  return <svg><path /></svg>
}
```

```js
var _ref =
/*#__PURE__*/

  return _ref;
}
```

```js
var _ref =
/*#__PURE__*/ (() => 
```

- Babel version(s): 7.x

yannickcr:
| Fixed Issues?            | Fixes #11037 
| Major: Breaking Change?  | I don't think, plugin output is different but it should not be a breaking change
| Tests Added + Pass?      | Updated output snapshots

As described in https://github.com/babel/babel/issues/11037, the current code generated by @babel/plugin-transform-react-constant-elements is not tree-shakable since the `#__PURE__` annotation only works on function calls (https://github.com/terser/terser/issues/513).

The proposed solution (https://github.com/terser/terser/issues/513#issuecomment-554685420) is to wrap the hoisted values in an IIFE to make them tree-shakable. This is what this change intent to do.

No new tests were added but the existing snapshots were updated to match the new output.

### DIFF
--- a/packages/babel-plugin-transform-react-constant-elements/test/fixtures/constant-elements/append-to-end-when-declared-in-scope-2/output.mjs
+++ b/packages/babel-plugin-transform-react-constant-elements/test/fixtures/constant-elements/append-to-end-when-declared-in-scope-2/output.mjs
@@ -1,17 +1,25 @@
 var _ref =
 /*#__PURE__*/
-<div>child</div>;
+(() => <div>child</div>)();
 
 const AppItem = () => {
   return _ref;
 };
 
+var _ref3 =
+/*#__PURE__*/
+(() => <p>Parent</p>)();
+
+var _ref4 =
+/*#__PURE__*/
+(() => <AppItem />)();
+
 var _ref2 =
 /*#__PURE__*/
-<div>
-        <p>Parent</p>
-        <AppItem />
-      </div>;
+(() => <div>
+        {_ref3}
+        {_ref4}
+      </div>)();
 
 export default class App extends React.Component {
   render() {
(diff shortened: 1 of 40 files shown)


nicolo-ribaudo:
The problem that this PR solver is _not_ that `terser` is unable to remove compiled JSX elements, it's that we don't add the `__PURE__` annotation to nested elements.

Try pasting this code in Terser's [REPL](https://try.terser.org/):
```js
(() => {

let _ref =
/*#__PURE__*/
React.createElement("svg1", null, React.createElement("path1", null));

let _ref2 =
/*#__PURE__*/
(() => React.createElement("svg2", null, React.createElement("path2", null)))();

})();
```

You'll notice that both `svg1` and `svg2` are removed, but only `path2` and not `path1`.

packages/babel-plugin-transform-react-constant-elements/test/fixtures/constant-elements/append-to-end-when-declared-in-scope-2/output.mjs
```diff
@@ -1,17 +1,25 @@
...
 var _ref2 =
 /*#__PURE__*/
-<div>
-        <p>Parent</p>
-        <AppItem />
-      </div>;
+(() => <div>
+        {_ref3}
+        {_ref4}
+      </div>)();
```
nicolo-ribaudo: This should be kept as a single element, not as 3 separate elements.
The arrow function is not needed if the elements don't have impure children anyway.

JLHwung:
Oops, I am also working on this issue and #11105 is just submitted.

yannickcr:
@JLHwung Oh! Seems that unlike me you're managed to skip hoisting for children elements.

Let's continue on your PR then, I'll close this one 🙂

PR #11105, opened 2020-02-06
Wrap hoisted JSX elements into IIAFE so pure annotation can work

### LINKED ISSUE #11037: @babel/plugin-transform-react-constant-elements generates non-tree-shakeable output

JLHwung:
| Fixed Issues?            | Fixes #11037 
| Major: Breaking Change?  |
| Minor: New Feature?      |

`@babel/plugin-transform-react-constant-elements` adds pure annotation to the hoisted JSX elements. 
```jsx
var _ref2 =
/*#__PURE__*/
<div>child</div>;

const AppItem = () => {
  return _ref2;
}
```
However, in reality it never work as expected with bundlers/minifiers because `terser` only supports pure annotation on call expression (https://github.com/terser/terser/issues/513#issuecomment-554685420). This PR wraps the hoisted elements into an IIAFE so Rollup can remove unused constant elements.
```jsx
var _ref2 =
/*#__PURE__*/
(() => <div>child</div>)();
```

Note that this PR will cause unnecessary hoisting under certain scenarios, (see comments below). To mitigate this side effect we may have babel-traverse exposed the Path Hoister API, but it could not be done in v7 due to our semver requirement.

cc @babel/react

### DIFF
--- a/packages/babel-plugin-transform-react-constant-elements/test/fixtures/constant-elements/append-to-end-when-declared-in-scope-3/output.js
+++ b/packages/babel-plugin-transform-react-constant-elements/test/fixtures/constant-elements/append-to-end-when-declared-in-scope-3/output.js
@@ -1,10 +1,10 @@
 var _ref2 =
 /*#__PURE__*/
-<div>child</div>;
+(() => <div>child</div>)();
 
 var _ref3 =
 /*#__PURE__*/
-<p>Parent</p>;
+(() => <p>Parent</p>)();
 
 (function () {
   class App extends React.Component {
@@ -17,10 +17,13 @@ var _ref3 =
   const AppItem = () => {
     return _ref2;
   },
+        _ref4 =
+  /*#__PURE__*/
+  (() => <AppItem />)(),
         _ref =
   /*#__PURE__*/
-  <div>
+  (() => <div>
           {_ref3}
-          <AppItem />
-        </div>;
+          {_ref4}
+        </div>)();
 });
--- a/packages/babel-plugin-transform-async-to-generator/test/fixtures/regression/7178/output.js
+++ b/packages/babel-plugin-transform-async-to-generator/test/fixtures/regression/7178/output.js
@@ -6,9 +6,9 @@ function action() {
 
 var _ref =
 /*#__PURE__*/
-React.createElement(Contact, {
+(() => React.createElement(Contact, {
   title: title
-});
+}))();
 
 function _action() {
   _action = babelHelpers.asyncToGenerator(function* () {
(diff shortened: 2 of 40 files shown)


packages/babel-plugin-transform-react-constant-elements/test/fixtures/constant-elements/append-to-end-when-declared-in-scope-3/output.js
```diff
@@ -17,10 +17,13 @@ var _ref3 =
...
+  /*#__PURE__*/
+  (() => <AppItem />)(),
         _ref =
   /*#__PURE__*/
-  <div>
+  (() => <div>
           {_ref3}
-          <AppItem />
-        </div>;
+          {_ref4}
```
JLHwung: The `<AppItem />` here is unnecessarily hoisted to the same scope of `_ref` because the Path Hoister can not neglect the introduced IIAFE. Note that when the JSX element is _not_ hoisted to the top level, its children may has opportunity hoisted to top level. But we can't know which scope the element will be hoisted to until we have run `path.hoist()`, after that the AST is changed and it's too later to determine whether such hoisting is favourable.

nicolo-ribaudo: What if we added a new parameter to `path.hoist()`, to say "hoist this path but only if it can go at least up to this scope"?

JLHwung: Yeah, we can do it but due to the semver requirements we can not ship it in patch/minor releases unless we are doing dirty version checks.

Do we have plans to change semver to minor-version compatibilities in Babel 8?

nicolo-ribaudo: If someone uses an old `@babel/traverse` version, their JSX elements will be hoisted singularly because the "minimum scope" parameter is ignored. It's not optimal, but it will still perfectly work.

gaearon:
This seems like it adds overhead to the startup path, which is bad. However, the current transform is already non-ideal in that sense (it adds startup overhead with no lazyness). We couldn't get it working at FB so I don't think we care strongly about how it works.

nicolo-ribaudo:
Actually, adding a `lazy` option (similar to the commonjs plugin) seems a good idea to me.

JLHwung:
Closing it in favor of #11126

PR #11106, opened 2020-02-06
Centralize @babel/eslint-* tests

kaicataldo:
| Fixed Issues?            | 
| Patch: Bug Fix?          | 
| Patch: Chore?          | :+1:
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

I've been trying to figure out how to structure tests (and which packages should be responsible for which tests), and have decided to advocate for the following:

* `@babel/eslint-parser`: Only tests the shape of the AST (and asserts that it matches Espree's)
* `@babel/eslint-plugin`: Only test core rules that we have to create an `@babel/*` copy of
* `@babel/eslint-tests`: Integration/performance tests (typescript-eslint has a similar setup). This includes testing with ESLint itself, the ESLint core rules that we don't have to override, popular plugins, etc.

I've moved all of the integration-y (by the definition above) tests into `@babel/eslint-tests` to centralize them there. Just to be clear, `@babel/eslint-tests` should never be published. I made it a package because we need to install deps for testing.

Thoughts/concerns?

I'm not sure why we weren't linting any of the `@babel/eslint-*` test files, but I enabled that as well.


eslint/babel-eslint-parser/test/fixtures/rules/syntax-error.js
```diff
@@ -1,6 +0,0 @@
-class ClassName {
```
kaicataldo: I don't think this is actually being used anywhere?

eslint/babel-eslint-tests/test/helpers/verifyAndAssertMessages.js
```diff
@@ -0,0 +1,70 @@
...
+    );
+  }
+
+  messages.forEach((message, i) => {
+    const formattedMessage = `${message.line}:${message.column} ${
+      message.message
+    }${message.ruleId ? ` ${message.ruleId}` : ""}`;
+    const expectedMessage = expectedMessages[i];
+
+    if (expectedMessage instanceof RegExp) {
```
kaicataldo: Added the ability to pass this function a `RegExp` so we can enable the `xit`ed test below.

eslint/babel-eslint-tests/test/integration/eslint/eslint.js
```diff
@@ -90,9 +13,9 @@ describe("verify", () => {
     );
   });
 
-  xit("Readable error messages (issue #3)", () => {
+  it("Readable error messages (issue #3)", () => {
     verifyAndAssertMessages("{ , res }", {}, [
-      "1:3 Parsing error: Unexpected token",
+      /1:2 Parsing error:.*Unexpected token \(1:2\)/,
```
kaicataldo: This error reports an absolute file path, which is why I'm assuming it was disabled. With `RegExp` support we can now enable this!

kaicataldo:
Hm, looks like CI is failing in Windows when the machine is trying to install yarn.

eslint/babel-eslint-tests/test/helpers/verifyAndAssertMessages.js
```diff
@@ -0,0 +1,69 @@
...
+          Actual: ${formattedMessage}
+        `,
+        );
+      }
+    } else if (formattedMessage !== expectedMessage) {
+      throw new Error(
+        `
+          Message ${i} does not match:
+          Expected: ${expectedMessage}
+          Actual:   ${formattedMessage}
```
nicolo-ribaudo: Ideally, we should be able to re-use `jest.each` instead of writing our own logic.

eslint/babel-eslint-tests/test/helpers/verifyAndAssertMessages.js
nicolo-ribaudo: Can we use `jest`'s methods instead of writing our own errors checking logic?

kaicataldo: I think that would be great. Mind if we do that in a follow up PR?

nicolo-ribaudo: It's ok!

PR #11109, opened 2020-02-07
Fix rewriteThis in helper-module-transforms for computed class elements

### LINKED ISSUE #10784: rewriteThis in helper-module-transforms misses computed class elements
## Bug Report

**Current Behavior**
The transformed code does not throw an expected error.

**Input Code**
- REPL or Repo link if applicable: https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=KYDwDg9gTgLgBAYwDYEMDOa4GE4G84DaMAFgJZoB0AdigLbAC6AFAJR4C-c7QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=false&targets=&version=7.7.4&externalPlugins=%40babel%2Fplugin-transform-modules-commonjs%407.7.4

```js
export class C { [this.name]() {} }
```

**Expected behavior/code**
The transformed code should throw _can not read property `name` of undefined_. Because `this` is the top level context, which is `undefined` if the source type is `module`.

- Filename: `babel.config.js`

```js
{
  "plugins": ["@babel/plugin-transform-modules-commonjs"]
}
```

**Environment**

```

  System:
    OS: macOS 10.15.1
  Binaries:
    Node: 13.1.0 - /usr/local/bin/node
    Yarn: 1.19.2 - /usr/local/bin/yarn
    npm: 6.12.1 - /usr/local/bin/npm
  Monorepos:
    Lerna: 3.19.0
  npmPackages:
    @babel/cli: ^7.7.0 => 7.7.0 
    @babel/core: ^7.7.2 => 7.7.2 
    @babel/eslint-plugin-development: ^1.0.1 => 1.0.1 
    @babel/plugin-proposal-class-properties: ^7.7.0 => 7.7.0 
    @babel/plugin-proposal-export-namespace-from: ^7.5.2 => 7.5.2 
    @babel/plugin-proposal-nullish-coalescing-operator: ^7.4.4 => 7.4.4 
    @babel/plugin-proposal-numeric-separator: ^7.2.0 => 7.2.0 
    @babel/plugin-proposal-object-rest-spread: ^7.7.4 => 7.7.4 
    @babel/plugin-proposal-optional-chaining: ^7.6.0 => 7.6.0 
    @babel/plugin-transform-flow-strip-types: ^7.7.4 => 7.7.4 
    @babel/plugin-transform-for-of: ^7.7.4 => 7.7.4 
    @babel/plugin-transform-modules-commonjs: ^7.7.0 => 7.7.0 
    @babel/plugin-transform-runtime: ^7.6.2 => 7.6.2 
    @babel/preset-env: ^7.7.1 => 7.7.1 
    @babel/preset-flow: ^7.0.0 => 7.0.0 
    @babel/register: ^7.7.0 => 7.7.0 
    @babel/runtime: ^7.7.2 => 7.7.2 
    babel-eslint: ^11.0.0-beta.1 => 11.0.0-beta.1 
    babel-jest: ^24.9.0 => 24.9.0 
    babel-plugin-transform-charcodes: ^0.2.0 => 0.2.0 
    eslint: ^6.0.1 => 6.0.1 
    eslint-config-babel: ^9.0.0 => 9.0.0 
    gulp-babel: ^8.0.0 => 8.0.0 
    jest: ^24.9.0 => 24.9.0 
    lerna: ^3.19.0 => 3.19.0 
    rollup-plugin-babel: ^4.0.0 => 4.3.3 
```

**Possible Solution**
https://github.com/babel/babel/blob/a6e8b3843b53c6e106d9fb2e37fea60c306dd97c/packages/babel-helper-module-transforms/src/rewrite-this.js#L17

1) Should traverse computed `ClassMethod`
2) requeue the `key` child of `ClassProperty` if it is computed. This step is to fix the following case
```js
export class C { [this.name] = 42 }
```

sidntrivedi012:
| Fixed Issues?            | Fixes #10784  
| Tests Added + Pass?      | Not as of now
| Documentation PR Link    | 
| Any Dependency Changes?  |

@JLHwung please review if I correctly used `ClassMethod` to find if its computed or not ? Also, where do we have to requeue the `key` child of `ClassProperty` to ?


sidntrivedi012:
The transformed code now throws `can not read property name of undefined`.

Screenshot 2020-02-08 at 8 03 42 PM

packages/babel-helper-module-transforms/src/rewrite-this.js
```diff
@@ -17,6 +17,11 @@ const rewriteThisVisitor = {
   ClassProperty(path) {
     path.skip();
   },
+  ClassMethod(path) {
+    if (path.node.computed) {
+      //requeue the key child of classProperty here
+    }
```
nicolo-ribaudo: This is a no-op, I don't understand how it fixes the bug :thinking:

sidntrivedi012: @nicolo-ribaudo I searched Babel handbook for ClassMethod and found a similar implementation here - https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#being-aware-of-nested-structures . Thus, tried to add like mentioned there.

nicolo-ribaudo: Yes, it vists `ClassMethod`, but doesn't do anything when visiting it.
It's similar to how this code visits every array element, but doesn't do anything:
```js
var a = [1, 2, 3];
a.forEach(num => {
  if (num % 2 == 0) {
    // It's even
  }
});
```

I don't think that this issue is fixed, and you can check it by using `@babel/plugin-syntax-class-properties` instead of `@babel/plugin-proposal-class-properties` in the test.

Currently, the test is correctly failing only because `@babel/plugin-proposal-class-properties` moves `this` outside of the computed key.

packages/babel-plugin-transform-modules-commonjs/test/fixtures/rewritethis-computed-class-elements/input.mjs
```diff
@@ -0,0 +1 @@
+export class C { [this.name] = 42 }
```
nicolo-ribaudo: This needs to be inside a nested folder; otherwise, the file is executed instead of being transformed and compared with `output.js`.

Can you move it inside `packages/babel-plugin-transform-modules-commonjs/test/fixtures/misc/undefined-this-computed-class-key/`?

packages/babel-helper-module-transforms/src/rewrite-this.js
```diff
@@ -17,6 +17,11 @@ const rewriteThisVisitor = {
   ClassProperty(path) {
     path.skip();
```
nicolo-ribaudo: We have a method to skip everything but the computed key somewhere (it's used by `environmentVisitor`). We can re-use it here.

sidntrivedi012: Okay, so we can use that method inside the ClassProperty to find if the `ClassMethod` is computed or not ?

nicolo-ribaudo: We should do it both for class methods and for class properties

nicolo-ribaudo: Also, note that `ClassMethods` are already visited by the `Function` visitor: we need to handle them there.

sidntrivedi012: Ok. Thanks for this info. Adding to this, do we have to requeue the key somewhere (it was mentioned in the issue text)?

nicolo-ribaudo: Not skipping it should be enough. Currently skip+requeue is buggy.

sidntrivedi012: Ok, will send a patch commit asap. Exams starting tomorrow 😅

nicolo-ribaudo: Good luck!

sidntrivedi012: Thanks and same to you too 👍

sidntrivedi012:
>We have a method to skip everything but the computed key somewhere (it's used by environmentVisitor). We can re-use it here.
@nicolo-ribaudo Have used `skipAllButComputedKey` function here and for using that, imported the `helper-replace-supers` package. But still, the tests don't show the intended error 🤔 . PTAL. Thanks.

/cc: @JLHwung

packages/babel-plugin-transform-modules-commonjs/test/fixtures/misc/undefined-this-computed-class-key/input.mjs
```diff
@@ -0,0 +1 @@
+export class C { [this.name] = 42 }
```
JLHwung: Can you add a test case for computed class method?
```
export class C { [this.name]() {} }
```

sidntrivedi012: @JLHwung Changed the test to the one for computed class method. PTAL.

packages/babel-plugin-transform-modules-commonjs/test/fixtures/misc/undefined-this-computed-class-key/input.mjs
```diff
@@ -0,0 +1 @@
+export class C { [this.name]() {} }
```
JLHwung: Oh the class property test is good to have. You can write multiple tests:
```
undefined-this-computed-class-method-name
undefined-this-computed-class-property-name
```

packages/babel-plugin-transform-modules-commonjs/test/fixtures/misc/undefined-this-computed-class-key/output.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+Object.defineProperty(exports, "__esModule", {
+  value: true
+});
+exports.C = void 0;
+
+class C {
+  [this.name]() {}
```
JLHwung: It should be overwritten by `rewriteThis`. You can check if `skipAllButComputedKey` is called in the `Function` visitor, which includes `ClassMethod`.

sidntrivedi012: @JLHwung Oh! I did not see the `!` before `isArrowFunctionExpression` and thus, didn't change `path.skip()`. Have changed it now to use the `skipAllButComputedKey` function. Also, have re-added the tests for `class-properties`. PTAL.

packages/babel-helper-module-transforms/src/rewrite-this.js
```diff
@@ -12,10 +13,10 @@ const rewriteThisVisitor = {
     path.replaceWith(path.scope.buildUndefinedNode());
   },
   Function(path) {
-    if (!path.isArrowFunctionExpression()) path.skip();
+    if (!path.isArrowFunctionExpression()) skipAllButComputedKey(path);
```
nicolo-ribaudo: We should call `skipAllButComputedKey` only if `path.isMethod()` (otherwise keep `path.skip()`), because the `skipAllButComputedKey` function expects a class element as parameter.

packages/babel-helper-module-transforms/src/rewrite-this.js
```diff
@@ -13,9 +14,10 @@ const rewriteThisVisitor = {
   },
   Function(path) {
     if (!path.isArrowFunctionExpression()) path.skip();
+    if (path.isMethod()) skipAllButComputedKey(path);
```
nicolo-ribaudo: A method is never an arrow function, so it will always be skipped by the line before.
We should do
1. If it is a Method,
  i. skipAllButComputedKey
2. Else, if it is not an arrow,
  i. skip

packages/babel-helper-module-transforms/src/rewrite-this.js
```diff
@@ -12,10 +13,11 @@ const rewriteThisVisitor = {
     path.replaceWith(path.scope.buildUndefinedNode());
   },
   Function(path) {
-    if (!path.isArrowFunctionExpression()) path.skip();
+    if (path.isMethod()) skipAllButComputedKey(path);
+    if (!path.isArrowFunctionExpression() && !path.isMethod()) path.skip();
```
nicolo-ribaudo: Nit: you can use `else if` and avoid checking `isMethod` again.

sidntrivedi012: @nicolo-ribaudo Made the changes. PTAL now.

packages/babel-helper-replace-supers/src/index.js
```diff
@@ -26,7 +26,7 @@ function getPrototypeOfExpression(objectRef, isStatic, file, isPrivateMethod) {
   return t.callExpression(file.addHelper("getPrototypeOf"), [targetRef]);
 }
 
-function skipAllButComputedKey(path: NodePath) {
+export function skipAllButComputedKey(path) {
```
nicolo-ribaudo:
```suggestion
export function skipAllButComputedKey(path: NodePath) {
```

sidntrivedi012:
> 👍
> 
> Sidenote: we may want to consider where `skipAllButComputedKey` lives (separate from this PR of course), as it seems weird to need to import replace supers for it?

Sure, should I open a discussion issue for it if needed ?

PR #11110, opened 2020-02-08
Disallow babel.transformFromAst synchronously

Pranav2612000:
| Fixed Issues?            | Part of Babel 8 Release plan 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR Link    |

Allows babel.transformFromAst to run only asynchronously by removing the part of the code required for running it synchronously i.e. with callback functions.

### DIFF
--- a/packages/babel-core/src/transform-ast.js
+++ b/packages/babel-core/src/transform-ast.js
@@ -19,10 +19,6 @@ type TransformFromAst = {
     opts: ?InputOptions,
     callback: FileResultCallback,
   ): void,
-
-  // Here for backward-compatibility. Ideally use ".transformSync" if you want
-  // a synchronous API.
-  (ast: AstRoot, code: string, opts: ?InputOptions): FileResult | null,
 };
 
 const transformFromAstRunner = gensync<
@@ -46,12 +42,6 @@ export const transformFromAst: TransformFromAst = (function transformFromAst(
   if (typeof opts === "function") {
     callback = opts;
     opts = undefined;
-  }
-
-  // For backward-compat with Babel 6, we allow sync transformation when
-  // no callback is given. Will be dropped in some future Babel major version.
-  if (callback === undefined) {
-    return transformFromAstRunner.sync(ast, code, opts);
   }
 
   transformFromAstRunner.errback(ast, code, opts, callback);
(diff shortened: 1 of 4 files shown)


packages/babel-core/src/transform-ast.js
```diff
@@ -48,12 +44,6 @@ export const transformFromAst: TransformFromAst = (function transformFromAst(
     opts = undefined;
   }
 
-  // For backward-compat with Babel 6, we allow sync transformation when
-  // no callback is given. Will be dropped in some future Babel major version.
-  if (callback === undefined) {
-    return transformFromAstRunner.sync(ast, code, opts);
```
nicolo-ribaudo: Maybe we could throw a nice error message, like `Starting from Babel 8.0.0, the 'transformFromAst' function expects a callback. If you need to call it synchronously, please use 'transformFromAstSync' instead.`.

Pranav2612000:
Yeah! Sure, just wanted to check if this is the right way to go?
I'll do the same thing for the other functions and add a appropriate error message. Any other changes I should consider?

packages/babel-core/src/transform-ast.js
Pranav2612000: How does this look?
```
  if (callback === undefined) {
    throw new Error(
         "Starting from Babel 8.0.0, the 'transformFromAst' function expects a callback. If you 
         need to call it synchronously, please use 'transformFromAstSync",
    );
  }
```
I'll push the changes once you approve.

Pranav2612000:
Going ahead with doing the same procedure for the other functions.

packages/babel-core/src/transform-ast.js
nicolo-ribaudo: It looks good :+1:

packages/babel-core/src/transform-file.js
```diff
@@ -42,6 +42,20 @@ const transformFileRunner = gensync<[string, ?InputOptions], FileResult | null>(
...
+  if (callback === undefined) {
+    throw new Error(
+      "Starting from Babel 8.0.0, the 'transformFile' function expects a callback. If you need to call it synchronously, please use 'transformFileSync",
+    );
+  }
+
+  transformFileRunner.errback(code, opts, callback);
+}: Function);
 
 export const transformFile: TransformFile = transformFileRunner.errback;
```
nicolo-ribaudo: Can you delete this line? Otherwise `transformFile` is declared twice.

Pranav2612000: Yes. Thanks.

PR #11112, opened 2020-02-08
Respect preserveComments option in tempate.ast()

### LINKED ISSUE #9684: template.ast() doesn't respect preserveComments option.
## Bug Report

**Current Behavior**
```
const testAst = api.template.ast(`test(/* test */ 'test')`, {
  preserveComments: true
});
```

Will have no comment node in the resulting AST.

**Expected behavior/code**
The resulting AST should have comment node

**Environment**
- Babel version(s): 7.3.4
- Node/npm version: 10.15.2
- OS: Windows 10
- Monorepo: no
- How you are using Babel: API.

dentrado:
| Fixed Issues?            | Fixes #9684 

`validate(opts)` added `preserveComments: false` when `preserveComments` was not present in the validated options. This caused the `NO_PLACEHOLDER` options in `builder.js` to contain `preserveComments: false`. So since `NO_PLACEHOLDER` later is merged in to the rest of the options given to template.ast() the preserveComments option was always set to false.

This PR changes `validate(opts)` so it keeps `preserveComments` undefined if it was null/undefined.

### DIFF
--- a/packages/babel-template/test/index.js
+++ b/packages/babel-template/test/index.js
@@ -28,6 +28,18 @@ describe("@babel/template", function() {
     expect(generator(output).code).toBe(comments);
   });
 
+  it("should preserve comments with a flag", function() {
+    const output = template(comments, { preserveComments: true })();
+    expect(generator(output).code).toBe(comments);
+  });
+
+  it("should preserve comments with a flag when using .ast", function() {
+    const output1 = template.ast(comments, { preserveComments: true });
+    const output2 = template({ preserveComments: true }).ast(comments);
+    expect(generator(output1).code).toBe(comments);
+    expect(generator(output2).code).toBe(comments);
+  });
+
   describe("string-based", () => {
     it("should handle replacing values from an object", () => {
       const value = t.stringLiteral("some string value");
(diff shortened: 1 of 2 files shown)


packages/babel-template/test/index.js
```diff
@@ -219,7 +231,7 @@ describe("@babel/template", function() {
     });
   });
 
-  describe.only(".syntacticPlaceholders", () => {
+  describe(".syntacticPlaceholders", () => {
```
nicolo-ribaudo: 😬

kaicataldo: We should probably lint for this 😬

kaicataldo: Actually, it looks like we should be linting for this already: https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-focused-tests.md is a recommended rule.

nicolo-ribaudo: We added it yesterday :stuck_out_tongue: 
https://github.com/babel/babel/pull/11119

nicolo-ribaudo:
Thanks!

PR #11118, opened 2020-02-09
fix(parser): add missing @babel/types dependency

merceyz:

Jest (@SimenB) is testing Yarn 2 (https://github.com/facebook/jest/pull/9476) and their TypeScript build failed because `@babel/parser` is missing a dependency on `@babel/types`.

### DIFF
--- a/packages/babel-parser/package.json
+++ b/packages/babel-parser/package.json
@@ -33,6 +33,9 @@
     "charcodes": "^0.2.0",
     "unicode-12.0.0": "^0.7.9"
   },
+  "dependencies": {
+    "@babel/types": "^7.8.3"
+  },
   "bin": {
     "parser": "./bin/babel-parser.js"
   }


nicolo-ribaudo:
I'm wondering how it's possible that I didn't notice it while testing yarn 2 on this repository 🤔

merceyz:
It's only required in the typescript types.

https://github.com/babel/babel/blob/143d159982bcbe9a31c758b67a888d948c5ff376/packages/babel-parser/typings/babel-parser.d.ts#L8-L16

nicolo-ribaudo:
Oh right, I forgot about it. Related: https://github.com/babel/babel/pull/10315, https://github.com/babel/babel/pull/9924.

I don't think that we should add a whole dependency only used for TS types. I think that we have 3 options:
- Add `@babel/types` in `optionalPeerDependencies`. Users using TS can then add `@babel/types` to their dependencies.
- Publish a package _only_ with the type definitions
- Copy the `.d.ts` file from `@babel/types` to `@babel/parser` when publishing.

cc @babel/babel

ljharb:
Is there an `optionalPeerDependencies`? I think you'd need to add it in peerDeps, and then mark it optional in peerDepsMeta (which would make it required for older npm users).

nicolo-ribaudo:
> Is there an `optionalPeerDependencies`?

It was in my search history and I assumed that it existed :sweat_smile: 

> which would make it required for older npm users

We are dropping Node.js <10.13 in Babel 8, and v10.13.0 has npm v6.12.0 which supports peerDepsMeta. However, I think that this is sub-optimal and I would prefer my third suggestion.

arzyu (approved):

PR #11119, opened 2020-02-10
Add eslint-plugin-jest

JLHwung:
| Any Dependency Changes?  | Added `eslint-plugin-jest` to root devDependencies

This PR introduces `eslint-plugin-jest` to lint our test sources. It catches several test source bugs. Note that some rules are disabled as the change is not that straightforward and thus better toggle on later.

### DIFF
--- a/packages/babel-core/test/option-manager.js
+++ b/packages/babel-core/test/option-manager.js
@@ -94,8 +94,8 @@ describe("option-manager", () => {
       expect(() => {
         loadOptions({
           plugins: [[plugin, null]],
-        }).toThrow(/.plugins[0][1] must be an object, false, or undefined/);
-      });
+        });
+      }).toThrow(".plugins[0][1] must be an object, false, or undefined");
 
       expect(calls).toEqual([]);
     });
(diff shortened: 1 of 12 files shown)


packages/babel-core/test/option-manager.js
```diff
@@ -94,8 +94,8 @@ describe("option-manager", () => {
       expect(() => {
         loadOptions({
           plugins: [[plugin, null]],
-        }).toThrow(/.plugins[0][1] must be an object, false, or undefined/);
-      });
+        });
+      }).toThrow(".plugins[0][1] must be an object, false, or undefined");
```
JLHwung: `toThrow` is chained at wrong function. Unfortunately it will always pass and `eslint-plugin-jest` helps detecting such bugs.

packages/babel-traverse/test/scope.js
```diff
@@ -422,7 +422,11 @@ describe("scope", () => {
           }
         });
 
-        /*if (kind1 !== kind2) {
+        if (kind1 !== kind2) {
```
JLHwung: 👍 for https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-commented-out-tests.md

PR #11122, opened 2020-02-10
Implement "export type {}"

### LINKED ISSUE #10981: `import type` in TypeScript

nicolo-ribaudo:
| Fixed Issues?            | Partially fixes #10981
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Implement `export type { A }` and `export type { A } from "mod"` new TypeScript syntax in parser, transform and generator.
When this syntax is used, the AST nodes have an `exportKind: "type"` property, otherwise it is set to `exportKind: "value"`.

TODO (different PR):
- Add `exportKind: "type"` for type-only export decarations like `export type A = 2` or `export interface B {}`, etc.


packages/babel-parser/test/fixtures/typescript/export/export-type-from/input.ts
```diff
@@ -0,0 +1 @@
+export type { T } from './mod';
```
JLHwung: Can you add a new test
```ts
export type from "./mod"
```
where both `exportDefaultFrom` and `typescript` are enabled? So we can cover the situation when `this.isContextual("type")` is `true` but the lookahead char is not left bracket.

packages/babel-parser/test/fixtures/experimental/export-extensions/default-type-with-typescript/options.json
```diff
@@ -0,0 +1,5 @@
+{
+  "plugins": ["exportDefaultFrom", "typescript"],
+  "sourceType": "module",
+  "throws": "Unexpected token, expected \"=\" (1:17)"
```
nicolo-ribaudo: Did this work before this PR?

JLHwung: It throws before this PR because `isExportDefaultSpecifier` returns false

https://github.com/babel/babel/blob/865d5155c2ec91f1122e94b88e04ee902dfd6677/packages/babel-parser/src/plugins/typescript/index.js#L1614

It may be copied and revised from `flow` plugin. I think we should consider relax the constraint unless banned by the `export-default-from` proposal.

PR #11124, opened 2020-02-11
fix: search for browserslist if esmodules is falsy

### LINKED ISSUE #11123: babel skip searching for browserslist config in some unexpected condition
## Bug Report

**Current Behavior**

If we use targets like this for preset-env

```js
{
  targets: {
    esmodules: false
  }
}
```

babel won't search for browserslist config file 

**Input Code**

**Expected behavior/code**

Should still search for browserslist config file even if targets is set to esmodules: false

- Filename: `babel.config.js`

```js
{
  presets: [
    [
      '@babel/preset-env',
      {
        targets: {
          esmodules: false
        }
      }
    ]
  ]
}
```

**Environment**

```
- Babel version(s): v7.7.5
- Node/npm version: Node v10.18.0 / npm v6.13.4
- OS: OSX 10.12.3
- Monorepo: lerna
```

**Possible Solution**

fengzilong:
| Fixed Issues?            | Fixes #11123 
| Major: Breaking Change?  | Maybe for some user

As stated in issue #11123, with targets

```js
{
  presets: [
    [
      '@babel/preset-env',
      {
        targets: {
          esmodules: false
        }
      }
    ]
  ]
}
```

babel won't search for browserslist config

### DIFF
--- a/packages/babel-helper-compilation-targets/src/index.js
+++ b/packages/babel-helper-compilation-targets/src/index.js
@@ -184,13 +184,15 @@ export default function getTargets(
     targets.browsers = Object.keys(supportsESModules)
       .map(browser => `${browser} ${supportsESModules[browser]}`)
       .join(", ");
+  } else {
+    // remove falsy esmodules to fix `hasTargets` below
+    delete targets.esmodules
   }
 
   // Parse browsers target via browserslist
   const browsersquery = validateBrowsers(targets.browsers);
 
-  const hasTargets = targets.esmodules ||
-    Object.keys(targets).filter(value => value !== TargetNames.esmodules).length > 0;
+  const hasTargets = Object.keys(targets).length > 0;
   const shouldParseBrowsers = !!targets.browsers;
   const shouldSearchForConfig =
     !options.ignoreBrowserslistConfig && !hasTargets;
--- a/packages/babel-helper-compilation-targets/src/index.js
+++ b/packages/babel-helper-compilation-targets/src/index.js
@@ -186,7 +186,7 @@ export default function getTargets(
       .join(", ");
   } else {
     // remove falsy esmodules to fix `hasTargets` below
-    delete targets.esmodules
+    delete targets.esmodules;
   }
 
   // Parse browsers target via browserslist
--- a/packages/babel-helper-compilation-targets/src/index.js
+++ b/packages/babel-helper-compilation-targets/src/index.js
@@ -189,7 +189,8 @@ export default function getTargets(
   // Parse browsers target via browserslist
   const browsersquery = validateBrowsers(targets.browsers);
 
-  const hasTargets = Object.keys(targets).length > 0;
+  const hasTargets = targets.esmodules ||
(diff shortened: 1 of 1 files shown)


packages/babel-helper-compilation-targets/src/index.js
```diff
@@ -189,7 +189,8 @@ export default function getTargets(
   // Parse browsers target via browserslist
   const browsersquery = validateBrowsers(targets.browsers);
 
-  const hasTargets = Object.keys(targets).length > 0;
+  const hasTargets = targets.esmodules ||
```
JLHwung: I think we should normalize target `{ esmodules: false }` to `{}` in

https://github.com/babel/babel/blob/157bb6e8318902c96f0393bc6a3972627bc6d4f9/packages/babel-preset-env/src/normalize-options.js#L121

so if `target.esmodules` presents, if must be `true`.

fengzilong: Hmmm...what I worry about is that some package use `babel-helper-compilation-targets` directly, e.g. @vue/babel-preset-app. should we fix the bug for those packages

https://github.com/vuejs/vue-cli/blob/c8cecffedbf7b19cf930bb2821b5c352bc716a67/packages/%40vue/babel-preset-app/index.js#L17

JLHwung: I see. Since `target.esmodules` is used as a shortcut to set `target.browsers` to be those with esmoduels support. We can delete it after it is consumed.

fengzilong: Seems reasonable ❤️. Updated

packages/babel-helper-compilation-targets/src/index.js
```diff
@@ -184,6 +184,9 @@ export default function getTargets(
     targets.browsers = Object.keys(supportsESModules)
       .map(browser => `${browser} ${supportsESModules[browser]}`)
       .join(", ");
+  } else {
+    // remove falsy esmodules to fix `hasTargets` below
+    delete targets.esmodules;
```
nicolo-ribaudo: Should we also delete it if it is truthy?

fengzilong: User may invoke this method twice or more, keep the truthy value is safer for this case

```js
const getTargets = require( '@babel/helper-compilation-targets' ).default
const targets = { esmodules: true }
const result1 = getTargets( targets )
const result2 = getTargets( targets )
// result1 should be equal to result2 here
```

nicolo-ribaudo: Wouldn't they still be the same?

After the first call, targets has a `browsers` property which makes its behaviour identical to `esmodules: true`.

fengzilong: Oops, sorry I didn't realize targets.browsers have been assigned. Updated

nicolo-ribaudo:
Thanks!

PR #11125, opened 2020-02-11
Use require.resolve instead of the "resolve" package

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  | Yes, removed

The `paths` option is supported starting from node 8.9.0 (https://nodejs.org/api/modules.html#modules_require_resolve_request_options).

The upgrade was done using this small codemod (probably it's overkill for this PR, but I expected many more `resolve.sync` usages :joy:)
<details>

```js
function removeResolve({ types: t }) {
  return {
    visitor: {
      Program(path) {
        for (const el of path.get("body")) {
          if (
            el.isImportDeclaration() &&
            el.get("source").isStringLiteral({ value: "resolve" })
          ) {
            el.remove();
          }
        }
      },
      CallExpression(path) {
        if (!path.get("callee").matchesPattern("resolve.sync")) return;

        const args = [path.get("arguments.0").node];

        if (path.get("arguments").length > 1) {
          // unsupported
          if (!path.get("arguments.1").isObjectExpression()) return;

          const opts = t.objectExpression([]);
          for (const prop of path.get("arguments.1.properties")) {
            // unsupported
            if (!prop.isObjectProperty({ computed: false })) return;

            switch (prop.node.key.name) {
              case "basedir":
                opts.properties.push(
                  t.objectProperty(
                    t.identifier("paths"),
                    t.arrayExpression([prop.node.value])
                  )
                );
                break;

              // unsupported
              default: return;
            }
          }

          args.push(opts);
        }

        path.replaceWith(
          t.callExpression(
            t.memberExpression(
              t.identifier("require"),
              t.identifier("resolve")
            ),
            args
          )
        );
      }
    }
  };
}

```

</details>


packages/babel-core/src/config/files/configuration.js
```diff
@@ -135,7 +134,7 @@ export function* loadConfig(
   envName: string,
   caller: CallerMetadata | void,
 ): Handler<ConfigFile> {
-  const filepath = yield* resolve(name, { basedir: dirname });
+  const filepath = require.resolve(name, { paths: [dirname] });
```
nicolo-ribaudo: This is now always synchronous, but I don't think that it's a real problem.

lib/third-party-libs.js.flow
```diff
@@ -2,12 +2,11 @@
  * Basic declarations for the npm modules we use.
  */
 
-declare module "resolve" {
-  declare export default {
-    (string, {| basedir: string |}, (err: ?Error, res: string) => void): void;
-    sync: (string, {| basedir: string |}) => string;
-  };
-}
+declare var require: {
```
JLHwung: Do we still need this as `require` is Node.js builtin package?

nicolo-ribaudo: Yeah, flow doesn't support it.

PR #11126, opened 2020-02-11
Mark transpiled JSX elements as pure

### LINKED ISSUE #11037: @babel/plugin-transform-react-constant-elements generates non-tree-shakeable output

nicolo-ribaudo:
| Fixed Issues?            | https://twitter.com/wSokra/status/1227163427879161857, closes #11105, fixes #11037
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

`React.createElement` (and often also for other JSX implementation, but it's not guaranteed) calls are pure: they are meant to produce an object which then is used by the React (or other lib) runtime to actually instantiate the component.

We can mark them as such to help minifiers strip unused JSX elements.
Adding it to every component might not seem ideal, but:
1. It can be easily gzipped, since it is exactly the same characters every time
2. It helps minifier remove "real" code, and if something can't be removed minifiers will remove the comment anyway.

If some props aren't pure, minifiers won't remove them even if there is the annotation:

```jsx
import { Foo, Bar } from "components";

export default function Component() {
  <Foo className={getClass()} prop="xx" />;
  return <Bar className="xx"/>
}

// -> After Babel

import { Foo, Bar } from "components";

export default function Component() {
  /*#__PURE__*/ React.createElement(Foo, {
    className: getClass(),
    prop: "xx"
  });
  return /*#__PURE__*/ React.createElement(Bar, {
    className: "xx"
  });
}

// -> After Terser (and Prettier)

import { Foo, Bar } from "components";
export default function Component() {
  return getClass(), React.createElement(Bar, { className: "xx" });
}
```

This PR solves #11037 in a different way than #11105, because we mark JSX elements as pure no matter if they are hoisted or not.

The output from that issue would now be
```js
/*#__PURE__*/ React.createElement("svg", null, /*#__PURE__*/ React.createElement("path", null))
```
Which can be tree-shaken.


packages/babel-plugin-transform-react-jsx/test/fixtures/react/adds-appropriate-newlines-when-using-spread-attribute/output.js
```diff
@@ -1,3 +1,4 @@
+/*#__PURE__*/
 React.createElement(Component, babelHelpers.extends({}, props, {
```
sokra: Maybe `babelHelpers.extends` should get a pure annotation too?

packages/babel-plugin-transform-react-jsx/test/fixtures/react/should-avoid-wrapping-in-extra-parens-if-not-needed/output.js
```diff
@@ -1,4 +1,16 @@
...
+/*#__PURE__*/
+React.createElement("div", null, props.children);
+var x =
+/*#__PURE__*/
+React.createElement(Composite, null, props.children);
+var x =
+/*#__PURE__*/
+React.createElement(Composite, null,
+/*#__PURE__*/
+React.createElement(Composite2, null));
```
sokra: The output code would be a little bit more readable and less verbose if the comment would not be in a separate line but in the same line:

``` js
var x =
/*#__PURE__*/
React.createElement(Composite, null,
/*#__PURE__*/
React.createElement(Composite2, null));
```

vs

``` js
var x = /*#__PURE__*/ React.createElement(Composite, null, /*#__PURE__*/ React.createElement(Composite2, null));
```

But this isn't important...

nicolo-ribaudo: I agree, I'll do it in a separate PR since it isn't strictly related to adding the annotation to JSX elements.

packages/babel-plugin-transform-react-jsx/test/fixtures/react/adds-appropriate-newlines-when-using-spread-attribute/output.js
nicolo-ribaudo: Maybe also other helpers are pure, I'll check 👍

devongovett:
I think it’s worth being careful about this. One could make a custom JSX pragma that is not pure and this would break that. Maybe only do it for the default React.createElement? Or have a whitelist (e.g. preact, etc.)?

JLHwung:
Yeah, I agree on @devongovett , I think we should limit pure annotation when `options.pragma === "React.createElement"` and `options.pragmaFrag === "React.Fragment"`.

We can offer limited whitelist and finally a `pragmaIsPure` options for custom pragmas.

nicolo-ribaudo:
I don't think that we can have a whitelist, because the pragma isn't based on the imported module but on the variable name (for example, `h` for Preact).

@developit When this PR is merged, Preact can benefit from these changes if the `pure: true` option is added to https://preactjs.com/guide/v10/getting-started/#setting-up-jsx and https://preactjs.com/guide/v8/switching-to-preact/#via-babel.

@JLHwung I didn't add the check for `pragmaFrag` because fragments are transpiled as `pragmaFn(PragmaFrag, { props })`, so the function is still always the pragma.

packages/babel-plugin-transform-react-constant-elements/src/index.js
```diff
@@ -1,6 +1,5 @@
 import { declare } from "@babel/helper-plugin-utils";
 import { types as t } from "@babel/core";
-import annotateAsPure from "@babel/helper-annotate-as-pure";
```
JLHwung: nit: should remove `@babel/helper-annotate-as-pure` from package.json.

nicolo-ribaudo:
I was wondering if it would make sense to only add annotations to elements in the top level scope, since they are the ones preventing tree-shaking.

JLHwung:
@nicolo-ribaudo Rollup can also shakes non top level functions, example is [here](https://rollupjs.org/repl/?version=1.29.1&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMCU3QiUyMG0lMjAlN0QlMjBmcm9tJTIwJy4lMkZjb21wb25lbnQuanMnJTNCJTVDbm0oKSUzQiUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTdEJTJDJTdCJTIybmFtZSUyMiUzQSUyMmNvbXBvbmVudC5qcyUyMiUyQyUyMmNvZGUlMjIlM0ElMjIlNUNuJTVDbmV4cG9ydCUyMGZ1bmN0aW9uJTIwbSgpJTIwJTdCJTVDbiU1Q3R2YXIlMjBfcmVmMiUyMCUzRCU1Q24lMkYqJTIzX19QVVJFX18qJTJGJTVDblJlYWN0LmNyZWF0ZUVsZW1lbnQoJTVDJTIyc3ZnJTVDJTIyJTJDJTIwbnVsbCUyQyUyMCUyRiolMjNfX1BVUkVfXyolMkZSZWFjdC5jcmVhdGVFbGVtZW50KCU1QyUyMnBhdGglNUMlMjIlMkMlMjBudWxsKSklM0IlNUNuJTIwJTIwcmV0dXJuJTIwdHJ1ZSUzQiU1Q24lN0QlMjIlMkMlMjJpc0VudHJ5JTIyJTNBZmFsc2UlN0QlNUQlMkMlMjJvcHRpb25zJTIyJTNBJTdCJTIyZm9ybWF0JTIyJTNBJTIyY2pzJTIyJTJDJTIybmFtZSUyMiUzQSUyMm15QnVuZGxlJTIyJTJDJTIyYW1kJTIyJTNBJTdCJTIyaWQlMjIlM0ElMjIlMjIlN0QlMkMlMjJnbG9iYWxzJTIyJTNBJTdCJTdEJTdEJTJDJTIyZXhhbXBsZSUyMiUzQW51bGwlN0Q=).

nicolo-ribaudo:
Yeah, I'm wondering how common it is to have unused elements inside functions.
`env === "development" && <Foo />` would be removed even without the annotation.

packages/babel-plugin-transform-function-name/test/fixtures/function-name/export-default-arrow-renaming-3/output.js
```diff
@@ -7,11 +7,12 @@ exports["default"] = void 0;
 
 var _default = function _default(_ref) {
   var _onClick = _ref.onClick;
-  return React.createElement("div", {
-    onClick: function onClick() {
-      return _onClick();
-    }
-  });
+  return (/*#__PURE__*/React.createElement("div", {
+      onClick: function onClick() {
```
nicolo-ribaudo: I'll check why this is double indented :thinking:

nicolo-ribaudo: Fixed

packages/babel-preset-react/src/index.js
```diff
@@ -8,15 +8,19 @@ import transformReactJSXSelf from "@babel/plugin-transform-react-jsx-self";
...
-  const pragmaFrag = opts.pragmaFrag;
-  const throwIfNamespace =
-    opts.throwIfNamespace === undefined ? true : !!opts.throwIfNamespace;
+  const {
+    pragma,
+    pragmaFrag,
+    pure,
+    throwIfNamespace = true,
+    useSpread,
+    runtime = "classic",
```
nicolo-ribaudo: This is a breaking change from the "new JSX" PR, because that PR accepts `runtime: false` as an alias for `runtime: "classic"`.

If this PR ends up being merged in a different release, I'll have to revert this line.

nicolo-ribaudo:
I have finished adapting this PR after https://github.com/babel/babel/pull/11154 (I copied the same logic to the new transform)

packages/babel-preset-react/src/index.js
lunaruan: Yeah, this should be okay. No preference for whether `runtime: false` should alias to `runtime: 'classic'` as long as `runtime: 'classic'` is the default behavior

packages/babel-plugin-transform-react-jsx/src/transform-classic.js
```diff
@@ -46,29 +53,25 @@ export default declare((api, options) => {
...
         }
       }
 
       state.set("jsxIdentifier", createIdentifierParser(pragma));
       state.set("jsxFragIdentifier", createIdentifierParser(pragmaFrag));
       state.set("usedFragment", false);
-      state.set("pragmaSet", pragmaSet);
-      state.set("pragmaFragSet", pragmaFragSet);
+      state.set("pragmaSet", pragma !== DEFAULT.pragma);
+      state.set("pragmaFragSet", pragmaFrag !== DEFAULT.pragmaFrag);
```
Ailrun: This part causes an issue when users want to use custom `pragma` with default `pragmaFrag` (namely the `React.Fragment`).
This is the case in the [emotion babel preset](https://github.com/emotion-js/emotion/tree/master/packages/babel-preset-css-prop), and causes a problem in [gatsby](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-emotion/src/gatsby-node.js) too

PR #11129, opened 2020-02-11
Typescript: Preserve decorated definite class properties

jamescdavis:
| Fixed Issues?            |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is a follow-up to #8238, but for class properties marked as definite (using the non-null assertion `!`)

### Current behavior:
The `babel-plugin-transform-typescript` removes class properties marked as definite regardless if decorators are assigned to it or not.

```ts
// Original
class C {
  @foo
  d;
  @foo
  e = 3;
  @foo
  f!;
}

// Transformed
class C {
  @foo
  d;
  @foo
  e = 3;
  // <-- Missing `f` field
}
```

### Expected behavior:
The babel-plugin-transform-typescript should only remove a definite class property if it has no decorators.

```ts
// Original
class C {
  @foo
  d;
  @foo
  e = 3;
  @foo
  f!;
}

// Transformed
class C {
  @foo
  d;
  @foo
  e = 3;
  @foo
  f;
}
```

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -74,7 +74,9 @@ export default declare(
             );
           }
 
-          path.remove();
+          if (!node.decorators) {
+            path.remove();
+          }
         } else if (!allowDeclareFields && !node.value && !node.decorators) {
           path.remove();
         }
@@ -84,6 +86,7 @@ export default declare(
         if (node.readonly) node.readonly = null;
         if (node.optional) node.optional = null;
         if (node.typeAnnotation) node.typeAnnotation = null;
+        if (node.definite) node.definite = null;
       },
       method({ node }) {
         if (node.accessibility) node.accessibility = null;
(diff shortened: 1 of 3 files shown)


nicolo-ribaudo:
Thanks!

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -74,7 +74,11 @@ export default declare(
             );
           }
 
-          path.remove();
+          if (node.definite && node.decorators) {
+            node.definite = null;
+          } else {
+            path.remove();
+          }
```
JLHwung: nit: I feel like it's clearer to only check against `!node.decorators` here (like we did in the `else if` branch).
```js
if (!node.decorators) {
  path.remove();
}
```
and then reset `definite` later (like we did for `abstract`, `accessibility`, etc.)
```js
if (node.definite) node.definite = null;
```

jamescdavis: that's a *definite* improvement in clarity 😉

JLHwung:
@jamescdavis Can you rebase on upstream master? The e2e error will be fixed after rebasing.

PR #11133, opened 2020-02-12
Skip newlines around inline `#__PURE__` annotations

nicolo-ribaudo:
| Fixed Issues?            | Fixes&nbsp; https://github.com/babel/babel/pull/11126#discussion_r377603838
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR is mostly to improve the output readability, since the unindented newlines looked really bad.

I special-cased `/*#__PURE__*/` annotations because I think that we should preserve the old behavior for normal comments.

### DIFF
--- a/packages/babel-generator/src/printer.js
+++ b/packages/babel-generator/src/printer.js
@@ -9,6 +9,7 @@ import * as generatorFunctions from "./generators";
 const SCIENTIFIC_NOTATION = /e/i;
 const ZERO_DECIMAL_INTEGER = /\.0+$/;
 const NON_DECIMAL_LITERAL = /^0[box]/;
+const PURE_ANNOTATION_RE = /^\s*[@#]__PURE__\s*$/;
 
 export type Format = {
   shouldPrintComment: (comment: string) => boolean,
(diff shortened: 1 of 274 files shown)


packages/babel-generator/src/printer.js
```diff
@@ -584,15 +589,26 @@ export default class Printer {
...
-      this._printComment(comment);
+    if (
+      inlinePureAnnotation &&
+      comments.length === 1 &&
+      PURE_ANNOTATION_RE.test(comments[0].value)
+    ) {
+      this._printComment(
+        comments[0],
+        // Keep newlines if the comment marks a standalone call
+        this._buf.hasContent() && !this.endsWith("\n"),
```
nicolo-ribaudo: This is for cases like https://github.com/babel/babel/blob/master/packages/babel-plugin-proposal-async-generator-functions/test/fixtures/async-generators/expression/output.js#L1, where I think that we should keep the `\n`.

PR #11134, opened 2020-02-12
Mark `wrapNativeSuper` and `wrapRegExp` as pure

nicolo-ribaudo:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Technically there are also other pure helpers (e.g. `asyncToGenerator`), but annotating them won't help minifiers because their result is always directly used by the compiled output.

### DIFF
--- a/packages/babel-plugin-transform-classes/src/transformClass.js
+++ b/packages/babel-plugin-transform-classes/src/transformClass.js
@@ -6,6 +6,7 @@ import ReplaceSupers, {
 import optimiseCall from "@babel/helper-optimise-call-expression";
 import * as defineMap from "@babel/helper-define-map";
 import { traverse, template, types as t } from "@babel/core";
+import annotateAsPure from "@babel/helper-annotate-as-pure";
 
 type ReadonlySet<T> = Set<T> | { has(val: T): boolean };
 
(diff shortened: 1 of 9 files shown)


packages/babel-plugin-transform-classes/src/transformClass.js
```diff
@@ -565,11 +566,14 @@ export default function transformClass(
...
 
     if (classState.isDerived) {
-      const arg = classState.extendsNative
-        ? t.callExpression(classState.file.addHelper("wrapNativeSuper"), [
-            t.cloneNode(superName),
-          ])
-        : t.cloneNode(superName);
+      let arg = t.cloneNode(superName);
+      if (classState.extendsNative) {
+        arg = t.callExpression(classState.file.addHelper("wrapNativeSuper"), [
```
JLHwung: Can we add pure notation to the helper sources so we don't worry about if the helper needs pure annotation when added?

nicolo-ribaudo: I have thought about it, but the problem is that the annotation should be added to the `CallExpression`, and the `æddHelper` function only returns the callee.

PR #11136, opened 2020-02-13
fix @babel/traverse can't use path.remove() with noScope

### LINKED ISSUE #11128: @babel/traverse noScope can't path.remove()
When I upgraded from v7.7.7 to v7.8.4, I found that there was a problem with the noScope option of @babel/traverse.

Error message:
```
e:\mypath\babel\node_modules\@babel\traverse\lib\path\removal.js:39
  Object.keys(bindings).forEach(name => this.scope.removeBinding(name));
                                                   ^

TypeError: Cannot read property 'removeBinding' of null
    at e:\mypath\babel\node_modules\@babel\traverse\lib\path\removal.js:39:52
    at NodePath._removeFromScope (e:\mypath\babel\node_modules\@babel\traverse\lib\path\removal.js:39:25)
    at NodePath.remove (e:\mypath\babel\node_modules\@babel\traverse\lib\path\removal.js:22:8)
    at AssignmentExpression (e:\mypath\babel\test.js:6:14)
    at NodePath._call (e:\mypath\babel\node_modules\@babel\traverse\lib\path\context.js:55:20)
    at NodePath.call (e:\mypath\babel\node_modules\@babel\traverse\lib\path\context.js:42:17)
    at NodePath.visit (e:\mypath\babel\node_modules\@babel\traverse\lib\path\context.js:90:31)
    at TraversalContext.visitQueue (e:\mypath\babel\node_modules\@babel\traverse\lib\context.js:112:16)
    at TraversalContext.visitSingle (e:\mypath\babel\node_modules\@babel\traverse\lib\context.js:84:19)
```

Code to reproduce the problem:
```js
var traverse = require("@babel/traverse").default;
var parser = require("@babel/parser");

traverse(parser.parse("a=1"), {
    AssignmentExpression(path) {
        path.remove();
    },
    noScope: true,
});
```

liuxingbaoyu:
| Fixed Issues?            | Fixes #11128  
| Minor: New Feature?      |No
| Documentation PR Link    | 

### DIFF
--- a/packages/babel-traverse/src/path/removal.js
+++ b/packages/babel-traverse/src/path/removal.js
@@ -7,7 +7,9 @@ export function remove() {
   this._assertUnremoved();
 
   this.resync();
-  this._removeFromScope();
+  if (!this.opts || !this.opts.noScope) {
+    this._removeFromScope();
+  }
 
   if (this._callRemovalHooks()) {
     this._markRemoved();
--- a/packages/babel-traverse/test/removal.js
+++ b/packages/babel-traverse/test/removal.js
@@ -33,4 +33,16 @@ describe("removal", function() {
       expect(generateCode(rootPath)).toBe("x = () => {};");
     });
   });
+
+  it("remove with noScope", function() {
+    const ast = parse("a=1");
+    traverse(ast, {
+      AssignmentExpression: function(path) {
+        path.remove();
+      },
+      noScope: true,
+    });
+
+    expect(generate(ast).code).toBe("");
+  });
 });


packages/babel-traverse/src/path/removal.js
```diff
@@ -7,7 +7,9 @@ export function remove() {
   this._assertUnremoved();
 
   this.resync();
-  this._removeFromScope();
+  if (this.opts && !this.opts.noScope) {
```
nicolo-ribaudo: When is `this.opts` undefined?

liuxingbaoyu: I don't know ... CI reports an error. I didn't find out exactly what caused it. I also feel very confused, but in order not to affect compatibility, I added this judgment.
[CI Link](https://travis-ci.com/babel/babel/jobs/286667571)

packages/babel-traverse/test/removal.js
```diff
@@ -33,4 +33,15 @@ describe("removal", function() {
       expect(generateCode(rootPath)).toBe("x = () => {};");
     });
   });
+
+  it("remove with noScope", function() {
+    const ast = parse("a=1");
+    traverse(ast, {
+      AssignmentExpression: function(path) {
+        path.remove();
+      },
```
JLHwung: I guess we should add `noScope: true` here?

packages/babel-traverse/src/path/removal.js
nicolo-ribaudo: The check should be `!this.opts || !this.opts.noScope`. If `opts` is not defined, `noScope` is not true so the scope is created.

liuxingbaoyu: oh no! You're right.

packages/babel-traverse/test/removal.js
liuxingbaoyu: Yes, I'm a fool...

packages/babel-traverse/src/path/removal.js
liuxingbaoyu: I stumbled upon [this](https://github.com/babel/babel/blob/f544753bb8c9c7a470d98e897b089fd31b83d1f6/packages/babel-traverse/src/path/context.js#L114-L119) just now...

PR #11137, opened 2020-02-13
Breaking: align babel-eslint-parser & espree

kaicataldo:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

As I was switching our test suite to use Jest, I noticed that we're not checking for extraneous properties on nodes. I'm still looking into which properties might be useful to keep (even if they're not defined by ESTree), but this way we can be explicit about what extra properties we decide to keep.

This is a breaking change because it's removing some properties from nodes in the `@babel/eslint-parser` AST.


eslint/babel-eslint-parser/src/convert/convertAST.js
```diff
@@ -23,7 +19,15 @@ function convertNodes(ast, code) {
...
     },
     exit(path) {
-      const node = path.node;
+      const { node } = path;
+
+      if (Object.hasOwnProperty.call(node, "extra")) {
+        delete node.extra;
+      }
+
+      if (node.loc && Object.hasOwnProperty.call(node.loc, "identifierName")) {
```
kaicataldo: Seems unlikely ESLint would need this, but please let me know if I'm wrong.

eslint/babel-eslint-parser/src/convert/convertAST.js
```diff
@@ -23,7 +19,15 @@ function convertNodes(ast, code) {
       }
     },
     exit(path) {
-      const node = path.node;
+      const { node } = path;
+
+      if (Object.hasOwnProperty.call(node, "extra")) {
```
kaicataldo: Seems unlikely ESLint would need this, but please let me know if I'm wrong.

eslint/babel-eslint-parser/test/index.js
```diff
@@ -1,19 +1,43 @@
...
-import assertImplementsAST from "./helpers/assert-implements-ast";
 
-const babelOptions = {
+const BABEL_OPTIONS = {
   configFile: require.resolve(
     "@babel/eslint-shared-fixtures/config/babel.config.js",
   ),
 };
+const ALLOWED_PROPERTIES = [
+  "importKind",
```
kaicataldo: These properties seem useful for Flow - should we keep them for `eslint-plugin-flow`?

eslint/babel-eslint-parser/src/convert/convertAST.js
```diff
@@ -95,9 +100,9 @@ function convertNodes(ast, code) {
 function convertProgramNode(ast) {
   ast.type = "Program";
   ast.sourceType = ast.program.sourceType;
-  ast.directives = ast.program.directives;
   ast.body = ast.program.body;
   delete ast.program;
+  delete ast.errors;
```
kaicataldo: Extra property that ESLint doesn't use (though we are planning to implement a top-level `recoverableErrors` property - you can find more info [here](https://github.com/eslint/rfcs/tree/master/designs/2019-recoverable-error-handling)).

eslint/babel-eslint-parser/package.json
```diff
@@ -31,7 +31,6 @@
     "@babel/core": "^7.2.0",
     "@babel/eslint-shared-fixtures": "*",
     "eslint": "^6.0.1",
-    "espree": "^6.0.0",
```
kaicataldo: Thoughts on this? Seems like we should ensure we're using the version of Espree that's included with the version of ESLint we're testing against.

eslint/babel-eslint-parser/src/convert/convertAST.js
```diff
@@ -107,12 +107,7 @@ function convertProgramNode(ast) {
   if (ast.comments.length) {
     const lastComment = ast.comments[ast.comments.length - 1];
 
-    if (!ast.tokens.length) {
-      // if no tokens, the program starts at the end of the last comment
```
kaicataldo: This behavior doesn't match Espree's (it probably did at one point and has since changed).

eslint/babel-eslint-parser/test/index.js
```diff
@@ -306,13 +306,11 @@ describe("Babel and Espree", () => {
     expect(babylonAST.tokens[3].value).toEqual("#");
   });
 
-  // eslint-disable-next-line jest/no-disabled-tests
-  it.skip("empty program with line comment", () => {
+  it("empty program with line comment", () => {
```
kaicataldo: This is now aligned! 🎉

eslint/babel-eslint-parser/test/index.js
```diff
@@ -1,19 +1,41 @@
-import assert from "assert";
 import espree from "espree";
```
nicolo-ribaudo: If we are removing this from `package.json`, we should explicitly import the correct one.

```js
const espreePath = require.resolve("espree", {
  paths: [path.dirname(require.resolve("eslint"))]
]);
const espree = require(espreePath);
```

kaicataldo: That makes sense!

eslint/babel-eslint-parser/test/index.js
```diff
@@ -1,51 +1,85 @@
...
-const babelOptions = {
+const BABEL_OPTIONS = {
   configFile: require.resolve(
     "@babel/eslint-shared-fixtures/config/babel.config.js",
   ),
 };
+const ALLOWED_PROPERTIES = [
+  "importKind",
+  "exportKind",
+  "variance",
```
kaicataldo: This is used in [`eslint-plugin-flowtype`](https://github.com/gajus/eslint-plugin-flowtype/search?q=variance&unscoped_q=variance).

eslint/babel-eslint-parser/test/index.js
```diff
@@ -1,51 +1,85 @@
...
+  "typeArguments",
```
kaicataldo: This is _not_ used in [eslint-plugin-flowtype](https://github.com/gajus/eslint-plugin-flowtype/search?q=typearguments&unscoped_q=typearguments). Should we remove this?

eslint/babel-eslint-parser/test/index.js
```diff
@@ -1,51 +1,85 @@
...
-import assertImplementsAST from "./helpers/assert-implements-ast";
 
-const babelOptions = {
+const BABEL_OPTIONS = {
   configFile: require.resolve(
     "@babel/eslint-shared-fixtures/config/babel.config.js",
   ),
 };
+const ALLOWED_PROPERTIES = [
+  "importKind",
```
kaicataldo: I'm assuming we want to leave these properties since they might be useful for Flow-related ESLint rules. `exportKind` is already used in [`eslint-plugin-flowtype`](https://github.com/gajus/eslint-plugin-flowtype/search?q=exportKind&unscoped_q=exportKind), so it seems like we should include both properties.

eslint/babel-eslint-parser/test/index.js
nicolo-ribaudo: I think that we should keep it: if espree doesn't support flow, we don't need to align in this case.

eslint/babel-eslint-parser/package.json
ljharb: i think it should be a combo peer dep and dep, that way `npm ls` will fail (and `npm install` in npm 7+) when the peer dep isn't met.

eslint/babel-eslint-parser/test/index.js
ljharb: `eslint-plugin-import` definitely uses these.

eslint/babel-eslint-parser/src/convert/convertAST.js
JLHwung: AFAIK `node.extra` is only used internally in babel-parser. It is a collection of per-node parser features, `parenthesized`, `trailingComma`, etc. So we can safely remove `node.extra`.

eslint/babel-eslint-parser/package.json
kaicataldo: @nicolo-ribaudo @JLHwung @existentialism Thoughts?

nicolo-ribaudo:
@kaicataldo The failing tests are related to the changes in this PR.

kaicataldo:
Yikes, had a bad rebase. Fixing up now.

eslint/babel-eslint-parser/src/convert/convertAST.js
kaicataldo: Got it, thanks!

kaicataldo:
`reflog` for the win!

eslint/babel-eslint-parser/test/index.js
nicolo-ribaudo: We might want to add tests for third-party ESLint plugins in the e2e Circle CI workflow.

eslint/babel-eslint-parser/package.json
nicolo-ribaudo: Now that we use `require.resolve` to get ESLint's dep, this shouldn't be needed anymore.

eslint/babel-eslint-parser/src/convert/convertAST.js
```diff
@@ -1,14 +1,10 @@
...
 
 function convertNodes(ast, code) {
-  const state = { source: code };
   const astTransformVisitor = {
     noScope: true,
     enter(path) {
-      const node = path.node;
-
-      // private var to track original node type
-      node._babelType = node.type;
```
nicolo-ribaudo: I was wondering why we had this in the first place, but a quick search on GitHub ([search](https://github.com/search?p=1&q=_babelType&type=Code)) shows that it's only used in snapshot tests output: it can be safely removed.

PR #11141, opened 2020-02-13
breaking: remove useSpread and useBuiltIns option

### LINKED ISSUE #9652: Add option to preset-react / transform-react-jsx to output object spread syntax

JLHwung:
| Fixed Issues?            | Fixes #9652 
| Tests Added + Pass?      | Todo
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR removes `useSpread` and `useBuiltIns` options in both `@babel/plugin-transform-jsx` and `@babel/preset-react`. A warning message with migration config will be printed if users are using these options.

This PR also depends on #11740, I will rebased once it is merged to `next-8`.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11141"></a>


nicolo-ribaudo:
:+1: for removing them.

packages/babel-helper-builder-react-jsx/src/index.js
```diff
@@ -156,82 +156,26 @@ You can turn on the 'throwIfNamespace' flag to bypass this warning.`,
...
-    const objs = [];
-
-    const { useSpread = false } = file.opts;
-    if (typeof useSpread !== "boolean") {
-      throw new Error(
-        "transform-react-jsx currently only accepts a boolean option for " +
-          "useSpread (defaults to false)",
+    if ("useSpread" in file.opts) {
+      console.warn(
+        "useSpread is always enabled in Babel 8, please remove it from the config.\n" +
```
nicolo-ribaudo: Can we move this warning to the top-level plugin function?

JLHwung: I think it is okay to stay here otherwise we need duplicate these between plugin-react-jsx and preset-react. In Babel 9 we can even remove these checks.

JLHwung: Oh I realize that in presets we have to pass through `useSpread` and `useBuiltins` for this warning to pop up. It's a little bit awkward. I will move to plugin/preset then.

JLHwung:
The e2e test is expected, `create-react-app` has to remove the `useBuiltIns` options here when upgrading to babel 8. https://github.com/facebook/create-react-app/blob/8ba0ccb514589433bb7ccab7ed067aa5973543d2/packages/babel-preset-react-app/create.js#L101

nicolo-ribaudo:
@JLHwung I have pushed a commit to comment out that `useBuiltIns: true` in `create-react-app`: it's important that the test is green, otherwise we won't notice new failures.

However, I noticed that it's throwing about `useSpread`, but `create-react-app` never uses it :thinking:

scripts/integration-tests/e2e-create-react-app.sh
```diff
@@ -21,11 +21,20 @@ cd tmp/create-react-app || exit
 #                                   TEST                                       #
 #==============================================================================#
 
+# !!! WARNING !!!
```
hzoo, line 24: maybe a useless github action. but could be cool to comment/uncomment code to when a github issue is merged!

packages/babel-preset-react/src/index.js
```diff
@@ -12,8 +12,25 @@ export default declare((api, opts) => {
   const throwIfNamespace =
     opts.throwIfNamespace === undefined ? true : !!opts.throwIfNamespace;
   const development = !!opts.development;
-  const useBuiltIns = !!opts.useBuiltIns;
-  const { useSpread } = opts;
+  if ("useSpread" in opts) {
+    throw new Error(
+      '@babel/preset-react: "useSpread" is always enabled in Babel 8, please remove it from the config.\n',
```
hzoo, line 17: I wonder what it would take for us to just include this in babel itself, or point to running babel-upgrade if we want to implement that later

packages/babel-plugin-transform-react-jsx/src/index.js
```diff
@@ -15,6 +15,26 @@ export default declare((api, options) => {
   const JSX_ANNOTATION_REGEX = /\*?\s*@jsx\s+([^\s]+)/;
   const JSX_FRAG_ANNOTATION_REGEX = /\*?\s*@jsxFrag\s+([^\s]+)/;
 
+  if ("useSpread" in options) {
+    throw new Error(
+      'transform-react-jsx: "useSpread" is always enabled in Babel 8, please remove it from the config.\n',
```
hzoo, line 20: Although I guess we could also change the message. Not sure what would be better, "Since the "useSpread" is enabled by default in Babel 8, you can remove it from your config" @existentialism

packages/babel-preset-react/src/index.js
nicolo-ribaudo: That is already in preset-env, so it falls in "the future or preset-env (or its end)" discussion 😉

packages/babel-plugin-transform-react-jsx/src/index.js
nicolo-ribaudo: "Enabled by default" implies that you can disable it to me, and "you can remove it from your config" doesn't convey that it _must_ be removed.

packages/babel-plugin-transform-react-jsx/src/transform-classic.js
```diff
@@ -19,6 +19,27 @@ export default declare((api, options) => {
   const JSX_ANNOTATION_REGEX = /\*?\s*@jsx\s+([^\s]+)/;
   const JSX_FRAG_ANNOTATION_REGEX = /\*?\s*@jsxFrag\s+([^\s]+)/;
 
+  if ("useSpread" in options) {
+    throw new Error(
+      'transform-react-jsx: Since "useSpread" is always enabled in Babel 8, you must remove it from your config.',
```
existentialism, line 24: Should we make these messages a bit nicer for future? Maybe something like:

```suggestion
      'transform-react-jsx: Since Babel 8, "useSpread" is enabled by default. Please remove it from your config.',
```

JLHwung: See https://github.com/babel/babel/pull/11141#discussion_r443252782 for wording issues about "enabled by default" and "must" 😄

existentialism: Okay, howabout:

```suggestion
      'transform-react-jsx: Since Babel 8, an inline object with spread elements is always used, and the "useSpread" option is no longer available. Please remove it from your config.',
```

packages/babel-plugin-transform-react-jsx/src/transform-classic.js
```diff
@@ -19,6 +19,27 @@ export default declare((api, options) => {
...
+  if ("useSpread" in options) {
+    throw new Error(
+      'transform-react-jsx: Since Babel 8, an inline object with spread elements is always used, and the "useSpread" option is no longer available. Please remove it from your config.',
+    );
+  }
+
+  if ("useBuiltIns" in options) {
+    const useBuiltInsFormatted = JSON.stringify(options.useBuiltIns);
+    throw new Error(
+      `transform-react-jsx: Since "useBuiltIns" is removed in Babel 8, you must remove it from your config.
```
existentialism, line 31: Here too:

```suggestion
      `transform-react-jsx: Since Babel 8, an inline object with spread elements is always used, and the "useSpread" option is no longer available. Please remove it from your config.
```

PR #11142, opened 2020-02-13
Use Yarn 2

nicolo-ribaudo:
| Q                        | A 
| ------------------------ | ---
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  | :+1:
| License                  | MIT

PR #11143, opened 2020-02-13
Add Yarn 2 files to .gitignore

nicolo-ribaudo:

This is to avoid having a lot of untracked files when switching from `next-8-dev` to `master`, after merging https://github.com/babel/babel/pull/11142

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,10 @@
 .npmrc
 
+.yarn/cache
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
I'm merging this even if the other PR isn't ready, it doesn't hurt anyway :stuck_out_tongue:

PR #11148, opened 2020-02-16
Allow await when it is not in AsyncArrowHead

### LINKED ISSUE #10988: await should be allowed when it is not in AsyncArrowHead

arku:
| Fixed Issues?            | Fixes #10988 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR fixes the parser error thrown for the input `(await) => {}`. I tried to fix the other cases mentioned in the issue description and after struggling with delaying the await position check in case of async call expressions, I reached out to @JLHwung for help. He added that the delay check is indeed difficult and it is best to refactor the await position tracking with a scope based recording mechanism that is non-trivial.


packages/babel-parser/src/tokenizer/state.js
```diff
@@ -59,6 +59,7 @@ export default class State {
   // Flags to track
   inParameters: boolean = false;
   maybeInArrowParameters: boolean = false;
+  maybeInAsyncArrowHead: boolean = false;
```
nicolo-ribaudo: It's not clear to me what the difference between these two flags should be. Isn't the arrow head its parameters?

JLHwung: The `maybeInAsyncArrowHead` lives across function declarations.
```js
async (foo = function (await) {}) => {}
```
When parsing `await`, `maybeInAsyncArrowHead` is `true` but `maybeInArrowParameters` is `false`.

nicolo-ribaudo: We should add a comment in the code explaining it 👍

arku: @nicolo-ribaudo Done. Let me know if we can make it better.

packages/babel-parser/src/parser/expression.js
```diff
@@ -560,10 +560,15 @@ export default class ExpressionParser extends LValParser {
       stop: false,
     };
     do {
+      const oldMaybeInAsyncArrowHead = this.state.maybeInAsyncArrowHead;
+      if (state.maybeAsyncArrow) {
+        this.state.maybeInAsyncArrowHead = state.maybeAsyncArrow;
+      }
```
nicolo-ribaudo: Nit: inside the `if` `state.maybeAsyncArrow` is always `true`.

```suggestion
      if (state.maybeAsyncArrow) this.state.maybeInAsyncArrowHead = true;
```

arku: @nicolo-ribaudo Ah, yeah. Just updated the PR.

PR #11154, opened 2020-02-19
Add experimental version of the `babel-plugin-transform-react-jsx` transform

lunaruan:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    |
| Any Dependency Changes?  |

As part of React's [RFC to simplify element creation](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md), we want to update the `babel-plugin-transform-react-jsx` transform so that it uses `React.jsx` instead of `React.createElement` and adds to ability to auto import React for JSX.

The main differences between React.jsx/React.jsxDEV and React.createElement are:
1.) key is now passed as an explicit argument rather than through props
3.) children are now passed through props rather than as an explicit argument
4.) props must always be an object
5.) __source and and __self (added through separate Babel plugins) are now passed as separate arguments into React.jsxDEV rather than through props
6.) In production, we will use `jsxs(type, props, maybeKey)` if there MORE THAN ONE child in the AST and `jsx(type, props, maybeKey)` if  there is ONE OR FEWER children in the AST. 
7.) We want to deprecate key spread through props, so we will use `React.createElement` to transform the case where props spread comes before a key (ex. `<div {...props} key="Hi" />`) so that we differentiate between prop spread after a key (ex. `<div key={foo} {...props} />`, which will override the key) and warn appropriately.
8.) In development mode we will use `jsxDEV(type, props,  key, isStaticChildren, source, self)`, where isStaticChildren is true when you should use `jsxs` and false if you should use `jsx`. This will be specified using a `mode: development` option. (One concern here is that someone might accidentally forget to set this. We can make this option required so that this doesn't happen)

This PR also adds two options for auto import to React:
1.) `importSource`: The React module to import from. Defaults to react.
2.) `runtime: 'classic' | 'automatic'`. In Babel 7.9 this will default to `classic` and in Babel 8 it will default to `automatic`. In classic mode, we will use `createElement` without auto importing (ie. the current behavior). In automatic mode, we will use `jsx/jsxs` and auto import. If the sourceType is module, we will use named exports (`import {jsx as _jsx} from "react"`; JSX compiles to `_jsx` etc.). If the sourceType is script, we will use require (`var _react = _interopRequireWildcard(require("react"));`. jSX compiles to `_react.jsx` etc.)

Automatic runtime adds two pragmas (`jsxAutoImport` and `jsxImportSource`) that allow users to specify `autoImport` and `importSource` in the docblock. Classic runtime will still have `pragma` and `pragmaFrag`.

In addition, for Babel 8, development mode will require a separate Babel plugin, `@babel/plugin-transform-react-jsx-development`. This mode will automatically add `__source` and `__self` for both runtimes so that users will not need to use separate plugins. Adding the `source` and `self` plugins will be a no-op.


nicolo-ribaudo:
> 1.) importSource: The React module to import from. Defaults to react.
> 2.) autoImport: The type of import. Defaults to none.
> ...

I think that this option has too many possible values. There are only three possible cases:
1. The `jsx` function is a named export of the module (considering `import foo` as a shorthand for `import { default as foo }`, thus the name is `default`)
2. The `jsx` function is a property of a named exported object (again, `default` is a normal name).
3. The `jsx` function is a property of a global object.

(1) and (2) can be "collapsed" together, using the `object.property` notation that we are already using for the `pragma` option.
(3) can be obtained by disabling `importSource`.

The `require` vs `import` case should be automatically handled by `@babel/helper-module-imports`, which detects the current `sourceType`.

In these examples I'm using the `pragma` option name rather than `autoImport`, because I think that it fits better my proposed changes. Keep in mind that some of the examples in the following table aren't valid React imports, but they might be useful for other libraries.

Input code: `<div />`

| **sourceType** | **importSource** | **pragma** | **output** |
|:--:|:--:|:--:|:--:|
| `module` | `false` | `React.jsx` | `React.jsx("div")` |
| `script` | `false` | `React.jsx` | `React.jsx("div")` |
| `module` | `"react"` | `jsx` | `import { jsx } from "react";`<br/>`jsx("div");` |
| `script` | `"react"` | `jsx` | `const _react = require("react");`<br/>`(0, _react.jsx)("div");` |
| `module` | `"react"` | `default` | `import { default as _react } from "react";`<br/>`_react("div");` |
| `script` | `"react"` | `default` | `const _react = _interopRequireDefault(require("react"));`<br/>`(0, _react.default)("div");` |
| `module` | `"react"` | `default.jsx` | `import { default as _react } from "react";`<br/>`_react.jsx("div");` |
| `script` | `"react"` | `default.jsx` | `const _react = _interopRequireDefault(require("react"));`<br/>`(0, _react.default).jsx("div");` |
| `module` | `"react"` | `jsx.dev` | `import { _jsx } from "react";`<br/>`_jsx.dev("div");` |
| `script` | `"react"` | `jsx.dev` | `const _react = require("react");`<br/>`(0, _react.jsx).dev("div");` |

This avoids hard-coding the `jsx`/`jsxDEV` names since other libraries use other IDs (e.g. `h`). We can still use `jsx`/`jsxDEV` as the default values.

We would probably also need `pragmaStatic` and `pragmaFrag` options.

packages/babel-plugin-transform-react-jsx/src/transform-experimental.js
```diff
@@ -0,0 +1,815 @@
...
+    const objs = [];
+
+    // In order to avoid having duplicate "children" keys, we avoid
+    // pushing the "children" prop if we have actual children. However,
+    // the children prop may have side effects, so to be certain
+    // these side effects are evaluated, we add them to the following prop
+    // as a sequence expression to preserve order. So:
+    // <div children={x++} foo={y}>{child}</div> becomes
+    // React.jsx('div', {foo: (x++, y), children: child});
+    // duplicateChildren contains the extra children prop values
```
nicolo-ribaudo: We could consider throwing in the transform if there are both syntactic children and the `children` property.

If we don't want to, we don't need this custom logic anyway. This output has the same effect:
```js
React.jsx("div", {
  children: x++,
  foo: y,
  children: child,
});
```

packages/babel-plugin-transform-react-jsx/src/transform-experimental.js
```diff
@@ -0,0 +1,815 @@
...
+
+    const useBuiltIns = file.opts.useBuiltIns || false;
+    if (typeof useBuiltIns !== "boolean") {
+      throw new Error(
+        "transform-react-jsx currently only accepts a boolean option for " +
+          "useBuiltIns (defaults to false)",
+      );
+    }
+
+    if (useSpread && useBuiltIns) {
```
nicolo-ribaudo: We are going to remove these options in Babel 8, and default to `...`. We can avoid adding them to this new transform.

hzoo: just for reference, https://github.com/babel/babel/pull/11141 via [jsx label](https://github.com/babel/babel/pulls?q=is%3Apr+is%3Aopen+label%3A%22area%3A+jsx%22)

lunaruan:
Thanks for the review! I forgot to explain some things that might make this PR and the implementation make a bit more sense. This Babel transform is only meant to be used for React. `jsx`/`jsxs`/`jsxDEV` are not supposed to be used by anything other than the Babel transform. They are meant to be an implementation detail of React and the Babel transform and open to changing in the future, and are setting the stage for other changes. They also have a different function signature than `createElement` does. We never planned for this to replace the current Babel transform for Preact/Inferno/etc. I proposed some alternate suggestions for your concerns below. LMK if you have any other questions!

> We would probably also need `pragmaStatic`/`pramaMulti`, `pragmaDev` and `pragmaFrag` options.
> Or maybe `pragma: { single: "React.jsx", multiple: "React.jsxs", dev: "React.jsxDEV", frag: "React.Fragment" }`.

Distinguishing between `jsx`/`jsxs`/`jsxDEV`/`createElement` is an implementation detail for React and will change in the future. Therefore, we don't think that it makes sense to add a pragma option for all of them. In addition, the signature for `jsx`/`jsxs`/etc. is different than `createElement`, so Preact/Inferno/etc. will not be able to use the new transform the way they currently are.

To satisfy the people currently using the pragma with the `createElement` function signature, what do you think about one of these options? 
1.) We could leave the old Babel transform for Preact/Inferno/etc. and use this only for React.
2.) We could have a `pragma` that, when set, defaults to only using `createElement` and not `jsx`/`jsxs`/`jsxDEV`

> 
> In these examples I'm using the `pragma` option name rather than `autoImport`, because I think that it fits better my proposed changes. Keep in mind that some of the examples in the following table aren't valid React imports, but they might be useful for other libraries.
>

Using `source`/`importSource`/`pragma` to determine how to import React will mostly likely not work for a couple of reasons:
1.) We think the `pragma` field should not be part of `jsx`/`jsxs`/`jsxDEV` (see above).
2.) There are use cases for all the import options above so we can't get rid of them. (Ex. we want everyone to use named exports but typescript only works with namespace exports etc.)

It seems like your concern with the `autoImport` field is that it is too complicated. In that case, we could simplify it by removing the `requires` option like so:
* If `sourceType` is script and `autoImport` is not `default`/`namespace`/`namedExport`, we always go with the `var _react = require("react")`. 
* Otherwise, use `default`/`namespace`/`namedExport` as specified.

What do you this of this alternate proposal?

> I'm not sure I follow. Why doesn't this work?
> 
> ```js
> const props = { key: "foo" };
> React.jsx("div", { ...props}, "Hi");
> ```
> 
> You'll still see that the second parameter contains a `key` property, because the "real" `key` is passed as a different parameter.

What you are suggesting is our end goal; we want to ignore the "props" `key` and just use the "real" `key`. However right now, people can use the spread `key` as an actual `key`. While we are incrementally deprecating this behavior, we need to preserve the current behavior, which means that `<div {...props} key={'hi'} />` (`key` is `'hi'`) is different than `<div key={'hi'} {...props} />` (`key` is `'foo'`). After we deprecate key spread, this two would be the same and we would remove `createElement`.

packages/babel-plugin-transform-react-jsx/src/transform-experimental.js
lunaruan: Sounds good! I'll delete them for now and default to `...` in a separate PR?

packages/babel-plugin-transform-react-jsx/src/transform-experimental.js
lunaruan: We actually did this at some point but it broke on a previous browser because duplicate keys caused an error, so we had to do it this way instead. I'll add that to the comment though to clarify.

sebmarkbage:
Note that part of the reason we want autoImport to be the default is that it prevents sketchy shadowing. E.g. this should not work:

```
import React from "somethingelse";
<Foo />
```
OR
```
function Foo(React) {
  <Foo />
}
```

The `jsx`/`jsxs`/`jsxDEV`/`createElement` implementations are just the defaults. The second step is enabling further compiler optimizations by statically knowing that all JSX is going to resolve to a predictable runtime so we can change the behavior of JSX just by knowing what will render it. I.e. similar to babel helpers.

So specifying the actual implementation details in pragmas goes directly against that goal. We want to make the coupling of JSX to a particular implementation stronger to enable compilers to operate on the assumption that it can compile JSX however it wants as long as it knows which runtime it's targeting. E.g. an optimizing Babel plugin can see that this is using importSource "react" and know how to instead use "react/jsx-optimized-helpers" with a completely different implementation, but if it sees importSource "preact" it would ignore it since it doesn't know what that the runtime is compatible with the optimization. Similarly another plugin can do the inverse.

Allowing people to specifying the fine grained implementation details would undermine that ecosystem goal.

packages/babel-plugin-transform-react-jsx/src/transform-experimental.js
```diff
@@ -0,0 +1,753 @@
...
+  defaultExport: "defaultExport", // import _default from "react";
+  namedExports: "namedExports", // import { jsx } from "react";
+};
+
+const JSX_AUTO_IMPORT_ANNOTATION_REGEX = /\*?\s*@jsxAutoImport\s+([^\s]+)/;
+const JSX_IMPORT_SOURCE_ANNOTATION_REGEX = /\*?\s*@jsxImportSource\s+([^\s]+)/;
+
+// We want to use React.createElement, even in the case of
+// jsx, for <div {...props} key={key} /> to distinguish it
+// from <div key={key} {...props} />. This is an intermediary
```
jridgewell: Can you explain why this is necessary?

When I try to reason about it, it seems like the current implementation is the opposite of what should happen. We can reliably tell that `<div {...props} key={KEY} />` is supposed to be `KEY`. In `<div key={KEY} {...props} />`, we can't tell if the key is `KEY` or was passed in from the spread props. So to maintain the current behavior while you deprecate, you should be using `createElement` only in the second case.

nicolo-ribaudo:
Thanks for explaining that `jsx`/`jsxs`/etc are an implementation detail and are not meant to be something that users should think about. I'm ok with them not being configurable.

I still think that `autoImport` is unnecessary:
- `none` can be obtained by allowing `importSource: false`
- `namespace` and `namedExports` do the same thing, just with different syntax.
- `require` should be detected automatically by the `inject-imports` helper
- `default` would be wrong (according to https://github.com/facebook/react/pull/18102)

> 1.) We could leave the old Babel transform for Preact/Inferno/etc. and use this only for React.
> 2.) We could have a pragma that, when set, defaults to only using createElement and not jsx/jsxs/jsxDEV

I prefer 1. Depending on what the other React-like libraries decide to do (if they want to keep using `createElement` or not) we'll just have to decide if we want the plugin to be named `@babel/transform-react-jsx/compat` or `@babel/transform-jsx` with a mandatory `pragma`/`importSource` option.
Anyway, we can defer this discussion to Babel 8.

packages/babel-plugin-transform-react-jsx/src/transform-experimental.js
sebmarkbage: For `<div key={key} {...props} />` we'll look at `props` to determine if it has a key and if so override the key passed in. So we preserve the behavior by keeping the runtime behavior + warn and then later change the runtime behavior.

For `<div {...props} key={key} />` we don't know if it is `<div {...props} key={undefined} />` and we can't distinguish that from `<div {...props} />` at runtime because they both pass `undefined` to the `key` argument.

packages/babel-plugin-transform-react-jsx/src/transform-experimental.js
sebmarkbage: There's a period in ECMAScript's lifetime when duplicate keys wasn't allowed in strict mode. Unfortunately they are still around. Babel should probably have a transform to remove duplicate keys from any object literal.

We could error but that would make it harder to upgrade. Since we did run into this, we know people write this code today. Arguably it shouldn't error for the same reasons ECMAScript doesn't error on duplicate keys.

packages/babel-plugin-transform-react-jsx/src/transform-experimental.js
nicolo-ribaudo: It makes sense :+1:
The comment should probably be updated to mention that it's needed for the `undefined` case.

<details>
<summary>Half joking</summary>

> because they both pass undefined to the key argument.

`arguments.length`? :stuck_out_tongue: 
</details>

nicolo-ribaudo:
I was thinking about the `jsxAutoImport` and `jsxImportSource` options: do you think that they would be needed by React users? Currently, the `jsxPragma` comment is mostly needed for non-react users.

If you don't think that they would be needed in the real world, we could only allow the option in the config file. If someone asks for them, we can always add them later.

I'm sorry if I'm playing the part of the one who just wants to remove features from the PR, but I'm trying to keep it as simple as possible by avoiding options which might not be necessary :sweat_smile:

lunaruan:
> I still think that autoImport is unnecessary:

Hey! We talked about it a bit and we decided that we don't need the `autoImport/jsxAutoImport` options after all. We decided that the new transform will always auto import. If users don't want to auto import, they will use the `createElement` version of the Babel transform instead. What do you think about this:
1. `version` is `experimental` and `sourceType` is `module`: we use `jsx/jsxs` and named exports
2. `version` is `experimental` and `sourceType` is `script`: we use `jsx/jsxs` and require
3. `version` is `current`: we use `createElement` and don't auto import

`jsxVersion` can be defined in the docblock to specify whether or not we want to auto import.
 
(In Babel 8, we can swap the versions so that `jsx` becomes `current` and `createElement` becomes `legacy`.)

> I was thinking about the jsxAutoImport and jsxImportSource options: do you think that they would be needed by React users? Currently, the jsxPragma comment is mostly needed for non-react users.

We do still need `jsxImportSource` because some users might use React for their app but have a some files that use something other than React elsewhere on the page.

lunaruan:
> The only thing that I'm not sure about is the name of the option (the implementation idea is 💯).

What do you think about `runtime: "classic" | "automatic"`, with the default being "automatic"? We don't want the implementation details to be apparent to the user, so we don't want to explicitly call things `createElement` or `jsx`, and one of the main highlights of this transform is the ability to automatically import `jsx/jsxs` so users don't have to.

We can default to "classic" in 7.9 and default to "automatic" in 8.0.

 The last thing: from a code point of view, (1) and (2) don't need to be handled separately. You can call addNamed from @babel/helper-module-imports, and it automatically decides if it should use import or require.

The reason why we handle this separately is because would prefer to use `var React = require("react")` over `var _jsx = require("react").jsx` for requires because we found that it was faster, but we want to use `import  {jsx as _jsx} from "react"` for imports.

existentialism:
> What do you think about runtime: "classic" | "automatic"

:+1:

nicolo-ribaudo:
Sounds good 👍

marvinhagemeister:
That's super awesome! Simplifying `createElement` is something we talked about a lot in the Preact team and we came to similar conclusions. We have proposals (non-public) that are very similar to `jsx/jxs`. Looping in @jviide as he has played around a lot with those ideas in [htm](https://github.com/developit/htm) which is basically a runtime jsx version based on tagged template literals (works with React too).

### About Pragma

Reading through the code I think we can fully support the proposed features. The main reason we need configurable pragmas today is because of React's usage of a `default` export. We need that to be configurable today is to change `React.createElement` to just `createElement`. That's pretty much solved once React switches to named imports. Right now I don't see any reason that would prevent us from supporting `jsx`, `jsxs` and `jsxDEV`.

In our case we'll likely alias `jsxDEV` to just `jsx` as all warnings during development can be enabled at runtime by importing `preact/debug` regardless of how the application was transpiled/bundled. It works by attaching callbacks to various stages of our renderer at runtime, but that's an implementation detail of how Preact works internally.

### Auto imports

The only config option I'd maybe like to propose is for the import source itself. We'd like to change `import { jsx } from "react"` to `import { jsx } from "preact"`. It's not a hard requirement for us as we could always write a custom plugin to swap the import source ourselves. Nonetheless I think it may be a nice thing to have configurable.

### Refs

The changes proposed here allow us to simplify our `createElement` implementation even further by not having to have extra conditions to pull of "special" props like `key`. Currently `ref` needs the same special treatment as `ref` and I'm happy to see that it is not treated as a separate argument to `jsx/jsxs` here. My hope is that we will keep it as a normal prop, because that would automatically solve the need for a `forwardRef`-like component.

```jsx
// Today
const Foo = forwardRef((props, ref) => <div ref={ref) />);

// vs
const Foo = props => <div ref={props.ref} />;
```

iirc that was part of the original proposal @sebmarkbage made for the future direction of `createElement` but I can't find it right now in my bookmarks :see_no_evil: 

### Conclusion

All in all we're very excited about these changes. They allow us to simplify our code even further and save more precious bytes. We're fully on board with them and we really appreciate reaching out to us. Great work all around! :raised_hands: :100:

_I'll probably have a go at implementing `jsx/jsxs/jsxDEV` for Preact over the weekend :tada:_

nicolo-ribaudo:
@lunaruan You can disable the failing test on windows (like it was done at `packages/babel-preset-react/test/fixtures/preset-options/development/options.json`). Also, please create a windows version of that test.

nicolo-ribaudo:
@marvinhagemeister WDYT about the `createElement` fallback needed by react to warn about `key` in spread props?

If it's not needed by Preact, I can think of two possibilities:
- Automatically disable it if `importSource` is set (maybe make it opt-in in this case)
- Always disable it, and instead of having the warning in React we can put it directly in the compiled output (@lunaruan wdyt?):
```js
_jsx("div", { foo: 1, ..._warnIfKey(props) }, key);

// warnIfKey helper implementaition
function _warnIfKey(props) {
  if ("key" in props) console.warn(...);
}
```

marvinhagemeister:
@nicolo-ribaudo Personally I'm fine with moving users off of spreading keys, but I'm admittedly a bit scared of the amount of support questions we may get.

I think for us and similar frameworks the most preferable solution would be opt-in. Knowing that, we're very likely at the mercy of the React ecosystem either way, because most warnings/mismatches come from older third party components. Some of those may break unexpectedly (for both React and Preact) if they rely on `props.key` directly.

_Looping in @DanielRosenwasser from the TypeScript team because many Preact users compile jsx via TS directly._

packages/babel-helper-builder-react-jsx-experimental/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "lib/index.js",
+  "dependencies": {
+    "@babel/helper-module-imports": "^7.8.3",
+    "@babel/types": "^7.8.3",
+    "esutils": "^2.0.0"
+
```
nicolo-ribaudo: Nit: please remove this empty line.

packages/babel-plugin-transform-react-jsx-development/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@babel/plugin-transform-react-jsx-development",
```
nicolo-ribaudo: Why not an option in the main transform? :thinking:

lunaruan: We decided that because development mode (jsxDEV) is sufficiently different enough from production mode (jsx) in both arguments and maybe potential future optimization, it would be better to make the split more explicit. This might also simplify Preact's use case if they only want to use `jsxDEV`

lunaruan:
Hey! I just pushed some changes to the transform. The major ones are:

1.) Split up `plugin-transform-react-jsx` and `plugin-transform-react-jsx-development`. This is because the DEV mode and the prod mode is sufficiently different enough we wanted to make the split more explicit.
2.) Removed the `autoImport` option. Now the transform defaults to named exports if source type is `module` and require if source type is `script`.
3.) Removed the `version` option and created `runtime: "classic" | "automatic"` `@jsxRuntime` options and a 
4.) Embedded the modules `plugin-transform-react-jsx` and `plugin-transform-react-jsx-development` into the new development transform. Now, when using the new transform, we automatically filter out the added `source` and `self` from these plugins if you have them enabled and re-add them inside the development transform.
5.) Refactored to add the `"classic"` code into the new jsx transform so that people can still use that mode in the future if they need.

lunaruan:
Thanks for commenting @marvinhagemeister! Regarding your comments/concerns:

> The only config option I'd maybe like to propose is for the import source itself. We'd like to change import { jsx } from "react" to import { jsx } from "preact". It's not a hard requirement for us as we could always write a custom plugin to swap the import source ourselves. Nonetheless I think it may be a nice thing to have configurable.

We are proposing an option `importSource` that allows you to specify where you'd like to import from. Does this field work for you?

> The changes proposed here allow us to simplify our createElement implementation even further by not having to have extra conditions to pull of "special" props like key.
>
> Currently ref needs the same special treatment as key and I'm happy to see that it is not treated as a separate argument to jsx/jsxs here. My hope is that we will keep it as a normal prop, because that would automatically solve the need for a forwardRef-like component.

Yes, actually our goal is to eventually deprecate `forwardRef` as per [this RFC](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md)

> If it's not needed by Preact, I can think of three possibilities:
>
> * Make it opt-out
> * Automatically disable it if importSource is set (maybe make it opt-in in this case)
> *Always disable it, and instead of having the warning in React we can put it directly in the compiled output (@lunaruan wdyt?):

Just substituting the `createElement` case with `jsx` will not work because of what @sebmarkbage said before (we cannot distinguish between `<div {...props} key={undefined} />` and `<div {...props} />` at runtime, even though they are not equivalent). An alternate to disabling `createElement` is just to not support the `createElement` case altogether, ie. requiring `key` to be before `props` spread. This is the direction we are eventually heading for anyway. Our eventual goal is to make key its own concept separate from props. Therefore, we don't want key to be lumped in with props (ex. after a props spread).

nicolo-ribaudo:
> This is the direction we are eventually heading for anyway.

Do you mean in this PR, or in the future?

lunaruan:
> Do you mean in this PR, or in the future?

In the future (Babel 9 maybe)

packages/babel-plugin-transform-react-jsx/src/transform-experimental.js
nicolo-ribaudo: Yeah, Babel already has that transform. If people need to support a browser which throws for duplicate keys, it's handled by preset-env.

Since we are going to enable object spread by default in the jsx transform, most people will have preset-env anyway.

packages/babel-plugin-transform-react-jsx/src/transform-experimental.js
nicolo-ribaudo: Would it be possible to default to ... already in this PR?

Andarist:
> (In Babel 8, we can swap the versions so that jsx becomes current and createElement becomes legacy.)

Could this be somehow automated? For a considerable amount of time, many people will still be at older React versions and will have to manually pass options to this transform. Could Babel query FS and check the currently installed React version? I know this is not something that is usually done here (it increases coupling and stuff), but maybe could be considered here to improve DX?

> The reason why we handle this separately is because would prefer to use var React = require("react") over var _jsx = require("react").jsx for requires because we found that it was faster

Interesting, do you maybe know why? Shouldn't, theoretically, a static binding perform better than a dynamic~ lookup?

> The main differences between React.jsx/React.jsxDEV

Is the main difference that `jsxDEV` can accept those extra self/source "props"? Is the guidance for library authors to always use the "prod mode" and the dev one should be solely used in apps based on the `NODE_ENV` of the build process?

> Currently, the jsxPragma comment is mostly needed for non-react users.

Mostly - yes, but there are **popular** React libraries that use jsx pragma to provide extra functionality (mainly Emotion and other css-in-js libraries). We use in Emotion to resolve css prop at runtime (convert it to injected className), but other than that we forward props to `React.createElement`. 

One of the main reasons why we do this using a pragma and not the Babel plugin is that pragma works "everywhere" - so it "just works" when using TS or when used in Babel-constrained tools like CRA.

What would be the guidance on this one? The use case looks a little bit unusual at first but it has actually allowed us to solve problems.

nicolo-ribaudo:
> Could this be somehow automated? For a considerable amount of time, many people will still be at older React versions and will have to manually pass options to this transform. Could Babel query FS and check the currently installed React version? I know this is not something that is usually done here (it increases coupling and stuff), but maybe could be considered here to improve DX?

It's an option that you have to set one time while upgrading to Babel 8.0.0, going through the migration guide. People creating a new app with Babel 8.0.0, won't have to worry about it because in a new app they'll probably use the latest React version.

Also, we tried automating it for `@babel/runtime` but couldn't find a reliable way to do so.

> Interesting, do you maybe know why? Shouldn't, theoretically, a static binding perform better than a dynamic~ lookup?

I was also surprised by this, but decided to just trust them :joy:

@lunaruan I just realized that you can obtain both `var React = require("react"); React.jsx` and `var _jsx = require("react").jsx; _jsx` using `@babel/helper-module-imports`. So there is no reason to special-case the CommonJS implementation.

https://runkit.com/nicolo-ribaudo/babel-helper-module-imports-for-react-jsx

> Mostly - yes, but there are popular React libraries that use jsx pragma to provide extra functionality (mainly Emotion and other css-in-js libraries)

Yes, sorry :sweat_smile:
What I meant is that it's mostly needed for people not directly using the `React.createElement` function and most of those libraries would still need the pragma rather than the autoimport.

Andarist:
> What I meant is that it's mostly needed for people not directly using the React.createElement function and most of those libraries would still need the pragma rather than the autoimport.

Well, we probably would love to migrate to new JSX functions - rather than staying on `createElement` but we can't do that if you don't allow us to customize this with pragma.

OTOH right now this is just a matter of a single pragma (`/** @jsx jsx */`) and the new ones would require 3 (if we count them by import specifiers), so actually what would solve our problem is a pragma for "import source" - `@jsxImportSource`, `@jsxSource`? It's not as good as `@jsx jsx` as it's much longer, but I guess we would have to propose it only for tools like CRA where Babel customization is not an option and just ask people to customize the Babel transform with `"importSource"` where it can be done. WDYT?

nicolo-ribaudo:
Yes, the pragma should only be the "backup plan" when the config can't be edited.

Does this sound good to you?
1. If `/* @jsx .. */` is set, fallback to the old `creteElement`-like style.
    ```js
    /** @jsx jsx */

    <div css={{ ... }} />
    ```
2. If the `/* @jsxImportSource .. */`/`/* @jsxSource .. */`(*), `emotion` would expose an entry point the `jsx` and `jsxs` function with the new signature.
    ```js
    /** @jsxSource @emotion/core/jsx */

    <div css={{ ... }} />
    ```

(*) I slightly prefer the shorter one, but don't have strong opinions about it.

packages/babel-helper-builder-react-jsx-experimental/src/index.js
```diff
@@ -0,0 +1,844 @@
...
+        if (shouldUseCreateElement(path)) {
+          imports.createElement = true;
+        } else if (path.node.children.length > 1) {
+          const importName = options.development ? "jsxDEV" : "jsxs";
+          imports[importName] = true;
+        } else {
+          const importName = options.development ? "jsxDEV" : "jsx";
+          imports[importName] = true;
+        }
+      },
```
nicolo-ribaudo: If we have already found all the possible import names, can we call `path.stop()` to speed up the traversal?

packages/babel-helper-builder-react-jsx-experimental/src/index.js
```diff
@@ -0,0 +1,844 @@
...
+      });
+
+      return importMap;
+    } else {
+      // var _react = require("react")
+      return addNamespace(path, state.source, {
+        importedInterop: "uncompiled",
+      }).name;
+    }
+  }
```
nicolo-ribaudo: Currently, this function returns sometimes an object, sometimes a string. It would be better to unify them.

You can simplify this function like this:
```js
function addAutoImports(path, state) {
  const importMap = {};
  Object.keys(imoprts).forEach(importName => {
    importMap[importName] = addNamed(path, imoprtName, state.source, {
      importedInterop: "uncompiled",
      ensureLiveReference: true,
    });
  });
}
```

And the instead of `createIdentifierName(path, name, importName)` use `t.cloneNode(importName[name])`.

packages/babel-helper-builder-react-jsx-experimental/src/index.js
```diff
@@ -0,0 +1,844 @@
...
+    if (isModule(path) && importName) {
+      const identifierName = `${importName[name]}`;
+      return identifierName;
+    } else {
+      return `${importName}.${name}`;
+    }
+  }
+
+  function getImportNames(parentPath) {
+    const imports = {};
```
nicolo-ribaudo: Since this is only used to set properties as `true`, and then in `addAutoImports` we don't even check what the value of the properties is but only their presence, it seems the perfect use case for a `new Set()`.

Andarist:
Yes, I think that would address our issue quite nicely and I also prefer the shorter version.

The only problem is that our current implementation based on `createElement` is called `jsx`, but I don't expect you to deal with that here, that's on us 😅

packages/babel-plugin-transform-react-jsx/src/index.js
```diff
@@ -1,100 +1,15 @@
...
+/* eslint-disable-next-line @babel/development/plugin-name */
+import transformExperimental from "./transform-experimental";
 import { declare } from "@babel/helper-plugin-utils";
-import jsx from "@babel/plugin-syntax-jsx";
-import helper from "@babel/helper-builder-react-jsx";
-import { types as t } from "@babel/core";
 
 export default declare((api, options) => {
-  api.assertVersion(7);
+  const runtime = options.runtime || "classic";
```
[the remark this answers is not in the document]
lunaruan: We do throw an error if `runtime` isn't `"classic"` or `"automatic"`. It's in the helper Babel plugin instead of the main plugin though because we'll get rid of this file in v8.0.0. (I inlined all the "classic" code in the new helper Babel plugin)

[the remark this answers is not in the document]
existentialism: Let's add a comment here to remind ourselves :)

packages/babel-helper-builder-react-jsx-experimental/src/index.js
lunaruan: This won't exactly work because it will add a separate `require` for every import.

packages/babel-plugin-transform-react-jsx/src/transform-experimental.js
lunaruan: What is the transform accounts for duplicate keys? I guess since this is for Babel 8 it's probably okay to introduce this as a breaking change, since we're also doing things like defaulting to spread anyway. @sebmarkbage

nicolo-ribaudo: It's https://babeljs.io/docs/en/babel-plugin-transform-duplicate-keys.

You can see it in action here: https://babeljs.io/repl/#?browsers=ie%2011&build=&builtIns=false&spec=false&loose=false&code_lz=G4QwTgBCELwQ3gKAhAHgLggRgDTLZgEyIC-QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Cenv&prettier=false&targets=&version=7.8.6&externalPlugins=%40babel%2Fplugin-transform-async-to-generator%407.8.3

It produces a sub-optimal output, but:
- The only popular browser that needs it is IE ([these browsers support them](https://github.com/babel/babel/blob/8f281f578db68ef2092ce35b8db003bdb346a9e4/packages/babel-compat-data/data/plugins.json#L199-L208))
- If we want to produce a more optimized output by merging the first duplicate key into the following one, it should be done in `@babel/plugin-transform-duplicate-keys` and not here
- Duplicate keys are super rare anyway

packages/babel-helper-builder-react-jsx-experimental/src/index.js
nicolo-ribaudo: Oh ok.
I'd like to make `addNamed` work better with multiple named imports in this case, but I/you/anyone can do it in a separate PR.

lunaruan: Okay sounds good. I'll just leave this for now then and update after `addNamed` is fixed.

nicolo-ribaudo: We'll probably merge this PR without this change, to avoid blocking it.

packages/babel-plugin-transform-react-jsx-development/test/fixtures/linux/auto-import-dev/output.mjs
```diff
@@ -0,0 +1,31 @@
+import { createElement as _createElement } from "react";
+import { jsxDEV as _jsxDEV } from "react";
+import { Fragment as _Fragment } from "react";
```
nicolo-ribaudo: This has the same problem as the one you mentioned for `require()`. Again, we can fix it separately.

lunaruan: Sounds good. I have a workaround in my diff for now.

packages/babel-plugin-transform-react-jsx-development/test/fixtures/linux/auto-import-dev/exec.js
```diff
@@ -0,0 +1,51 @@
+var actual = transform(
```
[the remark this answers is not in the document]
lunaruan: Is there a way to run the tests when the os is windows?

marvinhagemeister:
> We are proposing an option importSource that allows you to specify where you'd like to import from. Does this field work for you?

@lunaruan  Awesome, sounds good to me :+1: 

Played around with an implementation for `jsx/jsxs/jsxDEV` in the past few days. It's working really great and allows us to remove a few branches which saves us some bytes :tada: 

I'm wondering if the split between `jsx` and `jsxs` can be made redundant by leveraging the same `isStaticChildren` argument from `jsxDEV` and than do the branching inside `jsx`. Function declarations cost us more bytes because it's longer and we'd have to add an additional export to our module. In my understanding code bases will make use of both `jsx` and `jsxs`.

```js
jsx(type, props, key)
jsxs(type, props, key)
jsxDEV(type, props, key, isStaticChildren, source, self)

// vs only 2 functions
jsx(type, props, key, isStaticChildren)
jsxDEV(type, props, key, isStaticChildren, source, self)
```

Not sure what the React implementation of `jsx/jsxs` looks like, so I don't have the full picture.

gaearon:
>Mostly - yes, but there are popular React libraries that use jsx pragma to provide extra functionality (mainly Emotion and other css-in-js libraries). We use in Emotion to resolve css prop at runtime (convert it to injected className), but other than that we forward props to React.createElement.

Why does emotion use a pragma? Is there a reason it doesn't only change props? E.g. transform `<div css={foo} />` to `<div className={something(foo)} />` and let the JSX transform do the rest?

lunaruan:
> I'm wondering if the split between `jsx` and `jsxs` can be made redundant by leveraging the same `isStaticChildren` argument from `jsxDEV` and than do the branching inside `jsx`. Function declarations cost us more bytes because it's longer and we'd have to add an additional export to our module. In my understanding code bases will make use of both `jsx` and `jsxs`.

We thought about doing this, but adding an extra argument incurs extra runtime cost during each call. We also might add extra arguments in the future, so we want to avoid passing `isStaticChildren` in as an argument because it's more of an implementation detail.

Andarist:
> Why does emotion use a pragma? 

It resolves at runtime, rather than during build time - so it's more beginner-friendly as this works:
```
const props = { css: { color: 'hotpink' } }
;<div {...props} />
``` 
While we probably could consider not supporting this particular scenario (I believe most css prop usage happens "inline") there are also other benefits:
- we also grab the context value from inside our `jsx` function to resolve styles
- it works in CRA (cant customize Babel configs in CRA after all)
- it also just works in TS (TS doesnt allow for transformer plugins inside tsconfig.json).

There might be also some other benefits I forgot right now, but those are the main ones.

lunaruan:
@nicolo-ribaudo We did some preliminary tests of the `runtime: classic` version of the transform inside our codebase and it looks good so far. We will test the `runtime: automatic` version soon. LMK if you have any other comments/concerns in the meantime!

packages/babel-plugin-transform-react-jsx/src/index.js
```diff
@@ -1,100 +1,18 @@
+/* eslint-disable-next-line @babel/development/plugin-name */
```
nicolo-ribaudo: Is this needed? I thought that we would need it at line 6, or disable it for the whole file :thinking:

packages/babel-plugin-transform-react-jsx/src/index.js
```diff
@@ -1,100 +1,18 @@
...
+import transformStable from "./transform-stable";
+/* eslint-disable-next-line @babel/development/plugin-name */
+import transformExperimental from "./transform-experimental";
```
nicolo-ribaudo: Please rename these two files to `transform-classic` and `transform-automatic`.

nicolo-ribaudo:
Can you also add the `runtime` option to `@babel/preset-react`? It also already has a `development` that we can use to enable the new `...-development` plugin.

lunaruan:
@marvinhagemeister We want to move `jsx/jsxs` and `jsxDEV` into separate modules because we want to clarify that users aren't supposed to call them directly and only use them in the Babel transform. We also want to split out jsxDEV in particular because we don't want it to be imported in production and add to bundle size. We are talking about moving `jsx/jsxs into a `react/jsx-runtime` file and `jsxDEV` to a `react/jsx-dev-runtime` file. This would mean if you specify `react` as the import source, this would:
* import `jsxDEV` from `react/jsx-dev-runtime`
* import `jsx/jsxs` from `react/jsx-runtime`
* import `createElement` from `react` (we are planning on deprecating this case eventually anyway so there's no need to move it to any of thte runtimes)
* We will likely also re-export `Fragment` from the new modules as well 

Would this work for you?

packages/babel-plugin-transform-react-jsx-development/test/fixtures/linux/classic-runtime-with-source-self-plugins/output.js
```diff
@@ -0,0 +1,38 @@
+var _jsxFileName = "<CWD>/packages/babel-plugin-transform-react-jsx-development/test/fixtures/linux/classic-runtime-with-source-self-plugins/input.js",
+    _jsxFileName2 = "<CWD>/packages/babel-plugin-transform-react-jsx-development/test/fixtures/linux/classic-runtime-with-source-self-plugins/input.js";
```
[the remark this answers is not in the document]
lunaruan: The first one is injected because of `transform-react-jsx-source` plugin, and it's not used because we filter out the injected source and self and insert our own. We could throw if someone uses the new transform and source or self, but that seems a bit aggressive. What doo you think?

packages/babel-plugin-transform-react-jsx-development/test/fixtures/linux/handle-fragments-with-key/output.js
```diff
@@ -0,0 +1,8 @@
+var _react = require("react");
+
+var _jsxFileName = "<CWD>/packages/babel-plugin-transform-react-jsx-development/test/fixtures/linux/handle-fragments-with-key/input.js";
+
+var x = _react.jsxDEV(React.Fragment, {}, "foo", false, {
```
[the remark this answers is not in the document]
lunaruan: In this case, because users call `React.Fragment` directly, they would have to import React directly (similar to if they call createElement directly). Let me rewrite the test to reflect this.

packages/babel-plugin-transform-react-jsx/src/index.js
lunaruan: The lint rule complains because `transform-classic` and `transform-automatic` are not Babel plugins. We could disable it for the whole file but it seems more specific to only disable it on the lines where we need it.

packages/babel-helper-builder-react-jsx-experimental/src/index.js
```diff
@@ -0,0 +1,739 @@
...
+  return {
+    JSXNamespacedName(path, state) {
+      const throwIfNamespace =
+        state.opts.throwIfNamespace === undefined
+          ? true
+          : !!state.opts.throwIfNamespace;
+      if (throwIfNamespace) {
+        throw path.buildCodeFrameError(
+          `Namespace tags are not supported by default. React's JSX doesn't support namespace tags. \
+You can turn on the 'throwIfNamespace' flag to bypass this warning.`,
```
JLHwung: It seems clearer to me that we just say what our users should do
```suggestion
You can set `throwIfNamespace: false` to bypass this warning.`,
```

packages/babel-helper-builder-react-jsx-experimental/src/index.js
```diff
@@ -0,0 +1,739 @@
...
+import * as t from "@babel/types";
+import { addNamed, addNamespace, isModule } from "@babel/helper-module-imports";
+
+export function helper(babel, options) {
+  const FILE_NAME_VAR = "_jsxFileName";
+
+  const JSX_SOURCE_ANNOTATION_REGEX = /\*?\s*@jsxImportSource\s+([^\s]+)/;
+  const JSX_RUNTIME_ANNOTATION_REGEX = /\*?\s*@jsxRuntime\s+([^\s]+)/;
+
+  const JSX_ANNOTATION_REGEX = /\*?\s*@jsx\s+([^\s]+)/;
```
JLHwung: Although `JSX_ANNOTATION_REGEX` exists in current source, but it seems to me that the `JSX_ANNOTATION_REGEX` implicitly requires an extra whitespace before the closing `*/`:
```js
JSX_ANNOTATION_REGEX.exec("/**@jsx dom*/")[1]
// returns `dom*/`
```

packages/babel-helper-builder-react-jsx-experimental/src/index.js
lunaruan: Sounds good. This warning was taken from the original Babel transform so I'll update that one as well!

packages/babel-helper-builder-react-jsx-experimental/src/index.js
lunaruan: It seems like `file.ast.comments`, which is what we're using, already strips out the the open and closing comment tags. So actually the comment `/**@jsx dom*/` would be parsed as `JSX_ANNOTATION_REGEX.exec("*@jsx dom")`, which is okay.

marvinhagemeister:
> > I'm wondering if the split between `jsx` and `jsxs` can be made redundant by leveraging the same `isStaticChildren` argument from `jsxDEV` and than do the branching inside `jsx`. Function declarations cost us more bytes because it's longer and we'd have to add an additional export to our module. In my understanding code bases will make use of both `jsx` and `jsxs`.
> 
> We thought about doing this, but adding an extra argument incurs extra runtime cost during each call. We also might add extra arguments in the future, so we want to avoid passing `isStaticChildren` in as an argument because it's more of an implementation detail.

I see. Admittedly that doesn't back up our experience of optimizing Preact over the years as we have several functions with many arguments. Maybe @mathiasbynens or @bmeurer can give a definitive answer here?

> @marvinhagemeister We want to move jsx/jsxs and jsxDEV into separate modules because we want to clarify that users aren't supposed to call them directly and only use them in the Babel transform.

That's really great as it was one of the ideas that came up in our team as feedback. My current implementation is right in core, but many voiced their preference of moving that to a module :+1:

> We also want to split out jsxDEV in particular because we don't want it to be imported in production and add to bundle size. We are talking about moving jsx/jsxs into a react/jsx-runtime file and jsxDEV to a react/jsx-dev-runtime file. This would mean if you specify react as the import source, this would:
>
> - import jsxDEV from react/jsx-dev-runtime
> - import jsx/jsxs from react/jsx-runtime
> - import createElement from react (we are planning on deprecating this case eventually anyway so there's no need to move it to any of the runtimes)

The less files and exports we have to stub, the better for us. In our case all three will be very likely the same function, so having to juggle not just exports, but also different import paths is an overhead for us.

I'm very curious to see if there are any implementations of those for React to see somewhere. Off the top of my head I have a hard time thinking of any dev specific behaviour other than passing `source` and `self` around. Is there somewhere a pending PR against React?

Andarist:
> I'm very curious to see if there are any implementations of those for React to see somewhere.

https://github.com/facebook/react/blob/d2158d6ccbcbd998d1cd1c06ac6c6682cbeafd04/packages/react/src/React.js#L68-L72

nicolo-ribaudo:
@lunaruan I agree with @marvinhagemeister: I think that it would be better to have a single entry point. Wouldn't `jsxDEV` be easily three-shaken/DCE'ed in production anyway?

lunaruan:
@marvinhagemeister @nicolo-ribaudo We have a nontrivial amount of DEV specific behavior in React (ex. warnings, source, and self). We are also on planning on more optimizations in the future and we also want to have `react/jsx-runtime` be a stable version of `jsx/jsxs` whereas the parameters for `jsxDEV` might change in the future. What is your use case for `jsxDEV`? Could you just use `react/jsx-runtime` (ie `jsx/jsxs`) instead?

nicolo-ribaudo:
Oh right, even in development mode Preact could set `importSource` to the same value they use in production mode.

packages/babel-preset-react/test/fixtures/preset-options/development-runtime-automatic-windows/output.js
```diff
@@ -0,0 +1,10 @@
+var _react = require("react");
+
+var _jsxFileName = "C:\\Users\\travis\\build\\babel\\babel\\packages\\babel-preset-react\\test\\fixtures\\preset-options\\development-runtime-automatic-windows\\input.js";
```
nicolo-ribaudo: TODO: (Unrelated) We should fix this to use `<CWD>`, like on linux.

packages/babel-standalone/package.json
```diff
@@ -87,6 +87,7 @@
     "@babel/plugin-transform-react-inline-elements": "^7.8.3",
     "@babel/plugin-transform-react-jsx": "^7.8.3",
     "@babel/plugin-transform-react-jsx-compat": "^7.8.3",
+    "@babel/plugin-transform-react-jsx-development": "^7.8.3",
```
nicolo-ribaudo: Can you also add it to `packages/babel-standalone/scripts/pluginConfig.json`?

marvinhagemeister:
@lunaruan Good point! That option didn't cross my mind to just alias the importSource for the dev version to the same file as the prod one 👍

nicolo-ribaudo:
Oh, the failing test is caused by this PR, but you just need to update the fixture output.

packages/babel-helper-builder-react-jsx-experimental/src/index.js
```diff
@@ -0,0 +1,779 @@
...
+      // isStaticChildren, __source, and __self are only used in development
+      // automatically include __source and __self in this plugin
+      // so we can eliminate the need for separate Babel plugins in Babel 8
+      source = makeSource(path, file);
+      self = t.thisExpression();
+      args.push(
+        key === undefined ? path.scope.buildUndefinedNode() : key,
+        t.booleanLiteral(path.node.children.length > 1),
+        source ?? path.scope.buildUndefinedNode(),
+        self ?? path.scope.buildUndefinedNode(),
```
JLHwung: When will `self` be nullish?

nicolo-ribaudo:
This will be released on Friday!

marvinhagemeister:
@nicolo-ribaudo Awesome, thanks or the heads up!

Andarist:
@lunaruan i see that on master new jsx factories are still copying the props, is this going to be changed before the final release?

gaearon:
No, as it would be a breaking change. We are leaving this for a future major. (See https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#spread-only)

Andarist:
As far as I can see new jsx factories and old `createElement` do not share those implementation parts so it should not be a breaking change for new factories to adhere to the planned semantics right away. Or am I missing something here?

gaearon:
Yeah the goal is to get everyone on the new transform soon so we can *then* introduce depreciations for now-detectable patterns and *after* that we can break them in a React major.

PR #11158, opened 2020-02-20
Fix scope of function body when var redeclares param

### LINKED ISSUE #10471: Default function parameters is a inline function
```js
var x = 1
function foo(x, y = function () { x = 2 }) {
	var x = 3
	y()
	console.log(x);
}
foo() 
console.log(x);  // 3 1 
```
after bable 
```js
"use strict";

var x = 1;

function foo(x) {
  var y = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : function () {
    x = 2;
  };
  var x = 3;
  y();
  console.log(x);
}

foo();
console.log(x);
// 2 1 
```

### LINKED ISSUE #10798: the `var` declaration is missing with the default function parameters scope
## Bug Report

**Current Behavior**

Playground:
https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABFApgZygClJRBeRTASnwD5EBvAKEUQDcBDAJ0Qf0QEYBuGxCBNHAA2KAHRC4Ac0wMiPAL4lqtRiwAOTFHXgg07AAwKuQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015&prettier=false&targets=&version=7.7.4&externalPlugins=

**Input Code**
```js
function test(func = () => {
  var a = 1;
  console.log(a);
}) {
  var previous = 0;
};
```
Translate these with es2015 presets, the result is:
```js
"use strict";

function test() {
  var func = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : function () {
    var a = 1;
    console.log(a);
  };
  return function () {
    previous = 0;
  }();
}

;
```
We can see that the generated function returns a closure, and the various `previous` is not declared in the closure.

in strict mode, we will get the exception:
`Uncaught ReferenceError: previous is not defined`

**Expected behavior/code**
The `var` expression won't be removed.

Babel playground with es2015 preset config.

image

openorclose:
| Fixed Issues?            |  Fixes #10471 Fixes #10798
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 

The below is valid js, but babel currently throws a redeclaration error:
```js
function f(a = 1) {
  var a = 1;
}
```
Link: https://babeljs.io/repl#?browsers=%3E0.5%25%2C%20not%20ie%20%3C%3D11%2C%20not%20safari%205.1%2C%20not%20samsung%204&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABMAFAQ0QXkQRgJSIDeAUIogG5oBOiG2OA3MQL5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Ces2015%2Cenv&prettier=false&targets=&version=7.8.4&externalPlugins=%40babel%2Fplugin-transform-react-jsx%407.8.3

babel-traverse registers a `var` or function declaration that redeclare a function parameter as a constant violation.

So now during function parameter transformation, if we encounter a parameter whose bindings contain constant violations, we check if they are `var` or `function` declarations. If we find empty `var` declarations (e.g. `var x;`), we delete them since such declarations do not override the parameter. If not, so we have var declarations with initalisers or function declarations, and we need to wrap the function body in an iife to allow these differently scoped declarations.

However `callDelegate` always hoists up these `var` declarations into the real function scope instead of letting them remain in the IIFE function scope, so let's add another parameter to `callDelegate` that allows us to prevent this hoisting.

We only need to check for `var` and function declarations redefinition, because redeclaration of parameters via `const` or `let` is a parse error.


packages/babel-traverse/test/fixtures/rename/parameter-redeclaration/output.js
```diff
@@ -1,9 +1,9 @@
-function f(z, b) {
+function f(a, b) {
```
openorclose: Test case change here.

The test input.js was:

```js
function f(a, b) { // a1
  var a = "redeclared"; // a2
  return b;
}

function g(a) {
  function a() {
    return "function, redeclared";
  }
  return a();
}
```

and a plugin was used on it to rename `a` to `z`.

However, at lines a1 and a2, they refer to different `a`'s, so they definitely shouldnt be both renamed to `z`.

openorclose:
On a separate note, there isn't `function parameter scope` property on a Function, so parameters of the function and local variables of the function all get lumped together in the same scope, so we have to choose between the current behaviour, where params take precedence over function body `var`s, or this pr, where we let the `var`s "override" the function params.

nicolo-ribaudo:
Can you also add this test, which should return 2?
```js
function f(a = 2) {
  var a;
}
```

openorclose:
> Can you also add this test, which I think should return 2?
> 
> ```js
> function f(a = 2) {
>   var a;
>   return a;
> }
> ```

Oh right, this pr will return the wrong value for the above.

openorclose:
> Can you also add this test, which I think should return 2?
> 
> ```js
> function f(a = 2) {
>   var a;
>   return a;
> }
> ```

Added and fixed, thanks for the test case!

Updated the implementation and the pr description, whenever we see an empty var declaration that clashes with a parameter it gets deleted.

openorclose:
Can I get a review on this (I can't seem to request reviewers)?

packages/babel-plugin-transform-parameters/src/params.js
```diff
@@ -64,6 +64,31 @@ export default function convertFunctionParams(path, loose) {
   for (let i = 0; i < params.length; i++) {
     const param = params[i];
 
+    for (const name of Object.keys(param.getBindingIdentifiers())) {
+      const constantViolations = scope.bindings[name]?.constantViolations;
```
JLHwung: nit: `scope.bindings[name]` is always truthy because param bindings should always be registered in the function scope. And we don't have to empty check `.constantViolations` because it will always be initialized as an empty array. So it suffices to
```js
for (const redeclarator of scope.bindings[name].constantViolations) {}
```

packages/babel-plugin-transform-parameters/test/fixtures/parameters/var-same-as-param-closure/input.js
```diff
@@ -0,0 +1,8 @@
+var x = 1
```
JLHwung: nit: you can change this test to `exec.js` and use `expect` assertions:

```js
function foo(x, y = function () { x = 2 }) {
  var x = 3
  y()
  expect(x).toBe(3);
}
foo()
expect(x).toBe(1);
```

Babel fixture test runner will apply transform to `exec.js` and evaluate the assertions above.

packages/babel-helper-call-delegate/src/index.js
```diff
@@ -32,7 +36,9 @@ export default function(path: NodePath, scope = path.scope) {
   let args = [];
 
   // todo: only hoist if necessary
```
JLHwung: nit: I think the todo item here is done.

packages/babel-plugin-transform-parameters/src/params.js
openorclose:
```
Summary of all failing tests
 FAIL  packages/babel-plugin-transform-typescript/test/index.js
  ● babel-plugin-transform-typescript/class › parameter properties with parameters

    TypeError: babel/packages/babel-plugin-transform-typescript/test/fixtures/class/parameter-properties-with-parameters/input.ts: Cannot read property 'constantViolations' of undefined
```

Thanks! I tried this out and the above test failed, seems like `scope.bindings[name]` is undefined for the particular case.

packages/babel-plugin-transform-parameters/test/fixtures/parameters/var-same-as-param-closure/input.js
openorclose: Thanks!

packages/babel-helper-call-delegate/src/index.js
openorclose: Ok, will remove comment

openorclose:
The following test failed when merging latest master.

Can this be updated?

```
Summary of all failing tests
 FAIL  packages/babel-preset-env/test/fixtures.js (5.254s)
  ● babel-preset-env/debug › browserslists last 2 versions not ie

    Expected /home/circleci/babel/packages/babel-preset-env/test/fixtures/debug/browserslists-last-2-versions-not-ie/stdout.txt to match transform output.
    To autogenerate a passing version of this file, delete the file and re-run the tests.

    Diff:

    - Expected
    + Received

    @@ -3,29 +3,29 @@
      Using targets:
      {
        "android": "79",
        "chrome": "79",
        "edge": "79",
    -   "firefox": "71",
    +   "firefox": "72",
        "ios": "13",
        "opera": "65",
        "safari": "12.1",
        "samsung": "9.2"
      }
      
      Using modules transform: auto
      
      Using plugins:
    -   proposal-nullish-coalescing-operator { "android":"79", "chrome":"79", "edge":"79", "firefox":"71", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    -   proposal-optional-chaining { "android":"79", "chrome":"79", "edge":"79", "firefox":"71", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    -   syntax-json-strings { "android":"79", "chrome":"79", "edge":"79", "firefox":"71", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    -   syntax-optional-catch-binding { "android":"79", "chrome":"79", "edge":"79", "firefox":"71", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    -   syntax-async-generators { "android":"79", "chrome":"79", "edge":"79", "firefox":"71", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    -   syntax-object-rest-spread { "android":"79", "chrome":"79", "edge":"79", "firefox":"71", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    -   transform-dotall-regex { "firefox":"71" }
    -   proposal-unicode-property-regex { "firefox":"71" }
    -   transform-named-capturing-groups-regex { "firefox":"71" }
    +   proposal-nullish-coalescing-operator { "android":"79", "chrome":"79", "edge":"79", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    +   proposal-optional-chaining { "android":"79", "chrome":"79", "edge":"79", "firefox":"72", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    +   syntax-json-strings { "android":"79", "chrome":"79", "edge":"79", "firefox":"72", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    +   syntax-optional-catch-binding { "android":"79", "chrome":"79", "edge":"79", "firefox":"72", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    +   syntax-async-generators { "android":"79", "chrome":"79", "edge":"79", "firefox":"72", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    +   syntax-object-rest-spread { "android":"79", "chrome":"79", "edge":"79", "firefox":"72", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    +   transform-dotall-regex { "firefox":"72" }
    +   proposal-unicode-property-regex { "firefox":"72" }
    +   transform-named-capturing-groups-regex { "firefox":"72" }
        transform-template-literals { "safari":"12.1" }
    -   transform-modules-commonjs { "android":"79", "chrome":"79", "edge":"79", "firefox":"71", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    -   proposal-dynamic-import { "android":"79", "chrome":"79", "edge":"79", "firefox":"71", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    +   transform-modules-commonjs { "android":"79", "chrome":"79", "edge":"79", "firefox":"72", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
    +   proposal-dynamic-import { "android":"79", "chrome":"79", "edge":"79", "firefox":"72", "ios":"13", "opera":"65", "safari":"12.1", "samsung":"9.2" }
```

JLHwung:
@openorclose Can you rebase on the upstream master? It should be fixed now.

openorclose:
> @openorclose Can you rebase on the upstream master? It should be fixed now.

Rebased!

packages/babel-plugin-transform-parameters/src/params.js
openorclose: Don't think I should fix the Typescript part here, so ill leave the checks in for now?

existentialism:
Nice work @openorclose!

openorclose:
Thanks for the reviews!

PR #11159, opened 2020-02-20
feat: enhance babel-core assertVersion error message

PatNeedham:

I experienced this the other day, where I got this error message to appear:
image

Even though my `@babel/core`, `@babel/plugin-*`, and `@babel/preset-*` dev dependencies were all v7.x.x. 

That specific error only went away after installing `@babel/cli`  (got substituted with a more easily addressable error):
image

### DIFF
--- a/packages/babel-core/src/config/helpers/config-api.js
+++ b/packages/babel-core/src/config/helpers/config-api.js
@@ -86,7 +86,9 @@ function assertVersion(range: string | number): void {
       `it is likely that something in your build process is loading the ` +
       `wrong version. Inspect the stack trace of this error to look for ` +
       `the first entry that doesn't mention "@babel/core" or "babel-core" ` +
-      `to see what is calling Babel.`,
+      `to see what is calling Babel. It is also possible a globally ` +
+      `installed version of babel-cli is being used if neither that package ` +
+      `nor @babel/cli appear inside the package.json`,
   );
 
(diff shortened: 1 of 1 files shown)


packages/babel-core/src/config/helpers/config-api.js
```diff
@@ -86,7 +86,9 @@ function assertVersion(range: string | number): void {
       `it is likely that something in your build process is loading the ` +
       `wrong version. Inspect the stack trace of this error to look for ` +
       `the first entry that doesn't mention "@babel/core" or "babel-core" ` +
-      `to see what is calling Babel.`,
+      `to see what is calling Babel. It is also possible a globally ` +
+      `installed version of babel-cli is being used if neither that package ` +
+      `nor @babel/cli appear inside the package.json`,
```
nicolo-ribaudo, line 91: This is never true in this package. Since we will release this as Babel >7.9.x, it's impossible that the globally installed `babel-cli` is being used because it only loads `babel-core` 6.

You should update the message in `@babel/helper-plugin-utils` instead.

nicolo-ribaudo:
Closing as this PR doesn't work as-is (see https://github.com/babel/babel/pull/11159#pullrequestreview-388568057). If you want to open a new PR for `@babel/helper-plugin-utils`, it would be welcome :)

Issue #1036, opened 2015-03-16
AMD equivalent of commonStrict


matthewp:
Shouldn't there be one?  Since Babel compiles to the commonjs amd wrapper style it's inconsistent that es modules set module.exports rather than exports.default. Could there be an `amdStrict` perhaps?

stefanpenner:
we have been disabling babel's modules, and using esperantojs for this functionality.

sebmck:
There is?

```javascript
export default foo();
```

```sh
$ babel --modules amdStrict
```

```javascript
define(["exports"], function (exports) {
  "use strict";

  export["default"] = foo();
})
```

matthewp:
Woops, did not know about this!  Thanks! Is it documented.

PR #11167, opened 2020-02-24
Use Yarn 2

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  | :+1:

Prev: https://github.com/babel/babel/pull/11142

---

## NOTE to whoever rebases this PR

To solve conflicts related to the `workspace:` protocol in `package.json`s, use this script:
```js
const cp = require("child_process");
const fs = require("fs");

cp.execSync("git checkout --ours .");

const allPkgs = ["codemods", "eslint", "packages"].flatMap(
  dir => fs.readdirSync(dir).map(d => `${dir}/${d}/package.json`)
).filter(p => {
  try {
    fs.readFileSync(p);
    return true;
  } catch {
    return false;
  }
});

const allNames = allPkgs.map(p => require(process.cwd() + "/" + p).name);

for (const pkg of allPkgs) {
  const p = JSON.parse(fs.readFileSync(pkg, "utf8"));
  let updated = false;

  for (const d of ["dependencies", "devDependencies"]) {
    if (!p[d]) continue;

    for (const [name, version] of Object.entries(p[d])) {
      if (allNames.includes(name) &&!version.startsWith("workspace:")) {
        p[d][name] = `workspace:${version}`;
        updated = true;
      }
    }
  }

  console.log(pkg, updated);
  if (updated) fs.writeFileSync(pkg, JSON.stringify(p, null, 2) + "\n");
}
```

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11167"></a>


.circleci/config.yml
```diff
@@ -2,16 +2,35 @@ version: 2.1
...
 
-  # node_modules shouldn't be cached.
-  # See https://github.com/yarnpkg/berry/issues/988
   - &save-yarn-cache
     paths:
       - .yarn/cache
     key: v2-yarn-cache
 
+  - &save-node-modules-cache
+    # Cache big node_modules folders
```
JLHwung: (😢We are cursed by not using PnP)

nicolo-ribaudo: Actually, after https://github.com/yarnpkg/berry/pull/955 we don't need to cache it anymore!

nicolo-ribaudo:
Ready!

Thanks @arcanis and @larixer for the help :heart:

.circleci/config.yml
JLHwung: That PR is merged, guess we can get rid of the cache?

nicolo-ribaudo:
New failure which wasn't there before the rebase 🙃

arcanis:
Interesting, I tried locally but can't repro it 🤔 Can you try pushing an empty commit to run the test again while preserving the logs for future reference? I wonder if it's consistent. Printing the n_m content might also help shed some light.

nicolo-ribaudo:
@arcanis The problem was caused by accidentally caching `node_modules` on Windows.

Currently, Travis CI automatically caches it for JavaScript projects: it might be good to reach out to them asking to automatically cache `.yarn/cache` (and not `node_modules`) when using Yarn 2.

larixer:
@nicolo-ribaudo Please note, that I'm sure that node-modules linker in berry's `master` is broken at the moment (I have missed the results of require and peer dependency guarantees checks on installed packages), but it will be fixed via this PR:
https://github.com/yarnpkg/berry/pull/1015
Also, I think that caching `node_modules` should work and actually speed up linking. I think both `.yarn/cache` and `node_modules` caching should be beneficial in case of `n_m` linker

nicolo-ribaudo:
Yeah, I'm using that PR (not master).

larixer:
@nicolo-ribaudo I'm hesitant about if it is right to reuse the cached `node_modules` from the broken version of `node-modules` plugin?

nicolo-ribaudo:
So far I have disabled `node_modules` cache so it's not what is causing problems. I will enable it again when CI is green. (Maybe, because the additional cache size might not be worth 5s of saved linking time)

nicolo-ribaudo:
The windows build is failing because of a permission problem. The `node_modules/@babel/helpers` symlink created by Yarn in the root of the monorepo throws EPERM when accessed.

arcanis:
^ Regarding this spawn, the issue is likely that the `yarn.cmd` is a cmd file, and that cmd files cannot be spawned without using `cross-spawn`. Npm, pnpm, and Yarn 1.x used to have [abominations](https://github.com/npm/cmd-shim) to workaround this problem 😟

nicolo-ribaudo:
Could I run `spawn("node", [".yarn/releases/yarn-sources, "node", ...args])`?

arcanis:
Yep, should be fine. Btw, do you have an idea why the job failed? Seems like there's a lingering process, but I'm not sure how to debug that on Travis 🤔

nicolo-ribaudo:
Travis' SSH only works on linux, so I'm debugging everything with `console.log` :grin:

arcanis:
We're so close but I don't see anything damning in the logs 😕 Last resort, I'd probably go with bisecting the test script to find which line exactly is triggering the hang

larixer:
@arcanis @nicolo-ribaudo I see `gpg-agent` in the process list. Is it possible we face this issue:
https://travis-ci.community/t/build-doesnt-finish-after-completing-tests/288/23
?

nicolo-ribaudo:
I have added back `YARN_GPG` (we were already using it with Yarn 1), but a few days ago the build finished even without that option :thinking: 

Anyway, let's see.

arcanis:
@nicolo-ribaudo Fixed with #1036 - try using `yarn set version from sources --branch 1036`

nicolo-ribaudo:
I can't, because I'm currently already using the `hoisting-fixes` branch :sweat_smile:

nicolo-ribaudo:
YEEEEEEEEEEEEEEEEEEESSSSSSSSSSSSSSSSSSSSSSSSSSSSS it green :heart_eyes: :tada: :unicorn: :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:  :heart_eyes: :tada: :unicorn:

nicolo-ribaudo:
TODO: Wait for a yarn release, and https://github.com/babel/babel/pull/11142#discussion_r383011521

arcanis:
💃

arcanis:
> since it is a prerelease, `workspace:^7.8.x` doesn't match it. (@arcanis is it expected?)

Looking at the code yes (the standard `semver.satisfies` function doesn't allow it), but we have a dedicated algorithm for this kind of comparison that ignores rcs for this kind of situation (we use it for peer dependencies, for example). I'll look to use it for the `workspace:` protocol.

nicolo-ribaudo:
@arcanis Sorry, I deleted my comment before seeing your response :sweat_smile: 

---

The last commits show that `make new-version` works with the lerna patches. I'll remove them before merging.

CI is failing and it's expected: `v7.8.8-aaaaaaaaaaaaaaaa.1` is a prerelease, and isn't matched by `7.0.0-0` that we are internally using in the `assertVersion` function.

lerna.json
```diff
@@ -22,11 +22,8 @@
         "*.txt",
         "test/**",
         "**/test/**",
-        "codemods/**",
         "# We ignore every JSON file, except for native-modules, built-ins and plugins defined in babel-preset-env/data.",
-        "@(!(native-modules|built-ins|plugins|package)).json",
-        "# Until the ESLint packages version are aligned with Babel's, we ignore them",
-        "eslint/**"
```
nicolo-ribaudo: For Babel 8 we'll want to publish these packages.

nicolo-ribaudo:
@arcanis `yarn` doesn't do anything after rebasing :thinking: :
```
babel on  yarn-2-compat [$] via ⬢ v13.10.1 took 3s 
➜ yarn
➤ YN0000: ┌ Resolution step

babel on  yarn-2-compat [$!] via ⬢ v13.10.1 
➜ 
```

merceyz:
> via ⬢ v13.10.1

That's a broken version of node, the fix is at https://github.com/nodejs/node/pull/32111 which seems to have landed in node v13.11.0

nicolo-ribaudo:
Oh right, thanks!

arcanis:
I think the Node version can be unpinned - the regression got fixed in the 13.11 as I understand it!

nicolo-ribaudo:
Ok thanks!
Also, I need to rebase this PR after the new release :upside_down_face:

.circleci/config.yml
```diff
@@ -40,12 +30,12 @@ aliases:
 executors:
   node-executor:
     docker:
-      - image: circleci/node:13
+      - image: circleci/node:13.8
```
JLHwung: nit: it should be okay to unpin node.js version.

nicolo-ribaudo: Circle CI doesn't support 13.11 yet: https://circleci.com/docs/2.0/circleci-images/#nodejs

Makefile
```diff
@@ -168,7 +156,7 @@ bootstrap-flow:
 test-flow:
 	$(NODE) scripts/parser-tests/flow
 
-test-flow-ci: build-bundle-ci bootstrap-flow
+test-flow-ci: build-bundle bootstrap-flow
```
JLHwung: The `-ci` variant of make rules specifies the correct dependencies of recipes, so `make build-bundle-ci` ensures that it will run after `bootstrap` while `make build-bunle` (run by human) always assume `make bootstrap` or `make watch` has been run. The correct dependencies are critical to `make -j` on CI.

packages/babel-node/test/index.js
```diff
@@ -112,7 +112,7 @@ const buildTest = function(binName, testName, opts) {
 
     args = args.concat(opts.args);
 
-    const spawn = child.spawn(process.execPath, args);
+    const spawn = child.spawn("node", args);
```
JLHwung: What if `node` is not resolvable from `PATH` env? i.e. User can invoke babel-node via absolute path of `node` bin.

nicolo-ribaudo: I think I changed this because `process.execPath` was not working, let me check again.

Makefile
nicolo-ribaudo: Now that `make bootstrap-only` would be the same as `yarn install` on CI, and thus every command needs it, it would always be executed first because it's needed by every job.

For this reason, I simplified the `Makefile` by moving `yarn install` to the Travis and CircleCI config files. This also makes it possible to show then in different "steps" on CircleCI, making it clearer how long they take.

Do you think that it would still be better to move `yarn` back to the `makefile`, and recreate the `-ci` rules I removed?

.circleci/config.yml
nicolo-ribaudo: TODO: I will wait to merge this PR until it's supported.

Makefile
JLHwung: `bootstrap-flow` can run independently with `bootstrap`. Both of them are network operations so it should still be faster than running sequentially.

However, since `bootstrap-flow` is not that slow, I am good if the execution time benefit is marginal.

.circleci/config.yml
larixer: Looks like CircleCI support 13.11 and 13.12 now.

.travis.yml
```diff
@@ -1,19 +1,20 @@
 git:
   depth: 5
 language: node_js
+
 cache:
-  yarn: true
+  npm: false
+  directories:
+    - .yarn/cache
```
larixer, line 8: I have experimented a bit with Yarn 2 on Travis CI. I came to conclusion that Travis cache is inefficient, it is faster to make Travis not cache anything, neither downloaded archives in `~/.yarn/berry` nor `node_modules`. I have got the best overall perfs with Travis, by setting:
```
cache: false
```
and using Yarn 2 settings below:
```
nodeLinker: node-modules
enableGlobalCache: true
compressionLevel: 0
```
`compressionLevel: 0` makes Yarn 2 store cache in zip files uncompressed, thus not spending much CPU time on compression. 
`enableGlobalCache: true` makes Yarn 2 store zip files only in one place inside `~/.yarn/berry/cache`, it normal y stores copy of them also in `<project_root>/.yarn/cache` via copy-on-write operation (which is not available on all the OSes).

You might have different results, just wanted to share some ideas with you.

nicolo-ribaudo:
@arcanis The e2e test is failing because
- `packages/babel-core` depends on `@babel/helpers@workspace:^7.9.0`
- The top-level `package.json` has a `devDependency` on `@babel/core@7.0.0` (from the registry)
- `@babel/core@7.0.0` has a dependency on `@babel/helpers@^7.0.0`, and I want it to use _exatly_ `@babel/helpers@7.0.0`. If in my top-level dependencies there is any other package depending on `@babel/helpers`, I want it to use 7.0.0 too.
- I added `resolutions: { "@babel/helpers": "7.0.0" }` to my top-level `package.json`, but it also affected `packages/babel-core`'s dependnecy.

arcanis:
The following would do what you wish, but I still see two diffs in the output:

```ts
  pkg.devDependencies['@babel/core'] = '7.0.0';

  for (const pkgName of Object.keys(pkg.devDependencies)) {
    const pkgVersion = pkg.devDependencies[pkgName];
    pkg.resolutions[pkgName + '@npm:' + pkgVersion + '/@babel/helpers'] = '7.0.0';
  }

  fs.writeFileSync('./package.json', JSON.stringify(pkg, null, 2));
```

image

image

nicolo-ribaudo:
Yeah, the second output is caused by a new version of a dependency.

nicolo-ribaudo:
> The following would do what you wish, but I still see two diffs in the output:

Not exactly: there is still a `@babel/helpers@7.9.0` in the dependency tree of the top-level package.json.

nicolo-ribaudo:
@CompuIves May ask for help? :sweat_smile::pray:

Is it possible to configure the exact command to run in `.codesandbox/ci.json` to install the dependencies? I would need to run exactly `yarn` or `yarn install`.

CompuIves:
Yep! That would work. You can set an `installCommand` in the config, more info on that is here:

https://codesandbox.io/docs/ci#configuration.

So I think in your case you would do `install`? I was thinking that it would already run `install` :thinking:.

CompuIves:
Ahh, I see a command is already set here: https://github.com/babel/babel/blob/master/.codesandbox/ci.json#L2

You could replace this maybe with install, and set the build command to `codesandbox`.

nicolo-ribaudo:
I tried here: https://github.com/babel/babel/pull/11167/commits/ddac2c50ea84c9a6f6f24ab087050725417aadbc

But it runs `yarn run install` instead of `yarn install`: https://ci.codesandbox.io/status/babel/babel/pr/11167/builds/25493

CompuIves:
Ohh interesting. So it doesn't work in Yarn v2 if I understand correctly? Okay, that's a valid use-case. I'll remove `--pure-lockfile`.

nicolo-ribaudo:
Maybe it makes sense to first run `yarn --version` and the only remove the option when using Yarn 2? (cc @arcanis)

arcanis:
This would be ideal, yes. For context, `--pure-lockfile` got removed since it's almost never what people were looking for (outdated lockfiles are still allowed, they just won't be persisted on disk). The useful flag was `--frozen-lockfile`, which got renamed into `--immutable`.

CompuIves:
Ah, that makes sense. I removed it for now, could you maybe make a commit to test whether the new build works?

nicolo-ribaudo:
It still throws about `--non-interactive`.

@arcanis Is simply running `yarn install` safe, or should CodeSandbox run `yarn --version` and then choose between `yarn install --frozen-lockfile --non-interactive` and `yarn --immutable`? (assuming that's it's possible in CodeSandbox' install process)

arcanis:
Non-interactive [is already implied on CI](https://github.com/yarnpkg/yarn/blob/25d5526c8995443b03c113c44f9c8e9a79c2c074/src/config.js#L483), so it can be safely removed.

CompuIves:
It's now removed! If you create an empty commit the build should succeed now.

nicolo-ribaudo:
It works :tada: :tada: :tada:

nicolo-ribaudo:
(Ignore the coverage drop, it's a bug in the action)

nicolo-ribaudo:
# :warning: Use "rebase and merge", it will make it easier to fix conflicts

nicolo-ribaudo:
I'm counting @arcanis's review for our two-:white_check_mark: policy, since as a Yarn maintainer he is one of the most qualified people out there to review this PR!

kachkaev:
As of now, CONTRIBUTING.md on  `next-8-dev` and `next-8-rebased` still says:

> *Yarn*: Make sure that Yarn 1 is [installed](https://classic.yarnpkg.com/en/docs/install) with version >= `1.19.0`.

Should the instructions be changed in a follow-up PR? The doc should probably include tips on how to use yarn2 without uninstalling yarn1 because this is what most devs are still using.

merceyz:
Yarn 1 will shell out to the local version in the repository

arcanis:
The documentation is still up to date - as of this PR the Yarn version is stored within the repository, external contributors only need *a* global Yarn, regardless of its version.

PR #11173, opened 2020-02-26
Add `onlyRemoveTypeImports` option to `transform-typescript`

### LINKED ISSUE #10981: `import type` in TypeScript

kaicataldo:
| Fixed Issues?            | Fixes #10981
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is based on the [`feat-7.9.0/ts-import-type`](https://github.com/babel/babel/tree/feat-7.9.0/ts-import-type) branch, and I believed the test failures are from that branch.

With the addition of type-only imports and exports to TypeScript, we are now able to explicitly mark imports as types. This PR adds an `onlyExplicitTypeImports` option to `plugin-transform-typescript`, which allows the user to opt into only removing imports that are unambiguously types. We will probably want to make this the default behavior in the future.

Questions for reviewers:
1. Are there any missing test cases? 
1. Would it be worth adding a second test file that explicitly sets the new option to `false`?


nicolo-ribaudo:
@kaicataldo I have rebased `feat-7.9.0/...` and your branch on master to make CI pass.

kaicataldo:
`onlyRemoveTypeImports` seems like the clearest option name to me.

nicolo-ribaudo:
Kai doesn't like using newly learned words! :stuck_out_tongue:

kaicataldo:
Maybe I did know the word at some point and it was elided from my vocabulary.

nicolo-ribaudo:
Could you also add tests for `import {} from "x"` and `import "x"`, since they are explicitly pointed out in a comment in the code?

nicolo-ribaudo:
Merged into #11171

JLHwung:
We also need
1) add this option to `preset-typescript`.
2) update docs on https://babeljs.io/docs/en/babel-preset-typescript#options and https://babeljs.io/docs/en/babel-plugin-transform-typescript#options

PR #11175, opened 2020-02-26
replace `gulp-watch` by `gulp.watch`

JLHwung:
| Fixed Issues?            | When running `make watch` for a long time after several git checkout, multiple file watchers are registered
| Any Dependency Changes?  | `gulp-watch` is removed

When running `make watch` for a long time and after several `git checkout`, `gulp-watch` will register multiple file watchers for a single file. It seems to me the file watcher is leaking.

Replacing `gulp-watch` with builtin `gulp.watch` seems to fix this issue.

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -5,7 +5,6 @@ const through = require("through2");
 const chalk = require("chalk");
 const newer = require("gulp-newer");
 const babel = require("gulp-babel");
-const gulpWatch = require("gulp-watch");
 const fancyLog = require("fancy-log");
 const filter = require("gulp-filter");
 const gulp = require("gulp");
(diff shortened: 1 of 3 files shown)


nicolo-ribaudo:
We have had this problem since forever, but it was easier to restart `make watch` than to fix it :laughing:

Gulpfile.js
```diff
@@ -239,10 +238,6 @@ gulp.task("build-no-bundle", () => buildBabel());
 gulp.task(
   "watch",
   gulp.series("build-no-bundle", function watch() {
-    gulpWatch(
-      defaultSourcesGlob,
-      { debounceDelay: 200 },
-      gulp.task("build-no-bundle")
-    );
+    gulp.watch(defaultSourcesGlob, gulp.task("build-no-bundle"));
```
JLHwung: The `gulp.watch` equivalent of `{ debouceDelay: 200 }` is { delay: 200 }`. According to https://gulpjs.com/docs/en/api/watch#options `200` is the default `delay` value, so we remove option argument here.

PR #11178, opened 2020-02-27
Implement support for `declare` on class fields with Flow

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | https://github.com/babel/website/pull/2196
| Any Dependency Changes?  |

Flow has recently landed support for `declare` fields, similar to TypeScript. (https://github.com/facebook/flow/commit/11b7adbf85ea237cfc01d1b8dc6f6dcdbc299157)

This PR implements parser support, and adds an `allowDeclareFields` option to the transform plugin to mirror the one of `@babel/plugin-transform-typescript`. In Babel 8 it will be enabled by default.

Even if it's not released yet in Flow, I'm adding this to the v7.9.0 milestone since Flow is released about once a week.

cc @mroch


packages/babel-parser/test/fixtures/flow/class-properties/declare-static-field/input.js
```diff
@@ -0,0 +1,3 @@
+class A {
+  declare static foo
+}
```
JLHwung: Can you add a new test case swapping `declare` and `static`? It is allowed in TypeScript but not flow.
```js
class A {
  static declare foo
}
```

packages/babel-parser/test/fixtures/flow/class-properties/declare-field-computed/input.js
```diff
@@ -0,0 +1,3 @@
+class A {
+  declare [foo]
+}
```
JLHwung: Flow does not support declared computed class fields yet: https://flow.org/try/#0G4QwTgBAZg9jEF4IEYDcAoAxgGxAZzwgEEIBvdCCAEwFMdwaIBtWGAXXQF8g

nicolo-ribaudo: It's not released yet.

Looking at the source code it doesn't seem to be disallowed (https://github.com/facebook/flow/commit/11b7adbf85ea237cfc01d1b8dc6f6dcdbc299157#diff-ca4a151bb3ce5bd8c11f1bc3b00036ebR619), but I don't see any test to confirm my assumption.

cc @@mroch should they be supported?

mroch: change the version selector in the top right to `master` and it parses.

It does error with "Computed property keys not supported.", but that's a type error, not a parse error (you can tell the difference on the CLI, but on Try Flow it's only in the JSON tab (`"kind": "infer"`))

nicolo-ribaudo: Ok, thanks :+1:

nicolo-ribaudo:
I split the error into two different errors to match Flow (https://github.com/facebook/flow/blob/cb99c11e0f620a981b6cea5c5ba2f02c7263679b/src/parser/parse_error.ml#L325-L326).

JLHwung:
@nicolo-ribaudo Can you also update https://babeljs.io/docs/en/babel-preset-flow#options?

Issue #11183, opened 2020-02-27
[Parser]: First token after a class is considered strict mode


pvdz:
I guess when parsing the `}` closing token of a class, the next token is still parsed as the strict-mode context from that class, rather than using the outer strict mode status.

In script mode, this (a legacy octal) passes in Babel (because it applies annexb by default):
```js
05
```
This fails after a class:
```js
class x {}
05
```
But passes with anything in between:
```js
class x {}
y
05
```

(All the above cases should pass in non-strict script mode under annexb rules)

Babel Parser version: 7.8.4

PR #11186, opened 2020-02-28
fix: token after strict mode block is evaluated in strict mode

kaicataldo:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/11183
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

We're currently eating the closing bracket token of the class body before restoring the outer scope's strict mode state. This PR changes it so that we restore the out strict mode state before eating this final token.

Question: Does this seem like the correct way to fix this?

### DIFF
--- a/packages/babel-parser/test/fixtures/es2015/regression/11183/input.js
+++ b/packages/babel-parser/test/fixtures/es2015/regression/11183/input.js
@@ -1,0 +1,5 @@
+class X {}
+05
+
+function x() { 'use strict' }
+05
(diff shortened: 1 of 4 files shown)


packages/babel-parser/test/fixtures/es2015/regression/11183/input.js
```diff
@@ -0,0 +1,17 @@
+class A {}
```
JLHwung: Can you also a test case about function declaration, which is failing on master
```js
function A() { "use strict"; }
05
```

kaicataldo:
Thanks for the review! Will take a look later this afternoon.

kaicataldo:
Waiting to land https://github.com/babel/babel/pull/11188 first since the two PRs actually overlap a bit.

JLHwung:
FYI https://github.com/acornjs/acorn/commit/44e52b237b72cca6654d513344e6c33c2194e2ff is how acorn fixes this issue. That is a neat fix, maybe we can port it here?

kaicataldo:
This is ready for re-review!

PR #11188, opened 2020-02-28
fix: non-directive "use strict" should not enable parsing in strict mode

kaicataldo:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/10412
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Since I was in the parser looking at strict mode parsing bugs, I decided to take a stab at this! Building on the work started in https://github.com/babel/babel/pull/10451 (hope that's okay - it looks like an orphaned PR), this changes the strict mode check to be less brittle and to ensure that `"use strict"` is actually a directive when setting strict mode when function bodies are parsed.


packages/babel-parser/test/fixtures/esprima/invalid-syntax/migrated_0222/output.json
```diff
@@ -13,8 +13,7 @@
     }
   },
   "errors": [
-    "SyntaxError: Octal literal in strict mode (1:36)",
-    "SyntaxError: Octal literal in strict mode (1:57)"
```
kaicataldo: Does anyone else see why this might have changed? There are two octal literals in this test, so it seems correct to report two errors.

kaicataldo: Need to figure out why this second error was removed, as there are two octal literals in this test.

packages/babel-parser/src/parser/util.js
```diff
@@ -192,29 +190,6 @@ export default class UtilParser extends Tokenizer {
     }
   }
 
-  strictDirective(start: number): boolean {
```
kaicataldo: This is now dead code.

packages/babel-parser/src/parser/statement.js
```diff
@@ -854,6 +863,7 @@ export default class StatementParser extends ExpressionParser {
     directives: ?(N.Directive[]),
     topLevel: boolean,
     end: TokenType,
+    afterBlockParse?: Function,
```
nicolo-ribaudo: Can you refine this type annotation? (Also in the other functions)

```suggestion
    afterBlockParse?: (hasStrictModeDirective: boolean) => void,
```

packages/babel-parser/src/parser/util.js
```diff
@@ -4,12 +4,10 @@ import { types as tt, type TokenType } from "../tokenizer/types";
 import Tokenizer from "../tokenizer";
 import State from "../tokenizer/state";
 import type { Node } from "../types";
-import { lineBreak, skipWhiteSpace } from "../util/whitespace";
+import { lineBreak } from "../util/whitespace";
 import { isIdentifierChar } from "../util/identifier";
 import * as charCodes from "charcodes";
 
-const literal = /^('|")((?:\\?.)*?)\1/;
```
nicolo-ribaudo: Deleted regexp -> Good news :laughing:

packages/babel-parser/test/fixtures/esprima/invalid-syntax/migrated_0222/output.json
nicolo-ribaudo: <details>
<summary> :warning: Spoiler :warning: </summary>

This is probably because before this PR, we first checked if there was a `use strict` directive _and then_ tokenized the invalid literals.
Now, we first tokenize the invalid literals, then set `this.state.strict` to `true`, and in `parseBlockOrModuleBlockBody` only throw for the first invalid one. It should be fixed there.

</details>

packages/babel-parser/src/parser/util.js
nicolo-ribaudo: <details>
<summary> :warning: Spoiler :warning: </summary>

This is probably because before this PR, we first checked if there was a `use strict` directive _and then_ tokenized the invalid literals.
Now, we first tokenize the invalid literals, then set `this.state.strict` to `true`, and in `parseBlockOrModuleBlockBody` only throw for the first invalid one. It should be fixed there.

</details>

kaicataldo:
This is ready for re-review!

packages/babel-parser/test/fixtures/esprima/invalid-syntax/migrated_0222/output.json
kaicataldo: This has now been fixed.

packages/babel-parser/src/tokenizer/state.js
```diff
@@ -143,7 +143,7 @@ export default class State {
 
   // TODO
   containsOctal: boolean = false;
-  octalPosition: ?number = null;
+  octalPosition: number | null = null;
```
kaicataldo: This doesn't look like it can ever be `undefined`.

packages/babel-parser/src/parser/statement.js
```diff
@@ -838,14 +847,20 @@ export default class StatementParser extends ExpressionParser {
...
+    let oldStrict = null;
 
     while (!this.eat(end)) {
-      if (!parsedNonDirective && this.state.containsOctal && !octalPosition) {
-        octalPosition = this.state.octalPosition;
+      // Track octal literals that occur before a "use strict" directive.
+      if (
+        !parsedNonDirective &&
+        this.state.containsOctal &&
+        this.state.octalPosition !== null
```
nicolo-ribaudo: Question: is this just needed for flow, or it's possible that `containsOctal` is `true` but `octalPosition` is false?

If it's just for flow, I have started to use this pattern in other parts of the parser:
```js
/*:: declare var invariant; */ // <-- At the beginning of the file

// ...

if (!parsedNonDirective && this.state.containsOctal) {
  /*:: invariant(this.state.octalPosition !== null) */
  octalPositions.push(this.state.octalPosition);
}
```

kaicataldo: Thanks for this feedback! I actually think we can get rid of `this.state.octalPosition` altogether. Please let me know what you think!

packages/babel-parser/src/parser/statement.js
```diff
@@ -838,14 +847,19 @@ export default class StatementParser extends ExpressionParser {
...
 
     while (!this.eat(end)) {
-      if (!parsedNonDirective && this.state.containsOctal && !octalPosition) {
-        octalPosition = this.state.octalPosition;
+      // Track octal literals that occur before a "use strict" directive.
+      if (!parsedNonDirective && this.state.containsOctal) {
+        // This octal literal will always occur in a directive and will
+        // therefore be escaped. Subtract two from the current position to
+        // account for the "\" character.
+        octalPositions.push(this.state.pos - 2);
```
nicolo-ribaudo: What if it's not at the end of the string, or if it has more than one digit?

`"foo \03 bar"`

nicolo-ribaudo: Maybe we could remove `containsOctal`, since it's the same as `octalPosition !== null`.

kaicataldo: Goooooood call. I guess another option could be to not track this at all and use a regex check on directives? Not sure how big a hit that would be performance-wise, but it would allow us to remove all this extra "global" state.

nicolo-ribaudo: It would probably be ok, because it super rare to have another directive before `"use strict"` anyway.

kaicataldo: After exploring this some more, I think we should just go with removing `containsOctal` for now. Removing global state doesn't feel like a good enough reason to recalculate the position, imo.

kaicataldo: I ended up changing `octalPosition` to `octalPositions` and made it an array, because the current behavior only reports the last octal literal found in a directive (since it doesn't get checked until after the token has been parsed). With an array, we can track the positions of multiple octal literals and warn on each of them.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -1296,11 +1291,10 @@ export default class Tokenizer extends LocationParser {
               return null;
             } else if (this.state.strict) {
               this.raise(codePos, Errors.StrictOctalLiteral);
-            } else if (!this.state.containsOctal) {
-              // These properties are only used to throw an error for an octal which occurs
+            } else {
+              // This properties are only used to throw an error for an octal which occurs
```
existentialism:
```suggestion
              // This property is only used to throw an error for an octal which occurs
```

🙃

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -1033,11 +1032,7 @@ export default class Tokenizer extends LocationParser {
       this.input.charCodeAt(start) === charCodes.digit0;
     if (octal) {
       if (this.state.strict) {
-        this.raise(
-          start,
-          // todo: merge with Errors.StrictOctalLiteral
-          "Legacy octal literals are not allowed in strict mode",
-        );
+        this.raise(start, Errors.StrictOctalLiteral);
```
kaicataldo: I believe this change unifies the language for all the octal literal strict mode errors.

packages/babel-parser/src/tokenizer/index.js
kaicataldo: https://media.giphy.com/media/2wZvHDi5BgZ9VugXlo/source.gif

packages/babel-parser/src/parser/statement.js
```diff
@@ -870,6 +877,22 @@ export default class StatementParser extends ExpressionParser {
       body.push(stmt);
     }
 
+    // Throw an error for any octal literals found before a
+    // "use strict" directive. Strict mode will be set at parse
+    // time for any literals that occur after the directive.
+    if (this.state.strict && octalPositions.length) {
+      for (const pos of octalPositions.flat()) {
```
nicolo-ribaudo: This isn't supported in node <11.

packages/babel-parser/src/parser/statement.js
```diff
@@ -838,14 +847,16 @@ export default class StatementParser extends ExpressionParser {
...
-    let oldStrict;
-    let octalPosition;
+    let oldStrict = null;
 
     while (!this.eat(end)) {
-      if (!parsedNonDirective && this.state.containsOctal && !octalPosition) {
-        octalPosition = this.state.octalPosition;
+      // Track octal literals that occur before a "use strict" directive.
+      if (!parsedNonDirective && this.state.octalPositions.length) {
+        octalPositions.push(this.state.octalPositions);
```
```suggestion
        octalPositions.push.apply(octalPositions, this.state.octalPositions);
```
or
```suggestion
        octalPositions.push(...this.state.octalPositions);
```

And you can avoid using `.flat()`

packages/babel-parser/src/parser/statement.js
kaicataldo: Ah, right. Thanks for catching that!

nicolo-ribaudo:
@kaicataldo Please merge this PR yourself, I don't want to steal attribution :pray:

nicolo-ribaudo:
The e2e failure doesn't seem related.

PR #11190, opened 2020-02-28
Fix printing parentheses around optional chains

### LINKED ISSUE #11187: Forced calls of OptionalCalls do not print correctly
## Bug Report

It's possible to force the calling of an `OptionalCallExpression` by wrapping it in parentheses:

```js
foo?.();     // <- That's an optional call
foo?.()();   // <- That's an optional call of an optional call

(foo?.())(); // <- That's a forced call of an optional call
```

This roughly translates to the following:

```js
foo == null ? undefined : foo();     // <- That's an optional call
foo == null ? undefined : foo()();   // <- That's an optional call of an optional call

(foo == null ? undefined : foo())(); // <- That's a forced call of an optional call
```

Now, the bug: When we print these without transforming them (meaning we're going to output ES2020 syntax), the forced call loses its parentheses:

```js
// Input
(foo?.())();

// Output
foo?.()();
```

This is incorrect! If we were to try parsing that output, it'd be an optional call of an optional call, not a forced call of an optional call.

**Current Behavior**

Outputs `foo?.()()`, which is incorrect

**Input Code**
- [REPL](https://babeljs.io/repl/#?browsers=&build=&builtIns=false&spec=false&loose=true&code_lz=PTAEDEHsCcGMFMAmpYEMA27SQGalQHbYAOALgJaQEYoboBQAFDpJAPwB0jAlNzwNz16ICDATIAtvAkAjeNGx5CJClRppMTFuy68OM1AC9-QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=true&targets=&version=7.8.6&externalPlugins=)

```js
(foo?.())();
```

**Expected behavior/code**

Outputs `(foo?.())()`

**Possible Solution**

We need to check if parenthesis are necessary when outputting an `OptionalCallExpression`. We already do this for `OptionalMemberExpression`s (which have a very similar behavior).

https://github.com/babel/babel/blob/d13fd7c7bdb51995e8c9171b7047fc93402f8c3a/packages/babel-generator/src/node/parentheses.js#L254-L259

You can verify that if you input `(foo?.bar).baz` (a forced member of an optional member), it will have parenthesis in the output.

We need to test the following to cases:

```js
// Forced call of an optional call
(foo?.())();

// Forced member of an optional call
(foo?.()).baz;
```

sag1v:
| Fixed Issues?            | Fixes #11187 
| Documentation PR Link    | 

Updated `OptionalMemberExpression` to compare the `callee` for `callExpression`  or `object` for `memberExpression` against the `node`.
This should fix a case such as:
```js
foo[bar?.baz];
```
That transforms to:
```js
foo[(bar?.baz)];
```

Added a `OptionalCallExpression` as well with the same checks as `OptionalMemberExpression`.

Should solve cases of:
```js
(foo?.())();
(foo?.()).baz;
```


existentialism:
@sag1v nice work!

PR #11192, opened 2020-03-01
Refactor: add parser message template

JLHwung:

This PR collects the scattered babel parser error messages into a centralized message store. By doing so we can implement [expression scops](https://docs.google.com/document/d/1FAvEp9EUK-G8kHfDIEo_385Hs2SUBCYbJ5H-NnLvq8M/edit) which records different types of errors when parsing ambiguous patterns.

The naming conventions of error messages strictly follows v8's [message template](https://cs.chromium.org/chromium/src/v8/src/common/message-template.h) if they exists, otherwise I coin some phrase so PTAL on the message template. 🙏

Todo
- [x] flow plugin
- [x] typescript plugin
- [x] estree plugin
- [x] jsx plugin

### DIFF
--- a/packages/babel-parser/src/parser/location.js
+++ b/packages/babel-parser/src/parser/location.js
@@ -16,6 +16,7 @@ type ErrorContext = {
   code?: string,
 };
 
+// The Errors key follows https://cs.chromium.org/chromium/src/v8/src/common/message-template.h unless it does not exist
 export const Errors = Object.freeze({
   ArgumentsDisallowedInInitializer:
     "'arguments' is not allowed in class field initializer",
--- a/packages/babel-parser/src/parser/location.js
+++ b/packages/babel-parser/src/parser/location.js
@@ -213,7 +213,7 @@ export default class LocationParser extends CommentsParser {
 
   raiseWithData(
     pos: number,
-    data: {
+    data?: {
       missingPlugin?: Array<string>,
       code?: string,
     },
@@ -226,6 +226,7 @@ export default class LocationParser extends CommentsParser {
       ` (${loc.line}:${loc.column})`;
     return this._raise(
       Object.assign(
+        {},
         {
           loc,
           pos,
--- a/packages/babel-parser/src/parser/location.js
+++ b/packages/babel-parser/src/parser/location.js
@@ -16,7 +16,179 @@ type ErrorContext = {
   code?: string,
 };
 
-export const Errors = {};
+export const Errors = Object.freeze({
+  ArgumentsDisallowedInInitializer:
+    "'arguments' is not allowed in class field initializer",
+  AsyncFunctionInSingleStatementContext:
+    "Async functions can only be declared at the top level or inside a block",
+  AwaitBindingIdentifier:
+    "Can not use 'await' as identifier inside an async function",
+  AwaitExpressionFormalParameter:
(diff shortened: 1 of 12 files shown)


packages/babel-parser/src/parser/location.js
```diff
@@ -26,33 +207,39 @@ export default class LocationParser extends CommentsParser {
...
+    errorTemplate: string,
+    ...params: any
   ): Error | empty {
     const loc = this.getLocationForPosition(pos);
-
-    message += ` (${loc.line}:${loc.column})`;
-    // $FlowIgnore
-    const err: SyntaxError & { pos: number, loc: Position } = new SyntaxError(
+    const message =
+      errorTemplate.replace(/%(\d+)/g, (_, i: number) => params[i]) +
```
JLHwung: The error message now supports naive `%0`-style templates.

packages/babel-parser/src/plugins/flow.js
```diff
@@ -42,6 +43,79 @@ const reservedTypes = new Set([
   "void",
 ]);
 
+/* eslint sort-keys: "error" */
+const flowErrors = Object.freeze({
+  AmbiguousConditionalArrow:
```
nicolo-ribaudo: Idea: maybe we should prefix these errors with `Flow`?

JLHwung:
I am not sure if I should do this for TypeScript plugin because the way ts compiler indexes diagnostics is quite brutal: https://github.com/microsoft/TypeScript/blob/9c71eaf59040ae75343da8cdff01344020f5bba2/lib/diagnosticMessages.generated.json

nicolo-ribaudo:
We should probably diverge from their error codes, and be consistent with the other parts of our parser.

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -58,6 +59,42 @@ type ParsingContext =
   | "TypeMembers"
   | "TypeParametersOrArguments";
 
+const tsErrors = Object.freeze({
```
existentialism: Consistency nit, but `tsErrors` vs `TsErrors`? (Since it's `Errors` above)

existentialism:
Time for a lint rule that prevents `this.raise` to be called without a constant 😝

packages/babel-parser/src/parser/location.js
```diff
@@ -26,33 +208,40 @@ export default class LocationParser extends CommentsParser {
...
+      ` (${loc.line}:${loc.column})`;
+    return this._raise(
+      Object.assign(
+        {},
+        {
+          loc,
+          pos,
+        },
+        data,
+      ),
```
nicolo-ribaudo: The first empty object isn't needed.

```suggestion
      Object.assign({ loc, pos }, data),
```

JLHwung: It was required by Flow otherwise `{ loc, pos }` is a sealed object. https://also.github.io/notes/2016/03/13/flow-object-assign.html

If we are okay with that I can `FlowIgnore` it lol.

nicolo-ribaudo: Maybe this?

```suggestion
      Object.assign(({ loc, pos }: Object), data),
```

PR #11193, opened 2020-03-01
Pass URLs to import() rather than paths

nicolo-ribaudo:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Native `import()` expects an URL, but we were passing it an absolute file path. It worked on linux, but fails on windows because `C:` is parsed as an URL protocol.

https://github.com/nodejs/node/issues/31710

We don't have tests for this, because Jest doesn't support native `import()` yet.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,5 +1,7 @@
 "use strict";
 
+const currentNodeSupportsURL = !!require("url").pathToFileURL;
+
 module.exports = function(api) {
   const env = api.env();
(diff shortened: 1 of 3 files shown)


nicolo-ribaudo:
All the failures are probably because `resolve` doesn't support URLs :thinking:

PR #11194, opened 2020-03-01
Allow defining the moduleIds-related option in the transform plugins

### LINKED ISSUE #5473: [7.0] Consider moving root AMD/UMD/SystemJS options to be plugin/preset options?
Raising this for discussion for now because I'm curious, but not at all convinced what the right move would be.

We currently have the following options to `babel-core`:
```
// Boolean to enable inserting names for AMD/UMD/SystemJS modules, since they are
// normally anonymous because the loader already knows the filename to load it anyway.
"moduleIds",

// Function to modify the AMD/UMD/SystemJS module name once babel-core has generated it.
"getModuleId",

// Provide a default base prefix for AMD/UMD/SystemJS module paths.
"moduleRoot",

// An explicit string name to skip `babel-core`s programmatic generation.
"moduleId",

// Used as the base for module names. Defaults to the "filename" passed to babel.
"filenameRelative"
```

These are all then exposed via `state.file.getModuleName()` from `babel-core`.

We own these module transforms and we want to make the usage as easy as possible, but it also seems strange to me to have these module arguments hardcoded into `babel-core`. If we do it this way "because it is easiest" that also means that while we get a free pass, noone else could write a plugin that needs arguments like this in an easy, so we avoid dogfooding our own APIs by sidestepping them entirely.

Depending on the usecase, it seems like all but "moduleId" are things that would be pretty static for users, and now that we have `.babelrc.js`, people can do

```
module.exports = {
  plugins: [
    ['transform-es2015-modules-systemjs', {
      moduleIds: true,
      moduleRoot: 'myApp',
      getModuleId (name) {
        return name + "suffix";
      },
    }],
  ],
};
```

They could no longer pass the string ones via `babel --module-ids` and such, so that's one potential downside. It also means people couldn't just do

```
{
  presets: ['es2015', {modules: 'systemjs'}],
}
```

and then use the core Babel arguments to configure the rest separately.

At least some of this is an argument for supporting plugin options in the CLI like https://github.com/babel/babel/issues/4800

@guybedford I'd be curious for your thoughts as the most go-to person for non-CommonJS module formats.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #5473
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR is the first step towards https://github.com/babel/babel/issues/5473. I didn't move `filenameRelative` since it's not mentioned in the [modules-related docs](https://babeljs.io/docs/en/options#amd-umd-systemjs-module-options), and it's used as a default value for `sourceFileName`'s option.

When this PR is merged, I will open one in the `next-8-dev` branch to remove the global options. This will make it easier to migrate before upgrading to Babel 8.


packages/babel-core/package.json
```diff
@@ -46,6 +46,7 @@
     "@babel/template": "^7.8.6",
     "@babel/traverse": "^7.8.6",
     "@babel/types": "^7.8.6",
+    "@babel/helper-module-transforms": "^7.8.6",
```
existentialism: Nit: alpha sort

PR #11198, opened 2020-03-03
fix(babel-parser): chain off optionally chained keys named class and function

### LINKED ISSUE #11197: Transpilation fails when chaining off optionally chained keys that share names with certain reserved keywords (e.g. class and function)
## Bug Report

**Current Behavior**

I have a statement that goes something like the following: `const a = foo?.class.bar`. This does not transpile, and I'm quite sure I've found the reason why: Because I'm trying to chain `bar` off the optionally chained `class`. 

I've been playing around in the REPL using only `@babel/plugin-proposal-optional-chaining v7.6.0` plugin:
https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAhjAvDAZiEB-AdMANnCCLAIzgCcg&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=false&targets=&version=7.8.6&externalPlugins=%40babel%2Fplugin-proposal-optional-chaining%407.6.0

The following expressions do *not* work in babel (but works fine in the Chrome console, v80):
```javascript
foo?.class.baz
foo?.function.baz
foo?.bar?.class.bar
foo?.function?.baz
```

The following expressions *do* work:
```javascript
foo?.bar.class.bar
foo?.class
foo?.['function'].bar
foo?.['class'].bar
foo.class.bar
foo?.let.bar
foo?.const.bar
foo?.var.bar
```

*From this, I've formed the following hypothesis*: 
- Using the reserved keywords `class` or `function` directly after an optional chaining operator, followed by a regular or optional chain leads to babel being unable to transpile the expression, e.g. `foo?.class.baz`. 
- The effect disappears when using bracket notation around `function` or `class`, e.g. `foo?.['class'].bar`. 
- Using other keywords (e.g. `var`, `let` and `const`) or just not chaining off the `class` or `function` expression also leads to the effect disappearing, e.g. `foo?.let.bar` and `foo.class.bar`. 
- And if there's no optional chaining operator in the first place, it also works fine, e.g. `foo.class.bar`

**Expected behavior/code**
I expect the code to transpile and to behave as if using non-reserved keywords, i.e. the way the non-transpiled code behaves in Chrome

Vages:
| Fixed Issues?            |  Fixes #11197
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    |
| Any Dependency Changes?  |

Added the ability to parse expressions such as those described in #11197. (This may be considered a minor upgrade, but I would not, as it does not extend the current API as much as it restores it to its expected behavior.)

### DIFF
--- a/packages/babel-parser/src/tokenizer/index.js
+++ b/packages/babel-parser/src/tokenizer/index.js
@@ -233,7 +233,7 @@ export default class Tokenizer extends LocationParser {
       return;
     }
 
-    if (curContext.override) {
+    if (curContext?.override) {
       curContext.override(this);
     } else {
       this.getTokenFromCode(this.input.codePointAt(this.state.pos));
--- a/packages/babel-parser/src/tokenizer/index.js
+++ b/packages/babel-parser/src/tokenizer/index.js
@@ -233,8 +233,9 @@ export default class Tokenizer extends LocationParser {
       return;
     }
 
-    if (curContext?.override) {
-      curContext.override(this);
+    const override = curContext?.override;
+    if (override) {
+      override(this);
     } else {
       this.getTokenFromCode(this.input.codePointAt(this.state.pos));
     }
--- a/packages/babel-parser/test/fixtures/es2020/optional-chaining/chaining-off-optionally-chained-keys-named-class-or-function/input.js
+++ b/packages/babel-parser/test/fixtures/es2020/optional-chaining/chaining-off-optionally-chained-keys-named-class-or-function/input.js
@@ -1,0 +1,4 @@
+foo?.class.bar
+foo?.function.bar
+foo?.bar?.class.bar
+foo?.function?.bar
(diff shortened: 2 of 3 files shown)


nicolo-ribaudo:
Thanks!

Vages:
The build seems to fail. It seems to work fine for node 6, 8 and 12, but 10 errors. Could any of you maintainers retry it? Perhaps we should try replacing `const override = curContext?.override;` with `const override = curContext && curContext.override;` to see if that helps.

nicolo-ribaudo:
I restarted it, lets see.

existentialism:
Nice work, @Vages!

PR #11200, opened 2020-03-03
Don't create `TSParenthesizedType` when `createParenthesizedExpressio…

### LINKED ISSUE #9546: [breaking] Remove `TSParenthesizedType` unless createParenthesizedExpressions is enabled
I don't understand why we generate `TSParenthesizedType` nodes by default. Example: [AST Explorer](https://astexplorer.net/#/gist/d54a52be17b8ae39d3f3b6b394084602/c948055ee400984a7471e011bd08e0a3cd1184da). It isn't needed (it isn't part of the AST, but more a CST thing) and we don't use it from Flow.

Also, we should add it to Flow when that option is enabled.

Ref: #8025

smelukov:
…ns` disabled (#9546)

| Fixed Issues?            | Fixes #9546 
| Documentation PR Link    | 
| Any Dependency Changes?  |


nicolo-ribaudo:
The first failure (`generation/typescript › types conditional infer`) is the `@babel/generator` I explained in the issue, and it should be fixed.

The second one (`generation/typescript › types parenthesized`) actually looks OK, and that output file should be re-generated (you can delete `output.js` and run `yarn jest babel-generator`).

nicolo-ribaudo:
Could you rebase this PR? We have merged the commit from your other PR to the `next-8-dev` branch.

smelukov:
@nicolo-ribaudo done

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -773,6 +773,13 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         case tt.bracketL:
           return this.tsParseTupleType();
         case tt.parenL:
+          if (!this.options.createParenthesizedExpressions) {
+            this.expect(tt.parenL);
+            const type = this.tsParseType();
+            this.expect(tt.parenR);
+            return type;
+          }
```
nicolo-ribaudo: Can you use `addExtra` to add info about parenths, like we do here?

https://github.com/babel/babel/blob/f9d4b799b74d4cc381679a3124f2c8c58745f4c0/packages/babel-parser/src/parser/expression.js#L1335-L1336

smelukov:
rebased ✅

oliverdunk:
I think there might be one remaining failure, because `types/tuple-optional/output.json` needs regenerating with the `extra` property?

Other than that, looks great!

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -773,6 +773,16 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         case tt.bracketL:
           return this.tsParseTupleType();
         case tt.parenL:
+          if (!this.options.createParenthesizedExpressions) {
+            const startPos = this.state.start;
+            this.expect(tt.parenL);
```
JLHwung, line 778: It suffices to move the cursor only `this.next()` instead of `this.expect(tt.parenL)` because the switch case has implied `this.match(tt.parenL)`.

nicolo-ribaudo:
Merged at https://github.com/babel/babel/commit/62290aa1badf7846168cb6c26a9ca63c738f14ce

PR #10971, opened 2020-01-09
Add numeric separator to `shippedProposals`

Wetinee:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #10921 
| Patch: Bug Fix?          |  No
| Major: Breaking Change?  | No
| Minor: New Feature?      | Yes
| Tests Added + Pass?      | Not yet
| Documentation PR Link    | Not yet
| Any Dependency Changes?  | Yes
| License                  | MIT

I've added numeric separator in babel-preset-env and babel-preset-dev-env, but failed some tests and I haven't found out why.  Your help is greatly appreciated.

PR #11201, opened 2020-03-03
Allow using `preset-env` with newer versions of `compat-data`

nicolo-ribaudo:
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Currently, adding a new plugin to `@babel/compat-data` breaks old `@babel/preset-env` versions. This is because `preset-env` iterates over `compat-data`'s plugins, and throws if the plugin isn't defined in `preset-env`'s `available-plugins.js` file.

This was found in https://github.com/babel/babel/pull/10971#pullrequestreview-364543352

Instead of removing the error (which is useful to keep the two lists in sync), I propose to set `compat-data`'s version using `~`. This is enough to solve the problem, because adding new plugins to `compat-data` is a new feature, requiring a minor version.

This PR should be merged before 7.9.0: 7.8.0 was the first version with `@babel/compat-data`, and we should fix it in 7.8.x so that people using `"@babel/preset-env": "~7.8.0"` don't get hit by this bug.

### DIFF
--- a/packages/babel-preset-env/package.json
+++ b/packages/babel-preset-env/package.json
@@ -11,7 +11,7 @@
   "repository": "https://github.com/babel/babel/tree/master/packages/babel-preset-env",
   "main": "lib/index.js",
   "dependencies": {
-    "@babel/compat-data": "^7.8.6",
+    "@babel/compat-data": "~7.8.6",
     "@babel/helper-compilation-targets": "^7.8.6",
     "@babel/helper-module-imports": "^7.8.3",
     "@babel/helper-plugin-utils": "^7.8.3",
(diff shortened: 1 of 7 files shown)


packages/babel-preset-env/package.json
```diff
@@ -11,7 +11,7 @@
   "repository": "https://github.com/babel/babel/tree/master/packages/babel-preset-env",
   "main": "lib/index.js",
   "dependencies": {
-    "@babel/compat-data": "^7.8.6",
+    "@babel/compat-data": "~7.8.6",
```
JLHwung: `@babel/compat-data` is also depended by `@babel/helper-compilation-targets`. By doing so we risk having two versions of `@babel/compat-data` if one installs preset-env + helper-compilation-targets 7.8.0 while compat-data 7.9.0 has been published.

Can we fix this on the preset-env side? So it can work with higher `compat-data` as long as we would not drop entries in minor releases. We can introduce an extra unit test as a sync reminder.

PR #11208, opened 2020-03-04
Rephrase parser error message

JLHwung:

Rephrase some parser error messages and resolve todos introduced in #11192 . PTAL and comment if you have any ideas. 🙏

(Yes we will do todos)


JLHwung:
The codecov error is a false negative. This time codecov returns different HTTP code when a repository token secret is not available.

This one returns 400 (expected) https://github.com/babel/babel/pull/11201/checks?check_run_id=483631598#step:6:37

This one returns some code other than 400 
https://github.com/babel/babel/runs/486123611?check_suite_focus=true#step:6:37

I will add an empty commit tomorrow to see if it is fixed.

packages/babel-parser/src/parser/location.js
```diff
@@ -165,18 +163,16 @@ export const Errors = Object.freeze({
...
-    "You can only use decorators on an export when exporting a class",
+    "A decorated export must export a class declaration",
   UnsupportedDefaultExport:
     "Only expressions, functions or classes are allowed as the `default` export.",
   UnsupportedImport: "import can only be used in import() or import.meta",
   UnsupportedMetaProperty: "The only valid meta property for %0 is %0.%1",
-  //todo: remove Stage 2 as we are likely to forget updating when it progressed
   UnsupportedParameterDecorator:
-    "Stage 2 decorators cannot be used to decorate parameters",
+    "Decorators cannot be used to decorate parameters",
```
nicolo-ribaudo: Actually, `legacy` decorators can be used to decorate parameters.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -1033,11 +1033,7 @@ export default class Tokenizer extends LocationParser {
       this.input.charCodeAt(start) === charCodes.digit0;
     if (octal) {
       if (this.state.strict) {
-        this.raise(
-          start,
-          // todo: merge with Errors.StrictOctalLiteral
-          "Legacy octal literals are not allowed in strict mode",
-        );
+        this.raise(start, Errors.StrictOctalLiteral);
```
nicolo-ribaudo: Here we might want to change the `StrictOctalLiteral` message, because `0o1` is perfectly valid in strict mode.

packages/babel-parser/src/parser/location.js
```diff
@@ -35,8 +35,7 @@ export const Errors = Object.freeze({
   ConstructorClassField: "Classes may not have a field named 'constructor'",
   ConstructorClassPrivateField:
     "Classes may not have a private field named '#constructor'",
-  // todo: rephrase to get/set accessor
-  ConstructorIsAccessor: "Constructor can't have get/set modifier",
+  ConstructorIsAccessor: "Constructor can't have get/set accessor",
```
nicolo-ribaudo:
```suggestion
  ConstructorIsAccessor: "Constructor can't be a get/set accessor",
```

existentialism: Howabout

```suggestion
  ConstructorIsAccessor: "Constructor cannot have a get/set accessor",
```

nicolo-ribaudo: In
```js
class A {
  get foo() {}
}
```
I think that `foo` _is_ the accessor, not that it _has_ the accessor.

packages/babel-parser/src/parser/location.js
JLHwung: I think it is okay because this error is not visible to users if they are using `legacy` decorators. Like we did in "Legacy octal literals are not allowed in strict mode", if an error is thrown on a legacy elements, we should prefix its name. Otherwise we are referring to the latest proposal.

packages/babel-parser/src/tokenizer/index.js
JLHwung: Yeah, `0o1` is octal literal but not _legacy_ octal literals. I have changed `StrictOctalLiteral` to be "Legacy octal literals are not allowed in strict mode".

nicolo-ribaudo:
@kaicataldo FYI, this probably conflicts with your PR.

packages/babel-parser/src/parser/location.js
```diff
@@ -35,8 +35,7 @@ export const Errors = Object.freeze({
   ConstructorClassField: "Classes may not have a field named 'constructor'",
   ConstructorClassPrivateField:
     "Classes may not have a private field named '#constructor'",
-  // todo: rephrase to get/set accessor
-  ConstructorIsAccessor: "Constructor can't have get/set modifier",
+  ConstructorIsAccessor: "Constructor may not be a get/set accessor",
```
existentialism: Howabout:

```suggestion
  ConstructorIsAccessor: "Class constructor may not be an accessor",
```

PR #11217, opened 2020-03-05
Create Npm_error.md issue template

nicolo-ribaudo:

Every time we release a new version we get bug reports about this, but it's not something that we can fix.


nicolo-ribaudo:
I mean, it could happen that those reports are valid (it happened for `@babel/parser` a few months ago).

PR #6320, opened 2017-09-26
Add option to indicate when files are being processed

jmunson:
| Q                        | A 
| ------------------------ | ---|
| Fixed Issues             | None|
| Patch: Bug Fix?          | No|
| Major: Breaking Change?  |No| 
| Minor: New Feature?      | Yes|
| Tests Added/Pass?        | Existing tests still pass, did not add new tests to cover this options output|
| Spec Compliancy?         | doesn't change anything spec related|
| License                  | MIT|
| Doc PR                   | No|
| Any Dependency Changes?  | No|

I added the flag -n or --noisy to add output when files are being processed in watch mode.
This should make it easier to integrate with external debugging tools like [VSCode](https://code.visualstudio.com/docs/editor/tasks#_background-watching-tasks) that can be configured to wait for a background watching task to finish processing the files before it attempts to run or debug them.

PR #11220, opened 2020-03-06
Log after subsequent compilations in --watch mode

### LINKED ISSUE #7926: No babel-cli watch mode console output after first compilation
## Bug Report

**Current Behavior**

The babel-cli in `--watch` mode does not output any messages following the first compilation. It seems as though the CLI is not doing anything even though files are compiling.

**Input Code**

```sh
babel src -d lib --watch
```

**Expected behavior/code**

The CLI should display an output summary after every `--watch` mode compilation.

N/A.

**Environment**
- Babel version(s): v7.0.0-beta.41 - v7.0.0-beta.46
- OS: macOS v10.13.4
- Monorepo: No

**Possible Solution**

This issue began with https://github.com/babel/babel/pull/7439 in [v7.0.0-beta.41](https://github.com/babel/babel/releases/tag/v7.0.0-beta.41). The babel-cli has a new a new `--verbose` flag for seeing exactly what files were compiled, but even with the verbose output of what files are being compiled there are no subsequent summary messages. An output summary should be displayed every compilation, regardless of the `--verbose` flag.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #7926, Closes #6320
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is a revision of #6320.

1. I used the same message that we already use for the first compilation
1. I removed the `--noisy` option. We already have `--quiet` and `--verbose`, and a third option would be too much (unless we implement something like `--loglevel`). I made it the default behavior since it just replicates what `@babel/cli` already does.
1. I used `lodash/debounce` instead of a custom implementation.


packages/babel-cli/src/babel/dir.js
```diff
@@ -172,16 +186,29 @@ export default async function({
...
+
+          try {
+            await handleFile(
+              filename,
+              filename === filenameOrDir
+                ? path.dirname(filenameOrDir)
+                : filenameOrDir,
+            );
+
+            compiledFiles++;
```
JLHwung: Will `compiledFiles` be incremented twice if a single file is modified twice?

nicolo-ribaudo: With the current implementation, if it is modified twice in a very short period of time, yes.

Do you think that the memory/complexity of using a `Set` with the file paths instead of the simple counter? (This looks like a rhetorical question, but it's not :sweat_smile:)

JLHwung: On devtools like Babel I prefer correctness over performance unless it becomes our bottleneck.

On the other hand, thinking of the DX, does `Successfully compiled # files with babel` really deliver value? I mean as a developer I would not care how many files are compiled, especially when in watch mode. In Babel repo we are printing the compiled path, execution timestamp in watch mode. If one is changing a single file using babel-cli, it will keep printing `successfully compiled 1 file with Babel` so babel-cli can not assure users that the file _has_ been compiled after they saved.

What about printing things like "successfully compiled after 346ms", not ideal but at least they are more likely to differ across different runs.

I am okay to leave it as-is and address this in a separate DX PR.

existentialism: Yeah I'm up for adjusting the message too, in a follow-up.

nicolo-ribaudo: Ok :+1:

nicolo-ribaudo: It now logs
```
Successfully compiled 3 files with Babel (123ms).
```

packages/babel-cli/src/babel/dir.js
```diff
@@ -138,24 +139,50 @@ export default async function({
...
+      }
+
+      const diff = process.hrtime(startTime);
+
+      console.log(
+        `Successfully compiled ${compiledFiles} ${
+          compiledFiles !== 1 ? "files" : "file"
+        } with Babel (${diff[0] * 1e3 + Math.round(diff[1] / 1e6)}ms).`,
+      );
+      compiledFiles = 0;
```
JLHwung, line 160: I think we should also reset `compiledFiles` if any of
- `cliOptions.quiet` is `true`.
- error is thrown in `await handle(filename)`.

nicolo-ribaudo: > cliOptions.quiet is true.

- Why? If `quiet` is true, we never read this variable so it doesn't really matter.

> error is thrown in await handle(filename).

- In this case, the `compiledFiles` variable would just be thrown away because it's defined in the same scope as the one containing `await handle(filename)`.

JLHwung: If `babel` is running on watch mode, any errors in the first build will not reset `compiledFiles` nor halt the execution. The `compiledFiles` will be accumulated until `logSuccess()` is called. It may confuse users if they revise a single file to fix the syntax error and `babel` prints `Successfully compiled 2366 files`.

JLHwung: Oh my bad! I thought `handle(filename)` will be returned for some unknown mental process. You are right!

PR #11248, opened 2020-03-12
Handle private access chained on an optional chain

jridgewell:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

See the resolution from https://github.com/tc39/proposal-class-fields/pull/301.

We intend to allow devs to access private properties in the optional chains, eg `foo?.bar.#baz`. We're still debating whether to allow the private property to _start_ the optional chain, eg `foo?.#baz`.

--- Edits By @JLHwung ---

Todo list tracking `?. PrivateIdentifier` support
- [x] `foo?.#x`
- [x] `foo?.#m()`
- [x] `foo.#m?.()`
- [x] `foo?.#m?.()`
- [x] add test cases for loose mode


packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -241,18 +246,28 @@ const privateNameHandlerSpec = {
 };
 
 const privateNameHandlerLoose = {
-  handle(member) {
+  get(member) {
```
nicolo-ribaudo: Out of curiosity, is this small refactoring needed to implement this feature?

(I like the new code, I'm not asking to remove it)

packages/babel-plugin-proposal-class-properties/test/fixtures/private/optional-chain-with-transform/output.js
```diff
@@ -0,0 +1,43 @@
...
+    (_ref14 = (_ref24 = _ref13 = (_ref12 = (_o14 = _o6 = o) === null || _o14 === void 0 ? void 0 : _o14.Foo, _o6 === void 0 || _o6 === null) ? void 0 : _classStaticPrivateFieldSpecGet(_ref12, Foo, _self)) === null || _ref24 === void 0 ? void 0 : _ref24.self, _ref13 === void 0 || _ref13 === null) ? void 0 : _classStaticPrivateFieldSpecGet(_ref14, Foo, _x);
+    (_ref17 = (_ref25 = _ref16 = (_ref15 = (_o15 = _o7 = o) === null || _o15 === void 0 ? void 0 : _o15.Foo, _o7 === void 0 || _o7 === null) ? void 0 : _classStaticPrivateFieldSpecGet(_ref15, Foo, _self).self) === null || _ref25 === void 0 ? void 0 : _ref25.self, _ref16 === void 0 || _ref16 === null) ? void 0 : _classStaticPrivateFieldSpecGet(_ref17, Foo, _x);
+    (_ref20 = (_ref26 = _ref19 = (_ref27 = (_ref18 = (_o16 = _o8 = o) === null || _o16 === void 0 ? void 0 : _o16.Foo, _o8 === void 0 || _o8 === null) ? void 0 : _classStaticPrivateFieldSpecGet(_ref18, Foo, _self)) === null || _ref27 === void 0 ? void 0 : _ref27.self) === null || _ref26 === void 0 ? void 0 : _ref26.self, _ref19 === void 0 || _ref19 === null) ? void 0 : _classStaticPrivateFieldSpecGet(_ref20, Foo, _x);
```
nicolo-ribaudo: Wow, this is a mess :joy:

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -37,9 +37,113 @@ const handle = {
   handle(member) {
     const { node, parent, parentPath } = member;
 
+    if (member.isOptionalMemberExpression()) {
+      const root = member.find(({ node, parent, parentPath }) => {
+        if (parentPath.isOptionalMemberExpression()) {
+          return parent.optional || parent.object !== node;
```
nicolo-ribaudo: When is `parent.object !== node`? (Please, write a comment in the code)

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -37,9 +37,113 @@ const handle = {
...
+
+      const rootParentPath = root.parentPath;
+      if (
+        rootParentPath.isUpdateExpression({ argument: node }) ||
+        rootParentPath.isAssignmentExpression({ left: node })
+      ) {
+        throw member.buildCodeFrameError(`can't handle assignment`);
+      }
+      if (rootParentPath.isUnaryExpression({ operator: "delete" })) {
+        throw member.buildCodeFrameError(`can't handle delete`);
```
nicolo-ribaudo: Those two errors are also handled by the parser, right?

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -37,9 +37,113 @@ const handle = {
...
+
+      let nearestOptional = member;
+      for (;;) {
+        if (nearestOptional.isOptionalMemberExpression()) {
+          if (nearestOptional.node.optional) break;
+          nearestOptional = nearestOptional.get("object");
+          continue;
+        } else if (nearestOptional.isOptionalCallExpression()) {
+          if (nearestOptional.node.optional) break;
+          nearestOptional = nearestOptional.get("object");
```
nicolo-ribaudo: Shouldn't this be `callee`?

packages/babel-plugin-proposal-class-properties/test/fixtures/private/optional-chain/output.js
```diff
@@ -0,0 +1,43 @@
...
+
+    const o = {
+      Foo: Foo
+    };
+    const deep = {
+      very: {
+        o
+      }
+    };
+    (_ref = (_o = o)?.Foo, _o === void 0 || _o === null) ? void 0 : _classStaticPrivateFieldSpecGet(_ref, Foo, _x);
```
nicolo-ribaudo: Can we produce this simpler output for `o?.Foo.#x`?

```js
(_o = o) ?? _classStaticPrivateFieldSpecGet(_o.Foo, Foo, _x);
```

packages/babel-helper-create-class-features-plugin/src/fields.js
jridgewell: Yes. The old code didn't actually use `memberExpressionToFunctions` (because it was unnecessary), and the optional behavior is implemented inside `memberExpressionToFunctions`. These changes are what's necessary for that API.

packages/babel-helper-member-expression-to-functions/src/index.js
jridgewell: Done.

packages/babel-helper-member-expression-to-functions/src/index.js
jridgewell: `delete foo.#bar` is handled by parser, but it's technically legal for `delete super.foo` which is also handled by this. I had written member-expression-to-functions to be generic enough to use on any kind of AST, so I added the error cases here.

I also plan to propose `foo?.bar = 1` eventually, it's just not included the original optional chaining proposal. So eventually, we'd need to add this.

packages/babel-helper-member-expression-to-functions/src/index.js
jridgewell: Yup, thanks! Looks like I need more tests.

packages/babel-plugin-proposal-class-properties/test/fixtures/private/optional-chain-with-transform/output.js
jridgewell: I know! It took me forever to parse this to verify.

I have a pending PR that adds new AST node support to `generateUidBasedOnNode` that helps a little bit, but it's still terrible.

packages/babel-plugin-proposal-class-properties/test/fixtures/private/optional-chain/output.js
jridgewell: Unfortunately not, since that would only execute `_classStaticPrivateFieldSpecGet`'s branch when `o` was nullish! Then `_o.Foo` would be guaranteed to throw.

We'd need an inverse operator, something like "[not-nullish and](https://github.com/tc39/proposal-nullish-coalescing/issues/61)".

nicolo-ribaudo: Oh right :man_facepalming: 

I think that we can still simplify it a bit, by avoiding the double nullish check (the manual one and the one in `?.`) and a temporary variable:
```js
(_o = o) === null || _o === void 0 ? void 0 : _classStaticPrivateFieldSpecGet(_o.Foo, Foo, _x)
```

(Please put the `null` check first, to that `=== void 0 ? void 0 :` matches the existing optional chaining transform and can be gzipped more)

jridgewell: Yah... why didn't I think of that?

packages/babel-parser/src/parser/expression.js
```diff
@@ -610,15 +610,16 @@ export default class ExpressionParser extends LValParser {
...
-        ? this.parseIdentifier(true)
         : this.parseMaybePrivateName(true);
       node.computed = computed;
 
       if (node.property.type === "PrivateName") {
         if (node.object.type === "Super") {
           this.raise(startPos, Errors.SuperPrivateField);
         }
+        if (optional) {
+          this.raise(node.property.start, Errors.OptionalChainingNoPrivate);
```
JLHwung, line 621: Can you update this PR as long as `?. PrivateIdentifier` has been accepted?

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -37,9 +86,153 @@ const handle = {
...
+        if (startingOptional.isOptionalMemberExpression()) {
+          if (startingOptional.node.optional) break;
+          startingOptional = startingOptional.get("object");
+          continue;
+        } else if (startingOptional.isOptionalCallExpression()) {
+          if (startingOptional.node.optional) break;
+          startingOptional = startingOptional.get("callee");
+          continue;
+        }
+      }
```
nicolo-ribaudo, line 153: Can we add a `throw new Error("Internal error")` at the end of this loop's body? It's better to throw than to accidentally have an infinite loop.

JLHwung:
I am converting this PR to draft as transforming `?. PrivateIdentifier` are still WIP.

nicolo-ribaudo:
Whops I clicked "Approve" on the wrong PR

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -37,9 +86,171 @@ const handle = {
...
+      );
+
+      if (context) {
+        const endParent = endParentPath.node;
+        endParentPath.replaceWith(
+          t.optionalCallExpression(
+            t.optionalMemberExpression(
+              endParent.callee,
+              t.identifier("call"),
+              endParent.computed,
```
JLHwung, line 234: It should be always `false` because we are transforming `foo.bar(qux)` to `foo.bar.call(foo.bar, qux)`

JLHwung:
✅On changes by Justin. But it will need approval other than myself for my changes.

Thanks for @nicolo-ribaudo who inspires me on the `#m?.()` support!

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/optional-chain-before-member-call-with-transform/output.js
```diff
@@ -32,41 +32,41 @@ class Foo {
...
+    (_call2 = (_ref20 = fn === null || fn === void 0 ? void 0 : (_classPrivateFieldLoo4 = _classPrivateFieldLooseBase(fn().Foo, _self)[_self]).getSelf) == null ? void 0 : _ref20.call(_classPrivateFieldLoo4)) === null || _call2 === void 0 ? void 0 : _classPrivateFieldLooseBase(_call2.self, _m)[_m]();
+    (_getSelf5 = (_ref21 = fn === null || fn === void 0 ? void 0 : _classPrivateFieldLooseBase(fn().Foo, _self)[_self]) == null ? void 0 : _ref21.getSelf()) === null || _getSelf5 === void 0 ? void 0 : _classPrivateFieldLooseBase(_getSelf5.self, _m)[_m]();
+    (_getSelf6 = (_ref22 = fn === null || fn === void 0 ? void 0 : _classPrivateFieldLooseBase(fn().Foo, _self)[_self]) == null ? void 0 : _ref22.getSelf == null ? void 0 : _ref22.getSelf()) === null || _getSelf6 === void 0 ? void 0 : _classPrivateFieldLooseBase(_getSelf6.self, _m)[_m]();
```
nicolo-ribaudo, line 69: In the future we'll figure out how to use `== null` here.

JLHwung: We could add `loose` support to https://github.com/babel/babel/blob/b457f52ab50945138d53af86766c0309581b42ba/packages/babel-helper-member-expression-to-functions/src/index.js#L143

packages/babel-helper-member-expression-to-functions/src/index.js
JLHwung: The current implementation also disallows `foo?.#bar.qux`, which is allowed in the spec. That said I am happy to fix this issue in patch releases before it gets too big.

PR #3608, opened 2016-07-27
Set sourceRoot in babel-register transform to fix paths in stacks

danez:
This is the followup of #3523 which was reverted. 

This change fixes the stacktraces by setting `sourceRoot` in the options. This does not touch the `sources` or `file` property in the map, but adds a property called `sourceRoot`. By spec the sourceRoot is prepended to `sources` and `file`.

> Resolving Sources
> If the sources are not absolute URLs after prepending of the “sourceRoot”, the sources are resolved relative to the SourceMap (like resolving script src in a html document). 

Printing absolute paths is also more in line with node itself, which prints absolute paths in its traces.

I tested the previously [failing project](https://github.com/aureooms/es-itertools) with this change and it worked.

i also thought about tests, but the main blocker right now is that there is no way to unregister babel-register which messes then with other tests. As soon as we merge #3139 it might get possible to write tests and after the test to properly unhook.

PR #11249, opened 2020-03-12
Add path separator to `@babel/register` sourceRoot

andrewdotn:
| Documentation PR Link    | 

The [Source Maps R3 standard][] states that `sourceRoot` is “prepended to
the individual entries in the ‘source’ field.” If `sources` contains file
names, and `sourceRoot` is intended to refer to a directory but doesn’t end
with a trailing slash, any consumers of the source map are in for a bad
day.

[Source Maps R3 standard]: https://sourcemaps.info/spec.html

In particular, with the new `NODE_V8_COVERAGE` feature built-in to node,
using `@babel/register` results in bogus coverage data, with
`…/srcfoo.js` coming out where `…/src/foo.js` was intended.

Thanks to @bcoe for writing up [Source maps in Node.js][], it was a big
help in tracking down what was going wrong.

[Source maps in Node.js]: https://medium.com/@nodejs/source-maps-in-node-js-482872b56116

References: https://github.com/bcoe/c8/issues/161, https://github.com/babel/babel/pull/3608

### DIFF
--- a/packages/babel-register/src/node.js
+++ b/packages/babel-register/src/node.js
@@ -41,7 +41,7 @@ function compile(code, filename) {
   const opts = new OptionManager().init(
     // sourceRoot can be overwritten
     {
-      sourceRoot: path.dirname(filename),
+      sourceRoot: path.dirname(filename) + path.sep,
       ...deepClone(transformOpts),
       filename,
     },
--- a/packages/babel-register/test/fixtures/source-map/foo/bar.js
+++ b/packages/babel-register/test/fixtures/source-map/foo/bar.js
@@ -1,0 +1,3 @@
+function add(a, b) {
+  return a + b;
+}
--- a/packages/babel-register/test/fixtures/source-map/index.js
+++ b/packages/babel-register/test/fixtures/source-map/index.js
@@ -1,0 +1,7 @@
+const { retrieveSourceMap } = require('source-map-support');
+
+const path = require.resolve('./foo/bar');
+
+require('./foo/bar');
+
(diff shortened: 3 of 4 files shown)


packages/babel-register/src/node.js
```diff
@@ -41,7 +41,7 @@ function compile(code, filename) {
   const opts = new OptionManager().init(
     // sourceRoot can be overwritten
     {
-      sourceRoot: path.dirname(filename),
+      sourceRoot: path.dirname(filename) + path.sep,
```
JLHwung: The `sourceRoot` should be a valid URL. (C.f. https://github.com/mozilla/source-map/issues/340#issuecomment-392879514) Ideally we should use `url.filePathToURL` shipped in node.js 10.12 or [file-url](https://github.com/sindresorhus/file-url) to convert `path.dirname` into file URL and then add a trailing slash, because `\` is not valid in URL.

Sidenote: Although the spec says `sourceRoot` is supposed to prepend sources, the `source-map` library support sourceRoot without slash: https://github.com/mozilla/source-map/blob/7a263408de1b60e60b6220ca3e36d9b3e4aa047d/lib/util.js#L413

andrewdotn: Babel is currently outputting `C:\Users\...` for `sourceRoot` on windows, and `path.sep` is consistent with that.

Changing `sourceRoot` to be a file URL isn’t something that I have time to code and test right now.

I hope that won’t block the merging of this bug fix.

PR #11250, opened 2020-03-12
Add stricter Optional Chain node validation

jridgewell:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  | yes
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Optional chains cannot start with `optional: false`. Somewhere in the `object`/`callee` tree, there must be an optional with `optional: true`.

### DIFF
--- a/packages/babel-types/src/definitions/utils.js
+++ b/packages/babel-types/src/definitions/utils.js
@@ -188,7 +188,8 @@ export function assertShape(shape: { [string]: FieldOptions }): Validator {
 
 export function assertOptionalChainStart(): Validator {
   function validate(node) {
-    for (let current = node; current; ) {
+    let current = node;
+    while (node) {
       const { type } = current;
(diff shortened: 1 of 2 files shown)


packages/babel-types/src/definitions/utils.js
```diff
@@ -186,6 +186,33 @@ export function assertShape(shape: { [string]: FieldOptions }): Validator {
...
+        if (current.optional) return;
+        current = current.object;
+        continue;
+      }
+
+      break;
+    }
+
+    throw new TypeError(
+      `Non-optional ${node.type} must chain from an optional OptionalMemberExpression or OptionalCallExpression`,
```
JLHwung: Can we print `current.type` when `current.optional` is falsy?

jridgewell: Done.

PR #11257, opened 2020-03-13
fix: remove bindings when removing a node

### LINKED ISSUE #11256: `Cannot read property 'type' of null` again
## Bug Report

**Current Behavior**
This is a follow up to #11210 . When the assignment expression is wrapped in a block scope that is compiled to IIFE, the error occurs again.

```
TypeError: unknown: Cannot read property 'type' of null
    at convertFunctionParams (/path/to/node_modules/@babel/plugin-transform-parameters/lib/params.js:88:24)
    ...
```

**Input Code**

```js
class A {
  a = b => {
    {
      ({ b } = {});
    }
  };
}
```

```json
// options.json
{
  "plugins": [
    "proposal-class-properties",
    "transform-parameters",
    "transform-destructuring"
  ]
}
```

**Expected behavior/code**
Should not throw error.

**Environment**

```
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 13.10.1 - /path/to/node_modules/.bin/node
    Yarn: 1.22.0 - /usr/local/bin/yarn
    npm: 6.13.7 - /path/to/node_modules/.bin/npm
  npmPackages:
    @babel/runtime: ^7.1.5 => 7.8.4
    babel-plugin-macros: ^2.8.0 => 2.8.0
    eslint: ^6.8.0 => 6.8.0
```
- Babel version(s): all latest, @babel/core@7.8.8
- Node/npm version: Node 13/npm 6
- Monorepo: no

**Possible Solution**

Possible cause: In `replaceWithMultiple` the original node is removed, but its bindings are not removed as expected, leaving a node with `removed: true, node: null` in the `constantViolations` and breaks `plugins-transform-parameters`.

The bindings are found from `path.node`:

https://github.com/babel/babel/blob/420f2ee69aad0316c6f49e9dfdc6a02c8f9d4fc8/packages/babel-traverse/src/path/removal.js#L24-L27

But `path.node` is always `null` when it is being removed:

https://github.com/babel/babel/blob/420f2ee69aad0316c6f49e9dfdc6a02c8f9d4fc8/packages/babel-traverse/src/path/replacement.js#L55-L59

I found a tricky way to solve this:
- save the node to be removed before it is set to `null`
- pass it to `remove`
- search bindings: first `this.node = node`, then search, finally `this.node = null`

gera2ld:
| Fixed Issues?            | Fixes #11256
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-plugin-transform-parameters/test/fixtures/regression/11256/input.mjs
+++ b/packages/babel-plugin-transform-parameters/test/fixtures/regression/11256/input.mjs
@@ -1,0 +1,7 @@
+export default class A {
+  a = b => {
+    {
+      ({ b } = {});
+    }
+  }
+}
--- a/packages/babel-plugin-transform-parameters/test/fixtures/regression/11256/options.json
+++ b/packages/babel-plugin-transform-parameters/test/fixtures/regression/11256/options.json
@@ -1,0 +1,7 @@
+{
+  "plugins": [
+    "proposal-class-properties",
+    "transform-parameters",
+    "transform-destructuring"
+  ]
+}
--- a/packages/babel-plugin-transform-parameters/test/fixtures/regression/11256/output.mjs
+++ b/packages/babel-plugin-transform-parameters/test/fixtures/regression/11256/output.mjs
@@ -1,0 +1,13 @@
+function _defineProperty(obj, key, value) { if (key in obj) { Object.defineProperty(obj, key, { value: value, enumerable: true, configurable: true, writable: true }); } else { obj[key] = value; } return obj; }
+
+export default class A {
+  constructor() {
+    _defineProperty(this, "a", b => {
+      {
+        var _ref = {};
+        b = _ref.b;
+      }
+    });
+  }
+
+}
--- a/packages/babel-traverse/src/path/replacement.js
+++ b/packages/babel-traverse/src/path/replacement.js
@@ -49,6 +49,11 @@ export function replaceWithMultiple(nodes: Array<Object>) {
   nodes = this._verifyNodeList(nodes);
   t.inheritLeadingComments(nodes[0], this.node);
   t.inheritTrailingComments(nodes[nodes.length - 1], this.node);
+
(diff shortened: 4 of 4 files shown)


packages/babel-traverse/src/path/removal.js
```diff
@@ -2,13 +2,14 @@
 
 import { hooks } from "./lib/removal-hooks";
 import { REMOVED, SHOULD_SKIP } from "./index";
+import * as t from "@babel/types";
 
-export function remove() {
+export function remove(node?: Object) {
```
nicolo-ribaudo: Is it possible to fix this without modifying the public API? Maybe we could remove the bindings in `replaceWithMultiple` before setting `this.node` to `null`.

Alternatively, we could leave `@babel/traverse` as-is (this PR doesn't fix the problem anyway, for example when removing `if (foo) var bar;`), and check if the binding exists when using it.

gera2ld: > check if the binding exists when using it

Does it mean it's expected that `this._removeFromScope()` does not remove all bindings, and all plugins should be aware of that?

Then I'll add a check for `node.removed` in `plugin-transform-paramters`.

gera2ld:
For some reason `node.removed` is not updated correctly in `constantViolations`, so I tried to remove the bindings before `this.remove()`.

packages/babel-traverse/src/path/removal.js
nicolo-ribaudo: Yes, `_removeFromScope` is only expected to remove bindings of the current declaration (`this.node`).

Maybe we could fix it by calling `this._removeFromScope()` in `path.replaceWithMultiple`, before setting `this._node` to `null`?

> all plugins should be aware of that?

Unfortunately yes :man_shrugging: :slightly_frowning_face: 
Until we manage to fully automate our scope handling, plugins should be aware that it could not be in sync with the AST.

PR #11261, opened 2020-03-14
Memoize call expressions in optional chains in loose mode

### LINKED ISSUE #11084: [plugin-proposal-optional-chaining] Loose mode produces unoptimized code for class methods
## Bug Report

**Current Behavior**
The `loose` mode of `optional-chaining` plugin compiles class methods in non-optimized way repeating the call to the method instead of remembering its result.

It does not happen for function call though.

**Input Code**
- REPL or Repo link if applicable: [Runkit](https://runkit.com/lodin/optional-chaining-repeating-call-to-object-method).

```js
a.get(b)?.();
```
compiles to:
```js
a.get(b) == null ? void 0 : a.get(b)();
```

**Expected behavior/code**
The result of method call should be memoized as it happens in the strict mode.

```js
var a$get;

(a$get = a.get(b)) == null ? void 0 : a$get();
```

- Filename: `babel.config.js`

```js
{
    plugins: [[require('@babel/plugin-proposal-optional-chaining'), {loose: true}]]
}
```

**Environment**

- Babel version(s): 7.8.4
- Node/npm version: v13.5.0/6.13.4
- OS: Windows 10
- Monorepo: Lerna
- How you are using Babel: cli/programmatically

oliverdunk:
| Fixed Issues?            | Fixes #11084  
| Documentation PR Link    | 

Previously, in loose mode, we weren't memoizing optional call expressions. This resulted in code like `a.get(b)?.();` being compiled to `a.get(b) == null ? void 0 : a.get(b)();`, which repeats a function call.

This MR changes the behaviour so that memoization is only skipped when no work is done by the call's callee.

### DIFF
--- a/packages/babel-plugin-proposal-optional-chaining/src/index.js
+++ b/packages/babel-plugin-proposal-optional-chaining/src/index.js
@@ -83,8 +83,8 @@ export default declare((api, options) => {
 
           // Ensure call expressions have the proper `this`
           // `foo.bar()` has context `foo`.
-          if (isCall && isSimpleMemberExpression(chain)) {
-            if (loose) {
+          if (isCall) {
+            if (loose && isSimpleMemberExpression(chain)) {
               // To avoid a Function#call, we can instead re-grab the property from the context object.
               // `a.?b.?()` translates roughly to `_a.b != null && _a.b()`
               node.callee = chain;
--- a/packages/babel-plugin-proposal-optional-chaining/src/index.js
+++ b/packages/babel-plugin-proposal-optional-chaining/src/index.js
@@ -7,6 +7,15 @@ export default declare((api, options) => {
 
   const { loose = false } = options;
 
+  function isSimpleMemberExpression(expression) {
+    return (
+      t.isIdentifier(expression) ||
+      (t.isMemberExpression(expression) &&
+        !expression.computed &&
+        isSimpleMemberExpression(expression.object))
+    );
+  }
+
   return {
     name: "proposal-optional-chaining",
     inherits: syntaxOptionalChaining,
@@ -50,9 +59,10 @@ export default declare((api, options) => {
 
           let ref;
           let check;
-          if (loose && isCall) {
+          if (loose && isCall && isSimpleMemberExpression(node.callee)) {
             // If we are using a loose transform (avoiding a Function#call) and we are at the call,
-            // we can avoid a needless memoize.
+            // we can avoid a needless memoize. We only do this if the callee is a simple member
+            // expression, to avoid multiple calls to nested call expressions.
             check = ref = chain;
           } else {
             ref = scope.maybeGenerateMemoised(chain);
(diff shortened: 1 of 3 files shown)


packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -50,9 +59,10 @@ export default declare((api, options) => {
 
           let ref;
           let check;
-          if (loose && isCall) {
+          if (loose && isCall && isSimpleMemberExpression(node.callee)) {
```
jridgewell: I think this can be as simple as `!t.isCallExpression(chain)` and still get the same behavior.

Note, `chain` is just a local variable for `node.callee`.

jridgewell: Nit:
```suggestion
          if (loose && isCall && isSimpleMemberExpression(chain)) {
```

oliverdunk: Easy change, done!

oliverdunk:
@jridgewell Please ignore that request for review. I need to make sure we're probably setting the call context.

oliverdunk:
@jridgewell Found and fixed my bug! It would be great if you could take a look :)

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -72,12 +83,12 @@ export default declare((api, options) => {
 
           // Ensure call expressions have the proper `this`
           // `foo.bar()` has context `foo`.
-          if (isCall && t.isMemberExpression(chain)) {
```
jridgewell: We'll still need this.

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -72,12 +83,12 @@ export default declare((api, options) => {
...
-            if (loose) {
+          if (isCall) {
+            if (loose && isSimpleMemberExpression(chain)) {
               // To avoid a Function#call, we can instead re-grab the property from the context object.
               // `a.?b.?()` translates roughly to `_a.b != null && _a.b()`
               node.callee = chain;
-            } else {
+            } else if (chain.object) {
```
jridgewell: With the above change, we'll be able to revert this.

packages/babel-plugin-proposal-optional-chaining/src/index.js
oliverdunk: Ah, I was wondering why the chain.object wasn't needed before. Do you think isMemberExpression is a bit too specific? Shouldn't we allow optional member expressions too?

jridgewell: Any optional members will have been converted to regular members by https://github.com/babel/babel/blob/d4d6ec8f2b8ebfb32381a1e8dcadbadec3ae5736/packages/babel-plugin-proposal-optional-chaining/src/index.js#L41

oliverdunk:
@jridgewell I've addressed your feedback again. Really appreciate all of the help! It's early days when it comes to my understanding of Babel - but you and Nico have been incredibly welcoming.

nicolo-ribaudo:
Thanks Oliver!

PR #11262, opened 2020-03-14
Extract for-of iterator handling to a helper

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  |

This greatly reduces the code size (*) when for-of is used multiple times across the
codebase. Also, makes it easier to read what's happening in a compiled loop.

(*) I tested the difference with 3 for loops _before_ adding the human-friendly error message about the missing `Symbol.iterator`. I didn't use `transform-runtime` or `external-helpers`, so these numbers include the helper size.

These are the minified sizes (with terser):
| | spec | loose |
|-|--------|---------|
| **old** | 563 bytes | 465 bytes |
| **new** | 457 bytes | 349 bytes |

Since the error message is a bit more than 100 bytes long, the size improvement probably needs 4 loops rather than one now. However, keep in mind that if I added the human-friendly error message to the old implementation it would have added 100 bytes 3 times, making the size improvement even bigger.

<details>
<summary>Size test code</summary>

```javascript
loop: for (let x of a) {
  break loop;
}
for (var x of a) {
  a = null;
}
for (x of a) {
  continue;
}
```

</details>

Part 1/n


packages/babel-plugin-transform-for-of/src/no-helper-implementation.js
```diff
@@ -0,0 +1,192 @@
+import { template, types as t } from "@babel/core";
+
+// This is the legacy implementation, which inlines all the code.
+// It must be kept for compatibility reasons.
+// TODO (Babel 8): Remove this code.
+
+export default function transformWithoutHelper(loose, path, state) {
```
nicolo-ribaudo: This file contains the old plugin code needed when the helper isn't available. I just copy-pasted it without any edit.

packages/babel-plugin-transform-block-scoping/test/fixtures/general/hoisting/output.js
```diff
@@ -10,29 +10,19 @@ var _loop = function (i) {
...
-var _iteratorError = undefined;
+var _iteratorHelper = babelHelpers.createForOfIteratorHelper(nums),
+    _step;
 
 try {
-  for (var _iterator = nums[Symbol.iterator](), _step; !(_iteratorNormalCompletion = (_step = _iterator.next()).done); _iteratorNormalCompletion = true) {
+  for (_iteratorHelper.s(); !(_step = _iteratorHelper.n()).done;) {
     var i = _step.value;
     var x;
     var f;
```
nicolo-ribaudo: This is a bug unrelated to this PR: `f` and `x` should be declared in the `_loop` function.

packages/babel-helpers/src/helpers.js
```diff
@@ -1003,6 +1003,61 @@ helpers.nonIterableRest = helper("7.0.0-beta.0")`
...
+          if (didErr) throw err;
+        }
+      }
+    };
+  }
+`;
+
+helpers.createForOfIteratorHelperLoose = helper("7.9.0")`
+  export default function _createForOfIteratorHelperLoose(o) {
+    var i = 0;
```
jridgewell: Nit: Don't share the variable here.

packages/babel-plugin-transform-for-of/src/index.js
```diff
@@ -189,156 +160,74 @@ export default declare((api, options) => {
...
 
-        // add the value declaration to the new loop body
-        if (declar) {
-          block.body.push(declar);
-        }
+        node.body.body.unshift(declar);
+
+        const nodes = builder.build({
+          CREATE_ITERATOR_HELPER: state.addHelper(builder.helper),
+          ITERATOR_HELPER: scope.generateUidIdentifier("iteratorHelper"),
```
jridgewell: Nit: 
```suggestion
          ITERATOR_HELPER: scope.generateUidIdentifier("iterator"),
```

packages/babel-plugin-transform-for-of/src/index.js
```diff
@@ -189,156 +160,74 @@ export default declare((api, options) => {
...
+          STEP_KEY: t.identifier(stepKey),
+          OBJECT: node.right,
+          BODY: node.body,
+        });
+        const container = builder.getContainer(nodes);
 
-        // push the rest of the original loop body onto our new body
-        block.body = block.body.concat(node.body.body);
+        t.inherits(container[0], node);
+        t.inherits(container[0].body, node.body);
```
jridgewell: Does `container` ever contain more than one element? If not, we can replace the `replaceWithMultiple` with `replaceWith` and drop the `parent.isLabeledStatement()` check.

packages/babel-helpers/src/helpers.js
nicolo-ribaudo: I mostly did it for code size, since this is a runtime helper, and `i` could stand both for `index` and `iterator` :sweat_smile:

packages/babel-plugin-transform-for-of/src/index.js
nicolo-ribaudo: `container` always contains a single node. However:
1. We are doing `replaceWithMultiple(nodes)`, and  in non-loose mode `nodes` contains 2 nodes (the var declaration and the try statement)
2. We can't drop the `parent.isLabeledStatement()` check because in non-loose mode it the label must be moved to the inner loop. If we only did `path.replaceWith()` it would be attached to the `try` statement.

packages/babel-plugin-transform-for-of/src/index.js
```diff
@@ -189,156 +160,74 @@ export default declare((api, options) => {
...
+        let declar;
+
+        const stepKey = scope.generateUid("step");
+        const stepValue = t.memberExpression(
+          t.identifier(stepKey),
+          t.identifier("value"),
+        );
+
+        if (
+          t.isIdentifier(left) ||
```
JLHwung: nit: It suffices to check `t.isVariableDeclaration` first and assuming it's `LVal` in the else branch, because we have validated `ForOfStatement.left` even when `BABEL_TYPES_8_BREAKING` is not enabled:

https://github.com/babel/babel/blob/1ba41f208430bdf382b75b56e9aad7436ca522ad/packages/babel-types/src/definitions/es2015.js#L333

packages/babel-plugin-transform-block-scoping/test/fixtures/general/hoisting/output.js
jridgewell: I think they have to stay here. If we moved the `var` inside the `_loop` function, they would no longer declare the variable in the function scope (it's not a block `let`/`const`).

packages/babel-plugin-transform-for-of/src/index.js
jridgewell: >  If we only did path.replaceWith() it would be attached to the try statement.

I think that would actually be equivalent. The `finally` block should still fire correctly.

nicolo-ribaudo: But then, it doesn't work with `continue`.

jridgewell: Ohhh. I was only thinking about `break`.

PR #11263, opened 2020-03-14
Add for-of fallback for arrays in browsers without symbol support

nicolo-ribaudo:
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is similar to the `assumeArray` option: it makes arrays work in browsers without `Symbol` support. It's safe to apply this by default, because for non-arrays it still goes in the `Symbol.iterator` path.

This PR isn't really useful per se, but it's needed for Part 3 of this series.

**Part 2/n**

Depends on:
- [x] Part 1 (https://github.com/babel/babel/pull/11262)


packages/babel-helpers/src/helpers.js
```diff
@@ -1043,8 +1043,24 @@ helpers.createForOfIteratorHelper = helper("7.9.0")`
...
+      if (Array.isArray(o)) {
+        var i = 0;
+        var F = function(){};
+        return {
+          s: F,
+          n: function() {
+            if (i >= o.length) return { done: true };
+            return { done: false, value: o[i++] };
+          },
+          e: F,
```
jridgewell: This `e` method needs to throw the passed in error.

nicolo-ribaudo: Thanks, good catch!

PR #11266, opened 2020-03-15
Add "allowArrayLike" support to the for-of transform

nicolo-ribaudo:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | https://github.com/babel/website/pull/2234
| Any Dependency Changes?  |

When enabled, this option makes Babel treat array-like objects as if
they were iterables. This makes it possible to avoid a `Symbol` polyfill when
using for-of on DOM lists, strings or `arguments` in browsers without `Symbol.iterator`
support.

- [x] Depends on #11263


packages/babel-plugin-transform-for-of/src/index.js
```diff
@@ -6,14 +6,20 @@ import transformWithoutHelper from "./no-helper-implementation";
 export default declare((api, options) => {
   api.assertVersion(7);
 
-  const { loose, assumeArray } = options;
+  const { loose, assumeArray, allowArrayLike } = options;
```
JLHwung: Might worthy to note that this option requires `babel/core @ 7.9.0`, otherwise it is an no-op.

PR #11268, opened 2020-03-15
Allow rest/spread on polyfilled or builtin iterables without `Symbol` support

### LINKED ISSUE #9277: Spread transformation requires Symbol.iterator now

nicolo-ribaudo:
| Fixed Issues?            | Fixes #9277
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Currently, when `Symbol` is not supported, we allow using rest/spread with:
- arrays
- strings
- arguments

With this PR, it will be also possible to use it with
- maps
- sets
- binary arrays

While in old browsers es6 builtins would still need to be polyfilled, it's way
easier to polyfill them because `Symbol` cannot be reliably polyfilled.

I didn't use `instanceof` because:
- it doesn't work with polyfills not attached to the global scope
- when using Babel to load polyfills, it would force the inclusion of `Map` and `Set` polyfills even if they are not used

Downside: the current approach of relying on `toString || construcor.name` doesn't work with subclasses.

---

Depends on https://github.com/babel/babel/pull/11264 (mostly to avoid merge conflicts, it shouldn't be a problem because the other PR is straightforward).
I'm marking this as a draft until that PR is merged, but this is ready for review.

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -909,7 +909,7 @@ helpers.slicedToArray = helper("7.0.0-beta.0")`
     return (
       arrayWithHoles(arr) ||
       iterableToArrayLimit(arr, i) ||
-      unsupportedIterableToArray(arr) ||
+      unsupportedIterableToArray(arr, i) ||
       nonIterableRest()
     );
   }
@@ -925,7 +925,7 @@ helpers.slicedToArrayLoose = helper("7.0.0-beta.0")`
     return (
       arrayWithHoles(arr) ||
       iterableToArrayLimitLoose(arr, i) ||
-      unsupportedIterableToArray(arr) ||
+      unsupportedIterableToArray(arr, i) ||
       nonIterableRest()
     );
   }
@@ -964,11 +964,10 @@ helpers.toConsumableArray = helper("7.0.0-beta.0")`
 `;
 
 helpers.arrayWithoutHoles = helper("7.0.0-beta.0")`
+  import arrayLikeToArray from "arrayLikeToArray";
+
   export default function _arrayWithoutHoles(arr) {
-    if (Array.isArray(arr)) {
-      for (var i = 0, arr2 = new Array(arr.length); i < arr.length; i++) arr2[i] = arr[i];
-      return arr2;
-    }
+    if (Array.isArray(arr)) return arrayLikeToArray(arr);
   }
 `;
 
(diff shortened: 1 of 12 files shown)


packages/babel-helpers/src/helpers.js
```diff
@@ -997,6 +1011,21 @@ helpers.iterableToArrayLimitLoose = helper("7.0.0-beta.0")`
   }
 `;
 
+helpers.unsupportedIterableToArray = helper("7.9.0")`
+  export default function _unsupportedIterableToArray(o) {
+    if (!o) return;
+    if (typeof o === "string") return Array.from(o);
+    var n = Object.prototype.toString.call(o).slice(8, -1);
+    if (n === "Object" && o.constructor) n = o.constructor.name;
```
nicolo-ribaudo: This should work with `es6-shim`. As an alternative, we could detect the presence of a [`_es6-shim iterator_`](https://github.com/paulmillr/es6-shim/blob/0d47be15894bb0f95068545aed69d388be8ce7d0/es6-shim.js#L281) property.

ljharb: By using Array.from, it already works with es6-shim without having to hardcode anything.

ljharb:
My `es-get-iterator` and `iterate-value` packages may help here - they work on every version of every browser and node, and are also tested with es6-shim, core-js 2, and core-js 3.

nicolo-ribaudo:
@ljharb I would prefer to avoid loading them because they can't be easily inlined (when not using `@babel/runtime`) and are way bigger than the helper I implemented in this PR.

I think that we should see if this fix is "good enough", and if we get other bug reports we can iterate.

packages/babel-helpers/src/helpers.js
JLHwung: `Array.from` is not supported on IE11. Can we use `arrayLikeToArray`?

nicolo-ribaudo: The problem is that Maps and Sets are not array-like.
I decided to use `Array.from` because because it is way more easily polyfilled than symbols.

WDYT about using `arrayLikeToArray` for strings/arguments/binary arrays, and `Array.from` for `Map`/`Set`?

ljharb: It's included with es6-shim tho, and imo it's completely reasonable to require anything faithfully polyfillable/shimmable, but symbols are not that.

@nicolo-ribaudo's suggestion works for me as well.

packages/babel-helpers/src/helpers.js
```diff
@@ -1020,6 +1033,28 @@ helpers.iterableToArrayLimitLoose = helper("7.0.0-beta.0")`
...
 
+helpers.unsupportedIterableToArray = helper("7.9.0")`
+  import arrayLikeToArray from "arrayLikeToArray";
+
+  export default function _unsupportedIterableToArray(o, minLen) {
+    if (!o) return;
+    if (typeof o === "string") return arrayLikeToArray(o, minLen);
+    var n = Object.prototype.toString.call(o).slice(8, -1);
+    if (n === "Object" && o.constructor) n = o.constructor.name;
+    if (n === "Map" || n === "Set") return Array.from(n);
```
umanghome, line 1044: @nicolo-ribaudo This should really be
```js
    if (n === "Map" || n === "Set") return Array.from(o);
```
because `n` is a string!

PR #11283, opened 2020-03-18
fix: ensure explicit file extensions in "exports"

guybedford:
Node.js requires "exports" to use explicit file extensions.

That this was working was actually a bug in the modules implementation that will be patched soon, and I'm working on the PR for this.

Posting this PR first thing so that hopefully Babel can avoid any compat issues here when that fix finds its way downstream to a future Node.js release.

Happy to answer any questions further on this too.


existentialism:
@guybedford thanks!

nicolo-ribaudo:
Oh, thanks!

PR #11284, opened 2020-03-18
fix: async arrow functions should not be allowed after binary operator.

### LINKED ISSUE #11203: [parser] Async arrow cannot be rhs of binary operator
## Bug Report

Copied from https://github.com/acornjs/acorn/issues/914

```js
4 + async() => 2
```

[REPL](https://babeljs.io/repl#?browsers=node%206&build=&builtIns=false&spec=false&loose=false&code_lz=CwAg1CCGDOCeB2BjAFAShAXgHwgExA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=script&lineWrap=false&presets=env%2Cenv&prettier=true&targets=Node-13&version=7.8.6&externalPlugins=)

vedantroy:
| Fixed Issues?            |  Fixes #11203 
| Documentation PR Link    | 

Previously, code such as `1 + async () => 2` was parsed succesfully.

The root cause of the issue is inside `parseExprAtom`, which is called by `parseExprSubscripts`. The `switch` statement matches to `case tt.name:` and returns an Identifier node with name "async" (skipping all the if stmts inside the case). Then, `parseExprSubscripts` will reach `return this.parseSubscripts(expr, startPos, startLoc);`, which parses the rest of the expression and returns an `ArrowFunctionExpression`.

The issue is fixed by adding an extra parameter to `this.parseSubscripts` that indicates an async arrow function declaration is forbidden at the location. This prevents Babel from parsing `async() => 2` as an async function declaration. Instead, it will throw once it reaches the `=` sign, because it will expect a semicolon since it is parsing `async()` as a function call (instead of a async arrow function declaration).


packages/babel-parser/src/parser/expression.js
```diff
@@ -545,18 +549,25 @@ export default class ExpressionParser extends LValParser {
...
     base: N.Expression,
     startPos: number,
     startLoc: Position,
     noCalls?: ?boolean,
+    asyncArrowFuncForbidden?: ?boolean,
   ): N.Expression {
     const state = {
       optionalChainMember: false,
-      maybeAsyncArrow: this.atPossibleAsync(base),
+      maybeAsyncArrow: this.atPossibleAsync(base) && !asyncArrowFuncForbidden,
```
JLHwung: Can we simplify a bit to
```suggestion
      maybeAsyncArrow: this.atPossibleAsync(base) && startPos === this.state.potentialArrowAt,
```

nicolo-ribaudo: I find your suggestion also easier to understand :+1:

JLHwung: I realize that `atPossibleAsync` is only used to detect whether there is an async arrow ahead. I suggest we rename `atPossibleAsync` to `atPossibleAsyncArrow`. And the condition `base.start === this.state.potentialArrowAt` can be added to current `atPossibleAsync`.

By doing so we can also fix another related case for free
```typescript
4 + async<number>() => 2 
// This typescript snippet should throw but now it is parsed successfully
```

vedantroy:
Fixed! Also, out of curiosity/so I can better understand the parser, I have a few questions. It would be helpful if someone could answer them.

What does the `noCalls` parameter do in `parseSubscripts`?

Also, why is `this.input.slice(base.start, base.end) === "async"` needed if we know
```
base.type === "Identifier" &&
base.name === "async"
```

Lastly, what is a scenario where `this.state.lastTokEnd` would not equal `base.end` in `atPossibleAsyncArrow`? This question might be too general.

Also, about `atPossibleAsyncArrow`/the old `atPossibleAsync`.

It seems like `atPossibleAsyncArrow` is called in the following situation:
The code is `foo()` and the base expression is .

`atPossibleAsyncArrow` will check if `foo` is

JLHwung:
Great questions!

> What does the noCalls parameter do in parseSubscripts?

`noCalls` is used to indicate whether a left parenthesis after an identifier should be considered as a CallExpression. i.e. When parsing a NewExpression `new C()`, we set `noCalls` to be true so the `(` after `C` is not a CallExpression.

https://github.com/babel/babel/blob/2e6f9583ca738ca50a21ead8c86b86db6a888f0f/packages/babel-parser/src/parser/expression.js#L1435

> Why is this.input.slice(base.start, base.end) === "async" needed if we know (the other two conditions)

This is a very good one! I think we can remove `base.name === "async"` check because `this.input.slice(base.start, base.end) === "async" && base.type === "Identifier"` implies `base.name === "async"`. However the opposite direction can't follow because identifier may contain escaped sequence:
```js
var \u{61}sync
// base.name === "async"
```

> What is a scenario where this.state.lastTokEnd would not equal base.end in

One of such scenarios is that the expression atom can be parenthesized
```js
(async);
// base.end points to 'c'
// this.state.lastTokEnd points to ')'
```

packages/babel-parser/src/parser/expression.js
```diff
@@ -748,7 +748,7 @@ export default class ExpressionParser extends LValParser {
     return this.finishNode(node, "TaggedTemplateExpression");
   }
 
-  atPossibleAsync(base: N.Expression): boolean {
+  atPossibleAsyncArrow(base: N.Expression): boolean {
     return (
```
JLHwung: The conditions here corresponds to `atPossibleAsync`. I was meant to add `base.start === this.state.potentialArrowAt` here so it can be rebranded as `atPossibleAsyncArrow`.
```js
atPossibleAsyncArrow(base: N.Expression): boolean {
  return (base.type === "Identifier" &&
  base.name === "async" &&
  this.state.lastTokEnd === base.end &&
  !this.canInsertSemicolon() &&
  base.end - base.start === 5 &&
  base.start === this.state.potentialArrowAt)
}
```
The first 5 conditions are about `atPossibleAsync` and the last one is `hasPossibleArrow` so it combines into `atPossibleAsyncArrow`.

vedantroy: I think I understand. I'll add the extra condition to this `atPossibleAsyncArrow` method, and delete the `atPossibleAsync` method. This means I will also change

```
if (!noCalls && this.atPossibleAsync(base)) {
```
in the Typescript plugin to
```
if (!noCalls && this.atPossibleAsyncArrow(base)) {
```

nicolo-ribaudo: If you are going with the `base.end - base.start === 5` route, could you add a comment in the code that it's needed to check that there are no escape sequences?

JLHwung: Half joking: If `this.state.lastTokEnd >= base.end && base.start >= this.state.potentialArrowAt` is always true (I don't know). We can even further simplify it to
```js
return (base.type === "Identifier" &&
  base.name === "async" &&
  !this.canInsertSemicolon() &&
  this.state.lastTokEnd - this.state.potentialArrowAt === 5)
)
```

vedantroy:
Updated. I also added a test for Typescript, since this PR will fix a bug there too.

nicolo-ribaudo:
> Should I add a test for async arrow functions after unary operators?

If it was broken before, then yes please!

PR #11288, opened 2020-03-19
Use the correct `this` in `__self` for JSX elements in arrows

nicolo-ribaudo:
| Fixed Issues?            | Reported on Twitter
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The problem was that we were injecting `__self` _after_ transpiling arrow functions, and thus it didn't receive the outer `this` but the one of the new function expression.

This PR fixes the problem by injecting `__self` and `__source` in `Program:enter`.

Probably windows tests will fail because I tried to manually edit them, I'll fix them after CI tells me the necessary updated.


packages/babel-plugin-transform-react-jsx/test/fixtures/nextReactClassic/does-not-add-source-self-windows/exec.js
```diff
@@ -1,31 +0,0 @@
-var actual = transform(
```
nicolo-ribaudo: I merged this test to the one that runs on linux. Also, I made it input/output rather than calling `transform` in `exec.js`.

packages/babel-helper-builder-react-jsx-experimental/src/index.js
```diff
@@ -760,40 +773,33 @@ You can set \`throwIfNamespace: false\` to bypass this warning.`,
...
+    }
 
     return props.length > 0 ? t.objectExpression(props) : t.nullLiteral();
   }
+
+  function sourceSelfError(path, name) {
+    const pluginName = `transform-react-jsx-${name.slice(2)}`;
+
+    return path.buildCodeFrameError(
+      `Duplicate ${name} prop found. You are most likely using the deprecated ${pluginName} Babel plugin. __source and __self will be set automatically in automatic runtime. Please remove transform-react-jsx-source and transform-react-jsx-self from your Babel config.`,
```
existentialism:
```suggestion
      `Duplicate ${name} prop found. You are most likely using the deprecated ${pluginName} Babel plugin. Both __source and __self are automatically set when using the automatic runtime. Please remove transform-react-jsx-source and transform-react-jsx-self from your Babel config.`,
```

lunaruan:
Should we also fix the `transform-react-jsx-self` plugin so that everyone using that also gets the fix?

nicolo-ribaudo:
I can prepare another PR for the standalone plugins, but given that they have been proven for years and that they shouldn't be used anymore I don't think that it's worth it 🤷‍♀️

lunaruan:
`__self` is used primarily for a string ref deprecation warning, and we are worried that people will take some time to upgrade to Babel 8.0 and use the new transform. We would like them to be able to get this warning ASAP so that they can work on removing string refs before they are deprecated, so we would prefer that it is also added to the self transform. Is this okay with you?

Thanks again for doing this!

PR #11289, opened 2020-03-19
Add @babel/helper-validator-identifier

### LINKED ISSUE #10434: the unicode definition of t.isValidIdentifier is outdated
## Bug Report

**Current Behavior**
The following snippet returns false because `esutils` are still using Unicode 9.0.0. 

**Input Code**
- REPL or Repo link if applicable: https://runkit.com/embed/h5oakw0vn4bs

```js

t.isValidIdentifier("\u{1B170}");
```

**Expected behavior/code**
Return true, which also aligns to babel-parser.

**Environment**
- Babel version(s): v7.6.1
- Node/npm version: Node 12
- OS: macOS 10.14, Chrome 76
- Monorepo: no
- How you are using Babel: reading source code and try to get rid of `esutils`

**Possible Solution**
Consider move `isIdentifierStart` and `isIdentifierChar` from tokenizer to babel-types, and then get rid of `esutils`.

\u{1B170} (𛅰) is a [Nüshu](https://en.wikipedia.org/wiki/N%C3%BCshu) character, encoded since Unicode 10.0.0 and has `ID_Start` and `ID_Continue` properties.

Screen Shot 2019-09-12 at 18 51 56

JLHwung:
| Fixed Issues?            | Fixes #10434
| Minor: New Feature?      | Yes, new package is introduced
| Any Dependency Changes?  |

This PR extracts most utilities of `babel-parser/utils/identifier` into a dedicated package. By doing so we can replace `esutils` since we only use a small faction (validation) of it.

This PR also fixes the issue caused by out-of-dated Unicode version bundled by `esutils`.


packages/babel-helper-validator-identifier/src/keyword.js
```diff
@@ -0,0 +1,98 @@
...
+  strictBind: ["eval", "arguments"],
+};
+const keywords = new Set(reservedWords.keyword);
+const reservedWordsStrictSet = new Set(reservedWords.strict);
+const reservedWordsStrictBindSet = new Set(reservedWords.strictBind);
+
+/**
+ * Checks if word is a reserved word in non-strict mode
+ */
+export const isReservedWord = (word: string, inModule: boolean): boolean => {
```
nicolo-ribaudo: Nit: any reason not to use a function declaration? Mostly for consistency with the rest of the file.

JLHwung: Oh I just copy it from `babel-parser`. Can refactor to function declaration.

packages/babel-types/package.json
```diff
@@ -14,7 +14,8 @@
   "dependencies": {
     "esutils": "^2.0.2",
     "lodash": "^4.17.13",
-    "to-fast-properties": "^2.0.0"
+    "to-fast-properties": "^2.0.0",
+    "@babel/helper-validator-identifier": "^7.9.0"
```
existentialism: Nit: alphasort

packages/babel-helper-validator-identifier/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "name": "@babel/helper-validator-identifier",
+  "version": "7.9.0",
+  "description": "Validate identifier/keywords name",
+  "repository": "https://github.com/babel/babel/tree/master/packages/babel-helper-validator-identifier",
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "./lib/index.js",
+  "exports": "./lib/index.js",
```
JLHwung: Following practice from https://github.com/nodejs/node/pull/32098/files#diff-8e67f407bc32a0569e25d7ecaff6e494R208

`"exports": string` is supported on every version of node.js 13.

Issue #11291, opened 2020-03-20
7.9.0: TS regression with import type as an identifier


hzoo:
## Bug Report

Reported via Twitter DM, related to https://github.com/babel/babel/pull/11053

**Current Behavior**

Throws an error with an import with identifier type.

**Input Code**
- [REPL](https://babeljs.io/repl#?browsers=Last%201%20Chrome%20Version&build=&builtIns=false&spec=false&loose=false&code_lz=JYWwDg9gTgLgBDAnmApnAZlCI4HJ0QS4DcQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Ctypescript&prettier=false&targets=&version=7.9.0&externalPlugins=)

```js
import type from 'foo';
```

as opposed to

```js
import type T from './mod';
import type { A, B } from './mod';
import type * as Types from './mod'; 
```

Guess:

We are assuming the check for `type` is the type word vs. an identifier named type so we need to check that first (like if `from` occurs after or something like that)

https://github.com/babel/babel/blob/bb6a1580d208e57f5a75b9a4e5df398837936f15/packages/babel-parser/src/plugins/typescript/index.js#L1869-L1870

kaicataldo:
I'm working on this :)

DanielRosenwasser:
CC @andrewbranch

PR #11296, opened 2020-03-20
fix: parse value imports named type as values

kaicataldo:
| Fixed Issues?            | fixes https://github.com/babel/babel/issues/11291
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR updates the logic for parsing import types to make sure that `type` is actually a keyword rather than an identifier for a value import.

Questions:
1. I'm not sure why so many `options.json` files changes when I ran the tests with `OVERWRITE=true`. Should I revert those changes?
1. Am I missing any cases?


packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -1866,7 +1866,17 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         return this.tsParseImportEqualsDeclaration(node);
       }
 
-      if (this.eatContextual("type")) {
+      const ahead = this.lookahead();
```
existentialism: We should probably only do the lookahead if `this.match(tt.name)`.

Does it make sense to combine this with the above block?

```
if (this.match(tt.name)) {
  const ahead = this.lookahead();

  if (ahead.type === tt.eq) {
    return this.tsParseImportEqualsDeclaration(node);
  }

  if (
    this.state.value === "type" && 
    // import type, { a } from "b";
    ahead.type !== tt.comma &&
    // import type from "a";
    !(ahead.type === tt.name && ahead.value === "from")
  ) {
    node.importKind = "type";
  }
} else {
  node.importKind = "value";
}
```

packages/babel-parser/test/fixtures/estree/typescript/import-require/output.json
```diff
@@ -43,6 +43,7 @@
             "column": 32
           }
         },
+        "importKind": "value",
```
existentialism: I don't think we want to set `importKind` on `TSImportEqualsDeclaration`s

kaicataldo: For my own understanding, why is that? Isn't it an imported value?

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -1862,14 +1862,25 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
-        node.importKind = "value";
+      if (this.match(tt.name)) {
+        const ahead = this.lookahead();
+
+        if (ahead.type === tt.eq) {
+          return this.tsParseImportEqualsDeclaration(node);
+        }
+
+        if (
+          this.state.value === "type" &&
```
JLHwung: should also check `!this.state.containsEsc`, which is also included in `isContexual`.
```js
import typ\u{65} typescript from "typescript";
```

Note, it seems that the TS [REPL](https://www.typescriptlang.org/play/?ts=Nightly#code/JYWwDg9gTgLgBDAnmAOgVwN4DYCsBfBZAUwGcBjKYMeAMyghDgCIkxSKqYmBuIA) does not agree with me, which should be a bug.

nicolo-ribaudo: Yeah, let's assume that escapes are disallowed for now and report it to the TS team (cc @DanielRosenwasser)

packages/babel-parser/test/fixtures/flow/anonymous-function-types/bad_01/options.json
```diff
@@ -1,3 +1,8 @@
 {
+  "sourceType": "module",
+  "plugins": [
+    "jsx",
+    "flow"
+  ],
   "throws": "Unexpected token, expected \";\" (1:29)"
```
nicolo-ribaudo: Could you revert all these unrelated changes to the tests? :sweat_smile:

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -1862,14 +1862,25 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+        }
+
+        if (
+          this.isContextual("type") &&
+          // import type, { a } from "b";
+          ahead.type !== tt.comma &&
+          // import type from "a";
+          !(ahead.type === tt.name && ahead.value === "from")
+        ) {
+          this.eat(tt.name);
```
nicolo-ribaudo: Nit: `this.next()` is enough, since we are not using the return value of `this.eat` here.

packages/babel-parser/test/fixtures/flow/anonymous-function-types/bad_01/options.json
kaicataldo: Sure thing.

packages/babel-plugin-transform-typescript/test/fixtures/imports/import-named-type/output.mjs
```diff
@@ -0,0 +1 @@
+
```
jridgewell: Shouldn't this have an output?

kaicataldo: Yes - fixing now

hzoo:
Wrote in Slack, but I think we need to check against `export type` as well? (doesn't have to be in this pr tho)

packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -1862,14 +1862,25 @@ export default (superClass: Class<Parser>): Class<Parser> =>
     checkDuplicateExports() {}
 
     parseImport(node: N.Node): N.AnyImport {
-      if (this.match(tt.name) && this.lookahead().type === tt.eq) {
-        return this.tsParseImportEqualsDeclaration(node);
-      }
+      if (this.match(tt.name) || this.match(tt.star) || this.match(tt.braceL)) {
```
kaicataldo: I _think_ this covers all cases?

packages/babel-plugin-transform-typescript/test/fixtures/imports/import-named-type/output.mjs
kaicataldo: I forgot to set the `onlyRemoveTypeImports` option for these tests 🤦‍♂

packages/babel-parser/src/plugins/typescript/index.js
nicolo-ribaudo: Yeah it should.

PR #11299, opened 2020-03-20
Class cannot extend async function

### LINKED ISSUE #11209: [parser]: A class should not extend an async function
## Bug Report

copied from https://github.com/acornjs/acorn/issues/917

The following snippet throw unexpected token `=>`.

```js
class A extends async () => {} {}
class B extends async function () {} {}
```

[REPL](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAECC0FMAeAXBA7AJjSBPdw0AFAJTQC8AfNAN4C-tdAUKJDAEKKobbR4HQAZgFcCKAJYB7dMTL1GQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=script&lineWrap=false&presets=env&prettier=false&targets=Node-6&version=7.8.7&externalPlugins=).

vedantroy:
| Fixed Issues?            | Fixes #11209 
| Documentation PR Link    | 
| Any Dependency Changes?  |

Prevents `class B extends async function () {} {}`.

Half of issue #11209 will be fixed by https://github.com/babel/babel/pull/11284, since that PR makes it a syntax error for a class to extend an async arrow function.

This PR fixes the other half of the issue, making it impossible for a class to extend an anonymous non-arrow async function.

Note: I don't know if this fix should be in Babel though, because it seems like more of a type/runtime issue rather than a syntax/parser issue?

### DIFF
--- a/packages/babel-parser/src/parser/statement.js
+++ b/packages/babel-parser/src/parser/statement.js
@@ -1648,7 +1648,19 @@ export default class StatementParser extends ExpressionParser {
   }
 
   parseClassSuper(node: N.Class): void {
-    node.superClass = this.eat(tt._extends) ? this.parseExprSubscripts() : null;
+    const superClass = this.eat(tt._extends)
+      ? this.parseExprSubscripts()
+      : null;
+    if (superClass !== null) {
+      if (superClass.type == "FunctionExpression" && superClass.async) {
+        this.raise(
+          superClass.start,
+          Errors.ExtendsValueNotConstructor,
+          this.input.slice(superClass.start, superClass.end),
+        );
+      }
+    }
+    node.superClass = superClass;
   }
 
   // Parses module export declaration.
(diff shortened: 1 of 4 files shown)


packages/babel-parser/src/parser/statement.js
```diff
@@ -1648,7 +1648,19 @@ export default class StatementParser extends ExpressionParser {
   }
 
   parseClassSuper(node: N.Class): void {
-    node.superClass = this.eat(tt._extends) ? this.parseExprSubscripts() : null;
+    const superClass = this.eat(tt._extends)
+      ? this.parseExprSubscripts()
+      : null;
+    if (superClass !== null) {
```
vedantroy: I had to use 2 if stmts instead of 1 because Flow can't recognize that `superClass !== null` if I do `if (superClass !== null && ...` (I think).

nicolo-ribaudo:
> Note: I don't know if this fix should be in Babel though, because it seems like more of a type/runtime issue rather than a syntax/parser issue?

When you have this doubt, I suggest running this code in your browser's console, without calling the function:
```js
function fn() {
  // code you want to test
}
```
for example
```js
function fn() {
  class A extends async function() {} {}
}
```

Since we are not calling the function, that code won't be executed and it won't throw any runtime error: it throws if and only if there is a syntax error.

As you can test, that code doesn't throw because it's valid syntax (even if invalid at runtime), and thus it shouldn't be fixed here.

Issue #11303, opened 2020-03-21
[transform-parameter] Incorrect IIFE when shadowed variable initialized from parameters


huanzhiyazi:
- **First, check a function like this:**

```js
const dateSample = (date, defValue = 0) => {
  var date = new Date(date, defValue);
  return date;
}
```

- **And then transfer it with Babel 7:**

```js
var dateSample = function dateSample(date) {
  var defValue = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : 0;
  return function () {
    var date = new Date(date, defValue);
    return date;
  }();
};
```

The input param `date` is discarded for function scope chain.

- **But if you transfer the original code with Babel 6, the output code is:**

```js
var dateSample = function dateSample(date) {
    var defValue = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : 0;

    var date = new Date(date, defValue);
    return date;
};
```

Yeah! You can run it perfectly.

I am not sure if it is a bug! But I think the original code is not a good coding habit, which should be processed as an error to warn the developers.

JLHwung:
The Babel 7 behaviour is expected. I would not consider it a bug yet it is indeed sub-optimal.

> The input param date is discarded for function scope chain.

When there is a variable declaration shadowing the function parameter binding, (i.e. `var date` inside `date`-param function), we will wrap the function body inside an [IIFE](https://en.wikipedia.org/wiki/Immediately_invoked_function_expression) so that the scope of the `var`-binding is _always_ up to its function body. That means it should not interfere with the desugared parameter initializer. For example (taken from `packages/babel-plugin-transform-parameters/test/fixtures/parameters/var-same-as-param-closure/exec.js`)

```js
var x = 1
function foo(x, y = function () { x = 2 }) {
  var x = 3
  y()
  expect(x).toBe(3);
}
foo()
expect(x).toBe(1);

```
If we do not wrap `var x = 3; y()` into an IIFE, `x` will be changed to `2` by default `y` implementation.

> I think the original code is not a good coding habit, which should be processed as an error to warn the developers.

True. But babel placed correctness over legibility. The babel processed code is not supposed to be of good readability.

Note that in your case, the IIFE _is_ redundant, and we could definitely improve it a bit using the following heuristics:

Do not wrap function body as IIFE if
1) the parameters contain either Assignment Pattern or Identifier
2) at least one of parameters is Assignment Pattern
3) The initializer of the assignment pattern is a literal (does not refer to any other params)

[...]

PR #11307, opened 2020-03-21
Define default value for vars shadowing params

### LINKED ISSUE #11256: `Cannot read property 'type' of null` again

nicolo-ribaudo:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/11303, fixes #11256
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The best way to handle params's horrible shadowing behavior is by... relying on it :grin:


packages/babel-plugin-transform-parameters/src/params.js
```diff
@@ -1,4 +1,3 @@
-import callDelegate from "@babel/helper-call-delegate";
```
nicolo-ribaudo: This is now unused and can be removed from the repo.

nicolo-ribaudo:
I ended up also pushing a workaround for https://github.com/babel/babel/issues/11256 (a fix would require major changes to how we handle the scope)

packages/babel-plugin-transform-parameters/test/fixtures/parameters/var-no-init-for-init-same-as-param/exec.js
```diff
@@ -1,5 +1,5 @@
 function foo(a = 2) {
   for (var a, i = 0; i < 1; i++);
-  expect(a).toBe(undefined);
+  expect(a).toBe(2);
```
JLHwung: Good catch!

packages/babel-plugin-transform-parameters/src/params.js
```diff
@@ -52,18 +51,22 @@ export default function convertFunctionParams(path, loose) {
...
   };
 
   const body = [];
   const params = path.get("params");
 
-  let firstOptionalIndex = null;
-
-  for (let i = 0; i < params.length; i++) {
-    const param = params[i];
+  let isSimpleParameterList = true;
```
JLHwung: What about
```js
const isSimpleParameterList = params.every(param => param.isIdentifier());
```
It will return `true` when `params` is empty.

packages/babel-plugin-transform-parameters/src/params.js
```diff
@@ -52,18 +51,22 @@ export default function convertFunctionParams(path, loose) {
   const { node, scope } = path;
 
   const state = {
-    iife: false,
-    scope: scope,
+    stop: false,
+    needsOuterBinding: false,
+    shadowedParams: new Set(),
```
JLHwung: We can move `shadowedParams` out of `state` if we don't intend it accessed by `visitIdentifier`.

packages/babel-plugin-transform-parameters/test/fixtures/regression/11303/exec.js
```diff
@@ -0,0 +1,6 @@
+function test(date, defValue = 1) {
```
JLHwung: nit: lol they don't have to be `date` and `defValue`.

packages/babel-plugin-transform-parameters/src/params.js
```diff
@@ -23,27 +22,27 @@ const buildSafeArgumentsAccess = template(`
...
+    (scope.getBinding(name) === state.scope.parent.getBinding(name) &&
+      state.scope.hasOwnBinding(name))
+  ) {
+    state.needsOuterBinding = true;
+    path.stop();
+  }
+};
 
 const iifeVisitor = {
-  ReferencedIdentifier(path, state) {
```
JLHwung: I guess we can simplify to
```js
const iifeVisitor = {
  "ReferencedIdentifier|BindingIdentifier": visitIdentifier,
}
```
If there is a BindingIdentifier in the function scope, it should not satisfy `scope.getBinding(name) === state.scope.parent.getBinding(name)` so essentially it will be noop.

packages/babel-plugin-transform-parameters/src/params.js
```diff
@@ -178,13 +171,34 @@ export default function convertFunctionParams(path, loose) {
...
 
   return true;
 }
+
+function buildScopeIIFE(shadowedParams, body) {
+  const args = [];
+  const params = [];
+
+  for (const name of shadowedParams) {
+    // We create then twice; the other option is to use t.cloneNode
```
existentialism:
```suggestion
    // We create them twice; the other option is to use t.cloneNode
```

packages/babel-plugin-transform-parameters/src/params.js
```diff
@@ -23,47 +22,40 @@ const buildSafeArgumentsAccess = template(`
...
+  "ReferencedIdentifier|BindingIdentifier"(path, state) {
     const { scope, node } = path;
+    const { name } = node;
+
     if (
-      node.name === "eval" ||
-      !isSafeBinding(scope, node) ||
-      !isSafeBinding(state.scope, node)
+      name === "eval" ||
+      (scope.getBinding(name) === state.scope.parent.getBinding(name) &&
```
jridgewell: Isn't this just a `!scope.hasOwnBinding` check?

JLHwung: Nope, `state.scope` does not equal to `scope`.

PR #11309, opened 2020-03-21
Use prettier 2

nicolo-ribaudo:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

https://prettier.io/blog/2020/03/21/2.0.0.html

---

Years ago, when I was a young developer moving the first steps on GitHub, I really _hated_ that some developers didn't use a space between `function` and `(`: we put the space after `function` when there is a name, why shouldn't we do the same when the name is missing?

Then Prettier came and removed my spaces. It was really ugly, but I had to live with it because Prettier's benefits outweighed my disgust.

Then I grew up, and Prettier was part of my life. How is it possible that I put spaces between `function` and `(`? We don't put the space before `(` when there is a name: why should we put it when the name is missing?

Now Prettier 2 is coming and it adds those ugly spaces. They make my eye bleed, I don't want to look at this diff. Why do they do this to me?

**EDIT (2021-03-21):** I am now an experienced developer, but the space between `function` and `(` is not something I'm willing to remove. This is a hill I'm ready to die on. What was I thinking when I didn't want those spaces? The younger me was really crazy.

### DIFF
--- a/packages/babel-core/src/config/partial.js
+++ b/packages/babel-core/src/config/partial.js
@@ -135,30 +135,32 @@ export default function* loadPrivatePartialConfig(
   };
 }
 
-export const loadPartialConfig = gensync<[any], PartialConfig | null>(function*(
-  inputOpts: mixed,
-): Handler<PartialConfig | null> {
-  const result: ?PrivPartialConfig = yield* loadPrivatePartialConfig(inputOpts);
-  if (!result) return null;
+export const loadPartialConfig = gensync<[any], PartialConfig | null>(
+  function* (inputOpts: mixed): Handler<PartialConfig | null> {
+    const result: ?PrivPartialConfig = yield* loadPrivatePartialConfig(
+      inputOpts,
+    );
+    if (!result) return null;
 
-  const { options, babelrc, ignore, config } = result;
+    const { options, babelrc, ignore, config } = result;
 
-  (options.plugins || []).forEach(item => {
-    if (item.value instanceof Plugin) {
-      throw new Error(
-        "Passing cached plugin instances is not supported in " +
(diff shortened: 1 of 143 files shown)


packages/babel-core/src/config/partial.js
```diff
@@ -135,30 +135,32 @@ export default function* loadPrivatePartialConfig(
...
-
-  return new PartialConfig(
-    options,
-    babelrc ? babelrc.filepath : undefined,
-    ignore ? ignore.filepath : undefined,
-    config ? config.filepath : undefined,
-  );
-});
+export const loadPartialConfig = gensync<[any], PartialConfig | null>(
+  function* (inputOpts: mixed): Handler<PartialConfig | null> {
```
nicolo-ribaudo: Ok, I have to admit I already like this space after `*`

lydell:
FWIW, there’s [prefer-arrow-callback](https://eslint.org/docs/rules/prefer-arrow-callback) in ESLint (with autofix!) if you want to avoid the space-or-no-space thing in more cases.

nicolo-ribaudo:
Oh, we have to wait for Babel 8 because of Node 6 support :sob:

jridgewell:
> Oh, we have to wait for Babel 8 because of Node 6 support 😭

Upgrading a dev dependency should be fine. The environment you develop babel in can be increased without semver issues, it's only the environment that uses babel that needs to be maintained.

nicolo-ribaudo:
The problem is that we use prettier to build `@babel/types`, so the build fails on travis.

PR #11314, opened 2020-03-22
Fix 11281

### LINKED ISSUE #11281: `object-rest-spread` does not transpile correctly when binding is referenced in param initializer
## Bug Report

**Input Code**
- [REPL](https://babeljs.io/repl#?browsers=chrome%2080&build=&builtIns=false&spec=false&loose=false&code_lz=BQbwBAhgNGB08HkwF8YCMwF4wJgYy0gEosA-MEZIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Cenv&prettier=false&targets=&version=7.8.7&externalPlugins=%40babel%2Fplugin-proposal-object-rest-spread%407.8.3%2C%40babel%2Fplugin-external-helpers%407.8.3)

```js
({ a, ...O }, b = O, c = a) => {}
```
**Current Behavior**
The code above is transformed to
```js
(_ref, b = O, c = a) => {
  let {
    a
  } = _ref,
      O = babelHelpers.objectWithoutProperties(_ref, ["a"]);
};
```
where initializers `b = O` and `c = a` no longer refers to the destructured params. 

**Expected behavior/code**
It should transformed to
```js
(function (_ref) {
  let { a } = _ref,
      O = babelHelpers.objectWithoutProperties(_ref, ["a"]);
  var b = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : O;
  var c = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : a;
});
```

- Filename: `babel.config.js`

```js
{
  "plugins": { "@babel/proposal-object-rest-spread" }
}
```

**Environment**: REPL

**Possible Solution**
I haven't come up with a simple solution. Suggestions are welcome.

Solution 1: Fall back to `transform-parameters` when such pattern is detected

i) A function param declarator contains a RestElement.
ii) The declarator is also referenced by another params' initializers.

This solution will downgrade unnecessary codes, i.e.
```js
({ a, ...O }, b = 2, c = O) => {}
```
will be transformed to
```js
(function (_ref) {
  var a = _ref.a,
      O = babelHelpers.objectWithoutProperties(_ref, ["a"]);
  var b = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : 2;
  var c = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : O;
});
```
even when `b = 2` is supported on the target platform.

Solution 2: Extract `convertSingleParam` from `convertParams` defined in `transform-parameters`, apply `convertSingleParam` if one of params satisfies the pattern mentioned above. 

This solution should prevent downgrading the code as much as possible. i.e.
```js
({ a, ...O }, b = 2, c = O) => {}
```
should be transformed to
```js
(function (_ref, b = 2) {
 let { a } = _ref,
      O = babelHelpers.objectWithoutProperties(_ref, ["a"]);
   var c = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : O;
});
```

This issue is discovered when I am investigating #11278.

vedantroy:
| Fixed Issues?            | Fixes #11281 
| Documentation PR Link    | 

Fixes the bug where `({ a, ...R }, b = R, c = a) => {}` was not transpiled correctly if `transform-function-parameters` was not enabled.

I used solution 2 as described by @JLHwung in the issue post. This solution loops over each param with a default initializer `a = b`/nested default initializer (like `{a = b}` and roughly only transforms it if it is referencing a id created in an object rest pattern or if the previous parameter has been processed.

For example, only the last 2 parameters in this function will be transformed.
`({...b}, e, c = 2, a = b, f) => {}`

Looking at the "parameters-extra" tests will clarify.

**Problem**: Because I extracted `convertParam` from `extractFunctionParams` in the `babel-plugin-transform-parameters` package, I had to import that package into this plugin. However, I couldn't do that without breaking the package-lock.json, so I'm not sure what to do.

My code is kind of long :disappointed:, but hopefully it's the correct approach :smiley:

**Edit**: Looks like CI is failing because of the package.json issue I mentioned above.


vedantroy:
Update: I didn't realize, but this code is still a bit broken, so I'm going to close this PR for now.

PR #11315, opened 2020-03-22
[preset-typescript] Fix private members type annotations

### LINKED ISSUE #11310: [preset-typescript] Type annotations are not removed for private fields
## Bug Report

**Current Behavior**
When `@babel/preset-typescript` is used along with `@babel/plugin-syntax-class-properties` , it does not remove type annotations for private fields and methods.

**Input Code**
[Private fields typings - Runkit](https://runkit.com/lodin/private-fields-typings)

```js
class Foo {
  #bar: string = 'baz';
}
```

**Expected behavior/code**
A pure JS code that can be executed in the modern environment. No Typescript annotations should left.

```js
class Foo {
  #bar = 'baz';
}
```

- Filename: no (inline configuration)

```js
{
  presets: [require('@babel/preset-typescript')],
  plugins: [
    require('@babel/plugin-syntax-class-properties'),
  ]
}
```

**Environment**
- Babel version(s): 7.9.0
- Node/npm version: Node 13.5.0 / npm unknown (Runkit)
- OS: unknown (Runkit; probably, Linux)
- Monorepo: no
- How you are using Babel: `babel.transform()`

**Possible Solution**
I assume that it happens because of code that starts at [line 327](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-typescript/src/index.js#L327). There are checks only for class property and class method but not for class private property and class private method.

The strange thing is that for a private method everything works correct though it fails for a private property.

Lodin:
| Fixed Issues?            | Fixes #11310
| Documentation PR Link    | 

This PR fixes the bug when typescript transformer leaves type annotations for private properties. It does two main things:
- Removes annotation for property that has initializer
- Removes annotation and keeps private property declared if there is no initializer.
  ```typescript
  class Foo {
    #bar: string;
  }
  ```
  By default typescript transformer removes such properties, but we need to preserve them according to the spec.
  ```js
   class Foo {
    #bar;
   }
   ```

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -78,7 +78,12 @@ export default declare(
           if (!node.decorators) {
             path.remove();
           }
-        } else if (!allowDeclareFields && !node.value && !node.decorators) {
+        } else if (
+          !allowDeclareFields &&
+          !node.value &&
+          !node.decorators &&
+          !t.isClassPrivateProperty(node)
+        ) {
           path.remove();
         }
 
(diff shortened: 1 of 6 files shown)


nicolo-ribaudo:
Ignore the CI failure, is caused by an outdated base branch.

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -325,13 +330,16 @@ export default declare(
           // class transform would transform the class, causing more specific
           // visitors to not run.
           path.get("body.body").forEach(child => {
-            if (child.isClassMethod()) {
+            if (child.isClassMethod() || child.isClassPrivateMethod()) {
```
nicolo-ribaudo: Note: this doesn't actually fix anything, but I'm ok with keeping it since we could add new things to the `method` handler, which could be also allowed with private methods.

Lodin: Yes, I left it for consistency and for possible further changes

PR #11316, opened 2020-03-22
Better file extension handling for TypeScript and Flow presets

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR includes two different commits, conceptually separated, but the second one relies on the first one. The first one could be backported to Babel 7 if needed, but the second one is a breaking change.
1. `Don't require a filename for test/include/exclude _functions_`
2. `Better file extension handling for TS/Flow/JSX`

The first commit removes the requirement to provide a filename to Babel when a preset specifies a _function_ in `test`/`include`/`exclude`. By doing so, it's possible for presets to provide a fallback behavior in case of a missing file name.
I didn't remove the restriction for strings or regular expressions, since it's not clear what the default fallback should be.

The second commit provides a better UX when using Flow, TypeScript or JSX. The two main changes are that the plugins/presets don't opaquely disable other plugins, and that a filename isn't required anymore. I also changed the options they take regarding file names:
> - When using the TS preset, the filename is not required anymore
> - When both the TS and Flow plugins are enabled on the same file, an error is thrown rather then giving the precedence to TS
> - The TypeScript plugin doesn't disable JSX anymore
> - By default, the Flow preset only enables the plugin when applied to a non-ts file
> - The .isTSX option has been removed from the TS pluginsand preset, since users can directly enable the JSX plugin.
> - The .allExtensions option has been removed from the TS plugin, since it was used to replace the extension test with the removed .isTSX option
> - A new option, .ignoreExtensions, has been added to both the TS and Flow plugins, to make it possible to run them on JS and TS files respectively

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11316"></a>


JLHwung:
@nicolo-ribaudo Can you rebase?

packages/babel-preset-typescript/src/index.js
```diff
@@ -1,59 +1,61 @@
...
     },
   ) => {
     api.assertVersion(7);
 
-    if (typeof allExtensions !== "boolean") {
-      throw new Error(".allExtensions must be a boolean, or undefined");
+    if (typeof allExtensions !== "undefined" || typeof isTSX !== "undefined") {
+      throw new Error(
+        "The .allExtensions and .isTSX options have been removed.\n" +
+          "If you want to disable JSX detection based on the file extension, " +
```
existentialism, line 24:
```suggestion
          "If you want to disable file extension-based JSX detection, " +
```

packages/babel-preset-typescript/src/index.js
```diff
@@ -21,7 +21,7 @@ export default declare(
     if (typeof allExtensions !== "undefined" || typeof isTSX !== "undefined") {
       throw new Error(
         "The .allExtensions and .isTSX options have been removed.\n" +
-          "If you want to disable JSX detection based on the file extension, " +
+          "If you want to disable file extension-based JSX detection, " +
```
nicolo-ribaudo, line 24: English needs parentheses, here the fact that ` ` between "file" and "extension" has a higher precedence than `-` feels... weird :stuck_out_tongue:

PR #11319, opened 2020-03-23
Test Babel compiled using an old version of `@babel/core`

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  |

This should catch regressions where we use a new helper without a fallback
for older versions, or when we use a new `@babel/traverse` method.

### DIFF
--- a/scripts/integration-tests/e2e-babel-old-version.sh
+++ b/scripts/integration-tests/e2e-babel-old-version.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+#==============================================================================#
+#                                  SETUP                                       #
+#==============================================================================#
+
+# Start in scripts/integration-tests/ even if run from root directory
+cd "$(dirname "$0")" || exit
+
+source utils/local-registry.sh
+source utils/cleanup.sh
+
+# Echo every command being executed
+set -x
+
+# Go to the root of the monorepo
+cd ../..
+
+#==============================================================================#
+#                                   TEST                                       #
+#==============================================================================#
+
+startLocalRegistry "$PWD"/scripts/integration-tests/verdaccio-config.yml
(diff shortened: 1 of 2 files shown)


scripts/integration-tests/e2e-babel-old-version.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# Go to the root of the monorepo
+cd ../..
+
+#==============================================================================#
+#                                   TEST                                       #
+#==============================================================================#
+
+startLocalRegistry "$PWD"/scripts/integration-tests/verdaccio-config.yml
+yarn upgrade --scope @babel
```
JLHwung: Is `yarn upgrade` working again in our repo?

nicolo-ribaudo: We only need to upgrade the dependencies of the top-level `package.json`, since they are what we use to compile.

nicolo-ribaudo: Oh ignore me, it doesn't work.

JLHwung: We have replaced `yarn upgrade` by an ad-hoc script in #11021 due to https://github.com/yarnpkg/yarn/issues/7797. I am not sure if that issue has been fixed or it affects only `yarn upgrade-interactive` so `yarn upgrade` is good.

nicolo-ribaudo: It doesn't give problems here because we are not using yarn workspaces.

nicolo-ribaudo:
Thanks for all the reviews and thanks to me for the PR!

I didn't enable the test :rofl:

existentialism:
Tests can't fail if they don't run!

PR #11322, opened 2020-03-23
Compact parser fixture loc info

JLHwung:

Another MX (Maintinaer Experience) PR.

When working on babel parser, the fixture is typically too long to skim. One of obstacles are the `loc` info occupies many lines. This PR compacts the fixtures a bit so now it should be easier to read the test fixtures.

For example, see if you can read this fixture now: https://github.com/babel/babel/blob/bd783c4543d74ffa8f2a248ea449eda64d4fdf77/packages/babel-parser/test/fixtures/comments/basic/array-pattern-trailing-comma/output.json

Please review the first commit only as the whole diff turns out too large for GitHub.

The idea is stolen from the Flow repo.


JLHwung:
@nicolo-ribaudo CI is green now. 😄

Issue #11320, opened 2020-03-23
Inserted identifier clashes with existing identifier from the inner scope


Andarist:
## Bug Report

- [x] I would like to work on a fix!

**Current Behavior**

Inserted identifier clashes with the existing one **if** the scope gets crawled before generating the inserted UID.

**Input Code**
```js
require('@babel/core').transform(`

  export default () => {
    var _jsx;
    return <div />;
  };

`, {
  configFile: false,
  babelrc: false,
  plugins: [api => ({ visitor: { Program(p) { p.scope.crawl() } } })],
  presets: [['@babel/preset-react', { runtime: 'automatic' }]],
}).code
```

**Expected behavior/code**
actual vs expected
```diff
-import { jsx as _jsx } from "react/jsx-runtime";
+import { jsx as _jsx2 } from "react/jsx-runtime";
export default (() => {
  var _jsx;

-  return /*#__PURE__*/_jsx("div", {});
+  return /*#__PURE__*/_jsx2("div", {});
});
```

**Babel Configuration**

```js
{
  configFile: false,
  babelrc: false,
  plugins: [api => ({ visitor: { Program(p) { p.scope.crawl() } } })],
  presets: [['@babel/preset-react', { runtime: 'automatic' }]],
}
```

**Environment**

Babel 7.9, the rest is quite irrelevant as the issue is not env-specific and is easy to reproduce.

nicolo-ribaudo:
This might be a bug in `@babel/helper-module-imports`, or in the logic used to collect all the already used identifiers. I don't think that it lives in the jsx helper.

PR #11323, opened 2020-03-23
Fixed issue with programPath.scope.references not being registered back correctly after scope re-crawling

Andarist:
| Fixed Issues?            | fixes https://github.com/babel/babel/issues/11320
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 

References were not correctly registered back on subsequent crawls for cases when bindings were already registered on some inner scopes.

### DIFF
--- a/packages/babel-traverse/test/scope.js
+++ b/packages/babel-traverse/test/scope.js
@@ -303,6 +303,24 @@ describe("scope", () => {
 
       expect(path.scope.bindings.a).toBe(path.get("body[0]").scope.bindings.a);
     });
+
+    it("references after re-crawling", function() {
+      const path = getPath("function Foo() { var _jsx; }");
+
+      path.scope.crawl();
+      path.scope.crawl();
+
+      expect(path.scope.references._jsx).toBeTruthy();
+    });
+
+    test("generateUid collision check after re-crawling", function() {
+      const path = getPath("function Foo() { var _jsx; }");
+
+      path.scope.crawl();
+      path.scope.crawl();
+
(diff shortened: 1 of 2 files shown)


packages/babel-traverse/test/scope.js
```diff
@@ -303,6 +303,15 @@ describe("scope", () => {
...
       expect(path.scope.bindings.a).toBe(path.get("body[0]").scope.bindings.a);
     });
+
+    it("references after re-crawling", function() {
+      const path = getPath("function Foo() { var jsx; }");
+
+      path.scope.crawl();
+      path.scope.crawl();
+
+      expect(path.scope.references.jsx).toBeTruthy();
```
nicolo-ribaudo: Note: I have no idea of what `references` on the `scope` means since it's clearly not related to `binding.referencePaths`. However, this PR aligns the post-crawl state to the pre-crawl one.

Andarist: I'm not sure if this is the full story but those references seem to be a simple `Record<name, true>` of everything that could clash with generated identifiers (or other nodes), since it's not possible to just check some single scope to be sure if some name can be considered "sage", but it rather requires full knowledge about the module/script

Andarist:
> @Andarist Would it be possible to also add a test using generateUid, which is how you found ths bug?

Sure, should I add it to the same file? Or is there a better place for a test like this?

nicolo-ribaudo:
It's ok in this file.

PR #11325, opened 2020-03-24
fix: check parentheses between optional chain and other types

JLHwung:
| Fixed Issues?            | Seems like they have not been reported.

Rule of thumb: copy `isCallExpression` and `isMemberExpression` references whenever the `base` itself is not an optional chain.


packages/babel-generator/src/node/parentheses.js
```diff
@@ -59,8 +59,12 @@ export function UpdateExpression(node: Object, parent: Object): boolean {
   return (
     // (foo++).test(), (foo++)[0]
     t.isMemberExpression(parent, { object: node }) ||
+    // (foo++)?.test(), (foo++)?.[0]
+    t.isOptionalMemberExpression(parent, { object: node }) ||
```
jridgewell: Perhaps we should be adding an alias for these, so we don't need to duplicate everywhere.

nicolo-ribaudo: Yeah, I agree.

JLHwung: It's really difficult to come up with a proper name other than `CallExpression` itself. 😄

nicolo-ribaudo: `MemberExpressionish` 😛

jridgewell: Yah, I was thinking `isMemberish` and `isCallish`.

PR #11326, opened 2020-03-24
Correctly transpile when default parameter initializer references binding in rest pattern

### LINKED ISSUE #11281: `object-rest-spread` does not transpile correctly when binding is referenced in param initializer

vedantroy:
Note: If the title seems familiar, it is because I copied it from https://github.com/babel/babel/pull/11317, which I closed because I rewrote the code to a large degree, so the commit history was useless. Third time is the charm :upside_down_face:
| Fixed Issues?            | Fixes #11281 
| Documentation PR Link    | 

Fixes the bug where `({ a, ...R }, b = R, c = a) => {}` was not transpiled correctly if `transform-function-parameters` was not enabled.

I fixed it by modifying `convertFunctionParams` to accept a function to transform rest elements.


packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -1,6 +1,7 @@
 import { declare } from "@babel/helper-plugin-utils";
 import syntaxObjectRestSpread from "@babel/plugin-syntax-object-rest-spread";
 import { types as t } from "@babel/core";
+import { convertFunctionParams } from "@babel/plugin-transform-parameters";
```
nicolo-ribaudo: You need to add this as a dependency in package.json

vedantroy: Fixed. For some reason I thought adding it as a dependency wouldn't work because on my previous branch when I tried `npm i -S @babel/plugin-transform-parameters` and `npm i -S @babel/helper-call-delegate` npm threw up due package-lock.json conflict.

JLHwung: The correct (by the book) recipe is
```sh
lerna add @babel/helper-call-delegate --scope=@babel/plugin-proposal-object-rest-spread
```
But I feel like it's too verbose and most of time I manually revise `package.json` and then run `yarn lerna bootstrap`.

We should state it clearly in `CONTRIBUTING.md`.

packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -213,8 +218,67 @@ export default declare((api, opts) => {
...
+            if (paramHasRestElement[i]) {
+              replaceRestElement(param.parentPath, param);
+            }
+          }
+        } else {
+          const shouldTransformParam = idx =>
+            idx >= i - 1 || paramHasRestElement[idx];
+          convertFunctionParams(
+            path,
+            false, // TODO: What should this value be?
```
JLHwung: I think it is okay to pass through `loose` here.

packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -213,8 +218,67 @@ export default declare((api, opts) => {
       // function a({ b, ...c }) {}
       Function(path) {
         const params = path.get("params");
-        for (let i = params.length - 1; i >= 0; i--) {
-          replaceRestElement(params[i].parentPath, params[i]);
+        const paramHasRestElement = [];
```
JLHwung: It suffices to model `paramHasRestElement` as `Set<number>` and to only include the index of which has rest element.

packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -213,8 +218,67 @@ export default declare((api, opts) => {
...
+          if (idInRest) assignPath.stop();
+        };
+
+        let i;
+        for (i = 0; i < params.length && !idInRest; ++i) {
+          const param = params[i];
+          if (!paramHasRestElement[i]) {
+            if (param.isAssignmentPattern()) AssignmentPatternHandler(param);
+            else {
+              param.traverse({ AssignmentPattern: AssignmentPatternHandler });
```
JLHwung: we need to skip scopables when doing sub-traverse:
```js
Scope(path) {
  path.skip()
}
```

i.e.
```js
({ ...R }, f = R => R) => {}
```

packages/babel-plugin-proposal-object-rest-spread/src/index.js
vedantroy: Changed. I used an array of booleans instead of a set since the code was about the same length and I figured array access would be (very marginally) faster than set access.

vedantroy:
Any more feedback?

packages/babel-plugin-proposal-object-rest-spread/src/index.js
```diff
@@ -209,12 +214,76 @@ export default declare((api, opts) => {
...
+        }
+
+        // if true, a parameter exists that has an id in its initializer
+        // that is also an id bound in a rest parameter
+        // example: f({...R}, a = R)
+        let idInRest = false;
+
+        const IdentifierHandler = function(idPath) {
+          if (
+            (idPath.parentKey === "right" || idPath.parentKey === "value") &&
```
JLHwung: It does not suffice to check `parentKey` to determine if `idPath` is _referenced_. i.e.
```js
({ ...R }, a = f(R)) => {}
({ ...R }, { [R.key]: a }) => {}
```

We can check if an identifier is referencing the rest param binding via `scope`, like what we did in
https://github.com/babel/babel/blob/548cb3ee89552ffc08ee5625b084bd33f8107530/packages/babel-plugin-transform-parameters/src/params.js#L26
```js
{
  "ReferencedIdentifier|BindingIdentifier"(path, state) {
      // state.scope is the function scope
      // path.scope is the scope of current traversed identifier
      scope.getBinding(name) === state.scope.getBinding(name) && // referencing the binding accessible from function scope
      state.scope.hasOwnBinding(name) && // the binding is owned by the function scope
      idsInRestParams.has(name) // it a binding introduced by the rest param
    }
}
```

And similar to #11349, if we are doing on this approach, we shall skip type annotations and type parameters.

vedantroy: Fixed. I swear every time I think I need to do something manually, Babel's path/visitor api proves me wrong and gives me a shorter solution.

PR #11327, opened 2020-03-24
Optional chaining propagate receiver when parenthesized

### LINKED ISSUE #10802: Optional Chaining does not propagate receiver when parenthesized
## Bug Report

**Current Behavior**

**Input Code**

```js

```

**Expected behavior/code**

- Filename: `babel.config.js`

```js
{
}
```

**Possible Solution**

iansu:
| Fixed Issues?            | Fixes #10802 
| Documentation PR Link    | 

### DIFF
--- a/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/call/output.js
+++ b/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/call/output.js
@@ -1,0 +1,8 @@
+"use strict";
+
+const foo = {
+  bar: function () {
+    console.log('hello');
+  }
+};
+(foo === null || foo === void 0 ? foo.bar.bind(foo) : foo.bar)();
--- a/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/call/output.js
+++ b/packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/call/output.js
@@ -1,8 +1,9 @@
 "use strict";
 
 const foo = {
-  bar: function () {
-    console.log('hello');
+  bar() {
+    return this;
   }
+
 };
 (foo === null || foo === void 0 ? foo.bar.bind(foo) : foo.bar)();
--- a/packages/babel-plugin-proposal-optional-chaining/src/index.js
+++ b/packages/babel-plugin-proposal-optional-chaining/src/index.js
@@ -25,6 +25,7 @@ export default declare((api, options) => {
       "OptionalCallExpression|OptionalMemberExpression"(path) {
         const { parentPath, scope } = path;
         let isDeleteOperation = false;
+        let isCallExpression = false;
         const optionals = [];
 
         let optionalPath = path;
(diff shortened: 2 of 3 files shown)


packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/call/output.js
```diff
@@ -0,0 +1,9 @@
+"use strict";
+
+const foo = {
+  bar() {
+    return this;
+  }
+
+};
+(foo === null || foo === void 0 ? foo.bar() : foo.bar)();
```
JLHwung: When `foo` is nullish, the conditional expression should return `undefined` instead of `foo.bar()`, otherwise we should return `foo.bar()`. One of transform results can be
```js
(foo === null || foo === void 0) ? (void 0)() : foo.bar()
```

Since optional chaining has been shipped in Chrome 80, one can also check the execution results between before and after.

```js
var foo = undefined
(foo?.bar)();
// TypeError: (intermediate value) is not a function
(foo === null || foo === void 0 ? foo.bar() : foo.bar)();
// TypeError: Cannot read property 'bar' of undefined
```

```js
var foo = {
 bar() { return this._bar },
 _bar: 42
}
(foo?.bar)();
// 42;
(foo === null || foo === void 0 ? foo.bar() : foo.bar)();
// undefined
```

nicolo-ribaudo:
@JLHwung @jridgewell How should we handle arguments, using the non-bind mode?

```js
foo?.bar(arg1, arg2(), { x: y }, function (err, result) {
  if (err) alert("error!");
  else alert("ok");
});
```

JLHwung:
> @JLHwung @jridgewell How should we handle arguments, using the non-bind output?
> 
> ```js
> foo?.bar(arg1, arg2(), { x: y }, function (err, result) {
>   if (err) alert("error!");
>   else alert("ok");
> });
> ```

I think we can pass through arguments to the happy path while leaving `(void 0)()` intact, assuming user can not tell the runtime difference between `(void 0)()` and `(void 0)(foo)`. This approach also prevents potentially bloated code size.

```js
(foo === null || foo === void 0) ? (void 0)() : foo.bar(arg1, arg2(), { x: y }, function (err, result) {
  if (err) alert("error!");
  else alert("ok");
})
```

nicolo-ribaudo:
Unfortunately, the difference can be observed. This code logs 2:

```js
try {
  (undefined?.foo)(console.log(2));
} catch {}
```

I think that we should use `bind` when we have arguments, and leave it out when the function either doesn't have any argument, or all its arguments are pure.

e.g.
```js
a?.b()         => (a === null || a === void 0 ? (void 0)() : a.b())      
a?.b(foo())    => (a === null || a === void 0 ? void 0 : a.b.bind(a))(foo())
a?.b(1, 2)     => (a === null || a === void 0 ? (void 0)() : a.b(1, 2))
```

jridgewell:
That's disappointing. I consider this to be such an anti-pattern that I think it's fine to just do the unoptimized `foo.bind(bar)` in all cases (even when we could better optimize it).

iansu:
I've switched this back to using `bind`.

PR #11330, opened 2020-03-25
Remove `output-file-sync`

JLHwung:
| Any Dependency Changes?  | Remove unused `output-file-sync`

### DIFF
--- a/packages/babel-plugin-transform-runtime/scripts/build-dist.js
+++ b/packages/babel-plugin-transform-runtime/scripts/build-dist.js
@@ -1,7 +1,7 @@
 "use strict";
 
 const path = require("path");
-const outputFile = require("output-file-sync");
+const fs = require("fs");
 const helpers = require("@babel/helpers");
 const babel = require("@babel/core");
 const template = require("@babel/template");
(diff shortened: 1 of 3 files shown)


packages/babel-plugin-transform-runtime/scripts/build-dist.js
```diff
@@ -12,6 +12,10 @@ const transformRuntime = require("../");
 const runtimeVersion = require("@babel/runtime/package.json").version;
 const corejs2Definitions = require("../lib/runtime-corejs2-definitions").default();
 const corejs3Definitions = require("../lib/runtime-corejs3-definitions").default();
+const outputFileSync = function(filePath, data) {
+  fs.mkdirSync(path.dirname(filePath), { recursive: true });
```
nicolo-ribaudo: This only supports node 10, I'm wondering why it doesn't file on Travis with node 6 and 8 :thinking:

JLHwung: Because it targets to Babel 8? 🤓

nicolo-ribaudo: 🙃

PR #11345, opened 2020-03-27
Ignore abstract methods when decorating class

### LINKED ISSUE #10514: preset-typescript + decorator on abstract base class overwrites members with undefined
## Bug Report

**Current Behavior**
Decorating an abstract base class causes implementors to have `undefined` function refs even though the function is implemented.

Given this scenario: 
```js
abstract class A {

  @decorate()
  private decoratedMember: boolean

  protected abstract myMethod(): void
}

class B extends A {
  private config: any;

  constructor(config: any) {
    super();
    this.config = config;
  }

  public runMyMethod(): void {
    this.myMethod(); // <--- Error: undefined is not a function
  }

  protected myMethod(): void {
   // Do stuff
  }
}

const b = new B();
b.runMyMethod();
```

**Expected behavior/code**
Base class definitions should not initialize abstract methods to `void 0` when decorators are used. This will allow implementors of abstract classes that use decorators to properly define concrete functions. 

**Babel Configuration **
.babelrc

```js
{
  "presets": [
    [
      "@babel/preset-env",
      {
        "targets": {
          "electron": "6"
        }
      }
    ],
    "@babel/preset-typescript"
  ],
  "ignore": ["**/*.d.ts"],
  "plugins": [
    "@babel/plugin-transform-react-jsx",
    [
      {
        "decoratorsBeforeExport": true
      }
    ],
  ]
}
```

**Environment**
- Babel version(s): 7.5.5
- Node/npm version: 12.7
- OS:  OSX 10.14.6
- Monorepo: lerna

oliverdunk:
| Fixed Issues?            | Fixes #10514 
| Documentation PR Link    |

### Description

Implementations of abstract methods were not callable on classes, when the abstract class being implemented contained a decorated field.

### The bug

When compiling a decorated class, we replace the classes definition with a call to a helper. This helper takes an array of objects, each representing something it needs to define in the class it builds.

To generate one of these objects, we take the node, and pass it through [extractElementDescriptor](https://github.com/babel/babel/blob/16cf27a762346638fdacd8611ec607212a6fb7fd/packages/babel-helper-create-class-features-plugin/src/decorators.js#L50). The object this function generates includes a `type` key, which describes which type of definition the helper needs to add. This is set using `isMethod ? node.kind : "field"`.

We were including abstract methods in the set of nodes that we were generating objects from. Since `isMethod` was false for these, we were telling the decorate helper to produce a field. This resulted in the abstract class defining a field, whose name overlapped with the method being implemented. As a result, the method could not be accessed.

### My fix

Abstract methods are filtered out, so the decorate helper no longer tries to define them.

### DIFF
--- a/packages/babel-helper-create-class-features-plugin/src/decorators.js
+++ b/packages/babel-helper-create-class-features-plugin/src/decorators.js
@@ -128,7 +128,9 @@ export function buildDecoratedClass(ref, path, elements, file) {
 
   const classDecorators = takeDecorators(node);
   const definitions = t.arrayExpression(
-    elements.map(extractElementDescriptor.bind(file, node.id, superId)),
+    elements
+      .filter(element => !element.node.abstract)
+      .map(extractElementDescriptor.bind(file, node.id, superId)),
   );
 
   let replacement = template.expression.ast`
--- a/packages/babel-helper-create-class-features-plugin/src/decorators.js
+++ b/packages/babel-helper-create-class-features-plugin/src/decorators.js
@@ -129,6 +129,7 @@ export function buildDecoratedClass(ref, path, elements, file) {
   const classDecorators = takeDecorators(node);
   const definitions = t.arrayExpression(
     elements
+      // Ignore TypeScript's abstract methods (see #10514)
       .filter(element => !element.node.abstract)
       .map(extractElementDescriptor.bind(file, node.id, superId)),
   );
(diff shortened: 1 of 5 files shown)


packages/babel-helper-create-class-features-plugin/src/decorators.js
```diff
@@ -128,7 +128,9 @@ export function buildDecoratedClass(ref, path, elements, file) {
 
   const classDecorators = takeDecorators(node);
   const definitions = t.arrayExpression(
-    elements.map(extractElementDescriptor.bind(file, node.id, superId)),
+    elements
+      .filter(element => !element.node.abstract)
```
nicolo-ribaudo: Maybe it's worth adding a comment here that this is needed for TS.

nicolo-ribaudo:
@oliverdunk Could you try rebasing and re-generating `output.js`? The tests are failing because we recently (7.9.0) modified the `_iterableToArray`, causing a difference in the output for this test.

oliverdunk:
> @oliverdunk Could you try rebasing and re-generating `output.js`? The tests are failing because we recently (7.9.0) modified the `_iterableToArray`, causing a difference in the output for this test.

No problem, done.

PR #11346, opened 2020-03-28
Set correct async/generator in IIFE for params

nicolo-ribaudo:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Current behavior:
```js
function* f(a = 1) {
  var a = yield;
}
```

```js
function* f() {
  let a = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : 1;
  return function (a) {
    var a = yield;
  }(a);
}
```

### DIFF
--- a/packages/babel-plugin-transform-parameters/test/fixtures/regression/scope-gen-async/output.js
+++ b/packages/babel-plugin-transform-parameters/test/fixtures/regression/scope-gen-async/output.js
@@ -1,0 +1,23 @@
+function f() {
+  let a = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : 1;
+  return function* (a) {
+    var a = yield a;
+    return a;
+  }(a);
+}
+
+function g() {
+  let a = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : 1;
+  return async function (a) {
+    var a = await a;
+    return a;
+  }(a);
+}
+
+function h() {
+  let a = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : 1;
+  return async function* (a) {
+    var a = await (yield a);
+    return a;
+  }(a);
+}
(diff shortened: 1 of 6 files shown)


packages/babel-plugin-transform-parameters/test/fixtures/regression/scope-gen-async/output.js
```diff
@@ -0,0 +1,23 @@
+function f() {
+  let a = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : 1;
+  return function* (a) {
+    var a = yield a;
+    return a;
+  }(a);
+}
+
+function g() {
+  let a = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : 1;
```
jridgewell: The default expression could actually throw a sync error, which needs to be caught sand turned into a rejected promise.

packages/babel-plugin-transform-parameters/test/fixtures/regression/scope-gen-async/output.js
```diff
@@ -0,0 +1,23 @@
+function f() {
+  let a = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : 1;
```
jridgewell: Similar problem here, I don’t think a generator would throw until the first `next` call. I may be wrong though.

nicolo-ribaudo: Generators throw immediatly (`(function* (x = a.b) {})()` throws)

nicolo-ribaudo:
@ulrichb This fixes your error about `await` outside of an async function.

ulrichb:
@nicolo-ribaudo Thank you very much!

nicolo-ribaudo:
The problem is that we don't have a place where we can put a `var _this = this` declaration.

PR #11349, opened 2020-03-28
Don't check type annotations when deciding params scope

### LINKED ISSUE #11344: Naming conflict in callback typing produces unnecessry complex state machine
## Bug Report

**Current Behavior**

**Input Code**

```ts
async function someAsync(
    factory: () => Promise<string>,
    options: {
        someCallback?: (theSameName: number) => void;
    } = {},
): Promise<string> {

    const theSameName = await factory();

    return theSameName;
}
```

[REPL Link](https://babeljs.io/en/repl#?browsers=ie%2011&build=&builtIns=false&spec=false&loose=false&code_lz=IYZwngdgxgBAZgV2gFwJYHsIxOgtgUwEFxoAKAKBivmCmXQCcwAuGUgShgF4A-GABQZ5UIfAB4QyBqggBzHgBpK1dAAc0mEKwDey6tRwEAwsAA2pgEa0A1gH5WpZAAt8AZWAEAch_ysICXAt8Bk5eGAA3dFQAEwBuPRgAX24YbUSldlZBYVEJKRl5VPIEqE1kGGc3H28CFOAAd2BUcrhaeiYOeISGfGQEBixK9y8feMSgA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Ctypescript&prettier=false&targets=&version=7.9.0&externalPlugins=)

**Expected behavior/code**

When the argument of the `someCallback` _typing_ has the same name as the local in the body, then the generated state machine becomes weird (two nested `regeneratorRuntime.wrap()`). See the following text comparison of the REPL output:

image

Expected: No difference when the argument name in the callback _typing_ is the same as a local.

Just the TypeScript preset, see REPL link above.

**Additional info**

This is a follow-up issue of https://github.com/babel/babel/issues/11336#issuecomment-605075901, where I get _"Module parse failed: Cannot use keyword 'await' outside an async function"_ errors in the CRA Babel preset.

nicolo-ribaudo:
| Fixed Issues?            | Fixes #11344
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Type annotations caused unnecessary usage of IIFE, because we checked if they referenced any value binding.

### DIFF
--- a/packages/babel-plugin-transform-parameters/src/params.js
+++ b/packages/babel-plugin-transform-parameters/src/params.js
@@ -37,9 +37,8 @@ const iifeVisitor = {
     }
   },
   // type annotations don't use or introduce "real" bindings
-  "TypeAnnotation|TSTypeAnnotation"(path) {
-    path.skip();
-  },
+  "TypeAnnotation|TSTypeAnnotation|TypeParameterDeclaration|TSTypeParameterDeclaration": path =>
+    path.skip(),
 };
 
 export default function convertFunctionParams(path, loose) {
--- a/packages/babel-plugin-transform-parameters/src/params.js
+++ b/packages/babel-plugin-transform-parameters/src/params.js
@@ -36,6 +36,10 @@ const iifeVisitor = {
       path.stop();
     }
   },
+  // type annotations don't use or introduce "real" bindings
+  "TypeAnnotation|TSTypeAnnotation"(path) {
+    path.skip();
+  },
 };
 
 export default function convertFunctionParams(path, loose) {
--- a/packages/babel-plugin-transform-parameters/test/fixtures/regression/11344-flow/input.js
+++ b/packages/babel-plugin-transform-parameters/test/fixtures/regression/11344-flow/input.js
@@ -1,3 +1,7 @@
 function a(b: (c) => void = {}) {
   let c;
 }
+
+function d(e = <T>() => {}) {
+  let T;
+}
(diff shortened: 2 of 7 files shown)


packages/babel-plugin-transform-parameters/src/params.js
```diff
@@ -36,6 +36,10 @@ const iifeVisitor = {
       path.stop();
     }
   },
+  // type annotations don't use or introduce "real" bindings
+  "TypeAnnotation|TSTypeAnnotation"(path) {
```
JLHwung: Ideally we may expose `bindingType` of `checkLVal` to `parseBindingList` so we do not register bindings of lval when they are inside a type annotation.

BTW we should also skip TypeParameters.
```ts
function a(b = <T extends (c: any) => void>(): void => {}) {
  let c;
}
```

nicolo-ribaudo: Since identifiers are the most often created/moved nodes, I think that it would be safer to add a `skip` like this one to `@babel/traverse` rather than relying on what was initially parsed.

packages/babel-plugin-transform-parameters/test/fixtures/regression/11344-flow/input.js
```diff
@@ -0,0 +1,7 @@
+function a(b: (c) => void = {}) {
+  let c;
+}
+
+function d(e = <T>() => {}) {
```
JLHwung: nit: This test will be passing without current change since we are not registering bindings when parsing type parameters.

PR #11354, opened 2020-04-01
fix auto import error with whitespace JSXText

lunaruan:
The new JSX auto import React has a bug where it ignores valid whitespace text when calculating whether to import `jsx` or `jsxs`. This fixes it by stripping JSXText the same way that `t.react.buildChildren` strips JSXText so that the children calculated are consistent with each other.

### DIFF
--- a/packages/babel-helper-builder-react-jsx-experimental/src/index.js
+++ b/packages/babel-helper-builder-react-jsx-experimental/src/index.js
@@ -303,12 +303,8 @@ You can set \`throwIfNamespace: false\` to bypass this warning.`,
       "JSXElement|JSXFragment"(path) {
         if (path.type === "JSXFragment") imports.add("Fragment");
         const openingPath = path.get("openingElement");
-        const validChildren = openingPath.parent.children.filter(
-          child =>
-            !t.isJSXEmptyExpression(child) &&
-            !(t.isJSXText(child) && child.value.trim() === ""),
-        );
 
+        const validChildren = t.react.buildChildren(openingPath.parent);
         let importName;
         if (path.type === "JSXElement" && shouldUseCreateElement(path)) {
           importName = "createElement";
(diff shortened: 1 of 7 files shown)


packages/babel-helper-builder-react-jsx-experimental/src/index.js
```diff
@@ -303,12 +303,15 @@ You can set \`throwIfNamespace: false\` to bypass this warning.`,
       "JSXElement|JSXFragment"(path) {
         if (path.type === "JSXFragment") imports.add("Fragment");
         const openingPath = path.get("openingElement");
-        const validChildren = openingPath.parent.children.filter(
-          child =>
-            !t.isJSXEmptyExpression(child) &&
-            !(t.isJSXText(child) && child.value.trim() === ""),
-        );
-
+        const validChildren = openingPath.parent.children.filter(child => {
```
JLHwung, line 306: Can we reuse [`t.react.buildChildren`](https://github.com/babel/babel/blob/master/packages/babel-types/src/builders/react/buildChildren.js)?

lunaruan: `t.react.buildChildren` mutates the node. I can refactor it to return a value instead of mutating if that's okay with you.

JLHwung: > t.react.buildChildren mutates the node.

Oh really? I thought it is side-effect free. I will double check later.

> I can refactor it to return a value instead of mutating if that's okay with you.

I am okay with that if it turns out we does not depend on its mutation behaviors. BTW `t.react.buildChildren` is undocumented so it is likely only used by our plugins.

lunaruan: Ah actually maybe you're right. Let me update this diff

PR #11361, opened 2020-04-01
fix: do not add function name when parent's operator is not `=`

JLHwung:

Per https://tc39.es/ecma262/#sec-assignment-operators-runtime-semantics-evaluation, when operator is not `=`, the anonymous function expression will not be applied with NamedEvaluation.

### DIFF
--- a/packages/babel-helper-function-name/src/index.js
+++ b/packages/babel-helper-function-name/src/index.js
@@ -179,7 +179,7 @@ export default function({ node, parent, scope, id }, localBinding = false) {
         return;
       }
     }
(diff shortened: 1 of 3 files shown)


nicolo-ribaudo:
Does this fix https://github.com/babel/babel/issues/11362?

JLHwung:
@nicolo-ribaudo Not really. The issue 11362 refers to the transform plugin, which converts `f ??= function() {}` into `f ?? (f = function() {})` so the `.name` is not preserved.

Theoretically this PR fixes cases of `transform-function-name` + `syntax-logical-assignment-operators`.

PR #11370, opened 2020-04-02
logical-assignment: Do not assign names to anonymous functions

### LINKED ISSUE #11362: `proposal-logical-assignment` should not assign anonymous function name
## Bug Report

**Current Behavior**
The following snippet, after transform will print string `"f"` to the console

**Input Code**
- [REPL](https://babeljs.io/repl#?browsers=chrome%2080&build=&builtIns=false&spec=false&loose=false&code_lz=G4QwTgBAZg3AUACihA_CgvNArgOwMYAuAlgPY4ICUEA3gL4UB0OIAtgKZA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=script&lineWrap=false&presets=env%2Cstage-1&prettier=false&targets=&version=7.9.0&externalPlugins=%40babel%2Fplugin-syntax-class-properties%407.8.3)

```js
var f;
console.log((f ??= function() {}).name)
```

**Expected behavior/code**
According to https://tc39.es/proposal-logical-assignment/#sec-assignment-operators-runtime-semantics-evaluation, the anonymous function is not applied with NamedEvaluation, so it should print an empty string to the console.

- Filename: `babel.config.js`

```js
{
  "plugins": ["@babel/proposal-logical-assignment-operators"]
}
```

**Environment**
REPL.

**Solution**
Transform `f ??= function() {}` to `f = f ?? function() {}` may fix this issue. Or we leave a caveat note that we are not supporting function name in this plugin.

arku:
| Fixed Issues?            | Fixes #11362
| Documentation PR Link    | 
| Any Dependency Changes?  |


Andarist:
A bunch of stuff can be introduced to make the function anonymous, the easiest ones would be:
1. unwrap from array
```js
var _f;

var f;
((_f = f) !== null && _f !== void 0 ? _f : f = [function () {}][0]).name;
```
2. introduce a helper
```js
function ensureAnonymous(fn) { return fn }
var _f;

var f;
((_f = f) !== null && _f !== void 0 ? _f : f = ensureAnonymous(function () {})).name;
```

Both of those are valid es5 so I would consider them being better alternatives. I don't think that all current transforms (but I might be wrong) are that much concerned about `.name`, it's more of a "best effort" from what I remember. So maybe it's not worth doing crazy stuff to cover this? Or maybe a crazy hack could be hidden behind `spec: true` option?

packages/babel-plugin-proposal-logical-assignment-operators/src/index.js
```diff
@@ -39,13 +39,26 @@ export default declare(api => {
...
-          t.logicalExpression(
-            operator.slice(0, -1),
-            lhs,
-            t.assignmentExpression("=", left, right),
-          ),
-        );
+        const isRHSAnonymousFunction = t.isFunctionExpression(right, {
+          id: null,
+        });
+        if (isRHSAnonymousFunction && operator === "??=") {
```
jridgewell, line 45: We need to apply this to all the operators, not just `??=`.

packages/babel-plugin-proposal-logical-assignment-operators/src/index.js
```diff
@@ -39,13 +39,26 @@ export default declare(api => {
...
+          path.replaceWith(
+            t.assignmentExpression(
+              "=",
```
jridgewell, line 48: Doing `a = a ?? b` breaks the semantics of the operator.

jridgewell:
https://github.com/tc39/proposal-logical-assignment/issues/23#issuecomment-608700711 mentions that we can also use a `SequenceExpression` to do this:

```js
// input
foo ??= function() {};

// output
foo ?? (foo = (0, function() {}));

File: packages/babel-plugin-proposal-logical-assignment-operators/src/index.js (hunk 1 above)
Comment (arku) [reply packages/babel-plugin-proposal-logical-assignment-operators/src/index.js:45]: Ah, yeah.

**Comment by arku** [comment]
@jridgewell I have addressed the suggested changes. Do you mind taking another look?

File: packages/babel-plugin-proposal-logical-assignment-operators/src/index.js [hunk 3]
```diff
@@ -39,11 +39,18 @@ export default declare(api => {
           }
         }
 
+        const isRHSAnonymousFunction = t.isFunctionExpression(right, {
```
nicolo-ribaudo, line 42: We should also do it for arrow functions:
```js
a ??= () => {};
```
shouldn't get an implicit name.

packages/babel-plugin-proposal-logical-assignment-operators/test/fixtures/logical-assignment/anonymous-function/input.js
```diff
@@ -0,0 +1,8 @@
+var a;
+console.log((a ||= function () {}).name);
+
+var b = 1;
+console.log((b &&= function () {}).name);
```
nicolo-ribaudo, line 5: Since this is just an input/output test, we don't need the "runtime logic". This should be enough:

```js
a ||= function () {};
a &&= function () {};
a ??= function () {};
```

Also, can you add:
1. Another test folder for arrow functions
1. Another test folder for named functions
1. One `exec.js` file in each of those three folders, to test the _runtime_ behaviors:

```js
var a, b = true, c;

a ||= function () {};
b &&= function () {};
c ??= function () {};

expect(a.name).toBe("");
expect(b.name).toBe("");
expect(c.name).toBe("");
```

arku: @nicolo-ribaudo Cool, I didn't know about the assertion type of tests. Let me update my PR.

packages/babel-plugin-proposal-logical-assignment-operators/src/index.js
arku: Good catch! Let me fix it.

packages/babel-plugin-proposal-logical-assignment-operators/test/fixtures/logical-assignment/anonymous-functions-exec/options.json
```diff
@@ -0,0 +1,6 @@
+{
+  "plugins": [
+    "proposal-logical-assignment-operators",
+    "proposal-nullish-coalescing-operator"
```
nicolo-ribaudo, line 4: Note for reviewers: the "exec" and "transform" tests are in different folders because they have different configs.

packages/babel-plugin-proposal-logical-assignment-operators/src/index.js
```diff
@@ -39,11 +39,19 @@ export default declare(api => {
           }
         }
 
+        const isRHSAnonymousorArrowFunction =
+          t.isFunctionExpression(right, {
```
jridgewell, line 43: Nit: Can't this just be `t.isFunction`? That should capture both arrow and regular function expressions.

arku: @jridgewell My bad. Missed this comment.

jridgewell:
@babel/babel Any idea what's going on with the tests?

nicolo-ribaudo:
@JLHwung It is related to `exports`, maybe you know :thinking:

JLHwung:
Rebased on upstream. The exports error should be fixed in recent releases.

PR #11372, opened 2020-04-03
Add private-property-in-object support

jridgewell:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

https://github.com/tc39/proposal-private-fields-in-in

The parsing is a little loose, but it's going to take a lot of effort to restrict it to only `in` binary expressions.


packages/babel-parser/src/parser/expression.js
```diff
@@ -1138,6 +1138,9 @@ export default class ExpressionParser extends LValParser {
           this.registerTopicReference();
           return this.finishNode(node, "PipelinePrimaryTopicReference");
         }
+
+        this.expectPlugin("privateIn");
+        return this.parseMaybePrivateName(true);
```
nicolo-ribaudo, line 1143: Can we add
```js
if (!this.match(tt._in)) this.raise
```
?

JLHwung: And we should also check if the private name is parenthesized.
```js
(#p) in C
```
seems invalid.

jridgewell: Done and done.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -100,6 +100,34 @@ const privateNameVisitor = {
...
+    }
+
+    const { id, static: isStatic } = privateNamesMap.get(name);
+
+    if (isStatic) {
+      path.replaceWith(template.expression.ast`${right} === ${this.classRef}`);
+      return;
+    }
+
+    path.replaceWith(template.expression.ast`${id}.has(${right})`);
```
ljharb, line 127: just curious; do these field WeakMaps have `WeakMap.prototype.has` copied onto them? or can i intercept babel private fields, even in spec mode, by replacing `WeakMap.prototype.has`?

jridgewell: We do not secure the weak
maps, so this will do a prototype reference. Note that even `get` and `set` calls go through the prototype, too. We can't ensure the order of our evaluation, so the `WeakMap` could have already been patched by the time we reach the class definition.

ljharb: Of course - you can't protect against first-run code, but to be spec compliant you'd have to at least protect against later code. `Object.defineProperties(id, Object.getOwnPropertyDescriptors(WeakMap.prototype))` at `id` creation would address that.

ljharb:
@JLHwung I would assume `obj.#constructor` is a syntax error, so that `#constructor in x` should be as well?

packages/babel-helper-create-class-features-plugin/src/fields.js
nicolo-ribaudo: Currently, all of our output assumes that builtins haven't been modified.

nicolo-ribaudo:
I think that for `#constructor` we will already throw when handling private names not in scope, because since it can't be declared it will ever be in scope.

packages/babel-parser/src/parser/location.js
```diff
@@ -125,6 +125,7 @@ export const Errors = Object.freeze({
     "Topic reference was used in a lexical context without topic binding",
   PrimaryTopicRequiresSmartPipeline:
     "Primary Topic Reference found but pipelineOperator not passed 'smart' for 'proposal' option.",
+  PrivateInExpectedIn: "Expected `in` keyword to follow private name",
```
nicolo-ribaudo, line 128: Maybe we could also suggest how it should be used?
```suggestion
  PrivateInExpectedIn: "Private names are only allowed in property accesses (`obj.#%0`) or in `in` expressions (`#%0 in obj`)",
```

packages/babel-parser/test/fixtures/experimental/class-private-methods/asi-failure-generator/options.json
```diff
@@ -1,4 +1,7 @@
 {
-  "throws": "Unexpected token (3:3)",
-  "plugins": ["classProperties", "classPrivateMethods"]
-}
+  "throws": "This experimental syntax requires enabling the parser plugin: 'privateIn' (3:3)",
```
nicolo-ribaudo, line 2: Can we throw this error only if the token after `#foo` is actually `in`?

ljharb:
I've just added an early error for `#constructor` to [the proposed spec](https://github.com/tc39/proposal-private-fields-in-in/commit/b798be485fc84e392b10903278d01da5a0bc33b4); thanks for the ping :-)

packages/babel-helper-create-class-features-plugin/src/fields.js
ljharb: oof, ok - it's totally fine to assume they haven't been modified at module evaluation time (you don't really have a choice) but it's *not* fine to assume they haven't been modified later. that should be fixed asap (but not in this PR, obviously)

packages/babel-parser/src/parser/expression.js
```diff
@@ -1138,6 +1138,19 @@ export default class ExpressionParser extends LValParser {
           this.registerTopicReference();
           return this.finishNode(node, "PipelinePrimaryTopicReference");
         }
+
+        if (this.hasPlugin("privateIn")) {
+          node = this.parseMaybePrivateName(true);
+          this.classScope.usePrivateName(node.id.name, node.start);
```
JLHwung, line 1144: Can you add a new test case for `#constructor in c`? We check `#constructor` only in `parseClassPropertyName`.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -100,6 +100,34 @@ const privateNameVisitor = {
...
+
+    if (loose) {
+      const { id } = privateNamesMap.get(name);
+      path.replaceWith(template.expression.ast`
+        Object.prototype.hasOwnProperty.call(${right}, ${id})
+      `);
+      return;
+    }
+
+    const { id, static: isStatic } = privateNamesMap.get(name);
```
JLHwung, line 120: The private name is not necessarily defined in current class scope.
```js
class A {
  #p;
  B = class B {
    #q;
    m() { #q in B }
  }
}
```
When traversing into `#q in B` while we are still at `class A`, `privateNamesMap.has("q")` will return `undefined`.

---

It can be even trickier given that private name can scatter in nested class.
```js
class B {
  static #p = 42;
  static C = class C {
    static #p = 42;
    static pIn(obj) {
      return obj.#p;
    }
  }
  static pIn (obj) {
    return B.C.pIn(obj)
  }
}
```
Intuitively both `B.pIn(B)` and `B.pIn(B.C)` should return `true` since both `B` and `B.C` has declared `#p`.

ljharb: In the latter case, `#p in B` would be `false`, since each `#p` is a different one - in other words, inside `class C`, there is *no access available* to B's `#p`, because C's `#p` shadows it.

jridgewell: Ooooo, this is an excellent catch. It's actually unhandled in the overall private fields visitor! I'm going to open a specific PR for it, and then I'll rebase that into here.

jridgewell: This now depends on https://github.com/babel/babel/pull/11405, which we should merge first.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1158,7 +1171,11 @@ export default class ExpressionParser extends LValParser {
     const isPrivate = this.match(tt.hash);
 
     if (isPrivate) {
-      this.expectOnePlugin(["classPrivateProperties", "classPrivateMethods"]);
```
JLHwung, line 1161: I think throwing missing `privateIn` error for the common private property case will confuse users.
```js
class C {
  #p = 42;
}
```

Instead we can expect `privateIn` when we see `tt._in` after a private name.

jridgewell: I don't understand the suggestion. I need to expect one of the 3 plugins here, or else I won't be able to parse the private name.

JLHwung: I just realize that this line is never failed due to https://github.com/babel/babel/pull/11478/files#diff-80da704a8d2fc8d339ab57f154fb225eL429, which I address in #11478.

Base on #11478 and current implementation, if users are enabling `classPrivateProperties` but without `privateIn`, the following case does not throw missing `privateIn` error. ([REPL](https://babeljs.io/repl/build/21259/#?browsers=node%206&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEDC0G8BQ1oGIAO0C80CMA3CtALYAUAlItAE4CmALgK40B26WAluwwBacwAvkkFA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=script&lineWrap=false&presets=env%2Cstage-3%2Cenv&prettier=true&targets=Node-13&version=7.9.0%2Bpr.11372&externalPlugins=))
```js
class C {
  #p = 42;
  m() { return #p in this }
}
```
because we never really check `privateIn` as long as `classPrivateProperties` is enabled.

Can you add a test case to `packages/babel-parser/test/fixtures/experimental/_no-plugin`?

jridgewell: We check at https://github.com/babel/babel/pull/11372/files#diff-56fea09eea2de51b8291db400d2f69beR1142. Added a test, we get `SyntaxError: Unexpected token` at the `#` in `#foo in {}`.

JLHwung: Can we throw `the privateIn plugin is expected` when we see `#foo in {}` but `privateIn` is missing? Just like we [did](https://babeljs.io/repl/build/21296/#?browsers=node%2013&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEDC0G8BQ1oGIAO0C80CMA3EgL5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Ctypescript%2Cenv&prettier=true&targets=&version=7.9.0%2Bpr.11478&externalPlugins=) for `classPrivateProperties`.

jridgewell: There are several conflicting desires here:

1. Parsing smart pipelines, and giving good errors
2. Parsing invalid `#foo` expressions (not followed by `in`)
3. Parsing `#foo in {}` expressions, but not having plugin enabled

I don't have a good solution to all three of these. If you can push a commit, that'd be awesome.

JLHwung: Pushed a commit that should address 2) and 3).

For 1) we are printing good errors when `|>` is seen but no `pipeline` plugin. I think current situation is acceptable and since they are competing proposals we can give better errors after any of them advances.

jridgewell:
Excellent, thanks for the help!

nicolo-ribaudo:
@jridgewell Could you add a docs page for this new plugin? You can copy one of the existing pages for other simple plugins.

jridgewell:
https://github.com/babel/website/pull/2235

packages/babel-parser/src/parser/expression.js
```diff
@@ -1138,6 +1139,25 @@ export default class ExpressionParser extends LValParser {
           this.registerTopicReference();
           return this.finishNode(node, "PipelinePrimaryTopicReference");
         }
+
+        if (isIdentifierStart(this.input.codePointAt(this.state.end))) {
```
JLHwung, line 1143: This is insufficient as the first character of an identifier could be a unicode escape sequence.
```js
class Foo {
  #a;
  static isFoo(obj) {
    return #\u{61} in obj;
  }
}
```

packages/babel-parser/test/fixtures/experimental/private-in/private-in-escaped-sequence/output.json
```diff
@@ -0,0 +1,102 @@
+{
+  "type": "File",
+  "start":0,"end":68,"loc":{"start":{"line":1,"column":0},"end":{"line":6,"column":1}},
+  "errors": [
+    "SyntaxError: Invalid Unicode escape (2:3)",
```
JLHwung, line 5: 🤦‍♂️That's too bad: Emoji can't be a valid identifier name.

nicolo-ribaudo:
Note: this is not merged to `master` yet.

PR #11373, opened 2020-04-03
Fixed useBuiltIns and modules validation when using 'false' as option

### LINKED ISSUE #11352: Invalid warning for not declaring 'corejs' option when using 'useBuiltIns': 'false'
## Bug Report

**Current Behavior**
When using `@babel/preset-env` and setting `useBuiltIns: 'false'` and not declaring a `corejs` option, the following warning occurs:

> WARNING: We noticed you're using the `useBuiltIns` option without declaring a 
        core-js version. Currently, we assume version 2.x when no version 
        is passed. Since this default version will likely change in future 
        versions of Babel, we recommend explicitly setting the core-js version 
        you are using via the `corejs` option.\n" +
        You should also be sure that the version you pass to the `corejs` 
        option matches the version specified in your `package.json`'s 
        `dependencies` section. If it doesn't, you need to run one of the 
        following commands:
        npm install --save core-js@2    npm install --save core-js@3
         yarn add core-js@2              yarn add core-js@3

First, I don't know if this should be considered a bug, because the problem occurs only when setting the `useBuiltIns` option to the string `'false'` and not `false`. But if you set it accidentaly to a string value I think It should behave the same as when setting it to a `boolean` one.

**Expected behavior/code**
The warning should not occur.

- Filename: `babel.config.js`

```js
{
  presets: [
    "@babel/env",
    {
      useBuiltIns: "false"
    }
  ]
}
```

**Environment**

- Babel version: 7.9.0
- Babel preset-env version: 7.9.0
- Node/npm version: 10.17.0
- OS: Ubuntu 18.04.3 LTS
- How you are using Babel: [loader]

**Possible Solution**
Upon inspecting the `@babel/preset-env` source code, in the file `normalize-options.js` in the function `normalizeCoreJSOption` the following check is used 

``if (useBuiltIns && corejs === undefined) {``

Using a more thorough check for `useBuiltIns` will fix the warning.

JMarkoski:
| Fixed Issues?            | Fixes #11352 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Fixes the case when 'false' as string is provided for `useBuiltIns` or `modules` option by not converting the supplied option to string when it's 'false', and comparing it directly.

### DIFF
--- a/packages/babel-preset-env/src/normalize-options.js
+++ b/packages/babel-preset-env/src/normalize-options.js
@@ -167,8 +167,7 @@ export const validateModulesOption = (
   modulesOpt: ModuleOption = ModulesOption.auto,
 ) => {
   invariant(
-    ModulesOption[modulesOpt.toString()] ||
-      ModulesOption[modulesOpt.toString()] === ModulesOption.false,
+    ModulesOption[modulesOpt.toString()] || modulesOpt === ModulesOption.false,
     `Invalid Option: The 'modules' option must be one of \n` +
       ` - 'false' to indicate no module processing\n` +
       ` - a specific module type: 'commonjs', 'amd', 'umd', 'systemjs'` +
@@ -184,7 +183,7 @@ export const validateUseBuiltInsOption = (
 ) => {
   invariant(
     UseBuiltInsOption[builtInsOpt.toString()] ||
-      UseBuiltInsOption[builtInsOpt.toString()] === UseBuiltInsOption.false,
+      builtInsOpt === UseBuiltInsOption.false,
     `Invalid Option: The 'useBuiltIns' option must be either
     'false' (default) to indicate no polyfill,
     '"entry"' to indicate replacing the entry polyfill, or
--- a/packages/babel-preset-env/test/normalize-options.spec.js
+++ b/packages/babel-preset-env/test/normalize-options.spec.js
@@ -6,6 +6,7 @@ const {
   checkDuplicateIncludeExcludes,
   validateBoolOption,
   validateModulesOption,
+  validateUseBuiltInsOption,
   normalizePluginName,
 } = normalizeOptions;
 describe("normalize-options", () => {
@@ -242,9 +243,35 @@ describe("normalize-options", () => {
       }).toThrow();
     });
 
+    it("`'false'` option is invalid", () => {
+      expect(() => {
+        validateModulesOption("false");
+      }).toThrow();
+    });
+
     it("array option is invalid", () => {
       expect(() => {
         validateModulesOption([]);
+      }).toThrow();
+    });
+  });
+
+  describe("validateUseBuiltInsOptions", () => {
+    it("usage option is valid", () => {
+      expect(validateUseBuiltInsOption("usage")).toBe("usage");
+    });
+
+    it("entry option is valid", () => {
+      expect(validateUseBuiltInsOption("entry")).toBe("entry");
+    });
+
+    it("`false` option returns false", () => {
+      expect(validateUseBuiltInsOption(false)).toBe(false);
+    });
+
+    it("`'false'` option is invalid", () => {
+      expect(() => {
+        validateUseBuiltInsOption("false");
       }).toThrow();
     });
   });


existentialism:
@JMarkoski thanks for the PR! can we add a test for this [here](https://github.com/babel/babel/blob/master/packages/babel-preset-env/test/normalize-options.spec.js#L210)?

JMarkoski:
@existentialism Thanks, I'm very happy to contribute!  I've added the missing tests for  `validateUseBuiltInsOption` and for `validateModulesOption` when the input is 'false'. I think it's okay now.

existentialism:
@JMarkoski great job!

bencripps:
Is this considered a breaking change? 

I notice that for our application, we now have a bunch of broken builds because we aren't setting this property explicitly. 

Even more troublesome is that for dependencies that we need to transpile, we also run into this issue because their local babel-config files take precedence over the application specific configurations.

In our builds we now see the following errors:
```
   Invalid Option: The 'modules' option must be one of
     - 'false' to indicate no module processing
     - a specific module type: 'commonjs', 'amd', 'umd', 'systemjs' - 'auto' (default) which will automatically select 'false' if the current
       process is known to support ES module syntax, or "commonjs" otherwise```

Our babel config looks as follows:

```js
module.exports = {
    presets: [['@babel/preset-env', { modules: true }], '@babel/preset-react']
};
```

JLHwung:
Per https://babeljs.io/docs/en/babel-preset-env#modules, preset-env never accepts `true` as a valid option. As the error message is shown, please specify the module type that your app targets to.

Besides, the only behaviour change introduced in this PR is throwing on `modules: "false"`, which is also never accepted but in previous versions it is processed as `modules: false`, this is considered a bug fix.

Should you have more questions, please submit a new issue. Thanks.

JMarkoski:
Also, just to add, you say that you don't set the option explicitly, but in the configuration you set it to `{modules: true}`. If you don't set the `modules` option explicitly it will automatically select 'false' if the current process is known to support ES module syntax, or `'commonjs'` otherwise, as the error says.

bencripps:
Eee sorry, that was a mistype, from me trying to debug the issue. Our actual config is:

```js
module.exports = {
    presets: ['@babel/preset-env', '@babel/preset-react']
};```

Previously we were on @babel/preset-env@7.8.7 -- when we upgraded to @babel/preset-env@7.9.5 we started seeing the error above.

**Comment by JLHwung** [comment]
```js
module.exports = {
    presets: ['@babel/preset-env', '@babel/preset-react']
};
```
In the config above, `preset-env` will assume `modules: "auto"` and I can not reproduce the error you mentioned.

> we also run into this issue because their local babel-config files take precedence over the application specific configurations.

Can you share their local babel config files? Or better can you construct a reproduction repo? This can greatly help investigate the issues.

bencripps:
Their local file looks as follows:

```js
{
    "presets": ["es2015", "react", "stage-2"],
    "compact": false,
    "env": {
      "test": {
        "plugins": ["transform-runtime"],
        "sourceMaps": "both"
      }
    }
}
```

I can try to construct a reproducible repo. Curious is the assumption of "auto" a new behavior, or has it always been this way?

bencripps:
It's also interesting to note, that if I comment out the line check in `validateModulesOption`, found in the newest @babel/preset-env the build succeeds, without any compilation errors, which makes me think we don't have a code issue, and it's just this check which is throwing.

```
const validateModulesOption = (modulesOpt = _options.ModulesOption.auto) => {
  // comment out the validation 
  // (0, _invariant.default)(_options.ModulesOption[modulesOpt.toString()] || modulesOpt === _options.ModulesOption.false, `Invalid Option: The 'modules' option must be one of \n` + ` - 'false' to indicate no module processing\n` + ` - a specific module type: 'commonjs', 'amd', 'umd', 'systemjs'` + ` - 'auto' (default) which will automatically select 'false' if the current\n` + `   process is known to support ES module syntax, or "commonjs" otherwise\n`);
  return modulesOpt;
};
```

bencripps:
I can confirm that the issue doesn't happen in our source code, and that the error only happens because one of our dependencies is relying on a previous version of the babel libraries (as you can see from the above config).

Is it possible that this change causes an unexpected issue for our (admittedly odd) use-case? 

Unfortunately, we don't have access to change their code to use a newer version of babel.

JMarkoski:
> Curious is the assumption of "auto" a new behavior, or has it always been this way?

I can confirm that `@babel/preset-env@7.8.7` and `@babel/preset-env@7.9.5` both have the `'auto'` option.

Apart from that, I don't know that might be the issue, can you construct a reproduction repo to investigate further?

bencripps:
I've tried to build a minimally reproducible version but it's tough since many of the packages (including dev dependencies) are private.

However, in my debugging to see what "value" is causing the problem, I updated the error statement to include some extra information. It looks like it's throwing, but that the value it's getting is correct? It seems to be "false", of type string, which should be allowed? 

```bash
Invariant Violation: [BABEL] /Users/bcripps/indeed/mosaic-provider-base-nodejs/node_modules/@indeed/frontend-components-react/components/Modal/index.js: Invalid Option:
    ArgValue: false
    TypeOf Arg: string 
    The 'modules' option must be one of
     - 'false' to indicate no module processing
     - a specific module type: 'commonjs', 'amd', 'umd', 'systemjs' - 'auto' (default) which will automatically select 'false' if the current
       process is known to support ES module syntax, or "commonjs" otherwise
```

JMarkoski:
Then this is the expected behavior. This PR fixes exactly this issue, where you specify `'false'` as the option which is `string` instead of a boolean `false`.

bencripps:
Ah interesting! When I add this

```js
  console.log('typeof moduleOpt ' +  typeof modulesOpt)
```

I see

```bash
typeof moduleOpt string
typeof moduleOpt string
typeof moduleOpt string
typeof moduleOpt string
typeof moduleOpt string
typeof moduleOpt string
typeof moduleOpt string
typeof moduleOpt string
```

but none of our configs specify this parameter at all? Do you know what might be converting it to "false"?

bencripps:
I found it! We have some (🤦) internal build code that's injecting a string value :/ 

Thank you so much for your help in debugging this!

JMarkoski:
Great! You are welcome, I am very glad if I helped.

PR #11375, opened 2020-04-03
Fixed generateUid creating clashing ids after scope re-crawling

### LINKED ISSUE #11057: @babel/plugin-transform-block-scoping breaks when scope.crawl() is used by another plugin
## Bug Report

Hello, I've found some compatibility issues between `@babel/plugin-transform-block-scoping` and two community plugins (`babel-plugin-transform-react-remove-prop-types`, `babel-plugin-remove-debug`). The common thing between these two plugins is that they are using `scope.crawl()` method. When I comment out that method from `babel-plugin-transform-react-remove-prop-types` the bug seems to disappear. I've read parts of the babel-handbook but didn't find documentation for the crawl method. What does it do?

I've prepared a minimal repository for reproducing this issue with custom plugin where `scope.crawl()` is called.
```json
{
    "plugins": [
        "@babel/plugin-transform-block-scoping",
        "./babel-plugin-tester"
    ]
}
```

Note that `@babel/plugin-transform-block-scoping` is included in preset-env, which is used by create-react-app. In my case this created a bug into production code which was not present in development and test code. In my opinion this is quite dangerous, especially if apps are missing proper integration tests. Developers can't trust their unit tests and development mode functionality.

I'm just getting to know babel's implementation so let's leave this unchecked for now.

**Current Behavior**

This code block is transformed incorrectly:
```js
// Input
case 'B': {
    const { unicorn } = action;
    const unicorns = state.unicorns.map(_unicorn =>
        _unicorn.id === unicorn.id ? unicorn : _unicorn
    );

    return { ...state, unicorns, someCaseBAttributes: true };
}
```
Note how `var { unicorn: _unicorn } = action;` and `map(_unicorn` have conflicting variable names.
```js
// Output
case 'B': {
    var { unicorn: _unicorn } = action;

    var _unicorns = state.unicorns.map(_unicorn =>
        _unicorn.id === _unicorn.id ? _unicorn : _unicorn
    );

    return { ...state, unicorns: _unicorns, someCaseBAttributes: true };
}
```

When used with webpack this makes terser go crazy:
```js
case 'B':
    var { unicorn: _unicorn } = action,
        _unicorns = state.unicorns.map(
            _unicorn => (_unicorn.id, _unicorn.id, _unicorn)
        );
    return { ...state, unicorns: _unicorns, someCaseBAttributes: !0 };
```

**Input Code**
https://github.com/AriPerkkio/babel-preset-react-app-transpile-bug
```bash
$ yarn # Install deps
$ yarn babel # Runs babel on index.js
$ yarn terser # Runs terser on babel-output.js
$ yarn test # Runs test case against original index.js and dist/terser-output.js
```

**Expected behavior/code**

Variable names should not conflict. When `scope.crawl()` is not called, output is as expected:
```js
case 'B': {
    var { unicorn: _unicorn2 } = action;

    var _unicorns = state.unicorns.map(_unicorn =>
        _unicorn.id === _unicorn2.id ? _unicorn2 : _unicorn
    );

    return { ...state, unicorns: _unicorns, someCaseBAttributes: true };
}
```

Any configuration with `@babel/plugin-transform-block-scoping` and other plugin using `scope.crawl()`, such as create-react-app's `babel-preset-react-app`. Here are three examples:
- https://github.com/AriPerkkio/babel-preset-react-app-transpile-bug/blob/master/.babelrc
- https://github.com/AriPerkkio/babel-preset-react-app-transpile-bug/blob/remove-props-plugin/.babelrc
- https://github.com/AriPerkkio/babel-preset-react-app-transpile-bug/blob/another-plugin/.babelrc

**Environment**

```
  System:
    OS: Linux 4.19 Debian GNU/Linux 9 (stretch) 9 (stretch)
  Binaries:
    Node: 13.7.0 - /usr/bin/node
    Yarn: 1.21.1 - /usr/bin/yarn
    npm: 6.13.6 - /usr/bin/npm
  npmPackages:
    @babel/cli: ^7.8.3 => 7.8.3 
    @babel/core: ^7.8.3 => 7.8.3 
    @babel/plugin-transform-block-scoping: ^7.8.3 => 7.8.3 

```

**Possible Solution**

I'm not familiar with `scope.crawl()` method, but either plugins should not call it, or `@babel/plugin-transform-block-scoping` should not break when it's called.

Original issues:
- https://github.com/facebook/create-react-app/issues/8322
- https://github.com/AriPerkkio/asus-merlin-simple-vpn-client-setup/issues/1

Andarist:
| Fixed Issues?            | fixes #11057
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    |
| Any Dependency Changes?  |

Bunch of references were not registered back on the program parent after re-crawling because they were not visited by the `collectorVisitor`. I've managed (mostly) to move the code parts responsible for registering those references during the first `crawl` call into the mentioned `collectorVisitor` - this should allow them to be always uniformly, each time.

### DIFF
--- a/packages/babel-traverse/src/scope/index.js
+++ b/packages/babel-traverse/src/scope/index.js
@@ -155,6 +155,7 @@ const collectorVisitor = {
   For(path) {
     for (const key of (t.FOR_INIT_KEYS: Array)) {
       const declar = path.get(key);
+      // delegate block scope handling to the `BlockScoped` method
       if (declar.isVar()) {
         const parentScope =
           path.scope.getFunctionParent() || path.scope.getProgramParent();
@@ -253,6 +254,31 @@ const collectorVisitor = {
       }
     }
   },
+
+  CatchClause(path) {
+    path.scope.registerBinding("let", path);
+  },
+
+  Function(path) {
+    if (
+      path.isFunctionExpression() &&
+      path.has("id") &&
+      !path.get("id").node[t.NOT_LOCAL_BINDING]
+    ) {
+      path.scope.registerBinding("local", path.get("id"), path);
+    }
+
+    const params: Array<NodePath> = path.get("params");
+    for (const param of params) {
+      path.scope.registerBinding("param", param);
+    }
+  },
+
+  ClassExpression(path) {
+    if (path.has("id") && !path.get("id").node[t.NOT_LOCAL_BINDING]) {
+      path.scope.registerBinding("local", path);
+    }
+  },
 };
 
 let uid = 0;
@@ -279,6 +305,7 @@ export default class Scope {
     this.path = path;
 
     this.labels = new Map();
+    this.inited = false;
   }
 
   /**
@@ -761,7 +788,10 @@ export default class Scope {
   }
 
   init() {
-    if (!this.references) this.crawl();
+    if (!this.inited) {
+      this.inited = true;
+      this.crawl();
+    }
   }
 
   crawl() {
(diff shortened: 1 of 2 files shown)


packages/babel-traverse/src/scope/index.js
```diff
@@ -761,7 +788,10 @@ export default class Scope {
   }
 
   init() {
-    if (!this.references) this.crawl();
```
Andarist, line 764: this check was super implicit and confusing (nested scopes don't even gather `references` so this object stays always empty for them, but is initialized in `.crawl`)

packages/babel-traverse/src/scope/index.js
```diff
@@ -773,50 +803,24 @@ export default class Scope {
...
-        const node = path.get(key);
-        if (node.isBlockScoped()) this.registerBinding(node.node.kind, node);
-      }
-    }
-
-    // FunctionExpression - id
-
-    if (path.isFunctionExpression() && path.has("id")) {
-      if (!path.get("id").node[t.NOT_LOCAL_BINDING]) {
+    // TODO: explore removing this as it should be covered by collectorVisitor
```
Andarist, line 806: attempt to remove this has resulted in 14~ failing tests, most of the updated snapshots seemed OK, but some didn't. It requires more investigation and I'm afraid that I won't have time to do this

nicolo-ribaudo: :+1:

PR #11376, opened 2020-04-04
Add @babel/eslint-plugin-development-internal

kaicataldo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

As discussed, this PR implements a custom ESLint rule that enforces using error messages imported from a specific, configured module.

I still need to add a README and want to add more test cases, but figured I should get a WIP PR up so that I can get some early feedback. Comments/open questions:

1. Does this API make sense? Should we also add an option that allows you to enforce specific named imports as well?
1. Does it make sense to create a new plugin for this? Seemed like the best way to have custom rules in the repo until https://github.com/eslint/rfcs/pull/9 is implemented (which is still a ways out because we're working on ESLint v7).
1. Thoughts on the name? Got a better idea? :)
1. I generally believe in making rules as small as possible (i.e. checking one thing) and so I limited this rule to checking that the error message comes from the given `errorModule`. We could potentially add more validation, but in this case I think Flow covers our bases.
1. I moved the `RuleTester` utility into the shared fixtures package.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11376"></a>


eslint/babel-eslint-plugin-internal/src/rules/dry-error-messages.js
```diff
@@ -0,0 +1,144 @@
...
+    return {
+      // Check imports up front so that we don't have to check them for every ThrowStatement.
+      ImportDeclaration(node) {
+        if (isErrorModule(errorModule, filename, node.source.value)) {
+          for (const spec of node.specifiers) {
+            importedBindings.add(spec);
+          }
+        }
+      },
+      "ThrowStatement[argument.type='CallExpression'] [callee.type='MemberExpression']"(
```
nicolo-ribaudo, line 111: Nice syntax! Question: is `ThrowStatement > CallExpression[callee.type='MemberExpression']` supported?

eslint/babel-eslint-plugin-internal/src/rules/dry-error-messages.js
```diff
@@ -0,0 +1,144 @@
...
+        node,
+      ) {
+        if (
+          node.callee.object?.type !== "ThisExpression" ||
+          node.callee.property?.name !== "raise"
```
nicolo-ribaudo, line 116: When can `object/property` be undefined, given the selector we are using?

eslint/babel-eslint-plugin-internal/src/rules/dry-error-messages.js
```diff
@@ -0,0 +1,144 @@
...
+      }
+    }
+  }
+
+  return false;
+}
+
+export default {
+  meta: {
+    type: "suggestion",
```
nicolo-ribaudo, line 79: Does `suggestion` make CI fail? (I'd like it to do so)

eslint/babel-eslint-plugin-internal/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@babel/eslint-plugin-internal",
```
nicolo-ribaudo, line 2: Nit: I'd prefer to call this `@babel/eslint-plugin-development-internal`, since it's more similar to `@babel/eslint-plugin-development` than to `@babel/eslint-plugin`.

nicolo-ribaudo:
Good job! I think that this rule is ok as-is; if we need to enforce more things we can always update it/create other rules.

eslint/babel-eslint-plugin-internal/src/rules/dry-error-messages.js
kaicataldo: In this case, `"suggestion"` is referring to what type of rule this is (i.e. it enforces a code style choice rather than a possible code error). It can still be configured as `"off"`, `"warning"`, or `"error"` in our config :)

eslint/babel-eslint-plugin-internal/src/rules/dry-error-messages.js
kaicataldo: Yes, it is! That's definitely clearer, will update :+1:

eslint/babel-eslint-plugin-internal/src/rules/dry-error-messages.js
kaicataldo: Good catch - this looks like a holdover from a previous iteration. Removing and adding tests to cover this.

nicolo-ribaudo:
@kaicataldo Is this still "WIP"?

kaicataldo:
I was planning to add a readme as well as enable the rule, but I’m also fine with doing that in a follow up PR since folks already reviewed this.

existentialism:
@kaicataldo lets add the readme and enable it

nicolo-ribaudo:
GitHub now allows converting WIP PRs to Draft PRs :smile:

kaicataldo:
Apologies for the delay! Will be getting to this in the next day or two now that I'm healthy again.

kaicataldo:
Is there a way to include this in the top level `package.json` before we publish the package? It doesn't look like we do that with anything else at the moment (though I think it would be nice to do with `@babel/eslint-parser` as well!

nicolo-ribaudo:
If just specifying `*` as the version doesn't work you can try `"@babel/eslint-plugin-development-internal": "link:./eslint/babel-eslint-plugin-development-internal"`

kaicataldo:
Any ideas as to why Travis is failing? I'm unable to replicate this locally, but it looks like `@babel/eslint-plugin-development` isn't linking correctly/is missing in `node_modules`. Could this be a caching issue?

kaicataldo:
Ah, just realized this won't handle a path like `path.resolve(__dirname, 'path/to/dir')` gracefully, since it will it try to append `.js` to it. Would it be better to just use `require.resolve()` and mock the method in tests?

eslint/babel-eslint-plugin-development/src/index.js
```diff
@@ -2,7 +2,7 @@ import noDeprecatedClone from "./rules/no-deprecated-clone";
 import noUndefinedIdentifier from "./rules/no-undefined-identifier";
 import pluginName from "./rules/plugin-name";
 
-export default {
+module.exports = {
```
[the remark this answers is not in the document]
JLHwung: 🤔Shouldn't eslint refers to the built version `lib/index.js`?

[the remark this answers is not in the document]
kaicataldo: Yeah, it does! Since ESLint core doesn't have knowledge of the interop it doesn't look for `exports.default`. Leaving `module.exports` here means it won't transform it in `lib/index.js`.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -420,12 +420,14 @@ export default class Tokenizer extends ParserErrors {
       // misleading
       this.expectPlugin("recordAndTuple");
       if (this.getPluginOption("recordAndTuple", "syntaxType") !== "hash") {
+        /* eslint-disable @babel/development-internal/dry-error-messages */
         throw this.raise(
```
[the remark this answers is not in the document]
kaicataldo: Fixed up!

packages/babel-parser/src/parser/util.js
```diff
@@ -178,16 +182,10 @@ export default class UtilParser extends Tokenizer {
       this.state.yieldPos !== -1 &&
       (this.state.awaitPos === -1 || this.state.yieldPos < this.state.awaitPos)
     ) {
-      this.raise(
-        this.state.yieldPos,
-        "Yield cannot be used as name inside a generator function",
-      );
+      this.raise(this.state.yieldPos, Errors.YieldBindingIdentifier);
```
kaicataldo, line 185: Found some cases where we could consolidate these messages! :tada:

packages/babel-parser/src/parser/util.js
```diff
@@ -178,16 +182,10 @@ export default class UtilParser extends Tokenizer {
...
     }
     if (this.state.awaitPos !== -1) {
-      this.raise(
-        this.state.awaitPos,
-        "Await cannot be used as name inside an async function",
-      );
+      this.raise(this.state.awaitPos, Errors.AwaitBindingIdentifier);
```
kaicataldo, line 188: Found some cases where we could consolidate these messages! :tada:

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -1148,7 +1150,7 @@ export default class Tokenizer extends ParserErrors {
       if (next === charCodes.plusSign || next === charCodes.dash) {
         ++this.state.pos;
       }
-      if (this.readInt(10) === null) this.raise(start, "Invalid number");
+      if (this.readInt(10) === null) this.raise(start, Errors.InvalidNumber);
```
kaicataldo, line 1153: Found some cases where we could consolidate these messages! :tada:

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -1171,7 +1173,7 @@ export default class Tokenizer extends ParserErrors {
       // disallow floats, legacy octal syntax and non octal decimals
       // new style octal ("0o") is handled in this.readRadixNumber
       if (isFloat || octal || isNonOctalDecimalInt) {
-        this.raise(start, "Invalid BigIntLiteral");
+        this.raise(start, Errors.InvalidBigIntLiteral);
```
kaicataldo, line 1176: Added this error message to `Errors`.

kaicataldo:
Hmmm, looks like CI is still failing :(

.travis.yml
```diff
@@ -17,7 +17,7 @@ before_script:
 
 script:
   - if [ "$JOB" = "test" ]; then make -j test-ci; fi
-  - if [ "$JOB" = "lint" ]; then make -j code-quality-ci; fi
+  - if [ "$JOB" = "lint" ]; then make code-quality-ci; fi
```
kaicataldo, line 20: This was failing because `node_modules/*` was being deleted while the linting task was running, leading to ESLint not finding plugins.

My experience with `make` is limited, so maybe there's a better way to fix this? Maybe doing all the prep work ahead of time in its own `make` target and then running the checks in parallel after that?

kaicataldo: Okay, I think I fixed this by bootstrapping everything as a prerequisite and then running the linting jobs afterwards in parallel. One thing I'm not entirely sure of is if the sub `make` targets are also running in parallel.

This should be ready to go now!

kaicataldo:
Was hoping to land this before I go on vacation, but please feel free to merge/fix up if you'd like to before I get back. Otherwise, I'll follow up when I return!

kaicataldo:
This is ready for final review.

PR #11377, opened 2020-04-04
Transform ES2015 Unicode Escapes to ES5

jridgewell:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | https://github.com/babel/website/pull/2261
| Any Dependency Changes?  |

Inspired by the ASCIIfier conversation at TC39, I decided to find out how Babel handled unicode surrogates. And apparently we just didn't ever write a transform for them. Maybe using non-ascii identifiers is already rare enough.

This isn't 100% full proof, since there are surrogate pairs that are valid identifiers that can't be encoded into ES5. And, Tagged Template Literals record can't be changed, because it would change their `raw` values during runtime.


jridgewell:
As a follow up, could someone that's familiar with preset-env hook this up to http://kangax.github.io/compat-table/es6/#test-Unicode_code_point_escapes?

existentialism:
@jridgewell 

> As a follow up, could someone that's familiar with preset-env hook this up to

will do!

packages/babel-plugin-transform-unicode-escapes/src/index.js
```diff
@@ -0,0 +1,87 @@
...
+
+        value.cooked = replaceUnicodeEscapes(value.cooked);
+
+        const raw = replaceUnicodeEscapes(value.raw);
+        if (raw === value.raw) return;
+
+        const grandParent = parentPath.parentPath;
+        if (grandParent.isTaggedTemplateExpression()) {
+          throw path.buildCodeFrameError(
+            `Can't replace Unicode escape inside tagged template literal`,
```
nicolo-ribaudo, line 80:
```suggestion
            `Can't replace Unicode escape inside tagged template literals. You can enable "@babel/plugin-transform-template-literals" to compile them to classic strings.`,
```

packages/babel-plugin-transform-unicode-escapes/src/index.js
```diff
@@ -0,0 +1,87 @@
...
+          parentPath.isMemberExpression({ property: node }) ||
+          parentPath.isOptionalMemberExpression({ property: node })
+        ) {
+          parentPath.node.computed = true;
+          path.replaceWith(str);
+          return;
+        }
+
+        throw path.buildCodeFrameError(
+          `Can't represent "${name}" as a bare identifier`,
```
nicolo-ribaudo, line 57: When can this happen? Since `𝒜\` is not a valid identifier, isn't it invalid also when represented as `\u{1d49c}`?

packages/babel-plugin-transform-unicode-escapes/src/index.js
```diff
@@ -0,0 +1,87 @@
...
+
+export default declare(api => {
+  api.assertVersion(7);
+
+  const surrogate = /[\ud800-\udfff]/g;
+  const unicodeEscape = /(\\+)u\{([0-9a-fA-F]+)\}/g;
+
+  function escape(code) {
+    let str = code.toString(16);
+    // Sigh, node 6 doesn't have padStart
```
nicolo-ribaudo, line 12: Maybe add a "TODO" in the comment so that we don't forget about it for Babel 8.

jridgewell: Done.

packages/babel-plugin-transform-unicode-escapes/src/index.js
JLHwung: Per https://www.ecma-international.org/ecma-262/5.1/#sec-7.6, in ES5 the Identifier name aligns to Unicode version 3.0, in which only BMP and PUA Plane are defined, that means 𝒜, introduced later in Unicode version 3.1 and allocated to SMP, was never considered a valid identifier name.

Maybe we can state it clearly that any non-BMP characters are not accepted as identifier name in ES5 and so we cannot transform 𝒜 as an identifier.

jridgewell: I'm not sure I understand. `𝒜` is a valid identifier in ES6, and can be represented as both `𝒜` and `\u{1d49c}` in ES6.

```js
var 𝒜 = 1;
console.log(\u{1d49c}); // => 1
```

In ES5, `𝒜` is interpreted as two chars, `\ud835` and `\udc9c` (not the single char `\u{1d49c}`). These individual chars must validate as an identifier. Because individual surrogates don't have Unicode categories, neither is considered a part of `ID Start` or `ID Continue`. So it's not possible to take the bare identifier `𝒜` and output `\ud835\udc9c`, since it'll be invalid.

packages/babel-plugin-transform-unicode-escapes/src/index.js
jridgewell: Done.

packages/babel-plugin-transform-unicode-escapes/src/index.js
JLHwung: > ECMAScript implementations may recognise identifier characters defined in later editions of the Unicode Standard. If portability is a concern, programmers should only employ identifier characters defined in Unicode 3.0.

`𝒜` was introduced in Unicode 3.1. So `var 𝒜 = 1` may be invalid if the implementation does not align to Unicode >3.0, i.e. Node.js v0.10.

nicolo-ribaudo: @jridgewell Thanks for the explanation, I didn't know that in identifier you must validate the single characters rather than the whole code point.

JLHwung: @nicolo-ribaudo lol that is why we have long ugly regex in `@babel/helper-validator-identifier`. And I have thought of replacing that by `\p{ID_Start}\p{ID_Continue}+` but I give up since it will imply different behaviour between Node.js versions.

existentialism:
added support in preset-env/standalone (will prolly fail test but verified it passes after rebase)

existentialism:
@jridgewell i updated the preset-env mappings after your compat table PR landed

jridgewell:
Thanks!

PR #11382, opened 2020-04-05
Unify parens printing for postfix-like expressions

nicolo-ribaudo:
| Fixed Issues?            | [repl](https://babeljs.io/repl/#?browsers=safari%2010.1&build=&builtIns=false&spec=false&loose=false&code_lz=BQQwBA_GBGYFxgMYEoCEBuAUKSN5LQDoATdIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Cenv&prettier=false&targets=&version=7.9.0&externalPlugins=%40babel%2Fplugin-syntax-typescript%407.8.3)
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This blocks https://github.com/babel/babel/pull/10961

### DIFF
--- a/packages/babel-generator/src/node/parentheses.js
+++ b/packages/babel-generator/src/node/parentheses.js
@@ -32,7 +32,7 @@ const isClassExtendsClause = (node: Object, parent: Object): boolean =>
   (t.isClassDeclaration(parent) || t.isClassExpression(parent)) &&
   parent.superClass === node;
 
-const isPostfixExpression = (node: Object, parent: Object) =>
+const hasPostfixPart = (node: Object, parent: Object) =>
   ((t.isMemberExpression(parent) || t.isOptionalMemberExpression(parent)) &&
     parent.object === node) ||
   ((t.isCallExpression(parent) ||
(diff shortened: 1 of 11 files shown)


packages/babel-generator/src/node/parentheses.js
```diff
@@ -32,6 +32,15 @@ const isClassExtendsClause = (node: Object, parent: Object): boolean =>
   (t.isClassDeclaration(parent) || t.isClassExpression(parent)) &&
   parent.superClass === node;
 
+const isPostfixExpression = (node: Object, parent: Object) =>
```
JLHwung, line 35: AFAIK the postfix expression is used to describe a way of how the operator and its operands are arranged, so I am confused about the term here.

The Spec provides a term [`LeftHandSideExpression` ](https://tc39.es/ecma262/#prod-LeftHandSideExpression) and I think we can stick with that.

nicolo-ribaudo: `LeftHandSideExpression` doesn't work, because it also covers `PrimaryExpression`.

I initially called this "postfix" because TS's `!` is a postfix operator, and because in `foo()` and `foo.bar`, `()` and `.bar` are "postfixes" of the main `foo` expression :sweat_smile:

Maybe `hasPostfixPart`?

existentialism: I'm fine with `hasPostfixPart`.

JLHwung: Good for me 🌝

PR #11384, opened 2020-04-06
Output minimal strings by default

jridgewell:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This changes our behavior from always outputting ASCII-safe strings to outputting Unicode. It's smaller, supported since at least ES5, and maybe foreign language programmers will be less surprised with the output.

Maybe we want to add a simple `ascii` option to toggle the behavior back?


packages/babel-generator/src/index.js
```diff
@@ -56,11 +55,16 @@ function normalizeOptions(code, opts): Format {
     jsescOption: {
       quotes: "double",
       wrap: true,
+      minimal: true,
       ...opts.jsescOption,
     },
     recordAndTupleSyntaxType: opts.recordAndTupleSyntaxType,
   };
 
+  if (format.jsonCompatibleStrings) {
```
JLHwung, line 64:
```suggestion
  if (opts.jsonCompatibleStrings) {
```

nicolo-ribaudo:
Ref: https://github.com/babel/babel/issues/247

packages/babel-generator/src/index.js
jridgewell: Lol. I went back and forth with just doing this inline in the options, forget to fix this.

jridgewell:
I think there's a bit of compatibility risk for lone surrogates, which aren't transformed into `\uHHHH` format. We could fix this, so we're always serving valid utf-8.

> it seems that it was done not for compatibility with old engines but for non-utf8 encodings. Is it safe to output non-ASCII characters by default?

In strings, yes. The risk is that users will have improperly setup servers (not serving with `Content-Type: text/javascript; charset=utf-8` header). In which case, just fix your server.

Also, we still retain the ability to output ASCII only, you just need to configure it. The default should probably match what the user typed. And for foreign language programmers, they're probably typing stings containing non-ASCII chars. Being able to actually read the output is probably going to much nicer trying to understand whatever `"\u1234"` means.

mathiasbynens:
Yep, @jridgewell nailed it. Note that folks who cannot alter their server configuration to add `charset=utf-8` to the `Content-Type` header can instead use `<script charset=utf-8 src=...>`. The nice thing about shipping ASCII-only code is that you don't need to do any of these, but that's mostly nice for frameworks/libraries, i.e. code that you'd expect people to copy. For your own apps, you tend to have control over either the HTML or the server configuration.

mathiasbynens:
I think what you really want is to output non-ASCII symbols as-is unless they're lone surrogates or unless they're whitespace or non-printable.

jridgewell:
> output non-ASCII symbols as-is unless they're lone surrogates or unless they're whitespace or non-printable.

Is there a `jsesc` setting for this?

packages/babel-generator/src/generators/types.js
```diff
@@ -1,6 +1,8 @@
 import * as t from "@babel/types";
 import jsesc from "jsesc";
 
+const surrogates = /[\uD800-\uDBFF][\uDC00-\uDFFF]|([\uD800-\uDBFF]|[\uDC00-\uDFFF])/g;
```
JLHwung, line 4:
```suggestion
const surrogates = /([\uD800-\uDFFF])/gu;
```
given that it is ES2015 and should be supported by Node.js 6.

jridgewell: Cool, done!

packages/babel-generator/src/index.js
```diff
@@ -56,11 +55,16 @@ function normalizeOptions(code, opts): Format {
     jsescOption: {
       quotes: "double",
       wrap: true,
+      minimal: true,
```
JLHwung, line 58: Note that sometimes user may prefer to preserve the escaped form, especially for the space characters. i.e.
```
"\u{00a0}" // NBSP 
"\u{2005}" // 1/4 em
```
and
https://github.com/angular/angular/blob/9e70bcb34f91d439f5203dc22a44f323d02c4648/packages/compiler/src/ml_parser/html_whitespaces.ts#L19

I suggest we plug in the `minimal` option only when there is no escape sequence inside.

nicolo-ribaudo: I think that it would be better to implement this in `jsesc` (https://github.com/babel/babel/pull/11384#issuecomment-610014446), since it would also benefit non-Babel users.

jridgewell: We're only able to detect what the user actually typed via the `raw` value, which is handled before printing the cooked value. If we only have the cooked value, we can only guess what the user intended.

JLHwung: For the whitespaces, I think we can add them to the whitelist of the [`minimal` option](https://github.com/mathiasbynens/jsesc#minimal).

Since it changes the default print behaviour, I consider this as a breaking change.

jridgewell: The `raw` value already captures this appropriately, though. We only run the cooked value through `jsesc` when there's not a raw value. At that point, we have no idea what the user actually wrote in the source file.

mathiasbynens: Indeed. I do think we want to always escape things like NBSP though.

Maybe we should special-case all non-ASCII entries in `require('unicode-13.0.0/General_Category/Space_Separator/symbols.js');` in jsesc’s `minimal` option.

mathiasbynens:
`jsesc@3.0.0` now escapes lone surrogates even when `minimal: true` thanks to @jridgewell!

packages/babel-generator/src/index.js
jridgewell:
```
[ '0x20',
  '0xa0',
  '0x1680',
  '0x2000',
  '0x2001',
  '0x2002',
  '0x2003',
  '0x2004',
  '0x2005',
  '0x2006',
  '0x2007',
  '0x2008',
  '0x2009',
  '0x200a',
  '0x202f',
  '0x205f',
  '0x3000' ]
```

I think we should continue to print regular space. I could understand escaping the rest, though. @mathiasbynens, are you open to doing that in jsesc?

mathiasbynens: Agreed we wouldn't want to escape any ASCII spaces. We should take non-ASCII entries only.

Definitely open to a patch that does this in jsesc's `minimal` option. We could use [`src/data.js`](https://github.com/mathiasbynens/jsesc/blob/master/src/data.js) to export a list of special-cased symbols, based on reading `require('unicode-13.0.0/General_Category/Space_Separator/symbols.js');` (or `/code-points.js`) and filtering out all ASCII entries.

jridgewell: Updated to v3.0.1, which now escapes non-ASCII whitespace.

jridgewell:
>  AFAIK the non-minimal (defaults in babel v7) behaviour is unchanged between 2 and 3, isn't it?

Yes. But it might just be easier to land everything on next-8-dev. If we were to upgrade the v3 and a user were using `minimal`, they would get changed behavior.

packages/babel-generator/src/generators/types.js
JLHwung, line 4: nit: Since we are targeting to Node.js 10 from Babel 8, it suffices to use `/\p{Surrogate}/gu`.

jridgewell: Ugh, I force pushed the wrong commit. This should be gone now, since jsesc does everything needed.

packages/babel-generator/src/index.js
```diff
@@ -55,11 +55,16 @@ function normalizeOptions(code, opts): Format {
...
       ...opts.jsescOption,
     },
     recordAndTupleSyntaxType: opts.recordAndTupleSyntaxType,
   };
 
+  if (opts.jsonCompatibleStrings) {
```
nicolo-ribaudo, line 64: This option has been removed: https://github.com/babel/babel/pull/9958

PR #11395, opened 2020-04-09
chore: update stage preset in babel-standalone

JLHwung:
| Fixed Issues?            | Unblocks https://github.com/babel/website/issues/2224
| Patch: Bug Fix?          | I think we can ship it in patch release, if not I am okay with minor
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

- Removed unused syntax plugins since they have been enabled by default in babel-parser
- Updated proposal stages
- Added Record&Tuple syntax support


packages/babel-standalone/src/preset-stage-2.js
```diff
@@ -18,8 +18,6 @@ export default (_: any, opts: Object = {}) => {
         { legacy: decoratorsLegacy, decoratorsBeforeExport },
       ],
       babelPlugins.proposalFunctionSent,
-      babelPlugins.proposalExportNamespaceFrom,
-      babelPlugins.proposalNumericSeparator,
```
JLHwung, line 22: They are moved to stage 3.

PR #11404, opened 2020-04-10
Skip TSAsExpression when transforming spread in CallExpression

### LINKED ISSUE #11400: Typescript: object method with type assertion and spread arguments
## Bug Report

**Current Behavior**
When compiling Typescript, asserting an object's method as some type and immediately invoking it with spread arguments results in calling the method with `.apply(void 0, ...`, which results in an undefined value for `this` inside the method (despite being called on the object).

This does not occur if you drop the type assertion, or if you don't use the spread operator. Only if both are used.

**Input Code**
- [REPL link](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=BQEw9g5gdARghgJwNYAI4Gc0DsCeBKYKIxCdPIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Ctypescript&prettier=false&targets=&version=7.9.0&externalPlugins=)

```js
// using es2015-loose for brevity, but also reproducible with es2015

(dog.bark as any)(...args)
// ^ compiles to `dog.bark.apply(void 0, args)` - wrong

(dog.bark as any)(args)
// ^ compiles to `dog.bark(args);` - correct

(dog.bark)(...args)
// ^ compiles to `(_dog = dog).bark.apply(_dog, args);` - correct
```

**Expected behavior/code**
The method should be invoked with the object as the `this` arg.

Reproducible in the REPL with only es2015 and typescript presets.

**Environment**
N/A - see REPL link

oliverdunk:
| Fixed Issues?            | Fixes #11400
| Documentation PR Link    | 
| Any Dependency Changes?  |

In babel-plugin-transform-spread, when transforming spreaded arguments in a CallExpression, we weren't handling the case where the callee was a TSAsExpression. This meant we incorrectly determined the this of our apply call, transforming:

`(dog.bark as any)(...args)`

... into ...

`dog.bark.apply(void 0, args)`

I've fixed that by exploring deeper, into the expression of the TSAsExpression.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11404"></a>


oliverdunk:
Not sure if there's a better solution here. It feels like the ideal would be passing the entire path to a "get me this without the TypeScript" function.

nicolo-ribaudo:
Other tests:
```js
class A {
  #x;
  
  fn() {
    (this.#x as any)();
  }
}
```

```js
(a.b as any)?.()
```

And also with Flow type casts (`(a.b: any)(...args)`) and old-style TS casts (`(<any> a.b)(...args)`).

> It feels like the ideal would be passing the entire path to a "get me this without the TypeScript" function.

I agree! Missing the correct this with type casts is a common pitfall and a shared function would be useful. What about a new `@babel/helper-get-call-context` package?

packages/babel-plugin-transform-spread/src/index.js
```diff
@@ -94,7 +94,14 @@ export default declare((api, options) => {
...
         if (!hasSpread(args)) return;
 
-        const calleePath = path.get("callee");
+        let calleePath = path.get("callee");
+
+        if (calleePath.isTSAsExpression()) {
+          do {
+            calleePath = calleePath.get("expression");
+          } while (calleePath.isTSAsExpression());
+        }
```
nicolo-ribaudo, line 103: Using `if` before a `do-while` is exactly like using a `while`:
```suggestion
        while (calleePath.isTSAsExpression()) {
          calleePath = calleePath.get("expression");
        }
```

oliverdunk:
Thanks for the feedback @nicolo-ribaudo! I've created the package, and added a bunch of extra tests - let me know what you think.

I wasn't able to test `(a.b as any)?.()`, because I don't think `babel-plugin-transform-spread` supports optional call expressions. Do you agree? If so, I can open an issue to fix that.

I have two quick tweaks which I will make tomorrow morning:

- My new package should probably support optional call expressions (it nearly does, but I just realised that I made a mistake by asserting I was passed a call expression).
- It looks like the tests are failing. I'll try merging with master, unless you can see that something else is going wrong.

oliverdunk:
> Thanks for the feedback @nicolo-ribaudo! I've created the package, and added a bunch of extra tests - let me know what you think.
> 
> I wasn't able to test `(a.b as any)?.()`, because I don't think `babel-plugin-transform-spread` supports optional call expressions. Do you agree? If so, I can open an issue to fix that.
> 
> I have two quick tweaks which I will make tomorrow morning:
> 
> * My new package should probably support optional call expressions (it nearly does, but I just realised that I made a mistake by asserting I was passed a call expression).
> * It looks like the tests are failing. I'll try merging with master, unless you can see that something else is going wrong.

Made changes as promised.

nicolo-ribaudo:
@oliverdunk The tests I proposed in https://github.com/babel/babel/pull/11404#issuecomment-612395596 where _without_ spread :stuck_out_tongue: 

The caller is also needed by the optional chaining plugin.

packages/babel-plugin-proposal-private-methods/test/fixtures/call-context/ts-as-class-field/output.js
```diff
@@ -0,0 +1,17 @@
...
+class A {
+  constructor() {
+    _x.set(this, {
+      writable: true,
+      value: void 0
+    });
+  }
+
+  fn() {
+    _classPrivateFieldGet(this, _x)();
```
nicolo-ribaudo, line 12: This still call still doesn't have the correct `this`!

oliverdunk: Thanks, admittedly I didn't look close enough and thought it was right! That's a tough one, because we're going in the opposite direction. The transformation is done by `babel-helper-member-expressions-to-functions`, which takes a member, and then checks to see if its parent is a call expression.

I was able to fix it with the following diff: https://hastebin.com/onotaqusiw.diff

It seems like this would be an issue for the entire helper. Maybe the helper should be a type stripper, instead of just something for calls? It would export a function for each direction.

The final tricky bit would be things like this:

`if (parentPath.isAssignmentExpression({ left: node })) {`

We could probably change that to this?

`if (parentPath.isAssignmentExpression({ left: previousParent })) {`

With previousParent set to node initially, and then the last parent each time we go a level up?

oliverdunk:
I'm not sure if I need to do something to get the tests working with a new package?

packages/babel-plugin-proposal-private-methods/test/fixtures/call-context/ts-as-class-field/output.js
oliverdunk: It crossed my mind that if we move the logic to a plugin, we can't keep track of the previous parent. Maybe this logic has to be duplicated in this case.

packages/babel-plugin-proposal-optional-chaining/package.json
```diff
@@ -21,6 +21,7 @@
   "devDependencies": {
     "@babel/core": "^7.9.0",
     "@babel/helper-plugin-test-runner": "^7.8.3",
-    "@babel/plugin-transform-block-scoping": "^7.8.3"
+    "@babel/plugin-transform-block-scoping": "^7.8.3",
+    "@babel/helper-get-call-context": "7.9.6"
```
nicolo-ribaudo, line 25: Use `^`. I don't think that it's what is causing problems, but it's the only difference I see.

packages/babel-helper-get-call-context/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "version": "7.9.6",
+  "description": "Helper function to get call context",
+  "repository": "https://github.com/babel/babel/tree/master/packages/babel-helper-get-call-context",
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "lib/index.js",
+  "dependencies": {
+    "@babel/traverse": "^7.9.6"
```
nicolo-ribaudo, line 12: Can you make this a `devDependency`, since we only use it for the type and we strip types before publishing?

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -35,6 +37,11 @@ export default declare((api, options) => {
           const { node } = optionalPath;
           if (node.optional) {
             optionals.push(node);
+            chains.push(
+              optionalPath.isOptionalCallExpression()
+                ? getCallContext(optionalPath)["node"]
+                : node["object"],
```
nicolo-ribaudo, line 43:
```suggestion
                ? getCallContext(optionalPath).node
                : node.object,
```

packages/babel-plugin-proposal-private-methods/test/fixtures/call-context/ts-as-class-field/output.js
nicolo-ribaudo: Maybe we could keep this for a follow-up PR then

oliverdunk: Sounds good. I have a first draft, so I'll open a WIP MR with that once this is merged.

packages/babel-helper-get-call-context/package.json
oliverdunk: Ah, makes sense.

oliverdunk:
We should probably strip types from optional member expressions, too! 😅 

Does replacing `@babel/helper-get-call-context` with `@babel/helper-skip-types` sound ok?

packages/babel-plugin-proposal-optional-chaining/package.json
oliverdunk: No difference unfortunately. Perhaps https://github.com/babel/babel/pull/11581 could be the fix?

nicolo-ribaudo:
> We should probably strip types from optional member expressions, too! 😅

What do you mean?

Btw, for some other PRs (e.g. https://github.com/babel/babel/pull/11552) it might be useful to add a function to this package that, give the callee, returns the call (the opposite of the one introduced by this PR).

oliverdunk:
> What do you mean?

I'm thinking of this case: `(a?.b as ExampleType)?.c`

packages/babel-plugin-transform-spread/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.8.3",
-    "@babel/helper-plugin-test-runner": "^7.8.3"
+    "@babel/helper-plugin-test-runner": "^7.8.3",
+    "@babel/helper-get-call-context": "7.9.6"
```
nicolo-ribaudo, line 23:
```suggestion
    "@babel/helper-get-call-context": "^7.9.6"
```

oliverdunk:
I'm pushed! Hopefully the tests pass.

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -1,4 +1,7 @@
 import { declare } from "@babel/helper-plugin-utils";
+import skipTransparentExprWrappers, {
+  getCallContext,
+} from "@babel/helper-skip-transparent-expr-wrappers";
```
nicolo-ribaudo, line 4: Nit: I would slightly prefer to only use named exports in `@babel/helper-skip-transparent-expr-wrappers`, to have nicer imports.

packages/babel-plugin-proposal-optional-chaining/test/fixtures/transparent-expr-wrappers/ts-as-call-context/output.js
```diff
@@ -0,0 +1,3 @@
+var _a$b, _a;
+
+(_a$b = ((_a = a).b as any)) === null || _a$b === void 0 ? void 0 : _a$b.call(_a);
```
nicolo-ribaudo, line 3: @JLHwung WDYT about keeping the types? It's a nice thing to do, but I'd like to hear your opinion regarding whether or not it's worth the code complexity.

packages/babel-plugin-proposal-optional-chaining/src/index.js
oliverdunk: Sorted!

packages/babel-plugin-proposal-optional-chaining/test/fixtures/transparent-expr-wrappers/ts-as-call-context/output.js
JLHwung: I am good on changes here! Ideally types should only be removed via preset-flow/typescript.

oliverdunk:
@nicolo-ribaudo @JLHwung I've just merged with master, would love your feedback here. I had to change my approach slightly to accommodate https://github.com/babel/babel/pull/11623.

The most notable was that instead of building up a chains array in the initial setup, I now calculate the chain and chainWithTypes using the node, during the loop. This means working with a node instead of a path, but has the benefit of staying up to date, since @JLHwung's new code makes significant changes to the chain during the loop.

oliverdunk:
Oh, I threw in an extra test too! Wanted to make sure we skip over types when determining if we're in a CallExpression that needs the bind.

packages/babel-plugin-proposal-optional-chaining/test/fixtures/transparent-expr-wrappers/ts-parenthesized-expression-member-call/input.ts
```diff
@@ -0,0 +1 @@
+((o?.Foo.m) as ExampleType)()
```
nicolo-ribaudo, line 1: Nit: please add a trailing new line in these fails :pray:

packages/babel-plugin-transform-spread/test/fixtures/call-context/parenthesized-expressions/output.js
```diff
@@ -0,0 +1,3 @@
+var _a;
+
+((_a = a).b.apply)(_a, babelHelpers.toConsumableArray(args));
```
nicolo-ribaudo, line 3: It would be nice if this was transpiled to
```js
((_a = a).b).apply(_a, babelHelpers.toConsumableArray(args));
```

packages/babel-helper-skip-transparent-expr-wrappers/src/index.js
```diff
@@ -0,0 +1,32 @@
...
+      `Expected type "CallExpression" or "OptionalCallExpression" ` +
+        `but instead got "${callPath.type}".`,
+    );
+  }
+
+  const calleePath = callPath.get("callee");
+  return skipTransparentExprWrappers(calleePath);
+}
+
+export function isTransparentExprWrapper(node: Node) {
```
JLHwung, line 16: Can you add a comment to show the idea/concept of `TransparentExprWrapper`? Since it is not in the spec, the context of how this concept is come up will benefit new contributors.

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -62,28 +65,36 @@ export default declare((api, options) => {
 
           const isCall = t.isCallExpression(node);
           const replaceKey = isCall ? "callee" : "object";
-          const chain = node[replaceKey];
+
+          const chainWithTypes = node[replaceKey];
```
JLHwung, line 69: nit: I think the naming `chainWithTypes` does not cover the `ParenthesizedExpression` cases so I prefer `chain` instead. Maybe we can reuse `skipTransparentExprWrappers` here:

```js
const chain = skipTransparentExprWrappers(path.get("replaceKey")).node
```

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -130,12 +141,14 @@ export default declare((api, options) => {
             if (!loose || !isSimpleMemberExpression(object)) {
               // memoize the context object in non-loose mode
               // `(a?.b.c)()` to `(a == null ? undefined : (_a$b = a.b).c.bind(_a$b))()`
-              baseRef = scope.maybeGenerateMemoised(object);
+              baseRef = scope.maybeGenerateMemoised(
+                skipTransparentExprWrappers(object),
```
JLHwung, line 145: `object` is an ASTNode but not NodePath.

packages/babel-plugin-proposal-optional-chaining/test/fixtures/transparent-expr-wrappers/ts-parenthesized-expression-member-call/input.ts
oliverdunk: Of course, done!

packages/babel-plugin-transform-spread/test/fixtures/call-context/parenthesized-expressions/output.js
oliverdunk: That's a lot nicer, I made it happen 🙌

packages/babel-helper-skip-transparent-expr-wrappers/src/index.js
oliverdunk: Good call, done!

packages/babel-plugin-proposal-optional-chaining/src/index.js
oliverdunk: We actually need chain still (it's redeclared below), so can't reuse that name. The idea is that further down, we sometimes care about the types, and other times we want to ignore them. I'd be open to an alternative - `chainWithoutWrappers` perhaps?

packages/babel-plugin-proposal-optional-chaining/src/index.js
oliverdunk: Oh wow, thanks for catching that. I wonder why it works. Is .get() also defined on AST nodes?

Will look at this tomorrow, I think I might need a new function in the helper package.

oliverdunk:
Thanks for the feedback! Found some time to take a look at it.

@nicolo-ribaudo, I've hopefully addressed all of your comments. Would love for you to take another look.

@JLHwung, I need to fix the helper call where I wrongly pass an ASTNode. I'll ping you when that's done. In the meantime, I did leave a response to another bit of feedback you left.

Also changed the base branch to `main`, good call on the switch!

packages/babel-plugin-proposal-optional-chaining/src/index.js
oliverdunk: Sorted :)

packages/babel-helper-skip-transparent-expr-wrappers/README.md
```diff
@@ -0,0 +1,17 @@
+# @babel/helper-skip-transparent-expr-wrappers
```
existentialism, line 1: Nit: I think we should expand expr -> expression, @babel/helper-skip-transparent-expression-wrappers

oliverdunk:
Thanks for the reviews @existentialism and @kaicataldo! I went ahead and pushed the rename.

oliverdunk:
The changelog comes from the PR title, right? If so, we might want something broader:

> Introduce @babel/helper-skip-transparent-expression-wrappers package, to skip irrelevant types and parentheses in plugins

packages/babel-helper-skip-transparent-expression-wrappers/src/index.js
```diff
@@ -0,0 +1,37 @@
+import * as t from "@babel/types";
+import type { NodePath } from "@babel/traverse";
+
+export function getCallContext(callPath: NodePath): NodePath {
```
JLHwung, line 4: I think `getCallContext` is a consumer of `helper-skip-transparent-expression-wrappers` and it should not be in this package.

Do we have a test case on `Expected type "CallExpression" or "OptionalCallExpression" `? I think we can go with `skipTransparentExprWrappers(path.get("callee"))` and not worry about `path.type`, they should be validated by transformers or implied by how the `path` is filtered. i.e. `"CallExpression|OptionalCallExpression"(path)`.

oliverdunk: I'm alright with that. I think it made more sense when the scope of the package was smaller.

oliverdunk:
@JLHwung All sorted 😄

packages/babel-plugin-transform-spread/test/fixtures/call-context/flow-type-cast/options.json
```diff
@@ -0,0 +1,7 @@
+{
+  "presets": [
```
JLHwung, line 2: Can you use syntax plugin only? So we can test the types are indeed preserved.
```
{
  "plugins": ["syntax-flow"]
}
```

oliverdunk: Good call! I tweaked that, and also did the same for the two other tests.

packages/babel-plugin-transform-spread/test/fixtures/call-context/flow-type-cast/options.json
```diff
@@ -1,7 +1,10 @@
 {
-  "presets": [
+  "plugins": [
     [
-      "flow"
+      "transform-spread"
```
JLHwung, line 4: Can you add plugin `external-helpers` here? So we don't have to update this test fixture when the helper code is changed.

oliverdunk: @JLHwung This doesn't seem to work for me. I get an output that doesn't have the spread transformation. Any ideas?

JLHwung: Oh my bad. The `plugins` is overwritten completely.

```json
{
  "plugins": ["external-helpers", "transform-spread", "syntax-flow"]
}
```
should works.

oliverdunk: Resolved by https://github.com/babel/babel/pull/11404/commits/e6c3a4c5f0d381f2eeca881749efcd9eb56774db. Thanks for the tip!

oliverdunk:
@JLHwung Removed the inline helper functions! This should be good for another look.

I'm happy to keep tweaking this, but at the same time, I won't be offended if you want to jump in with a few commits. If you still have some tidy up in mind it might be quicker :)

oliverdunk:
@JLHwung Thanks for the merge!

JLHwung:
@oliverdunk Thank you for all the efforts!

PR #11405, opened 2020-04-11
Fix nested classes reference private fields

jridgewell:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    |
| Any Dependency Changes?  |

We failed to properly handle nested private field, when only a single private field was redeclared. [Example](https://babeljs.io/repl/#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEBiD29oG8BQ1oGIBmjoF5oBGAbnSwCMwAnA4s8gFwFMJGAKeRgC2eoEoU5DKEgwAcqxYATIRnmUadAEwMFGFm3aC06-TwCWEAHQ5EZPRu5HTVahfUBfYdGcZnzoA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=stage-2&prettier=false&targets=&version=7.9.0&externalPlugins=)

When we hit a nested class that redeclares a field, we [skip the nested class](https://github.com/babel/babel/blob/070ec201bb1ab8bf78fbb3c7aff2337056bf0ae3/packages/babel-helper-create-class-features-plugin/src/fields.js#L94-L98). And when we finally traverse that nested class, we'll only have the privates that are declared by it. We've lost the ancestor's private fields!

### DIFF
--- a/packages/babel-helper-create-class-features-plugin/src/fields.js
+++ b/packages/babel-helper-create-class-features-plugin/src/fields.js
@@ -7,21 +7,24 @@ import optimiseCall from "@babel/helper-optimise-call-expression";
 
 import * as ts from "./typescript";
 
-export function buildPrivateNamesMap(props) {
-  const privateNamesMap = new Map();
+export function buildPrivateNamesMap(privateNamesMap, props, depth) {
   for (const prop of props) {
-    const isPrivate = prop.isPrivate();
-    const isMethod = !prop.isProperty();
-    const isInstance = !prop.node.static;
-    if (isPrivate) {
+    if (prop.isPrivate()) {
+      const isMethod = !prop.isProperty();
+      const isInstance = !prop.node.static;
       const { name } = prop.node.key.id;
-      const update = privateNamesMap.has(name)
-        ? privateNamesMap.get(name)
-        : {
-            id: prop.scope.generateUidIdentifier(name),
-            static: !isInstance,
-            method: isMethod,
-          };
+
+      let update = privateNamesMap.get(name);
+      if (!update || update.depth !== depth) {
+        update = {
+          id: prop.scope.generateUidIdentifier(name),
+          static: !isInstance,
+          method: isMethod,
+          depth,
+        };
+        privateNamesMap.set(name, update);
+      }
+
       if (prop.node.kind === "get") {
         update.getId = prop.scope.generateUidIdentifier(`get_${name}`);
       } else if (prop.node.kind === "set") {
@@ -29,13 +32,16 @@ export function buildPrivateNamesMap(props) {
       } else if (prop.node.kind === "method") {
         update.methodId = prop.scope.generateUidIdentifier(name);
       }
-      privateNamesMap.set(name, update);
     }
   }
-  return privateNamesMap;
 }
 
-export function buildPrivateNamesNodes(privateNamesMap, loose, state) {
+export function buildPrivateNamesNodes(
+  privateNamesMap,
+  loose,
+  state,
+  privateDepth,
+) {
   const initNodes = [];
 
   for (const [name, value] of privateNamesMap) {
@@ -45,7 +51,16 @@ export function buildPrivateNamesNodes(privateNamesMap, loose, state) {
     // In spec mode, only instance fields need a "private name" initializer
     // because static fields are directly assigned to a variable in the
     // buildPrivateStaticFieldInitSpec function.
-    const { id, static: isStatic, method: isMethod, getId, setId } = value;
+    const {
+      id,
+      static: isStatic,
+      method: isMethod,
+      getId,
+      setId,
+      depth,
+    } = value;
+    if (depth !== privateDepth) continue;
+
     const isAccessor = getId || setId;
     if (loose) {
       initNodes.push(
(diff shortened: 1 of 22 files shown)


packages/babel-helper-create-class-features-plugin/src/index.js
```diff
@@ -49,141 +51,158 @@ export function createClassFeaturePlugin({
...
-        let constructor;
-        let isDecorated = hasOwnDecorators(path.node);
-        const props = [];
-        const elements = [];
-        const computedPaths = [];
-        const privateNames = new Set();
-        const body = path.get("body");
+          const loose = isLoose(this.file, feature);
+          const privateNamesMap = new Map(
+            privateNamesStack[privateNamesStack.length - 1],
```
nicolo-ribaudo, line 62: Since nodes can be re-queued and skipped, it's not guaranteed that they are visited in-order.

Instead of a stack, I think we could use a `WeakMap` from class nodes to `privateNamesMap`, and get the parent class using `path.findParent(p => p.isClass())`.

packages/babel-helper-create-class-features-plugin/src/index.js
```diff
@@ -30,6 +30,8 @@ export { FEATURES, injectInitialization };
 const version = pkg.version.split(".").reduce((v, x) => v * 1e5 + +x, 0);
 const versionKey = "@babel/plugin-class-features/version";
 
+const privateNamesStack = [new Map()];
```
nicolo-ribaudo, line 33: When using a `WeakSet`, please put this in the `createClassFeaturePluign` function so that when the plugin finishes running it can be garbage-collected, even if the AST was still retained.

nicolo-ribaudo:
> When we hit a nested class that redeclares a field, we skip the nested class.

I was wondering if it would be easier to fix it there. The `Class` visitor in `privateNameVisitor` could become something like this:

```js
  Class(path) {
    const { privateNamesMap } = this;
    const body = path.get("body.body");

    // First, evaluate the things in the outer scope.
    path.traverse(privateNameInnerVisitor, this);

    const visiblePrivateNames = new Map(privateNamesMap);
    for (const prop of body) {
      if (prop.isPrivate()) visiblePrivateNames.delete(prop.node.key.id.name);
    }

    path.traverse(privateNameVisitor, { ...this, privateNamesMap: visiblePrivateNames });
    path.skip();
  },
```

jridgewell:
> I was wondering if it would be easier to fix it there. The Class visitor in privateNameVisitor could become something like this:

I was trying to avoid processing the class multiple times. But this is easier to do, given the issues with out-of-order requeueing. Updated.

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/nested-class-redeclared/input.js
```diff
@@ -0,0 +1,15 @@
+class Foo {
```
JLHwung, line 1: Can you add a case on redeclared private names with computed key reference? The following case should throw since `#foo` is shadowed.
```js
class Foo {
  #foo = 1;

  test() {
    class Nested {
      #foo = 2;
      [this.#foo] = 3;
    }

    this.#foo;
  }
}
```
Given that it is not directly related to this PR, we can fix it in a separate one.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -85,24 +85,36 @@ const privateNameVisitor = {
...
+    // If the class doesn't redeclare any private fields, we can continue with
+    // our overall traversal.
+    if (visiblePrivateNames.size === privateNamesMap.size) {
+      return;
     }
+
+    // This class redeclares some private field. We need to process the outer
+    // environment with access to all the outer privates, then we can process
+    // the inner environment with only the still-visible outer privates.
+    path.traverse(privateNameNestedVisitor, this);
```
nicolo-ribaudo, line 103: @JLHwung Maybe it would be enough to remove this line to fix it?

JLHwung: I don't think so but I haven't tried. The nested private name is still allowed as long as it does not accessing outer class private identifiers. I think we shall whitelist computed element names for proper private name resolving behaviour.

```js
class Foo {
  #foo = 1;

  test() {
    class Nested {
      #foo = 2;
      bar = this.#foo;
    }

    new Nested;
  }
}
(new Foo).test();
```

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/nested-class-redeclared/input.js
jridgewell: Shouldn't we wait till the other PR to add the test case? It's going to be much harder to fix.

jridgewell: Actually, I have a fix.

JLHwung: Oh I meant we can add the new case and the fixes in a separate PR, my bad. As long as you come up with a fix, that's awesome! 😄

PR #11408, opened 2020-04-11
Add "Ping Asignees" GitHub action

nicolo-ribaudo:

- Ping assignees after 2 weeks to check if they are still working on the bug
- Unassign them after 1 additional week if they don't reply

The action source is at https://github.com/babel/actions/blob/v2/check-assigned-issues/main.js

I have tested this action at https://github.com/babel/test-github-actions/issues/56 (private repo), using 5 and 2 minutes instead of 14 and 7 days.

### DIFF
--- a/.github/workflows/ping-assignees.yml
+++ b/.github/workflows/ping-assignees.yml
@@ -15,7 +15,7 @@ jobs:
           token: ${{ secrets.BOT_TOKEN }}
           bot_name: babel-bot
           days_before_warning: 14
-          days_before_unassigning: 7
+          days_before_unassigning: 70
           comment: >
             Hi, @${username}! It has been 2 weeks since your last update: are
             you still working on this issue?
--- a/.github/workflows/ping-assignees.yml
+++ b/.github/workflows/ping-assignees.yml
@@ -1,0 +1,31 @@
+name: Ping Assignees
+
+on:
+  # Every day at 0:00
+  schedule: 0 0 * * *
+
+jobs:
+  ping_assignees:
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
e2e failure is not related.

.github/workflows/ping-assignees.yml
```diff
@@ -0,0 +1,31 @@
...
+    name: Ping Assignees
+    runs-on: ubuntu-latest
+    steps:
+      - name: Ping Assignees
+        uses: babel/actions/check-assigned-issues@v2
+        with:
+          token: ${{ secrets.BOT_TOKEN }}
+          bot_name: babel-bot
+          days_before_warning: 14
+          days_before_unassigning: 70
```
nicolo-ribaudo, line 18: If we leave 7, I think that this action will unassign issues inactive for 3 weeks (2 + 1) without first giving the warning. In one week I will decrease it back to 7.

nicolo-ribaudo:
I also need to check if the issue has any linked PR.

oliverdunk:
RE: https://github.com/babel/actions/blob/97afac23ab21e3e2849567127c5cd16d8aba1b09/check-assigned-issues/main.js#L9 - it looks like there's a typo in the `required` option? I guess `@actions/core` doesn't do any checking there.

Forgive me for being nosy and looking at a draft PR!

nicolo-ribaudo:
I don't think we need this PR anymore -- we rarely need to ping people, and the time needed to do it is less than the time I would need to get this PR ready.

PR #11410, opened 2020-04-13
Throw error for TypeScript `declare const enum`

### LINKED ISSUE #10785: String valued enums differ from tsc
## Bug Report

I think this should have been fixed by #7160.

**Input Code**
```ts
declare const enum E {
        e0 = "e0",
}

function f()
{
        console.log(E.e0);
}
```

**Expected behavior/code**
This is `tsc`'s output:
```js
function f() {
    console.log("e0" /* e0 */);
}
```

This is babel's output using `@babel/preset-typescript`:
```js
function f() {
  console.log(E.e0);
}
```

- Babel version(s): v7.7.4
- Node/npm version: Node 12 / npm 6.12.0
- OS: Gentoo Linux
- Monorepo: Just test file
- How you are using Babel: `npm run babel -- --config-file /path/to/babelrc --out-file a.js a.ts` with `babelrc` looking like this:
```json
{
  "presets": [
    "@babel/preset-typescript"
  ]
}
```

dosentmatter:
| Fixed Issues?            | Fixes #10785 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

In #10785 @fiesh reported that `declare const enum` was transpiling as if it was `declare enum`. Since we don't support `const enum`, an error should be thrown whether it has `declare` or not.

### DIFF
--- a/packages/babel-plugin-transform-typescript/src/enum.js
+++ b/packages/babel-plugin-transform-typescript/src/enum.js
@@ -3,13 +3,13 @@ import { template } from "@babel/core";
 
 export default function transpileEnum(path, t) {
   const { node } = path;
+  if (node.const) {
+    throw path.buildCodeFrameError("'const' enums are not supported.");
+  }
+
   if (node.declare) {
     path.remove();
     return;
-  }
-
-  if (node.const) {
-    throw path.buildCodeFrameError("'const' enums are not supported.");
   }
 
   const name = node.id.name;
--- a/packages/babel-plugin-transform-typescript/test/fixtures/declarations/const-enum/input.ts
+++ b/packages/babel-plugin-transform-typescript/test/fixtures/declarations/const-enum/input.ts
@@ -1,0 +1,1 @@
+declare const enum E {}
--- a/packages/babel-plugin-transform-typescript/test/fixtures/declarations/const-enum/options.json
+++ b/packages/babel-plugin-transform-typescript/test/fixtures/declarations/const-enum/options.json
@@ -1,0 +1,1 @@
+{ "throws": "'const' enums are not supported." }


nicolo-ribaudo:
I'm not sure if this should be categorized as a breaking change (and thus go in Babel 8) or as a bugfix.

dosentmatter:
@nicolo-ribaudo, I guess it depends how you interpret the docs.

In docs, we say https://babeljs.io/docs/en/babel-plugin-transform-typescript#caveats
> This plugin does not support `const enum`s because those require type information to compile.

We don't specify whether `const enum`s includes `declare const enum`.

I think `const enum` and `declare const enum` are the same thing anyway. `declare` makes it so no code is generated and it is only a hint to the compiler. It doesn't affect `const enum` since `const enum` already generates 0 code.

See:
https://stackoverflow.com/questions/35019987/what-does-declare-do-in-export-declare-class-actions
https://www.typescriptlang.org/docs/handbook/enums.html#ambient-enums

`tsc` has same output for both:
image

--

I think it's better to treat it as a bug. Because it seems we just accidentally put the two `if (node.const) {` and `if (node.declare) {` in the wrong order.

The current code, transpiles `declare const enum` correctly because it just removes it.

But the problem is with accessors. It's transpiling enum access as if it were `declare enum`. If you look at https://github.com/babel/babel/issues/10785, you'll see that

```
declare const enum Enum {
  A = 1,
}
console.log(Enum.A)
```
transpiles to
```
console.log(Enum.A)
```

If we supported `const enum`, it would be
```
console.log(1)
```

Since we don't transpile the accessors correctly, I think it's better to treat this as a bug. I'm not sure if we could consider the current functionality "working". It doesn't match the TS spec and I don't think we documented it.

If someone did depend on `declare const enum` working, they can fix it by changing to `declare enum`, which is the more correct for in TypeScript.

PR #11412, opened 2020-04-13
Fix evaluation order with object spread

Zzzen:
| Fixed Issues?            |  
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

TLDR: currently, the code below is not transpiled correctly. [REPL](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=G4QwTgBA1hC8EG8IgFwQIwBoICM0CYIBfAbgChRIB7ORZNAZmwDpWps9pmQBqH48mQDGVAHYBnKgBsApsylUA5gAoIVbnFjx0EAPwQARFSgGIaI1SoAHcQYCUQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Cenv&prettier=true&targets=&version=7.8.4&externalPlugins=)

```js
var k = { a: 1, b: 2 };
var o = { a: 3, ...k, b: k.a++ };

```
expected `o.a` to be 1.

Long story: I found weird `__assign` js code generated by `tsc`. Out of curiosity, I checked the relevant transformer code and feed the test to babel 💥. This fix is migrated from TypeScript. [Original Issue](https://github.com/microsoft/TypeScript/issues/31469)


nicolo-ribaudo:
Can we limit this deoptimization to when `path.isPure()` is false for some property after the spread?

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-spread/assignment/output.js
```diff
@@ -4,10 +4,10 @@ function _objectSpread(target) { for (var i = 1; i < arguments.length; i++) { va
...
 
-z = _objectSpread({
+z = _objectSpread(_objectSpread({
   x
-}, y);
+}), y);
 z = {
   x,
-  w: _objectSpread({}, y)
+  w: _objectSpread(_objectSpread({}), y)
```
nicolo-ribaudo, line 12: This change seems unnecessary.

Zzzen: Yeah, `scope.isPure` is not working as intended. What's the difference between `scope.isPure` and `path.isPure`?

Zzzen: :dizzy_face: My bad , `scope.isPure` is working.

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-spread/expression/output.js
```diff
@@ -4,11 +4,11 @@ function _objectSpread(target) { for (var i = 1; i < arguments.length; i++) { va
 
 function _defineProperty(obj, key, value) { if (key in obj) { Object.defineProperty(obj, key, { value: value, enumerable: true, configurable: true, writable: true }); } else { obj[key] = value; } return obj; }
 
-_objectSpread({
+_objectSpread(_objectSpread(_objectSpread(_objectSpread(_objectSpread({
   x
-}, y, {
+}), y), {}, {
```
Zzzen, line 9: x/y/a/b is not defined, thus not pure. Should we define them?

packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-spread/side-effect/output.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+var output = _objectSpread(_objectSpread(_objectSpread({}, pureA), {}, {
+  get foo() {},
+
+  get bar() {}
+
+}, pureB, {}, pureC, {}), impureFunc(), {}, pureD, {
+  pureD
+});
```
nicolo-ribaudo, line 35: I would expect this output to be
```js
var output = { ...pureA, get foo() {}, get bar() {}, ...pureB, ...pureC, ...impureFunc(), ...pureD, pureD }

var output = _objectSpread(
  _objectSpread(
    {},
    pureA,
    { get foo() {}, get bar() {} },
    pureB,
    {},
    pureC,
    {}
  ),
  impureFunc(),
  {},
  pureD,
  { pureD }
);
```

I think that it's a bug in the `isPure` implementation, that only handles `isClassMethod` instead of `isMethod` (that also catches object methods/accessors).

Zzzen: 🙆‍♂️ I'm gonna give it a try.

Issue #5089, opened 2017-01-11
Use cardinal for syntax highlighting code in babel-code-frame


mohsen1:
Currently, syntax highlighting is done inside babel-code-frame. Instead of maintaining your own colorizing use [cardinal](https://www.npmjs.com/package/cardinal) instead.

DrewML:
@mohsen1 Will happily accept a PR if it's keeping feature parity but improving the output.

Agathver:
Hi
We would like to work on this as the part of RGSoC 2017
@meghap9937 @lipsa7
This would be our teams first PR to get familiar with the code.

nicolo-ribaudo:
I'm closing this issue, since we just updated `js-tokens` and we are happy with how syntax highlighting is currently handled.

PR #11413, opened 2020-04-13
Improve syntax highlighting

lydell:
| Fixed Issues?            | (Nothing to link to.)
| Major: Breaking Change?  | Yes! Node.js 10+ required.
| Minor: New Feature?      | Not really
| Tests Added + Pass?      | Pass (there are no tests for token colors)
| Any Dependency Changes?  | js-tokens@4 → js-tokens@6

### Before

Screenshot from 2020-04-13 19-20-01

### After

Screenshot from 2020-04-13 19-20-13

### Changes

- Bugfix: Highlight await, implements, interface, let, package, private, protected, public, static, yield. (By using `isStrictReservedWord` instead of `isReservedWord`.)
- Update: Use js-tokens@6. This brings several improvements:
    - Support for JSX:
        - No more hacky workaround regex in this package.
        - JSXText is not highlighted as JS code anymore.
        - `>` and `}` inside JSXText are highlighted as invalid.
        - Multiline JSX strings are supported.
    - Invalid variable names are highlighted as invalid.
    - Regex vs division is properyly identified.
    - Support for BigInt syntax.
    - Support for highlighting tokens within template literal interpolation, and infinite nesting.

js-tokens@6 uses [Unicode Property Escapes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes), which is why **Node.js 10 or later is required.**

If you feel like doing a little bit of issues spring cleaning, you can probably close https://github.com/babel/babel/issues/5089.


lydell:
Travis failed, but it seems unrelated to this PR? https://travis-ci.com/github/babel/babel/jobs/318870243#L900-L930

nicolo-ribaudo:
@lydell `async` isn't a keyword (it's always a valid variable name), but we could special-case it in `@babel/code-frame`.

Also, we have fixed the failing tests on master.

lydell:
Thanks! I added support for as, async, from, get, of, set. TIL.

Those can of course be false positives, but the highlighter will give false positives for `x.if` and `({if: true})` also. I think it’s better to highlight too much than too little when it comes to keywords.

PR #11423, opened 2020-04-15
Support custom parserOpts for babel parser in @babel/eslint-parser

vikr01:
| Major: Breaking Change?  | N/A
| Documentation PR Link    | 

The README for `@babel/eslint-parser` states:

> - `babelOptions` passes through Babel's configuration [loading](https://babeljs.io/docs/en/options#config-loading-options) and [merging](https://babeljs.io/docs/en/options#config-merging-options) options (for instance, in case of a monorepo). When not defined, @babel/eslint-parser will use Babel's default configuration file resolution logic.

However, this hasn't exactly been true, because you can't override *all* of the babel parser's options. With this, there's now capability to use neat features like the `flowComments` plugin to lint flow comments just like flow typings.

### DIFF
--- a/eslint/babel-eslint-parser/src/configuration.js
+++ b/eslint/babel-eslint-parser/src/configuration.js
@@ -30,18 +30,34 @@ export function normalizeBabelParseConfig(options) {
     exclude: options.babelOptions.exclude,
     ignore: options.babelOptions.ignore,
     only: options.babelOptions.only,
-    parserOpts: {
-      allowImportExportEverywhere: options.allowImportExportEverywhere,
-      allowReturnOutsideFunction: true,
-      allowSuperOutsideMethod: true,
-      ranges: true,
-      tokens: true,
-      plugins: ["estree"],
-    },
+    parserOpts: normalizeBabelParserOptions(options),
     caller: {
       name: "@babel/eslint-parser",
     },
   };
+
+  function normalizeBabelParserOptions(options) {
+    const parserOpts = options.babelOptions.parserOpts || {};
+    const parserPlugins = parserOpts.plugins || [];
+
+    const normalizedParserOptions = Object.assign(
+      {
+        allowReturnOutsideFunction: true,
+        allowSuperOutsideMethod: true,
+        ranges: true,
+        tokens: true,
+      },
+      parserOpts,
+    );
+
+    normalizedParserOptions.plugins = ["estree", ...parserPlugins];
+    if (options.allowImportExportEverywhere !== undefined) {
+      normalizedParserOptions.allowImportExportEverywhere =
+        options.allowImportExportEverywhere;
+    }
+
+    return normalizedParserOptions;
+  }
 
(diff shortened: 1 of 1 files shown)


eslint/babel-eslint-parser/src/configuration.js
```diff
@@ -30,19 +30,35 @@ export function normalizeBabelParseConfig(options) {
...
+        allowReturnOutsideFunction: true,
+        allowSuperOutsideMethod: true,
+        ranges: true,
+        tokens: true,
+      },
+      parserOpts,
+    );
+
+    normalizedParserOptions.plugins = ["estree", ...parserPlugins];
+    if (options.allowImportExportEverywhere !== undefined) {
```
vikr01, line 54: So if `options.allowImportExportEverywhere` exists, it'll use that over `parserOpts. allowImportExportEverywhere` to prevent breaking changes.

eslint/babel-eslint-parser/src/configuration.js
```diff
@@ -30,19 +30,35 @@ export function normalizeBabelParseConfig(options) {
...
+      {
+        allowReturnOutsideFunction: true,
+        allowSuperOutsideMethod: true,
+        ranges: true,
+        tokens: true,
+      },
+      parserOpts,
+    );
+
+    normalizedParserOptions.plugins = ["estree", ...parserPlugins];
```
vikr01, line 53: Seems like the `"estree"` plugin is necessary for the eslint parser here, so I figured we shouldn't override it.

vikr01:
@danez You reviewed #9168 from me before, might you be able to review this?

I’d ideally like for us to start using this at Facebook. 🙂

nicolo-ribaudo:
@vikr01 We might want to do https://github.com/babel/babel/pull/11334#issuecomment-607522569 instead

existentialism:
> With this, there's now capability to use neat features like the flowComments plugin to lint flow comments just like flow typings.

Somewhat related, but for those who are using a Babel config, we could always support `flowComments` by augmenting the flow preset to enable comment parsing and/or a new syntax plugin... then with `babel-eslint@11` it'd just read the config?

(And yeah, I guess for the whole "passing config programmatically" story, we should roll with https://github.com/babel/babel/pull/11334#issuecomment-607522569)

vikr01:
@existentialism Agree — I think enabling the `flowComments` syntax via the flow preset would be nice.

vikr01:
@nicolo-ribaudo That looks to be for babel config plugins, this is for the syntax plugins. Don’t the two work differently, or am I mistaken?

nicolo-ribaudo:
@vikr01 My proposal was to allow passing a full Babel config, so you will also be able to pass `parserOpts` without special-casing it in our code.

JLHwung:
In https://github.com/babel/babel/pull/11639 we allowed pass through all babel options to `@babel/eslint-parser`. Closing this PR. @vikr01 Thanks for contributing!

PR #11426, opened 2020-04-15
Update chokidar to v3

eps1lon:
| Any Dependency Changes?  | `@babel/cli#chokidar`

Alternate to #10741 which was closed in favor of #10747. Unfortunately #10747 can't be backported to v7. 

Bumping chokidar is important because it has a transitive dependency on `kind-of@^3.2.2`  (via `chokidar#braces#snapdragon-*`). `kind-of` prior to 6.0.2 has a security vulnerability: https://github.com/advisories/GHSA-6c8f-qphg-qjgp.

### DIFF
--- a/packages/babel-cli/package.json
+++ b/packages/babel-cli/package.json
@@ -29,7 +29,7 @@
     "source-map": "^0.5.0"
   },
   "optionalDependencies": {
-    "chokidar": "^2.1.8"
+    "chokidar": "^3.3.1 || ^2.1.8"
   },
   "peerDependencies": {
     "@babel/core": "^7.0.0-0"


packages/babel-cli/package.json
```diff
@@ -29,7 +29,7 @@
     "source-map": "^0.5.0"
   },
   "optionalDependencies": {
-    "chokidar": "^2.1.8"
+    "chokidar": "^3.3.1 || ^2.1.8"
```
JLHwung, line 32: It will be still resolved to 3.3.1 on node.js 6, which is unsupported. Since `chokidar` is an optional dependency, you can install with `--no-optional` or use resolution to map it to `3.3.1`.

eps1lon: > you can install with --no-optional

That will still create a lockfile entry for it which triggers github security alerts.

> use resolution to map it to 3.3.1.

I'm doing that right now.

> It will be still resolved to 3.3.1 on node.js 6, which is unsupported.

JLHwung: > Are there plans in the future to closer align with node lts?

Yeah we will bump node requirements to `>= 10.13` on Babel 8: https://github.com/babel/babel/pull/10747

PR #11428, opened 2020-04-16
Implement `plugin-transform-react-pure-annotations `and add to `preset-react`

### LINKED ISSUE #11399: Automatically annotate `React.forwardRef` as PURE
## Feature Request

When building an app with React, `React.forwardRef` calls prevent tree shaking unused components. This can be fixed by adding a `/*#PURE*/` annotation so terser knows it's safe to remove. I think `@babel/preset-react` could do this automatically.

Either `@babel/plugin-transform-react-jsx` or a separate plugin could add `/*#PURE*/` comments when `forwardRef` calls are seen.

A separate plugin outside the babel monorepo. But then it would be up to individual apps to adopt that plugin rather than getting this for free just by upgrading babel.

Would happen automagically. 😉

devongovett:
| Fixed Issues?            | Closes #11399
| Documentation PR Link    | 
| Any Dependency Changes?  |

This implements `@babel/plugin-transform-react-pure` as described in #11399, and adds it by default to `@babel/preset-react`. This plugin adds `/*#__PURE__*/` annotations before many top-level React function calls, which allows them to be tree shaken/dead code eliminated by terser and other minifiers.

For example, libraries making use of `React.forwardRef` or `React.memo` could be tree shaken if unused. For apps using large libraries of components that use these APIs, this could make a big difference to overall bundle size.

This is built into `@babel/preset-react` by default for maximum impact across the ecosystem. However, it can be disabled by setting the `{"pure": false}` option to the preset, which also already disables pure annotations on compiled JSX.


nicolo-ribaudo:
The CI failure doesn't seem related to this PR, but it doesn't fail on master and it fails on every node version :thinking:

nicolo-ribaudo:
Ok, I restarted travis on master and it failed.

devongovett:
Updated the plugin name. I'm not sure what's up with the tests. They do pass locally...

PR #11432, opened 2020-04-16
Don't include "resolve" in @babel/standalone

nicolo-ribaudo:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Starting from https://github.com/browserify/resolve/pull/217, `resolve` unconditionally uses the `fs` object (even if we don't call any `resolve` function).

This is a problem in `@babel/standalone` since `fs` isn't available there. However, `resolve` shouldn't have been included in `@babel/standalone` in the first place :shrug:

### DIFF
--- a/packages/babel-plugin-transform-runtime/package.json
+++ b/packages/babel-plugin-transform-runtime/package.json
@@ -12,7 +12,8 @@
     "babel-plugin"
   ],
   "browser": {
-    "./lib/get-runtime-path/index.js": "./lib/get-runtime-path/browser.js"
+    "./lib/get-runtime-path/index.js": "./lib/get-runtime-path/browser.js",
+    "./src/get-runtime-path/index.js": "./src/get-runtime-path/browser.js"
   },
   "dependencies": {
     "@babel/helper-module-imports": "^7.8.3",
--- a/packages/babel-plugin-transform-runtime/package.json
+++ b/packages/babel-plugin-transform-runtime/package.json
@@ -11,6 +11,9 @@
   "keywords": [
     "babel-plugin"
   ],
+  "browser": {
+    "./lib/get-runtime-path/index.js": "./lib/get-runtime-path/browser.js"
+  },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


existentialism:
Is there any way to add a test for this?

nicolo-ribaudo:
I tried to search for a rollup option to disallow some modules, but I couldn't find it.

packages/babel-plugin-transform-runtime/package.json
```diff
@@ -11,6 +11,10 @@
   "keywords": [
     "babel-plugin"
   ],
+  "browser": {
```
JLHwung, line 14: Seems like it suffices to set [`preferBuiltins: false`](https://github.com/rollup/plugins/tree/master/packages/node-resolve#preferbuiltins) when building babel-standalone, so it can warn us on builtin imports.

JLHwung: Oh but it will not cover cases like `resolve` is imported. Maybe we can do something like
```js
{
  name: "warn-browser-unsupported-modules",
  resolveId(...args) {
    const [source] = args;
    if (["resolve"].includes(source)) {
      throw new Error(`${source} can not be bundled into @babel/standalone!`);
    }
    return this.resolve(...args);
  }
}
```

nicolo-ribaudo: Yeah, I'm merging this for now and I will investigate. I'm surprised that I can't find a plugin that does what we need :thinking:

JLHwung: @nicolo-ribaudo You may checkout https://github.com/dumbmatter/rollup-plugin-blacklist but I am not strongly for that.

nicolo-ribaudo: May I ask why don't you like it?

PR #11434, opened 2020-04-16
[`preset-env`] Add `browserslistEnv` option

### LINKED ISSUE #8834: Option to specify env for browserslist
## Feature Request

I am using `babel-preset-env` to transpile my JS source into two bundles. One for ES5 browsers and another for modern browsers. I would like to use browserslistrc for this purpose with two [environments](https://github.com/browserslist/browserslist#environments) defined for `es5` and `es6`. Selecting the correct environment requires setting the `BROWSERSLIST_ENV` environment variable or passing the `env` option. I cannot use the `BROWSERSLIST_ENV` as both versions are transpiled by the same gulp job.

I would like to use the [env](https://github.com/browserslist/browserslist#js-api) option of browserslist, which is currently not supported by babel.

I would like to introduce a new option to `babel-preset-env` , which will be passed as the `env` option to browserslist.

Alternative is to use `BROWSERSLIST_ENV` or `NODE_ENV`. But for a gulp watch job which handles multiple environments, this is not really an option.

AndrewLeedham:
Adds a new option to `preset-env`, which allows the user to specify an environment which is passed forward to browserslist. Continuation of the work from https://github.com/babel/babel/pull/9161 based on this comment: https://github.com/babel/babel/pull/9161#issuecomment-586529537

| Fixed Issues?            |  Fixes #8834, Closes #9161
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | https://github.com/babel/website/pull/2227
| Any Dependency Changes?  |


nicolo-ribaudo:
The failing tests should be fixed on master.

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -297,5 +313,10 @@ export default function normalizeOptions(opts: Options) {
     spec: validateBoolOption(TopLevelOptions.spec, opts.spec, false),
     targets: normalizeTargets(opts.targets),
     useBuiltIns: useBuiltIns,
+    browserslistEnv: validateStringOption(
+      TopLevelOptions.browserslistEnv,
+      opts.browserslistEnv,
+      "production",
```
JLHwung, line 319: Let's unset the default values here, otherwise it will be a breaking change.

Here is how `browserslist` picks its env https://github.com/browserslist/browserslist/blob/adcb789638f8d50fbdd6fe53db164e9cd4bb4444/node.js#L66

Basically it is `env > NODE_ENV > BROWSERSLIST_ENV > "production"`. Now if users are using the following config
```
[development]
last 1 chrome

[production]
last 1 version
```

Running `NODE_ENV=development babel --presets env` will have `browserslist` picked `development` as the `env`, however after the defaults are assigned here, `browserslist` will now pick `production` over `process.env.NODE_ENV`.

AndrewLeedham: Fair point

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -154,6 +154,22 @@ export const validateBoolOption = (
...
+export const validateStringOption = (
+  name: string,
+  value?: string,
+  defaultValue?: string,
+) => {
+  if (typeof value === "undefined") {
+    value = defaultValue;
+  }
+
+  if (typeof value !== "string" && typeof value !== "undefined") {
```
JLHwung, line 166:
```suggestion
  else if (typeof value !== "string") {
```

AndrewLeedham: done

packages/babel-helper-compilation-targets/test/.browserslistrc
```diff
@@ -0,0 +1,2 @@
+[custom]
```
JLHwung, line 1: nit: Let's move it to a separate folder (i.e. `test/custom-browserslists-env/.browserslistrc`) so that it will not potentially interfere with other tests.

existentialism: As I mentioned in the previous PR, I definitely think we should do this

PR #11437, opened 2020-04-17
Support data-native-module="true" to generate native <script type="module">

dfabulich:
If you want to use your browser's native ES module support, the `<script>` tag needs to have a `type="module"` attribute. Now, babel standalone can generate native module scripts with `<script type="text/babel" data-native-module="true">`

```
<!doctype html>
<html>
<head>
  <title>Babel Test</title>
</head>
<body>
  <script src="./babel.js"></script>
  <script type="text/babel" data-presets="react" data-native-module="true">
    /** @jsx h */
    import { h, Component, render } from 'https://unpkg.com/preact?module';
    render(<h1>Hello World!</h1>, document.body);
  </script>
</body>
</html>
```

| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Tests Added + Pass?      | 
| Documentation PR Link    | 
| Any Dependency Changes?  |


nicolo-ribaudo:
I'd like to discuss about this a bit more. I think that a `data-type="module"` (or `data-compiled-type`) option would be better, and `@babel/register` should do something like `scriptEl.setAttribute("type", scriptEl.getAttribute("data-type"))`.

existentialism:
`data-type` might be _too_ generic, `data-source-type`?

nicolo-ribaudo:
`data-source-type` seems like it implies something about the type of the original source code rather than the compiled one.

I proposed `data-type` because the "real" attribute is called `type`. Maybe one of these?
- `data-final-type`
- `data-compiled-type`
- `data-output-type`

existentialism:
Good point.

I guess `data-type` is fine... short of `data-script-type` 😆

PR #11449, opened 2020-04-20
Set exprAllowed to false for star token

### LINKED ISSUE #11435: Anonymous generator function expression with typescript generic fails to parse
## Bug Report

**Current Behavior**

Anonymous generator functions with typescript generics fail to parse:

```typescript
const fn = function* <T>(input: T): Generator<number> {
  yield 2;
}
```

`SyntaxError: Unexpected token, expected "(" (1:21)`

When adding a name to the function it works as expected:

```typescript
const fn = function* fn<T>(input: T): Generator<number> {
  yield 2;
}
```

TSC parses anonymous generator functions with generics [as expected](https://www.typescriptlang.org/play/index.html?ssl=1&ssc=1&pln=4&pc=1#code/MYewdgzgLgBAZmGBeeBXMwoEtwCoYA8AKgHwAUWYADqlAFwxECUDA4gKZjsBOAhlCG4EwqALYAjHiRgBvAFAwYATyzsANgBMYAJgDccgL5ygA).

**Input Code**
- [REPL](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAZmGBeeBXMwoEtwCoYA8AKgHwAUWYADqlAFwxECUDA4gKZjsBOAhlCG4EwqALYAjHiRgBvAFAwYATyzsANgBMYAJgDccgL5ygA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2%2Ctypescript&prettier=false&targets=&version=7.9.0&externalPlugins=)

```typescript
const fn = function* <T>(input: T): Generator<number> {
  yield 2;
}
```

**Expected behavior/code**

Should parse as an anonymous generator function with generic.

**Additional info**

Appears that the current behavior is correct for function declarations, but not function expressions. TSC [throws a parse error](https://www.typescriptlang.org/play/index.html#code/GYVwdgxgLglg9mAVAAgDwBUB8AKGYAOIUAXMugJSkDiApmDQE4CGUcDqYIAtgEaObIA3gChkyAJ4waAGwAmyAEwBuYQF9hQA) when the function is parsed as a declaration, but not when parsed as an expression (as above).

**Environment**

```
- Babel version(s): v7.9.4
- Node/npm version: 12
- OS: macOS
- Monorepo: N/A
- How you are using Babel: `@babel/parser`

### PR DESCRIPTION
**Pull request by existentialism**
| Fixed Issues?            | Fixes #11435 

### CODE REVIEW

**Comment by nicolo-ribaudo** [comment]
Do we have a test for `a * /b/`? (where `/` starts an expression after `*`).

**Comment by existentialism** [comment]
@nicolo-ribaudo added some, couldn't find any

File: packages/babel-parser/test/fixtures/es2015/generators/yield-regex/input.js [hunk 1]
```diff
@@ -0,0 +1 @@
+function* x() { yield y * /z/ }
```
nicolo-ribaudo, line 1: In this test we are not actually yielding a regexp. Did you mean something like `yield* /z/`?

existentialism: yeah sorry

nicolo-ribaudo:
@existentialism Could you update the flow tests whitelist file? https://github.com/facebook/flow/blob/master/src/parser/test/flow/async_generators/migrated_0007.js isn't failing anymore.

existentialism:
@nicolo-ribaudo good catch

PR #11450, opened 2020-04-20
Update compat data

JLHwung:
| Fixed Issues?            | Unblocks #11451
| Any Dependency Changes?  | Updated `compat-table` to latest commit


packages/babel-compat-data/data/plugin-bugfixes.json
```diff
@@ -1,125 +1,125 @@
 {
   "transform-async-to-generator": {
     "chrome": "55",
+    "opera": "42",
```
JLHwung, line 4: The change of support matrix is related to object key orders only.

hzoo: dang that's annoying haha

PR #11456, opened 2020-04-21
remove `< x.y.z` engines specification

JLHwung:
| Documentation PR Link    | 

Here we remove the `< x.y.z` in the root project engine specifications. This also aligns to what we did in `@babel/core`. Hopefully it should fix our failing master due to the release of Node.js 14.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ env:
 
 install: skip
 before_script:
-  - curl -o- -L https://yarnpkg.com/install.sh | bash
   - if [ "$TRAVIS_OS_NAME" = "windows" ]; then choco install make; fi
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -13,7 +13,6 @@ env:
 
 install: skip
 before_script:
-  - curl -o- -L https://yarnpkg.com/install.sh | bash
```
JLHwung, line 16: Also removed `yarn` update because:

1) after these years the bundled `yarn` version in CI is reasonably new.
2) `yarnpkg.com` could fail, i.e. https://travis-ci.com/github/JLHwung/babel/jobs/321733324

PR #11467, opened 2020-04-22
chore: test against node.js 14

JLHwung:

Update our build matrix

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,12 +40,12 @@ aliases:
 executors:
   node-executor:
     docker:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -40,12 +40,12 @@ aliases:
 executors:
   node-executor:
     docker:
-      - image: circleci/node:13
+      - image: circleci/node:latest
```
JLHwung, line 43: Currently the `latest` tag is still resolved to `13.10`. After CircleCI updates their docker image, we will move to `14` automatically.

PR #11468, opened 2020-04-22
Remove backward compatibility for union types

Beraliv:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | No 
| Patch: Bug Fix?          | No
| Major: Breaking Change?  | Yes ⚠️
| Minor: New Feature?      | No
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  | No
| License                  | MIT

Several requests regards TypeScript union types saved backward compatibility
Here I remove them for major update (Babel 8)

Related issue:
* https://github.com/babel/babel/issues/9530

Related PRs:
* https://github.com/babel/babel/pull/11378
* https://github.com/babel/babel/pull/11448

PR #11477, opened 2020-04-24
Disable e2e-babel-old-version test

nicolo-ribaudo:

Since Babel 8 is expected to break compatibility with Babel 7.0.0, it doesn't make sense to test it.

For example, it (correctly) fails in https://github.com/babel/babel/pull/11468


nicolo-ribaudo:
That's why I commented the code instead of deleting it!

PR #11479, opened 2020-04-24
Canary build of packages using CodeSandbox CI + mini repl

hzoo:
Done with @nicolo-ribaudo! Got to learn more about Next, CodeSandbox, babel-standalone 😄  

> Did some tests earlier on local fork https://github.com/hzoo/babel/pull/3

Current Babel template (forked @CompuIves one in https://github.com/babel/babel/pull/10639): https://codesandbox.io/s/babel-repl-9d0wg

- not using babel-standalone since I don't want to bundle (takes too long), so it just publishes the built files
- we just manually required all the plugins/presets so they are available

---

cc @CompuIves 

Integrating this got me thinking about some feature requests:
- being able to change the codesandbox bot comment
- can you programmatically generate templates based on the PR

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,5 +1,5 @@
 {
-  "installCommand": "bootstrap",
+  "installCommand": "codesandbox",
   "buildCommand": false,
-  "sandboxes": ["kypop"]
+  "sandboxes": ["kypop", "hk4lv"]
 }
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,6 +1,6 @@
 {
   "installCommand": "codesandbox",
   "buildCommand": false,
-  "sandboxes": ["kypop", "hk4lv"],
+  "sandboxes": ["9d0wg"],
   "packages": ["packages/*"]
 }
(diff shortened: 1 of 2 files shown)


.codesandbox/ci.json
```diff
@@ -1,5 +1,6 @@
 {
-  "installCommand": "bootstrap",
+  "installCommand": "codesandbox",
```
JLHwung, line 2: Can we just run `build`? The `build-no-bundle` run `babel` on every package while `build` will rollup `babel-parsers`.

nicolo-ribaudo: The problem is that `build` was too slow and froze codesandbox

hzoo: yeah we purposely didn't because we don't need it in this case, and it was too slow for some reason: https://ci.codesandbox.io/status/hzoo/babel/pr/1

JLHwung: I guess it comes from building `babel-standalone`.

An ugly workaround is `BABEL_COVERAGE=true make build`, which will skip building standalone.

nicolo-ribaudo:
Henry, do you remember that thing about ✔️s before merging PRs? 😛

hzoo:
Oh yeah I was assuming that, but noted 👍

PR #11485, opened 2020-04-26
Parse eventual-send (tildot) syntax

### LINKED ISSUE #11348: Proper eventual-send (wavy-dot, aka tildot) support
## Feature Request

The only problem is wanting to land the new ["wavy-dot" eventual send syntax](https://github.com/tc39/proposal-wavy-dot) (currently stage1) in Babel.

Add and implement an `["eventualSend"]` parser option.

`babel-plugin-syntax-eventual-send` and `babel-plugin-proposal-eventual-send` would be the next steps.

I had a quick hack to get this working (without a clean factoring) in #10115.  This current proposal is to implement it the right way according to Babel conventions.

Enabling `babel-plugin-syntax-eventual-send` will enable the `"eventualSend"` parser plugin.  That will emit new nodes in alignment with the [proposed syntax](https://github.com/tc39/proposal-wavy-dot#proposed-syntax):

```js
// foo~.bar
{ type: "AdjectiveMemberExpression", adjective: '~', object: foo, property: bar, computed: false }
// foo~.[i]
{ type: "AdjectiveMemberExpression", adjective: '~', object: foo, property: i, computed: true }
// foo~.(...args)
{ type: "AdjectiveCallExpression", adjective: '~', callee: foo, arguments: args }
```

The reason for the `Adjective` prefix (and `adjective` property) is to enable future expansion into combinations of optional chaining with eventual send, for adjectives `'~?'` and `'?~'` (which yields a combinatorial explosion if another syntax proposal follows this pattern).

Secondly, a new `babel-plugin-proposal-eventual-send` will actually implement `AdjectiveCallExpression|AdjectiveMemberExpression` in terms of HandledPromise static methods, including the needed pipelining semantics of optimising a member followed by call as a "method", and rewriting `ExpressionStatement` expressions whose value is discarded as "send-only" methods.

Shout out to: @erights, @FUDCo, @kriskowal.
I'd appreciate feedback from: @loganfsmyth, @jridgewell

michaelfig:
| Fixed Issues?            | Fixes #11348
| Documentation PR Link    | 

This implements the parser changes described in #11348 to implement the stage1 TC39 eventual-send syntax proposal described in that issue.

I hope to land these changes soon, to experiment with plugins that use the tildot syntax.  When the proposal is more mature, we will make a new PR to merge the appropriate plugins.


nicolo-ribaudo:
Is the difference between `(a~.b).c` and `a~.b.c` observable?

michaelfig:
> Is the difference between `(a~.b).c` and `a~.b.c` observable?

No, it is not.

nicolo-ribaudo:
Ok, then we don't need the `eventual: boolean` property on `EventualMemberExpression`. For optional chaining, is needed to disambiguate between `a?.b.c` (two `OptionalMemberExpression`s, the inner one has `optional: true` and the outer one has `optional: false`) from `(a?.b).c` (one `OptionalMemberExpression` with `optional: true`, and one `MemberExpression`.

`a~.b.c` can be represented as an `EventualMemberExpression` and a `MemberExpression`.

michaelfig:
> Ok, then we don't need the `eventual: boolean` property on `EventualMemberExpression`.

Thanks for catching this.  I've updated accordingly.

> Also, since this PR is quite big, can we keep the transform implementation for a separate PR 🙏

You bet!

packages/babel-parser/src/parser/expression.js
```diff
@@ -556,6 +556,7 @@ export default class ExpressionParser extends LValParser {
   ): N.Expression {
     const state = {
       optionalChainMember: false,
+      eventualMember: false,
```
nicolo-ribaudo, line 559: Is this still used now that you have removed the `eventual` property from the AST?

packages/babel-generator/src/generators/expressions.js
```diff
@@ -139,6 +139,43 @@ export function OptionalCallExpression(node: Object) {
...
+    computed = true;
+  }
+  this.token("~.");
+
+  if (computed) {
+    this.token("[");
+    this.print(node.property, node);
+    this.token("]");
+  } else {
+    if (!node.optional) {
```
JLHwung, line 160: The `optional` property has been removed from AST so `!node.optional` is always `true`.

michaelfig:
I'll have to defer to @erights on these details.

> How does this proposal interact with optional chains? In `a?.b~.c.d`, are `~.c` and `.d` accessed when `a` is nullish? Or is it equivalent to `(a?.b)~.c.d`? Is there an operator to do `?.` and `~.` at the same time (e.g. `a?~.b`)?

> I think that this heavily affects the AST design.

Agreed.

> I looked at the proposal's README, and realized that my question was incomplete. Since `a~.b()` and `(a~.b)()` are different (from what I understood, they are `a.[[ApplyMethodSend]]('b', [])` and `a.[[GetSend]]('b')()`), we would probably need a new `EventualMemberCallExpression` node if we go with the current AST design.

I had been thinking that `EventualMemberExpression` and `EventualCallExpression` would be sufficient and that the equivalent of `EventualMemberCallExpression` could be introduced during the AST visitor after the parser is done.  That may not be the case.

packages/babel-parser/src/tokenizer/index.js
```diff
@@ -707,6 +707,22 @@ export default class Tokenizer extends LocationParser {
     }
   }
 
+  readToken_tilde(): void {
+    // '~'
+    const next = this.input.charCodeAt(this.state.pos + 1);
+    const next2 = this.input.charCodeAt(this.state.pos + 2);
```
JLHwung, line 713: Can you move `next2` inside the `next is dot` branch? So we don't introduce an extra lookahead when next is not dot.

packages/babel-parser/src/parser/expression.js
michaelfig: No, it isn't.  Thanks!

packages/babel-parser/src/types.js
```diff
@@ -537,6 +537,14 @@ export type MemberExpression = NodeBase & {
   computed: boolean,
 };
 
+export type EventualMemberExpression = NodeBase & {
```
JLHwung, line 540: w.l.o.g. I will take `EventualMemberExpression` as an example, likewise `EventualCallExpression`.

I am hesitant about the extra node type `EventualMemberExpression` and likely-in-the-future `EventualOptionalMemberExpression` / `OptionalEventualMemberExpression` given that optional chaining has reached stage 4.

The `wavy-dot` proposal introduces infix operators `~.` similar to optional chaining. I understand that it may seem reasonable to add a new node type, which is what we did in `OptionalMemberExpression`. Essentially `OptionalMemberExpression` is a `MemberExpression`. The introduction of the new node type `OptionalMemberExpression` is due to 1) the new optional semantic and 2) the short-circuit behaviour. The latter requires that a new node type has to be introduced because we need to know when the base is nullish, we should skip which parts of the whole optional chains.

`EventualMemberExpression` is also a `MemberExpression`. Similar to `OptionalMemberExpression`, it introduces new eventual semantics. Besides that I don't see there are extra behaviours which we should track on the AST level. Therefore I propose to add an `eventual: boolean` property to the existing `MemberExpression`

```js
interface MemberExpression <: Expression, Pattern {
  type: "MemberExpression";
  object: Expression | Super;
  property: Expression;
  computed: boolean;
  eventual: boolean;
}
```

Since the eventual semantic is additive, we can also add `eventual: boolean` once the proposal supports optional chaining.

nataraj3:
Who are you


Sent from my Samsung Galaxy smartphone.
-------- Original message --------From: Michael FIG <notifications@github.com> Date: 27/04/2020  07:49  (GMT+05:30) To: babel/babel <babel@noreply.github.com> Cc: Subscribed <subscribed@noreply.github.com> Subject: Re: [babel/babel] Parse eventual-send (tildot) syntax (#11485) 

@michaelfig commented on this pull request.



In packages/babel-parser/src/parser/expression.js:
> @@ -556,6 +556,7 @@ export default class ExpressionParser extends LValParser {
   ): N.Expression {
     const state = {
       optionalChainMember: false,
+      eventualMember: false,

No, it isn't.  Thanks!

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub, or unsubscribe.
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/babel/babel/pull/11485#discussion_r415465284",
"url": "https://github.com/babel/babel/pull/11485#discussion_r415465284",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

packages/babel-parser/src/types.js
michaelfig: I'll remove for `EventualCallExpression` and `EventualMemberExpression`.  I think we still need `EventualMemberCallExpression`, which I would rename to `EventualSendExpression`.  I'll wait on that until you post your updated suggestion.

michaelfig:
> Who are you

I work with Mark S. Miller at Agoric.  If you need to know more than that, take it to keybase.io.

packages/babel-parser/src/types.js
michaelfig: Also, just a note:

`.?~` is sensible (only eventual-send if the lhs is not nullish), but `.~?` is not sensible (since eventual-send always returns a Promise, it cannot be nullish).

michaelfig: > Updates: As @nicolo-ribaudo points out `(a~.b)()` is `a~.b()` is not equivalent. So adding `eventual: boolean` is not enough. I will post an updated suggestion tomorrow. I am still negative on adding new node types.

@JLHwung have you had a chance to think about this?  I definitely want to implement the PR in the correct way, but am blocked until I get some design advice.

Thanks!

JLHwung: @michaelfig My approach is now based on the `ChainingExpression` approach proposed by ESLint team: See https://gist.github.com/JLHwung/6ec08a87e4da88874c50788c37d6fdf4 for detailed examples.

Note that this approach is tremendously different from current babel's approach and we will discuss if we will go for that path in the upcoming babel meetings. Before any decision is made, I would suggest we hold on a bit until more feedback from the AST shape of `wavy-dot`.

michaelfig: [From the above, I found estree/estree#204 which is long, but seems to be giving some traction to https://gist.github.com/mysticatea/5453e721892b9c5a621e445f5fcdf62f]

michaelfig: > @michaelfig My approach is now based on the `ChainingExpression` approach proposed by ESLint team: See https://gist.github.com/JLHwung/6ec08a87e4da88874c50788c37d6fdf4 for detailed examples.
> 
> Note that this approach is tremendously different from current babel's approach and we will discuss if we will go for that path in the upcoming babel meetings. Before any decision is made, I would suggest we hold on a bit until more feedback from the AST shape of `wavy-dot`.

@JLHwung, has this discussion settled?  If so, what does it mean for the kind of AST shape I should target?

michaelfig:
@nicolo-ribaudo and @JLHwung, I'd like to see this PR move forward again.  Would you be able to help me finish the AST design?

> > How does this proposal interact with optional chains?

Optional chains can short-circuit eventual sends.

> In `a?.b~.c.d`, are `~.c` and `.d` accessed when `a` is nullish?

They are short-circuited.  `a?.b~.c.d` when `a` is nullish is just `a`.

> Or is it equivalent to `(a?.b)~.c.d`?

`(a?.b)~.c.d` is implemented as `a?.b.[[GetSend]]('c').d` which when `a` is nullish results in `Promise.rejected(TypeError(`Cannot read property 'c' of undefined`)).d` which is just `undefined` since promises usually don't have a `d` property.

> Is there an operator to do `?.` and `~.` at the same time (e.g. `a?~.b`)?

Yes, we would like that.  `a?~.b` would result in `a` if `a` is nullish, otherwise `a~.b`.

> > I looked at the proposal's README, and realized that my question was incomplete. Since `a~.b()` and `(a~.b)()` are different (from what I understood, they are `a.[[ApplyMethodSend]]('b', [])` and `a.[[GetSend]]('b')()`)

That's correct.

> we would probably need a new `EventualMemberCallExpression` node if we go with the current AST design.

That seems right to me.

JLHwung:
Let's transfer the discussion about composition between optional chain and wavy-dot to the proposal repo. I think we have to hold off a bit until the draft clearly specifies the proposed syntax change to [LeftHandSideExpression](https://tc39.es/ecma262/#prod-LeftHandSideExpression). 

Because

> this heavily affects the AST design

as @nicolo-ribaudo mentioned before.

PR #11492, opened 2020-04-27
docs: update AST spec

JLHwung:

Added `BigIntLiteral` and logical assignment operators. Also refined some interface declarations.

I leave `OptionalCallExpression` undocumented as in ESTree it is still highly debated and we should decide in the future whether we will go with the ESTree design.

### DIFF
--- a/packages/babel-parser/ast/spec.md
+++ b/packages/babel-parser/ast/spec.md
@@ -10,6 +10,7 @@ These are the core @babel/parser (babylon) AST node types.
   - [StringLiteral](#stringliteral)
   - [BooleanLiteral](#booleanliteral)
   - [NumericLiteral](#numericliteral)
+  - [BigIntLiteral](#bigintliteral)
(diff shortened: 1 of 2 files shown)


packages/babel-parser/ast/spec.md
```diff
@@ -1024,7 +1034,7 @@ interface ParenthesizedExpression <: Expression {
 }
 ```
 
-An expression wrapped by parentheses.
+An expression wrapped by parentheses. By default `@babel/parser` does not create this node, unless passed by the `createParenthesizedExpressions: true` option.
```
existentialism, line 1037:
```suggestion
An expression wrapped by parentheses. By default `@babel/parser` does not create this node, unless the `createParenthesizedExpressions: true` option is passed.
```

nicolo-ribaudo:
I wonder if we could somehow auto-generate this file by adding JSDoc to the definitions in `@babel/types`

PR #11500, opened 2020-04-29
Update Flow to 0.123.0

nicolo-ribaudo:
| Documentation PR Link    | 
| Any Dependency Changes?  | devDep :arrow_up: 

This was harder than usual :confused:

### DIFF
--- a/packages/babel-helper-compilation-targets/src/index.js
+++ b/packages/babel-helper-compilation-targets/src/index.js
@@ -13,7 +13,7 @@ import {
 } from "./utils";
 import { browserNameMap } from "./targets";
 import { TargetNames } from "./options";
-import type { Targets, InputTargets, Browsers } from "./types";
+import type { Target, Targets, InputTargets, Browsers } from "./types";
 
 export type { Targets, InputTargets };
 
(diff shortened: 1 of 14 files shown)


packages/babel-helper-compilation-targets/src/index.js
```diff
@@ -170,30 +168,34 @@ type ParsedResult = {
...
-  delete targets.esmodules;
+  delete inputTargets.esmodules;
 
   // Parse browsers target via browserslist
-  const browsersquery = validateBrowsers(targets.browsers);
+  const browsersquery = validateBrowsers(inputTargets.browsers);
+  delete inputTargets.browsers;
+
+  // $FlowIgnore
+  let targets = (inputTargets: Targets);
```
nicolo-ribaudo, line 195: It was complaining a lot because `esmodules`/`browsers` are not part of `Targets`, which should contain the _resolved_ targets.

packages/babel-helper-compilation-targets/src/options.js
```diff
@@ -14,7 +15,4 @@ export const TargetNames = {
   android: "android",
   electron: "electron",
   samsung: "samsung",
-
-  // deprecated
-  uglify: "uglify",
```
nicolo-ribaudo, line 19: In practice, we handle this by deleting `targets.uglify` before using this object.

nicolo-ribaudo:
Ok I broke something

nicolo-ribaudo:
I had to refactor a bit `@babel/preset-env` to make the flow errors disappear without breaking anything, but I'm satisfied of the result. Now we have a better separation between the input targets (with `esmodules` and `browsers`) and the resolved targets, so we don't have to manually exclude `browsers` when iterating.

PR #11508, opened 2020-05-02
Add persistent cache to babel-core

### LINKED ISSUE #10838: Add persistent caching support to babel core
## Feature Request

Babel takes a significant amount of time when run from command line. This became especially frustrating as I was using it to transpile TypeScript. It was nice that I could configure different tools to support that simply by enabling their support for Babel. But it seemed wasteful that they kept performing the same conversions over and over again on subsequent command line statements etc. Caching should enable old fashioned make-like incremental building shared between all tools using Babel.

I would like to see an option added to enable caching to disk. Both for CLI calls and through Babel config files etc. Care will need to be taken to use proper hash keys, to ensure cache validity (e.g: catch incompletely written files), to reasonably optimize performance and to purge deprecated files (e.g: LRU caching). Unfortunately with different tools using Babel with different options cache might not be efficiently shared between those.

Caching is supported by babel-loader and babel-register but those (AFAIK) are not supported for use by tools like Karma, Rollup, es-dev-server and Storybook.

#5372 seems similar, but not the same as I'd like to be able to configure this for all tools at once.

When using the CLI I'm thinking about arguments like:

- `--cache` - to enable caching
- `--cacheDir`- to specify where to store cache files (default: node_modules/.cache/babel)
- `--cacheSize` - to specify maximum cache size

And similar options in config files and for related methods.

** Proof of concept **
I quickly hacked together a proof of concept implementation which reduced test execution time from about 14s to 7s. I've attached the modified @babel/core v7 lib/transformation/index.js below with rudimentary caching support added to the runSync() function
[babel-core-transformation-caching.zip](https://github.com/babel/babel/files/3935479/babel-core-transformation-caching.zip)
.

jridgewell:
| Fixed Issues?            | Fixes #10838
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Tests Added + Pass?      | Not yet
| Documentation PR Link    |
| Any Dependency Changes?  |

This adds a persistent file cache. Testing on my work project, this reduces test transpiling from ~50s to ~20, and a preset-env transpile from 160s to 110s.

Each file is salted with the resolved configuration for the file, and with an optional shared "salt" object that can be passed via configuration.

We may want to open up the API to allow generic file-caches to be passed in, but that'll take a bit more work.


packages/babel-core/package.json
```diff
@@ -52,6 +52,7 @@
     "gensync": "^1.0.0-beta.1",
     "json5": "^2.1.2",
     "lodash": "^4.17.13",
+    "mkdirp": "^1.0.4",
```
JLHwung, line 55: nit: please use `make-dir` to be consistent with `@babel/cli`.

jridgewell: Done.

packages/babel-core/src/config/validation/option-assertions.js
```diff
@@ -154,6 +155,45 @@ export function assertCallerMetadata(
...
+        const propLoc = access(saltLoc, prop);
+        const value = salt[prop];
+        if (
+          value != null &&
+          typeof value !== "boolean" &&
+          typeof value !== "string" &&
+          typeof value !== "number"
+        ) {
+          // NOTE(logan): I'm limiting the type here so that we can guarantee that
+          // the "caller" value will serialize to JSON nicely. We can always
```
nicolo-ribaudo, line 183: Nit: this isn't "caller"

PR #11512, opened 2020-05-02
Use ?. where it represents the intended semantics

nicolo-ribaudo:

Let's test our `?.` implementation in a real codebase! After https://github.com/babel/babel/pull/11500, it can be correctly type-checked by Flow.

I replaced this using a few regular expressions, to only replace `&&` where on the right there was a property access, a method call, or a comparison.

Also, I noticed that `?.` plays very bad with `.indexOf()` because it returns `-1` instead of `null`.

### DIFF
--- a/packages/babel-core/src/config/index.js
+++ b/packages/babel-core/src/config/index.js
@@ -17,7 +17,8 @@ export type { PartialConfig } from "./partial";
 
 const loadOptionsRunner = gensync<[mixed], Object | null>(function*(opts) {
   const config = yield* loadFullConfig(opts);
-  return config?.options;
+  // NOTE: Do not use ?., since we want to return "null" explicitly
+  return config ? config.options : null;
 });
 
 const maybeErrback = runner => (opts: mixed, callback: Function) => {
--- a/packages/babel-core/src/config/index.js
+++ b/packages/babel-core/src/config/index.js
@@ -17,7 +17,7 @@ export type { PartialConfig } from "./partial";
 
 const loadOptionsRunner = gensync<[mixed], Object | null>(function*(opts) {
   const config = yield* loadFullConfig(opts);
-  return config ? config.options : null;
+  return config?.options;
 });
 
 const maybeErrback = runner => (opts: mixed, callback: Function) => {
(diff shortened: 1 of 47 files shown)


packages/babel-core/src/transform-file.js
```diff
@@ -25,15 +25,7 @@ type TransformFile = {
...
-    let options;
-    if (opts == null) {
-      options = { filename };
-    } else if (opts && typeof opts === "object") {
-      options = {
-        ...opts,
-        filename,
-      };
-    }
+    const options = { ...opts, filename };
```
nicolo-ribaudo, line 28: This isn't really related, but since `opts` can only be `null` or an object these two things have the same behavior.

packages/babel-generator/src/buffer.js
```diff
@@ -315,10 +315,10 @@ export default class Buffer {
     const origFilename = targetObj.filename;
 
     targetObj.identifierName =
-      (prop === "start" && loc && loc.identifierName) || null;
-    targetObj.line = pos ? pos.line : null;
-    targetObj.column = pos ? pos.column : null;
-    targetObj.filename = (loc && loc.filename) || null;
+      (prop === "start" && loc?.identifierName) || null;
```
nicolo-ribaudo, line 318: Here I left `null` for the `prop !== "start"` case.

packages/babel-standalone/src/index.js
```diff
@@ -207,7 +207,7 @@ function onDOMContentLoaded() {
 
 // Listen for load event if we're in a browser and then kick off finding and
 // running of scripts with "text/babel" type.
-if (typeof window !== "undefined" && window && window.addEventListener) {
+if (typeof window !== "undefined" && window?.addEventListener) {
```
nicolo-ribaudo, line 210: Here `typeof` is still necessary in case `window` isn't defined.

JLHwung:
FYI there is codemod: https://github.com/NullVoxPopuli/optional-chaining-codemod

nicolo-ribaudo:
I love writing regexps! :stuck_out_tongue:

hzoo:
Nice! We should try to make some codemods for all proposals if possible!

packages/babel-core/src/config/index.js
```diff
@@ -17,6 +17,7 @@ export type { PartialConfig } from "./partial";
 
 const loadOptionsRunner = gensync<[mixed], Object | null>(function*(opts) {
   const config = yield* loadFullConfig(opts);
+  // NOTE: Do not use ?., since we want to return "null" explicitly
   return config ? config.options : null;
```
existentialism, line 21: could always do `config?.options ?? null`

PR #11514, opened 2020-05-04
[helpers] Add a private function name within `createSuper`

cpojer:

The `createSuper` helper was changed sometime between 7.4 and 7.9 to add a new private function within `createSuper`. We noticed in a large codebase that when a custom error instance gets thrown (like `throw new class CustomError extends Error {}`), the top of the stack frame blames to this internal function instead of the callsite. In this diff I propose adding a function name so that it isn't anonymous and can be filtered out by error reporting systems (when using source maps correctly). This diff should only add one byte in minified output for users of the ES2015 class transform.


jridgewell:
@cpojer: Can you update the tests? Run `OVERWRITE=1 make test-only`

cpojer:
Somehow I can't locally run the tests on my Facebook company laptop, possibly related to some watchman and Jest setup issue. Any chance you could update the test? 😬

nicolo-ribaudo:
@cpojer Can you allow edits from maintainers?

cpojer:
@nicolo-ribaudo should have already been enabled:
Screen Shot 2020-05-04 at 21 00 31

existentialism:
@nicolo-ribaudo @cpojer pushed updated fixtures

nicolo-ribaudo:
CI failure related to `preset-env` should be fixed on master.

PR #11520, opened 2020-05-04
Use single object spread call in loose mode

jridgewell:
| Fixed Issues?            | https://github.com/babel/babel/pull/11471#issuecomment-623399901
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    |
| Any Dependency Changes?  |

Does 2 things:

- Makes every `exec.js` test run in a new VM context (necessary for the new tests)
- Reuses the same object spread call in `loose` mode, making the output considerably smaller.

```js
// Old
var output = _extends(_extends(_extends(_extends(_extends(_extends({}, pureA), {}, {
  get foo() {},
  get bar() {}
}, pureB), pureC), impureFunc()), pureD), {}, {
  pureD
});

// New
var output = _extends({}, pureA, {
  get foo() {},
  get bar() {}
}, pureB, pureC, impureFunc(), pureD, {
  pureD
});
```


packages/babel-plugin-proposal-object-rest-spread/test/fixtures/object-spread-loose-builtins/no-object-assign-exec/exec.js
```diff
@@ -0,0 +1,43 @@
+"use strict";
+Object.defineProperty(Object.prototype, 'NOSET', {
```
jridgewell, line 2: These were polluting the shared context, causing the other tests to fail.

packages/babel-helper-transform-fixture-test-runner/src/index.js
```diff
@@ -19,31 +19,47 @@ import checkDuplicatedNodes from "babel-check-duplicated-nodes";
...
-const testContext = vm.createContext({
-  ...helpers,
-  process: process,
-  transform: babel.transform,
-  setTimeout: setTimeout,
-  setImmediate: setImmediate,
-  expect,
-});
-testContext.global = testContext;
+const contextModuleCache = new WeakMap();
```
JLHwung, line 22: Comparing the CI build time between [this PR](https://travis-ci.com/github/babel/babel/builds/163788731) and [master](https://travis-ci.com/github/babel/babel/builds/163835216), the unit test now costs 30% longer time. My gut is it comes from the cost of creating every single context and evaluating `@babel/polyfill` and `babel-helpers-in-memory`.

Instead of running common modules for every single `exec` test, can we leverage the [`cacheData`](https://nodejs.org/api/vm.html#vm_vm_runincontext_code_contextifiedobject_options) option? We can generate code cache after running common modules, and supply it to every single `exec` test context.

jridgewell:
> Instead of running common modules for every single exec test, can we leverage the cacheData option? We can generate code cache after running common modules, and supply it to every single exec test context.

Excellent idea! Done.

jridgewell:
> Can we separate this into a new PR?

Done.

cpojer:
Thank you for making this change, excited to try it out once it is released :)

PR #11531, opened 2020-05-07
Isolated exec tests

jridgewell:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Split from https://github.com/babel/babel/pull/11520, this run each `exec.js` test with a fresh global context to avoid pollution.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11531"></a>


jridgewell:
Some notes running a perf analysis:

`buildHelpers` takes a long time during initial load

Screen Shot 2020-05-07 at 1 01 40 AM

Some `run` calls do multiple `createContext` calls

Screen Shot 2020-05-07 at 1 02 36 AM

jridgewell:
Switching to the bundled `@babel/polyfill/dist/polyfill.min.js` fixes the perf regression entirely.

packages/babel-core/test/fixtures/plugins/multiple-definition-evaluation/exec.js
```diff
@@ -1,8 +1,8 @@
-var code = multiline([
```
JLHwung, line 1: nit: Can we also remove `multiline` from `packages/babel-helper-transform-fixture-test-runner/src/helpers.js`? It is never exported from `index.js`.

jridgewell: Not without a major bump. It's installed into the exec context, so it's available to tests. It's probably unused.

hzoo: If it's a concern around people using `babel-helper-transform-fixture-test-runner` as a dep I think we've never supported that as an API so it's fine (https://www.npmjs.com/package/@babel/helper-transform-fixture-test-runner). It's internal only essentially (we have no docs on it either), not that it's a good thing

PR #11532, opened 2020-05-07
Use rollup to bundle @babel/polyfill

jridgewell:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This results in several kb file savings when compressed:

| `polyfill.js` |   Before |    After |
|---------------|---------:|---------:|
| Size          | 239,724b | 208,909b |
| Size (Brotli) |  49,258b |  45,159b |

| `polyfill.min.js` |  Before |   After |
|-------------------|--------:|--------:|
| Size              | 99,310b | 78,250b |
| Size (Brotli)     | 29,243b | 25,380b |


nicolo-ribaudo:
Before merging this PR, we should check that the bundled polyfill passes the `core-js` test suite.

Rollup does many optimizations that are mostly not observable according to the standard, but often cause problems with feature detection.

hzoo:
Maybe could also inject this version of the polyfill into our own test runner, but would need to build that before any other tests are run

nicolo-ribaudo:
Also, please note that `@babel/polyfill` has been deprecated since 3303b079c596e56909adff4b834f098f5f1d8e2a (more than one year ago).

jridgewell:
I'll close since this is deprecated.

Issue #11166, opened 2020-02-24
Deep optional chaining breaks in function argument destructuring


maxdavidson:
## Bug Report

- [ ] I would like to work on a fix!

**Current Behavior**

Optional chaining breaks when destructuring deep chains inside the function arguments, due to incorrect hoisting of temporary variables.

[REPL](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=FAMwrgdgxgLglgewgAgBYFMA2mEAoDewyyAhgDRHIBGyAvKQPwB0VzUFAvgJTKHEBO6GGH4oADAG5gHYMAzY8-UgC5eHAAXcpQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Cenv&prettier=true&targets=Node-13&version=7.8.4&externalPlugins=)

**Input Code**

```js
function hello({
  a,
  b = a?.b?.c,
}) {
  return 0;
}
```

**Output Code**
```js
"use strict";

function hello({
  a,
  b = a === null || a === void 0 ? void 0 : (_a$b = a.b) === null || _a$b === void 0 ? void 0 : _a$b.c
}) {
  var _a$b;

  return 0;
}
```

**Expected behavior/code**

The code should assign a value to `b`, and not fail with `_a$b is not defined`.

**Babel Configuration**

```js
// .babelrc
{
  "presets": [
    [
      "@babel/preset-env",
      {
        "targets": { "node": "current" }
      }
    ]
  ]
}

```

**Environment**

- Babel version(s): 7.8.4
- Node/npm version: 13.8.0/6.13.7

**Possible Solution**

Either move the hoisted variable to the module scope, or use a temporary variable for the argument and perform the destructuring inside the function body.

jridgewell:
We keep having bugs related to injecting temp variables in a param scope.

nicolo-ribaudo:
Maybe in Babel 8 we could add a "transparent" scope (transparent similar to how we handle block scopes) to represent params scope.

Zzzen:
😕  What is the expected output though?

nicolo-ribaudo:
Probably something like this:
```js
function hello({
  a,
  b = (() => {
    var _a$b;
    return a === null || a === void 0 ? void 0 : (_a$b = a.b) === null || _a$b === void 0 ? void 0 : _a$b.c;
[...]

PR #11533, opened 2020-05-07
Fix temp vars injection for `?.` and `??` in function parameters

nicolo-ribaudo:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/11166
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11533"></a>


packages/babel-traverse/src/scope/index.js
```diff
@@ -447,13 +447,17 @@ export default class Scope {
    * Possibly generate a memoised identifier if it is not static and has consequences.
    */
 
-  maybeGenerateMemoised(node: Object, dontPush?: boolean): ?Object {
+  maybeGenerateMemoised(
+    node: Object,
+    dontPush?: boolean,
+    getInjectionPath?: NodePath => void,
+  ): ?Object {
```
nicolo-ribaudo, line 454: I don't like this API, but it's the best I could come up with in a non-breaking way :confused:

packages/babel-traverse/src/scope/index.js
```diff
@@ -864,6 +868,14 @@ export default class Scope {
   }) {
     let path = this.path;
 
+    if (path.isPattern()) {
+      // We are inside some function parameters
+      throw path.buildCodeFrameError(
+        "[Internal error] It's not possible to inject a variable in function parameters. " +
+          "Please check in the stack trace of this error which plugin is causing it.",
```
nicolo-ribaudo, line 875: This might cause new errors, but it's better to explicitly fail than to generate invalid code.

For example, it will throw when compiling `function f(a = x()?.b) {}` with a new version of `@babel/traverse` and an old version of the optional chaining plugin.

packages/babel-traverse/src/scope/index.js
```diff
@@ -864,6 +868,14 @@ export default class Scope {
   }) {
     let path = this.path;
 
+    if (path.isPattern()) {
+      // We are inside some function parameters
```
JLHwung, line 872: `path.isPattern` does not implies that we are inside function parameters. It can also be an object pattern in a block statement, etc.

nicolo-ribaudo: In that case, the `path` would be the block statement (here `path` is `scope.path`).

packages/babel-traverse/src/scope/index.js
existentialism, line 875:
```suggestion
          "Please check this error's stack track to see which plugin is causing it.",
```

Sidebar: it'd be much nicer to tell them which plugin it was

nicolo-ribaudo: Should we attempt parsing the error stack trace using a regular expression?

packages/babel-traverse/src/scope/index.js
```diff
@@ -872,7 +872,7 @@ export default class Scope {
       // We are inside some function parameters
       throw path.buildCodeFrameError(
         "[Internal error] It's not possible to inject a variable in function parameters. " +
-          "Please check in the stack trace of this error which plugin is causing it.",
+          "Please check this error's stack track to see which plugin is causing it.",
```
nicolo-ribaudo, line 875:
```suggestion
          "Please check this error's stack trace to see which plugin is causing it.",
```

existentialism: 🤦

packages/babel-traverse/src/scope/index.js
```diff
@@ -864,6 +868,16 @@ export default class Scope {
   }) {
     let path = this.path;
 
+    if (path.isPattern()) {
+      // We are inside some function parameters.
```
existentialism, line 872:
```suggestion
      // We are inside a function's parameters.
```

packages/babel-traverse/src/scope/index.js
```diff
@@ -864,6 +868,16 @@ export default class Scope {
...
+      // When using a destructuring outside of funtion parameters (e.g. in a variable
```
existentialism, line 873:
```suggestion
      // When using destructuring outside of function's parameters (e.g. in a variable
```

packages/babel-traverse/src/scope/index.js
JLHwung: Me too. Can we not merge this to v8? Since we can have a cleaner fix without maintaining a confusing API (`scope.maybeGenerateMemoised` is available in plugins).

nicolo-ribaudo: It's easier to revert this PR in the v8 branch, so that v8 can still be "master + some commits"

nicolo-ribaudo:
@JLHwung WDYT about adding a `NodePath#maybeMemoize(): { memoized: true, id: Node, path: NodePath } | { memoized: false }` method, where `path` in the return value is different from `this` when we have to inject a new scope?

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -92,7 +98,13 @@ export default declare((api, options) => {
               // Otherwise, we need to memoize the context object, and change the call into a Function#call.
               // `a.?b.?()` translates roughly to `(_b = _a.b) != null && _b.call(_a)`
               const { object } = chain;
-              let context = scope.maybeGenerateMemoised(object);
+              let context = scope.maybeGenerateMemoised(
```
JLHwung, line 101: Also need to sync with https://github.com/babel/babel/blob/32e7bb40279e3bc91d94cb640d24f11f632cd698/packages/babel-helper-member-expression-to-functions/src/index.js#L219

dsl101:
Sorry if it's poor etiquette to ask here, but, did this get stuck? I could really do with this fix, and one test appears to have been running for 26 days, which doesn't seem right...

PR #6440, opened 2017-10-07
Misc

hzoo:
Using yarn for the internal package bootstrap (not workspaces)

make bootstrap  211s
make bootstrap  100s

PR #11537, opened 2020-05-08
Update preset-env tests, and add yarn.lock deps

jridgewell:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    |
| Any Dependency Changes?  |

Hopefully, this'll stop the tests failing every time the browser compat tables are updated. We'll need to manually bump the deps used to generate our tables.


nicolo-ribaudo:
THANKS! 😁

Do we need to add the lockfile to .npmignore?

existentialism:
Yeah, let's add it

lerna.json
```diff
@@ -37,6 +37,6 @@
   ],
   "npmClient": "yarn",
   "npmClientArgs": [
-    "--no-lockfile"
+    "--pure-lockfile"
```
JLHwung, line 40: According to https://github.com/yarnpkg/yarn/issues/5847#issuecomment-537521943, `--pure-lockfile` will not modify `yarn.lock` even if we manually update `package.json`. I think can just remove `npmClientArgs` and resort to the default behaviour.

@hzoo BTW `--no-lockfile` was introduced at https://github.com/babel/babel/pull/6440, do you have any concerns about why `--no-lockfile` was applied?

hzoo: haha was so long ago, I assume it was because it was adding a lockfile to every package when you run `make bootstrap` instead of one at the top level?

jridgewell: I chose `--pure-lockfile` so that lerna wouldn't create a lock file in every package, but would still read from it. If we want to use lock files everywhere, I'd be fine with removing.

hzoo: Oh I meant that we didn't want lockfiles everywhere so I believe that's why `--no-lockfile` was originally added. yeah maybe pure is the way to go then but yeah need to update manually otherwise.

Issue #11534, opened 2020-05-07
@babel/standalone can't parse any quantifiers in unicode regexes


rhendric:
## Bug Report

- [ ] I would like to work on a fix!

**Current Behavior**

It's right there in the title. Something about the @babel/standalone compilation process causes it to omit critical parts of regjsparser, such that any regex literal with a `u` flag and a quantifier will cause Babel to incorrectly raise a syntax error.

Can be reproduced [on babeljs.io](https://babeljs.io/repl#?code_lz=PQQwVMCuQ&presets=es2015).

**Input Code**

```js
/a*/u
```

**Expected behavior/code**

It should parse valid regexes correctly.

**Babel Configuration (babel.config.js, .babelrc, package.json#babel, cli command, .eslintrc)**

```js
{ presets: ['es2015'] }
```

**Environment**

- Babel version(s): 7.9.6
- Node/npm version: N/A
- OS: N/A
- Monorepo: N/A
- How you are using Babel: standalone

nicolo-ribaudo:
Rollup is removing this code:
https://github.com/jviereck/regjsparser/blob/03d4486ec43184c1e6a18a871308d7ba1825c992/parser.js#L488-L495

PR #11538, opened 2020-05-08
Downgrade rollup to 1.27.9

nicolo-ribaudo:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/11534
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This is a partial revert of #11233. Due to a bug in rollup's tree-shaking logic, `@babel/standalone` didn't contain some necessary code. (https://github.com/babel/babel/issues/11534#issuecomment-625742700)

We cannot upgrade to Rollup 2 yet because it requires Node.js 10.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "prettier": "^1.19.1",
     "pump": "^3.0.0",
     "rimraf": "^2.6.3",
-    "rollup": "^1.32.1",
+    "rollup": "1.27.9",
     "rollup-plugin-babel": "^4.4.0",
     "rollup-plugin-commonjs": "^10.1.0",
     "rollup-plugin-node-builtins": "^2.1.2",
--- a/packages/babel-standalone/test/babel.js
+++ b/packages/babel-standalone/test/babel.js
@@ -196,5 +196,13 @@
 }`);
       });
     });
+
+    describe("regressions", () => {
+      it("#11534 - supports quantifiers in unicode regexps", () => {
+        expect(() =>
+          Babel.transform("/a*/u", { presets: ["es2015"] }),
+        ).not.toThrow();
+      });
+    });
   },
 );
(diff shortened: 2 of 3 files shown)


hzoo:
> We cannot upgrade to Rollup 2 yet because it requires Node.js 10.
Are we building standalone on every node version? Maybe we could just do it on latest?

nicolo-ribaudo:
We need to build standalone to test it :shrug:
We can try to update rollup in the `next-8-dev` branch.

JLHwung:
@nicolo-ribaudo Currently `@babel/standalone` is only tested on latest node. If we interpret `node >= 6` support as:

Babel can work on Node.js 6 and it is tested on Node.js 6, but Babel may not be built on `node >= 6`.

Then we can consider upgrading `rollup` and ESLint, etc. but leave `jest` stay on v24 to run tests only. I believe none of core team members are developing babel on node.js 6, and this change should ideally invisible to users.

nicolo-ribaudo:
@JLHwung It seems to be fixed when using https://rollupjs.org/repl/?version=2.8.2

Regarding the dependencies updates: I think we use rollup to build the parser even on Node.js 6.

3cp:
Suggest to have a look at https://github.com/zeit/ncc

I was troubled by various rollup issues (especially break nodejs v14) before, moved to ncc with zero config, never want to go back.

nicolo-ribaudo:
It lists "Only supports Node.js programs as input / output" as a design goal, and our output code must run on browsers 🤷

3cp:
It generates commonjs code. As long as you don't use nodejs api, there is no reason the code cannot run on browser. I got few projects using ncc to generate runnable code in both browser and nodejs.

3cp:
Example here
https://github.com/buttonwoodcx/bcx-validation/blob/55a17d6a10ab29c45fa0bf2dd37f1532f7159260/package.json#L31-L32

ncc doesn't support output to stdout, so there are two steps in my build.
1. ncc to pack all files into one js.
2. babel to transpile the bundle to es5 for IE11.

PR #11544, opened 2020-05-12
Refine babel core types

JLHwung:

- Generally improved flow typings in `@babel/core`.
- Enabled flow annotation on `packages/babel-core/src/config/validation/plugins.js`, which
catches a bug introduced by #10402 where `assertVisitorMap` is not working as intended.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11544"></a>

### DIFF
--- a/packages/babel-core/src/config/files/configuration.js
+++ b/packages/babel-core/src/config/files/configuration.js
@@ -151,7 +151,7 @@ export function* loadConfig(
  * Read the given config file, returning the result. Returns null if no config was found, but will
  * throw if there are parsing errors while loading a config.
  */
-function readConfig(filepath, envName, caller) {
+function readConfig(filepath, envName, caller): Handler<ConfigFile | null> {
   const ext = path.extname(filepath);
   return ext === ".js" || ext === ".cjs" || ext === ".mjs"
     ? readConfigJS(filepath, { envName, caller })
@@ -236,7 +236,7 @@ const readConfigJS = makeStrongCache(function* readConfigJS(
 
 const packageToBabelConfig = makeWeakCacheSync(
   (file: ConfigFile): ConfigFile | null => {
-    const babel = file.options[("babel": string)];
+    const babel: mixed = file.options[("babel": string)];
 
     if (typeof babel === "undefined") return null;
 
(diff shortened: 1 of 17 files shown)


packages/babel-core/test/__snapshots__/config-loading.js.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`@babel/core config loading programmatic plugins/presets should throw when plugin contains \`enter\` handler 1`] = `"[BABEL] /Users/jh/code/babel/packages/babel-core/test/fixtures/config-loading/folder/example.js: .visitor cannot contain catch-all \\"enter\\" or \\"exit\\" handlers. Please target individual nodes."`;
```
JLHwung, line 3: Just realize that since `.enter` and `.exit` are not allowed in `visitors` since v7, we could possibly remove https://github.com/babel/babel/blob/07347544fda9c88064379ec705a3c42224aa4c0a/packages/babel-traverse/src/context.js#L27

packages/babel-core/src/config/files/configuration.js
```diff
@@ -299,7 +299,7 @@ const readIgnoreConfig = makeStaticFileCache((filepath, content) => {
   };
 });
 
-function throwConfigError() {
+function throwConfigError(): void {
```
nicolo-ribaudo, line 302:
```suggestion
function throwConfigError(): empty {
```

packages/babel-core/test/__snapshots__/config-loading.js.snap
nicolo-ribaudo: Is it currently possible to use them if calling `traverse` directly?

hzoo: yeah I think so, it was added to core, although we could revisit that (change it back) too

JLHwung: Yeah it is possible to run `require("@babel/traverse").traverse({ enter() {} })`. Guess we should also remove that in Babel 8, to be aligned with what `@babel/core` limits on plugins.

nicolo-ribaudo: I think that `enter` for advanced users (= people using `traverse` directly) could be useful, and also it's the default behavior of all the other traversal methods.

nicolo-ribaudo:
I'm merging this with a single review, since it's mostly about internal typings.

PR #10955, opened 2020-01-03
LiteralPropertyName should allow BigIntLiteral

JLHwung:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | LiteralPropertyName does not allow BigIntLiteral
| Patch: Bug Fix?          | Yes
| Tests Added + Pass?      | Yes
| License                  | MIT

Tracing through the spec:

```
LiteralPropertyName => NumericLiteral => DecimalBigIntegerLiteral
LiteralPropertyName => NumericLiteral => NonDecimalIntegerLiteral BigIntLiteralSuffix
```
Related: 
https://github.com/tc39/test262/pull/2457
https://bugs.chromium.org/p/v8/issues/detail?id=10083

CI is red due to
- [ ] https://github.com/microsoft/TypeScript/issues/35982

PR #11547, opened 2020-05-12
refactor: add isLiteralPropertyName to parser utils

### LINKED ISSUE #11555: BigInt Literal not parsed in Typescript interface declaration
## Bug Report

- [] I would like to work on a fix!
- Not right now, but in a week or two if that's ok!

**Current Behavior**
BigInt literal is not being parsed in Typescript interface declaration:
https://babeljs.io/en/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=AQSwdgLgpgTgZgQwMZWAcSmWInAN4BQwxUAXMAIwAMYA3EQL4EED0LANlBMNAM4TkMWGDmABefGUo0mnbqgnUwQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=typescript&prettier=false&targets=&version=7.9.6&externalPlugins=%40babel%2Fplugin-syntax-bigint%407.8.3

**Input Code**
```typescript
 interface Generic {
   e: 10n;
 }

let test: Generic = {e: 10n}
```

**Expected behavior/code**
Should be parsed.

Whatever config the repl I linked is using. It is using the typescript preset and @babel/plugin-syntax-bigint

**Environment**
Whatever environment the repl is using.

Working in Typescript playground: https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=7&pc=3#code/AQSwdgLgpgTgZgQwMZWAcSmWInAN4BQwxUAXMAIwAMYA3EQL4EEA2UEw0AzhORljBzAAvPjKUaDIA

This is true even if the BigInt syntax plugin is installed. I suspect the Typescript plugin needs to be made aware of the BigInt plugin or something.

I also noticed that in the Babel typescript definitions (the official ones, not the ones in DefinitelyTyped), the following definition exists:
```typescript
export interface TSLiteralType extends BaseNode {
  type: "TSLiteralType";
  literal: NumericLiteral | StringLiteral | BooleanLiteral;
} 
```

This definition seems to be missing "BigIntLiteral" as a possible type for the literal.

JLHwung:
| Fixed Issues?            | BigIntLiteral should be allowed in name of async method and accessors, also fixes #11555 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR is a follow-up to #10955, found this issue when reading sources.


packages/babel-parser/src/parser/util.js
```diff
@@ -259,6 +259,22 @@ export default class UtilParser extends Tokenizer {
       this.raise(doubleProto, Errors.DuplicateProto);
     }
   }
+
+  /**
+   * Test if current token is a literal property name
+   * LiteralPropertyName:
```
existentialism, line 265: Should this also include `BigIntLiteral`?

JLHwung: The production is extracted from the [spec](https://tc39.es/ecma262/#prod-LiteralPropertyName) where `BigIntLiteral` is covered by `NumericLiteral`. I am happy to add that if it helps.

existentialism: Nah the link is fine

packages/babel-parser/src/parser/util.js
```diff
@@ -259,6 +259,24 @@ export default class UtilParser extends Tokenizer {
...
+   *   StringLiteral
+   *   NumericLiteral
+   *   BigIntLiteral
+   */
+  isLiteralPropertyName(): boolean {
+    return (
+      this.match(tt.name) ||
+      this.match(tt.string) ||
+      this.match(tt.num) ||
+      this.match(tt.bigint)
```
nicolo-ribaudo, line 277: Q: things like `true` or `null` fall in the `tt.name` case, right?

JLHwung: The fall in the special `tt._true` and `tt._false`. I move all keywords to `isLiteralPropertyName` since they are valid identifier name.

existentialism: Nice catch!

JLHwung:
Note for myself: work on a fix on #11555 based on this PR.

PR #11552, opened 2020-05-12
fix: ensure (a?.b)() has proper this

### LINKED ISSUE #10802: Optional Chaining does not propagate receiver when parenthesized

JLHwung:
| Fixed Issues?            | Fixes #10802, closes #11327 

It has been out there for quite a while and the fix seems more tricky than previously proposed in #11327 and #10802.

Unrelated: Can we introduce a granular `loose`-like `assumeNoAccessor` options in all the transformer plugins? Because I feel that the accessor-related memoizing code could bloat the code size. This option should also help `rest-spread` and `destructuring` transforms.

### DIFF
--- a/packages/babel-plugin-proposal-optional-chaining/src/index.js
+++ b/packages/babel-plugin-proposal-optional-chaining/src/index.js
@@ -108,7 +108,32 @@ export default declare((api, options) => {
               );
             }
           }
-
+          let replacement = replacementPath.node;
+          // Ensure (a?.b)() has proper `this`
+          if (
+            t.isMemberExpression(replacement) &&
+            replacement.extra?.parenthesized &&
+            replacementPath.parentPath.isCallExpression()
+          ) {
+            // `(a?.b)()` to `(a == null ? undefined : a.b.bind(a))()`
+            const { object, property } = replacement;
+            let memoisedObject;
+            if (!loose) {
+              // memoize the context object in non-loose mode
+              // `(a?.b.c)()` to `(a == null ? undefined : (_a$b = a.b).c.bind(_a$b))()`
+              const context = scope.maybeGenerateMemoised(object);
+              if (context) {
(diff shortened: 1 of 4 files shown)


JLHwung:
@nicolo-ribaudo Do you think I'd better rebase this PR onto `feat-7.10/private-stuff`? I just realize that this PR probably does not solve the following cases
```js
class C {
  static #m() { return this }
  test(C) {
    expect((C?.#m)()).toBe(C)
  }
}

C.test(C)
```

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -108,7 +108,32 @@ export default declare((api, options) => {
               );
             }
           }
-
+          let replacement = replacementPath.node;
+          // Ensure (a?.b)() has proper `this`
+          if (
+            t.isMemberExpression(replacement) &&
+            replacement.extra?.parenthesized &&
```
nicolo-ribaudo, line 115: We should also handle the case where we have `ParenthesizedExpression` nodes.

packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/parenthesized-member-call/exec.js
```diff
@@ -0,0 +1,25 @@
+let count = 0;
```
JLHwung, line 1: Consider rewrite the test into something similar with https://github.com/babel/babel/blob/785b6e8bde1419b3ea7d219cd8caf9dab043edfb/packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/optional-chain-private-call-with-transform/exec.js

for better coverage.

PR #11553, opened 2020-05-12
Fix _removeFromScope bug

### LINKED ISSUE #11551: Removing a VariableDeclarator causes same-named variables in outer scope's bindings to be removed
## Bug Report

**Current Behavior/Expected Behavior/Potential Solution**
I am developing a custom plugin. The details of the plugin are irrelevant to this bug.

Given this code:

```js
const foo = '1'
const main = () => {
	const foo = '2'
}
```

If the inner `foo` declaration is removed via `path.remove()`, then the binding for `foo` in the inner scope should be removed, and the binding for `foo` in the outer scope should _not_ be removed.

Instead, both the inner and outer bindings are removed.

Here is a codesandbox that does not demonstrate the issue unfortunately. For some reason all of the `scope`s in this demo are empty/have no bindings. I assume it is something related to codesandbox. Maybe someone knows how to get it working?

[Codesandbox](https://codesandbox.io/s/black-waterfall-bhcv8?file=/src/index.js)

I believe that here is what happens when the inner `foo` is removed via `path.remove()`:

- Binding for inner `foo` is removed from inner scope: https://github.com/babel/babel/blob/master/packages/babel-traverse/src/path/removal.js#L11
- Removal hooks get called: https://github.com/babel/babel/blob/master/packages/babel-traverse/src/path/removal.js#L14
- Since it is the last `VariableDeclarator` in the `VariableDeclaration`, `remove()` is called on the `VariableDeclaration`: https://github.com/babel/babel/blob/master/packages/babel-traverse/src/path/lib/removal-hooks.js#L25
- In `_removeFromScope` in the `VariableDeclaration`:
  - `this.getBindingIdentifiers()` includes the `foo` binding because the inner `VariableDeclaratior` hasn't actually been removed yet: https://github.com/babel/babel/blob/master/packages/babel-traverse/src/path/removal.js#L25
  - `this.scope.removeBinding('foo')` is called, but this time it gets removed from the outer scope because `this.scope.removeBinding` scans upwards for the binding, and the `foo` binding has already been removed from the inner scope: https://github.com/babel/babel/blob/master/packages/babel-traverse/src/path/removal.js#L26

Changing `removeBinding` to `removeOwnBinding` seems to fix this use case, but I'm unsure if that would cause problems in other use cases. Any other ideas? Is there good test coverage on scope tracking?

calebeby:
| Fixed Issues?            | Fixes #11551 
| Tests Added + Pass?      | y
| Documentation PR Link    | n

Context in #11551

### DIFF
--- a/packages/babel-traverse/src/path/removal.js
+++ b/packages/babel-traverse/src/path/removal.js
@@ -23,7 +23,7 @@ export function remove() {
 
 export function _removeFromScope() {
   const bindings = this.getBindingIdentifiers();
-  Object.keys(bindings).forEach(name => this.scope.removeBinding(name));
+  Object.keys(bindings).forEach(name => this.scope.removeOwnBinding(name));
 }
 
 export function _callRemovalHooks() {


calebeby:
I could use help fixing the failing test. I know that babel's TS implementation removes things that it thinks are only used as types, but I don't know why this change is causing that to fail :thinking:

nicolo-ribaudo:
I think that the problem is that for classes and functions, `path.scope` is the inner scope, while we should also remove the binding from the outer one.

PR #11554, opened 2020-05-12
Hide internal `@babel/core` functions in config errors

### LINKED ISSUE #11540: `@babel/core` `parseSync` does not handle `exclude` in config
## Bug Report

**Current Behavior**

- [REPL](babeljs.io/repl), [Codesandbox](https://codesandbox.io/s/babel-repl-custom-plugin-7s08o?file=/src/index.js), or GitHub Repo link if applicable: https://github.com/SimenB/babel-parse-exclude

**Input Code**

```js
```

**Expected behavior/code**
Passing an empty string to `parseSync` will throw if config file contains `exclude`.

`Error: Configuration contains string/RegExp pattern, but no filename was passed to Babel`

This is sorta documented (https://babeljs.io/docs/en/options#filename) but I find it very confusing 😀 There aren't really any parse info inferred from filename (I think?), at least not when we pass `parserOpts`.

- Filename: `babel.config.js`

```js
module.exports = {exclude: /react-native/};
```

**Environment**

```
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 12.16.3 - ~/.nvm/versions/node/v12.16.3/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v12.16.3/bin/npm
  npmPackages:
    @babel/core: ^7.9.6 => 7.9.6

```
- Babel version(s):  7.9.6
- Node/npm version: v12
- OS:  osx
- Monorepo:  no
- How you are using Babel: `@babel/core`'s `parseSync`

**Possible Solution**

I would expect this error to be suppressed when we use `parse` API, same as when we use `@babel/parser` directly.

nicolo-ribaudo:
| Fixed Issues?            | Closes #11540
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR introduces some logic to hide `@babel/core` internals in error stack traces for errors coming from outside `@babel/core` itself (for example, coming from config files or plugins/presets): `@babel/core` introduces a lot of frames (and all duplicated, because of how we use gensync), which are not really useful.

Unexpected errors thrown inside `@babel/core` itself still have the complete stack trace, because they are bugs and we need debug info.

If you want to see the old stack traces, you can replace `const SUPPORTED = !!Error.captureStackTrace;` with `const SUPPORTED = false;` and run the tests.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11554"></a>


liuxingbaoyu:
This PR looks great! The current stack provides little useful information.

nicolo-ribaudo:
@liuxingbaoyu Since you use windows, when you have time could you clone this PR, run `yarn jest babel-core/test/errors-stacks` and see why the tests are not throwing when they should? :pray:

liuxingbaoyu:
This is the same as https://github.com/babel/babel/pull/14671#discussion_r920480036, by the way I also found a typo there although it works fine.😂

packages/babel-core/test/errors-stacks.js
```diff
@@ -0,0 +1,292 @@
+import * as babel from "../lib/index.js";
+
+// TODO: Remove this in Babel 8, once we drop Node.js 8
```
liuxingbaoyu, line 3:
```suggestion
```
Seems like this should be removed at the same time, I didn't understand it when I submitted it just now.

liuxingbaoyu:
To be honest, the CI exception from the fork repo surprised me.😅
This is extremely strange, I often have similar problems too in my personal repo, but almost none in the main repo.😕

nicolo-ribaudo:
Thank you! :smile:

packages/babel-core/src/config/files/configuration.ts
```diff
@@ -263,17 +269,19 @@ const readConfigJSON5 = makeStaticFileCache((filepath, content): ConfigFile => {
...
-    err.message = `${filepath}: Error while parsing config - ${err.message}`;
-    throw err;
+    throw new ConfigError(
+      `Error while parsing config - ${err.message}`,
+      filepath,
+    );
   }
 
-  if (!options) throw new Error(`${filepath}: No config detected`);
+  if (!options) throw new ConfigError(`No config detected`, filepath);
```
liuxingbaoyu, line 278: Maybe there is some lint rule to make it a normal string, but in this PR I think it's ok.🤫

liuxingbaoyu (approved):

packages/babel-core/src/config/files/configuration.ts
ehoogeveen-medweb: There's [@typescript-eslint/quotes](https://typescript-eslint.io/rules/quotes), which can be configured with `allowTemplateLiterals: false` to avoid unnecessary template literals (despite the name, it does not forbid all usage of template literals). To be [compatible with Prettier](https://github.com/prettier/eslint-config-prettier#quotes), `avoidEscape: true` must also be set (even if you extend `eslint-config-prettier`, individual rule configurations take precedence).

packages/babel-core/src/errors/rewrite-stack-trace.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+export function beginHiddenCallStack<Fn extends Function>(fn: Fn): Fn {
+  if (!SUPPORTED) return fn;
+
+  return Object.defineProperty(
+    function (this: any) {
+      setupPrepareStackTrace();
+      return fn.apply(this, arguments);
+    } as any as Fn,
```
JLHwung, line 104: The type casting can be avoided if we parameterize the function's input and outputs.

```ts
export function beginHiddenCallStack<T, A extends any[], R>(
  fn: (this: T, ...args: A) => R,
) {
  if (!SUPPORTED) return fn;

  return Object.defineProperty(
    function (this: T, ...args: A) {
      setupPrepareStackTrace();
      return fn.apply(this, args);
    },
    "name",
    { value: STOP_HIDNG },
  );
}
```

nicolo-ribaudo: We still need to type cast somewhere, because the functions that we receive in the config have type `Function` which is not compatible with the "explicit" function types. However, I moved them to the config validation logic so that the "internals" are properly typed.

packages/babel-core/src/errors/rewrite-stack-trace.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  const { prepareStackTrace = defaultPrepareStackTrace } = Error;
+
+  Error.stackTraceLimit += STACK_TRACE_LIMIT_DELTA;
+
+  Error.prepareStackTrace = function stackTraceRewriter(err, trace) {
+    let newTrace = [];
+
+    const isExpected = expectedErrors.has(err);
+    let status = isExpected ? "hiding" : "unknown";
```
JLHwung, line 140: Nit: Can you restrict the typing here? It seems to me `status` can only be `"hiding" | "showing" | "unknown"`.

packages/babel-core/src/errors/rewrite-stack-trace.ts
```diff
@@ -0,0 +1,171 @@
...
+ * - If c() throws an unexpected error, then its shown call stack will be "c, d, e, f"
+ * - If d() throws an expected error, then its shown call stack will be "e, f"
+ * - If d() throws an unexpected error, then its shown call stack will be "d, e, f"
+ * - If e() throws an error, then its shown call stack will be "e, f"
+ *
+ * Additionally, an error can inject additional "virtual" stack frames using the
+ * injcectVirtualStackFrame(error, filename) function: those are added on the top
+ * of the existig stack, after hiding the possibly hidden frames.
+ * In the example above, if we called injcectVirtualStackFrame(error, "h") on the
+ * expected error thrown by c(), it's shown call stack would have been "h, e, f".
```
JLHwung, line 39:
```suggestion
 * expected error thrown by c(), its shown call stack would have been "h, e, f".
```
May extend the example a bit, for example, what will be the shown call stack if we called `injcectVirtualStackFrame(error, "h")` and `injcectVirtualStackFrame(error, "i")`?

JLHwung (approved):

PR #11571, opened 2020-05-16
Add a check for privateMap's existence

### LINKED ISSUE #11414: Babel throws confusing error when private property is used as computed key
## Bug Report

**Current Behavior**
The following snippet, after transformed, throws
```
Cannot read property 'get' of undefined
```

**Input Code**
- [REPL](https://babeljs.io/repl#?browsers=chrome%2080&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEAqD2BzRICm0DeAoa0DEADgE7wHQC80ARACYCWEYARmjVQNw7QDaALgBYMAdIRIEAuhWgAzMCAiosAXyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=true&sourceType=script&lineWrap=false&presets=env%2Cstage-3%2Cenv&prettier=false&targets=Node-13&version=7.9.0&externalPlugins=%40babel%2Fplugin-proposal-object-rest-spread%407.9.5)

```js
class Toggle {
  #prop = "disabled";
  [this.#prop] = false
}
```

**Expected behavior/code**
It should throw
```
attempted to get private field on non-instance
```
**Environment**
REPL

**Possible Solution**

We should also check if `privateMap` is `undefined` in https://github.com/babel/babel/blob/a6e8b3843b53c6e106d9fb2e37fea60c306dd97c/packages/babel-helpers/src/helpers.js#L1121

Found this issue when reviewing PR 11405.

AjayPoshak:
| Fixed Issues?            | Fixes #11414  
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-core/src/config/files/configuration.js
+++ b/packages/babel-core/src/config/files/configuration.js
@@ -71,8 +71,9 @@ export function* findRelativeConfig(
         loc,
         envName,
         caller,
-        packageData.pkg && packageData.pkg.dirname === loc
-          ? packageToBabelConfig(packageData.pkg)
+        packageData.pkg?.dirname === loc
+          ? // $FlowIgnore - packageData.pkg is not null
+            packageToBabelConfig((packageData.pkg: ConfigFile))
           : null,
       );
     }
--- a/packages/babel-core/src/config/index.js
+++ b/packages/babel-core/src/config/index.js
@@ -17,7 +17,8 @@ export type { PartialConfig } from "./partial";
 
 const loadOptionsRunner = gensync<[mixed], Object | null>(function*(opts) {
   const config = yield* loadFullConfig(opts);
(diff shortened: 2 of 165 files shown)


nicolo-ribaudo:
Instead of explicitly handling this case, would it be possible to always ensure that the `WeakMap` is always defined, as proposed at https://github.com/babel/babel/issues/11414#issuecomment-613272365?

nicolo-ribaudo:
Could you add a new test with the code mentioned in https://github.com/babel/babel/issues/11414? :pray:

Also, CI failure should is fixed on master.

PR #11572, opened 2020-05-17
Add targets option to @babel/plugin-transform-runtime

### LINKED ISSUE #8675: Unnecessary  require with transform runtime?
## Problem

My target environment is chrome 67, but still `corejs/promise` is required.

## Config

**package.json**

```json

{
  "name": "play-with-babel",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "devDependencies": {
    "@babel/cli": "^7.0.0",
    "@babel/core": "^7.0.0",
    "@babel/plugin-transform-runtime": "^7.0.0",
    "@babel/preset-env": "^7.0.0"
  },
  "dependencies": {
    "@babel/runtime": "^7.0.0",
    "@babel/runtime-corejs2": "^7.0.0"
  }
}

```
**babel.config.js**

```js
"use strict";

module.exports = function (api) {
  const env = api.env()
  const envOpts = {
    loose: true,
    modules: false,
    useBuiltIns: 'usage',
    targets: {
      chrome: "67"
    },
    debug: true
  }
  const config = {
    sourceType: "script",
    comments: false,
    presets: [
      ["@babel/preset-env", envOpts]
    ],
    plugins: [
      ["@babel/plugin-transform-runtime", {
        corejs: 2 //  <= avoid polluting global environment
      }]
    ]
  };
  return config;
};

```
## Input 

```js
const p = Promise.resolve(1)
```

## Output

```js
// as you can see,  **const** is not transformed which is correct

// but promise should not be imported right?

var _Promise = require("@babel/runtime-corejs2/core-js/promise"); 

const p = _Promise.resolve(1);
```

### LINKED ISSUE #9363: plugin-transform-runtime doesn't honor preset-env targets when corejs option is set
## Bug Report

I've observed that plugin-transform-runtime does not conditionally transform a feature based on preset-env's configured targets when the corejs option is set. The specific example I ran into was when using `Object.entries`, which is supported on Node 8, but not supported on Node 6.

**Current Behavior**
When preset-env is configured with `{ targets: { node: 6 }, useBuiltIns: 'usage' }`, then an `Object.entires` polyfill is included in the transpiled output. If the node version is changed to 8, then this polyfill is no longer included. If plugin-transform-runtime is configured with `{ corejs: 2 }`, then corejs's `Object.entries` implementation is included in the transpiled output when the node version is set to 6 or 8.

**Input Code**
```javascript
Object.entries({foo: 'bar'});
```

See https://repl.it/@blicksky/babel-transform-runtime-tests for a set of test cases that reproduces this scenario.

**Expected behavior/code**
When preset-env is configured with `{ targets: { node: 8 }, useBuiltIns: 'usage' }` and plugin-transform-runtime is configured with `{ corejs: 2 }`, then usages of `Object.entries` in the input code should not be replaced with corejs's implementation in the transpiled output.

```js
{
    babelrc: false,
    presets: [
        ['@babel/preset-env', {
            targets: { node: 8 },
            useBuiltIns: 'usage'
        }]
    ],
    plugins: [
        ['@babel/plugin-transform-runtime', {
            corejs: 2
        }]
    ]
}
```

**Environment**
- Babel version(s):
    - @babel/core: 7.2.2,
    - @babel/preset-env: 7.2.3,
    - @babel/plugin-transform-runtime: 7.2.0
    - @babel/runtime-corejs2: 7.2.0
- Node/npm version: Node 9.7.1
- OS: observed behavior on both Linux and MacOS
- Monorepo: N/A
- How you are using Babel: observed behavior via both the CLI and API

### LINKED ISSUE #9853: Using @babel/runtime-corejs2 and @babel/runtime-corejs3 leads to larger bundle sizes
## Bug Report

**Current Behavior**
When configuring `@babel/plugin-transform-runtime` to use `corejs2` or `corejs3`, bundle sizes increase x7 (corejs2) and x11 (corejs3)

**Input Code**

```js
const obj = {
  a: 1,
  b: 2,
};

const { b, ...rest } = obj;
const newObj = {...rest, a: b };
```

**Expected behavior/code**
I expect that the bundle sizes stay relatively close, when switching from `@babel/runtime` to `@babel/runtime-corejs2` or `@babel/runtime-corejs3`

***Configs to use `@babel/runtime`***
```js
module.exports = {
  presets: [
    ["@babel/preset-env", {
      modules: false,
      useBuiltIns: 'usage',
      corejs: 3
    }]
  ],
  plugins: [
  ]
}
```

***Configs to use `@babel/runtime-corejs2`***
```js
module.exports = {
  presets: [
    ["@babel/preset-env", {
      modules: false,
      useBuiltIns: 'usage',
      corejs: 2
    }]
  ],
  plugins: [
    ["@babel/plugin-transform-runtime", {
      corejs: 2
    }]
  ]
}
```

***Configs to use `@babel/runtime-corejs3`***
```js
module.exports = {
  presets: [
    ["@babel/preset-env", {
      modules: false,
      useBuiltIns: 'usage',
      corejs: 3
    }]
  ],
  plugins: [
    ["@babel/plugin-transform-runtime", {
      corejs: 3
    }]
  ]
}
```

**Environment**
- Babel version(s): 7.4.3 (all babel packages)
- Node/npm version: Node v10
- OS: OSX 10.13.6
- Monorepo: no
- How you are using Babel: CLI, Webpack loader

**Possible Solution**

N/A

Repro repo: https://github.com/GeorgeTaveras1231/corejs-size-regression-repro

Snapshot:

Screen Shot 2019-04-12 at 11 41 52 AM

TomerAberbach:
| Fixed Issues?            | Fixes #9363, Fixes #8675, Fixes #9853
| Documentation PR Link    | [Link](https://github.com/babel/website/pull/2253)

`@babel/plugin-transform-runtime` is great because it can be used to polyfill features without polluting the global namespace (i.e. ponyfills). This is the ideal choice for libraries because they shouldn't pollute the global namespace. However, `@babel/plugin-transform-runtime` does not support only polyfilling the features required by the target browser/node versions like `@babel/preset-env` does. This results in unnecessarily large bundles containing unused polyfills.

This pull request adds a `targets` option to `@babel/plugin-transform-runtime` that behaves similarly to `@babel/preset-env`'s `targets` field. It also adds `configPath` and `ignoreBrowserslist` options like `@babel/preset-env` has.

With regards to dependencies, `@babel/compat-data`, `@babel/helper-compilation-targets`, and `core-js-compat` dependencies were added to `@babel/plugin-transform-runtime`.

Lastly, I'm aware that [Babel is rethinking its polyfilling story](https://github.com/babel/babel-polyfills) so this change may become obsolete in the future. However, I think people have been waiting a long time for this feature and it could be a while before Babel's new polyfilling approach is ready; so I think this feature is useful as a short-term solution.

Looking forward to your feedback!


packages/babel-plugin-transform-runtime/test/fixtures/runtime-corejs3/built-in-static-methods-esmodules-target/output.js
```diff
@@ -0,0 +1,148 @@
...
+
+var _Math$hypot = require("@babel/runtime-corejs3/core-js-stable/math/hypot");
+
+var _JSON$stringify = require("@babel/runtime-corejs3/core-js-stable/json/stringify");
+
+Array.from;
+Array.isArray;
+Array.of;
+Date.now;
+_JSON$stringify;
```
developit, line 33: I was originally thrown off by this, but it's because not all ESM browsers follow the well-formed JSON.stringify spec revision. Overhead here is 2kb.

TomerAberbach:
> One additional useful thing would be the ability to prevent the inclusion of a given polyfill (like `JSON.stringify`/`json/stringify`).

I think that's a good idea. We could mirror `@babel/preset-env`'s [`include`](https://babeljs.io/docs/en/babel-preset-env#include) and [`exclude`](https://babeljs.io/docs/en/babel-preset-env#exclude) options (except they'd be arrays of methods and identifiers rather than plugin names). Although I think I'd prefer doing that in a separate PR (assuming this one gets merged) to avoid this PR getting too big.

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -1,12 +1,38 @@
...
+corejs3Polyfills["es.array.values"] = corejs3Polyfills["es.array.iterator"];
+corejs3Polyfills["es.string.trim-left"] =
+  corejs3Polyfills["es.string.trim-start"];
+corejs3Polyfills["es.string.trim-right"] =
+  corejs3Polyfills["es.string.trim-end"];
+corejs3Polyfills["es.symbol.for"] = corejs3Polyfills["es.symbol"];
+corejs3Polyfills["es.symbol.key-for"] = corejs3Polyfills["es.symbol"];
+corejs3Polyfills["es.object.get-own-property-symbols"] =
+  corejs3Polyfills["es.symbol"];
+corejs2Polyfills["es6.symbol.iterator"] = corejs2Polyfills["es6.symbol"];
```
JLHwung, line 34: We can add a new `es6.symbol.iterator` item in

https://github.com/babel/babel/blob/2f31ecf85d85cb100fa08d4d9a09de0fe4a117e4/packages/babel-compat-data/scripts/data/corejs2-built-in-features.js#L294

IMO in old browsers `es6.symbol.iterator` should have more support than the general `es6.symbol` feature as it does not depend on the other well known symbols.

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -83,7 +135,14 @@ export default declare((api, options, dirname) => {
   }
 
   function hasMapping(methods, name) {
-    return has(methods, name) && (proposals || methods[name].stable);
+    return (
+      has(methods, name) &&
+      (proposals || methods[name].stable) &&
+      (polyfillPaths.has(methods[name].path) ||
+        methods[name]?.types?.some(type =>
```
JLHwung, line 142:
```suggestion
        methods[name].types?.some(type =>
```
nit: `methods.name` should not be nullish if `has(methods, name)` is `true`.

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -1,12 +1,38 @@
...
+} from "@babel/helper-compilation-targets";
+import corejs2Polyfills from "@babel/compat-data/corejs2-built-ins";
+import corejs3Polyfills from "core-js-compat/data";
 
 import getCoreJS2Definitions from "./runtime-corejs2-definitions";
 import getCoreJS3Definitions from "./runtime-corejs3-definitions";
 import { typeAnnotationToString } from "./helpers";
 import getRuntimePath from "./get-runtime-path";
 
+// corejs sometimes lumps multiple functions/identifiers under one name
```
JLHwung, line 16: Can you add a comment that this part should be align to [packages/babel-preset-env/src/polyfills/corejs3/built-in-definitions.js](https://github.com/babel/babel/blob/master/packages/babel-preset-env/src/polyfills/corejs3/built-in-definitions.js)?

We can move these definitions to `@babel/compat-data` as, in this case, it is reusabled by both `preset-env` and `transform-runtime`.

packages/babel-plugin-transform-runtime/src/index.js
TomerAberbach: True. Thanks!

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -17,12 +43,38 @@ export default declare((api, options, dirname) => {
   const {
     corejs,
     helpers: useRuntimeHelpers = true,
-    regenerator: useRuntimeRegenerator = true,
+    regenerator = true,
     useESModules = false,
     version: runtimeVersion = "7.0.0-beta.0",
     absoluteRuntime = false,
+    targets: optionsTargets = {},
+    configPath = process.cwd(),
```
JLHwung, line 51: While I am aware that introducing `configPath` and `ignoreBrowserslistConfig` is to align the interface of `preset-env`, I prefer to hold off a bit as we are considering changing the browserslists related options in Babel 8. https://github.com/nicolo-ribaudo/babel-rfcs/blob/top-level-targets/rfcs/0000-top-level-browserslist.md

Instead of introducing multiple `browserslists` options in 7.10 and then deprecate them in Babel 8, we can add `targets` only and iterate later.

packages/babel-plugin-transform-runtime/src/index.js
TomerAberbach: Just to make sure I fully understand that file's format, would I add something like the following?
```json
"es6.symbol.iterator": "well-known symbols / Symbol.iterator"
```

packages/babel-plugin-transform-runtime/src/index.js
TomerAberbach: Done.

packages/babel-plugin-transform-runtime/src/index.js
TomerAberbach: Done.

TomerAberbach: Unresolving because I realized I assumed you'd want `ignoreBrowserslistConfig` set to `true` in `getTargets`. Do you think that's the right move?

My logic is that if we set `ignoreBrowserslistConfig` to `false`, then users have no way to have their browserslist config ignored other than removing it, which could mess with any other tooling they use. If we set `ignoreBrowserslistConfig` to `true`, then it's a little inconvenient for users because they may have to specify their targets in two configs, but at least we haven't put them in an impossible situation.

JLHwung: 👍That looks reasonable to me.

packages/babel-plugin-transform-runtime/src/index.js
TomerAberbach: Done.

TomerAberbach: Based on the failed tests, was I supposed to take `"well-known symbols / Symbol.iterator"` out of `"es6.symbol"`? And then regenerate the snapshots for the tests?

TomerAberbach: I actually don't think I can make this change. Making this change causes `@babel/preset-env` to emit `require("core-js/modules/es6.symbol.iterator");` statements, but [`core-js@2` doesn't have a module with that filename](https://github.com/zloirock/core-js/tree/v2/modules).

We might be stuck doing `corejs2Polyfills["es6.symbol.iterator"] = corejs2Polyfills["es6.symbol"];`.

TomerAberbach: @JLHwung just tagging in case you didn't get a notification about this. Let me know what your thoughts are!

JLHwung: Good catch. Why we need `corejs2Polyfills["es6.symbol.iterator"] = ...` if it turns out core-js@2 does not have one?

TomerAberbach: It's used to decide whether to replace `something[Symbol.iterator]()` calls with the `getIterator(something)` helper, `Symbol.iterator in something` usages with the `isIterable(something)` helper, and `something[Symbol.iterator]` usages with the `getIteratorMethod(something)` helper. Look for `polyfillPaths.has("symbol/iterator")` in the file to see what I mean.

These helper methods are available in the [`fn` directory of `core-js@2`](https://github.com/zloirock/core-js/tree/v2/fn).

TomerAberbach: @JLHwung went ahead and reverted the changes.

TomerAberbach:
@JLHwung is there any interest in this PR anymore? If not, then I can close it

nicolo-ribaudo:
@TomerAberbach As a side effect of https://github.com/babel/babel/pull/12845, now this plugin supports `targets` (even if it was not originally intended, and I'd still recommend using `babel-plugin-polyfill-corejs3` directly):

```js
const input = String.raw`
Array.from(3);
foo.flatMap(fn);
`;

const out = babel.transform(input, {
  configFile: false,
  sourceType: "module",
  targets: "chrome 60",
  plugins: [[transformRuntime, { corejs: 3 }]],
});

console.log(out.code);
```
logs
```js
import _flatMapInstanceProperty from "@babel/runtime-corejs3/core-js-stable/instance/flat-map";
Array.from(3);

_flatMapInstanceProperty(foo).call(foo, fn);
```

while

```js
const input = String.raw`
Array.from(3);
foo.flatMap(fn);
`;

const out = babel.transform(input, {
  configFile: false,
  sourceType: "module",
  targets: "chrome 40",
  plugins: [[transformRuntime, { corejs: 3 }]],
});

console.log(out.code);
```
logs
```js
import _Array$from from "@babel/runtime-corejs3/core-js-stable/array/from";
import _flatMapInstanceProperty from "@babel/runtime-corejs3/core-js-stable/instance/flat-map";

_Array$from(3);

_flatMapInstanceProperty(foo).call(foo, fn);
```

Thanks for the PR anyway!

PR #11115, opened 2020-02-09
Bump Jest to 25.1.0

nicolo-ribaudo:
| Q                        | A 
| ------------------------ | ---
| Tests Added + Pass?      | Yes
| Any Dependency Changes?  | Yes
| License                  | MIT

[Changelog](https://jestjs.io/blog/2020/01/21/jest-25)

This is targeting the `next-8-dev` branch because it doesn't support Node.js 6.

I'd like to experiment using V8's native coverage tool in the future.

PR #11579, opened 2020-05-18
Update prettier to v2

zxbodya:
This is needed in context of migration to typescript #11578 to allow using `import type` typescript syntax.

| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  | Yes, updating prettier

### DIFF
--- a/scripts/utils/formatCode.js
+++ b/scripts/utils/formatCode.js
@@ -1,11 +1,22 @@
 "use strict";
-const prettier = require("prettier");
 
-module.exports = function formatCode(code, filename) {
-  filename = filename || __filename;
-  const prettierConfig = prettier.resolveConfig.sync(filename);
-  prettierConfig.filepath = filename;
-  prettierConfig.parser = "babel";
+// TODO: Remove this `if` in Babel 8
+// Prettier only supports Node.js 10+, so we can fallback to not formatting
+// o CI on older Node.js versions
 
-  return prettier.format(code, prettierConfig);
-};
+if (process.env.CI && parseInt(process.version.node, 10) < 10) {
+  module.exports = function formatCode(code) {
+    return code;
+  };
+} else {
+  const prettier = require("prettier");
+
+  module.exports = function formatCode(code, filename) {
+    filename = filename || __filename;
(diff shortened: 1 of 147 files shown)


existentialism:
FWIW, we've held off on this for the same reasons as jest@25 (https://github.com/babel/babel/pull/11115).

zxbodya:
> FWIW, we've held off on this for the same reasons as jest@25 (#11115).

should be a bit better, as I understand it is used only during lint and when generating some code during build…

so, at least at the first sight, it should not be needed when running the test (as an option might be good enough to add polyfill during the build, or skip regenerating the files(which are committed anyway)

scripts/utils/formatCode.js
```diff
@@ -1,11 +1,22 @@
...
-  const prettierConfig = prettier.resolveConfig.sync(filename);
-  prettierConfig.filepath = filename;
-  prettierConfig.parser = "babel";
+// TODO: Remove this `if` in Babel 8
+// Prettier only supports Node.js 10+, so we can fallback to not formatting
+// o CI on older Node.js versions
 
-  return prettier.format(code, prettierConfig);
-};
+if (process.env.CI && parseInt(process.versions.node, 10) < 10) {
```
nicolo-ribaudo, line 7: We use prettier format the generated type definition files. However, we don't really need to do this on CI and we can skip it on Node.js 6 and 8 without introducing problems.

zxbodya: looks good for me, was thinking about disabling code generation completely - but, I think, having it without prettier is a better option

nicolo-ribaudo:
For reviewers: 19a7acf4bd6351574f61d93b72653de71f2771a2 is what makes CI work on Node.js 6

nicolo-ribaudo:
Let's wait for Brian's go-ahead to check if we can merge this (https://github.com/babel/babel/pull/11579#issuecomment-630190639)

existentialism:
:shipit: !

nicolo-ribaudo:
Thanks @zxbodya!

PR #11580, opened 2020-05-18
Update typescript dependency

zxbodya:
This is needed in the context of migration to typescript #11578 to allow using `import type` typescript syntax.

Also, updated generated types in `@babel/types` to have `assert` annotations.

| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  | yes, updated typescript

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11580"></a>

### DIFF
--- a/packages/babel-types/scripts/generators/typescript.js
+++ b/packages/babel-types/scripts/generators/typescript.js
@@ -128,9 +128,8 @@ for (const typeName of t.TYPES) {
 
   lines.push(
     `export function is${typeName}(node: object | null | undefined, opts?: object | null): ${result};`,
-    // TypeScript 3.7: https://github.com/microsoft/TypeScript/pull/32695 will allow assert declarations
     // eslint-disable-next-line max-len
-    `// export function assert${typeName}(node: object | null | undefined, opts?: object | null): asserts ${
+    `export function assert${typeName}(node: object | null | undefined, opts?: object | null): asserts ${
       result === "boolean" ? "node" : result
     };`
   );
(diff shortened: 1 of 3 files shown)


packages/babel-types/scripts/generators/typescript.js
```diff
@@ -303,9 +301,8 @@ lines.push(
   // the MemberExpression implication is incidental, but it follows from the implementation
   // eslint-disable-next-line max-len
   `export function matchesPattern(node: Node | null | undefined, match: string | ReadonlyArray<string>, allowPartial?: boolean): node is MemberExpression`,
-  // TypeScript 3.7: ": asserts n is T"
   // eslint-disable-next-line max-len
-  `export function validate<T extends Node, K extends keyof T>(n: Node | null | undefined, key: K, value: T[K]): void`,
+  `export function validate<T extends Node, K extends keyof T>(n: Node | null | undefined, key: K, value: T[K]): asserts n is T`,
```
JLHwung, line 305: This change bumps minimum TypeScript version to 3.7 for `@babel/types` typings users, which is a breaking one. Can you separate `@babel/types` change into another PR?

zxbodya: once #11578 is landed, it would require typescript 3.8 at least (because of `import type`), that is why I thought it makes sense to have it here as well…

if you have a preference about it - I can move it to be together the rest of migration to ts

zxbodya:
changes are already made in other PRs

PR #11588, opened 2020-05-19
add showConfig support

### LINKED ISSUE #10538: Feature Request: a flag to print the current config
There doesn't appear to be an easy way to know which config settings Babel is using when processing your file. This is kind of insane given that Babel is known for being such a huge headache to configure, and I'd bet that the majority of the issues opened in this repository are likely frustrating misconfiguration issues.

It seems like…

- https://github.com/babel/babel/issues/2960 was opened in 2015, but then closed at some point and no one noticed?
- https://github.com/babel/babel/pull/4806 was submitted in 2016, and got reviewed but fell through the cracks again?
- https://github.com/babel/babel/issues/6856 was opened in 2017, but seems to have gone dormant too?
- https://github.com/babel/babel/pull/7201 was submitted in 2018, with a few comments but never got to the goal?

Please add this! I know it will make my life using Babel easier, instead of incredibly frustrating. And I feel like it will probably drastically reduce the number of issues you get, making your lives easier too. 😉 

TypeScript has solved this nicely with their `--showConfig` flag:

```bash
$ tsc --emitDeclarationOnly --declarationMap --outDir ./lib --showConfig
{
    "compilerOptions": {
        "allowSyntheticDefaultImports": true,
        "declaration": true,
        "esModuleInterop": true,
        "module": "esnext",
        "moduleResolution": "node",
        "sourceMap": true,
        "strict": true,
        "suppressImplicitAnyIndexErrors": true,
        "target": "esnext",
        "emitDeclarationOnly": true,
        "declarationMap": true,
        "outDir": "./lib"
    },
    "files": [
        "./src/index.ts"
    ],
    "include": [
        "./src/**/*.ts"
    ]
}
```

And it's the first thing I use when I can't figure out why something is not compiling as expected.

### LINKED ISSUE #10617: RFC: Print Babel Configuration

JLHwung:
| Fixed Issues?            | Fixes #10617, closes #11018, closes #10538
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | https://github.com/babel/website/pull/2302
| Any Dependency Changes?  |

`showConfig` will print the config chain applied on a specific file path by order of descending priority. It will also print the effective `overrides` and `env` config applied to a single file.

**Example output** ([input](https://github.com/JLHwung/babel/blob/print-config/packages/babel-cli/test/fixtures/babel/env%20-%20SHOW_CONFIG_FOR/in-files/my-config.js), [Full output](https://github.com/JLHwung/babel/blob/print-config/packages/babel-cli/test/fixtures/babel/env%20-%20SHOW_CONFIG_FOR/stdout.txt))

(The `<CWD>` and `<ROOTDIR>` is _not_ presented in the raw output, it is replaced by Babel's fixture runner.)

```
Babel configs on "<CWD>/src/index.js" (ascending priority):
config <CWD>/my-extended.js
{
  "sourceMaps": false,
  "presets": [
    "@foo/babel-preset-1"
  ]
}

config <CWD>/my-config.js
{
  "sourceType": "script",
  "plugins": [
    "[Function: (api) => ({/n  name: /"@foo//" + __dirname,/n  visitor ... ]"
  ],
  "extends": "./my-extended.js"
}

config <CWD>/my-config.js .env["test"]
{
  "plugins": [
    [
      "@foo/babel-plugin-3",
      {
        "noDocumentAll": true
      },
      "@foo/babel-plugin-three"
    ]
  ]
}

config <CWD>/my-config.js .overrides[0]
{
  "test": "src/index.js",
  "plugins": [
    [
      "@foo/babel-plugin-2",
      {
        "noDocumentAll": true
      }
    ]
  ]
}

config <CWD>/my-config.js .overrides[0].env["test"]
{
  "plugins": [
    "@foo/babel-plugin-1",
    [
      {
        "name": "@foo/inline-babel-plugin-1",
        "visitor": {}
      },
      {
        "noDocumentAll": true
      }
    ]
  ]
}

config <CWD>/.babelrc
{}

programmatic options from @babel/cli
{
  "sourceFileName": "./src/index.js",
  "presets": [
    "<ROOTDIR>/packages/babel-preset-react"
  ],
  "plugins": [
    "<ROOTDIR>/packages/babel-plugin-transform-arrow-functions",
    "<ROOTDIR>/packages/babel-plugin-transform-strict-mode",
    "<ROOTDIR>/packages/babel-plugin-transform-modules-commonjs"
  ],
  "configFile": "./my-config.js",
  "caller": {
    "name": "@babel/cli"
  },
  "filename": "./src/index.js"
}
```
**Known limits**
- It does not expand the `presets`
- It does not print the resolved config
- It does not tell users why certain file is ignored/excluded

These limits can be addressed in other PRs.

**Additional**
This PR includes commits in #11544.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11588"></a>


packages/babel-core/src/config/config-chain.js
```diff
@@ -712,3 +744,33 @@ function matchPattern(
...
+  formatter: $Values<typeof ChainFormatter>,
+  context: ConfigContext,
+  validatedFile?: ValidatedFile,
+): string {
+  function formatHeading(type, context, validatedFile) {
+    let heading = "";
+    if (type === ChainFormatter.Programmatic) {
+      heading = "programmatic options";
+      if (context.caller?.name) {
+        heading += " from " + context.caller.name;
```
JLHwung, line 759: It will print `programmatic options (from babel-loader)` when caller name is given.

packages/babel-core/src/config/config-chain.js
```diff
@@ -209,16 +229,30 @@ export function* buildRootChain(
...
       if (!result) return null;
+      babelRcReport = babelrcLogger.output();
 
       mergeChain(fileChain, result);
     }
   }
 
+  if (context.showConfig) {
+    console.log(
+      // print config by the order of descending priority
```
nicolo-ribaudo, line 249: Maybe we should print in the output that in case of conflicts the first config overrides the second?

packages/babel-core/src/config/partial.js
```diff
@@ -81,22 +81,29 @@ export default function* loadPrivatePartialConfig(
...
-        : undefined,
+    filename,
     cwd: absoluteCwd,
     root: absoluteRootDir,
     envName,
     caller,
+    showConfig:
+      showConfig ||
+      path.resolve(cwd, process.env.BABEL_SHOW_CONFIG_FOR_PATH ?? "") ===
+        filename,
```
nicolo-ribaudo, line 106: Maybe we should also allow setting `showConfig` to a string? I'm trying to think about one usecase where having a boolean and logging the config for _every_ file is useful, but I can't think of any common one :thinking:

packages/babel-core/src/config/printer.js
```diff
@@ -0,0 +1,133 @@
...
+
+function descriptorToConfig(d: UnloadedDescriptor): string | {} | Array<mixed> {
+  let name = d.file?.request;
+  if (name == null) {
+    if (typeof d.value === "object") {
+      name = d.value;
+    } else if (typeof d.value === "function") {
+      // If the unloaded descriptor is a function, i.e. `plugins: [ require("my-plugin") ]`,
+      // we print the first 50 characters of the function source code and hopefully we can see
+      // `name: 'my-plugin'` in the source
```
hzoo, line 78: Good hack, although I don't believe this will be the case for most plugins?

Even testing on https://codesandbox.io/s/lively-star-xv8bt, it will just print the code instead, the default plugin on ASTExplorer wouldn't even show the name until char 85.

JLHwung: Good catch. Well since an object has no key orders, I don't think there exists an integer N such that the plugin name is included in the `(0, N)` substring.

My initial thought here is to provide user at least something more descriptive/recognizable than a generic `[Function]` string. I have come up with several approach:

- Print Function name. It could not help much because we don't have such convention that the default export should be a named function
- Use regex to extract `name: "blablabla"` from function source. It doesn't work with dynamic name and has false positive.
- Actually load the descriptor. We have to initialize the API object earlier, and it may introduce breaking changes as we can end up loading a plugin twice.

So I end up with this approach and hope it could serve as an indicator so that users could know what this plugin is -- most of the time these are ad-hoc plugins not in `node_modules`.

hzoo: > Use regex to extract name: "blablabla" from function source. It doesn't work with dynamic name

I think it would be really rare/weird to have a dynamic name, it's just a string? Yeah I don't think it'll be an issue in general but for standalone/repl which is a diff usecase altogether.

nicolo-ribaudo: I think that this is something that we can revisit with experience: I believe that the hard-to-understand function name might not be a problem for a few reasons:
1. Most people use `"babel-plugin-foo"` or `require.resolve("babel-plugin-foo")`
2. `.slice(0, 50)` always includes the plugin function name (or at least the first 41 characters if it's longer). This can be useful for plugins directly defined in the config file, [as we do](https://github.com/babel/babel/blob/e498bee10f0123bb208baa228ce6417542a2c3c4/babel.config.js#L168).
3. If the plugin is an actual anonymous function and the `.slice(0, 50)` substring doesn't contain any useful information, the person can easily see in which config file it is defined and check what's used at that index in the `plugins` array directly in the config file.

packages/babel-core/src/config/config-chain.js
JLHwung: I tend to keep the output succinct. The implementation detail of config merging should be referenced from babel docs but not a debug output.

To reason how these config merges, we can print the resolved config in later PRs. Think of the `Styles` tab v.s. the `Computed` tab in the Chrome DevTools / Elements.

packages/babel-core/src/config/partial.js
JLHwung: I am not sure if logging the config for _every file_ is useful, because it can overwhelm quite easily for even a small project.

When people is not using `@babel/cli`, I prefer `env.BABEL_SHOW_CONFIG_FOR_PATH` over `showConfig: boolean` because environment variable is more robust, since people may end up adding `showConfig` to a babel config that is, unfortunately, _not_ loaded by `@babel/core` at all, and I definitely don't want people debugging `showConfig` itself.

nicolo-ribaudo: > I am not sure if logging the config for every file is useful, because it can overwhelm quite easily for even a small project.

Than maybe we don't need the boolean `showConfig`, and only rely on the env var?

> Do you think we should throw if path.resolve(cwd, process.env.BABEL_SHOW_CONFIG_FOR_PATH ?? "") does not even exist? This is the case when people specify it wrong and currently no config will be printed.

:+1:

JLHwung: I have removed `showConfig: boolean` support to make this PR minimal. We can add `showConfig: boolean` later if it is indeed a popular usage.

packages/babel-core/src/config/files/configuration.js
```diff
@@ -299,6 +299,20 @@ const readIgnoreConfig = makeStaticFileCache((filepath, content) => {
   };
 });
 
+export function* resolveShowConfigPath(
+  dirname: string,
+): Handler<string | null> {
+  const targetPath = process.env.BABEL_SHOW_CONFIG_FOR_PATH;
```
JLHwung, line 305: Do you think the following candidates are easier for typing?
```
BABEL_SHOW_CONFIG_FOR
BABEL_CONFIG_FOR
```

nicolo-ribaudo: I slightly prefer `BABEL_SHOW_CONFIG_FOR`

packages/babel-core/src/config/files/configuration.js
```diff
@@ -299,6 +299,23 @@ const readIgnoreConfig = makeStaticFileCache((filepath, content) => {
...
+export function* resolveShowConfigPath(
+  dirname: string,
+): Handler<string | null> {
+  const targetPath = process.env.BABEL_SHOW_CONFIG_FOR;
+  if (targetPath != null) {
+    const absolutePath = path.resolve(dirname, targetPath);
+    const stats = yield* fs.stat(absolutePath);
+    if (!stats.isFile()) {
+      throw new Error(
+        `${absolutePath}: BABEL_SHOW_CONFIG_FOR must refers to a regular file, directory is not supported.`,
```
existentialism, line 311:
```suggestion
        `${absolutePath}: BABEL_SHOW_CONFIG_FOR must refer to a regular file, directories are not supported.`,
```

packages/babel-core/src/config/config-chain.js
```diff
@@ -650,22 +726,28 @@ function shouldIgnore(
   dirname: string,
 ): boolean {
   if (ignore && matchesPatterns(context, ignore, dirname)) {
-    debug(
-      "Ignored %o because it matched one of %O from %o",
-      context.filename,
+    const message = `No config is applied on "${
```
existentialism, line 729: > No config applies to `<>` because it matches one of `<>`

existentialism:
Should we add a test for .only?

packages/babel-core/src/config/validation/options.js
```diff
@@ -187,6 +191,7 @@ export type ValidatedOptions = {
   inputSourceMap?: RootInputSourceMapOption,
   envName?: string,
   caller?: CallerMetadata,
+  showConfig?: boolean,
```
nicolo-ribaudo, line 194: Should the changes in this file be reverted because of https://github.com/babel/babel/pull/11588/commits/a4d31fa32ab4ded7e314529260357f0db6402d14?

JLHwung: Fixed(up) in cb3b76bba74e131c50fc4d2c17325a80132a15aa.

JLHwung:
@nicolo-ribaudo @existentialism WDYT we print config in the order of ascending priority instead?

I realize that although we have ordered config sources in descending order(programatic -> babelrc -> babel.config),  the config items in these sources are actually printed in ascending, i.e.
```
config <CWD>/my-config.js
{
  "sourceType": "script",
  "plugins": [
    "[Function: (api) => ({/n  name: /"@foo//" + __dirname,/n  visitor ... ]"
  ],
  "extends": "./my-extended.js"
}

config <CWD>/my-config.js .env["test"]
{
  "plugins": [
    [
      "@foo/babel-plugin-3",
      {
        "noDocumentAll": true
      },
      "@foo/babel-plugin-three"
    ]
  ]
}

config <CWD>/my-config.js .overrides[0]
{
  "test": "src/index.js",
  "plugins": [
    [
      "@foo/babel-plugin-2",
      {
        "noDocumentAll": true
      }
    ]
  ]
}
where `.overrides[0]` will overwrite `.ent["test"]` , which will overwrite base.

The inconsistence here will make people confusing about reasoning how they will be merged. I can go with either descending or ascending (mentally think of `Object.assign`, where the latter always overwrite the former) but we should be consistent here.

nicolo-ribaudo:
In general I prefer descending order (so that I see the most important thing first), but in this case I agree that it would probably be better to sort them in ascending order. However, I'd ask to output a line at the beginning saying the last thing is the one with the highest priority.

alexkreidler:
Just a follow up: is there a reason you removed (https://github.com/babel/babel/pull/11588/commits/bfdf5cf5a725976c35c6ba5eb056a0da59021479) the showConfig boolean option?

That seemed like the easiest way to use this feature.

nicolo-ribaudo:
The reason is that it would log the config for every file processed by Babel, generating an output so big that it would hardly be useful.

PR #11597, opened 2020-05-22
Fix comments for smartPipeline topic-forbidding contexts

lazytype:
| Patch: Bug Fix?          | just comment strings
| Documentation PR Link    | no

I think these comments were probably hastily copied and pasted between different statement types without updating the comments to be specific to that statement type. I updated the comments to reflect what's actually the case to the best of my ability.

### DIFF
--- a/packages/babel-parser/src/parser/statement.js
+++ b/packages/babel-parser/src/parser/statement.js
@@ -659,9 +659,7 @@ export default class StatementParser extends ExpressionParser {
 
       clause.body =
         // For the smartPipelines plugin: Disable topic references from outer
-        // contexts within the function body. They are permitted in function
-        // default-parameter expressions, which are part of the outer context,
-        // outside of the function body.
+        // contexts within the catch clause's body.
         this.withTopicForbiddingContext(() =>
           // Parse the catch clause's body.
           this.parseBlock(false, false),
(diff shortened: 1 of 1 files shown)


packages/babel-parser/src/parser/statement.js
```diff
@@ -1197,8 +1195,7 @@ export default class StatementParser extends ExpressionParser {
     this.expect(tt.braceL);
 
     // For the smartPipelines plugin: Disable topic references from outer
-    // contexts within the class body. They are permitted in test expressions,
-    // outside of the class body.
+    // contexts within the class body.
```
mAAdhaTTah, line 1198: I'm not sure this shortening is an improvement.

mAAdhaTTah:
The rest of these seem good tho.

packages/babel-parser/src/parser/statement.js
lazytype: I'm believe the existing comment was copy-pasted from `parseWhileStatement`, where the mention of a "test expression" makes sense since `while` loops have one, i.e.
```while (<test_expression>) {```
but since `class` bodies do not, removing mention of "test expression" is removing objectively false information from the comment (which I found to be quite confusing when I first encountered the comment).

mAAdhaTTah: Ah, fair.

PR #11600, opened 2020-05-23
Add hack style pipeline proposal

xixixao:
| Fixed Issues?            | Addresses https://github.com/tc39/proposal-pipeline-operator/issues/84| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | (I'll add this shortly)
| Any Dependency Changes?  |

This addresses the discussion starting at https://github.com/tc39/proposal-pipeline-operator/issues/167#issuecomment-631878219.

It adds the Hack-style only pipeline proposal. It's a strict version of the smart proposal, it's simpler, and I hope it has a better chance to go through the proposal process.

Even if this isn't your favorite version, or you don't believe this is the right path forward, surely we can agree it would be good to level the playing field to have this proposal implemented in Babel.

## Implementation details

I dropped the indirect eval. I don't see a reason to special eval for pipelines, the point of this proposal is that it is extremely simple and interacts well with existing code. Therefore this code:

```
var x = 5;
return "console.log(x)" |> eval;
```
should work just as well as
```
var x = 5;
return eval("console.log(x)");
```

## Tests

I converted the "bare" test to an error test (the other proposals are unfortunately missing tests for errors, so the coverage there isn't great).
I changed the eval test to test the "direct" eval.


aadamsx:
More examples please.

xixixao:
@aadamsx sorry, what do you mean by more examples? This PR includes a full test suite. What other kind of example would you like to see?

mAAdhaTTah:
To clarify: Hack is Smart without "bare style"?

nicolo-ribaudo:
According to the linked issue, it seems that hack doesn't use `#` but binds the piped value to the `$` binding 🤔

nicolo-ribaudo:
@tabatkins - RE: https://github.com/tc39/proposal-pipeline-operator/issues/167#issuecomment-632384719

Do you plan to talk about this during the June meeting? I tried looking for slides or examples in the agenda but I couldn't find them.

xixixao:
@mAAdhaTTah correct, this is like the "smart" proposal but without supporting "bare" calls.

@nicolo-ribaudo `$$` as topic reference works for Hack, but clearly doesn't for JS. `#` seems to be the most likely candidate, and it's already used for the smart proposal.

tabatkins:
Talking with the other pipeline champions, we'll be targetting the *July* meeting for pipeline discussions.

js-choi:
There’s now a [formal Hack-pipes proposal](https://github.com/js-choi/proposal-hack-pipes), which is replacing the now-withdrawn smart-mix pipes proposal. The next steps are:

1. First we will need to resolve this fork’s conflicts with Babel latest.
2. Then we will need to adjust its code to match the [new formal specification](https://jschoi.org/21/es-hack-pipes/), which has a bunch of grammatical simplifications.
3. We also will need to remove smart-mix functionality, since smart mix’s proposal has been withdrawn.
4. After this pull request gets merged into Babel latest, we should try to add support for different placeholder tokens like `%` and `@`, in another pull request.

I’ve been talking with @xixixao, @mAAdhaTTah, and @tabatkins about this. We’ll work on this when we can.

nicolo-ribaudo:
Thanks!

xixixao:
Updated to @js-choi 's new version of this PR.

js-choi:
Thanks, @xixixao. 

The current code now should have no conflicts with Babel upstream. In addition, the obsolete smart-mix plugin mode has been removed. ASTs also have been simplified; they just use `BinaryExpression` without any extra node types.

I think it would be a good idea to have the Hack plugin mode require a specified topic token, with no default token. (The default topic token is currently `#`, which had not been well received by TC39 in the past.)

To that end, I’m working on adding support for specifying the topic token right now, after which I’ll work on supporting `@`, `%`, and `?`.

packages/babel-parser/src/parser/error-message.js
```diff
@@ -129,18 +129,14 @@ export const ErrorMessages = Object.freeze({
...
-  PipelineHeadSequenceExpression:
-    "Pipeline head should not be a comma-separated sequence expression",
-  PipelineTopicUnused:
-    "Pipeline is in topic style but does not use topic reference",
-  PrimaryTopicNotAllowed:
-    "Topic reference was used in a lexical context without topic binding",
-  PrimaryTopicRequiresSmartPipeline:
-    "Primary Topic Reference found but pipelineOperator not passed 'smart' for 'proposal' option.",
+  PipeBodyCannotBeArrow:
+    'Unexpected arrow "=>" after pipeline body; arrow function acting as pipe body must be parenthesized due to operator precedence',
```
JLHwung, line 133: Nit: we are now [enforcing](https://github.com/babel/babel/pull/13130) consistent error message format. They should end with a period(`.`).

nicolo-ribaudo:
Can we keep adding the hack proposal and removing the smart one in separate PRs? We cannot remove the smart proposal until the next major, but the hack proposal can be implemented in a minor.

js-choi:
> Can we keep adding the hack proposal and removing the smart one in separate PRs? We cannot remove the smart proposal until the next major, but the hack proposal can be implemented in a minor.

Alright, noted; I’ll add smart-mix mode back in a future commit.

packages/babel-parser/src/parser/error-message.js
js-choi: Fixed in 5c4caf25271052d7a8a317f893cff58b0499e2d0. I changed only pipe-related errors; I’ll defer changing the other errors to #13130.

xixixao:
Superceded by https://github.com/babel/babel/pull/13191

PR #11612, opened 2020-05-26
Make error message prefixes more descriptive

eps1lon:
~Sorry for the drive-by contribution. Didn't want to forget about it. I'll look at possible CI failures later.~ Updated tests

I noticed when testing errors for babel macro with `babel-plugin-tester` that the error messages would be prefixed with `unknown:`. It wasn't clear to me whether this was an issue on my side with custom errors or babel-plugin-tester. 

Turns out this is expected if the code does not come from an actual file. Hopefully this is less confusing for future users if we prefix it with "unknown file" instead of "unknown". It's not obvious to me that "unknown" refers to a file in the error context.

It might even make more sense to not prefix the filename and instead append the filename e.g.

```
Some error message from plugin X
  in /path/to/file/that/is/transformed.js
```

| Fixed Issues?            | found none using keywords `unknown error`
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      | Probably
| Documentation PR Link    | 
| Any Dependency Changes?  |

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11612"></a>

### DIFF
--- a/packages/babel-core/src/config/full.ts
+++ b/packages/babel-core/src/config/full.ts
@@ -220,7 +220,9 @@ function enhanceError<T extends Function>(context: ConfigContext, fn: T): T {
       // There are a few case where thrown errors will try to annotate themselves multiple times, so
       // to keep things simple we just bail out if re-wrapping the message.
       if (!/^\[BABEL\]/.test(e.message)) {
-        e.message = `[BABEL] ${context.filename || "unknown"}: ${e.message}`;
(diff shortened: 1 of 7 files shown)


nicolo-ribaudo:
Related - https://github.com/babel/babel/pull/11554

nicolo-ribaudo (approved):
I like this, I'll take care of rebasing once https://github.com/babel/babel/pull/11554 is merged.

liuxingbaoyu (approved):

PR #11623, opened 2020-05-27
fix: ensure (a?.b)() has proper this

### LINKED ISSUE #10802: Optional Chaining does not propagate receiver when parenthesized

JLHwung:
| Fixed Issues?            | Fixes #10802, closes #11327 

It has been out there for quite a while and the fix seems more tricky than previously proposed in #11327 and #10802.

This PR supersedes #11552 .
 
Todo
- [x] `(a?.#m)()`
- [x] `parserOpts.createParenthesizedExpression`


packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -167,13 +167,19 @@ const handle = {
       const parentIsOptionalCall = parentPath.isOptionalCallExpression({
         callee: node,
       });
+      const isParenthesizedMemberCall =
+        parentPath.isCallExpression({ callee: node }) &&
+        node.extra?.parenthesized;
```
JLHwung, line 172: The current approach does not support `parserOpts.createParenthesizedExpressions`. When developing I realize that the whole transform does not take `ParenthesizedExpressions` into account, I suggest we address it in a separate PR before this one gets too big to review.

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -23,14 +23,23 @@ export default declare((api, options) => {
...
         let isDeleteOperation = false;
         const optionals = [];
 
         let optionalPath = path;
         while (
           optionalPath.isOptionalMemberExpression() ||
-          optionalPath.isOptionalCallExpression()
+          optionalPath.isOptionalCallExpression() ||
+          optionalPath.isParenthesizedExpression() ||
+          optionalPath.isTSNonNullExpression()
```
JLHwung, line 42: By doing so we also fix cases like `a!!?.b` are not transformed currently.

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -323,6 +330,13 @@ const privateNameHandlerLoose = {
     });
   },
 
+  boundGet(member) {
+    return t.callExpression(
+      t.memberExpression(this.get(member), t.identifier("bind")),
+      [member.node.object],
```
jridgewell, line 336: Don't we need to clone?

packages/babel-helper-create-class-features-plugin/src/fields.js
```diff
@@ -241,6 +241,13 @@ const privateNameHandlerSpec = {
     ]);
   },
 
+  boundGet(member) {
+    return t.callExpression(
+      t.memberExpression(this.get(member), t.identifier("bind")),
```
jridgewell, line 246: Don't we need to memoize the member?

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -23,14 +23,23 @@ export default declare((api, options) => {
...
       "OptionalCallExpression|OptionalMemberExpression"(path) {
-        const { parentPath, scope } = path;
+        const { scope } = path;
+        let parentPath;
+        // unwrap parenthesis around parent path
+        for (
+          { parentPath } = path;
+          parentPath.type === "ParenthesizedExpression";
+          { parentPath } = parentPath
+        );
```
jridgewell, line 33: Maybe we should call it something other than `parentPath`?

```suggestion
        const parentPath = path.findParent(p => !p.isParenthesizedExpression());
```

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -113,7 +120,36 @@ export default declare((api, options) => {
...
+              // it must be unwrapped from parenthesized expression.
+              replacementPath.parentPath !== parentPath) &&
+            parentPath.isCallExpression()
+          ) {
+            // `(a?.b)()` to `(a == null ? undefined : a.b.bind(a))()`
+            const { object } = replacement;
+            let baseRef;
+            if (!loose) {
+              // memoize the context object in non-loose mode
+              // `(a?.b.c)()` to `(a == null ? undefined : (_a$b = a.b).c.bind(_a$b))()`
```
jridgewell, line 138: Do we need to check if there's a call expression in the path? `(o?.Foo().m)();` will double call, I think.

packages/babel-plugin-proposal-optional-chaining/src/index.js
nicolo-ribaudo: @oliverdunk Should we also check for this in your PR?

oliverdunk: Makes sense, I'll add that now 👍

PR #11624, opened 2020-05-27
Fix standalone tag when data-type is not set.

### LINKED ISSUE #11621: In-browser JSX translation broken with 7.9 -> 7.10 upgrade.
## Bug Report

**Current behavior**

In browser JSX translation does not work for Babel 7.10.0.

A script tag in HTML like the following works fine.
<script src="normal_script.js"></script>

A tag like this does not work:
<script src="react_script.jsr" type="text/babel"></script>

By *does not work* I mean that no Javascript in that file will execute. Not even simple JS that does not require any translation. For example, if react_script.jsr contained only this line:

`console.log("hi` world");

it would fail to execute.

This issue is remedied back to normal if following change is made to the HTML.

<script src="https://unpkg.com/@babel/standalone/babel.min.js"></script>

changed to

<script src="https://unpkg.com/@babel/standalone@7.9.0/babel.min.js"></script>

**Expected behavior**

The above not to break as described.

**Environment**

```
- Babel version(s): v7.10.0
- OS: Windows 10 and Linux 4.19
- Browser: Chrome and Firefox

**Possible Solution**
Temp workaround: use v7.9.0 of babel.

### LINKED ISSUE #11628: babel stand alone version 7.10.0 doesnt work
7.10 doent seems do do anything at all. null behavoir. I had to revert to 7.9

dfabulich:
Fixes #11621.

| Fixed Issues?            | Fixes #11621, fixes #11628
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

I thought `getAttribute` returned `undefined` when the attribute was unset, but it actually returns null. 🤦‍♂️

I manually tested this patch with and without a `data-type` attribute. (It appears that the babel-standalone script tag has no tests?? I might add some later, but for now, I want to fix this bug right away.)

### DIFF
--- a/packages/babel-standalone/src/transformScriptTags.js
+++ b/packages/babel-standalone/src/transformScriptTags.js
@@ -53,7 +53,7 @@ function buildBabelOptions(script, filename) {
  */
 function run(transformFn, script) {
   const scriptEl = document.createElement("script");
-  if (typeof script.type !== "undefined") {
+  if (script.type) {
     scriptEl.setAttribute("type", script.type);
   }
   scriptEl.text = transformCode(transformFn, script);


nicolo-ribaudo:
Thanks!
Yeah, we should figure out how to test this.

existentialism:
Yeah let's go with a falsy check

dfabulich:
Switched to a truthy check. `if (script.type)`

I'm a bit surprised to see that y'all had an empty string; I tested the previous version on Chrome 83, Safari 13.1, and Firefox 76, both with and without a `data-type` attribute.

existentialism:
@dfabulich thanks for the quick fix!

Issue #11499, opened 2020-04-29
Better parser error when using jsx


hzoo:
Was checking https://github.com/babel/babel/pull/11478 and testing out syntax and noticed we don't error with the normal "use this plugin/preset" like we do for proposals with JSX itself. @nicolo-ribaudo mentioned it's probably since jsx is implemented as a normal [plugin](https://github.com/babel/babel/tree/master/packages/babel-parser/src/plugins) so we didn't have `this.expectPlugin("recordAndTuple");`

```sh
unknown: Unterminated regular expression (1:5)

> 1 | <a></a>
    |      ^
```

```sh
unknown: Support for the experimental syntax 'pipelineOperator' isn't currently enabled (1:3):

> 1 | a |> b
    |   ^

Add @babel/plugin-proposal-pipeline-operator (https://git.io/vb4SU) to the 'plugins' section of your Babel config to enable transformation.
```

nicolo-ribaudo:
This bug needs different steps in order to be fixed, but they can be done in the same PR.

1. First, we should make `@babel/parser` throw a developer-friendly error.
   Consider these examples:
   ```js
   require("@babel/parser").parse("<div></div>", {});
   require("@babel/parser").parse("a |> b", {});
   ```
    The first one throws `SyntaxError: Unexpected token (1:0)`, while the second one throws a more actionable `SyntaxError: This experimental syntax requires enabling the parser plugin: 'pipelineOperator' (1:2)`.
    Why this difference? If you search for the `"pipelineOperator"` string in `packages/babel-parser/src`, you'll see that (as @hzoo mentioned above) we are calling the `this.expectPlugin("pipelineOperator")` function [when we find the "pipeline" (`|>`) operator](https://github.com/babel/babel/blob/bb180eba9f3a2d0dc69e62beb788bfb8a9215a47/packages/babel-parser/src/parser/expression.js#L365-L366).

   We can do something similar for JSX, but it's harder. The main problem is that checks for ECMAScript proposals (such as `|>`) are inlined in the parser code, while checks for JSX code are only enabled when the `jsx` plugin is enabled.

   However, we know that JSX elements are _atomic expressions (parsed by the [`parseExprAtom`](https://github.com/babel/babel/blob/bb180eba9f3a2d0dc69e62beb788bfb8a9215a47/packages/babel-parser/src/parser/expression.js#L877) method) and that no other expression can start with `<`.
[...]

PR #11629, opened 2020-05-27
Add jsx parser missing plugins errors #11499

HiMrHu:
| Fixed Issues?            | #11499 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-parser/test/fixtures/jsx/errors/_no-plugin-ts-type-param/options.json
```diff
@@ -0,0 +1,4 @@
+{
+  "throws": "This experimental syntax requires enabling the parser plugin: 'jsx' (1:0)",
```
nicolo-ribaudo, line 2: Maybe we could throw `this.expectOnePlugin(["jsx", "flow", "typescript"])` to give an error that is also correct in this case?

PR #11631, opened 2020-05-27
Fix moduleAttributesVersion errors with stage-0 preset in babel standalone

### LINKED ISSUE #11628: babel stand alone version 7.10.0 doesnt work

hamlim:
| Fixed Issues?            | Fixes #11628
| Patch: Bug Fix?          | ✅ 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Changes:

* Expose `moduleAttributesVersion` config option on the `stage-0` preset for babel standalone
* Forward that `moduleAttributesVersion` onto the config for `stage-1` preset
* Fix version mismatch in the `module-attributes` plugin and `plugin-utils` file (one was expecting `version: 'apr-2020'` and the other was expecting `version: 'may-2020'`, I **assumed** that may is the correct version here) See: https://github.com/babel/babel/blob/master/packages/babel-parser/src/plugin-utils.js#L95-L101
* Updated config in the `example.htm` file for testing the changes locally

Testing steps:

* `make prepublish-build-standalone`
* `npx serve ./packages/babel-standalone`
* Visit localhost and navigate to `examples/example.htm`
* Verify that babel **doesn't** hit an error when transforming the code from the textarea

### DIFF
--- a/packages/babel-plugin-syntax-module-attributes/src/index.js
+++ b/packages/babel-plugin-syntax-module-attributes/src/index.js
@@ -3,11 +3,11 @@ import { declare } from "@babel/helper-plugin-utils";
 export default declare((api, { version }) => {
   api.assertVersion(7);
 
-  if (typeof version !== "string" || version !== "apr-2020") {
+  if (typeof version !== "string" || version !== "may-2020") {
     throw new Error(
       "The 'moduleAttributes' plugin requires a 'version' option," +
         " representing the last proposal update. Currently, the" +
-        " only supported value is 'apr-2020'.",
+        " only supported value is 'may-2020'.",
     );
   }
 
--- a/packages/babel-standalone/examples/example.htm
+++ b/packages/babel-standalone/examples/example.htm
@@ -1,38 +1,49 @@
 <!DOCTYPE html>
 <html>
-<head>
-  <meta charset="utf-8" />
-  <title>babel-standalone example</title>
-</head>
-<body>
-  Input:
-  <textarea id="input" style="width: 100%" rows="10">
+  <head>
+    <meta charset="utf-8" />
+    <title>babel-standalone example</title>
+  </head>
+  <body>
+    Input:
+    <textarea id="input" style="width: 100%" rows="10">
 const getMessage = () => 'Hello World';
 const someDiv = <div>{getMessage()}</div>;
-</textarea>
+</textarea
+    >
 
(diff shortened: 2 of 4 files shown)


packages/babel-plugin-syntax-module-attributes/src/index.js
```diff
@@ -3,11 +3,11 @@ import { declare } from "@babel/helper-plugin-utils";
 export default declare((api, { version }) => {
   api.assertVersion(7);
 
-  if (typeof version !== "string" || version !== "apr-2020") {
+  if (typeof version !== "string" || version !== "may-2020") {
```
nicolo-ribaudo, line 6: Good catch!

packages/babel-standalone/examples/example.htm
```diff
@@ -1,38 +1,49 @@
 <!DOCTYPE html>
 <html>
```
nicolo-ribaudo, line 2: Could you revert the changes to this file?

nicolo-ribaudo:
I now we don't have many tests for `standalone`, but could you add one to https://github.com/babel/babel/blob/master/packages/babel-standalone/test/babel.js? 🙏

hamlim:
> I now we don't have many tests for `standalone`, but could you add one to [/packages/babel-standalone/test/babel.js@`master`](https://github.com/babel/babel/blob/master/packages/babel-standalone/test/babel.js) [](/babel/babel/blob/HEAD@{2020-05-27T17:28:57Z}/packages/babel-standalone/test/babel.js)? 🙏

Sure! Would something like asserting that stage-0 doesn't throw when passing `moduleAttributesVersion` work?

hamlim:
Added a small regression test 👍

packages/babel-standalone/examples/example.htm
```diff
@@ -1,38 +1,49 @@
...
+      function transform() {
+        try {
+          outputEl.innerHTML = Babel.transform(inputEl.value, {
+            presets: [
+              "es2015",
+              "react",
+              [
+                "stage-0",
+                {
+                  decoratorsBeforeExport: false,
```
JLHwung, line 34: I don't think we need to change the example once we offers default value to these options. The example should serve as a minimal working config.

PR #11633, opened 2020-05-27
[hotfix] Use same targets for fields as for private methods

nicolo-ribaudo:
| Fixed Issues?            | https://twitter.com/IAmTrySound/status/1265639257571540993
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

```js
babel.transformSync("class A { x }", {
  configFile: false,
  presets: [
    ["@babel/env", {
      targets: { node: 12 },
      shippedProposals: true,
    }],
  ],
});
```

The above code doesn't work with Babel 7.10.0 because
1. Class fields are supported, so the plugin is not enabled
2. Private methods are not supported, so the plugin is enabled and we use `create-class-features`
3. When transforming a class with the `create-class-features`, we check if all the eeded features are enabled. We don't support transforming private methods without transforming private fields, so we throw an error

I'm working on removing the dependency between the two plugins, but this quick and dirty hack makes preset-env work until I open a better PR. I would like to merge this one for the next patch release (today?).

cc @TrySound


TrySound:
Yay, thanks!

TrySound:
@nicolo-ribaudo I just got the same problem with numeric separator in node 13

nicolo-ribaudo:
That is unexpected 🤔

The problem fixed by this PR is that the private methods and class fields plugins depend on each other, and on Node.js 12 only one of them was enabled.

Numeric separators don't have such hidden dependency 🤔

TrySound:
This is weird, I got syntax error similar to class properties.

nicolo-ribaudo:
What is the exact error?

TrySound:
`[!] Error: Identifier directly after number` points to object with numeric separator.

This is very weird, syntax plugin is included. Here's debug output
```
@babel/preset-env: `DEBUG` option

Using targets:
{
  "node": "13.12"
}

Using modules transform: false

Using plugins:
  syntax-numeric-separator { "node":"13.12" }
  proposal-class-properties { "node":"13.12" }
  proposal-private-methods { "node":"13.12" }
  proposal-nullish-coalescing-operator { "node":"13.12" }
  proposal-optional-chaining { "node":"13.12" }
  syntax-json-strings { "node":"13.12" }
  syntax-optional-catch-binding { "node":"13.12" }
  syntax-async-generators { "node":"13.12" }
  syntax-object-rest-spread { "node":"13.12" }
  syntax-dynamic-import { "node":"13.12" }
  syntax-top-level-await { "node":"13.12" }

Using polyfills: No polyfills were added, since the `useBuiltIns` option was not set.
```

By the way same problem with 14.3. Though I cannot reproduce it in isolated example.

nicolo-ribaudo:
Could you share your lockfile in a gist?

TrySound:
Oh, I get it. Sorry, it's rollup syntax error 🤦

Issue #11622, opened 2020-05-27
Odd 'loose' mode configuration must be the same...


dimaqq:
`'loose' mode configuration must be the same for both @babel/plugin-proposal-class-properties and @babel/plugin-proposal-private-methods`

I'm getting this error even though I don't use either of the proposals explicitly.
Thus, it's kinda hard for me to configure their `loose` modes 😆 
I think they are brought in through a preset.

My config is:
```
  "babel": {
    "plugins": [
      "babel-plugin-macros"
    ],
    "presets": [
      [
        "@babel/preset-env",
        {
          "useBuiltIns": false,
          "shippedProposals": true,
          "debug": false
        }
      ]
    ]
  },
  "babelMacros": {
    "styledComponents": {
      "pure": true
    }
  },
```

I've recently done `yarn upgrade` and this error showed up.
Summary of babel versions that got upgraded in `yarn.lock`:
* 7.9.6 → 7.10.0: core, runtime, compat-data, generator, helpers, parser, traverse, types, helper-compilation-targets, helper-create-class-features-plugin, helper-replace-supers, ...
* 7.8.6 → 7.10.0: template
* 7.8.3 → 7.10.0: helper-member-expressions-to-function, plugin-transform-spread
* 7.9.0 → 7.10.0: plugin-transform-for-of
* ...

existentialism:
Both `@babel/plugin-proposal-class-properties` and `@babel/plugin-proposal-private-methods` are enabled via `shippedProposals`. 

Are you sure you don't have another babel config (like a .babelrc file) that is enabling loose mode for class-properties?

dimaqq:
I don't have another config... at least not anything I've written.
I am using CRA + config overrides, where I tell webpack(?) to look for babel config in `package.json`. I guess it's possible that CRA set some options somewhere...

I've tried:
* explicit `"loose": false` in `"@babel/preset-env"` options: does not help 👎 
* explicit `"loose": true` in `"@babel/preset-env"` options: helps, I can build again 👍

I guess I'm stuck in `loose` mode now.
(not that it worries me, I've only learned about it today)

Would it help if I cut down my setup to a small-ish public repo?

existentialism:
CRA sets class properties to loose :)

https://github.com/facebook/create-react-app/blob/master/packages/babel-preset-react-app/create.js#L151

dbayerlein-bhs:
I have the same problem with the lastest beta of Storybook.

danielbayerlein:
[...]

PR #11634, opened 2020-05-27
Class features loose should have precedence over preset-env

nicolo-ribaudo:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/11622
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Please suggest me a better method to fix this if possible 😭


packages/babel-helper-create-class-features-plugin/src/features.js
```diff
@@ -18,49 +24,67 @@ export const FEATURES = Object.freeze({
...
   //   @babel/plugin-class-properties { loose: false }
   // is transformed in loose mode.
   // We only enabled the feature if it was previously disabled.
-  if (!hasFeature(file, feature)) {
+  if (!hasFeature(file, feature) || canIgnoreLoose(file, feature)) {
     file.set(featuresKey, file.get(featuresKey) | feature);
-    if (loose) file.set(looseKey, file.get(looseKey) | feature);
+    if (
+      loose ===
+      "#__internal__@babel/preset-env__prefer-true-but-false-is-ok-if-it-prevents-an-error"
```
JLHwung, line 49: Lol I can't prevent myself from thinking of https://github.com/facebook/react/blame/80bff5397bf854750dbe7c286f61654ea58938c5/src/umd/ReactUMDEntry.js#L21

packages/babel-helper-create-class-features-plugin/src/features.js
```diff
@@ -18,49 +24,67 @@ export const FEATURES = Object.freeze({
...
+
+    const loose = isLoose(file, mask);
+
+    if (canIgnoreLoose(file, mask)) {
+      continue;
+    } else if (resolvedLoose === !loose) {
+      throw new Error(
+        "'loose' mode configuration must be the same for @babel/plugin-proposal-class-properties, " +
+          "@babel/plugin-proposal-private-methods and " +
+          "@babel/plugin-proposal-private-property-in-object (whe they are enabled).",
```
JLHwung, line 77:
```suggestion
          "@babel/plugin-proposal-private-property-in-object (when they are enabled).",
```

packages/babel-preset-env/src/index.js
```diff
@@ -301,10 +301,25 @@ export default declare((api, opts) => {
...
   const pluginUseBuiltIns = useBuiltIns !== false;
   const plugins = Array.from(pluginNames)
-    .map(pluginName => [
-      getPlugin(pluginName),
-      { spec, loose, useBuiltIns: pluginUseBuiltIns },
-    ])
+    .map(pluginName => {
+      if (
+        pluginName === "proposal-class-properties" ||
+        pluginName === "proposal-private-methods"
```
JLHwung, line 307: I think we can use `classFeaturePlugins = ["proposal-class-properties", "proposal-private-methods", "proposal-private-property-in-object"]` to align with the class feature plugin factory. Although it is not included in preset-env yet but we are likely to forget adding them when `privateIn` reaches stage 3.

existentialism:
Thought about this a bit... I think this is fine. Can we possibly emit a warning when the loose mode from `@babel/preset-env` is overridden by another plugin's option?

nicolo-ribaudo:
@existentialism I have added the warning. You can see in the tests added in the last commit that it can still be explicitly silenced by explicitly enabling the plugins with the correct loose mode.

existentialism:
Is it possible to show them which plugin overrode?

I think something like:

> Though the `loose` option was set to `true` in your `@babel/preset-env` config, it will not be used since the 'loose' mode option was set for {plugin name}. The `loose` option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled).

Might be slightly more helpful in both letting them know what is going on, and how to fix it.

nicolo-ribaudo:
Better?

existentialism:
I think it is, you?

We might be able to shorten it, but thats more bikeshedding for later. I think it describes the problem well and how to fix it.

nicolo-ribaudo:
Yeah, we don't need to shorten it. It's more important to exactly explain what's going on and how to fix it.

PR #11639, opened 2020-05-28
feature: babel-eslint-parser passes through config options

kaicataldo:
| Fixed Issues?            | refs https://github.com/babel/babel/pull/11334
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

As discussed in https://github.com/babel/babel/pull/11334, this PR allows for users to pass through Babel configuration options directly to Babel's parse method, allowing for `@babel/eslint-parser` to be used without a Babel configuration file.

We already have a few tests that cover this code path (see [here](https://github.com/babel/babel/blob/master/eslint/babel-eslint-parser/test/index.js#L7-L11)). I feel like this is sufficient for now, but please let me know if you would rather I added more test cases.

cc @jharris4


jharris4:
@kaicataldo thanks so much for this, it's really awesome to see this finally get implemented!

If I still lived in Brooklyn (and if it weren't for covid) I'd buy you a beer. In the meantime is there a spot I can make a one off contribution towards your OSS work?

eslint/babel-eslint-parser/src/configuration.js
```diff
@@ -41,6 +26,7 @@ export function normalizeBabelParseConfig(options) {
     caller: {
       name: "@babel/eslint-parser",
     },
+    ...options.babelOptions,
```
nicolo-ribaudo, line 29: I would prefer to deeply merge the caller and the parser options, similarly to how we do for `caller` in `babel-loader` ([source](https://github.com/babel/babel-loader/blob/adac4592274e151ebf0313a33da553a19a8bf39d/src/injectCaller.js#L6)):

```js
    ...options.babelOptions,
    parserOpts: {
      allowImportExportEverywhere: options.allowImportExportEverywhere,
      allowSuperOutsideMethod: true,
    ...options.babelOptions.parserOpts,
      plugins: ["estree", ...options.babelOptions.parserOpts?.plugins ?? []],
      ranges: true,
      tokens: true,
    },
    caller: {
      name: "@babel/eslint-parser",
    ...options.babelOptions.caller,
    },
```

By doing so, we make sure to keep out defaults (especially for the parser options) even if someone overwrites some of them.

kaicataldo: Makes sense!

kaicataldo:
@jharris4 Thanks for the kind words ❤️

I currently have a [GitHub Sponsors profile](https://github.com/sponsors/kaicataldo), but it unfortunately doesn't allow for one time donations.

Hit me up if you end up back in Brooklyn post-pandemic - would love to hang out!

kaicataldo:
Thanks for the reviews, all!

PR #11641, opened 2020-05-29
Add support for printing ImportAttribute

existentialism:
| Any Dependency Changes?  |

Small follow up to https://github.com/babel/babel/pull/10962!


packages/babel-generator/test/fixtures/types/ImportAttribute/input.js
```diff
@@ -0,0 +1 @@
+import foo from "foo.json" with type: "json";
```
JLHwung, line 1: Can you also add a test case for dynamic import? Though currently it should be supported as well.
```js
import("foo.json", { with: { type: "json" } });
```

PR #11652, opened 2020-05-30
fix: implement early errors for record and tuple

JLHwung:
| Fixed Issues?            | Implements early errors for record and tuple

Implements early syntax errors specified in https://github.com/tc39/proposal-record-tuple#syntax-errors

- Allow only `ObjectProperty` and `SpreadElement` in `RecordExpression`.
- Disallow holes in `TupleExpression`
<s>I don't have a good idea on throwing recoverable errors other than passing through `isTuple` everywhere.</s>They are now recoverable.
- Disallow `__proto__` as key in `RecordExpression`. 

@rickbutton BTW do you think we can deprecate `syntaxType: bar`? It seems to me `{|` is longer proposed because I don't see a spec text on that.


nicolo-ribaudo:
What if in `parseExprListItem` we _always_ report "Empty holes are not allowed here" when `allowEmpty` is false?

JLHwung:
@nicolo-ribaudo `parseExprListItem(false, ?)` is also used in `parseCallExpressionArguments`. Do we have mental concept of holes in call arguments?

nicolo-ribaudo:
> Do we have mental concept of holes in call arguments?

You are right, we don't. However, it's an error we can easily recover from: we just need to find a good error message for it. Maybe `A comma can only be directly followed by another comma in array literals, creating a hole`?

packages/babel-parser/src/parser/error-message.js
```diff
@@ -163,7 +166,7 @@ export const ErrorMessages = Object.freeze({
     "Private names can only be used as the name of a class element (i.e. class C { #p = 42; #m() {} } )\n or a property of member expression (i.e. this.#p).",
   UnexpectedReservedWord: "Unexpected reserved word '%0'",
   UnexpectedSuper: "super is only allowed in object methods and classes",
-  UnexpectedToken: "Unexpected token '%'",
+  UnexpectedToken: "Unexpected token '%0'",
```
JLHwung, line 169: 🤦‍♂️

packages/babel-parser/src/parser/expression.js
```diff
@@ -2088,7 +2103,10 @@ export default class ExpressionParser extends LValParser {
     allowPlaceholder: ?boolean,
   ): ?N.Expression {
     let elt;
-    if (allowEmpty && this.match(tt.comma)) {
+    if (this.match(tt.comma)) {
+      if (!allowEmpty) {
+        this.raise(this.state.pos, Errors.UnexpectedToken, ",");
```
JLHwung, line 2108: I resort to the most generic error message. I still feel awkward mentioning holes when throwing `f(,)`. Any suggestions are definitely welcome.

At least the error is now recoverable.

rickbutton:
@JLHwung thank you for pinging me on this! The changes you have listed look good to me. On syntax options, there is still some discussion in https://github.com/tc39/proposal-record-tuple/issues/10 on which exact sigils to use for the literal syntax. I would be fine with removing the `bar` syntaxType option, but it is possible that the syntax will still change (although the champion group prefers `#`).

JLHwung:
@nicolo-ribaudo Can we consider this PR has also approval from Rick Button?

nicolo-ribaudo:
Yes :+1:

PR #11653, opened 2020-05-30
refactor: split locationParser into ParserErrors and error message

JLHwung:

Rename `packages/babel-parser/src/parser/location.js` into `error.js` and extract error message templates into a dedicated file.


JLHwung:
@kaicataldo I think this change is related to https://github.com/babel/babel/pull/11376/files#diff-e4403a877d80de653400d88d85e4801aR72.

kaicataldo:
Yeah, will update it once this is merged!

PR #11662, opened 2020-06-02
refactor: do not rely on AST extra properties in plugins

JLHwung:
| Fixed Issues?            | See explainers below

As a follow-up to #11623, this PR refines `(a?.b)()` detection and ensures that plugins never rely on `extra` AST properties. Also fixes a regression introduced in #11623 that `f((a?.b))` is incorrectly transformed into `f((a == null ? void 0 : a.b.bind(a)))`.


packages/babel-plugin-proposal-optional-chaining/test/fixtures/general/parenthesized-member-call-loose/output.js
```diff
@@ -32,8 +32,8 @@ class Foo {
     (o == null ? void 0 : o.Foo.m.bind(o.Foo))().toString();
     ((_o$Foo$self$getSelf = ((_o$Foo = o.Foo) == null ? void 0 : _o$Foo.self.getSelf.bind(_o$Foo.self))()) == null ? void 0 : _o$Foo$self$getSelf.m.bind(_o$Foo$self$getSelf))();
     ((_o$Foo$self$getSelf2 = ((_o$Foo$self = o.Foo.self) == null ? void 0 : _o$Foo$self.getSelf.bind(_o$Foo$self))()) == null ? void 0 : _o$Foo$self$getSelf2.m.bind(_o$Foo$self$getSelf2))();
-    ((_fn$Foo$self$getSelf = ((_fn = fn()) == null ? void 0 : (_fn$Foo = _fn.Foo) == null ? void 0 : _fn$Foo.self.getSelf.bind(_fn.Foo.self))()) == null ? void 0 : _fn$Foo$self$getSelf.m.bind(_fn$Foo$self$getSelf))();
-    ((_fn$Foo$self$getSelf2 = (fn == null ? void 0 : (_fn$Foo$self2 = _fn$Foo$self = fn().Foo.self) == null ? void 0 : _fn$Foo$self2.getSelf.bind(_fn$Foo$self))()) == null ? void 0 : _fn$Foo$self$getSelf2.m.bind(_fn$Foo$self$getSelf2))();
```
JLHwung, line 36: ✌️Unnecessary memoizers removed.

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -24,8 +24,19 @@ export default declare((api, options) => {
     visitor: {
       "OptionalCallExpression|OptionalMemberExpression"(path) {
         const { scope } = path;
-        const parentPath = path.findParent(p => !p.isParenthesizedExpression());
+        // maybeParenthesized points to the outermost parenthesizedExpression
+        // or the path itself
+        let maybeParenthesized = path;
+        const parentPath = path.findParent(
+          p => !(p.isParenthesizedExpression() && (maybeParenthesized = p)),
+        );
```
nicolo-ribaudo, line 32: Nit: could you refactor this into something longer but easier to understand?

```suggestion
        const parentPath = path.findParent(p => {
          if (!p.isParenthesizedExpression()) return true;
          maybeParenthesized = p;
        });
```

packages/babel-plugin-proposal-optional-chaining/src/index.js
```diff
@@ -116,14 +128,7 @@ export default declare((api, options) => {
...
           // Ensure (a?.b)() has proper `this`
-          if (
-            t.isMemberExpression(replacement) &&
-            (replacement.extra?.parenthesized ||
-              // if replacementPath.parentPath does not equal parentPath,
-              // it must be unwrapped from parenthesized expression.
-              replacementPath.parentPath !== parentPath) &&
-            parentPath.isCallExpression()
-          ) {
+          if (i === 0 && parentIsCall) {
```
jridgewell, line 131: What happens if we remove the `i` check?

JLHwung: Good question!

Since `parentIsCall` is computed only once for the `baseRef`, if we don't limit this branch with `i === 0`, which means applying to the baseRef only, it will generate multiple memoizing assignment during looping all the optional chain elements.

This change fixes an issue in https://github.com/babel/babel/pull/11662/files#diff-778d6015479bf103795200e754865e9dL30 where `fn().Foo.self` is memoized twice.

kaicataldo: Do we want to add this in a comment?

PR #11664, opened 2020-06-02
Add support for es6-shims iterables in babel helpers

ms:
In addition to looking for Symbol.iterator, look for custom property `"_es6-shim iterator"`. This property is added by es6-shims to iterables in environments where `Symbol` is not natively supported.

Add a shared helper to extract the iterator method from iterables. Use it where a sync iterator is expected (iterable spreads, for of, etc.).

Note: in `iterableToArray` specifically, the `Array.from` method needs to be equally aware of the custom iterator property. In practice, it needs to be the es6-shim version. Otherwise we return an empty array, which is arguably worse than throwing an exception for unsupported
iterables. This should always be the case unless some other code modified the global Array.from _after_ es6-shims ran.

| Minor: New Feature?      | Yes, support for es6-shims iterables
| Tests Added + Pass?      | No test added, unsure how to test

### DIFF
--- a/packages/babel-helpers/src/helpers.js
+++ b/packages/babel-helpers/src/helpers.js
@@ -82,6 +82,7 @@ helpers.jsx = helper("7.0.0-beta.0")`
 `;
 
 helpers.asyncIterator = helper("7.0.0-beta.0")`
+  import extractIteratorMethod from "extractIteratorMethod";
   export default function _asyncIterator(iterable) {
     var method
     if (typeof Symbol !== "undefined") {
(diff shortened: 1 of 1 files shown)


packages/babel-helpers/src/helpers.js
```diff
@@ -992,9 +991,18 @@ helpers.maybeArrayLike = helper("7.9.0")`
   }
 `;
 
+helpers.extractIteratorMethod = helper("7.0.0-beta.0")`
+  export default function _extractIteratorMethod(iter) {
+    if ("_es6-shim iterator_" in Object(iter)) return iter["_es6-shim iterator_"];
+    if (typeof Symbol !== "undefined" && Symbol.iterator in Object(iter))
+      return iter[Symbol.iterator];
+  }
```
ljharb, line 999: I wrote https://www.npmjs.com/package/es-get-iterator which includes what is actually required to reliably get the iterator method - https://github.com/ljharb/es-get-iterator/blob/master/index.js is surely too large to consider including in babel, but imo that is the correct approach, and it'd be ideal for babel to use it.

If you end up sticking with this simpler approach, be sure to also check for the string "@@iterator" which is used in older Firefoxes.

PR #11669, opened 2020-06-03
Add `delete obj?.#x.a` support

### LINKED ISSUE #11657: Babel should transform `delete obj?.#x.a`
## Bug Report

**Current behavior**

- [REPL](https://babeljs.io/repl#?browsers=chrome%2080&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEDC0G8BQ1oQC5nQS2NAxAB7QC806AFthANwpqY57oCmGAFAJSL2oAmLEC1bQA9gCMAVgH4AdEVkA7UYpaFqrRejqoAvkl1A&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Cstage-3%2Cenv&prettier=false&targets=&version=7.10.2&externalPlugins=)

**Input Code**

```js
delete obj?.#x.a
```

**Expected behavior**
It should be transformed something like

```js
obj == null ? true : (delete _privateGet(obj, Foo, x).a)
```

Note that `true` should be returned instead of `undefined` when `obj` is nullish.

**Environment**
REPL

**Additional Context**
Follow-up to https://github.com/babel/babel/pull/11248/files#r430537770

JLHwung:
| Fixed Issues?            | Fixes #11657 

This PR is a follow-up to #11248, which supports `delete obj?.#x.a` in a similar way of #10806.

This PR includes commits in #11662, I will rebase once that one is merged.

### DIFF
--- a/packages/babel-plugin-proposal-class-properties/test/fixtures/private/optional-chain-delete-property-with-transform/exec.js
+++ b/packages/babel-plugin-proposal-class-properties/test/fixtures/private/optional-chain-delete-property-with-transform/exec.js
@@ -1,0 +1,102 @@
+class Foo {
+  static #x = 1;
+  static #self = Foo;
+  static self = Foo;
+  static getSelf() { return this }
+
+  static test() {
+    const o = { Foo: Foo };
+    const deep = { very: { o } };
+    function fn() {
+      return o;
+    }
+    function fnDeep() {
+      return deep;
+    }
+    expect(() => delete Foo?.#self.unicorn).toThrow(TypeError);
+    expect(() => delete deep?.very.o?.Foo.#self.unicorn).toThrow(TypeError);
+
+    expect(() => delete o?.Foo.#self.unicorn).toThrow(TypeError);
+    expect(() => delete o?.Foo.#self.self.unicorn).toThrow(TypeError);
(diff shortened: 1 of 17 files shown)


packages/babel-plugin-proposal-class-properties/test/fixtures/private/optional-chain-delete-property-with-transform/exec.js
```diff
@@ -0,0 +1,102 @@
...
+  static test() {
+    const o = { Foo: Foo };
+    const deep = { very: { o } };
+    function fn() {
+      return o;
+    }
+    function fnDeep() {
+      return deep;
+    }
+    expect(() => delete Foo?.#self.unicorn).toThrow(TypeError);
```
JLHwung, line 16: Since `unicorn` is non configurable, `delete Foo.#self.unicorn` will throw `TypeError`, so we are sure that the regular path is invoked indeed.

nicolo-ribaudo: clever!

PR #11670, opened 2020-06-04
Update index.js

dueka:
The ":" on line 93 was causing function to throw an error of "Unexpected token, exprected ",". I have edited line 93 to change the ":" into a ",".

| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | `Fixes #1, Fixes #2` 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

PR #11673, opened 2020-06-04
removed type annotation in js file

FBosler:
Small Bug fix with .ts type annotation being used in .js file.


JLHwung:
Duplicate of #11670

PR #11676, opened 2020-06-04
Properly parse `export default from` when `exportDefaultFrom` is not enabled

### LINKED ISSUE #7293: 'export default from' can not be correctly compiled

JLHwung:
| Fixed Issues?            | Fixes #7293, closes #7309

This PR contains two fixes. I include these two because they are highly related and the change should be acceptable to fit in a single PR.

The first fix addresses to `export default from;` when `exportDefaultFrom` is _not_ enabled, which is allowed by the [spec](https://tc39.es/ecma262/#prod-ExportDeclaration). We do a second lookahead after we have seen `export default from` via a new `nextTokenStartSince` method, and test if it starts with `'` or `"` which iff. starts a string literal.

The second fix ban `export default from` when `exportDefaultFrom` is _enabled_, which is forbidden by the [spec draft](https://tc39.es/proposal-export-default-from/#prod-ExportDeclaration). Though I don't find any detailed explainers but I guess it is to address the following ASI hazard.

```js
export default from
"bar";
```


kaicataldo:
LGTM, but looks like Flow is currently failing the build.

packages/babel-parser/test/fixtures/experimental/export-extensions/default-from-identifier-escaped/input.js
```diff
@@ -0,0 +1 @@
+export default \u{66}rom;
```
JLHwung, line 1: 🤦‍♂️IMO it should be addressed in the export-default-from proposal that while `export default from` is banned,  `export default \u{66}rom` is still valid.

packages/babel-parser/src/parser/statement.js
```diff
@@ -1911,6 +1925,17 @@ export default class StatementParser extends ExpressionParser {
       if (isDefault) {
         // Default exports
         this.checkDuplicateExports(node, "default");
+        if (this.hasPlugin("exportDefaultFrom")) {
+          const declaration = ((node: any): N.ExportDefaultDeclaration)
+            .declaration;
+          if (
+            declaration.type === "Identifier" &&
+            declaration.name === "from" &&
+            declaration.end - declaration.start === 4 // does not contain escape
```
JLHwung, line 1934: `export default (from)` is allowed in https://tc39.es/proposal-export-default-from/#prod-ExportDeclaration

nicolo-ribaudo:
Related - https://github.com/tc39/proposal-export-default-from/issues/2

packages/babel-parser/src/parser/statement.js
```diff
@@ -1857,10 +1857,24 @@ export default class StatementParser extends ExpressionParser {
...
+      return true;
+    }
+    // lookahead again when `export default from` is seen
+    if (this.match(tt._default) && hasFrom) {
+      const nextAfterFrom = this.input.charCodeAt(
+        this.nextTokenStartSince(next + 4),
+      );
+      return (
+        nextAfterFrom === charCodes.quotationMark ||
+        nextAfterFrom === charCodes.apostrophe
```
nicolo-ribaudo, line 1874: I don't understand why we need to special-case these two characters and why not, for example `(` or `*`.

packages/babel-parser/src/plugins/placeholders.js
```diff
@@ -251,6 +251,21 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return super.parseExport(node);
     }
 
+    isExportDefaultSpecifier(): boolean {
+      if (this.match(tt._default)) {
+        const next = this.nextTokenStart();
+        if (this.isUnparsedContextual(next, "from")) {
+          if (
+            this.input.substr(this.nextTokenStartSince(next + 4), 2) ===
+            tt.placeholder.label
```
nicolo-ribaudo, line 260:
```suggestion
            
            this.input.startsWith(tt.placeholder.label, this.nextTokenStartSince(next + 4))
```

Since you are not hard-coding `%%` but using `tt.placeholder.label`, let's also not hard-code it's length (`2`)!

packages/babel-parser/src/parser/statement.js
JLHwung: If we see a string literal after from, it is considered a fromClause and thus it is indeed export default fromClause instead of export default AssignmentExpression.

nicolo-ribaudo: I was reading quaotationMark as questionMark 😁

Issue #11474, opened 2020-04-24
any in ClassDeclaration fields in TypeScript type in @babel/types since 7.8.0


grubersjoe:
## Bug Report

- [ ] I would like to work on a fix!

**Current Behavior**

I updated `@babel/types` from `~7.7.0` to the latest version `7.9.5` and noticed that all of a sudden I get new TypeScript errors. The Problem is that beginning with `@babel/types@7.8.0` a lot of fields of the `ClassDeclaration` type use `any` instead of more specific types used before (compare below). Because I use the `noImplicitAny` option of the TypeScript compiler new errors occur. 

```ts
// Current type declaration (^7.8.0)
export interface ClassDeclaration extends BaseNode {
  type: "ClassDeclaration";
  id: any;
  superClass: any;
  body: any;
  decorators: any;
  abstract: boolean | null;
  declare: boolean | null;
  implements: any;
  mixins: any;
  superTypeParameters: any;
  typeParameters: any;
}
```

**Expected behavior/code**

I would expect that an minor upgrade does not break types. I'm not sure if there was a specific reason to set this to any or if it is simply a bug in type generation or somewhere else.

```ts
// Old type declaration (~7.7.0)
export interface ClassDeclaration extends BaseNode {
  type: "ClassDeclaration";
  id: Identifier | null;
  superClass: Expression | null;
  body: ClassBody;
  decorators: Array<Decorator> | null;
  abstract: boolean | null;
  declare: boolean | null;
  implements: Array<TSExpressionWithTypeArguments | ClassImplements> | null;
  mixins: any | null;
  superTypeParameters: TypeParameterInstantiation | TSTypeParameterInstantiation | null;
  typeParameters: TypeParameterDeclaration | TSTypeParameterDeclaration | Noop | null;
}
```

**Environment**

```
System:
    OS: Linux 5.6 Arch Linux
Binaries:
    Node: 12.16.1 - ~/.nvm/versions/node/v12.16.1/bin/node
    Yarn: 1.22.4 - /usr/bin/yarn
    npm: 6.13.4 - ~/.nvm/versions/node/v12.16.1/bin/npm
 npmPackages:
    @babel/types: ~7.8.0 => 7.8.7 
Monorepo: no
How you are using Babel: babel-loader
```

**Possible Solution**

Use the more specific types as before.

Should I provide more information/context?

grubersjoe:
Closing this, since it was fixed in #11628

PR #11687, opened 2020-06-06
improve node type definitions to avoid any's in generated types

zxbodya:
Updates node type definitions in babel-types, to avoid `any` in generated types.

| Fixed Issues?            |  #11474
| Patch: Bug Fix?          | yes(improve generate types)
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-types/src/definitions/core.js
+++ b/packages/babel-types/src/definitions/core.js
@@ -281,10 +281,12 @@ defineType("File", {
     },
     comments: {
       validate: assertEach(assertNodeType("Comment")),
+      optional: true,
     },
     tokens: {
       // todo(ts): add Token type
       validate: assertEach(Object.assign(() => true, { type: "any" })),
+      optional: true,
     },
   },
 });
--- a/packages/babel-types/src/definitions/core.js
+++ b/packages/babel-types/src/definitions/core.js
@@ -84,10 +84,13 @@ defineType("BinaryExpression", {
         const expression = assertNodeType("Expression");
         const inOp = assertNodeType("Expression", "PrivateName");
 
-        return function (node, key, val) {
+        const validator = function (node, key, val) {
           const validator = node.operator === "in" ? inOp : expression;
           validator(node, key, val);
         };
+        // todo(ts): can be discriminated union by `operator` property
+        validator.oneOfNodeTypes = ["Expression", "PrivateName"];
+        return validator;
       })(),
     },
     right: {
@@ -276,6 +279,13 @@ defineType("File", {
     program: {
       validate: assertNodeType("Program"),
     },
+    comments: {
+      validate: assertEach(assertNodeType("Comment")),
+    },
+    tokens: {
+      // todo(ts): add Token type
+      validate: assertEach(Object.assign(() => true, { type: "any" })),
+    },
   },
 });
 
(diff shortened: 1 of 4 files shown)


packages/babel-types/src/definitions/core.js
```diff
@@ -276,6 +279,13 @@ defineType("File", {
     program: {
       validate: assertNodeType("Program"),
     },
+    comments: {
+      validate: assertEach(assertNodeType("Comment")),
+    },
+    tokens: {
+      // todo(ts): add Token type
+      validate: assertEach(Object.assign(() => true, { type: "any" })),
+    },
```
nicolo-ribaudo, line 288: These should both be optional.

packages/babel-types/src/definitions/experimental.js
```diff
@@ -28,8 +29,20 @@ defineType("AwaitExpression", {
...
+          validate: Object.assign(() => true, {
+            oneOfNodeTypes: ["Expression"],
+          }),
+        },
+        callee: {
+          validate: Object.assign(() => true, {
+            oneOfNodeTypes: ["Expression"],
+          }),
+        },
+      }
```
nicolo-ribaudo, line 45: This is a breaking change just because we don't validate those fields in Babel 7.

packages/babel-types/src/definitions/experimental.js
```diff
@@ -226,6 +243,14 @@ defineType("Import", {
...
 defineType("ImportAttribute", {
   visitor: ["key", "value"],
+  fields: {
+    key: {
+      validate: assertNodeType("Identifier"),
+    },
+    value: {
+      validate: assertNodeType("StringLiteral"),
+    },
+  },
```
nicolo-ribaudo, line 253: This is _technically_ a breaking change, but this node is so new that I'd be fine with considering it a bug fix.

packages/babel-types/src/definitions/experimental.js
zxbodya: so, for type definitions it is ok to keep it as `Expression`?

nicolo-ribaudo: Yeah

packages/babel-types/src/definitions/experimental.js
JLHwung: I think generally such changes should be considered a bug fix because the validator should try its best to align with the parser behaviour.

zxbodya:
rebased it on current master(resolving conflicts because of prettier update)

nicolo-ribaudo:
Thanks!

packages/babel-types/src/definitions/core.js
```diff
@@ -276,6 +279,15 @@ defineType("File", {
     program: {
       validate: assertNodeType("Program"),
     },
+    comments: {
+      validate: assertEach(assertNodeType("Comment")),
+      optional: true,
+    },
```
G-Rath, line 285: Just FYI this seems to have broken Jests `toMatchInlineSnapshot` matcher for tests written in TypeScript (https://github.com/facebook/jest/issues/10208)

zxbodya: good catch, this is not intended, effect… probably this should be considered a breaking change and so to go to babel 8

PR #11688, opened 2020-06-07
fix build config to work the same when running on windows

zxbodya:
While debugging failing windows build for #11578 (https://travis-ci.com/github/babel/babel/jobs/345463718) found a problem bundling `babel-standalone`: it was failing when trying to bundle `./lib/config/files/index.js` instead of `./src/config/files/index-browser.ts` from `babel-core`

Fixes for it are in two places:
- `scripts/rollup-plugin-babel-source.js` - `load` there was not handling windows path separator correctly
- `babel.config.js` - filter fix in rollup config, some modules failed to compile because babel config overrides were not applied on windows for the same path separator issue.

However, I still do not understand - how current master version can build correctly on windows… It looks like there is some other place in rollup to use `browser` field, but then - why it would change after converting sources to typescript

| Fixed Issues?            | 
| Patch: Bug Fix?          | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11688"></a>

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,5 +1,11 @@
 "use strict";
 
+const path = require("path");
+
+function normalize(src) {
+  return src.replace(/\//, path.sep);
+}
+
 module.exports = function (api) {
   const env = api.env();
 
@@ -93,7 +99,9 @@ module.exports = function (api) {
       "packages/*/test/fixtures",
       ignoreLib ? "packages/*/lib" : null,
       "packages/babel-standalone/babel.js",
-    ].filter(Boolean),
+    ]
+      .filter(Boolean)
+      .map(normalize),
     presets: [["@babel/env", envOpts]],
     plugins: [
       // TODO: Use @babel/preset-flow when
@@ -115,14 +123,14 @@ module.exports = function (api) {
         test: [
           "packages/babel-parser",
           "packages/babel-helper-validator-identifier",
-        ],
+        ].map(normalize),
         plugins: [
           "babel-plugin-transform-charcodes",
           ["@babel/transform-for-of", { assumeArray: true }],
         ],
       },
       {
-        test: ["./packages/babel-cli", "./packages/babel-core"],
+        test: ["./packages/babel-cli", "./packages/babel-core"].map(normalize),
         plugins: [
           // Explicitly use the lazy version of CommonJS modules.
           convertESM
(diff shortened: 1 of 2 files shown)


babel.config.js
```diff
@@ -115,14 +122,14 @@ module.exports = function(api) {
         test: [
           "packages/babel-parser",
           "packages/babel-helper-validator-identifier",
-        ],
+        ].map(normalize),
```
nicolo-ribaudo, line 125: In Babel 8 I'd like to normalize the paths internally in `@babel/core`

nicolo-ribaudo:
> However, I still do not understand - how current master version can build correctly on windows… It looks like there is some other place in rollup to use browser field, but then - why it would change after converting sources to typescript

I admit that I have no idea :thinking:

nicolo-ribaudo:
https://github.com/zxbodya/babel/blob/997d00ed08f882156bdb0c5217b8f42b2e6babfa/packages/babel-core/package.json#L42

It seems that you didn't change it to `.ts` in your branch. Maybe it's what is causing the problem?

zxbodya:
> https://github.com/zxbodya/babel/blob/997d00ed08f882156bdb0c5217b8f42b2e6babfa/packages/babel-core/package.json#L42
> 
> It seems that you didn't change it to `.ts` in your branch. Maybe it's what is causing the problem?

Good catch, indeed changing it to `.ts` - made it work.

zxbodya:
> Good catch, indeed changing it to `.ts` - made it work.

It looks - `load` in `scripts/rollup-plugin-babel-source.js` can be removed completely… - I tried building on windows with fixes from PR and without them - both results in exactly the same `babel-standalone` file.

Also, apparently if not providing custom `load` - changes in `babel.config.js` are not really needed(probably somewhere in rollup paths are normalized for files it resolves).

@nicolo-ribaudo what you think? should I revert changes `babel.config.js` and instead of fixing - to remove `load` from `scripts/rollup-plugin-babel-source.js` to avoid confusion because of it?

JLHwung:
> However, I still do not understand - how current master version can build correctly on windows… 

Currently our Travis Windows CI will skip standalone build, `@babel/standalone` is only built on Circle CI (Linux).

@zxbodya Can you rebase?

zxbodya:
@JLHwung  rebased it.
however: check my previous comment - might be this change it not really needed

PR #11689, opened 2020-06-07
Instantiate presets before plugins

nicolo-ribaudo:
| Documentation PR Link    | 

This is only needed for https://github.com/babel/babel/pull/12151, but I'm sending them as two separate PR to make it easier to review them one by one (since the diffs overlap, but are conceptually separate).

The e2e failure makes me think that I got plugins ordering wrong, I'll check.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11689"></a>

### DIFF
--- a/packages/babel-core/src/config/full.js
+++ b/packages/babel-core/src/config/full.js
@@ -24,6 +24,7 @@ import {
   validate,
   type CallerMetadata,
   checkNoUnwrappedItemOptionPairs,
+  type PluginItem,
 } from "./validation/options";
 import { validatePluginObject } from "./validation/plugins";
 import makeAPI from "./helpers/config-api";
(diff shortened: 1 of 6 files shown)


nicolo-ribaudo:
Wow, preset ordering is hard. They _sometimes_ run bottom to top, and _sometimes_ top to bottom :upside_down_face: 
In the old code this wasn't explicit, but was handled due to a specific order of push/unshift calls. I have now added an if to switch between push and unshift because I couldn't figure out how to properly reflect the old behavior.

packages/babel-core/test/api.js
```diff
@@ -423,6 +423,8 @@ describe("api", function () {
         "argone;",
         "five;",
         "six;",
+        "twentyone;",
+        "twentytwo;",
```
nicolo-ribaudo, line 427: Before the last commit, this was
```js
twentyone
twentytwo
five
six
```

And it was causing a failure in the CRA e2e test.

packages/babel-core/src/config/full.js
```diff
@@ -199,12 +188,32 @@ export default gensync<[any], ResolvedConfig | null>(function* loadFullConfig(
     .map(plugins => ({ plugins }));
   opts.passPerPreset = opts.presets.length > 0;
 
+  if (context.filename?.includes("RuntimeErrorContaine")) {
+    console.log(initialPluginsDescriptors.map(p => p.key));
```
JLHwung, line 192: 🤔Is that for debug purpose?

JLHwung (approved):
LGTM other than a nit.

packages/babel-core/src/config/full.js
nicolo-ribaudo: Yes 🙃
I'll remove it tomorrow

existentialism (approved):

nicolo-ribaudo:
I'm marking this as "do not merge" until I finish the other PR, so that I'm sure that what I did here is only and everything I need.

nicolo-ribaudo:
Merging into `feat-7.12.0/config-loading-updates`

PR #11696, opened 2020-06-09
Export babel toolings in @babel/core and @babel/standalone

### LINKED ISSUE #11599: Export babel tooling in @babel/standalone and @babel/core
## Feature Request

I am implementing a tool to transform an expression using `parser.parseExpression`, `traverse`, `types.X` and `generate(exp)`. This tool should be directly used on the web however @babel/standalone doesn't export these toolings.

Directly export babel tooling in @babel/standalone(and @babel/core, since @babel/standalone just imports from it).

I tried to use `babel.transform` from @babel/standalone. But an expression will be treated as a statement by babel or trigger a parsing error(like `{ a, b: 3, ...obj }`).

ajihyf:
| Fixed Issues?            | Fixes #11599  
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

I am implementing a tool to transform an expression using `parser.parseExpression`, `traverse`, `types.X` and `generate(exp)`. This tool is used on the web browser however @babel/standalone doesn't export these toolings. Thus I want to directly export babel tooling in @babel/standalone(and @babel/core, since @babel/standalone just imports from it).


NullVoxPopuli:
I ran in to a need for this today :)

liuxingbaoyu:
Thank you for your contribution!
Sorry, due to unknown reasons, CI cannot be passed after rebasing. I will temporarily close this PR, and I may open a new PR based on this later.

liuxingbaoyu (approved):

liuxingbaoyu:
Thanks!

kungfooman:
This is very helpful and allows a better testing of all PRs directly in the console of babel-bot's web preview :rainbow: 

image

nicolo-ribaudo:
I dislike how this can create confusion between `import { parse } from "@babel/core"; parse(...)` and `import { parser } from "@babel/core"; parser.parse(...)`, that do two completely different things.

Given that the need is only in `@babel/standalone`, how do you feel about only exposing them there as `Babel.packages.parser`, `Babel.packages.generator`, etc?

liuxingbaoyu:
I agree with this!
What do you think of `Babel.parser` and `Babel.generator`?

liuxingbaoyu:
> Given that the need is only in @babel/standalone

Does this mean doing this in `@babel/core` as well?
`import { packages } from "@babel/core";` doesn't seem very good.

NullVoxPopuli:
hm, I'm mostly hoping for a way to _not_ using `@babel/standalone`, but would like to easily use `@babel/core` (and friends) to have a minimal browser-compiler with a set few plugins (rather than the _all of them_ that standalone includes)

liuxingbaoyu (approved):
Now this PR is no longer exported from `@babel/core`, but I think that doesn't matter and it can be treated as another issue.

nicolo-ribaudo:
I don't think we should export them from `@babel/core` at all. If you are already using individual unbundled packages, you can just import `@babel/parser`&co directly.

Issue #11469, opened 2020-04-22
Trailing comments of ClassMethod are reported as trailing comments of BlockStatement when the method has no code inside but has a comment


SukantPal:
Starting from this code:

```js
class ExampleAST {
  /** @ctor */
  constructor() {
    /** @member */
    this.member = 'value';
  }

  /* Trailing comment */
}
```

The `/* Trailing comment */` is in the `trailingComments` of the ClassMethod "constructor". However, if you comment out the only line of code in the constructor:

```js
class ExampleAST {
  /** @ctor */
  constructor() {
    /** @member */
    // this.member = 'value';
  }

  /* Trailing comment */
}
```

Then, the trailing comment appears in the trailing comments of the `BlockStatement` inside the constructor.

Furthermore, if you remove both comments from inside the constructor, then the trailing comments work as expected (trailing comments of the ClassMethod, not BlockStatement).

https://astexplorer.net/#/gist/3d8ffdbbe3c3f2929911f50419821e6e/1f6b623281bdc6a0b320aa1e115edcdca483dfe7

PR #11697, opened 2020-06-09
Fix innercomments

shaodahong:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/11469
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |


packages/babel-parser/src/parser/comments.js
```diff
@@ -264,7 +264,17 @@ export default class CommentsParser extends BaseParser {
       ) {
         node.innerComments = trailingComments;
       } else {
-        node.trailingComments = trailingComments;
+        // TrailingComments maybe contain innerComments
+        const findIndex = trailingComments.findIndex(
```
kaicataldo, line 268: Can we name this variable `firstTrailingCommentIndex`? We could probably get ride of the comment then!

packages/babel-parser/src/parser/comments.js
```diff
@@ -264,7 +264,17 @@ export default class CommentsParser extends BaseParser {
...
+          comment => comment.end >= node.end,
+        );
+
+        if (findIndex !== -1 && findIndex !== 0) {
```
kaicataldo, line 272: Looks like we need tests for when:

* Trailing comments _only_ contain inner comments
* When will the second half of this (`findIndex !== 0`) evaluate truthy? Can we add a test for that?

shaodahong: > Trailing comments only contain inner comments

Done.

> When will the second half of this (findIndex !== 0) evaluate truthy? Can we add a test for that?

Have it already, look like

```js
class A {
  test() {}
  // this.member = 'value';
}
```

if `findIndex === 0`mean  both **trailingComments**, so we can skip

packages/babel-parser/test/fixtures/comments/regression/11469/output.json
```diff
@@ -0,0 +1,181 @@
+{
+  "type": "File",
+  "start":0,"end":193,"loc":{"start":{"line":1,"column":0},"end":{"line":18,"column":1}},
+  "errors": [
+    "SyntaxError: Identifier 'A' has already been declared (9:6)",
```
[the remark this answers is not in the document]
shaodahong: Thanks, I fixed it

packages/babel-parser/test/fixtures/comments/regression/11469/output.json
```diff
@@ -0,0 +1,177 @@
...
+      "value": " this.member = 'value';",
+      "start":109,"end":134,"loc":{"start":{"line":11,"column":4},"end":{"line":11,"column":29}}
+    },
+    {
+      "type": "CommentLine",
+      "value": " this.member = 'value';",
+      "start":166,"end":191,"loc":{"start":{"line":17,"column":2},"end":{"line":17,"column":27}}
+    }
+  ]
+}
```
nicolo-ribaudo, line 177: Nit: 
```suggestion
}

```

existentialism:
@shaodahong thanks!

Issue #11646, opened 2020-05-29
Babel should transform class private properties mixed with logical assignment operators


JLHwung:
## Bug Report

This is a good first issue for first-time Babel contributors.

**Current behavior**
A clear and concise description of the behavior.

- [Codesandbox](https://codesandbox.io/s/admiring-shape-r8657)

throws
```
unknown: Property operator expected value to be one of ["+","-","/","%","*","**","&","|",">>",">>>","<<","^","==","===","!=","!==","in","instanceof",">","<",">=","<="] but got "??"
```

**Input Code**

```js
class C {
  #x;
  m() { this.#x ??= 42 }
};
```

**Expected behavior**
It should be transpiled successfully

**Babel Configuration (babel.config.js, .babelrc, package.json#babel, cli command, .eslintrc)**
- Filename: `babel.config.js`

```js
{
  "presets": [
    ["@babel/preset-env", {
      "targets": "Edge 16",
      "shippedProposals": true
    }]
  ],
  "plugins": ["@babel/plugin-proposal-logical-assignment-operators"]
}
```

**Environment**
CodeSandbox

**Possible Solution**

When transforming private class properties, babel will call `packages/babel-helper-member-expression-to-functions/src/index.js` which transforms private property usage to helper calls. For example,
```js
this.#x += 42
```
is approximately transformed to
```
_privateSet(this, x, _privateGet(this, x) + 42)
```

As you can see `_privateGet(this, x) + 42` is a binary expression. But `??` is not a valid binary expression, so the error is thrown. To fix it, we should construct logical expression for logical assignment operators, namely `??=`, `||=` and `&&=`.

If you don't know how to clone Babel, follow these steps: (you need to have `make` and `yarn` available on your machine).
1. Write a comment there to let other possible contributors know that you are working on this bug.
1. Fork the repo
1. Run `git clone https://github.com/<YOUR_USERNAME>/babel.git && cd babel`
1. Run `yarn && make bootstrap`
1. Wait :hourglass_flowing_sand: 
1. Run `make watch` (or `make build` whenever you change a file)
1. Add a test (only `input.js`; `output.js` will be automatically generated) to `packages/babel-plugin-proposal-class-properties/test/fixtures/private/logical-assignment` (tip: you can copy from contents of `fixtures/private/assignment`)
1. Update the code!
1. `yarn jest class-properties` to run the tests
    - If some test outputs don't match but the new results are correct, you can delete the bad `output.js` files and run the tests again
[...]

PR #11702, opened 2020-06-10
add support for logical assignments with private properties

ryzokuken:
| Fixed Issues?            | Fixes https://github.com/babel/babel/issues/11646
| Documentation PR Link    | 

Patches the logic for handling assignment operators and adds support for
handling the logical assignment operators appropriately.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11702"></a>

### DIFF
--- a/packages/babel-helper-member-expression-to-functions/src/index.js
+++ b/packages/babel-helper-member-expression-to-functions/src/index.js
@@ -315,29 +315,42 @@ const handle = {
 
     // MEMBER = VALUE   ->   _set(MEMBER, VALUE)
     // MEMBER += VALUE   ->   _set(MEMBER, _get(MEMBER) + VALUE)
+    // MEMBER ??= VALUE   ->   _get(MEMBER) ?? _set(MEMBER, VALUE)
     if (parentPath.isAssignmentExpression({ left: node })) {
       if (this.simpleSet) {
         member.replaceWith(this.simpleSet(member));
         return;
       }
 
-      const { operator, right } = parent;
-      let value = right;
+      const { operator, right: value } = parent;
 
-      if (operator !== "=") {
-        // Give the state handler a chance to memoise the member, since we'll
-        // reference it twice. The second access (the set) should do the memo
-        // assignment.
-        this.memoise(member, 2);
-
-        value = t.binaryExpression(
-          operator.slice(0, -1),
-          this.get(member),
-          value,
-        );
+      if (operator === "=") {
+        parentPath.replaceWith(this.set(member, value));
+      } else {
+        const operatorTrunc = operator.slice(0, -1);
+        if (t.LOGICAL_OPERATORS.includes(operatorTrunc)) {
+          // Give the state handler a chance to memoise the member, since we'll
+          // reference it twice. The first access (the get) should do the memo
+          // assignment.
+          this.memoise(member, 1);
+          parentPath.replaceWith(
+            t.logicalExpression(
+              operatorTrunc,
+              this.get(member),
+              this.set(member, value),
+            ),
+          );
+        } else {
+          // Here, the second access (the set) is evaluated first.
+          this.memoise(member, 2);
+          parentPath.replaceWith(
+            this.set(
+              member,
+              t.binaryExpression(operatorTrunc, this.get(member), value),
+            ),
+          );
+        }
       }
-
-      parentPath.replaceWith(this.set(member, value));
       return;
     }
 
--- a/packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/logical-assignment/options.json
+++ b/packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/logical-assignment/options.json
@@ -1,0 +1,6 @@
+{
+  "plugins": [
+    ["proposal-logical-assignment-operators", { "loose": true }],
+    ["proposal-class-properties", { "loose": true }]
+  ]
+}
--- a/packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/logical-assignment/output.js
+++ b/packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/logical-assignment/output.js
@@ -1,0 +1,42 @@
+function _classPrivateFieldLooseBase(receiver, privateKey) { if (!Object.prototype.hasOwnProperty.call(receiver, privateKey)) { throw new TypeError("attempted to use private field on non-instance"); } return receiver; }
+
+var id = 0;
+
+function _classPrivateFieldLooseKey(name) { return "__private_" + id++ + "_" + name; }
+
+var _nullish = _classPrivateFieldLooseKey("nullish");
+
+var _and = _classPrivateFieldLooseKey("and");
+
+var _or = _classPrivateFieldLooseKey("or");
+
+class Foo {
+  constructor() {
+    Object.defineProperty(this, _nullish, {
+      writable: true,
+      value: 0
+    });
+    Object.defineProperty(this, _and, {
+      writable: true,
+      value: 0
+    });
+    Object.defineProperty(this, _or, {
+      writable: true,
+      value: 0
+    });
+  }
+
+  self() {
+    return this;
+  }
+
+  test() {
+    var _classPrivateFieldLoo, _classPrivateFieldLoo2, _classPrivateFieldLoo3, _classPrivateFieldLoo4;
+
+    (_classPrivateFieldLoo = _classPrivateFieldLooseBase(this, _nullish))[_nullish] ?? (_classPrivateFieldLoo[_nullish] = 42);
+    (_classPrivateFieldLoo2 = _classPrivateFieldLooseBase(this, _and))[_and] && (_classPrivateFieldLoo2[_and] = 0);
+    (_classPrivateFieldLoo3 = _classPrivateFieldLooseBase(this, _or))[_or] || (_classPrivateFieldLoo3[_or] = 0);
+    (_classPrivateFieldLoo4 = _classPrivateFieldLooseBase(this.self(), _nullish))[_nullish] ?? (_classPrivateFieldLoo4[_nullish] = 42);
+  }
+
+}
--- a/packages/babel-plugin-proposal-class-properties/test/fixtures/private/logical-assignment/input.js
+++ b/packages/babel-plugin-proposal-class-properties/test/fixtures/private/logical-assignment/input.js
@@ -1,0 +1,17 @@
+class Foo {
+  #nullish = 0;
+  #and = 0;
+  #or = 0;
+
+  self() {
+    return this;
+  }
+
+  test() {
+    this.#nullish ??= 42;
+    this.#and &&= 0;
+    this.#or ||= 0;
+
+    this.self().#nullish ??= 42;
+  }
+}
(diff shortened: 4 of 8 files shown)


packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -306,11 +306,21 @@ const handle = {
         // assignment.
         this.memoise(member, 2);
 
-        value = t.binaryExpression(
-          operator.slice(0, -1),
-          this.get(member),
-          value,
-        );
+        const operatorTrunc = operator.slice(0, -1);
+        switch (operatorTrunc) {
```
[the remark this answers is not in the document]
ryzokuken: Sorry, I'm new to the codebase, didn't know these constants existed. Will fix.

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -306,11 +306,21 @@ const handle = {
...
+          case "??":
+          case "||":
+          case "&&":
+            value = t.logicalExpression(operatorTrunc, this.get(member), value);
+            break;
+          default:
+            value = t.binaryExpression(
+              operator.slice(0, -1),
```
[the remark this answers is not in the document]
ryzokuken: Oops, I must've left that in while debugging.

ryzokuken:
> We already support private properties and logical assignment, just not together. Would it be ok for you to mark this as a "Bug fix" (since it's just an integration problem) and merge it in a patch release?

Right, I was unsure, thanks for letting me know.

ryzokuken:
Thanks @nicolo-ribaudo @JLHwung and others, I fixed the code and the PR body. Hope this is better.

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -306,11 +306,12 @@ const handle = {
         // assignment.
         this.memoise(member, 2);
 
-        value = t.binaryExpression(
-          operator.slice(0, -1),
-          this.get(member),
-          value,
-        );
+        const operatorTrunc = operator.slice(0, -1);
+        if (t.LOGICAL_OPERATORS.includes(operatorTrunc)) {
```
jridgewell, line 310: Ooo, can we fix `https://github.com/babel/babel/blob/master/packages/babel-plugin-proposal-logical-assignment-operators/src/index.js#L16`, too?

packages/babel-plugin-proposal-class-properties/test/fixtures/private/logical-assignment/output.js
```diff
@@ -0,0 +1,37 @@
...
+    });
+
+    _or.set(this, {
+      writable: true,
+      value: 0
+    });
+  }
+
+  test() {
+    _classPrivateFieldSet(this, _nullish, _classPrivateFieldGet(this, _nullish) ?? 42);
```
jridgewell, line 30: This is a fine first start, but it's [technically incorrect](https://babeljs.io/repl/build/23793/#?browsers=ie%2011&build=&builtIns=false&spec=false&loose=true&code_lz=MYGwhgzhAEBiD29oG8BQ1oHMCmAXaAxAEZgBOAFAJQrobSl4CupAdtAAwDctAvrRHkIkKAN2po60XAAtS8AO7QW2RQFFScigAMI0-IxAATFgHJ8RbNGBgQIbIa2VuGPrVzYIuKjUkyAlhAAdMRk0AD8YQC80ACMztB8rsqKCPBUge6eVJxAA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Cstage-2%2Cenv&prettier=false&targets=&version=7.10.2%2Bpr.11702&externalPlugins=).

We need to perform the logical operator first, then perform the assignment in the RHS of the logical. We should test with a deep property (`this.self.#bar ??= 1`), to ensure the memoization works correctly.

packages/babel-helper-member-expression-to-functions/src/index.js
ryzokuken: Will do.

packages/babel-plugin-proposal-class-properties/test/fixtures/private/logical-assignment/output.js
ryzokuken: I see your point. Right now I'm clubbing these alongside `MEMBER += VALUE   ->   _set(MEMBER, _get(MEMBER) + VALUE)`. Do you propose that for logical assignment we have a different transformation `MEMBER ??= VALUE   ->   _get(MEMBER) ?? _set(MEMBER,  VALUE)`?

ryzokuken: I checkout out the proposal explainer (https://github.com/tc39/proposal-logical-assignment), and that's exactly what the proposal says!

jridgewell: Yup, exactly.

ryzokuken: @jridgewell gimme a quick second.

ryzokuken: @jridgewell that definitely took more than a second, but it should be better now, PTAL.

packages/babel-plugin-proposal-class-properties/test/fixtures/private/logical-assignment/output.js
```diff
@@ -0,0 +1,35 @@
...
+    });
+
+    _or.set(this, {
+      writable: true,
+      value: 0
+    });
+  }
+
+  test() {
+    _classPrivateFieldGet(this, _nullish) ?? _classPrivateFieldSet(this, _nullish, 42);
```
jridgewell, line 30: We still need to add a test case for deep property access (`this.self().#foo ??= bar`). [repl](https://babeljs.io/repl/build/23799/#?browsers=chrome%2080&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEBiD29oG8BQ1oGIB2BXEIAlhABbQC80ADANzrT0QCmIAZgBQCUK9GATkwAuuPtmiCSxOhgC-qeoKYRBXHhgwTiAOmZsuWnPiKloAflOUALACZp0OTKA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=stage-3&prettier=false&targets=&version=7.10.2%2Bpr.11702&externalPlugins=) That shows us we still have one last bug, because the memoization isn't being done on the correct `this.self()` access.

packages/babel-plugin-proposal-class-properties/test/fixtures/private/logical-assignment/input.js
```diff
@@ -0,0 +1,11 @@
+class Foo {
```
JLHwung, line 1: Please also duplicate the test to `test/fixtures/private-loose`.

JLHwung:
Friendly ping @ryzokuken :)

ryzokuken:
Sorry @JLHwung, I'll finish it up this weekend.

ryzokuken:
@jridgewell I pushed the test you asked for, but I still don't quite understand how to fix the memoization issue. Could you please elaborate what needs to be done in that regard?

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/logical-assignment/output.js
```diff
@@ -26,10 +26,17 @@ class Foo {
...
+    return this;
+  }
+
   test() {
+    var _this$self;
+
     _classPrivateFieldGet(this, _nullish) ?? _classPrivateFieldSet(this, _nullish, 42);
     _classPrivateFieldGet(this, _and) && _classPrivateFieldSet(this, _and, 0);
     _classPrivateFieldGet(this, _or) || _classPrivateFieldSet(this, _or, 0);
+    _classPrivateFieldGet(_this$self, _nullish) ?? _classPrivateFieldSet(_this$self = this.self(), _nullish, 42);
```
nicolo-ribaudo, line 39: `_this$self` should be defined in the get call here.

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/logical-assignment/options.json
```diff
@@ -0,0 +1,6 @@
+{
+  "plugins": [
+    "proposal-logical-assignment-operators",
```
JLHwung, line 3: Can you add `loose: true` to both plugins?
```json
{
  "plugins": [
    ["proposal-logical-assignment-operators", { "loose": true }],
    ["proposal-class-properties", { "loose": true }]
  ]
}
```
The `private-loose` folder is test cases about `loose` options so we are sure that both `loose` and non-loose (default) are working.

JLHwung:
@ryzokuken The memoization is done in https://github.com/babel/babel/blob/1dd94e813ed8161ed0efc7e99fd3cd4277c6ab02/packages/babel-helper-member-expression-to-functions/src/index.js#L285-L288

The second argument `2` means _when_ we should memoize the member call. i.e. `obj.self().#a += 1` is roughly transformed to

```js
// obj.self().#a = obj.self().#a + 1
classFieldSet($objSelf = obj.self(), a, classFieldGet($objSelf, a) + 1)
```

via
```js
this.set(member, t.binaryExpression("+", this.get(member), value))
```

We should memoize `obj.self()` in `classFieldSet` because it is the first argument of `classFieldSet` and thus evaluated before `classFieldGet`. However since in Babel we have called `this.get` and `this.set` sequentially, `2` is a counter to indicate that we should memoize the member in the _second_ call of `[this.get, this.set]` sequence. (Think about TTL in ICMP protocol)

However `obj.self().#a ??= 1` is different here, it should be transformed as
```js
// obj.self().#a ?? (obj.self().#a = 1)
classFieldGet($objSelf = obj.self(), a) ?? classFieldSet($objSelf, a, 1)

// current PR
classFieldGet($objSelf, a) ?? classFieldSet($objSelf = obj.self(), a, 1)
```

via
```js
t.LogicalExpression("??", this.get(member), this.set(member))
```

Note that the call sequence is unchanged -- `this.get, this.set`. But we should memoize in `classFieldGet` because it is now evaluated _before_ `classFieldSet`. To achieve that we can decrement the memoize counter for logical expressions.

```js
// add some comment about why 1 is chosen for logical expressions
this.memoise(member, 1);
```

jridgewell:
Thank you for explaining @JLHwung, that's extremely thorough!

JLHwung:
@ryzokuken 🏓️ Logical Assignment is seeking advancement in July meeting, will be great if we can ship this bug fix after that.

ryzokuken:
@JLHwung my bad, I'll get that done.

ryzokuken:
Update: I've made the change, I'll push this in a bit, sorry for the delay. I got super busy with all sorts of weird things.

ryzokuken:
@JLHwung does this look good enough? Let me know if it does and I'll autosquash and force push.

JLHwung:
@ryzokuken Can you update the test case (config/fixture) w.r.t. 
https://github.com/babel/babel/pull/11702#pullrequestreview-429232140
and
https://github.com/babel/babel/pull/11702#discussion_r447211024 ?

packages/babel-plugin-proposal-class-properties/test/fixtures/private-loose/logical-assignment/output.js
```diff
@@ -0,0 +1,42 @@
...
+    return this;
+  }
+
+  test() {
+    var _this$self;
+
+    _classPrivateFieldGet(this, _nullish) ?? _classPrivateFieldSet(this, _nullish, 42);
+    _classPrivateFieldGet(this, _and) && _classPrivateFieldSet(this, _and, 0);
+    _classPrivateFieldGet(this, _or) || _classPrivateFieldSet(this, _or, 0);
+    _classPrivateFieldGet(_this$self, _nullish) ?? _classPrivateFieldSet(_this$self = this.self(), _nullish, 42);
```
jridgewell, line 39: This test still needs to be updated. I believe the transform code is correct, the test is just out of date now.

ryzokuken:
@jridgewell @JLHwung the tests should be updated now. Sorry for the delay again, I noticed the stage advancement yesterday.

JLHwung:
The e2e was fixed on upstream main. Can you rebase?

ryzokuken:
Thanks for all the help on this, everyone!

PR #11703, opened 2020-06-10
fix:added check for forXstatement pattern

### LINKED ISSUE #11272: Private Name in the left ForOfStatement is not transformed
## Bug Report

**Current Behavior**
The following snippet, when transformed via preset-env and stage-3 preset, will throw
```
Property left of ForOfStatement expected node to be of a type ["VariableDeclaration","LVal"] but instead got "CallExpression"
```

**Input Code**
- [REPL](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEDC0G8BQ1oGIAOBuF0C2AFAJSK6oBmA9gE7QEAuAFgJYQB0m0l50A2gLpEcqAL5IRQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Cstage-3%2Cenv&prettier=false&targets=&version=7.8.7&externalPlugins=)

```js
class C {
  #p;
  m() {
    for (this.#p of []);
  }
}
```

**Expected behavior/code**
It should be transformed into something like
```js
// helpers is omissioned
var C = /*#__PURE__*/function () {
  function C() {
    _classCallCheck(this, C);

    _p.set(this, {
      writable: true,
      value: void 0
    });
  }

  _createClass(C, [{
    key: "m",
    value: function m() {
      for (var _i = 0, _arr = []; _i < _arr.length; _i++) {
        i = _arr[_i];
        _classPrivateFieldSet(this, _p, i);
      };
    }
  }]);

  return C;
}();
```

**Possible Solution**

As Justin noted, we can fix it by detecting such patterns in `@babel/helper-member-expression-to-functions`, so babel will apply `destructureSet` private name handler, which will return a member expression -- a valid LHS.

https://github.com/babel/babel/blob/1ba41f208430bdf382b75b56e9aad7436ca522ad/packages/babel-helper-member-expression-to-functions/src/index.js#L110

1. Write a comment there to let other possible contributors know that you are working on this bug.
1. Fork the repo
1. Run `make watch`
1. Add a test (only `input.js`; `output.js` will be automatically generated)
1. Update the code!
1. `yarn jest babel-plugin-proposal-class-properties` to run the tests

This issue is likely due to plugin ordering: the `private-class-properties` runs before `for-of-transforms` after `this.#p` is transformed to helper calls which is not a valid LHS, the for-of transforms throws on unsupported for-of left.

wlawt:
| Fixed Issues?            |  Fixes #11272
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Issue Ref: [Private Name in the left ForOfStatement is not transformed, #11272 ](https://github.com/babel/babel/issues/11272)

The problem was that the private label `#a` runs before the `for-of` transform, and that has been fixed by adding a pattern check at the start of the `if` statement. Test cases included the original one from #11272 and the following:

**Input**
```
class Rainbow {
  tColors = ['Red', 'Orange', 'Yellow', 'Green', 'Blue', 'Indigo', 'Violet'];
  #color;

  PrintColors() {
    for (this.#color of this.tColors) {
      console.log(this.#color)
    }
  }
}
```

**Output**
```
var _color = new WeakMap();

var Rainbow = /*#__PURE__*/function () {
  "use strict";

  function Rainbow() {
    babelHelpers.classCallCheck(this, Rainbow);
    babelHelpers.defineProperty(this, "tColors", ['Red', 'Orange', 'Yellow', 'Green', 'Blue', 'Indigo', 'Violet']);

    _color.set(this, {
      writable: true,
      value: void 0
    });
  }

  babelHelpers.createClass(Rainbow, [{
    key: "PrintColors",
    value: function PrintColors() {
      for (babelHelpers.classPrivateFieldDestructureSet(this, _color).value of this.tColors) {
        console.log(babelHelpers.classPrivateFieldGet(this, _color));
      }
    }
  }]);
  return Rainbow;
}();
```

### DIFF
--- a/packages/babel-helper-member-expression-to-functions/src/index.js
+++ b/packages/babel-helper-member-expression-to-functions/src/index.js
@@ -329,7 +329,8 @@ const handle = {
       return;
     }
 
-    // for ( MEMBER of ARR) && for ( MEMBER in ARR)
+    // for (MEMBER of ARR)
+    // for (MEMBER in ARR)
     // { KEY: MEMBER } = OBJ -> { KEY: _destructureSet(MEMBER) } = OBJ
     // { KEY: MEMBER = _VALUE } = OBJ -> { KEY: _destructureSet(MEMBER) = _VALUE } = OBJ
     // {...MEMBER} -> {..._destructureSet(MEMBER)}
--- a/packages/babel-helper-member-expression-to-functions/src/index.js
+++ b/packages/babel-helper-member-expression-to-functions/src/index.js
@@ -337,7 +337,8 @@ const handle = {
     // [MEMBER = _VALUE] = ARR -> [_destructureSet(MEMBER) = _VALUE] = ARR
     // [...MEMBER] -> [..._destructureSet(MEMBER)]
     if (
-      parentPath.isForXStatement() ||
+      (parentPath.isForXStatement() &&
+        parentPath.parentPath.isAssignmentPattern({ left: node })) ||
       // { KEY: MEMBER } = OBJ
       (parentPath.isObjectProperty({ value: node }) &&
         parentPath.parentPath.isObjectPattern()) ||
--- a/packages/babel-helper-member-expression-to-functions/src/index.js
+++ b/packages/babel-helper-member-expression-to-functions/src/index.js
@@ -329,6 +329,7 @@ const handle = {
       return;
     }
 
+    // for ( MEMBER of ARR) && for ( MEMBER in ARR)
     // { KEY: MEMBER } = OBJ -> { KEY: _destructureSet(MEMBER) } = OBJ
     // { KEY: MEMBER = _VALUE } = OBJ -> { KEY: _destructureSet(MEMBER) = _VALUE } = OBJ
     // {...MEMBER} -> {..._destructureSet(MEMBER)}
@@ -337,6 +338,7 @@ const handle = {
     // [MEMBER = _VALUE] = ARR -> [_destructureSet(MEMBER) = _VALUE] = ARR
     // [...MEMBER] -> [..._destructureSet(MEMBER)]
     if (
+      // for ( MEMBER of ARR) && for ( MEMBER in ARR)
       parentPath.isForXStatement({ left: node }) ||
       // { KEY: MEMBER } = OBJ
       (parentPath.isObjectProperty({ value: node }) &&
--- a/packages/babel-helper-member-expression-to-functions/src/index.js
+++ b/packages/babel-helper-member-expression-to-functions/src/index.js
@@ -337,8 +337,7 @@ const handle = {
     // [MEMBER = _VALUE] = ARR -> [_destructureSet(MEMBER) = _VALUE] = ARR
     // [...MEMBER] -> [..._destructureSet(MEMBER)]
     if (
-      (parentPath.isForXStatement() &&
-        parentPath.parentPath.isAssignmentPattern({ left: node })) ||
+      parentPath.isForXStatement({ left: node }) ||
       // { KEY: MEMBER } = OBJ
       (parentPath.isObjectProperty({ value: node }) &&
         parentPath.parentPath.isObjectPattern()) ||
--- a/packages/babel-helper-member-expression-to-functions/src/index.js
+++ b/packages/babel-helper-member-expression-to-functions/src/index.js
@@ -339,7 +339,8 @@ const handle = {
     // [MEMBER = _VALUE] = ARR -> [_destructureSet(MEMBER) = _VALUE] = ARR
     // [...MEMBER] -> [..._destructureSet(MEMBER)]
     if (
-      // for ( MEMBER of ARR) && for ( MEMBER in ARR)
+      // for (MEMBER of ARR)
+      // for (MEMBER in ARR)
       parentPath.isForXStatement({ left: node }) ||
       // { KEY: MEMBER } = OBJ
       (parentPath.isObjectProperty({ value: node }) &&
--- a/packages/babel-helper-member-expression-to-functions/src/index.js
+++ b/packages/babel-helper-member-expression-to-functions/src/index.js
@@ -337,6 +337,7 @@ const handle = {
     // [MEMBER = _VALUE] = ARR -> [_destructureSet(MEMBER) = _VALUE] = ARR
     // [...MEMBER] -> [..._destructureSet(MEMBER)]
     if (
+      parentPath.isForXStatement() ||
       // { KEY: MEMBER } = OBJ
       (parentPath.isObjectProperty({ value: node }) &&
         parentPath.parentPath.isObjectPattern()) ||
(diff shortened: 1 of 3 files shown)


packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -337,6 +337,8 @@ const handle = {
     // [MEMBER = _VALUE] = ARR -> [_destructureSet(MEMBER) = _VALUE] = ARR
     // [...MEMBER] -> [..._destructureSet(MEMBER)]
     if (
+      (parentPath.isForXStatement() &&
+        parentPath.parentPath.isAssignmentPattern({ left: node })) ||
```
JLHwung, line 341: If a `path` is a `ForXStatement`, how can its `parentPath` be an `AssignmentPattern`? Note that the children of an `Pattern` is always an expression. A statement can not be an expression.

Here is an AST example of `for (o.p of arr);`
```json
// for (o.p of arr);
{
  "type": "ForOfStatement",
  "await": false,
  "left": {
    "type": "MemberExpression",
    "object": { "type": "Identifier", "name": "o" },
    "property": { "type": "Identifier", "name": "p" }
   },
  "right": { "type": "Identifier", "name": "arr" }
}
```

The `parentPath` here refers to the `NodePath` of `ForOfStatement`. `NodePath` is a wrapper of the AST node that provides `parent` association with other `NodePath`. In this PR We would like to make sure that `member`, which refers to `MemberExpression` here, is indeed the `left` of the underlying AST node of `parentPath`.

So we can use
```js
parentPath.isForXStatement() && parentPath.node.left === node
```

The query function `isForXStatement` accepts an additional parameter as query, so the code above can be simplified as

```js
parentPath.isForXStatement({ left: node })
```

wlawt: ok, when I run the test I seem to get the same results, which shouldn't happen?

**input**
```
class D {
  #arr;
  f() {
    for (const el of this.#arr);
  }
}
```

**output**
```
var _arr = new WeakMap();

var D = /*#__PURE__*/function () {
  "use strict";

  function D() {
    babelHelpers.classCallCheck(this, D);

    _arr.set(this, {
      writable: true,
      value: void 0
    });
  }

  babelHelpers.createClass(D, [{
    key: "f",
    value: function f() {
      for (var el of babelHelpers.classPrivateFieldGet(this, _arr)) {
        ;
      }
    }
  }]);
  return D;
}();
```

JLHwung: The output is expected. Can you add a test case from the original issue?
```js
class D {
  #arr;
  f() {
    for (const this.#arr of [1, 2]);
  }
}
```

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -337,6 +337,7 @@ const handle = {
     // [MEMBER = _VALUE] = ARR -> [_destructureSet(MEMBER) = _VALUE] = ARR
     // [...MEMBER] -> [..._destructureSet(MEMBER)]
     if (
```
jridgewell, line 339: Nit: Please add a comment for both for-of and for-in, like the above.

packages/babel-plugin-proposal-class-properties/test/fixtures/private/1-helpermemberexpressionfunction/input.js
```diff
@@ -0,0 +1,6 @@
+class D {
+  #arr;
+  f() {
+    for (const el of this.#arr);
```
jridgewell, line 4: These should be reversed.

Please also add a test for for-in.

JLHwung: It is a regression check on the first draft implementation: https://github.com/babel/babel/pull/11703/files/d4c0a5e8b75306f918cf2a2307e4f65130f9ea05#diff-decaf745088978f783d453709a8a8479

jridgewell: Ooo, ok. So we need 2 new test cases (for-of and for-in) with it appearing in the `left`.

hzoo: 👍 can just do it in the next line, or leave a comment that it shouldn't be transformed

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -329,6 +329,7 @@ const handle = {
       return;
     }
 
+    // for ( MEMBER of ARR) && for ( MEMBER in ARR)
```
JLHwung, line 332:
```suggestion
    // for (MEMBER of ARR)
    // for (MEMBER in ARR)
```
nit: `&&` could cause confusion since it is a valid JavaScript operator.

packages/babel-helper-member-expression-to-functions/src/index.js
```diff
@@ -337,6 +338,8 @@ const handle = {
     // [MEMBER = _VALUE] = ARR -> [_destructureSet(MEMBER) = _VALUE] = ARR
     // [...MEMBER] -> [..._destructureSet(MEMBER)]
     if (
+      // for ( MEMBER of ARR) && for ( MEMBER in ARR)
```
JLHwung, line 341:
```suggestion
      // for (MEMBER of ARR)
      // for (MEMBER in ARR)
```

kaicataldo:
Looks like there are failing tests - do you mind fixing those up?

wlawt:
> Looks like there are failing tests - do you mind fixing those up?

@existentialism said it had to do with an upstream dep breaking the tests, not sure how I'd go about those

JLHwung:
CI failure is unrelated and will be fixed by https://github.com/babel/babel/pull/11709.

packages/babel-plugin-proposal-class-properties/test/fixtures/private/1-helpermemberexpressionfunction/input.js
hzoo: personally I think just doing them all in the same class is fine

```js
for (const el of this.#arr);
for (this.#arr of [1, 2]);
for (this.#arr in [1,2,3]);
```

the empty arr seems to be the same case?

and I would rename `1-helpermemberexpressionfunction` to something like for-of-member-expression or something

existentialism:
@wlawt great work! thank you!

packages/babel-plugin-proposal-class-properties/test/fixtures/private/1-helpermemberexpressionfunction/input.js
wlawt: for sure, will remember for next time!

PR #11722, opened 2020-06-16
Add better parser error when using jsx

### LINKED ISSUE #11499: Better parser error when using jsx

penguingovernor:
| Q                       | A                                                               |
| ----------------------- | --------------------------------------------------------------- |
| Fixed Issues?           | Fixes #11499                                                    |
| Patch: Bug Fix?         | Yes                                                             |
| Major: Breaking Change? | No                                                              |
| Minor: New Feature?     | No                                                              |
| Tests Added + Pass?     | Yes                                                             |
| Documentation PR Link   |  |
| Any Dependency Changes? | No                                                              |
| License                 | MIT                                                             |

Added tests to `babel-parser` to check when the `<` symbol is used in the following circumstances:

1. No jsx plugin is loaded
2. When valid JS code is used (ie `"use strict" < div`)
3. When type parameters are used.

This PR also address issue #11499 and adds a `tt.relational` test case in [`parseExprAtom`](https://github.com/babel/babel/blob/bb180eba9f3a2d0dc69e62beb788bfb8a9215a47/packages/babel-parser/src/parser/expression.js#L877).
If that test case is triggered then a `expectOnePlugin(["jsx", "flow", "typescript"])` error is thrown.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11722"></a>


packages/babel-parser/test/fixtures/jsx/errors/_no_plugin/input.js
```diff
@@ -0,0 +1 @@
+<div></div>
```
JLHwung, line 1: Can you also add the following test case?
```jsx
<div>{name}</div>
```

penguingovernor: Done and done (:

packages/babel-parser/test/fixtures/jsx/errors/_no-plugin-ts-type-param/input.js
```diff
@@ -0,0 +1 @@
+<div>() => {}
```
JLHwung, line 1: This case is parsed as flow type annotation. Can you add an extra test but `flow` is _not_ enabled?

penguingovernor: Can do :)

packages/babel-parser/test/fixtures/jsx/errors/_no-plugin-type-param/options.json
```diff
@@ -0,0 +1,4 @@
+{
+  "plugins": [],
```
JLHwung, line 2: Can you add `jsx` here? I think it will still throw syntax error.

penguingovernor: It does throw that error 

```
SyntaxError: Unexpected token (1:13)
    > 1 | <div>() => {}
        |              ^
```

What would the correct error message (if any) be in that case?

nicolo-ribaudo: If we can throw something like "expected closing JSX element" it's better, otherwise `Unexpected token` is ok for now.

penguingovernor: Sounds good to me, pushing that final test

existentialism:
Should we retarget this onto `main`?

PR #11724, opened 2020-06-16
fix: do-statementlist behavior

### LINKED ISSUE #9404: @babel/plugin-proposal-do-expressions: StatementList behavior
## Bug Report

**Current Behavior**
Compilation result:
```js
x = 1;
x = (1, void 0);
```

**Input Code**
A module with the following contents
```js
x = do { 1; };
x = do { 1;;;; };
```

**Expected behavior/code**
Expected compilation result:
```js
x = 1;
x = 1;
```
This would match what what the ECMAScript spec currently says:
>The value of a StatementList is the value of the last value-producing item in the StatementList. For example, the following calls to the eval function all return the value 1:
>
> eval("1;;;;;")
> eval("1;{}")
> eval("1;var a;")

Chrome and Node repls both comply with the spec's assertion above.

I'm not sure this plugin attempts to evaluate StatementLists the same way. Its description is
> the final statement completion value inside the block becomes the completion value of the do expression.

"final statement completion value" could mean either the last value-producing statement (as above), or could mean the syntactically latest-appearing in the block. I'm not sure which is intended.

I'm not even sure the do-expression spec is precise enough right now to justify my expectation here. It suggests:
```
(do { <stmt> };) equivalent to { <stmt> }
```
but that doesn't parse:
```
$ npx babel --plugins @babel/plugin-proposal-do-expressions index.js
{ SyntaxError: /Users/swilson/code/playground/index.js: Unexpected token, expected "," (2:17)

  1 | x = do { 1; };
> 2 | x = (do { 1;;;; };);
    |                  ^
  3 |
  4 |
    at Parser.raise (/Users/swilson/code/playground/node_modules/@babel/parser/lib/index.js:3834:17)
    at Parser.unexpected (/Users/swilson/code/playground/node_modules/@babel/parser/lib/index.js:5142:16)
    at Parser.expect (/Users/swilson/code/playground/node_modules/@babel/parser/lib/index.js:5134:28)
    at Parser.parseParenAndDistinguishExpression (/Users/swilson/code/playground/node_modules/@babel/parser/lib/index.js:6413:14)
    at Parser.parseExprAtom (/Users/swilson/code/playground/node_modules/@babel/parser/lib/index.js:6211:21)
    at Parser.parseExprSubscripts (/Users/swilson/code/playground/node_modules/@babel/parser/lib/index.js:5848:23)
    at Parser.parseMaybeUnary (/Users/swilson/code/playground/node_modules/@babel/parser/lib/index.js:5828:21)
    at Parser.parseExprOps (/Users/swilson/code/playground/node_modules/@babel/parser/lib/index.js:5717:23)
    at Parser.parseMaybeConditional (/Users/swilson/code/playground/node_modules/@babel/parser/lib/index.js:5690:23)
    at Parser.parseMaybeAssign (/Users/swilson/code/playground/node_modules/@babel/parser/lib/index.js:5635:21)
  pos: 32,
  loc: Position { line: 2, column: 17 },
  code: 'BABEL_PARSE_ERROR' }
```

Thanks in advance!

package.json dependencies:
```json
{
  "@babel/cli": "^7.2.3",
  "@babel/core": "^7.2.2",
  "@babel/plugin-proposal-do-expressions": "^7.2.0"
}
```
CLI command:
```sh
npx babel --plugins @babel/plugin-proposal-do-expressions index.js
```

**Environment**
- Babel version(s): both what's in the repl, and the above version info
- Node/npm version: 11.7.0/6.5.0
- OS: macOS 10.13.6
- Monorepo: No

wlawt:
| Fixed Issues?            | Fixes #9404  
| Documentation PR Link    | 

Added check for multi-use of semicolons where it was returning `void 0` before. 

**Input** 
```
expect(do {
  x = do { 1; };
  z = do { 1;;;; };
  w = (do { 1;;;; });
})
```

**Output**
```
x=1
z=1
w=1
```

The parsing issue addressed in #9404 seems to be correct and ok (reference: [9404 comment](https://github.com/babel/babel/issues/9404#issuecomment-644746463))


packages/babel-types/src/converters/gatherSequenceExpressions.js
```diff
@@ -67,13 +68,16 @@ export default function gatherSequenceExpressions(
     } else if (isEmptyStatement(node)) {
       // empty statement so ensure the last item is undefined if we're last
       ensureLastUndefined = true;
```
JLHwung, line 70: If _all_ of `nodes` are empty statements, it should `ensureLastUndefined`. i.e. both `eval("{ ; }")` and `eval("{ ;;; }")` returns `undefined`. Note that you can mentally think of `do { ; }` as `eval("{ ; }")`.

However, when empty statements follow any value producing statements, it should not `ensureLastUndefined`. i.e. `eval("{ 3;; }")` returns `3`.

Therefore we should check 

- If the empty statement is the _first_ of `nodes`, we should `ensureLastUndefined`, otherwise it is a no-op because it can always be omitted. Conceptually it is more like `ensureFirstUndefined`.

I don't think need to track `multiColons` because
- the specific value of multiColons is _not_ used as a number later
- Transformers operates on AST level (`EmptyStatement` v.s. `colon`) while colon is a grammar structure and should be addressed by tokenizer instead

packages/babel-plugin-proposal-do-expressions/test/fixtures/do-expressions/semicolons.js
```diff
@@ -0,0 +1,5 @@
+expect(do {
```
JLHwung, line 1: This is not a proper `expect` usage because it does not have any matchers. See https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/valid-expect.md

I guess you intended to
```js
expect(do { 1;; }).toBe(1);
expect(do { ;; }).toBe(undefined);
// ...
```

packages/babel-types/src/converters/gatherSequenceExpressions.js
wlawt: yup that makes sense, just have a question: if accessing the `path` isn't available to check if it's `inList` or `key` is there a way to do it with the array of `nodes` being passed to the function? 

I've also explored the `scope` that gets passed and trying to use it's `path` but `inList` and `key` are `undefined` for it.

packages/babel-types/src/converters/gatherSequenceExpressions.js
```diff
@@ -66,7 +66,12 @@ export default function gatherSequenceExpressions(
       exprs.push(body);
     } else if (isEmptyStatement(node)) {
       // empty statement so ensure the last item is undefined if we're last
-      ensureLastUndefined = true;
+      // checks if node is first
+      // checks if input is `;` or `;;` which will return `()` in where
+      // the length will always be 2
+      if (nodes.indexOf(node) === 0 || nodes.length === 2) {
```
JLHwung, line 72: I don't think we need to check specifically `nodes.length === 2`? i.e. `do { ;;; }` should also return `undefined`.

wlawt: yeah i added the `node.length` because `()` always threw an error regardless of the amount of `;` given. 

just to clarify what does no-op mean ? 

and do we need to explicitly mark it `false` since we're in the condition that it is `true`?

JLHwung: > () always threw an error regardless of the amount of ; given.

It is syntax error if there is only empty statement in the `()`, i.e. `()`, `(;)`, `(;;)` should throw.

> just to clarify what does no-op mean ?

no-op means we do not change `ensureLastUndefined`, as you can see it is reset as `false` in the beginning of loop. We can check empty statement before validating.

Let `eLU` be `ensureLastUndefined`, we can illustrate how `eLU` is changed between EmptyStatements ant other statements.

```
;                ;;;;;                              42                  ;;;;;
|--> eLU = true  |--> eLU = (previous eLU) = true   |--> eLU = false    |--> eLU= (previous eLU) = false
```

wlawt: if `()` , `(;)` , `(;;)` should throw would it be a different issue because checking if `node` is first fails the cases just mentioned?

JLHwung: Babel currently throws syntax error for `()`, `(;)`, which is expected. Generally when you are working on AST, you may assume that it is always valid combinations of AST nodes, in this case, it is `EmptyStatement` nested in `BlockStatement`. The combinations of punctuators, i.e. `(`, `)` should be cared by the parser, and by the time when you have a valid AST, it can not be a representation of the source like `()`.

Generally parentheses do not affect the semantic meaning, it is a valid combination only if there is an expression wrapped inside parentheses: https://tc39.es/ecma262/#prod-ParenthesizedExpression
```
ParenthesizedExpression[Yield, Await]:
  ( Expression[+In, ?Yield, ?Await] )
```

wlawt: right, ok the changes i just did keeps the checking if `node` is first and letting it throw if `()` , `(;)`

JLHwung: I think I understand what you mean by `()` and `(;)`

In the example `{ ;; }`, the parse AST is
```
[ EmptyStatement, EmptyStatement ]
```
In the first loop, `ensureLastUndefined` is set to `true`.
In the second loop, `ensureLastUndefined` is set to `false` in the beginning of the loop body: 
https://github.com/babel/babel/blob/6812a4297d1f0ecdad2ab0a67f752d9201f65b7b/packages/babel-types/src/converters/gatherSequenceExpressions.js#L28

Since `exprs` is an empty array, `ensureLastUndefined` is `false`, `sequenceExpression(exprs)` will be printed as `()`, which will throw syntax error because sequence expression can not be empty.

We can reset `ensureLastUndefined` only if `node` is not an `EmptyStatement`.

```js
for (const node of nodes) {
  if (!isEmptyStatement(node)) {
    ensureLastUndefined = false;
  }
// ...
}
```

So this ensures that when we are processing an EmptyStatement following any statements, it becomes an no-op, which does _not_ change `ensureLastUndefined`.

If we only check `node is first of nodes && nodes.length === 2`, it can not handle `do { ;;; }`: When processing the third empty statement, `ensureLastUndefined` will still be reset to `false` and `gatherSequenceExpressions` will still generate an empty sequence expression.

wlawt: That makes so much sense, thank u!

packages/babel-types/test/converters.js
```diff
@@ -320,7 +320,10 @@ describe("converters", function () {
     it("gathers empty statements", function () {
       const node = parseCode(";");
       const sequence = t.toSequenceExpression([undefinedNode, node], scope);
-      expect(generateCode(sequence.expressions[1])).toBe("undefined");
+      expect(() => {
+        generateCode(sequence.expressions[1]);
+      }).toThrow("Cannot read property '1' of undefined");
+      // expect(generateCode(sequence.expressions[1])).toBe("undefined");
```
JLHwung, line 326: The test is failing because the test assumption does not hold if the empty statement is following a value producing statement, in this test case, `undefined ;` will _not_ produce a sequence expression because the empty statement is not the leading statement. Therefore `sequence.expression` is `undefined`.

However, asserting an error thrown because it is in fact _not_ an sequence expression can be confusing to other contributors. Moreover, we know that it still _gathers_ empty statement only if the `EmptyStatement` is the first element:

So we need to revise the test and clearly state its new behaviour

```js
it("gathers leading empty statement", function() {
  const node = parseCode(";");
  const sequence = t.toSequenceExpression([node, t.cloneNode(node)], scope);
  expect(generateCode(sequence)).toBe("undefined");
});
```

wlawt: Oh I see, also is the difference between `toBeUndefined()` vs. `toBe("undefined")`  just the type it returns?

packages/babel-types/test/converters.js
```diff
@@ -320,7 +320,7 @@ describe("converters", function () {
     it("gathers empty statements", function () {
```
JLHwung, line 320: Can you also change the test title here? Because we have changed the behaviour.

JLHwung:
The e2e CI error has been fixed on `main`.

PR #11728, opened 2020-06-17
fix(do-expr): SwitchStatement with IfStatement cases

### LINKED ISSUE #11608: do-expressions and nested breaks
## Bug Report

```js
console.log(
  do {
    switch ("1") {
      case "1":
        if (true) {
          "out";
          break;
        }
    }
  }
);

console.log(
  do {
    x: {
      "1";
      break x;
    }
  }
);
```

This should print `1` twice. Instead it prints `undefined` twice.

- [REPL](https://babeljs.io/repl#?browsers=ie%209&build=&builtIns=usage&spec=false&loose=false&code_lz=MYewdgziA2CmB00QHMAUAoABJgJiTA3ltphAO4CWALsABaaoBEAjIwJSHEmbACGEsTC0YAuLt0wUAZgyoAnAK6wORCRKYgFVdgG5xagEZzYvANZ612AL7ib19Gz3pQkGAiRpieTtwAeIn3VhR31MIxNTTF8LEjtMGxCgA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Ces2015%2Ces2016%2Ces2017%2Creact%2Cstage-0%2Cstage-1%2Ctypescript%2Cenv&prettier=false&targets=Node-8.9&version=7.9.6&externalPlugins=)

barronwei:
| Fixed Issues? | Fixes #11608 
| Patch: Bug Fix? | Yes
| Major: Breaking Change? | No
| Minor: New Feature? | No 
| Tests Added + Pass? | Yes
| Documentation PR Link |
| Any Dependency Changes? | No

This PR checks for whether the last statement in the `BlockStatement` case for `getCompletionRecords` is a `BreakStatement`. 

As of right now, the code just pops off the last statement of a `BlockStatement` and pushes that onto `completionRecords`. 

This last statement could be a `BreakStatement` in either a `IfStatement` or `LabeledStatement`, rendering the return value to be `undefined`.

This PR correctly handles:

```
const x = (n) => do {
  switch (n) {
    case 0:
      if (true) {
        'out';
        break;
      }
      
   }
}
```

```
const x = do {
  y: {
    'out';
    break y;
  }
}
```

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11728"></a>


packages/babel-traverse/src/path/family.js
```diff
@@ -81,14 +81,21 @@ export function getCompletionRecords(): NodePath[] {
     paths = addCompletionRecords(this.get("alternate"), paths);
   } else if (this.isDoExpression() || this.isFor() || this.isWhile()) {
     paths = addCompletionRecords(this.get("body"), paths);
-  } else if (this.isProgram() || this.isBlockStatement()) {
+  } else if (this.isProgram()) {
+    paths = addCompletionRecords(this.get("body").pop(), paths);
+  } else if (this.isBlockStatement()) {
```
JLHwung, line 86: We can not remove `BreakStatement` because it will lose the completion record when the `prevSibling` is neither an ExpressionStatement nor a BlockStatement. i.e.
```js
do { switch(0) { case 0: {
  if (true) { break; } "unreachable"; 
  } }
}
```
If `break` is removed, `"unreachable"` will be incorrectly included in the completion record.

The completion record tracking of SwitchStatement was introduced at #10070. The `BreakStatement` is handled in https://github.com/babel/babel/blob/e498bee10f0123bb208baa228ce6417542a2c3c4/packages/babel-traverse/src/path/family.js#L52-L60

The `BreakStatement` is searched in `consequent` and `consequent[type=BlockStatement] > body`

https://github.com/babel/babel/blob/e498bee10f0123bb208baa228ce6417542a2c3c4/packages/babel-traverse/src/path/family.js#L29-L41

Apparently this routine can not find `BreakStatement` nested in `IfStatement` of the `SwitchCase`. We can do a tree search for `BreakStatement`, which resembles `getCompletionRecords`.

```js
function findBreak(statements) {
  let breakStatement;
  if (!Array.isArray(statements)) {
    statements = [statements];
  }

  for (const statement of statements) {
    if (statement.isBlockStatement()) {
      breakStatement = findBreak(statement.get("body"));
    } else if (statement.isIfStatement()) {
      breakStatement =
        findBreak(statement.get("consequent")) ??
        findBreak(statement.get("alternate"));
    } else if (statement.isCatchClause() || statement.isLabeledStatement()) {
      breakStatement = findBreak(statement.get("body"));
    // todo: add other statements except Iteration and Switch, which is breakable statement.
    } else if (statement.isBreakStatement()) {
      breakStatement = statement;
    }
    if (breakStatement) {
      return breakStatement;
    }
  }
  return null;
}
```
By doing so we can reuse current `completionRecordForSwitch` routine which already handled `BreakStatement`.

packages/babel-plugin-proposal-do-expressions/test/fixtures/do-expression-switch-case/condition-before-break/input.js
```diff
@@ -0,0 +1,10 @@
+const x = (n) => do {
+  switch (n) {
+    case 0:
+      if (true) {
+        'out';
+        break;
+      }
```
JLHwung, line 7: Can you add a test case where `BreakStatement` does not follow an `ExpressionStatement`?
```js
case 1:
  if (true) {
    break;
  }

```

packages/babel-plugin-proposal-do-expressions/test/fixtures/do-expressions/labeled-break/input.js
```diff
@@ -0,0 +1,6 @@
+const x = do {
+  y: {
+    'out';
+    break y;
```
JLHwung, line 4: We can not remove labelled break statement without more execution context, i.e.
```js
// returns 42
do { y: { "out"; break y; } 42 }
```
I don't come up with a satisfiable fix regarding to the LabeledStatement without significant rewriting `getCompetionsRecords`. I suggest we separate this fix into another PR and this PR is focused on SwitchStatement.

barronwei: Will do, but does `getCompletionRecords` not get the last node of the `BlockStatement`, which would be `42` in this case?

JLHwung: If we replace `"out"; break y;` by `return "out"`, the compiled output will be
```js
(() => { y: { return "out"; } return 42 } })()
// returns "out"
```
I don't think we correctly handle the labeled break statement in `getCompletionRecords`, so I suggest we address `LabelStatement` in another PR.

barronwei: Oh I see. I forgot about the loop. Thanks for the feedback, and sounds good!

packages/babel-traverse/src/path/family.js
barronwei: What is the reasoning behind returning `getCompletionRecords` as opposed to appending to `paths` in the `isFunction` case?

JLHwung: This is a good question! Each function owns a new execution context. So when a `Function` is met, we should skip finding `break` inside the function body because the statements inside can not interfere with completion records (control flows) outside. i.e.

```js
{
  a: 42;
  () => { break a }
}
```
is invalid because the label `a` is _not_ defined in the execution context own by the arrow function.

barronwei: That completely makes sense. I have made the necessary changes. Please let me know if there needs to be any others!

barronwei:
Is there anything that needs to be changed?

JLHwung:
@barronwei Generally we have two-approval process for PR. I will ping some core members to see if they can review.

cc @existentialism @kaicataldo @nicolo-ribaudo

kaicataldo (approved):
LGTM, thanks!

PR #11734, opened 2020-06-19
Account for ConfigItem being generated by another copy of Babel

developit:
| Fixed Issues?            | developit/microbundle#657
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | 

This fixes `ConfigItem` unboxing when `core.createConfigItem()` is used to generate a ConfigItem that is then passed to a different copy of Babel. This is common in cases where npm fails to dedupe modules, or when differing versions of Babel are in-use.

```
// good - uses the version of babel that is doing the transforming:
module.exports = function preset(api) {
  api.createConfigItem(...)
}
// bad - uses whatever version this module installed:
import api from '@babel/core';
module.exports = babelLoader.custom({
  options() {
    api.createConfigItem(..)
  }
});
```

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11734"></a>

### DIFF
--- a/packages/babel-core/src/config/item.js
+++ b/packages/babel-core/src/config/item.js
@@ -49,6 +49,8 @@ export function getItemDescriptor(item: mixed): UnloadedDescriptor | void {
 
 export type { ConfigItem };
 
+const CONFIG_ITEM_BRAND = Symbol.for("@babel/core@7 - ConfigItem");
+
 /**
  * A public representation of a plugin/preset that will _eventually_ be load.
  * Users can use this to interact with the results of a loaded Babel
@@ -67,7 +69,7 @@ class ConfigItem {
   /**
    * Used to detect ConfigItem instances from other Babel instances.
    */
-  [Symbol.for("@babel/core@7 - ConfigItem")] = true;
+  [CONFIG_ITEM_BRAND] = true;
 
   /**
    * The resolved value of the item itself.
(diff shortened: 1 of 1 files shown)


developit:
Another suggestion, rather than changing all the instanceof checks, might be to throw or warn when this behavior is detected:

```js
import core from '@babel/core';  // could be a separate copy of Babel (determined by npm)

export function custom() {
  return {
    config() {
      core.createConfigItem(...)
      // "Error: detected ConfigItem instances from a different version of Babel"
    }
  }
}
```

packages/babel-core/src/config/item.js
```diff
@@ -64,6 +64,11 @@ class ConfigItem {
    */
   _descriptor: UnloadedDescriptor;
 
+  /**
+   * Used to detect ConfigItem instances from other Babel instances.
+   */
+  _isConfigItem: true;
```
nicolo-ribaudo, line 70:
```suggestion
  [Symbol.for("@babel/core@7 - ConfigItem")]: true;
```

And we can remove the `instanceof` check and only rely on this.

packages/babel-core/src/config/item.js
```diff
@@ -40,7 +40,7 @@ export function createConfigItem(
 }
 
 export function getItemDescriptor(item: mixed): UnloadedDescriptor | void {
-  if (item instanceof ConfigItem) {
+  if (item && item[Symbol.for("@babel/core@7 - ConfigItem")] === true) {
```
JLHwung, line 43:
```suggestion
  if (item?.[Symbol.for("@babel/core@7 - ConfigItem")] === true) {
```

packages/babel-core/src/config/item.js
```diff
@@ -64,6 +64,11 @@ class ConfigItem {
    */
   _descriptor: UnloadedDescriptor;
 
+  /**
```
nicolo-ribaudo, line 67:
```suggestion
  // TODO(Babel 8): Check if this symbol needs to be updated
  /**
```

nicolo-ribaudo (approved):

existentialism (approved):

PR #11739, opened 2020-06-22
[cjs] Skip duplicate reexported bindings in namespace reexports

### LINKED ISSUE #11710: Transpiling (non-conflicting) namespace reexports emits invalid code
## Bug Report

**Current behavior**

```
TypeError: Cannot redefine property: c
```

Namespace reexports are transpiled like this:
```js
for(key ...){
  Object.defineProperty(exports, key, {
    enumerable: true,
    get: function () {
      return imported[key];
    }
  });
}
```

But the spec allows for reexporting a symbol twice in the same file if it is actually the same binding (see example code below).

This means that `Object.defineProperty(exports, "c", ...)` is called twice, causing the error.

One of the packages that do this is `@fluentui/react` (this came up in https://github.com/parcel-bundler/parcel/issues/4399)

**Input Code**

- https://github.com/mischnic/parcel-issue-4399 (`lib/` already contains the Babel output of `src-valid`)

```js
export * from "./a.js";
export * from "./b.js";

// ---- everything below is just for context ---

// a.js
export { c } from './c.js'; // <-- 
export const a = "a";
// b.js
export { c } from './c.js'; // <-- 
export const b = "b";

// c.js
export function c() {
	return false;
}
```

**Expected behavior**

The first file should export `a, b, c` (and not throw).

```sh
babel --plugins @babel/plugin-transform-modules-commonjs --delete-dir-on-start src-valid -d lib
```

**Environment**

```
  System:
    OS: macOS 10.15.4
  Binaries:
    Node: 14.4.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - /usr/local/bin/npm
  npmPackages:
    @babel/cli: ^7.10.1 => 7.10.1
    @babel/core: ^7.10.2 => 7.10.2
    @babel/plugin-transform-modules-commonjs: ^7.10.1 => 7.10.1
```

**Possible Solution**
Not use `Object.defineProperty`? `configurable: true` ? But all of these have other sideffects

**Additional context**

Technically, f there are conflicting reexports (see `src-invalid`), there should be a SyntaxError (!).

But even [Rollup doesn't handle this correctly](https://rollupjs.org/repl/?version=2.15.0&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMColMjBmcm9tJTIwJTVDJTIyLiUyRmEuanMlNUMlMjIlM0IlNUNuZXhwb3J0JTIwKiUyMGZyb20lMjAlNUMlMjIuJTJGYi5qcyU1QyUyMiUzQiUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTdEJTJDJTdCJTIybmFtZSUyMiUzQSUyMmEuanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyZXhwb3J0JTIwJTdCJTIwYyUyMCU3RCUyMGZyb20lMjAnLiUyRmMxLmpzJyUzQiU1Q25leHBvcnQlMjBjb25zdCUyMGElMjAlM0QlMjAlNUMlMjJhJTVDJTIyJTNCJTIyJTJDJTIyaXNFbnRyeSUyMiUzQWZhbHNlJTdEJTJDJTdCJTIybmFtZSUyMiUzQSUyMmIuanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyZXhwb3J0JTIwJTdCJTIwYyUyMCU3RCUyMGZyb20lMjAnLiUyRmMyLmpzJyUzQiU1Q25leHBvcnQlMjBjb25zdCUyMGIlMjAlM0QlMjAlNUMlMjJiJTVDJTIyJTNCJTIyJTJDJTIyaXNFbnRyeSUyMiUzQWZhbHNlJTdEJTJDJTdCJTIybmFtZSUyMiUzQSUyMmMxLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGZ1bmN0aW9uJTIwYygpJTIwJTdCJTVDbiU1Q3RyZXR1cm4lMjBmYWxzZSUzQiU1Q24lN0QlMjIlN0QlMkMlN0IlMjJuYW1lJTIyJTNBJTIyYzIuanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyZXhwb3J0JTIwZnVuY3Rpb24lMjBjKCklMjAlN0IlNUNuJTVDdHJldHVybiUyMHRydWUlM0IlNUNuJTdEJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmVzJTIyJTJDJTIybmFtZSUyMiUzQSUyMm15QnVuZGxlJTIyJTJDJTIyYW1kJTIyJTNBJTdCJTIyaWQlMjIlM0ElMjIlMjIlN0QlMkMlMjJnbG9iYWxzJTIyJTNBJTdCJTdEJTdEJTJDJTIyZXhhbXBsZSUyMiUzQW51bGwlN0Q=) (and Babel cannot work across multiple files anyway). https://github.com/rollup/rollup/issues/3629

mischnic:
| Fixed Issues?            | Fixes #11710
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes (I think)

Namespace reexports: skip already added reexports on the `exports` object if they are strictly equal (as suggested in https://github.com/babel/babel/issues/11710#issuecomment-647175235).

Questions:

1. Not sure if this change could cause other side effects?
	- Currently, loose mode wouldn't throw any errors on duplicate bindings (in non-loose `Object.defineProperty` does this automatically). However, the order of the objects should be the same (because both skip if already set).
2. I don't think there are any (integration) tests which aren't snapshot tests?

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11739"></a>

### DIFF
--- a/packages/babel-helper-module-transforms/src/index.js
+++ b/packages/babel-helper-module-transforms/src/index.js
@@ -241,7 +241,13 @@ function buildNamespaceReexport(metadata, namespace, loose) {
           EXPORTS[key] = NAMESPACE[key];
         });
       `
-    : template.statement`
+    : // Also skip already assigned bindings if they are strictly equal
+      // to be somewhat more spec-compliant when a file has multiple
+      // namespace re-exports that would cause a binding to be exported
+      // multiple times. However, multiple bindings of the same name that
+      // export the same primitive value are silently skipped
+      // (the spec requires an "ambigous bindings" early error here).
+      template.statement`
         Object.keys(NAMESPACE).forEach(function(key) {
           if (key === "default" || key === "__esModule") return;
           VERIFY_NAME_LIST;
--- a/packages/babel-helper-module-transforms/src/index.js
+++ b/packages/babel-helper-module-transforms/src/index.js
@@ -236,6 +236,7 @@ function buildNamespaceReexport(metadata, namespace, loose) {
         Object.keys(NAMESPACE).forEach(function(key) {
           if (key === "default" || key === "__esModule") return;
           VERIFY_NAME_LIST;
+          if (key in EXPORTS && EXPORTS[key] === NAMESPACE[key]) return;
 
           EXPORTS[key] = NAMESPACE[key];
         });
@@ -244,6 +245,7 @@ function buildNamespaceReexport(metadata, namespace, loose) {
         Object.keys(NAMESPACE).forEach(function(key) {
           if (key === "default" || key === "__esModule") return;
           VERIFY_NAME_LIST;
+          if (key in EXPORTS && EXPORTS[key] === NAMESPACE[key]) return;
 
           Object.defineProperty(EXPORTS, key, {
             enumerable: true,
--- a/packages/babel-plugin-transform-modules-amd/test/fixtures/amd/export-from/output.js
+++ b/packages/babel-plugin-transform-modules-amd/test/fixtures/amd/export-from/output.js
@@ -6,6 +6,7 @@ define(["exports", "foo"], function (_exports, _foo) {
   });
   Object.keys(_foo).forEach(function (key) {
     if (key === "default" || key === "__esModule") return;
+    if (key in _exports && _exports[key] === _foo[key]) return;
     Object.defineProperty(_exports, key, {
       enumerable: true,
       get: function () {
--- a/packages/babel-plugin-transform-modules-amd/test/fixtures/loose/export-from/output.js
+++ b/packages/babel-plugin-transform-modules-amd/test/fixtures/loose/export-from/output.js
@@ -4,6 +4,7 @@ define(["exports", "foo"], function (_exports, _foo) {
   _exports.__esModule = true;
   Object.keys(_foo).forEach(function (key) {
     if (key === "default" || key === "__esModule") return;
+    if (key in _exports && _exports[key] === _foo[key]) return;
     _exports[key] = _foo[key];
   });
 });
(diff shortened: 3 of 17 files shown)


packages/babel-helper-module-transforms/src/index.js
```diff
@@ -236,6 +236,7 @@ function buildNamespaceReexport(metadata, namespace, loose) {
         Object.keys(NAMESPACE).forEach(function(key) {
           if (key === "default" || key === "__esModule") return;
           VERIFY_NAME_LIST;
+          if (key in EXPORTS && EXPORTS[key] === NAMESPACE[key]) return;
```
nicolo-ribaudo, line 239: Since this is technically not spec compliant, could you add a comment explaining the reason to do it?

mischnic: Done (directly above the template string)

mischnic:
> Can you update the test fixtures?

Done

nicolo-ribaudo (approved):

nicolo-ribaudo:
Thanks!

PR #11740, opened 2020-06-22
chore: add rejects assertion to fixture-test-runner

JLHwung:
| Minor: New Feature? | `"rejects": string` is supported
| Major: Breaking Change?  | `"throws": true` is no longer accepted in test fixture options

This PR adds `rejects: string` supports to our fixture runner. When `rejects` is specified, it will check if an error is thrown asynchronously from the fixture tests. This feature is depended by #11141 due to the fact that we brings async config support to `@babel/core`.

Since `assert.rejects` is supported since node v10.0.0, this feature targets to Babel 8 only.

It also removes currently unused `"throws": boolean` support, to make it indeed a breaking change. 😄

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11740"></a>

### DIFF
--- a/packages/babel-helper-transform-fixture-test-runner/src/index.js
+++ b/packages/babel-helper-transform-fixture-test-runner/src/index.js
@@ -228,6 +228,7 @@ function run(task) {
         !expected.code &&
         outputCode &&
         !opts.throws &&
+        !opts.rejects &&
         fs.statSync(path.dirname(expected.loc)).isDirectory() &&
         !process.env.CI
       ) {
@@ -385,13 +386,22 @@ export default function (
             if (dynamicOpts) dynamicOpts(task.options, task);
 
             const throwMsg = task.options.throws;
+            const rejectMsg = task.options.rejects;
             if (throwMsg) {
               // internal api doesn't have this option but it's best not to pollute
               // the options object with useless options
               delete task.options.throws;
 
-              assert.throws(runTask, function (err) {
-                return throwMsg === true || err.message.indexOf(throwMsg) >= 0;
+              return assert.throws(runTask, function (err) {
+                assert.ok(err.message.indexOf(throwMsg) >= 0);
+                return true;
+              });
+            } else if (rejectMsg) {
+              delete task.options.rejects;
+
+              return assert.rejects(runTask, function (err) {
+                assert.ok(err.message.indexOf(rejectMsg) >= 0);
+                return true;
               });
             } else {
               if (task.exec.code) {
--- a/packages/babel-helper-transform-fixture-test-runner/src/index.js
+++ b/packages/babel-helper-transform-fixture-test-runner/src/index.js
@@ -393,14 +393,14 @@ export default function (
               delete task.options.throws;
 
               return assert.throws(runTask, function (err) {
-                assert.ok(err.message.indexOf(throwMsg) >= 0);
+                assert.ok(err.message.includes(throwMsg));
                 return true;
               });
             } else if (rejectMsg) {
               delete task.options.rejects;
 
               return assert.rejects(runTask, function (err) {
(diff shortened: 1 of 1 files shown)


packages/babel-helper-transform-fixture-test-runner/src/index.js
```diff
@@ -341,6 +342,48 @@ const toEqualFile = () => ({
...
+  }
+
+  const throwMsg = task.options.throws;
+  if (throwMsg) {
+    // internal api doesn't have this option but it's best not to pollute
+    // the options object with useless options
+    delete task.options.throws;
+
+    assert.throws(runTask, function (err) {
+      assert.ok(err.message.indexOf(throwMsg) >= 0);
```
JLHwung, line 357:
```
return err.message.indexOf(throwMsg) >= 0;
```
is also changed to `assert.ok`. See https://nodejs.org/api/assert.html#assert_assert_throws_fn_error_message


>  Avoid returning anything from validation functions besides `true`.
    Otherwise, it's not clear what part of the validation failed. Instead, throw an error about the specific validation that failed (as done in this example) and add as much helpful debugging information to that error as possible.

packages/babel-helper-transform-fixture-test-runner/src/index.js
```diff
@@ -341,6 +342,48 @@ const toEqualFile = () => ({
...
+      return true;
+    });
+  } else {
+    if (task.exec.code) {
+      const result = runTask;
+      if (result && typeof result.then === "function") {
+        return result;
```
nicolo-ribaudo, line 364: Do we need the sync and async handles, or can we handle `.rejects` here?

JLHwung: I don't think we can handle `.rejects` here because `task.exec` is not defined. In #11141 it was used to reproduce the `"useSpread"` error message in test fixtures.

`assert.rejects` does behave like `assert.throws` except for the asynchronous behaviour, but I tend to be specific about the error synchrony.

JLHwung:
This PR has been simplified to only call `assert.rejects` when `rejects` is specified. The test wrapping is unchanged. PTAL.

packages/babel-helper-transform-fixture-test-runner/src/index.js
```diff
@@ -385,13 +386,22 @@ export default function (
...
+            const rejectMsg = task.options.rejects;
             if (throwMsg) {
               // internal api doesn't have this option but it's best not to pollute
               // the options object with useless options
               delete task.options.throws;
 
-              assert.throws(runTask, function (err) {
-                return throwMsg === true || err.message.indexOf(throwMsg) >= 0;
+              return assert.throws(runTask, function (err) {
+                assert.ok(err.message.indexOf(throwMsg) >= 0);
```
existentialism, line 396: Nit: indexOf -> includes?

packages/babel-helper-transform-fixture-test-runner/src/index.js
```diff
@@ -385,13 +386,22 @@ export default function (
...
             const throwMsg = task.options.throws;
+            const rejectMsg = task.options.rejects;
             if (throwMsg) {
               // internal api doesn't have this option but it's best not to pollute
               // the options object with useless options
               delete task.options.throws;
 
-              assert.throws(runTask, function (err) {
-                return throwMsg === true || err.message.indexOf(throwMsg) >= 0;
+              return assert.throws(runTask, function (err) {
```
JLHwung, line 395: Though `assert.rejects` shares same behaviour with `assert.throws` on synchronous errors, which means we can simply replace `assert.throws` by `assert.rejects` and extend `throws` so that it contains both sync and async errors. However I think it is still valuable to differentiate between them.

PR #11741, opened 2020-06-23
Allow plugins to indicate dependencies on random files

### LINKED ISSUE #8497: Allow plugins to indicate dependencies on random files
## Feature Request

In contexts where Babel is running alongside a filesystem watcher, there are cases where a plugin may load an arbitrary file. Take a plugin like https://www.npmjs.com/package/babel-plugin-inline-import for example. Or some usecases of babel-plugin-macros.

It would be ideal if Babel had a way for plugins to indicate that these dependencies exist, so that if callers of Babel wanted to take that into account, they could. For example, Webpack has [`.addDependency`](https://webpack.js.org/api/loaders/#this-adddependency), so that changes to that file will cause the loader to re-execute when the dependency is changed. We could easily use this API in `babel-loader` to handle this. Implementing a similar API in `@babel/cli`'s watcher wouldn't be too bad either.

I don't know how other Babel wrappers would fair, but if someone wanted to help investigate that, it would certainly help. Off the top off my head we'd want to check, at the very least, if the following have ways to register additional files to watch:

* Meteor
* Jest
* Browserify
* Gulp

### Edit

We may also want to include in this work logic to help us take dependency file content into account for the purposes of cache invalidation. For example, `babel-loader`'s `cacheDirectory` option will not invalidate caches, if say a `.browserlistrc` file changes `preset-env`'s plugin list: https://github.com/babel/babel-loader/issues/690

vedantroy:
| Fixed Issues?            | Fixes #8497
| Tests Added + Pass?      | No (existing tests pass, but need help on adding new tests)
| Documentation PR Link    | 

I took a shot at fixing https://github.com/babel/babel/issues/8497. ~~The fix seems surprisingly simple, so I might be missing something obvious.~~ (was missing something obvious)

This PR adds a `addExternalDependency` function to Babel's plugin API. Plugins can call this function with 2 values:
- the path of the external dependency
- the path of the dependent file (usually the current file the plugin is processing).

Calling `addExternalDependency` establishes a 2 way-mapping. Specifically, @babel/core exports 2 methods:
- `getExternalDependencies` - This method returns a map of external dependency names to their dependent files
- `getDependencies` - This method returns a map of source file names to the non-source files (external dependencies) that depend on them.

This PR integrates the above information with @babel/cli's --watch flag. Whenever a file is compiled in watch mode, @babel/cli will check the above information and watch any new external dependencies.

This is a draft PR because there's a few things missing:
- ~~tests (I'm not sure how to test this, given that it's not a normal code transformation)~~
   - there are cli tests, but writing watch tests seems much more complicated. I will look more into this. The ideal test would be one that loads a plugin that relies on a non-source file, and then transforming that non-source file during the test to ensure the source file is also transformed. This seems complicated, but also the most comprehensive option.
- I need to make the code work with flow better
- should I rename dependencies/externalDependencies to sourceFile/nonSourceFile

This PR could work in conjunction with a PR to babel-loader. Specifically, babel-loader could use `getExternalDependencies` to call webpack's `addDependency`.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11741"></a>

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11741"></a>

### DIFF
--- a/packages/babel-cli/src/babel/util.ts
+++ b/packages/babel-cli/src/babel/util.ts
@@ -192,11 +192,9 @@ export function onDependencyFileChanged(
         for (const dependent of externalFileDeps.get(filePath)) {
           await callback(dependent);
         }
-        await callback(filePath);
       }
-    } else {
-      await callback(filePath);
     }
+    return await callback(filePath);
   }
   ["add", "change"].forEach(type => getWatcher().on(type, onFileChanged));
 }
--- a/packages/babel-cli/src/babel/util.ts
+++ b/packages/babel-cli/src/babel/util.ts
@@ -4,6 +4,10 @@ import path from "path";
 import fs from "fs";
 import { createRequire } from "module";
 
+/**
+ * Set the file permissions of dest to the file permissions
+ * of src.
+ */
 export function chmod(src: string, dest: string): void {
   try {
     fs.chmodSync(dest, fs.statSync(src).mode);
@@ -60,7 +64,7 @@ const CALLER = {
   name: "@babel/cli",
 };
 
-export function transform(
+export async function transform(
   filename: string,
   code: string,
   opts: any,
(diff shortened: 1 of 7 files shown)


packages/babel-core/src/config/full.js
```diff
@@ -223,6 +226,7 @@ const loadDescriptor = makeWeakCache(function* (
       ...context,
       ...makeAPI(cache),
     };
+    api.addExternalDependency = fileName => dependencies.add(fileName);
```
JLHwung, line 229: Can you move `addExternalDependency` to `packages/babel-core/src/config/helpers/config-api.js`? We should also add it to the `PluginAPI` interface.

packages/babel-core/src/index.js
```diff
@@ -22,6 +22,7 @@ export {
   loadOptions,
   loadOptionsSync,
   loadOptionsAsync,
+  getExternalDependencies,
```
JLHwung, line 25: It is okay to export from a nested sub folder like `./config/helpers/config-api.js`.

vedantroy:
@JLHwung I'm integrating this with @babel/cli, so the watch method will recompile a file when an external dependency that is associated with it is changed.

Do you know why `disableGlobbing` is set to `true`, [here](https://github.com/babel/babel/blob/65911144e9eee760052e20fed9780bd78922f942/packages/babel-cli/src/babel/file.js#L219), but not set to `true` [here](https://github.com/babel/babel/blob/65911144e9eee760052e20fed9780bd78922f942/packages/babel-cli/src/babel/dir.js#L193).

JLHwung:
@vedantroy 

> Do you know why disableGlobbing is set to true, here, but not set to true here.

Good catch, I believe `disableGlobbing` should be set for the latter, too.

JLHwung:
> Which version is right / or is there a reason that file.js has additional checks?

I think it should also be applied for `dir.js`. The changes you mentioned before was made by me one years ago. I have no idea if I had special reasons other than that I simply overlooked them 😄. 

> I'm consolidating some of the code in dir.js and file.js in utils.js

Can you address them in a separate PR? So we can fast track them in patch release.

> I think right now this bug does not occur in practice because the first check (!util.isCompilableExtension) returns false, so the if statement never triggers.

Correct, if a folder is specified in `filenames`, users should always specified `--extensions` otherwise babel doesn't know it should compile `foo/*.ts`. The `filenames.includes` is working only for `babel foo.ts` cases, since `--extensions ts` is implied from the arguments.

 > there doesn't seem to be any enforcement of ordering

It could happen, though we have used `awaitFinish` in `chokidar` so that every compilation should has a 50ms time window during which the file can be considered unchanged. But we can't ensure a compilation task can be finished in 50ms. I suggest we can explore them in a separate PR beside the changes we mentioned above.

vedantroy:
@JLHwung -- Another question while I'm integrating the external dependency tracking with "dir.js" is:

What is happening in [these 3 lines of code](https://github.com/babel/babel/blob/cfaa70dcf48a0ea9a6c7adb4ae1f7925934b3933/packages/babel-cli/src/babel/dir.js#L214)?

Also what are the inputs of [handleFile](https://github.com/babel/babel/blob/cfaa70dcf48a0ea9a6c7adb4ae1f7925934b3933/packages/babel-cli/src/babel/dir.js#L100)? I see `src` and `base`, but, I don't know exactly what they mean. (Especially, `base`)

packages/babel-core/src/config/full.js
vedantroy: Fixed.

packages/babel-core/src/index.js
vedantroy: Fixed.

vedantroy:
@JLHwung -- the PR is ready for review now!

As a sidenote, this PR has a [video doc](https://getvideodoc.com/) attached to it. You can view it [here](https://app.getvideodoc.com/babel/babel/1).

You can also download the client [here](https://getvideodoc.com/download). Using the client, you can do `vdoc blame <file name>` to  get a line-by-line view of a file and its corresponding video docs.

Typing `vdoc blame` on any file modified in this PR will show the video doc linked above as well as the lines of code it corresponds to. I hope this will make it easier for future contributors to modify this part of the codebase.

vedantroy:
@JLHwung any updates on this? I do think this would be helpful for plugin/macro authors. At the same time, I understand that the Babel core team is busy, so if right now's not a good time, I can check back in a month or two :smiley:

goatandsheep (approved):

goatandsheep:
@sosukesuzuki anywhere we can improve on this to get this out?

vedantroy:
@goatandsheep I'm not a maintainer, just the PR author, but it seems good to add some tests to verify this new feature works.

MauriceArikoglu:
Any news when this will hit prod? @JLHwung @vedantroy @goatandsheep @nicolo-ribaudo

JLHwung:
@vedantroy @MauriceArikoglu Sorry I completely forgot this PR. I will rebase on the latest main and it should be ready for review.

MauriceArikoglu:
@JLHwung thank you! Can I be of any help in the review process?

packages/babel-cli/src/babel/util.ts
```diff
@@ -71,26 +75,132 @@ export function transform(
...
+    const externalFileDeps = babel.getExternalDependencies();
+    if (externalFileDeps.has(filePath)) {
+      if (sourceFilesAreCompiledIntoASingleFile) {
+        // When using --out-file, Babel traverses all the files every time
+        // so there's no point in calling the callback multiple times. The callback
+        // for --out-file knows to recompile no matter what if it receives null.
+        return await callback(null);
+      } else {
+        for (const dependent of externalFileDeps.get(filePath)) {
+          await callback(dependent);
```
JLHwung, line 193: This can be improved by `Promise.all`: the external deps can be compiled in parallel.

MauriceArikoglu: @JLHwung can it actually? what if a dependency depends on another dependency?

JLHwung (approved):

MauriceArikoglu:
🚀 🚀 🚀 

Thanks to @JLHwung and @vedantroy!

packages/babel-cli/src/babel/util.ts
vedantroy: I think this should be changed to `Promise.all`, alas I don't have tons of bandwidth to do this right now so if @JLHwung wants to change it, that'd be great.

nicolo-ribaudo (changes requested):
I downloaded this locally to play a bit with it, and there are a few changes that we need to do:
- `api.addExternalDependency` should only take a single parameter: the dependency path. It should already know the path of the current file.
- the `dependencies` and `externalDependencies` set contain the same data, they are just inverted; I don't understand why we need both of them.

---

What should happen when the dependencies of a file change? Assume we have this plugin, which replaces the `INLINE_ME` variable with the file contents.
```js
function inlineFilePlugin(api, { inlineFile }) {
  const { types: t } = api;

  api.addExternalDependency(inlineFile, "./demo.js");

  return {
    visitor: {
      Identifier(path) {
        if (path.node.name === "INLINE_ME") {
          path.replaceWith(t.stringLiteral(readFileSync(inlineFile, "utf8")));
        }
      },
    },
  };
}
```

the user might be running it like this:
```js
babel.transformFileSync("./demo.js", {
  plugins: [[inlineFilePlugin, { inlineFile: "./file1.txt" }]],
});
```

The user has a watcher that uses `babel.getExternalDependencies()` to re-trigger their compiler step. Whenever `file1.txt` changes, it re-compiles `demo.js`.

At some point the config changes to `{ inlineFile: "./file2.txt" }`: when `file1.txt` changes, it should not re-compile `demo.js` anymore because it's not a dependency anymore.

I was going to propose fixing this by re-setting `demo.js`'s dependencies whenever it's compiled, but then I realized that the user might want to compile both with `{ inlineFile: "./file1.txt" }` and `{ inlineFile: "./file2.txt" }` for different build targets that run in "parallel".

The solution to all of this is to _not_ store dependencies in a shared global, but to return them in a `Set` in the result of `babel.transform*`: Babel integrations are then free that set however they want, avoiding:
- global state that leaks across unrelated build pipelines
- "phantom" dependencies that should not be marked as dependencies anymore.

---

I can do these changes if you want; I also want to add tests.

nicolo-ribaudo:
I opened an alternative PR: https://github.com/babel/babel/pull/14065

raqso (approved):

PR #11748, opened 2020-06-26
Support private properties for semi rule

simonkotwicz:
| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

Corresponding PR for eslint-plugin-babel repo: https://github.com/babel/eslint-plugin-babel/pull/194#issuecomment-649805154

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11748"></a>

### DIFF
--- a/eslint/babel-eslint-plugin/src/rules/semi.js
+++ b/eslint/babel-eslint-plugin/src/rules/semi.js
@@ -76,30 +76,38 @@ function report(context, node, missing) {
   });
 }
 
+function checkForSemicolon(node, context) {
+  const options = context.options[1];
+  const exceptOneLine = options && options.omitLastInOneLineBlock === true;
+
+  const sourceCode = context.getSourceCode();
+  const lastToken = sourceCode.getLastToken(node);
+
+  if (context.options[0] === "never") {
+    if (isUnnecessarySemicolon(context, lastToken)) {
+      report(context, node, true);
+    }
+  } else {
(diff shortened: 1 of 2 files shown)


eslint/babel-eslint-plugin/src/rules/semi.js
```diff
@@ -76,30 +76,38 @@ function report(context, node, missing) {
...
+        report(context, node, true);
+      }
+    }
+  }
+}
+
 export default ruleComposer.joinReports([
   rule,
   context => ({
     ClassProperty(node) {
```
JLHwung, line 106: @kaicataldo Does `matches` [selector](https://eslint.org/docs/developer-guide/selectors) support
```
":matches(ClassProperty, ClassPrivateProperty)"
```
?

kaicataldo: Yes! As an example, you can see this core rule: https://github.com/eslint/eslint/blob/master/lib/rules/func-call-spacing.js#L171

We should be able to combine both of these selectors into one, e.g. `"ClassProperty, ClassPrivateProperty"(node) {}`.

JLHwung: Good to know that `A, B` is supported, guess we should update https://eslint.org/docs/developer-guide/selectors#what-syntax-can-selectors-have.

Issue #10746, opened 2019-11-21
Babel 8 Release Plan


nicolo-ribaudo:
We plan to release a new major version in the beginning of 2020.

This release won't have all the migration pain which there was while migrating from Babel 5 to Babel 6 and then from Babel 6 to Babel 7. We plan to only make a few breaking changes, and provide an easy migration strategy for each of them. Because of this, you won't see ~80 prereleases like we did in for Babel 7.0.0, but we plan to only release a few of them.

Babel 8.0.0 will _only_ contain breaking changes: we will release a minor version the same day, containing all the bug fixes and new features that would otherwise be released in 8.0.0.

This document is still a work in progress, but you can already start by applying the suggested migration strategies to your own codebase.

- ### Compilation breaking changes
  - [Don't remove uninitialized class fields when using Flow/TS](https://github.com/babel/babel/issues/10039) (pr for flow: [#10120](https://github.com/babel/babel/pull/10120))
    - **Area**: Flow and TypeScript transforms
    - **Impact**: High (only for flow and TypeScript users)
    - **Migration**:
      Flow users can already update their code to use the flow comments syntax if they don't want to initialized to `undefined`. This is officially [recommended](https://github.com/facebook/flow/issues/6811#issuecomment-496808192) by the Flow team:
      ```js
      class A {
        foo: ?string; // initialized to undefined
        /*:: bar: number */ // type-only
      }
      ```
      
      TypeScript users can use the new `declare` syntax, which introduced in TypeScript 3.7 and Babel 7.7:
      ```js
      class A {
        foo: string | void; // initialized to undefined
        declare bar: number; // type-only
      }
      ```
      Note that while this syntax is enabled by default in `@babel/parser` when using the `typescript` plugin, if you are using `@babel/plugin-transform-typescript` or `@babel/preset-typescript` you must enable the `declareFields` option:
      
      ```json5
      {
        presets: [
          ["@babel/typescript", { "declareFields": true }]
        ]
      }
      ```
  - [Require `@babel/plugin-proposal-dynamic-import` when transforming `import()` to SystemJS](https://github.com/babel/babel/blob/78cd63d9cfcd96e6a151c58fed392c3ee757d861/packages/babel-plugin-transform-modules-systemjs/src/index.js#L183-L185)
    - **Area**: `@babel/plugin-transform-modules-systemjs`
    - **Impact**: Medium
[...]

PR #11751, opened 2020-06-26
Drop core-js 2 support

JLHwung:
| Fixed Issues?            | https://github.com/babel/babel/issues/10746#issuecomment-573402372
| Documentation PR Link    | 
| Any Dependency Changes?  |

This PR drops support of core-js 2 in `compat-data`, `preset-env` and `transform-runtime`. It should be reviewed commit-by-commit.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11751"></a>


packages/babel-preset-env/src/normalize-options.js
```diff
@@ -48,11 +46,7 @@ const getValidIncludesAndExcludes = (
   new Set([
     ...allPluginsList,
     ...(type === "exclude" ? modulePlugins : []),
-    ...(corejs
-      ? corejs == 2
-        ? [...Object.keys(corejs2Polyfills), ...defaultWebIncludes]
```
JLHwung, line 53: I like the idea of _not_ applying web polyfills for node-only targets. However these `defaultWebIncludes` are applied _unconditionally_. That is also why the core-js 3 support does not have `defaultWebIncludes`. I think we should bring back and properly filter out web-only polyfills for node-only targets. Surely it falls into a new PR.

packages/babel-preset-env/test/fixtures/debug/entry-no-corejs-shippedProposals/input.mjs
```diff
@@ -1 +1 @@
-import '@babel/polyfill';
+import 'core-js';
```
JLHwung, line 1: The CoreJS3EntryPlugin has thrown warning for `@babel/polyfill` imports and it does _not_ replace these imports to `core-js` submodules. I think this behaviour is acceptable and we can preserve it in Babel 8.

The test case is about when `core-js` is _not_ specified and `@babel/polyfill` is imported, it should be replaced via the usage plugin by a series of `core-js` submodule imports. Since we have also changed the `core-js` defaults to 3, we shall also revise this case to showcase that `import "core-js"` will be replaced, too.

packages/babel-preset-env/src/normalize-options.js
```diff
@@ -218,19 +212,19 @@ export function normalizeCoreJSOption(
   let rawVersion;
 
   if (useBuiltIns && corejs === undefined) {
-    rawVersion = 2;
+    rawVersion = "3.6";
```
JLHwung, line 215: We should also convey this change in the release plan, that if you specify `useBuiltIns` but leave `corejs` empty, it will now defaults to `3.6`.

packages/babel-preset-env/test/normalize-options.spec.js
```diff
@@ -87,6 +87,18 @@ describe("normalize-options", () => {
       });
     });
 
+    it("should throw removed option if corejs version is 2", () => {
+      [2, 2.1].forEach(corejs => {
+        ["entry", "usage"].forEach(useBuiltIns => {
+          expect(() =>
+            normalizeOptions.default({ useBuiltIns, corejs }),
+          ).toThrowError(
+            /Since Babel 8, the core-js@2 support has been dropped. Please use `corejs: "3.6"`/,
```
nicolo-ribaudo, line 96: Can we also mention that if they _really_ want `core-js@2`, they can use `babel-plugin-polyfill-corejs2` from https://github.com/babel/babel-polyfills?

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -35,15 +34,21 @@ export default declare((api, options, dirname) => {
 
   const corejsVersion = rawVersion ? Number(rawVersion) : false;
 
-  if (![false, 2, 3].includes(corejsVersion)) {
+  if (corejsVersion === 2) {
     throw new Error(
-      `The \`core-js\` version must be false, 2 or 3, but got ${JSON.stringify(
+      "Since Babel 8, the core-js 2 support has been dropped. Please use 'corejs: 3'.",
```
nicolo-ribaudo, line 39: Also let's mention the separate plugin here.

packages/babel-plugin-transform-runtime/src/index.js
```diff
@@ -35,15 +34,21 @@ export default declare((api, options, dirname) => {
...
+      `The \`core-js\` version must be false or 3, but got ${JSON.stringify(
         rawVersion,
       )}.`,
     );
   }
 
-  if (proposals && (!corejsVersion || corejsVersion < 3)) {
+  if (proposals && !corejsVersion) {
     throw new Error(
       "The 'proposals' option is only supported when using 'corejs: 3'",
```
nicolo-ribaudo, line 53: Maybe we should remove the version number, since `proposals` is now supported by every supported `core-js` version?

```suggestion
      "The 'proposals' option is only supported when using core-js",
```

nicolo-ribaudo:
The other breaking change I would like to see is to drop `@babel/runtime-corejs3` and directly inject imports to `core-js-pure`.

Note that if we internally make `@babel/runtime` and `useBuiltIns` delegate to https://github.com/babel/babel-polyfills (and I think that we should, so that we don't have to maintain the same duplicated thing) it will happen automatically.

PR #11752, opened 2020-06-27
[regression] Don't validate file.comments in `@babel/types`

nicolo-ribaudo:
| Fixed Issues?            | Fixes https://github.com/babel/babel/pull/11687#discussion_r446474079, https://github.com/facebook/jest/issues/10208
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

In https://github.com/babel/babel/pull/11687 we introduced a new valid check, but it breaks Jest. Let's defer it to Babel 8.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11752"></a>

### DIFF
--- a/packages/babel-types/src/definitions/core.js
+++ b/packages/babel-types/src/definitions/core.js
@@ -281,8 +281,10 @@ defineType("File", {
     },
     comments: {
       validate: !process.env.BABEL_TYPES_8_BREAKING
-        ? Object.assign(() => {}, { each: { type: "Comment" } })
-        : assertEach(assertNodeType("Comment")),
+        ? Object.assign(() => {}, {
+            each: { oneOfNodeTypes: ["CommentBlock", "CommentLine"] },
+          })
+        : assertEach(assertNodeType("CommentBlock", "CommentLine")),
       optional: true,
     },
     tokens: {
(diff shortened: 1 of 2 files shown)


packages/babel-types/src/definitions/core.js
```diff
@@ -280,7 +280,9 @@ defineType("File", {
       validate: assertNodeType("Program"),
     },
     comments: {
-      validate: assertEach(assertNodeType("Comment")),
+      validate: !process.env.BABEL_TYPES_8_BREAKING
+        ? Object.assign(() => {}, { each: { type: "Comment" } })
+        : assertEach(assertNodeType("Comment")),
```
zxbodya, line 285: one thing, I noticed, when looking into referred issue:

`assertNodeType("Comment")` results in correct typescript types, but actual check is not correct instead it should be `assertNodeType("CommentBlock", "CommentLine")`

packages/babel-types/test/regressions.js
```diff
@@ -0,0 +1,9 @@
+import * as t from "../lib";
+
+describe("regressions", () => {
+  it("jest .toMatchInlineSnapshot usef 'Line' for comments", () => {
```
JLHwung, line 4:
```suggestion
  it("jest .toMatchInlineSnapshot used 'Line' for comments", () => {
```

PR #11755, opened 2020-06-27
Allow unknown/any in TS catch clause param

existentialism:
| Fixed Issues?            |
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11755"></a>

https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#unknown-on-catch


packages/babel-parser/src/plugins/typescript/index.js
```diff
@@ -2667,4 +2669,26 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+        switch (type.typeAnnotation.type) {
+          case "TSAnyKeyword":
+          case "TSUnknownKeyword":
+            param.type = type;
+            break;
+
+          default:
+            this.raise(
+              type.typeAnnotation.start,
+              TSErrors.UnexpectedCatchClauseParamType,
```
nicolo-ribaudo, line 2687: I don't think that we can report this error at parse time, consider this example:

```js
type A = any;

try {

} catch (e: A) {}
```

[Playground Link](https://www.typescriptlang.org/play?useDefineForClassFields=true&ts=4.0.0-beta&ssl=1&ssc=1&pln=5&pc=18#code/C4TwDgpgBAglC8UCGA7EBuAsAKB8ATiFAN444C+UAxksFQBZQAUEAXLAJQnlA)

existentialism: Yeah. We could at least check against primitive types?

JLHwung: > check against primitive types

It is open ended: Should we throw `number | string`? What about `number | any`?

I don't think the primary types on caught param is popular, compared to

```js
try { } catch (e: Error) {}
```

but we don't know if `Error` is `any` so we can't throw at parsing time. I tend to leave it as wont-fix as it falls into "runtime error" in the TS type-checking phrase.

PR #11777, opened 2020-07-01
chore: build standalone once in prepublish step

JLHwung:
| Fixed Issues?            | Removed duplicate build-standalone step in `make publish`

Currently when `make prepublish` runs, `@babel/standalone` will be built twice

1. `make prepublish` => `make build` => `make build-standalone`
2. `npm run-script prepublishOnly` => `make prepublish-build-standalone`

The first step skips minification while the second one _includes_ it. This PR removes the `scripts` hook in `@babel/standalone` and run `prepublish-build-standalone` when `prepublish` recipe is executed.


JLHwung:
✅ As we can see in https://app.circleci.com/pipelines/github/babel/babel/3237/workflows/0018a857-7362-4563-a44e-6da6443c9d07/jobs/25052 that the published `@babel/standalone` contains both minified and raw versions. This behaviour is not changed.

PR #11779, opened 2020-07-02
Update Yarn 2 to 2.0.0-rc.36

JLHwung:

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11779"></a>

Note that this PR targets to `next-8-dev`.


.yarnrc.yml
```diff
@@ -9,5 +9,6 @@ plugins:
 unsafeHttpWhitelist:
   - localhost
 
-# This path should always match the one in .yarnrc
-yarnPath: .yarn/releases/yarn-2.0.0-rc.33.js
+yarnPath: .yarn/releases/yarn-2.0.0-rc.36.js
+
+httpTimeout: 120000
```
JLHwung, line 14: `httpTimeout` is not available in Yarn 2.0.0-rc.33, so I update to the latest version.

nicolo-ribaudo:
Oh did you notice a timed out build with yarn 2? I dropped it because I only saw the problem with yarn 1.

JLHwung:
@nicolo-ribaudo Oh I thought it is dropped because it is not supported on yarn 2 rc.33. I am happy to remove `httpTimeout` if things are going well on gitpod.

nicolo-ribaudo:
I didn't even notice that there wasn't a way to increment the timeout in the RC I was using 😅

If you saw a gitpod build failing then let's merge this PR, otherwise we can wait and see if the problem appears in the future.

JLHwung:
@nicolo-ribaudo This PR has been revised to only update Yarn 2 from `rc.33` to `rc/36`, the lockfile format is changed a bit.

PR #11781, opened 2020-07-02
chore: improve Yarn 2 cache on CI

JLHwung:
| Fixed Issues?            | Add `@actions/cache` to our GitHub coverage workflow

Based on example: https://github.com/actions/cache/blob/master/examples.md#node---yarn

The cache is working as expected. We can compare the Yarn  fetch time between `main` and this PR.

| [main](https://github.com/babel/babel/pull/11962/checks?check_run_id=988776208#step:4:57) | [This PR](https://github.com/babel/babel/pull/11781/checks?check_run_id=988851094#step:6:43) |
| --- | --- |
| 2.4m | 0.91s |

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11781"></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,8 @@ jobs:
             ${{ runner.os }}-yarn-
       - name: Update Yarn cache
         if: steps.yarn-cache.outputs.cache-hit != 'true'
+        env:
+          YARN_ENABLE_SCRIPTS: false # disable post-install scripts
         run: |
           yarn install --immutable
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,10 @@
-name: Node CI
+name: CI
 
 on: [push, pull_request]
 
 jobs:
-  validate-yarn-cache:
+  prepare-yarn-cache:
+    name: Prepare Cache
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
@@ -28,8 +29,9 @@ jobs:
         run: |
           yarn install --immutable
 
-  test:
-    needs: validate-yarn-cache
+  test-coverage:
+    name: Test on Node.js Latest
+    needs: prepare-yarn-cache
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
@@ -52,7 +54,8 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
 
   legacy-node-version-test:
-    needs: validate-yarn-cache
+    name: Test on Node.js
+    needs: prepare-yarn-cache
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,8 +40,6 @@ jobs:
         with:
           path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
           key: ${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-yarn-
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage.yml
```diff
@@ -12,6 +12,16 @@ jobs:
         uses: actions/setup-node@v2-beta
         with:
           node-version: "*"
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
```
nicolo-ribaudo, line 17: Does this also work with Yarn 2?

JLHwung: It works on Yarn 2 only: https://github.com/actions/cache/blob/main/examples.md#node---yarn-2

.github/workflows/ci.yml
```diff
@@ -3,13 +3,45 @@ name: Node CI
 on: [push, pull_request]
 
 jobs:
+  validate-yarn-cache:
```
JLHwung, line 6: This step ensures the test jobs later will always be served with proper yarn cache.

arcanis: I'm not sure it'll be in the default distribution. One problem with the v1 was that installs could be done in various slightly different ways that made the surface for unforeseen edge cases much larger. I'd like to avoid that with one standardized path as much as possible.

That being said, a plugin to do this would be trivial, so we probably can find something eventually, once we've figured out a good way to develop/host such plugins (it's already possible, but there's no comprehensive documentation).

merceyz: You can sort of get this if you set this specific job to install using PnP, which links so fast you shouldn't really notice.
Set the env variable `YARN_NODE_LINKER=pnp`

JLHwung: @merceyz We opt-out of PnP because Flow does not support PnP: https://github.com/yarnpkg/berry/issues/634#issuecomment-565582736.

merceyz: Indeed, but for this specific job, validate-yarn-cache, its job is to generate a valid cache, not run Flow, so installing using PnP would do that without the cost of linking. The cache for PnP and node-modules is the same.

JLHwung: Awesome: https://github.com/babel/babel/pull/11781/checks?check_run_id=991006280#step:6:27 The pnp linker costs only 4s compared to 20s of `node-modules`.

merceyz: You can reduce it even more if you also set `YARN_ENABLE_SCRIPTS=false`, you can do this since you don't need to run postinstalls to get a valid cache

JLHwung: Impressive: https://github.com/babel/babel/pull/11781/checks?check_run_id=991025473#step:6:26 3s on the linker stage now. @merceyz Thank you!

.github/workflows/ci.yml
```diff
@@ -3,13 +3,46 @@ name: Node CI
...
```
nicolo-ribaudo, line 6: Nit: maybe this should be called `prepare-yarn-cache`, since "validate" implies that it would fail the build if, for example, the cache is old.

nicolo-ribaudo (approved):

.github/workflows/ci.yml
JLHwung: I have added a `"name"` section in https://github.com/babel/babel/pull/12002/files/d484dbba30163e84d5b8cc610efea652b50c6a67..303ec4a3229d17092d770e01cbe6a77bcc72a075#diff-e9f950f17198d3d5e3122a44230a09b9R7

I can port the name changes here and also rename the job id.

PR #11782, opened 2020-07-02
chore: refine yarn cache config

JLHwung:

The Yarn Cache is not working properly because the cache directory of Yarn 1 is _not_ `~/.yarn-cache`, as we can see from this [job log](https://app.circleci.com/pipelines/github/babel/babel/3253/workflows/ef0651e9-e19b-4e0d-b9f5-e26e4de7029c/jobs/25141/parallel-runs/0/steps/0-102), the cache is only 32B.

Ideally we should run `yarn cache dir` and use this value as the paths that we should cache. However Circle-CI does not support template on cache paths.

Here is comparison of `yarn install` time between this PR and `main`.

| [main](https://app.circleci.com/pipelines/github/babel/babel/3253/workflows/ef0651e9-e19b-4e0d-b9f5-e26e4de7029c/jobs/25141/parallel-runs/0/steps/0-104) | [This PR](https://app.circleci.com/pipelines/github/babel/babel/3258/workflows/b4d98cd1-1b70-4bb8-8921-cf2ba12127d8/jobs/25178/parallel-runs/0/steps/0-104) |
| --- | --- |
| 39.76s | 6.03s |

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11782"></a>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,10 +4,12 @@ aliases:
   - &restore-node-modules-cache
     keys:
       - v1-yarn-deps-{{ checksum "yarn.lock" }}
+      - v1-yarn-deps-
 
   - &restore-yarn-cache
     keys:
-      - v1-yarn-cache
+      - v1-yarn-cache-{{ checksum "yarn.lock" }}
+      - v1-yarn-cache-
 
   - &save-node-modules-cache
     paths:
@@ -16,8 +18,8 @@ aliases:
 
   - &save-yarn-cache
     paths:
-      - ~/.yarn-cache
-    key: v1-yarn-cache
+      - "{{ .Environment.YARN_CACHE_FOLDER }}"
+    key: v1-yarn-cache-{{ checksum "yarn.lock" }}
 
   - &artifact_babel
     path: ~/babel/packages/babel-standalone/babel.js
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -4,10 +4,12 @@ aliases:
   - &restore-node-modules-cache
     keys:
       - v1-yarn-deps-{{ checksum "yarn.lock" }}
+      - v1-yarn-deps-
 
   - &restore-yarn-cache
     keys:
-      - v1-yarn-cache
+      - v1-1-yarn-cache-{{ checksum "yarn.lock" }}
```
JLHwung, line 11: The `v2-` has been used as prefix in Babel 8. So we resort to `v1-1`.

.circleci/config.yml
```diff
@@ -4,10 +4,12 @@ aliases:
   - &restore-node-modules-cache
     keys:
       - v1-yarn-deps-{{ checksum "yarn.lock" }}
+      - v1-yarn-deps-
```
nicolo-ribaudo, line 7: Why do we need this key without the hash?

JLHwung: Good question!

It is intended to fetch the latest cache generated successfully by any branch if the `yarn.lock` is changed. Chances are that there could be some cache contents that can still be reused by `yarn`, i.e. some libraries that are not still resolved by the changed lock file.

See also https://circleci.com/docs/2.0/caching/#restoring-cache

PR #11789, opened 2020-07-03
Import per-function lodash modules

### LINKED ISSUE #11726: @babel/core - Get rid of (only 4 instances) lodash to fix security issue
## Feature Request

I would like to remove lodash as a dependency in @babel/core. There has been a very long period where lodash had the chance to fix a prototype pollution vulnerability, and even after long period of security companies advicing privately, the security report was published publicly. And still, the maintainer is openly ignoring the security problem. https://github.com/lodash/lodash/issues/4738

This has lead many dependents to get off lodash, rightfully so. Security should not be ignored for this long, especially with this many weekly downloads.

@babel/core is a pretty huge dependent of lodash, but there are actually only 4 instances where lodash is used:
- escapeRegExp --> ./src/config/pattern-to-regex.js
- escapeRegExp --> ./test/config-chain.js
- sortBy --> ./src/transformation/block-hoist-plugin.js
- cloneDeep --> ./src/transformation/normalize-file.js

Safe to say, lodash is just helpers, and you never actually NEED them, you can just use regular JavaScript or create your own helper if you need it. Reducing dependencies is always a smart security/risk move.

If the babel maintainers agree, I would be happy to try and provide a pull request to replace the instances of lodash in @babel/core so we can get rid of it, and subsequently fix the security vulnerability that so many people face who are using @babel/core. 

Let me know!

jayaddison:
| Fixed Issues?            | Fixes #11726 

Reading [some](https://github.com/babel/babel/issues/11726#issuecomment-653098556) [comments](https://github.com/babel/babel/issues/11726#issuecomment-653266404) on #11726, it sounds like it should be possible to replace the full `lodash` dependency included in many of the `babel` packages with smaller, per-function `lodash` function modules.

To explain that differently: where code previously used `import clone from "lodash/clone"` from the full [`lodash`](https://www.npmjs.com/package/lodash) package, instead it should be possible to `import clone from "lodash.clone"` using the smaller [`lodash.clone`](https://www.npmjs.com/package/lodash.clone) package.

There are some concerns and gotchas here:

- :warning: Package versions for the per-function `lodash` modules vary, and have **not** yet been analyzed for feature-parity against the current `lodash` release (4.17.15)
- :information_source: The `lodash.extends` package does not exist, but `lodash.assignIn` is an [alias](https://github.com/lodash/lodash/blob/4.17.15/dist/lodash.js#L16678) for `lodash.extends` so we can use the corresponding package
- :information_source: It was necessary to revert https://github.com/babel/babel/pull/11376 to build and test this locally, and local test results still look patchy at the time of writing

Any ideas for checking the freshness and correctness of the per-function packages would be appreciated; performing that verification should be considered a blocker for this approach.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11789"></a>


packages/babel-types/package.json
```diff
@@ -17,8 +17,10 @@
   "types": "lib/index.d.ts",
   "dependencies": {
     "@babel/helper-validator-identifier": "^7.10.4",
-    "lodash": "^4.17.13",
-    "to-fast-properties": "^2.0.0"
```
JLHwung, line 21: `to-fast-properties` is incorrectly removed.

packages/babel-cli/src/babel/util.js
```diff
@@ -2,7 +2,7 @@
 
 import readdirRecursive from "fs-readdir-recursive";
 import * as babel from "@babel/core";
-import includes from "lodash/includes";
+import includes from "lodash.includes";
```
existentialism, line 5: Can we replace this with Array.includes and eliminate the dep altogether?

https://github.com/jayaddison/babel/blob/dependencies/lodash-submodules/packages/babel-cli/src/babel/util.js#L50

packages/babel-types/package.json
jayaddison: Nice catch, thanks!

packages/babel-cli/src/babel/util.js
jayaddison: That seems reasonable - there are a few other places where `includes` is in use; are there any general rules for when it's safe to use the prototype function?  `babel-traverse` looks like it includes the `Array` type spec, if that's relevant:

https://github.com/jayaddison/babel/blob/dependencies/lodash-submodules/packages/babel-traverse/src/index.js#L93

existentialism: Should be safe to replace all the cases being used on an array, since node support for .includes is v6!

packages/babel-cli/src/babel/options.js
```diff
@@ -4,7 +4,7 @@ import fs from "fs";
 
 import commander from "commander";
 import { version } from "@babel/core";
-import uniq from "lodash/uniq";
+import uniq from "lodash.uniq";
```
kaicataldo, line 7: Is there any chance we can use `Array.from(new Set([]))` and get rid of this dep?

packages/babel-plugin-transform-block-scoping/package.json
```diff
@@ -14,7 +14,8 @@
   "main": "lib/index.js",
   "dependencies": {
     "@babel/helper-plugin-utils": "^7.10.4",
-    "lodash": "^4.17.13"
+    "lodash.assignin": "^4.2.0",
+    "lodash.values": "^4.3.0"
```
kaicataldo, line 18: Can we use `Object.values()` and get rid of this dep?

packages/babel-types/src/builders/builder.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import loClone from "lodash/clone";
+import loClone from "lodash.clone";
```
kaicataldo, line 2: Can we use `Object.assign()` or spread operator (`const clone = { ...obj }` or `const clone = [...arr]`) since this is a shallow clone? That would allow us to remove this dep.

JLHwung:
I suggest we replace some of lodash usage by builtin functions, which should fall into a separate PR. After that PR is merged, we can replaced the more-than-trivial lodash usage by per-function modules.

https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore may be helpful.

jayaddison:
> I suggest we replace some of lodash usage by builtin functions, which should fall into a separate PR. After that PR is merged, we can replaced the more-than-trivial lodash usage by per-function modules.

That sounds like a good approach to me.  I might have time to try that over the weekend; either way, anyone should feel free to create the builtin-replacement-PR and I'll merge `main` back in here and resolve (expected) conflicts after it's merged.

ark120202:
Note that per-module lodash packages apparently won't be supported in the future, see https://github.com/lodash/lodash/issues/3838

packages/babel-cli/src/babel/options.js
jayaddison: That looks feasible, yep - thanks!

packages/babel-types/src/builders/builder.js
jayaddison: Although `uniq` and `values` look fairly safe to replace (see https://github.com/babel/babel/pull/11790), `clone` seems a little more subtle (for example, due to the fact that `lodash/clone` handles different types automatically, and also handles basic types).

I haven't yet applied any changes there, although with a bit more scrutiny and care it's probably possible to remove the usage of `lodash/clone` here, yep.

jayaddison:
Ah, thanks @ark120202.  That could pour a bit of cold water on this plan.

packages/babel-plugin-transform-block-scoping/package.json
jayaddison: NB: I attempted this in #11790, but it seems that `Object.values` is only fully supported from Node 7.x onwards.

packages/babel-helper-define-map/package.json
```diff
@@ -15,6 +15,6 @@
   "dependencies": {
     "@babel/helper-function-name": "^7.10.4",
     "@babel/types": "^7.10.4",
-    "lodash": "^4.17.13"
+    "lodash.has": "^4.5.2"
```
jayaddison, line 18: `lodash.has` is similar to `lodash.get`, and `lodash.get` appears in [You-Dont-Need-Lodash-Underscore](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore).

That said, the [YDNLU native JavaScript replacement](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_get) is relatively complex (9 lines of non-trivial code including regular expressions) and generates an error when key paths do not exist.  On balance it may make sense to maintain readability for now, although that's just my two cents and I'd welcome feedback.

merceyz:
As @ark120202 mentioned the per method packages are deprecated and wont be updated in the future, https://lodash.com/per-method-packages, using them also results in larger installs as it's most likely some other dependency has `lodash` as one of its dependencies, meaning the main lodash package and the per method packages will be installed

packages/babel-cli/src/babel/file.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 
 import convertSourceMap from "convert-source-map";
-import defaults from "lodash/defaults";
+import defaults from "lodash.defaults";
```
existentialism, line 4: We should be able to remove `defaults`, yeah?

```
const res = await util.transform(
      cliOptions.filename,
      code,
      {
        ...babelOptions,
        sourceFileName: "stdin",
      },
);
```

jayaddison: Possibly! - we'd want to be careful about the argument ordering to the spread operator:

```
> var defaults = {'option': 'valueFromDefault'};
> var cliArgs = {'option': 'commandLineArgument'};
> {...defaults, ...cliArgs}
{ option: 'commandLineArgument' }
> {...cliArgs, ...defaults}
{ option: 'valueFromDefault' }
```

Also a subtle difference is that `lodash.defaults` mutates the first argument and returns it as a value, whereas the spread operator returns a new shallow clone; that's probably fine and safe in most cases, but worth being careful about.

nicolo-ribaudo: In our case the mutation isn't a problem, since the first object is always defined inline in the call.

packages/babel-helper-fixtures/src/index.js
```diff
@@ -1,6 +1,6 @@
-import cloneDeep from "lodash/cloneDeep";
-import clone from "lodash/clone";
-import extend from "lodash/extend";
+import cloneDeep from "lodash.clonedeep";
+import clone from "lodash.clone";
```
nicolo-ribaudo, line 2: We could avoid `clone`:

```js
      options: { ...rootOpts },
```

packages/babel-helper-define-map/package.json
nicolo-ribaudo: Let's keep it, since we are using it with a deep path.

packages/babel-helper-fixtures/src/index.js
```diff
@@ -1,6 +1,6 @@
...
+import extend from "lodash.assignin";
```
nicolo-ribaudo, line 3: `Object.assign`?

packages/babel-cli/src/babel/file.js
jayaddison: Okie doke, thanks - https://github.com/babel/babel/pull/11797 contains an attempt to clean up the use of `lodash.defaults` from the codebase (including this callsite).

packages/babel-helper-fixtures/src/index.js
jayaddison: Thanks for the confirmation @nicolo-ribaudo - this is now removed in https://github.com/babel/babel/pull/11811, along with one other `clone` callsite

packages/babel-helper-fixtures/src/index.js
jayaddison: `lodash.extend` replaced with `Object.assign` here (and for one other remaining callsite) in https://github.com/babel/babel/pull/11812

packages/babel-types/src/builders/builder.js
jayaddison: (removal of `lodash/clone` now handled via https://github.com/babel/babel/pull/11811)

packages/babel-core/src/config/pattern-to-regex.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import path from "path";
-import escapeRegExp from "lodash/escapeRegExp";
```
jayaddison, line 3: It might be possible to replace `lodash/escapeRegExp` with a function [provided by the MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Escaping), although I don't know whether it'd make sense to replicate that per-babel-package or whether it should go into a utility/common package if so.

coreyfarrell: Could we instead consider replacing it with [escape-string-regexp](https://www.npmjs.com/package/escape-string-regexp).  That module is well tested, not being deprecated and gets 19 million weekly downloads (it's not below scrutiny).

TrySound: Keep in mind it requires major bump
https://github.com/sindresorhus/escape-string-regexp/blob/master/.travis.yml

coreyfarrell: Ah sorry for missing that, I guess this suggestion would only apply to babel 8 then.

TrySound:
lodash/cloneDeep can be replaced with great [klona](https://github.com/lukeed/klona). Node v8 is required, will be a breaking change.

TrySound:
`lodash/has` perhaps can be replaced with [dlv](https://github.com/developit/dlv) which is more like `lodash/get`.

TrySound:
[is-plain-object](https://github.com/jonschlinkert/is-plain-object) is very small replacement for `lodash/isPlainObject`. Supports node 0.10.

jayaddison:
> lodash/cloneDeep can be replaced with great [klona](https://github.com/lukeed/klona). Node v8 is required, will be a breaking change.

Thanks @TrySound - do you have a sense for how `klona` compares to the widely-deployed [`clone`](https://www.npmjs.com/package/clone) library?

A few notes on https://github.com/lukeed/klona/issues/13 make it sound like there's functionality yet to be implemented, and `clone` seems to provide support for some of those via [API options](https://github.com/pvorb/clone#api).

jayaddison:
> `lodash/has` perhaps can be replaced with [dlv](https://github.com/developit/dlv) which is more like `lodash/get`. Node v6 is supported.

Wow, `dlv` has an impressive commitment towards dependency size reduction :)

I don't know what to suggest here.  There's certainly an argument for keeping the bundled dependencies small, but I also think there's a strong safety argument for using libraries that are stable and avoid JavaScript tricks that could potentially lead to unexpected gotchas/situations in future (in particular the use of an `undef` argument seems like a neat size-reduction measure, but potentially problematic in future).

The lowest-risk option might be not to change the existing implementation (i.e. continue to use `lodash/has`) until there's a compelling reason to (and yep, I realize that'd block the goal of removing `lodash` completely, but it'd be better to maintain software safety and quality rather than risk user impact).

jayaddison:
I think I'm going to close this and a few of the related `lodash`-replacement pull requests since the original concerns regarding `lodash` maintenance seem to have subsided (at least for now), and so the PRs just exist as extra overhead for everyone.

Generally I'd delete the branches and then also the source repository once done; if I remember this makes it tricky to retrieve the commits later, so.. does anyone have opinions about whether that's reasonable to do?  (I don't think it'd be super challenging for someone to rewrite these if needed -- and as far as I know GitHub will still show the PR diff so it'd be possible to recreate them that way too).

yumetodo:
@jayaddison AFAIK, After you delete the `jayaddison:dependencies/lodash-submodules`, the `pull/11789/head` branch will still exist.

jayaddison:
Thanks @yumetodo - let's hope that should be the case; I seem to remember running into problems even with the pr refs previously, but forget exactly what happened.  Probably user error :)

jayaddison:
All good: here's how I confused myself previously: https://github.com/getpelican/pelican-themes/pull/659#issuecomment-566149036 (it was possible to recover by opening a new pull request)

nicolo-ribaudo:
Hey thanks for closing these PRs: we usually keep them open for years until someone from the team finally reviews and merges them, but since this is not necessary anymore (we might still remove lodash in the future, but it's not urgent) I appreciate that you avoided keeping them around forever :sweat_smile:

PR #11790, opened 2020-07-04
Reduce dependency on lodash functions: includes, uniq, repeat, isinteger

jayaddison:
| Fixed Issues?            | Relates to #11789 

This change incorporates feedback from #11789 - in particular, the following `lodash` functions are replaced with JavaScript built-in functions:

- `lodash/includes` is replaced by `Array.prototype.includes`
- ~~`lodash/values` is replaced by `Object.values`~~ `Object.values` is fully-supported [from Node 7.x](https://node.green/#ES2017-features-Object-static-methods-Object-values)
- `lodash/uniq` is replaced by `Array.from(new Set(...))`
- `lodash.repeat` is replaced by `String.prototype.repeat`
- `lodash.isinteger` is replaced by `Number.isInteger`

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11790"></a>

### DIFF
--- a/packages/babel-helper-transform-fixture-test-runner/src/index.js
+++ b/packages/babel-helper-transform-fixture-test-runner/src/index.js
@@ -351,7 +351,12 @@ export default function (
   const suites = getFixtures(fixturesLoc);
 
   for (const testSuite of suites) {
-    if (suiteOpts.ignoreSuites.includes(testSuite.title)) continue;
+    if (
+      suiteOpts.ignoreSuites &&
+      suiteOpts.ignoreSuites.includes(testSuite.title)
+    ) {
+      continue;
+    }
 
     describe(name + "/" + testSuite.title, function () {
       jest.addMatchers({
@@ -360,8 +365,9 @@ export default function (
 
       for (const task of testSuite.tests) {
         if (
-          suiteOpts.ignoreTasks.includes(task.title) ||
-          suiteOpts.ignoreTasks.includes(testSuite.title + "/" + task.title)
+          suiteOpts.ignoreTasks &&
+          (suiteOpts.ignoreTasks.includes(task.title) ||
+            suiteOpts.ignoreTasks.includes(testSuite.title + "/" + task.title))
         ) {
           continue;
         }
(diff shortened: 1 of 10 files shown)


packages/babel-helper-transform-fixture-test-runner/src/index.js
```diff
@@ -352,7 +351,12 @@ export default function (
   const suites = getFixtures(fixturesLoc);
 
   for (const testSuite of suites) {
-    if (includes(suiteOpts.ignoreSuites, testSuite.title)) continue;
+    if (
+      suiteOpts.ignoreSuites &&
+      suiteOpts.ignoreSuites.includes(testSuite.title)
```
existentialism, line 356:
```suggestion
      suiteOpts.ignoreSuites?.includes(testSuite.title)
```

nicolo-ribaudo:
> lodash/values is replaced by Object.values Object.values is fully-supported from Node 7.x

After that this PR and the other one are merged to `main`, I will also merge them to the `next-8-dev` branch: Babel 8 will require Node.js 8+, so we can then remove other lodash modules.

packages/babel-helper-transform-fixture-test-runner/src/index.js
jayaddison: Thanks!  I *think* this also means we can get rid of the `suiteOpts.ignoreSuites` condition on the previous line?  (I haven't used the optional chaining operator much).

I've pushed a change that applies that simplification here and also in the `ignoreTasks` block a few lines further down.

packages/babel-traverse/src/index.js
```diff
@@ -90,7 +89,7 @@ traverse.hasType = function (
   blacklistTypes: Array<string>,
```
nicolo-ribaudo, line 89: Could you mark this type as nullable? You correctly used `blacklistTypes?.includes` because we don't always pass this parameter, but it should be reflected in the signature:

```suggestion
  blacklistTypes?: Array<string>,
```

hzoo:
Awesome!

I think for `Object.values` we can just do this instead? https://github.com/babel/babel/pull/11790/commits/9fd3679d6db03066daee09fad0050e5292a32aa1#diff-11a6b6c946b2660882f17e01aa180067L548

```diff
- const args = values(outsideRefs).map(id => t.cloneNode(id));
+ const args = Object.keys(outsideRefs).map((ref) => t.cloneNode(outsideRefs[ref]));
```

jayaddison:
Yep, thanks - good idea!  See follow-up #11798 for a replacement of `lodash.values`.

PR #11791, opened 2020-07-04
babel-traverse: prefer clearer, reduced-bias option naming

jayaddison:
| Documentation PR Link    | This babel-traverse option appears undocumented currently

This proposed change attempts to clarify the naming of a `babel-traverse` option (noticed during development of #11790) by using language that aims to be clearer for non-native English speakers and has reduced modern connotations of bias.

Since the option doesn't appear to be widely used ([ref](https://github.com/search?p=3&q=babel-traverse+blacklist+filename%3A%2A.js&type=Code)) and doesn't appear to be documented([ref](https://babeljs.io/docs/en/babel-traverse), [ref](https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#babel-traverse)), this changeset doesn't attempt to maintain backwards compatibility, although a release note to indicate that it is a breaking change would be useful.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11791"></a>

### DIFF
--- a/packages/babel-traverse/src/path/context.js
+++ b/packages/babel-traverse/src/path/context.js
@@ -51,9 +51,9 @@ export function _call(fns?: Array<Function>): boolean {
   return false;
 }
 
-export function isBlacklisted(): boolean {
-  const blacklist = this.opts.blacklist;
-  return blacklist && blacklist.indexOf(this.node.type) > -1;
+export function isDenylisted(): boolean {
+  const denylist = this.opts.denylist;
+  return denylist && denylist.indexOf(this.node.type) > -1;
 }
 
 export function visit(): boolean {
@@ -61,7 +61,7 @@ export function visit(): boolean {
     return false;
   }
 
-  if (this.isBlacklisted()) {
+  if (this.isDenylisted()) {
     return false;
   }
 
--- a/packages/babel-traverse/src/path/context.js
+++ b/packages/babel-traverse/src/path/context.js
@@ -52,9 +52,12 @@ export function _call(fns?: Array<Function>): boolean {
 }
 
 export function isDenylisted(): boolean {
-  const denylist = this.opts.denylist;
+  const denylist = this.opts.denylist ?? this.opts.blacklist;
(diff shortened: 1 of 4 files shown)


packages/babel-traverse/src/path/context.js
```diff
@@ -51,17 +51,17 @@ export function _call(fns?: Array<Function>): boolean {
   return false;
 }
 
-export function isBlacklisted(): boolean {
-  const blacklist = this.opts.blacklist;
-  return blacklist && blacklist.indexOf(this.node.type) > -1;
+export function isDenylisted(): boolean {
+  const denylist = this.opts.denylist;
```
nicolo-ribaudo, line 55: Even if they are not documented, we should try to keep backwards compatibility.

```suggestion
  const denylist = this.opts.denylist ?? this.opts.blacklist;
```

packages/babel-traverse/src/visitors.js
```diff
@@ -275,7 +275,7 @@ function shouldIgnoreKey(key) {
   if (key === "enter" || key === "exit" || key === "shouldSkip") return true;
 
   // ignore other options
-  if (key === "blacklist" || key === "noScope" || key === "skipKeys") {
+  if (key === "denylist" || key === "noScope" || key === "skipKeys") {
```
nicolo-ribaudo, line 278:
```suggestion
  if (
    key === "denylist" ||
    key === "noScope" ||
    key === "skipKeys" ||
    // TODO: Remove in Babel 8
    key === "blacklist"
  ) {
```

packages/babel-traverse/src/path/context.js
```diff
@@ -51,17 +51,17 @@ export function _call(fns?: Array<Function>): boolean {
...
+  return denylist && denylist.indexOf(this.node.type) > -1;
 }
```
nicolo-ribaudo, line 57:
```suggestion
}

// TODO: Remove in Babel 8
export { isDenylisted as isBlacklisted };
```

PR #11797, opened 2020-07-06
Replace lodash 'defaults' usage with ES6 Spread initializer

jayaddison:

This change is a follow-up to #11790 and proposes removal of usages of [`lodash/defaults`](https://lodash.com/docs/4.17.15#defaults) from the codebase, to be replaced by the ES6 [spread operator](https://tc39.es/proposal-object-rest-spread/).

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11797"></a>


jayaddison:
:warning: Nope, this isn't ready as-is; the ordering of arguments are still a problem in at least one case, I think.

jayaddison:
(...and perhaps even the logic around `undefined` property handling via `lodash.defaults` could make it hard/impossible to simply drop-in the spread operator as a replacement)

packages/babel-cli/src/babel/file.js
```diff
@@ -177,22 +170,17 @@ export default async function ({
...
+            sourceFileName: sourceFilename,
+            // Since we're compiling everything to be merged together,
+            // "inline" applies to the final output file, but not to the individual
+            // files being concatenated.
+            sourceMaps:
+              babelOptions.sourceMaps === "inline"
+                ? true
+                : babelOptions.sourceMaps,
+            ...babelOptions,
+          });
```
jayaddison, line 183: This change is problematic too, and may be the cause of some test failures.

packages/babel-cli/src/babel/dir.js
```diff
@@ -48,15 +47,10 @@ export default async function ({
...
-      const res = await util.compile(
-        src,
-        defaults(
-          {
-            sourceFileName: slash(path.relative(dest + "/..", src)),
-          },
-          babelOptions,
-        ),
-      );
+      const res = await util.compile(src, {
```
existentialism, line 50: Shouldn't it be the other way around? Since `defaults`:

> Once a property is set, additional values of the same property are ignored.

jayaddison: I'm finding it a little tricky to reason about the differences - in one way, yep, it seems like it may simply be a case of reversing the argument order.

As I understand it though, the spread operator will *always* use the keys discovered in later arguments to override the contents of previously-declared keys.

So `sourceFileName` from the default here would always end up overriding any values provided in `babelOptions`.  Are they always guaranteed to by the defaults and non-user-supplied?

packages/babel-cli/src/babel/file.js
```diff
@@ -127,16 +127,10 @@ export default async function ({
...
-      code,
-      defaults(
-        {
-          sourceFileName: "stdin",
-        },
-        babelOptions,
-      ),
-    );
+    const res = await util.transform(cliOptions.filename, code, {
+      sourceFileName: "stdin",
```
jayaddison, line 131: @existentialism I guess this is a simpler example here.  We always want `stdin` to be applied here, I suppose?

If that's the case, then yes these are definitely out-of-order.

packages/babel-cli/src/babel/dir.js
existentialism: FWIW, it may not be the _intent_, but as far as I can tell, `sourceFileName` in this instance cannot be overridden by `babelOptions`?

jayaddison: Yep, I think you're correct there, and I've updated the argument ordering here and the other callsite.

Because I'm still feeling a little uncertain about the change, I've marked it as a draft for now and will re-read the code again a bit later.

kaicataldo:
Do we want to do this? The documentation for `_.defaults` says 

> Once a property is set, additional values of the same property are ignored.

So if any of the objects we’re merging in contain the same key twice, this would change the behavior.

existentialism:
@kaicataldo check out the now collapsed/outdated discussion

jayaddison:
I do think this is still a little risky and would agree with @kaicataldo's concern, even if technically the code should be safe from a readthrough and with tests passing.

existentialism:
I'm fine with closing and sticking with `_.defaults` (though I don't see it as being that risky)

jayaddison:
Would changes like this tend to be caught by manual testing of release candidates / betas before a full release?

If the CLI & library paths here get a decent amount of pre-release exercise I think that'd tend to de-risk the changes a lot.

packages/babel-helper-transform-fixture-test-runner/src/index.js
```diff
@@ -376,9 +375,10 @@ export default function (
               run(task);
             }
 
-            defaults(task.options, {
+            task.options = {
               sourceMap: !!(task.sourceMappings || task.sourceMap),
-            });
+              ...task.options,
+            };
```
coreyfarrell, line 381: This introduces some slight changes to the behavior though I'm not sure it matters in this module / context.  Prior to this PR `task.options` got mutated, this PR causes it to instead be cloned and `task` gets mutated.  I think the exact replacement for existing code would actually be:
```js
if ('sourceMap' in task.options === false) {
  task.options.sourceMap = !!(task.sourceMappings || task.sourceMap);
}
```

I'm not familiar with this module to say if this is actually an issue so I'm not actually requesting any change be made, I leave that up to other reviewers to decide if this matters.

jayaddison: That makes sense to me.  Of the `lodash` function replacements so far, this is the one I'm least confident about in terms of behaviour change, safety, and also readability to some extent.  I've only update this single callsite so far, but will re-consider the other locations too.

jayaddison: (adding additional test coverage is probably the best answer to the safety and behaviour change concerns, and I should look into that too)

packages/babel-traverse/src/scope/index.js
```diff
@@ -952,11 +951,11 @@ export default class Scope {
    */
 
   getAllBindings(): Object {
-    const ids = Object.create(null);
+    let ids = Object.create(null);
 
     let scope = this;
     do {
-      defaults(ids, scope.bindings);
+      ids = { ...scope.bindings, ...ids };
```
coreyfarrell, line 958:
```suggestion
      for (const key of Object.keys(scope.bindings)) {
        if (key in ids === false) {
          ids[key] = scope.bindings[key];
        }
      }
```

This is probably a the safer replacement as it ensures that ids (the return value) keeps the null prototype.  From what I can tell `scope.bindings` should was also created by `Object.create(null)` so this will ensure prototype pollution cannot alter the result.

packages/babel-traverse/src/scope/index.js
```diff
@@ -952,11 +951,11 @@ export default class Scope {
    */
 
   getAllBindings(): Object {
-    const ids = Object.create(null);
+    let ids = Object.create(null);
```
coreyfarrell, line 954:
```suggestion
    const ids = Object.create(null);
```

Once the other suggested change is made.

nicolo-ribaudo (approved):

TrySound (approved):

existentialism (approved):

nicolo-ribaudo:
Thanks!

PR #11798, opened 2020-07-06
Reduce dependency on lodash functions: values, extends

jayaddison:

As mentioned by @hzoo in https://github.com/babel/babel/pull/11790#issuecomment-654334560, we can remove the (only) use of `lodash.values` within the codebase with a slightly longer-form but logically-equivalent iteration over the target object properties followed by retrieval of each object value.

- `lodash/values` is replaced by `Object.keys` in combination with `map` and object property access-by-key
- `lodash/extend` is replaced by use of native JavaScript objecty property iteration

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11798"></a>

### DIFF
--- a/packages/babel-plugin-transform-block-scoping/src/index.js
+++ b/packages/babel-plugin-transform-block-scoping/src/index.js
@@ -338,6 +338,12 @@ const loopVisitor = {
   },
 };
 
+function extendMap(map, obj) {
+  for (const key of Object.keys(obj)) {
+    map.set(key, obj[key]);
+  }
+}
+
 class BlockScoping {
   constructor(
     loopPath?: NodePath,
@@ -700,10 +706,7 @@ class BlockScoping {
       const init = this.loop.left || this.loop.init;
       if (isBlockScoped(init)) {
         declarators.push(init);
-        const names = t.getBindingIdentifiers(init);
-        for (const name of Object.keys(names)) {
-          this.outsideLetReferences.set(name, names[name]);
-        }
+        extendMap(this.outsideLetReferences, t.getBindingIdentifiers(init));
       }
     }
 
@@ -752,9 +755,7 @@ class BlockScoping {
       // declaration, rather than (for example) mistakenly including the
       // parameters of a function declaration. Fixes #4880.
       const keys = t.getBindingIdentifiers(declar, false, true);
-      for (const key of Object.keys(keys)) {
-        this.letReferences.set(key, keys[key]);
-      }
+      extendMap(this.letReferences, keys);
       this.hasLetReferences = true;
     }
 
--- a/packages/babel-plugin-transform-block-scoping/src/index.js
+++ b/packages/babel-plugin-transform-block-scoping/src/index.js
@@ -543,7 +543,7 @@ class BlockScoping {
     this.hoistVarDeclarations();
 
     // turn outsideLetReferences into an array
-    const args = [...outsideRefs.values()].map(node => t.cloneNode(node));
+    const args = Array.from(outsideRefs.values(), node => t.cloneNode(node));
     const params = args.map(id => t.cloneNode(id));
 
     const isSwitch = this.blockPath.isSwitchStatement();
--- a/packages/babel-plugin-transform-block-scoping/src/index.js
+++ b/packages/babel-plugin-transform-block-scoping/src/index.js
@@ -193,7 +193,7 @@ const letReferenceBlockVisitor = traverse.visitors.merge([
 const letReferenceFunctionVisitor = traverse.visitors.merge([
   {
     ReferencedIdentifier(path, state) {
-      const ref = state.letReferences[path.node.name];
+      const ref = state.letReferences.get(path.node.name);
 
       // not a part of our scope
       if (!ref) return;
@@ -248,7 +248,7 @@ const continuationVisitor = {
     if (path.isAssignmentExpression() || path.isUpdateExpression()) {
       for (const name of Object.keys(path.getBindingIdentifiers())) {
         if (
-          state.outsideReferences[name] !==
+          state.outsideReferences.get(name) !==
           path.scope.getBindingIdentifier(name)
         ) {
           continue;
@@ -446,7 +446,7 @@ class BlockScoping {
     const letRefs = this.letReferences;
 
     for (const key of letRefs.keys()) {
-      const ref = letRefs[key];
+      const ref = letRefs.get(key);
       const binding = blockScope.getBinding(ref.name);
       if (!binding) continue;
       if (binding.kind === "let" || binding.kind === "const") {
@@ -477,7 +477,7 @@ class BlockScoping {
     for (const key of letRefs.keys()) {
       // just an Identifier node we collected in `getLetReferences`
       // this is the defining identifier of a declaration
-      const ref = letRefs[key];
+      const ref = letRefs.get(key);
 
       // todo: could skip this if the colliding binding is in another function
       if (scope.parentHasBinding(key) || scope.hasGlobal(key)) {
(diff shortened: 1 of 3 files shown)


jayaddison:
(NB: Using `Object.values` was considered as an alternative during https://github.com/babel/babel/pull/11790 but was discounted since that prototype method isn't reliably available until Node 7.x)

packages/babel-plugin-transform-block-scoping/src/index.js
```diff
@@ -545,7 +544,9 @@ class BlockScoping {
     this.hoistVarDeclarations();
 
     // turn outsideLetReferences into an array
-    const args = values(outsideRefs).map(id => t.cloneNode(id));
```
JLHwung, line 548: Note that `Object.keys()` will throw on nullish values but we are good here because it is initialized as `Object.create(null)`.

I think an alternative way is refactoring `outsideRefs` as a `Map()`, so we will have `.keys()` and `.values()`, and what's more the `outsideRefs`, by design, does not have fixed shape, so V8 will finally gave up due to [megamorphic shapes](https://erdem.pl/2019/08/v-8-function-optimization#inline-cache-ic).

jayaddison: Thanks @JLHwung - just to confirm, that would include refactoring the variable referenced by `this.outsideLetReferences`, as well as the local variable `outsideRefs`?

kaicataldo: Is there a reason we can’t use `Object.values().map` here?

jayaddison: @kaicataldo We should be set for `Object.values` from Babel 8 onwards; but `main`'s using Node [6.x](https://github.com/babel/babel/blob/3e55270135410c5b0f75de7b9a283565357fe595/package.json#L82) and `Object.values` [isn't available there](https://node.green/#ES2017-features-Object-static-methods-Object-values) currently.

JLHwung: @jayaddison Yep it may end up with more efforts so it's not a blocker and more like good-to-have.

jayaddison: Ok; it does look like a little more work, but not too much, and it could be a valuable cleanup (and performance improvement if I understand the V8 implications correctly).  I'll give that a try here.

jayaddison: (NB: unfortunately I can't easily run tests locally at the moment, so this may involve a few attempts.  I'll spend more time getting those to work if this ends up with repeat CI failures..)

jayaddison:
:question: Looks like I've introduced an unterminated loop here somehow; the CI unit tests appear to be timing out.

packages/babel-plugin-transform-block-scoping/src/index.js
```diff
@@ -519,20 +517,18 @@ class BlockScoping {
...
         ) {
-          delete outsideRefs[id.name];
-          delete this.letReferences[id.name];
+          outsideRefs.delete(id.name);
 
           this.scope.rename(id.name);
 
-          this.letReferences[id.name] = id;
-          outsideRefs[id.name] = id;
+          outsideRefs.set(id.name, id);
```
jayaddison, line 531: Aha.  I should have noticed that these variables (`this.letReferences`, `outsideRefs`) aren't referring to the same object (I did think it was strange, so I de-duplicated them; but now I realize they're not duplicates in the first place..)

packages/babel-plugin-transform-block-scoping/src/index.js
```diff
@@ -545,7 +543,7 @@ class BlockScoping {
     this.hoistVarDeclarations();
 
     // turn outsideLetReferences into an array
-    const args = values(outsideRefs).map(id => t.cloneNode(id));
+    const args = [...outsideRefs.values()].map(node => t.cloneNode(node));
```
AviVahl, line 546: Tip: `Array.from` supports a second parameter for this use case:

```js
Array.from(outsideRefs.values(), node => t.cloneNode(node));
```
Bit cleaner (and probably faster? iterates once?)

jayaddison: Neat, thanks :)  Change applied.

jayaddison:
I'm pretty confused about how/where I might have introduced the performance regression here - I'd appreciate any ideas and suggestions on how to track that down :)

AviVahl:
Could be that the `Map`s are costing way more than objects if they contain low number of items.
Also, in:
```js
  for (const key of Object.keys(obj)) {
    if (Object.prototype.hasOwnProperty.call(obj, key)) {
      map.set(key, obj[key]);
    }
  }
```
You could remove the `if` and just set the key, as `Object.keys` returns own enumerable keys. The `Object.prototype.hasOwnProperty` lookup itself on every iteration is also costly (if not removed, at least destructured once outside the function).

packages/babel-plugin-transform-block-scoping/src/index.js
jayaddison: @JLHwung Based on some performance regression we seem to be hitting when using `Map`, that approach is looking a little more challenging at the moment.  If we have any more ideas for how to speed things up (or if I've introduced any mistakes here that have led to the performance issues) it might still be possible.

jayaddison:
Thanks @AviVahl - the timeouts during continuous integration had been occurring before the `if` condition was introduced, so I'm not sure that it'll turn out to be the cause, but it's good to know that it's redundant code and can be removed -- I've gone ahead and reverted it.

JLHwung:
@jayaddison Sorry for the late reply. The performance regression, as I can observe, is not introduced by `Map` but an implicit infinite loops.

One of key differences between `Object.keys()` and `Map.prototype.keys()` is that the latter returns an iterator, which means if you add new element to the Map in a loop, the iterator _will_ eventually visit the new element.

In the following example, the for loop is infinite because we keep adding new key to the map. (0, -1, 0, -1 ...) though the original map has only one element ( 0 => undefined ).

```js
const map = new Map([[0]]);
for (const k of map.keys()) {
  a.delete(k);
  a.set(~k, undefined);
}
```

Now let's take a look at our codebase,

https://github.com/babel/babel/blob/bea1b563776e9dadd8b5fda968602ab391fd7ea8/packages/babel-plugin-transform-block-scoping/src/index.js#L526-L541

hangs because in every loop `scope.rename()` generates a new name, assigns to `id.name` and this name is added into `outsideRefs`. Since `scope.rename()` avoids name conflicts, the loop will never terminate.

To mitigate we can take a snapshot of all the keys in `outsideRefs`

```js
[...outsideRefs.keys()]
```

jayaddison:
Brilliant, thanks @JLHwung - oddly an infinite loop was my initial guess for the CI problems but I couldn't figure out where that might have been happening, your explanation makes complete sense.

It looks like there are now some legitimate unit test failures here; I'm finishing up for the day but will take a look at those soon (and will finally make another effort to get unit tests running correctly locally).

JLHwung:
@jayaddison I am really appreciated your efforts!

> It looks like there are now some legitimate unit test failures here

I believe they are due to https://github.com/babel/babel/blob/e498bee10f0123bb208baa228ce6417542a2c3c4/packages/babel-plugin-transform-block-scoping/src/tdz.js#L23

jayaddison:
> @jayaddison I am really appreciated for your efforts!
> 
> > It looks like there are now some legitimate unit test failures here
> 
> I believe they are due to
> 
> https://github.com/babel/babel/blob/e498bee10f0123bb208baa228ce6417542a2c3c4/packages/babel-plugin-transform-block-scoping/src/tdz.js#L23

Glad to help - and thanks, yep - that was it!

PR #11801, opened 2020-07-07
fix: handle block-level function declaration (#10046)

### LINKED ISSUE #10046: plugin-transform-block-scoping renames function declaration
## Bug Report

**Current Behavior**
@babel/plugin-transform-block-scoping (and maybe some other plugin in preset-env) incorrectly renames a FunctionDeclaration in a nested block even though it should behave like a `var` declaration.

**Input Code**
[REPL](https://babeljs.io/repl#?babili=false&browsers=%3E%202%25&build=&builtIns=false&spec=false&loose=true&code_lz=JYMwFALgTgrgpgSgAQG8BQSkhgOwMYTAD2OSsOYy6mmUcEMUpAqsDhABwCCUUAhgE8AdCChEAtmADaARgA0SAEwKAzAF0EAbgxIAvmn1ps-QiSQQ4AZwiVUOugyZlclbYbwlLRADZwh3ogBzSCsbBC00IA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=&prettier=false&targets=Node-8&version=7.4.5&externalPlugins=%40babel%2Fplugin-transform-block-scoping%407.4.4)

```js
if(true) {
  function run() {
    return true;
  }
}

function test() {
  return run();
}
```

*Note*: turning `function run() {` into `var run = function() {` behaves correctly.

**Expected behavior/code**
Same as Input

**Actual output**
```js
if (true) {
  function _run() { // <-------
    return true;
  }
}

function test() {
  return run();
}
```

```js
{
  "plugins": ["@babel/plugin-transform-block-scoping"]
}
```

or
```js
{
}
```

**Environment**
- Babel version(s): 7.4.0
- Node/npm version: Node 10
- OS: macOS
- Monorepo: -
- How you are using Babel: via babel-core

vitorveiga:
| Fixed Issues?            | Fixes #10046, resolves #10050 
| Documentation PR Link    | 

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11801"></a>


vitorveiga:
Hello. 

@JLHwung did you have a chance to review this PR? 

Best Regards

JLHwung (changes requested):
I think it suffices to check if the following conditions are both `true`.

1) `binding.kind` is "hoisted"
2) `key` is not defined in parent scope

<details>

```js
const binding = scope.getOwnBinding(key);
if (binding) {
  if (binding.kind === "hoisted" && !scope.parent.hasOwnBinding(key)) {
    continue;
  }
  scope.rename(ref.name);
}

```
</details>

The first condition is actually a shortcut for `(binding && binding.path && !binding.path.isFunctionDeclaration())`. As of the second condition, we don't have to check `isVar` because the fact that it is one of `letRefs` implies there must be a `let`/`const` binding in current scope chain. Since the `binding.kind` is `"hoisted"`, it _must_ be defined in upper scope so we can simply check whether an upper scope own a binding named as `key`.

vitorveiga:
Thanks for your review :) 

Oh ok.. didn't know that ```binding.kind === 'hoisted'``` is in fact a shortcut for checking if its a function declaration. 
On the other hand, I think we need to continue with ```isVar``` condition because we need to rename the variable if the upper scope has a const or let variable with the same name. However, if its a `var` we should skip the renaming operation.

The conditions were updated!

JLHwung:
> I think we should keep the isVar condition because we need to rename the variable if the upper scope has a const or let variable with the same name. However, if its a var we should skip the renaming operation.

Note that the renaming is nested in `for (const key of letRefs.keys())` loop. If it is a var, it will not be included in `letRefs` so in the loop, if `!parentBinding` is `false`, which means `key` is _owned_ in parent scope, then `isVar(parentBinding.path.parent)` must be `false` otherwise it contradicts with the fact that `key` is of `letRefs`.

(You can also click on "Details" to see my implementation in https://github.com/babel/babel/pull/11801#pullrequestreview-521265154)

vitorveiga:
Note that the implementation in #11801, fails on the unit-test packages/babel-plugin-transform-block-scoping/test/fixtures/general/issue-10046-collision-var/input.js due to the variable declaration on the program node.

In that particular unit test, the ```!parentBinding``` is false and without ```isVar``` condition the rename will happen turning the function declaration into to ```var _run = function() {}```.

> if !parentBinding is false, which means key is owned in parent scope, then isVar(parentBinding.path.parent) must be false otherwise it contradicts with the fact that key is of letRefs.

Maybe there is some implementation bug on letRefs map, but it seems that `isVar(parentBinding.path.parent)` is true on that use case.

vitorveiga:
Hello @JLHwung  did you have a chance to read my last comment? Thanks

JLHwung (approved):

nicolo-ribaudo (approved):

nicolo-ribaudo:
This PR caused some regressions: https://app.circleci.com/pipelines/github/babel/babel/4967/workflows/27043ada-f8d2-482e-810b-2593a17cf087/jobs/35381

PR #11802, opened 2020-07-07
Fix break/continue when switch is nested inside loop

### LINKED ISSUE #11690: Incorrect switch break logic
## Bug Report

Switch break does not work.

Babel **ES2015 Preset** puts "break" in another "switch", so the original "switch" is not "break", and the logic is wrong.

The issue can reproduce on [REPL](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=BQMwrgdgxgLglgewsOMCmBbAzgGgARwCUeA3gFB55RJYx4A2aEeAvAetgHSMQDmMACwDcZCnhAIATnmBCCeADwMmcgNSqipMZWoRa7TKwPYA2nAC6I7XiwB3VFAEzUmEwAZzxcpR9UAhlhoeACMAFxavr66-oG6ACZGLhgmwZaikT5wIM4cKeasLGyQcWggcBBocV7WGZJoMGCSECIZlAC-6a2BMAAqcBhoCGAwoJCwiMjVrT5-jJIjsUhVLa1t-MGEVtMARnV-ANYrvh0ZUAFBAEzh3q2zaPPAF5s1PrtoB0c-J5ElIH5g9BgoRelDeH2s3w6bUIwBMeXwJgu5nM62eQA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=es2015&prettier=false&targets=&version=7.10.2&externalPlugins=).

In the following code, case 1 does not break.

**Input Code**

```js
(function(items, i) {
  const len = items.length;

  for (; i < len; ++i) {
    const item = items[i];

    switch (item[0]) {
      case 1: {
        const second = item[1];

        if (item[1] === undefined) {
          return;
        }

        setTimeout(function() {
          alert(second);
        }, 1);

        break;
      }
      case 2: {
        alert(2);
        break;
      }
      default:
        break;
    }
  }
})([[1], [2]], 1);
```

**Output Code**

```js
"use strict";

function _typeof(obj) { "@babel/helpers - typeof"; if (typeof Symbol === "function" && typeof Symbol.iterator === "symbol") { _typeof = function _typeof(obj) { return typeof obj; }; } else { _typeof = function _typeof(obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; }; } return _typeof(obj); }

(function (items, i) {
  var len = items.length;

  for (; i < len; ++i) {
    var item = items[i];

    switch (item[0]) {
      case 1:
        {
          var _ret = function () {
            var second = item[1];

            if (item[1] === undefined) {
              return {
                v: void 0
              };
            }

            setTimeout(function () {
              alert(second);
            }, 1);
            return "break";
          }();

          switch (_ret) {
            case "break":
              break;

            default:
              if (_typeof(_ret) === "object") return _ret.v;
          }
        }

      case 2:
        {
          alert(2);
          break;
        }

      default:
        break;
    }
  }
})([[1], [2]], 1);
```

**Environment**

- Babel version(s): v7.10.2

### LINKED ISSUE #11800: Bug: switch statements containing closures are broken
## Bug Report

**Current behavior**
Switch statements containing closures are generating broken code causing adjacent cases to run. See this minimal reproduction with the default settings:

https://babeljs.io/en/repl#?browsers=defaults&build=&builtIns=false&spec=false&loose=false&code_lz=GYewTgBAFAxiB2BnALhAhhEwIG0CMAugJQQDeAUBBIgO4CWyMAFtMmAK4CmJFVVMaRJwhsuALjKU-_BCggAjCAF4IeANxTpUEkoB8CjdKp1sUUdwVhOaANaGjceMjrwu9qgF9NEAUIjA0ABshCV4jZCYwEBoIeE4YgFEwKLAoACJ2eCs0ZjR5QM40oncIL09yLyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2%2Cenv&prettier=false&targets=&version=7.10.4&externalPlugins=

**Input Code**

```js
for (const a of [1]) {
  switch (true) {
    case true: {
      const b = 1;
      () => b;
      if (true) break;
      continue;
    }
    case false: {
      throw new Error("unreachable");
    }
  }
}
```

**Output Code**

```js
for (var _i = 0, _arr = [1]; _i < _arr.length; _i++) {
  var a = _arr[_i];

  switch (true) {
    case true:
      {
        var _ret = function () {
          var b = 1;

          (function () {
            return b;
          });

          if (true) return "break";
          return "continue";
        }();

        switch (_ret) {
          case "break":
            break;

          case "continue":
            continue;
        }
      }

    case false:
      {
        throw new Error("unreachable");
      }
  }
}
```

**Expected behavior**
No error is thrown (the break statement functions as intended, and the false case does not run).

Defaults. See the REPL.

**Environment**
Defaults. See the REPL.

**Possible Solution**
If you want to use a `break` statement here, it needs to be a labeled break to break the outer loop.

```js
switch (_ret) {
  case "break":
    break outer;

  case "continue":
    continue;
}
```

Alternatively, use an if instead of a nested switch.

```js
if (_ret === "break") break;
if (_ret === "continue") continue;
```

existentialism:
| Fixed Issues?            | Fixes #11800, fixes #11690 

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11802"></a>

### DIFF
--- a/packages/babel-plugin-transform-block-scoping/src/index.js
+++ b/packages/babel-plugin-transform-block-scoping/src/index.js
@@ -844,16 +844,7 @@ class BlockScoping {
 
   buildHas(ret: string) {
     const body = this.body;
-
-    let retCheck;
     const has = this.has;
-
-    if (has.hasReturn) {
-      // typeof ret === "object"
-      retCheck = buildRetCheck({
-        RETURN: t.identifier(ret),
-      });
-    }
 
     if (has.hasBreakContinue) {
       for (const key of Object.keys(has.map)) {
@@ -866,8 +857,13 @@ class BlockScoping {
       }
     }
 
+    // typeof ret === "object"
     if (has.hasReturn) {
-      body.push(retCheck);
+      body.push(
+        buildRetCheck({
+          RETURN: t.identifier(ret),
+        }),
+      );
     }
   }
 }
--- a/packages/babel-plugin-transform-block-scoping/src/index.js
+++ b/packages/babel-plugin-transform-block-scoping/src/index.js
@@ -847,7 +847,6 @@ class BlockScoping {
 
     let retCheck;
     const has = this.has;
-    const cases = [];
 
     if (has.hasReturn) {
       // typeof ret === "object"
@@ -858,41 +857,17 @@ class BlockScoping {
 
     if (has.hasBreakContinue) {
       for (const key of Object.keys(has.map)) {
-        cases.push(t.switchCase(t.stringLiteral(key), [has.map[key]]));
-      }
-
-      if (has.hasReturn) {
-        cases.push(t.switchCase(null, [retCheck]));
-      }
-
-      if (cases.length === 1) {
-        const single = cases[0];
         body.push(
           t.ifStatement(
-            t.binaryExpression("===", t.identifier(ret), single.test),
-            single.consequent[0],
+            t.binaryExpression("===", t.identifier(ret), t.stringLiteral(key)),
+            has.map[key],
           ),
         );
-      } else {
-        if (this.loop) {
-          // https://github.com/babel/babel/issues/998
-          for (let i = 0; i < cases.length; i++) {
-            const caseConsequent = cases[i].consequent[0];
-            if (t.isBreakStatement(caseConsequent) && !caseConsequent.label) {
-              if (!this.loopLabel) {
-                this.loopLabel = this.scope.generateUidIdentifier("loop");
-              }
-              caseConsequent.label = t.cloneNode(this.loopLabel);
-            }
-          }
-        }
+      }
+    }
 
-        body.push(t.switchStatement(t.identifier(ret), cases));
-      }
-    } else {
-      if (has.hasReturn) {
-        body.push(retCheck);
-      }
+    if (has.hasReturn) {
+      body.push(retCheck);
     }
   }
 }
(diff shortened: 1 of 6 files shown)


packages/babel-plugin-transform-block-scoping/src/index.js
```diff
@@ -858,41 +857,17 @@ class BlockScoping {
...
-        body.push(t.switchStatement(t.identifier(ret), cases));
-      }
-    } else {
-      if (has.hasReturn) {
-        body.push(retCheck);
       }
     }
+
+    if (has.hasReturn) {
+      body.push(retCheck);
```
JLHwung, line 870: nit: We can merge it with line 852
```js
if (has.hasReturn) {
      // typeof ret === "object"
      retCheck = buildRetCheck({
        RETURN: t.identifier(ret),
      });
    }
```

packages/babel-plugin-transform-block-scoping/test/fixtures/general/for-break-continue-return/output.js
```diff
@@ -15,18 +15,11 @@
...
-      case "continue":
-        continue;
-
-      case "break":
-        break _loop;
-
-      default:
-        if (typeof _ret === "object") return _ret.v;
-    }
+    if (_ret === "continue") continue;
```
JLHwung, line 21: Not related to this PR: the string literal here is not minimizer-friendly. We should revise https://github.com/babel/babel/blob/b1a8e72e169b967968c403242a7073b8122d0288/packages/babel-plugin-transform-block-scoping/src/index.js#L266-L272

later, e.g. change it to `"b"`/`"c"`.

PR #11807, opened 2020-07-08
Disallow duplicated AST nodes

### LINKED ISSUE #11506: plugin-proposal-decorators doesn't clone `id` identifier
## Bug Report

**Current Behavior**
After `@babel/plugin-proposal-decorators` and `@babel/plugin-proposal-class-properties` ran, these two identifiers are the same object in the AST (see test script below), so mutating the top-level one has unwanted side-effects.

Babel output:
```js
// ....
let Foo /* <-- 1 */ = _decorate(null, function (_initialize) {
  class Foo /* <-- 2*/ {
    constructor() {
      _initialize(this);
    }

  }

  return {
    F: Foo,
    d: [{
      kind: "field",
      decorators: [property({
        type: String,
        reflect: true
      })],
      key: "content",
      value: void 0
    }]
  };
});

window.customElements.define("my-element", Foo);
```

This also has the effect that `programScope.rename("Foo", "Bar")` renames (2) and the reference in `F: Foo` as well even though these belong to a different binding.

Taking the output above, reparsing it and testing the renaming again, the inner `Foo`s are left untouched. So the AST in is a invalid (?) state.

**Input Code**

```js
class Foo {
  @property({ type: String, reflect: true }) content
}
window.customElements.define("my-element", Foo);
```

**Expected behavior/code**
The two `Foo` should be two different identifier objects.

```js
const babel = require("@babel/core");
const { default: traverse } = require("@babel/traverse");
const { default: generate } = require("@babel/generator");

let INPUT = `
class Foo {
  @property({ type: String, reflect: true }) content
}
window.customElements.define("my-element", Foo);`;

(async () => {
	let ast = babel.transformSync(INPUT, {
		plugins: [
			[
				{
					decoratorsBeforeExport: false,
				},
			],
		],
		code: false,
		ast: true,
	}).ast;

	traverse(ast, {
		Program(path) {
			let binding = path.scope.getBinding("Foo");
			let decl = binding.path;

			let declaratorId = decl.node.declarations[0].id;
			let bindingId = binding.id;
			let classId =
				decl.node.declarations[0].init.arguments[1].body.body[0].id;

			console.log(declaratorId == bindingId); // true, correct
			console.log(declaratorId == classId); // true, wrong
		},
	});

	console.log(generate(ast).code);
})();
```

**Environment**

- Babel version(s): `@babel/plugin-proposal-class-properties@7.8.3`
- Node/npm version: Node 14
- OS: macos

**Possible Solution**
Somewhere in https://github.com/babel/babel/blob/master/packages/babel-helper-create-class-features-plugin/src/index.js, an identifier should be cloned instead of being reused.

JLHwung:
| Fixed Issues?            | Fixes #11506 

As a follow-up to #7149, this PR enables duplicated node checks against all testcases and then fixes all duplicate AST errors.

I didn't add a regression test of #11506 because it is covered in [this test case](https://github.com/babel/babel/blob/v7.10.4/packages/babel-plugin-proposal-decorators/test/fixtures/element-descriptors/original-own-field/exec.js). I realized that the duplicated node checks are not effective when I am investigating #11506, which should have been captured by our test runner.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11807"></a>

### DIFF
--- a/packages/babel-helper-create-class-features-plugin/src/misc.js
+++ b/packages/babel-helper-create-class-features-plugin/src/misc.js
@@ -72,8 +72,14 @@ export function injectInitialization(path, constructor, nodes, renamer) {
   if (isDerived) {
     const bareSupers = [];
     constructor.traverse(findBareSupers, bareSupers);
+    let isFirst = true;
     for (const bareSuper of bareSupers) {
-      bareSuper.insertAfter(nodes);
+      if (isFirst) {
+        bareSuper.insertAfter(nodes);
+        isFirst = false;
+      } else {
+        bareSuper.insertAfter(nodes.map(n => t.cloneNode(n)));
+      }
     }
   } else {
     constructor.get("body").unshiftContainer("body", nodes);
--- a/packages/babel-helper-transform-fixture-test-runner/src/index.js
+++ b/packages/babel-helper-transform-fixture-test-runner/src/index.js
@@ -146,6 +146,7 @@ function run(task) {
   function getOpts(self) {
     const newOpts = merge(
       {
+        ast: true,
         cwd: path.dirname(self.loc),
         filename: self.loc,
         filenameRelative: self.filename,
--- a/packages/babel-plugin-transform-parameters/src/params.js
+++ b/packages/babel-plugin-transform-parameters/src/params.js
@@ -229,6 +229,6 @@ function buildScopeIIFE(shadowedParams, body) {
   }
 
   return t.returnStatement(
-    t.callExpression(t.arrowFunctionExpression(params, body), params),
+    t.callExpression(t.arrowFunctionExpression(args, body), params),
   );
 }
--- a/packages/babel-plugin-transform-parameters/src/params.js
+++ b/packages/babel-plugin-transform-parameters/src/params.js
@@ -229,6 +229,6 @@ function buildScopeIIFE(shadowedParams, body) {
   }
 
   return t.returnStatement(
-    t.callExpression(t.arrowFunctionExpression(args, body), params),
+    t.callExpression(t.arrowFunctionExpression(params, body), args),
   );
 }
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -134,7 +134,9 @@ export default declare(
               );
             }
 
-            return template.statement.ast`this.${id} = ${id}`;
+            return template.statement.ast`this.${t.cloneNode(
+              id,
+            )} = ${t.cloneNode(id)}`;
           });
 
           injectInitialization(classPath, path, assigns);
--- a/packages/babel-plugin-transform-typescript/src/index.js
+++ b/packages/babel-plugin-transform-typescript/src/index.js
@@ -134,9 +134,8 @@ export default declare(
               );
             }
 
-            return template.statement.ast`this.${t.cloneNode(
-              id,
-            )} = ${t.cloneNode(id)}`;
+            return template.statement.ast`
+              this.${t.cloneNode(id)} = ${t.cloneNode(id)}`;
           });
 
           injectInitialization(classPath, path, assigns);
--- a/packages/babel-plugin-transform-typescript/src/namespace.js
+++ b/packages/babel-plugin-transform-typescript/src/namespace.js
@@ -157,8 +157,10 @@ function handleNested(path, t, node, parentExport) {
 
   if (parentExport) {
     fallthroughValue = template.expression.ast`
-      ${parentExport}.${realName} || (
-        ${parentExport}.${realName} = ${fallthroughValue}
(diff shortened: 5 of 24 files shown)


packages/babel-helper-transform-fixture-test-runner/src/index.js
```diff
@@ -147,6 +147,7 @@ function run(task) {
   function getOpts(self) {
     const newOpts = merge(
       {
+        ast: true,
```
JLHwung, line 150: 😄Finally get duplicate AST checks working!

packages/babel-plugin-transform-parameters/src/params.js
```diff
@@ -229,6 +229,6 @@ function buildScopeIIFE(shadowedParams, body) {
   }
 
   return t.returnStatement(
-    t.callExpression(t.arrowFunctionExpression(params, body), params),
+    t.callExpression(t.arrowFunctionExpression(args, body), params),
```
JLHwung, line 232: @nicolo-ribaudo I believe this is what you intended for. Only one question: do you prefer `args` in `arrowFunctionExpression` or `params` in `arrowFunctionExpression`? 😄

packages/babel-helper-create-class-features-plugin/src/misc.js
```diff
@@ -72,8 +72,14 @@ export function injectInitialization(path, constructor, nodes, renamer) {
   if (isDerived) {
     const bareSupers = [];
     constructor.traverse(findBareSupers, bareSupers);
+    let isFirst = true;
     for (const bareSuper of bareSupers) {
-      bareSuper.insertAfter(nodes);
+      if (isFirst) {
```
JLHwung, line 77: The test [`legacy-class-prototype-properties/child-classes-properties`](https://github.com/babel/babel/blob/v7.10.4/packages/babel-plugin-proposal-decorators/test/fixtures/legacy-class-prototype-properties/child-classes-properties/exec.js) is [failing](https://github.com/JLHwung/babel/runs/851654026#step:4:767) if we change `bareSuper.insertAfter(nodes.map(n => t.cloneNode(n)))` unconditionally. The current approach happens to make our test cases happy and also slightly faster because we get rid of redundant clone.

However, I think it reveals an issue which is not caught in our current test cases.

I suggest we address that in a separate PR or even let it stay in the backlog since it is related to legacy decorators only.

I am not familiar with decorators part, suggestions are definitely welcome.

nicolo-ribaudo: It is probably failing because in the decorators plugin we use a `WARNING_CALLS` WeakSet to track warnings: if we clone the node we cannot find it anymore in the set.

packages/babel-plugin-transform-parameters/src/params.js
nicolo-ribaudo: According to the spec, a `CallExpression` takes an `ArgumentsList` and an `ArrowFunctionExpression` has some `ArrowParameters`, so it's the other way around :nerd_face:

packages/babel-plugin-transform-typescript/src/index.js
```diff
@@ -134,7 +134,9 @@ export default declare(
               );
             }
 
-            return template.statement.ast`this.${id} = ${id}`;
+            return template.statement.ast`this.${t.cloneNode(
+              id,
+            )} = ${t.cloneNode(id)}`;
```
nicolo-ribaudo, line 139: Nit: I really dislike how this is formatted. Does prettier accept something like this?

```suggestion
            return template.statement.ast`
              this.${t.cloneNode(id} = ${t.cloneNode(id)}
            `;
```

packages/babel-plugin-transform-typescript/src/namespace.js
```diff
@@ -157,15 +157,17 @@ function handleNested(path, t, node, parentExport) {
...
   if (parentExport) {
     fallthroughValue = template.expression.ast`
-      ${parentExport}.${realName} || (
-        ${parentExport}.${realName} = ${fallthroughValue}
+      ${parentExport}.${t.cloneNode(realName)} || (
+        ${t.cloneNode(parentExport)}.${t.cloneNode(
+      realName,
+    )} = ${fallthroughValue}
       )
     `;
```
nicolo-ribaudo, line 165: Also this is unreadable.

```js
const memberExpr = t.memberExpression(parentExport, realName);
fallthroughValue = template.expression.ast`
  ${memberExpr} || (
    ${t.cloneNode(memberExpr)} = ${fallthroughValue}
  )
`;
```

packages/babel-plugin-transform-parameters/src/params.js
JLHwung, line 232:
```suggestion
    t.callExpression(t.arrowFunctionExpression(params, body), args),
```

PR #11810, opened 2020-07-09
Replace lodash 'sortBy' usage with Array.prototype.sort

jayaddison:

This is likely a lower-priority change now that `lodash` 4.17.19 has been released and resolves a vulnerability that caused package audit warnings and noise for many users; that was the original motivation for removing some of these functions as per #11789.

That said, there still seems to be scope for further dependency reduction, so this changeset is an attempt to remove one more `lodash` reference.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11810"></a>


packages/babel-core/src/transformation/block-hoist-plugin.js
```diff
@@ -41,24 +39,14 @@ const blockHoistPlugin = {
...
-          if (priority == null) priority = 1;
-          if (priority === true) priority = 2;
-
-          // Higher priorities should move toward the top.
-          return -1 * priority;
-        });
+        if (node.body.every(node => node?._blockHoist == null)) return;
+        node.body
+          .filter(node => node?._blockHoist == null)
+          .forEach(node => (node.priority = 1));
```
jayaddison, line 45: Nope, this isn't a great approach; on a fresh readthrough, `priority` isn't a property that previously existed on these nodes (`priority` in the existing implementation is a purely local variable), so it's not wise to *set* it as a property on them either.

packages/babel-core/src/transformation/block-hoist-plugin.js
```diff
@@ -41,24 +39,13 @@ const blockHoistPlugin = {
...
-          if (priority == null) priority = 1;
-          if (priority === true) priority = 2;
-
-          // Higher priorities should move toward the top.
-          return -1 * priority;
-        });
+        if (node.body.every(node => node?._blockHoist == null)) return;
+        const priority = node => {
+          if (node?._blockHoist == null) return 1;
+          if (node?._blockHoist === true) return 2;
```
coreyfarrell, line 45:
```suggestion
          if (node?._blockHoist == null) return -1;
          if (node?._blockHoist === true) return -2;
```

This is potentially what caused the test failures, previous code set `priority` then returned the negated value.

jayaddison: Ah, good catch :)  Thanks!

jayaddison:
Still puzzling.  One possibility is that the sort order may be reversed at the moment, based on some unit test failures where the received statements are flipped around from the expected statements:

```
Expected: "import _source from \"source\"; _source;"
Received: "_source; import _source from \"source\";"
```

I'm just doing some sorting via the node shell with `lodash.sortBy` and `Array.prototype.sort` to try to determine whether that's the cause.  The comparator function could be the source of the trouble (and perhaps just reversing the division operands around is the solution, but I'd like to confirm that first).

jayaddison:
Yep, the current `b - a` comparator ordering when used with `Array.prototype.sort` _doesn't_ match `lodash/sortBy`; so I think these need to be flipped around.  A quick justification:

```javascript
> sortBy = require('lodash/sortBy')
[Function]
> arr = [{"a": 1}, {"a": -1}, {"a": 3}]
[ { a: 1 }, { a: -1 }, { a: 3 } ]
> sortFn = x => x.a * -1;
[Function: sortFn]
> sortBy(arr, sortFn)
[ { a: 3 }, { a: 1 }, { a: -1 } ]
> arr.sort((a, b) => sortFn(b) - sortFn(a))
[ { a: -1 }, { a: 1 }, { a: 3 } ]
> 
```

nicolo-ribaudo:
Thanks for investigating!

jayaddison:
I'm a bit puzzled about why the [latest Travis checks](https://github.com/babel/babel/pull/11810/checks?check_run_id=854661491) are failing here for Node < v12.  There must be a language feature in use that's only supported in later versions?

coreyfarrell:
> I'm a bit puzzled about why the [latest Travis checks](https://github.com/babel/babel/pull/11810/checks?check_run_id=854661491) are failing here for Node < v12. There must be a language feature in use that's only supported in later versions?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Browser_compatibility

Apparently `Array.prototype.sort` wasn't a stable sort before node.js 12.  It looks like lodash deals with this by wrapping array items so the internal comparitor has access to each item's initial index.

The following code demonstrates this problem:
```js
const items = [];

for (let i = 0; i < 20; i++) {
  items.push({idx: i, value: 'a'});
}

items[10].value = 'b';
items.sort((a, b) => {
  if (a.value < b.value) {
    return -1;
  }

  if (a.value > b.value) {
    return 1;
  }

  // This if statement makes the sort stable for node.js < 12 but requires
  // knowledge of the original indexes.
  if (process.versions.node.split('.')[0] < 12) {
    return a.idx - b.idx;
  }

  return 0;
});

console.log(items);
```

jayaddison:
Well, today's turning into a good day for learning software archaeology; thank you @coreyfarrell :)

coreyfarrell:
@jayaddison I just did some more experimentation with my sort stability demo, found an easier way to get it working for older node.  See my previous message I've edited the example code to use a `Map` when needed.

jayaddison:
@coreyfarrell Nice!  Just to rephrase it as I understand it: for Node versions prior to v12, we store a map keyed on each node object and storing the array index for that object as a value.  If that map exists during sorting, then it's used as a tie-breaker to ensure stable sort behaviour.

Nice approach.  I'll start incorporating that here.

jayaddison:
(we might also want to keep the 'Sorting with map' section from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Browser_compatibility in mind, particularly if any of the performance concerns mentioned there crop up)

packages/babel-core/src/transformation/block-hoist-plugin.js
```diff
@@ -41,24 +39,25 @@ const blockHoistPlugin = {
...
+        const priority = node => {
+          if (node?._blockHoist == null) return -1;
+          if (node?._blockHoist === true) return -2;
+          return -1 * node?._blockHoist;
+        };
+        node.body.sort(
+          (a, b) =>
+            priority(a) - priority(b) ||
+            stabilityMap.get(a) - stabilityMap.get(b),
+        );
```
coreyfarrell, line 60:
```suggestion
        node.body.sort(
          (a, b) => {
            const result = priority(a) - priority(b);
            if (!result && sortIsUnstable) {
              return stabilityMap.get(a) - stabilityMap.get(b);
            }
            return result;
        });
```
The existing code adds the overhead to node.js 12+ even if not needed and does a lookup on an empty map (`undefined - undefined` is NaN).

Side note in my example code `new Map` was only run in Node.js < 12, current versions of Node.js can avoid even creating the Map object.

jayaddison: Yup, good call.  I was trying to streamline the logic a little, but any elegance in the code isn't worth the overhead of the lookups.  I'll push a change for that shortly.

packages/babel-core/src/transformation/block-hoist-plugin.js
```diff
@@ -41,23 +39,23 @@ const blockHoistPlugin = {
...
+            : null;
+        const priority = node => {
+          if (node?._blockHoist == null) return -1;
+          if (node?._blockHoist === true) return -2;
+          return -1 * node?._blockHoist;
+        };
+        node.body.sort((a, b) => {
+          const result = priority(a) - priority(b);
+          if (result) return result;
+          if (stabilityMap) return stabilityMap.get(a) - stabilityMap.get(b);
```
coreyfarrell, line 58: `return 0` after this?  Not sure what `Array#sort` does with undefined return where `result === 0 && !stabilityMap`.

jayaddison: Sounds good - what do you think about `return result`?  (even though it'd be guaranteed to be zero in that case, it links the ownership/origin of that value back to the calculation for anyone reading the code)

coreyfarrell: Seems fine, I don't have a strong opinion but my original suggestion accomplished this in a different way:
```js
const result = priority(a) - priority(b);
if (!result && stabilityMap) {
  return stabilityMap.get(a) - stabilityMap.get(b);
}
return result;
```

jayaddison: Ok, yep - I like that there are only two `return` statements there.  I guess I have an aversion to checking two conditions in the `if` block if there's a way we can avoid it.

How about

```javascript
const result = priority(a) - priority(b);
if (stabilityMap) {
  return result || stabilityMap.get(a) - stabilityMap.get(b);
}
return result;
```

coreyfarrell: I have no objection to that.

jayaddison: Cool.  Thanks for the discussion :)

packages/babel-core/src/transformation/block-hoist-plugin.js
```diff
@@ -43,8 +43,10 @@ const blockHoistPlugin = {
         // TODO: Babel 9 (?) - remove stabilityMap once Node >= 12 is guaranteed
         // Array sorting is not stable in earlier Node releases
         // See: https://v8.dev/blog/array-sort for details
-        const stabilityMap =
-          process.versions.node.split(".")[0] < 12
+        const nodeVersion = typeof process !== "undefined"
+            ? process.versions.node.split(".")[0]
+            : undefined;
+        const stabilityMap = nodeVersion && nodeVersion < 12
```
nicolo-ribaudo: If `nodeVersion` is not defined (browsers) we should still use the map.

packages/babel-core/src/transformation/block-hoist-plugin.js
```diff
@@ -41,23 +39,29 @@ const blockHoistPlugin = {
...
+        if (!node.body.find(node => node?._blockHoist != null)) return;
+        // TODO: Babel 9 (?) - remove stabilityMap once Node >= 12 is guaranteed
+        // Array sorting is not stable in earlier Node releases
+        // See: https://v8.dev/blog/array-sort for details
+        const nodeVersion =
+          typeof process !== "undefined"
+            ? process.versions.node.split(".")[0]
+            : undefined;
+        const stabilityMap =
+          nodeVersion && nodeVersion < 12
```
nicolo-ribaudo, line 51: If `nodeVersion` is not defined (browsers) we should still use the map because the sorting might not be stable.

```suggestion
          !nodeVersion || nodeVersion < 12
```

nicolo-ribaudo, line 51: If `nodeVersion` is not defined (browsers) we should still use the map because the sorting might not be stable.

```suggestion
          !nodeVersion || nodeVersion < 12
```

packages/babel-core/src/transformation/block-hoist-plugin.js
```diff
@@ -41,23 +39,29 @@ const blockHoistPlugin = {
...
+            ? new Map(node.body.map((n, idx) => [n, idx]))
```
nicolo-ribaudo, line 52:
```suggestion
            ? new Map(node.body.entries())
```

packages/babel-core/src/transformation/block-hoist-plugin.js
```diff
@@ -41,23 +39,29 @@ const blockHoistPlugin = {
   visitor: {
     Block: {
       exit({ node }) {
-        let hasChange = false;
-        for (let i = 0; i < node.body.length; i++) {
-          const bodyNode = node.body[i];
-          if (bodyNode?._blockHoist != null) {
-            hasChange = true;
-            break;
+        if (!node.body.find(node => node?._blockHoist != null)) return;
```
nicolo-ribaudo, line 42: Nit: 
```suggestion
        if (!node.body.some(node => node?._blockHoist != null)) return;
```

packages/babel-core/src/transformation/block-hoist-plugin.js
jayaddison: :100:

packages/babel-core/src/transformation/block-hoist-plugin.js
```diff
@@ -41,23 +39,29 @@ const blockHoistPlugin = {
...
+        if (!node.body.some(node => node?._blockHoist != null)) return;
+        // TODO: Babel 9 (?) - remove stabilityMap once Node >= 12 is guaranteed
+        // Array sorting is not stable in earlier Node releases
+        // See: https://v8.dev/blog/array-sort for details
+        const nodeVersion =
+          typeof process !== "undefined"
+            ? process.versions.node.split(".")[0]
+            : undefined;
+        const stabilityMap =
+          !nodeVersion || nodeVersion < 12
```
coreyfarrell, line 51:
```suggestion
            : 0;
        const stabilityMap =
          nodeVersion < 12
```

Feel free to ignore this suggestion, just wanted to point out that the conditional can be simplified if you make browsers produce `nodeVersion === 0`.

jayaddison: Thanks @coreyfarrell - I think I'd prefer to keep the meaning/semantics of `nodeVersion` closer to the ground truth (i.e. it'll genuinely have an undefined value in non-Node environments), even though I'd admit that it's (hopefully) unlikely that anyone would be line-by-line debugging the value.  I'm sure I've probably used tricks like this to simplify my own code elsewhere too, so it's a bit arbitrary of me not to accept.  But I think I'd prefer the more verbose condition in this case.

packages/babel-core/src/transformation/block-hoist-plugin.js
coreyfarrell: `Array#entries()` actually gives the reverse of what we need for each element.
```js
const items = ['a', 'b', 'c'];
console.log(new Map(items.map((n, idx) => [n, idx])));
// Map { 'a' => 0, 'c' => 1, 'b' => 2 }
console.log(new Map(items.entries()))
// Map { 0 => 'a', 1 => 'c', 2 => 'b' }
```

We need a Map keyed on the values of `node.body`, not a Map keyed on the indexes.

jayaddison:
NB: There's a change to the `semver.intersects` logic from `semver` v6.0.0 onwards; that might be worth looking at, particulary in relation to the comments here: https://github.com/babel/babel/blob/5bbad8936b05a2b6739dcade6464dbc361118aa9/packages/babel-core/src/transformation/file/file.js#L141-L158

jayaddison:

PR #11811, opened 2020-07-09
Replace lodash 'clone' usage with ES6 Spread initializer

jayaddison:

Continues removal of lodash function dependencies identified per #11789.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11811"></a>

### DIFF
--- a/packages/babel-traverse/src/visitors.js
+++ b/packages/babel-traverse/src/visitors.js
@@ -1,6 +1,5 @@
 import * as virtualTypes from "./path/lib/virtual-types";
 import * as t from "@babel/types";
-import clone from "lodash/clone";
 
 /**
  * explode() will take a visitor object with all of the various shorthands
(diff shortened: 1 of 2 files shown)


packages/babel-traverse/src/visitors.js
```diff
@@ -106,7 +105,7 @@ export function explode(visitor) {
       if (existing) {
         mergePair(existing, fns);
       } else {
-        visitor[alias] = clone(fns);
+        visitor[alias] = { ...fns };
```
coreyfarrell, line 108: @jayaddison Sorry I missed this while it was open but I think this could introduce an edge case behavior change if `fns` has any non-enumerable values in the prototype chain.  A better replacement might be:
```js
visitor[alias] = Object.create(Object.getPrototypeOf(fns));
Object.assign(visitor[alias], fns);
```

Here the cloned object is created with the prototype of `fns`, then `Object.assign` copies only the own iterable properties from fns.

jayaddison: No problem at all, thanks for catching this.  Fixup opened in #11820.

PR #11815, opened 2020-07-09
eslint-parser: ES2020 features

JLHwung:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #11711 
| Minor: New Feature?      | Yes
| Tests Added + Pass?      | Yes
| Any Dependency Changes?  | `eslint-visitor-keys` is bumped to `^1.3.0` which includes the `ChainExpression` visitor keys.
| License                  | MIT

This PR adds estree parsing support on the following features:

- optional chaining expression.
- nullish coalescing
- `export * ns` (credits to @existentialism)

The nullish coalescing is out of the box since ESTree and Babel shares the AST shape. Thus I have enabled test.

The `@babel/parser` `estree` plugin now supports parsing optional chaining and `export * ns`.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11815"></a>

PR #11819, opened 2020-07-10
Compare export * directly to espree in eslint-parser

existentialism:

Piggybacking on https://github.com/babel/babel/pull/11815

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11819"></a>

### DIFF
--- a/eslint/babel-eslint-parser/src/convert/convertAST.js
+++ b/eslint/babel-eslint-parser/src/convert/convertAST.js
@@ -91,11 +91,19 @@ function convertNodes(ast, code) {
     "typeParameters",
   ];
 
+  // Make sure we visit `exported` key to remove `identifierName` from loc node
+  const oldExportAllDeclarationKeys = t.VISITOR_KEYS.ExportAllDeclaration;
+  t.VISITOR_KEYS.ExportAllDeclaration = t.VISITOR_KEYS.ExportAllDeclaration.concat(
+    "exported",
+  );
+
   traverse(ast, astTransformVisitor, null, state);
 
   // These can be safely deleted because they are not defined in the original visitor keys.
   delete t.VISITOR_KEYS.Property;
   delete t.VISITOR_KEYS.MethodDefinition;
+
+  t.VISITOR_KEYS.ExportAllDeclaration = oldExportAllDeclarationKeys;
 }
 
 function convertProgramNode(ast) {
--- a/packages/babel-types/src/definitions/es2015.js
+++ b/packages/babel-types/src/definitions/es2015.js
@@ -249,7 +249,7 @@ defineType("ClassDeclaration", {
 });
 
 defineType("ExportAllDeclaration", {
-  visitor: ["source", "exported"],
+  visitor: ["source"],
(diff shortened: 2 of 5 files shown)


existentialism:
Note bumping eslint-scope is necessary to support ExportAllDeclaration => https://github.com/eslint/eslint-scope/pull/56

packages/babel-types/src/definitions/es2015.js
```diff
@@ -249,7 +249,7 @@ defineType("ClassDeclaration", {
 });
 
 defineType("ExportAllDeclaration", {
-  visitor: ["source"],
+  visitor: ["source", "exported"],
```
nicolo-ribaudo, line 252: What is this needed for?

existentialism: We won't remove identifierName from the location if we don't visit it.

nicolo-ribaudo: Oh ok, I was confused because it isn't in `fields`. Since it is ESTree-specific and not part of the Babel AST, maybe we should monkey patch it like we do it [here](https://github.com/babel/babel/blob/ffb42488ba1a9d05a41d028cd72ae8c77dfe2071/eslint/babel-eslint-parser/src/convert/convertAST.js#L86)?

existentialism: Good catch!

eslint/babel-eslint-parser/src/convert/convertAST.js
```diff
@@ -91,11 +91,19 @@ function convertNodes(ast, code) {
     "typeParameters",
   ];
 
+  // Make sure we visit `exported` key to remove `identifierName` from loc node
+  const oldExportAllDeclarationKeys = t.VISITOR_KEYS.ExportAllDeclaration;
```
JLHwung, line 95: We should add `ExportAllDeclaration` to `eslint/babel-eslint-parser/src/visitor-keys.js`, too.

JLHwung:
Do you mind rebase your PR onto `babel-eslint-2020` and submit this PR to my repo? So we can track `export *` in #11815.

eslint/babel-eslint-parser/src/convert/convertAST.js
existentialism: I don't think we need to: https://github.com/eslint/eslint-visitor-keys/blob/master/lib/visitor-keys.json#L71?

JLHwung: Well only some of them are picked:

https://github.com/babel/babel/blob/ffb42488ba1a9d05a41d028cd72ae8c77dfe2071/eslint/babel-eslint-parser/src/visitor-keys.js#L7-L13

existentialism: Yeah, guess we have to add back ESLint's version since Babel's will overwrite it

existentialism:
@JLHwung pushed to your branch

PR #11820, opened 2020-07-10
Fixup: ensure any prototype properties of cloned 'fns' object are retained

jayaddison:

This change is a fixup and continuation of changes applied in https://github.com/babel/babel/pull/11811 to remove usages of the `lodash/clone` function.

Ref: https://github.com/babel/babel/pull/11811#discussion_r452826637

### DIFF
--- a/packages/babel-traverse/src/visitors.js
+++ b/packages/babel-traverse/src/visitors.js
@@ -105,7 +105,8 @@ export function explode(visitor) {
       if (existing) {
         mergePair(existing, fns);
       } else {
-        visitor[alias] = { ...fns };
+        visitor[alias] = Object.create(Object.getPrototypeOf(fns));
+        Object.assign(visitor[alias], fns);
       }
     }
   }


jayaddison:
Hrm.  On the one hand, [`mergePairs`](https://github.com/babel/babel/blob/cf01c15dafe42a840aa4b4c47479e21ddfa25582/packages/babel-traverse/src/visitors.js#L285-L289) seems to operate only using enumerable keys when copying from a `src` object; I read that as a sign that the intent of the code is only to care about enumerable properties.

On the other hand, `visitors` is [exported](https://github.com/babel/babel/blob/cf01c15dafe42a840aa4b4c47479e21ddfa25582/packages/babel-traverse/src/index.js#L11) by the top-level `babel-traverse` module which could be an argument in favour of keeping things strictly the same behaviour-wise -- or it could be time for a wake-up call to anyone somehow relying on non-enumerable properties being exposed.

I'd probably be in favour of simplifying the set of properties (just in terms of simplicity and reducing undefined/unexpected behaviours outside the intent of the code) -- as long as that's not going to set half the JS ecosystem on fire.

cc @coreyfarrell just in case you missed any of this; this is a discussion around maintaining edge-case behaviour after replacing `lodash/clone`.

coreyfarrell:
In the strictest view dropping the prototype could be a breaking change, but other interpretations could view it is a bug fix.  I don't have a strong opinion about this patch being needed for v7, v8, both or neither.  This edge case will not cause trouble for any babel plugin I've ever worked on.  My primary concern was that the edge case be noticed so it can be decided what if anything to do.  @jayaddison thanks for posting this PR!

jayaddison:
Thanks @coreyfarrell!

If we're ok without needing this (at least until we discover any use cases where it does/could cause problems), then I think from a simplicity point of view it'd be preferable to close this.  I think/hope we'd be able to reference and re-apply this later quickly if needed (hopefully there are enough relevant search terms in the GitHub discussion and topic.. I might add one or two more).

PR #11823, opened 2020-07-10
Delegate to the new `babel-plugin-polyfill-corejs3` package

nicolo-ribaudo:
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11823"></a>

This PR should be reviewed commit by commit.

**Breaking changes**:
- The new polyfill packages require Node.js 10
- `@babel/runtime` now injects imports to `core-js-pure` instead of `@babel/runtime-corejs3/core-js/` (since the new package doesn't use `@babel/runtime-corejs3`). This will make it easier for people to move to `babel-plugin-polyfill-corejs3` in the future, since it won't cause other changes in their code.

The output fixtures should be the same; I'll mark this PR as ready after that I'm sure that it doesn't remove any necessary import.


packages/babel-preset-env/test/fixtures/corejs3/usage-browserslist-config-ignore/stdout.txt
```diff
@@ -32,8 +32,22 @@ Using plugins:
...
+Using targets: {
+  "android": "61",
+  "chrome": "61",
+  "edge": "16",
+  "firefox": "60",
+  "ios": "10.3",
+  "opera": "48",
+  "safari": "10.1",
+  "samsung": "8.2"
+}
```
nicolo-ribaudo, line 46: The targets are now duplicated. I could inject a secret parameter used only by `preset-env` to avoid printing them in the polyfill provider, but I don't think that it's worth the complexity.

nicolo-ribaudo:
Moved to https://github.com/babel/babel/pull/12583

PR #11832, opened 2020-07-13
Forward deopt node path

johanholmerin:
| Fixed Issues?            | 
| Documentation PR Link    | 

Currently, when evaluating array values, object keys, and object properties, if a deopt is found, the location is discarded and the location in the array or the object will be used. I suggest this information is kept. It will make it easier to understand the cause of the deopt.

```javascript
{
  a: { // <- deopt currently points to this object
    b // <- I suggest pointing to this identifier
  }
}
```

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11832"></a>

### DIFF
--- a/packages/babel-core/test/evaluation.js
+++ b/packages/babel-core/test/evaluation.js
@@ -21,25 +21,6 @@ describe("evaluation", function () {
     });
   }
 
-  function addDeoptTest(code, type, expectedType) {
-    it(type + " deopt: " + code, function () {
-      const visitor = {};
-
-      visitor[type] = function (path) {
-        const evaluate = path.evaluate();
-        expect(evaluate.confident).toBeFalsy();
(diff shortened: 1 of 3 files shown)


packages/babel-core/test/evaluation.js
```diff
@@ -21,6 +21,25 @@ describe("evaluation", function () {
     });
   }
 
+  function addDeoptTest(code, type, expectedType) {
```
JLHwung, line 24: nit: Can you move these tests to `packages/babel-traverse/test/evaluation.js`?

johanholmerin:
Fixed

johanholmerin:
Do you need anything from me before merging?

JLHwung:
@johanholmerin We have a two-approval policy on merging PR. ping @babel/core-team.

PR #11836, opened 2020-07-14
Always retain lines for async arrow

### LINKED ISSUE #11830: Adding a comment to an async function before the first argument can result in a syntax error
## Bug Report

**Current behavior**
If you place a comment before the first (and only) argument of an async function (but after the async keyword), and the function has only a single argument, and if you use the await keyword inside that function, the compiled javascript won't run because the function isn't marked as async.

You get the following error when you run the compiled code:
```
babel-async-bug-repro/actual.js:5
  return foo(await a);
             ^^^^^

SyntaxError: missing ) after argument list
    at Module._compile (internal/modules/cjs/loader.js:720:23)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)
    at Module.load (internal/modules/cjs/loader.js:643:32)
    at Function.Module._load (internal/modules/cjs/loader.js:556:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:839:10)
    at internal/main/run_main_module.js:17:11
```

Here's a full reproduction
- https://github.com/xolvio/babel-async-comment-bug

**Input Code**

```js
const x = async (
  // some comment <- this comment is the problem
  a
) => {
  return foo(await a);
};

function foo(a) {
  return a;
}
```

**Expected behavior**
The compiled code should run without error. Presumably by making sure the async keyword is always on the same line as the function declaration

**Babel Configuration (.babelrc)**

The default config will generate the error (as of the day I posted this), also the following config will generate the error (note the borwserlist config is necessary to generate the error, otherwise regeneratorRuntime is used instead of emitting the async/await keywords):

- Filename: `.babelrc`
```js
{
}
```

- Filename: `package.json`
```js
  ...
  "scripts": {
    "build": "babel input.js --out-file actual.js",
    "test": "node actual.js"
  },
  "devDependencies": {
    "@babel/cli": "^7.10.4",
    "@babel/core": "^7.10.4",
    "@babel/preset-env": "^7.10.4"
  },
  "browserslist": [
    "node 12",
    "last 1 Chrome version"
  ]
```

**Environment**

```
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 12.6.0 - /usr/local/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
  npmPackages:
    @babel/cli: ^7.10.4 => 7.10.4 
    @babel/core: ^7.10.4 => 7.10.4 
    @babel/preset-env: ^7.10.4 => 7.10.4 
```

**Possible Solution**
I assume that always wrapping the arguments of an async function in parenthesis (and keeping the async keyword on the same line as the opening parentheses) would fix the issue.

cwohlman:
First stab at fixing #11830, fixes the bug but has other issues.

| Fixed Issues            | Fixes #11830 
| Major: Breaking Change?  | No - Although the current implementation breaks some tests
| Tests Added + Pass?      | No - Added passing tests but breaking other tests
| Documentation PR Link    | 

Since the fix is to always include parentheses for arrow functions, it breaks the following tests:

* `generation/edgecase › single arg async arrow with retainlines`
* `generation/types › ArrowFunctionExpression`

I'm not sure what we should do here, ideally I guess we should only wrap the param in parentheses if there's a comment proceeding the param, though I don't have any idea how to do that.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11836"></a>


packages/babel-generator/src/generators/methods.js
```diff
@@ -109,7 +109,8 @@ export function ArrowFunctionExpression(node: Object) {
   if (
     node.params.length === 1 &&
     t.isIdentifier(firstParam) &&
-    !hasTypes(node, firstParam)
+    !hasTypes(node, firstParam) &&
+    !node.async
   ) {
     if (
       this.format.retainLines &&
```
JLHwung, line 116: We can move `node.async` to here
```
this.format.retainLines || node.async
```
so we are instructing that if `async` presents, we try to retain lines and print `(` if the function body is in different lines with the async arrow.

cwohlman:
Thanks @JLHwung, looks like that does the trick.

PR #11837, opened 2020-07-14
chore: use repository HEAD when pulling third party repos

JLHwung:

The `test262` is [planning](https://github.com/tc39/test262/issues/2699) to rename `master` into `main`. In this PR we removes the `--branch` option when cloning third-party repos. By doing so `git` will always clone the repository `HEAD`, which is the default branch HEAD as of GitHub repos -- so we are agnostic about upstream default branch name.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11837"></a>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,7 +88,7 @@ jobs:
       - run:
           name: Setup Test Runner
           command: |
-            git clone --recurse-submodules https://github.com/babel/babel-test262-runner
+            git clone --depth=1 --recurse-submodules https://github.com/babel/babel-test262-runner
             cd babel-test262-runner
             npm ci
             npm i tap-mocha-reporter --save-dev
--- a/packages/babel-compat-data/scripts/download-compat-table.sh
+++ b/packages/babel-compat-data/scripts/download-compat-table.sh
@@ -16,5 +16,5 @@ fi
 
 rm -rf build/compat-table
 mkdir -p build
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -88,7 +88,7 @@ jobs:
       - run:
           name: Setup Test Runner
           command: |
-            git clone --recurse-submodules https://github.com/babel/babel-test262-runner
+            git clone --depth=1 --recurse-submodules https://github.com/babel/babel-test262-runner
```
JLHwung, line 91: Note that `--depth=N` implies `--single-branch`.

packages/babel-compat-data/scripts/download-compat-table.sh
```diff
@@ -16,5 +16,5 @@ fi
 
 rm -rf build/compat-table
 mkdir -p build
-git clone --branch=gh-pages --single-branch --shallow-since=2020-04-01 https://github.com/kangax/compat-table.git build/compat-table
-cd build/compat-table && git checkout -qf $COMPAT_TABLE_COMMIT
+git clone --single-branch --shallow-since=2020-04-01 https://github.com/kangax/compat-table.git build/compat-table
+cd build/compat-table && git checkout -q $COMPAT_TABLE_COMMIT
```
JLHwung, line 20: `-f` is to suppress warning when working tree is different than HEAD. This is highly unlikely since we have removed `build/compat-table` before cloning.

PR #11842, opened 2020-07-15
Replace lodash 'escapeRegExp' with escape-string-regexp library

jayaddison:

From [previous discussion](https://github.com/babel/babel/pull/11789#discussion_r452863345) about removing `lodash` dependencies, it sounds like we can use the widely-used [escape-string-regexp](https://www.npmjs.com/package/escape-string-regexp) library to replace the functionality provided by `lodash/escapeRegExp`.

As noted in that discussion, this module requires Node 10 or greater, and that won't be guaranteed in `babel` until version 8, so this changeset is based against the `next-8-dev` branch.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11842"></a>

### DIFF
--- a/packages/babel-core/package.json
+++ b/packages/babel-core/package.json
@@ -53,6 +53,7 @@
     "@babel/types": "workspace:^7.10.4",
     "convert-source-map": "^1.7.0",
     "debug": "^4.1.0",
+    "escape-string-regexp": "^4.0.0",
     "gensync": "^1.0.0-beta.1",
     "json5": "^2.1.2",
     "lodash": "^4.17.13",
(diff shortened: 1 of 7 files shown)


JLHwung:
@jayaddison Updating the `yarn.lock` should fix the e2e error.

nicolo-ribaudo:
Thanks!

jayaddison:
Eh, sort of a nitpick of myself here, but I realized I should have said 'yes' to the 'dependency changes' question in the pull request description Q&A.  I'll edit that in the description now (even though it's post-review and post-commit).

nicolo-ribaudo:
Oh thanks! That table is only used by us to add PR labels, which are used in the changelog and by reviewers!

jayaddison:
Good to know, thanks @nicolo-ribaudo!

PR #11838, opened 2020-07-14
refactor: replace caniuse-db by mdn-browser-compat-data

JLHwung:
| Q                        | A 
| ------------------------ | ---
| Patch: Bug Fix?          | `es6.modules` compat-data should include `node: 13.2`.
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

In this PR I have replaced `caniuse-db` by `mdn-browser-compat-data` when building the modules support in `@babel/compat-data`. The rationale here is

- `caniuse` [states](https://caniuse.com/#feat=mdn-javascript_statements_export) that the support data of `es6-modules` are from `mdn/browser-compat-data`
- I have submitted a PR for [`export * as ns` support](https://github.com/mdn/browser-compat-data/pull/6394), which enables us to include `export-namespace-from` in the `preset-env`.
- The `mdn-browser-compat-data` is easier to process than `caniuse-db` (no more `getLowestImplementedVersions`)
- `caniuse` does not have `nodejs` support data

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11838"></a>

PR #11849, opened 2020-07-17
Add `export-namespace-from` to `preset-env`

### LINKED ISSUE #11363: Add `proposal-export-namespace-from` to `preset-env`
## Bug Report

**Current Behavior**
When `@babel/preset-env` is enabled, export namespace syntax causes Babel to emit an error:

```
SyntaxError: /[...]/index.js: Unexpected export specifier type
> 1 | export * as ns from './foo'
```

A different, less descriptive error is emitted when `{loose: true}` is passed to the preset:

```
TypeError: /[...]/index.js: Cannot read property 'name' of undefined
```

**Input Code**
See [this repo](https://github.com/billyjanitsch/babel-export-ns-bug) for a minimal reproduction. Run `npm i && npm run build` to observe the error. Removing `@babel/preset-env` from the config prevents the error. So does providing `{modules: false}` to the preset.

```js
export * as ns from './foo'
```

**Expected behavior/code**
I would expect this code to transpile without error.

- Filename: `babel.config.js`

```js
module.exports = {
  // comment out this line to fix the build:
  presets: ['@babel/env']
}
```

**Environment**

```
  System:
    OS: macOS Mojave 10.14.6
  Binaries:
    Node: 13.11.0 - ~/.config/nvm/13.11.0/bin/node
    npm: 6.14.4 - ~/.config/nvm/13.11.0/bin/npm
  npmPackages:
    @babel/cli: ^7.8.4 => 7.8.4 
    @babel/core: ^7.9.0 => 7.9.0 
    @babel/preset-env: ^7.9.0 => 7.9.0
```

JLHwung:
| Fixed Issues?            | Closes #11363
| Documentation PR Link    | 
| Any Dependency Changes?  | `preset-env` has two new deps

- Added compat-data from `mdn-browser-compat-data`
- Added a new caller support flag: `supportsExportNamespaceFrom`, we should also update `babel-loader` and [`@rollup/plugin-babel`](https://github.com/rollup/plugins/blob/master/packages/babel/src/index.js) since both two have supported `export * as ns`.
  * WebPack: https://github.com/webpack/webpack/pull/11117
  * Rollup: https://github.com/rollup/rollup/pull/3193
- When `modules: false` (preserve ES modules) is specified, compat-data is used to determine whether `export * as ns` should be transpiled

This PR includes commits from #11838, I will rebase once it gets merged.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11849"></a>


billyjanitsch:
Thanks @JLHwung! Sorry I didn't get around to working on this.

packages/babel-preset-env/src/index.js
```diff
@@ -273,11 +286,20 @@ export default declare((api, opts) => {
...
       modules !== "auto" || !api.caller?.(supportsDynamicImport),
+    shouldTransformExportNamespaceFrom: !(
+      (modules === "auto" && api.caller?.(supportsExportNamespaceFrom)) ||
+      (modules === false &&
+        !isRequired("proposal-export-namespace-from", transformTargets, {
+          compatData,
+          includes: include.plugins,
+          excludes: exclude.plugins,
+        }))
+    ),
```
nicolo-ribaudo, line 297: Nit: This is a bit hard to follow. Could we move the `!` inside to match the same `!`/`||` pattern used in the lines above?

```suggestion
    shouldTransformExportNamespaceFrom: 
      (modules !== "auto" || !api.caller?.(supportsExportNamespaceFrom)) &&
      (modules !== false || isRequired("proposal-export-namespace-from", transformTargets, {
          compatData,
          includes: include.plugins,
          excludes: exclude.plugins,
        }))
    ),
```

Or maybe even this doesn't change the readability 🤔😅

JLHwung: Well I initially unwrap `!` but these two `modules !== false || something` seems pretty mind-jiggling to me. I end up with current patterns. I can wrap these conditions into a new `shouldSkipExportNamespaceFrom` and it should be more straightforward.

JLHwung:
I am merging this PR to `feat-7.11.0/stage-4` feature branch.

PR #11855, opened 2020-07-20
Replace lodash 'isPlainObject' usage with is-plain-object library

jayaddison:

As per feedback in https://github.com/babel/babel/pull/11789#issuecomment-658982855, this is another instance of a `lodash` function that can be replaced by a standalone library.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11855"></a>

### DIFF
--- a/packages/babel-types/package.json
+++ b/packages/babel-types/package.json
@@ -17,7 +17,7 @@
   "types": "lib/index.d.ts",
   "dependencies": {
     "@babel/helper-validator-identifier": "^7.10.4",
-    "is-plain-object": "^3.0.1",
+    "is-plain-object": "^4.0.0",
     "lodash": "^4.17.19",
     "to-fast-properties": "^2.0.0"
   },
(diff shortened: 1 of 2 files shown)


jayaddison:
The two implementations seem to differ when determining whether an object created by using `Object.create(null)` is a plain object:

```javascript
> isplainobject_standalone = require('is-plain-object');
[Function: isPlainObject]
> isplainobject_lodash = require('lodash/isPlainObject');
[Function: isPlainObject]
> var x = Object.create(null);
undefined
> isplainobject_standalone(x);
false
> isplainobject_lodash(x);
true
```

TrySound:
Fixed this. Try v4
https://github.com/jonschlinkert/is-plain-object/releases/tag/v4.0.0

jayaddison:
(switching to draft because I'm not sure what factors should determine the approach between this and https://github.com/babel/babel/pull/11859)

kaicataldo:
Any reason we can't just use [lodash.isplainobject](https://www.npmjs.com/package/lodash.isplainobject) instead? I don't think we need to get rid of lodash entirely, but it would be great if we could just install the utils we're using individually.

TrySound:
`lodash.*` packages are deprecated and will not be supported in lodash v5.

kaicataldo:
Ah, okay, that's a very good reason :+1:

jayaddison:
Closing given a marginal preference for #11859 instead.

PR #11856, opened 2020-07-20
Replace lodash 'isRegExp' usage with NodeJS util isRegExp (or a fallback when unavailable)

jayaddison:

Similar to https://github.com/babel/babel/pull/11855/, this changeset replaces a further `lodash` dependency (in this case, `isRegExp`) with a standalone alternative (the `is-regexp` library function).

Edit: also for context, adding a reference to https://github.com/babel/babel/pull/11789 (this is one of a collection of pull requests relating to reducing dependency upon `lodash`)

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11856"></a>


packages/babel-types/src/converters/valueToNode.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import isPlainObject from "lodash/isPlainObject";
-import isRegExp from "lodash/isRegExp";
+import isRegExp from "is-regexp";
```
JLHwung, line 3: `_.isRegExp` will call optimized node.js builtin `util.isRegExp` before `toString` comparison. Given that most Babel users are using node.js, I would suggest in this case we create our own `isRegExp` helper.
```js
function isRegExp(value) {
  try {
    return require("util").isRegExp(value);
  } catch {
    return Object.prototype.toString.call(value) === '[object RegExp]';
  }
}

Comment (JLHwung) [reply packages/babel-types/src/converters/valueToNode.js:3]: Oh `util.isRegExp` is deprecated, maybe we should target this change to Babel 8 via https://nodejs.org/api/util.html#util_util_types_isregexp_value.

Comment (jayaddison) [reply packages/babel-types/src/converters/valueToNode.js:3]: Is it worth somehow pushing the `require` operation to a higher (module?) scope so that the `try...require` is only evaluated once?

Comment (jayaddison) [reply packages/babel-types/src/converters/valueToNode.js:3]: Ah, I missed your second comment before replying.  Yep, I'll retarget this against Babel 8 when applying a change to use a built-in NodeJS regexp test.

Comment (JLHwung) [reply packages/babel-types/src/converters/valueToNode.js:3]: > Is it worth somehow pushing the require operation to a higher (module?) scope so that the try...require is only evaluated once?

Good idea, we can even get rid of try catch here 
```
var nodeUtil = (function() {
  try {
    return require('util');
  } catch {}
}());

isRegExp = value =>
   nodeUtil ? nodeUtil.isRegExp(value) : Object.prototype.toString.call(value) === '[object RegExp]';
```

Comment (jayaddison) [reply packages/babel-types/src/converters/valueToNode.js:3]: Hrm.  After a little more time to pause I'm less certain what to do here.  The `process.binding` pattern doesn't appear elsewhere in the codebase, and although `nodejs` may be relatively well-adopted by `babel` users, I don't see too many references to platform-specific code (just [one](https://github.com/babel/babel/blob/e2b33c2cb3272cf46d4ca29097dc331d7570fca5/packages/babel-core/src/config/files/module-types.js#L60) or [two](https://github.com/babel/babel/blob/e2b33c2cb3272cf46d4ca29097dc331d7570fca5/packages/babel-register/src/cache.js#L45) mentions of nodejs-related workarounds).

I'm not averse to applying those changes if they're definitely the way to go, but I'd prefer to understand a bit more about the options and implications first.

Comment (coreyfarrell) [reply packages/babel-types/src/converters/valueToNode.js:3]: Note that `process.binding` is also deprecated (see [DEP0103](https://nodejs.org/dist/latest/docs/api/deprecations.html#deprecations_dep0103_process_binding_util_is_typechecks) and [DEP0111](https://nodejs.org/dist/latest/docs/api/deprecations.html#DEP0111)).  For using Node.js methods I would probably go for:
```js
const nodeUtilTypes = util.types || util;
```

This way you would use `util.types.isRegExp` in Node.js 10+, `util.isRegExp` in older versions.  Although `util.isRegExp` is deprecated in Node.js 4+ it's never going to be removed or produce a warning in Node.js < 10 so it seems like a reasonable fallback.

JLHwung: As of `process.binding()`, I meant to avoid interactions with bundlers -- that `require("utils")` is intended to be failed on browsers. Since we have offered a fallback, we don't expect the bundlers do anything about `require("utils")` So users don't have to provide `node: { utils: empty }` to Webpack or use `rollup-plugin-node-polyfills` that may even try to polyfill `utils.isRegExp`.

As a note, `@babel/standalone` bundles `@babel/types` and it is run in browsers. Building `@babel/standalone` does not fail because we have `rollup-plugin-node-polyfills`, but I don't expect `@babel/types` _bundling_ users have to change their bundler config in a non-breaking releases.

kaicataldo:
Any reason we can't just use [lodash.isregexp](https://www.npmjs.com/package/lodash.isregexp) instead? I don't think we need to get rid of lodash entirely, but it would be great if we could just install the utils we're using individually.

TrySound:
https://github.com/lodash/lodash/issues/3838

AmyShackles:
Probably a stupid question, but why not just test for the constructor?
```
let reg = /literal/gi
let reg2 = new RegExp(/constructorWithRegexPattern/, "gi");
let reg3 = new RegExp("constructorWithString", "gi");
let reg4 = RegExp(/factoryNotationWithRegexPattern/, "gi");
let reg5 = RegExp("factoryNotationWithString", "gi");
console.log(reg.constructor.name) // "RegExp"
console.log(reg2.constructor.name) // "RegExp"
console.log(reg3.constructor.name) // "RegExp"
console.log(reg4.constructor.name) // "RegExp"
console.log(reg5.constructor.name) // "RegExp"
```

It seems like you would just need to test for undefined/null and then check if the constructor was equivalent to RegExp.  Then you wouldn't need to bring in a util or use lodash (for this particular piece of code).

```
if (value && value.constructor === RegExp) {...}
```

JLHwung:
@AmyShackles The `constructor` property can be overwritten for non-primitive values.
```js
const foo = {};
foo.constructor = RegExp;
foo.constructor === RegExp // returns true
```

packages/babel-types/src/converters/valueToNode.js
```diff
@@ -16,6 +15,18 @@ import {
...
+try {
+  const util = require("util");
+  nodeUtilTypes = util.types || util;
+} catch {}
+
+const isRegExp = value => {
+  return nodeUtilTypes
+    ? nodeUtilTypes.isRegExp(value)
+    : Object.prototype.toString.call(value) === "[object RegExp]";
+};
```
ExE-Boss, line 28: The `nodeUtilTypes` check only needs to be performed once:

```suggestion
const isRegExp = nodeUtilTypes?.isRegExp || value => {
  return Object.prototype.toString.call(value) === "[object RegExp]";
};
```

jayaddison: Accepted, thanks!  (locally rather than via the GitHub UI, because I figured a merge from `main` would be worthwhile at the same time)

jayaddison: Hrm.. looks like a small issue with the optional chaining operator.  Reading up a little on the syntax and compatibility of it now.. hopefully a small fix.

jayaddison: As best I can tell, the problem was that the optional chaining operator (ES2020) wasn't available at compile-time (Node 6.x), so it's now replaced with a truthiness check.

packages/babel-types/src/converters/valueToNode.js
jayaddison: @JLHwung @coreyfarrell The code structure has been refactored a bunch since the previous comments in this discussion; in particular the top-level `util` module is always imported, and then a local `isRegExp` helper is assigned based on a series of preferentially-ordered fallbacks.  Hopefully that meets most of the optimization goals mentioned while also catering for older Node versions; glad for further feedback.

jayaddison:

PR #11778, opened 2020-07-01
upgrade: ESLint v7

kaicataldo:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | 
| Patch: Bug Fix?          | :+1:
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | Yes
| Documentation PR Link    | 
| Any Dependency Changes?  |
| License                  | MIT

This updates all of our packages to have a peer dependency on ESLint v7 (the current release line), as well as to use v7 to lint the repo. The biggest difference is some stricter validation in the RuleTester (which is why I needed to update tests).

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11778"></a>

PR #11857, opened 2020-07-20
Upgrade to ESLint v7

kaicataldo:
| Fixed Issues?            |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

This updates all of our packages to have a peer dependency on ESLint v7 (the current release line), as well as to use v7 to lint the repo. The biggest difference is some stricter validation in the RuleTester (which is why I needed to update tests).

Opening this in favor of #11778.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11857"></a>


kaicataldo:
Removing `babel-eslint` from the top-level package.json leads to installation failing (locally as well).

>  error An unexpected error occurred: "ENOENT: no such file or directory, lstat '/home/runner/work/babel/babel/eslint/babel-

I think it's because `babel-eslint` has `semver` as a dependency and so the linked package is able to resolve the module. Is there another way we can consume the `babel-eslint-*` packages in the top-level `package.json` that will correctly install the deps?

kaicataldo:
I _think_ I've found a way to use the local versions of the `babel-eslint-*` packages while still linking them in dev. Please let me know what you think!

scripts/integration-tests/e2e-babel-old-version.sh
```diff
@@ -21,6 +21,10 @@ cd ../..
 #==============================================================================#
 
 startLocalRegistry "$PWD"/scripts/integration-tests/verdaccio-config.yml
+
+# Install dependencies in individual packages so that we can link them at the top level.
+for package in eslint/*/; do yarn --production --cwd $$package; done
```
kaicataldo, line 26: Alternative we could make this a Make target

kaicataldo:
CI is currently failing because ESLint v7 has dropped support for Node <10 . This means that installation is failing on Node 6/8. I think it makes the most sense to use Yarn's `--ignore-engines` flag in Node 6/8. Thoughts?

kaicataldo:
Actually, it looks like we already do this: https://github.com/babel/babel/blob/main/Makefile#L265. Hmmm...

kaicataldo:
Oh I see, it's from the new Yarn installs I added 😄. Adding the flag now.

Makefile
```diff
@@ -258,7 +258,7 @@ bootstrap-only: lerna-bootstrap
 
 yarn-install: clean-all
 	# Install dependencies in individual packages so that we can link them at the top level.
-	for package in eslint/*/; do yarn --production --cwd $$package; done
+	for package in eslint/*/; do yarn --cwd $$package; done
```
hzoo: what's the difference between $$package and $package? (just looking at the other files)

JLHwung: I think this is how `$` should be escaped in Makefile.

kaicataldo: That's right! I had to look this up the first time I edited this file haha

PR #11859, opened 2020-07-21
Replace lodash 'isPlainObject' usage with is-plain-obj library

jayaddison:

This is an alternative to https://github.com/babel/babel/pull/11855.  What's the best selection between two libraries?  There are often `n` different factors to weigh up.

Some considerations here are:

- Popularity is comparable between `is-plain-object` and `is-plain-obj`
- `is-plain-obj` appears to have aimed towards compatibility with `lodash` in the past, and has had a stable release under that basis for a longer period of time
- `is-plain-obj` has a smaller unpacked size
- Test coverage appears broadly similar between the two; `is-plain-obj` may test more scenarios while `is-plain-object` provides in-browser-environment test coverage

Rather than attempt to sway a decision by offering only a single pull request it might be safest to open two and allow discussion.

In retrospect it might've been worth more fully exploring and evaluating the options available during a [previous thread](https://github.com/babel/babel/pull/11789#issuecomment-658982855).

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11859"></a>


jayaddison:
(switching to draft because I'm not sure what factors should determine the approach between this and https://github.com/babel/babel/pull/11855)

TrySound:
is-plain-obj requires node 8
https://github.com/sindresorhus/is-plain-obj/blob/master/package.json#L13

jayaddison:
> is-plain-obj requires node 8

That's a good consideration.  Honestly I don't feel well-experienced enough with the JS dependency ecosystem or `babel`'s roadmap to determine what to do at the moment, so I'll step back for a bit!

jayaddison:
After taking some time away from this and reconsidering, I think the approach I'd suggest to replace `isPlainObject` is to use `is-plain-obj` as proposed here in #11859, instead of `is-plain-object` as proposed in #11855.

It might seem a tiny and arbitrary choice, but I don't love the fact that in #11855 the external dependency had to be updated to match the application's use case (but thanks @TrySound for discovering and applying that change upstream, which should benefit other users).  That seems to indicate (to me at least) that there's less stability in the interface and behaviour of the `is-plain-object` library -- and that might be fine in some cases, but for a single-purpose library function it does raise some questions.  By contrast, `is-plain-obj` seems to prefer stability and avoidance of unnecessary releases, and core ownership and authorship of the library seem stronger based on the library's commit and merge history.

The main downside to the `is-plain-obj` approach is the [required upgrade](https://github.com/babel/babel/pull/11859#issuecomment-661784567) to Node 8 or later.  A counter-argument to that is that we're already on track to require a Babel 8 release (and therefore Node >= 10) before dropping `lodash` completely due to #11842.

jayaddison:

PR #11862, opened 2020-07-21
Correctly check reserved word for PropertyDefinition: IdentifierReference

### LINKED ISSUE #10411: [parser] `arguments` and `eval` as object shorthand should be valid in strict mode
## Bug Report

The object key shorthand syntax is checking the ident for being a keyword and triggers an error. This is not correct for `eval` and `arguments`, which are only read (which is legal).

**Current Behavior**

It will throw a syntax error `Unexpected keyword 'eval' (1:2)`

**Input Code**

https://astexplorer.net/#/gist/15eb31fbb1024faa6f42a4940d8beb4a/06e20b27a9291d5356faa340e430c0ded986ccb8

It will properly apply sloppy/strict mode exceptions (script vs module goal). So `eval` won't throw with script goal.

```js
({arguments})
```
```js
({eval})
```

Not sure if same bug, but definitely related is the next one, where the `eval` identifier is a regular object key, not a assignment destructuring pattern (because it assigns to `.x` of the object literal). Let me know if this is the same cause, otherwise I'll file a separate issue for this.

```
({...{eval}.x} = {});
```

**Expected behavior/code**

It should not throw an error because it's a regular object key and a read from `eval` / `arguments`, which is allowed.

**Environment**

Babel 7.5.5

**Possible Solution**

Add exception path for `eval` and `arguments` when doing the keyword check.

JLHwung:
| Fixed Issues?            | Fixes #10411

For object shorthand property, e.g. `({ arguments })` is allowed because
```
PropertyDefinition -> IdentifierReference -> Identifier -> IdentifierName but not ReservedWord
```
and `arguments` and `eval` is _not_ ReservedWord

For destructuring patterns, e.g. `({ arguments = 1 }) = {}` is allowed (in sloppy mode) because
```
PropertyDefinition -> CoverInitializedName -> IdentifierReference -> Identifier -> IdentifierName but not ReservedWord
```

If the fourth argument of `this.checkReservedWord` is `true`, it should be applied on BindingIdentifier, where `eval` and `arguments` are forbidden. Therefore the fix here is to align the spec and pass `false` instead.

This PR fixes a downstream issue: https://github.com/angular/angular-cli/issues/17439

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11862"></a>


nicolo-ribaudo:
> For destructuring patterns, e.g. `({ arguments = 1 }) = {}` is allowed because

I don't think so: the destructuring pattern must match the "AssignmentPattern" production, and:

#### 12.15.7.1 Static Semantics: Early Errors
_AssignmentProperty : IdentifierReference Initializer[opt]_
- It is a Syntax Error if AssignmentTargetType of IdentifierReference is not `simple`.

#### 12.1.3 Static Semantics: AssignmentTargetType
_IdentifierReference : Identifier_
1. If this _IdentifierReference_ is contained in strict mode code and StringValue of _Identifier_ is **"eval"** or **"arguments"**, return `invalid`.
2. Return `simple`.

JLHwung:
> If this IdentifierReference is contained in strict mode code and StringValue of Identifier is "eval" or "arguments", return invalid.

As for destructuring, we will check the BindingIdentifier ReservedWord in `checkLVal` later: Thus
```js
"use strict";
const { arguments } = foo();
```
still [throws](https://github.com/babel/babel/pull/11862/files#diff-4496a10e403e1fa4b802d2bb36bbe7a6). 
Since we can not differentiate between `CoverInitializedName` and `AssignmentProperty` until we see an `=` or `=>`, I think the change here is okay.

PR #11871, opened 2020-07-23
Parser refactoring

JLHwung:
| Fixed Issues?            | [REPL 1](https://babel.dev/repl#?browsers=&build=&builtIns=false&spec=true&loose=false&code_lz=PTAEGcAsHsFcBsAmoAukBO0DuAoAhuAJ4B2AxqAB44C8AfJUA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Ces2015%2Creact%2Cstage-3%2Cenv&prettier=false&targets=&version=7.10.5&externalPlugins=), [REPL 2](https://babel.dev/repl#?browsers=&build=&builtIns=false&spec=true&loose=false&code_lz=PTAEGcAsHsFcBsAmoB20Aup2QE7QO4BQApgB4AO0OmAhuAJ4oA6sA3gGwDMAvqAGZ4AtqABEfaNBEBuIA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Cstage-1&prettier=false&targets=&version=7.10.5&externalPlugins=)
| Documentation PR Link    | 
| Any Dependency Changes?  |

Refactors `@babel/parser` a bit. Focused on code readability and syntax upstream spec links.
Happens to fix a bug when adding grammar production comments.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11871"></a>


packages/babel-parser/test/fixtures/flow/type-annotations/20/output.json
```diff
@@ -34,7 +34,6 @@
                 },
                 "computed": false,
                 "kind": "set",
-                "variance": null,
```
JLHwung, line 37: The `variance` property exists in ClassProperty, and we don't have `variance` for ObjectMethod. This PR removes `variance` of `fooProp` accessor for the following cases:
```js
a={set fooProp(value:number){}}
```

I think the new behaviour is preferred and it aligns to what we did for non-accessors.

packages/babel-parser/src/parser/statement.js
```diff
@@ -1281,10 +1297,7 @@ export default class StatementParser extends ExpressionParser {
       prop.static = false;
       classBody.body.push(this.parseClassProperty(prop));
       return true;
-    } else if (containsEsc) {
```
JLHwung, line 1284: This branch is unreachable because `parseClassMemberFromModifier` is guarded by `isContextual` call, which must return `false` when `containsEsc`.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1672,12 +1741,27 @@ export default class ExpressionParser extends LValParser {
...
-      isAsync = false;
+    if (
+      !isPattern &&
+      !isGenerator &&
+      !containsEsc &&
+      this.maybeAsyncOrAccessorProp(prop)
+    ) {
+      // https://tc39.es/ecma262/#prod-AsyncMethod
+      // https://tc39.es/ecma262/#prod-AsyncGeneratorMethod
+      if (prop.key.name === "async" && !this.hasPrecedingLineBreak()) {
```
JLHwung, line 1752: Parsing `async` and `set`, `getter` have been unified in `parsePropertyDefinition` -- so they can share same sanity checks. Thus we don't need passing `containsEsc` (A tokenizing level states) to `parseObjectMethod`, instead `isAccessor` is pass.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1223,19 +1273,22 @@ export default class ExpressionParser extends LValParser {
...
     // here is that parseIdentifier will remove an item from the expression stack
     // if "function" or "class" is parsed as identifier (in objects e.g.), which should not happen here.
-    let meta = this.startNode();
-    this.next();
-    meta = this.createIdentifier(meta, "function");
+    this.next(); // eat `function`
 
-    if (this.prodParam.hasYield && this.eat(tt.dot)) {
+    if (this.prodParam.hasYield && this.match(tt.dot)) {
+      const meta = this.createIdentifier(
```
JLHwung, line 1287: Avoid creating an extra `meta` node when `tt.dot` is not met, instead we create `meta` when we have to. The same technique is also applied for `parseNewOrNewTarget`.

packages/babel-parser/src/parser/expression.js
```diff
@@ -594,163 +602,205 @@ export default class ExpressionParser extends LValParser {
     state: N.ParseSubscriptState,
```
JLHwung, line 602: `parseSubscript` has been broken into smaller pieces. I find it more easier to go through after refactoring: https://github.com/JLHwung/babel/blob/ee4d63ef00e5090f44ffedee718dd18fdc6e21e0/packages/babel-parser/src/parser/expression.js#L597

JLHwung:
I have changed the PR label because the bug fixed here is edging cases. It is more of a refactoring which is very likely invisible to users.

packages/babel-parser/src/parser/lval.js
```diff
@@ -314,20 +320,22 @@ export default class LValParser extends NodeUtils {
...
+  // https://tc39.es/ecma262/#prod-BindingElement
   parseMaybeDefault(
     startPos?: ?number,
     startLoc?: ?Position,
     left?: ?Pattern,
   ): Pattern {
-    startLoc = startLoc || this.state.startLoc;
-    startPos = startPos || this.state.start;
-    left = left || this.parseBindingAtom();
+    startLoc = startLoc ?? this.state.startLoc;
```
nicolo-ribaudo, line 335: ~~Maybe now we can use `??=`? :smile:~~

Flow :disappointed:

packages/babel-parser/src/parser/expression.js
```diff
@@ -482,28 +480,28 @@ export default class ExpressionParser extends LValParser {
   }
 
   // Parse unary operators, both prefix and postfix.
-
-  parseMaybeUnary(refExpressionErrors: ?ExpressionErrors): N.Expression {
+  // https://tc39.es/ecma262/#prod-UnaryExpression
+  parseUnary(refExpressionErrors: ?ExpressionErrors): N.Expression {
```
nicolo-ribaudo, line 484: I think that this was called `parseMaybeUnary` (like `parseMaybeConditional`, `parseMaybeAssign`) because it could return something which is not a unary expression, like a simple identifier.

I know that _technically_ Identifier is a child of UnaryExpression in the spec, but it's just a side effect of how it's specified.

packages/babel-parser/src/parser/expression.js
```diff
@@ -1919,6 +1994,43 @@ export default class ExpressionParser extends LValParser {
     return node;
   }
 
+  // parse an array literal or tuple literal
+  // https://tc39.es/ecma262/#prod-ArrayLiteral
+  // https://tc39.es/proposal-record-tuple/#prod-TupleLiteral
+  parseArrayLike(
+    close: TokenType,
+    isTuple: boolean,
+    canBePattern: boolean,
```
nicolo-ribaudo, line 2003: Nit: `isTuple` and `canBePattern` are swapped in the `parseObjectLike` function. It might be good to also swap them here to avoid confusion.

PR #11879, opened 2020-07-27
chore: use modules: "auto"

JLHwung:

Use `modules: "auto"` (defaults) when building Babel. See https://github.com/babel/website/pull/2301 for the rationale.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11879"></a>

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -7,9 +7,7 @@ module.exports = function (api) {
 
   const envOptsNoTargets = {
     loose: true,
-    modules: false,
     shippedProposals: true,
-    exclude: ["transform-typeof-symbol"],
   };
   const envOpts = Object.assign({}, envOptsNoTargets);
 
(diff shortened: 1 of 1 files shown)


babel.config.js
```diff
@@ -39,6 +37,7 @@ module.exports = function (api) {
     case "standalone":
       includeRegeneratorRuntime = true;
       unambiguousSources.push("packages/babel-runtime/regenerator");
+      envOpts.exclude = ["transform-typeof-symbol"]; // exclude slow plugins
```
JLHwung, line 40: `transform-typeof-symbol` is skipped since node.js 0.12, we only need to exclude this plugin when building `@babel/standalone`.

hzoo: We shouldn't even need to exclude since loose + preset-env should do this automatically actually haha

hzoo: oh here: https://github.com/babel/babel/blob/3a0aa8e829ee8b0af9b01df5f581aad6a3f126c0/packages/babel-preset-env/src/get-option-specific-excludes.js#L3

PR #11883, opened 2020-07-28
feat(babel‑types): Add type definitions for Node assertion methods

ExE-Boss:
| Fixed Issues?            | 
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 

This adds **TypeScript** and **Flow** type definitions for the `assert${typeName}` functions.

It also makes use of the `typesVersions` field, which was introduced in **TypeScript 3.1** to make the `assert${typeName}` functions into assertion functions without breaking backwards compatibility, like what I did with `@types/babel‑types` in <https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45830>.

This also now generates `/** @deprecated */` **JSDoc** comments for deprecated node type checking functions.

---

review?(@JLHwung, @MichaReiser, @nicolo-ribaudo)

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11883"></a>


packages/babel-types/package.json
```diff
@@ -15,6 +15,11 @@
   },
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
+  "typesVersions": {
+    ">=3.7": {
+      "lib/*.d.ts": ["lib/*.ts37.d.ts"]
```
JLHwung, line 20: nit: I think we can swap the name of `ts37.d.ts` with `index.d.ts`. So `index.d.ts` always align to latest TS, and we can have `lib/legacy.d.ts` as fallback.
```json
{
  "types": "lib/legacy.d.ts",
  "typesVersions": {
    ">=3.7": {
      "*": "lib/index.d.ts"
    }
}
```

ExE-Boss: The thing is, importing `lib/index.js` directly would cause old **TypeScript** versions to try to load `lib/index.d.ts`, but that would then have the **TypeScript 3.7** syntax.

Using `index.ts3.7.d.ts` also ensures that we can easily add `index.tsX.Y.d.ts` (e.g.: `index.ts4.0.d.ts`) in the future.

JLHwung: We will disable submodule imports in Babel 8. And we could drop ts <3.0 support, too. So `tsc` will always respect `typeVersions` and `types` and we can enforce that `lib/index.d.ts` always aligns to latest TS while `typeVersions` serves a fallback for older `ts>=3` versions.

ExE-Boss: `typesVersions` is only supported since **TypeScript 3.1**.

nicolo-ribaudo (approved):

nicolo-ribaudo:
I didn't notice this PR and did https://github.com/babel/babel/commit/a9bc9becc527e74bf2848b048d556144bad3f5c3 yesterday. However, this PR still added some improvements (`assert*` definitions for older TS versions, and `@deprecated` annotations) so I have rebased it.

Thanks! :smile:

PR #11894, opened 2020-07-29
Prepare to publish `@babel/eslint-*` packages

### LINKED ISSUE #10709: ESLint-related packages migration - TODO

JLHwung:
| Fixed issue | Closes #10709, resolves https://github.com/babel/babel-eslint/issues/787
| Any Dependency Changes?  | Bump `node` requirements of `@babel/eslint-*` packages.

This PR includes preparations need to publish `@babel/eslint-*` packages. It includes commits in #11857, I will rebase once that PR gets merged.

Some noticeable changes compared to #11857:

- Require `node` to be `^10.13.0 || ^12.13.0 || >=14.0.0`. It covers only supported node versions as of July 2020
- Disable submodule exports
- Require `babel` to be `>=7.11.0` since it is the Babel version that these packages are tested against

Note that `@babel/eslint-shared-fixtures` and `@babel/eslint-tests` are still private packages. I can not think of any use case of them and they looks to me more like a side-effect of sharing internal utils across different packages in a monorepo.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11894"></a>

### DIFF
--- a/eslint/babel-eslint-parser/package.json
+++ b/eslint/babel-eslint-parser/package.json
@@ -2,9 +2,8 @@
   "name": "@babel/eslint-parser",
   "version": "7.10.4",
   "description": "ESLint parser that allows for linting of experimental syntax transformed by Babel",
-  "author": "Sebastian McKenzie <sebmck@gmail.com>",
+  "author": "The Babel Team (https://babeljs.io/team)",
   "license": "MIT",
-  "private": true,
   "repository": {
     "type": "git",
     "url": "https://github.com/babel/babel.git",
@@ -15,11 +14,11 @@
   },
   "homepage": "https://babeljs.io/",
   "engines": {
-    "node": ">=10.9"
+    "node": "^10.13.0 || ^12.13.0 || >=14.0.0"
   },
   "main": "lib/index.js",
   "peerDependencies": {
-    "@babel/core": ">=7.10.4",
+    "@babel/core": ">=7.11.0",
     "eslint": ">=7.5.0"
   },
   "dependencies": {
--- a/eslint/babel-eslint-parser/package.json
+++ b/eslint/babel-eslint-parser/package.json
@@ -16,7 +16,12 @@
   "engines": {
     "node": "^10.13.0 || ^12.13.0 || >=14.0.0"
   },
-  "main": "lib/index.js",
+  "main": "./lib/index.js",
+  "type": "commonjs",
+  "exports": {
+    ".": "./lib/index.js",
+    "./package.json": "./package.json"
+  },
   "peerDependencies": {
     "@babel/core": ">=7.11.0",
     "eslint": ">=7.5.0"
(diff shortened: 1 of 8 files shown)


eslint/babel-eslint-parser/package.json
```diff
@@ -15,12 +14,17 @@
   },
   "homepage": "https://babeljs.io/",
   "engines": {
-    "node": ">=10.9"
+    "node": "^10.13.0 || ^12.13.0 || >=14.0.0"
```
JLHwung, line 17: I feel like we can enforce the same ranges for Babel 8: that we don't have to fight against early node v13 versions that may complain about incorrect `exports` usage.

eslint/babel-eslint-plugin-development-internal/package.json
```diff
@@ -18,15 +18,12 @@
   "author": "Kai Cataldo <kai@kaicataldo.com>",
   "license": "MIT",
   "private": true,
-  "engines": {
```
JLHwung, line 21: I removed the `engines` requirement here because its requirement can be implied from its peerDeps: `@babel/eslint-parser`.

eslint/babel-eslint-parser/package.json
```diff
@@ -23,7 +23,7 @@
     "./package.json": "./package.json"
   },
   "peerDependencies": {
-    "@babel/core": ">=7.11.0",
+    "@babel/core": ">=7.10.0",
```
kaicataldo, line 26: Why do we need to change this back to 7.10.0?

JLHwung: Footgun: https://github.com/babel/babel/runs/925434089#step:4:751

kaicataldo: Gotcha! So will we bump this when we do the `7.11.0` release?

JLHwung: I think we still support `7.10`? Maybe we should remove runtime `peerDependencies` checks, they should be addressed when installing packages.

kaicataldo:
Oops, reviewed this after it was merged

PR #11896, opened 2020-07-30
update: hardcode @babel/eslint-parser min supported version check

kaicataldo:
| Fixed Issues?            |
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

The original intent of this was to ensure that we're using a version of `@babel/core` that has the `parse()` API. The current implementation leads to peer dependency changes causing breaking changes.

Ideally, we'd like users to use the same version of `@babel/core` and `@babel/eslint-parser` to ensure compatibility, but we can figure out if we want to do anything about that (logging a message, etc.) later.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11896"></a>

### DIFF
--- a/eslint/babel-eslint-parser/src/index.js
+++ b/eslint/babel-eslint-parser/src/index.js
@@ -15,7 +15,8 @@ import visitorKeys from "./visitor-keys";
 let isRunningMinSupportedCoreVersion = null;
 
 function baseParse(code, options) {
-  const minSupportedCoreVersion = ">=7.0.0";
+  // Ensure we're using a version of `@babel/core` that includes `parse()` and `tokTypes`.
+  const minSupportedCoreVersion = ">=7.2.0";
 
   if (typeof isRunningMinSupportedCoreVersion !== "boolean") {
     isRunningMinSupportedCoreVersion = semver.satisfies(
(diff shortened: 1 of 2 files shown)


eslint/babel-eslint-parser/src/index.js
```diff
@@ -12,19 +12,22 @@ import convert from "./convert";
 import analyzeScope from "./analyze-scope";
 import visitorKeys from "./visitor-keys";
 
-let isRunningSupportedVersion;
+let isRunningMinSupportedCoreVersion = null;
 
 function baseParse(code, options) {
-  if (typeof isRunningSupportedVersion !== "boolean") {
-    isRunningSupportedVersion = semver.satisfies(
+  const minSupportedCoreVersion = ">=7.0.0";
```
hzoo, line 18: Ok so parse was added first in 7.0?

kaicataldo: `@babel/core#parse()` was added in `7.0.0-beta.51`, but we need to bump this to `7.2.0` so that we have access to `@babel/parser#tokTypes`.

existentialism:
:shipit:

kaicataldo:
Thanks for reviewing this so fast!

PR #11899, opened 2020-07-31
Update `@babel/*` deps

JLHwung:
| Any Dependency Changes?  | `@babel/*` is updated

In this PR we also removes extra install step for `@babel/eslint-*`.

We have run into issues during publish: Let's say we want to publish 7.12. After `lerna version`, running `yarn install` will request unpublished 7.12 `@babel/*` deps from eslint packages, although we can choose the published version 7.11 resolved as 7.12, it will incur changes on `yarn.lock` and thus the publish will fail because uncommitted yarn.lock file. See https://app.circleci.com/pipelines/github/babel/babel/3558/workflows/05ae840c-67c4-4312-9776-51f33f445325/jobs/26769 for the reproducing publish error.

I have replaced two eslint packages from `link:` to published version. Lerna should handle them properly. Two private eslint packages are reserved as link. It happens to work because they don't have dependencies.

Also simplified babel config by removing included plugins.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11899"></a>


babel.config.js
```diff
@@ -105,8 +105,6 @@ module.exports = function (api) {
       // TODO: Use @babel/preset-flow when
       // https://github.com/babel/babel/issues/7233 is fixed
       "@babel/plugin-transform-flow-strip-types",
-      ["@babel/proposal-class-properties", { loose: true }],
```
hzoo, line 108: nice!

Makefile
```diff
@@ -252,8 +252,6 @@ endif
 bootstrap-only: lerna-bootstrap
 
 yarn-install: clean-all
-	# Install dependencies in individual packages so that we can link them at the top level.
-	for package in eslint/*/; do yarn --ignore-engines --cwd $$package; done
```
kaicataldo, line 256: Don't we still need this until Babel 8?

JLHwung: Running yarn install in sub folders incurs changes to the yarn lock files after lerna version, lerna will abort publishing with uncommitted changes. See also the PR descriptions.

Issue #7347, opened 2018-02-07
[transform-modules-umd] When there is a default export, set browser global to the export itself


shreyasminocha:
I am exporting a function as the default function. I require the `babel-plugin-transform-modules-umd` to set the `global.actual` variable to `mod.exports.default` rather than `mod.exports` as it currently does.

### Input Code
```js
export function foo() {
    return new Bar();
}
```

### Expected Behavior

```js
(function (global, factory) {
    if (typeof define === "function" && define.amd) {
        define(["exports"], factory);
    } else if (typeof exports !== "undefined") {
        factory(exports);
    } else {
        var mod = {
            exports: {}
        };
        factory(mod.exports);
        global.Bar = mod.exports.default;
    }
})(this, function (exports) {
    "use strict";
 
    Object.defineProperty(exports, "__esModule", {
          value: true
    });
 
    exports.default = foo;
});
```

### Current Behavior

At the moment, `global.actual` will be set to `mod.exports`.

### Context

I'm trying to rewrite a popular javascript library in es6. I want the exported function to be accessible through `global.actual`.

kottenator:
@shreyasminocha - I've found a workaround: `babel-plugin-add-module-exports`:

```js
// .babelrc
{
  "presets": ["env"],
  "plugins": [
    "add-module-exports",
    "transform-es2015-modules-umd"
  ]
}
```

Issue #2212, opened 2015-08-14
Kill CommonJS default export behaviour


sebmck:

cesarandreu:
What's the planned behavior going forward?

sebmck:
@cesarandreu To always export a default to `exports.default`.

jmm:
@sebmck Is this in reference to the issue discussed in #2047? I've been finding myself agreeing with 
@UltCombo that that behavior can be a headache.

UltCombo:
:+1: 
Adding `.default` to your CJS `require()`s can be slightly bothersome, but not nearly as much as having an unexpected breaking change due to simply adding a named export to the imported module.

sebmck:
@jmm Not really. It's the fact that people are (wrongly) doing:

**a.js**

```javascript
import { foo, bar } from "./b";
```

**b.js**

```javascript
export default {
  foo: "foo",
  bar: "bar"
};
```

And Babel is letting them.

UltCombo:
@sebmck That's just another symptom of bad implicit magic, imho. If you add a named export to `b.js` then it will be a breaking change to `a.js` (although it should be broken since the beginning).

jmm:
@sebmck Oh ok, I wasn't aware of that behavior. :+1: to killing that off (obviously).

jquense:
Hi! first, thanks for all the hard work and excellent tool. is the idea that commonJS consumers would need to `require().default` for all modules compiled with Babel? Reduced magic is nice, but that seems like it would be a deal breaker for folks publishing to the node ecosystem where such a convention doesn't exist. Perhaps it's worth it but if it's helpful feedback, I think if that was the behavior I'd stop using the module syntax rather then ask consumers to use `.default`.

UltCombo:
@jquense or you can tell the consumers to prefer the ES2015 module syntax over the CJS one. :wink:

jquense:
@UltCombo, I really can't tell anyone building a node app to also use Babel to transpile their app to avoid some  on idiomatic import with my module.  they just won't use my module instead

UltCombo:
@jquense I don't really see a problem as long as the API usage is properly documented for CJS and ES Modules. You may as well use a named export to give a more meaningful property name than `default`.

jquense:
[...]

Issue #2047, opened 2015-07-22
CommonJS module formatter: default export is assigned differently depending on whether there are any named exports


UltCombo:
This sample code:

```js
export default 1;
export const x = 2;
```

Generates:

```js
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports["default"] = 1;
var x = 2;
exports.x = x;
```

Which is fine.
Now if we remove the named export:

```js
export default 1;
```

Generates:

```js
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports["default"] = 1;
module.exports = exports["default"];
```

The `module.exports` is overwritten by the default export.
This still works fine if importing the module via `import` statements (due to the `interopRequireDefault` helper), but it is rather confusing and error-prone if importing the CJS module via `require()` calls.

That is, sometimes the default export is returned directly by `require('module')`, other times you will need to access the `require('module').default` property, all depending on an arbitrary factor that may change anytime (adding/removing named exports), which causes a breaking change for all consumers.

loganfsmyth:
This is expected behavior: http://babeljs.io/docs/usage/modules/#interop

As documented there:

> If you don't want this behaviour then you can use the commonStrict module formatter.

UltCombo:
Thanks for the quick reply.
I thought the strict module formatters were going to be removed (https://github.com/babel/babel/issues/587#issuecomment-71347804), I assume this has changed?

I'll be using the `commonStrict` formatter again now, thanks. :smile:

UltCombo:
So, looks like the `commonStrict` module formatter does not attempt any interop at all -- I had forgot this bit.

In my ideal use case, I'd like to keep the ES/CJS modules interop, but have the CJS module formatter expose the default export in a consistent way.

Well, I guess for this rare use case (where I need dynamic `require()`s, can't use ES Modules syntax), I can just as well import and apply the `interop-require` helper from the babel-runtime package manually.

danielo515:
I am also missing this feature. The reasoning behind removing it still unclear to me. Removing this feature almost breaks all the modules that trusted babel for transpile to regular CJS modules.

loganfsmyth:
You can get this behavior in Babel 6 via https://www.npmjs.com/package/babel-plugin-add-module-exports

danielo515:
[...]

PR #11903, opened 2020-08-01
plugin-transform-modules-umd: add option "legacyDefaultOnlyExport"

KingSora:
| Patch: Bug Fix?          |No
| Minor: New Feature?      |Yes
| Documentation PR Link    | 

Good day!

This Contribution is made on a Windows System, I had many problems to create a setup where I could run all the `make` scripts, so please point out mistakes. Its also my first PR to babel.

I made this codechanges to make it possible to have a "legacy default only export". Whats that? If the module export (or re-exports) only one thing and this export is a default export, the `_exports` object is set to the exported object directly.

This is related to issues like #7347, it shall also replace the `babel-plugin-add-module-exports` plugin, which was [dropped in babel 7](https://babeljs.io/docs/en/v7-migration-api#export-changes).

How does it work? You can pass a option called `legacyDefaultOnlyExport` to the `plugin-transform-modules-umd` plugin. This option will then check if the module has only one `default` export (or re-export) and only then it will change the output. If you have more exports the option will do nothing and the output won't be changed.

The `ESModuleHeader`:
```js
Object.defineProperty(_exports, "__esModule", {
  value: true
});
```
also won't be added, as well as the `_exports.default = void 0;` line.
Please take a look at the tests for more information.

General example:
```js
// input:
export default 42;
```
```js
// output:
(function (global, factory) {
  if (typeof define === "function" && define.amd) {
    define(["exports"], factory);
  } else if (typeof exports !== "undefined") {
    factory(exports);
  } else {
    var mod = {
      exports: {}
    };
    factory(mod.exports);
    global.input = mod.exports;
  }
})(typeof globalThis !== "undefined" ? globalThis : typeof self !== "undefined" ? self : this, function (_exports) {
  "use strict";

  var _default = 42;
  _exports = _default;
});
```

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11903"></a>


packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js
```diff
@@ -57,6 +57,38 @@ export function hasExports(metadata: ModuleMetadata) {
...
 
+/**
+ * Check if the module has a default export (or reexport) and no other named exports (or reexports).
+ */
+export function hasDefaultExportOnly(metadata: ModuleMetadata) {
+  const { local, source } = metadata;
+
+  if (local.size === 0) {
+    const sourceReexports = new Map();
+    for (const [, srcMetadata] of source) {
```
oliverdunk, line 68: Could all of this be replaced with something like:

```
if (source.length > 1) return false;

const [, srcMetadata] = source[0];
if (srcMetadata.reexports.length > 1) return false;

const [exportName, importName] = reexports[0];
return exportName === "default";
```

There's maybe a balance to making it more readable, but it seems a shame to build a Map when we only care about the first item.

packages/babel-helper-module-transforms/src/index.js
```diff
@@ -64,7 +66,11 @@ export function rewriteModuleStatementsAndPrepareHeader(
   }
 
   const headers = [];
-  if (hasExports(meta) && !strict) {
+  const defaultExportOnly = hasDefaultExportOnly(meta);
+  const doLegacyDefaultOnlyExport =
+    legacyDefaultOnlyExport && defaultExportOnly;
```
oliverdunk, line 71: Is it worth changing this to `legacyDefaultOnlyExport && hasDefaultExportOnly(meta)`, so we only run the check if we need to?

KingSora:
@oliverdunk Thanks for your review, I'll look into your suggestions as soon as I'm at home.
I'm not sure about the formatting changes, because I definitely didn't touch any of the files myself, this was done with `make fix`. I could revert them, but is this desired, because if someone would run `make fix`, those changes would pop up again (maybe this is also happening only on my end because I use Windows?)

packages/babel-helper-module-transforms/src/normalize-and-load-metadata.js
KingSora: Well this code is to analyze whether you have a `default reexport`. So I have to check all the imports, because you can have 10 imports and not 10 reexports. So I definitely need the loop. Then we need to check whether the import has a `reexport` and if the `reexport` is not a `default` reexport, we can return false, and if yes we can set `hasDefaultReexport` to `true`, but we need to continue the loop because its possible to have further reexports.

Code would be like this:
```js
let hasDefaultReexport = false;
for (const [, srcMetadata] of source) {
  const { reexports } = srcMetadata;
  if (reexports) {
    for (const [exportName, importName] of reexports) {
      if (exportName === 'default') {
        hasDefaultReexport = true;
      }
      else {
        return false;
      }
    }
  }
}

return hasDefaultReexport;
```

JLHwung: The revised code is more readable. Can also remove unused `importName`.

KingSora: Pushed the changes!

packages/babel-helper-module-transforms/src/index.js
KingSora: Its updated, thanks! :)

JLHwung:
I would like to push back because this option introduces different behaviours when new `export` is added. 

Consider a library `forty-two@1.0` compiled with `@babel/transform-modules-umd, { legacyDefaultOnlyExport: true }`. 
```js
export default 42;
```

becomes
```js
module.exports = 42;
```
So downstream users can use `require("forty-two");` directly.

`forty-two@1.1` introduces a new feature.
```js
export default 42;
export const name = "forty-two";
```
But now `require("forty-two");` is no more `42`, it returns `{ default: 42, name: "forty-two" }`, thus it is a breaking change for library users. But if we look back to our source: adding a named exports `export const name` is definitely not a breaking change.

I believe consistency weighs more than cost of typing `.default`. And if you use `import`, bundlers like Rollup and WebPack will transform to `interpolate(require("forty-two"))` where `interpolate` will unwrap `.default` for you. So you don't have to type `.default` at all.

@KingSora Thanks for your efforts on this PR!

KingSora:
@JLHwung I understand your argument, and you're right that this option should be used with caution aka. "if you know what you're doing", but currently I have literally no option to bundle my library with a bundler like `rollup` without a option like this.

Why is that? - because the browser global will be always the exported object, so even if I have only a default export, and I want may lbrary to be directly a global variable like `window.myLibrary` I can't do that because babel will force it to `window.myLibrary.default`. Currently my only option with a bundler like `rollup` is to use their `umd` wrapper which is providing such a option, but even then the problem isn't solved, because the `rollup-babel-plugin` is warning me, that if I run babel after rollup wrapped my code inside their umd wrapper, the semantics of their umd wrapper could be changed by the changes babel does. So currently its really hard to bundle a library with a default only export with babel + rollup or babel + webpack or anything like that, because the public interface in the browser gets screwed up every time. 

Before you reject this PR please really take this argument into consideration, because if you google this problem, you'll find not just a few articles which have the same problem.

JLHwung:
> babel could throw and error (or a warning message) if the library has named exports and legacyDefaultOnlyExport: true is used. Then users wouldn't run into possibly unwanted behavior without a notification at least.

This is not a solution: The published `module.exports` _is_ published and resides in people's node module. You can change `legacyDefaultOnlyExport` to be `true` to make Babel happy, but the unwanted library interface changes are inevitable for users. And what we are meant to is adding a named exports should _not_ break interface after umd transform.

> I have only a default export, and I want my library to be directly a global variable like window.myLibrary I can't do that because babel will force it to window.myLibrary.default.

If you intend to ship `window.myLibrary`, then it is conceptually a namespace export instead of default export. And a namespace export is more preferable than default exports because bundlers can do tree-shaking and remove unused exports. Let's take a look at the example of library `forty-two`

```js
const FortyTwo = {
  value: 42,
  en: "forty two",
  fr: "quarante-deux",
  zh: "四十二",
}
export default FortyTwo
// transformed as
window.FortyTwo.default = FortyTwo;
```

You can replace default exports by several named exports
```js
export const value = 42;
export const en = "forty two";
export const fr = "quarante-deux";
export const zh = "四十二";
// transformed as
window.FortyTwo.value = 42;
window.FortyTwo.en = "forty two";
// ...
```

In this case the `FortyTwo` becomes a namespace that contains all exports or your library. Bundlers can shake unused languages if users are not referring it, compared to a non-shakeable default export. Users can refer `window.FortyTwo` without the fear of it breaking because new named exports are added.

In the past Babel 6 has transformed `export default foo` as `module.exports = foo`. It was dangerous and it allowed people to
```js
import { foo, bar } from "qux";
export default {
  foo: "foo",
  bar: "bar"
}
```
So it is dropped intentionally in https://github.com/babel/babel/issues/2212 and also discussed in https://github.com/babel/babel/issues/2047. I think our attitude towards this behaviour is unchanged as Sebastian stated before:

> I am not willing to support this in Babel, even as an option, as it's proved to be very dangerous for refactoring and semantics.

nicolo-ribaudo:
Please note that "dropped babel-plugin-add-module-exports in Babel 7" means that we don't use it when transpiling Babel's own source code. It's not maintained by us, but it looks like it works with UMD and Babel 7: https://github.com/59naga/babel-plugin-add-module-exports

PR #11907, opened 2020-08-03
Return a list of files that were read from loadPartialConfig

devongovett:
This adds a `files` list to the return value of `babelCore.loadPartialConfig` which lists all files that were read in the process of loading the config. Some of these are currently returned in the `babelrc`, `babelignore`, and `configFile` fields, but other files such as those that were extended from another config are not returned, making it difficult for other tools that consume babel (e.g. Parcel, webpack) to properly watch these extended configs and invalidate caches. The `files` key should solve this as it includes all files that influenced the config, not just the top-level ones.

In addition, I've added a `showIgnoredFiles` option to `loadPartialConfig` which always returns a `PartialConfig` result even when the file is ignored, rather than returning `null`. This allows tools to properly invalidate caches when whatever config file that marked the compiled file as ignored changes. An `isIgnored` boolean is returned as part of the `PartialConfig` result to indicate that the file is ignored. Given that changing the `null` result to return a value would be a breaking change, this is behind an option. Open to better name suggestions though! 😉 

This is really important for us in Parcel so that we can get proper caching working with Babel. At the moment there's really no good way to do that without support from Babel itself, or reinventing all of its config loading. This should also help other tools like babel-loader - found a [todo comment](https://github.com/babel/babel-loader/blob/13a82386473986517a6e9dcfb1a7b75b47379d64/src/index.js#L209-L213) for it. 😄 

Totally open to alternative suggestions. Really made this to get the conversation started.

| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 
| Any Dependency Changes?  |

### DIFF
--- a/packages/babel-core/src/config/config-chain.js
+++ b/packages/babel-core/src/config/config-chain.js
@@ -40,6 +40,7 @@ export type ConfigChain = {
   plugins: Array<UnloadedDescriptor>,
   presets: Array<UnloadedDescriptor>,
   options: Array<ValidatedOptions>,
+  files: Array<string>,
 };
 
 export type PresetInstance = {
@@ -71,6 +72,7 @@ export function* buildPresetChain(
     plugins: dedupDescriptors(chain.plugins),
     presets: dedupDescriptors(chain.presets),
     options: chain.options.map(o => normalizeOptions(o)),
+    files: [],
   };
 }
 
@@ -128,6 +130,8 @@ export type RootConfigChain = ConfigChain & {
   babelrc: ConfigFile | void,
   config: ConfigFile | void,
   ignore: IgnoreFile | void,
+  isIgnored: boolean,
+  files: Array<string>,
 };
 
 /**
@@ -202,6 +206,7 @@ export function* buildRootChain(
       : null;
 
   let ignoreFile, babelrcFile;
+  let isIgnored = false;
   const fileChain = emptyChain();
   // resolve all .babelrc files
   if (
@@ -215,14 +220,18 @@ export function* buildRootChain(
       context.caller,
     ));
 
+    if (ignoreFile) {
+      fileChain.files.push(ignoreFile.filepath);
+    }
+
     if (
       ignoreFile &&
       shouldIgnore(context, ignoreFile.ignore, null, ignoreFile.dirname)
     ) {
-      return null;
+      isIgnored = true;
     }
 
-    if (babelrcFile) {
+    if (babelrcFile && !isIgnored) {
       const validatedFile = validateBabelrcFile(babelrcFile);
       const babelrcLogger = new ConfigPrinter();
       const result = yield* loadFileChain(
@@ -231,10 +240,16 @@ export function* buildRootChain(
         undefined,
         babelrcLogger,
       );
-      if (!result) return null;
-      babelRcReport = babelrcLogger.output();
+      if (!result) {
+        isIgnored = true;
+      } else {
+        babelRcReport = babelrcLogger.output();
+        mergeChain(fileChain, result);
+      }
+    }
 
-      mergeChain(fileChain, result);
+    if (babelrcFile && isIgnored) {
+      fileChain.files.push(babelrcFile.filepath);
     }
   }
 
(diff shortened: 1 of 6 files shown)


packages/babel-core/src/config/partial.js
```diff
@@ -181,17 +195,23 @@ class PartialConfig {
   babelrc: string | void;
   babelignore: string | void;
   config: string | void;
+  isIgnored: boolean;
+  files: Array<string>;
```
JLHwung, line 199: Should it be a `Set<string>`? There could be multiple `extends` in different config items:
```js
[{
  test: "foo.js",
  extends: "root-config"
}, {
  test: "bar.js",
  extends: "root-config"
}]
```

packages/babel-core/src/config/config-chain.js
```diff
@@ -257,12 +272,14 @@ export function* buildRootChain(
   );
 
   return {
-    plugins: dedupDescriptors(chain.plugins),
-    presets: dedupDescriptors(chain.presets),
-    options: chain.options.map(o => normalizeOptions(o)),
+    plugins: isIgnored ? [] : dedupDescriptors(chain.plugins),
+    presets: isIgnored ? [] : dedupDescriptors(chain.presets),
+    options: isIgnored ? [] : chain.options.map(o => normalizeOptions(o)),
+    isIgnored,
```
nicolo-ribaudo, line 278: I would love if this was a string enum instead of a boolean.

In https://github.com/babel/babel/pull/11409 I was working on moving the extensions handling from `@babel/cli`/`@babel/register`/etc to `@babel/core` to support specifying file extensions in Babel config files.
However, I had a problem when making it work with the `--copy-files` and `--no-copy-ignored` options: my PR doesn't give a way of knowing _why_ a file is not transpiled (if it is ignored or if it has an unsupported extension).

If we made this option an enum, it would be then easy to extend it adding a third state.

devongovett: Something like `fileAction: 'transpile' | 'ignored' | 'unsupported'`? Not really sure what to name it haha. 😄

nicolo-ribaudo: I avoided suggesting any name on purpose, since I couldn't find one 😂

Maybe `fileHandling`, since "ignoring" isn't really an action but it's _not_ doing the action?

(Or maybe someone else will have a better alternative 🤞)

existentialism: I'm okay with `fileHandling`.

devongovett: Updated. Added `unsupported` to the type definition as well but it's not implemented yet. Can remove for now if you prefer.

devongovett:
Anything I can do to move this forward? 😄

JLHwung (approved):
Look good to me. If we get enough approvals, will need to update https://babeljs.io/docs/en/babel-core#loadpartialconfig

devongovett:
I'm happy to send a PR for that!

kaicataldo (approved):

devongovett:
https://github.com/babel/website/pull/2333

nicolo-ribaudo:
@devongovett @JLHwung @existentialism  @kaicataldo When I asked `fileHandling` to be a string so that it can allow three values (transpile, ignored and unsupported), I was thinking that I could use it for [`9114ffc` (#12151)](https://github.com/babel/babel/pull/12151/commits/9114ffcc94cb125a7b8391a2055c2f566717a024): my idea was to load the config for a file, check if it's "ignored" or "unsupported", and act accordingly.

However, while implementing it I realized that it's not possible because `loadPartialConfig` does not resolve presets, and thus it doesn't have info about the supported `extensions`.

I found another way of detecting ignored vs unsupported: if `loadPartialConfig` returns `null` it's ignored; otherwise if `transformFile` returns `null` it's unsupported.

For this reason, I propose to change back the design of this PR to `isIgnored: boolean` rather than `fileHandling: string`. I can work on it this weekend if you agree.

JLHwung:
@nicolo-ribaudo I don't have strong preference here. But I believe

> I found another way of detecting ignored vs unsupported: if loadPartialConfig returns null it's ignored; otherwise if transformFile returns null it's unsupported.

should be documented somewhere. If `fileHandling: StatusEnum` is more friendly to API users, we can go with current design.

nicolo-ribaudo:
Sure, I'll document it alongside with the `extensions` option docs!

PR #11911, opened 2020-08-04
Use fastest-levenshtein instead of levenary

ka-weihe:
`levenary` is incredibly slow. `fastest-levenshtein` should be used instead since it is much faster (up to 10x faster) and it has no dependencies - unlike `levenary`. 

### Benchmarks
                      50 words, length max=15 min=5 avr=8.4
         181,539 op/s » fastest-levenshtein
          55,712 op/s » levenary
                      20 sentences, length max=106 min=30 avr=64.4
          26,641 op/s » fastest-levenshtein
           3,356 op/s » levenary

| Fixed Issues?            | 
| Patch: Bug Fix?          |
| Major: Breaking Change?  |
| Documentation PR Link    | 

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11911"></a>

### DIFF
--- a/packages/babel-helper-compilation-targets/package.json
+++ b/packages/babel-helper-compilation-targets/package.json
@@ -23,8 +23,8 @@
   "dependencies": {
     "@babel/compat-data": "^7.10.4",
     "browserslist": "^4.12.0",
+    "fastest-levenshtein": "^1.0.10",
     "invariant": "^2.2.4",
-    "levenary": "^1.1.1",
     "semver": "^5.5.0"
   },
   "peerDependencies": {
--- a/packages/babel-helper-compilation-targets/src/index.js
+++ b/packages/babel-helper-compilation-targets/src/index.js
@@ -1,7 +1,7 @@
 // @flow
 
 import browserslist from "browserslist";
-import findSuggestion from "levenary";
+import { closest as findSuggestion } from "fastest-levenshtein";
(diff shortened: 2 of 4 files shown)


JLHwung:
`levenary` is 50% smaller than `fastest-levenstein`.

https://bundlephobia.com/result?p=fastest-levenshtein@1.0.10
https://bundlephobia.com/result?p=levenary@1.1.1

Given that `levenary` is only used when Babel users have invalid options, so the code path is an one-time thing. In other words, `levenary` will not be executed at all if users provide valid options. Therefore, I don't think it is worthy here to optimize `levenary`.

ka-weihe:
It's 269 bytes extra to improve a piece of code by a lot and get one less dependency. If this was literally any other piece of code of Babel (no matter how insignificant) and an improvement gave a 10x speedup, would you really not approve it because of a mere 269 bytes?

JLHwung:
An 1000x speedup can not benefit most users if it is not executed at all:
```js
if (false) { very_fast_sort(very_huge_array) }
```
However babel's dependencies are downloaded unconditionally. And as is shown in your benchmark, the `levenary` is not so low that people would complain that Babel's error message does not show up. In this case, the bundled size outweighs performance improvement.

I am not meant to discourage you for performance improvements. I will be really appreciated if you can optimize the critical paths of Babel --- e.g. traversing AST.

PR #8115, opened 2018-06-04
fix flow-double-nullable-type-cant-be-parsed

HassanAli321:
| Q                        | A 
| ------------------------ | ---
| Fixed Issues?            | Fixes #7924 
| Patch: Bug Fix?          | yes
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Tests Added + Pass?      | 
| Documentation PR         | 
| Any Dependency Changes?  |
| License                  | MIT

PR #11912, opened 2020-08-04
rescan gt/lt token after TsAsExpression is parsed

### LINKED ISSUE #11446: Binary operator after TypeScript `as` assertion causes wrong SyntaxError
## Bug Report

**Current Behavior**

If a bianry operator like `>=` appears directly after TypeScript `as` assertion, a wrong SyntaxError occurs.

**Input Code**
```ts
function foo(x: any) {
	return x as number >= 0;
}
```

The above code causes the following SyntaxError:
```
{ SyntaxError: test.ts: Unexpected token (2:22)
  1 | function foo(x: any) {
> 2 |   return x as number >= 0;
    |                       ^
  3 | }
  4 | 
```

**Expected behavior/code**

No SyntaxError should occur.

- Filename: `.babelrc`

```js
{
}
```

**Environment**
```
  System:
    OS: Linux 4.4 Ubuntu 14.04.5 LTS, Trusty Tahr
  Binaries:
    Node: 10.15.1 - ~/local/lib/node-v10.15.1-linux-x64/bin/node
    Yarn: 1.22.4 - ~/local/lib/node-v10.15.1-linux-x64/bin/yarn
    npm: 6.9.0 - ~/local/lib/node-v10.15.1-linux-x64/bin/npm
  npmPackages:
    @babel/cli: ^7.8.4 => 7.8.4 
    @babel/core: ^7.9.0 => 7.9.0 
    @babel/preset-typescript: ^7.9.0 => 7.9.0 
```

JLHwung:
| Fixed Issues?            | Fixes #11446 
| Documentation PR Link    | 
| Any Dependency Changes?  |

When parsing mixture of BinaryExpression and TsAsExpression, e.g.
```ts
a as boolean <= b
```
The `<=` is read when we are parsing TsType `boolean`, thus it is read as a single `<` (tt.relational) token instead of a compound of `<=`. Here we introduce a `reScan_lt_gt` method and re-scan `<`/`>` so `<=` can be correctly tokenized. Note that the TypeScript compiler adopted the same approach here:

https://github.com/microsoft/TypeScript/blob/624736418132374a6213bd4c66310017ee17d38a/src/compiler/parser.ts#L4329

This PR also fixes the following issue
```ts
a as boolean ?? b
```
because `??` is read as a pair of `?` questional instead of nullishCoalescing. Since it is used in Flow exclusively (#8115), I moved the code path to the flow plugin.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11912"></a>

### DIFF
--- a/packages/babel-parser/src/plugins/flow.js
+++ b/packages/babel-parser/src/plugins/flow.js
@@ -2135,6 +2135,9 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         (code === charCodes.greaterThan || code === charCodes.lessThan)
       ) {
         return this.finishOp(tt.relational, 1);
+      } else if (this.state.inType && code === charCodes.questionMark) {
+        // allow double nullable types in Flow: ??string
+        return this.finishOp(tt.question, 1);
       } else if (isIteratorStart(code, next)) {
         this.state.isIterator = true;
         return super.readWord();
--- a/packages/babel-parser/src/plugins/typescript/index.js
+++ b/packages/babel-parser/src/plugins/typescript/index.js
@@ -1884,6 +1884,8 @@ export default (superClass: Class<Parser>): Class<Parser> =>
           node.typeAnnotation = this.tsNextThenParseType();
         }
         this.finishNode(node, "TSAsExpression");
+        // rescan `<`, `>` because they were scanned when this.state.inType was true
+        this.reScan_lt_gt();
         return this.parseExprOp(
           node,
           leftStartPos,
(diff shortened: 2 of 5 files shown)


packages/babel-parser/src/plugins/flow.js
```diff
@@ -2135,6 +2135,9 @@ export default (superClass: Class<Parser>): Class<Parser> =>
         (code === charCodes.greaterThan || code === charCodes.lessThan)
       ) {
         return this.finishOp(tt.relational, 1);
+      } else if (this.state.inType && code === charCodes.questionMark) {
```
existentialism, line 2138: Do we have tests for `??string`?

JLHwung: We do, it was previously fixed in #8115. I forgot to mention that.

existentialism: Cool, just wanted to double check :)

PR #11919, opened 2020-08-05
refactor: simplify smart pipeline parsing

JLHwung:

I found this issue when I try to add more test coverages to `@babel/parser`. As you can see in https://codecov.io/gh/babel/babel/src/bc7a811fce3ceeea393229299c1cdb63858608e6/packages/babel-parser/src/parser/expression.js#L2526

The `PipelineBareConstructor` case branch is not covered at all. I realized that these branches are not reachable so we can simplify the parser paths here.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11919"></a>

### DIFF
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -2479,17 +2479,10 @@ export default class ExpressionParser extends LValParser {
     startPos: number,
     startLoc: Position,
   ): N.PipelineBody {
-    const pipelineStyle = this.checkSmartPipelineBodyStyle(childExpression);
-
-    this.checkSmartPipelineBodyEarlyErrors(
-      childExpression,
-      pipelineStyle,
-      startPos,
-    );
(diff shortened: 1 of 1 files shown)


packages/babel-parser/src/parser/expression.js
```diff
@@ -2479,77 +2479,47 @@ export default class ExpressionParser extends LValParser {
...
-    pipelineStyle: N.PipelineStyle,
     startPos: number,
   ): void {
     if (this.match(tt.arrow)) {
       // If the following token is invalidly `=>`, then throw a human-friendly error
       // instead of something like 'Unexpected token, expected ";"'.
       throw this.raise(this.state.start, Errors.PipelineBodyNoArrow);
-    } else if (
-      pipelineStyle === "PipelineTopicExpression" &&
-      childExpression.type === "SequenceExpression"
```
JLHwung, line 2509: `childExpression.type === "SequenceExpression"` implies the fact that `pipelineStyle === "PipelineTopicExpression"`.

PR #11922, opened 2020-08-05
simplify isLookaheadRelational method

JLHwung:

Simplify `isLookaheadRelational` to get rid of an uncovered line: https://codecov.io/gh/babel/babel/src/bc7a811fce3ceeea393229299c1cdb63858608e6/packages/babel-parser/src/parser/util.js#L42

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11922"></a>

### DIFF
--- a/packages/babel-parser/src/parser/util.js
+++ b/packages/babel-parser/src/parser/util.js
@@ -6,7 +6,6 @@ import State from "../tokenizer/state";
 import type { Node } from "../types";
 import { lineBreak } from "../util/whitespace";
 import { isIdentifierChar } from "../util/identifier";
-import * as charCodes from "charcodes";
 import { Errors } from "./error";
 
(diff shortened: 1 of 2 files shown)


packages/babel-parser/src/parser/util.js
```diff
@@ -35,18 +34,6 @@ export default class UtilParser extends Tokenizer {
     return this.match(tt.relational) && this.state.value === op;
   }
 
-  isLookaheadRelational(op: "<" | ">"): boolean {
-    const next = this.nextTokenStart();
-    if (this.input.charAt(next) === op) {
-      if (next + 1 === this.input.length) {
```
JLHwung, line 41: This check is removed in `isLookaheadToken_lt` because if `next + 1` is the `length`, `afterNext` is `NaN`, thus it always satisfies
```
afterNext !== charCodes.lessThan && afterNext !== charCodes.equalsTo
```

PR #11923, opened 2020-08-05
Add more parser test cases

JLHwung:

This PR aims for improving `@babel/parser`'s test coverage. Ideally the only uncovered lines should be `throw new Error("unreachable")`.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11923"></a>

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11923"></a>


packages/babel-parser/src/parser/statement.js
```diff
@@ -1784,19 +1765,9 @@ export default class StatementParser extends ExpressionParser {
 
   maybeParseExportDeclaration(node: N.Node): boolean {
     if (this.shouldParseExportDeclaration()) {
-      if (this.isContextual("async")) {
```
JLHwung, line 1787: These checks are redundant because `shouldParseExportDeclaration` will check `isAsyncFunction`
https://github.com/babel/babel/blob/a827ca41f3c07a1ffddeaa00998c772a61a60b68/packages/babel-parser/src/parser/statement.js#L1918-L1938
When `this.isContextual("async")` is true, `isAsyncFunction()` must be true and `async` must be followed by a `function` token, therefore `this.unexpected(next, tt._function);` is never reached.

packages/babel-parser/src/parser/statement.js
```diff
@@ -1999,15 +1970,10 @@ export default class StatementParser extends ExpressionParser {
     const currentContextDecorators = this.state.decoratorStack[
       this.state.decoratorStack.length - 1
     ];
+    // If node.declration is a class, it will take all decorators in the current context.
```
existentialism, line 1973: Nit:

```suggestion
    // If node.declaration is a class, it will take all decorators in the current context.
```

packages/babel-parser/test/expressions/experimental/top-level-await/script/input.js
```diff
@@ -0,0 +1,2 @@
+// prettier-ignore
+await +42
```
nicolo-ribaudo, line 2: Can you add the same test for "unambiguous" source type?

JLHwung: Addressed in https://github.com/babel/babel/pull/11923/commits/cb65cb3fc4edf48ac13fc83722222ec5d6d388dd.

PR #11930, opened 2020-08-07
Refactor [In] production parameter tracking

JLHwung:

This PR is a follow-up to #10956. In this PR we remove the `noIn` passed down to `parseMaybeAssign` and replace it with `this.prodParam.hasIn`, which is only set in `parseMaybeAssignAllowIn`, `parseMaybeAssignDisallowIn` and `parseExpression`. This aligns to the fact that in the spec `[+In]` and `[~In]` exist only in `AssignmentExpression` and `Expression`.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11930"></a>

Blocks #11931.

### DIFF
--- a/packages/babel-parser/src/parser/expression.js
+++ b/packages/babel-parser/src/parser/expression.js
@@ -44,6 +44,7 @@ import {
 import { ExpressionErrors } from "./util";
 import {
   PARAM_AWAIT,
+  PARAM_IN,
   PARAM_RETURN,
   PARAM,
   functionFlags,
@@ -153,11 +154,9 @@ export default class ExpressionParser extends LValParser {
   // the AST node that the inner parser gave them in another node.
 
   // Parse a full expression.
-  // - `noIn`
+  // - `disallowIn`
   //   is used to forbid the `in` operator (in for loops initialization expressions)
-  //   When `noIn` is true, the production parameter [In] is not present.
-  //   Whenever [?In] appears in the right-hand sides of a production, we pass
-  //   `noIn` to the subroutine calls.
+  //   When `disallowIn` is true, the production parameter [In] is not present.
 
   // - `refExpressionErrors `
   //   provides reference for storing '=' operator inside shorthand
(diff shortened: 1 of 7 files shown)


packages/babel-parser/src/parser/expression.js
```diff
@@ -2049,7 +2057,12 @@ export default class ExpressionParser extends LValParser {
     trailingCommaPos: ?number,
   ): N.ArrowFunctionExpression {
     this.scope.enter(SCOPE_FUNCTION | SCOPE_ARROW);
-    this.prodParam.enter(functionFlags(isAsync, false));
+    let flags = functionFlags(isAsync, false);
```
JLHwung, line 2060: We need to inherit `[In]` when the arrow function body is `{`-less (concise body).

packages/babel-parser/src/parser/expression.js
```diff
@@ -153,44 +154,80 @@ export default class ExpressionParser extends LValParser {
...
-        node.expressions.push(this.parseMaybeAssign(noIn, refExpressionErrors));
+        node.expressions.push(this.parseMaybeAssign(refExpressionErrors));
       }
       this.toReferencedList(node.expressions);
       return this.finishNode(node, "SequenceExpression");
     }
     return expr;
   }
 
+  // Setting [~In] parameter for assignment expression
```
JLHwung, line 196:
```suggestion
  // Set [~In] parameter for assignment expression
```

packages/babel-parser/src/parser/expression.js
```diff
@@ -153,44 +154,80 @@ export default class ExpressionParser extends LValParser {
...
+    return this.disallowInAnd(() =>
+      this.parseMaybeAssign(
+        refExpressionErrors,
+        afterLeftParse,
+        refNeedsArrowPos,
+      ),
+    );
+  }
+
+  // Setting [In] parameter for assignment expression
```
JLHwung, line 211:
```suggestion
  // Set [+In] parameter for assignment expression
```

PR #11931, opened 2020-08-07
fix: ExpressionBody should respect [In] parameter

### LINKED ISSUE #11202: [parser] Illegal arrow expression containing `in` as `for` lhs
## Bug Report

Copied from https://github.com/acornjs/acorn/issues/922

REPL: [Safe link](https://babeljs.io/repl#?browsers=node%206&build=&builtIns=false&spec=false&loose=false&code_lz=GYewTgBAFAHhC8A-CMCWA7CBPA3DglDkA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=script&lineWrap=false&presets=env&prettier=true&targets=Node-13&version=7.8.6&externalPlugins=)

(change `xin` to be `x in` and you will see REPL is stuck 😱)

JLHwung:
| Fixed Issues?            | Fixes #11202

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11931"></a>


packages/babel-parser/src/parser/expression.js
```diff
@@ -2110,7 +2110,8 @@ export default class ExpressionParser extends LValParser {
     this.state.inParameters = false;
 
     if (isExpression) {
-      node.body = this.parseMaybeAssignAllowIn();
+      // https://tc39.es/ecma262/#prod-ExpressionBody
+      node.body = this.parseMaybeAssign();
```
JLHwung, line 2114: After the `[In]` parameter tracking refactor, the fix is straightforward: `parseMaybeAssign` parses `AssignmentExpression[?In]`, while `parseMaybeAssignAllowIn` parses `AssignmentExpression[In]`.

According to the spec we should replace `parseMaybeAssignAllowIn` to `parseMaybeAssign`.

nicolo-ribaudo (approved):

PR #11932, opened 2020-08-08
fix: add ImportExpression visitorKeys

### LINKED ISSUE #10904: Template string failing with Cannot read property 'range' of null
It throws error when dynamic import, template literal inline and rule `template-curly-spacing` as follow:

#### Tested on:
`babel-eslint@10.0.3` (with `@babel/parser@7.7.7`)
`eslint@6.7.2`

#### Sample code:
```javascript
import(`${'package.json'}`)
```

#### .eslintrc.json
```json
{
  "parser": "babel-eslint",
  "rules": {
    "template-curly-spacing": ["error", "never"]
  }
}
```

The problem is coming from `@babel/parser@7.7.7`, which added `ImportExpression`.

_Originally posted by @yitomok in https://github.com/babel/babel-eslint/issues/799#issuecomment-567598343_

JLHwung:
| Fixed Issues?            | Fixes #10904

As a follow-up to #10828, this PR adds `ImportExpression` to `VISITOR_KEYS`.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11932"></a>

### DIFF
--- a/eslint/babel-eslint-parser/test/index.js
+++ b/eslint/babel-eslint-parser/test/index.js
@@ -569,7 +569,7 @@ describe("Babel and Espree", () => {
 
     it("Dynamic Import", () => {
       parseAndAssertSame(`
-        const a = import('a');
+        const a = import(moduleName);
       `);
     });
   });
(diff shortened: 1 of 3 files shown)


eslint/babel-eslint-parser/test/index.js
```diff
@@ -569,7 +569,7 @@ describe("Babel and Espree", () => {
 
     it("Dynamic Import", () => {
       parseAndAssertSame(`
-        const a = import('a');
+        const a = import(moduleName);
```
JLHwung, line 572: The test example is revised so that if we do not update `VISITOR_KEYS`, the identifier will not be traversed and the test will fail due to mismatch AST. Note that when traversing AST we remove `loc.identifierName`.

PR #11938, opened 2020-08-09
ESM Distributions with Babel build

### LINKED ISSUE #11701: Publish as modules (instead of compiling to cjs)
> Progress tracked at https://github.com/babel/babel/projects/16

## Feature Request

**Is your feature request related to a problem?**

Would like to be able to consume babel packages directly as modules (as they are written), instead of just cjs. (In my case, may want to import Babel itself into something like snowpack/vite which tries to only support esm). This would be for anything trying to use babel for tooling like the repl/etc so not a common usecase.

Also good that we dogfood what other packages/libraries are doing, even though we aren't really a tool for the browser it would be good to attempt to understand the issues with doing all this anyway.

> :warning: **EDIT**: See https://github.com/babel/babel/issues/11701#issuecomment-777118858

Use [package exports](https://nodejs.org/api/esm.html#esm_package_entry_points). When building we change our config to do `modules: false` in preset-env. Currently we do `"main": "lib/index.js",` for most packages.

Could output a whole folder like `dist` or `module` instead of `lib` or output `.mjs` in the same `lib` folder.

```json
{
  "main": "./lib/main.js",
  "module": "./lib/module.mjs",
  "exports": {
    ".": {
      "import": "./lib/module.mjs",
      "default": "./lib/main.js"
    },
    "./other": {
      "import": "./lib/other.mjs",
      "default": "./lib/other.js"
    }
  }
}
```

**Describe alternatives you've considered**.

Don't think there's an alternative other than consuming the source via github which is difficult because this is a monorepo. We only publish cjs.

yukukotani:
| Fixed Issues?            | Related to #11701, but not resolved all yet (Rollup is remaining)
| Major: Breaking Change?  | No (Maybe?)
| Documentation PR Link    | 

This PR adds configuration to build and provide ESM distributions of packages that is currently built with Babel (not Rollup).

All of packages except below will support ESM.
- babel-cli
- babel-standalone
- babel-runtime
- babel-runtime-corejs2
- babel-runtime-corejs3

Using `.module.js` instead of `.mjs` because webpack 4 doesn't resolve `.mjs` same way to `.js`.
Ref: https://github.com/preactjs/preact/pull/1425 https://github.com/formatjs/formatjs/issues/1395

I don't 100% understand Babel's build process yet. Please let me know if there is any misunderstanding!

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11938"></a>

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -28,6 +28,14 @@ function swapSrcWithLib(srcPath) {
   return parts.join(path.sep);
 }
 
+function swapJsWithMjs(srcPath) {
+  if (srcPath.slice(-3) === ".js") {
+    return srcPath.slice(0, -3) + ".mjs";
+  } else {
+    return srcPath;
+  }
+}
+
 function getIndexFromPackage(name) {
   return `${name}/src/index.js`;
 }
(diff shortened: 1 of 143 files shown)


Monchi:
Working on fixing e2e

Monchi:
Ready for review!

hzoo:
Wow, thanks for this work and the other PRs (excited about Skypack/snowpack too)!

Monchi:
@JLHwung Thanks for pointing out!

> Publishing a new esm build could cause two copies of @babel/core running in the tools:

It was a blind spot. So why do you in favor of first approach? My understanding is that Babel is stateless and getting no pain by dual package.

nicolo-ribaudo:
It ended up being more complex that expected, but Babel 8 is now being published as ESM.

Thank you anyway for trying to help with this!

PR #11941, opened 2020-08-10
fix: push new token context when braceHashL is seen

### LINKED ISSUE #11926: Records & Tuples: records as React props
Hey, reposting this records & tuples bug, as @nicolo-ribaudo asked me to do here https://github.com/tc39/proposal-record-tuple/issues/195

---

This does not work:

https://rickbutton.github.io/record-tuple-playground/#eyJjb250ZW50IjoiaW1wb3J0IFJlYWN0IGZyb20gXCJodHRwczovL2Nkbi5za3lwYWNrLmRldi9yZWFjdFwiO1xuaW1wb3J0IFJlYWN0RE9NIGZyb20gXCJodHRwczovL2Nkbi5za3lwYWNrLmRldi9yZWFjdC1kb21cIjtcblxuXG5jb25zdCBIZWxsbyA9ICh7dXNlcn0pID0+IHtcbiAgcmV0dXJuIDxwPkhlbGxvIHt1c2VyLm5hbWV9PC9wPlxufTtcblxuUmVhY3RET00ucmVuZGVyKFxuICAgIDxIZWxsbyB1c2VyPXsje25hbWU6IFwiU2ViYXN0aWVuXCJ9fS8+LFxuICAgIGRvY3VtZW50LmJvZHksXG4pO1xuXG4iLCJzeW50YXgiOiJoYXNoIiwiZG9tTW9kZSI6dHJ1ZX0=

```jsx
const Hello = ({user}) => {
  return <p>Hello {user.name}</p>
};

ReactDOM.render(
    <Hello user={#{name: "Sebastien"}}/>,
    document.body,
);
```

image

---

But this does:

https://rickbutton.github.io/record-tuple-playground/#eyJjb250ZW50IjoiaW1wb3J0IFJlYWN0IGZyb20gXCJodHRwczovL2Nkbi5za3lwYWNrLmRldi9yZWFjdFwiO1xuaW1wb3J0IFJlYWN0RE9NIGZyb20gXCJodHRwczovL2Nkbi5za3lwYWNrLmRldi9yZWFjdC1kb21cIjtcblxuXG5jb25zdCBIZWxsbyA9ICh7dXNlcn0pID0+IHtcbiAgcmV0dXJuIDxwPkhlbGxvIHt1c2VyLm5hbWV9PC9wPlxufTtcblxuY29uc3QgdXNlciA9ICN7bmFtZTogXCJTZWJhc3RpZW5cIn1cblxuUmVhY3RET00ucmVuZGVyKFxuICAgIDxIZWxsbyB1c2VyPXt1c2VyfS8+LFxuICAgIGRvY3VtZW50LmJvZHksXG4pO1xuXG4iLCJzeW50YXgiOiJoYXNoIiwiZG9tTW9kZSI6dHJ1ZX0=

```jsx
const Hello = ({user}) => {
  return <p>Hello {user.name}</p>
};

const user = #{name: "Sebastien"}

ReactDOM.render(
    <Hello user={user}/>,
    document.body,
);
```

JLHwung:
| Fixed Issues?            | Fixes #11926 

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11941"></a>


packages/babel-parser/test/fixtures/experimental/record-and-tuple/disambiguate-regex-hash/input.js
```diff
@@ -0,0 +1 @@
+(#{})/foo
```
JLHwung, line 1: This case fails on current `main`, too. Because a token context is not pushed and `tt.braceR.updateContext` incorrectly set `this.state.exprAllowed = true` (when `context.length` is 1) so `/foo` is parsed as unterminated regex.

packages/babel-parser/src/tokenizer/context.js
```diff
@@ -140,3 +141,8 @@ tt.backQuote.updateContext = function () {
 tt.star.updateContext = function () {
   this.state.exprAllowed = false;
 };
+
+tt.braceHashL.updateContext = function () {
```
nicolo-ribaudo, line 145: Nit: can you add a comment that we don't do this for `braceBarL` because we only pop the context when we see `}` and not `|}`?

JLHwung, line 145:
```suggestion
// we don't need to update context for tt.braceBarL because we do not pop context for tt.braceBarR
tt.braceHashL.updateContext = function () {
```

PR #11944, opened 2020-08-10
Simplify tokenizer update context

JLHwung:

A side-output when working on #11941.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11944"></a>


packages/babel-parser/src/tokenizer/types.js
```diff
@@ -155,7 +155,8 @@ export const types: { [name: string]: TokenType } = {
   plusMin: new TokenType("+/-", { beforeExpr, binop: 9, prefix, startsExpr }),
   // startsExpr: required by v8intrinsic plugin
   modulo: new TokenType("%", { beforeExpr, binop: 10, startsExpr }),
-  star: createBinop("*", 10),
+  // unset `beforeExpr` as it can be `function *`
+  star: new TokenType("*", { binop: 10 }),
```
JLHwung, line 159: So we can remove
```js
tt.star.updateContext = function () {
  this.state.exprAllowed = false;
};
```
because `exprAllowed` is always implied from `beforeExpr` is there isn't a customized `updateContext` handler.

PR #11955, opened 2020-08-12
Fix parsing type casted generic flow arrow exprs

### LINKED ISSUE #11716: Flow arrow expression with type params inside type cast reported as parse error
## Bug Report

When parsing Flow, arrow expressions with type parameters inside type casts are reported as a parse error. For example:

### Input
```
// @flow
(<T>() => {}: any)
```

### Actual Output
```
(<T>() => {}: any)
            ^
Unterminated JSX contents (2:12)
```

### Expected Output
Expected no errors, as Flow parses this successfully: https://astexplorer.net/#/gist/7c46443529fac6438b16343f614cdaf2/3525323e76f13496801ac9894a69d2837d2b230a

From a quick initial investigation this appears to be because a `TypeCastExpression` is returned instead of an `ArrowFunctionExpression` in `parseMaybeAssign`, causing this check to fail: https://github.com/babel/babel/blob/master/packages/babel-parser/src/plugins/flow.js#L2693.

existentialism:
| Fixed Issues?            | Fixes #11716

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11955"></a>


packages/babel-parser/src/plugins/flow.js
```diff
@@ -2679,23 +2679,44 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+              this.resetStartLocationFromNode(result, typeParameters);
+
+              return result;
+            },
           );
-          arrowExpression.typeParameters = typeParameters;
-          this.resetStartLocationFromNode(arrowExpression, typeParameters);
+
+          // The above can return a TypeCastExpression when the arrow
+          // expression is not wrapped in parens.
```
JLHwung, line 2696:
```suggestion
          // expression is not wrapped in parens. See also `this.parseParenItem`.
```

packages/babel-parser/src/plugins/flow.js
```diff
@@ -2679,23 +2679,44 @@ export default (superClass: Class<Parser>): Class<Parser> =>
...
+          const expr = this.maybeUnwrapTypeCastExpression(arrowExpression);
```
JLHwung, line 2697: Can you add a new test case?
```js
<T>(() => {}: any);
```
should throw.

existentialism: Nice catch!

I'll take a look at the two other issues you found tomorrow.

JLHwung: Yeah. Those two are not blockers, we can address those two in a separate PR.

nicolo-ribaudo (commented):
Can you add the same tests but with `enableParenthesizedExpressions`?

nicolo-ribaudo (approved):

PR #11214, opened 2020-03-05
Ghost copy of #11167

arcanis:
This PR isn't for merge - I'd just like to run the repo tests on #11167 🙂

PR #11962, opened 2020-08-14
Use Yarn 2

JLHwung:

This PR backports yarn 2 related PRs (#11167, #11779, #11782 and #11822) to `main`.
- Added `yarn-plugin-deduplicate` (`yarn deduplicate`)
- Moved some devDependencies to leaf packages.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11962"></a>


package.json
```diff
@@ -13,10 +13,10 @@
   "devDependencies": {
     "@babel/cli": "^7.10.4",
     "@babel/core": "^7.11.0",
-    "@babel/eslint-config-internal": "link:./eslint/babel-eslint-config-internal",
-    "@babel/eslint-parser": "^7.11.0",
-    "@babel/eslint-plugin-development": "^7.11.0",
-    "@babel/eslint-plugin-development-internal": "link:./eslint/babel-eslint-plugin-development-internal",
+    "@babel/eslint-config-internal": "portal:./eslint/babel-eslint-config-internal",
+    "@babel/eslint-parser": "portal:./eslint/babel-eslint-parser",
```
JLHwung, line 17: Thanks to the `portal:` protocol available only in Yarn 2, Babel can be linted by latest `@babel/eslint-parser`. cc @kaicataldo

scripts/integration-tests/e2e-babel-old-version.sh
```diff
@@ -43,10 +43,7 @@ node "$PWD"/scripts/integration-tests/utils/bump-babel-dependencies.js
   fs.writeFileSync('./package.json', JSON.stringify(pkg, null, 2));
 "
 
-rm yarn.lock
```
JLHwung, line 42: The `e2e-babel-old-version` test fails because the `browserslist` lock is also removed: https://app.circleci.com/pipelines/github/babel/babel/3678/workflows/a3f16b77-88e7-4beb-904c-a7bad43e9591/jobs/27418

package.json
merceyz: You could use the workspace protocol instead, which does a better job here

JLHwung: This is the root `package.json`, which is not part of any workspace. Use `workspace` protocol here will throw
```
Workspace not found
```

scripts/integration-tests/e2e-babel-old-version.sh
nicolo-ribaudo: Can we instead modify package.json pinning only browserslist?

package.json
merceyz: The root is technically a workspace, but anyways, it should work as the eslint folders are declared as a workspaces. Jest does this as well https://github.com/facebook/jest/blob/c9c8dba4dd8de34269bdb971173659399bcbfd55/package.json#L56, If you tried before https://github.com/babel/babel/pull/11962/commits/4e94e2a56eb96bf5566eb829a2bf23a02cd2df0a it makes sense that it failed

.yarnrc.yml
```diff
@@ -0,0 +1,12 @@
+enableTransparentWorkspaces: false
+
+nodeLinker: node-modules
```
merceyz, line 3:
```suggestion
nodeLinker: node-modules
enableGlobalCache: true
```
Since you're using the `node-modules` linker you can enable the global cache to speed up installs
https://yarnpkg.com/configuration/yarnrc#enableGlobalCache

package.json
JLHwung: I tried after 3d1e4c3 and it failed. I guess it is related to `enableTransparentWorkspaces: false` in `yarnrc`.

.yarnrc.yml
JLHwung: Can I assume the global cache folder is `$HOME/.yarn/cache`? I have checked on https://yarnpkg.com/configuration/yarnrc but didn't find out. It would be great if the global cache folder can be obtained programmatically so we don't have to hardcode.

scripts/integration-tests/e2e-babel-old-version.sh
JLHwung: We didn't upgrade `browserslist` in this PR so it is implicitly pinned in the lock file. It was pinned as `4.12.0`, same as the version declared in the deleted `packages/babel-compat-data/yarn.lock`. I don't think we need to pin it on package.json except we meant to avoid a security flaw or certain version does not keep backward compatibility.

package.json
merceyz: Hm, just tested with master and got the same result, however changing it to this works:
```diff
diff --git a/package.json b/package.json
index be3015833..11d9a134b 100644
--- a/package.json
+++ b/package.json
@@ -13,10 +13,10 @@
   "devDependencies": {
     "@babel/cli": "^7.10.4",
     "@babel/core": "^7.11.0",
-    "@babel/eslint-config-internal": "portal:./eslint/babel-eslint-config-internal",
-    "@babel/eslint-parser": "portal:./eslint/babel-eslint-parser",
-    "@babel/eslint-plugin-development": "portal:./eslint/babel-eslint-plugin-development",
-    "@babel/eslint-plugin-development-internal": "portal:./eslint/babel-eslint-plugin-development-internal",
+    "@babel/eslint-config-internal": "workspace:*",
+    "@babel/eslint-parser": "workspace:*",
+    "@babel/eslint-plugin-development": "workspace:*",
+    "@babel/eslint-plugin-development-internal": "workspace:*",
     "@babel/plugin-proposal-dynamic-import": "^7.10.4",
     "@babel/plugin-proposal-object-rest-spread": "^7.11.0",
     "@babel/plugin-transform-flow-strip-types": "^7.10.4",
```

Or if you want to be explicit:
```diff
diff --git a/package.json b/package.json
index be3015833..b31abfce9 100644
--- a/package.json
+++ b/package.json
@@ -13,10 +13,10 @@
   "devDependencies": {
     "@babel/cli": "^7.10.4",
     "@babel/core": "^7.11.0",
-    "@babel/eslint-config-internal": "portal:./eslint/babel-eslint-config-internal",
-    "@babel/eslint-parser": "portal:./eslint/babel-eslint-parser",
-    "@babel/eslint-plugin-development": "portal:./eslint/babel-eslint-plugin-development",
-    "@babel/eslint-plugin-development-internal": "portal:./eslint/babel-eslint-plugin-development-internal",
+    "@babel/eslint-config-internal": "workspace:eslint/babel-eslint-config-internal",
+    "@babel/eslint-parser": "workspace:eslint/babel-eslint-parser",
+    "@babel/eslint-plugin-development": "workspace:eslint/babel-eslint-plugin-development",
+    "@babel/eslint-plugin-development-internal": "workspace:eslint/babel-eslint-plugin-development-internal",
     "@babel/plugin-proposal-dynamic-import": "^7.10.4",
     "@babel/plugin-proposal-object-rest-spread": "^7.11.0",
     "@babel/plugin-transform-flow-strip-types": "^7.10.4",

```

.yarnrc.yml
merceyz: > Can I assume the global cache folder is $HOME/.yarn/cache?

Almost, it's located in `~/.yarn/berry/cache` on unix and `~/AppData/Local/Yarn/Berry/cache/` on Windows

> It would be great if the global cache folder can be obtained programmatically so we don't have to hardcode.

If you have the `enableGlobalCache: true` in your config then `yarn config get cacheFolder` points to it

scripts/integration-tests/e2e-babel-old-version.sh
nicolo-ribaudo: I meant to pin it in only in this test, like we do for `@babel/core`

package.json
JLHwung: @merceyz Thanks! I go with `workspace:*` for simplicity. It seems like a bug to me if `"workspace:eslint/babel-eslint-parser"` is working while `"workspace:./eslint/babel-eslint-parser"` is bailing. Should I file an issue for that?

merceyz: Yes please :+1:

scripts/integration-tests/e2e-babel-old-version.sh
JLHwung: Ah I see, the `e2e-test` here is meant to update all deps. I went ahead and pinned it globally.

.circleci/config.yml
```diff
@@ -1,25 +1,13 @@
...
-    paths:
-      - node_modules
-    key: v1-yarn-deps-{{ checksum "yarn.lock" }}
+    key: v2-yarn-cache-{{ checksum "yarn.lock" }}
 
   - &save-yarn-cache
     paths:
-      - ~/.cache/yarn
-    key: v1-1-yarn-cache-{{ checksum "yarn.lock" }}
+      - .yarn/cache
```
merceyz, line 9: With the global cache enabled this is incorrect

.travis.yml
```diff
@@ -1,14 +1,16 @@
 git:
   depth: 5
 language: node_js
+
 cache:
-  yarn: true
+  npm: false
+  directories:
+    - .yarn/cache
```
merceyz, line 8: Same here

.circleci/config.yml
merceyz, line 9: With the global cache enabled this is incorrect

.travis.yml
```diff
@@ -37,13 +39,13 @@ matrix:
         # https://travis-ci.community/t/build-doesnt-finish-after-completing-tests/288/9
         - YARN_GPG=no
       cache:
-        yarn: true
+        npm: false
         directories:
           - $HOME/AppData/Local/Temp/chocolatey
+          - .yarn/cache
```
merceyz, line 45: Same here

.travis.yml
JLHwung: Nice catch!

merceyz:
The latest Yarn from sources has some nice performance improvements compared to 2.1.1, might be worth updating to that https://github.com/yarnpkg/berry/compare/2020-07-09a...master
```
yarn set version from sources
```

JLHwung:
@merceyz Thanks to let us know! I prefer the tagged version as current performance is already greatly improved over `lerna bootstrap`. We will update when 2.2.0 is published.

package.json
JLHwung: Filed https://github.com/yarnpkg/berry/issues/1712.

.yarn/plugins/@yarnpkg/plugin-deduplicate.cjs
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable */
+module.exports = {
+name: "@yarnpkg/plugin-deduplicate",
```
JLHwung, line 3: @eps1lon I just realized that `yarn-plugin-deduplicate` is not licensed under any known licenses. Can you grant permission of usage for the Babel project?

eps1lon: It's now licensed under MIT. Thanks for the notice :+1:

package.json
```diff
@@ -30,11 +30,8 @@
     "@rollup/plugin-json": "4.0.1",
     "babel-jest": "^24.9.0",
     "babel-plugin-transform-charcodes": "^0.2.0",
-    "browserify": "^16.2.3",
```
JLHwung, line 33: They are moved to `@babel/polyfill` because it is only used when building `dist` of `@babel/polyfill`.

packages/babel-polyfill/scripts/build-dist.sh
```diff
@@ -1,17 +1,14 @@
 #!/bin/sh
 set -ex
 
-BROWSERIFY_CMD="../../node_modules/browserify/bin/cmd.js"
```
JLHwung, line 4: Do not have any assumption on node_modules structures and invoke via `yarn` instead.

packages/babel-polyfill/package.json
```diff
@@ -21,5 +21,11 @@
   "dependencies": {
     "core-js": "^2.6.5",
     "regenerator-runtime": "^0.13.4"
+  },
+  "devDependencies": {
```
JLHwung, line 25: They are pinned as the version that previously locked in `yarn.lock`. We can update them in subsequent PRs but given that `@babel/polyfill` is deprecated, this is of very low priority.

merceyz (changes requested):
The version of `@yarnpkg/plugin-constraints` isn't the latest tagged release (2.1.0), unsure which version it is but it should be updated

arcanis:
For context, it used to be used here: https://github.com/babel/babel/pull/11214/files#diff-8ca351b0cde465d2c1b11fe6dd12b1a5R25-R35

JLHwung:
> For context, it used to be used here: https://github.com/babel/babel/pull/11214/files#diff-8ca351b0cde465d2c1b11fe6dd12b1a5R25-R35

Thanks for the comment. Yes it was used to bump the babel dependencies of a given repository. Later we come up with a [custom script](https://github.com/babel/babel/blob/180e9c00e34c08e430305342ef16b67078374bcf/scripts/integration-tests/utils/bump-babel-dependencies.js) for this purpose. So currently it is not used.

packages/babel-register/test/fixtures/browserify/register.js
```diff
@@ -1,4 +1,4 @@
 process.env.BABEL_DISABLE_CACHE = "true";
-require("@babel/register").default({
```
JLHwung, line 2: This test did _not_ test `packages/babel-register` but instead it is tested against some random `@babel/register` in `node_modules`.

nicolo-ribaudo (approved):
Approving again since there have been many changes

packages/babel-polyfill/package.json
JLHwung: `browserify` has been updated to `16.5.2` which bumps `browser-resolve` to 2.0.0 that supports Yarn PnP: https://github.com/browserify/browser-resolve/blob/master/CHANGELOG.md#browser-resolve-change-log

nicolo-ribaudo:
To make it easier to review this PR, I suggest using this link: https://github.com/babel/babel/pull/11962/files/dc3b917a52614e5107c1b7a8ca16b23b3eaa8fc9..ac1366665d34621bcf669be29c789647149cf265?file-filters%5B%5D=.cjs&file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.sh&file-filters%5B%5D=.yml&hide-deleted-files=true

It hides the commits that have already been reviewed in the `next-8-dev` branch. Also, you can skip reading the _contents_ of the file in `.yarn/plugins` since those are just dependencies (but yarn plugins need to be committed because you might need to run them _before_ `yarn install`).

.yarnrc.yml
```diff
@@ -1,10 +1,12 @@
...
+
 enableTransparentWorkspaces: false
 
 nodeLinker: node-modules
 
 plugins:
-  - path: .yarn/plugins/@yarnpkg/plugin-constraints.js
-    spec: "@yarnpkg/plugin-constraints"
+  - path: .yarn/plugins/@yarnpkg/plugin-deduplicate.cjs
+    spec: "https://raw.githubusercontent.com/eps1lon/yarn-plugin-deduplicate/latest/bin/%40yarnpkg/plugin-deduplicate.js"
```
JLHwung, line 9: In the future we will remove this plugin and use built-in `yarn dedupe` ([PR](https://github.com/yarnpkg/berry/pull/1558)) which is expected to ship in Yarn 2.2.0.

eslint/babel-eslint-config-internal/package.json
```diff
@@ -19,6 +19,6 @@
   },
   "peerDependencies": {
     "@babel/eslint-parser": "^7.10.4",
-    "eslint-plugin-flowtype": "^3.0.0"
+    "eslint-plugin-flowtype": "^4.6.0"
```
JLHwung, line 22: `eslint-plugin-flowtypw@4.6.0` is installed in root package.json so we provided `4.6.0` for `^3.0.0`, since `@babel/eslint-config-internal` is private, we bump the peerDependencies to what we used in root package.json.

JLHwung:
Rocket

JLHwung:
:shipit:

PR #11970, opened 2020-08-17
fix: ensure defaults are set in @babel/eslint-parser

### LINKED ISSUE #11902: @babel/eslint-parser does not have default parserOptions (but babel-eslint did)
## Bug Report

I just tried out replacing `babel-eslint` with the `@babel/eslint-parser`, and the first thing I encountered were a few instances of the error:

```
Parsing error: ImportDeclaration should appear when the mode is ES6 and in the module context
```

After doing a little digging, I realized that I needed to set the parserOptions in my eslint config like this:

```
parserOptions: {
  "ecmaVersion": 6,
  "sourceType": "module",
  "ecmaFeatures": {
    "modules": true
  }
}
```

I'm filing this as a bug report because `babel-eslint` provides default parserOptions [here](https://github.com/babel/babel-eslint/blob/6398dac79448a9ac2c8bd3109795433f86799529/lib/index.js#L27) but the new `@babel/eslint-parser` package does not.

I'd vote that either these defaults should be used for `@babel/eslint-parser` as well, or else there should be a very prominent message in the package's README explaining that defaults are no longer provided (I bet that'd save more issues like this one being opened in the future :-) )

kaicataldo:
| Fixed Issues?            | Fixes #11902
| Major: Breaking Change?  |
| Minor: New Feature?      |
| Documentation PR Link    | 
| Any Dependency Changes?  |

ESLint sets `ecmaVersion: undefined` in the config options passed to custom parsers when the `ecmaVersion` isn't defined in the user's config. Since the key exists, it ends up overriding the default value :(

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11970"></a>

### DIFF
--- a/eslint/babel-eslint-parser/src/configuration.js
+++ b/eslint/babel-eslint-parser/src/configuration.js
@@ -9,7 +9,19 @@ export function normalizeESLintConfig(options) {
     requireConfigFile: true,
   };
 
-  return Object.assign(defaultOptions, options);
+  // ESLint sets ecmaVersion: undefined when ecmaVersion is not set in the config.
+  // Prune to ensure that defaults are respected.
+  const prunedOptions = Object.entries(options).reduce(
+    (options, [key, value]) => {
+      if (value !== undefined) {
+        options[key] = value;
+      }
+      return options;
+    },
+    {},
+  );
+
+  return Object.assign(defaultOptions, prunedOptions);
 }
 
 export function normalizeBabelParseConfig(options) {
--- a/eslint/babel-eslint-parser/src/configuration.js
+++ b/eslint/babel-eslint-parser/src/configuration.js
@@ -1,27 +1,24 @@
 import { loadPartialConfig } from "@babel/core";
 
 export function normalizeESLintConfig(options) {
-  const defaultOptions = {
-    babelOptions: {},
-    ecmaVersion: 2020,
-    sourceType: "module",
-    allowImportExportEverywhere: false,
-    requireConfigFile: true,
+  const {
+    babelOptions = {},
+    // ESLint sets ecmaVersion: undefined when ecmaVersion is not set in the config.
+    ecmaVersion = 2020,
+    sourceType = "module",
+    allowImportExportEverywhere = false,
+    requireConfigFile = true,
+    ...otherOptions
+  } = options;
+
+  return {
+    babelOptions,
+    ecmaVersion,
+    sourceType,
+    allowImportExportEverywhere,
+    requireConfigFile,
+    ...otherOptions,
   };
-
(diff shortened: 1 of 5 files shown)


eslint/babel-eslint-tests/test/integration/eslint/eslint.js
```diff
@@ -1,3 +1,5 @@
+import eslint from "eslint";
+import * as parser from "../../../../babel-eslint-parser";
```
JLHwung, line 2: Can we add `@babel/eslint-parser` to the `dependencies` of `@babel/eslint-tests`?

eslint/babel-eslint-tests/test/integration/eslint/eslint.js
```diff
@@ -1879,3 +1881,23 @@ describe("verify", () => {
     );
   });
 });
+
+describe("ESLint config", () => {
```
JLHwung, line 1885: I slightly prefer to place them in a new test file `eslint/config.js` as the other tests in `eslint.js` are about verifying lint results.

eslint/babel-eslint-parser/src/configuration.js
```diff
@@ -9,7 +9,19 @@ export function normalizeESLintConfig(options) {
...
+    (options, [key, value]) => {
+      if (value !== undefined) {
+        options[key] = value;
+      }
+      return options;
+    },
+    {},
+  );
+
+  return Object.assign(defaultOptions, prunedOptions);
```
JLHwung, line 24: What do you think avoiding `prunedOptions` and `Object.assign`?

```js
  const {
    babelOptions = {},
    ecmaVersion = 2020,
    sourceType = "module",
    allowImportExportEverywhere = false,
    requireConfigFile = true,
    ...otherOptions
  } = options;

  return {
    babelOptions,
    ecmaVersion,
    sourceType,
    allowImportExportEverywhere,
    requireConfigFile,
    ...otherOptions
  };
```

kaicataldo: Sure!

eslint/babel-eslint-tests/test/integration/eslint/eslint.js
kaicataldo: I don't have a strong opinion either way! Happy to move into a separate file.

eslint/babel-eslint-tests/test/integration/eslint/eslint.js
kaicataldo: Ah, yeah - I thought it wouldn't update when changes were made to the parser, but I just confirmed that's not the case.

PR #11978, opened 2020-08-19
fix: do not mutate newTypes

JLHwung:

#11973 introduced a bug that `newTypes` may be mutated in default exports. This is unwanted.

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11978"></a>


JLHwung:
I am gonna merge it since it is a small fix.

PR #11982, opened 2020-08-19
Parse string input to @babel/code-frame

### LINKED ISSUE #11981: @babel/code-frame assumes location input will contain numbers
## Bug Report

`@babel/code-frame` assumes location input contains integers. String input can prevent `linesBefore` from truncating output due to unintentional string concatenation on [this line](https://github.com/babel/babel/blob/main/packages/babel-code-frame/src/index.js#L57).

```js
import {codeFrameColumns} from '@babel/code-frame';

const result = codeFrameColumns(
  fs.readFile(filename, 'utf-8'), // file with < 322 lines
  { start: { line: '32' },
  { linesBelow: 2 },
);
```

This is obviously user-error (and would be caught if using static types), but results in confusion rather than an error. It's simple enough to protect against within `@babel/code-frame`.

**Expected behavior**
`@babel/code-frame` should truncate input according to `linesAbove`/`linesBelow`.

**Possible Solution**

Use `parseInt` on location input before using in arithmetic expressions

micburks:
| Fixed Issues?            | Fixes #11981

Fixes #11981 by using `parseInt` on location before using in arithmetic expressions

<a href="https://gitpod.io/#https://github.com/babel/babel/pull/11982"></a>

### DIFF
--- a/packages/babel-code-frame/src/index.js
+++ b/packages/babel-code-frame/src/index.js
@@ -48,10 +48,10 @@ function getMarkerLines(
     ...loc.end,
   };
   const { linesAbove = 2, linesBelow = 3 } = opts || {};
-  const startLine = startLoc.line;
-  const startColumn = startLoc.column;
-  const endLine = endLoc.line;
-  const endColumn = endLoc.column;
+  const startLine = parseInt(startLoc.line);
+  const startColumn = parseInt(startLoc.column);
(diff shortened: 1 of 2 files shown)


JLHwung:
I am conservative on this PR, see https://github.com/babel/babel/issues/11981#issuecomment-677880886.

PR #12007, opened 2020-08-25
Fixed refs to decorated class from itself methods

### LINKED ISSUE #11131: [@babel/plugin-proposal-decorators] legacy: true, non-transformed class not picked up inside methods
## Bug Report

**Current Behavior**

When using legacy: true (didn't check without it) and using a class decorator that transforms the class into another one ("wraps it"), the original class will be used instead of the original one.

**Input Code**

https://codesandbox.io/s/boring-morning-yeel2

```js
function myClassDecorator(clazz) {
  return class extends clazz {};
}

function getA() {
  // here it returns the wrapped class correctly
  return A;
}

@myClassDecorator
class A {
  getA1() {
    // apparently in this case it will pick up the original class rather than the transformed one
    return A;
  }

  getA2() {
    return getA(); // should be no different than `getA1`, but it is :-/
    // in this case it correctly returns the wrapped class
  }

  getA3 = () => {
    // apparently in this case it will pick up the original class rather than the transformed one
    return A;
  };
}

const a1 = new A().getA1();
if (a1 !== A) {
  console.error("a1: different class"); // triggers
}

const a2 = new A().getA2();
if (a2 !== A) {
  console.error("a2: different class"); // does not trigger
}

const a3 = new A().getA3();
if (a3 !== A) {
  console.error("a3: different class"); // triggers
}
```

**Expected behavior/code**
I expect it to return / use the wrapped class rather than the original one in all instances (like TS does)

babel-preset-react-app package when running in TS mode
(basically @babel/plugin-proposal-decorators with legacy set to true)
```

**Environment**
<!--- Tip: Instead of filling out the questions below, you can run `npx envinfo --preset babel` and paste the result below ``` -->
```

```
Babel version(s): core 7.8.4, babel/plugin-proposal-decorators 7.8.3

System:
    OS: Windows 10 10.0.18363
  Binaries:
    Node: 13.6.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD

**Possible Solution**

The generated code:

```js
var _class, _temp;

function myClassDecorator(clazz) {
  return class extends clazz {};
}

function getA() {
  return A;
}

let A = myClassDecorator(_class = (_temp = class A {
  constructor() {
    this.getA3 = () => {
      // apparently in this case it will pick up the original class rather than the transformed one
      return A;
    };
  }

  getA1() {
    // apparently in this case it will pick up the original class rather than the transformed one
    return A;
  }

  getA2() {
    return getA(); // should be no different than `getA1`, but it is :-/
  }

}, _temp)) || _class;

const a1 = new A().getA1();

if (a1 !== A) {
  console.error("a1: different class"); // triggers
}

const a2 = new A().getA2();

if (a2 !== A) {
  console.error("a2: different class"); // does not trigger
}

const a3 = new A().getA3();

if (a3 !== A) {
  console.error("a3: different class"); // triggers
}
```

this line:
```let A = myClassDecorator(_class = (_temp = class A {```

makes me think the method is picking up the `class A` rather than the `let A`, so maybe the solution could be to do what TS does:

http://www.typescriptlang.org/play/#code/GYVwdgxgLglg9mABAWwJ4GEA2BDAzrgEQFMI4AnbKcgCghwC96AuRbMVAShbdUQG8AUIkRkiUEGSR08uREQAeUImAAms6Y34BfANwCtAgaEiwEiAOZiAgtQ78hIsRKRW9BgQAE0WGcVIUqMgFpfEQre2FLKCsARlsI4UQAeiTWAAc07FEwKExeGCQoAAsYdTwiRBgoRAB3GExMRDSYCABrRBA0xGKK8hhzAuxGkNkAoqIybqK2KYqoCjBcYHJkIhVEBCIHYVFxSTC9YXdI6wAmeMFExz2kKJsOHWTU3CK4EEx1gCMKsDhEFRgwGAE2U1WKMwABncYhCADSIT4gapVSqyJgAWiSDmOFmsAGZEABeRDxQkAPgSwhS6Uy2Vy+UKJTKuAqKLqDSaLXanVmGzI-UGwxwoTGEymMx63QWSxWaw2YC2V12zgO2LchlIi2q2BiRMQCpqYVsADpobY9ICSTrEABCQnEqx2S6ITW4OCYIjGiZkGgAIh1LABQJBORdwtwvoeTyl-UsZFw+g1CFw2tOeoNRo4prO5oEluo2DTdodTocrvdnu9fsLgcBwLpYZkkce1JUcCIsl+YP55jjieCye1BOJGfu2eieNz+ewBOLYVLwnLHq9ZB9ZGo-rxteDDZGzej81jEwTBiAA

```js
...
var A_1;
function myClassDecorator(clazz) {
    return class extends clazz {
    };
}
function getA() {
    return A;
}
let A = A_1 = class A {
    constructor() {
        this.getA3 = () => {
            // apparently in this case it will pick up the original class rather than the transformed one
            return A_1;
        };
    }
    getA1() {
        // apparently in this case it will pick up the original class rather than the transformed one
        return A_1;
    }
    getA2() {
        return getA(); // should be no different than `getA1`, but it is :-/
    }
};
A = A_1 = __decorate([
    myClassDecorator
], A);
```

Amareis:
| Fixed Issues?            | Fixes #11131 
| Documentation PR Link    | 

Problem described in #11131
Solved by inserting in all methods (including #private ones and constructor) `let ClassName = _someGeneratedVariable`, where `_someGeneratedVariable` hold "final" class, after all decorators applied.

### DIFF
--- a/packages/babel-plugin-proposal-decorators/test/fixtures/legacy-class-inner-refs/exec.mjs
+++ b/packages/babel-plugin-proposal-decorators/test/fixtures/legacy-class-inner-refs/exec.mjs
@@ -1,0 +1,42 @@
+let Child
+
+function dec(cls) {
+  Child = class extends cls {}
+  return Child
+}
+
+function dec2(cls) {
+  return class extends cls {}
+}
+
+@dec
+@dec2
+class Parent {
+  constructor() {
+    this.prop = Parent
+  }
+
+  static static() {
+    return Parent
+  }
+
+  method() {
+    return Parent
+  }
+
+  #priv() {
+    return Parent
+  }
+
+  public() {
+    return this.#priv()
+  }
+}
+
+expect(Parent.static()).toBe(Child)
+
+const p = new Parent()
+
+expect(p.prop).toBe(Child)
+expect(p.method()).toBe(Child)
+expect(p.public()).toBe(Child)
--- a/packages/babel-plugin-proposal-decorators/src/transformer-legacy.js
+++ b/packages/babel-plugin-proposal-decorators/src/transformer-legacy.js
@@ -66,9 +66,33 @@ function applyClassDecorators(classPath) {
   const decorators = classPath.node.decorators || [];
   classPath.node.decorators = null;
 
+  const className = classPath.node.id;
+  let externalName;
+  if (className) {
+    externalName = classPath.scope.generateDeclaredUidIdentifier(
+      "decorated" + className.name,
+    );
+    const fix = path => {
+      if (path.node.body.body.length) {
+        path.node.body.body.unshift(
+          t.variableDeclaration("let", [
+            t.variableDeclarator(
+              t.cloneNode(className),
+              t.cloneNode(externalName),
+            ),
+          ]),
+        );
+      }
+    };
+    classPath.scope.traverse(classPath.node, {
+      ClassMethod: fix,
+      ClassPrivateMethod: fix,
+    });
+  }
+
   const name = classPath.scope.generateDeclaredUidIdentifier("class");
 
-  return decorators
+  const res = decorators
     .map(dec => dec.expression)
     .reverse()
     .reduce(function (acc, decorator) {
@@ -78,6 +102,8 @@ function applyClassDecorators(classPath) {
         INNER: acc,
       }).expression;
     }, classPath.node);
+  if (!externalName) return res;
+  return t.assignmentExpression("=", externalName, res);
 }
 
 function hasClassDecorators(classNode) {
--- a/packages/babel-plugin-proposal-decorators/test/fixtures/legacy-class-inner-refs/options.json
+++ b/packages/babel-plugin-proposal-decorators/test/fixtures/legacy-class-inner-refs/options.json
@@ -1,0 +1,8 @@
+{
+  "presets": ["env"],
+  "plugins": [
+    ["proposal-decorators", { "legacy": true }],
+    ["proposal-private-methods", { "loose": true }],
+    ["proposal-class-properties", { "loose": true }]
+  ]
+}
--- a/packages/babel-plugin-proposal-decorators/test/fixtures/legacy-decl-to-expression/class-decorators/output.mjs
+++ b/packages/babel-plugin-proposal-decorators/test/fixtures/legacy-decl-to-expression/class-decorators/output.mjs
@@ -1,7 +1,7 @@
-var _class, _class2;
+var _decoratedB, _class, _decoratedA, _class2;
 
-let A = dec(_class2 = class A {}) || _class2;
+let A = _decoratedA = dec(_class2 = class A {}) || _class2;
 
 export { A as default };
 
-let B = dec(_class = class B {}) || _class;
+let B = _decoratedB = dec(_class = class B {}) || _class;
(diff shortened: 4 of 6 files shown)


packages/babel-plugin-proposal-decorators/test/fixtures/legacy-class-inner-refs/exec.mjs
```diff
@@ -0,0 +1,42 @@
...
+  constructor() {
+    this.prop = Parent
+  }
+
+  static static() {
+    return Parent
+  }
+
+  method() {
+    return Parent
```
JLHwung, line 24: Can you add test cases on `property`? i.e.
```js
@dec
@dec2
class Parent {
  property = Parent,
}
```
What will `p.property` be? Same for private properties.

Amareis: >Solved by inserting in all methods (including static, #private ones and constructor) let ClassName = _someGeneratedVariable, where _someGeneratedVariable hold "final" class, after all decorators applied.
>References in class properties still isn't supported.

Property isn't working currently (will hold undecorated class). For full solution we should replace all class references in read positions of class body to outer reference, but I ain't good enough at babel for such tricks.

SuperOleg39: Hi! Catch issue with property yesterday, similar to @JLHwung  case, but with static property)

example source:
```
function dec(cls) {
  return cls
}

@dec
class Decorated {
  static property = Decorated
}
```

and output:
```
"use strict";

var _class, _class2, _temp;

function dec(cls) {
  return cls;
}

let Decorated = dec(_class = (_temp = _class2 = class Decorated {}, _class2.property = Decorated, _temp)) || _class;
```

This code produce runtime error `ReferenceError: Cannot access 'Decorated' before initialization`

SuperOleg39: @Amareis about class properties, fix references require changes in `@babel/plugin-proposal-class-properties` plugin?

Amareis: > @Amareis about class properties, fix references require changes in `@babel/plugin-proposal-class-properties` plugin?

Don't think so. If we get full solution with replacing of original class reference in class body, this will be working without any additional efforts.

xaviergonz:
@nicolo-ribaudo what's needed for this PR to get merged? just curious

zachdaniel:
Would love to see this merged :)

maccman:
Let's get this merged

xaviergonz:
I'd say one good reason could be able to use babel to transpile typescript without a compatibility issue.,since this plugin is often used to do so. If the old behavior wants to be kept (understandable), then at least it could be offered as a "legacy-ts" mode or similar. Using the new stage 3 proposal would not fix the typescript compatibility issue.

nicolo-ribaudo:
TypeScript will support standard decorators in the next version: https://github.com/microsoft/TypeScript/issues/49074

Amareis:
5 years passed...

